PR #1084, opened 2015-07-21
rpi-ft5406: Use firmware API

notro:
I don't have the hardware so this patch has not been functionally tested.
I have only tested that it detects that no hardware is attached.


pelwell:
Looks OK to me.

popcornmix:
Ping @ghollingworth can you test this?

PR #1086, opened 2015-07-24
dwc_otg: Add ARCH_BCM2835 support

notro:
Tested on ARCH_BCM2708 and ARCH_BCM2835.


popcornmix:
Looks okay to me. Ping @pelwell @P33M

P33M:
No objections.

drivers/irqchip/irq-bcm2835.c
```diff
@@ -56,7 +56,7 @@
 #include "irqchip.h"
 
 /* Put the bank and irq (32 bits) into the hwirq */
-#define MAKE_HWIRQ(b, n)	((b << 5) | (n))
+#define MAKE_HWIRQ(b, n)	(((b) << 5) | (n))
 #define HWIRQ_BANK(i)		(i >> 5)
 #define HWIRQ_BIT(i)		BIT(i & 0x1f)
```
pelwell, line 61: i should get the same parenthetical treatment as b, no?

drivers/irqchip/irq-bcm2835.c
```diff
@@ -72,9 +72,13 @@
 					| SHORTCUT1_MASK | SHORTCUT2_MASK)
 
 #define REG_FIQ_CONTROL		0x0c
+#define REG_FIQ_ENABLE		0x80
```
pelwell, line 75: Does this indentation look better in an editor?

notro, line 75: Yes. You can see it if you click on View (as whole file).

pelwell:
I just have the one unanswered point (use of parentheses within macros - the commit fixes one by necessity but there are two on adjacent lines that could also be fixed), but otherwise it looks good.

notro:
I fixed that one because otherwise I would get a wrong result when I use it later:

``` c
#define FIQ_START       (NR_IRQS_BANK0 + MAKE_HWIRQ(NR_BANKS - 1, 0))
```

The patch is accepted in mainline, so it's too late to change it now.

PR #1096, opened 2015-08-04
bcm2835 camera driver planar/packed stride length

floft:
I was having issues in Gstreamer with YUV420 and it appears it is from the stride length or bytes per line reported by the driver being incorrect when it's giving planar data. I found [a discussion](http://comments.gmane.org/gmane.comp.video.gstreamer.devel/54508) about this that led me to look at [mx6-camif.h](https://github.com/veo-labs/linux-veobox/blob/veobox/drivers/staging/media/imx6/capture/mx6-camif.h#L67) and [mx6-camif.c](https://github.com/veo-labs/linux-veobox/blob/veobox/drivers/staging/media/imx6/capture/mx6-camif.c#L159) where they specified the number of bits per Y pixel. Making this change made Gstreamer happy; however, I'm not entirely sure how this should be done. It seems that the YUV420 is planar when I'm using it on the Raspberry Pi with Gstreamer, but are there cases where it actually is packed?

### DIFF
--- a/drivers/media/platform/bcm2835/bcm2835-camera.c
+++ b/drivers/media/platform/bcm2835/bcm2835-camera.c
@@ -88,12 +88,13 @@ static const struct v4l2_fract
 /* video formats */
 static struct mmal_fmt formats[] = {
 	{
-	 .name = "4:2:0, packed YUV",
+	 .name = "4:2:0, planar, YUV",
 	 .fourcc = V4L2_PIX_FMT_YUV420,
 	 .flags = 0,
 	 .mmal = MMAL_ENCODING_I420,
 	 .depth = 12,
 	 .mmal_component = MMAL_COMPONENT_CAMERA,
+	 .ybbp = 1,
 	 },
 	{
 	 .name = "4:2:2, packed, YUYV",
@@ -102,6 +103,7 @@ static struct mmal_fmt formats[] = {
 	 .mmal = MMAL_ENCODING_YUYV,
 	 .depth = 16,
 	 .mmal_component = MMAL_COMPONENT_CAMERA,
+	 .ybbp = 2,
 	 },
 	{
 	 .name = "RGB24 (LE)",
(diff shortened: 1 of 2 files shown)


popcornmix:
Ping @6by9
Note: 3.18 tree is no longer the latest, but we can discuss this here.

6by9:
Thanks for the patch. It looks like it is addressing a valid issue, but if you're fixing it, it also ought to be fixed in vidioc_g_fbuf to return the correct value.

You've not removed the old line (was 897, now 911), so f->fmt.pix.bytesperline is being assigned and then assigned again.

Rereading the spec (http://hverkuil.home.xs4all.nl/spec/media.html#idp52868032) 

> For compressed formats the bytesperline value makes no sense. Applications and drivers must set this to 0 in that case.

So actually ybpp should be 0 for JPEG, MJPEG and H264. 

Otherwise it looks good.

floft:
How's this look? Is that how to fix vidioc_g_fbuf?

6by9:
Yes, looks fine now - thanks. 

As popcornmix says, 3.18 is no longer the latest, though it appears still to be updated. I'll let him specify which branch he wants this to be pushed to first.
I suspect his first request will be that these two commits want to be squashed into a single commit - we don't need the history of how we got to a working solution.

(If I were being really fussy then I'd ask you to check the patch with check_patch.pl (it looks clean though), and also add a "signed off" line to the commit text (git commit -s). However it looks like the admins on the Pi kernel aren't as fussy as upstream.)

6by9:
Looks good to me.
@popcornmix Up to you which branch this gets merged to and by whom. It's valid on all branches.

popcornmix:
@floft (and @6by9) thanks. Bug fixes are always welcome.
I've cherry-picked this to 4.1 (current stable) and 4.2 trees. It will appear in next rpi-update kernel (which is now 4.1 tree).

PR #1097, opened 2015-08-13
BCM270x: Use firmware API

notro:
Boot tested on Pi1 and Pi2.

ARCH_BCM2835 will be in a PR of it's own.


popcornmix:
Looks okay - thanks.

PR #1098, opened 2015-08-13
BCM2708: Use brcm,bcm2835-system-timer

notro:
When booting with Device Tree use drivers/clocksource/bcm2835_timer.c

Tested with and without DT.


notro:
Oh, I forgot about the warning, I'll see if I can fix that.

```
  CC      drivers/clocksource/bcm2835_timer.o
drivers/clocksource/bcm2835_timer.c:39:0: warning: "MAX_TIMER" redefined [enabled by default]
 #define MAX_TIMER 3
 ^
In file included from arch/arm/mach-bcm2708/include/mach/irqs.h:26:0,
                 from ./arch/arm/include/asm/irq.h:7,
                 from ./arch/arm/include/asm/hardirq.h:6,
                 from include/linux/hardirq.h:8,
                 from include/linux/interrupt.h:12,
                 from drivers/clocksource/bcm2835_timer.c:22:
arch/arm/mach-bcm2708/include/mach/platform.h:186:0: note: this is the location of the previous definition
 #define MAX_TIMER                       2
 ^
```

pelwell:
Is that symbol even used?

notro:
No it doesn't look like it is. I'm doing a build to be sure.

notro:
I have removed MAX_TIMER from arch/arm/mach-bcm2708/include/mach/platform.h

pelwell:
That works for me.

PR #1099, opened 2015-08-15
Enable ARCH_BCM2835 for downstream use

notro:
This PR makes ARCH_BCM2835 functionally equivalent to ARCH_BCM2708.

spi-bcm2708 is dropped now that spi-bcm2835 is the default on 4.1.
i2c-bcm2708 is kept, because i2c-bcm2835 has not seen any downstream use yet and it doesn't work with a I2C OLED display I have.

The Device Tree files are copied from their 2708 counterparts.
#### Issues

Onboard audio is disabled by default regardless of whether the module is loaded or not.

apb_pclk and uart0_pclk are defined both in bcm2708_common.dtsi and in clk-bcm2835 resulting in an error in the boot messages from the latter.
I haven't found a way to disable a DT clock, the status property is not checked and changing the compatible string results in no serial output.
This is a purely cosmetic issue.

```
[    0.212758] apb_pclk not registered
[    0.212849] uart0_pclk not registered
[    0.212900] uart0_pclk alias not registered
```
#### Test details

Serial & Revision

```
$ cat /proc/cpuinfo
...
Hardware        : BCM2835
Revision        : 0010
Serial          : 00000000a1725dd1
```

Thermal

```
$ cat /sys/class/thermal/thermal_zone0/temp && vcgencmd measure_temp
39007
temp=39.0'C

```

cpufreq

```
$ head -1 /boot/config.txt
arm_freq=900
$ dmesg | grep cpufreq
[    1.893675] bcm2835-cpufreq: min=700000 max=900000

```

vc_mem

```
$ sudo vcdbg log msg 2>&1 | grep Loading
000915.390: Loading 'kernel.img' from SD card
001159.040: Loading 'bcm2835-rpi-b-plus.dtb' from SD card
```

/dev/vcio

```
$ sudo /opt/vc/src/hello_pi/hello_fft/hello_fft.bin 12
rel_rms_err = 7.8e-07, usecs = 259, k = 0
```

bcm2708_fb Colour palette

```
# Audio is necessary: /boot/config.txt
# dtparam=audio=on

$ ./ch-egg-port-sdl-source_v1.1/ch-egg.elf

```

bcm2708_fb Ioctl: FBIOPAN_DISPLAY and FBIO_WAITFORVSYNC

```
$ ./fbtestXIV

```

bcm2708_fb Blanking

```
$ setterm -blank 1

```

SPI is tested with a PiScreen display.

Uart1 is tested.


notro:
@pelwell The mmc drivers get a slightly different build on ARCH_BCM2835. Is it possible to have the same build on all?

pelwell:
It was never clear to me why the builds would handle SDIO interrupts differently:
- On 270x builds, the non-threaded IRQ handler calls mmc_signal_sdio_irq that wakes up the MMC-framework-provided sdio_irq_thread, that then calls process_sdio_pending_irqs.
- On 2835 builds, the non-threaded IRQ handler defers to the threaded IRQ handler, which calls sdio_run_irqs that in turn calls process_sdio_pending_irqs directly.

I found this patch (https://github.com/raspberrypi/linux/commit/bf3b5ec66bd03d66e9ea729aaca013ea1047a797) that introduced sdio_run_irqs, but I don't know why the use int the mmc driver is conditional on the architecture. The differences in the sdhost driver are academic since we don't (and probably never will) use the SDIO capabilities - that code has never been run. The 2835 configuration seems to be the more up-to-date since it is using the new facility, so perhaps I could try making the change unconditionally. Unfortunately I'm not in a position to test SDIO functionality, which makes me reluctant to commit a change like that.

notro:
There is only one in-kernel user of this new code, and that's the sdhci subsystem: [mmc: sdhci: convert to new SDIO IRQ handling](https://github.com/torvalds/linux/commit/781e989cf593c71d26bdca74f5e77b3651fc060e)
It looks like the code in bcm2835-mmc was copied from that patch.
However this statement is missing in our code, but it seem to only have impact on some resume path: `mmc->caps2 |= MMC_CAP2_SDIO_IRQ_NOTHREAD;`

I added a printk to bcm2835_mmc_thread_irq(), but it didn't fire.

So I suggest we just rip out this conditional code. It's been a year since the patch was added and there has been no other drivers converting to it, so it can't be that important.
Even though Russel had this to say:

```
...
rather than allowing the SDIO core code to buggily spawn its own
thread.

```

I think it's important that the mmc driver behaves in the same way on all architectures if possible.

pelwell:
See https://github.com/raspberrypi/linux/pull/1102

pelwell:
I've succeeded in getting the upstream interrupt controller to work for DT-enabled Pi 1 builds. Without DT it falls back to the old armctrl code. This patch (https://gist.github.com/pelwell/d41ec7fb7db4d1136eb5) should be applied after all of the others in this PR - perhaps you could append it, @notro?

pelwell:
For Pi 2 there will be a race between the upstreaming of the BCM2836 support and our removal of non-DT support. If BCM2836 wins then we can produce a similar patch for Pi 2, otherwise we just remove the old armctrl driver and switch wholesale.

Tested on Pi 1 and Pi 2, with and without Device Tree.

pelwell:
@notro Can you rebase against 4.2-rc7, and consider appending my patch?

notro:
Building fails on ARCH_BCM2835:

```
drivers/usb/host/dwc_otg/dwc_otg_hcd_linux.c: In function ‘hcd_init_fiq’:
drivers/usb/host/dwc_otg/dwc_otg_hcd_linux.c:453:13: error: ‘INTERRUPT_VC_USB’ undeclared (first use in this function)
  enable_fiq(INTERRUPT_VC_USB);
             ^
drivers/usb/host/dwc_otg/dwc_otg_hcd_linux.c:453:13: note: each undeclared identifier is reported only once for each function it appears in
```

Why was this change necessary:

``` diff
-   init_FIQ(FIQ_START);
+   init_FIQ(NUMBER_IRQS);
```

I'm restarting the build now. I'll let you know when I have tried it.

pelwell:
> Why was this change necessary:

Because of lines like this (https://github.com/notro/linux/commit/bab25463bb5842694f73de42bdf7edec600f2b7c#diff-b412d3553bc228bafd652cac486b891fR107):

```
static inline unsigned int hwirq_to_fiq(unsigned long hwirq)
{
    hwirq -= NUMBER_IRQS;
```

and this (https://github.com/notro/linux/commit/bab25463bb5842694f73de42bdf7edec600f2b7c#diff-b412d3553bc228bafd652cac486b891fR206):

```
            irq = irq_create_mapping(intc.domain,
                    MAKE_HWIRQ(b, i) + NUMBER_IRQS);
```

i.e. You have mapped the FIQs starting NUMBER_IRQS after the IRQs, and init_FIQ indicates where the FIQs start. Without this change it can't associate the arriving FIQ interrupts with the correct device.

notro:
The irq-bcm2835.c change doesn't work on ARCH_BCM2835. Lots of:

```
[    7.305673] Transfer to device 4 endpoint 0x1 frame 769 failed - FIQ timed out. Data may have been lost.
```

Reverting the change made it work again.

I remember that getting the fiq patch working was trial and error, so it's hard to come up with the rationale for this now :-)

I guess you needed this change to make it work on ARCH_BCM2708?

pelwell:
Yes, it was necessary. I'll try the 2835 build tomorrow.

notro:
Rebased

pelwell:
I have an updated patch [here](https://gist.github.com/pelwell/8ad7c6710eed7d44ab41) that builds and works on ARCH_BCM2835 as well, and without changing any of the upstream code.

There were two related problems:
1) The FIQ offset (the parameter to init_FIQ()) is in Linux IRQ numbers (not the underlying hardware IRQ values). This should be the number of registered IRQs (see 2), which is why FIQ_START was correct for BCM2835 builds.

2) The IRQ map used by irq-bcm2835 puts the ARMCTRL interrupts (BANK0) at 0, and the other banks at 32 and 64. Since BANK0 only includes 8 interrupt sources, this leaves a hole in the hardware IRQ numbers between BANK0 and BANK1. (The downstream code puts the short bank at the end, so no gap).
CONFIG_SPARSE_IRQ is set on ARCH_BCM2835, which causes 16 legacy IRQs to be allocated and no gaps to appear in the mapped IRQs:

```
IRQ BANK0: 16-23
IRQ BANK1: 24-55
IRQ BANK2: 56-87
FIQ BANK0: 88-95
FIQ BANK1: 96-127
FIQ BANK2: 128-159
```

You will see there is a constant offset of 72 between corresponding FIQs and IRQs, and 72 is FIQ_START.

On MACH_BCM2708, CONFIG_SPARSE_IRQ is not set, so there are no legacy IRQs and it tries to use the identity mapping for IRQ numbers. The allocator avoids 0 as an illegal value, so maps BANK0 to 1-8. But it isn't necessary to skip 32 or 64, so the mapping ends up looking like this:

```
IRQ BANK0: 1-8
IRQ BANK1: 32-63
IRQ BANK2: 64-95
FIQ BANK0: 96-103
FIQ BANK1: 128-159
FIQ BANK2: 160-191
```

Here the FIQ->IRQ offset is 95 for BANK0 and 96 for BANK1 and BANK2. This explains why NUMBER_IRQs (96) worked as the argument to init_FIQ, but it wouldn't have worked for the BANK0 FIQs and breaks ARCH_BCM2835 builds.

The workaround I have found is to provide an implementation of arch_dynirq_lower_bound (a weak function) that pushes the IRQ BANK0 numbers up, pulls the FIQ BANK1+2 numbers down to 104-167, and leaves the other numbers alone:

```
IRQ BANK0: 24-31
IRQ BANK1: 32-63
IRQ BANK2: 64-95
FIQ BANK0: 96-103
FIQ BANK1: 104-135
FIQ BANK2: 136-167
```

This restores the fixed offset, and keeps it at 72, matching ARCH_BCM2835 and removing the need to change irq-bcm2835.

notro:
Thanks for the detailed explanation.
I have added your commit.
Tested on ARCH_BCM2708 and ARCH_BCM2835

pelwell:
I'm happy with this. @popcornmix?

popcornmix:
This PR breaks booting for me with bcm2709_defconfig. NFS boot fails. See:
http://pastebin.com/UgUHpcCL

popcornmix:
@pelwell    looks like "BCM2708: Use upstream interrupt driver on all Pi1's" is the commit that breaks booting on Pi2.

pelwell:
There's a squash patch [here](https://gist.github.com/pelwell/bfbd359c76272224091a) that should theoretically restore normal service.

popcornmix:
squash patch does fix the bcm2709_defconfig build.

notro:
Squashed and tested on ARCH_BCM2708.

PR #1100, opened 2015-08-17
BCM270X_DT: Add pwm and pwm-2chan overlays

pelwell:
From the README entries:
  Legal pin,function combinations for each channel:
    PWM0: 12,4(Alt0) 18,2(Alt5) 40,4(Alt0)            52,5(Alt1)
    PWM1: 13,4(Alt0) 19,2(Alt5) 41,4(Alt0) 45,4(Alt0) 53,5(Alt1)
  N.B.:
    1) Pin 18 is the only one available on all platforms, and
       it is the one used by the I2S audio interface.
       Pins 12 and 13 might be better choices on an A+, B+ or Pi2.
    2) The onboard analogue audio output uses both PWM channels.
    3) So be careful mixing audio and PWM.
    4) Currently the clock must have been enabled and configured
       by other means.


popcornmix:
No objections. Feel free to merge.

PR #1102, opened 2015-08-17
bcm2835-mmc/sdhost: Remove ARCH_BCM2835 differences

pelwell:
The bcm2835-mmc driver (and -sdhost driver that copied from it)
contains code to handle SDIO interrupts in a threaded interrupt
handler rather than waking the MMC framework thread. The change
follows a patch from Russell King that adds the facility as the
preferred way of working.

However, the new code path is only present in ARCH_BCM2835
builds, which I have taken to be a way of testing the waters
rather than making the change across the board; I can't see
any technical reason why it wouldn't be enabled for MACH_BCM270X
builds. So this patch standardises on the ARCH_BCM2835 code,
removing the old code paths.

### DIFF
--- a/drivers/mmc/host/bcm2835-mmc.c
+++ b/drivers/mmc/host/bcm2835-mmc.c
@@ -684,12 +684,10 @@ void bcm2835_mmc_send_command(struct bcm2835_host *host, struct mmc_command *cmd
 	}
 
 	timeout = jiffies;
-#ifdef CONFIG_ARCH_BCM2835
 	if (!cmd->data && cmd->busy_timeout > 9000)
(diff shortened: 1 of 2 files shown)


notro:
Should we add this for completeness? The mmc core might act on it in the future in a way that affects us.

``` c
mmc->caps2 |= MMC_CAP2_SDIO_IRQ_NOTHREAD;
```

This line has received some extra spaces in addition to the tab:

``` diff
+       timeout += 10 * HZ;
```

pelwell:
Agreed.

pelwell:
Our resident SDIO guru has verified that an SDIO device works with the patch applied, so I'll merge it.

PR #1104, opened 2015-08-17
rpisense-fb: add low-light mode and gamma control

XECDesign:


popcornmix:
Looks fine to me. @pelwell any objections?

Issue #1110, opened 2015-08-19
DMA: Firmware: please provide more than 5 DMA channels to the linux kernel


msperl:
please either provide more static channels so that we can run spi + i2s + USB + MMC concurrently on the system or provide a "registration" system to allow for better sharing of DMA between VC and ARM (say via mbox messages).

HiassofT:
+1 for that request

I'm currently facing the issue that the Cirrus audio card driver no longer works with the 4.1 kernel.

Only 3 DMA channels out of the 5 are available, the driver needs SPI to configure the wm5102 (SPI needs 2 DMA channels) and the bcm2708-i2s also needs 2 DMA channels because the wm5102 supports both playback and record. So it's short by 1 DMA channel and bcm2708-i2s won't load...

I noticed a hardcoded limit of 5 in bcm2708-dmaengine.c inside the `#ifdef CONFIG_DMA_BCM2708_LEGACY` block of bcm2835_dma_probe. Does that have anything to do with the 5 channel limit perhaps?

pelwell:
I am investigating. It's my first time looking at the DMA driver(s), but the initial observation is that we are using the LEGACY_API mode which pre-configures 5 channels. This may be significant...

msperl:
But then the allocation should fail and not loop forever in kernel context (or so it seems)...

Do we need legacy mode?

I know that we can not use the upstream DMA driver as it does not contain all the features - there is a merge request by @notro, but that never got merged upstream...

HiassofT:
With the Cirrus driver DMA allocation fails rather gracefully. bcm2708-i2s gets the playback channel and then errors out when trying to acquire the record channel:

```
[    4.695533] bcm2708-i2s 3f203000.i2s: Missing dma channel for stream: 1
[    4.702319] bcm2708-i2s 3f203000.i2s: ASoC: pcm constructor failed: -22
```

At this point all 5 dma channels are marked in_use, if I rmmod bcm2708-i2s channel 4 is freed - nothing too unusual so far.

The hard lock up #1108 might be caused by something else.

msperl:
So I have done my testing and have come to the conclusion:
- assuming the sequence: modprobe snd_soc_rpi_dac; modprobe spi-bcm2835
  - I can load a device that does not trigger DMA transfers and everything works fine (modprobe ads7846)
  - but as soon as I load a device that DOES make use of DMA the system is "failing" possibly overwriting memory or similar (modprobe fb_s6d02a1)
[...]

Issue #1108, opened 2015-08-19
DMA: when spi and i2s dma enabled drivers are both loaded the system will crash


msperl:
when both:
- snd-soc-bcm2708-i2s
- spi-bcm2835

are loaded the rpi becomes unresponsive and no messages are printed on the serial console.
Only a reset/power cycle will bring the system back online.

There seems to be an issue with the dma-engine code that will trigger this issue...

HiassofT:
I'm seeing a similar issue here (RPi2 with kernel 4.1) with the following config (no external hardware required)

```
dtparam=spi=on
dtoverlay=rpi-dac
```

Raspbian boots up fine and the drivers are loaded, but as soon as I try to play back an audio file (eg `speaker-test -c 2`) the RPi locks up hard - no response to pings.

If I disable SPI or switch to spi-bcm2708 no lockups occur.

pelwell:
I think I may have found the problem. Channel 12 shouldn't be marked as being available because it doesn't have a dedicated interrupt line; instead, it shares channel 11's (along with 13 & 14). The driver doesn't cope with this, so only channel 11 from this group should be used. The interrupt which would logically belong to channel 12 (<1 28>) is actually a global interrupt for ALL channels. Enabling could easily cause a lockup.

See https://github.com/raspberrypi/linux/pull/1113

msperl:
Stupid question: Bulk DMA channel - It does not say anywhere in the documentation anything about this!
Searching for BULK/Bulk/bulk in the PDF does not give any result...

msperl:
OK - found it:

```
DMA channel 0 and 15 are fitted with an external 128 bit 8 word read FIFO. This enables efficient memory to memory transfers to be performed. This FIFO allows the DMA to 
```

HiassofT:
A quick test with #1113 shows that the rpi-dac + spi test no longer locks up. Thanks a lot @pelwell!

pelwell:
Thanks for finding the problem at such a convenient time.

pelwell:
Can we close this now?

PR #1113, opened 2015-08-20
bcm2708-dmaengine: Use more DMA channels (but not 12)

pelwell:
1) Only the bcm2708_fb drivers uses the legacy DMA API, and
it requires a BULK-capable channel, so all other types
(FAST, NORMAL and LITE) can be made available to the regular
DMA API.

2) DMA channels 11-14 share an interrupt. The driver can't
handle this, so don't use channels 12-14 (12 was used, probably
because it appears to have an interrupt, but in reality that
interrupt is for activity on ANY channel). This may explain
a lockup encountered when running out of DMA channels.

The combined effect of this patch is to leave 7 DMA channels
available + channel 0 for bcm2708_fb via the legacy API.

See: https://github.com/raspberrypi/linux/issues/1110
     https://github.com/raspberrypi/linux/issues/1108

### DIFF
--- a/arch/arm/boot/dts/bcm2708_common.dtsi
+++ b/arch/arm/boot/dts/bcm2708_common.dtsi
@@ -59,11 +59,10 @@
 				     <1 24>,
 				     <1 25>,
 				     <1 26>,
-				     <1 27>,
-				     <1 28>;
+				     <1 27>;
 
 			#dma-cells = <1>;
-			brcm,dma-channel-mask = <0x7f35>;
+			brcm,dma-channel-mask = <0x0f35>;
 		};
 
 		intc: interrupt-controller {
--- a/drivers/dma/bcm2708-dmaengine.c
+++ b/drivers/dma/bcm2708-dmaengine.c
@@ -184,7 +184,7 @@ static void vc_dmaman_init(struct vc_dmaman *dmaman, void __iomem *dma_base,
 }
 
 static int vc_dmaman_chan_alloc(struct vc_dmaman *dmaman,
-				unsigned preferred_feature_set)
+				unsigned required_feature_set)
 {
 	u32 chans;
 	int chan = 0;
@@ -193,10 +193,8 @@ static int vc_dmaman_chan_alloc(struct vc_dmaman *dmaman,
 	chans = dmaman->chan_available;
 	for (feature = 0; feature < BCM_DMA_FEATURE_COUNT; feature++)
 		/* select the subset of available channels with the desired
-		   feature so long as some of the candidate channels have that
-		   feature */
-		if ((preferred_feature_set & (1 << feature)) &&
-		    (chans & dmaman->has_feature[feature]))
+		   features */
+		if (required_feature_set & (1 << feature))
 			chans &= dmaman->has_feature[feature];
 
 	if (!chans)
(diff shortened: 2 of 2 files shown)


popcornmix:
Okay - feel free to merge.

msperl:
Can you also merge it into 4.2?

pelwell:
It's also on 4.2.

kukabu:
I'm not sure but look like this commit break sdhost-overlay and spi-dma-overlay
foundation kernel without this pacth (for apt-get) - https://gist.github.com/kukabu/da83930e40d54cb7c454
my custom kernel with this patch - https://gist.github.com/kukabu/f5fc74ea8f0f93481f0e
dmesg diff - https://gist.github.com/kukabu/dd61a4d6f45b343dceab

How do I debug the overlay?

pelwell:
It's working for me, so one of us is doing something wrong. Can you post your config.txt (just the active bits - you can delete the comments) and the output of:

```
sudo vcdbg log msg |& grep -viE "(hdmi|gpioman):"
```

please?

kukabu:
config.txt - https://gist.github.com/kukabu/9a7f8e508c58618d75b5
vcdbg log msg - https://gist.github.com/kukabu/5629b5c2b4368b0199c5

@pelwell thanks, if it work for you then I something broken :(

pelwell:
```
001291.049: Failed to load overlay 'sdhost-overlay'
001295.702: Failed to load overlay 'spi-dma-overlay'
```

The "-overlay" part is inferred and should not be used - remove it. However, at the moment it still ought to work, so that shouldn't be the problem.

Oh - now I see:

```
001231.524: Kernel trailer DDTK property says yes
```

DDTK is an unreleased development feature in the loader that will cause it to look for different filenames. `dtoverlay=sdhost` will look for `sdhost.dtbo` instead of `sdhost-overlay.dtb`, which isn't useful to you at the moment. This mode is selected by a flag in the kernel trailer (added by an unreleased version of mkknlimg).

How have you built this kernel? What does `scripts/knlinfo kernel.img` show (you may need to change the paths)?

Also, do you really want to turn on both i2c interfaces? i2c0 (or i2c1 on a Rev 1 Model B) is intended for the VideoCore processor. The recommended setting for people using i2c is:

```
dtparam=i2c_arm=on
```

which will automatically select the correct interface on each Pi. If you want to explicitly allow the ARM to access the VideoCore I2C then use:

```
dtparam=i2c_vc=on
```

but be aware that it will stop the camera from working.

pelwell:
OK, so that version _is_ released (in the tools repo), but it isn't one found in any of the kernel trees. I suspect you are trying to enable some extra kernel features and ended up setting CONFIG_OF_OVERLAY. This enables the new kernel dynamic Device Tree support, but it isn't very useful because it doesn't include a userspace interface; apart from some self tests, only one driver uses it in the 4.2 tree. Note that this isn't required in a standard Pi kernel because all of the overlay functionality is in the loader.

Your options are:
1) Don't define CONFIG_OF_OVERLAY.
2) Use the version of mkknlimg found in the scripts folder of the kernel tree.

For reference, the latest documentation for kernel building can be found [here](https://www.raspberrypi.org/documentation/linux/kernel/building.md).

I think any apparent link to this bcm2708-dmaengine commit is just coincidental.

kukabu:
I found the my mistake. CONFIG_OF_OVERLAY=y in .config.
@pelwell thanks and excuse my disturbing you.

PR #1077, opened 2015-07-19
BCM270X_DT: mz61581: Revert to spi-bcm2708

notro:
The MZ61581 display does not work with spi-bcm2835 and software chip selects. It works before the commit: spi: bcm2835: transform native-cs to gpio-cs on first spi_setup

Revert to spi-bcm2708 until the cause has been detected and the issue resolved.

Also increase reliability by adding a patch for missing display controller reset.
This fbtft patch is now in linux-next and will probably land in 4.3.

PR #1125, opened 2015-09-04
Rpi 4.1.y spi bcm2835 patches clock-polarity issues 

msperl:
merge from 4.3-pre back to 4.1 to fix #1077.

### DIFF
--- a/drivers/spi/spi-bcm2835.c
+++ b/drivers/spi/spi-bcm2835.c
@@ -480,7 +480,7 @@ static int bcm2835_spi_transfer_one_poll(struct spi_master *master,
 					 struct spi_device *spi,
 					 struct spi_transfer *tfr,
 					 u32 cs,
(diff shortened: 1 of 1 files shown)


msperl:
@notro: please check

pelwell:
Looks OK to me.

notro:
I haven't got time to actually test it now, but it looks like what I tested previously.
Thanks for getting this fixed Martin. This means that I don't have to provide a prebuilt spi-dma kernel for the fbtft users anymore :-)

msperl:
@notro: it is exactly what you have tested before - it is exactly upstream, so it should work...

msperl:
It may be valuable to push this into 4.2 as well, as it will only show up officially in 4.3!

popcornmix:
Was ahead of you there: https://github.com/raspberrypi/linux/tree/rpi-4.2.y

PR #1127, opened 2015-09-08
Fixes issue preventing DMA being used

rlbartle:
Fixes issue preventing DMA being used whenever the rx/tx_buf is not word aligned.

### DIFF
--- a/drivers/spi/spi-bcm2835.c
+++ b/drivers/spi/spi-bcm2835.c
@@ -386,14 +386,14 @@ static bool bcm2835_spi_can_dma(struct spi_master *master,
 	/* otherwise we only allow transfers within the same page
 	 * to avoid wasting time on dma_mapping when it is not practical
 	 */
-	if (((size_t)tfr->tx_buf & PAGE_MASK) + tfr->len > PAGE_SIZE) {
+	if (((size_t)tfr->tx_buf & PAGE_SIZE-1) + tfr->len > PAGE_SIZE) {
 		dev_warn_once(&spi->dev,
 			      "Unaligned spi tx-transfer bridging page\n");
 		return false;
 	}
-	if (((size_t)tfr->rx_buf & PAGE_MASK) + tfr->len > PAGE_SIZE) {
+	if (((size_t)tfr->rx_buf & PAGE_SIZE-1) + tfr->len > PAGE_SIZE) {
 		dev_warn_once(&spi->dev,
-			      "Unaligned spi tx-transfer bridging page\n");
+			      "Unaligned spi rx-transfer bridging page\n");
 		return false;
 	}
 


popcornmix:
ping @msperl @pelwell

msperl:
What problem does this solve?
Which driver sends unaligned data?
Also I would recommend to first upstream the patch and then only merge it into the foundation.

pelwell:
Regardless of whether the patch is needed - and likely it is needed otherwise @rlbartle probably wouldn't have spotted or submitted it - the code is incorrect without it and correct with it, which surely is an end in itself.

msperl:
Still - it would help share with which driver the issue was detected (because there are potentially some means to run DMA even for those transfers - but without a use-case that I can test myself it is hard to verify).

Anyway: now looking at the defines it seems as if I have mistaken the MASK for its inverse, so in principle it is correct.

Still, as we are using an upstream driver I recommend you push the change upstream first to keep the differences minimal - especially from a merging perspective...

msperl:
Note that I would add brackets around (PAGE_SIZE - 1)

rlbartle:
Ok I'll try a pull upstream.

My driver is at [axpi-fpga](https://github.com/rlbartle/axpi-fpga)

msperl:
If you prefer I can also push a patch for you with you reporting...
I wonder if it would be sensible to work on those things to enable transfers for unalligned transfers passing multiple pages...

msperl:
Unfortunately linux upstream does not work via github...

I have created a patch that I can submit upstream, but to report/attribute it correctly I would need your (@rlbartle) email...

rlbartle:
Thanks, it's robert@axium.co.nz.

msperl:
will send it out to the correct mailing list today...

msperl:
sent patch to the spi and rpi mailinglists - you are in cc

msperl:
note that I still would recommend that you align your transfers to word, so that you can always be assured to have an optimal transfer speeds - even if you transfer more than one page size.

The reason behind this is that the spi-framework will split your transfer into a list of individual transfers of single pages. so if you start your transfer at PageBoundry+10 and you transfer 4096 bytes it will create a list of:
- [ Pageboundry+10 : Pageboundry+4096 [ - so 4086 bytes
- [ Pageboundry+4096 : Pageboundry+4106 [ - so 10 bytes

The problem behind that is that the DMA can only do word-sized transfers so for the first frame it would transfer 1022 words (=4088) bytes into the FIFO and not 4086 as required, which leaves 2 bytes in the FIFO.

You could say that these frames are bordering each other, but that is not necessarily the case in physical memory (even if it is true in all cases right now) and that is why we have split at page boundaries.

There are ways around this, but it would mean a lot of coding for edge-cases.

We would need to reset the transfer (also on the SPI side internally) to clean the remaining 2 bytes from the fifo and then restart the DMA process.

I guess we can only argue for that to get merged if there is an upstream-driver that really requires unaligned access and which can not be made aligned easily...

I hope this explains the why of the current situation better....

msperl:
got applied by Mark into the spi tree immediately - should propagate into the kernel...
(note that the patch includes brackets, so is slightly different to this merge request)

popcornmix:
@msperl can you point me at the commit? I can then cherry-pick it.

msperl:
It is in the spi fork (//git.kernel.org/pub/scm/linux/kernel/git/broonie/spi.git) - I will tell you when it gets into repo of linus (it is already in spi/for-linus)...
(unless you want to merge from the spi fork, where the commit is: 2a3fffd45822070309bcf0b1e1dae624d633824a)

pelwell:
I've cherry-picked the patch from linux-next onto rpi-4.1.y and rpi-4.2.y.

Please close if you are happy with the result.

PR #1130, opened 2015-09-12
rtl8192c_rf6052: PHY_RFShadowRefresh(): fix off-by-one

marckleinebudde:
Signed-off-by: Marc Kleine-Budde mkl@blackshift.org


popcornmix:
Ideally should be submitted to realtek:
http://www.realtek.com.tw/downloads/downloadsView.aspx?Langid=1&PNid=21&PFid=48&Level=5&Conn=4&DownTypeID=3&GetDown=false&Downloads=true

But looks valid, so I've cherry-picked it.

PR #1132, opened 2015-09-13
Revert "BCM270X_DT: mz61581: Revert to spi-bcm2708"

notro:
This reverts commit 1820cd05d93b2d465d1616202772efe5bf0d11fe.

The spi-bcm2835 driver has been fixed, so now we can use it again.

Signed-off-by: Noralf Trønnes noralf@tronnes.org

### DIFF
--- a/arch/arm/boot/dts/overlays/README
+++ b/arch/arm/boot/dts/overlays/README
@@ -368,6 +368,8 @@ Params: speed                    Display SPI bus speed
 
         fps                      Delay between frame updates
 
+        txbuflen                 Transmit buffer length (default 32768)
+
         debug                    Debug output level {0-7}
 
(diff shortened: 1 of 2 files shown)


notro:
Don't pull this yet, I have to change txbuflen to get good throughput. The dma mapping/engine overhead is high.

notro:
Done updating PR.

notro:
For the SPI DMA curious:
A full display update is 300kB.
Using a 4k transmit buffer gives a measured throughput of 8MB/s at 128MHz (75 spi transfers).
Using a 32k buffer gives a throughput of 11 MB/s (10 spi transfers).

For comparison, the out-of-tree [spi-bcm2708](https://github.com/notro/spi-bcm2708) can do 12MB/s with 4k buffers, but relies on the legacy DMA API.

PR #1134, opened 2015-09-15
Use dtb-dirs feature for overlays.

dlech:
The kernel makefiles have a dtb-dirs target that is for vendor subdirectories.

Using this fixes install_dtbs, which previously did not install the overlays.


pelwell:
Nice. Thanks.

[ For other readers, the make target is 'dtbs_install' ]

PR #1135, opened 2015-09-15
mmc: spi: disable dma mapping code, as it does not work

msperl:
since:
    commit 0589342c27944e50ebd7a54f5215002b6598b748
    Author: Rob Herring rob.herring@calxeda.com
    Date:   Tue Oct 29 23:36:46 2013 -0500

```
     of: set dma_mask to point to coherent_dma_mask

     Platform devices created by DT code don't initialize dma_mask pointer to
     anything. Set it to coherent_dma_mask by default if the architecture
     code has not set it.
```

the mmc-spi driver no longer detects the sd-cards any longer.

This patch disables the offending section in mmc_spi
so that the driver works again and detects the sd card normally.

There is no real negative impact as the spi-framework
does no longer rely on driver-provided DMA mappings but
maps the transfers transparently.


msperl:
this also applies to 4.1.y

pelwell:
Do you think this is a Pi-specific issue?

Have you submitted this patch upstream?

msperl:
maybe - there was one longer discussion where I was asked to bisect the issue (and I got to that specific commit), but the thread ended without any conclusion.

My personal solution was to have the above patch (because it is unclear where things fall over) - as you provide an overlay for it I guess you may want that the driver works as well...

So I wanted to share this observation and the patch - others may want to use it as well...

Note that it also happens in the upstream 4.2, so it is somewhat generic...

Ruffio:
@msperl this points to 4.2.y. If it should get merged, then it must point to 4.4.y. Is it still relevant?

Ruffio:
@msperl is this still relevant? It is pointing to v. 4.2 and currently we are running 4.4.. IMHO it should be closed...

msperl:
well - it still is not working, but if you want to ignore that issue, then you should remove the spi-mmc module from the kernel

mi-hol:
@msperl @Ruffio  the way I understand Rob Herring's commit comment at top, is that all ARM Linux distributions would suffer from the same issue. Hence I'd agree with @pelwell 's question ' is a Pi-specific issue?' 
Issue resolution will need to be approached differently.
I wonder if it should be related to #1801 brought up by @Electron752 
Any thoughts?

Electron752:
I'm not sure how it relates to issue #1801 as I can't see the exact change at this point.

I have hit a bunch of issue with the DMA mask needing to be explicitly set otherwise things don't work at all. Especially on arm64.   I just assumed it was some kind of policy decision to get people to cleanup their drivers...

pelwell:
I don't think it is related (the change is here: https://github.com/raspberrypi/linux/pull/1135/commits/aef50a19f335ae3d2eb43ac5496c7f03930b394f).

pelwell:
Submitted upstream.

JamesH65:
Did this ever get accepted?

pelwell:
It died when the initial response wasn't a "Yes" and other things took over.

JamesH65:
@pelwell Did upstream ever take this?

pelwell:
No, but with nobody (including the author) showing an interest I didn't put a lot of effort into it.

PR #1152, opened 2015-10-02
Add support for the HiFiBerry DAC+ Pro.

hifiberry:
The HiFiBerry DAC+ and DAC+ Pro products both use the existing bcm sound driver with the DAC+ Pro having a special clock device driver representing the two high precision oscillators.

An addition bug fix is included for the PCM512x codec where by the physical size of the sample frame is used in the calculation of the LRCK divisor as it was found to be wrong when using 24-bit depth sample contained in a little endian 4-byte sample frame.


pelwell:
It's a large change, but apart from the pcm512x tweak it is Hifiberry specific, and the code looks OK to me. No objections from me.

PR #1160, opened 2015-10-11
Fix I2S issues with bcm2835-dma

HiassofT:
dreq setup was missing for slave transfers and for some (yet unknown) reason transfer sizes larger than 32k cause repated clicking on I2S soundcards.

These changes bring bcm2835-dma in line with bcm2708-dmaengine, but only limit cyclic lite transfers to 32k - other lite transfers can still use 64k-4.

See also discussion in #1153

ping  @notro


notro:
Looks good, thanks @HiassofT

PR #1164, opened 2015-10-12
Merge clk-bcm2835 driver

anholt:
As part of getting ready for merging the vc4 driver to the RPi kernel, I need the ability to set the pixel clock and HDMI state machine clocks.  I've been working on upstreaming this driver, and the subsystem maintainer said he hoped to merge it last week, but it's still not up in clk-next.  So, instead, this is a cherry-pick of those queued changes plus a couple new ones to get the driver to build on the 2708/2709 arches.

Note that it'll probe and read out the clock rate of all the clocks it defines, but won't do anything with them until you hook some driver up to use them (due to the CLK_IGNORE_UNUSED flags).  We could potentially use it in the 2708_common dt for other clocks, but I figured we should start with merging the code at all before worrying about potentially disrupting other drivers.


anholt:
Of course, as soon as I submit this, clk/clk-bcm2835 gets updated with my driver merged (so it should appear in 4.4).  I'm still OK with using this pull request, since the rpi trees do rebases all the time anyway.

popcornmix:
Currently rpi-4.3.y isn't stable for pi2 (I think it is related to interrupt controller changes).
I'll give this a test on pi1 which I think is stable.

I don't imagine a problem merging this in assuming it doesn't break normal usage.
@pelwell I assume you're okay with this?

anholt:
Which tree should I be working against for getting things landed for next week?  I was assuming it was just "grab the current one."  (I'm trying to get as much of the merging done as I can, so we can focus on integration issues next week)

popcornmix:
rpi-4.1.y is our current stable (for both apt-get updates and rpi-updates).
rpi-4.2.y is the more experimental tree which uses more upstream drivers (e.g. we switched to upstream dma driver last week). It is used by OSMC so needs to be fairly stable.
rpi-4.3.y is currently unstable. All drivers seem to work on both Pi1 and Pi2, but on Pi2 there are panics. Sometimes on boot, and sometimes it runs for quite a while.

So, I guess the first question is which upstream drivers do you need and which trees are they present in? In theory we can probably backport things we need, but it's better if they are already there.

Is Pi2 support essential now? We could focus on getting Pi1 integrated until Pi2 is working (I'm sure Pi2 can be fixed on 4.3).

The other question is how likely the integration is to break the existing kernels. As no one is using 4.3 yet, then there is no real worry. 4.2 and (more so) 4.1 we care about keeping them stable more. Obviously we can work on a branch, but that creates some extra merging work.

I think if Pi2 was reliable I'd definitely prefer working on the 4.3 tree.

Anyway, Pi1 4.3 kernel has built with this PR and boots. I'll have a play and see if it seems stable.

anholt:
Until 4.4, I don't think I'm picky about which tree we work in, just as long as it's something we'll be able to get into users' hands easily (I'm with Eben on this -- I definitely want people to be able to start playing with the new driver in Raspbian).

popcornmix:
Okay, perhaps target 4.2. That is stable and we wouldn't have a problem with bumping apt-get/rpi-update to it in the near future.

anholt:
OK.  Should I do a new pull-request to it for clocks?

popcornmix:
Yep

popcornmix:
I'll close this for now. Once everything is working in 4.2, I'll cherry-pick to 4.3.

PR #1165, opened 2015-10-12
Merge bcm2835-clock driver to 4.2

anholt:
Rebased from 4.3 to 4.2.  Working on vc4 itself, next.


pelwell:
As I said on #1164: Given that nothing is using bcm2835-cprman yet - it has no clients in bcm2708_common.dtsi - it should be disabled by default; it can than be enabled using an overlay, or in a derived .dts file once it is being used.

popcornmix:
@anholt Are you happy to make @pelwell suggested change? We can then merge this.

anholt:
Yeah, I did that in my tree here and I've been working on the new series with vc4.  Once I trust that things are working for consuming cprman again, I'll send the updated pull req.

I also wrote a patch for converting the DT's fixed clocks to use cprman, which didn't boot and I haven't sorted that out yet (rpi-merge-clk-4.2-allclocks).

anholt:
I've force-pushed with the DT change having status=disabled.  I've tested KMS on top of it at this point.

pelwell:
Thanks, Eric.

PR #1166, opened 2015-10-13
Multichannel pcm

ghost:
This raises the number of channels to 8, and adds support for the ALSA channel mapping API.

It also includes some other changes, which are strictly speaking not related (see commit list).


popcornmix:
I'd like to get this in, but hopefully remove a few of the uglier bits.
I've just pushed an updated firmware which should support using this vc_vchi_audioserv_defs.h
http://paste.ubuntu.com/12991296/

It bumps the audioserv version to 3. As long as kernel reports 3 it will apply channelmap from VC_AUDIO_CONFIG_T. channelmap is ignored is kernel reports version of 2. As the extra parameter doesn't change the union size, it shouldn't break anything.

Can you remove the gencmd and try using the new scheme for setting channelmap.

I'd also like to make firmware handle non-power-of-two number of channels.

ghost:
> Can you remove the gencmd and try using the new scheme for setting channelmap.

Yes, I'll try to get to it soon.

> I'd also like to make firmware handle non-power-of-two number of channels.

In audioserv version 3?

popcornmix:
> In audioserv version 3?

Not in yet. It can probably be added to version 3 as I doubt anyone other us will be relying on features form version 3 in the next week or so.

ghost:
Replaced with #1257. (Not going to rely on github to pick the correct base, so I opened a new PR.)

PR #1170, opened 2015-10-19
New overlay for PiScreen 2 R

mwilliams03:
This is a new overlay for the PiScreen2r.   Which is a new version of PiScreen (PiScreen 2) which features a new chipset.
This was created with the assistance of @notro


pelwell:
The overlay looks good to me, but it needs a change to the Makefile to build it and an entry in the README. If you make those additions, squash them into a single commit and `git push -f` then I'll happily merge it.

mwilliams03:
Done... and thank you.

pelwell:
I've squashed these together and merged them.

PR #1173, opened 2015-10-21
rpi_display: add backlight driver and overlay

P33M:
Add a mailbox-driven backlight controller for the Raspberry Pi DSI
touchscreen display. Requires updated GPU firmware to recognise the
mailbox request.

cc: Gordon Hollingworth gordon@raspberrypi.org


pelwell:
checkpatch has a few issues with this commit. So far we haven't been sticklers for Signed-off-by lines and updatng MAINTAINERS, but we do hold ourselves and others to the line length and whitespace rules. Here's the checkpatch output in its entirety:

```
WARNING: please write a paragraph that describes the config symbol fully
#82: FILE: drivers/video/backlight/Kconfig:268:
+config BACKLIGHT_RPI

WARNING: added, moved or deleted file(s), does MAINTAINERS need updating?
#103: 
new file mode 100644

WARNING: line over 80 characters
#146: FILE: drivers/video/backlight/rpi_backlight.c:39:
+   ret = rpi_firmware_property(gbl->fw, RPI_FIRMWARE_FRAMEBUFFER_SET_BACKLIGHT,

ERROR: trailing whitespace
#152: FILE: drivers/video/backlight/rpi_backlight.c:45:
+^I$

ERROR: space required before the open parenthesis '('
#153: FILE: drivers/video/backlight/rpi_backlight.c:46:
+   if(brightness < 0) {

WARNING: struct of_device_id should normally be const
#216: FILE: drivers/video/backlight/rpi_backlight.c:109:
+static struct of_device_id rpi_backlight_of_match[] = {

WARNING: DT compatible string "rpi,rpi-backlight" appears un-documented -- check ./Documentation/devicetree/bindings/
#217: FILE: drivers/video/backlight/rpi_backlight.c:110:
+   { .compatible = "rpi,rpi-backlight" },

WARNING: DT compatible string vendor "rpi" appears un-documented -- check ./Documentation/devicetree/bindings/vendor-prefixes.txt
#217: FILE: drivers/video/backlight/rpi_backlight.c:110:
+   { .compatible = "rpi,rpi-backlight" },

total: 3 errors, 6 warnings, 183 lines checked

NOTE: Whitespace errors detected.
      You may wish to use scripts/cleanpatch or scripts/cleanfile

0001-rpi_display-add-backlight-driver-and-overlay.patch has style problems, please review.

NOTE: If any of the errors are false positives, please report
      them to the maintainer, see CHECKPATCH in MAINTAINERS.
```

pelwell:
My overlay checking script is very picky about the formatting of the overlays README file. TABs are banned, there should be two blank lines between overlays, and `<none>` must be `<None>`.

Oh, and there must be an overlay to go with the documentation.

P33M:
Pushed an update. Still the question about whether we even need the .check_fb function - I don't know enough about the internals of framebuffer/backlight to say one way or the other.

notro:
Without check_fb, the display will blank regardless of which framebuffer received the blank command.
But the way it's implemented here it can't distinguish between fb's because gbl->fbdev is not set anywhere so it's always NULL.
Looking at [gpio_backlight](http://lxr.free-electrons.com/source/drivers/video/backlight/gpio_backlight.c#L111) implementation, gbl->fbdev is only set from platform_data and not when using DT.

check_fb, if it exists, is called from fb_notifier_callback() in [drivers/video/backlight/backlight.c](http://lxr.free-electrons.com/source/drivers/video/backlight/backlight.c#L40)

[fb_blank()](http://lxr.free-electrons.com/ident?i=fb_blank) is responsible for generating the fb notifier blanking event (this event will only be sent if the display driver has a fb_blank() function and it returns 0).

P33M:
This is where it gets complicated - the backlight is only applicable to the DSI screen yet Linux doesn't know if the framebuffer is being output on the HDMI port or the DSI port. GPU firmware calls can explicitly specify output displays but this is independent of the fb interfaces.

We can sort-of work around this using the firmware - it can be made to, depending on whether the LCD is attached and enabled, add this node to the DT?

notro:
Are you saying that both displays are using the bcm2708_fb driver and that the firmware decides which display gets the output?
If so, can't you just let the bcm2708_fb blanking (RPI_FIRMWARE_FRAMEBUFFER_BLANK) control the DSI backlight?

notro:
I forgot that this driver can set brightness and not just on/off, so RPI_FIRMWARE_FRAMEBUFFER_BLANK doesn't cut it.

How will DSI be handled when the vc4 driver is out? Will it get it's own driver?

P33M:
The vc4 kms driver currently targets HDMI outputs only. DSI/DPI are separate peripherals thus would require release of documentation for those peripherals. DSI's going to remain blob-side only for the time being.

P33M:
My impression is that firmware should add the dt node if the DSI display is attached and active (config.txt:ignore_lcd != 0).

I can't see any obvious way to link up the fbdev with the backlight in the DT case - looking through all other users of gpio-backlight or pwm-backlight, they seem to just default to saying "yes" to check_fb.

pelwell:
If the driver is effectively going to just say yes then I would prefer to make that explicit in the code.

P33M:
Which would be achieved by removing ->check_fb.

notro:
> I can't see any obvious way to link up the fbdev with the backlight in the DT case

One way to solve this is to let the display driver get the backlight device using of_find_backlight_by_node(). Now info->bl_dev can be set by the display driver and the backlight driver can do this:

``` c
static int rpi_backlight_check_fb(struct backlight_device *bl,
                                  struct fb_info *info)
{
        return info->bl_dev == bl;
}
```

The major problem with this approach is that it triggers deferred probing because fbdev is placed before backlight in the linking order, so when the display driver is probed the backlight isn't available yet.
The display driver then needs to return -EPROBE_DEFER and try again later when the backlight has been probed. This happens in late_initcall (3sec after the video driver was first probed).
There are discussions going on to try and address this deferred probing problem for video drivers (and others) but no solution has been agreed upon yet.

The only situation where it's a problem that we can't match fb to backlight is when we have multiple framebuffer devices AND we need to control blanking individually.
I suggest you just drop check_fb.

pelwell:
@P33M - can you make it so?

P33M:
Yes. One more tweak to the firmware to make everything automagic, too.

P33M:
Driver update pushed.

popcornmix:
@pelwell happy? I have no objections.
Note: this PR is against 4.3 tree. I don't think that matters - I'll cherry pick it to 4.2 and 4.1 when merged.

Issue #1147, opened 2015-09-28
How to support the 7" touch display from raspberrypi org on a custom kernel?


DieterReuter:
We do use a custom built kernel 4.1.x for the Raspberry Pi and would like to support the new 7" touch display. Until now I do have the pure display supported but without the touch feature.

I just need a pointer what kernel config is necessary, and/or what DTS overlay file we have to include and maybe how to activate it. 

Thanks for any help or pointers!

popcornmix:
Add:

```
CONFIG_INPUT_TOUCHSCREEN=y
CONFIG_TOUCHSCREEN_RPI_FT5406=m
```

and make sure you ship rpi-ft5406-overlay.dtb

The touchscreen overlay is enabled automatically when the LCD is detected by firmware.

procount:
Would I be able to add this touchscreen driver to the Kernel used with NOOBS (3.18.7-rescue-v7)?
Would it be easy or a complicated backport (I have no experience of backporting)? Which files do I need to port - just: drivers/input/touchscreen/rpi-ft5406.c & Kconfig?
Or is the preferred method to use a later v4.1.x Kernel for NOOBS?

EDIT: On further investigation I would need to add: 
include/linux/platform_data/mailbox-bcm2708.h
drivers/input/touchscreen/Makefile
from commit fa2b33b614fead97a2eb8b950fb1c77e73f0eb28

DieterReuter:
Thanks for the answers. I had all the right kernel setting still applied, but was missing to enable the Device Tree support on the kernel with the script `mkknlimg`. Once this step was included in our build script, the 7" touch screen works just fine.

procount:
Hmmm. I'd forgotten DT support - I don't think that is in v3.18. Is this a showstopper, or is there a workaround?

pelwell:
3.18 is the first released Pi kernel that supported DT, so that in itself shouldn't be a problem.

procount:
OK down the rabbit hole then! :) Is this the correct way?
Step 1 - Enable DT support in buildroot kernel with Boot Options/ Flattended DeviceTree Support
Step 2 - Tag the kernel image as DT aware using mkknlimg
Step 3 - Test kernel DT is working
Step 4 - Add the RPI_FT5406 touchscreen into the kernel
Step 5 - cross fingers & test

pelwell:
[...]

PR #1178, opened 2015-10-28
ARCH_BCM270X: Drop ATAGS support

notro:
This PR makes Device Tree mandatory for booting the ARCH_BCM2708 kernel.

We have used Device Tree since we [switched to 3.18](https://www.raspberrypi.org/forums/viewtopic.php?f=29&t=97216) in January and it's time to remove [ATAGS](http://stackoverflow.com/questions/21014920/arm-linux-atags-vs-device-tree) support.

This will cut down on the amount of patches that we have to keep around and apply on each new kernel release. It is also a necessary step on the way to get one kernel image that can boot both Pi1 and Pi2 (instead of having kernel.img and kernel7.img).

A /boot/config.txt that disables Device Tree will not boot anymore:

```
device_tree=
```

There will be no visible sign that the kernel didn't receive a Device Tree since it won't know which drivers to load and hence can't display an error (a custom kernel with EARLY_PRINTK enabled would print an error on the serial console).


popcornmix:
Great I'll give this a test. Been looking forward to this day. (Even more so when 2709 gets the same treatment).

popcornmix:
You could cherry-pick https://github.com/raspberrypi/linux/pull/1163 and then drop bcm2708-i2s.

popcornmix:
> 67 files changed with 356 additions and 5,887 deletions.

Awesome! Still boots and seems to work.

notro:
> Been looking forward to this day

Me too. ARCH_BCM2835 is now just around the corner.

> You could cherry-pick #1163 and then drop bcm2708-i2s.

Will do.

notro:
**Version 2**

Removed bcm2708-i2s driver. bcm2835-i2s patches were already applied to rpi-4.3.y.

Removed the 4M init_dma_coherent_pool_size so we're back to the default 256k.

Added reboot_part parameter.
I haven't tested it with NOOBS, but this leads to poweroff:

```
echo 63 | sudo tee /sys/module/bcm2708/parameters/reboot_part
sudo reboot
```

Any other value (1,5) leads to a regular reboot. Does the bootloader boot partiton 0 if the asked for partition isn't bootable?

Added aux enable node so we can enable uart1 from board_bcm2835.

popcornmix:
If a partition doesn't exist it falls back to partition zero.

notro:
I discovered that the kernel was hung with no message on the console 30 min after last round of testing.
Checking back now after one hour and it's hung again.
I will add back init_dma_coherent_pool_size() tomorrow and see if that helps.

notro:
I rebuilt the kernel from scratch and haven't been able to reproduce the problem.

notro:
I'm not sure what I should do with reboot_part.
I have added sysfs symlinks `module/{bcm2708,bcm2709} -> module/bcm2835_wdt`, but just realised that the bcm2709 symlink won't work because I can't turn bcm2709 into multiplatform yet.
Is it possible to let NOOBS use `/sys/module/bcm2835_wdt/parameters/reboot_part` directly?
That would remove the need for those symlinks.

popcornmix:
If necessary NOOBS can use a new mechanism when choosing next partition to boot from. The kernel version and NOOBS version tend to be tied together (updating NOOBS kernel is not generally supported).

There will be other users of reboot_part. We use it in production tests to signal results from a previous boot. No doubt others are using it for multiboot purposes. But if it needs to change, then it needs to change. It's not too hard to test for both possible paths.

Ping @lurch @ghollingworth that a future kernel update may require NOOBS changes.

lurch:
Thanks for the heads up. NOOBS v1.4.x is still using a 3.18 kernel ( https://github.com/raspberrypi/noobs/blob/master/buildroot/.config#L309 ) without DT enabled.

If/when it becomes necessary to change the reboot_part logic though, it'll be easy to modify: https://github.com/raspberrypi/noobs/blob/master/recovery/util.cpp#L143

pelwell:
On the subject of updating NOOBS, are you aware of @procount's efforts to port NOOBS to 4.1? You can read the saga of the odyssey [here](https://github.com/raspberrypi/linux/issues/1147#issuecomment-152034107).

lurch:
@pelwell Yup, we've been communicating via email :-) It's just a shame that I'm so busy at the moment that I haven't been able to give him more help.

procount:
An odyssean saga? - definitely! But I'm just a small fish standing on the "shoulders of giants", grateful for any tidbits thrown my way.

popcornmix:
@notro v4.3 is out, so this needs a rebase.

notro:
The Pi was dead again today, no error messages. It had just been sitting idle.
I'll start bisecting.

popcornmix:
I've noticed that without this PR the USB hardware correctly accesses memory through the 0x4 alias bus address. After this PR it accesses memory incorrectly through the 0x0 alias.

The GPU debugger catches bad accesses:

```
0x0000_0001  L2_ALIAS_EX         L2 bad address exception                 
   [    0] 0x1        ALIASEX             Address exception                        occurred
0x8000_00c0  L2_ALIAS_EX_ID      L2 bad address ID                        
   [ 7- 0] 0xc0       AXIID               Axi ID of bad access                     USB_ID
   [   31] 0x1        WRITE               Read (0) or Write (1)                    Write
0x0f85_4d30  L2_ALIAS_EX_ADDR    L2 bad address                           
   [31- 0] 0xf854d30  ADDR                Address of the bad access                0xf854d30
```

Not sure if that is the cause of your instability, but it is something suspicious.

notro:
The 'Convert to multiplatform' patch changes the address logic, so I guess it's the culprit.
If so it's a bit strange that this didn't surface when I did my earlier bcm2835 work. Maybe I didn't have long enough run times.
Can you checkout the commit before the multiplatform one and try that with the debugger?

popcornmix:
The difference between 0x0 and 0x4 alias is quite minor (whether it goes through GPU's L1 cache). Most peripherals like USB can't see the L1 cache, so may get the same data in either case (although we are in undefined behaviour territory). Each access does generate an exception on the GPU, so at the very least kernel accesses will cripple the gpu performance.

If the same error applies to Pi2 (0x0 alias instead of 0xC alias), then it is likely to be far more obviously broken as peripherals can see the GPU's L2 cache (but don't want to).

I'll try to do the test. Note: with v4.3 there are a few build errors with platform include files. I'm hacking it by adding extra header includes, but I suspect there is a better solution.

popcornmix:
I can confirm that the GPU exceptions start with the multiplatform commit.

notro:
Can you give me an example statement from the usb driver which triggers this exception?

notro:
I've spotted something that could be the source of the problem.
If DWC_DMA_ALLOC() is the way dma buffers are allocated, then it doesn't get the required device translation (DT property dma-ranges).
dma_ctx is NULL, but it should be the device that requests the buffer. Both vchiq and bcm2708_fb had the same issue, but for those it was easy to spot since they didn't work at all.

drivers/usb/host/dwc_common_port/dwc_os.h

``` c
#define DWC_DMA_ALLOC(_size_,_dma_) __DWC_DMA_ALLOC(NULL, _size_, _dma_)
```

drivers/usb/host/dwc_common_port/dwc_common_linux.c

``` c
void *__DWC_DMA_ALLOC(void *dma_ctx, uint32_t size, dwc_dma_t *dma_addr)
{
#ifdef xxCOSIM /* Only works for 32-bit cosim */
    void *buf = dma_alloc_coherent(dma_ctx, (size_t)size, dma_addr, GFP_KERNEL);
#else
    void *buf = dma_alloc_coherent(dma_ctx, (size_t)size, dma_addr, GFP_KERNEL | GFP_DMA32);
#endif
    if (!buf) {
        return NULL;
    }

    memset(buf, 0, (size_t)size);
    return buf;
}
```

notro:
**Version 3**

Removed bcm2835_wdt sysfs symlinks. NOOBS have to use the new path when it switches to kernel version >4.1: /sys/module/bcm2835_wdt/parameters/reboot_part

Pass in the device when allocating dma memory in dwc_otg to get the correct bus address.

@popcornmix Please test with the debugger to verify that the exceptions are indeed gone.

HiassofT:
Did some tests, this is already looking really good, great job!

Only a few minor things:

Currently I have to patch mkknlimg so the kernel doesn't get flagged as 2835, otherwise it won't boot. Manually specifying the device_tree file in config.txt also works, but that's a PITA as I swap between B and B+ quite often during testing. Some idea how to improve 2835 detection in mkknlimg?

I also get a warning in dmesg every time an I2S audio device is closed. Looks like dma_free_coherent now doesn't like to be called with IRQs disabled:

```
[   72.211019] ------------[ cut here ]------------
[   72.211101] WARNING: CPU: 0 PID: 0 at include/asm-generic/dma-mapping-common.h:274 bcm2835_dma_desc_free+0xc0/0xd0()
[   72.211118] Modules linked in: snd_soc_pcm512x_i2c snd_soc_pcm512x regmap_i2c snd_soc_iqaudio_dac snd_soc_bcm2835_i2s regmap_mmio i2c_bcm2708 snd_soc_core snd_compress bcm2835_gpiomem snd_pcm_dmaengine snd_pcm snd_timer snd uio_pdrv_genirq uio fuse
[   72.211212] CPU: 0 PID: 0 Comm: swapper Not tainted 4.3.0+ #1
[   72.211226] Hardware name: BCM2708
[   72.211289] [<c001704c>] (unwind_backtrace) from [<c0013d88>] (show_stack+0x20/0x24)
[   72.211338] [<c0013d88>] (show_stack) from [<c032605c>] (dump_stack+0x20/0x28)
[   72.211374] [<c032605c>] (dump_stack) from [<c00267bc>] (warn_slowpath_common+0x8c/0xc4)
[   72.211403] [<c00267bc>] (warn_slowpath_common) from [<c00268b0>] (warn_slowpath_null+0x2c/0x34)
[   72.211434] [<c00268b0>] (warn_slowpath_null) from [<c036d3e4>] (bcm2835_dma_desc_free+0xc0/0xd0)
[   72.211466] [<c036d3e4>] (bcm2835_dma_desc_free) from [<c036d544>] (bcm2835_dma_terminate_all+0xc0/0x2b8)
[   72.211519] [<c036d544>] (bcm2835_dma_terminate_all) from [<bf07f0f8>] (snd_dmaengine_pcm_trigger+0x70/0x180 [snd_pcm_dmaengine])
[   72.211777] [<bf07f0f8>] (snd_dmaengine_pcm_trigger [snd_pcm_dmaengine]) from [<bf0a1d8c>] (soc_pcm_trigger+0xb4/0x134 [snd_soc_core])
[   72.212040] [<bf0a1d8c>] (soc_pcm_trigger [snd_soc_core]) from [<bf060648>] (snd_pcm_do_stop+0x64/0x68 [snd_pcm])
[   72.212170] [<bf060648>] (snd_pcm_do_stop [snd_pcm]) from [<bf0603d4>] (snd_pcm_action_single+0x48/0x88 [snd_pcm])
[   72.212296] [<bf0603d4>] (snd_pcm_action_single [snd_pcm]) from [<bf063670>] (snd_pcm_drain_done+0x24/0x2c [snd_pcm])
[   72.212428] [<bf063670>] (snd_pcm_drain_done [snd_pcm]) from [<bf066cdc>] (snd_pcm_update_state+0xe0/0x110 [snd_pcm])
[   72.212571] [<bf066cdc>] (snd_pcm_update_state [snd_pcm]) from [<bf066ec4>] (snd_pcm_update_hw_ptr0+0x1b8/0x3ac [snd_pcm])
[   72.212706] [<bf066ec4>] (snd_pcm_update_hw_ptr0 [snd_pcm]) from [<bf067164>] (snd_pcm_period_elapsed+0xac/0xd4 [snd_pcm])
[   72.212803] [<bf067164>] (snd_pcm_period_elapsed [snd_pcm]) from [<bf07f3fc>] (dmaengine_pcm_dma_complete+0x54/0x58 [snd_pcm_dmaengine])
[   72.212847] [<bf07f3fc>] (dmaengine_pcm_dma_complete [snd_pcm_dmaengine]) from [<c036c5d0>] (vchan_complete+0xdc/0x140)
[   72.212887] [<c036c5d0>] (vchan_complete) from [<c002ac24>] (tasklet_action+0x88/0xe0)
[   72.212916] [<c002ac24>] (tasklet_action) from [<c002a0ac>] (__do_softirq+0x108/0x3a8)
[   72.212942] [<c002a0ac>] (__do_softirq) from [<c002a6e0>] (irq_exit+0xcc/0x120)
[   72.212968] [<c002a6e0>] (irq_exit) from [<c00648c4>] (__handle_domain_irq+0x60/0xb4)
[   72.213004] [<c00648c4>] (__handle_domain_irq) from [<c0010928>] (handle_IRQ+0x2c/0x30)
[   72.213032] [<c0010928>] (handle_IRQ) from [<c0009488>] (bcm2835_handle_irq+0x3c/0x58)
[   72.213064] [<c0009488>] (bcm2835_handle_irq) from [<c05f4144>] (__irq_svc+0x44/0x7c)
[   72.213080] Exception stack(0xc08a3f08 to 0xc08a3f50)
[   72.213102] 3f00:                   00000000 00000000 c08a3f68 c08a5028 00000002 c08a2000
[   72.213128] 3f20: c08a4158 00000000 c0909e84 c09093c7 c09093c7 c08a3f64 c08a3f58 c08a3f58
[   72.213145] 3f40: c00109e4 c00109e8 60000013 ffffffff
[   72.213174] [<c05f4144>] (__irq_svc) from [<c00109e8>] (arch_cpu_idle+0x30/0x4c)
[   72.213206] [<c00109e8>] (arch_cpu_idle) from [<c00569b0>] (default_idle_call+0x34/0x48)
[   72.213233] [<c00569b0>] (default_idle_call) from [<c0056b70>] (cpu_startup_entry+0x1ac/0x200)
[   72.213276] [<c0056b70>] (cpu_startup_entry) from [<c05eecd4>] (rest_init+0x80/0x98)
[   72.213320] [<c05eecd4>] (rest_init) from [<c0842ce4>] (start_kernel+0x390/0x408)
[   72.213339] ---[ end trace aa5f118f8db1fe91 ]---
```

Ah, and could you please add commit 0bbc1e244c1ff0c446697ad6c15596ba5d50b0d6 (limit cyclic lite transfers to 32k)? Haven't found the time yet to dig further into that..

notro:
Thanks for giving this a spin.

> Looks like dma_free_coherent now doesn't like to be called with IRQs disabled

[Documentation/DMA-API.txt](https://www.kernel.org/doc/Documentation/DMA-API.txt)

```
void
dma_free_coherent(struct device *dev, size_t size, void *cpu_addr,
         dma_addr_t dma_handle)

Free a region of consistent memory you previously allocated.  dev,
size and dma_handle must all be the same as those passed into
dma_alloc_coherent().  cpu_addr must be the virtual address returned by
the dma_alloc_coherent().

Note that unlike their sibling allocation calls, these routines
may only be called with IRQs enabled.
```

It seems that the documentation has been realised in code with this 4.3 commit: https://github.com/torvalds/linux/commit/6894258eda2f9badc28c878086c0e54bd5b7fb30

How to solve this? I don't know.

> Ah, and could you please add commit 0bbc1e2

Will do.

HiassofT:
> > Ah, and could you please add commit 0bbc1e2
> 
> Will do.

Thanks!

BTW: you can drop the dreq/slave_id commit, the DMA channels are now
properly set up in the DT and this is no longer needed.

pelwell:
If the image support bcm283x and bcm270x then I think the firmware should load a bcm270x DTB by default. There are a few ways to achieve this:

1) If the image support both bcm283x and bcm270x, don't report bcm283x.
2) Have a command-line parameter to suppress bcm283x reporting.
3) Create a bcm270x tag, set it as appropriate, and put the DTB selection logic in the firmware.

Any thoughts?

notro:
> 2) Have a command-line parameter to suppress bcm283x reporting.

This is currently possible with the --dtok parameter. It turns of auto detection.

> 3) Create a bcm270x tag, set it as appropriate, and put the DTB selection logic in the firmware.

I like this one. The tag decides the default dtb.
If both tags are present, the bootloader can check to see which dtb's are available and use that one. If both are present, use 270x.

popcornmix:
Latest version of this PR is not provoking GPU asserts.
Are you still seeing crashes when sitting idle?

notro:
That's good to hear.

I have been running 24 hours now, mostly idle, running an dma instrumented kernel.
During that time I've done 10 big network transfers since that previously have triggered many of my lockups. I did a transfer, then some minutes later it had locked up.

I'm planning to do a regular build this evening and test that.

Here's a list of major dma allocations made during boot: https://gist.github.com/notro/703d51809931024e0100

HiassofT:
I also like the idea with the additional 270x tag, but - if possible -
we should look for a way so that the kernel can also be booted with
older firmware versions.

One way or the other we need some method to distinguish between
270x only / 283x only / both-enabled kernels. When we have that
we could for some transition period drop the 283x tag on both-enabled
kernels in mkknlimg (until people have updated their firmware).

Would that work?

notro:
I have had 2 lockups, now trying without the multiplatform step.

popcornmix:
I've cherry-picked the I2S audio commits from this PR to rpi-4.3.y. See:
http://forum.kodi.tv/showthread.php?tid=231092&pid=2156514#pid2156514

It's not generated any conflicts, but you should find that 11 commits have been dropped from this PR.

notro:
I have been running close to 48 hours now on 2 Pi's without the multiplatform patch.
One has been sitting idle and one has done a network file copy each 10 minutes.

I'll update the PR later today.

pelwell:
To solve the image tagging problem I propose to create two new tags - 270X and 283X (note the capital 'X's):
1) The old '283x' tag will continue to be set to 1 for images built only for ARCH_BCM2835.
2) The new '270X' tag will be set to 1 if built for ARCH_BCM2708/9, otherwise 0.
3) The new '283X' tag will be set to 1 if built for ARCH_BCM2835/6, otherwise 0.

Old firmware will only look for '283x', so it will load bcm283\* dtbs for builds for just ARCH_BCM2385, otherwise it will use the usual bcm270\* dtbs.

New firmware will first look for the new tags, using bcm270\* if 270X == 1, bcm283\* if 283X == 1, falling back to the behaviour of the old firmware based on 283x. As @notro said, if both flags are present it will prefer the bcm270\* firmware if present, but will use the bcm283\* if it has to and can.

Of course, the effect of pr1178 is to remove the distinction between bcm2708 and bcm2835, so 270X will have the same value as 283X in subsequent builds. I don't see that as necessarily a problem, but if we did want to maintain the distinction then we could make the DT_MACHINE_START declarations conditional on the relevant CONFIG_ARCH_ option.

I will give mkknlimg command-line overrides for all tags, just in case.

notro:
**Version 4**

Removed multiplatform patches.
Without multiplatform there's no pressing need to change mkknlimg for the sake of this PR.

popcornmix:
With latest version of PR do you believe it is stable again?
How far off is a Pi2 version? If the answer is not far, then I'd prefer to merge both together.

notro:
> With latest version of PR do you believe it is stable again?

Actually it's hard to say for sure since I don't know why it locked up. Apart from the io map removal, we are only removing code that's not used or move code that shouldn't cause runtime problems.
So the only thing that could I see that could cause problems is the io map removal, and while writing this I realise that enabling uart1 will crash the kernel without that map. ARCH_BCM2835 has different uart1 enabling code.

I think I will remove the io map commit as well, to be absolutely certain that this won't lockup.
So we'll just let this PR deal with removing ATAGS and nothing else.

> How far off is a Pi2 version?

It will just take a couple of days, so I can add it here.

notro:
**Version 5**

Removed 'bcm2708: Remove I/O map' since it can potentially contribute to the lockups.
Removed watchdog patches since they are not strictly necessary now that we don't have the multiplatform patch.
Added ARCH_BCM2709 ATAGS removal.

popcornmix:
Excellent, I'm running this now on Pi2 with no obvious issues. I'll shout if I spot anything broken.
I'll see if I can get this into @MilhouseVH's OpenELEC test build tonight which will bring a reasonable number of testers.

pelwell:
https://github.com/raspberrypi/linux/pull/1195 contains an update to mkknlimg that adds per-platform tags as discussed. This will allow the firmware to make a more informed choice about which dtb file to load.

popcornmix:
@notro have you seen the watchdog module work?
Previously you could (as root):

```
modprobe bcm2708_wdog
touch /dev/watchdog # enable the watchdog
```

and pi would reboot after about 15 seconds (if you didn't keep writing to /dev/watchdog).
I've not succeeded in making bcm2835_wdt behave the same way.

notro:
It works for me (2835 build).

```
~$ dmesg | grep watchdog
[    1.659874] bcm2835-wdt 20100000.watchdog: Broadcom BCM2835 watchdog timer
~$ sudo touch /dev/watchdog
[   61.190099] watchdog watchdog0: watchdog did not stop!
~$ Uncompressing Linux... done, booting the kernel.
```

popcornmix:
Does it work with 2709 build?
That doesn't have WDT built in (only `CONFIG_BCM2835_WDT=m`)

notro:
Have you enabled the watchdog?

```
dtparam=watchdog=on
```

popcornmix:
Ah yes, dtparam method works. It does mean a script can't just (easily) enable the watchdog directly, although perhaps we should enable it by default in the device tree.

BTW: I've pushed rpi-4.3.y_rebase tree which is identical to rpi-4.3.y with this PR but without the "add loads of platform devices/ATAGS stuff then remove the platform devices/ATAGS stuff" commits.

Reduces the patches needed considerably.

notro:
> It does mean a script can't just (easily) enable the watchdog directly, although perhaps we should enable it by default in the device tree.

Yes, we might as well do that since that will be the case when we switch to 2835.
At the same time the module/driver can be made builtin since it will always load with the device present.

> Reduces the patches needed considerably.

It's music to my ears :-)

notro:
> rpi-4.3.y_rebase tree

**squash: include ARCH_BCM2708 / ARCH_BCM2709**
No need to patch drivers/mailbox/mailbox.c now that we've dropped ATAGS.

**Main bcm2708/bcm2709 linux port**
arch/arm/mach-bcm2708/include/mach/entry-macro.S is not used anymore with MULTI_IRQ_HANDLER enabled. @pelwell moved that code to armctrl.c (which we now have dropped).

popcornmix:
Okay, removed those.

popcornmix:
ping @ujfalusi for thoughts on https://github.com/raspberrypi/linux/pull/1178#issuecomment-154164418
The bcm2835_dma_desc_free call you added seems to be called from interrupt context which isn't allowed. Any ideas of how to avoid this?

HiassofT:
> The bcm2835_dma_desc_free call you added seems to be called from interrupt context which isn't allowed. Any ideas of how to avoid this?

I had a look at the code and noticed two things:

In bcm2835_dma_terminate_all bcm2835_dma_desc_free is called inside
a spin_lock_irqsave block.

https://github.com/notro/linux/blob/multi2708/drivers/dma/bcm2835-dma.c#L656

That would be easy to fix but won't help much alone since
bcm2835_dma_terminate_all is already called with IRQs disabled.

I went up the call-tree and noticed that IRQs are disabled 
quite early in the alsa code in snd_pcm_period_elapsed.

https://github.com/notro/linux/blob/multi2708/sound/core/pcm_lib.c#L1881

No idea how or where to properly fix this...

ujfalusi:
Hi,

On Sat, Nov 14, 2015 at 3:15 PM, popcornmix notifications@github.com
wrote:

> ping @ujfalusi https://github.com/ujfalusi for thoughts on #1178
> (comment)
> https://github.com/raspberrypi/linux/pull/1178#issuecomment-154164418
> The bcm2835_dma_desc_free call you added seems to be called from interrupt
> context which isn't allowed. Any ideas of how to avoid this?
> 
> Yes, I did added the bcm2835_dma_desc_free() to bcm2835_dma_terminate_all()
> in commit f93178291712772983845700b12fc1c8b32f2eb1
> It supposed to fix a memory leak, but I have missed the fact that there is
> a dma_free_coherent().
> The memleak is there w/o my patch and with my patch dma_free_coherent()
> will cause problem.
> Looking at the code it might worth to try to convert the driver to use
> dma_pool_create()/dma_pool_zalloc()/dma_pool_free() ?
> You could revert my patch as well, but then you will leak sizeof(struct
> bcm2835_dma_cb) \* nr_periods every audio start/stop cycle...

## 

Péter

notro:
I came to the same conclusion that a DMA pool could solve this, because [dma_pool_free()](http://lxr.free-electrons.com/ident?i=dma_pool_free) doesn't free the dma allocation:

```
 void dma_pool_free(struct dma_pool *pool, void *vaddr, dma_addr_t dma)
{
...
        /*
         * Resist a temptation to do
         *    if (!is_page_busy(page)) pool_free_page(pool, page);
         * Better have a few empty pages hang around.
         */
...
}
```

Some dma drivers that use dma_pool_free in their .desc_free function:
http://lxr.free-electrons.com/ident?i=zx_dma_free_desc
http://lxr.free-electrons.com/ident?i=pxad_free_desc
http://lxr.free-electrons.com/ident?i=pl08x_desc_free
http://lxr.free-electrons.com/ident?i=mdc_list_desc_free
http://lxr.free-electrons.com/ident?i=idma64_vdesc_free
http://lxr.free-electrons.com/ident?i=fsl_edma_free_desc
http://lxr.free-electrons.com/ident?i=jz4780_dma_desc_free
http://lxr.free-electrons.com/ident?i=sun6i_dma_free_desc

ujfalusi:
Hi,
Can you test the following patch:
http://lists.infradead.org/pipermail/linux-arm-kernel/2015-November/385073.html

Unfortunately I do not have access to Raspberry Pi, but in theory this should fix the bcm2835_dma. FWIW it compiles cleanly ;)

Thanks,
Peter

pelwell:
I've just pushed my mkknlimg changes, so my earlier mkknlimg patch (https://github.com/notro/linux/commit/e44f65e62027a1f58aa7333df4d62d3914ffdddd) is empty and must be dropped.

popcornmix:
@ujfalusi tthanks for patch. There are a few compile errors: http://pastebin.com/qj3tV1N1
error_cb label is in a different function to goto call, and control_block_size, control_block_base, control_block_base_phys are not defined.

HiassofT:
Thanks a lot, Peter, from a very quick test your patch seems to be working fine!

I didn't have much time so I applied the patch on top of this PR, manually resolved the conflicts and changed the slave sg code as well.

Here's a gist with the full bcm2835-dma.c file and the diff.
https://gist.github.com/HiassofT/58ba3b68d27983c0c0be

@notro, @popcornmix could you have a look at that?

ujfalusi:
On Mon, Nov 16, 2015 at 3:23 PM, popcornmix notifications@github.com
wrote:

> @ujfalusi https://github.com/ujfalusi tthanks for patch. There are a
> few compile errors: http://pastebin.com/qj3tV1N1
> error_cb label is in a different function to goto call, and
> control_block_size, control_block_base, control_block_base_phys are not
> defined.
> 
> Interesting. The patch builds fine on top of linux-next and 4.4-rc1

## 

Péter

popcornmix:
@HiassofT your patch does build for me and I could play audio on an iqaudio dacplus without errors in dmesg log.
@ujfalusi Can you check @HiassofT's patch looks okay?

HiassofT:
> Interesting. The patch builds fine on top of linux-next and 4.4-rc1

No problem with your patch. This RPi tree contains local changes
and enhancements to bcm2835-dma which aren't upstreamed yet, so
conflicts are to be expected :)

One of these enhancements is SG DMA support, which is used by the
MMC driver, so any possible bugs should show up pretty quickly.

popcornmix:
@HiassofT I've got a number of 
`[  250.889312] pcm512x 1-004c: No SCLK, using BCLK: -2`
messages in dmesg. Is that expected?

ujfalusi:
On Mon, Nov 16, 2015 at 3:45 PM, popcornmix notifications@github.com
wrote:

> @HiassofT https://github.com/HiassofT your patch does build for me and
> I could play audio on an iqaudio dacplus without errors in dmesg log.
> @ujfalusi https://github.com/ujfalusi Can you check @HiassofT
> https://github.com/HiassofT's patch looks okay?

The file https://gist.github.com/HiassofT/58ba3b68d27983c0c0be looks good.
I see that you have also changed the bcm2835_dma_prep_slave_sg() there. I'm
sure if something is wrong with the dma_pool usage it will show up rather
fast.
Can someone reply to my patch upstream [1] with Tested-by or Acked-by - or
if I need to change something, so it will make it to 4.4? I did not
intended to break bcm2835_dma...

[1]
http://lists.infradead.org/pipermail/linux-arm-kernel/2015-November/385073.html

Thanks,
Péter

notro:
**Version 6**

Removed mkknlimg patch.

HiassofT:
> @HiassofT I've got a number of 
> `[  250.889312] pcm512x 1-004c: No SCLK, using BCLK: -2`
> messages in dmesg. Is that expected?

I'm getting these as well and if I understand the pcm512x code
correctly they are to be expected in configurations like the
iqaudio card (no local oscillator driving SCLK, pcm512x
clocking from I2S BCLK line driven by bcm2835-i2s instead).

popcornmix:
This PR has been in the last two nights of Milhouse OpenELEC builds without complaints. I've also been running for a few days. It's clearly a good thing so lets merge it.

FYI: I've just pushed a 4.4 kernel that uses the rebased/squashed version of this. As expected the rebasing was considerably quicker than usual with the smaller number of custom patches. If it wasn't for the v4l2 buffer change it would have been trivial.

Many thanks @notro this PR is very much appreciated. Makes the rpi kernel a much cleaner place.

pelwell:
Yes indeed - thank you, @notro.

notro:
You're welcome.

popcornmix:
@notro 

> Removed 'bcm2708: Remove I/O map' since it can potentially contribute to the lockups.

Does the IO map leave mmu pages in a different state? I'm wondering if the lack of an mmu flag like "strongly ordered" may result in mostly working, but unreliably behaviour?

notro:
I really don't know, memory handling at this level is beyond my understanding.
But I didn't experience this lockup problem when I did my ARCH_BCM2835 work, so I'm planning to try rpi-4.0.y and rpi-4.1.y to verify that those are good.
It will be at least a week until I have time to do that.

popcornmix:
> I really don't know, memory handling at this level is beyond my understanding.

Yes, mine too. Perhaps someone who knows more will chime in.
There is a CONFIG_ARM_PTDUMP debug option which could provide some debug info for page tables.

taudac:
Which defconfig shall be used to build rpi-4.3.y for the raspberry pi 2?
`bcm2835_defconfig` or `bcm2709_defconfig`?

ping @notro

pelwell:
Use `bcm2709_defconfig` for now, but once the bcm2836 platform is ready `bcm2836_defconfig` will become the correct choice.

taudac:
@pelwell Thanks.

Shall the rpi2 kernel now be named `kernel.img` or `kernel7.img` as it used to be?

pelwell:
kernel7.img is still correct.

pelwell:
The guide [here](https://www.raspberrypi.org/documentation/linux/kernel/building.md) might be helpful.

clivem:
I have 4 days of uptime without any issues, using 4.3.y ad3300ed3855845edda171525d0c780532c98586, plus dma pool patch from gist above, outputting I2S audio 24/7.

HiassofT:
dma pool patch is running fine here as well, had it on my RPi1+2 during the last week.

I also tested the original dma pool patch with mainline 4.4-rc1 and -rc2, worked fine too, and posted details and code on how to test with mainline to the linux-arm-kernel list:

http://lists.infradead.org/pipermail/linux-arm-kernel/2015-November/387442.html

notro:
I have now tried downstream bcm2835_defconfig on rpi-4.1.y and rpi-4.2.y, both running for 3 days without lockup. Then I tried rpi-4.3.y and got a `page domain fault` during bootup caused by systemd (I have debug on the kernel command line which results in heavy systemd logging) and less that one hour later the Pi was unresponsive.
Since there's no mention of this problem upstream, I'll start disabling downstream drivers to see if I can get to a state that doesn't lockup.

clivem:
As above, I haven't had any issues running with the dma pool patch on Pi2B, but with a 2708 build and A+ I am getting kernel hard locks......
[snd_pcm_kernel_panic.txt](https://github.com/raspberrypi/linux/files/50842/snd_pcm_kernel_panic.txt)

clivem:
rpi-4.3.y bcm2708 on A+ WITHOUT dma pool patch......
[bcm2835_dma_desc_free_without_dma_pool.txt](https://github.com/raspberrypi/linux/files/50931/bcm2835_dma_desc_free_without_dma_pool.txt)

notro:
dwc_otg seems to be the reason for lockups. If I switch to dwc2 it doesn't lockup (44 hours runtime).
Disabling fiq with dwc_otg also seems to work around the problem (21 hours runtime): `dwc_otg.fiq_enable=0 dwc_otg.fiq_fsm_enable=0`
The lockup usually happens after I do a network file copy a couple of hours after boot.
I will return to the 2836 work and see how this behaves on the Pi2.

The page failure problem was related to CONFIG_CPU_SW_DOMAIN_PAN being enabled. If only I had remembered this sooner that @popcornmix mentioned this back in october.
A recent [discussion](https://lkml.org/lkml/2015/12/3/86) concludes that CONFIG_CPU_SW_DOMAIN_PAN and CONFIG_UACCESS_WITH_MEMCPY are currently incompatible:

```
> > There's a few ways to solve this:
> >
> > 1. Make the thread switching code save and restore the DACR register as
> >    it would do for domains.  This imposes an overhead on every single
> >    context switch whether or not we happen to be in this _single_
> >    troublesome code.  (Patch attached - as there's several, I'm attaching
> >    them.)
> >
> > 2. Add additional code to the uaccess-with-memcpy stuff to reset the
> >    DACR value prior to using memcpy() or memset().  (Patch attached.)
> >
> > 3. Make uaccess-with-memcpy depend on !CPU_SW_DOMAINS_PAN (suggested by
> >    Will)
> >
> > 4. Delete the uaccess-with-memcpy code (also suggested by Will.)
> >
> > I think the best thing I can do is say... "Discuss amongst yourselves" :)
>
> Personally, I'd advocate for #2 or #4.  Prior commit 0f64b247e6 I was
> already leaning towards #4.
>
> So if some people are still relying on uaccess-with-memcpy and #2 fixes
> it then it's all good.  I'd suggest surrounding the DACR accesses with
> #ifdef CONFIG_CPU_SW_DOMAIN_PAN in the final patch.

Last time this was discussed, people were unhappy about removing it
as they were seeing performance advantages with it enabled.  Of course,
that was with it being buggy.

I think at this point, short of deleting it, I'd opt for (2) so the
overhead is attributed to the appropriate place, and not spread across
the entire system.  That should then be the prelude for another round
of performance evaluation to see whether it is still advantageous to
have the uaccess-with-memcpy code before making a final decision whether
to revert (2) and apply (3) instead, or to go with (4).
```

popcornmix:
I believe uaccess-with-memcpy was quite beneficial, so removing it is not something I'm keen on. Lets see what they decide upstream.

@pelwell has a patch to use the upstream interrupt controller with 2836. It seems good to me except fiq is not supported yet. If we do get fiq working with that, I wonder if the dwc unreliability will magically resolve?

notro:
I have fiq support with 2836: [irqchip: irq-bcm2835: Add 2836 FIQ support](https://github.com/notro/linux/commit/01d0ea76de50580100ebceaf4676a2d858b08e84)
https://github.com/notro/linux/commits/bcm2836

popcornmix:
I tried Phil's upstream interrupt controller patches and your fiq patch:
http://paste.ubuntu.com/13940188/

but dwc_otg fails with fiq enabled. It does work with fiq disabled.
Have you had upstream interrupt controller and fiq working with dwc_otg?

notro:
Yes I have 2836 running now with fiq enabled (a bit unreliable network transfer though).
You seem to be missing the syscon DT node which the irq driver needs:

```
    syscon@40000000 {
           compatible = "brcm,bcm2836-arm-local", "syscon";
           reg = <0x40000000 0x100>;
    };
```

I use Eric's [timer patch](https://github.com/notro/linux/commit/b1ca75fe92ad3e4de88dbc239c0b130f82db6a91) and it enables syscon (I have removed the conditional here):

``` diff
diff --git a/arch/arm/mach-bcm/Kconfig b/arch/arm/mach-bcm/Kconfig
index edbe6fe..3d76e13 100644
--- a/arch/arm/mach-bcm/Kconfig
+++ b/arch/arm/mach-bcm/Kconfig
@@ -115,6 +115,7 @@ config ARCH_BCM2835
    select ARM_TIMER_SP804
    select CLKSRC_OF
    select FIQ
+   select MFD_SYSCON
    select PINCTRL
    select PINCTRL_BCM2835
    help
```

Looking at your patch I can see why I have problems with mmc dma on 2836.

notro:
And I can see that @pelwell discovered the reason for the spurious interrupts, that's good!

notro:
I see that bcm2709_defconfig has CONFIG_CACHE_L2X0 disabled. Why is that? ARCH_BCM2709 does select [MIGHT_HAVE_CACHE_L2X0](https://github.com/raspberrypi/linux/blob/rpi-4.3.y/arch/arm/mm/Kconfig#L906) in [arch/arm/Kconfig](https://github.com/raspberrypi/linux/blob/rpi-4.3.y/arch/arm/Kconfig)
bcm2835_defconfig also selects MIGHT_HAVE_CACHE_L2X0 through ARCH_MULTI_V6_V7, so it is enabled.
I'm wondering what to do in a combined Pi1/Pi2 kernel image and I'm having stability issues in my 2836 kernel on Pi2.

popcornmix:
Pretty sure we don't want CONFIG_CACHE_L2X0. That is an outer L2 cache (external to the arm cores). We have an inner L2 cache (integrated with the cores).

notro:
Is it ok to enable CONFIG_NEON on Pi1?

diff -u arch/arm/configs/bcmrpi_defconfig arch/arm/configs/bcm2709_defconfig

``` diff
+# CONFIG_CACHE_L2X0 is not set
+CONFIG_SMP=y
+CONFIG_HAVE_ARM_ARCH_TIMER=y
+CONFIG_VMSPLIT_2G=y
+CONFIG_NEON=y
+CONFIG_KERNEL_MODE_NEON=y
-CONFIG_LATENCYTOP=y
-CONFIG_CRYPTO_CRYPTD=m
-CONFIG_CRYPTO_SHA512=m
-CONFIG_CRYPTO_SHA1_ARM=m
-CONFIG_CRYPTO_AES_ARM=m
+CONFIG_CRYPTO_SHA1_ARM_NEON=m
+CONFIG_CRYPTO_AES_ARM_BS=m
```

popcornmix:
There is no Neon support on pi1 so it's probably not good to enable it (unless it gets disabled in some way).

popcornmix:
@notro have you had fiq + upstream interrupt controller working with bcm2709_defconfig platform?

notro:
Yes, this is what I used: https://github.com/notro/linux/commits/irq2836-43
Looking at the patches I see that I adjust the irq numbering like @pelwell did on 2708.
If you're not trying to get the irq drivers to co-exist, try adding `select SPARSE_IRQ` instead of using arch_dynirq_lower_bound().

notro:
Yes, SPARSE_IRQ is what I used in the drop atags patchset: https://github.com/raspberrypi/linux/commit/9b4b8da51ac4982bf8a5ebe5b288cfbc9560dc79

popcornmix:
I want to drop downstream interrupt controller in 4.4 (which has drop atags).
https://github.com/raspberrypi/linux/tree/rpi-4.4.y_irq has Phil's upstream irq patch plus your FIQ patch. Can you see anything missing?

notro:
I guess you have this error in the kernel log: Failed to get local register map. FIQ is disabled for cpus > 1
Because you seem to missing the syscon node: https://github.com/notro/linux/blob/ceac641527b8eec687a81c0d44d5bb29f0e15693/arch/arm/boot/dts/bcm2709.dtsi#L45

I see that @anholt doesn't use a syscon node in his 2836 v2 patch like he did in his dev branch that I looked at.
irq-bcm2836 claims the reg address we need, so I'm not sure what the best solution is here.

popcornmix:
Yes, I get that error in kernel log.

notro:
It should work if you add that node, but I'm not sure if this is the approved way of doing it.

popcornmix:
I'll give it a try. For now I'll settle for working. Approved can come later.
Do I need: https://github.com/notro/linux/commit/ceac641527b8eec687a81c0d44d5bb29f0e15693#diff-68483e7094a0b0a646c8462503f8262eR21

notro:
If all the cpu's come up, I guess you're ok, but there's some smp code in irq-bcm2836 and I don't know how that plays with the current 2709 smp code:

``` c
static void
bcm2836_arm_irqchip_smp_init(void)
{
#ifdef CONFIG_SMP
    /* Unmask IPIs to the boot CPU. */
    bcm2836_arm_irqchip_cpu_notify(&bcm2836_arm_irqchip_cpu_notifier,
                       CPU_STARTING,
                       (void *)smp_processor_id());
    register_cpu_notifier(&bcm2836_arm_irqchip_cpu_notifier);

    set_smp_cross_call(bcm2836_arm_irqchip_send_ipi);
#endif
}
```

Maybe it just does the same operation twice, I haven't studied it.

notro:
> Maybe it just does the same operation twice, I haven't studied it.

I mean: maybe irq2836 and 2709 combined will so some smp operations twice.

popcornmix:
Adding https://github.com/raspberrypi/linux/commit/0d3f1f5961a80a8469e2aabf4ecec8763c9826cf doesn't affect the "Failed to get local register map. FIQ is disabled for cpus > 1" error.
I feel I need something else to use that node. Do I need CONFIG_MFD_SYSCON? Do I need something like: https://github.com/notro/linux/commit/ceac641527b8eec687a81c0d44d5bb29f0e15693#diff-68483e7094a0b0a646c8462503f8262eR104 ?

popcornmix:
CONFIG_MFD_SYSCON does seem to make the fiq work. Just need to test for stability.

popcornmix:
Upstream interrupt controller was in last night's milhouse OpenELEC build without complaints.
rpi-4.4.y tree is now -rc6 and using upstream interrupt controller and seems good.
Thanks @notro and @pelwell

PR #1183, opened 2015-11-01
Build i2c_gpio module and add a device tree overlay to configure it.

ali1234:
This allows to use the software/bit-banging i2c driver on Raspberry Pi GPIOs. GPIO pins and clock speed can be selected at boot time with overlay params.

This is useful for talking to I2C devices which need I2C_M_NOSTART or clock stretching.


pelwell:
Thanks - this looks like it could be helpful for some users. It isn't so easy for another overlay to make use of it because of the way labels defined within overlays aren't exported (there is a workaround that involves targetting a fragment at the `/__symbols__` node), but I can revisit this when I expand the overlay functionality.

popcornmix:
No objections.

PR #1195, opened 2015-11-12
scripts: Multi-platform support for mkknlimg and knlinfo

pelwell:
The firmware uses tags in the kernel trailer to choose which dtb file
to load. Current firmware loads bcm2835-*.dtb if the '283x' tag is true,
otherwise it loads bcm270*.dtb. This scheme breaks if an image supports
multiple platforms.

This patch adds '270X' and '283X' tags to indicate support for RPi and
upstream platforms, respectively. '283x' (note lower case 'x') is left
for old firmware, and is only set if the image only supports upstream
builds.


lurch:
I've got no idea what the '--ddtk' flag is for, but does it need to be added to the Usage string?

pelwell:
It does nothing useful yet, and documenting it will only lead to more questions...

popcornmix:
4.3 is dead and this seems to be present in 4.4 so closing.

Issue #1196, opened 2015-11-12
  bcm2835-v4l2 works incorrectly without preview


pbugalski:
This issue was discovered as a result of this discussion: https://www.raspberrypi.org/forums/viewtopic.php?f=43&t=62364&p=813478
Generally, if you capture video with preview on the screen module works fine. Unfortunately, if you turn off preview, white balance and exposition parameters are ignored and set to automatic.
To reproduce run:
v4l2-ctl -v width=1024,height=768,pixelformat=I420
v4l2-ctl --set-ctrl=white_balance_auto_preset=0,red_balance=5000,blue_balance=1500
v4l2-ctl --set-ctrl=brightness=55,contrast=15,saturation=0,horizontal_flip=1,vertical_flip=1,power_line_frequency=1,sharpness=0,color_effects=0,rotate=0,color_effects_cbcr=32896
v4l2-ctl --set-ctrl=auto_exposure=1,exposure_time_absolute=300,auto_exposure_bias=12,iso_sensitivity=1
v4l2-ctl -p 10
v4l2-ctl --stream-mmap=3 --stream-skip=2 --stream-to=./test.yuv --stream-count=4

As a result video will be captured with automatic white-balance (instead of reddish). If you turn on overlay, result will be correct. 
Also system log says that something goes wrong without overlay (preview):

pbugalski:
[bcm2835-v4l.txt](https://github.com/raspberrypi/linux/files/32910/bcm2835-v4l.txt)

pbugalski:
The reason of problem is described in RaspiCam Documentation (https://github.com/raspberrypi/userland/blob/master/host_applications/linux/apps/raspicam/RaspiCamDocs.odt):
"The preview display is optional, but can be used full screen or directed to a specific rectangular area on the display. If preview is disabled, the null_sink component is used to 'absorb' the preview frames. It is necessary for the camera to produce preview frames even if not required for display, as they are used for calculating exposure and white balance settings."

I have patch for bcm2835-v4l, which solves this issue.

popcornmix:
Ping @6by9

6by9:
You've quoted a stack dump, but without the line immediately before it which will tell you what the warning was.
It looks identical to https://github.com/raspberrypi/linux/issues/817 where an example stack for that error is given, complete with the 
[...]

PR #1197, opened 2015-11-12
bcm2835-v4l2 works incorrectly without preview

pbugalski:
Redirect preview from camera to null_sink if no preview is required. It is required by camera for calculating exposure and white balance settings.


6by9:
Please hold off for now. This is a workaround for a firmware bug that I will hopefully find time to address in the next week.

pbugalski:
Sure. The issue discussion is here: https://github.com/raspberrypi/linux/issues/1196
I still think workaround is better idea than malfunction, but I can hold off. This workaround is from raspistill/raspivid, so it isn't any new idea.

6by9:
Firmware is fixed, so this PR is not required now. Please close if you agree.

PR #1199, opened 2015-11-14
dts: Added overlay for gpio_ir_recv driver

hsteinhaus:
This PR adds a DTS that allows the use of the gpio_ir_recv kernel driver.

The gpio_ir_recv  driver is a rc-core based driver for TSOP-style IR receivers connected to a GPIO pin and is part of the Linux kernel. In contrast to lirc_rpi, no LIRC is needed for IR signal decoding or RC key mapping. These tasks are done by proven kernel infrastructure instead.

In contrast to the LIRC/lirc_rpi approach, the gpio_ir_recv driver brings a better RC responsiveness and a more consistent timing between RC button pressed and and key event and does not suffer from the typical double-key problems often reported from the lirc_rpi/LIRCd approach.

### DIFF
--- a/arch/arm/boot/dts/overlays/Makefile
+++ b/arch/arm/boot/dts/overlays/Makefile
@@ -17,6 +17,7 @@ dtb-$(RPI_DT_OVERLAYS) += at86rf233-overlay.dtb
 dtb-$(RPI_DT_OVERLAYS) += bmp085_i2c-sensor-overlay.dtb
 dtb-$(RPI_DT_OVERLAYS) += dht11-overlay.dtb
 dtb-$(RPI_DT_OVERLAYS) += enc28j60-overlay.dtb
+dtb-$(RPI_DT_OVERLAYS) += gpio-ir-overlay.dtb
 dtb-$(RPI_DT_OVERLAYS) += gpio-poweroff-overlay.dtb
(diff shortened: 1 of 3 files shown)


popcornmix:
Looks reasonable to me. @pelwell ?

pelwell:
It looks OK to me (I'm away and can't test it though).

Just FYI the strings "off", "down" and "up" translate to the values 0, 1 and 2 when assigning to an integer dtparam.

hsteinhaus:
> Just FYI the strings "off", "down" and "up" translate to the values 0, 1 and 2 when assigning to an integer dtparam.

Wow, this devicetree stuff is really magic ;)

I will change the comments and doc accordingly in a follow-up commit shortly.

popcornmix:
Can you squash the commits together and force push? (`git rebase -i HEAD~2` and `git push -f`).

pelwell:
Thanks for squashing - we don't receive notifications for a push, hence the slow response.

My automated checker spotted that a parameter name was incorrect in the README, and while I was correcting that I spotted the absence of "dtoverlay=" in the Load: section, so I've patched it up and pushed the new commit. I've also improved the checker and corrected some similar issues elsewhere in the README - I can see where you copied it from.

Thanks for the contribution.

PR #1206, opened 2015-11-20
SDIO-overlay: add poll_once-boolean parameter

pboettch:
Add paramter to toggle sdio-device-polling done every second or once at boot-time in the sdio-overlay.dts .

Can also be applied to other branches of more recent versions.


pelwell:
Thanks. I've pushed this into rpi-4.1.y, rather than rpi-4.1.y_rebase.

PR #1213, opened 2015-12-04
VC4 backwards-compatible code changes

anholt:
I've now submitted the 3D driver upstream: http://lists.freedesktop.org/archives/dri-devel/2015-December/095834.html

This is a merge to get downstream closer to upstream, but without requiring ABI or DT changes.

Remaining differences between upstream and downstream:
- Upstream has MSAA support (requires updated Mesa packages, changes for MSAA haven't landed in master yet).
- Upstream changed "hpd-gpio" to "hpd-gpios".  It would be nice to sync that before our DT gets too settled.
- Upstream is using generic PM domains instead of firmware calls from vc4.  I still need to test GPU reset with the new code.
- Upstream doesn't have the "wait for rendering to finish" out-of-memory fallback yet.

I'm currently working on landing the MSAA bits, and I'll try to get a branch together for rebuilding the .debs.


popcornmix:
Builds and runs okay for me. @pelwell happy to merge?

pelwell:
I'm happy for you to change hpd-gpio to hpd-gpios whenever you want. Although it isn't really the done thing, our DT has been through several incompatible changes, but because the kernel is always packaged with the DT and overlays this hasn't been a problem. Just put the overlay change and any code change in the same commit.

PR #1228, opened 2015-12-15
VC4: Update to the stable ABI.

anholt:
The vc4 3D support is now merged to drm-next, and this updates the downstream driver to match the ABI that's landing in 4.5.   Pulling this requires an updated Mesa build, either git master or the 11.1 release that just happened.  I have a further change in the 11.1-vc4 branch (https://github.com/anholt/mesa/tree/11.1-vc4) that should improve the stability of X, which should land in 11.1.1.  There's also an update to the X Server backport tree (http://cgit.freedesktop.org/~anholt/xserver/log/?h=raspbian-backport) that fixes a massive memory leak in my glamor-delay-shareable work in the previous backport build.

### DIFF
--- a/arch/arm/boot/dts/overlays/vc4-kms-v3d-overlay.dts
+++ b/arch/arm/boot/dts/overlays/vc4-kms-v3d-overlay.dts
@@ -68,7 +68,7 @@
 				      <0x7e808000 0x100>;
 				interrupts = <2 8>, <2 9>;
 				ddc = <&i2c2>;
-				hpd-gpio = <&gpio 46 GPIO_ACTIVE_HIGH>;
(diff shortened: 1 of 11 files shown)


popcornmix:
Should this work with just a kernel update? I don't have an easy way of bumping Mesa (until @spl237 makes a new package). Just using the updated kernel fails to start X:
`[   45.175103] [drm:vc4_use_bo [vc4]] *ERROR* Trying to use shader BO as something other than a shader`

anholt:
Nope, it really does require the updated Mesa.

popcornmix:
Oh well, I'll trust it works. I have confirmed it doesn't break without the dtoverlay enabled.

Issue #1212, opened 2015-12-03
RPi zero Support for USB Gadget Mode (pi shows up as keyboard, disk, net, camera etc)


hh:
Having a $5 usb gadget that could function as a combination of anything in https://github.com/torvalds/linux/tree/master/drivers/usb/gadget/function would be pretty grand.

Think about your phone, how it works to connect a micro-B to charge and  connect to it as a gadget to your computer, then when you connect a micro-A cable you plug mice, keyboard, and thumb drives into your phone.

As a usb host, you can plug a phone into it (just like you do your dekstop or laptop) and flash it. Format a usb drive, or bootstrap embedded devices like intel edison.

As a usb gadet, you can plug it into a computer and have it show up as a usb keyboard, disk, network, and other interesting things. (send some keystrokes to hit F12, and boot to a usb installation disk). 

I'm looking to see if we can't get the RPi zero to function as both.

It looks like it may have worked at one point: https://github.com/raspberrypi/linux/issues/881#issuecomment-161411866

I just wanted to gauge the interest.

imbens:
That would also allow you to code on the Pi Zero with only a single cable (for power and network) to a PC.

harjoc:
Apparently the data pins on the port used for power aren't connnected to the SoC:

http://raspberrypi.stackexchange.com/questions/8587/can-you-transfer-data-through-the-microusb

hh:
I'm not talking about using the port that is for power only. 

This is specifically for the second usb micro port on the RPi zero that does have the data pins connected (you usually use an OTG cable to connect to it in host mode). I'm wanting to use that same port in gadget mode.

hh:
It looks like the RPi can be powered over that second port (the one we want to use in gadget mode). So @imbens idea for allowing us to code with a single cable seems valid.

hh:
For most OTG supported ports, it depends on what you plug into it that decides if the port is in host mode or gadget mode. We just need to get confirmation that the RPi zero port is wired the same way for the port that has it's usb data pins connected (not the one dedicated to power).

["A device with a micro-A plug inserted becomes an OTG A-device, and a device with a micro-B plug inserted becomes a B-device. The type of plug inserted is detected by the state of the pin ID ."](https://en.wikipedia.org/wiki/USB_On-The-Go#OTG_micro_plugs) 

[...]

PR #1239, opened 2015-12-26
rpi-4.4.y: Add USB gadget support to Pi Zero

notro:
Build the dwc2 usb driver as a module to be able to use the USB On-the-Go functionality (supports both host and device/peripheral mode).
Add a specific Device Tree file for Pi Zero with dwc2 enabled in otg mode. Also fix the inverted ACT led.

This requires a change in the bootloader to give the Zero a dtb of it's own. Currently it uses this:

```
~$ sudo vcdbg log msg
001289.085: Loading 'bcm2708-rpi-b-plus.dtb' from SD card
```

This PR can be tested by setting the dtb manually in /boot/config.txt:

```
device_tree=bcm2708-rpi-zero.dtb
```

**Testing**

This will give a serial console on a connected computer:

```
sudo modprobe g_serial
sudo systemctl start getty@ttyGS0.service
```

On my Windows 8 computer, the USB serial port shows up as: ELMO GMAS (COM5)

Hotswapping to a keyboard works.

**Controller fifo**

The fifo assignment isn't optimal. This PR uses only 1024 of the 4080 available dynamic fifo entries:

```
[    8.779556] dwc2 20980000.usb: EPs: 8, dedicated fifos, 4080 entries in SPRAM
```

Currently assigned in Device Tree:

```
&usb {
    g-np-tx-fifo-size = <32>;
    g-rx-fifo-size = <256>;
    g-tx-fifo-size = <256 128 128 64 64 64 32>;
};
```

I don't know how to best distribute the remaining entries.

**Issue**

When I disconnect the keyboard, I get a warning:

```
[   93.639089] g_serial gadget: Gadget Serial v2.4
[   93.639120] g_serial gadget: g_serial ready
[   93.642186] dwc2 20980000.usb: bound driver g_serial
<connect computer>
[  101.877358] dwc2 20980000.usb: new device is high-speed
[  101.893135] dwc2 20980000.usb: new address 5
[  101.908600] g_serial gadget: high-speed config #2: CDC ACM config
<disconnect computer and connect keyboard>
[  158.138483] usb 1-1: new low-speed USB device number 2 using dwc2
[  158.355959] usb 1-1: New USB device found, idVendor=046d, idProduct=c313
[  158.356001] usb 1-1: New USB device strings: Mfr=1, Product=2, SerialNumber=0
[  158.356019] usb 1-1: Product: USB Multimedia Keyboard
[  158.356036] usb 1-1: Manufacturer: LITEON Technology
[  158.378340] input: LITEON Technology USB Multimedia Keyboard as /devices/platform/soc/20980000.usb/usb1/1-1/1-1:1.0/0003:046D:C313.0001/input/input0
[  158.441305] hid-generic 0003:046D:C313.0001: input,hidraw0: USB HID v1.10 Keyboard [LITEON Technology USB Multimedia Keyboard] on usb-20980000.usb-1/input0
[  158.454182] input: LITEON Technology USB Multimedia Keyboard as /devices/platform/soc/20980000.usb/usb1/1-1/1-1:1.1/0003:046D:C313.0002/input/input1
[  158.509498] hid-generic 0003:046D:C313.0002: input,hidraw1: USB HID v1.10 Device [LITEON Technology USB Multimedia Keyboard] on usb-20980000.usb-1/input1
<disconnect keyboard>
[  186.039220] usb 1-1: USB disconnect, device number 2
[  186.219073] ------------[ cut here ]------------
[  186.219241] WARNING: CPU: 0 PID: 67 at drivers/usb/dwc2/gadget.c:176 dwc2_hsotg_init_fifo+0x18c/0x1ac [dwc2]()
[  186.219258] Modules linked in: evdev usb_f_acm u_serial g_serial libcomposite cfg80211 rfkill dwc2 udc_core bcm2835_gpiomem bcm2835_wdt uio_pdrv_genirq uio i2c_dev snd_bcm2835 snd_pcm snd_timer snd fuse
[  186.219359] CPU: 0 PID: 67 Comm: kworker/u2:2 Not tainted 4.4.0-rc6+ #2
[  186.219371] Hardware name: BCM2708
[  186.219435] Workqueue: dwc2 dwc2_conn_id_status_change [dwc2]
[  186.219508] [<c0016ccc>] (unwind_backtrace) from [<c0013b3c>] (show_stack+0x20/0x24)
[  186.219548] [<c0013b3c>] (show_stack) from [<c032682c>] (dump_stack+0x20/0x28)
[  186.219584] [<c032682c>] (dump_stack) from [<c00222c8>] (warn_slowpath_common+0x8c/0xc4)
[  186.219613] [<c00222c8>] (warn_slowpath_common) from [<c00223bc>] (warn_slowpath_null+0x2c/0x34)
[  186.219697] [<c00223bc>] (warn_slowpath_null) from [<bf0ae6ac>] (dwc2_hsotg_init_fifo+0x18c/0x1ac [dwc2])
[  186.219831] [<bf0ae6ac>] (dwc2_hsotg_init_fifo [dwc2]) from [<bf0b04e4>] (dwc2_hsotg_core_init_disconnected+0x60/0x35c [dwc2])
[  186.219937] [<bf0b04e4>] (dwc2_hsotg_core_init_disconnected [dwc2]) from [<bf0a6824>] (dwc2_conn_id_status_change+0xe8/0x21c [dwc2])
[  186.220001] [<bf0a6824>] (dwc2_conn_id_status_change [dwc2]) from [<c0039c70>] (process_one_work+0x13c/0x490)
[  186.220029] [<c0039c70>] (process_one_work) from [<c003a13c>] (worker_thread+0x178/0x538)
[  186.220064] [<c003a13c>] (worker_thread) from [<c003fee8>] (kthread+0xe0/0xfc)
[  186.220097] [<c003fee8>] (kthread) from [<c000f788>] (ret_from_fork+0x14/0x2c)
[  186.220114] ---[ end trace 900ecfdec3d84a77 ]---
<connect computer>
[  189.792998] dwc2 20980000.usb: new device is high-speed
[  189.808759] dwc2 20980000.usb: new address 13
[  189.831151] g_serial gadget: high-speed config #2: CDC ACM config

```

**Failed on 4.1**

I also tried to do this PR on rpi-4.1.y but it didn't work and I didn't try to find out why. Got this error:

```
[    7.463269] dwc2 20980000.usb: no platform data or transceiver defined
```

This PR is based on work done by several people in #1212


notro:
I only get the warning on keyboard disconnect if I have g_serial loaded.

popcornmix:
Thanks. No objections. @pelwell ?

I did mention the inverted ACT led shortly before launch, but I think Mike liked the inverted, mostly on behaviour due to the absence of a PWR led, so it wasn't fixed then. I still think fixing it is probably the right thing to do, but we'll see what @pelwell and others think.

MrEngman:
The ACT led is great how it works at the moment. Ideal as there is no PWR led. Please don't change it.

notro:
> The ACT led is great how it works at the moment. Ideal as there is no PWR led.

In what way ideal, that it works excatly as a PWR led? The trigger can be changed to `default-on` to mimick a power led.

The problem with the inverted behaviour is that trigger `none` turns the led on and `default-on` turns it off. `heartbeat` gives a strange pulsing, and `mmc0` is steady on apart from small "inverted blinks" when there's activity instead of a blink. `cpu0` is also strange like mmc0.

There's a DT overlay parameter (act_led_activelow) to change the inversion, but I think that the expected behaviour should be the default.

Available triggers:

```
~$ cat /sys/class/leds/led0/trigger
none kbd-scrollock kbd-numlock kbd-capslock kbd-kanalock kbd-shiftlock kbd-altgrlock kbd-ctrllock kbd-altlock kbd-shiftllock kbd-shiftrlock kbd-ctrlllock kbd-ctrlrlock [mmc0] timer oneshot heartbeat backlight gpio cpu0 default-on input
```

Change trigger from boot with overlay parameter: act_led_trigger

Change trigger immediately:

```
~$ echo cpu0 | sudo tee /sys/class/leds/led0/trigger
cpu0
```

MrEngman:
I didn't say it works "exactly" as a PWR led, but it does provide a good indication the Pi zer0 is powered. It appears to default to ON indicating the Pi is powered and flashes off when the SD card is being accessed/written.

My Pi zero sat next to me is idle at the moment and the ACT led is lit showing it is being powered. If it is changed to behave like the earlier Pi's that have a seperate PWR led then it would be off and I would not know if it was powered on or not.

Or is this a case of it ain't broke fix it so it is, like the changes to the wifi made over the last few months.

clivem:
I think the original Zero LED behaviour should stay! I'm more interested in a visual indication that the Zero is being powered than its SDCARD is being accessed. What we have now is a good compromise, only having a single LED available to work with.

notro:
How about this change which will turn the led on, but keep the correct behaviour for the other triggers?

``` diff
 &leds {
    act_led: act {
        label = "led0";
-       linux,default-trigger = "mmc0";
+       linux,default-trigger = "default-on";
        gpios = <&gpio 47 1>;
    };
 };
```

I agree that it's useful to have a power indicator.

Some thoughts on using the act led:
If I were to make a product based around a Zero, I would have turned the led on using dt-blob.bin, indicating that the board is powered (firmware loaded and sdcard funtional).
Then I would have used the trigger "timer" in the dtb causing the led to blink during kernel loading.
At last I would have turned the led steady on when the board was fully funtional.

This way I would now that if the led never lights, there's a power/sdcard problem. If it lights up I know the firmware is loaded.
If it doesn't start blinking, the kernel has failed to load.
If it stays blinking, I know that my app or whatever hasn't started succesfully.
There is also one more diagnostic step, and that would be to switch to the hearbeat pattern as the first thing when userspace loads. To show that systemd has started.
But I have never shipped a product so I don't know how useful it is to pinpoint these different stages for diagnostic purposes.

pelwell:
I have no objections to the Pi Zero dtb. I think the inverted activity LED, although different, is a sensible use of a single LED.

mobluse:
I think this should be tested and working on Pi Model A before it is merged, because maybe there are many A:s that can use this if it was adapted to these.

ED6E0F17:
I increased "g-np-tx-fifo-size" because 16 seemed "a bit small" compared to the legacy code, but from inspection of the code it seems to remain unused. My personal preference is to make "g-rx-fifo-size" as large as possible, and set "g-tx-fifo-size"  based on actual testing.

notro:
Regarding OTG on model A

Have anyone tried setting `dr_mode = "peripheral";` in the Device Tree?
Maybe it's enough, I don't know. If it doesn't work then we need to know which register holds the value of the USB_OTGID pin and see how it's used in the driver (AFAICT this pin is tied low on both A and B).

[drivers/usb/dwc2/core.c#L477](http://lxr.free-electrons.com/source/drivers/usb/dwc2/core.c#L477)

``` c
static int dwc2_core_reset(struct dwc2_hsotg *hsotg)
{
        if (hsotg->dr_mode == USB_DR_MODE_HOST) {
                gusbcfg = readl(hsotg->regs + GUSBCFG);
                gusbcfg &= ~GUSBCFG_FORCEDEVMODE;
                gusbcfg |= GUSBCFG_FORCEHOSTMODE;
                writel(gusbcfg, hsotg->regs + GUSBCFG);
        } else if (hsotg->dr_mode == USB_DR_MODE_PERIPHERAL) {
                gusbcfg = readl(hsotg->regs + GUSBCFG);
                gusbcfg &= ~GUSBCFG_FORCEHOSTMODE;
                gusbcfg |= GUSBCFG_FORCEDEVMODE;
                writel(gusbcfg, hsotg->regs + GUSBCFG);
        } else if (hsotg->dr_mode == USB_DR_MODE_OTG) {
                gusbcfg = readl(hsotg->regs + GUSBCFG);
                gusbcfg &= ~GUSBCFG_FORCEHOSTMODE;
                gusbcfg &= ~GUSBCFG_FORCEDEVMODE;
                writel(gusbcfg, hsotg->regs + GUSBCFG);
        }
```

How should a type A to type A cable be wired to test this? Straight through?

lurch:
> How should a type A to type A cable be wired to test this? Straight through?

I'm still away from home so can't check, but IIRC when I was experimenting with rpiboot on a ModelA, I used a cable like [this](http://www.amazon.co.uk/Cables-Usb-Male-Cable-Black/dp/B001U3ZP0Q) and carefully removed the outer shielding and snipped through the 5V wire.

mobluse:
About Pi A: I tested with an adapter similar to this: http://www.amazon.com/niceEshop-Changer-Adapter-Coupler-Converter/dp/B00BPHBCX6 . It has pin 1 connected to pin 1, 2 to 2, 3 to 3, 4 to 4, shield to shield (one contact is up side down compared to the other). It gives this with dmesg in Ubuntu 14.04 LTS and similar in Raspbian:
[382925.640098] usb 6-1: new full-speed USB device number 4 using uhci_hcd
[382925.806151] usb 6-1: New USB device found, idVendor=0a5c, idProduct=2763
[382925.806163] usb 6-1: New USB device strings: Mfr=1, Product=2, SerialNumber=0
[382925.806171] usb 6-1: Product: BCM2708 Boot
[382925.806178] usb 6-1: Manufacturer: Broadcom

I also looked at the schematics for Pi A and B on page 3 and USB_OTGID/K16 is connected to ground: https://www.raspberrypi.org/documentation/hardware/raspberrypi/schematics/Raspberry-Pi-Rev-2.1-Model-AB-Schematics.pdf

gbaman:
Using a similar adapter to @mobluse, just gave it a go with my A+. Unfortunately not having any luck getting it to appear on attached computer. Have tried this pull request and also tried using a kernel with dwc2 set to Gadget only mode. Both without any luck on fresh Raspbian images.

notro:
Device mode on A+

I haven't made that cable yet and I don't know when I get the time, but I believe I have found a solution to overriding the USB_OTGID pin. So if someone want to test it, here's what I intend to try:
[drivers/usb/dwc2/hcd.h](https://github.com/raspberrypi/linux/blob/rpi-4.4.y/drivers/usb/dwc2/hcd.h#L380):

``` diff
 /*
  * Returns the mode of operation, host or device
  */
 static inline int dwc2_is_host_mode(struct dwc2_hsotg *hsotg)
 {
+   if (hsotg->dr_mode == USB_DR_MODE_HOST)
+       return 1;
+   else if (hsotg->dr_mode == USB_DR_MODE_PERIPHERAL)
+       return 0;
+
    return (readl(hsotg->regs + GINTSTS) & GINTSTS_CURMODE_HOST) != 0;
 }
 static inline int dwc2_is_device_mode(struct dwc2_hsotg *hsotg)
 {
-   return (readl(hsotg->regs + GINTSTS) & GINTSTS_CURMODE_HOST) == 0;
+   return !dwc2_is_host_mode(hsotg)
 }
```

This is also needed in DT: `dr_mode = "peripheral";`

The GINTSTS reg is available in debugfs, host mode is determined by bit 0:

```
# keyboard connected on zero
~$ sudo grep GINTSTS /sys/kernel/debug/20980000.usb/regdump
GINTSTS = 0x04000021

# keyboard disconnected
~$ sudo grep GINTSTS /sys/kernel/debug/20980000.usb/regdump
GINTSTS = 0x04008022
```

popcornmix:
@notro can you rebase and drop the LED commit for now.
Feel free to submit it as separate PR if you are still keen, but no point holding up this PR for it.

notro:
> and drop the LED commit for now.

Yes there was a strong opposition against that, and I do see the advantage as a combined poweron/disk activity indicator. And it's possible to change the polarity with a dt parameter so I have changed my position on this.

@ED6E0F17 mentioned a patch [usb: dwc2: gadget: fix TX FIFO size and address initialization](https://www.spinics.net/lists/linux-usb/msg134539.html) which I had a look at but it wasn't a shallow rabbithole so I dropped that. Some notes I made: https://gist.github.com/notro/1abf4b2d89580d80489a

I'll see if I can rebase later today. I have made a dwc_otg overlay for the Zero, but I'm wondering if I should do a dwc2 overlay as well which can be used as a debug aid with usb problems to rule out dwc_otg as the problem. It can also have parameters to set the fifo properties.

popcornmix:
I could imagine a dwc2 overlay being useful on a Pi2, for debugging USB issues.
That driver may improve with time (perhaps getting upstream fiq support) so may be a more viable option in the future.

popcornmix:
@P33M can you comment on the fifo sizes? Is it something we can (and should) be setting, or is it just read only?

notro:
**Version 2**
- Go back to inverted act led
- Add dwc_otg overlay so the Pi Zero can use the dwc_otg driver in host mode with it's fiq support
- Add dwc2 overlay with parameters for fifo sizes

I won't pursue Device mode on A+. I'm returning to Pi2 and ARCH_BCM2835 for another go.

P33M:
According to the databook, the register for device periodic per-endpoint FIFO size is a fixed quantity with a variable start address. The fixed quantity is read-only and in the top 16 bits of DPTXFSIZn (offset 0x104 + (EPn-1) \* 4). The power-on reset value for the bottom bits is a bit complicated and I'm not sure I can post without disclosing internal core details.

However, these bottom bits are writeable which means you can change the start address for each of these endpoint FIFOs as long as the value written doesn't exceed the value loaded on reset.

As far as I'm concerned, treating the periodic FIFO sizes in device mode as read-only (incl. POR value which implies a contiguous block of SRAM) is the right thing to do.

popcornmix:
@notro we'd like to get this in. Can you rebase?
I wonder whether defaulting to dwc2 on Pi Zero is correct. Currently it's not an issue as you need to choose the pi zero dtb explicitly, but it may make sense to have firmware choose that automatically which would then mean Pi Zero users who don't care about gadget support will get dwc2 driver which may not work so well. Any thoughts?

Anyway, I'd like to get this in soon and a `BRANCH=next` firmware build out for testing.

clivem:
No, please do not default to dwc2! ;) People who want that need to specify it via config.txt, IMHO.

pelwell:
Relax:

> Currently it's not an issue as you need to choose the pi zero dtb explicitly

clivem:
Phil.... LOL. I am relaxed! Or I was until I read the @popcornmix comment, "may make sense to have firmware choose that automatically which would then mean Pi Zero users who don't care about gadget support will get dwc2 driver which may not work so well." And he's right, about the not working so well part. Trying to use a mt7601u wireless dongle was a complete train-wreck for me with dwc2. ;)

pelwell:
We wouldn't make that switch unless we thought dwc2 was solid. An interim step is likely to be to load the zero dtb automatically, but the defer the dwc2 selection to an overlay.

clivem:
@pelwell ...... which makes perfect sense to me! Any thought of dwc2 by default, was the scary part! ;)

pelwell:
We don't hate you _that_ much.

notro:
> @notro we'd like to get this in. Can you rebase?

I have very little time to work on the Pi these days and I have a kernel build in my tree that is not trivial to stash away.
Can you use the patches you already applied to rpi-4.4.y and just drop the zero dts patch?

> I wonder whether defaulting to dwc2 on Pi Zero is correct.

I assumed dwc2 was stable since it had moved out of staging, but @clivem has proven otherwise at least for one use case. Yes it's probably best to make it optional, also not having fiq support generates a lot of "slow" interrupts.

This means that the only difference between bcm2708-rpi-b-plus.dts and bcm2708-rpi-zero.dts is the pwr_led node. So maybe we should just drop the zero dts for now.
I wonder if this would work:
bcm2708-rpi-zero.dts

```
#include "bcm2708-rpi-b-plus.dts"

/ {
    model = "Raspberry Pi Model Zero";
};

/ {
    __overrides__ {
        /delete-property/ pwr_led_gpio;
        /delete-property/ pwr_led_activelow;
        /delete-property/ pwr_led_trigger;
    };
};

/delete-node/ &pwr_led;

```

popcornmix:
Okay.
`sudo BRANCH=next rpi-update`
should get the latest kernel with these commits.

NicoHood:
Ahm just a small question:
I tried the command above (with a clean full jessie image) and this installs kernel 4.4.0-rc8+ . I then set the device tree in config.txt and loaded the `g_ether` module via `/etc/modules`.

```
#config.txt (appended, the rest was kept unchanged)
device_tree=bcm2708-rpi-b.dtb
dtoverlay=mmc
```

Its not working the same way as @gbaman s [tutorial](https://gist.github.com/gbaman/50b6cca61dd1c3f88f41) did. So something must be missing, possibly the modules?

clivem:
Erm..... I haven't actually looked at that next branch, but I think your config.txt should be more like ....

device_tree=bcm2708-rpi-zero.dtb
dtoverlay=dwc2
and "dtoverlay=mmc" if you'd prefer to revert back to mmc rather than sdhost driver.

..... if you want to use g_ether on Zero.

You might want to wait for @popcornmix to reply before you blindly follow my advice and risk ending up with something that doesn't boot. ;)

NicoHood:
Okay so an even smaller tutorial now, but with more details after installing added.
https://github.com/NicoHood/NicoHood.github.io/wiki/Raspberry-Pi-USB-Gadget-Tutorial

Thanks for all the help! I just wrote down all my notes, most work was done by other great people!

<hr>

Also an still unanswered question is, if the pi zero is protected if I power it from the OTG port and the nativ power port at the same time. Otherwise switching from host to device at runtime (with separate PSU) is dangerous and might kill one of the devices connected to it.

Edit: Here is a solution: https://github.com/raspberrypi/linux/issues/1212#issuecomment-208219343

nmaas87:
Hey @NicoHood 

regarding the question about the MAC address,
it should work like this (change in /etc/network/interfaces)

```
allow-hotplug usb0
iface usb0 inet static
address 192.168.7.2
netmask 255.255.255.0
network 192.168.7.0
broadcast 192.168.7.255
gateway 192.168.7.1
dns-nameservers 8.8.8.8
hwaddress ether 00:01:04:1b:2C:1F
```

According to this site: http://www.youritronics.com/how-to-set-the-mac-address-from-etcnetworkinterfaces-in-debian/ (have not tested it, yet)

Regarding the powering problem, I think the only way to be sure would be to see, wheter there are some kind of protection diodes sepearting both (potential) power sources or such. Problem is, the electrical schematics of the pi zero are not (yet) available online (at least not where the rest of the stuff is found at: https://www.raspberrypi.org/documentation/hardware/raspberrypi/schematics/README.md )

NicoHood:
You can do

``` bash
echo "options g_ether host_addr=00:11:22:33:44:55" | sudo tee /etc/modprobe.d/g_ether.conf

#or
sudo nano /etc/modprobe.d/g_ether.conf
options g_ether host_addr=00:11:22:33:44:55
options g_ether dev_addr=00:66:77:88:99:00
```

Which works perfectly for me. I dont know which one is more reliable. the "host" setting sets the mac, that the connected host PC sees. The "dev" setting (just replace host with dev) sets the mac that the pi sees itself for its own usb0 adapter. This is confusing me. Shouldnt they be the same anyways? You can modify them in the way that the host sees the pi with a different mac than the pi sees itself.

Too bad about the schematics. Someone who inspected the pcb told me to better be carefully since it looks to him that the lines are just 1:1 connected. Thats a pity.

nmaas87:
@NicoHood 
Thanks for the heads up, regarding the MAC address :+1:  :)! 

IMHO:
I think the host_addr and dev_addr are important in the fact that they _are_ different. Yes, normally you are completly right, a ethernet/network interface is only allowed to have one mac address. But, I think the OTG Ethernet does emulate a "complete network" in that sense that you can see the OTG connection between host pc and pi zero not only as ONE network interface, but instead two.
So you got the ethernet interface on your pizero, which has the dev_addr and you got another ethernet interface on the pc, which needs to have a different mac, the host_addr. And those two virtual ethernet interfaces do communicate with each other.

HOST PC -------------------------------------------   OTG Ethernet  ----------------------------------------------- PiZero 
equals
HOST PC --- HOST Virtual Ethernet Card --- Virtual Wire ---- PiZero Virtual Ethernet Card ---- PiZero

So that would make sense - but I could be complelty wrong here.

About the OTG Power - yes, that would make sense - regarding the little space they got on the pcb as well as the price point (yes, not really expensive to do stuff like this - but every cm² and cent/penny counts!). So - I would really refrain from testing OTG + USB Power on a PiZero. Chances are the poor bugger would just die (maybe not instantly, but in given time :/)

NicoHood:
You are totally right, that MAC explanation makes sense. So I think we got a nice module solution here. I like it more than the network interface setting.

I just have no idea why the module doesnt load at startup. I could provide additional information if you like to. This new feature looks very promising and runs very smooth so far.

nmaas87:
Ok, I tried your small tutorial and can reproduce your problem:
While I could do the update without problems, the reboot after the kernel update already "broke" my setup ;). After that, I inserted

```
# This is required in the config.txt (make sure to use "zero" now)
device_tree=bcm2708-rpi-zero.dtb
# Regarding the readme this is applied automatically for a pi zero (and yes it works without)
# dtoverlay=dwc2
# The new kernel causes sd card errors for me, so i use the old sd card driver
# dtoverlay=mmc
```

into my /boot/config.txt and tried to boot again: Nothing. Or well - at least something: My windows machine at least did mention that there was an usb device which could not be recognized, which proofed your idea that the module is not loaded correctly via the /etc/modules file (yes, I got the g_ether entry there - of course - as well :)). Last try was to add modprobe g_ether to my /etc/rc.local - and - voilá - it came up - not a problem :). So you're right - I don't know what it is - but the entry in /etc/modules is somehow "ignored" (technically, I am sure it is not ignored, but some dependency is not yet loaded at that point of the boot process or something else happens which causes the g_ether to be not loaded if done via /etc/modules...). So I think maybe we could get some more infos via dmesg and other log files. Problem: I did neither got my Serial TTL device, nor my mini usb cable with me - so debugging that matter while beeing in need of the virtual ethernet will be a little bit hard :/.

Ah, I got the problem. Just a moment, I figure out what is _really_ needed, but I got it working again via "only /etc/modules" :)

nmaas87:
OK, solution for the "g_ether does not work, even though included in /etc/modules" is to include following lines in in /etc/modules:

```
dwc2
g_ether
```

dwc2 is needed and not loaded automatically(!). So the basic problem is: modprobe tries to include g_ether, but the dependency of it (dwc2) is missing and that fails. If you're trying to do that via sudo modprobe (probably later in the boot process) - dwc2 was already loaded by some other stuff - at least that is me - guessing what is going on.

@NicoHood Could you test that? Would that also work for your setup?

shrx:
Why do we need the full Jessie image instead of lite?

nmaas87:
@shrx I think the lite Jessie image should work too - basically the needed stuff is included within the kernel - so if you upgrade that, you should have it all together :). I used the latest version of the stripped down Minibian Jessie and it worked ( https://minibianpi.wordpress.com/ ).

notro:
> dwc2 is needed and not loaded automatically

It is, but later in the boot process. When /etc/modules is parsed it is not loaded yet, so since g_ether depends on it, it has to be manually loaded.

I believe `udev Coldplug all Devices` is the one autoloading dwc2, but this service is started after `Load Kernel Modules` and takes longer to finish:

```
[    7.603262] systemd[1]: Starting Load Kernel Modules...
         Starting Load Kernel Modules...
[    7.635960] systemd[1]: Starting udev Coldplug all Devices...
         Starting udev Coldplug all Devices...
<...>
[  OK  ] Started Load Kernel Modules.
[    8.060231] systemd[1]: Started Load Kernel Modules.
<...>
[  OK  ] Started udev Coldplug all Devices.
[    8.339931] systemd[1]: Started udev Coldplug all Devices.
<...>
         Starting /etc/rc.local Compatibility...
<...>
[  OK  ] Started /etc/rc.local Compatibility.
```

/etc/modules is hooked up to systemd through a symlink:

```
~$ ls -l /etc/modules-load.d/modules.conf
lrwxrwxrwx 1 root root 10 Sep  5 22:23 /etc/modules-load.d/modules.conf -> ../modules
```

Another thing:
Now I remember why I wanted dwc2 to be the default on Zero, because coupled with a raspian systemd service that would only run only on a zero in usb device mode, it could load g_serial automatically and give an out-of-the-box serial console: Just connect the Zero to a computer and get a console. But all in due time...

NicoHood:
I updated the tutorial above now with a lot more details!

It works great, and with the patch @nmaas87 suggested everything works now at boot. Plug in your pi and it always connects with the same mac now.

@shrx I guess we dont. I just tried with the full install. Feel free to test the lite image. I guess it should work without any problems.

I guess its quite easy to use OTG now ;)
The power issue is still now solved though.
I also never tried to switch back to a non-zero pi. Not sure what will happen then with this kinda dev state.

clivem:
@notro LOL. You're making me nostalgic! I remember how cool I thought it was when I plugged my first BeagleBone into a PC and got a serial console....... But the fact doesn't change, without dwc2 receiving some love, while it may be the answer for world peace in device mode, it leaves something to be desired, (that's the politest way of putting it), in host mode. wifi dongles suddenly disconnecting from USB bus is the tip of an iceberg. Heaven help you, (train crash leading to toxic chemical spill), trying to use it for USB audio! ;)

notro:
> I remember how cool I thought it was when I plugged my first BeagleBone into a PC and got a serial console

Me too, it was so easy to get started.

Your comment jogged my Solution-Finder-Unit and I realised that my goal might not be so far away after all.
config.txt is easily available for editing, so a user could add: dtoverlay=dwc2
Then the service could do:

```
if zero() and dwc2() and device_mode() then
    modprobe g_serial
end
```

A keyboard plugged in after boot will still work, and if this behaviour is not desirable, the service can just be disabled.

So here's project for me when I'm done banging my head agaist getting Pi2 mainline support working without issues (unless someone can't wait that long and finds a way to actually implement this).

> Heaven help you, (train crash leading to toxic chemical spill), trying to use it for USB audio! ;)

My focus on an easily available serial console gave me tunnel vision, so I'm glad there's a review process :-)

NicoHood:
Just a small question:
I cannot use ping without sudo on the g_ether module. Is this normal?

``` bash
pi@raspberrypizero:~$ ping 8.8.8.8
ping: icmp open socket: Operation not permitted
```

nmaas87:
@NicoHood Not a problem because of g_ether I think. Test if the SUID bit is set correctly on your ping command, as described here: http://ben.goodacre.name/tech/Ping:_icmp_open_socket:_Operation_not_permitted_(Linux)
On my Minibian Jessie - it worked right out of the box :)
Edit: It did work right out of the box (ping 8.8.8.8) without root access, however, the SUID bit was not set on my PiZero (I have no idea how this could work, because network access is always a kind of root thing...). On my Ubuntu Box, however, the SUID bit is set on /bin/ping :).

t3chguy:
For those wanting to test it, it seems as though you'll need to use
`sudo BRANCH=next rpi-update d17f1116bcd7f087451dfaed855bd7625c1e49a7`
unless I am missing something, the dtb file referenced for device tree is missing in the latest commit to next of rpi-firmware.

NicoHood:
I now also tried the usb gadget. It seems to work fine, I tried to install an OS into the emulated usb drive, but it failed after booting. But it was used on a slow "random" sd card, I will try this again on a better one.

The question is now, how to enable the gadgets by default on a zero, but disable them in a pi2? We currently have to edit the config.txt. Is there a way to add something like `if` inside those files, or can the be done with a kernel fix?

Also can we automatically enable the dwc2 module if g_ether is loaded for example? Would this work too? Then people do not have to type both of them in the `/etc/modules`. 

I really would like to see this feature more flexible across different pi models. This is really awesome. And some of the responsible pi people should solve the power issue by providing a schematic, so we know if connecting a PSU in device mode hurts any device.

NicoHood:
I just tried again (with a different sd card, from the beginning) and after applying:

```
echo "device_tree=bcm2708-rpi-zero.dtb" | sudo tee -a /boot/config.txt
```

The raspi stops booting and the rainbow screen only shows up. If i remove the line (with an sd card reader) it boots up again.

Using g_ether without those line does not work, but the line itself now also doesnt work anymore. So what has changed?

I first setup the device with a pi2, however the 4.4 kernel all boots fine on a zero, just the device tree setting makes it not boot up anymore.

Edit:
I put in my old sd card and I noticed that `bcm2708-rpi-zero.dtb` is just missing on the boot partition. Maybe because I did the rpi-upgrade on the pi2, not the zero? I guess this is a bug?

t3chguy:
@NicoHood no, the zero specific .dtb is no more as I found out. Just `dtoverlay=dwc2` is enough after `sudo BRANCH=next rpi-update`, then either manually load one of the g_ modules or stick

```
dwc2
g_ether
```

inside `/etc/modules`

notro:
What's the link to the Raspian github issue tracker? I can't find it.
I have a proposal on how to easily enable a usb console.

lurch:
> What's the link to the Raspian github issue tracker?

I guess https://github.com/RPi-Distro/repo/issues is the nearest match?

popcornmix:
Correct. For raspbian build of default debian packages it is:
https://www.raspbian.org/RaspbianBugs

For raspberry pi specific packages and init scripts (not from debian) then:
https://github.com/RPi-Distro/repo/issues

notro:
Thanks. https://github.com/RPi-Distro/repo/issues/17

NicoHood:
@t3chguy thanks! I updated the tutorial above. `echo "dtoverlay=dwc2" | sudo tee -a /boot/config.txt` is now all you need.

Is there a chance to load this overlay by default or is there any disadvantage for this?

There is also another bug that `/etc/modprobe.d/g_ether.conf` is not recognized sometimes. I think its mostly after first loading g_ether. If you remove it and load it again it will be applied.

lurch:
> Is there a chance to load this overlay by default or is there any disadvantage for this?

See the messages from @clivem earlier in this discussion.

NicoHood:
So its unstable for host mode on some devices and it will be kept like this?

Cant we load/unload those overlays at runtime? (Sorry i never did such kernel things before).
Is there any chance to see this somehow improved in the future or is this just the limitation we have?

braian87b:
Hey, what about using with usbip client as host on Raspberry Pi Zero (as usbip server running in another rasbperry pi with a device connected), then wirelessly you can plug Raspberry Pi Zero on a computer and see the device if it where connected directly to the computer. Can this be done?

lurch:
@braian87b I've read your comment several times, and I'm still not quite sure what you're asking for?

I _think_ what you're suggesting is having a PiZero in gadget-mode connected over USB to a PC, and then having the PiZero communicating wirelessly with another Raspberry Pi, and then having a USB device you connect to this second Raspberry Pi "tunnelled" wirelessly to the PiZero, and then exposing it as a 'virtual' USB device to the PC which the PiZero is connected to?

However the PiZero only has one USB OTG port, which can operate in either host or device mode. So if you've got the Pi connected in device-mode to a PC, there's then no 'spare' USB host port into which to plug a Wifi dongle. There _are_ ways of communicating with a wifi chip over the Pi's GPIO pins e.g. https://shop.pimoroni.com/products/esp8266-phat but that only runs at 1mbps, whereas USB2 runs at up to 480mbps (even USB1 is 12mbps).

braian87b:
@lurch Right... sorry, I didn't notice... thanks!

dtaht:
I was catching up on this discussion in light of some discussion elsewhere ( https://plus.google.com/u/0/+EricRaymond/posts/DR2zNNyNs3Z ) 

and I see my pi2 had in 4.4.12+/   

 for dwc2.ko 

but not in my latest update (4.4.12-v7). Does the OTG stuff work on the pi2 or pi3?

lurch:
> Does the OTG stuff work on the pi2 or pi3?

No. The USB hub in the LAN9514 'gets in the way' and means that the single USB port in the SoC on those devices can only work in host (master) mode. It's only because the PiZero has a single USB port (i.e. no USB hub) that it's also able to (optionally) work in device (slave) mode.

Dygear:
Anyone happen to have documentation on how to use the `g_serial` option. With only
 * Raspberry Pi Zero (W),
 * Formatted SD Card with Jessie Lite
 * A USB Cable.

I want to just connect it to my desktop computer and connect to it via `ttyACM0: USB ACM device`.

The [CHIP does this out of the box](https://bbs.nextthing.co/t/setting-up-chip-as-a-headless-server-with-minimal-tools/1505), so I was hoping to get the same sort of interface out of the Pi Zero. Can someone point me in the right direction for that?

(I'm willing to be a guinea pig as I have 2 Pi Zero Ws and 1 Pi Zero - All of them I intend on using as headless servers. So never having to get a USB keyboard and Mini HDMI to HDMI connector is a bonus.)

t3chguy:
@Dygear you can use this for reference https://bit.ovh/2016/01/31/Raspberry-Pi-Zero-Gadget-Mode/

You can ignore rpi-update and run all those commands in a chroot of the mounted sd card (or even looped iso/img) 

You will have to create an appropriate symlink so systemd starts the tty/getty, and ignore the interfaces related commands and replace g_ether with g_serial

I have notes on my laptop, I'll update when I get home

bastijr:
我现在需要PI zero实现一个功能：使用USB连接PIzero和iPhone使得iPhone是host，PIzero是主？请问这个已经实现了吗？是否有开源的代码？

bastijr:
I now need PI zero to implement a function: use USB Wire to connect PIzero and iPhone so that iPhone is the host, PIzero is the slave? I want to known if the funtion has been finished? Is there open source code?

pelwell:
This is a question, not an issue. Please open a topic in the Raspberry Pi forums, stating what functionality the PiZero should provide - mass storage, etc.

PR #1240, opened 2015-12-27
MMC: Do not use mmc_debug if CONFIG_MMC_BCM2835 is not set

JanLuca:
If CONFIG_MMC_BCM2835 was not set the compiling of the kernel failed
since mmc_debug was not defined but used in drivers/mmc/core/quirks.c.

This patch add a ifdef-check for CONFIG_MMC_BCM2835 to the change of
commit 64d395457f793250d2e582eeb38cc3403b1db98c


JanLuca:
This patch applies to other branches, too (all branches with commit 64d395457f793250d2e582eeb38cc3403b1db98c)

PR #1255, opened 2016-01-11
SDIO-overlay: Add parameter to change the SDIO bus width

al177:
Adds a parameter to the SDIO overlay, bus_width=, that can be used to limit the SDIO bus to 1 bit mode.  It is necessary to limit the driver to 1 bit mode when the bus is wired for 1 bit mode but the device advertises 4 bit mode.

Tested that this doesn't affect the bus width on the remapped SDIO bus or on the bcm2835-sdhost bus when not defined and when the SDIO overlay is not loaded.

### DIFF
--- a/arch/arm/boot/dts/overlays/README
+++ b/arch/arm/boot/dts/overlays/README
@@ -613,6 +613,8 @@ Params: overclock_50             Clock (in MHz) to use when the MMC framework
         poll_once                Disable SDIO-device polling every second
                                  (default on: polling once at boot-time)
 
+        bus_width                Set the SDIO host bus width (default 4 bits)
+
 
 Name:   smi
 Info:   Enables the Secondary Memory Interface peripheral. Uses GPIOs 2-25!
--- a/arch/arm/boot/dts/overlays/sdio-overlay.dts
+++ b/arch/arm/boot/dts/overlays/sdio-overlay.dts
@@ -11,6 +11,7 @@
 			pinctrl-names = "default";
 			pinctrl-0 = <&sdio_pins>;
 			non-removable;
+			bus-width = <4>;
 			status = "okay";
 		};
 	};
@@ -28,5 +29,6 @@
 
 	__overrides__ {
 		poll_once = <&sdio_mmc>,"non-removable?";
+		bus_width = <&sdio_mmc>,"bus-width:0";
 	};
 };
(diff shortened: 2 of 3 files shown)


pelwell:
I have no objections to these. @popcornmix ?

popcornmix:
Will the bus width still be 4 when booted with device tree disabled?

al177:
I haven't tested with devicetree totally disabled (not sure how to do that on a Pi setup) but the [if at 1464 of bcm2835-mmc.c](https://github.com/raspberrypi/linux/blob/rpi-4.1.y/drivers/mmc/host/bcm2835-mmc.c#L1464) should already take care of the case where there is no dt node for mmc at all.

I did test that mmc0 comes up at 4 bits wide when the sdio overlay isn't loaded, and that mmc1 width comes up at 4 bits wide when the parameter isn't provided.

popcornmix:
`device_tree=` in config.txt would disable device tree.

pelwell:
I'm sure it will be ok, but you would need to confirm that the performance was unchanged since the interface will still work in 1-bit mode.

al177:
I verified that mmc0 comes up in 4-bit mode when device tree is disabled.

I checked with a scope that one of my SDIO targets (a 4 bit capable device) is in 1-bit mode when requested by checking that SD_D2 and SD_D3 are idle, and in 4-bit mode as selected by bus_width=4 and when omitting bus_width all four data lines are active.

I also verified that the pins mapped to mmc1 were all idle when the SDIO overlay wasn't loaded.

SD card r/w performance seems unchanged for my card (Samsung 32GB Evo+) for each of these cases at ~17MB/s for write and ~22MB/s for read.  Those numbers would not be possible in 1-bit SDIO mode.

In all of these cases, /sys/kernel/debug/mmc1/ios reports the bus as running at the expected rate and /sys/kernel/debug/mmc0/ios always reports a bus width of 4.

campag:
When in 1-bit mode, SD_D2 & SD_D3 aren't used, yet the overlay still configures those pins. Should there be a separate overlay for 1-bit mode which doesn't configure the corresponding GPIOs 26 & 27, or is there another way to do that in the existing overlay?

pelwell:
The existing parameter mechanism isn't flexible enough to support a variable number of pins in a clean way - a second overlay is far simpler.

PR #1261, opened 2016-01-18
VC4 backport to 4.1

anholt:
This branch pulls the current rpi-4.4.y VC4 code and its dependencies back to 4.1.  I've tested on a 2709 build not using the vc4-overlay (since I use u-boot, which I suspect can't handle the overlays).


popcornmix:
Kernel boots with kms overlay disabled.
I've cherry-picked the kms overlay from 4.2, but so far not booting.
Framebuffer doesn't initialise and my uart is not currently working.
I suspect there is more missing DT stuff.

pelwell:
There was a problem with a recent I2C commit that broke the HDMI support. I've pushed a fix for this, and with the overlay commits added to the mix the new driver loads and presents a console as expected.

PR #1265, opened 2016-01-24
Add master volume control for HiFiBerry DAC+

hifiberry:
The chipset driver (pcm512x.c) allows acces to the master volume control in a way that clipping is possible. The new "Master" volume control in the HiFiBerry DAC+ driver limits volume to +0dB and makes it easier to use for unexperienced users as 100% equals 0dB (max volume without clipping).

### DIFF
--- a/sound/soc/bcm/hifiberry_dacplus.c
+++ b/sound/soc/bcm/hifiberry_dacplus.c
@@ -29,7 +29,7 @@
 #include <sound/pcm.h>
 #include <sound/pcm_params.h>
 #include <sound/soc.h>
-#include <sound/jack.h>
+#include <sound/tlv.h>
 
 #include "../codecs/pcm512x.h"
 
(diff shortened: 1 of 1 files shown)


pelwell:
I'm not sufficiently invested in the audio card ecosystem to make a call on the merits of this patch without further input. @iqaudio - what is your opinion?

While you are debating whether the approach this patch takes is correct, I have some minor issues with the patch as it stands:
1) The repeated calls to snd_rpi_hifiberry_dacplus_ctl_set_map are a bit hacky, but I can see that they are convenient.
2) The indendation of rpi_hifiberry_dacplus_snd_controls is excessive.
3) It's a shame you couldn't add three extra spaces to the initialisation of the controls member of snd_rpi_hifiberry_dacplus.

iqaudio:
Phil, sorry for delay in replying - traveling this week. 

We will always be at the mercy of anyone changing the codec drivers, but the names have changed for good reason and are, through discussion, now stable.

The particular device allows upto (iirc) +23db which I'd exposed as 100% volume would cause the majority of users an issue and for this reason we first set 100% at +4db but still the majority of users wanted 0db at 100%. This we did through simple driver change.

The main variable mixer name "digital" is following the asoc gatekeepers recommendation - not my choice - but this has been adopted by the majority of the software vendors. 

It should be noted that if advanced users want to push the volume > 0db it would be a simple utility to adjust the chip's volume registers if needed.

I'm happy with the status quo as we've all seen the environment stabilise but what Daniel wants or feels best for his customers is up to him.

Daniel, if you want to just take the code from the pi-dacplus driver. No issue.

Best regards,

Gordon

pelwell:
Thank you, Gordon.

@hifiberry Do you still want to go ahead with this change, after addressing my nitpicks?

clivem:
Can I suggest that now that Gordon has spoken, that Daniel uses a module param. Then his "experienced users" who can be trusted not to be applying 24db gain via the Digital control, to the latest CD they bought and ripped yesterday, which has been brickwall mastered to 0dbfs already.... can be instructed to create a module file with the module option disabled. eg.
[raspberrypi-kernel-4_4-sound-soc-bcm-berry_dacplus-max_vol_v2.txt](https://github.com/raspberrypi/linux/files/103541/raspberrypi-kernel-4_4-sound-soc-bcm-berry_dacplus-max_vol_v2.txt)

clivem:
Daniel, using dt rather than module option .......
[raspberrypi-kernel-4_4-sound-soc-bcm-berry_dacplus-max_vol.txt](https://github.com/raspberrypi/linux/files/104150/raspberrypi-kernel-4_4-sound-soc-bcm-berry_dacplus-max_vol.txt)

clivem:
Oops.... Left a reference to the iqaudio in that previous patch....
[raspberrypi-kernel-4_4-sound-soc-bcm-berry_dacplus-max_vol.txt](https://github.com/raspberrypi/linux/files/104201/raspberrypi-kernel-4_4-sound-soc-bcm-berry_dacplus-max_vol.txt)

hifiberry:
Ok, I will do it that way. I hope the volume control name really stays at it is now.

clivem:
Daniel, I also hope that volume control name never gets renamed! It has caused me pain also. But based on what Gordon has said, (and a couple of other conversations I have had), there doesn't seem to be any intention to want to change it again.

The more immediate problem, IMHO, is "clipping". I have been supporting users with your hardware, twice in the last two weeks..... Although I have no wish to insult those users...... Are you familiar with Spinal Tap and the amplifier volume control that doesn't stop at 10, but goes to 11? LOL. That's what we have here. A users instinct, (especially when using playback software that applies volume gain in software before it gets to mixer), is to max each mixer control...... Leaving aside the techies and audiophools who understand the concept of decibels, a lot of users purchasing your hardware don't know the difference between a decibel and a doughnut! These people do need the "elf and safety" of someone who knows, protecting them from what they don't know. But of course, with a way to over-ride the default behaviour for the people who do understand and can safely have full access to everything the underlying hardware is capable of, without causing the supplier of the hardware (or any software that runs on it or is using it) a support nightmare!

I had a conversation last night, where we were talking about early CD's...... Some 1980's CD's that were mastered with plenty of headroom, where you could want to apply gain in the digital domain. But then it was pointed out to me, that people are already using replay gain, and some sort of software gain adjustment, based on those replay gain tags before the audio gets to hardware mixer...... The conversation then turned to the exposing of de-emphasis filter via mixer control and how many people would manually set that even if they knew that they had an early CD with it applied. Then I was told something I wasn't aware of, that if that CD is ripped with iTunes, it automatically applies filter during the rip, so the ripped CD copy wouldn't need that filter applied by DAC hardware anyway.......

PeterPablo:
As a user I am very much looking forward to the new default behavior of limiting the gain to 0 dB, as I currently can not practically leverage the hardware mixer control using the software `shairport sync` of @MikeBrady: If I would currently completely turn up the volume of my laptop (Mac) to 100%, the volume (see `alsamixer` of the mixer control `Digital`) would also change to 100% which at the moment of writing corresponds to +24 dB and thus leads to strong audible clipping!

For the sake of consistency and compatibility I see no reason for (yet) a new mixer name.

hifiberry:
As there is now a new patch here

https://github.com/DigitalDreamtimeLtd/linux/commit/0ee32f650e9f3ed70dac13f46e30141b2513101b

I will close this pull request.

PR #1270, opened 2016-01-27
solved issue of mirroring screen after rotation.

sphonala:
changed init code for rotation & new tft voltage problem.


popcornmix:
We'll need more information on the problem and what this does.

Really this is an upstream file and the patch should be submitted upstream:

I can't judge whether this is a safe change and won't cause problems for other users.
Once accepted upstream we can pull the change in.

Ping @notro who may know about this.

sphonala:
@notro, we have solved issue of 90 degree rotation for tinylcd 3.5 tft and voltage problems for new tinylcd tft's. the present init code for fb_tinylcd.c having an issue of 90 degree rotation, image showing its mirror after 90 degree rotation and also solved the voltage issue for new tft's with changing write_reg(par, 0xC1, 0x45, 0x07); to write_reg(par, 0xC1, 0x01, 0x01);
thanks.

tinylcd:
@notro commit this changes @sphonala to fb_tinylcd.c

notro:
fbtft is practically dead now.
The fbdev maintainer won't accept any new fbdev drivers: [No more new fbdev drivers, please](https://lkml.org/lkml/2015/9/24/253), so it can't be moved out of staging.
And Greg Kroah-Hartman (staging maintainer) didn't include any fbtft patches in the [[GIT PULL] Staging driver patches for 4.5-rc1](http://lkml.iu.edu/hypermail/linux/kernel/1601.1/03960.html) change set.
This means that at some point it will be deleted, since staging is a temporary place for drivers to stay while they're cleaned up for inclusion in their proper place.

So there's no point in sending fbtft patches. Bug fixes might get in, but this seems to be a new display.

I have been working on and off on a [DRM](https://en.wikipedia.org/wiki/Direct_Rendering_Manager) version of fbtft and I have a hack with one display working (sort of), but it's still some work to be done before I can post some patches for review.
This is my work so far: https://github.com/notro/tinydrm

notro:
Notify some more people about the current fbtft status: @ladyada @awatterott @mwilliams03

ladyada:
ok thanks for the notice! hopefully the raspi fork can keep it in?

Ferroin:
The intent is to get a rewrite using the DRM (Direct Rendering Manager) framework pushed upstream.  Personally, I don't think it's likely that fbtft will be removed completely upstream until that happens, as the intent to eventually rewrite it using DRM has been stated on the mailing list.

tinylcd:
sad to know about "The fbdev maintainer won't accept any new fbdev drivers", hopefully the raspi fork can keep it in.
@notro Bug fixes might get in, but this seems to be a new display.
by these changes, voltage & rotation issues are fixed. these changes are now work on both new & old set of displays.

notro:
As @Ferroin says, if I can come up with a drm version in a reasonable time, fbtft won't go away before there's an alternative.
But that's just part of the story, because even if I make an alternative "library" solution, the individual displays have to be ported/supported. It isn't difficult to do, but it's takes time to do the testing and prepare patches. It's unlikely that I will do this even for all the displays I got myself. At least not in the required timeframe. Hence the heads up for some of the businesses that use these drivers. There's no problem yet, but it might be.

@tinylcd 
Ok, if marked as a bug fix it should probably be accepted. However I can't do it.

ladyada:
@notro np for the displays i have access to: HX5378D, ILI934x, SSDxxxx OLEDs, ST7735, etc. i've already done the code on an Arduino, so no biggie for me to re-port them again :) 
without putting any pressure on you - do you have some ETA on when a DRM version may be available for first test? if not, do you mind pinging me when there's something to test?

notro:
@ladyada

> do you have some ETA on when a DRM version may be available for first test?

No. I have worked on and off with this for over a year now, but this time around I plan to keep going until I have sent an RFC patchset for review.
The intention is to get some early feedback to see if I'm heading in the right direction.

> if not, do you mind pinging me when there's something to test?

Sure.

I will also post updates in this thread: https://github.com/notro/tinydrm/issues/2

Ruffio:
This PR is pointing to 4.1.y, is this ever going to be merged?

popcornmix:
No would need to be resubmitted to 4.4 tree if still desired, although in general we are reluctant to keep patches to upstream code in this tree (we'd prefer the changes went upstream).

PR #1277, opened 2016-02-01
I2S fixes + PCM3060 codec support

jonronen:
There are two changes in this pull request:
1. Fixing the Raspberry Pi's I2S driver init. The driver's init code was written inside bcm2709.c which, in addition to being architecturally wrong, prevented the driver from loading in kernel 4.1 because of device tree issues. The init code was moved to the driver's source code, where both the SoC device and the driver are loaded together.
2. Adding support for the TI PCM3060 audio codec. This codec provides high quality audio input and output and can be used for sound effects, sampling, playback, etc. A board based on this codec can be found here: https://oshpark.com/shared_projects/9mRzWVHy with schematics here: https://raw.githubusercontent.com/jonronen/geda-stuff/master/audio_shield/pcm3060_rpi/pcm3060.png


Ruffio:
This PR is pointing to 4.1.y, is it ever going to be merged?

popcornmix:
No. Would need to be resubmitted to 4.4 if still required.

PR #1282, opened 2016-02-04
Allow up to 24db digital gain to be applied when using IQAudIO DAC+

clivem:
24db_digital_gain DT param can be used to specify that PCM512x
codec "Digital" volume control should not be limited to 0db gain,
and if specified will allow the full 24db gain.

### DIFF
--- a/arch/arm/boot/dts/overlays/README
+++ b/arch/arm/boot/dts/overlays/README
@@ -362,8 +362,21 @@ Params: <None>
 
 Name:   iqaudio-dacplus
 Info:   Configures the IQaudio DAC+ audio card
-Load:   dtoverlay=iqaudio-dacplus
-Params: <None>
+Load:   dtoverlay=iqaudio-dacplus,<param>=<val>
+Params: 24db_digital_gain       Allow gain to be applied via the PCM512x codec
+                                Digital volume control. Enable with
+                                "dtoverlay=iqaudio-dacplus,24db_digital_gain"
+                                (The default behaviour is that the Digital
+                                volume control is limited to a maximum of
+                                0dB. ie. it can attenuate but not provide
+                                gain. For most users, this will be desired
+                                as it will prevent clipping. By appending
+                                the 24db_digital_gain parameter, the Digital
+                                volume control will allow up to 24dB of
+                                gain. If this parameter is enabled, it is the
+                                responsibility of the user to ensure that
+                                the Digital volume control is set to a value
+                                that does not result in clipping/distortion!)
 
 
 Name:   lirc-rpi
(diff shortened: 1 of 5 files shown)


clivem:
I've added a similar patch for the HiFiBerry DAC+, but that should not be applied without an OK from Daniel first! He was working on a patch himself and might prefer whatever he has to this.

arch/arm/boot/dts/overlays/README
```diff
@@ -273,8 +273,21 @@ Params: <None>
...
+Params: 24db_digital_gain       Allow gain to be applied via the PCM512x codec
+                                Digital volume control. Enable with 
+                                "dtoverlay=hifiberry-dacplus,24db_digital_gain"
+                                (The default behaviour is that the Digital
+                                volume control is limited to a maximum of
+                                0db. ie. it can attenuate but not provide
+                                gain. For most users, this will be desired 
+                                as it will prevent clipping. By appending 
+                                the 24db_digital_gain parameter, the Digital
+                                volume control will allow up to 24db of
```
PeterPablo: You might want to write `dB` (capital "B"), as is [usually done](https://en.wikipedia.org/wiki/Decibel) for the unit decibel. This would apply to several places in your patch.

PeterPablo:
Ping @hifiberry, so Daniel is noticed about this patch.

hifiberry:
Looks good. Thank you for this.

popcornmix:
Any objections to merging? @pelwell ?

pelwell:
@DigitalDreamtime Thanks for these commits. I think the use of the capital in "dB" only needs to apply to the README. All other DT parameters are lower case, and the Linux coding style expects lower case variable names.

If you can partially revert your fixups, squash ("fixup") them into the original commits and "git push -f" them I will merge them. Otherwise I'll patch them up and commit them tomorrow.

pelwell:
i.e. I push them but with your name on.

popcornmix:
@DigitalDreamtime if you are fixing up the commit, can you fix the trailing whitespace?
`git rebase --whitespace=fix HEAD~2`
should do it.

clivem:
You guys OK with that. Lowercase "db" in dt and code. --whitespace=fix, squash and rebase.

PeterPablo:
Will these two patches be backporded/cherry-picked into rpi-4.1.y or is the answer a simple no, due to the approaching switch to the 4.4 kernel branch? (I would like to test this and give feedback and am wondering wether `sudo BRANCH=next rpi-update` is the way to go.)

clivem:
I was assuming 4.4.y was going to be the "stable" branch "real soon now". Need small mod to apply to 4.1.y. Phil, do you want me to submit a pull for 4.1.y?

pelwell:
I think that would be a waste of effort, Clive. Peter - we welcome any testing of the BRANCH=next builds; Dom will push out an update tonight including these patches.

popcornmix:
`BRANCH=next rpi-update` now includes this PR.

PeterPablo:
This works as advertised and prevents clipping, if be incautious. Thank you @clivem!

PR #1292, opened 2016-02-15
Fix mkdir error in makefile

LuisHsu:
I just found a bug in makefile:1123
It remove the $(MODLIB)/source,$(MODLIB)/kernel and $(MODLIB)/build folders without re-make it


popcornmix:
This file is unchanged from upstream, so really any problem should be reported there.
See: https://www.kernel.org/pub/linux/docs/lkml/reporting-bugs.html

PR #1293, opened 2016-02-16
Updated smsc95xx driver to check for a valid MAC address in eeprom first

Craig-J-Roberts:
I'm using multiple USB->Ethernet Adaptors based on the same chip as the RPi has. I've found they all show up with the same MAC address on boot. This Pull Request solves this by checking for a valid MAC on eeprom before setting the MAC address based on the smsc95xx.macaddr parameter. It's a re-order of existing code, not additional code.


popcornmix:
Tested and it didn't affect the MAC address on Pi. Seems okay to me.

JamesH65:
Is this an upstream issue as well? Or is this driver Pi only?

popcornmix:
This alters a patch we already added to allow MAC address to be set from command line options, so it doesn't affect upstream.

PR #1299, opened 2016-02-22
Rpi 4.4.y SPI auxiliaries with overlays

frasersdev:
Following on from https://github.com/raspberrypi/linux/pull/1295

This Pull request enables the SPI 1 & 2 auxiliaries  to compatible raspberry PIs
- Backports the Auxiliary clock from 4.5 to 4.4
- Minor config of the driver build environment to ensure they get built for CONFIG_ARCH_BCM2708 & CONFIG_ARCH_BCM2709 devices.
- Adds the aux spi driver into the defconfigs as a module.
- Adds the auxiliary and spi1/2 devices into the device tree in a disabled state
- Provides device tree overlays which enables the devices and gives users a degree of control over how they are setup.

This pull request resolves issue https://github.com/raspberrypi/linux/issues/1034


frasersdev:
one sec, there is a typo

frasersdev:
One thing that changed from the example I did was the (re)naming of the client enable (CE) lines (instead of client select CS). This is because the broadcom peripheral document uses 'CE' and that is what most people refer to them as on the PI. Its printed on scores of boards and headers etc. What do you think?

pelwell:
I'm not keen on that aspect of the change - the README doesn't use CE consistently which tells me that you aren't convinced either. The DT bindings for SPI refer to chipselects and use the abbreviation cs, so I think it is less confusing to keep that nomenclature.

By the way, I've made a few cosmetic tweaks to the third patch - you can find an updated version https://gist.github.com/pelwell/972aab58a718081502be, and if you are happy with my changes then you could update this PR with it and any CE->CS changes before I merge it.

frasersdev:
ok, commit updated.
- reverted ce->cs
- aux patch included (tested & working fine BTW)

no other changes!

I'm happier with my README in this attempt. 
I guess what was bothering me was there needed to be some reference back the the broadcom nomenclature as a reader may not make the connection that CS=CE (e.g. non-english speaker) - I think (hope) the reference aside the default numbering is enough without complicating everything.

pelwell:
It's a shame you edited your originals rather than my updated versions, but I've reapplied my changes and pushed the result.

AlexSudnikov:
Hey everyone ! Sorry for a dumb question but...how do i enable theese aux SPI's. I mean , i only get spidev0.0 and spidev0.1 ( once a proper dtparam is added to the config ).
But if try to activate spi1 or spi2 ( following DTB's  __symbols__ names ) , like this:
dtparam=spi1=on
dtparam=spi2=on
Nothing happens...maybe someone could point out what is the problem , thx ! )
Btw, i'm using Rpi-3 model B , Linux 4.9.41-v7+

pelwell:
If you run `dtoverlay -a | grep spi` in a Raspbian shell you'll get:
```
  enc28j60-spi2
  raspidac3
  sc16is752-spi1
  spi-gpio35-39
  spi-rtc
  spi0-cs
  spi0-hw-cs
  spi1-1cs
  spi1-2cs
  spi1-3cs
  spi2-1cs
  spi2-2cs
  spi2-3cs
```
The overlays named `spi1-*` and `spi2-*`enable SPI1 and SPI2 with different numbers of chip-select/chip-enable lines.
```
$ dtoverlay -h spi1-2cs
Name:   spi1-2cs

Info:   Enables spi1 with two chip select (CS) lines and associated spidev
        dev nodes. The gpio pin numbers for the CS lines and spidev device node
        creation are configurable.
        N.B.: spi1 is only accessible on devices with a 40pin header, eg:
              A+, B+, Zero and PI2 B; as well as the Compute Module.

Usage:  dtoverlay=spi1-2cs,<param>=<val>

Params: cs0_pin                 GPIO pin for CS0 (default 18 - BCM SPI1_CE0).
        cs1_pin                 GPIO pin for CS1 (default 17 - BCM SPI1_CE1).
        cs0_spidev              Set to 'disabled' to stop the creation of a
                                userspace device node /dev/spidev1.0 (default
                                is 'okay' or enabled).
        cs1_spidev              Set to 'disabled' to stop the creation of a
                                userspace device node /dev/spidev1.1 (default
                                is 'okay' or enabled).
```
Note:
1. Although we found the overlays and their usage from a shell, the above usage steps are meant to go in `config.txt`.
2. Each interface is only available on a small number of pin groups - SPI1 is on GPIOs 16-21, while SPI2 is on GPIOs 40-49. Since only GPIOs 0-27 are brought out to the 40-pin header, SPI2 is only usable on a Compute Module.

AlexSudnikov:
Ah, i see. Thanks for a reply pelwell !) 
The only thing that is still a bit unclear to me is...how do i communicate with these aux. devices , i mean there is a python spidev module , which is  used "for interfacing with SPI devices from user space via the spidev linux kernel driver " , however one still has to specify the exact device ( /dev/spidev0.0 for instance ) 
to communicate with. So the question is how does one actually use these chip select lines...but i think it's all about API :) Anyways, thx !

PR #1300, opened 2016-02-23
ASoC: bcm: add missing .owner fields in sound card drivers

HiassofT:
If snd_soc_card.owner is not set the kernel won't do usage refcounting
and one can remove the card driver module while it's in use (eg playback
active) - which leads to a kernel crash.

The missing owner field also prevents ALSA slot ordering
(options snd slots=module-name1,module-name-2,...) from working with
the I2S cards as it has no module name to match against.

Fix these issues by setting the .owner field in the snd_soc_card structs.

@popcornmix @pelwell if you merge this please also cherry-pick this commit into the 4.1 and 4.5 trees

### DIFF
--- a/sound/soc/bcm/hifiberry_amp.c
+++ b/sound/soc/bcm/hifiberry_amp.c
@@ -61,6 +61,7 @@ static struct snd_soc_dai_link snd_rpi_hifiberry_amp_dai[] = {
 
 static struct snd_soc_card snd_rpi_hifiberry_amp = {
 	.name         = "snd_rpi_hifiberry_amp",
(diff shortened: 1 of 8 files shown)


clivem:
Good catch. And off-topic, but thanks anyway. It also told me what I needed to do to be able to "position" sun4i_codec driver on Cubietruck using "slots". I half realised what the problem was with that when "cat /proc/asound/modules" returned null rather than a module name, but this gave it to me on a plate, so thanks for that. ;)

pelwell:
Merged, with thanks.

PR #1301, opened 2016-02-24
Add 1-bit SDIO with the minimum pins required for that mode: GPIOs 22…

campag:
…-25.

For increased compatibility with other hardware, it's useful to force SDIO into 1 bit mode & free up GPIOs 26, 27 (SD_D2, SD_D3). Variable pin numbers isn't possible, so an overlay separate from the existing 4-bit/1-bit dual mode overlay is proposed.

### DIFF
--- a/arch/arm/boot/dts/overlays/README
+++ b/arch/arm/boot/dts/overlays/README
@@ -654,6 +654,27 @@ Params: overclock_50            SD Clock (in MHz) to use when the MMC framework
         bus_width               Set the SDIO host bus width (default 4 bits)
 
 
+Name:   sdio-1bit
+Info:   Selects the bcm2835-sdhost SD/MMC driver, optionally with overclock,
+        and enables 1-bit SDIO via GPIOs 22-25.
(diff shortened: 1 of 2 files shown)


pelwell:
Can you explain why this requires 4 pins and not the expected 3 (CLK, CMD + DAT0)?

campag:
I believe DAT1 pin is specified/used as INT in one bit mode. Whether that is used in practice on the Pi's mmc peripheral when in one bit mode I have not checked. All the better if we can spare another pin - I'll take a look.

campag:
Attempting to run 1-bit mode with just three wires caused lock-ups. It seems the fourth D1/INT is essential.

pelwell:
Thanks for the clarification.

campag:
@pelwell looks like I forgot to add the overlay to the Makefile in the PR, so the .dtbo is not being compiled.

pelwell:
Fixed in 4.4. 4.1 is closed now.

Issue #1315, opened 2016-03-01
Configuring I2C pin allocation


6by9:
Just a question on how you'd like stuff done.
I'm wanting to add device tree parameters for setting the I2C GPIO pins, mainly so I can get to the camera via i2c0 and GPIOs 28&29. 

Would you prefer they were added as overrides on the base dt, or by following the approach in https://www.raspberrypi.org/forums/viewtopic.php?f=107&t=120938&p=825883 (following @pelwell 's recommendation in https://www.raspberrypi.org/forums/viewtopic.php?p=779990#p779990)?

I have lifted the overlays from that forum post and have a patch sorted (with appropriate credit to the author), but wanted to confirm that was your preferred approach before creating a PR.
Thanks.

pelwell:
Create a PR with an overlay (or as many as you need). The risks with overlays are non-existent - if you don't use them they do nothing - and they are slightly less "approved" than a mod to the base DTBs would be.

6by9:
Ta. Will push later.

PR #1316, opened 2016-03-02
DT: Add overlays to configure I2C pins

6by9:
Lifted from
https://www.raspberrypi.org/forums/viewtopic.php?f=107&t=120938&p=825883
so not claiming this to be my own work.
Adds overlays i2c0-bcm2708 and i2c1-bcm2708 that allow the pin
allocations for i2c-0 and i2c-1 to be changed.

Done as overlays as discussed under #1315.


pelwell:
I've merged this, after fixing a typo and tweaking the formatting.

6by9:
Thanks Phil.
Is it worth pulling those to the 4.1 and/or other branches? :-)

pelwell:
It will happen.

pelwell:
And it has.

PR #1328, opened 2016-03-06
Enable DMA channels 12-14

### LINKED ISSUE #1327: RPi3: not enough DMA channels for SPI+I2S
I received a report that the Cirrus audio card, which needs both SPI and I2S, doesn't work on the RPi3:
http://openelec.tv/forum/124-raspberry-pi/69870-wolfson-audio-card-support?start=270#159249

Looking at dmesg http://sprunge.us/Fjdc it seems like we have the same problem again like last summer (issue #1110) - the kernel is short of one DMA channel and requesting the second DMA channel for bcm2835-i2s fails

```
[    4.551417] bcm2708-i2s 3f203000.i2s: Missing dma channel for stream: 1
```

On the RPi1/2 when SPI+I2S is used only one DMA channel is free, and since on RPi3 mmc-bcm2835 seems to want 2 DMA channels as well we have a problem.

As a quick workaround using spi-bcm2708 (which doesn't use DMA) instead of spi-bcm2835 on kernel 4.1 seems to work fine, but it looks we don't have a similar option on kernel 4.4 (or for people who need DMA-supported SPI).

Ideally we'd have at least one more DMA channel available, that would solve this problem.

I don't have my RPi3 yet so currently can't test myself but I think the issue should be easily reproducible with "dtparam=spi=on,i2s=on".

ping @popcornmix @pelwell @msperl @notro

HiassofT:
This should fix #1327 

The patches to bcm2835-dma are by @msperl who recently submitted them for upstreaming:
http://archive.arm.linux.org.uk/lurker/message/20160305.105211.40e3c813.en.html

I just merged them into the rpi-4.4.y tree and added the DT changes (drop me a line if you prefer to have the shared-mask and shared-irq settings removed from bcm2708_common.dtsi, they reflect the default values in bcm2835-dma).

I don't have a RPi3 yet, but did some testing on RPi2 with the sdio overlay enabled (so DMA usage should be the same as on RPi3) and the Cirrus Logic audio card (which needs SPI and I2S) and so far everything seems to work fine: a total of 10 DMA channels instead of 7 showing up in /sys/class/dma and both playback and recording with the card worked so the 3 additional DMA channels seem to be working.


pelwell:
Given that these patches are headed upstream but hasn't made it yet, and given that there are workarounds for people who have run out of DMA channels (`dtparam=sd_force_pio` in today's rpi-update release disables the use of DMA by the SD card driver, freeing up two), we'd rather wait until they are accepted upstream than have to subsequently change them, and the author agrees. It isn't that we don't trust the code, it is just that there is already enough patch management to do with each new kernel version.

If the lack of these patches is causing significant problems then feel free to state your case with more feeling.

PR #1340, opened 2016-03-11
Remove I2S config from bt_pins.

clivem:
Remove I2S config from bt_pins. Causes issues with clock alignment when I2S is
used by an external DAC via GPIO header.

### DIFF
--- a/arch/arm/boot/dts/bcm2710-rpi-3-b.dts
+++ b/arch/arm/boot/dts/bcm2710-rpi-3-b.dts
@@ -45,9 +45,9 @@
 	};
 
 	bt_pins: bt_pins {
-		brcm,pins =     <28 29 30 31 43>;
-		brcm,function = <6 6 6 6 4>;   /* alt2:PCM alt0:GPCLK2 */
-		brcm,pull =     <0 0 0 0 0>;
+		brcm,pins = <43>;
+		brcm,function = <4>; /* alt0:GPCLK2 */
+		brcm,pull = <0>;
 	};
 
 	uart0_pins: uart0_pins {
--- a/arch/arm/boot/dts/overlays/pi3-miniuart-bt-overlay.dts
+++ b/arch/arm/boot/dts/overlays/pi3-miniuart-bt-overlay.dts
@@ -29,7 +29,7 @@
 		target = <&uart1>;
 		__overlay__ {
 			pinctrl-names = "default";
-			pinctrl-0 = <&uart1_pins>;
+			pinctrl-0 = <&uart1_pins &bt_pins>;
 			status = "okay";
 		};
 	};


pelwell:
Thanks, Clive.

HiassofT:
Thanks!

@pelwell could you also pick this into 4.1?

pelwell:
@popcornmix Could you do this?

popcornmix:
Done

pelwell:
Thank you.

bonezuk:
Thanks, Clive =D
I've just finished testing the latest 4.4.5+ (plus latest firmware) on both the DAC-Pro and Digi boards with the pi3-disable-bt overlay removed and both sound cards are working as expected. Fantastic stuff. =D
I'm in the process of verifying the 4.1 kernel to see if it to works.

bonezuk:
Confirmed that the fix is working on the 4.1 kernel as well for Digi and DAC-Pro.

clivem:
@bonezuk You're welcome, Stuart. But to be honest, I'm just the idiot that submitted the patch. It's @HiassofT that deserves the credit for the detective work that led to the patch.

PS. Please accept my apology for the outburst in your I2S master/slave issue.

bonezuk:
Cheers to @HiassofT as well.
No problem for the outburst... can understand where your coming from ;-)

BGSputnik:
Cheers, I appreciate all your explanation and comments since it helped me a lot to setup my first system.
I am running the latest Jessie, with the DAC+pro and Kodi v16.1. 
It all works except that I see a failure during booting regarding the Bluetooth.
I am also unable to enable it once the boot up has completed.
I followed all your steps described including Kernel upgrade etc. but something must have disturbed my initial setup hence the booting problem. 
Any guidance as in how to rectify this issue ?
Highly appreciated,

pelwell:
What's in your config.txt?

I'm assuming you have a Pi3 - are you using a 2.5A power supply? Do you ever see the red light flicker off? What does `vcgencmd get_throttled` report?

BGSputnik:
Hello Pelwell, Thanks for your prompt reply.
Yes, I am using the P3 and it was shipped with a 2 Amp power supply. The red LED is constantly on and I dont think I ever had power issues, for some projects I even connect the Sense Hat which should take even more power then just the Kodi app on a "naked" Pi.
When I start up the OS the screen is fine and I only see a purple line within the Kodi app.
The vcgencmd get_throttled shows a value of 0x50005
The red LED is off the green one flashes (most of the time I see however the rainbow square on the right, top side of the screen.
My config:
dtoverlay=hifiberry-dacplus

hdmi_cvt=1024 600 60 3 0 0 0 
hdmi_mode=87
start_x=1
gpu_mem=128

Any ideas ?

pelwell:
The throttled bits say you are underpowered - see [this forum post](https://www.raspberrypi.org/forums/viewtopic.php?f=63&t=147781&start=50#p972790) for an explanation. The rainbow square is telling you the same thing.

Having an insufficient power supply will limit the processor speeds and can cause malfunction - we recommend a 2.5A supply for a Pi3. I think your Bluetooth issue is probably power-related - at startup time the current draw is quite high and I have known marginal supplies to cause problems then.

Issue #324, opened 2013-07-04
3.10 does not compile (implicit declaration... vc_cma*)


moonman:
Hello,

Here's the partial compile log:

drivers/char/broadcom/vc_cma/vc_cma.c: In function 'vc_cma_init':
 drivers/char/broadcom/vc_cma/vc_cma.c:1088:2: error: implicit declaration of function 'create_proc_entry' [-Werror=implicit-function-declaration]
 drivers/char/broadcom/vc_cma/vc_cma.c:1088:20: error: assignment makes pointer from integer without a cast [-Werror]
 drivers/char/broadcom/vc_cma/vc_cma.c:1095:19: error: dereferencing pointer to incomplete type
 drivers/char/broadcom/vc_cma/vc_cma.c: In function 'vc_cma_exit':
 drivers/char/broadcom/vc_cma/vc_cma.c:1134:38: error: dereferencing pointer to incomplete type
 cc1: all warnings being treated as errors
 distcc[13074] ERROR: compile drivers/char/broadcom/vc_cma/vc_cma.c on localhost failed
 make[4]: **\* [drivers/char/broadcom/vc_cma/vc_cma.o] Error 1
 make[3]: **\* [drivers/char/broadcom/vc_cma] Error 2
 make[2]: **\* [drivers/char/broadcom] Error 2
 make[1]: **\* [drivers/char] Error 2
 make: **\* [drivers] Error 2

Full log can be found here: http://archlinuxarm.org:81/builder/in-log/linux-raspberrypi-latest-3.10.0-2-armv6.log.html.gz 

Kernel config is here: https://github.com/archlinuxarm/PKGBUILDs/blob/master/core/linux-raspberrypi-latest/config

CrimsonKnight13:
I can confirm that I have the same issue.

P33M:
https://github.com/torvalds/linux/commit/80e928f7ebb958f4d79d4099d1c5c0a015a23b93

This is why. Function has been removed from upstream.

moonman:
What helped is I disabled bcm_vc_cma driver in config and it compiles, though I'm not too sure what functionality will be lost without this driver. Since default config does not enable this driver I guess it is not that important.

akolesen:
The following patch fixes it. Please, review:
https://github.com/raspberrypi/linux/pull/338

moonman:
Will build the kernel tonight. Thanks for working on this!

PR #1343, opened 2016-03-11
 usb gadget on rpi A

prudy:
It is direct fix for issue 'usb gadget on rpi A #324'.

### DIFF
--- a/arch/arm/boot/dts/overlays/dwc2-overlay.dts
+++ b/arch/arm/boot/dts/overlays/dwc2-overlay.dts
@@ -8,7 +8,7 @@
 		target = <&usb>;
 		#address-cells = <1>;
 		#size-cells = <1>;
-		__overlay__ {
+		dwc2_usb: __overlay__ {
 			compatible = "brcm,bcm2835-usb";
 			reg = <0x7e980000 0x10000>;
 			interrupts = <1 9>;
@@ -21,9 +21,9 @@
 	};
 
 	__overrides__ {
-		dr_mode = <&usb>, "dr_mode";
-		g-np-tx-fifo-size = <&usb>,"g-np-tx-fifo-size:0";
-		g-rx-fifo-size = <&usb>,"g-rx-fifo-size:0";
-		g-tx-fifo-size = <&usb>,"g-tx-fifo-size:0";
+		dr_mode = <&dwc2_usb>, "dr_mode";
+		g-np-tx-fifo-size = <&dwc2_usb>,"g-np-tx-fifo-size:0";
+		g-rx-fifo-size = <&dwc2_usb>,"g-rx-fifo-size:0";
+		g-tx-fifo-size = <&dwc2_usb>,"g-tx-fifo-size:0";
 	};
 };
--- a/drivers/usb/dwc2/gadget.c
+++ b/drivers/usb/dwc2/gadget.c
@@ -2254,6 +2254,7 @@ void dwc2_hsotg_core_init_disconnected(struct dwc2_hsotg *hsotg,
 {
 	u32 intmsk;
 	u32 val;
+	u32 usbcfg;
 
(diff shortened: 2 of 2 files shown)


pelwell:
The overlay commit looks good to me, although the commit message could mention the overlay by name. @P33M Do you approve of the OTG change?

P33M:
@prudy: Can you rebase the patch?

The OTG change looks ok, but this is an upstream driver. Are we happy to accept this out-of-tree?

popcornmix:
We would prefer the dwc patch is submitted upstream and we can pick it once it is accepted.

prudy:
It makes sense, but don't you feedback changes to the upstream?

prudy:
let split dtb and dwc2 changes, so only dtb will apply locally as rpi specific.

pelrun:
@prudy, is "dtoverlay=dwc2,dr_mode=peripheral" all I need to switch this on, or is something extra required? Haven't quite got this working on my A+ yet.

prudy:
Two ways:
- compile dwc2 in gadget-only mode and use 'dtoverlay=dwc2' or
- compile dwc2 in dual-mode and use 'dtoverlay=dwc2,dr_mode=peripheral'

In any case this dtb patch is needed but also kernel patch to gadget. The latter is on its way in upstream or you can take it from my branch.

pelrun:
I had already recompiled the dwc2 driver in dual mode (with all the patches, obviously since I'm posting on this pr), and the overlay is already in the rpi tree, and it's behaving as expected when I configure some libcomposite devices, but nothing happens when I connect to the pi with an a-a cable. Both ends act as if I'd not connected the cable at all.

prudy:
If nothing happens it means the patches are not active. I'd check:
- using ko modules (so not built-in) do not forget to modprobe the right module
- the rpi firmware is fresh enough to handle DT
- kernel is using DT, dtb are built and put onto sdcard overlay/
- the config.txt is set to use DT (device_tree=bcm2708-rpi-b.dtb)
- nothing else is using the USB
- with the above in place, use 'dtdebug=1' in config, so you can use vcdbg to check if dr_mode is applied correctly
  (or even easier, compile dwc2 in gadget-only mode, so it is in the only right mode)

See this post with kernel configuration:
https://www.raspberrypi.org/forums/viewtopic.php?f=29&t=129653

In the startup log you should see this:
"Configuration mismatch. dr_mode forced to"

and this:
[    2.307172] g_serial gadget: Gadget Serial v2.4
[    2.314621] g_serial gadget: g_serial ready
[    2.324993] dwc2 20980000.usb: bound driver g_serial

eventually this when plugging in the cable:
[   37.540037] dwc2 20980000.usb: new device is high-speed
[   37.693650] dwc2 20980000.usb: new device is high-speed
[   37.791228] dwc2 20980000.usb: new address 4

pelrun:
As far as I can tell, everything is running as intended (no "mismatch" message because I have dr_mode set correctly in config.txt) and I've added debug messages to prove that the force dev bits are being preserved... but inserting the cable (I've tried both a home-soldered one and a purchased one) does nothing. At this point I think I'll probably wait until this makes it into the official Pi kernel before trying again.

ED6E0F17:
Linux 4.5 has an extra patch that will check the hardware capabilities of the model "A", and block otg mode:

EDIT: Confirmed this patch is not a problem, and gadget mode is working on model A

SloMusti:
Any idea how to most effectively enable OTG mode on A+ with latest linux and work around the OTG mode block?

ED6E0F17:
As far as I know it should work with "dtoverlay=dwc2","dr_mode=peripheral" from linux 4.6 and above.

The patch I mentioned above may only check what the chip can do, not which cable is connected but I broke the sdcard slot on my Model A, so I replaced that with a Zero and cannot actually test what might happen on an A+.

PR #1365, opened 2016-03-23
Add I2C_MUX and I2C_MUX_PCA954x and overlay to configure muxes

6by9:
Adds kernel configs for I2C muxes and a dt overlay for PCA9548
that adds the 8 muxed I2C buses and mux device.

i2c-mux.ko is 7898 bytes.
i2c-mux-pca954x.ko is 8668 bytes.
So neither is large.

Overlay loads PCA9548A on address 0x70 (can be addressed from 0x70 to 0x77) to add 8 extra I2C busses. I can't see an easy way to make the address configurable via a dt param and get it to modify the mux@70 section. Happy to change if there is an easy way.
I haven't worked out if there's a way to set the base /dev/i2cN node number, but not sure if that is necessary or not.


arch/arm/boot/dts/overlays/README
```diff
@@ -360,6 +360,11 @@ Params: i2c_gpio_sda            GPIO used for I2C data (default "23")
                                 (default "2" = ~100kHz)
 
 
+Name:   i2c-mux-pca9548a
+Info:   Adds support for an NXP PCA9548A I2C multiplexer on i2c_arm
+Load:   dtoverlay=i2c-mux-pca9548a
+Params: <None>
+
```
pelwell, line 368: You're going to need an extra blank line here. Sorry.

6by9, line 368: No problem.

6by9:
For context, this came out of forum thread https://www.raspberrypi.org/forums/viewtopic.php?f=44&t=140054

Extra blank line added and repo force updated.

pelwell:
To parameterise the address you could add the following (untested) bit of dts:

```
    __overrides__ {
        addr = <&i2cmux>,"reg:0";
    };
```

This would have the affect of changing the address, but without changing the name of the node. It's not ideal, but it's the best you can do for now.

6by9:
Thanks. I'd seen I could do that, but the mismatch caused me to hesitate. Do you think it's worth adding?

6by9:
Is it worth making this overlay always turn on i2c_arm too? I'm not sure what it'll do without the parent node (trying it now...)

pelwell:
> Do you think it is worth adding?

Yes, I do. When I get round to revamping the overrides syntax we can tidy up the result, but it would be good to have the facility there now.

> Is it worth making this overlay always turn on i2c_arm too?

Yes to this too. It just removes another source of user error.

It looks like /dev/i2c-\* numbering is working already, but if you want more control I've a feeling that the global I2C bus ID is derived from the Device Tree aliases, so try adding `i2c<n> = &<label on new i2c bus node>;`.

6by9:
> > Do you think it is worth adding?
> 
>  Yes, I do.

Done. You got it right even without testing.

It looks like i2c_arm gets enabled automatically by the current overlay, although I'm not too sure on the mechanism for how.

The numbering is working for me in that it takes the next available node numbers (3-10). I was trying to head off the weird situations where you want to avoid particular nodes. I'm not quite following you on the aliases section, so I think I'll skip that bit for now.

pelwell:
I've stripped the TABs from the README and merged it manually.

pelwell:
I also moved the changes to bcm2835_defconfig to bcmrpi_defconfig instead, and removed the whitespace at the end of the overlay.

6by9:
Thank you :-)
I can't say I fully understand device tree, but that wasn't too painful. I had a quick look at the alias stuff but it didn't fall into place, so it will have to remain as just assuming the next available /dev/i2cN numbers.

6by9:
Just in case I come back to it later, it looks like you are spot on with using aliases for naming based on http://www.spinics.net/lists/linux-i2c/msg18901.html
I obviously just got the incantation wrong.

PR #1366, opened 2016-03-23
Add support for the DS1339 I²C RTC chip in the i2c-rtc DeviceTree overlay.

nboullis:


pelwell:
I've shortened the commit message and removed the unicode character, but otherwise I've pushed the commit unmodified. Thanks.

nboullis:
Thanks @pelwell !

PR #1393, opened 2016-04-05
V4L2 driver updates

6by9:
A few updates to the V4L2 driver for the Pi Camera.
- Correct the support for ISO control.
- Add option to specify the /dev/videoN node number.
- Support for multiple cameras on the CM.
- Support for specifying where the overlay is on the screen.
- Compliance test failure.

A quick review by others would be appreciated, just to make sure I haven't missed something silly in the init/exit loops. (I'm afraid I haven't done as much testing of that as I perhaps ought to have done)

### DIFF
--- a/drivers/media/platform/bcm2835/bcm2835-camera.c
+++ b/drivers/media/platform/bcm2835/bcm2835-camera.c
@@ -1825,7 +1825,7 @@ void bcm2835_cleanup_instance(struct bm2835_mmal_dev *dev)
 static struct v4l2_format default_v4l2_format = {
 	.fmt.pix.pixelformat = V4L2_PIX_FMT_JPEG,
 	.fmt.pix.width = 1024,
-	.fmt.pix.bytesperline = 1024,
+	.fmt.pix.bytesperline = 0,
 	.fmt.pix.height = 768,
 	.fmt.pix.sizeimage = 1024*768,
 };
--- a/drivers/media/platform/bcm2835/bcm2835-camera.c
+++ b/drivers/media/platform/bcm2835/bcm2835-camera.c
@@ -666,6 +666,30 @@ static struct vb2_ops bm2835_mmal_video_qops = {
 	IOCTL operations
    ------------------------------------------------------------------*/
 
+static int set_overlay_params(struct bm2835_mmal_dev *dev,
+			      struct vchiq_mmal_port *port)
+{
+	int ret;
+	struct mmal_parameter_displayregion prev_config = {
+	.set = MMAL_DISPLAY_SET_LAYER | MMAL_DISPLAY_SET_ALPHA |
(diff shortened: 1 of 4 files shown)


drivers/media/platform/bcm2835/bcm2835-camera.c
```diff
@@ -54,6 +54,10 @@ int bcm2835_v4l2_debug;
 module_param_named(debug, bcm2835_v4l2_debug, int, 0644);
 MODULE_PARM_DESC(bcm2835_v4l2_debug, "Debug level 0-2");
 
+int video_nr = -1;
```
popcornmix: static?

drivers/media/platform/bcm2835/bcm2835-camera.c
```diff
@@ -54,9 +56,10 @@ int bcm2835_v4l2_debug;
 module_param_named(debug, bcm2835_v4l2_debug, int, 0644);
 MODULE_PARM_DESC(bcm2835_v4l2_debug, "Debug level 0-2");
 
-int video_nr = -1;
-module_param_named(video_nr, video_nr, int, 0644);
-MODULE_PARM_DESC(video_nr, "videoX start number, -1 is autodetect");
+#define UNSET (-1)
+static int video_nr[] = {[0 ... (MAX_BCM2835_CAMERAS - 1)] = UNSET };
```
popcornmix, line 60: Ah - static!

drivers/media/platform/bcm2835/bcm2835-camera.c
```diff
@@ -1792,8 +1895,10 @@ static int __init bm2835_mmal_init(void)
 free_dev:
 	kfree(dev);
 
-	v4l2_err(&dev->v4l2_dev,
-		 "%s: error %d while loading driver\n",
+	for ( ; camera > 0; camera--)
+		bcm2835_cleanup_instance(gdev[camera]);
```
popcornmix, line 1935: Should this cleanup gdev[0]?
Should you set gdev[camera] = NULL; ?

6by9, line 1935: Possibly. 
This is failure to load, so won't that unload the module due to the failure and therefore deallocate gdev.

Probably safest to reset it to NULL anyway - I'll update it.

drivers/media/platform/bcm2835/bcm2835-camera.c
6by9: Perhaps same argument should be applied to bcm2835_v4l2_debug above.  As you've seen, I've copy/pasted from somewhere else in the next patch and it has become static :-)

6by9:
@popcornmix I think that addresses the things you raised, and adds an extra patch making the module parameters static variables, except bcm2835_v4l2_debug which is used in multiple files).

PR #1396, opened 2016-04-07
Enable hid-betopff module

jochenberger:
Add force feedback support for Betop based devices
https://github.com/raspberrypi/linux/blob/rpi-4.1.y/drivers/hid/hid-betopff.c


pelwell:
We're usually happy to add config options that just cause modules to be built, but you should add it to bcm2709_defconfig as well. If you modify your patch using `git commit --amend` you can use `git push -f`-to update your PR, so that there is one commit instead of two.

jochenberger:
PR updated

pelwell:
I see no reason not to merge this - it just builds a 7.1 KB module. @popcornmix ?

PR #1406, opened 2016-04-14
Add support for Digital Dreamtime Akkordion (using OEM IQAudIO DAC module)

clivem:
This pull request is to add support for the Digital Dreamtime Akkordion music player, IQAudIO edition, which is "powered" by an OEM IQAudIO DAC module.

(2nd attempt. Remove the quotes. Closed the first.)

### DIFF
--- a/arch/arm/boot/dts/overlays/Makefile
+++ b/arch/arm/boot/dts/overlays/Makefile
@@ -13,6 +13,7 @@ ifeq ($(CONFIG_ARCH_BCM2835),y)
 endif
 
 dtbo-$(RPI_DT_OVERLAYS) += ads7846.dtbo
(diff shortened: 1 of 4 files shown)


pelwell:
`git push -f` is your friend. But otherwise it looks OK.

clivem:
LOL. I won't tell you why it actually all went wrong. You might start thinking I'm the sort of fool that would  type "rm -rf /" without giving it much thought.......

pelwell:
Luckily your work was backed up on github...

clivem:
LOL. Which was a problem in itself, I couldn't access github from git cmdline. Kept returning error codes, I've never seen before.....

PR #1412, opened 2016-04-16
w1: add ability to change temperature sensor precision

x29a:
The DS18B20 temperature sensor is quite popular with the rpi and yet it is not possible to change the precision with which the sensors are converting their temperature values.

Since some sensors come preconfigured with a low precision, people are stuck to e.g. 9bit when using the rpi when 12bit are possible.

This has been tested on an rpi1 running at 4.1.13+ with a DS18B20 sensor. Other sensors support the commands as well but are disabled for now because of lack of testability.

### DIFF
--- a/Documentation/w1/slaves/w1_therm
+++ b/Documentation/w1/slaves/w1_therm
@@ -33,7 +33,15 @@ temperature conversion at a time.  If none of the devices are parasite
 powered it would be possible to convert all the devices at the same
 time and then go back to read individual sensors.  That isn't
 currently supported.  The driver also doesn't support reduced
-precision (which would also reduce the conversion time).
+precision (which would also reduce the conversion time) when reading values.
+
+Writing a value between 9 and 12 to the sysfs w1_slave file will change the 
+precision of the sensor for the next readings. This value is in (volatile) 
+SRAM, so it is reset when the sensor gets power-cycled.
+
+To store the current precision configuration into EEPROM, the value 0
+has to be written to the sysfs w1_slave file. Since the EEPROM has a limited
+amount of writes (>50k), this command should be used wisely.
 
 The module parameter strong_pullup can be set to 0 to disable the
 strong pullup, 1 to enable autodetection or 2 to force strong pullup.
--- a/drivers/w1/slaves/w1_therm.c
+++ b/drivers/w1/slaves/w1_therm.c
@@ -217,7 +217,7 @@ static inline int w1_therm_eeprom(struct device *device)
 	if (ret != 0)
 		goto post_unlock;
 
-	if(!sl->family_data) {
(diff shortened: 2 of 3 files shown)


popcornmix:
The 4.1 kernel tree is not longer used.
rpi-update and future apt-get updates will come from 4.4.
Please create a new PR for 4.4 branch.

flaggz:
how exactly do I write the resolution bit in the w1_slave after the x29a commits?

x29a:
Note, this change got accepted upstream with 4.7, i also updated the [docs](https://github.com/raspberrypi/linux/blob/rpi-4.7.y/Documentation/w1/slaves/w1_therm#L38), so its basically:
`echo 12 > w1_slave`
for 12 bits, same for 9, 10 and 11. Other values are not supported.

flaggz:
i have changed to 11bit precision by plugging the sensor on arduino and when plugged on raspberry the resolution is still 12bit.
does the RPI writes EEPROM on each boot?

x29a:
im not sure if the arduino method writes the precision to SRAM or EEPROM, you would have to elaborate but that is out of the scope of this PR.

In order to save the precision of 11bit permanently with the w1 changes, you would have to write a 0 to the w1_slave, so something like
`echo 11 > w1_slave && echo 0 > w1_slave`

then check again with arduino and after unplugging the sensor/reboot the device.

Please note again, the changes were introduced with 4.7, which version are you running? self compiled with the patch applied?

flaggz:
It can write both SRAM and EEPROM. This is the important instruction:
  reg_cmd = 0x5F;  // Configuration Register (resolution) 7F=12bits 5F=11bits 3F=10bits 1F=9bits
  ds.reset();
  ds.select(addr);
  ds.write(0x4E); // Write scratchpad command
  ds.write(0);// TL data
  ds.write(0);// TH data
  ds.write(reg_cmd); 
  ds.reset(); // This "reset" sequence is mandatory
  ds.select(addr);// it allows the DS18B20 to understand the copy scratchpad to EEPROM command
  ds.write(0x48); // Copy Scratchpad command

Right now I’m using standard 4.4. I would prefer to not use dev version if I can.
Just asking if before the patch Raspberry is doing anything at SRAM or EEPROM level when plugging devices.

flaggz:
updated to 4.9 with $ sudo rpi-update BRANCH=next
rebooted
/sys/bus/w1/devices/28-80000008152c $ sudo echo 11 > w1_slave
-bash: w1_slave: Permission denied

pelwell:
The output of the echo command running as root is received by the shell running as user pi, and the shell is trying to write it to w1_slave. You need the whole pipeline run as root:
```
sudo sh -c "echo 11 > w1_slave"
```

PR #1415, opened 2016-04-17
Update tlv320aic23.c

alisabri:
"NULL" s should be NULL


popcornmix:
This is an upstream linux file, so the patch ought to be submitted upstream:

Ruffio:
@alisabri @popcornmix could one of you submit the bug at kernel.org? I don't have enough knowledge/technical details to do so :-(

pelwell:
Already accepted upstream - back-ported to rpi-4.9.y.

PR #1419, opened 2016-04-20
vc4 updates for 4.4

anholt:
Here's a giant merge of post-4.4 development for vc4.  Just touch-tested with a pi2 so far (X, glxgears).  Note that I'm using the overlay for downstream merge testing now, which I wasn't before.

The only commit I'm not really pleased with is the CMA move in the overlay.  Also, it would be really nice to be able to customize the CMA area based on the board in use so that we could enable pi1 with a smaller CMA area, but I don't know how we would do that with a single overlay file.


popcornmix:
I get a WARN on boot.

```
[   16.118823] [<80018724>] (unwind_backtrace) from [<80014058>] (show_stack+0x20/0x24)
[   16.118836] [<80014058>] (show_stack) from [<80320044>] (dump_stack+0xd4/0x118)
[   16.118849] [<80320044>] (dump_stack) from [<80025300>] (warn_slowpath_common+0x98/0xc8)
[   16.118861] [<80025300>] (warn_slowpath_common) from [<800253ec>] (warn_slowpath_null+0x2c/0x34)
[   16.118899] [<800253ec>] (warn_slowpath_null) from [<7f133a9c>] (drm_update_vblank_count+0x188/0x368 [drm])
[   16.118972] [<7f133a9c>] (drm_update_vblank_count [drm]) from [<7f133cec>] (vblank_disable_and_save+0x70/0x80 [drm])
[   16.119034] [<7f133cec>] (vblank_disable_and_save [drm]) from [<7f133d70>] (vblank_disable_fn+0x74/0xa0 [drm])
[   16.119072] [<7f133d70>] (vblank_disable_fn [drm]) from [<80081bcc>] (call_timer_fn+0x40/0x1a0)
[   16.119084] [<80081bcc>] (call_timer_fn) from [<80081eec>] (run_timer_softirq+0x1c0/0x324)
[   16.119093] [<80081eec>] (run_timer_softirq) from [<80029040>] (__do_softirq+0x18c/0x3cc)
[   16.119103] [<80029040>] (__do_softirq) from [<80029620>] (irq_exit+0xdc/0x140)
[   16.119115] [<80029620>] (irq_exit) from [<80071090>] (__handle_domain_irq+0x70/0xc4)
[   16.119129] [<80071090>] (__handle_domain_irq) from [<80010a74>] (handle_IRQ+0x2c/0x30)
[   16.119140] [<80010a74>] (handle_IRQ) from [<8000951c>] (bcm2836_arm_irqchip_handle_irq+0xb8/0xbc)
[   16.119152] [<8000951c>] (bcm2836_arm_irqchip_handle_irq) from [<805b78c4>] (__irq_svc+0x44/0x5c)
```

http://paste.debian.net/439666/

popcornmix:
We suspect the cma issue is because device tree by default loads at top of non highmem memory (e.g. ends at 768M).
`device_tree_address=0x1f400000`
would load the device tree at 500M which avoids the issue.
We should be able to make this happen automatically, so the CMA commit can be dropped if preferred.

anholt:
I don't see any reason not to include the CMA commit -- the size is the only thing we care about, and the base just has to be whatever works from among 0, 256, 512, or 768.

anholt:
The vblank warning has been around for a while, not special to this merge.  I think it's that we don't have a frame count/current scanline register from HVS, so I need to go test and fix this path that nobody else uses.

popcornmix:
Is there any disadvantage to allocating more to CMA?
I'm aware that some games need more than 256M.
What is the disadvantage to setting, say, 512M@256M?
I understand that unused CMA can still be used for user pages - is there a disadvantage?
Also where does the 256M alignment requirement come from? Something in the v3d hardware?

anholt:
The V3D hardware, to a rough approximation, can't have its addresses cross a 256MB boundary due to a HW bug.  Thus we choose 256MB for size and align it to a 256MB boundary somewhere, to get the maximum size possible.

popcornmix:
I'm not getting a console framebuffer with overlay enabled.
http://paste.debian.net/439877/

```
[    9.396923] bcm2835-clk 3f101000.cprman: pllh: rate out of spec: 0 vs (600000000, 3000000000)
[    9.396949] Division by zero in kernel.
```

Any thoughts? Booting to desktop does work.

ED6E0F17:
It would be better for DRM & DRM_VC4 to be built-in to the kernel, rather than as modules.

```
CONFIG_DRM=m
CONFIG_DRM_UDL=m
CONFIG_DRM_VC4=m
```

anholt:
@popcornmix I'm assuming you have an HDMI monitor plugged in?  Is the firmware configuring HDMI at boot?  Anything in config.txt other than the overlay and avoid_warnings?  Asking this to see if I can replicate your environment.

The divide in set_rate is of the rate being passed in, which would suggest that the mode->clock is 0?  Next step would be to put a drm_mode_debug_printmodeline(mode) near the top of vc4_hdmi_encoder_mode_set().  Also, setting drm.debug=0x14 on kernel command line should help with modeset debugging.

popcornmix:
Yes, monitor is plugged into HDMI. Firmware chooses:
`state 0x12000a [HDMI CEA (16) RGB lim 16:9], 1920x1080 @ 60.00Hz, progressive`
Interestingly I do get the framebuffer console at work, but not at home.
I have exactly the same monitor at work and home.
The monitor at home came up with firmware "rainbow" splash screen, but went to power save mode as kernel booted. 
config.txt is clean on both devices apart from the needed overlay and avoid_warnings lines.
I'll have to test more at home as I'm not seeing the issue at work now.

PR #1432, opened 2016-04-24
ARM: bcm2708: add generic fbtft overlay with reset and dc gpios

msperl:
add a generic fbtft overlay with reset and dc-gpios.


pelwell:
The README must not use TABs, and lines should not exceed 80 characters. Capitalising the first word of the parameter descriptions would also match the others.

Otherwise it looks OK.

msperl:
General kernel policy is to use tabs wherever possible, so why do we beg to differ?

pelwell:
To make it easy for the dtoverlay/dtparam commands to locate the information for specific overlays and parameters.

pelwell:
Closing for lack of activity. Fix the README and resubmit against 4.9 if still wanted.

PR #1434, opened 2016-04-24
drm/vc4: Add DT parameters to control CMA usage

pelwell:
Platforms with less memory available may struggle to find 256MB aligned
on a 256MB boundary. Add parameters to allow less to be requested.
Example: dtoverlay=vc4-kms-v3d,cma-128

See: https://github.com/raspberrypi/linux/pull/1431

Signed-off-by: Phil Elwell phil@raspberrypi.org

### DIFF
--- a/arch/arm/boot/dts/overlays/README
+++ b/arch/arm/boot/dts/overlays/README
@@ -1023,8 +1023,12 @@ Name:   vc4-kms-v3d
 Info:   Enable Eric Anholt's DRM VC4 HDMI/HVS/V3D driver. Running startx or
         booting to GUI while this overlay is in use will cause interesting
         lockups.
-Load:   dtoverlay=vc4-kms-v3d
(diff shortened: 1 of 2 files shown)


msperl:
If it works - I am happy with that...

kukabu:
why is it necessary? why not add the 'cma=ХХХ' via the /boot/cmdline.txt?

pelwell:
It isn't necessary, but doing it this way is neater because you only need to edit one file, and it doesn't require the extra knowledge that CMA is controlled from the kernel command line.

It also isn't necessary to use a DT overlay to add an I2C Real Time Clock, but some people find:

```
dtoverlay=i2c-rtc,ds1307
```

preferable to:

```
echo ds1307 0x68 > /sys/class/i2c-adapter/i2c-1/new_device
```

PR #1435, opened 2016-04-24
Added 1-wire slave device support for 3A 2100H

saper-2:
The 3A 2100H device is a Chinese clone of DS2413 - it uses a own family code 0x85 so the standard DS2413 module does not recognize this chip. I have created a driver for this chip - basically it's copy of ds2413 code with changed device id and renamed functions so it work with 3A 2100H.

Now I made all changes under linux (on Rpi3) and I compiled the kernel from my repo so I'm pretty sure there shouldn't be any problems. To enable my driver, generate `.config` file by `bcmrpi_defconfig` or `bcm2709_defconfig`, then add to the `.config` line: `CONFIG_W1_SLAVE_2100H=m` or use `make menuconfig` and enable my device under 1-wire slave device drivers.
I have described also whole process of adding my module to cloned RaspberryPi kernel repo at https://github.com/saper-2/rpi-ds2413-2100h .

I'm sorry for previous pull request (windows messed up cloned git repo - win have problem with "aux.c/aux.h" files :disappointed:  ).

### DIFF
--- a/drivers/w1/slaves/Kconfig
+++ b/drivers/w1/slaves/Kconfig
@@ -38,6 +38,12 @@ config W1_SLAVE_DS2413
 	  Say Y here if you want to use a 1-wire
 	  DS2413 Dual Channel Addressable Switch device support
 
+config W1_SLAVE_2100H
+	tristate "Dual Channel Addressable Switch 0x85 family support (clone DS2413: 3A 2100H)"
+	help
+	  Say Y here if you want to use a 1-wire
+	  3A 2100H (DS2413 clone) - Dual Channel Addressable Switch device support
+
 config W1_SLAVE_DS2406
 	tristate "Dual Channel Addressable Switch 0x12 family support (DS2406)"
 	select CRC16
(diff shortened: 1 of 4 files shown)


pelwell:
This code isn't Raspberry Pi-specific, so it should be submitted upstream:

Having said that, I doubt it will be accepted upstream because, apart from the function name changes, there is only one difference - the ID. Try editing the existing ds2413 driver and adding:

```
MODULE_ALIAS("w1-family-" __stringify(W1_FAMILY_2100H));

...

static struct w1_family w1_family_85 = {
    .fid = W1_FAMILY_2100H,
    .fops = &w1_f83_fops,
};

static int __init w1_f3a_init(void)
{
    int ret;

    ret = w1_register_family(&w1_family_3a);
    if (ret == 0)
        ret = w1_register_family(&w1_family_85);
    return ret;
}
```

That would make a much smaller patch, with no duplications, so is more likely to be accepted.

6by9:
This had already been discussed on the forums at https://www.raspberrypi.org/forums/viewtopic.php?f=45&t=132705&p=895533 with Phil's approach suggested as correct (although I see I made a typo and had 2 returns in one function).

pelwell:
My patch was intended as an experiment, and it lacked the matching unregister call, but between us I think we have it covered.

Ruffio:
@saper-2 is this still relevant or has it been fixed?

saper-2:
It is not issue, but as mentioned this should be sent to kernel.org , unfortunately doing this is really too much for me (I read kernel.org patch submission)... *(Xmas are coming so I might think about it - if I won't forget :stuck_out_tongue: )*
I'm closing it, if somebody will want to add support for 2100H then he/she can go to my repo and check what & how.

PR #1436, opened 2016-04-25
dts: add overlay for pitft22

agherzan:
Add the pitft22 overlay from adafruit Adafruit-Pi-Kernel-o-Matic repo:
    https://github.com/adafruit/Adafruit-Pi-Kernel-o-Matic

Signed-off-by: Petter Mabäcker petter@technux.se
Signed-off-by: Andrei Gherzan andrei@resin.io


pelwell:
Please include an entry in the overlays/README file. You should be able to adapt this one:

```
Name:   pitft28-capacitive
Info:   Adafruit PiTFT 2.8" capacitive touch screen
Load:   dtoverlay=pitft28-capacitive,<param>=<val>
Params: speed                   Display SPI bus speed

        rotate                  Display rotation {0,90,180,270}

        fps                     Delay between frame updates

        debug                   Debug output level {0-7}
```

The README rules are: No TABs, no lines longer than 80 characters, two blank lines between entries.

pelwell:
Ah, this is rpi-4.1.y, which is essentially closed. We won't be making any more releases from it, but if you have a good reason for wanting the commit in this branch then we will consider it.

agherzan:
I understand. Didn't know that 4.1 was closed. We need it for people using 4.1 and want support for pitft 22 in meta-raspberrypi. We will keep that patch in the repository then for this specific version.

PR #1438, opened 2016-04-25
Rpi 4.6.y

Twitch0815:
Requesting xpad be added to kernel so that Retropie can correctly use Xbox one controllers. 

https://github.com/recalbox/recalbox-buildroot/pull/250

When trying to install on current kernel 
Your kernel headers for kernel 4.1.19-v7+ cannot be found at
/lib/modules/4.1.19-v7+/build or /lib/modules/4.1.19-v7+/source.


popcornmix:
Is this meant to be a pull request?

Is it meant to be applied to 4.4, or 4.6?

ED6E0F17:
The Commit message links to a patch that has already been merged in 4,4, to enable this device:

 https://github.com/raspberrypi/linux/blob/rpi-4.4.y/drivers/input/joystick/xpad.c#L127

The OP may be asking for "CONFIG_JOYSTICK_XPAD=m"

Edit:  "CONFIG_JOYSTICK_XPAD=m" is already included with the 4.4 defconfig.

ED6E0F17:
As far as I can tell this is a request to add xpad support to linux 4.1, but 4.1 already has xbox1 xpad support. 

Recalbox need to update their build system to use a newer kernel, rather than patching an older one.

popcornmix:
Closing as this doesn't make sense to me.
Feel free to re-open after rebasing and with an updated description.

PR #1446, opened 2016-05-01
Fixed the compile error with spi_bcm2708 driver

allenh1:
Needed the spi_bcm2708 driver to work for a CS project. It is now working. I also cleaned up some compiler warnings.


drivers/spi/Kconfig
```diff
@@ -87,10 +87,18 @@ config SPI_AXI_SPI_ENGINE
 	tristate "Analog Devices AXI SPI Engine controller"
 	depends on HAS_IOMEM
 	help
-	  This enables support for the Analog Devices AXI SPI Engine SPI controller.
-	  It is part of the SPI Engine framework that is used in some Analog Devices
-	  reference designs for FPGAs.
-
+	  This enables support for the Analog Devices AXI SPI Engine SPI
```
popcornmix: You shouldn't be editing these lines

popcornmix:
Why the changes to bcm2835-sdhost.c?

What was the issue with upstream spi driver?

allenh1:
The changes to `bcm2835-sdhost.c` just resolve some warnings -- I can remove those if you want.

As for the upstream spi driver, the upstream driver was not there. After dragging in the old source, a compiler error was introduced. It has been fixed, and the spi driver has since been tested.

allenh1:
Caught up with current commits, removed unnecessary changes.

allenh1:
Caught up with rasberrypi/linux/rpi-4.6.y

Ruffio:
@allenh1 why is this targeted to 4.6.y when 4.4.y is currently in use and there is no plan to switch to 4.6.y within short time?

allenh1:
Because I was using 4.6.y and the driver was not in 4.6.y. I believe it was in 4.4.y, last I checked.

On Aug 29, 2016 4:43 PM, Rasmus Christiansen notifications@github.com wrote:

@allenh1https://github.com/allenh1 why is this targeted to 4.6.y when 4.4.y is currently in use and there is no plan to switch to 4.6.y within short time?

##

Ruffio:
@allenh1 I think that only PR to 4.4.y would be accepted, but I could be mistaken.

popcornmix:
We may accept PRs to newer branches where that makes sense (e.g. upstream changes mean that only the newer branch is affected).

4.6 is already EOL upstream, so we are unlikely to continue to support it - 4.7 would be preferred.

But the real question is why spi-bcm2708 is required. This is the deprecated downstream SPI driver. The intention is that everyone uses the upstream spi-bcm2835 driver. If there are problems with the upstream driver then it would be preferable to make that driver work, rather than resurrecting the old driver.

Ruffio:
@allenh1 is this PR going to be 'fixed' according to @popcornmix comments? If not, then I don't think it will ever be accepted and therefore it should be 'closed'.

allenh1:
I'll fix it as requested. Just been putting off the rebase of my life XD

On Dec 6, 2016 11:23 AM, Rasmus Christiansen <notifications@github.com> wrote:

@allenh1<https://github.com/allenh1> is this PR going to be 'fixed' according to @popcornmix<https://github.com/popcornmix> comments? If not, then I don't think it will ever be accepted and therefore it should be 'closed'.

-

pelwell:
Closing for lack of activity and the fact that we are now on 4.9.

allenh1:
That's fine

PR #1452, opened 2016-05-02
switch to clockmanager by default

msperl:
note that his requires the patchset clkmgr-backport-critical (https://github.com/raspberrypi/linux/pull/1451) to be applied to work 100% for all tested use-cases.


pelwell:
We are now using clockmgr on 4.9 and above.

PR #1457, opened 2016-05-06
Rpi 4.4.y: HDMI hotplug detection, overlay cleanup

anholt:
I finally got a rework of the overlay to work (mostly just irritating debug from overlays not validating well and silently failing to apply at runtime), so I could get HDMI hotplug detection fixed.  The commits on this branch should apply to 4.5.y as well.

### DIFF
--- a/arch/arm/boot/dts/bcm2708_common.dtsi
+++ b/arch/arm/boot/dts/bcm2708_common.dtsi
@@ -1,4 +1,7 @@
+#include "dt-bindings/clock/bcm2835.h"
 #include <dt-bindings/clock/bcm2835-aux.h>
+#include "dt-bindings/power/raspberrypi-power.h"
+#include "dt-bindings/gpio/gpio.h"
 #include "skeleton.dtsi"
 
 / {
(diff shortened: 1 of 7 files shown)


popcornmix:
@pelwell are you happy with DT changes?

arch/arm/boot/dts/bcm2708_common.dtsi
```diff
@@ -261,6 +282,24 @@
 			status = "disabled";
 		};
 
+		pixelvalve2: pixelvalve@7e807000 {
+			compatible = "brcm,bcm2835-pixelvalve2";
+			reg = <0x7e807000 0x100>;
```
pelwell: I'd like a `status = "disabled";` here.

pelwell:
Yes, apart from the fact that the pixelvalve2 node is enabled by default. The overlay already contains the necessary status="okay" to enable it, so I just want a single line added.

anholt:
Re-pushed the branch with pixelvalve fixed up.

pelwell:
Looks good to me.

PR #1459, opened 2016-05-07
Add ASoC support for boards/HATs using ESS Sabre ES9023 DAC

clivem:
Phil/Dom, Don't actually want this pulled, (at least not without review/comment first by Berry guys)...

This is an attempt at code re-use by having common board driver (in this case for the ESS Sabre ES9023 DAC variants), and using "compatible" and dt for ALSA card/dai/dai_stream names, rather than everyone cut and pasting a complete, existing driver, to add a new board. 

Manufacturers not bothered about having company names in ALSA strings just use the generic (es9023-dac) overlay. Those bothered, (and I have added examples for my DD Akkordion and HB DAC+ Light), create their own overlay, set "names" in dt, and add a "compatible" to es9023_dac card driver. 

Most of the current ES9023 boards "piggyback" on the hifiberry_dac/pcm5102a combo. @bonezuk and @hifiberry, please let me know which way you want to go. If you want to stick with using your "old" dac driver and pcm5102a codec combo for your ES9023 DAC+ Light boards, I will drop the specific overlay for you from the pull request. But I wanted you to take a look at it.


hifiberry:
@clivem : The approach totally makes sense. However, at the moment I would like to stick with the PCM5102 driver with the DAC+ Light.
Did you think about creating a fully "generic" driver that supports output/input and sample rates all configurable by device tree? E.g. this might be interesting for people that uses old 44.1kHz DACs.

A generic module with input/output capabilities would also be helpful to connect I2S microphones without adding another driver.

clivem:
> However, at the moment I would like to stick with the PCM5102 driver with the DAC+ Light.

@hifiberry No problem with me. I didn't want you to feel left out! ;) I've removed the overlay for hifiberry-dacpluslight from this patch set.

> Did you think about creating a fully "generic" driver that supports output/input and sample rates all configurable by device tree? E.g. this might be interesting for people that uses old 44.1kHz DACs.

I think I already have exactly what you are describing. It's called the NOS driver, because it's what I use to drive the obsolete stuff, like TDA1541, AD1862/5, etc. etc. 

I have a patch set of about 40 patches, that are I2S audio related. Starting with the dedicated driver for the ES9023, was starting with what I thought was the least contentious patch set, to show the "approach". I have 3 stages of rationalising the PCM5xxx situation, which I suspect I'm going to have trouble getting both you and Gordon on-board with.

I do want to point one thing out to you. That you are using the pcm5102a codec driver for both pcm5102a and es9023, and that shouldn't become the reason for not wanting to move forward on 384k support. The conflict here being, that the datasheet for es9023 gives 192k as PCM max, while all PCM5xxx family support 384k. That is the reason I have made the user specify a dt param to get 384k support from es9023 codec.... Not every 9023 chip can do it, and I have no idea what value OSC all the manufacturers are using for their ES9023 boards, mostly 50M I suspect. But there is no reason not to want to support 352k8/384k "out-of-the-box" for all the PCM5's. The Pi is capable. The DAC codec is capable. These sort of things become problems, or at least things you need to think about, at the point you are using a codec driver, for something other than the intended hardware.

hifiberry:
We would like to support you with all the patches. I guess, it would be good to have a better idea where your are heading to. Send me a message with your personal contact details. Then we can have a short call and see how we can help each other in the best way.

pelwell:
The codec is a generic component and should be pushed upstream. The overlays look correct, but you could remove all of that repetition by including the generic es9023 overlay in the badged versions as starting points. And why would you bind the names into the driver? Since the whole point is to encourage re-use just read the names from the overlays, or am I missing something?

On a side note, I hope there is a commercial agreement in place w.r.t. the "MamboBerry" name, otherwise the objections to "BoomBerry" ring a bit hollow.

sound/soc/bcm/es9023_dac.c
```diff
@@ -0,0 +1,259 @@
...
+	priv->dai.ops = &snd_rpi_es9023_dac_ops;
+
+	priv->card.dai_link = &priv->dai;
+	priv->card.dev = &pdev->dev;
+	priv->card.num_links = 1;
+	priv->card.owner = THIS_MODULE;
+
+	priv->bclk_ratio_int_div = of_property_read_bool(np,
+					"es9023-dac,bclk-ratio-int-div");
+
```
pelwell, line 223: Wouldn't it be smaller, simpler and more efficient to have one function that determines the board variant, returning a `const struct es9023_dac_variant*` from which you can pick the relevant field rather than four virtually identical functions to perform the lookup and then return one value from the structure?

Or just read the strings from DT.

clivem:
> On a side note, I hope there is a commercial agreement in place w.r.t. the "MamboBerry" name, otherwise the objections to "BoomBerry" ring a bit hollow.

This is not the place.... But AFAIK there has never been any attempt to register MamboBerry as a trademark, and therefore no-one objecting to it. I believe that is why Aaron switched from BoomBerry to JustBoom, because his UK/EU trademark application for BoomBerry was challenged.

With regard to your comments on this patch series. It has been superseded by the "simple" patch set, for which I'm stuck in the reeds, (as per usual), because upstream seemed less than enthusiastic about Martin's "simple framework" patch set, on which I am building.

Anyway, can we please leave this PR "open" at the moment as a placeholder. I'm not expecting or asking you to "pull" it.

clivem:
@pelwell On the basis that this is not a pull request that you are going to action, I've just pushed 2 patches to this PR, so you can see what it would look like with "simple". The es9023_dac machine driver gives way to simple-card. The es9023 codec is pushed upstream. 2x generic overlays downstream. One for pcm5102a variants, one for es9023. Plus the 2 overlays for custom bclk ratio, bclk_int_div.

pelwell:
Interesting. If the simple framework could be expanded just a bit more, allowing selection of external crystals, for example - then we might be able to do away with more drivers, but the difficulties include finding a useful set of operations (modifying alsa params, writing to registers?), ideally without having to change the codecs, whilst keeping the complexity low enough to get past the maintainers.

clivem:
> If the simple framework could be expanded just a bit more, allowing selection of external crystals

@pelwell Two steps in front of you. ;) But unless/until the most basic "simple framework" is accepted, before we even start adding more functionality to the framework, it's a non-starter..... Hence the reason for starting with examples of the brain-dead hardware (rather than software I2C) DAC's and just the bclk_div via the framework, as a demonstration, rather than anything more complicated.

Hopefully, Martin will come back from his travels, and we'll be able to move forward on a bunch of stuff.

PR #1462, opened 2016-05-09
IQAudIO: Add auto-mute for amp to dac driver

clivem:
Phil/Dom, this is another PR that needs third-party input rather than actually pulling.

@iqaudio Gordon, this was about to slip through the cracks. I'm not sure whether I ever sent you the 3rd generation patch, which rather than a "oneshot" to unmute amp at startup when driver module loads, is unmuting when audio device opens and muting when audio device closed. It achieves what you wanted, while accommodating those who already modified userspace software to mute/unmute the amp when the playback software opens/closes the audio device. Thoughts?

"dtoverlay=iqaudio-dacplus,auto_mute_amp"

### DIFF
--- a/arch/arm/boot/dts/overlays/README
+++ b/arch/arm/boot/dts/overlays/README
@@ -524,6 +524,12 @@ Params: 24db_digital_gain       Allow gain to be applied via the PCM512x codec
                                 responsibility of the user to ensure that
                                 the Digital volume control is set to a value
                                 that does not result in clipping/distortion!)
+        auto_mute_amp           If specified, unmute/mute the IQAudIO amp when
+                                starting/stopping audio playback. eg.
+                                dtoverlay=iqaudio-dacplus,auto_mute_amp
+        auto_mute_amp_gpio      Allow a user-specified GPIO for amp
+                                unmute/mute. (default is GPIO22.) eg.
+                                dtoverlay=iqaudio-dacplus,auto_mute_amp_gpio=22
 
 
 Name:   lirc-rpi
--- a/arch/arm/boot/dts/overlays/iqaudio-dacplus-overlay.dts
+++ b/arch/arm/boot/dts/overlays/iqaudio-dacplus-overlay.dts
@@ -30,14 +30,29 @@
 
 	fragment@2 {
 		target = <&sound>;
-		frag2: __overlay__ {
+		iqaudio_dac: __overlay__ {
 			compatible = "iqaudio,iqaudio-dac";
+			pinctrl-names = "default";
+			pinctrl-0 = <&iqaudio_dac_mute_amp_pins>;
 			i2s-controller = <&i2s>;
 			status = "okay";
 		};
 	};
 
+	fragment@3 {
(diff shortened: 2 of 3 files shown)


HiassofT:
It might be better to do the amp (un-)muting via DAPM/set_bias_level to avoid possible clicks.

By default DAPM waits 5 seconds (pmdown_time) before shutting down playback streams so a close/stop immediately followed by open/start doesn't trigger an amp mute+unmute.

Something like this: https://gist.github.com/HiassofT/33beeaf5a78d2819a324a11e2e3423fe

clivem:
Thanks Matthias, for the hint. I'll modify the patch later this evening.

iqaudio:
Sorry for delay in replying.

Intention is that unmute is applied to mask any electrical power on only.

A user may also (via a physical button) control the mute state. 

1) Apply power and amp is muted to mask any power on "thump"

2) Pi starts up.

3) amp is taken out of mute state.

4) user may play music etc and press mute button at any time. Say when home phone rings.

5) media application should continue to play with amp muted. This may be a playlist, or similar with Alsa level open / close.

6) user can unmute via button and music should be audible.

If auto mute is at every open / close then the above use case will fail. This was not the intention of the unmute at startup.

This is amp section mute, not Alsa mute.

Hope that helps clarify.

Best regards,

Gordon

clivem:
OK. I'm going to leave this alone. I think we are talking at cross purposes. ALSA mute of DAC chip is still available, and that will not cause amp to mute. I also wasn't aware of any physical buttons on your amp boards to control amp mute. So it sounds like I am implementing something I don't completely understand with outdated hardware. I'll close this. Sorry for the noise.

clivem:
@iqaudio If I have time tomorrow, I will create a new pull request for the original "oneshot" unmute amp when module loads. As I said, that was the original functionality. I'll try and get the "auto-mute" functionality made available via piCorePlayer distro for the guys already using it. Which means I can put the invert back and a couple of other things that were left out as not directly applicable to your "official" amp, but was applicable to the diy guys who are controlling amps via a GPIO while using your DAC board. ;)

iqaudio:
Sorry.

We intent to use a rotary encoder for volume and the encoders switch (via gpio) to toggle gpio22. The same can be done by linking gpio22 to ground or 3v3

Gordon

clivem:
@iqaudio Gordon, if you can find 10 mins tomorrow, can you test the latest patch, where I have added the once only "dtoverlay=iqaudio-dacplus,unmute_amp" (at startup) back. That should give you what you want....

The "dtoverlay=iqaudio-dacplus,auto_mute_amp" gives your users what they want. ;)

Rotary encoders with switches.... LOL Let me think about that. But that's an either/or. Either enable a dt driver param to control amp mute, or don't use dt param and control manually, whether that be from rotary encoder switch or some other manual switch......

Oops, won't let me re-open after I force pushed. Patch here.....
https://github.com/DigitalDreamtimeLtd/linux/commit/c2889af77eaf97cb439a2749ce8bfcb669f1f928

PR #1463, opened 2016-05-09
Rpi 4.4.y: Changes from drm-vc4 next

anholt:
This series is a followup from the HPD series (and includes it), bringing in changes from drm-vc4-next.  There are also two important new patches from Mario Kleiner on Friday that I haven't merged to -next just yet, but which I have tested on this branch.

The DPI support isn't quite finished -- my overlay (not included) ended up hanging at boot, and i haven't worked out what's going wrong yet.  However, without the overlay everything's fine and this brings us in line with -next (where I have tested DPI before).


anholt:
Re-push was to fix using &clocks instead of &cprman in the DPI node.

I'm successfully booting with DPI in my rpi-4.4.y-vc4-dpi branch and the new overlay included, but neither the firmware nor my driver are successfully displaying on the panel, so I suspect I've just broken the cable of my panel.

PR #1472, opened 2016-05-12
Rpi 4.4.y clk driver fixes

anholt:
Here are a few fixes that are on their way into the clk-next tree (Stephen Boyd had one comment that I hope I've answered to his satisfaction).  Patch 1 should make having drivers reference the VPU clock from cprman reliable.  Patch 2 should keep -EPROBE_DEFER or clock disables of siblings below PLLD_PER from taking down the network.  Patch 3 should fix a user's bug report I have about HDMI flickering (and I've had flickering on DSI because of it, as well).


joaodriessen:
Dear Pelwell & Popcornmix, please forgive the selfish question, any idea on when this pull request and #1463 will be reviewed/merged?

PR #1476, opened 2016-05-16
New driver for the AudioInjector audio input and output hat.

flatmax:
This audio hat has both input and output audio and allows other hats to be stacked on top.

Contains the ALSA machine driver and necessary alterations to the Kconfig and Makefile to drive the audioinjector pi hat.
Adds the dts overlay and updates the Makefile and README.
Updates the relevant defconfig files to enable building for the Raspberry Pi.


sound/soc/bcm/audioinjector-pi-hat.c
```diff
@@ -0,0 +1,140 @@
...
+		.of_match_table = audioinjector_pi_hat_of_match,
+       },
+       .probe          = audioinjector_pi_hat_probe,
+       .remove         = audioinjector_pi_hat_remove,
+};
+
+module_platform_driver(audioinjector_pi_hat_driver);
+MODULE_AUTHOR("Matt Flax <flatmax@flatmax.org>");
+MODULE_DESCRIPTION("AudioInjector.net Pi Hat");
+MODULE_LICENSE("GPL V2");
```
pelwell: This must say "GPL v2" (note the lower-case "v") otherwise it won't link.

sound/soc/bcm/audioinjector-pi-hat.c
```diff
@@ -0,0 +1,140 @@
...
+			dai->platform_of_node = i2s_node;
+		} else
+			if (!dai->cpu_of_node) {
+				dev_err(&pdev->dev, "Property 'i2s-controller' missing or invalid\n");
+				return -EINVAL;
+			}
+	}
+
+	if ((ret = snd_soc_register_card(card))) {
+		printk("audioinjector_pi_hat_probe error\n");
```
pelwell: I don't think you need this line - isn't the following dev_err enough? printk is frowned on these days.

sound/soc/bcm/audioinjector-pi-hat.c
```diff
@@ -0,0 +1,140 @@
+/*
+ * ASoC Driver for AudioInjector Pi Hat
+ *
+ *  Created on: 13-May-2016
+ *      Author: flatmax@flatmax.org
```
pelwell: It would be polite to mention another author here, since nobody in their right minds writes modules like this without reading another. The "i2s-controller" support looks familiar, for a start.

pelwell:
I've made a few in-line comments - apart from those the code looks OK.

Nice job on the overlay and README as well - I didn't find a single niggle.

I have one question though - this driver is _so_ simple, could you not use the [simple-card](https://github.com/raspberrypi/linux/blob/rpi-4.4.y/Documentation/devicetree/bindings/sound/simple-card.txt) driver instead?

clivem:
I want to make several comments, but in the interests of not making more enemies can we please start thinking about some naming rules for I2S audio HAT overlays, specifically I would like to see the codec name in the overlay name...... so eg. "audioinjector-pi-hat", what I'd like to see going forward is ...

<product>-<codec>-audio .... so the overlay name for this would be "audioinjector-wm8731-audio" ;)

Sorry for bringing this up in a random PR. It isn't directly aimed at the submitter of this PR, but it will be much easier going forward if we have some kind of "loose" naming scheme, but a naming policy all the same.

pelwell:
That would mean renaming the existing overlays, which won't be popular.

But on the subject of HATs, I do hope the AudioInject HAT has an EEPROM, otherwise it shouldn't be called a HAT (see the [HAT FAQ](https://github.com/raspberrypi/hats/blob/master/README.md)).

clivem:
> That would mean renaming the existing overlays, which won't be popular.

No, existing stuff, already accepted will remain for compatibility. But new additions need to conform.

I have rescheduled my call with Daniel tonight, but eg for HB. they will keep their existing hifiberry_dac overlay, limited to 192k for compatibility, but their users will be able to also use simple-pcm5102a and simple-es9023a overlays. (Which will be new additions, using simple-card, not hifiberry_dac driver.) If we make a specific (for commercial reasons) new overlay for HB as a migration path, it would be hifiberry_pcm5102a, but hifiberry_dac overlay will also remain for time to come for compatibility.

I do have half a plan which I need to get a couple more people on board with, before presenting to you. But in the interim, let's start thinking about naming for new additions.... Please.

I am aware that there are going to be another 5 audio HAT submissions over the coming weeks, not just the two you got this week and one last week. This is going to get even more messy very quickly, with no sort of naming policy.

clivem:
> But on the subject of HATs, I do hope the AudioInject HAT has an EEPROM, otherwise it shouldn't be called a HAT 

More than that, it's not just good enough to have an unprogrammed EEPROM on the board, it actually needs to be programmed to conform to HAT spec, in which case overlay is "audioinjector-wm8731-audio-hat", otherwise just "audioinjector-wm8731-audio".... @pelwell Thoughts????

pelwell:
You could argue that a HAT with a programmed EEPROM doesn't need an overlay to be maintained in the kernel tree, but I'm not getting into this now...

sound/soc/bcm/audioinjector-pi-hat.c
flatmax: Thanks for pointing this out.

clivem:
> You could argue that a HAT with a programmed EEPROM doesn't need an overlay to be maintained in the kernel tree

Yes, you could, but from a maintenance point of view, I would argue that it should be submitted, so if anything needs to change in the eeprom overlay, perhaps as the result of a major kernel update, a user can be instructed to use ....

```
force_eeprom_read=0
dtoverlay=audioinjector-wm8731-audio-hat
```

... in his config.txt until the eeprom can be re-flashed with an updated overlay. ;)

sound/soc/bcm/audioinjector-pi-hat.c
flatmax: I have added the history of this machine driver.

flatmax:
I have now got 2 extra commits addressing the initial requests for code changes, then the suggested renames.
I think this all actually makes the code and commits more descriptive and meaninful. Thanks for your review and discussion.

Oh - I forgot to mention that the simple-card.c has it's  snd_soc_dai_set_sysclk hard wired to clk_id=0. As the audioinjector pi sound card uses a crystal on the codec for low jitter high quality audio (codec master), the wm8731 codec driver requires a clk_id=1. Otherwise I would be able to use the simple-card.c driver ... nice idea !

Matt

pelwell:
The changes look good, but you didn't comment on my simple-card question.

flatmax:
oh yep ... that is a great idea, using simple-card.c.
However, there is a problem, simple-card.c hard wires the clk_id in the function call "snd_soc_dai_set_sysclk"  to 0. When the wm8731 is put into master mode, it requires a clk_id=WM8731_SYSCLK_XTAL=1 ... this tells the codec driver that there is a crystal attached and alters the DAPM (as well as setting the frequency and certain constraints). For that reason, simple-card.c will not work - as it is hard wired to a clk_id of 0.

If I am not mistaken, this is the only stumbling block ? Great idea ... configure everything from the device tree, I am going to use your suggestion in future where possible !

thanks
Matt

pelwell:
Yes, once again simple-card proves to be slightly too simple to be useful.

OK - three final requests:

1) Squash these three commits down to one using `git rebase -i HEAD~3`, marking the 2nd and 3rd commits for "fixup".

2) Change the commit message so there is a blank line between the first line (the Subject) and the rest of the description - think of it like an email. Without that change, the git log message is many lines long. The upstream kernel has a 52-character limit for commit message, but for Pi-specific commits 70-72 is acceptable - short enough to get a git log --oneline into 80 characters.

3) Use `git push -f` to update the PR.

flatmax:
Thank you, now committed.

PR #1485, opened 2016-05-23
Added Overlay for Microchip MCP23017 I2C gpio expander

MikeDK:
This overlay makes use of the gpio_mcp23s08 kernel driver and is currently only intended for the I2C device MCP23017 due to lack of other devices to test.

I2C address and interrupt pin can be parametrized -> described in README.

### DIFF
--- a/arch/arm/boot/dts/overlays/README
+++ b/arch/arm/boot/dts/overlays/README
@@ -567,6 +567,15 @@ Params: gpio_out_pin            GPIO for output (default "17")
                                 (default "off")
 
 
+Name:   mcp23017
+Info:   Configures the MCP23017 I2C port expander
+Load:   dtoverlay=mcp23017,<param>=<val>
+Params: gpiopin                 Gpio pin connected to the INTA output of the
+                                MCP23017 (default: 4)
+
+        addr                    I2C address of the MCP23017 (default: 0x20)
+
+
 Name:   mcp2515-can0
 Info:   Configures the MCP2515 CAN controller on spi0.0
 Load:   dtoverlay=mcp2515-can0,<param>=<val>
--- a/arch/arm/boot/dts/overlays/mcp23017-overlay.dts
+++ b/arch/arm/boot/dts/overlays/mcp23017-overlay.dts
@@ -1,0 +1,54 @@
+// Definitions for MCP23017 Gpio Extender from Microchip Semiconductor
+
+/dts-v1/;
+/plugin/;
+
+/ {
+	compatible = "brcm,bcm2835", "brcm,bcm2708", "brcm,bcm2709";
+
+	fragment@0 {
+		target = <&i2c1>;
+		__overlay__ {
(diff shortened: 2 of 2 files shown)


arch/arm/boot/dts/overlays/README
```diff
@@ -586,6 +586,14 @@ Params: oscillator              Clock frequency for the CAN controller (Hz)
 
         interrupt               GPIO for interrupt signal
 
+
```
pelwell: Delete this trailing whitespace.

arch/arm/boot/dts/overlays/README
```diff
@@ -586,6 +586,14 @@ Params: oscillator              Clock frequency for the CAN controller (Hz)
 
         interrupt               GPIO for interrupt signal
 
+        
+Name:   mcp23017
+Info:   Configures the MCP23017 I2C port expander
+Load:   dtoverlay=mcp23017,<param>=<val>
+Params: gpiopin                 Gpio pin connected to the INTA output of the MCP23017
```
pelwell: This line is too long, and doesn't say what the default value is.

arch/arm/boot/dts/overlays/README
```diff
@@ -586,6 +586,14 @@ Params: oscillator              Clock frequency for the CAN controller (Hz)
 
         interrupt               GPIO for interrupt signal
 
+        
+Name:   mcp23017
```
pelwell: Alphabetical order, please - this belongs before mcp2515

pelwell:
I've commented in-line. Fix those, squash down to one commit and `git push -f`, and I'll merge it.

MikeDK:
done ;-)

MikeDK:
Question: who shall do the device tree compiling?
I can see that my .dts is included in the kernel, and the new readme is also in the last firmware, but there is no .dtb(o) file in the overlays folder ;)

MikeDK:
PS: Or shall I compile it, and make a pull request on the firmware repo?

popcornmix:
The firmware will be generated with whatever is built by the kernel, so no need to PR there.
What seems to be missing is the Makefile change to actually build this overlay. See:
https://github.com/raspberrypi/linux/blob/rpi-4.4.y/arch/arm/boot/dts/overlays/Makefile#L47

MikeDK:
ahh thanks for that hint!
I have completely overseen the Makefile...

When I do the change, shall I squash down to one commit again, or shall I make a seperate pull request (as this one is already merged?)

pelwell:
It's OK - since it was also my oversight I've patched up the Makefile myself. I have a script that checks overlays and the README before merging, but it doesn't check the Makefile - I'll have to fix that.

pelwell:
I've now implemented the check, so this shouldn't happen again. It also means I get to be picky about the order of entries in the Makefile.

MikeDK:
Thanks a lot!

PR #1488, opened 2016-05-24
[8192cu] Fixes

fran6co:
(https://github.com/raspberrypi/linux/pull/1487) Summary of changes:
- Adds support for ISY - Edimax
- Adds support for On Networks - N150MA
- Adds support for Netgear N150 -WNA1000M
- Adds support for TP-Link TL-WN723N
- Suppress spurious messages
- Fixes https://github.com/pvaret/rtl8192cu-fixes/issues/23
- Fixes CONFIG_MULTI_VIR_IFACES
- Fixes compatiblity with 3.13
- Enables compiler warnings and general fixes
- Fixes NetworkManager not finding the device (https://github.com/raspberrypi/linux/pull/1525, https://github.com/raspberrypi/linux/pull/1489#issuecomment-224958014)

Reference repositories:
- https://github.com/pvaret/rtl8192cu-fixes
- https://github.com/diederikdehaas/rtl8812AU (realtek drivers share a common base, that's the latest I could find)

Issue #1319, opened 2016-03-03
kexec new kernel freeze after show Starting new kernel


huangjiansword:
step 1,build new kernel
I build a new kernel from source 3.18.7+ follow https://www.kernel.org/doc/Documentation/kdump/kdump.txt
CONFIG_KEXEC=y
CONFIG_SYSFS=y
CONFIG_DEBUG_INFO=Y
CONFIG_CRASH_DUMP=y
CONFIG_PROC_VMCORE=y
AUTO_ZRELADDR=y

then I copy new zImage from arch/arm/boot to SD Card /boot, and delete old kernel.img, then rename zImage to kernel.img

then I reboot the pi (A+), everything is OK.

and I checked config parameter with zless /proc/config.gz, it's all correct.

step2: 
I install kexec-tools at the new kernel pi.

sudo apt-get install kexec-tools

kexec's version is 2.0.3 ( but I tried to build kexec 2.0.11 and installed, also freeze at the same place)

step 3,run new kernel
 when I run this:
sudo kexec --type=zImage /boot/kernel.img --command-line="$(cat /proc/cmdline)" 

it show that the device is shutdown and  umount local filesystems

but freezed after show "kexec: Starting new kernel"

what's wrong with it?

huangjiansword:
```
pi@raspberrypi ~ $ cat /boot/cmdline.txt
dwc_otg.lpm_enable=0 console=ttyAMA0,115200 console=tty1 root=/dev/mmcblk0p2 rootfstype=ext4 elevator=deadline rootwait
```

```
pi@raspberrypi ~ $ cat /boot/config.txt
# For more options and information see
# http://www.raspberrypi.org/documentation/configuration/config-txt.md
# Some settings may impact device functionality. See link above for details

# uncomment if you get no picture on HDMI for a default "safe" mode
#hdmi_safe=1

# uncomment this if your display has a black border of unused pixels visible
# and your display can output without overscan
#disable_overscan=1

# uncomment the following to adjust overscan. Use positive numbers if console
# goes off screen, and negative if there is too much border
#overscan_left=16
#overscan_right=16
#overscan_top=16
#overscan_bottom=16

# uncomment to force a console size. By default it will be display's size minus
# overscan.
#framebuffer_width=1280
#framebuffer_height=720

# uncomment if hdmi display is not detected and composite is being output
#hdmi_force_hotplug=1

# uncomment to force a specific HDMI mode (this will force VGA)
#hdmi_group=1
[...]

Issue #631, opened 2014-07-07
I2C GPIO-PCA953X interrupt support ?


ghost:
Hello,
I am using the PCA953X GPIO driver to interface a PCA9555-equivalent IC via I2C-1 on a Pi Rev 2. The driver is able to provide a polling mechanism to read the Input pin values. I have an interrupt line wired up too. From the kernel driver (3.10 or 3.12), the interrupt line gets registered, but I dont see any interrupts. I was wondering if anybody has got the interrupt version of the driver working on a Pi. From what I can make out, this is a chained gpiochip and it seems to be failing in the irq_create_mapping scenario. What specific changes would be needed ?
   Hopefully this is the right forum to post this on. 
TIA for the help,
Ajit

Ruffio:
@aja91 has this issue been resolved? If yes, then please close this issue.

PR #1489, opened 2016-05-24
[8192cu] Enables cfg80211

fran6co:
Got some patches from https://github.com/pld-linux/rtl8812au

### DIFF
--- a/drivers/net/wireless/realtek/rtl8192cu/os_dep/linux/ioctl_cfg80211.c
+++ b/drivers/net/wireless/realtek/rtl8192cu/os_dep/linux/ioctl_cfg80211.c
@@ -331,7 +331,7 @@ static int rtw_cfg80211_inform_bss(_adapter *padapter, struct wlan_network *pnet
 	u8 *notify_ie;
 	size_t notify_ielen;
 	s32 notify_signal;
-	u8 buf[MAX_BSSINFO_LEN], *pbuf;
+	u8 *buf, *pbuf;
 	size_t len,bssinf_len=0;
 	struct rtw_ieee80211_hdr *pwlanhdr;
 	unsigned short *fctrl;
@@ -459,6 +459,7 @@ static int rtw_cfg80211_inform_bss(_adapter *padapter, struct wlan_network *pnet
 	DBG_8192C("notify_timestamp: %#018llx\n", notify_timestamp);
 */
 
+	buf = rtw_zmalloc(MAX_BSSINFO_LEN);
 	pbuf = buf;
 
 	pwlanhdr = (struct rtw_ieee80211_hdr *)pbuf;
@@ -485,6 +486,8 @@ static int rtw_cfg80211_inform_bss(_adapter *padapter, struct wlan_network *pnet
 
 	_rtw_memcpy(pbuf, pnetwork->network.IEs, pnetwork->network.IELength);
 	len += pnetwork->network.IELength;
+
+	rtw_mfree(buf, MAX_BSSINFO_LEN);
 
 	//#ifdef CONFIG_P2P
 	//if(rtw_get_p2p_ie(pnetwork->network.IEs+12, pnetwork->network.IELength-12, NULL, NULL))
@@ -577,7 +580,11 @@ int rtw_cfg80211_check_bss(_adapter *padapter)
 	bss = cfg80211_get_bss(padapter->rtw_wdev->wiphy, notify_channel,
 			pnetwork->MacAddress, pnetwork->Ssid.Ssid,
 			pnetwork->Ssid.SsidLength,
+#if LINUX_VERSION_CODE < KERNEL_VERSION(4, 1, 0)
 			WLAN_CAPABILITY_ESS, WLAN_CAPABILITY_ESS);
+#else
+			IEEE80211_BSS_TYPE_ESS, IEEE80211_PRIVACY_ANY);
+#endif
 
 	return 	(bss!=NULL);
 }
@@ -700,13 +707,16 @@ void rtw_cfg80211_indicate_disconnect(_adapter *padapter)
 	if (!padapter->mlmepriv.not_indic_disco) {
 		DBG_8192C("pwdev->sme_state(b)=%d\n", pwdev->sme_state);
 
-		if(pwdev->sme_state==CFG80211_SME_CONNECTING)
+		if (check_fwstate(pmlmepriv, WIFI_UNDER_LINKING)) {
 			cfg80211_connect_result(padapter->pnetdev, NULL, NULL, 0, NULL, 0,
 				WLAN_STATUS_UNSPECIFIED_FAILURE, GFP_ATOMIC/*GFP_KERNEL*/);
-		else if(pwdev->sme_state==CFG80211_SME_CONNECTED)
+		} else {
+#if LINUX_VERSION_CODE < KERNEL_VERSION(4,2,0)
 			cfg80211_disconnected(padapter->pnetdev, 0, NULL, 0, GFP_ATOMIC);
-		//else
-			//DBG_8192C("pwdev->sme_state=%d\n", pwdev->sme_state);
+#else
+			cfg80211_disconnected(padapter->pnetdev, 0, NULL, 0, false, GFP_ATOMIC);
+#endif
+        }
 
 		DBG_8192C("pwdev->sme_state(a)=%d\n", pwdev->sme_state);
 	}
@@ -1549,7 +1559,7 @@ static int cfg80211_rtw_set_default_key(struct wiphy *wiphy,
 
 static int cfg80211_rtw_get_station(struct wiphy *wiphy,
 				    struct net_device *ndev,
-				    u8 *mac, struct station_info *sinfo)
+				    const u8 *mac, struct station_info *sinfo)
 {
 	int ret = 0;
 	_adapter *padapter = wiphy_to_adapter(wiphy);
@@ -1589,16 +1599,32 @@ static int cfg80211_rtw_get_station(struct wiphy *wiphy,
 			goto exit;
 		}
 
+#if (LINUX_VERSION_CODE >= KERNEL_VERSION(4,0,0))
+		sinfo->filled |= BIT(NL80211_STA_INFO_SIGNAL);
+#else
 		sinfo->filled |= STATION_INFO_SIGNAL;
+#endif
 		sinfo->signal = translate_percentage_to_dbm(padapter->recvpriv.signal_strength);
 
+#if (LINUX_VERSION_CODE >= KERNEL_VERSION(4,0,0))
+		sinfo->filled |= BIT(NL80211_STA_INFO_TX_BITRATE);
+#else
 		sinfo->filled |= STATION_INFO_TX_BITRATE;
+#endif
 		sinfo->txrate.legacy = rtw_get_cur_max_rate(padapter);
 
+#if (LINUX_VERSION_CODE >= KERNEL_VERSION(4,0,0))
+		sinfo->filled |= BIT(NL80211_STA_INFO_RX_PACKETS);
+#else
 		sinfo->filled |= STATION_INFO_RX_PACKETS;
+#endif
 		sinfo->rx_packets = sta_rx_data_pkts(psta);
 
+#if (LINUX_VERSION_CODE >= KERNEL_VERSION(4,0,0))
+		sinfo->filled |= BIT(NL80211_STA_INFO_TX_PACKETS);
+#else
 		sinfo->filled |= STATION_INFO_TX_PACKETS;
+#endif
 		sinfo->tx_packets = psta->sta_stats.tx_pkts;
 
 	}
@@ -2419,7 +2445,7 @@ static int rtw_cfg80211_set_key_mgt(struct security_priv *psecuritypriv, u32 key
 	return 0;
 }
 
-static int rtw_cfg80211_set_wpa_ie(_adapter *padapter, u8 *pie, size_t ielen)
+static int rtw_cfg80211_set_wpa_ie(_adapter *padapter, const u8 *pie, size_t ielen)
 {
 	u8 *buf=NULL, *pos=NULL;
 	u32 left;
@@ -3089,7 +3115,9 @@ void rtw_cfg80211_indicate_sta_assoc(_adapter *padapter, u8 *pmgmt_frame, uint f
 			ie_offset = _REASOCREQ_IE_OFFSET_;
 
 		sinfo.filled = 0;
+#if (LINUX_VERSION_CODE < KERNEL_VERSION(4,0,0))
 		sinfo.filled = STATION_INFO_ASSOC_REQ_IES;
+#endif
 		sinfo.assoc_req_ies = pmgmt_frame + WLAN_HDR_A3_LEN + ie_offset;
 		sinfo.assoc_req_ies_len = frame_len - WLAN_HDR_A3_LEN - ie_offset;
 		cfg80211_new_sta(ndev, GetAddr2Ptr(pmgmt_frame), &sinfo, GFP_ATOMIC);
@@ -3387,7 +3415,11 @@ static const struct net_device_ops rtw_cfg80211_monitor_if_ops = {
 };
 #endif
 
+#if (LINUX_VERSION_CODE >= KERNEL_VERSION(4,1,0))
+static int rtw_cfg80211_add_monitor_if(_adapter *padapter, char *name, unsigned char name_assign_type, struct net_device **ndev)
+#else
 static int rtw_cfg80211_add_monitor_if(_adapter *padapter, char *name, struct net_device **ndev)
+#endif
 {
 	int ret = 0;
 	struct net_device* mon_ndev = NULL;
@@ -3418,6 +3450,9 @@ static int rtw_cfg80211_add_monitor_if(_adapter *padapter, char *name, struct ne
 	mon_ndev->type = ARPHRD_IEEE80211_RADIOTAP;
 	strncpy(mon_ndev->name, name, IFNAMSIZ);
 	mon_ndev->name[IFNAMSIZ - 1] = 0;
+#if (LINUX_VERSION_CODE >= KERNEL_VERSION(4,1,0))
+	mon_ndev->name_assign_type = name_assign_type;
+#endif
 	mon_ndev->destructor = rtw_ndev_destructor;
 
 #if (LINUX_VERSION_CODE>=KERNEL_VERSION(2,6,29))
@@ -3482,6 +3517,9 @@ static int
 	#else
 		char *name,
 	#endif
+#if (LINUX_VERSION_CODE >= KERNEL_VERSION(4,1,0))
+		unsigned char name_assign_type,
+#endif
 		enum nl80211_iftype type, u32 *flags, struct vif_params *params)
 {
 	int ret = 0;
@@ -3499,7 +3537,11 @@ static int
 		ret = -ENODEV;
 		break;
 	case NL80211_IFTYPE_MONITOR:
+#if (LINUX_VERSION_CODE >= KERNEL_VERSION(4,1,0))
+		ret = rtw_cfg80211_add_monitor_if(padapter, (char *)name, name_assign_type, &ndev);
+#else
 		ret = rtw_cfg80211_add_monitor_if(padapter, (char *)name, &ndev);
+#endif
 		break;
 
 #if (LINUX_VERSION_CODE >= KERNEL_VERSION(2,6,37)) || defined(COMPAT_KERNEL_RELEASE)
@@ -3760,7 +3802,7 @@ static int cfg80211_rtw_stop_ap(struct wiphy *wiphy, struct net_device *ndev)
 #endif //(LINUX_VERSION_CODE < KERNEL_VERSION(3,4,0))
 
 static int	cfg80211_rtw_add_station(struct wiphy *wiphy, struct net_device *ndev,
-			       u8 *mac, struct station_parameters *params)
+			       const u8 *mac, struct station_parameters *params)
 {
 	DBG_871X(FUNC_NDEV_FMT"\n", FUNC_NDEV_ARG(ndev));
 
@@ -3768,8 +3810,15 @@ static int	cfg80211_rtw_add_station(struct wiphy *wiphy, struct net_device *ndev
 }
 
 static int	cfg80211_rtw_del_station(struct wiphy *wiphy, struct net_device *ndev,
-			       u8 *mac)
+#if (LINUX_VERSION_CODE < KERNEL_VERSION(3,15,0))
+			       u8 *mac) {
+#elif (LINUX_VERSION_CODE < KERNEL_VERSION(4,1,0))
+                   const u8 *mac) {
+#else
+                   struct station_del_parameters *params)
 {
+    const u8 *mac = params->mac;
+#endif
 	int ret=0;
 	_irqL irqL;
 	_list	*phead, *plist;
@@ -3859,7 +3908,7 @@ static int	cfg80211_rtw_del_station(struct wiphy *wiphy, struct net_device *ndev
 }
 
 static int	cfg80211_rtw_change_station(struct wiphy *wiphy, struct net_device *ndev,
-				  u8 *mac, struct station_parameters *params)
+				  const u8 *mac, struct station_parameters *params)
 {
 	DBG_871X(FUNC_NDEV_FMT"\n", FUNC_NDEV_ARG(ndev));
 
@@ -4643,6 +4692,9 @@ static int cfg80211_rtw_mgmt_tx(struct wiphy *wiphy,
 #else
 	struct net_device *ndev,
 #endif
+#if (LINUX_VERSION_CODE >= KERNEL_VERSION(3,14,0))
+	struct cfg80211_mgmt_tx_params *params,
+#else
 	struct ieee80211_channel *chan,
 #if (LINUX_VERSION_CODE >= KERNEL_VERSION(2,6,38)) || defined(COMPAT_KERNEL_RELEASE)
 	bool offchan,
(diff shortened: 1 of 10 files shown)


popcornmix:
@MrEngman any comments?

fran6co:
Don't merge it yet, I'm having some problems with the connection it gets a bit spotty as it runs for some time.

fran6co:
Fixed, it seems to be working now.

MrEngman:
Hi @popcornmix,

First I must apologise for not responding sooner but I've not been too well for the last couple of weeks. Recovering now but then I received your request for comment and that gave me a real shock and I'm just recovering from that :)

I feel really honoured to be asked for my opinion.

I must say that I've been involved with wifi on the Pi pretty much since it first appeared, compiling a variety of out of tree drivers, and I guess am one of the reasons the 8192cu driver was included in the image, however, compiling wifi drivers and actually knowing how they work are two completely different matters. However, I do think it would be good to update the driver.

I have updated the driver on one of my Pi's and have been running tests transferring data back and forth from the Pi 2 B and a Win7 laptop for the last couple of days and it would appear to be running without any problems. @fran6co looks to have done a good job.

There is just one minor alteration I made. In the past I had noticed when compiling the kernel that there was a warning from the 8192cu driver and this was still there when I compiled the kernel before applying @fran6co's update - 

```
  CC [M]  drivers/net/wireless/realtek/rtl8192cu/os_dep/linux/os_intfs.o
drivers/net/wireless/realtek/rtl8192cu/os_dep/linux/os_intfs.c:999:2: warning: initialization from incompatible pointer type [enabled by default]
  .ndo_select_queue = rtw_select_queue,
  ^
drivers/net/wireless/realtek/rtl8192cu/os_dep/linux/os_intfs.c:999:2: warning: (near initialization for ‘rtw_netdev_ops.ndo_select_queue’) [enabled by default]
  CC [M]  drivers/net/wireless/realtek/rtl8192cu/os_dep/linux/usb_intf.o

```

I applied an additional patch to remove that warning. I have used the same patch on other Realtek drivers. The patch is -

```
diff -aur linux/drivers/net/wireless/realtek/rtl8192cu-old/os_dep/linux/os_intfs.c linux/drivers/net/wireless/realtek/rtl8192cu/os_dep/linux/os_intfs.c
--- linux/drivers/net/wireless/realtek/rtl8192cu-old/os_dep/linux/os_intfs.c    2015-12-30 16:52:02.256591427 +0000
+++ linux/drivers/net/wireless/realtek/rtl8192cu/os_dep/linux/os_intfs.c        2015-12-30 15:44:24.256765365 +0000
@@ -945,7 +945,14 @@
        return dscp >> 5;
 }

-static u16 rtw_select_queue(struct net_device *dev, struct sk_buff *skb)
+static u16 rtw_select_queue(struct net_device *dev, struct sk_buff *skb
+#if LINUX_VERSION_CODE >= KERNEL_VERSION(3, 13, 0)
+       , void *accel_priv
+#endif
+#if LINUX_VERSION_CODE >= KERNEL_VERSION(3, 14, 0)
+       , select_queue_fallback_t fallback
+#endif
+)
 {
        _adapter        *padapter = rtw_netdev_priv(dev);
        struct mlme_priv *pmlmepriv = &padapter->mlmepriv;


```

fran6co:
Thanks @MrEngman! Should I apply this patch in this PR or would be better to add it to https://github.com/raspberrypi/linux/pull/1488 ?

MrEngman:
I'm not sure, Perhaps you should wait for @popcornmix to suggest where.

fran6co:
Could you review those changes too?

popcornmix:
@fran6co probably best to add @MrEngman's patch to #1488.

@pelwell any objections to this?

fran6co:
@popcornmix  I did add his patch to #1488. https://github.com/raspberrypi/linux/pull/1488/commits/2bb47d2ed6fa196be7db2ac5f92f928eee65d333

drivers/net/wireless/realtek/rtl8192cu/os_dep/linux/ioctl_cfg80211.c
```diff
@@ -577,7 +580,11 @@ int rtw_cfg80211_check_bss(_adapter *padapter)
 	bss = cfg80211_get_bss(padapter->rtw_wdev->wiphy, notify_channel,
 			pnetwork->MacAddress, pnetwork->Ssid.Ssid,
 			pnetwork->Ssid.SsidLength,
+#if LINUX_VERSION_CODE < KERNEL_VERSION(4, 1, 0)
 			WLAN_CAPABILITY_ESS, WLAN_CAPABILITY_ESS);
+#else
+            IEEE80211_BSS_TYPE_ESS, IEEE80211_PRIVACY_ANY);
```
pelwell: Use TABs for indentation.

fran6co: Done.

pelwell:
checkpatch has lots of issues with it - see https://gist.github.com/pelwell/e42758ee8cbe2ca864c6eb0452df54f6.

fran6co:
@pelwell I didn't run checkpatch... I'll fix all those issue.

pelwell:
OK - when that's fixed and you've squashed them to one patch I'm happy for it to be merged.

fran6co:
Most of those errors and warnings that come from the realtek driver (for example C99 stuff). Should I create a new commit fixing all that stuff in all the realtek code?

popcornmix:
I'd say it's probably not worth fixing warnings from the existing realtek driver.
Ideally it will dropped one day (when upstream driver has improved sufficiently).

But any code you change would be better to match what upstream prefers.

fran6co:
There are some warning I can't fix, the code uses `extern` and `checkpatch` complains about it.

fran6co:
Same goes for the use of `LINUX_VERSION_CODE`

fran6co:
I think it should be all or nothing with this, the realtek driver doesn't comply to any of the Linux standards. I could clean and reformat the code to do so but if this driver is not going to be merged upstream I don't see the point. If I just clean this patch, it's going to be a mismatch of code styles in an already messy code.

popcornmix:
Yes, I think getting the whole realtek driver through checkpatch is too big a job.

@pelwell are you okay with how it is currently?

pelwell:
Yes. Let's hope it has a short life.

popcornmix:
Yes, we should start building the upstream driver, with it disabled by default, but easily enabled.
Encourage more users to use the upstream driver and switch the default at some point.

@MrEngman / @fran6co do you know which driver gets used if the kernel is built with both the upstream and realtek drivers?

fran6co:
Just tested, it loads the upstream one.

popcornmix:
Okay, it would have been easier if it were the realtek one.
I assume if you blacklist the upstream one you get the realtek one loaded?

fran6co:
Yeah, I'm using `blacklist rtl8192cu`.

clivem:
If you have time to play around with this, you probably also want to enable the upstream rtl8xxxu driver and test that with 8188CUS and 8192CU chipsets.

fran6co:
I wouldn't rely on the order, it's better to just blacklist the other drivers always.

popcornmix:
Okay, so creating a file `/etc/modprobe.d/rtl8192cu.conf` containing `blacklist 8192cu`
will stop the realtek driver from being used.

Or creating a file `/etc/modprobe.d/rtl8192cu.conf` containing `blacklist rtl8192cu`
will stop the upstream driver from being used.

So we could start building the updstream driver in addition, and also add the `blacklist rtl8192cu`  file, and if users get both those changes together, they will have no change in behaviour, but have the ability to rename `blacklist rtl8192cu` to `blacklist 8192cu` to switch to the upstream driver.

@MrEngman / @fran6co Can you confirm my understanding is correct?

It's a little tricky to do both of these simultaneously. We can do it through apt-get once we believe this it working correctly (and in fact could add the blacklist in advance as it should be harmless now).

rpi-update users may inadvertently switch to the upstream driver. We can make a note of this in the rpi-update warning text that is displayed (pointing to a forum thread discussing this change). Additional switchers to the upstream driver will be useful for testing purposes if updaters don't read the notice...

MrEngman:
Not sure if this is OK but finally managed to compile the upstream driver and ended up with 4 modules

```
./rtl8192cu/rtl8192cu.ko
./rtlwifi.ko
./rtl8192c/rtl8192c-common.ko
./rtl_usb.ko
```

Created the directories and installed the modules on my Pi and then rebooted without blacklisting either of the drivers and well -

lsmod showed

```
rtl_usb                12828  0
rtl8192c_common        58401  0
rtlwifi                85402  2 rtl_usb,rtl8192c_common
mac80211              532034  2 rtl_usb,rtlwifi
8192cu                569295  0
cfg80211              427855  3 mac80211,rtlwifi,8192cu
rfkill                 16037  2 cfg80211
```

So some of the upstream modules were loaded but not the main driver `rtl8192cu.ko` and instead the Realtek driver `8192cu.ko` was loaded. dmesg showed

```
[    6.205577] usbcore: registered new interface driver rtl8192cu
[    6.309218] Error: Driver 'rtl8192cu' is already registered, aborting...
[    6.416577] rtl8192cu 1-1.2:1.0 wlan1: renamed from wlan0
[    6.456133] systemd-udevd[157]: renamed network interface wlan0 to wlan1
```

so presumably the Realtek driver was registered first and as it uses the name `rtl8192cu` even though the driver is `8192cu.ko`, this stopped the upstream driver loading. I have seen similar before where two drivers were both loaded, I believe in that case MediaTek mt7610 or mt7601, but each had a slightly different name, although the wifi still worked.

I then created file `/etc/modprobe.d/rtl8192cu.conf` and added `blacklist 8192cu` and rebooted and the Pi came up with the upstream driver enabled, however the wifi led was permanently lit. Not nice AFAIAC, and I don't expect other users would be to pleased either.

I then removed `blacklist 8192cu` from `/etc/modprobe.d/rtl8192cu.conf` and edited `/etc/modprobe.d/8192cu.conf` and added `blacklist rtl8192cu` and rebooted. The Realtek wifi came up fine, led flashing as normal, much nicer !!

With the various drivers blacklisted lsmod showed the relevant correct modules were loaded unlike when they were both enabled.

On a similar note I am not keen on the staging 8188eu driver and compile my own 8188eu driver from a rather newer version Realtek driver which I much prefer to use, as do a number of other users. Or maybe I'm just too fussy. ;)

fran6co:
Squashed the commits

clivem:
@popcornmix **WARNING** I've got a feeling that the merging of this PR is the reason that none of the 6 Pi's with 8188 dongles I have in my test group, have re-appeared on the network since I updated the kernel an hour ago and rebooted them!

popcornmix:
@clivem would be good if you can confirm that reverting this commit fixes it.
Also any differences in dmesg or lsmod output would be useful.

clivem:
@popcornmix Yes, I'll investigate later. Sorry, not in the same physical location that they are, right now. And as is always the case, when you need it, none of those 8188 dongled units are one of the 3 units that have a USB serial console cable plugged into the NUC sat next to them. I think I need to make a bulk purchase of USB serial cables, and daisy chain a few USB hubs back to the NUC, so I have a serial console for each one of the units in the test group. One rainy day......

MrEngman:
Just compiled the updated kernel and the warning when compiling the 8192cu driver I mentioned earlier in this issue is still there. What happened to the patch to fix it?

Does the update not include the patches in #1488 where I think my patch ended up in?

I initially tested the updated driver with just #1489 but then also added #1488. No obvious differences in operation.

Just going to check with 8188eu wifi but don't expect to find any problems  But you never know.... ;)

clivem:
> Just going to check with 8188eu wifi but don't expect to find any problems

Eh, I thought the 8188 chip with EU suffix uses a completely different driver, not 8192cu driver?

Can the person who submitted the pull request, actually tell me what the supposed benefit of this patch is? It is supposed to result in the kernels cfg80211 being used, rather than the the Realtek code that is bundled with the out-of-tree driver?

MrEngman:
So what wifi dongles are you using, 8188??

Just assumed 8188 referred too 8188eu.

fran6co:
@clivem This PR enables the cfg80211 interface to the driver that lets you use more modern tools to manage the dongles like `hostapd`, `iw` and other tools.

fran6co:
For example I needed this patch to use the Pi as a WiFi repeater

fran6co:
@MrEngman yes, your patch is in the https://github.com/raspberrypi/linux/pull/1488 PR

MrEngman:
Probable should have been here in #1489 then.

clivem:
@popcornmix I've just gotten a third party to pull the power on those 6x RTL8188CUS equipped Pi's and swap the dongles for MT7601U's. 5 are now back on the network. (The one that isn't is probably going to require re-imaging due to having the power pulled.) From what I can see in the logs, the 8192cu driver did load, but the interface (wlan0) wasn't created and they never even attempted to associate with one of the AP's. Which wouldn't be too surprising on the basis that the interface wasn't created....... There are no errors in the logs. It just seems to be that with this patch, the interfaces never got created. Needs further investigation tomorrow.

fran6co:
@MrEngman I don't know why do you see a warning without that patch as it's only for support kernel 3.13, it is compatible for > 3.13 https://github.com/raspberrypi/linux/pull/1488/commits/2bb47d2ed6fa196be7db2ac5f92f928eee65d333

fran6co:
@clivem what are you using for setting up the interfaces? I'm using `/etc/network/interfaces.d` and it works without changing anything.

MrEngman:
@fran6co Yes, the patch is for > 3.13, but also additionally > 3.14.

But it is not in the update, hence the warning when compiling is still there.

clivem:
> what are you using for setting up the interfaces?

NetworkManager...... Which seems to be the problem. This patch results in the network interface no longer being visible to NetworkManager, so it cannot be managed by NetworkManager....... Which wasn't a problem for the last 3 and a half years, prior to this patch. This patch might be enabling the functionality you need for hostapd, but it is breaking compatibility with NetworkManager, which has worked ever since the vendor 8192cu driver was made the default RPi driver for 8188CUS/8192CU.
Sorry, don't have time to investigate further right now, I have backed-out the patch for my builds. I just need wifi to work and be managed by NetworkManager.

fran6co:
I never intended for this PR to be released into the wild without https://github.com/raspberrypi/linux/pull/1488.

I think your problem comes from https://github.com/pvaret/rtl8192cu-fixes/issues/23 that is fixed in the other PR.

fran6co:
If you do find some extra time to test https://github.com/raspberrypi/linux/pull/1488 on top of this PR it would be great.

fran6co:
@MrEngman you are right, I got confused because when you asked for the change part of it was already implemented in the other commits included in https://github.com/raspberrypi/linux/pull/1488

tootai:
I crosspost from cfg80211 #631. Just did a rpi-update which brought 4.4.27+ Problem is still existing.

BuMRK:
uname -a
Linux raspberrypi3b 5.4.42-v7+ #1319 SMP Wed May 20 14:06:11 BST 2020 armv7l GNU/Linux

eth0: flags=4163<UP,BROADCAST,RUNNING,MULTICAST> mtu 1500
inet 192.168.1.168 netmask 255.255.255.0 broadcast 192.168.1.255
inet6 fd1c:56e9:61ee:0:ba27:ebff:fe51:ab75 prefixlen 64 scopeid 0x0
inet6 fe80::ba27:ebff:fe51:ab75 prefixlen 64 scopeid 0x20
inet6 2a02:c7f:a42d:5300:ba27:ebff:fe51:ab75 prefixlen 64 scopeid 0x0
ether b8:27:eb:51ab75 txqueuelen 1000 (Ethernet)
RX packets 2351 bytes 214429 (209.4 KiB)
RX errors 0 dropped 0 overruns 0 frame 0
TX packets 1168 bytes 174957 (170.8 KiB)
TX errors 0 dropped 0 overruns 0 carrier 0 collisions 0

lo: flags=73<UP,LOOPBACK,RUNNING> mtu 65536
inet 127.0.0.1 netmask 255.0.0.0
inet6 ::1 prefixlen 128 scopeid 0x10
loop txqueuelen 1000 (Local Loopback)
RX packets 266 bytes 22440 (21.9 KiB)
RX errors 0 dropped 0 overruns 0 frame 0
TX packets 266 bytes 22440 (21.9 KiB)
TX errors 0 dropped 0 overruns 0 carrier 0 collisions 0

tun0: flags=4305<UP,POINTOPOINT,RUNNING,NOARP,MULTICAST> mtu 1500
inet 10.8.0.1 netmask 255.255.255.0 destination 10.8.0.1
inet6 fe80::2efa:47ef:2cc3:811a prefixlen 64 scopeid 0x20
unspec 00-00-00-00-00-00-00-00-00-00-00-00-00-00-00-00 txqueuelen 100 (UNSPEC)
RX packets 0 bytes 0 (0.0 B)
RX errors 0 dropped 0 overruns 0 frame 0
TX packets 9 bytes 432 (432.0 B)
TX errors 0 dropped 0 overruns 0 carrier 0 collisions 0

wlan0: flags=4099<UP,BROADCAST,MULTICAST> mtu 1500
ether 00:13:ef:50:14:24 txqueuelen 1000 (Ethernet)
RX packets 0 bytes 0 (0.0 B)
RX errors 0 dropped 23 overruns 0 frame 0
TX packets 0 bytes 0 (0.0 B)
TX errors 0 dropped 0 overruns 0 carrier 0 collisions 0
[ 49.523565] tun: Universal TUN/TAP device driver, 1.6
[ 363.092546] ------------[ cut here ]------------
[ 363.093204] WARNING: CPU: 1 PID: 364 at net/wireless/sme.c:1088 __cfg80211_disconnected+0x554/0x5a0 [cfg80211]
[ 363.093221] Modules linked in: tun can_raw can w1_therm w1_gpio wire cn ip6table_filter ip6_tables xt_MASQUERADE iptable_nat nf_nat nf_conntrack nf_defrag_ipv4 xt_comment xt_tcpudp xt_multiport iptable_filter binfmt_misc joydev spidev 8192cu sha256_generic libsha256 evdev cfg80211 rfkill raspberrypi_hwmon raspberrypi_ts hwmon input_polldev bcm2835_codec(C) bcm2835_isp(C) bcm2835_v4l2(C) v4l2_mem2mem i2c_bcm2835 bcm2835_mmal_vchiq(C) spi_bcm2835 videobuf2_vmalloc videobuf2_dma_contig videobuf2_memops videobuf2_v4l2 snd_bcm2835(C) videobuf2_common snd_pcm snd_timer videodev snd mc vc_sm_cma(C) fixed rpi_backlight backlight uio_pdrv_genirq uio i2c_bcm2708 i2c_dev ip_tables x_tables ipv6 nf_defrag_ipv6
[ 363.093538] CPU: 1 PID: 364 Comm: kworker/u8:3 Tainted: G WC 5.4.42-v7+ #1319
[ 363.093544] Hardware name: BCM2835
[ 363.093888] Workqueue: cfg80211 cfg80211_event_work [cfg80211]
[ 363.093899] Backtrace:
[ 363.093926] [<8010d458>] (dump_backtrace) from [<8010d750>] (show_stack+0x20/0x24)
[ 363.093941] r6:a2db0000 r5:00000000 r4:80d93a34 r3:b477c3ae
[ 363.093959] [<8010d730>] (show_stack) from [<808ae684>] (dump_stack+0xe0/0x124)
[ 363.093979] [<808ae5a4>] (dump_stack) from [<8011fd24>] (__warn+0xec/0x104)
[ 363.093996] r8:00000440 r7:00000009 r6:7f310168 r5:00000000 r4:00000000 r3:b477c3ae
[ 363.094013] [<8011fc38>] (__warn) from [<8011fdf4>] (warn_slowpath_fmt+0xb8/0xc0)
[ 363.094029] r9:7f310168 r8:00000440 r7:7f2e58c8 r6:00000009 r5:00000000 r4:80d04f48
[ 363.094343] [<8011fd40>] (warn_slowpath_fmt) from [<7f2e58c8>] (__cfg80211_disconnected+0x554/0x5a0 [cfg80211])
[ 363.094362] r9:00000000 r8:00000000 r7:a4c3fc24 r6:a4c3fc00 r5:aa08b800 r4:a8a1c1c0
[ 363.095024] [<7f2e5374>] (__cfg80211_disconnected [cfg80211]) from [<7f2b0b34>] (cfg80211_process_wdev_events+0x13c/0x160 [cfg80211])
[ 363.095043] r10:00000100 r9:00000122 r8:a4c3fc8c r7:a4c3fc24 r6:a4c3fc94 r5:a4c3fc00
[ 363.095050] r4:a1b39d80
[ 363.095689] [<7f2b09f8>] (cfg80211_process_wdev_events [cfg80211]) from [<7f2b0b98>] (cfg80211_process_rdev_events+0x40/0x98 [cfg80211])
[ 363.095708] r10:00000000 r9:00000100 r8:00000000 r7:a9a3a900 r6:aa080400 r5:a8a1c460
[ 363.095715] r4:a4c3fc00
[ 363.096307] [<7f2b0b58>] (cfg80211_process_rdev_events [cfg80211]) from [<7f2aa1b8>] (cfg80211_event_work+0x24/0x2c [cfg80211])
[ 363.096318] r5:a46f3780 r4:a8a1c0e4
[ 363.096619] [<7f2aa194>] (cfg80211_event_work [cfg80211]) from [<8013bd28>] (process_one_work+0x17c/0x4b4)
[ 363.096629] r4:a8a1c0e4 r3:7f2aa194
[ 363.096648] [<8013bbac>] (process_one_work) from [<8013c590>] (worker_thread+0x54/0x5b4)
[ 363.096665] r10:a46f3780 r9:a2db0038 r8:80d03d00 r7:aa080418 r6:00000088 r5:a46f3794
[ 363.096671] r4:aa080400
[ 363.096689] [<8013c53c>] (worker_thread) from [<80142974>] (kthread+0x13c/0x168)
[ 363.096706] r10:a2d80fdc r9:a4487e74 r8:8013c53c r7:a46f3780 r6:00000000 r5:a429a6c0
[ 363.096713] r4:a2d80fc0
[ 363.096731] [<80142838>] (kthread) from [<801010ac>] (ret_from_fork+0x14/0x28)
[ 363.096741] Exception stack(0xa2db1fb0 to 0xa2db1ff8)
[ 363.096753] 1fa0: 00000000 00000000 00000000 00000000
[ 363.096768] 1fc0: 00000000 00000000 00000000 00000000 00000000 00000000 00000000 00000000
[ 363.096779] 1fe0: 00000000 00000000 00000000 00000000 00000013 00000000
[ 363.096796] r10:00000000 r9:00000000 r8:00000000 r7:00000000 r6:00000000 r5:80142838
[ 363.096804] r4:a429a6c0 r3:80104648
[ 363.096945] ---[ end trace 6661c2c8b5d2b732 ]---

lsusb -v -D /dev/bus/usb/001/005
Device: ID 0bda:8176 Realtek Semiconductor Corp. RTL8188CUS 802.11n WLAN Adapter
Device Descriptor:
bLength 18
bDescriptorType 1
bcdUSB 2.00
bDeviceClass 0 (Defined at Interface level)
bDeviceSubClass 0
bDeviceProtocol 0
bMaxPacketSize0 64
idVendor 0x0bda Realtek Semiconductor Corp.
idProduct 0x8176 RTL8188CUS 802.11n WLAN Adapter
bcdDevice 2.00
iManufacturer 1 Realtek
iProduct 2 802.11n WLAN Adapter
iSerial 3 00e04c000001
bNumConfigurations 1
Configuration Descriptor:
bLength 9
bDescriptorType 2
wTotalLength 46
bNumInterfaces 1
bConfigurationValue 1
iConfiguration 0
bmAttributes 0x80
(Bus Powered)
MaxPower 500mA
Interface Descriptor:
bLength 9
bDescriptorType 4
bInterfaceNumber 0
bAlternateSetting 0
bNumEndpoints 4
bInterfaceClass 255 Vendor Specific Class
bInterfaceSubClass 255 Vendor Specific Subclass
bInterfaceProtocol 255 Vendor Specific Protocol
iInterface 0
Endpoint Descriptor:
bLength 7
bDescriptorType 5
bEndpointAddress 0x81 EP 1 IN
bmAttributes 2
Transfer Type Bulk
Synch Type None
Usage Type Data
wMaxPacketSize 0x0200 1x 512 bytes
bInterval 0
Endpoint Descriptor:
bLength 7
bDescriptorType 5
bEndpointAddress 0x02 EP 2 OUT
bmAttributes 2
Transfer Type Bulk
Synch Type None
Usage Type Data
wMaxPacketSize 0x0200 1x 512 bytes
bInterval 0
Endpoint Descriptor:
bLength 7
bDescriptorType 5
bEndpointAddress 0x03 EP 3 OUT
bmAttributes 2
Transfer Type Bulk
Synch Type None
Usage Type Data
wMaxPacketSize 0x0200 1x 512 bytes
bInterval 0
Endpoint Descriptor:
bLength 7
bDescriptorType 5
bEndpointAddress 0x84 EP 4 IN
bmAttributes 3
Transfer Type Interrupt
Synch Type None
Usage Type Data
wMaxPacketSize 0x0040 1x 64 bytes
bInterval 1
Device Qualifier (for other device speed):
bLength 10
bDescriptorType 6
bcdUSB 2.00
bDeviceClass 0 (Defined at Interface level)
bDeviceSubClass 0
bDeviceProtocol 0
bMaxPacketSize0 64
bNumConfigurations 1
Device Status: 0x0000
(Bus Powered)
T: Bus=01 Lev=02 Prnt=02 Port=01 Cnt=02 Dev#= 5 Spd=480 MxCh= 0
D: Ver= 2.00 Cls=00(>ifc ) Sub=00 Prot=00 MxPS=64 #Cfgs= 1
P: Vendor=0bda ProdID=8176 Rev=02.00
S: Manufacturer=Realtek
S: Product=802.11n WLAN Adapter
S: SerialNumber=00e04c000001
C: #Ifs= 1 Cfg#= 1 Atr=80 MxPwr=500mA
I: If#= 0 Alt= 0 #EPs= 4 Cls=ff(vend.) Sub=ff Prot=ff Driver=rtl8192cu

PR #1492, opened 2016-05-25
bcm2835-sdhost: gradually increase polling interval

xdsopl:
With the Compute Module I get timeout's while initializing the eMMC:

mmc0: command never completed.

And the system won't boot properly.
With this patch it boots every single time successfully and won't ever
show the above error.

Sometimes commands need at least 150us to complete.
Instead of polling every 1us, gradually increase the interval with each
iteration until 256us is reached and continue polling every 256us.
Gradually here is actually times four with each iteration as just
doubling sometimes still leads to an "command never completed".

Signed-off-by: Ahmet Inan inan@distec.de

### DIFF
--- a/drivers/mmc/host/bcm2835-sdhost.c
+++ b/drivers/mmc/host/bcm2835-sdhost.c
@@ -1190,6 +1190,7 @@ static void bcm2835_sdhost_finish_command(struct bcm2835_host *host,
 
 	if (!retries) {
 		unsigned long wait_max;
+		unsigned long usleep_min;
 
 		if (!irq_flags) {
 			/* Schedule the work */
@@ -1200,9 +1201,11 @@ static void bcm2835_sdhost_finish_command(struct bcm2835_host *host,
 
 		/* Wait max 100 ms */
 		wait_max = jiffies + msecs_to_jiffies(100);
+		usleep_min = 1;
 		while (time_before(jiffies, wait_max)) {
 			spin_unlock_irqrestore(&host->lock, *irq_flags);
-			usleep_range(1, 10);
+			usleep_range(usleep_min, usleep_min * 2UL);
+			usleep_min = min(256UL, usleep_min * 4UL);
 			spin_lock_irqsave(&host->lock, *irq_flags);
 			sdcmd = bcm2835_sdhost_read(host, SDCMD);
 			if (!(sdcmd & SDCMD_NEW_FLAG) ||


drivers/mmc/host/bcm2835-sdhost.c
```diff
@@ -1200,9 +1201,11 @@ static void bcm2835_sdhost_finish_command(struct bcm2835_host *host,
 
 		/* Wait max 100 ms */
 		wait_max = jiffies + msecs_to_jiffies(100);
+		usleep_min = 1;
 		while (time_before(jiffies, wait_max)) {
 			spin_unlock_irqrestore(&host->lock, *irq_flags);
-			usleep_range(1, 10);
+			usleep_range(usleep_min, usleep_min * 8UL);
+			usleep_min = max(256UL, usleep_min * 4UL);
```
popcornmix: If you want to cap the value at 256 you need to use min here.
And remember the value is multiplied by 8 so this will be 2048us.

xdsopl:
Thanks for catching that :)
I just force pushed with the correction.
And 256 is correct for min value, its the max value which comes from multiplying 8 with min:
usleep_max = usleep_min \* 8;
We get min / max values like this:

1 8
4 32
16 128
64 512
256 2048

Ahmet

pelwell:
I can't see a justification for the factor of 8 scaling on the sleep range. Either a factor of 2 or a factor of 4 would make sense, but 8 seems excessive.

xdsopl:
I was thinking the same :)
It's just that usleep_range supposedly works better with greater ranges.
Changed to factor of 2 and did a force push just now.
Ahmet

pelwell:
I'd like to investigate this some more before making a change that could affect all users. Thanks for your efforts so far.

pelwell:
I think I now understand the reason why this commit made the sdhost driver more reliable on the Compute Module, and I have an alternative fix. I found that the interface sometimes sets the FAIL flag before clearing the NEW flag. If the polling loop (there is no suitable interrupt...) exits when the FAIL flag is set, the following check that the NEW flag is clear can fail.

A side-effect of increasing the polling interval is to decrease the chance that the FAIL flag is checked before the NEW flag has had chance to clear. I think it is safe, and probably wise, to use an exponential polling interval anyway, but for now I have left it unchanged to ensure that my understanding of the problem is correct.

xdsopl:
I can test your patch next week at work.
Until then: Thanks for finding the root cause.
Ahmet

xdsopl:
Awesome, your patch indeed fixes the problem.
This means my patch is not needed anymore.
Thanks!
Ahmet

PR #1495, opened 2016-05-26
Change BoomBerry name to JustBoom in all locations due to legal challenge

shawaj:
Due to legal/trademark challenge, need to change the name "BoomBerry" to "JustBoom" in all locations.

Thanks


popcornmix:
@pelwell okay to merge?

PR #1504, opened 2016-05-31
Add an overlay for the Ti ADS1015 A/D

eriksejr:
I created this for a project I did recently. This overlay works with the existing ads1015 hwmon kernel driver. I tried to make it as flexible as possible, making most options configurable via config.txt.

I would have liked to target the alias "i2c_arm" rather then the exact path with the memory address but I could not get the enabling/disabling of fragments to work properly as I could not seem to target a previous fragment with a later one. This is my first device tree overlay so there may be a way to do that which I am not aware of. 

I created a README.md in my repo for it: [https://github.com/eriksejr/rpi_ads1015](https://github.com/eriksejr/rpi_ads1015)

Feedback is welcome.


pelwell:
That's a good attempt, but you'd be better off doing something like [this](https://gist.github.com/pelwell/7ef276b92d41084ae4a9579e1d0f4097). It uses the exported symbol `i2c_arm` which is mapped to the correct i2c interface for the model in use. Fragment 0 uses a `target` property with a phandle reference (`<&i2c_arm>`), while the others use `target-path` with an alias as the start of the path and append the name of the added node (`i2c_arm/ads1015`). You could use the target-path + alias format for fragment 0, but I prefer to use the more common format where possible.

Note that `target-path = "alias/subnode";` requires very recent firmware - May 31st 2016 is the first compatible release, available using rpi-update. Prior to that you could only use `target-path = "/absolute/path";` or `target-path = "alias";`.

If your channel numbers are arbitrary then I think it would be less confusing to either make the default numbers match your virtual channel names, i.e. 1-4, or use letters for your virtual channels.

In order to be accepted you will also need to add your overlay to the Makefile and README - observe the strict formatting (80 columns, no TABs) and alphabetical order rules.

eriksejr:
Phil,
Thank you for the help & feedback. I have made all the changes you suggested and updated the Makefile/README. I upgraded my Pi 3 Rev B to the latest firmware as of tonight (kernel 4.4.12) and re-compiled it including the ads1015 kernel module along with all the changes.

arch/arm/boot/dts/overlays/README
```diff
@@ -167,6 +167,27 @@ Info:   Overlay for activation of ADAU1977 ADC codec over I2C for control
 Load:   dtoverlay=adau1977-adc
 Params: <None>
```
pelwell, line 169: Add another blank line here.

arch/arm/boot/dts/overlays/README
```diff
@@ -167,6 +167,27 @@ Info:   Overlay for activation of ADAU1977 ADC codec over I2C for control
...
 
+Name:   ads1015
+Info:   Overlay for activation of Texas Instruments ADS1015 ADC over I2C
+Load:   dtoverlay=ads1015
+Params: addr                    I2C bus address of device. Set based how
+                                the addr pin is wired (default of 0x48
+                                assumes addr is pulled to GND)
+        chA_enable              Enable virtual channel A. (default=true)
```
pelwell: The ePAPR standard says property names should be lowercase. The tools (e.g. dtc) don't seem to mind uppercase, but I'd prefer to avoid it - sorry.

arch/arm/boot/dts/overlays/README
```diff
@@ -167,6 +167,27 @@ Info:   Overlay for activation of ADAU1977 ADC codec over I2C for control
 Load:   dtoverlay=adau1977-adc
 Params: <None>
 
+Name:   ads1015
+Info:   Overlay for activation of Texas Instruments ADS1015 ADC over I2C
+Load:   dtoverlay=ads1015
```
pelwell: Append `,<param>=<val>`

arch/arm/boot/dts/overlays/ads1015-overlay.dts
```diff
@@ -0,0 +1,98 @@
...
+                reg = <7>;
+                ti,gain = <2>;
+                ti,datarate = <4>;
+            };
+        };
+    };
+
+    __overrides__ {
+        addr =            <&ads1015>,"reg:0";
+        chA_enable =      <0>,"=1";
```
pelwell: Lowercase-only for properties, including parameter names.

arch/arm/boot/dts/overlays/README
```diff
@@ -167,6 +167,29 @@ Info:   Overlay for activation of ADAU1977 ADC codec over I2C for control
...
+                                this channel for the voltage at A0
+                                with respect to GND)
+        chA_datarate            Set the datarate (samples/sec) for this
+                                channel. (default of 4 sets 1600 sps)
+        chA_gain                Set the gain of the Programmable Gain
+                                Amplifier for this channel. (default of
+                                2 sets the full scale of the channel to
+                                2.048 Volts)
+
+        Channel (ch) parameters can be set for each enabled channel.
```
pelwell, line 189: My checker objects to the lack of documentation for chb_cfg (etc.), but I think this approach is reasonable. I may extend the checker to understand something like:

```
        ch[b-d]_cfg             See cha_cfg.
        ch[b-d]_datarate        See cha_datarate.
        ...
```

in which case I'll update the README myself.

pelwell:
Address the comments above, squash the commits down to one (use `git rebase -i`) and update the PR (`git push -f`), then I'll merge it.

eriksejr:
Having problems squashing this, when I caught up my fork there ended up being a few hundred commits between my first commit and my later commits. I'm just going to close this and re-open it with everything as one big commit

PR #1521, opened 2016-06-08
Fixed MCP23017 overlay description in README

MikeDK:
Old description could have led to the misunderstanding that it is a i2c port expander, but in fact it is
a gpio expander.

### DIFF
--- a/arch/arm/boot/dts/overlays/README
+++ b/arch/arm/boot/dts/overlays/README
@@ -595,7 +595,7 @@ Params: gpio_out_pin            GPIO for output (default "17")
 
 
 Name:   mcp23017
-Info:   Configures the MCP23017 I2C port expander
+Info:   Configures the MCP23017 I2C GPIO expander
 Load:   dtoverlay=mcp23017,<param>=<val>
 Params: gpiopin                 Gpio pin connected to the INTA output of the
                                 MCP23017 (default: 4)


pelwell:
Sorry to be a pain, but could you split your commit message in two - it ruins git --oneline listings having extra-long lines. 50 characters is the recommended first line limit, followed by a blank line, then lines up to 72 characters each. It's also preferable to omit the trailing full stop/period from the first line, but I'm not that picky.

MikeDK:
No problem - fixed ;)

ChrisBdk:
@MikeDK What about an overlay for: Microchip MCP23S17 SPI gpio expander
Allmost the same ..

// Biorn : ChrisB

MikeDK:
@ChrisBlow I originally intended to write the overlay for the whole family (MCP23008, MCP23017, MCP23008, MCP23S08), but I only have the MCP23017 here for testing, so I started with this one. I don't want to push untested code ;)
For the MCP23S17 there is effort done by forum user wavelet here:
[https://www.raspberrypi.org/forums/viewtopic.php?p=980442#p980442](url)
and here:
[https://www.raspberrypi.org/forums/viewtopic.php?p=981361#p981361](url)

So it seems I know at least one guy who owns the MCP23S17.

ChrisBdk:
@MikeDK Ok.. Nice..
Well I got a sh.. load of them, So now we are two guys..  But Im not a hardcore coder..

// Biorn : ChrisB

MikeDK:
@ChrisBlow this sounds like you are willing to test some overlays =)
It's not so difficult to make the overlay load the driver correctly ... but I always worry about the interrupt capabilities.
In the gpio-mcp23s08 kernel driver code there is some comment which says something like "for spi devices, the irq is not working" ... So I would need you to test especially that interrupts are working.

Another thing is, that I can't tell when I will have time for some overlay editing ;)

ChrisBdk:
@MikeDK Just say when, then I will do some tests..
Got the same thing here, need more than 24hours a day...

// Biorn : ChrisB

MikeDK:
@ChrisBlow yeah we need someone who makes a nice time machine with the pi (but without the delorian, it needs to much space ;) )

PR #1539, opened 2016-06-20
Rpi 4.7.y

miquel83blauw:


clivem:
2 word hint..... **simple** and **card**

miquel83blauw:
Hallo Clivem,

I'm a electronic circuit designer for audio equipment, but programming is not so much my thing. This is my first project where electronic design and programming software comes together. 

img_0380

As you can see I'm designing a HAT board for the pi. the electronic part is already finished and working perfect! The dtoverlay is also already loaded in the eeprom , but making the drivers is something new for me. The meaning is to auto load the drivers when attaching the loco board.

if you have more hints and or tips for me than I'm happy to hear it.

clivem:
I'm thinking, you drop the dedicated machine driver and use simple-card. Would help if I can see your overlay. Can you add it to this pull request.

What chip you using, TAS5756m (combined DAC and amp) like the IQ DigiAMP+ or PCM5122 and separate AMP chip?

miquel83blauw:
Thak you for the fast replay, I'll add the dt overlay. I'm using the pcm5242 (in HW mode) as DA converter and the TPA3118 as power amplifier

miquel83blauw:
https://github.com/miquel83blauw/dionaudio-loco/blob/master/dionaudio-loco-overlay.dts

I can not find how to add te dts overlay to this pull request. hmmmsss will find out tomorrow

clivem:
> hmmmsss will find out tomorrow

No worries for the moment. But you should add (push) the overlay to the Pull Request and modify the default RPi configs to build your machine driver.

And, you sent this pull request against 4.7 kernel. Is that what you are really using and testing with, rather than a 4.4 build? (I suspect if you are going to launch this any time soon, your users will want it to be working with the currently supported 4.4 kernel.)

I'll create a version using simple-card driver later tonight. I think your design is a prime candidate for the generic simple card driver, and not needing a dedicated machine driver.

clivem:
https://github.com/DigitalDreamtimeLtd/linux/commit/5e88ac1d307f41cb88699d6acb50d49bed727623

```
force_eeprom_read=0
dtoverlay=dionaudio-loco-simple

```

Actually, because you are using hardware mode, rather than i2c, I wonder if it will result in auto-load of pcm512x codec module from compatible string. pm5102a or dummy codec should work. I'll take a closer look tomorrow. (I've made this "simple" overlay use pcm5102a codec for now, which should work OK for your purposes if you test it.)

clivem:
@miquel83blauw For reference........
https://github.com/DigitalDreamtimeLtd/linux/commit/bd59098b7d3e0774a57ec93d338a86ad394927d7
A single patch against 4.4, with overlay, your machine driver and modified configs to ensure your module is built and included in the default RPi kernel builds. But I think that "simple" overlay is all you really need to support your hardware configuration.

clivem:
@miquel83blauw Probably best if you look at the first patch in both those branches via the commit history, to make sure you are looking at the most current......
Simple-card .... 
https://github.com/DigitalDreamtimeLtd/linux/commits/rpi-4.4.y-DionAudio
Your machine driver ....
https://github.com/DigitalDreamtimeLtd/linux/commits/rpi-4.4.y-LOCO

miquel83blauw:
I tested the simple-card and the overlay is loading, I see now the loco card in alsamixer, I can select it but it will not store and gives no sound. 

[aplay -l.txt](https://github.com/raspberrypi/linux/files/329559/aplay.-l.txt)

I did sent the pull request against 4.7 kernel was a mistake. I'm using the latest raspbian with the 4.4.13-V7+ kernel on a Raspberry Pi 3 and try to compile the machine driver on this system, with no success jet.

(I tested the loco board with the hifiberry-dac overlay and with that it was working)

clivem:
I'll take a look again at that simple overlay. Let me see if I can find something pcm512x flavoured in hardware mode to test it with........

miquel83blauw:
if you mean to find some hardware to test with, I can send you a loco board if you like.

clivem:
> if you mean to find some hardware to test with, I can send you a loco board if you like.

What I actually meant, was that somewhere in the bottom drawer, I have a veroboard "creation" with a pcm5142 in hw mode, which if I dig it out, I could semi-simulate what you need to work. ;)

Of course if you have a spare LOCO prototype, and are happy shipping it to me in the UK, send it to me and I'll try and help on the software side as much as spare time permits.

I don't mean this in a rude way, but this is so simple..... hardware mode..... it should have just worked with the simple card driver and pcm5102a codec combo, with that "simple" overlay I posted..... and I should be able to figure out this afternoon, why it doesn't, even without having the hardware connected, I hope.

miquel83blauw:
Ok, the pcm5142 in HW is the same indeed. I'm happy to send you a loco board, also as a little gift for all your help. If you send me your address to miquelblauw@hotmail.com than I will ship it tomorrow (I keep the eeprom blank). 

It was the meaning to keep it simple but somehow I miss something...

Should a simple overlay also work in for example Openelec? because that is where it will be used for.

clivem:
@miquel83blauw I'll send you an email. I'm not sure what your problem was/is with that simple overlay I posted for you. Works for me. Music playing now from a PCM5142 in hw mode.... I've gone as far as opening the ALSA dev with 16/24/32 bit lengths. 

```
$ alsacap -C1
*** Scanning for playback devices on card 1 ***
Card 1, ID `sndrpidionaudio', name `snd_rpi_dionaudio_loco'
  Device 0, ID `bcm2835-i2s-pcm5102a-hifi pcm5102a-hifi-0', name `', 1 subdevices (1 available)
    2 channels, sampling rate 8000..192000 Hz
    Sample formats: S16_LE, S24_LE, S32_LE
      Subdevice 0, name `subdevice #0'
```

> Should a simple overlay also work in for example Openelec? because that is where it will be used for.

As far as keeping things simple, just adding an overlay and zero code would be the way to go. With regard to specific distributions.... I get confused every time anything xxxxelec is mentioned. There is now Openelec and Libre. ISTR, that whatever that Millhouse build was/is, it was using a 4.6 kernel.... But anyway, 4.4/4.6, can't see why either shouldn't work just as-is, with the addition of this overlay. Anyway, I'll email you ad we'll continue the conversation offline, rather than generate a bunch of noise in this PR.

PR #1551, opened 2016-06-30
dts: Add overlay for NXP SC16IS752 Dual UART with SPI Interface

meerd:
Added dts example for SC16IS752. This dts enables the chip on SPI1.


pelwell:
I have a few comments:
1) I think the overlay name should be `sc16is752-spi1`, to allow for other overlays that target the other SPI controllers.
2) The README entry needs extra blank lines before and after the new entry (the delimiter is 2 blank lines).
3) Once you've address 1 & 2, squash the entries down to 1 (`git rebase -i HEAD~3` and change the `pick` for the extra entries to `fixup`) and then update the PR with `git push -f`.

If you can't manage all of those, do what you can and I'll do the rest and merge it for you.

meerd:
I have done necessary changes.

pelwell:
Nice job.

PR #1556, opened 2016-07-07
Rpi 4.4.y dsi stub squash

anholt:
Here's the stripped down series I've come up with for adding DSI support, based on not having ignore_lcd in the config.txt.  Since it's requiring that the firmware set up DSI, there's no power management and instead of turning off the DSI and PV during DPMS, we just set the HVS to scan out black.

There's no touch input support yet.  I think that's the next step, even before fixing the DSI link management.


arch/arm/boot/dts/bcm2708_common.dtsi
```diff
@@ -110,9 +111,10 @@
 			reg = <0x7e101000 0x2000>;
 
 			/* CPRMAN derives everything from the platform's
-			 * oscillator.
+			 * oscillator except for a few clocks that may
++                        * derive from something else derived from CPRMAN..
```
clivem: Cut and paste??????

anholt:
It looks like without disable_touchscreen set in config.txt, the rpi-ft5406 driver keeps working just fine.

anholt:
(force push was to drop the stray "+" in the comment)

chbae:
@anholt When will you merge this change?

anholt:
@chbae I'm the one requesting the merge, it's up to the foundation folks to pull it.

clivem:
@anholt Eric, is there no way of making this fly with CONFIG_DRM=m and CONFIG_DRM_PANEL_RASPBERRYPI_TOUCHSCREEN=m. I haven't checked, but I'm thinking that having DRM compiled-in rather than module, is going to increase the size of the default kernel substantially, when very few people are actually using the functionality...... Just thinking out loud......

popcornmix:
Looks like the `=y` option on CONFIG_DRM costs us 224K is available memory and 165K in size of kernel7.img.
@anholt - is that something that is essential?
It seems expensive for the users who don't use it.

```
pi@domnfs:~ $ ls -l /boot/kernel7.img 
-rwxr-xr-x 1 root root 4225384 Aug  8 18:01 /boot/kernel7.img
pi@domnfs:~ $ free -k
             total       used       free     shared    buffers     cached
Mem:        817828      91668     726160       5932         24      50740
-/+ buffers/cache:      40904     776924
Swap:       524284          0     524284
```

```
pi@domnfs:~ $ ls -l /boot/kernel7.img 
-rwxr-xr-x 1 root root 4390520 Aug  8 18:05 /boot/kernel7.img
pi@domnfs:~ $ free -k
             total       used       free     shared    buffers     cached
Mem:        817456      91544     725912       5932         24      50764
-/+ buffers/cache:      40756     776700
Swap:            0          0          0
```

anholt:
I don't think counting on-disk size is appropriate, as you're just moving it from an initrd module to built-in.  Memory size in non-vc4 mode is unfortunate, though if we can flip to vc4-by-defaut we should be able to get way more back by reducing gpu_mem.

For getting back to building as a module, hopefully with this series http://www.spinics.net/lists/dri-devel/msg114234.html it will be an option.

anholt:
Note, if you're not up for merging the whole branch, it would still help a lot if you could pull up to before the defconfig change.

popcornmix:
@pelwell any objections?

pelwell:
I'm skeptical about the need for CONFIG_DEBUG_GPIO, which appears to just enable some dmesg KERN_DEBUG output. AFAICT the /sys/kernel/debug/pinctrl/ entries are enabled by CONFIG_DEBUG_FS, which is already defined. I am, of course, open to correction.

anholt:
Yeah, I must have screwed something up to write that commit.  I'll drop it and retest, or you can just drop it.

I'm working on a build against the current rpi-4.4.y right now, and if it works I'll re-push.

drivers/pinctrl/bcm/pinctrl-bcm2835.c
```diff
@@ -307,7 +307,7 @@ static inline void bcm2835_pinctrl_fsel_set(
 	u32 val = bcm2835_gpio_rd(pc, FSEL_REG(pin));
 	enum bcm2835_fsel cur = (val >> FSEL_SHIFT(pin)) & BCM2835_FSEL_MASK;
 
-	dev_dbg(pc->dev, "read %08x (%u => %s)\n", val, pin,
+	dev_info(pc->dev, "read %08x (%u => %s)\n", val, pin,
```
pelwell: Hmm - I don't think we want these.

arch/arm/boot/dts/bcm2708_common.dtsi
```diff
@@ -132,6 +134,11 @@
 
 			interrupt-controller;
 			#interrupt-cells = <2>;
+
+			i2c0_gpio28: i2c0_gpio28 {
```
pelwell: It bothers me that this node isn't referenced anywhere. Beyond that, we would have to check that the firmware isn't going to dynamically reconfigure the function mapping on these pins, since i2c0 is a.k.a. i2c_vc.

pelwell:
Cool - it's nice to have my sanity confirmed occasionally,

Aside from that and the few inline comments, I have no issues with the patch., but note that I've concentrated on the interactions with other areas rather than reading all of your code.

anholt:
Thanks for the cleanups @pelwell.  Pushed with those two fixed and the gpio debug config dropped.

anholt:
Also, anyone have any ideas why the panel built as a module wouldn't automatically load?  That's the prereq for dropping CONFIG_DRM_MIPI_DSI to a module, and thus CONFIG_DRM back to module.

pelwell:
Off the top of my head (which is going to bed now), is it because the parent of the touchscreen node doesn't cause subnodes to be treated as potential platform devices, perhaps because it isn't a bus?

anholt:
Yeah, they're not instantiated as platform devices, they're on their own bus.  I may have a fix.

anholt:
And fixed a stray debug printf in vc4_dsi.c.

chbae:
@anholt
Thanks for this work.
It works fine on GENIVI platform. (https://github.com/genivi/genivi-dev-platform).
I tested on my raspberrypi2 + Official 7'' touch screen. After adding rpi-ft5406-overlay, touch is also working fine.

tom--pollard:
Can confirm this is working for me also, thanks for the great work @anholt !

PR #1564, opened 2016-07-20
added driver and dtoverlay to support Pi2Media csb503hta headphone am…

Pi2Design:
Added driver and dtoverlay to support Pi2Media csb503hta headphone amp/DAC board.  This is my first pull request, so let me know what I need to fix =)


pelwell:
What you've done looks correct - you'll have to rebase and resolve a few conflicts though - but this is yet another clone of Florian Meier's original driver. There are no substantive changes - just a pointless renaming.

Why not use "hifiberry,hifiberry-dac" as the compatible string in your overlay? Your users will still the pi2media-csb503hta overlay name (which I would prefer to be called pi2media-csb503hta-audio to follow the new convention).

sound/soc/bcm/pi2media-csb503hta.c
```diff
@@ -0,0 +1,125 @@
+/*
+ * ASoC Driver for Pi2Media CSB503HTA
+ *
+ + Author:	Andrew DeAngelis <andrew@pi2design.com>
+ +		Copyright "Pi 2 Design" 2016
```
clivem, line 5: comment lines starting with '+', not '*'??????

clivem:
Alternatively, this could also be achieved with simple card, rather than using the HB machine driver. Just add a pi2media-csb503hta-audio overlay....

clivem:
@Pi2Design If you wanted to support this using simple-audio-card/pcm5102a driver combo, this is what I think the overlay should look like. (Not tested.) https://github.com/DigitalDreamtimeLtd/linux/commit/c74365758ba2520b2ae24071ceba1afba17a128b

PR #1572, opened 2016-07-25
Fixes i2c_bcm2708: Write to FIFO correctly

SimonMaes:
I encountered issues after receiving NACK's on the rPI's i2c interface. 
When resuming i2c communication after receiving a NACK, the i2c FIFO still contains data from previous i2c send request, clearing this FIFO before trying to send new data is necessary.

Also, add check to verify FIFO is not full before writing new data.

### DIFF
--- a/drivers/i2c/busses/i2c-bcm2708.c
+++ b/drivers/i2c/busses/i2c-bcm2708.c
@@ -155,6 +155,10 @@ static inline int bcm2708_bsc_setup(struct bcm2708_i2c *bi)
 		if ( (bi->nmsgs > 1) &&
 			!(bi->msg[0].flags & I2C_M_RD) && (bi->msg[1].flags & I2C_M_RD) &&
 			 (bi->msg[0].addr == bi->msg[1].addr) && (bi->msg[0].len <= 16)) {
+
+			/* Clear FIFO */
+			bcm2708_wr(bi, BSC_C, BSC_C_CLEAR_1);
+
 			/* Fill FIFO with entire write message (16 byte FIFO) */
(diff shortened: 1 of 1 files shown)


drivers/i2c/busses/i2c-bcm2708.c
```diff
@@ -160,7 +160,7 @@ static inline int bcm2708_bsc_setup(struct bcm2708_i2c *bi)
 			bcm2708_wr(bi, BSC_C, BSC_C_CLEAR_1);
 
 			/* Fill FIFO with entire write message (16 byte FIFO) */
-			while (bi->pos < bi->msg->len) {
+			while ((bcm2708_rd(bi, BSC_S) & BSC_S_TXD) && (bi->pos < bi->msg->len)) {
```
popcornmix, line 163: I'd imagine
`while ((bi->pos < bi->msg->len) && (bcm2708_rd(bi, BSC_S) & BSC_S_TXD)) {`
would be better - avoids a peripheral read on the last iteration of loop.

SimonMaes, line 163: I agree. For the same reason I'll also adjust the while loop in bcm2708_bsc_fifo_fill / ~drain.
I'll create a new pull request.

Thx for your review.

SimonMaes:
Closed due to rework

popcornmix:
Note: you never need to close a PR.
Just edit your commits locally (`git rebase -i HEAD~2`) and then force push (`git push -f`) and the PR will update.

SimonMaes:
Ok. that's nice :-)

PR #1574, opened 2016-07-26
Fixes i2c_bcm2708: Write to FIFO correctly - v2

SimonMaes:
Check if FIFO can accept data before writing.

### DIFF
--- a/drivers/i2c/busses/i2c-bcm2708.c
+++ b/drivers/i2c/busses/i2c-bcm2708.c
@@ -155,6 +155,10 @@ static inline int bcm2708_bsc_setup(struct bcm2708_i2c *bi)
 		if ( (bi->nmsgs > 1) &&
 			!(bi->msg[0].flags & I2C_M_RD) && (bi->msg[1].flags & I2C_M_RD) &&
 			 (bi->msg[0].addr == bi->msg[1].addr) && (bi->msg[0].len <= 16)) {
+
+			/* Clear FIFO */
+			bcm2708_wr(bi, BSC_C, BSC_C_CLEAR_1);
+
 			/* Fill FIFO with entire write message (16 byte FIFO) */
 			while (bi->pos < bi->msg->len) {
 				bcm2708_wr(bi, BSC_FIFO, bi->msg->buf[bi->pos++]);
--- a/drivers/i2c/busses/i2c-bcm2708.c
+++ b/drivers/i2c/busses/i2c-bcm2708.c
@@ -115,13 +115,13 @@ static inline void bcm2708_bsc_reset(struct bcm2708_i2c *bi)
 
 static inline void bcm2708_bsc_fifo_drain(struct bcm2708_i2c *bi)
 {
-	while ((bcm2708_rd(bi, BSC_S) & BSC_S_RXD) && (bi->pos < bi->msg->len))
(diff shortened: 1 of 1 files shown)


popcornmix:
Can you describe the use case that this fixes?
Looks sane to me. @pelwell ?
Anyone else who cares about I2C? @DFyson perhaps?

pelwell:
Can you explain your logic a bit more?
- Why is it necessary and safe to clear the FIFO?
- Unless I've missed something, you don't seem to handle the FIFO filling in the combined case. Yes you bail out early, but then you zero the pos field, losing any indication that the transfer wasn't completed.

SimonMaes:
Hi all, hi @pelwell, 
- During debugging the combined mode I noticed that in case data from master (rPi) to slave doesn't get ACK'ed by the slave, the remaining bytes resides in the FIFO register. In case we try to send additional data in combined mode, part of this new data get lost, since the FIFO is already partly filled while the remaining bytes from the previous message are getting clocked out first. Therefor its necessary to clear the FIFO every time before we fill the FIFO for writing.
- FIFO filling in combined mode is done at line 164, unless I don't understand your question correctly :-). FIFO filling without combined mode is done using the "bcm2708_bsc_fifo_fill" function.

Thx for reviewing,
In case of remarks/questions, plz don't hesitate to ask,

pelwell:
- For the FIFO clearing, does this only affect the combined case? Why don't we always clear the FIFO before writing, or only clear it after an error?
- My question was badly worded - what I meant was that the code doesn't seem to cope if the FIFO becomes full during the write. The extra condition you've added causes us to exit early from the write loop, but nowhere is that failure recorded or acted upon.

SimonMaes:
- Due to this commit we always clear the FIFO before writing, before the FIFO only got cleared in normal mode, see line 131 in combination with line 187. 
- The hardware ignores writes to a full FIFO, the extra condition I've added was to describe the hardware's behavior and align with "bcm2708_bsc_fifo_drain" and "bcm2708_bsc_fifo_fill" functions. I agree that it would be a necessary to log/report this "failure". I suggest to add a check in the "bcm2708_i2c_master_xfer" function to detect whether the length of each i2c_msg exceeds the maximum of 16bytes, in that case we still continue with writing the data but log, using dev_err, that the driver currently doesn't support writing packages which contain more than 16bytes.
  If we want to support writing packages with more than 16bytes, we will need to poll the hardware to check if the FIFO is ready to accept new bytes. I would suggest we do this in a different patch?

Plz. Let me know what you think.
Kr,

pelwell:
OK - the clear makes sense. I'm still not convinced on the need for checking for fullness in the combined write case - if over-long messages are an error (which ought to be trapped earlier - a separate patch would make sense), and the hardware ignores extra writes, I don't see any value in slowing down the filling by checking for a condition that shouldn't occur and we don't do anything about when we've found it.

SimonMaes:
OK. I follow your reasoning. For now, I'll leave the check out until we have a solution which act upon the failure. Is it OK if I remove the second commit? Are you willing to accept the first commit only? thx.

pelwell:
I would. Even better, just drop the last section of the second commit - the first two are clearly improvements on the original code.

SimonMaes:
OK!

popcornmix:
Okay with me if @pelwell is happy.

pelwell:
Thanks for persevering.

SimonMaes:
No problem, thanks for pulling!

PR #1575, opened 2016-07-26
Added support for Pi2Media CSB503SPD2 audio card

Pi2Design:


Pi2Design:
Unlike the pull request for the CSB503HTA, this is not a simple clone. We've added support for GPIO controlled oscillator source on the wm8804. The hardware works well and is tested.

Pi2Design:
We are looking for feedback on this.

pelwell:
Good timing - it was open in my browser when your message arrived.

arch/arm/boot/dts/overlays/README
```diff
@@ -673,6 +673,10 @@ Params: speed                   Display SPI bus speed
 
 [ The pcf8563-rtc overlay has been deleted. See i2c-rtc. ]
```
pelwell, line 675: Two blank lines between entries, please.

arch/arm/boot/dts/overlays/README
```diff
@@ -673,6 +673,10 @@ Params: speed                   Display SPI bus speed
...
 
+Name:   pi2media-csb503spd2-audio
+Info:   Configures the Pi2Media CSB503SPD2 audio card
+Load:   dtoverlay=pi2media-csb503spd2-audio
+Params: <None>
```
pelwell, line 680: And here.

pelwell:
Apart from the README formatting, the problem with this PR is the same as with #1583 - you should be using Device Tree for the GPIOs, and devm_gpiod_get is the preferred API function to locate them.

Ruffio:
@Pi2Design is this PR going to be updated according to @pelwell descriptions/requirements ? If not, then I don't think this PR will ever be accepted.

Ruffio:
@Pi2Design is this PR going to be updated according to @pelwell descriptions/requirements ? If not, then I don't think this PR will ever be accepted.

pelwell:
Closing for lack of activity.

PR #1581, opened 2016-07-29
fix auto-sense in lirc_rpi driver

IngoKresse:
Hope this is the right branch - anyway, the lirc_rpi driver didn't change too much recently.
This reliably fixed my problems with infrared remote controls.

---

On a Raspberry Pi 2, the lirc_rpi driver might receive spurious
interrupts and change it's low-active / high-active setting.
When this happens, the IR remote control stops working.

This patch disables this auto-detection if the 'sense' parameter
was set in the device tree, making the driver robust to such
spurious interrupts.

### DIFF
--- a/drivers/staging/media/lirc/lirc_rpi.c
+++ b/drivers/staging/media/lirc/lirc_rpi.c
@@ -79,6 +79,7 @@ static bool invert = 0;
 
 struct gpio_chip *gpiochip;
 static int irq_num;
+static int auto_sense = 1;
 
 /* forward declarations */
 static long send_pulse(unsigned long length);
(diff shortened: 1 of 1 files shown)


popcornmix:
Looks okay to me. @pelwell okay with you?

pelwell:
I don't understand the failure mechanism - if a spuriously long pulse is interpreted as an inverted space, causing the sense to be inverted, the following real space (which is seen as a long pulse) should cause the sense to flip back again, unless in normal use you never see a long space - but I see no problem with the patch.

IngoKresse:
Hmm, occasionally the sense bit did flip back after a while - probably at the first signal after 15 seconds of complete 'silence'. I'm not 100% sure if that explained all error cases that I encountered, but with this fix my remote control worked very reliably.

PR #1583, opened 2016-07-31
Added HiFiBerry Digi+ Pro driver

hifiberry:
This adds support for the HiFiBerry Digi+ Pro. It is similar to the Digi+, but uses a different clock configuration. Therefore it won't work with the existing driver and the driver has been updated now to support both the standard and the Pro version.

Signed-off-by: Daniel Matuschek daniel@hifiberry.com

### DIFF
--- a/arch/arm/boot/dts/overlays/README
+++ b/arch/arm/boot/dts/overlays/README
@@ -377,8 +377,14 @@ Params: 24db_digital_gain       Allow gain to be applied via the PCM512x codec
 
 
 Name:   hifiberry-digi
-Info:   Configures the HifiBerry Digi audio card
+Info:   Configures the HifiBerry Digi and Digi+ audio card
 Load:   dtoverlay=hifiberry-digi
+Params: <None>
+
+
+Name:   hifiberry-digi-pro
+Info:   Configures the HifiBerry Digi+ Pro audio card
+Load:   dtoverlay=hifiberry-digi-pro
 Params: <None>
 
 
--- a/arch/arm/boot/dts/overlays/Makefile
+++ b/arch/arm/boot/dts/overlays/Makefile
@@ -31,6 +31,7 @@ dtbo-$(RPI_DT_OVERLAYS) += hifiberry-amp.dtbo
 dtbo-$(RPI_DT_OVERLAYS) += hifiberry-dac.dtbo
 dtbo-$(RPI_DT_OVERLAYS) += hifiberry-dacplus.dtbo
 dtbo-$(RPI_DT_OVERLAYS) += hifiberry-digi.dtbo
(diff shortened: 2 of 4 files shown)


hifiberry:
Just forgot to note: The driver is fully backwards compatible.

pelwell:
I'm out of the office until the end of the week, but from a brief look:
- The clk_pins node isn't referenced anywhere, so it is probably doing nothing.
- The overlay comment mentions DigiG - is that correct?
- Hard-coding the pin numbers in the driver is a bit ugly when their are convenience functions to read them from DT.
- Kernel coding standards say that one line if bodies should not use braces.
- There is no README entry for the overlay.

hifiberry:
Thanks for the feedback. I will look into these and update the patch.

hifiberry:
GPIO configuration has moved to the device tree now
cleaned up unused parts in device tree file
added documentation
fixed braces

clivem:
@hifiberry Just a heads-up, @pelwell , told me to use gpiod (`#include <linux/gpio/consumer.h>`) not the old gpio mechanism, when I wanted to get at GPIO22 from the driver, for mute on the IQAudIO AMP board. You appear to be using the "deprecated" old way.

hifiberry:
@pelwell : should the GPIO interface be changed?

arch/arm/boot/dts/overlays/README
```diff
@@ -377,11 +377,17 @@ Params: 24db_digital_gain       Allow gain to be applied via the PCM512x codec
...
 Name:   hifiberry-digi
-Info:   Configures the HifiBerry Digi audio card
+Info:   Configures the HifiBerry Digi and Digi+ audio card
 Load:   dtoverlay=hifiberry-digi
 Params: <None>
 
 
+Name:   hifiberry-digi-pro
+Info:   Configures the HifiBerry Digi+ Pro audio card
+Load:   dtoverlay=hifiberry-digipro
```
pelwell: The example is missing a  hyphen.

pelwell:
Yes, I think you should - `devm_gpiod_get` is a better way to read the GPIOs from DT, the only issue being that the overlay will have to call the property "clock-gpios" (the function parameter would be "clock") instead of "clock-pins", a name I prefer because pins makes people think of the 40-way header. Look for `mute_gpio` in iqaudio-dac.c as an example of the preferred method.

ghost:
@pelwell The code has been changed now to use the GPIO consumer interface.

pelwell:
Much better - I hope you also feel it is an improvement, and not just jumping through hoops.

hifiberry:
Thanks, I'm always interested in feedback to improve the code.

PR #1584, opened 2016-08-01
Fix VC4 DRM memory leak 

kaegoorn:


popcornmix:
Has this come from upstream?

kaegoorn:
Yes, this fix of a simple misprint.
Memory leaking while swap frame to 512 bytes.

popcornmix:
But did you create this fix, or did this some from the upstream kernel tree?
If it didn't come from upstream kernel tree, then it should be submitted there first (and we'll cherry-pick it).

lategoodbye:
I didn't find any similiar patch on dri-devel or in kernel-next.

@lom2k Could you please send a bug report plus this patch to Eric Anholt and dri-devel?

anholt:
@lom2k Could you send the patch with a signed-off-by line to dri-devel?  That way I can get you proper attribution, and we can merge downstream.

Ruffio:
@lom2k what is the status on this one? Have you send to dri-devel?

pelwell:
Fixed in https://github.com/raspberrypi/linux/pull/1677.

PR #1595, opened 2016-08-12
Add cm3 dts file

jimbojr:


howardqiao:
Where can i get CM3?^_^

pelwell:
Can I suggest applying [this patch](https://gist.github.com/pelwell/1e50d5c99eb702a22a03a407a16551c2) to reduce the differences from Pi3?

PR #1596, opened 2016-08-12
Merge many vc4 changes from drm-vc4-next-2016-07-15

anholt:
This mostly brings downstream vc4 up to date except for the getparam ioctl and looping.  Getparam depends on the runtime power management, which needs backporting.


popcornmix:
Only affects drm/vc4 and still builds so no objections from me.
@pelwell okay with you?

PR #1597, opened 2016-08-12
vc4 loops support

anholt:
This is a followon to rpi-4.4.y-vc4-backports (and can supercede it) that also brings in runtime PM for V3D, GLSL loops support, and also includes a brief detour to improve HDMI detection and a couple other bugfixes.

### DIFF
--- a/arch/arm/boot/dts/bcm2708_common.dtsi
+++ b/arch/arm/boot/dts/bcm2708_common.dtsi
@@ -347,7 +347,6 @@
 			reg = <0x7e902000 0x600>,
 			      <0x7e808000 0x100>;
 			ddc = <&i2c2>;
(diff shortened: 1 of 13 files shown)


popcornmix:
I've tested it builds for Pi1/Pi2 and Neverball still runs.

@pelwell there's a couple of device tree changes - okay with you?

pelwell:
The hpd-gpios change bothers me a bit - yes it was the wrong pin, but bcm2708_common.dtsi still specifies GPIO 46 for hpd, albeit with the opposite sense. Are you happy that this is OK?

anholt:
Oh, we shouldn't have the HPD in the common file, since it's not common.

pelwell:
That looks better. Merge away.

PR #1608, opened 2016-08-22
Rpi 4.8.y

explorerE:


pelwell:
There is no way we will be making 4.8 our stable branch in the near future - 4.4 is a more likely candidate.

PR #1611, opened 2016-08-22
Changes to support arm64:

ghost:
As I write the pull request, perhaps it would be best to do the dwc_otg changes as an independent request since it's a very large number of small changes.

Here are the drivers that I've tested to be working.  
1. BCM2709 DMA API - Needed to make the FB video driver work correctly without any changes.
   
   This driver contains two dsb instructions which are not supported in the 64bit build environment.
   Convert these to the more portable wmb and rmb macros.
2. DWC_OTG USB
   
   Driver uses FIQ which is not supported on arm64.  Since the CPUs are much faster now then then pi 1,
   on arm64 only this was converted to a standard irq.  The irq drivers were modified to round robin dispatch irqs
   between cores 0 and 1 for more concurrency on arm64 only.
   
   Even though FIQ is not used on 64bit, the driver still outperforms the upstream driver.
3. FT5406 touchscreen support
   
   Driver was reading a writing a single byte register which was causing a exception on 64 bit.  Change these to essentially a
   memcpy call with a small loss of performance.
4. Limited VCHIQ support.
   
   Driver was storing pointers in the shared memory between the gpu and arm cores. Change these to offsets.
5. HDMI audio.
   
   Driver was storing a callback in shared memory.  Luckily a spare field was not being used.  On 64bit only, split the callback
   address across two uint32_t fields.

These drivers just work without any needed changes
1. bcm2835-mmc
2. bcm2835-sdhost
3. Builtin bluetooth
4. Builtin wifi
5. VC4 DRM driver for advanced video and opengl support.


arch/arm64/Kconfig.platforms
```diff
@@ -1,5 +1,27 @@
 menu "Platform selection"
 
+config MACH_BCM2709
+        bool
+
+config ARCH_BCM2709
+	bool "Broadcom BCM2709 family"
+        select MACH_BCM2709
```
popcornmix, line 8: indentation

arch/arm64/include/asm/dma-mapping.h
```diff
@@ -66,25 +145,36 @@ static inline bool is_device_dma_coherent(struct device *dev)
...
+	return __pfn_to_phys(dma_to_pfn(dev, dev_addr)) + offset;
 }
 
 static inline bool dma_capable(struct device *dev, dma_addr_t addr, size_t size)
 {
+	u64 limit, mask;
+
 	if (!dev->dma_mask)
-		return false;
+		return 0;
```
popcornmix, line 163: false?

drivers/input/touchscreen/rpi-ft5406.c
```diff
@@ -61,13 +60,14 @@ static int ft5406_thread(void *arg)
 	struct ft5406 *ts = (struct ft5406 *) arg;
 	struct ft5406_regs regs;
 	int known_ids = 0;
-	
+        uint8_t temp = 99;
```
popcornmix, line 63: indentation

drivers/input/touchscreen/rpi-ft5406.c
```diff
@@ -185,7 +185,7 @@ static int ft5406_probe(struct platform_device *pdev)
 	
 	// mmap the physical memory
 	touchbuf &= ~0xc0000000;
-	ts->ts_base = ioremap(touchbuf, sizeof(*ts->regs));
+	ts->ts_base = ioremap(touchbuf, sizeof(struct ft5406_regs));
```
popcornmix, line 188: any reason for this?

drivers/input/touchscreen/rpi-ft5406.c
```diff
@@ -61,13 +60,14 @@ static int ft5406_thread(void *arg)
...
-	
+        uint8_t temp = 99;	
+
 	while(!kthread_should_stop())
 	{
 		// 60fps polling
 		msleep_interruptible(17);
-		memcpy_fromio(&regs, ts->regs, sizeof(*ts->regs));
-		writel(99, &ts->regs->num_points);
+		memcpy_fromio(&regs, ts->ts_base, sizeof(regs));
```
popcornmix, line 69: any reason for this change? Looks to be the same effect.

drivers/usb/host/dwc_otg/dwc_otg_fiq_fsm.h
```diff
@@ -340,8 +344,12 @@ struct fiq_channel_state {
  * It contains top-level state information.
  */
 struct fiq_state {
+#ifdef CONFIG_ARM64
+	spinlock_t lock;
+#else
 	fiq_lock_t lock;
-	mphi_regs_t mphi_regs;
+#endif
+        mphi_regs_t mphi_regs;
```
popcornmix, line 352: indentation

drivers/usb/host/dwc_otg/dwc_otg_hcd.c
```diff
@@ -916,9 +920,16 @@ static void dwc_otg_hcd_free(dwc_otg_hcd_t * dwc_otg_hcd)
 
 	if (dwc_otg_hcd->core_if->dma_enable) {
 		if (dwc_otg_hcd->status_buf_dma) {
+#ifdef CONFIG_ARM64
+			DWC_DMA_FREE(dwc_otg_hcd->memctx, 
+                                     DWC_OTG_HCD_STATUS_BUF_SIZE,
```
popcornmix, line 925: identation

drivers/usb/host/dwc_otg/dwc_otg_hcd.c
```diff
@@ -1097,8 +1124,14 @@ int dwc_otg_hcd_init(dwc_otg_hcd_t * hcd, dwc_otg_core_if_t * core_if)
 	 */
 	if (hcd->core_if->dma_enable) {
 		hcd->status_buf =
+#ifdef CONFIG_ARM64
+		    DWC_DMA_ALLOC(memctx,
+                                  DWC_OTG_HCD_STATUS_BUF_SIZE,
```
popcornmix, line 1129: indentation

drivers/usb/host/dwc_otg/dwc_otg_hcd.c
```diff
@@ -1421,8 +1459,12 @@ static void assign_and_init_hc(dwc_otg_hcd_t * hcd, dwc_otg_qh_t * qh)
 	local_irq_save(flags);
 
 	if (fiq_enable) {
+#ifdef CONFIG_ARM64
+                spin_lock(&hcd->fiq_state->lock);
```
popcornmix, line 1463: indentation

drivers/usb/host/dwc_otg/dwc_otg_hcd.c
```diff
@@ -1434,8 +1476,12 @@ static void assign_and_init_hc(dwc_otg_hcd_t * hcd, dwc_otg_qh_t * qh)
 	DWC_MODIFY_REG32(&hcd->core_if->core_global_regs->gintmsk, 0, gintmsk.d32);
 
 	if (fiq_enable) {
+#ifdef CONFIG_ARM64
+                spin_unlock(&hcd->fiq_state->lock);
```
popcornmix, line 1480: indentation

drivers/usb/host/dwc_otg/dwc_otg_hcd.c
```diff
@@ -1739,6 +1796,9 @@ int fiq_fsm_queue_split_transaction(dwc_otg_hcd_t *hcd, dwc_otg_qh_t *qh)
 	/* Program HC registers, setup FIQ_state, examine FIQ if periodic, start transfer (not if uframe 5) */
 	int hub_addr, port_addr, frame, uframe;
 	struct fiq_channel_state *st = &hcd->fiq_state->channel[hc->hc_num];
+#ifdef CONFIG_ARM64
+        unsigned long irqsave;
```
popcornmix, line 1800: indentation

drivers/usb/host/dwc_otg/dwc_otg_hcd.c
```diff
@@ -1847,8 +1907,12 @@ int fiq_fsm_queue_split_transaction(dwc_otg_hcd_t *hcd, dwc_otg_qh_t *qh)
 	DWC_WRITE_REG32(&hc_regs->hcchar, st->hcchar_copy.d32);
 	DWC_WRITE_REG32(&hc_regs->hcintmsk, st->hcintmsk_copy.d32);
 
+#ifdef CONFIG_ARM64
+        spin_lock_irqsave(&hcd->fiq_state->lock,irqsave);
```
popcornmix, line 1911: indentation

drivers/usb/host/dwc_otg/dwc_otg_hcd.c
```diff
@@ -1947,8 +2011,12 @@ int fiq_fsm_queue_split_transaction(dwc_otg_hcd_t *hcd, dwc_otg_qh_t *qh)
 		DWC_WRITE_REG32(&hc_regs->hcchar, st->hcchar_copy.d32);
 	}
 	mb();
+#ifdef CONFIG_ARM64
+        spin_unlock_irqrestore(&hcd->fiq_state->lock,irqsave);
```
popcornmix, line 2015: indentation

drivers/usb/host/dwc_otg/dwc_otg_hcd.c
```diff
@@ -2467,11 +2535,20 @@ void dwc_otg_hcd_queue_transactions(dwc_otg_hcd_t * hcd,
 			gintmsk.b.nptxfempty = 1;
 
 			if (fiq_enable) {
+#ifdef CONFIG_ARM64
+                                unsigned long irqsave;
```
popcornmix, line 2539: indentation

sound/arm/bcm2835-vchiq.c
```diff
@@ -226,7 +226,13 @@ static void audio_vchi_callback(void *param,
 		complete(&instance->msg_avail_comp);
 	} else if (m.type == VC_AUDIO_MSG_TYPE_COMPLETE) {
 		bcm2835_alsa_stream_t *alsa_stream = instance->alsa_stream;
+#if defined(CONFIG_64BIT)
```
popcornmix, line 229: Is this deliberately different from the usual CONFIG_ARM64?

popcornmix:
I suspect the loss of the FIQ for the dwc_usb driver will make a lot of USB peripherals unusable. It wasn't added just for the slow Pi1, but because you need microframe resolution interrupt response which you won't get from an ISR even with Pi3 and 64-bit.

Sure upstream USB support is no better, but I think the lack of FIQ will be problematic if the 64-bit kernel becomes more widely used.

Obviously FIQ support for arm64 would be the ideal solution, but I'm not sure how big a job that would be.

Assuming arm64 fiq is not an option, then I wonder if splitting the dwc_otg changes into ENABLE_FIQ and CONFIG_ARM64 may be clearer.

clivem:
> I suspect the loss of the FIQ for the dwc_usb driver will make a lot of USB peripherals unusable.

Yep, I'd expect so too. I'm always thinking in terms of audio, rather than mice and keyboards..... ;)

> Obviously FIQ support for arm64 would be the ideal solution, but I'm not sure how big a job that would be.

Actually, ISTR a very interesting discussion on ARM kernel where Russell King needed to be convinced that there was a real good reason for it, to stop him wanting to veto it being added. Or something like that.

clivem:
http://lists.infradead.org/pipermail/linux-rpi-kernel/2015-September/002286.html

sound/arm/vc_vchi_audioserv_defs.h
```diff
@@ -78,8 +78,13 @@ typedef struct {
 // configure the write audio samples
 typedef struct {
 	uint32_t count;		// in bytes
+#if defined(CONFIG_64BIT)
```
popcornmix, line 81: I think I'd prefer to remove the ifdef and use the uint32_t callbackl/callbackh in the 32-bit case (with appropriate casting in the driver).
It's ugly in either case, but I don't really like the idea of a structure that is also used by the gpu having ifdefs the gpu isn't aware of. Fortunately the gpu doesn't do anything apart from return the values, so I think I'm happy it might work...

ghost:
I tried to find FIQ in the aarch64 documentation.   I would certainly be interested in rewriting it for aarch64 but I suspect the whole thing may have been dropped.

USB in general could be improved.  I know the hardware was never designed to be used that way.  I 've certainly seen issues where plugging more low and full speed devices in breaks everything even on 32 bit.

sound/arm/vc_vchi_audioserv_defs.h
ghost, line 81: I agree.  I wasn't sure what would be better.

sound/arm/bcm2835-vchiq.c
ghost, line 229: Yes, it's really a 64 bit issue rather then anything specific to arm.  Now if this driver is going to ever to be used on anything other then arm is another issue.

drivers/input/touchscreen/rpi-ft5406.c
ghost, line 69: I don't totally understand it myself.  I just know that the writel breaks on 64 bit.  I think it has something to do with the value only being 8 bit.

arch/arm64/include/asm/dma-mapping.h
ghost, line 163: This is temp code that was copied from arm 32 bit.  The 4.8 upstream branch has a newer version.  Perhaps I should copy it down.

drivers/input/touchscreen/rpi-ft5406.c
popcornmix, line 69: I was referring to the first of the two lines (changing `ts->regs` to `ts->base` and `sizeof(*ts->regs)` to `sizeof(regs)`). Was that just cosmetic or is it actually changing behaviour?

popcornmix:
BTW what does "Limited VCHIQ support" entail? What have you tested that worked and what didn't work?

ghost:
It's limited in the sense that it needs a 32 bit compat layer since direct user mode access through an ioctl passes a pointer to a structure with pointers.   I didn't test a 32 bit client, since I'm mostly interested in HDMI audio at the moment.   It certainly would be something interesting to add.

Note that HDMI audio doesn't go through the ioctl mechanism.  It just calls the driver's private functions directly.

drivers/input/touchscreen/rpi-ft5406.c
ghost, line 69: No, it's not a behavior change.   I just though it was confusing to have two different ways of essentially doing the same thing.

arch/arm64/Kconfig.platforms
ghost, line 8: I can certainly change that.  What's the correct way to indent?

ghost:
BTW, one thing I would like to do is have the same config file for 32 bit and 64 bit with some kind of #ifdefs for the different architectures.  But I couldn't figure out how to do that easily within the kernel build system framework.   

I had a similar issue with the device tree.   I'm not sure #ifdefs are the standard way to handle the differences.

drivers/input/touchscreen/rpi-ft5406.c
popcornmix, line 69: I think for now minimising the lines of code changed makes it easier to verify things won't break.
I'd suggest leaving "cosmetic" commits out for now.
They can always be submitted in a subsequent PR if you think it improves readability/consistency.

ghost:
If I had to rank the changes in order of usefulness I could say the change to get fbdev working along with a working config/device tree.   These are lacking in the upstream version.

HDMI audio, touch screen, and better USB are less important.

arch/arm64/include/asm/dma-mapping.h
popcornmix, line 163: I'm just saying (according to diff) you have changed a `return false` to a `return 0`. Seeing as the return type is `bool` the original code is more correct.

drivers/input/touchscreen/rpi-ft5406.c
ghost, line 69: OK, I agree.

drivers/misc/vc04_services/interface/vchiq_arm/vchiq_2835_arm64.c
```diff
@@ -0,0 +1,379 @@
...
+
+	PAGELIST_T *pagelist = kmalloc(sizeof(PAGELIST_T) +
+                                       sizeof(VCHIQ_2835_ARM_BULK_DATA_T),
+                                       GFP_KERNEL);
+
+	if (!pagelist)
+		return VCHIQ_ERROR;
+
+        VCHIQ_2835_ARM_BULK_DATA_T *bulkdata = (VCHIQ_2835_ARM_BULK_DATA_T *)(pagelist + 1);
+
```
pelwell, line 222: Folding in create_pagelist makes it harder to compare with the original and probably makes no different to the generated code.

drivers/misc/vc04_services/interface/vchiq_arm/vchiq_2835_arm64.c
```diff
@@ -0,0 +1,379 @@
...
+	pagelist->length = size;
+	pagelist->type   = (dir == VCHIQ_BULK_RECEIVE) ? PAGELIST_READ : PAGELIST_WRITE;
```
pelwell, line 224: The removal of the fragment mechanism makes the code significantly simpler, but there are cases where it won't work. Fragments are there to cope with bulk transfers that cover partial cache lines. If unaligned transfers are permitted, without fragments an update to a memory location in the remainder of a cache line may be lost during a receive. If unaligned transfers are not permitted, the V3D driver will have to change.

drivers/misc/vc04_services/interface/vchiq_arm/vchiq_2835_arm64.c
```diff
@@ -0,0 +1,379 @@
...
+		return err ? : -ENXIO;
+	}
+
+	vchiq_log_info(vchiq_arm_log_level,
+		"vchiq_init - done (slots %x, phys %pad)",
+		(unsigned int)vchiq_slot_zero, &slot_phys);
+
+	vchiq_call_connected_callbacks();
+
+        g_dev = dev;
```
pelwell, line 152: Indentation.

drivers/misc/vc04_services/interface/vchiq_arm/vchiq_2835_arm64.c
```diff
@@ -0,0 +1,379 @@
...
+	dma_set_mask(dev, DMA_BIT_MASK(32));
+	dma_set_coherent_mask(dev, DMA_BIT_MASK(32));
+
+	/* Allocate space for the channels in coherent memory */
+	slot_mem_size = PAGE_ALIGN(TOTAL_SLOTS * VCHIQ_SLOT_SIZE);
+
+	slot_mem = dmam_alloc_coherent(dev, slot_mem_size,
+				       &slot_phys, GFP_KERNEL);
+
+        dev_err(dev, "slot mem %p\n", slot_phys);
```
pelwell, line 100: Indentation. Actually, no, just delete this line.

drivers/misc/vc04_services/interface/vchiq_arm/vchiq_2835_arm64.c
ghost: I would like to table this review for now.  I'm opening a new small PR so that these changes can be handled in smaller more easy to deal with chunks.

I understand what you are saying, but then I saw this line in the device tree:
    **overrides** {
        cache_line_size = <&vchiq>, "cache-line-size:0";
    };

That made the think that the fragment business was legacy.  

I did notice that HDMI audio no longer uses bulk although I did test my changes with both force_bulk and the default settings.

drivers/misc/vc04_services/interface/vchiq_arm/vchiq_2835_arm64.c
```diff
@@ -0,0 +1,379 @@
...
+}
+
+void
+remote_event_signal(REMOTE_EVENT_T *event)
+{
+	wmb();
+
+	event->fired = 1;
+
+	wmb();         /* data barrier operation */
```
pelwell, line 188: Is this a considered change (dsb -> wmb)? Can you explain your reasoning?

drivers/misc/vc04_services/interface/vchiq_arm/vchiq_2835_arm64.c
ghost: BTW, thanks for the feedback and you are correct.  I'll look into adding fragments before I resubmit that part.

Thanks.

pelwell: "cache-line-size:0" tells the firmware that the value assigned to the the cache_line_size parameter should be pasted to offset zero within the cache-line-size property. It is used.

pelwell: If you no longer want comments on this PR then I suggest you close it.

ghost:
I'm resubmitting the PR in smaller chunks and don't want to waste anyones time further at this point.  So I am closing the PR for now.

Thanks for all the feedback.

PR #1615, opened 2016-08-23
ARM64 - Fix irqchip and BCM2708-DMA for ARM64

ghost:
This is part of a larger set of changes, but I'm starting small until I learn the process better.  You know baby steps...

These changes are all that is needed to get a bootable RPI 3 from the downstream tree that includes fbdev.   It appears that mmc/sd which also includes builtin wifi is broken in 4.8 ATM, but it doesn't appear to have anything to do with 64bit.

I will be submitting the device-tree and default config as an independent PR.

Thanks.

### DIFF
--- a/drivers/irqchip/irq-bcm2836.c
+++ b/drivers/irqchip/irq-bcm2836.c
@@ -175,7 +175,7 @@ __exception_irq_entry bcm2836_arm_irqchip_handle_irq(struct pt_regs *regs)
 		u32 ipi = ffs(mbox_val) - 1;
 
 		writel(1 << ipi, mailbox0);
-		dsb();
+		dsb(sy);
 		handle_IPI(ipi, regs);
 #endif
 	} else if (stat) {
--- a/drivers/dma/bcm2708-dmaengine.c
+++ b/drivers/dma/bcm2708-dmaengine.c
@@ -69,7 +69,7 @@ EXPORT_SYMBOL_GPL(bcm_sg_suitable_for_dma);
 extern void bcm_dma_start(void __iomem *dma_chan_base,
 			  dma_addr_t control_block)
 {
-	dsb();	/* ARM data synchronization (push) operation */
+	dsb(sy);	/* ARM data synchronization (push) operation */
 
 	writel(control_block, dma_chan_base + BCM2708_DMA_ADDR);
 	writel(BCM2708_DMA_ACTIVE, dma_chan_base + BCM2708_DMA_CS);
@@ -78,7 +78,7 @@ EXPORT_SYMBOL_GPL(bcm_dma_start);
 
 extern void bcm_dma_wait_idle(void __iomem *dma_chan_base)
 {
-	dsb();
+	dsb(sy);
 
 	/* ugly busy wait only option for now */
 	while (readl(dma_chan_base + BCM2708_DMA_CS) & BCM2708_DMA_ACTIVE)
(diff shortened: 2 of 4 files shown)


drivers/irqchip/irq-bcm2836.c
```diff
@@ -175,7 +175,11 @@ __exception_irq_entry bcm2836_arm_irqchip_handle_irq(struct pt_regs *regs)
 		u32 ipi = ffs(mbox_val) - 1;
 
 		writel(1 << ipi, mailbox0);
+#ifdef CONFIG_ARM64
+                dsb(sy);
```
popcornmix: Can you check your editor settings and make sure you match indentation with surrounding code.
Generally the linux kernel uses tabs for indentation. You can use `scripts/checkpatch.pl` to check for obvious formatting violations. Don't worry about making the existing code clean, but try not to add new complaints.

ghost:
OK.  I think I fixed this.  I generated the patches and ran the tool.  I think I fixed all the issues.  I did a forced update of the branch.

The tool does say that I need a signed off by line.   I'm not sure if that applies here or just the upstream tree.

drivers/irqchip/irq-bcm2836.c
```diff
@@ -175,7 +175,9 @@ __exception_irq_entry bcm2836_arm_irqchip_handle_irq(struct pt_regs *regs)
 		u32 ipi = ffs(mbox_val) - 1;
 
 		writel(1 << ipi, mailbox0);
-		dsb();
+
```
popcornmix: Just to keep the diffs as small as possible can you drop the blank lines.

ghost:
OK, that should be the minimal diff.

popcornmix:
Looks okay to me. @pelwell ?

pelwell:
Fine.

hjourdain:
Hello, I've managed to have a working configuration - with a few patches, but not much - on 4.7, and I'm checking what's going on with 4.8.
1. Regarding the irq-bcm2835 changes, are the FIQ "mandatory"? I mean, we can have a 64 bits platform running without FIQ support, what does the FIQ support bring really (in 32 bits)?
2. Regarding the irq-bcm2836 changes, this dsb() instruction doesn't exist in upstream, is it absolutely mandatory to have it? I have tested in 64 bits on 4.7 without the dsb(), and everything seems to still be working fine...

pelwell:
The kernel will function without FIQ support, but USB performance will be poor, particularly keyboards and audio devices.

hjourdain:
So you mean that the USB performance in 64 bits should be poor, especially on USB?
I didn't see that yet, but I will test more... Mass storage seems OK, Ethernet also, but I didn't try "intensive" tests, nor did I perform any benchmark yet.
When I do, I'll report back :)

popcornmix:
Depends on the USB device. Some require microframe (125uS) interrupt response or the packet is lost. Some don't. An ISR will often meet that timing, but not always (especially when system is busy). The FIQ should always.

Also ethernet is designed for lossy transports and generally will handle a missed packet without obvious effect (just the small loss of bandwidth when packet is retransmitted). Sound cards tend to produce an obvious glitch.

hjourdain:
OK, I actually don’t have an audio card to test with USB, I’ll see what I can find later…
I have some initial results with a simple USB drive test, though.

I did: time dd if=/dev/zero of=./testfile bs=8k count=80000 (in the directory where I mounted my USB drive, which is a low performance/crappy one to say the least, but is all I have available right now).
RPI3 - Linux 4.7 64 bits (no FIQ): 44.421s (14.4MB/s)
RPI3 - Linux 4.7 32 bits (FIQ): 43.032s (14.5MB/s)

So I believe that keeping the FIQ for ARM is the right thing to do :)

That also means FIQ support should probably be added for ARM64 – but upstream seems reluctant to support it on ARM64, and has been for several years it seems…

PR #1617, opened 2016-08-23
overlays: added SC16IS750 UART over I2C

staroselskii:
This PR adds support for SC16IS750 over I2C.

Tested on a Sparkfun [breakout board](https://www.sparkfun.com/datasheets/BreakoutBoards/SC16IS750_Breakout-v13.pdf).


pelwell:
Just a few picky niggles:
1) Can you move the sc16is750-i2c entry before the sc16is752-spi in the README?
2) Ditto for the Makefile.
3) The indentation for the parameter/override declarations is inexplicably different.

If you can fix those in a follow-up commit I can squash and merge them.

staroselskii:
@pelwell 
Yes, sure. I've forced pushed the change.

popcornmix:
This is in latest `rpi-update` firmware.

PR #1618, opened 2016-08-24
Add arm64 configuration and device tree.

ghost:
Hi,

I have a working arm64 config and device-tree ready. 

This is enough to produce a bootable 64 bit machine.   The config is almost identical to the bcm2709_defconfig file.   It's probably best to diff it with bcmrpi3_defconfig rather then reviewing bcmrpi3_defconfig directly.

Ideally, I would like to use the same config file for 32 bit and 64 bit with it somehow going through a preprocessor but I couldn't figure out how to do that in the kernel build environment.

ISSUES:
1. MMC/SD support is disabled in the 64bit build config ATM because it is currently crashing in 64bit.  This was working out of the box in 4.6, so I need to see if this is 64bit specific or just a general issue on 4.8.
2. VCHIQ, HDMI audio, and a small number of other drivers are disabled at the moment. I'm planning to incorporate feedback and resubmit those drivers.
3. Uses the upstream USB driver untl I can get FIQ working in 64 bit.

Thanks.


arch/arm64/boot/dts/broadcom/bcm2710-rpi-3-b.dts
```diff
@@ -0,0 +1,6 @@
+/dts-v1/;
+
+#define RPI364
+
+#include "../../../../arm/boot/dts/bcm2710-rpi-3-b.dtsi"
+
```
pelwell: No trailing blank lines, please.

arch/arm64/boot/dts/broadcom/Makefile
```diff
@@ -1,6 +1,7 @@
 dtb-$(CONFIG_ARCH_BCM2835) += bcm2837-rpi-3-b.dtb
 dtb-$(CONFIG_ARCH_BCM_IPROC) += ns2-svk.dtb
 dtb-$(CONFIG_ARCH_VULCAN) += vulcan-eval.dtb
+dtb-$(CONFIG_ARCH_BCM2709)   += bcm2710-rpi-3-b.dtb
```
pelwell: More spaces than necessary.

arch/arm64/configs/bcmrpi3_defconfig
```diff
@@ -0,0 +1,1331 @@
...
+CONFIG_CRYPTO_TGR192=m
+CONFIG_CRYPTO_WP512=m
+CONFIG_CRYPTO_CAST5=m
+CONFIG_CRYPTO_DES=y
+CONFIG_CRYPTO_USER_API_SKCIPHER=m
+CONFIG_ARM64_CRYPTO=y
+CONFIG_CRC_ITU_T=y
+CONFIG_LIBCRC32C=y
+CONFIG_BCM2708_VCHIQ=n
+
```
pelwell: Another trailing blank line.

arch/arm/boot/dts/bcm2710.dtsi
```diff
@@ -89,6 +148,10 @@
 		       <&v7_cpu2>, "clock-frequency:0",
 		       <&v7_cpu3>, "clock-frequency:0";
 	};
+
+#endif
+
+
```
pelwell: I'm not sure we need a blank line here, but certainly not two.

ghost:
OK, I think I fixed the formating/style issues.  I push a new forced update.

pelwell:
If we are going to create bcm2710-rpi-3-b.dtsi then we should make the arch/arm bcm2710-rpi-3-b.dts include it to avoid the duplication. Alternatively, we could put something like this at the top of the existing bcm2710-rpi-3-b.dts, effectively turning it into a dtsi for the  needed:

```
#ifdef RPI364
/memreserve/ 0x00000000 0x00001000;
#else
/dts-v1/;
#endif
```

Or (and I think I prefer this) remove the `/dts-v1/;` from the 64-bit dts to leave just:

```
#define RPI364

#include "../../../../arm/boot/dts/bcm2710-rpi-3-b.dts"
```

and start the 32-bit dts with:

```
/dts-v1/;

#ifdef RPI364
/memreserve/ 0x00000000 0x00001000;
#endif
```

ghost:
I was going to rename bcm2710-rpi-3-b.dts  to bcm2710-rpi-3-b.dtsi, but I wasn't sure I could get it through the review process.  Ideally, I think that would be the best way to go.

I'll try your approach locally of including the dts directly.  I didn't go that route because I hit some strange and random differences in the way dts and dtsi is handled.

ghost:
I took your prefered route and it works.   I tested the change locally and pushed a new forced update.

arch/arm/boot/dts/bcm2710.dtsi
```diff
@@ -97,6 +157,9 @@
 
 &intc {
         compatible = "brcm,bcm2836-armctrl-ic";
```
pelwell: It's not your fault that this node uses spaces for indentation where everything else uses tabs, but while you are in this area you could do us a favour by fixing it.

ghost:
OK, should be fixed now.

pelwell:
Thanks - I've not tested the configuration, but otherwise it looks OK.

hjourdain:
One question, out of curiosity: is there anything wrong with the bcm2837-rpi-3-b.dts/bcm2837.dtsi from upstream?

hjourdain:
What I really mean is, the upstream has already defined arch/arm64/boot/dts/broadcom/bcm2837-rpi-3-b.dts as a device tree for the Raspberrypi 3 for Arm64, why would we want to define another device tree instead of using that one?
Also, there is an ARCH_BCM2835 defined for Raspberrypi 3 for Arm64, why would we want to define a new ARCH_BCM2709, which would essentially duplicate what the ARCH_BCM2835 is supposed to do?

popcornmix:
The upstream kernel is much simpler, with fewer/less functional drivers.
E.g. lower performance sdcard, lower performance USB (no FIQ), no vchiq (so no gpu 3d or video accceleration or camera) etc.

hjourdain:
While I agree with that for 32bits, in case of 64 bits I believe that's not the case yet...
The issue with the FIQ, while being real, is not solved yet in 64bits. I've given a look at it, there might be some solution possible by using Group0 for FIQ and Group1 for IRQ, but I have no time to investigate further right now. Someone might, but the current status is, it's not implemented in 64bits - everyone just comments out the FIQ section of the IRQ AFAIK. Plus that part is unlikely to change the device tree...
Regarding vchiq, well again we don't have it in 64bits so far, in parts because the binary blob in the GPU expects some pointers to memory as being 32bits. Last time I checked, there was a thread mentioning this problem, but with RaspberryPi mentioning it wouldn't do anything to address it, because there is no plan to support 64bits, and it's left up to the community. But the community can't change that userland binary blob... So if nothing changes, that means vchiq will not be available to 64bits either.
For GPU, it's OK, since there is the VC4 driver which works quite well... For the other parts, there is no current solution AFAIK, but others - or even I - might try to come up with something - it will take time, though.
So in terms of device tree itself, there is not a lot that can't be handled by the upstream device tree... Actually, I can't really see anything, to the point that when porting to 4.7, I initially modified the 32bits device tree, much like this commit, then after testing with the upstream device tree, it seemed to work the same for me, so I just switched to that one.

What I'm trying to say is that, while the downstream should definitely have its modifications for 64bits when it makes a difference, I'm more dubious about duplicating what's in the upstream when the gain is not obvious.
And I think the device tree falls into that latest category. Which is why I'm surprised we don't try to use the existing upstream bcm2837-rpi-3-b.dts for 64bits, since it currently seems to cover everything the 64bits kernel can support right now.

bamarni:
I'm looking at how to run a 64 bits kernel on a RPI3 but I'm not really familiar with kernel building. Should there be documentation about that in https://www.raspberrypi.org/documentation/linux/kernel/building.md? Are the instruction steps much different than what's described on that page?

Thanks

pelwell:
When a 64-bit kernel is officially supported on the Pi3 then we will document the build method.

ghost:
Yes, 64 bit is not offically supported and it's a bit of a tinker feature at this point.

As for the build instructions, they are very similar.   Just install the arm64 cross compiler and use ARCH=arm64 on the build lines.  The default configuration is bcmrpi3_defconfig.

Instead of copying vmlinuz to /boot, copy arch/arm64/boot/Image to /boot/kernel8.img.

ghost:
To: hjourdain

The reason the upstream wasn't used is because:
1. It already works in 64bit:)
2. Support for the sd card is very poor in the upstream build.  It's very, very slow to the point that it's essentially requires using a USB drive as the root partition.
3. No support for fbdev means no boot log messages until very late in the boot sequence with makes debugging issues very, very difficult.

I actually have more patches available to get HDMI sound and a few other things to work, but I don't have a complete VCHIQ implementation yet so it's only good enough for HDMI sound.

bamarni:
@Electron752 : thanks for your hints. I've tried to build it but unfortunately my rpi 3 wouldn't boot anymore, I did the following : 

```
apt-get update
apt-get install -y bc build-essential gcc-aarch64-linux-gnu git unzip zip

git clone --depth=1 -b rpi-4.8.y https://github.com/raspberrypi/linux.git

cd linux
KERNEL=kernel8
make ARCH=arm64 CROSS_COMPILE=aarch64-linux-gnu- -j6 bcmrpi3_defconfig
make ARCH=arm64 CROSS_COMPILE=aarch64-linux-gnu- -j6 

wget https://downloads.raspberrypi.org/raspbian_lite_latest
unzip raspbian_lite_latest 
# fdisk -l 2016-05-27-raspbian-jessie-lite.img
mount -o loop,offset=70254592 2016-05-27-raspbian-jessie-lite.img /mnt
mount -o loop,offset=4194304,sizelimit=66060288 2016-05-27-raspbian-jessie-lite.img /mnt/boot

cp arch/arm64/boot/Image /mnt/boot/kernel8.img
cp arch/arm64/boot/dts/broadcom/bcm2710-rpi-3-b.dtb /mnt/boot/
make ARCH=arm64 CROSS_COMPILE=aarch64-linux-gnu- INSTALL_MOD_PATH=/mnt -j8 modules_install

umount /mnt/boot
umount /mnt
```

I've also tried to copy `lib/modules/firmware` too, copy the kernel as `kernel7.img` / `kernel.img` while removing the others. Do you know what I'm missing?

Edit : It's on ubuntu 16.04

ghost:
Can you elaborate on what you mean by it doesn't "boot".   At a
minimum, you should be able be to see the kernel startup messages on
the monitor.   Even without the modules, you should be able to get at
least that far.

Some possible ideas:
1. I'm a bit skeptical about how you are mounting the image partitions
   by specifying a offset like that.  Can you mount the SD card directly
   and confirm that the files were copied correctly?
2. I haven't tested on raspbian in awhile.  I use a native 64bit OS for
   esting(Debian sid arch64).  It's possible I forgot to enable all the
   32bit compatibility layers to get 32bit userland to work.

In the case of 2, you should still see boot messages on the monitor
before everything dies trying to load init or systemd.   

Finally, booting 32bit raspbian with a 64 bit kernel doesn't have a
whole bunch of benefit.  In fact, you will just break a  bunch of
stuff.  When I first started this exercise, I did some performance
tests and 64bit in general is slightly slower then 32 bit in many
cases. 
So unless you really need 64 bit swap files, you are going to see
absolutely no end user benefit to this.

bamarni:
I do not see anything at all on the screen, that's why I thought I missed something in the build process. I couldn't mount the sd card directly because I'm building the kernel in a vm from the cloud as I don't have a linux host, but I can confirm that files were copied properly, after unmounting and mounting again I could still see them.

I'm not looking for performance improvements, I've got some 64 bits programs I'd like to run on my rpi 3, I started looking at raspbian support as it's the official os and stumbled on this PR :)

ghost:
That’s puzzling then.  You should at least see some boot messages.

Can you confirm that your cmdline.txt file is correct?   You should have console=tty1 or something similar.  I think the default is to use the miniUART as the default terminal.   Also, make sure you don’t have an initrd or anything similar in your config.txt.   You don’t need initrd for booting with this config.

Another thing to try is to create a basic FAT32 boot partition and nothing else on the drive(no root or anything).  All you need to copy over is the foundation firmware files(*.bin *.dat *.elf) plus kernel8.img.   You should not need a config.txt but you will need to have a cmdline.txt as the default is to use the miniUART or you will need to modify the default kernel config.

Here is what my cmdline.txt contains:
console=tty1 root=/dev/mmcblk0p2 rootfstype=ext4 rootwait

Like I said, try this without a root partition(or modules).  You should at least see kernel boot messages.

pelwell:
Use `console=serial0` in cmdline.txt - the firmware will rewrite it as appropriate to use ttyAMA0 or ttyS0.

ghost:
Sorry, I forgot to add that in addition to the firmware files and
kernel8.img, you will need the device tree(bcm2710-rpi-3-b.dtb).  Newer
linux kernels will not boot without a device tree and if you don't have
it, you will not get far at all into the boot sequence..

bamarni:
@Electron752 : my cmdline.txt is similar than yours, I also had copied the device tree (cf. my previous snippet for the exact commands I ran).

I'm really not sure how to debug further as I'm not familiar with kernel building at all... Would that help if you could access the produced image? Just in case I've uploaded it online here : http://138.68.92.147/

ghost:
So I downloaded your image and copied your boot partition onto the sd
card of my pi3.  It boots just fine...

All I did was delete kernel.img and kernel7.img.

Also, you need to delete this section from your cmdline.txt
quiet init=/usr/lib/raspi-config/init_resize.sh

The quiet is why you are not seeing diagnostic messages and the init=
is why it isn't booting.

Here is what I changed the cmdline.txt to:
dwc_otg.lpm_enable=0 console=serial0,115200 console=tty1
root=/dev/mmcblk0p2 rootfstype=ext4 elevator=deadline fsck.repair=yes
rootwait

Hope this helps.

BTW, I made an exception this time but in general I don't run software
from random sources.   It's a good way to end up with malware/viruses.

bamarni:
After deleting kernel.img and kernel7.img, updating cmdline.txt I still get the same result, without any error messages :/

Just to compare I've tried with the default raspbian image from https://www.raspberrypi.org. It seems like `quiet init=/usr/lib/raspi-config/init_resize.sh` is just a trick so that people don't have to run `sudo raspi-config --expand-rootfs` anymore, the `init_resize.sh` script automatically removed `quiet init=/usr/lib/raspi-config/init_resize.sh` from my `cmdline.txt` before it rebooted. While it should have indeed help to diagnose boot errors, I don't think it's the problem.

I've also tried to put `kernel=kernel8.img` at the end of my `config.txt` file, as explained [here](https://www.raspberrypi.org/documentation/linux/kernel/building.md), but no more luck...

Thanks for your help @Electron752 , let me know if I can provide additional information.

ghost:
Obviously, something else is going on because it bootups fine for me.  
At the very minimum, you should be seeing some debug spew on the
monitor as it boots.

I really don't know how to help any more at this point.   I will say
that raspbian is not the best choice at this point if you want a 64 bit
kernel.   It's 32 bit and it uses some specifics of the hardware that
isn't implemented yet in 64 bit.

I use debian, and others have had alot of success with fedora and arch
linux.

bamarni:
Just figured out what was going on... I'm using this kind of hdmi switch on my tv (https://www.bhphotovideo.com/images/images2500x2500/Comprehensive_CSW_HD201C_CSW_HD201C_2x1_HDMI_Switcher_668874.jpg) and even though it works with the standard raspbian, the one with the 64 bits kernel I've compiled can't handle this apparently. After plugging the HDMI cable directly to the TV I could see all the log messages.

So I've deleted kernel.img / kernel7.img and updated cmdline.txt, I think the relevant error messages are (copying manually so there might be a typo or 2 :) ) : 

```
[...]
init[1]: unedfined instruction: pc=00000000f713ab50
Code: e1a02005 e0840000 e1a01004 eb0031cf (ee070fba)
Kernel panic - not syncing: Attempted to kill init! exitcode=0x00000004

CPU: 1 PID: 1 Comm: init Not tainted 4.8.0-rc7-v8+ #1
[...]
```

Do you think it's related to the kernel I've compiled or raspbian? I will also give it a shot with debian or ubuntu arm64 as you suggested.

ghost:
The TV issue might be a bug in 4.8 and not 64 bit specific.   

Most likely you are seeing crashes in the init process because I missed
enabling all the compatibility layers.   You can enable them in the
config if you wish and recompile the kernel  I probably should do
another pull request soon and update some of the issues I missed.  Like
I said debian 64bit booted fine on my pi 3.

Debian might be better to start with.  I don't know but I think I saw
somewhere on the Internet that some of the newer versions of ubuntu
were going to use an alternative graphics system other then X windows.
  I don't know if that requires video acceleration or if it works with
the old fbdev drivers.

The accelerated video driver for the pi 3 does work, but since raspbian
doesn't work so well at this time you need to munge the device tree
yourself to get the configuration correct.   Also the driver for
accelerated video loads rather late in the boot sequence so it isn't
good for starting out.

Glad to hear that you have some error messages now!

bamarni:
A last question if you don't mind, should I use the same method to install debian on my rpi3? Or is there a ready to use installer / image? I only need to run 64 bits programs and don't plan to use wifi / bluetooth etc.

Thanks

ghost:
If you have a working arm64 install of debian on another machine, then
yes you should just copy the files over and change a few config files.
 And of course replace the boot partitions and kernel modules.

I use debootstrap which is more the pain.   If you have a x86/amd64
install of debian, you can use the --foreign switch to create an
install for a different architecture(arm64 in this case). It runs part
of the install, then you boot the pi 3 and run debootstrap --second-
stage to finish the setup.   Another option is to use qemu-static to
emulate the rpi 3 to create the initial install, but I've found that
doing that is slow compared to an actual pi 3.

bamarni:
@Electron752 I've tried again with the last option you've mentioned. I'm getting closer as now it can boot and I can login :) However the filesystem is read-only, I think it's because an error occured during kernel boot. Here is a capture : 

img_20161009_191621

To generate the root filesystem I've followed instructions at https://wiki.debian.org/Arm64Qemu, do you have an idea about what's wrong here?

yuchenericwu2:
@bamarni, I'm running into the same issues that you're seeing. I did some digging and it looks like @Electron752 also hit the sdhost failure and opened a bug here: https://github.com/raspberrypi/linux/issues/1619.

That being said, the bug has since been closed and I'm not sure what the resolution is. From what I can gather the NULL being passed into dma_zalloc_coherent needs to be replaced with a driver context, but I don't know what that is...

Any help would be much appreciated.

ghost:
The bug doesn't stop any functionality.   It's just causes some extra debug spew.   

If the sd driver wasn't working, the filesystem would not be accessible at all instead of just read only.

I just debian linux running arm64 on the RPI 3 on 4 different RPI's doing various things that are on all the time.   I'm able to use it just fine in 64 bit mode.

If you have a repo case, I can try to reproduce it myself.

ghost:
The bug doesn't stop any functionality.   It's just causes some extra
debug spew.   

If the sd driver wasn't working, the filesystem would not be accessible
at all instead of just read only.

I just debian linux running arm64 on the RPI 3 on 4 different RPI's
doing various things that are on all the time.   I'm able to use it
just fine in 64 bit mode.

If you have a repo case, I can try to reproduce it myself.

ghost:
BTW, It looks like my outbound e-mail was down for a bit.  Here is my reply from October 5th, 2016.

The errors on the screen should not be enough to keep the filesystem read only.  Typically, a read only file system means that your fstab file is wrong.

Can you double check that your /etc/fstab is correct?

Here is what mine looks like from one of my arm64 RPI 3s.
/dev/mmcblk0p2   /      ext4  errors=remount-ro 0 1
/dev/mmcblk0p1   /boot  vfat  errors=remount-ro 2 2
/dev/mmcblk0p3   none   swap  sw                0 0

The last line is not needed if you don't have a swap partition setup.    

Also check your cmdline.txt file to make sure you have the correct root partition, but it looks like you are getting that far.

Are for the errors, I can give you brief explanation as to what is happening.
1. sdhost: fail to allocate log buffer

This is a benign errors in the logging system of the sd card driver.  It doesn't hurt anything except debug logging.
1. Missing clocks

This happens because platform drivers do not always have depend drivers loaded first.   In this case the clock driver.  When this errors happens the kernel core will retry loading the driver later in the boot sequence.

You should be able to see the sdhost/mmc drivers load later in the log.
1. brcmfmac is for the wifi and you are seeing this error because you haven't copied the firmware for the wifi from the raspbian disk image.  It just means you won't be able to use the wifi.

yuchenericwu2:
@Electron752 

Thanks for the quick response, you're right that the sdhost spew was just a red herring. My user space is failing to boot, not because of sdhost but because init contains undefined instructions.

For the record, I'm trying to load a 32-bit Raspbian OS into the 64-bit kernel that you have provided. Is this something that is supported?

Here is a screenshot of the error on boot:

img_0502

ghost:
The 64bit kernel isn't support at all by the PI Foundation.  It's just a tinker project.

32-bit raspbian doesn't work right now.   I think I missed a few of the compability layer switches when I submitted the pull request.   So 32bit OSs don't work right now.

You can use a 64 bit OS, but you need to either jump through a few hoops to create the bootstrap media or you need to install it on a different arm64 machine and then replace the kernel.

I can probably try 32 bit raspbian again sometimes soon and at least make sure it boots and then submit a new pull request.

yuchenericwu2:
Yep, I understand that this is all _bleeding edge_. Do you think it's just a matter of setting the right parameters in the .config file generated by make bcmrpi3_def? If so, I could try to tinker as well...

I'd prefer to keep using the Raspbian OS as opposed to Ubuntu, since I'm using a lot of peripherals and GPIO pins from the board.

ghost:
Yes, for the most part I think it's just enabling a few config settings.  I had it working at one point in the past.

Some settings that I can think of are:
CONFIG_ARMV8_DEPRECATED=y
CONFIG_COMPAT=y
CONFIG_SYSVIPC_COMPAT=y

I suspect if you enable those you will get much farther into the boot sequence.   Just to let you know through, that vchiq is not working in the public version at this time.

yuchenericwu2:
Awesome! CONFIG_ARMV8_DEPRECATED did the trick, thanks so much!

Is VCHIQ only required for certain video processing?

ghost:
It's also needed for HDMI audio.

ghost:
I created pull a pull request to make the necessary changes to the config to get raspbian to boot.  I tested it on raspbian jessie 2016-09-23.

While most things seem to be working, I wasn't able to get mathmatica to load because it does some kind of funky OS detection which seems to be broken by a 64bit kernel.

yuchenericwu2:
By the way, the 64-bit kernel has been broken at some point between 4.8.2 and 4.8.6. More specifically, rpi_backlight.ko no longer builds in 64-bit mode.

I disabled the module manually in bcmrpi3_defconfig by setting CONFIG_BACKLIGHT_RPI=n, but I'm not sure what the consequences that are.

ghost:
OK, thanks.

I'll take a look at it later today.

ghost:
@yuchenericwu2: It looks like some recent changes were made to the source tree and arm64 is fairly broken.  I don't think disabling CONFIG_BACKLIGHT_RPI is good enough at this point.   It might take me awhile to make a fix.

Probably your best option is to roll back your local git tree to 4.8.2 and rebuild if arm64 is important to you.   This is bleeding edge and all and it isn't actually supported.

yuchenericwu2:
My git-foo is a bit lacking but is it possible to checkout a commit that includes your changes and only the 4.8.2 commits?

Since the merge, the history seems to have gotten messed up.

ghost:
Not that I know of.  I'm still a bit new to git.

popcornmix:
`git rebase --onto v4.8.2 stable/linux-4.8.y rpi-4.8.y`

will create a 4.8.2 kernel with the pi specific patches.

ghost:
OK, I was able to fix this easier then I thought.
I've created pull request #1710 to fix the build issue.

I'll add here if/when it gets accepted.

adalric:
Can I ask how did yo guys managed to get Bluetooth and Wifi working?

PR #1620, opened 2016-08-24
ARM64 - Fix page table corruption in bcm2835-sdhost driver.

ghost:
Hi,

I was able to fix the MMC driver in ARM64.   I reenabled SD/MMC 
in the bcmrp3_defconfig kernel config.  I tested it and everything seems to 
be working now.

Here is the checkin description:
Fix page table corruption in bcm2835-sdhost driver.

The sg->page_link field is no longer directly convertable
to a page pointer since the base kernel stores extra info
in some of the unused bits.

On arm64, convert code to use the sg_page macro which is more
portable.

Reenable SD/MMC drivers in arm64 config.


drivers/mmc/host/bcm2835-sdhost.c
```diff
@@ -815,7 +815,11 @@ static void bcm2835_sdhost_prepare_dma(struct bcm2835_host *host,
 			if (sg_is_last(sg)) {
 				BUG_ON(sg->length < len);
 				sg->length -= len;
+#ifdef CONFIG_ARM64
+				host->drain_page = sg_page(sg);
```
popcornmix, line 818: If the ifdef needed? Can we use sg_page in both the arm32 and arm64 cases?

ghost: Yes, I just didn't want to be responsible for testing all the different combinations.  The macro is very simple through if you want to look at it.

popcornmix: Looks safe to me. I think we always want to avoid idefs where possible.

ghost: I mean I can remove the #ifdef if you want.  It should work.

ghost:
Ok, I removed the #ifdef and pushed a new forced update.

popcornmix:
@pelwell happy with the change?

pelwell:
Thanks, @Electron752.

ghost:
Your welcome.

This gives a good feature set for arm64 at this point.

PR #1628, opened 2016-08-31
rpi-4.7.y enable clock manager and modify all subsystems to use it

clivem:
Placeholder.... This is what I think we have to merge to move all of the subsystems over to using clock manager in rpi-4.7.y.

@msperl I think this is turning more into a "loose ends" to have clock manager enabled by default, and move all of the subsystems over to using it, rather than an "backport upstream clockmgr/I2S" thing. I'm still testing, but I'm not seeing any issues at the moment.

PR #1592, opened 2016-08-10
Rpi 4.4.y msperl i2s/clockmgr backport

clivem:
This is another of my PR's that isn't actually a PR.

@msperl Martin, I've just quickly hacked your last unmerged backport, (obviously sans the dma patchset, that is already in rpi-4.4.y)....... Is there still any interest in trying to get this back into 4.4.y? 

@pelwell / @popcornmix How much longer do you expect rpi-4.4.y to remain the "stable" release? Will rpi-4.6.y be the next "stable"? How far away is that?

PR #1629, opened 2016-09-01
rpi-4.7.y enable clock manager and modify all subsystems to use it - with reverts

msperl:
Make use of the clock framework for all drivers (no more opt in like with the 4.4 kernels, where lots of things stop working when the clock framework is enabled - especially i2s)

This is the equivalent patch to #1628, but here with reverted reverts instead of a single "bulk" revert.


clivem:
@popcornmix Unless @msperl has found any problems, I've tested to the point where I think we want to pull this into rpi-4.7.y branch.

The issue I saw with Zero's was entirely my own fault and not an issue. (I forgot to patch the BFQ scheduler, which I use as IO scheduler for SDCARD's into my bcm2708 "flavour" kernel builds. Doh!)

I'm not going to list boards, as I don't want it to look like I'm recommending one board over another, but I have tested using B, B+, 2B, 3B and Zero (basically everything other than CM), with 12 commonly available third party I2S HATs/boards, ie. the usual suspects. Obviously, my primary interest is I2S output, but I've checked serial console output, aux serial (BT on Pi3), wifi on Pi3, I2C, everything I use is working OK with clock manager using this patch set.

**NB. Other boards are available for the Zero! These were the two I used for testing.**
img_20160902_141813602_640x452

pelwell:
Why are we reverting the commit that adds mmap support to the I2S driver? That is a useful feature for some people, and is a stepping stone to enabling it by default.

clivem:
> Why are we reverting the commit that adds mmap support to the I2S driver? That is a useful feature for some people, and is a stepping stone to enabling it by default.

MMAP is enabled by default!

pelwell:
Well that's one way of finding out why it was disabled in the first place.

clivem:
> Well that's one way of finding out why it was disabled in the first place.

Well, we know why it was disabled in the first place. Someone reported problems with CAPTURE with specific hardware. I don't have that specific hardware to test myself but I talk to a man that does. Since the upstream dma backport made it into downstream code, it is no longer an issue having MMAP enabled. It's more of a PITA it being disabled by default now. SonicPi is dead in the water without MMAP, ISTR.

@msperl Do you want to go down the belt and braces path, having MMAP disabled by default in the I2S driver, and an overlay to enable it, how things used to be? 

MMAP can be disabled when you actually open the ALSA audio device. I have written software that is capable of doing that, without needing to depend on the driver defaults......

popcornmix:
There are a few complaints in log:

```
[    1.578861] sdhost-bcm2835 3f202000.sdhost: could not get clk
[    1.583089] mmc-bcm2835 3f300000.mmc: could not get clk
```

which @pelwell says is due to initialisation ordering, and they succeed after a deferred probe. We may want to hide those messages as they look like something is wrong.

But otherwise we're okay with this. Lets see if there are any issues in the next nightly LibreELEC build.

popcornmix:
Thanks for working on/testing this.

clivem:
> which @pelwell says is due to initialisation ordering, and they succeed after a deferred probe.

It tries to initialise some subsystems before clockmgr is initialised, so deferred probe. It's fine. I don't think we should hide those messages. I'd rather see them. ;)

clivem:
> Lets see if there are any issues in the next nightly LibreELEC build.

Sorry, I forgot. I know you already told me the answer to this.... Is the LibreELEC build using 4.6 or 4.7 tree?

popcornmix:
The wording and red colouration in dmesg will probably cause unwanted bug reports. Maybe there is a way to make it seem more normal.

LE uses 4.7 kernel.

clivem:
```
if (PTR_ERR(clk) != -EPROBE_DEFER) {
  dev_err(dev, "could not get clk\n");
}
```

msperl:
Note that I have posted a patch to the rpi and clock list to initialize clockmgr early - You could try that...

pelwell:
Yes, that's what my patch looks like (except without the deprecated braces around a single line block).

msperl:
Please get someone to test smi and pwm...

clivem:
> Note that I have posted a patch to the rpi and clock list to initialize clockmgr early - You could try that...

@msperl LOL. Martin, make it easy for a fool. Can you give me a link to the patch rather than me scouring mailing list archives?

msperl:
https://patchwork.kernel.org/patch/9128797/
I had to search as well (first in my emails, then in Google)

clivem:
Thanks. Hmmm. `State    Rejected, archived`

msperl:
Well, it was related to the second patch to claim the sdram clock early, which Eric wanted to get implemented differently (probably because the sdram clocks are only accessible from vc4)
Still that first part does early initialization

clivem:
I'd add a Tested-By and be asking the RPi guys to pull it, but seen as it is't going to receive upstream acceptance, it'll just remain part of my local, ever growing, patch set.

It didn't apply cleanly so I removed "err" variable and used the already defined "ret".

https://github.com/DigitalDreamtimeLtd/linux/commit/08f54e2b0136b15850941634a41374d1655d8ccb

popcornmix:
Yes, I also cherry-picked the patch to 4.7 tree. It avoids the deferred probe warnings.

msperl:
I will try to upstream the early init patch alone - glad to hear that it makes a difference on its own...

clivem:
> I will try to upstream the early init patch alone - glad to hear that it makes a difference on its own...

Cool! Makes perfect sense to me, that something on which everything else now depends, get's initialised early and avoids a bunch of deferred probes..... sdhost, mmc, serial....

msperl:
Well, if I remember correctly, there was one concern about possible drivers that during init enable a clock and then fail/defer on something else, which would disable a clock which could disable a necessary pll or pll-divider...

But then I guess that driver needs to get modified.

msperl:
So I guess this is finally done - now we require the support by Eric to get vc4 updates into the 4.7 kernel...

clivem:
> So I guess this is finally done

@msperl Yes, we're done. Will be interesting to see if there is any "fall-out" after the next LibreELEC build. 

The real question is, do we want to pull the "subsystems to use clockmgr rather than clocks directly, backport" into rpi-4.4.y, as this is going to be the "default" RPi kernel for the foreseeable future.... #1592 (If Pi guys are going to wait for next LTS. And if that is to be 4.9, that's not going to be before the end of the year.)

Anyway, this is finally "put to bed". Thank you Martin! You did all of the heavy lifting. I mostly did the "getting frustrated"..... ISTR, it was January when I started running the new dma code. (That took its time to make it into downstream.) And I've lost track of how many months ago, I was testing a rpi-4 tree "backport" of dma/I2S/clockmgr...... 

I'd like to return to trying to get that "simple framework" callback stuff into upstream simple-card, if you're up for it? I made a few more mods. Next week I'll make a new rpi-4.7.y branch.....

msperl:
@clivem as for the backport: the patches should be easy to apply, but I guess foundation will not want to move that way as there is a bigger risk in breaking things unexpectedly. The risk is still there with 4.7, but a move to a newer kernel makes it more obvious that things may break.

I wonder if we now can get the vc4 driver into 4.7, because going from there it should then be easy to move to the next stable kernel, as we are mostly upstream already....

But I guess that popcornmix will need to make that decision.

clivem:
> but I guess foundation will not want to move that way as there is a bigger risk in breaking things unexpectedly. The risk is still there with 4.7, but a move to a newer kernel makes it more obvious that things may break.

Agreed. But my concern is that we think we have put this to bed..... (I thought this was finished 3 months ago when I had everything I needed to work with clockmgr, working with a patched rpi-4.4 build.) If they really do wait for a 4.9 LTS release before moving from 4.4, this is going to come up again around Christmas, if there are any issues...... Don't know about you but I'd rather see them resolved now, rather than deferred for 3 months.

> I wonder if we now can get the vc4 driver into 4.7, because going from there it should then be easy to move to the next stable kernel, as we are mostly upstream already....

I wonder if we need to ask @popcornmix et @pelwell to make a "next" build from the rpi-4.7.y tree so it gets some wider testing. (I'm not convinced that OpenELEC is enough. The stuff those guys are testing. ie. the audio, I'm fairly sure we are on safe ground already. I'm not sure they are interested in VC4.) 

I think we should ping Eric and ask him to get VC4 "up to scratch" in this rpi-4.7.y tree and get a "next" build out there for the people who might want to live a little dangerously and not sit out the rest of the year with a rpi-4.4 kernel build. There are several people who did want I2S and VC4 at the same time. Make it easy for them with a "next" build they can rpi-update to, and they will test! ;)

clivem:
Having had cause to look closely at the logs on my CM test subject today, I noticed the enc28j60 spi ethernet not initialising on 4.7.y ....

```
Linux version 4.7.2-513.20160903git6e32787.fc20.armv6hl.bcm2708 (clivem@e5530) (gcc version 4.9.3 (crosstool-NG crosstool-ng-1.22.0-88-g8460611) )

kernel: enc28j60 spi0.0: enc28j60 Ethernet driver 1.02 loaded
kernel: enc28j60 spi0.0: enc28j60 chip not found
kernel: enc28j60: probe of spi0.0 failed with error -5
```

I wondered whether this was related to clockmgr.... 

Still works fine on my rpi-4.4.y plus clockmgr backport.... (So most likely something else changed between rpi-4.4.y and rpi-4.7.y.)

```
Linux version 4.4.19-504.20160906git571df49.fc20.armv6hl.bcm2708 (clivem@e5530) (gcc version 4.9.3 (crosstool-NG crosstool-ng-1.22.0-88-g8460611) )

kernel: enc28j60 spi0.0: enc28j60 Ethernet driver 1.02 loaded
kernel: net eth0: enc28j60 driver registered
NetworkManager[187]: <info> (eth0): driver 'enc28j60' does not support carrier detection.
NetworkManager[187]: <info> (eth0): new Ethernet device (driver: 'enc28j60' ifindex: 3)
```

I'll have to take a closer look later in the week.

msperl:
It is "correctly" detected, as otherwise eth0 does not probe - "enc28j60_chipset_init" fails...

You can load enc28j60 with the kernel argument debug=... (up to 0xffff)

That should give you messages like "chip RevID:" and that indicates that spi is working and the chip is detected...

popcornmix:
Ping @anholt - the rpi-4.7.y tree now uses the upstream clock framework.
It might be nice (not urgent) to get the vc4-kms-v3d merged in as hopefully we are closer to upstream and fewer cherry-picks are required. No immediate plan to bump to 4.7, but we can push out a "next" branch kernel for wider testing.

msperl:
For completeness: the clock patches have gone into clk-next, so they should show up in 4.9...

Issue #1605, opened 2016-08-18
Kernel Options for systemd 231


crazy4chrissi:
I compiled a kernel as described here:
https://www.raspberrypi.org/documentation/linux/kernel/building.md
This kernel would not boot properly on Arch Linux with systemd 231. The problem and its solution is described here:
https://www.danand.de/index.php/2016-08/solution-to-systemd-231-seccomp-error-on-allwinner-soc/
With these kernel options, systemd 231 worked well for me:
CONFIG_HAVE_ARCH_SECCOMP_FILTER=y
CONFIG_SECCOMP_FILTER=y
CONFIG_SECCOMP=y
CONFIG_OABI_COMPAT=n

popcornmix:
Arch produce their own kernel. Our default settings are chosen for raspbian.
If there are no downsides we would consider options that are useful for Arch, but `CONFIG_OABI_COMPAT` was required for https://github.com/raspberrypi/linux/issues/766

crazy4chrissi:
Okay, then probably let's postpone this until Raspbian embraces systemd 231, which might take a couple of years... ;-)  currently, systemd 231 is in Debian sid only https://packages.debian.org/sid/systemd

sandersr:
Same problem on Fedora 25 and official RPI kernel. Any reason why are these settings disabled:
CONFIG_HAVE_ARCH_SECCOMP_FILTER=y
CONFIG_SECCOMP_FILTER=y
CONFIG_SECCOMP=y
?

pelwell:
CONFIG_SECCOMP=y and CONFIG_OABI_COMPAT=y are both in the standard configurations, but not CONFIG_HAVE_ARCH_SECCOMP_FILTER=y or CONFIG_SECCOMP_FILTER=y.

As I said in [issue 1172](https://github.com/raspberrypi/linux/issues/1172):

> With all kernel options that aren't simply loadable modules we are concerned about increased kernel size and reduced performance. If you were to present "before and after" comparisons of free memory and performance (using some standard benchmarks) it may strengthen your case.

Ferroin:
The issue with CONFIG_SECCOMP_FILTER is actually that we have CONFIG_OABI_COMPAT=y, which is itself needed by the old version of GDB in Raspbian.  There was talk that it might get patched locally so it doesn't need that, but I don't think that actually happened.

clivem:
>  There was talk that it might get patched locally so it doesn't need that, but I don't think that actually happened.

[...]

Issue #1065, opened 2015-07-15
Consider adding RPi-specific Kconfig patch which pre-selects hardware-essential options


srcshelton:
Moving from an original RPi to an RPi2 has caused several issues when building custom kernels (especially in relation to the MMC/SDHC controller), partly because it's not immediately obvious what kernel options are mandatory for the RPi to boot with full functionality (e.g. SDHC, access to VideoCore, appropriate DT options set, etc.).

It would be a huge benefit to integrate a patch such as:

https://github.com/gtgt/gt-overlay/blob/master/patches/sys-kernel/pf-sources/3.11.1-4567_distro-Gentoo-Kconfig.patch

... which Gentoo uses to provide an optional but highly intuitive method of ensuring that the kernel has minimal functionality to try to guarantee that it will at least boot.  The use-case would be when other areas of the kernel config have been extensively customised (thus making running 'make defconfig' and importing the changes a long job) but where the hardware-specific choices need to be re-applied.

This would also provide a mechanism to black-list any known-bad options which, for a given release, are known to cause issues.

Thanks in advance,

Stuart

Ferroin:
+1 on this, making it harder for people to shoot themselves in the foot when building their own kernels is a good thing.

Belphemur:
:+1: It would facilitate a lot instead of juggling with defconfig.

Ruffio:
@srcshelton has your issue been resolved? If so, please close this issue. Thanks.

srcshelton:
Still not resolved, no.  I was going to submit a patch, but my RPi kernel builds broke just after I filed this issue (the system didn't want to see its SDHC hardware on boot) and I've not had chance to fix things since.

I've recently picked it up again, though, so I'm going to try a stock 4.4.y kernel shortly...

srcshelton:
PR #1631 - We're actually more limited in what can be achieved due to Kconfig only being able to `select` options to be built-in (with no supported method I could see for selecting tri-state options as modules or for deselecting other options without changing their dependencies) , but hopefully still of some use...

clivem:
> huge benefit to integrate a patch such as...

How so? You use the bcmrpi_defconfig or bcm2709_defconfig and your kernel will have all the neccesary options configured to boot. Am I missing something here? It feels like you are trying to fix something that never needed fixing.....

mi-hol:
@srcshelton is this still active or should it be closed?

JamesH65:
[...]

PR #1631, opened 2016-09-07
Enhancement suggested in #1065 (and potentially providing a solution for #1605)

srcshelton:
Add an additional Kconfig which can optionally automatically select relevant kernel options to ensure maximum Raspberry Pi compatibility when building custom kernels

Drawbacks:
- Kconfig provides no method to directly deselect options
- Kconfig only allows options to be set to 'Y', rather than enabling an option to be built as a module

(Deselection could be performed by editing other stock Kconfig files to depend on one of the new options provided here being in a certain state - but this is a more intrusive change)

### DIFF
--- a/Kconfig
+++ b/Kconfig
@@ -8,4 +8,6 @@ config SRCARCH
 	string
 	option env="SRCARCH"
 
+source "contrib/Kconfig"
+
 source "arch/$SRCARCH/Kconfig"
--- a/contrib/Kconfig
+++ b/contrib/Kconfig
@@ -1,0 +1,573 @@
+menu "Pre-select default options"
+
+	menu "Raspberry Pi"
+
+		config RPI_KCONFIG
+			bool "Raspberry Pi support"
+
+			default y
+
+			# Ensure that output fits onto an 80-column display with ncurses window-rendering:
+			#
+			########xxx      1         2         3         4         5         6         7    xxxxx8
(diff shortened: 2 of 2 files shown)


Electron752:
BTW, Is the reason you haven't tested on RPI 3 because you don't have one?  I happen to have one if you need someone to run a quick test.

Also, I think all you need to do to get arm64 to work is delete 
config ARCH_BCM2709
        bool "Broadcom BCM2709 family"
        select MACH_BCM2709
        select HAVE_SMP
        select ARM_AMBA
        select COMMON_CLK
        select ARCH_HAS_CPUFREQ
        select GENERIC_CLOCKEVENTS
        select MULTI_IRQ_HANDLER
        select SPARSE_IRQ
        select MFD_SYSCON
        select VC4
        select USE_OF
        select ARCH_REQUIRE_GPIOLIB
        select PINCTRL
        select PINCTRL_BCM2835
        help
          This enables support for Broadcom BCM2709 boards.

Change to ARCH_BCM2835 in the defconfig.

JamesH65:
@popcornmix @pelwell Any suggestions on what to do with this?

pelwell:
It's a wish at the moment - essentially, "please do some work so we don't have to". This would make sense if enough people would make use of it to reduce some support burden sufficiently to offset the cost of setting it up, but I have no confidence that that would be the case.

The bar would be lowered significantly if the request appeared as a pull request, but until then I think we can close this.

PR #1634, opened 2016-09-12
overlays: Add dpi18 overlay

pelwell:
Add support for 18-bit DPI displays. Although the dpi24 could be used,
this overlay leaves GPIOs 22-27 free for other uses.


pelwell:
@6by9 Is this what you had in mind?

6by9:
Yup, exactly that. I just hadn't done it, or had done it somewhere but failed to get it to work with my Kippah and 5" display. (Using the blob works fine, but I never have got the overlay to work. I've now switched to the DSI display so it wasn't bothering me).

PR #1637, opened 2016-09-14
Rpi 4.4.y firmware kms

anholt:
Here it is: vc4 running on top of dispmanx.  Just switch your dtoverlay line from vc4-kms-v3d to vc4-fkms-v3d to test.


pelwell:
I've made some cosmetic changes, after which I'm happy with the patch set. You should squash them into your commits and force push.

PR #1645, opened 2016-09-19
Allo Piano DAC boards: Initial 2 channel (stereo) support

clivem:
Add initial 2 channel (stereo) support for Allo Piano DAC (2.0/2.1) boards,
using allo-piano-dac-pcm512x-audio overlay and allo-piano-dac ALSA ASoC
machine driver.

@pelwell NB: The reason that I have advocated a dedicated machine driver for the Piano DAC boards, rather than re-using one of the other manufacturer PCM5122 machine drivers, is that there will be modifications to follow, to support the DUAL DAC nature of the Piano 2.1 board. But those modifications will depend on upstream changes to pcm512x codec driver, which will need to be accepted before moving forward.

The Allo Piano products have already gone on sale .....
[Piano DAC 2.0](http://www.allo.com/sparky/piano-dac.html)
[Piano DAC 2.1](http://www.allo.com/sparky/piano-dac-2-1.html)


clivem:
This patch should also apply cleanly to rpi-4.7.y branch. Please can you apply it there as well?

pelwell:
You must have a script to generate these by now - a bit of renaming here, a bit of reformatting there...

clivem:
> You must have a script to generate these by now - a bit of renaming here, a bit of reformatting there...

LOL. Not quite. It goes against my nature to have this proliferation of PCM512x machine drivers, when one will do! But I've given up trying to herd cats and have everyone play nice, and sing from the same hymn sheet.....

PR #1664, opened 2016-10-03
BCM270X_DT: Use bcm283x.dtsi, bcm2835.dtsi and bcm2836.dtsi

notro:
The mainline Device Tree files are quite close to downstream now.
This is a proposal to take advantage of that and take yet another "mainline" step.

I have made a diff file: bcm270x.dtsi instead of patching bcm283x.dtsi directly. I don't know what's the best maintainable solution. One benefit is that it is overly clear how we differ from mainline. The changes could also be included in bcm2708-rpi.dtsi as well, but IMO that's not so good since we loose the distinction made in mainline.

I haven't followed downstream so close the last months, so I don't know if this has any side effects on the overlays.
There's some more details in the commit message.

This is the diff introduced by this patch that applies to all dts. I have removed phandle changes that are due to node order changes and lines without change.

``` diff

@@ -34,7 +34,6 @@
    aliases {
-       clocks = "/clocks";
-       leds = "/soc/leds";
+       leds = "/leds";
-       spi2 = "/soc/spi@7e2150C0";
+       spi2 = "/soc/spi@7e2150c0";
@@ -65,19 +64,31 @@
-   clocks: clocks {
+   clocks {
-       clk_osc: clock@6 {
+       clk_osc: clock@3 {
-           reg = <0x6>;
+           reg = <0x3>;
+       };
+   };
+
+   leds: leds {
+       compatible = "gpio-leds";
+
+       act_led: act {
+           gpios = <0x8 0x10 0x1>;
+           label = "led0";
+           linux,default-trigger = "mmc0";
+           linux,phandle = <0x19>;
+           phandle = <0x19>;
        };
    };
@@ -110,12 +121,12 @@
-       cprman: cprman@7e101000 {
+       cprman: clocks: cprman@7e101000 {
@@ -179,53 +190,52 @@
        gpiomem {
-           status = "okay";
        };

-       gpu: gpu {
+       gpu: vc4: gpu {
@@ -234,15 +244,17 @@ hdmi
+           interrupts = <0x2 0x8 0x2 0x9>;
        };

        hvs: hvs@7e400000 {
+           interrupts = <0x2 0x1>;
        };
@@ -313,18 +325,6 @@
-       leds: leds {
-           compatible = "gpio-leds";
-
-           act_led: act {
-               gpios = <0x7 0x10 0x1>;
-               label = "led0";
-               linux,default-trigger = "mmc0";
-               linux,phandle = <0x19>;
-               phandle = <0x19>;
-           };
-       };
-
@@ -347,18 +347,21 @@
        pixelvalve0: pixelvalve@7e206000 {
+           interrupts = <0x2 0xd>;
        };

        pixelvalve1: pixelvalve@7e207000 {
+           interrupts = <0x2 0xe>;
        };

        pixelvalve2: pixelvalve@7e807000 {
+           interrupts = <0x2 0xa>;
        };
@@ -386,7 +389,6 @@ random
-           status = "okay";
        };
@@ -444,18 +446,18 @@
-       spi0: spi@7e204000 {
+       spi0: spi: spi@7e204000 {
@@ -480,17 +482,17 @@
-       spi2: spi@7e2150C0 {
+       spi2: spi@7e2150c0 {
@@ -510,6 +512,8 @@
        usb: usb@7e980000 {
+           #address-cells = <0x1>;
+           #size-cells = <0x0>;
@@ -517,6 +521,7 @@
        v3d: v3d@7ec00000 {
+           interrupts = <0x1 0xa>;
        };
@@ -536,7 +541,6 @@ watchdog
-           status = "okay";
        };

```

The diff is produced like this, the resulting diff is sorted:

```
ARCH=arm ./scripts/dtc/dtx_diff arch/arm/boot/dts/bcm2708-rpi-b.dts > ../bcm283x-dtsi/org/bcm2708-rpi-b.dts
<patch>
ARCH=arm ./scripts/dtc/dtx_diff arch/arm/boot/dts/bcm2708-rpi-b.dts > ../bcm283x-dtsi/new/bcm2708-rpi-b.dts
./scripts/dtc/dtx_diff ../bcm283x-dtsi/org/bcm2708-rpi-b.dts ../bcm283x-dtsi/new/bcm2708-rpi-b.dts > ../bcm283x-dtsi/bcm2708-rpi-b.diff
```

Resulting full diffs:
- [bcm2708-rpi-b.diff](https://gist.github.com/notro/bb34d239030f68b0e76a621030f7d065)
- [bcm2708-rpi-b-plus.diff](https://gist.github.com/notro/a9735cdc370b8916e7ac18af9bde7d23)
- [bcm2708-rpi-cm.diff](https://gist.github.com/notro/b5de2ccdbcf16e68ad82f892fa3f93b1)
- [bcm2709-rpi-2-b.diff](https://gist.github.com/notro/98bcb05097bd9ecc17fb6e23ef066916)
- [bcm2710-rpi-3-b.diff](https://gist.github.com/notro/091d440631df462e6c0510bd1d431722)
- [RPI364-bcm2710-rpi-3-b.diff](https://gist.github.com/notro/f439d7d36b4d230f05299e7df6fba30e)
- [bcm2710-rpi-cm3.diff](https://gist.github.com/notro/67d56da6b73ebc473ecd2f826b05242d)

I have only boot tested on Pi1 B+.

One thing I discovered is that we set timer clock-frequency in bcm2708.dtsi and bcm2709.dtsi, but apparently this should be avoided:

[Documentation/devicetree/bindings/arm/arch_timer.txt](https://www.kernel.org/doc/Documentation/devicetree/bindings/arm/arch_timer.txt)

```
** CP15 Timer node properties:

- compatible : Should at least contain one of
        "arm,armv7-timer"
        "arm,armv8-timer"

- clock-frequency : The frequency of the main counter, in Hz. Should be present
  only where necessary to work around broken firmware which does not configure
  CNTFRQ on all CPUs to a uniform correct value. Use of this property is
  strongly discouraged; fix your firmware unless absolutely impossible.
```


popcornmix:
We do set CNTFRQ here:
https://github.com/raspberrypi/tools/blob/master/armstubs/armstub7.S#L82
https://github.com/raspberrypi/tools/blob/master/armstubs/armstub8.S#L72

but that doesn't exist on 2708 (no arch timers).
So perhaps it can be removed from bcm2709.dtsi but not bcm2708.dtsi?

notro:
Sorry I messed up: we set it in bcm2709.dtsi and bcm2710.dtsi.
Yeah, it looks like we can remove it.

I'll wait and see what @pelwell thinks about this PR before I spend more time on it.

pelwell:
I think the changes look good. All overlays still apply. It's just a shame that the multiple inclusions make it harder to follow.

If the bindings say not to include "clock-frequency" then we should take it out.

notro:
> It's just a shame that the multiple inclusions make it harder to follow.

Do you see a better/different way of doing it?
Would you prefer that bcm270x.dtsi was instead a patch on bcm283x.dtsi? Perhaps it's better...

pelwell:
Without trying a few options I wasn't going to suggest there was a better way, just observing that we've spread the information over more levels in the hierarchy, which is good for avoiding replication and less good for legibility.

notro:
**Version 2**

Removed clock-frequency property on the bcm2709/bcm2710.dtsi timer nodes.

pelwell:
Looks good to me.

popcornmix:
No objection here. Can you rebase?

notro:
Rebased

popcornmix:
@MilhouseVH has reported an issue when building LibreELEC with this commit:

> Hi, got a problem with the latest RPi builds - if I include https://github.com/raspberrypi/linux/commit/6e8485a9965ad126a152677c73f8eb6fdbc10fbe in the kernel, then kodi crashes at startup: http://sprunge.us/OADS
> `vcdbg log msg 2>&1| pastebinit` => http://sprunge.us/bEfI
> `dmesg | pastebinit` => http://sprunge.us/jRfb (edited)
> Maybe a clue?
> 
> ```
> rpi22:~ # /usr/lib/kodi/kodi.bin --standalone -fs --lircdev /run/lirc/lircd
> failed to add service - already in use?
> Segmentation fault (core dumped)
> ```
> 
> I've also tried with a completely vanilla `/flash/config.txt` - same result. If I revert the linux commit, then Kodi works as normal
> This is a crashlog from RPi1 showing same error when the linux commit is included: http://sprunge.us/UHAZ - it boots fine without it

Any ideas?

notro:
At first glance, this doesn't look good:

```
[    1.908542] sdhost-bcm2835 3f202000.sdhost: could not get clk
[    1.910784] mmc-bcm2835 3f300000.mmc: could not get clk
```

I also have the sdhost message, have to dig further

notro:
Those messages were just deferred probing.

I don't get this one on Pi2:

```
[    0.047431] uart-pl011 3f201000.serial: could not find pctldev for node /soc/gpio@7e200000/uart0_pins, deferring probe
```

The node ordering in the DT has probably changed due to this patch. But I thought that driver load order was the important thing, since the devices are already present when we get to module loading.

Do you know what kodi is trying to do when it crashes?

popcornmix:
Yes, open display and init GL.
The firmware normally disables these features when it detects the nodes from the vc4-kms-v3d driver. It seems the new device tree is falsely triggering this.
@pelwell is just fixing the firmware to check more explicitly.

hello_triangle shows the failure more simply.

popcornmix:
Firmware has been pushed that fixes this.

MilhouseVH:
Thanks all, confirming LibreELEC now boots and starts Kodi on RPi1 and RPi2/3.

PR #1665, opened 2016-10-04
Rpi 4.4.y HDMI fixes

anholt:
Here's a series of fixes that I've submitted upstream and will probably land in 4.9 or 4.10.


popcornmix:
Quick test with vc4-kms-v3d, vc4-fkms-v3d and without seemed okay.
clk change looks safe to me.

@pelwell okay with you?

drivers/clk/bcm/clk-bcm2835.c
```diff
@@ -523,8 +523,13 @@ static long bcm2835_pll_rate_from_divisors(unsigned long parent_rate,
 static long bcm2835_pll_round_rate(struct clk_hw *hw, unsigned long rate,
 				   unsigned long *parent_rate)
 {
+	struct bcm2835_pll *pll = container_of(hw, struct bcm2835_pll, hw);
+	const struct bcm2835_pll_data *data = pll->data;
 	u32 ndiv, fdiv;
 
+	rate = max(data->min_rate, rate);
+	rate = min(data->max_rate, rate);
+
```
pelwell, line 531: Why not use clamp (http://lxr.free-electrons.com/source/include/linux/kernel.h#L773)?

pelwell (commented):
Fine by me apart from the clamp() suggestion.

drivers/clk/bcm/clk-bcm2835.c
anholt: because I grabbed the wrong commit when picking from the upstream tree.  it's functionally the same, though.

pelwell: Wouldn't it be better to avoid any potential future merge conflicts by pulling the right commit instead?

anholt: I was off on the user qpu branch and just now got a chance to fix this.

Note that github now allows maintainers to edit the contents of pull requests, which is a nice way of dealing with style nitpicks instead of round-tripping to the submitter.

PR #1666, opened 2016-10-04
bcm2835-i2s: Calculate bclk_ratio based on physical width

clivem:
S24_LE is a 24 bit payload in a 32 bit container. When we calculate the
bclk_ratio for S24_LE we need to use 2*physical_width, rather than
2*sample_width.

NB. Using a bclk_ratio of 48 for S24_LE resulted in playback at
incorrect (slow) speed.

(NB. For downstream. Whilst this doesn't affect downstream currently with the dedicated downstream machine drivers that set the bclk_ratio explicitly, using simple-card driver causes i2s driver to calculate the bclk_ratio, wrongly in the case of S24_LE.)


pelwell:
This change looks OK to me. Yes, it's an upstream file and not Pi-specific, but this isn't the first downstream change so I'm happy to merge it.

@popcornmix?

popcornmix:
Any plans for this to go upstream?
@HiassofT @msperl happy with this?

clivem:
Forget it.

PR #1667, opened 2016-10-07
Rpi 4.8.y

kraj:


pelwell:
Please explain how to demonstrate the problem without the patch.

kraj:
@pelwell are you referring to camera patch or the dtb one ?

pelwell:
At least the dtb patch, since it isn't a problem we've encountered in our builds.

kraj:
@pelwell I use yocto to build the kernel, its possible that its only visible with yocto environment.

scripts/Makefile.lib
```diff
@@ -313,7 +313,8 @@ $(obj)/%.dtb: $(src)/%.dts FORCE
 	$(call if_changed_dep,dtc)
 
 quiet_cmd_dtco = DTCO    $@
-cmd_dtco = $(CPP) $(dtc_cpp_flags) -x assembler-with-cpp -o $(dtc-tmp) $< ; \
+cmd_dtco = mkdir -p $(dir ${dtc-tmp}) ; \
+	$(CPP) $(dtc_cpp_flags) -x assembler-with-cpp -o $(dtc-tmp) $< ; \
 	$(objtree)/scripts/dtc/dtc -@ -H epapr -O dtb -o $@ -b 0 \
```
pelwell: By analogy with cmd_dtc, can you change $(objtree)/scripts/dtc/dtc to $(DTC)?

pelwell:
The bcm2835-camera.c change fixes a warning, so thanks for that.

The change to the arch/arm Makefile makes sense, and is harmless if not used (which we obviously aren't).

I'm happy to accept the change to the cmd_dtco in Makefile.lib by analogy to changes in cmd_dtc, but it would be good if you could also use $(DTC) as suggested in my line comment.

kraj:
@pelwell uploaded a v2 addressing your review comment. I also improved the commit message to remove local yocto related stuff from commit msg.

pelwell:
rpi-4.8.y had moved on, causing a conflict, so I rebased and pushed these patches manually.

Thanks!

PR #1668, opened 2016-10-07
[RFC] watchdog: bcm2835: Support setting reboot partition

notro:
Mainline doesn't have support for setting the reboot partition which NOOBS needs.
Before I try and get this upstream I want to get some feedback here first.

Instead of using an out-of-band way to set the partition (/sys/module/bcm2709/parameters/reboot_part) I have used the possibility to pass a reboot command string.
This way the partition number can be passed as an argument to the reboot command.

On Jessie (systemd), this becomes equivalent to poweroff:

```
$ sudo reboot 63
<snip>
[  OK  ] Reached target Shutdown.
[   46.766406] watchdog: watchdog0: watchdog did not stop!
[   47.350282] reboot: Restarting system with command '63'

```

NOOBS can do something like this:

``` c
#include <linux/reboot.h>
#include <sys/reboot.h>
#include <sys/syscall.h>
#include <unistd.h>

int main(int argc, char *argv[]) {
    if (argc != 2)
        return -1;

    syscall(SYS_reboot, LINUX_REBOOT_MAGIC1, LINUX_REBOOT_MAGIC2, LINUX_REBOOT_CMD_RESTART2, argv[1]);
}
```

```
$ gcc -o reboot_part reboot_part.c
$ sync; sudo ./reboot_part 63
[   85.406544] reboot: Restarting system with command '63'

```


notro:
I see that I forgot to remove the _reboot_part_ module parameter in bcm270{8,9}.c. NOOBS can check for the presence of this to know which method to choose.

popcornmix:
Tested on Pi1 and Pi2.
We like the change. I've cherry-picked the commits. Thanks.

PR #1675, opened 2016-10-10
RFC: QPU user shader support in vc4

anholt:
What I took away from my last visit was that the highest priority for being able to switch to vc4-fkms-v3d was having a solution for QPU user shaders like hello_fft.  This merge allows hello_fft to run without even recompiling.

The basic idea here is that we find the mailbox calls that hello_fft makes, and redirect them into vc4 instead.  This is pretty gross, but building a new, clean interface in vc4 would involve pretty major surgery to hello_fft.  Do we want to do that instead?  It would drop the dependency on vc_mem for setting up user shaders, but it would still be root-only, because I'm not planning on writing shader validation for user programs.

Testing status: I'm running hello_fft in a loop while piglit runs, and things seem fine (~50% through the piglit run).

I still need to do some research into the VPM setup to see if I need to dynamically change VPMBASE either for functionality of QPU programs or for performance of 3D, or if it's fine like this.  But before I go too much farther down polishing this path, are there general comments on the architecture here?

### DIFF
--- a/arch/arm/boot/dts/bcm2708_common.dtsi
+++ b/arch/arm/boot/dts/bcm2708_common.dtsi
@@ -375,6 +375,7 @@
 			compatible = "brcm,vc4-v3d";
 			reg = <0x7ec00000 0x1000>;
 			power-domains = <&power RPI_POWER_DOMAIN_V3D>;
+			brcm,firmware = <&firmware>;
 			status = "disabled";
 		};
 
--- a/drivers/firmware/raspberrypi.c
+++ b/drivers/firmware/raspberrypi.c
@@ -26,6 +26,13 @@ struct rpi_firmware {
 	struct mbox_chan *chan; /* The property channel. */
 	struct completion c;
 	u32 enabled;
+
+	struct vc4_dev *vc4;
+	int (*vc4_qpu_execute)(struct vc4_dev *vc4,
+			       u32 num_qpu,
+			       u32 control,
+			       u32 noflush,
+			       u32 timeout);
 };
 
 static struct platform_device *g_pdev;
@@ -66,6 +73,63 @@ rpi_firmware_transaction(struct rpi_firmware *fw, u32 chan, u32 data)
 EXPORT_SYMBOL_GPL(rpi_firmware_transaction);
 
 /**
+ * Peeks at the property request to see if it's something that we
+ * should pass off to vc4 instead.
+ */
+static int
+vc4_filter_property(struct rpi_firmware *fw, uint32_t *data, size_t tag_size)
+{
+	uint32_t tag = data[0];
+	int ret;
+
+	if (!fw->vc4)
+		return -ENOENT;
+
+	switch (tag) {
+	case RPI_FIRMWARE_EXECUTE_QPU: {
+		struct qpu_execute_packet {
+			u32 tag;
+			u32 bufsize;
+			u32 size;
+			u32 num_qpu;
+			u32 control;
+			u32 noflush;
+			u32 timeout_ms;
+		} *packet = (void *)data;
+
+		ret = fw->vc4_qpu_execute(fw->vc4,
+					  packet->num_qpu,
+					  packet->control,
+					  packet->noflush,
+					  packet->timeout_ms);
+
+		packet->num_qpu = (ret != 0);
+
+		return 0;
+	}
+
+	case RPI_FIRMWARE_SET_ENABLE_QPU: {
+		struct qpu_enable_packet {
+			u32 tag;
+			u32 bufsize;
+			u32 size;
+			u32 enable;
+		} *packet = (void *)data;
+		/* If vc4 is present, userspace doesn't get to control
+		 * when the QPUs are off or on.  Just hand back the
+		 * return value indicating success.
+		 */
+		packet->enable = 0;
+
+		return 0;
+	}
+
+	default:
+		return -ENOENT;
+	}
+}
+
+/**
  * rpi_firmware_property_list - Submit firmware property list
  * @fw:		Pointer to firmware structure from rpi_firmware_get().
  * @data:	Buffer holding tags.
@@ -86,6 +150,14 @@ int rpi_firmware_property_list(struct rpi_firmware *fw,
 	u32 *buf;
 	dma_addr_t bus_addr;
 	int ret;
+
+	/* NOTE: We're only handling filtering on the first property
+	 * here, and if it gets filtered then we skip the rest of
+	 * them.  This is enough for hello_fft.
+	 */
+	ret = vc4_filter_property(fw, data, tag_size);
+	if (ret != -ENOENT)
+		return ret;
 
 	/* Packets are processed a dword at a time. */
 	if (size & 3)
@@ -265,6 +337,22 @@ struct rpi_firmware *rpi_firmware_get(struct device_node *firmware_node)
 	return platform_get_drvdata(pdev);
 }
 EXPORT_SYMBOL_GPL(rpi_firmware_get);
+
+/**
+ * Called by the vc4 driver at its probe time, to request that QPU
+ * execution requests be redirected to it.
+ */
+void rpi_firmware_register_vc4(struct rpi_firmware *fw, struct vc4_dev *vc4,
+			       int (*qpu_execute)(struct vc4_dev *vc4,
+						  u32 num_qpu,
+						  u32 control,
+						  u32 noflush,
+						  u32 timeout))
+{
+	fw->vc4 = vc4;
+	fw->vc4_qpu_execute = qpu_execute;
+}
+EXPORT_SYMBOL_GPL(rpi_firmware_register_vc4);
 
 static const struct of_device_id rpi_firmware_of_match[] = {
 	{ .compatible = "raspberrypi,bcm2835-firmware", },
--- a/drivers/gpu/drm/vc4/vc4_drv.h
+++ b/drivers/gpu/drm/vc4/vc4_drv.h
@@ -70,6 +70,9 @@ struct vc4_dev {
 	 */
 	uint64_t finished_seqno;
 
+	uint64_t qpu_emit_seqno;
+	uint64_t qpu_finished_seqno;
+
 	/* List of all struct vc4_exec_info for jobs to be executed in
 	 * the binner.  The first job in the list is the one currently
 	 * programmed into ct0ca for execution.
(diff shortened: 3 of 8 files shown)


popcornmix:
Looks interesting. I'll see if I can get it running with qpu accelerated HEVC decode.

maazl:
I did not fully get how you did the hack, but I had another approach to be able to run shader code concurrently. Maybe the ideas could be combined.
Well, not really concurrently as there are some exclusive resources like VPM and semaphores that cannot reasonably be shared. But different user space applications may execute prepared shader code in a queue without knowing each others.

The basic idea is to move the mailbox code from hello_fft into the vcio device driver - in fact the way you avoided.
This addresses several issues. First of all the sharing of the GPU shaders but also the issue that all applications need to run as root (and therefore cannot reasonably be on nfs shares). And from my point of view even more important: resource cleanup. The driver keeps track of allocated GPU memory and releases it if the application terminates unexpectedly. Every time you press Ctrl-C in hello_fft or even just miss to prepend it with sudo you need to reboot to get the memory back.

The source code is available at http://maazl.de/project/vcio2/doc/index.html it compiles as part of the kernel as well as with DKMS. But it is quite old and has to be updated for newer kernels. Furthermore DKMS is quite useless becaues the Raspi kernels do not provide kernel headers in general. I plan to revive this work and provide a kernel patch for 4.4 kernels.

From the technical point of view the driver just adds some new IOCTLs to vcio. But for security reasons a new device vcio2 is used. Just because it is reasonable to allow access to the new driver API to some other group than root. The driver does some basic validation before executing shader code. Of course no full validation, because that would take much longer than to execute the code. It is still a major security issue to grant access the this device - but who cares on a device where sudo requires no password? The validation is only intended to prevent some crashes due to accidental, wrong calls during development. E.g. you can only map GPU memory into your process that _you_ have allocated.

The driver is compatible in the way that you still can execute the old hello_fft as root. But a new, modified hello_fft can use the driver and does not need root priviledges. (I already did the changes - no big deal).

Putting this new API together with your idea of redirecting the calls to vc4 would be even more sophisticated. From the vcio2's point of view it is even more straight forward to do so.

By the way: your hack probably will sadly fail for small transforms and small batches. hello_fft uses direct access to the V3D registers in this case. This is broken by design and there is nothing we can do to get this working reliable.
It is done for performance reasons because the way through the mailbox takes longer than the entire shader code for small batches. Do you know how the vc4 redirect behaves in this context? Does it use the mailbox channel too? - Hopefully not, because there are other drawbacks. The power management calls to change the clock frequencies go through the same mailbox channel. This prevents the cpufreq governor changes from executing while shader code is executing.
I did not fully get how you did the hack, but I had another approach to be able to run shader code concurrently. Maybe the ideas could be combined.
Well, not really concurrently as there are some exclusive resources like VPM and semaphores that cannot reasonably be shared. But different user space applications may execute prepared shader code in a queue without knowing each others.

The basic idea is to move the mailbox code from hello_fft into the vcio device driver - in fact the way you avoided.
This addresses several issues. First of all the sharing of the GPU shaders but also the issue that all applications need to run as root (and therefore cannot reasonably be on nfs shares). And from my point of view even more important: resource cleanup. The driver keeps track of allocated GPU memory and releases it if the application terminates unexpectedly. Every time you press Ctrl-C in hello_fft or even just miss to prepend it with sudo you need to reboot to get the memory back.

The source code is available at http://maazl.de/project/vcio2/doc/index.html it compiles as part of the kernel as well as with DKMS. But it is quite old and has to be updated for newer kernels. Furthermore DKMS is quite useless becaues the Raspi kernels do not provide kernel headers in general. I plan to revive this work.

From the technical point of view the driver just adds some new IOCTLs to vcio. But for security reasons a new device vcio2 is used. Just because it is reasonable to allow access to the new driver API to some other group than root. The driver does some basic validation before executing shader code. Of course no full validation, because that would take much longer than to execute the code. It is still a major security issue to grant access the this device - but who cares on a device where sudo requires no password? The validation is only intended to prevent some crashes due to accidental, wrong calls during development. E.g. you can only map GPU memory into your process that _you_ have allocated.

The driver is compatible in the way that you still can execute the old hello_fft as root. But a new, modified hello_fft can use the driver and does not need root priviledges. (I already did the changes - no big deal).

Putting this new API together with your idea of redirecting the calls to vc4 would be even more sophisticated. From the vcio2's point of view it is even more straight forward to do so.

By the way: your hack probably will sadly fail for small transforms and small batches. hello_fft uses direct access to the V3D registers in this case. This is broken by design and there is nothing we can do to get this working reliable.
It is done for performance reasons because the way through the mailbox takes longer than the entire shader code for small batches. Do you know how the vc4 redirect behaves in this context? Does it use the mailbox channel too? - Hopefully not, because there are other drawbacks. The power management calls to change the clock frequencies go through the same mailbox channel. This prevents the cpufreq governor changes from executing while shader code is executing.

anholt:
@maazl If we want proper memory management, then we should add an interface to vc4 for running user shaders.  Memory management is already handled well there, and then you wouldn't have to go through mad inter-driver hacks like this to get the QPU code executed.

If we want non-root access, then we need to do validation of the VPM writes.  I can't merge code for non-root upstream otherwise, so Raspberry Pi would be stuck carrying the patches forever.

But for now, I'm just looking to get to the point of "old code continues working as much as possible" so that we can switch to vc4 by default.

maazl:
@anholt 
Everything is fine, if the memory management is already available in vc4.

However, for the root issue: if a second device, maybe with restricted functionality, can be created we could come around the security restriction. The device could have root access by default - like other devices too. And if a user decides to change it's group and/or permissions then it is not the kernels fault. So the user has the choice either to run the applications as root or change the security.

Regarding shader code validation: it is impossible to do this with reasonable performance. You need to do a full emulation of the code including uniforms and memory reads because the VPM write address is likely the result of a more or less complex calculation. And even reading arbitrary memory is a security hole.
The only way around would be a higher language level like OpenCL - I could bet no one will do the job until the Raspberry Pi finally gets obsolete.

There is nothing to be said against the compatibility of the vc4 driver. If it provides an interface for user shaders sooner or later - even better.

But we should find some kind of standard solution with a reasonable compromise between usability, performance and safety to get the GPU shaders used by different applications. E.g. a resampler for PulseAudio or libgpufft.

ghost:
Hi, I just stumbled on this and it's a bit interesting to me.  I don't know much about 3D graphics other then the very basics.

What I don't understand here is if people are looking for a way to run arbitrary code on the GPU, or if they are talking about just getting hello_fft to work. Since hello_fft is only about 1.000 lines of C code, I would say fix hello_fft rather then try to make the existing stuff work.

As for running arbitrary GPU code, I don't see how that could possibly get validated.  Even it if was somehow emulated first, I don't think it could be validated.

As for root/non-root, it seems to me that it should be root only.   Raspbian logins in as root by default, so the only interesting case is if someone is trying to harden their system.  If they are they are doing it it is for a good reason.  And if they really don't want to login as root by default, perhaps they could stub out part of the application into a seperate exe that they mark as SUID?

Finally, I don't understand why this should go in VC4.  This isn't really a graphics issue, it's about running arbitrary code on the GPU.   Maybe you want to use part of hello_fft from some kind of sensor data analysis and you don't want a video monitor connected at all and you don't want VC4 loaded because of the extra memory requirements.   You only want the FFT part.

pelwell:
Andrew Holme, the author of hello_fft, describes the implementation and the motivation behind it [here](http://www.aholme.co.uk/GPU_FFT/Main.htm). As well as being a useful tool in its own right, hello_fft is an example of how one can use the QPUs, the processing elements that normally run shaders, for GPGPU/OpenCL/CUDA-style acceleration of parallel compute intensive tasks. This repurposing of the QPUs requires the cooperation of the VPU-based V3D driver, which is why any replacement ARM-based V3D driver should have a similar mechanism - we don't like to remove facilities unless absolutely necessary.

ghost:
Perhaps I better take this topic offline while I digest the article.

I think my only real question is simple:
It seems to me vector processing is a more general application then shaders.   Why should vector processing be implemented on top of shaders as opposed to shaders being implemented on top of vector processing?

I mean vc4 could itself call some other internal driver that just manages access to the QPU. And this QPU could have direct user mode access.  Why load that huge video component just to run a software radio for a simple headless device?

pelwell:
> It seems to me vector processing is a more general application then shaders. Why should vector processing be implemented on top of shaders as opposed to shaders being implemented on top of vector processing?

It isn't. If running shader code is all you want to do then the V3D driver will do that for you - it wouldn't be an OpenGL driver if it didn't.

The mechanism on which hello_fft is built allows you to write programs in QPU assembler. This isn't the same as VPU assembler - you aren't getting access to the main VPU processors - but you can do a huge amount of vector-based number crunching while playing nicely with the OpenGL stack.

@anholt is very familiar with the Linux goals (each component should do one thing and do it well) and the sometimes competing goal of maximum performance. If we end up with a monolithic solution it will be for good reasons.

ghost:
OK, I think I understand better here.

I guess I'll leave with food for thought.  If I have a humidity/temperature sensor with a long range software radio(for cost reasons), why would I want the extra RAM and FLASH ram requirements of having opengl, which probably implies X-Windows and who know what else just to send temperature and humidity data?

pelwell:
Remember that you would be in the minority - why would the GUI have to run slower just because some guy wants to use SRDS?

Seriously, if VC4 does go the monolithic route then there is nothing stopping somebody taking those elements required to just run code through the QPUs (trivial, I'm sure...) and packaging them as an alternate driver for the V3D hardware. Device Tree overlays would let you select your preferred driver at boot time.

ghost:
OK, this is quickly turning into a political land-mine which perhaps is better I not get to involved in while I still have my feet.

I just don't understand why it needs to be completely an XOR kind of deal here.   Some devices may want just have the radio, some devices may want just cool video, and some devices may need BOTH cool video and a cool cheap radio.   If you have specific drivers you would need to have three different sets.  One of each scenario.

Can you imagine how complex the device trees would be in a situation like that?

pelwell:
> I just don't understand why it needs to be completely an XOR kind of deal here.

Because ultimately one entity needs to be responsible for the allocation and programming of the QPUs, at least if you want a dynamic allocation (I don't know how feasible a static partitioning between two drivers would be). There are logically three entities - a module that allocates QPUs (A), a V3D driver module (B), and a module that presents a user-space interface to the QPUs (C). Both B and C rely on the services of A. In any particular implementation A, B and C may or may not be implemented as fewer than 3 drivers/modules, but the logical roles remain.

ghost:
I agree, but the difference would be long term testing and maintenance.   If this were implemented as three drivers where A is always needed, B and C can be switched out as needed.  

The interesting case if what happens if all three components take independent development paths.   Every time VC4 gets improved, then the combo driver would need a gazillion patches applied to it. Sounds like a big mess to me.

I originally started this because it just wasn't clear to me that representing the QPU with the shader abstraction is the best way to go.  Now that I see radios are involved, it completely doesn't make sense.

BTW, I have no interest in trying to implement or own any of the components we have kicked around.  I was simply trying to understand and now I see this whole issue is quite deep.

Ruffio:
@anholt @pelwell @ghost what is the status on this PR? Have we reached a conclusion of what to do?

Ruffio:
@anholt @pelwell @ghost what is the status on this PR? Have we reached a conclusion of what to do?

pelwell:
vc4 isn't my driver so I don't think it is up to me - you can replace all my comments above with "whatever Eric says" or "because Eric wants to do it that way".

JamesH65:
This stalled a year ago - any suggestions on how to progress it? (or close it..)

anholt:
I'm still interested, but we need confirmation that the plan to have vc4 own QPU execution and the firmware calls back to us for any internal QPU execution it needs is the way forward.

Ruffio:
@anholt and how do we get this confirmation?

PR #1678, opened 2016-10-11
bcm270x: Platform file cleanup

notro:
Once I got started I couldn't stop, so I did all the cleaning I could in the platform/board files.
Not much platform specific code left now.

Tested on Pi1 and Pi2 running overnight. Boot tested on Pi3.


popcornmix:
I like the look of dropping lots of code, so +1 from me.
Booted okay on Pi1 and Pi2.

@pelwell does it look okay to you?

arch/arm/mach-bcm2708/bcm2708.c
```diff
@@ -19,125 +19,23 @@
...
-				   NULL);
-	if (ret) {
-		pr_err("of_platform_populate failed: %d\n", ret);
-		BUG();
-	}
-
-	bcm2708_init_uart1();
-
 	system_rev = boardrev;
 	system_serial_low = serial;
```
pelwell, line 36: I think we can also lose the module parameters and "linux,serial" and "linux,revision" as bcm2835 does, which would leave the only difference being the cma initialisation and the board ID.

arch/arm/mach-bcm2709/bcm2709.c
```diff
@@ -19,162 +19,36 @@
...
-				   NULL);
-	if (ret) {
-		pr_err("of_platform_populate failed: %d\n", ret);
-		BUG();
-	}
-
-	bcm2709_init_uart1();
-
 	system_rev = boardrev;
 	system_serial_low = serial;
```
pelwell, line 36: As proposed for bcm2708, use "linux,serial" and "linux,revision" instead.

pelwell:
I think it's great - there's virtually nothing there. Which is mildly frustrating...

I'm happy for this to be merged, ideally with my proposed serial number change (I've tried it, and it works as expected).

popcornmix:
@notro Thanks for this - a very nice clean up.
I've pushed Phil's suggestion for using DT rather than modules params for serial and board rev.

notro:
Great, maybe we even can drop the serial number bit now that /serial-number is populated.

I'm gonna try and convert to ARCH_MULTIPLATFORM now. When I tried this last year, I got stopped by memory corruption issues, hopefully they have been magically resolved :-)

pelwell:
I considered relying on the "serial-number" property, but that would mean that system_serial_low would not be set (which may be technically irrelevant), and I didn't like the asymmetry. Plus, I thought it was better to reduce the differences with the upstream code. But none of those are strongly held views.

notro:
system_serial_low is only used to set system_serial if the latter is not already set. system_serial is the one displayed in /proc/cpuinfo.
If by upstream you refer to board_bcm2835.c, then the serial/revision code is a [downstream patch](https://github.com/raspberrypi/linux/commit/b20b6c9db3b0ac6b5daad77e06c8390e617c5aa3) added by me.
But we can deal with this later.

This is the last attempt I know about to get support for 'revision':
ARM: /proc/cpuinfo: DT: Add support for Revision
https://patchwork.kernel.org/patch/5712301/

PR #1686, opened 2016-10-15
ARM64: Modify default config to get raspbian to boot

ghost:
1. Enable emulation of deprecated instructions.
2. Enable ARM 8.1 and 8.2 features which are now detected at runtime.
3. Switch the default governer to powersave.
4. Include the watchdog timer driver in the kernel image rather then a module.

Tested with raspbian-jessie 2016-09-23.


popcornmix:
Looks fine to me. @pelwell ?

pelwell:
They all make sense to me apart from the powersave governor - why use this instead of the on-demand governor?

ghost:
I changed back to the powersaver governor mostly to minimize the difference between the 32 bit and the 64 bit config.   I was also having some issues in the past where it was switching back and forth very rapidly and occasionally locking in the firmware.

I also found that a standard 6 foot USB cable has a bit too much resistance and thus a bit too much voltage drop for the RPI 3.   I don't see the issue nearly as much after switching to 3 foot cables, but I don't know about others.  The powersaver governor is much less demanding.

I mostly want to get this check-in because it appears that most people still want to try raspbian with a 64 bit kernel, and the changes from the check-in will at least help people get started.

popcornmix:
@pelwell raspbian switches to the ondemand governor during boot

pelwell:
Well, the governor can always be changed at runtime, so I have no objections.

Thank you for your contributions.

pelwell:
Well, the governor can always be changed at runtime, so I have no objections.

Thank you for your contributions.

PR #1693, opened 2016-10-20
bcm270x: Get bus address from Device Tree

notro:
This is some more cleanup of bcm270x.
I was hoping to convert to ARCH_MULTIPLATFORM, but that didn't work out.

The main change in this PR is to get the bus/dma address from the Device Tree 'dma-ranges' property instead of hardcoding it.

Tested on Pi1 and Pi2.

@popcornmix please check with the GPU debugger to make sure that this doesn't give `L2 bad address exception` like it did when I tried this last year: https://github.com/raspberrypi/linux/pull/1178#issuecomment-153525832

Next step is to see if I can find which multi platform kconfig knob(s) that breaks the downstream drivers. CONFIG_ARM_PATCH_PHYS_VIRT=y is one suspect.


notro:
I will make a new PR that also converts to multi platform.

Issue #1172, opened 2015-10-19
Enable user namespaces and seccomp


CTassisF:
Hello.

I am trying to use Firejail[0] on my RaspberryPi running Raspbian but it shows two warnings:

Warning: user namespaces not available in the current kernel.
Warning: seccomp disabled, it requires a Linux kernel version 3.5 or newer.

Can you enable this features on the next kernel release?

My current uname -a is:
Linux RaspberryPi 4.1.7+ #817 PREEMPT Sat Sep 19 15:25:36 BST 2015 armv6l GNU/Linux

Thanks in advance,
César

[0] https://l3net.wordpress.com/projects/firejail/

Ferroin:
It is also worth noting that both of these are required for Chromium and Google Chrome to properly implement sandboxing of their various sub-components.

Ruffio:
@CTassisF has your issue been resolved? If so, please close this issue. Thanks.

mpx:
User namespaces are disabled from the current kernel:

```
root@pi4:~# uname -r
4.4.13-v7+
root@pi4:~# modprobe configs
root@pi4:~# zgrep -E 'CONFIG_(USER_NS|SECCOMP)' /proc/config.gz
# CONFIG_USER_NS is not set
CONFIG_SECCOMP=y
```

pelwell:
With all kernel options that aren't simply loadable modules we are concerned about increased kernel size and reduced performance. If you were to present "before and after" comparisons of free memory and performance (using some standard benchmarks) it may strengthen your case.

jcberthon:
@pelwell what kind of benchmark are you after? Disk block IO, file system IO, CPU cycles, etc.?

I want seccomp for Docker on the Raspberry Pi, it is an important protection. And user namespace as well, this allow for mapping user ID within a container to different user ID on the host. This is important for example for user ID 0 (aka root)! This is not limited to Docker, but LXC and LXD requires user namespace for running unprivileged containers.

So I'm ready to measure memory before and after, but for benchmark, what do you expect?

pelwell:
How about running dd from /dev/zero to /dev/null with three block sizes - 1, 4k and 1024k - and a count set to make each one take at least 10 seconds? I'm concerned about performance for non-sandboxed processes, but I would be curious to compare with sandboxed as well, so a 3x3 grid of results (MB/s for each of three sizes in a kernel without seccomp, with seccomp, and sandboxed) would be great.

jcberthon:
[...]

Issue #1555, opened 2016-07-06
How to enable Apparmor when build kernel?


zfyseu:
Hello, everyone!

I want to use Apparmor on raspbian.
Is there anyone know how to build the kernel so that I can use Apparmor?

Thanks!

popcornmix:
Some info here: https://github.com/raspberrypi/firmware/issues/468
But the options are not going to be added to the default raspbian kernel.

For info on how to build the kernel or whether other distributions may support this, it is best to ask on the [forum](www.raspberrypi.org/forums).

zfyseu:
Thanks , what does kernel options mean?

popcornmix:
Ask in the forum. This isn't the right place to discuss this as ~500 devs gets email notifications on every message here.

zfyseu:
Hello popcornmix!

I have followed the link "https://www.raspberrypi.org/documentation/linux/kernel/building.md" to cross compile my raspbian kernel.
But I don't know how and where to eidt kernel option to support apparmor?

Do you know how to do it ?
Thanks!

clivem:
> But the options are not going to be added to the default raspbian kernel.

I must be having a bad dream, having just looked at what got added to rpi-4.8.y tree configs? ;)

zfyseu:
edit  "/linux/arch/arm/configs/bcm2709_defconfig" by adding the following kernel options:

CONFIG_KEYS=y
CONFIG_SECURITY=y
CONFIG_SECURITYFS=y
CONFIG_SECURITY_NETWORK=y
CONFIG_SECURITY_PATH=y
CONFIG_SECURITY_APPARMOR=y
CONFIG_SECURITY_APPARMOR_BOOTPARAM_VALUE=1
CONFIG_SECURITY_APPARMOR_HASH=y
CONFIG_INTEGRITY=y
CONFIG_INTEGRITY_AUDIT=y

and then run the command "make ARCH=arm CROSS_COMPILE=arm-linux-gnueabihf- bcm2709_defconfig"

combs:
i was able to get apparmor running on a Pi3 with these flags!

Issue #1326, opened 2016-03-06
rpi3: TTYS0 does run too slow


msperl:
The new TTYS0 as the default interface does not run at 115.2k but instead at 71.8k using firmware version: 6e8b794818e06f50724774df3b1d4c6be0b5708c.

msperl:
So looking at the DT I see:
- we use a multiplier clock clk_uart1 for the auxiliar clock 
  - we multiply the core clock by 2
- core clock is 400MHz (cat /sys/kernel/debug/clk/core/clk_rate)
- so the multiplier clock (to make the uart driver happy, as the HW only uses a internal divider of 8 not 16 as typical other divices) is 800MHz. (cat /sys/kernel/debug/clk/clock/clk_rate)

But this seems to be incorrect - the actual clock should (still) be 500MHz.
115200/800*500 = 72000kHz
So this still means we run the VPU clock at 250MHz while we assume it runs at 400MHz.

So maybe this is a firmware issue not a kernel issue itself.

pelwell:
1) Remember that the downstream RPi DT isn't changing any clock settings, it is telling Linux what they are. By declaring clk_uart1 to be twice the actual frequency we are getting the uart driver to generate a divisor that is twice as large. This factor of two accounts for the discrepancy in the oversampling rate:

```
baudrate = core_clk/(16 * div) = core_clk/(8 * div'), where div' = div*2
```

2) The actual difference in baudrate stems from the fact that, by default, the core clock varies with turbo mode. The default turbo core clock is 400MHz, while the minimum, non-turbo clock is 250MHz. If you wish to use the mini uart, and not everybody does, then you need to stop the core clock from changing. If you have an adequate power supply to avoid under-voltage, and adequate cooling for the work load, then you can set `force_turbo=1` to keep the core clock at 400MHz. If instead either of those two pre-requisites is in doubt then your only option is to limit the core frequency to 250MHz (`core_freq=250`).

msperl:
The device tree shows:
- uart1_clk is a multiple of 2 of core
- core is a fixed clock with a value of 400MHz - i.e. this is hardcoded in the device-tree!

This results in a uart1_clk of 800MHz

Looking at the distinct registers for vpu, PLLC_CORE0 and PLLC.

It seems as if we really run:
- PLLC at 1GHz (19.2MHz \* 52.08333206176757812500)
- PLLC_core0 at 500MHz (divider of 2)
- vpu at 125MHz (divider of 4)

[...]

Issue #766, opened 2015-01-23
SIGILL on using GDB, in ld-linux-armhf.so.3


sneakernets:
Disclaimer: I'm using the latest firmware...

I am working on [zdoom](https://github.com/rheit/zdoom), a Doom source port. However, when I attach zdoom to GDB on the Raspberry Pi, I get this:

```
Program received signal SIGILL, Illegal instruction.
0xb6fe8acc in ?? () from /lib/ld-linux-armhf.so.3
```

`bt` produces nothing of worth.

This is interesting, in that the program "runs" fine otherwise, but not with `gdb`.

Compiler: gcc 4.6.3-14+rpi1
Linux: 3.18.3+

popcornmix:
I believe this is normal (I've seen this when debugging xbmc for a couple of years).
I suspect the code is deliberately executing the illegal instruction with a signal handler enabled (e.g. to determine whether optional instructions are supported).
See:
http://stackoverflow.com/questions/15071625/gdb-ddd-program-received-signal-sigill

sneakernets:
I actually checked for this too, but even when I use  `handle SIGILL pass nostop noprint`, the program dies in `gdb` with the same error. I suppose I should have noted that in the first post. 

This has been hair-pullingly frustrating to figure out on my own for the past two weeks, so I threw in the towel today and submitted it as an issue. :grin:

nbrachet:
I have experienced this issue too after a recent upgrade -- sadly I don't remember which one.

I traced this SIGILL down to a static initialization in my code. Still frustrating issue.

popcornmix:
@nbrachet A firmware/kernel upgrade, or other packages?

nbrachet:
@popcornmix I'm afraid both.

nbrachet:
If that helps the problem occurs whether I build with g++-4.6 or g++-4.8.

popcornmix:
If it was a firmware/kernel update that caused it, then downgrading should fix it. You can revert to an older version:
http://raspberrypi.stackexchange.com/questions/15917/restore-kernel-version/15919#15919

sneakernets:
If it helps, this started to be an issue around mid-december 2014.

nbrachet:
[...]

PR #1698, opened 2016-10-27
config: Add SECCOMP_FILTER and APPARMOR

jcberthon:
This change provides updated configuration for bcm2709 (RPi2/3) and bcmrpi (RPi1/Zero) to support user namespaces (which has already been added in rpi-4.6.y+ branches; https://github.com/raspberrypi/linux/commit/24438a33d86227d4796fb4d2b81e1ff8f0755fbf and https://github.com/raspberrypi/linux/commit/d7cbfb857fd6ba101da996c221ed5ecc50e522cf), seccomp filterring (warning, this deactivate CONFIG_OABI_COMPAT, see issue #1605, but I think this is for the better; https://github.com/raspberrypi/linux/commit/ad1274ccccfdf7849a2de5064cf544be0569d45a and https://github.com/raspberrypi/linux/commit/1dfa4648ad38b2c2b90310d2918c3371dbd746b6), AppArmor (https://github.com/raspberrypi/linux/commit/341d06b894e710531284f9bfb20283c574508621 and https://github.com/raspberrypi/linux/commit/9f2b55e696ddddbb4c180980fbdd6b26ff56dd6b).

As added bonus, I have added a few extra changes which offer better control to limit resources of a process or container. These are covered by all other commits of this pull request.

This pull request should address the problems raised in issue #1605 and in issue #1172. They will provide better security for the Raspbian system, and they can be used by container technologies (e.g. Docker but not limited to it) to enhance the containment, resource limitation and overall security. The impact of such change is very limited, there was close to no increase in memory by activating these configs, and the performance were a bit impacted only in worse case scenario with many many system calls. The impact has been documented in issue #1172 and is in my opinion totally acceptable given the advantages :-) .

_Note: to apply this pull request to newer branch (i.-e. rpi-4.6.y onward), the commits need to be cherry-pick as the username space commit will conflict with a change already implemented. This is also true for the commits https://github.com/raspberrypi/linux/commit/05fd8f389887c5717a95790c13084828ed1750dd and https://github.com/raspberrypi/linux/commit/25cf4608e8dc845cb5230ce24f5c1dca6df2ce80 which activate further cgroup resource management configs, but for some they are already present in newer branches. If wished, I can provide different pull request for each branch, just let me know how shall I proceed best._


pelwell:
Putting aside for a moment the performance, which looks pretty good (perhaps suspiciously so), there is a problem with the SECCOMP patches in that they disable CONFIG_OABI_COMPAT. As discussed in #1605, #652, #766 etc., this is still required for the Raspbian version of gdb, so we are not prepared to remove it just yet.

Debian Stretch is currently set to include gdb 7.11, which contains a patch removing the need for OABI_COMPAT. Stretch won't be ready until early next year, and it will take us some time to base our standard images on it.

Fortunately we have the ability to apply downstream patches to Rapsbian packages before release. @XECDesign already has the patch applied locally, and is prepared to update the published package to include it, at which point we will probably (I'm not the only person to convince) merge these changes and tell people who care to update their gdb.

jcberthon:
Regarding the performances, my suspicion is that by simply activating the seccomp filtering capabilities that should not impact the performance, except probably if we apply a seccomp filtering profile to a process. But by default seccomp filtering is not used by the system. So for me the little variations between the vanilla and the seccomp kernel are just "noise". What would be interesting is to to a benchmark within Docker when we apply a seccomp profile or when we disable it. But as I mentionned in the issue, the seccomp user space tools are too old in Debian Jessie (although there is a backported version that should work in the Debian Jessie armhf repo), so I cannot use seccomp filtering anyway with Docker.

For AppArmor, this is another story because it is active and running for many processes. And although there are no profile for dd, there is a small performance impact in the worse case scenarios.

As for test3 which looks almost the same across all environments, I suspect we are hitting other performance ceiling, so the kernel changes are not reflected in this test.

What about setting seccomp aside for the time being, but at least merging the other changes?

pelwell:
The gdb package has now been updated (`sudo apt-get update; sudo apt-get install gdb` to install - thanks, @XECDesign), so that hurdle has been removed. But since this PR, if merged, will affect all Raspberry Pis, we need some more scrutiny.

The next step might be to merge these commits, or a subset, into the "next" branch, although that may have to wait until the network booting beta has finished with it.

jcberthon:
That's great news!

All my testing were done on a Raspberry Pi 2 which I use mostly as a low power network server. So it's only one model out of many and for one use case. So it needs indeed further testing on the other platforms and when used in electronic projects (e.g. When using the GPIO I/f) or other use cases.

But I can't help much but continue testing my platform and use case. If you see anything else I could do, let me know. 

PS: as a side note, I do not expect much impact as the activated features are not used unless explicitly setup. Even AppArmor is not active if no profile is installed. But of course between what we expect and reality there is often a big difference 😉

popcornmix:
@jcberthon files in arch/arm/configs should only be created with `make savedefconfig` (i.e. only include non-default option), never manually edited. (e.g. `CONFIG_OABI_COMPAT=n` is not a valid syntax, the line should just be removed).

jcberthon:
Hi @popcornmix 

I didn't know, thanks for the info. Do you want me to correct my pull request? Or will you just correct it after merge?

ghost:
Hi,

I'm not sure this is the correct location to discuss this.  Maybe I should open a different issue/etc.

I'm wondering why the need for two completely different files bcmrpi_defconfig and bcm2709_defconfig.  Plus, I know this is still in the far future bcmrpi3_defconfig.   Especially when I can count on one hand the number of lines that are different.

What I was thinking is why not have a common file with most of the configuration that gets appended with an architecture specific version.   Maybe the cat files could then be run through savedefconfig.

Another option maybe to run the standard configuration through a preprocessor first to generate the base configurations.

djtm:
I can confirm there is no performance hit. I've done some further benchmarking, including a kernel with strong stack protection as well: 
http://openbenchmarking.org/result/1611266-SO-APPARMORS15

I even compiled in selinux (though I didn't actively use it, maybe systemd did). The results indicate that apparmor has pretty much no performance impact, strong stack protection has a negible impact (only apache static pages drops). Im my case it's active and running in the background ("Security Performance"). I've used the sources from this repository for all kernels.

A much strong impact comes from heat during the benchmark. The first run is usually the fastest (5 % deviation). I do have a closed case.

I'll next try to compile the software with stack protection as well though that's probably not as much of concern here?

**I believe security is a very important issue on these devices. There are many, they are often online 24/7 with several attack surfaces and limited oversight.** AppArmor and SELinux are very important tools to shield against attacks. As is stack protection. I believe it should be as easy as possible to users to use them. I would vote for including strong stack protection and maybe even SELinux, e.g. if the kernel is used in Fedora systems.

Compiling your own kernel is not always an option. When I upgraded the git repo with the kernel sources, I had a merge conflict, though I had not changed any source files. That makes upgrading the kernel even slower.

fpietrosanti:
The GlobaLeaks project would introduce support to run Whistleblowing Platform on Rasberry Pi out of the box if it would only support AppArmor by default. +1

djtm:
I've been runnig my system for three months with these features enabled (plus strong stack), compiled from source, and I've had no issues (except an unrelated wifi issue recently). Kodi works fine, too.

@jcberthon: If seccomp does not impact performance being disabled, I think there is no issue, since the user can control whether to enable it or not. I'm using the docker arm repo (`deb https://apt.dockerproject.org/repo raspbian-jessie main`) and I think it's got seccomp and apparmor enabled. Would some further benchmarks help?

jcberthon:
@djtm I'm also using Docker from the same repo. That particular built of Docker is NOT built with SECCOMP filtering support because it requires the seccomp user space tools to be version 2.2.1 or newer (as I mentioned in this comment https://github.com/raspberrypi/linux/issues/1172#issuecomment-252655607). However Debian and Raspbian Jessie is using an older version (2.1.1).

Debian has a backported repository in which there is a new version of the seccomp user space tools (backported from testing). However, I have tried to add this repo and install only the libseccomp2 and its dependencies and then recompile Docker to use it, but I can't remember now what was the problem but something did not work, so I reverted the changes.

Just execute `docker info` and check what `Security Options:` are reported. If you see seccomp you are all good, if you do not see it, it means it is not available on your platform.

On Raspbian Jessie with my custom Kernel and AppArmor installed:

```
$ docker info
(...)
Server Version: 1.13.1
Storage Driver: overlay2
(...)
Security Options:
 apparmor
Kernel Version: 4.9.5-v7-lowlat-tick-rtc1307+
```

jcberthon:
Note: I've seen that some of the propose changes in my commits (which was forked from the 4.4 branch) are now included in newer branch, so that if you take branch 4.9 there is less changes to be included.

So if you want me to redo the commits and propose a new PR for branch 4.9, just let me know.

JamesH65:
@jcberthon Please rebase over the latest kernel. @popcornmix @pelwell OK to incorporate this once rebased?

jcberthon:
Hi @JamesH65 

I will try to rebase it during the coming days.

JamesH65:
Thanks.

jcberthon:
I wanted to rebase, but that did not work. So I checked out rpi-4.9.y and then cherry picked the commits. Should have the same effect.

I update the pull request, including the title as now in the rpi-4.9.y branch the `USER_NS` option is already activated.

I'm currently building the kernel and will report if it still works. I do not expect surprises, because I already compile my kernel regularly from the rpi-4.9.y branch with those options plus a few extra ones.

jcberthon:
Hi @JamesH65 

I can confirm that with the above commits I can successfully build the kernel and run it.

LGTM

JamesH65:
ping @pelwell @popcornmix Seems to me from comments above that we can merge this?

jcberthon:
Any progress regarding merging?

JamesH65:
@popcornmix @pelwell This PR appears to have stalled...

jcberthon:
I'm ready to rebase or do other changes, just let me know.

jcberthon:
Hi @popcornmix @pelwell

How can I push further this PR? The commit 3f258a7 is also bringing missing config file to the Raspberry Pi 2 and 3 kernel, configs very useful for Docker containers.

I can of course rebase again on the top of the 4.9, and also propose this change to other branches if you prefer. Let me know how I can best work to get this PR forward.

JamesH65:
Having a clear out of PR's.  SECCOMP_FILTER is now turned on, but there appear to be no settings for APPARMOR in the current kernel to turn on. Should this PR's be closed?

This issue will be closed within 30 days unless further interactions are posted. If you wish this issue to remain open, please add a comment. A closed issue may be reopened if requested.

Pulsar33:
@**pelwell** : Hello
I'm new in Raspberry World, starting directly with a Pi4 and Raspbian Buster. I see that this old thread started in 2016 and I don't see any solution coming. In several papers about Buster, the apparmor availability is said to be [one of the important improvements](https://www.debian.org/releases/buster/amd64/release-notes/ch-whats-new.html#apparmor) (and I agree). But apparmor doesn't work (on Pi4 boards) and can't be activated (kernel settings missing).
So, [my bug report](https://github.com/raspberrypi/linux/issues/3196) is closed. Can you tell me more about this noncompliance of the kernel regarding the Buster specification ?
Best regards. Pulsar33

jcberthon:
Running a Debian based OS on a Raspberry Pi with AppArmor is of course working, if you compile your own kernel.

I do think that Raspbian should support AppArmor and I did not see a memory impact. Although I only have Raspberry 2 and 3 so with 1GB this is not really constraint. And with the Raspberry Pi 4 out, this is even less of a concern.

I would recommend to have AppArmor compiled in the Kernel. This is a good security improvement.

jcberthon:
If there is interest in having AppArmor, I can revive this PR and push another set of commit to just activate that.

Pulsar33:
@**jcberthon** : thank you for your answer !
IMHO, when you want to try a software that is new or poorly controlled by the community, you should limit it in one of two ways:
- either it has to access sensitive data, so it doesn't be able to access outside the machine
- either it has to access outside the machine so it doesn't be able to access sensitive data

It seems to me that apparmor allows these two strategies (among many others) to be implemented easily with few drawbacks. 

So yes, if you could push a set of commit for apparmor (which is promoted by Debian) it would be a great thing for me.
Best regards. Pulsar33

popcornmix:
@jcberthon I think last time this was looked into (years back) the user<->kernel cost was the main objection.
Running a benchmark like
```
sudo dd if=/dev/zero of=/dev/null bs=1 count=1M
```
before and after the change would be useful information to include with a PR.
It is difficult to justify a small performance cost that affects every user when the users wanting the change are a very tiny percentage. But if the overhead is low enough then it is easier to add.

combs:
> @jcberthon I think last time this was looked into (years back) the user<->kernel cost was the main objection. 
> It is difficult to justify a small performance cost that affects every user when the users wanting the change are a very tiny percentage. But if the overhead is low enough then it is easier to add.

@popcornmix  Further up we have a comment from @djtm with a benchmark, https://github.com/raspberrypi/linux/pull/1698#issuecomment-263079299

> I can confirm there is no performance hit. I've done some further benchmarking, including a kernel with strong stack protection as well:
> http://openbenchmarking.org/result/1611266-SO-APPARMORS15

popcornmix:
> @popcornmix Further up we have a comment from @djtm with a benchmark, #1698 (comment)

The fact there exists a benchmark without a performance hit doesn't mean no other benchmark is affected. This change will only affect benchmarks that interact with kernel, so your FFTs/matrix ops etc are not expected to be affected.

The one linked that is relevant is the TCP loopback test goes 7% slower.
I suspect the dd test will be similar.

Whether that slowdown will be significant in a system test (e.g. actually copying data from network to usb drive) is not clear, but any measurable performance hit needs to be weighed up carefully.

Pulsar33:
Thank you all to have a look to my request.

According to Ubuntu Wiki, users should be able to choose, using this :
```
sudo systemctl stop apparmor.service
sudo update-rc.d -f apparmor remove
```
or that :
```
sudo systemctl start apparmor.service
sudo update-rc.d apparmor defaults
```
The problem is that compiling one's own kernel is the only solution for Raspbian.
This is not so simple as it should be !
So when you speak of performance cost, does it exist even if apparmor is disabled ?

Best regards. Pulsar33

popcornmix:
> So when you speak of performance cost, does it exist even if apparmor is disabled ?

Yes I believe it does.

I have no problem with a performance cost that applies to something that is disabled by people who aren't using it. But I believe this also has a cost to people who aren't using it.

Pulsar33:
Hello,
Can you explain a little more please. I don't understand why and how.
BR. Pulsar33

pelwell:
Security systems like AppArmor work like a firewall by intercepting accesses and deciding whether or not they are permitted. Just as a firewall needs to examine packets, adding some latency, the kernel takes some time to perform its audit checks. Even when the feature is disabled at runtime, the kernel code still needs to check on every access whether filtering is enabled, and the increased use of function pointers etc. will change the kinds of optimisations the compiler can do, so you should be able to see how enabling the possibility of turning on the feature without actually doing so has an overhead.

You could imagine an alternative implementation where, rather than decide per access to filter or not, disabling the filtering would switch to a different version of the kernel at a coarser granularity - like having two alternate versions of a subsystem loaded simultaneously with only one active at a time - but this isn't how it works; in Linux, if you don't want to use a feature and you don't want any of the associated overhead, you don't enable it in the kernel build.

Pulsar33:
> Even when the feature is disabled at runtime, the kernel code still needs to check on every access whether filtering is enabled

Ok, thanks for the explanation. It seems to me that a logical test on a flag is not so heavy for our processors today. Yes indeed, it depends on how many times you do it, but anyway ... 
Did someone quantify this overhead (with apparmor disabled) ?

pelwell:
The overhead is going to depend on how invasive the changes are, and how often those bits of code are executed. As @popcornmix said:
> The one linked that is relevant is the TCP loopback test goes 7% slower. I suspect the dd test will be similar.

JamesH65:
> Did someone quantify this overhead (with apparmor disabled) ?

This is what we would need someone* to do prior to enabling it by default. We would need before and after benchmarks, along with with it turned on and off on a kernel with it enabled. The TCP loopback test mentioned by @popcornmix would probably be a good start. And for the correct coverage, it would also be necessary to know the impact on something like the Pi Zero as well as the more powerful models.

*Not us

djtm:
AppArmor as well as SELINUX, do not affect the system unless active and configured working for certain processes: https://unix.stackexchange.com/questions/73515/does-apparmor-decrease-the-system-performance 
Even an active default broad SELinux setup has minimal impact: https://www.phoronix.com/scan.php?page=news_item&px=Fedora-23-SELinux-Impact
So to be super extra careful, you could complile in apparmor and disable apparmor by default via cmdline. 

Also see: 
* https://www.usenix.org/legacy/publications/library/proceedings/sec02/full_papers/wright/wright_html/node21.html
* https://www.usenix.org/legacy/publications/library/proceedings/sec02/full_papers/wright/wright_html/node20.html

popcornmix:
@djtm if that's the case build the kernel with the options you want enabled.
Run the "dd" test, and report the numbers with existing and new kernel.
(with apparmor enabled/disabled, although the number we care about is the "out-of-the-box" setting that normal users will end up with when updating their kernel).

djtm:
Can someone provide this build for testing? I've never created one.

popcornmix:
https://www.raspberrypi.org/documentation/linux/kernel/building.md

Pulsar33:
Thanks for the link. I was not sure it was easy but work is in progress on Pi4 board (crossing the fingers) without modification to be able to compare the generated kernel against the original one.
Wait and see ...

Pulsar33:
Well, it's done. I wasn't able to know if the kernel compiled with "no change" was really correct because there was several differences between it and the official one. However, the board started normally. Then, I tried to use the makeconfig tool but wasn't able to identify what to change with it.
So I added the following defines according to [this post](https://github.com/raspberrypi/linux/issues/1555#issuecomment-260267726). Many thanks to @**zfyseu** and @**combs** as the board boots correctly and apparmor is active !
```
sudo aa-status
apparmor module is loaded.
12 profiles are loaded.
10 profiles are in enforce mode.
   /usr/bin/man
   /usr/lib/arm-linux-gnueabihf/lightdm/lightdm-guest-session
   /usr/lib/arm-linux-gnueabihf/lightdm/lightdm-guest-session//chromium
   libreoffice-senddoc
   libreoffice-soffice//gpg
   libreoffice-xpdfimport
   man_filter
   man_groff
   nvidia_modprobe
   nvidia_modprobe//kmod
2 profiles are in complain mode.
   libreoffice-oopslash
   libreoffice-soffice
0 processes have profiles defined.
0 processes are in enforce mode.
0 processes are in complain mode.
0 processes are unconfined but have a profile defined.
```
Here are the modifications (for Pi 4)
Edit "/linux/arch/arm/configs/bcm2711_defconfig" and add :
```
CONFIG_KEYS=y
CONFIG_SECURITY=y
CONFIG_SECURITYFS=y
CONFIG_SECURITY_NETWORK=y
CONFIG_SECURITY_PATH=y
CONFIG_SECURITY_APPARMOR=y
CONFIG_SECURITY_APPARMOR_BOOTPARAM_VALUE=1
CONFIG_SECURITY_APPARMOR_HASH=y
CONFIG_INTEGRITY=y
CONFIG_INTEGRITY_AUDIT=y
```
Then do : make bcm2711_defconfig

What's next ?
**How can I run the desired tests and report the results ?**

Good night
Pulsar33

jcberthon:
Hi @pelwell @popcornmix 

**I will refresh today that PR**. I will also make the test, probably on a Raspberry Pi 3B+ board with the vanilla Raspbian kernel (so no AppArmor and based on Kernel 4.19), and compile the same kernel release with AppArmor. I will then perform the `dd` benchmark with the vanilla, and with the AppArmor"-ed" kernel, but with the latter I will run it twice, once disabled and once enabled.

With this we will see if AppArmor has or not an impact when doing a lot of transition from user to kernel space. But you should notice though that this is a very peculiar benchmark and for instance (although this is not at all representative I know) in my usage of several Raspberry Pis at my home from network server (DNS, DHCP, NTP, printer, etc.), another as a IoT central server (running OpenHAB and Home Assistant and some bespoke software) and another as a IoT "sensor" (using various sensors such as temp, humidity, pressure, etc. and sniffing the WiFi network for intrusion) there is no visible performance impact on using AppArmor. I'm not saying there are no impact, but if they are my 2 RPi 2 and my RPi 3 are fast enough anyway that I don't see problems running those work loads.

A side note to @Pulsar33, AppArmor is not the perfect security tool. In your example where you want to try a software from the community and hope that AppArmor will protect you, I can tell you already it won't. In order for AppArmor to protect you from a rogue application, it requires to have a "profile" of that application behaviour. If it has not no profile, then the application can do what it wants, AppArmor will not forbid it. So basically AppArmor has a list of application it is monitoring, the others are not. And for instance on Ubuntu, an application as central nowadays as a web browser has no profile active, in that case that is the Firefox profile. So having AppArmor is good, espcially when you write your own profiles for new application you install. But this is far from being a silver bullet.

pelwell:
Thanks @jcberthon.

Without wishing to prejudge the outcome of the tests, it might be that there is a small but measurable overhead both in terms of RAM and CPU load even with AppArmor disabled at runtime. If that is the case, would anybody here feel strongly if AppArmor support were included in v7 and v8 kernels (2B/3B/3B+/3A+/4B/CM3 - devices with  more RAM and higher clock speeds) but not in the v6 kernel (A/B/A+/B+/0/0W/CM)?

Aranjedeath:
> 
> 
> Thanks @jcberthon.
> 
> Without wishing to prejudge the outcome of the tests, it might be that there is a small but measurable overhead both in terms of RAM and CPU load even with AppArmor disabled at runtime. If that is the case, would anybody here feel strongly if AppArmor support were included in v7 and v8 kernels (2B/3B/3B+/3A+/4B/CM3 - devices with more RAM and higher clock speeds) but not in the v6 kernel (A/B/A+/B+/0/0W/CM)?

Unless the numbers (overhead on CPU -- I don't think a few kb extra ram usage matters here) are horrible, it should probably be turned on everywhere? This thread (the original bug report, not this PR) was originally started with only the lower powered devices. That's where we want the security features.

djtm:
I agree with @Aranjedeath: Install and activate Apparmor by default on all systems. This is what everyone does: Debian, Ubuntu, Fedora, and even Android (selinux) - included and active by default. Apparently the performance is good. What amount of impact would even justify not activating default security for Raspbian and leaving users more exposed? Shouldn't performance issues rather be fixed and apparmor be manually disabled as needed? That's way easier than compiling and installing a new custom kernel at least every few weeks.

There could be a special kernel flavor kernel flavor, e.g. `-lightweight`,  for minimum resource use. Or there could be an extra kernel flavor, e.g. `-secure`, with higher security settings for users. Because asking users to compile their own kernel is not a realistic option for almost all users. Compiling it _once_ is one thing, but compiling it again and again regularly, that's an not manageable especially where you need security.

JamesH65:
@djtm. The reason all the big distros can enable this by default is because they are running on very fast x86 systems with a lot of memory, so any performance impact is relatively tiny.  Run it on a PiZero pr PI1 with 512MB of RAM and any performance impact can be relatively high in comparison. If that impact is low enough (which is all we are asking people to determine) then it can be enabled by default. But until we know that it's not going to crucify Pi0 etc performance, it's prudent to leave it disabled.

Pulsar33:
@**pelwell** said  :
> ... would anybody here feel strongly if AppArmor support were included in v7 and v8 kernels (2B/3B/3B+/3A+/4B/CM3 - devices with more RAM and higher clock speeds) but not in the v6 kernel (A/B/A+/B+/0/0W/CM)?

If the drawback is too high for small boards, we can do that, no ? 

@**djtm**  said : 
> There could be a special kernel flavor kernel flavor, e.g. `-lightweight`, for minimum resource use

This is a good idea too !

BR. Pulsar33

jcberthon:
Sorry, I could not complete the benchmark tests due to a small technical problem: `/boot` is full on my raspberry ;-)

That will need to wait for another day... and that day wasn't this past weekend.

*Update 2019-09-12: I wanted to do that this evening, the resizing and moving of the main ext4 partition worked like a charm. But for the FAT I cannot resize it, I need to backup, remove and recreate it, but I'm unsure that it will work nicely... So We will need to figure out another evening in the coming weeks...*

Pulsar33:
Hello,
@**jcberthon** said : 
> A side note to @Pulsar33, AppArmor is not the perfect security tool. In your example where you want to try a software from the community and hope that AppArmor will protect you, I can tell you already it won't.

Well I know that but there's no "**application level** firewall" for linux (only old unachieved projects) and apparmor or firejail are the only safety means I know able to act on an application by application basis. Unfortunately, Firejail is more easier to use but less stable and generic. So apparmor remains the only choice for me.
That's why I limit my wishes to the two cases mentioned above when I create my own profiles to test an application :

- either the application has to access sensitive data, 
so it shouldn't be able to access outside the machine
- or the application has to access outside the machine 
so it shouldn't be able to access sensitive data

For a new application, you can create a profile starting from an existing one and adding simple drastic rules to close data or net access.

On Pi4 board with Raspbian, I was able to enforce firefox-esr, after some editing on a firefox profile, and adding a rule to deny access to /media/**/. I didn't succeed enforcing Chromium at this time, having not found a good working official profile (strange isn't it ?)

I'm currently trying to enforce some applications (Image Magic, curl ...) starting from scratch, just in order to test what I say above ...

Have a nice day
Pulsar33

Pulsar33:
Hello
Were you able to run the tests ?
Can I help ?
Have a nice day
Pulsar33

jcberthon:
Hi @Pulsar33 and @pelwell 

I finally had some time to perform a few benchmark. I could not use my Raspberry Pi 3 because the `/boot/` partition was too small and I could not find an open source utility to resize the FAT32 partition successfully, all my attempts failed...

So here are the results:

# Platform

Raspberry Pi 2 Model B Rev 1.1 (CPU BCM2836)

Disk size of the kernels:

```
-rwxr-xr-x 1 root root 5450036 Oct  4 15:51 /boot/kernel7-4.19.75-aa.img
-rwxr-xr-x 1 root root 5324164 Oct  4 13:46 /boot/kernel7-4.19.75.img
```

# TL;DR;

I have built 2 kernels:
1. Raspberry Pi 4.19.75 "vanilla": no changed options. I build it so that I was sure to be using the same baseline for comparison.
2. Raspberry Pi 4.19.75 "AppArmor": I have then activated the AppArmor subsystem, but let the default Unix Discretionary Access Control. So one would need to explicitly boot with AppArmor kernel parameters to enable it.

I tested with both kernels. I did not try with the second kernel and AppArmor enabled because I lack time and we just want to see if there is performance regression and which amount.

The facts are: The `dd` performance test for 1B and 4KB is impacted (ca. 10% slower) when using the Kernel with AppArmor built-in but disabled. Similarly, the `sysbench` test when using the threads benchmark profile shows ca. 3% slower performance. For all other tests, there are no differences (`dd` with 1MB, `sysbench` for fileio or cpu profiles).

Conclusions: 

This is tested on a single platform. I have not conducted test on older or newer Raspberry Pis.

In my opinion, the performance impact weighted to the security gain is fully acceptable. In most cases users won't see a difference, no more memory used, no performance impact. In the unlikelyhood that an application performs constantly very tiny context switches between user and kernel space at a very high rate (the 1B and 4KB tests with `dd`) one could see 10% impact which is also acceptable. Anyway, this is my opinion that such a micro-benchmark is not representative. I have conducted it as requested and I bring the results. Now you can choose.

# Kernel 4.19.75 Vanilla

## Memory usage

```
$ dmesg | grep Memory
[    0.000000] Memory policy: Data cache writealloc
[    0.000000] Memory: 983752K/1015808K available (8192K kernel code, 629K rwdata, 2188K rodata, 1024K init, 821K bss, 23864K reserved, 8192K cma-reserved)
```

## `dd` benchmark

### "1 Byte" test

Test: `dd if=/dev/zero of=/dev/null bs=1 count=10M`

Results: **442 kB/s**

### "4 kByte" test

Test: `dd if=/dev/zero of=/dev/null bs=1 count=10M`

Results: **845 MB/s**

### "1 MByte" test

Test: `dd if=/dev/zero of=/dev/null bs=1 count=10M`

Results: **1004 MB/s**

## `sysbench` generic benchmark

### Threads

```
$ sysbench --num-threads=4 --test=threads run
sysbench 0.4.12:  multi-threaded system evaluation benchmark
Test execution summary:
    total time:                          5.6036s
    total number of events:              10000
    total time taken by event execution: 22.3988
    per-request statistics:
         min:                                  2.12ms
         avg:                                  2.24ms
         max:                                 31.90ms
         approx.  95 percentile:               2.26ms

Threads fairness:
    events (avg/stddev):           2500.0000/3.67
    execution time (avg/stddev):   5.5997/0.00
```

### CPU

```
$ sysbench --num-threads=4 --test=cpu run
sysbench 0.4.12:  multi-threaded system evaluation benchmark
Test execution summary:
    total time:                          55.3431s
    total number of events:              10000
    total time taken by event execution: 221.3106
    per-request statistics:
         min:                                 21.67ms
         avg:                                 22.13ms
         max:                                 72.65ms
         approx.  95 percentile:              23.03ms

Threads fairness:
    events (avg/stddev):           2500.0000/18.99
    execution time (avg/stddev):   55.3277/0.01
```

### File IO

```
$ sysbench --num-threads=4 --test=fileio --file-test-mode=seqwr run
sysbench 0.4.12:  multi-threaded system evaluation benchmark
Operations performed:  0 Read, 131072 Write, 128 Other = 131200 Total
Read 0b  Written 2Gb  Total transferred 2Gb  (10.274Mb/sec)
  657.56 Requests/sec executed

Test execution summary:
    total time:                          199.3296s
    total number of events:              131072
    total time taken by event execution: 753.9991
    per-request statistics:
         min:                                  0.11ms
         avg:                                  5.75ms
         max:                              16572.44ms
         approx.  95 percentile:               1.03ms

Threads fairness:
    events (avg/stddev):           32768.0000/539.87
    execution time (avg/stddev):   188.4998/0.00
```


# Kernel 4.19.75 AppArmor builded but disabled

Kernel boot message:

```
[    0.001495] AppArmor: AppArmor disabled by boot time parameter
```

## Memory usage

```
$ dmesg | grep Memory
[    0.000000] Memory policy: Data cache writealloc
[    0.000000] Memory: 983744K/1015808K available (8192K kernel code, 635K rwdata, 2232K rodata, 1024K init, 823K bss, 23872K reserved, 8192K cma-reserved)
```

The memory went down from 983752K to 983744K, so 8K less.

## `dd` benchmark

### "1 Byte" test

Test: `dd if=/dev/zero of=/dev/null bs=1 count=10M`

Results: **382 kB/s**

Impact: 13,6% slower

### "4 kByte" test

Test: `dd if=/dev/zero of=/dev/null bs=1 count=10M`

Results: **768 MB/s**

Impact: 9,1% slower

### "1 MByte" test

Test: `dd if=/dev/zero of=/dev/null bs=1 count=10M`

Results: **1001 MB/s**

Impact: 0,3% slower

## `sysbench` generic benchmark

### Threads

```
$ sysbench --num-threads=4 --test=threads run
sysbench 0.4.12:  multi-threaded system evaluation benchmark
Test execution summary:
    total time:                          5.7887s
    total number of events:              10000
    total time taken by event execution: 23.1307
    per-request statistics:
         min:                                  2.15ms
         avg:                                  2.31ms
         max:                                 26.20ms
         approx.  95 percentile:               2.29ms

Threads fairness:
    events (avg/stddev):           2500.0000/45.56
    execution time (avg/stddev):   5.7827/0.00
```

Impact (avg per/req): 3,1% slower  
Impact (avg exec time): 3,3% slower

### CPU

```
$ sysbench --num-threads=4 --test=cpu run
sysbench 0.4.12:  multi-threaded system evaluation benchmark
Test execution summary:
    total time:                          55.1544s
    total number of events:              10000
    total time taken by event execution: 220.5827
    per-request statistics:
         min:                                 21.67ms
         avg:                                 22.06ms
         max:                                 71.87ms
         approx.  95 percentile:              22.92ms

Threads fairness:
    events (avg/stddev):           2500.0000/23.77
    execution time (avg/stddev):   55.1457/0.00
```

Impact (avg per/req): 0,3% faster, so equal (noise)  
Impact (avg exec time): 0,3% faster, so equal (noise)

### File IO

```
$ sysbench --num-threads=4 --test=fileio --file-test-mode=seqwr run
sysbench 0.4.12:  multi-threaded system evaluation benchmark
Operations performed:  0 Read, 131072 Write, 128 Other = 131200 Total
Read 0b  Written 2Gb  Total transferred 2Gb  (10.413Mb/sec)
  666.43 Requests/sec executed

Test execution summary:
    total time:                          196.6791s
    total number of events:              131072
    total time taken by event execution: 755.6340
    per-request statistics:
         min:                                  0.11ms
         avg:                                  5.77ms
         max:                               7890.45ms
         approx.  95 percentile:               0.85ms

Threads fairness:
    events (avg/stddev):           32768.0000/762.99
    execution time (avg/stddev):   188.9085/0.00
```

Impact (req/s executed): 1,3% faster  
Impact (avg per/req): 0,3% slower, so equal (noise)  
Impact (avg exec time): 0,2% faster, so equal (noise)

jcberthon:
If you think the benchmark results are good, let me know. I can try to push a clean-up PR again.

Pulsar33:
@**jcberthon** : Many thanks for your work !
Not being an expert, I can't give a technical opinion, but it seems fine.
Hope all will be ok.
Best regards
Pulsar33

Pulsar33:
Hello
I am not used to the process of taking into account developments. What is the next step for this subject and under what timeline please?
BR
Pulsar33

Pulsar33:
@jcberthon  Hello
No more news since two months. Is there a problem ?
Best regards
Pulsar33

jcberthon:
Hi
I did not received any feedback so I guess this is on hold. And if the others are like me, they don’t have the time ;-)
I’ve to admit that I didn’t update my kernel on my Pi since I build it 2 months ago… bad practices I know but between work and my private life, I had no time touching my computer! And I don’t see that improving for some months. So do not expect much activity from me. 
Sorry :shrug:

Pulsar33:
@jcberthon Thank you for the answer and for all the job you've already done !
Hope someone will validate your work (I'm not able to do that myself, sorry)
Best regards
Pulsar33

Pulsar33:
@djtm @popcornmix @pelwell @pyllyukko Hello
Can some of you validate the job of @jcberthon so that we can use this capability please ?
Merry Christmas to everybody
Pulsar33

pelwell:
@popcornmix I'm of a mind to merge this (after squashing etc.), with the option to revert it if there are complaints.

pelwell:
We're going to hold off until the next stable release for Raspbian is made, but these options are likely to be available in a test kernel relatively soon (perhaps using the 5.4 tree as a test area).

Pulsar33:
Happy new year everybody !
Thank you @pelwell for your actions.
Best regards
Pulsar33

ghost (approved):

patrusso2:
Did this make it into Raspberry Pi OS?

Pulsar33:
> Did this make it into Raspberry Pi OS?

Hello,
I've just installed the last Raspios (previously Raspbian) available version and apparmor is still "unloaded".
Version 2020-05-27 : Linux kernel 4.19.118

Can someone tell why ?
Thank you.
Pulsar33

JamesH65:
It might be in 5.4, apt full-upgrade should get it, you can then check.

Pulsar33:
> It might be in 5.4, apt full-upgrade should get it, you can then check.

Thank you for the reply but ...
`pi@Polites:~ $ uname -a
 Linux Polites 5.4.51-v7l+ #1326 SMP Fri Jul 17 10:51:18 BST 2020 armv7l GNU/Linux
 pi@Polites:~ $ sudo aa-status
 apparmor module is not loaded.
`

Should I modify something ?
BR
Pulsar33

JamesH65:
OK, so its not in 5.4 then, it was just a guess that it might have been included.

chrishamm:
Can we expect any update on this? Our company is developing a plugin interface that will be operating on the Raspberry Pi and AppArmor would be ideal to limit executable permissions and to guarantee a certain level of security.

Please consider enabling at least the compilation of the AppArmor module in the default configuration, that way we don't need to provide our own custom kernel.

pelwell:
Things have moved on in 5.4 - some of the proposed changes are now part of the current defconfigs, and once configured in, apparmor is now enabled by default. There is a command line parameter to disable it, but that's the wrong way round for now.

If somebody proposes a configuration that makes apparmor an opt-in feature but otherwise leaves security as it was then it's likely to be merged fairly quickly.

chrishamm:
According to the [Arch wiki](https://wiki.archlinux.org/index.php/AppArmor#Installation) the AppArmor module has to be enabled using two kernel command line parameters. To leave AppArmor disabled by default, we can set `CONFIG_SECURITY_APPARMOR_BOOTPARAM_VALUE` and `CONFIG_LSM` (currently `0` and `"lockdown,yama,loadpin,safesetid,integrity"`) so I see no reason why support could not be built into the kernel.

It would be really great to see optional AppArmor support in the RPi kernel sometime soon!

pelwell:
That information doesn't apply to Linux 5.4, hence my comment.

chrishamm:
Well that's true but if you enable the AppArmor in the config, it isn't enabled by default when the kernel boots unless you explicitly activate it. I just built my own RPi kernel and I could confirm that:

```
root@raspberrypi:/home/pi# uname -a
Linux raspberrypi 5.4.58-v7-apparmor+ #2 SMP Sun Aug 23 00:15:28 CEST 2020 armv7l GNU/Linux
root@raspberrypi:/home/pi# cat /sys/module/apparmor/parameters/enabled
N
root@raspberrypi:/home/pi# cat /sys/kernel/security/lsm
capability
```

Instead of CONFIG_SECURITY_APPARMOR_BOOTPARAM_VALUE we have CONFIG_DEFAULT_SECURITY_APPARMOR which is unset unless explicitly enabled. For completeness here my kernel config: [config.txt](https://github.com/raspberrypi/linux/files/5113169/config.txt)

Of course I was able to enable AppArmor via the command line parameters `apparmor=1 lsm=apparmor` after installing the apparmor package. Only after that step the Linux module was enabled and a new `apparmor` directory showed up in `/sys/kernel/security`.

pelwell:
I'm sure some of you thought the day would never come, but https://github.com/raspberrypi/linux/commit/a3e07c5dfef2664fc1d900ebb4a3f91c04ab700b adds AppArmor support and the extra process/container controls. It will be in future kernel builds, and we'll be watching out for regressions.

popcornmix:
This is in latest rpi-update kernel.

Pulsar33:
Hello,
Thank you for your work however it seems not working for me : 
RPI 4B 4Gb kernel 5.4.51-v7l recompiled with apparmor which is active.
apt full-upgrade
Reboot => uname -a shows kernel 5.4.58-v7l+ but apparmor module is not loaded.
As suggested above, apparmor=1 lsm=apparmor added in cmdline.txt
Reboot => uname -a shows kernel 5.4.51-v7l back and apparmor module is not loaded
Currently restoring a backup with custom 5.4.51-v7l and apparmor activated

Did I miss something ?
BR
Pulsar33

pelwell:
You need 5.4.61 or later for it to include AppArmor. You can get it now by running `sudo rpi-update`, or build your own kernel, or wait for the apt package to be updated.

By the way, `apparmor=1` is not necessary.

Pulsar33:
Many thanks for this !
I'll test and report
BR
Pulsar33

bog-dan-ro:
It works okay for me.

Pulsar33:
Hello,
I'm not sure enough of my knowledge to do a [rpi-update](https://www.raspberrypi.org/documentation/raspbian/applications/rpi-update.md) so I tried to compile the latest kernel as I've already done. Here is what I have noticed :

**Starting with a custom kernel with apparmor ok**
```
 uname -a => kernel 5.4.58-v7l+	(custom kernel)
 sudo aa-status => 36 profiles are loaded
```
**Updating the system**
```
 sudo apt update
 sudo apt full-upgrade
 sudo reboot
 uname -a => kernel 5.4.51-v7l+
 sudo aa-status => apparmor module is not loaded.

```
**Compiling the last kernel**
```
 git clone --depth=1 https://github.com/raspberrypi/linux
 cd linux
 KERNEL=kernel7l
 make menuconfig => no change save exit
 make bcm2711_defconfig (without change)
 make -j4 zImage modules dtbs
 sudo make modules_install
 sudo cp arch/arm/boot/dts/*.dtb /boot/
 sudo cp arch/arm/boot/dts/overlays/*.dtb* /boot/overlays/
 sudo cp arch/arm/boot/dts/overlays/README /boot/overlays/
 sudo cp arch/arm/boot/zImage /boot/$KERNEL.img
 sudo reboot

```
**Checking the result**
```
 uname -a => 5.4.61-v7l+
 sudo aa-status => apparmor module is loaded. apparmor filesystem is not mounted.

```
Please what did I miss ?

Have a nice day
Pulsar33

pelwell:
I'm guessing you didn't add `lsm=apparmor` to `cmdline.txt`.

Pulsar33:
> I'm guessing you didn't add `lsm=apparmor` to `cmdline.txt`.

Nice ! Thank you very much.
I confirm it works for me now doing as said above and adding lsm=apparmor in /boot/cmdline.txt

Good job guys !
Best regards
Pulsar33

PR #1710, opened 2016-11-04
bcmrpi3_defconfig: Update config to be compatible with latest Kconfig…

ghost:
bcmrpi3_defconfig: Update config to be compatible with latest Kconfig changes.
This new config has been run through make savedefconfig so it looks a bit different.

NOTE: Both ARCH_BCM2835 and ARCH_BCM2709 are now set.


yuchenericwu2:
Awesome, thanks!

PR #1712, opened 2016-11-06
Make VCHIQ and HDMI work in arm64.

ghost:
This is a change to make VCHIQ and HDMI audio work with arm64.

All of the changes have been applied to linux-next at this point except for the dmac_map_area conversion to dma_map_sg and of course HDMI audio.

For now, I have an independent implementation of dma_map_sg for arm64 while arm32 uses the old dmac_map_area.  The reason why is that I would like dma_map_sg to get tested better and dma_map_sg has a small performance decrease with small messages.  The idea would be for the arm64 version to replace the arm32 version on both platforms in the future and that is what upstream is using.

For now, the broadcom character devices are still out of the build config until I can sort out what changes those drivers need. Also note that direct ioctls on vchiq are not completely working from 32 bit userland to a 64 bit kernel since it needs a compatibility layer.

I'm thinking that by getting a bigger change in downstream for now will increase interest in arm64 and possibly get others involved. But I can understand the idea of waiting until upstream is fixed before proceeding.

Thanks for your consideration.


pelwell:
With these commits I can boot to the Raspbian Pixel desktop. Initially I found USB wasn't working, but once I'd fixed my build script to pick up the right DTBs it was fine, with WiFi and Bluetooth functional.

We'd prefer to have the cherry-picked upstream commits, complete with the "Signed-off-by" annotations, but otherwise this looks good.

ghost:
I didn't do it that way because most of the upstream patches don't apply automatically due to subtle differences between upstream and downstream.   Also upstream is using drivers/staging and downstream is using drivers/misc.

Since I had to modify the paths and make other small changes to get the patches to apply, I didn't want to add the "Reviewed-by" and other tags because I didn't want to imply that the changes have been 100% reviewed. 

Since the implementations are a bit different, I was thinking for 4.8 downstream could continue to use the downstream version and once upstream makes it into the official release just completely switch over.

 Let me know what you want me to do or change.

popcornmix:
Perhaps you should move to 4.9 where we do use drivers/staging for vchiq and are a lot closer to upstream?
Also I could cherry-pick commits into 4.9 before the usual pi specific patches (i.e. rebase onto them). When the commits reach the branch we are rebasing onto these cherry-picks will disappear.

ghost:
Perhaps that makes sense.  

Do you know when you will be refreshing 4.9 to be updated with all the downstream patches such as the latest Kconfig changes.

popcornmix:
As far as I know 4.9 is up to date with 4.8 for 32-bit kernel use.
For 64-bit kernel config I'll cherry pick commits if you report they work (or you can PR).
I'd prefer not to cherry-pick without some indication they are still good.

ghost:
I'm looking at the two branches through github and I can see that as of right now, the ARCH_MULTIPLATFORM and USB power changes are in 4.8 but not 4.9.   It seems 4.8 has a few changes that 4.9 does not have.

OK, I'm thinking that if you can cherry-pick from linux-next, then perhaps it is best to just close the PR.

popcornmix:
ARCH_MULTIPLATFORM and USB power changes are in 4.9 tree.
It has been rebased, which means our downstream USB power code never exists, but tthe end result is the same.

ghost:
OK, thanks.

PR #1716, opened 2016-11-11
Use i2c-bcm2835 as default

notro:
I have submitted some patches upstream (now in [linux-next](https://git.kernel.org/cgit/linux/kernel/git/next/linux-next.git/log/drivers/i2c/busses/i2c-bcm2835.c)) which tries to bring i2c-bcm2835 up to speed so we can start to use it. And the only way to know if it really can replace i2c-bcm2708 is to make it default, since "no one" will try it otherwise.

I have added a debug patch that hopefully can be of aid in tracking down the cause if some device stops working.

There's one thing missing here and that's a fallback to i2c-bcm2708. Initially I planned to make a i2c-bcm2708-overlay.dts, but we have i2c{0,1}-bcm2708-overlay.dts, so I'm not sure what to do.


pelwell:
For the overlay, use `target = <&i2c_arm>;`, and call it `i2c-bcm2708`. The other overlays, which allow different pins to be used, will be misnamed, but that isn't a big problem.

notro:
Added i2c-bcm2708 fallback overlay

arch/arm/boot/dts/overlays/README
```diff
@@ -520,6 +520,12 @@ Params: ds1307                  Select the DS1307 device
                                 source
 
 
+Name:   i2c-bcm2708
```
pelwell: As with the Makefile, this item needs to move before i2c-gpio.

arch/arm/boot/dts/overlays/Makefile
```diff
@@ -41,6 +41,7 @@ dtbo-$(RPI_DT_OVERLAYS) += i2c-gpio.dtbo
 dtbo-$(RPI_DT_OVERLAYS) += i2c-mux.dtbo
 dtbo-$(RPI_DT_OVERLAYS) += i2c-pwm-pca9685a.dtbo
 dtbo-$(RPI_DT_OVERLAYS) += i2c-rtc.dtbo
+dtbo-$(RPI_DT_OVERLAYS) += i2c-bcm2708.dtbo
```
pelwell: Sorry, notro, my script complains this entry should come before i2c-gpio.

pelwell (changes requested):
Two nitpicks.

notro:
Fixed

pelwell:
My alphabet looks a bit different - I've pushed some changes to be squashed.

notro:
Sorry, the hyphen messed up my head. Can you squash this when pulling/merging?

pelwell:
If I go through github we can only merge as-is or squash down to one. Since most of these are upstream patches, it's easier if we keep them separate. If we are in agreement to apply your patches, we can always squash and apply them manually.

popcornmix:
@pelwell okay to merge this?
There is a rebase to 4.8.7 coming so I can squash the commits then.

pelwell:
I've not had chance to test the new code, but my tests would be limited anyway and I trust @notro. Yes, no objections.

PR #1717, opened 2016-11-12
Drop ARCH_BCM2708 and ARCH_BCM2709

notro:
I have taken another step towards mainline by removing ARCH_BCM2708/ARCH_BCM2709 completely, including 2708/2709/2710 in the compatible property.
I haven't done arm64, because I couldn't get bcmrpi3_defconfig right.

With this PR mkknlimg will tag the kernel for 2835, so if tagged it currently doesn't work.

Tested on Pi1 and Pi2, _not_ Pi3.

It would have been nice if we could move to the mainline dts files as well: bcm283{5,6}-rpi-*.dts
The bootloader could (if not tagged), use 2835 dtb's if present, or fall back to 2708.

What do you think?


Electron752:
I understand that arm64 is not supported and on the bleeding edge.  I was just wondering what isn't working out with bcmrpi3_defconfig?

Are you also changing the label of the 2708 fb driver, or is that being dropped as well?

Electron752:
BTW, Is the reason you haven't tested on RPI 3 because you don't have one? I happen to have one if you need someone to run a quick test.

Also, I think all you need to do to get arm64 to work is delete
config ARCH_BCM2709
bool "Broadcom BCM2709 family"
select MACH_BCM2709
select HAVE_SMP
select ARM_AMBA
select COMMON_CLK
select ARCH_HAS_CPUFREQ
select GENERIC_CLOCKEVENTS
select MULTI_IRQ_HANDLER
select SPARSE_IRQ
select MFD_SYSCON
select VC4
select USE_OF
select ARCH_REQUIRE_GPIOLIB
select PINCTRL
select PINCTRL_BCM2835
help
This enables support for Broadcom BCM2709 boards.

Change to ARCH_BCM2835 in the defconfig.

Electron752:
Here is a possible suggestion.   Perhaps the BCM2708/BCM2709 dtb should be an overlay on top of the upstream device tree.  That way the boot loader can pull it in if needed.

BTW, the overlay support seems to be a bit broken on arm64.  Perhaps the boot loader could be fixed to make overlays work for arm64 as well?

ghost:
I combined my needed changes with your's and opened PR #1719.

The bcmrpi3_defconfig looks a bit different.  The file from my branch should just replace the existing bcmrpi3_defconfig.   I tested the combined changes on a RPI 3.

ghost:
It looks like these PRs have been broken by the recent rebase.   Since my changes are small and I can just submit as a separate PR, I'm going to just close my version.

PeterPablo:
You should be able to rebase your branch and force push the changes. This way you wouldn't need to close this PR and open a new one.

popcornmix:
Yes:

```
git pull --rebase
git push -f
```

should be all that is needed to fix a PR - no need to open a new one.

Electron752:
I'm starting to have doubts that perhaps this is the best way to go anyway.  I have a few ideas I'm kicking around.  

If you guys still want this PR, that's up to you.

Electron752:
For example, perhaps the compatible string for bcm2710 should have both the old and new model IDs.

compatible = **"brcm,bcm2710","brcm,bcm2709",** "brcm,bcm2837", "brcm,bcm2836";

drivers/usb/host/dwc_otg/dwc_otg_fiq_fsm.c
```diff
@@ -78,7 +78,7 @@ void notrace _fiq_print(enum fiq_debug_level dbg_lvl, volatile struct fiq_state
  * fiq_fsm_spin_lock() - ARMv6+ bare bones spinlock
  * Must be called with local interrupts and FIQ disabled.
  */
-#if defined(CONFIG_ARCH_BCM2709) && defined(CONFIG_SMP)
+#if defined(CONFIG_SMP)
```
pelwell: There are now people building our downstream tree with random configurations and emailing us about build errors, so I'd be happier if we kept this implementation with its inline assembler conditional on CONFIG_ARCH_BCM2836.

pelwell:
I've appended a commit that keeps mkknlimg working as it used to, i.e. detecting builds from the RPi tree as wanting the downstread bcm27xx DTBs.

notro:
There is no CONFIG_ARCH_BCM2836 so I used CONFIG_ARCH_BCM2835.
Pi1 and Pi2 are differentiated with ARCH_MULTI_V6 and ARCH_MULTI_V7.

pelwell:
One final thought, on something I should have spotted earlier - shouldn't bcm2836.dtsi set `model = "BCM2836";` for completeness?

notro:
The files that include it, overwrites the property, so that means patching an upstream file without actually changing anything. 
But I agree that it should have been set for completness as you say, since bcm283x.dtsi sets `model = "BCM2835";`

What's your thoughts on my suggestion to switch to mainline dts files?
One of the main benefits that I see, is that the bootloader will treat downstream and upstream kernels the same. No need for u-boot or config.txt changes.
Another is that we will do it some day, so why not do it now, bundled with all the other changes we have made in this iteration?
It would put us in a place where the only difference between upstream and downstream are the downstream drivers (I know there are more differences but they are related to drivers and performance).

pelwell:
> What's your thoughts on my suggestion to switch to mainline dts files?

Not unless we modify the upstream DTBs to include FIQ-enhanced USB, the sdhost driver, DT parameters and the labels (and perhaps a few others).

notro:
Yes of course, in that case we would patch the upstream files.

The dtsi files would be patched by applying the diff-files we have:
bcm283x.dtsi <<- bcm270x.dtsi
bcm2835.dtsi <<- bcm2708.dtsi
bcm2836.dtsi <<- bcm2709.dtsi
bcm2835-rpi.dtsi <<- bcm2708-rpi.dtsi

And the downstream dts files will in reality just change names and the upstream version dropped:
bcm2708-rpi-b-plus.dts -> bcm2835-rpi-b-plus.dts
etc.

pelwell:
That plan is logically equivalent to deleting the upstream files. How is that a step forwards?

Apart from all of the downstream-only code, the primary difference to upstream is now the choice of DTB. By retaining the upstream files we make the differences plain, and by keeping them unmodified we reduce the merge conflicts when switching branches.

notro:
All the PR's I have made was done with the purpose of closing the gap to mainline. To minimize the difference. DTB names are one obvious difference.
That being said, this issue is not important for me personally. The downside is ofc as you say patches and merge conflicts.
Since I'm not the one doing that, I thought I'd ask before actually implementing it in this PR.

Anyways, if this PR goes in, it will mean that we have fully moved to ARCH_BCM2835 which is what I have been working towards for some time now. At that's a milestone for me personally :-)

pelwell:
> Anyways, if this PR goes in, it will mean that we have fully moved to ARCH_BCM2835 which is what I have been working towards for some time now. At that's a milestone for me personally :-)

And for that, and all your contributions, you have our thanks.

I'm happy for use to merge this PR. @popcornmix?

kukabu:
@notro I'm not sure but I think this patch breaks some overlays when compatible string contains "brcm,bcm27XX" only.

pelwell:
Really? I'll let you into a secret - currently nothing is checking the top-level compatible string in overlays.

kukabu:
It's ok then

notro:
I have updated the [Upstreaming](https://github.com/raspberrypi/linux/wiki/Upstreaming) page.

notro:
I did a fresh 4.8 Pi1 build, and get this:

```
$ sudo vcdbg log msg
debug_sym: OpenVideoCoreMemoryFileWithOffset: newHandle->vcSymbolTableOffset (0x20 - 0x1ec00000) > 8MB

Unable to open videocore memory access: -5

$ dmesg
[    0.000000] OF: fdt:Machine model: Raspberry Pi Model B Plus Rev 1.2

[    0.650445] raspberrypi-firmware soc:firmware: Attached to firmware from 2016-11-16 12:24

[    1.856713] vc_mem_init: called
[    1.856731] vc-mem: phys_addr:0x00000000 mem_base=0x1ec00000 mem_size:0x20000000(512 MiB)

[   65.689449] vc_mem_open: called file = 0xd898d960
[   65.689511] vc_mem_ioctl: called file = 0xd898d960
[   65.689525] vc_mem_ioctl: VC_MEM_IOC_MEM_SIZE=536870912
[   65.689537] vc_mem_ioctl: file = 0xd898d960 returning 0
[   65.689547] vc_mem_ioctl: called file = 0xd898d960
[   65.689554] vc_mem_ioctl: VC_MEM_IOC_MEM_BASE=515899392
[   65.689561] vc_mem_ioctl: file = 0xd898d960 returning 0
[   65.689568] vc_mem_ioctl: called file = 0xd898d960
[   65.689574] vc_mem_ioctl: VC_MEM_IOC_MEM_LOAD=515899392
[   65.689581] vc_mem_ioctl: file = 0xd898d960 returning 0
[   65.689589] vc_mem_ioctl: called file = 0xd898d960
[   65.689598] vc_mem_ioctl: VC_MEM_IOC_MEM_PHYS_ADDR=0x  (null)
[   65.689606] vc_mem_ioctl: file = 0xd898d960 returning 0
[   65.690616] vc_mem_release: called file = 0xd898d960
```

notro:
vchiq ping test doesn't work either:

```
$ vchiq_test -p
Ping test - service:echo, iters:1000, version 3
^C
```

Works fine on Pi2.

popcornmix:
@notro you are correct. There is a bug in the code that affects Pi1 more than Pi2 (bonus points for explaining why Pi2 wasn't broken...)
I've pushed a fix. (for the vcdbg issue - that won't be affecting vchiq)

popcornmix:
@pelwell `vchiq_test -p` seems to hang on 4.4 kernel too (on Pi1). Any ideas?

notro:
Yes, vcdbg is working now.

> bonus points for explaining why Pi2 wasn't broken...

I suck at this low level memory stuff. Some L2 cache alias thing maybe :-)

struct device is missing here:

```
static long vc_mem_copy()

    buf = dma_alloc_coherent(NULL, PAGE_ALIGN(size), &bus_addr,
                 GFP_ATOMIC);
```

This means that bus_addr contains the ARM physical address instead of the bus address on rpi-4.8.y.

popcornmix:
Added.

pelwell:
I'm still trying to get to the bottom of the vchiq_test -p stall.  The main protocol handling looks OK, but something is going wrong within the client library.

pelwell:
Got it - there was a race between the VCHI callback thread and the main test thread. This race is much more likely to occur on an SoC with a single ARM core, i.e. a Pi1.

I've pushed a fix to the userland repo.

PR #1721, opened 2016-11-13
Remove BCMRPI3_DEFCONFIG(Add Raspberry PI 3 Platform)

ghost:
This is becoming too political for me at this point.

If RPI3(ARM64) is important to people, I can update the configuration to be compatible with whatever is decided.

Thanks.


Electron752:
Perhaps I should explain for a moment...

I was thinking that longer term, bcmrpi_defconfig could be the base configuration.   bcm2709_defconfig would be a link to it.  Inside the Kconfig, it could have a "Raspberry PI 2" platform that add the few items extra needed for the RPI 2.

It seems that it's taking an awful lot of work to manage these configuration files, so perhaps reducing the number of files would simplify things.

PR #1631 has some of the ideas.

Electron752:
Ideally, it would be interesting if one defconfig could build on top of the other defconfig.  This just gives a sample of what is possible.

PR #1727, opened 2016-11-20
Audioinjector Fix bit offsets for equal volume and add 8 kHz operation

flatmax:
Applying to the audioinjector sound card only. This patch offsets channel
2 correctly from the LR clock. This ensures that channel 2 doesn't loose
any bits during capture. It also results in both channels 1 and 2 having
the same volume. This commit also adds 8 kHz operation.

Signed-off-by: Matt Flax <flatmax@flatmax.org>

### DIFF
--- a/sound/soc/bcm/audioinjector-pi-soundcard.c
+++ b/sound/soc/bcm/audioinjector-pi-soundcard.c
@@ -30,7 +30,7 @@
 #include "../codecs/wm8731.h"
 
 static const unsigned int bcm2835_rates_12000000[] = {
-	32000, 44100, 48000, 96000, 88200,
+	8000, 16000, 32000, 44100, 48000, 96000, 88200,
 };
 
 static struct snd_pcm_hw_constraint_list bcm2835_constraints_12000000 = {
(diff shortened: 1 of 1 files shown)


sound/soc/bcm/audioinjector-pi-soundcard.c
```diff
@@ -55,19 +55,21 @@ static int snd_audioinjector_pi_soundcard_hw_params(struct snd_pcm_substream *su
 
 	switch (params_rate(params)){
 		case 8000:
-			return snd_soc_dai_set_bclk_ratio(cpu_dai, 1508);
+			return snd_soc_dai_set_bclk_ratio(cpu_dai, 1);
```
pelwell, line 58: Did you mean 1500?

flatmax:
Any number above 1024 is not possible, check [1] below.

For that reason, by setting it to a value of 1, it is possible to use 8 
kHz audio, but it comes as stereo with a mono replication on both 
channels. However, if you consider that 8 kHz audio is most frequently 
used for voice and microphone recordings, mono replicated over stereo is 
fine. In fact this commit is specifically targeting a request from Farm 
radio in Tanzania, to enable 8 kHz natively for their purpose, which 
requires mono input and output.

thanks

Matt

[1] bclk_ratio is related to Bits 13:4 from the RXC_A Register, that is 
10 bits, with a maximum possible of 1024 clock cycles.

obbardc:
@popcornmix @flatmax @pelwell Can this pull request be implemented in the 4.9 branch? I'm not sure on the procedure of how to do this.. Cheers!

pelwell:
I think @popcornmix can cherry-pick it. I would, but I don't have access at the moment.

popcornmix:
``sound/soc/bcm/audioinjector-pi-soundcard.c`` is identical on 4.4, 4.8 and 4.9 branches, so this PR is already present.

obbardc:
@popcornmix Whoops, next time I will make sure to double-check! Cheers.

PR #1731, opened 2016-11-26
PITFT

ghost:


scottellis:
These overlays are already in the rpi-4.4 tree

  pitft28-capacitive-overlay.dts
  pitft28-resistive-overlay.dts
  pitft35-resistive-overlay.dts

Commit 1535181b "add adafruit pitft overlays" looks like it is trying to add these same overlays with different names:

pitft28c-overlay.dts
pitft28r-overlay.dts
pitft35r-overlay.dts

pelwell:
As @scottellis says, there are already overlays to support those displays, and Adafruit's instructions use them. If you still think you have something to contribute, please open a new PR with a minimal set of commits and a valid description.

PR #1739, opened 2016-12-02
Add overlays symlink and dtbo tagert for ARM64

Gerharddc:
This patch symlinks the overlays directory into ARM64 and adds the dtbo target needed to build them to the ARm64 makefile.

Signed-off-by: Gerhard de Clercq <gerharddeclercq@outlook.com>

### DIFF
--- a/arch/arm64/Makefile
+++ b/arch/arm64/Makefile
@@ -102,6 +102,9 @@ zinstall install:
 
 %.dtb: scripts
 	$(Q)$(MAKE) $(build)=$(boot)/dts $(boot)/dts/$@
+	
+%.dtbo: | scripts
+	$(Q)$(MAKE) $(build)=$(boot)/dts $(boot)/dts/$@
 
 PHONY += dtbs dtbs_install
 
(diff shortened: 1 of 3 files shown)


pelwell:
I have a few problems with this commit:

1. I'm not wildly keen on the symlink, but it isn't the first and I haven't thought of a better way yet, so I'm not going to object for that reason.

2. I think the symlink belongs in the broadcom directory, since they are Broadcom platform-specific. This then has the problem that the dtbs_install put them into broadcom/overlays/, but we have to solve that problem for the base dtbs.

3. You still need to add overlays to subdirs-y, otherwise the directory is never visited.

I'm prepared to overlook 1 and 2 and treat this is a stepping stone, but 3 needs to be fixed.

Gerharddc:
I also don't like the symlink but yeah what else. I see that some of the dts files in the broadcom directory actually just include the normal files in the arm folder themselves but I find this even less satisfying... 

I agree that the overlays should be in the broadcom folder but this will be horrible for build systems until the problem of the broadcom folder is solved. Luckily my system can easily be made to fetch the base dtbs from there but not overlays.

I didn't even notice the subdris-y problem because my system builds the overlays manually. I'll quickly add  that.

pelwell:
Thanks - I've squashed and merged your commits.

Issue #1738, opened 2016-12-01
rpi-ft5406 64 bit


Gerharddc:
I am trying to get the RPI touch screen working on a 64 bit image and have already got VC4 and DSI working, I only need the touch driver. This is proving difficult because the `writel(99, &ts->regs->num_points);` line is causing a crash related to paging to an invalid address. I have concluded that this is caused by some invalid address translation because the driver was hard-coded for 32 bit addresses.

I have already tried replacing all the direct pointer dereferencing in the code with the explicit io functions but am still facing the same issue.

I have decided that the problem lies in:
```
rpi_firmware_property(fw, RPI_FIRMWARE_FRAMEBUFFER_GET_TOUCHBUF,
&touchbuf, sizeof(touchbuf));
...
touchbuf &= ~0xc0000000;
ts->ts_base = ioremap(touchbuf, sizeof(*ts->regs));
```
With touchbuf being a u32. 

I am more than willing to fix this driver myself, I just can't find any information on the firmware interface and therefore have no idea what type of address translation needs to take place. For instance, does it still provide a 32 bit address when running a 64 bit kernel and what is the reason for `touchbuf &= ~0xc0000000`? Basically I just need know how I can map the RPI_FIRMWARE_FRAMEBUFFER_GET_TOUCHBUF to a 64 bit virtual address.

I guess one also needs to do the right translation when suggesting to RPI_FIRMWARE_FRAMEBUFFER_SET_TOUCHBUF.

pelwell:
The firmware deals in bus addresses. It has no knowledge of whether the OS is 32- or 64-bit, and it doesn't care.

The VPU address space is limited to 1GB, with the top two bits indicating certain addressing modes. The ARM's kernel space is configured to start at bus address 0, so the mask is to remove those upper bits and leave you with a bus address ready to map to a suitable kernel virtual address.

Gerharddc:
@pelwell Ok so the pointer at ts->ts_base from ioremap at the moment should be perfectly valid for use in for example iowrite32?

EDIT:
Or rather I guess, is the value of touchbuf currently correct for passing to ioremap?

pelwell:
I don't know off the top of my head, and I'm not about to start digging into 64-bit kernel internals until we decide to support it.

Gerharddc:
@pelwell Thanks for the support, I don't want to waste too much of your time. The last thing though, my error is as follow:

```
[...]

PR #1740, opened 2016-12-02
RPI-FT5406: Enable aarch64 support through explicit iomem interface

Gerharddc:
Signed-off-by: Gerhard de Clercq <gerharddeclercq@outlook.com>

### DIFF
--- a/drivers/input/touchscreen/rpi-ft5406.c
+++ b/drivers/input/touchscreen/rpi-ft5406.c
@@ -20,6 +20,7 @@
 #include <linux/input/mt.h>
 #include <linux/kthread.h>
 #include <linux/platform_device.h>
+#include <linux/stddef.h>
 #include <asm/io.h>
 #include <linux/dma-mapping.h>
 #include <soc/bcm2835/raspberrypi-firmware.h>
@@ -47,7 +48,6 @@ struct ft5406 {
 	struct input_dev       * input_dev;
 	void __iomem           * ts_base;
 	dma_addr_t		 bus_addr;
-	struct ft5406_regs     * regs;
 	struct task_struct     * thread;
 };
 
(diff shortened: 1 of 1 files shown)


popcornmix:
Needs a rebase after 4.8.12
I can't really see what effect removing regs pointer would have.
Are you sure it doesn't work just with the ``iowrite8`` change?

pelwell:
I'd argue the code is simpler without regs, so why not delete it?

popcornmix:
I'm not against the change, but I'd like to know what in the patch is a bug fix and what is cosmetic.

Gerharddc:
@popcornmix Ok yeah the code is the result of a problem I had as discussed on https://github.com/raspberrypi/linux/issues/1738 with @pelwell . From what I have just seen, the old code seems to actually work with the very latest firmware and the changes were needed for slightly older firmware to cope. I do think this is better than using "regs" though as it then doesn't treat iomem as a regular pointer (which isn't strictly legal but works here with the right firmware it seems) and as @pelwell said it's simpler without that extra pointer. Also, I'm not sure if rebase is a problem, this code should slot like this into any branch really?

I guess with the new firmware then everything is cosmetic but taking into account that this works even with the old firmware I guess it is more "rugged" or something.

Gerharddc:
I don't know what happened to the pull request but I see there is now something like 200 commits in it? Maybe I should resubmit it and then also just submit it as being more cosmetic?

popcornmix:
@Gerharddc it needs rebasing. Run:
```
git pull --rebase upstream rpi-4.8.y
git push -f
```
(assuming raspberrypi/linux repo is called upstream) and the PR should look good again.

Gerharddc:
@popcornmix  O, now I get what you meant, sorry for the misunderstanding. It's been done now though.

Electron752:
Any reason why this PR can't be accepted?

I did something very similar back in the July/August to get the touchscreen to work so I think the change should be safe.  And I think it's relatively safe to say this is a downstream only driver for the future so it makes sense for it to be here.

pelwell:
I'm happy with this change, provided whoever merges it adds a '6' to the commit message.

popcornmix:
Thanks. Added to 4.8 and 4.9 trees.

PR #1744, opened 2016-12-03
4.9 DSI panel

Gerharddc:
I have ported all the code needed to get the DSI panel working on this branch. I did it specifically for the RPi3 on aarch64 so the device tree modifactions are aimed at it and there is a commit or two related to simply getting this branch working on aarch64. I also had to permanently enable "cprman" to keep the board from freezing even without overlays, this can be fixed in future. For some reason DSI only wants to work with fkms at the moment so therefore it hasn't been added to the kms overlay as well.


Ruffio:
@Gerharddc If think that if this is going to have any change of getting accepted/merged, then all the conflicts must be resolved to begin with...

anholt:
The 4.9 DRM backport that just landed should cover this.  The FKMS mode doesn't need any of this DSI code, so that's why it only worked there.

pelwell:
Closing as out-of-date and forgotten.

PR #1745, opened 2016-12-04
Added spi2-enc28j60.dts overlay file that works on SPI2 interface

yidingalan:
The current enc28j60.dts only works on SPI0 interface and does not support SPI2. 

The new overlay file works on SPI2 and is already tested


pelwell:
Your PR didn't add the overlay to the Makefile, so it would never be built. It also didn't follow the convention/rule that the source file ends "-overlay.dts"; without this, the make rules don't add the necessary flags to build an overlay.

I've update the Makefile for you, but also changed the overlay name so that the most important feature (that it is an overlay for the enc28j60) comes first. After squashing the changes down to a single commit the result has now been pushed to rpi-4.4.y.

PR #1752, opened 2016-12-07
BCM270X_DT: overlays/*.dtb -> overlays/*.dtbo

staroselskii:
We now use dtbo files in overlays.


pelwell:
Thanks for pointing this out - I've corrected your correction.

PR #1756, opened 2016-12-09
Enable regulator support

HiassofT:
Several audio codecs and (headphone-) amplifiers, especially those with builtin regulators, require voltage/current regulator support in the kernel.

This feature hadn't been enabled before because it had the potential to break drivers that didn't list the (often required) supply properties in their DT overlays.

Fortunately things have changed and the kernel attaches a dummy regulator in those cases for quite a while now. This is especially important for HATs where the DT overlay is loaded from EEPROM.

Using a HiFiBerry Digi Pro (HAT, with overlay in EEPROM) on a kernel with regulator support only gives 2 harmless warnings but otherwise the device works fine:
```
[    5.363666] 1-003b supply PVDD not found, using dummy regulator
[    5.363836] 1-003b supply DVDD not found, using dummy regulator
[    5.375495] wm8804 1-003b: revision E
[    5.386010] snd-hifiberry-digi soc:sound: wm8804-spdif <-> 3f203000.i2s mapping ok
```

I've mainly tested with 4.8 and 4.9 kernels, but a quick test shows that it would be safe to add this to 4.4 as well - the hifiberry card works fine on this kernel, too.


clivem:
This is re-visiting #1479?

HiassofT:
@clivem yes, basically

lategoodbye:
Regarding to this [discussion](https://www.spinics.net/lists/arm-kernel/msg309116.html) simple regulators do not belong to a simple bus.

HiassofT:
@lategoodbye thanks for your feedback.

I'm a little confused now: I followed the DT structure of the Tegra210 which had regulators added in a simple-bus a few months ago 1b4c8420225cfcc9e94ac62a9779d3de672e96d7 

Is there some definitive guide where off-soc / on-board nodes should be added?

lategoodbye:
I assume the patch for Tegra210 has never been forwarded to the DT folks. There are a lot of bad examples.

Unfortunately i don't know about a guide about this topic. The devicetree mailing list should be more reliable in such cases.

HiassofT:
Digging through quite some mails on the DT list the consensus seems to be that non-addressable on-board devices are best added as root nodes to the DT. That also makes sense topology-wise, having a bus with pseudo addresses for the regulators looked odd to me.

I've moved the regulator nodes to the DT root and changed the names to end in _reg instead of _sys - that seemed more logical to me.

lategoodbye:
Looks good to me. Thanks

HiassofT:
superseded by #1810

PR #1765, opened 2016-12-13
Add support for Silicon Labs Si7013/20/21 humidity/temperature sensor.

ajtucker:
The SI7021 is a humidity and temperature sensor that uses the I2C bus and has a driver in the kernel, si7020.  I have wired one up to my Pi and used this device overlay to enable it at boot time.

The driver provides an IIO interface via sysfs at /sys/bus/iio/devices/iio:device? where the relative humidity and temperature can be read from the in_temp* and in_humidity* files.  Note that the values must be computed by taking the raw value, adding the offset and multiplying by the scale factor and then dividing by 1000, e.g.

temperature in degrees C = (in_temp_raw + in_temp_offset) * in_temp_scale / 1000

### DIFF
--- a/arch/arm/boot/dts/overlays/README
+++ b/arch/arm/boot/dts/overlays/README
@@ -543,6 +543,9 @@ Params: bmp085                  Select the Bosch sensortronic BMP085
 
         bmp280                  Select the Bosch sensortronic BMP280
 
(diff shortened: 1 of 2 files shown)


pelwell:
Thanks - apologies for the delay.

PR #1770, opened 2016-12-22
Added 3-wire functionality to spidev_test.c

hennichodernich:
3-wire mode requires half-duplex operation, write first, then read, using seperate transfers. rx_buf and tx_buf pointer have to be NULLed, respectively, otherwise the transfer will fail.

Did anyone ever check 3-wire mode with a 3-wire capable driver (e.g., spi-bcm2835)?


pelwell:
This one slipped through the net - sorry. The file in question is an unmodified upstream file - so changes should be submitted upstream.

PR #1783, opened 2017-01-08
bcm2835-i2s: Changes for allowing asymmetric sample formats.

gtrainavicius:
Hey,

I've noticed that when overdubbing audio with Audacity using 16-bit sample format for recording, it still uses 32 bits for playback, causing corrupt audio to get recorded. The bcm2835-i2s forces format symmetry between the streams, but it needn't be the case. (I alternatively tried to set .symmetric_formats=1 in bcm2835_i2s_dai, but Audacity wasn't able to detect this, and failed to record completely which would be unacceptable.)

I've made some changes that allow having asymmetric sample formats, and tested various combinations of simultaneous playback and recording with the pisound card, however, I don't own any other audio cards, so I can't test the changes using them.

Anyway, I tried to keep the changes to minimum, the only cases these changes should affect is when the user software is requesting asymmetric sample format for playback and capture, in all other cases, it should behave the same as without the changes.

Thank you,
Giedrius.

### DIFF
--- a/sound/soc/bcm/bcm2835-i2s.c
+++ b/sound/soc/bcm/bcm2835-i2s.c
@@ -310,6 +310,7 @@ static int bcm2835_i2s_hw_params(struct snd_pcm_substream *substream,
 	unsigned int sampling_rate = params_rate(params);
 	unsigned int data_length, data_delay, bclk_ratio;
 	unsigned int ch1pos, ch2pos, mode, format;
+	unsigned int previous_ftxp, previous_frxp;
 	unsigned int mash = BCM2835_CLK_MASH_1;
 	unsigned int divi, divf, target_frequency;
 	int clk_src = -1;
@@ -320,6 +321,7 @@ static int bcm2835_i2s_hw_params(struct snd_pcm_substream *substream,
 	bool frame_master =	(master == SND_SOC_DAIFMT_CBS_CFS
 					|| master == SND_SOC_DAIFMT_CBM_CFS);
 	uint32_t csreg;
+	bool packed;
 
 	/*
 	 * If a stream is already enabled,
@@ -465,26 +467,46 @@ static int bcm2835_i2s_hw_params(struct snd_pcm_substream *substream,
 		return -EINVAL;
 	}
 
-	/*
-	 * Set format for both streams.
-	 * We cannot set another frame length
-	 * (and therefore word length) anyway,
-	 * so the format will be the same.
-	 */
-	regmap_write(dev->i2s_regmap, BCM2835_I2S_RXC_A_REG, format);
-	regmap_write(dev->i2s_regmap, BCM2835_I2S_TXC_A_REG, format);
+	/* Set the format for the matching stream direction. */
+	switch (substream->stream) {
+	case SNDRV_PCM_STREAM_PLAYBACK:
+		regmap_write(dev->i2s_regmap, BCM2835_I2S_TXC_A_REG, format);
+		break;
+	case SNDRV_PCM_STREAM_CAPTURE:
+		regmap_write(dev->i2s_regmap, BCM2835_I2S_RXC_A_REG, format);
+		break;
+	default:
+		return -EINVAL;
+	}
 
 	/* Setup the I2S mode */
+	/* Keep existing FTXP and FRXP values. */
+	regmap_read(dev->i2s_regmap, BCM2835_I2S_MODE_A_REG, &mode);
+
+	previous_ftxp = mode & BCM2835_I2S_FTXP;
+	previous_frxp = mode & BCM2835_I2S_FRXP;
+
 	mode = 0;
 
-	if (data_length <= 16) {
-		/*
-		 * Use frame packed mode (2 channels per 32 bit word)
-		 * We cannot set another frame length in the second stream
-		 * (and therefore word length) anyway,
-		 * so the format will be the same.
-		 */
-		mode |= BCM2835_I2S_FTXP | BCM2835_I2S_FRXP;
+	/*
+	 * Retain the frame packed mode (2 channels per 32 bit word)
+	 * of the other direction stream intact. The formats of each
+	 * direction can be different as long as the frame length is
+	 * shared for both.
+	 */
+	packed = data_length <= 16;
+
+	switch (substream->stream) {
+	case SNDRV_PCM_STREAM_PLAYBACK:
+		mode |= previous_frxp;
+		mode |= packed ? BCM2835_I2S_FTXP : 0;
+		break;
+	case SNDRV_PCM_STREAM_CAPTURE:
+		mode |= previous_ftxp;
+		mode |= packed ? BCM2835_I2S_FRXP : 0;
+		break;
+	default:
+		return -EINVAL;
 	}
 
 	mode |= BCM2835_I2S_FLEN(bclk_ratio - 1);


pelwell:
This change looks reasonable to me - it compiles, and I2S playback still works. @HiassofT @msperl Any comments?

msperl:
more testing and upstreaming maybe?

gtrainavicius:
Let me know if there's anything I can do to help :)

HiassofT:
I'll run some playback/recording tests with the cirrus logic audio card
and then report back - could take a few days (busy with work)

sound/soc/bcm/bcm2835-i2s.c
```diff
@@ -465,26 +467,46 @@ static int bcm2835_i2s_hw_params(struct snd_pcm_substream *substream,
...
+	default:
+		return -EINVAL;
+	}
 
 	/* Setup the I2S mode */
+	/* Keep existing FTXP and FRXP values. */
+	regmap_read(dev->i2s_regmap, BCM2835_I2S_MODE_A_REG, &mode);
+
+	previous_ftxp = mode & BCM2835_I2S_FTXP;
+	previous_frxp = mode & BCM2835_I2S_FRXP;
```
flatmax, line 487: If these previous variables aren't used, they should be removed.

sound/soc/bcm/bcm2835-i2s.c
```diff
@@ -465,26 +467,46 @@ static int bcm2835_i2s_hw_params(struct snd_pcm_substream *substream,
...
+	switch (substream->stream) {
+	case SNDRV_PCM_STREAM_PLAYBACK:
+		regmap_write(dev->i2s_regmap, BCM2835_I2S_TXC_A_REG, format);
+		break;
+	case SNDRV_PCM_STREAM_CAPTURE:
+		regmap_write(dev->i2s_regmap, BCM2835_I2S_RXC_A_REG, format);
+		break;
+	default:
+		return -EINVAL;
+	}
```
flatmax, line 480: The format is the same, why split ?

sound/soc/bcm/bcm2835-i2s.c
```diff
@@ -465,26 +467,46 @@ static int bcm2835_i2s_hw_params(struct snd_pcm_substream *substream,
...
 
 	/* Setup the I2S mode */
+	/* Keep existing FTXP and FRXP values. */
+	regmap_read(dev->i2s_regmap, BCM2835_I2S_MODE_A_REG, &mode);
```
flatmax, line 484: The implications of restoring the mode are quite large and risky! Why not let the file set it up from scratch ?
All the lines which treat mode in the file do "mode |=" to alter mode. They all assume that mode starts as : mode = 0;

flatmax (commented):
There are unused variables in this patch. There is a risky introduction of restoring mode, rather then starting with it =0.

sound/soc/bcm/bcm2835-i2s.c
```diff
@@ -465,26 +467,46 @@ static int bcm2835_i2s_hw_params(struct snd_pcm_substream *substream,
...
+	 * Retain the frame packed mode (2 channels per 32 bit word)
+	 * of the other direction stream intact. The formats of each
+	 * direction can be different as long as the frame length is
+	 * shared for both.
+	 */
+	packed = data_length <= 16;
+
+	switch (substream->stream) {
+	case SNDRV_PCM_STREAM_PLAYBACK:
+		mode |= previous_frxp;
```
gtrainavicius, line 501: This is the location where previous_... variables are used.

sound/soc/bcm/bcm2835-i2s.c
gtrainavicius: They are used, see below.

sound/soc/bcm/bcm2835-i2s.c
gtrainavicius: Only the stream direction specific state is kept, not the whole mode.

There's mode = 0 line still there, see below.

sound/soc/bcm/bcm2835-i2s.c
```diff
@@ -465,26 +467,46 @@ static int bcm2835_i2s_hw_params(struct snd_pcm_substream *substream,
...
+	}
 
 	/* Setup the I2S mode */
+	/* Keep existing FTXP and FRXP values. */
+	regmap_read(dev->i2s_regmap, BCM2835_I2S_MODE_A_REG, &mode);
+
+	previous_ftxp = mode & BCM2835_I2S_FTXP;
+	previous_frxp = mode & BCM2835_I2S_FRXP;
+
 	mode = 0;
```
gtrainavicius, line 489: Here the mode is reset.

gtrainavicius:
@flatmax It looks like you've missed some lines, the comments you have made are not correct. Please see the places I've pointed out in the code.

sound/soc/bcm/bcm2835-i2s.c
gtrainavicius: hw_params is called once for each direction (playback, capture), depending on what the user software wants to do. The substream->stream stores the requested direction. So the whole idea of this pull request is to react and change only the state that is relevant to the requested direction.

pelwell:
The limited context around each difference can make it hard to follow. I wondered whether the two separate previous_f?xp variables could be replaced with previous_mode, with the masking applied when mode is being assembled - something like:
```
	switch (substream->stream) {
	case SNDRV_PCM_STREAM_PLAYBACK:
		mode = (previous_mode & BCM2835_I2S_FRXP) |
			(packed ? BCM2835_I2S_FTXP : 0);
		break;
	case SNDRV_PCM_STREAM_CAPTURE:
		mode = (previous_mode & BCM2835_I2S_FTXP) |
			(packed ? BCM2835_I2S_FRXP : 0);
		break;
```
but I do think the patch is correct.

sound/soc/bcm/bcm2835-i2s.c
flatmax: OK seems good.

sound/soc/bcm/bcm2835-i2s.c
flatmax: Yep I see it now.

sound/soc/bcm/bcm2835-i2s.c
flatmax: These can still be removed, see my new comment for line number 502

sound/soc/bcm/bcm2835-i2s.c
```diff
@@ -310,6 +310,7 @@ static int bcm2835_i2s_hw_params(struct snd_pcm_substream *substream,
 	unsigned int sampling_rate = params_rate(params);
 	unsigned int data_length, data_delay, bclk_ratio;
 	unsigned int ch1pos, ch2pos, mode, format;
+	unsigned int previous_ftxp, previous_frxp;
```
flatmax, line 313: These are still unnecessary.

gtrainavicius:
@pelwell My intention was to try to be explicit about what is kept from previous state, but I'd agree with your version too.

sound/soc/bcm/bcm2835-i2s.c
```diff
@@ -465,26 +467,46 @@ static int bcm2835_i2s_hw_params(struct snd_pcm_substream *substream,
...
+	 * of the other direction stream intact. The formats of each
+	 * direction can be different as long as the frame length is
+	 * shared for both.
+	 */
+	packed = data_length <= 16;
+
+	switch (substream->stream) {
+	case SNDRV_PCM_STREAM_PLAYBACK:
+		mode |= previous_frxp;
+		mode |= packed ? BCM2835_I2S_FTXP : 0;
```
flatmax, line 502: Replace this with : mode &= packed ? BCM2835_I2S_FTXP : 0;

sound/soc/bcm/bcm2835-i2s.c
```diff
@@ -465,26 +467,46 @@ static int bcm2835_i2s_hw_params(struct snd_pcm_substream *substream,
...
+		break;
+	case SNDRV_PCM_STREAM_CAPTURE:
+		mode |= previous_ftxp;
```
flatmax, line 505: Remove this line.

sound/soc/bcm/bcm2835-i2s.c
```diff
@@ -465,26 +467,46 @@ static int bcm2835_i2s_hw_params(struct snd_pcm_substream *substream,
...
+		mode |= packed ? BCM2835_I2S_FRXP : 0;
```
flatmax, line 506: Replace this with : mode &= packed ? BCM2835_I2S_FRXP : 0;

sound/soc/bcm/bcm2835-i2s.c
flatmax: Oh yep, however you can remove this line, not necessary.

sound/soc/bcm/bcm2835-i2s.c
flatmax: Yep, I see, but you can get rid of this for simplification because of changes suggested from line 502

sound/soc/bcm/bcm2835-i2s.c
```diff
@@ -465,26 +467,46 @@ static int bcm2835_i2s_hw_params(struct snd_pcm_substream *substream,
...
-		 * so the format will be the same.
-		 */
-		mode |= BCM2835_I2S_FTXP | BCM2835_I2S_FRXP;
+	/*
+	 * Retain the frame packed mode (2 channels per 32 bit word)
+	 * of the other direction stream intact. The formats of each
+	 * direction can be different as long as the frame length is
+	 * shared for both.
+	 */
+	packed = data_length <= 16;
```
flatmax, line 497: Not sure if it is good practice, but the packed variable can also be removed and this can be inline below.

flatmax (commented):
I see what you are saying, however it is better to simplify this commit further - check my suggestions.

gtrainavicius:
I'm sure it would work as you describe it, in my opinion being more explicit about the things that's happening (leaving the mode = 0; having explicit variables for what state is kept (previous_mode would work well too)) is better for readability.

Clearing most of mode variable in some switch by doing &= and keeping some of the state is a bit less explicit and easier to miss than an unconditional mode = 0;

Having more variables on stack should be really cheap on performance, it'd be great though if there was no warning for declaring the variable closer to where it's actually used rather than forcing it to be declared at the top of the function... :)

Anyway, up to you guys to decide what style to use.

HiassofT:
I've ran some tests with the Cirrus Logic audio card (play, rec, play+rec
at 16 and 32 bit) and everything worked fine.

The wm5102 codec has symmetric_samplebits=1 so I couldn't test the
asymmetric case - but from monitoring mode via printk it seems
to work as expected.

About coding style: IMO keep it as it is. Doing acrobatics with
bit shifts and masks make the code hard to read, I'm not a huge
fan of that.

So, from my side, I think this PR is good to merge.

pelwell:
Thanks, everyone.

gtrainavicius:
Thank you!

msperl:
please upstream the patch!

flatmax:
I notice that this has been committed. Don't you think this is far simpler ?
if  (data_length <= 16)
   mode &= BCM2835_I2S_FRXP 
else 
  mode=0;

Same for the other substream.

HiassofT:
Sorry, looks like I missed some nasty bugs when testing this - details are in issue #1799

pelwell:
Reverted.

PR #1790, opened 2017-01-12
ARM64: Make it work again on 4.9

Electron752:
Here are the changes that have been made:

1.  Correctly build overlays for arm64.  I have tested this with the official RPI touchscreen and can see that the overlay for the touchscreen is automatically loaded by the firmware and I can use the touchscreen.

2. Remove the bcm2709 platform for good.

3. Update the bcmrpi3_defconfig to work with 4.9.


popcornmix:
Okay by me. @pelwell ?

arch/arm64/boot/dts/broadcom/Makefile
```diff
@@ -1,7 +1,14 @@
...
+endif
+
 dtb-$(CONFIG_ARCH_BCM2835) += bcm2837-rpi-3-b.dtb
 dtb-$(CONFIG_ARCH_BCM_IPROC) += ns2-svk.dtb
 dtb-$(CONFIG_ARCH_VULCAN) += vulcan-eval.dtb
-dtb-$(CONFIG_ARCH_BCM2709) += bcm2710-rpi-3-b.dtb
+dtb-$(CONFIG_ARCH_BCM2835) += bcm2710-rpi-3-b.dtb
 
 always		:= $(dtb-y)
 subdir-y	:= $(dts-dirs)
```
pelwell, line 15: From what I have found, the clean rules are applied without regard for the configuration. This makes sense, since when you are changing configuration you want the old configuration to be cleaned, not the new one. Because overlays is only added to dts-dirs when ARCH_BCM2835 is set, this means that the .dtb files are never deleted.

`arch/arm/boot/dts/Makefile` has the following:
```
subdir-y        := overlays
clean-files     := *.dtb
```
I think you need at least the subdir-y change (albeit with ../overlays), and perhaps clean-files as well.

Electron752:
OK, I moved this out from the conditional in the arch/arm64/boot/dts/broadcom/Makefile:
dts-dirs += ../overlays

This should allow it to visit the overlays directory unconditionally.  I did a quick confirm that overlays are building correctly and the make clean is deleting the overlays as well.

I forced a new update.

pelwell:
Yes, that's better. Merged.

DieterReuter:
Wow, that was pretty fast.

DieterReuter:
I just tested it to build locally, but I got no `overlays` folder. Maybe there is still something missing.

Electron752:
I'm looking at https://github.com/raspberrypi/linux/tree/rpi-4.9.y/arch/arm64/boot/dts and I can see the overlays folder at the end of the page.   The overlays folder is just a symbolic link to arch/arm/boot/dts/overlays.  The built overlays should be in that folder after building.

I doubt this is the problem, but some file systems like FAT or NTFS don't support symbolic links.

DieterReuter:
In my build script I'm using the make flag `O=build/rpi64` to create the build artefacts in a different folder, maybe this collides with the link. I'm testing it right now w/o the `O=` flag.

pelwell:
Last time I looked at overlays on arm64 I found that they were being installed inside a 'broadcom' directory - perhaps that is the case for you. I didn't find a solution at the time, but adding overlays to subdirs-y in arch/arm64/boot/dts/Makefile (instead of ../overlays in arch/arm64/boot/dts/Makefile) might fix it.

Electron752:
I usually just manually copy the device tree and overlays files.  That seems to work for me.

I just tested running:
make ARCH=arm64 CROSS_COMPILE=aarch64-linux-gnu- INSTALL_DTBS_PATH=dtbs -j8 dtbs_install

What happens is the standard device tree is installed in dtbs/broadcom and the overlays are installed in dtbs/overlays.  It might be possible to fix the broadcom directory issue, but that's the way the upstream kernel works so perhaps it makes sense here as well.

Electron752:
I also tried running(Note the O= part):

make ARCH=arm64 CROSS_COMPILE=aarch64-linux-gnu- INSTALL_DTBS_PATH=dtbs -j8 O=build/rpi364 dtbs

I am indeed seeing the overlays in build/rpi364/arch/arm64/boot/dts/overlays.  So at least for me, it seems to be working.

PR #1791, opened 2017-01-13
ARM64: Make it work in 4.10 + Add HDMI Audio and VCHIQ support

Electron752:
This change makes arm64 work again in 4.10.  Here is a summary of the changes.

1. The change to fix overlays and bcmrpi3_defconfig has been cherry-picked from 4.9. 
2. A non-portable cast in vc04_services has been fixed.
3. SND_BCM2835 has been fixed for ARM64 now that VCHIQ works.
4. HDMI Audio and VCHIQ are now enabled in bcmrpi3_defconfig

I tested the audio by playing some FLAC audio files and a DVD rip from VLC on debian sid with the xfce4 desktop.   I enabled audio and VC4 video with the following config.txt parameters:

```
dtparam=audio=on
dtoverlay=vc4-kms-v3d
```

Ideally I would be great if this could somehow be back-ported to 4.9 since that's probably going to be the long term support release.  That would require back-porting the upstream 4.10 of vchiq to 4.9 through.


popcornmix:
I'm okay with this. Any objections @pelwell ?

pelwell:
No objections.

PR #1792, opened 2017-01-14
ARM64: Enable Kernel Address Space Randomization

Electron752:
This change enable address space randomization which is a great security feature in the ARM64 default build configuration.  Testing was done with the firmware given at https://github.com/raspberrypi/firmware/issues/694 which I understand will be included in the next public drop on github.  Testing included testing a randomization enabled kernel on an old firmware version to make sure nothing breaks when randomization isn't available.

It would be awesome if this change could also be applied to 4.9 and possibly 4.8 so that I don't have to create a new pull request just to update older branches.


popcornmix:
4.8 has reached [EOL](https://www.kernel.org/) so I don't think we'll continue updating it.
I'll push to 4.9 when merged here.
I'm okay with this. @pelwell?

pelwell:
Of course - i don't see any reason not to merge, now the firmware is going to support it. Let's have it in 4.9 as well.

PR #1794, opened 2017-01-15
ARM64/DWC_OTG: Port dwc_otg driver to ARM64

Electron752:
In ARM64, the FIQ mechanism used by this driver is not implemented.   As a workaround, a reqular IRQ is used instead of FIQ. The IRQ-CPU mapping is round robined on ARM64 to increase concurrency and allow multiple interrupts to be serviced at a time.  This reduces the need for FIQ.
   
I changed the default driver in the device tree to dwc_otg.  If it breaks on anybody, they can use the standard device tree overlays to switch back to dwc2.

Both dwc_otg and dwc2 are enabled in the build configuration(bcmrpi3_defconfig).
 
Ideally, this change could be applied to 4.9 as well as 4.10 if accepted.

Tests Run:
    
This mechanism is most likely to break when multiple USB devices are attached at the same time.  So the system was tested under stress.
    
Devices:
    
1. USB Speakers playing back a FLAC audio through VLC at 96KHz.(Higher then typical, but supported on my speakers).
    
2. sftp transferring large files through the built-in ethernet connection which is connected through USB.
    
3. Keyboard and mouse attached and being used.
    
Although I do occasionally hear some glitches, the music seems to play quite well.


popcornmix:
Getting fiq support for ARM64 would certainly be a preferable solution.
@P33M any comments on the ARM64 scheme used here?

Electron752:
I do have access to a bunch of different types of USB devices for testing and I have a USB bus analyzer.  So let me know if I can provide any data from specific tests to demonstrate this is working better then the DWC2 driver.  The important thing is that the changes in this attempt are all ARM64 specific, so they should be safe for ARM32.  

Ideally, I think the dwc2 driver could be improved to incorporate some of the ideas in dwc_otg, but I think getting dwc_otg working is a good short term solution.

I'm also thinking that instead of the FIQ hack, the Linux kernel really needs support for interrupt priorities and nested interrupts.  I'm guessing that would be a nightmare to get added to upstream through.

P33M:
I had toyed with the idea of round-robin IRQ dispatch but ultimately this
doesn’t fix the inherent latency with having non-nested IRQs. Round-robin
distribution of interrupts means you’re less likely to hit a core that’s
currently got them disabled, assuming equal distribution of critical
sections across tasks, but you could still potentially hit a core that’s
going to have IRQs disabled for too long.

Our realtime requirement is max 60uS latency between assertion and
servicing – which currently only the FIQ can guarantee as it’s never
blocked and the execution time is minimal.

The rt-preempt patchset has been around for ages but has never been
mainlined for various reasons. If this patchset were in the upstream
kernel, then it’d certainly be a candidate for replacing the FIQ – if and
only if the hard realtime requirement could be met in the face of SDRAM
bandwidth pressure. Absent this facility, I think we need to support FIQs
on aarch64.

Electron752:
I agree FIQ is the way to go and I think it's not unreasonable for me or someone else to make it work.  I actually have a JTAG debugger now, so writing and debugging very low level ASM is not completely unreasonable.

What I don't understand is why a sort of working dwc_otg is worse then dwc2?  What happens if the 60uS isn't quite met? I would assume some transfer just happens in the wrong location of the frame.  That seems better to me then the dwc2 approach where if alot of devices are attached at a time some of them can be completely starved.

BTW, dwc_otg isn't perfect with FIQ.  I found through experience the driver doesn't do a very good job of scheduling around the SOF markers which can cause broken ISO data.

Electron752:
Case in point, I just did a quick test with both dwc_otg and dwc2 with aarch64.

Here are the devices that are attached:
1. Internal Ethernet network
2. USB Speakers(96KHz)
3. Keyboard and Mouse

With dwc2, I'm finding that the speaker are not working at all.  In fact, I'm seeing a constant stream of these errors.
[   61.311617] dwc2 3f980000.usb: DWC OTG HCD URB Enqueue failed adding QTD. Error status -28
[   61.311644] usb 1-1.3: cannot submit urb 0, error -28: not enough bandwidth
[   61.319749] dwc2 3f980000.usb: dwc2_do_reserve: No host channel available for periodic transfer
[   61.319784] dwc2 3f980000.usb: DWC OTG HCD URB Enqueue failed adding QTD. Error status -28
[   61.319811] usb 1-1.3: cannot submit urb 0, error -28: not enough bandwidth
[   61.332738] dwc2 3f980000.usb: dwc2_do_reserve: No host channel available for periodic transfer
[   61.332774] dwc2 3f980000.usb: DWC OTG HCD URB Enqueue failed adding QTD. Error status -28
[   61.332801] usb 1-1.3: cannot submit urb 0, error -28: not enough bandwidth

With dwc_otg + my round robin hack everything is fully functional.

I do occasionally hear a pop and an error about a NYET, but it does work.
[  198.996657] Transfer to device 6 endpoint 0x2 frame 1321 failed - FIQ reported NYET. Data may have been lost.
[  199.119239] Transfer to device 6 endpoint 0x2 frame 1444 failed - FIQ reported NYET. Data may have been lost.
[  199.212720] Transfer to device 6 endpoint 0x2 frame 1537 failed - FIQ reported NYET. Data may have been lost.
[  200.755211] Transfer to device 6 endpoint 0x2 frame 1032 failed - FIQ reported NYET. Data may have been lost.

I haven't attached the bus analyzer but the NYETs could potentially be do to the SOF scheduling issue.

Electron752:
@popcornmix: Any reason this PR can't be accepted?

I realize P33M isn't 100% happy with it but it's a step forward even though it isn't perfect.   The changes are all arm64 specific, so the only real issue is long term maintenance. dwc_otg doesn't appear to be in "heavy" development at this time, so if arm64 gets broken by a later 32 bit change it isn't that big of a deal and can probably be fixed easily.

In some cases dwc_otg works better and in some cases dwc2 works better.    By getting dwc_otg at least partially working, it gives the option it easily switch between the two drivers with a simple config.txt overlay switch.

I am more then willing in the near feature to make an attempt to code up a full FIQ based replacement for 64 bit.  Of course, it would probably have to stay downstream as FIQ is out of the mainstream kernel architecture.

popcornmix:
Making FIQ work on ARM64 feels like the best solution, and we'd certainly be interested in a downstream patch if you do get that working (and upstream may be persuaded if it can be shown to work better).

However if this patch improves user experience for arm64 experimenters then we don't mind merging it.

Electron752:
Cool!!!  Thanks much!!!

PR #1795, opened 2017-01-15
Reapply "spi-bcm2835: Disable forced software CS"

ED6E0F17:
bcm2708_common.dtsi is not present in linux 4.9 device tree - there seem to be duplicated functions between bcm283* and bcm270*
````
"Select software CS in bcm2708_common.dtsi, and disable the automatic
conversion in the driver to allow hardware CS to be re-enabled with an
overlay."
````

### DIFF
--- a/arch/arm/boot/dts/bcm283x.dtsi
+++ b/arch/arm/boot/dts/bcm283x.dtsi
@@ -163,6 +163,7 @@
 			#address-cells = <1>;
 			#size-cells = <0>;
 			status = "disabled";
+			cs-gpios = <&gpio 8 1>, <&gpio 7 1>;
 		};
 
 		i2c0: i2c@7e205000 {


pelwell:
I think it is misleading to have a pinctrl configuration asking for ALT0 on pins 7 and 8, relying on the fact that they will become outputs when the cs-gpios are created, but I can live with it if it avoids patching another upstream file.

ED6E0F17:
It would be more direct if HW chip-select could be the default, and pushed upstream, with SW chip-select available as an overlay where it breaks. Some people wire up their hardware to the wrong pins, so it is useful to have Software CS in an overlay. ( https://github.com/PiInTheSky/lora-gateway/issues/31 )

pelwell:
I am happy for software CS to be the default since hardware CS has known problems, but I don't like the forced conversion in the driver which makes it impossible to request hardware CS. However, it's upstream now.

@popcornmix Happy to merge?

PR #1796, opened 2017-01-15
ARM64: Fix RTL8187/RTL8192CU wifi

Electron752:
These drivers include the ASM version of the ipv6 checksum routines from the base system.  This function has not been implemented in the base kernel for ARM64 probably because nobody has bothered to implement them.  The easy solution is to simply include a header from the base system that picks up the generic "C" version of this function.

This pull request is mostly being made to minimize the difference in build configurations between 32 bit and 64 bit.


popcornmix:
It looks like the include is fine with ARM32 so I'd be happier if you drop the ifdef and just add the include.

Electron752:
OK, give me a second.

Electron752:
OK, done.  The header is now included unconditionally.

pelwell:
OK by me.

Electron752:
It should be safe to send this to 4.10 as well.

Electron752:
BTW, thanks for merging this.

popcornmix:
Yes, added to 4.10.

PR #1805, opened 2017-01-19
config: Add CONFIG_USB_DWC2 and CONFIG_USB_GADGET

pelwell:
The introduction of CM3 makes gadget mode on 2709 a useful option,
so enable the building of the required modules. Note that these
modules are not loaded by default and must be enabled with a DT
overlay.


ED6E0F17:
Can you confirm that this enables USB_DWC2_DUAL_ROLE ?

popcornmix:
Yes I see CONFIG_USB_DWC2_DUAL_ROLE=y in .config after building with this change.

pelwell:
I've now added the remaining USB options that were in bcmrpi_defconfig but missing from bcm2709_defconfig.

Any objections to merging this?

popcornmix:
No objection.

PR #1810, opened 2017-01-22
Add driver for Wolfson / Cirrus Logic audio card

HiassofT:
I rewrote the original Wolfson / Cirrus driver almost from scratch, fixing a lot of bugs and adding several new features like setting / reading the AES channel status bits and most importantly removing the need to hack up upstream codec drivers.

The card needs regulator support enabled in the kernel, because the WM5102 chip's DCVDD input is supplied by an on-chip LDO. I've added the necessary configuration and DT changes as separate commits.

The original PR by Cirrus against kernel 3.12 hadn't set CONFIG_REGULATOR_DUMMY which caused several other soundcard drivers to break. Since kernel 3.13 this is implicitly set when enabling CONFIG_REGULATOR and the regulator framework will automatically add in dummy regulators if no regulators are specified in DT - in other words, devices with mandatory regulator properties in DT won't break if these properties aren't set.

Although the "Consolidate audio card overlays" commit isn't strictly needed I think it's a good idea to clean up the existing overlays and properly set the missing properties.

Several details of how to best expose the onboard 3.3V and 5V rails in DT have been discussed before in #1756, the regulator commits in this PR are identical to the ones in #1756.


popcornmix:
Well it's a lot cleaner that the first wolfson patch I saw.
I only have a slight reservation about enabling regulator support but I assume the issues we had in the past are resolved by the DT updates.

@pelwell any comments?

pelwell:
Provided scamming regulator support causes no problems (and I guess we won't know until we try) I'm happy with this commit. It will open the way for a few other codecs.

popcornmix:
Lets get this in and see what is reported in Milhouse builds.

cfstras:
Thanks for doing this!
As I couldn't find the information anywhere: when is this 4.9 kernel scheduled to appear in the Raspbian package repositories?

pelwell:
Once we're happy it doesn't break things. There's been an issue with I2C and SPI clocks that should be resolved with the next kernel build, then we'll take stock.

cfstras:
Okay, great. Is there a central place for release notes, or an easy way to read the changelogs for new packages?
👍 for all of your work!

pelwell:
They aren't collated, but the commit messages in the linux and firmware repos are there to be read. Don't expect many differences in Raspbian - we don't like changing too much at once.

PR #1812, opened 2017-01-23
ARM64: enable CONFIG_KEYS_COMPAT

bamarni:
Currently in 32-bit userspace calling the keyctl syscall returns an ENOSYS error.

This makes sure the compatibility wrapper is registered.

Signed-off-by: Bilal Amarni <bilal.amarni@gmail.com>


popcornmix:
This sounds like a change that should be made upstream. Could you submit it there?
No problem merging it here early, but would be good to know is it destined for upstream.

bamarni:
Sure, I'd be happy to submit it there too. I've never did it so I'd have to check the contribution instructions first.

popcornmix:
Some info here: https://kernel.org/doc/html/latest/process/submitting-patches.html

bamarni:
Thanks, I'll let you know here when it's done.

bamarni:
So I've emailed it yesterday to the list suggested by the maintainers script. It was probably a good fit for the trivial email too though? If it doesn't get any attention I'll try it there.

bamarni:
It has been applied as a more generic patch (in security/keys/Kconfig) by the keyring maintainers, so hopefully it will make its way to upstream at some point!

PR #1813, opened 2017-01-23
drm/vc4: Backport to 4.9

anholt:
Here's a giant backport for the 4.9 branch.  vc4-kms-v3d is working on my DSI panel.  TV out ought to work, but I haven't tested it.


popcornmix:
Update is working on HDMI.
What should I do to test DSI display?
What should I do to test composite display?

i.e. any special config.txt settings or mods to overlays needed to enable?

anholt:
DSI just comes up with no extra config.txt.  This is because I'm not using DSI transactions any more (never got them to work), so the firmware stealing our interrupt isn't a problem any more.

TV-out won't be enabled by default by fbdev or X, because we have no load detection of TV (sigh) so we report unknown connector state.  You may be able to bring up fbdev with kernel command line described in `Documentation/fb/modedb.txt` but I've just tested with xrandr on X.

popcornmix:
DSI is not coming up in my testing.
If I boot with DSI LCD and HDMI connected I see the firmware rainbow splash screen on LCD. It blanks and I see boot console text appear on HDMI.
If I unplug HDMI first, I see the firmware rainbow splash screen on LCD. It blanks and I see no more. If I reattach hdmi later I see the console (but at a lower resolution).

Is this an issue with our hotplug detect on pi3?

ktb83:
Here is what happens for me using this pull request (1813) with only DSI connected:
```
[    19.170] (II) glamor: EGL version 1.4 (DRI2):
[    19.228] (II) modeset(0): glamor initialized
[    19.230] (II) modeset(0): Output HDMI-1 has no monitor section
[    19.230] (II) modeset(0): Output Composite-1 has no monitor section
[    19.230] (II) modeset(0): Output DSI-1 has no monitor section
[    19.231] (II) modeset(0): EDID for output HDMI-1
[    19.231] (II) modeset(0): EDID for output Composite-1
[    19.232] (II) modeset(0): Printing probed modes for output Composite-1
[    19.232] (II) modeset(0): Modeline "720x480"x31.3   13.50  720 734 798 858  480 483 486 502 interlace (15.7 kHz e)
[    19.232] (II) modeset(0): EDID for output DSI-1
[    19.232] (II) modeset(0): Output HDMI-1 disconnected
[    19.232] (II) modeset(0): Output Composite-1 disconnected
[    19.232] (II) modeset(0): Output DSI-1 disconnected
[    19.232] (WW) modeset(0): No outputs definitely connected, trying again...
[    19.232] (II) modeset(0): Output HDMI-1 disconnected
[    19.232] (II) modeset(0): Output Composite-1 connected
[    19.232] (II) modeset(0): Output DSI-1 disconnected
[    19.232] (II) modeset(0): Using sloppy heuristic for initial modes
[    19.232] (II) modeset(0): Output Composite-1 using initial mode 720x480 +0+0
[    19.232] (==) modeset(0): Using gamma correction (1.0, 1.0, 1.0)
[    19.232] (==) modeset(0): DPI set to (96, 96)
[    19.232] (II) Loading sub module "fb"
[    19.232] (II) LoadModule: "fb"
[    19.232] (II) Loading /usr/lib/xorg/modules/libfb.so
[    19.243] (II) Module fb: vendor="X.Org Foundation"
```

anholt:
@ktb83 it sounds like the panel driver isn't getting loaded.  Is it in lsmod?  Attach dmesg?  babb24fec12ce391157be8d7355df5a8e991355e should have caused the driver to get loaded, though I'm wondering how this works without 4c4d16a7882ef86005cfadaa84370a8740d47836 from the downstream tree.

ktb83:
The following information is from an RPi3B running Debian Sid ARM64 with HDMI and DSI connected.

```
$ lsmod
rfcomm                 53248  2
cmac                   16384  1
bnep                   24576  2
hci_uart               32768  1
btbcm                  16384  1 hci_uart
bluetooth             397312  29 hci_uart,bnep,btbcm,rfcomm
brcmfmac              262144  0
vc4                   163840  2
drm_kms_helper        196608  2 vc4
evdev                  24576  4
brcmutil               20480  1 brcmfmac
joydev                 20480  0
drm                   442368  5 vc4,drm_kms_helper
cfg80211              651264  1 brcmfmac
syscopyarea            16384  1 drm_kms_helper
sysfillrect            16384  1 drm_kms_helper
sysimgblt              16384  1 drm_kms_helper
fb_sys_fops            16384  1 drm_kms_helper
rfkill                 32768  5 bluetooth,cfg80211
i2c_gpio               16384  0
i2c_algo_bit           16384  1 i2c_gpio
i2c_bcm2835            16384  0
spi_bcm2835            20480  0
rpi_ft5406             16384  0
uio_pdrv_genirq        16384  0
rpi_backlight          16384  0
uio                    24576  1 uio_pdrv_genirq
i2c_dev                20480  0
ip_tables              32768  0
x_tables               45056  1 ip_tables

```

 Xorg.0.log: 
[rpi3_arm64_DSI-HDMI_rpf4.9-pr1813_xorg0log.txt](https://github.com/raspberrypi/linux/files/728400/rpi3_arm64_DSI-HDMI_rpf4.9-pr1813_xorg0log.txt)

dmesg:
[rpi3_arm64_DSI-HDMI_rpf4.9-pr1813_dmesg.txt](https://github.com/raspberrypi/linux/files/728399/rpi3_arm64_DSI-HDMI_rpf4.9-pr1813_dmesg.txt)

For what it's worth, I also tried adding in `EXPORT_SYMBOL(of_device_uevent_modalias);` to drivers/of/device.c  without any further success.

anholt:
@ktb83 Wait, debian sid arm64?  What defconfig did you use?  Are you sure you built the panel driver?

@popcornmix would need to see your dmesg/xorg log, too.

popcornmix:
dmesg: http://paste.debian.net/910596/
xorg: http://paste.debian.net/910599/

ktb83:
@anholt Yes. bcmrpi3_defconfig. Sorry, I didn't realize that wasn't updated. So, I'm pretty sure that I didn't build the panel driver after looking at your commits and my .config. I've modified bcmrpi3_defconfig and I'm trying again.

anholt:
There's the problem:

    raspberrypi-touchscreen 3f700000.dsi.0: Unknown Atmel firmware revision: 0x00

So what's going wrong with i2c-gpio?  Does the pinmux for them show us still muxed to GPIO input/output, not the HW I2C controller?

anholt:
Re-pushed to fix up bcmrpi3_defconfig

ktb83:
@anholt Thanks. My build just finished and put me in the same place `[    4.360259] raspberrypi-touchscreen 3f700000.dsi.0: Unknown Atmel firmware revision: 0x00`

Edit: BTW, I don't understand your question, but if I can provide any helpful command output (i2cdetect, raspi-gpio get), feel free to request it.

Electron752:
@anhot: Thanks for fixing bcmrpi3_defconfig.

I have a touchscreen, so I'll do a build an see what I can do to possibly help.

Electron752:
Same issue here are well:
i2c-gpio soc:i2cdsi: using pins 28 (SDA) and 29 (SCL)
raspberrypi-touchscreen 3f700000.dsi.0: Unknown Atmel firmware revision: 0x00

Electron752:
Here is something interesting from the log:
/sys/kernel/debug/pinctrl/3f200000.gpio# cat pinmux-pins
Pinmux settings per pin
Format: pin (name): mux_owner gpio_owner hog?
pin 0 (gpio0): (MUX UNCLAIMED) (GPIO UNCLAIMED)
pin 1 (gpio1): (MUX UNCLAIMED) (GPIO UNCLAIMED)
pin 2 (gpio2): (MUX UNCLAIMED) (GPIO UNCLAIMED)
pin 3 (gpio3): (MUX UNCLAIMED) (GPIO UNCLAIMED)
pin 4 (gpio4): (MUX UNCLAIMED) (GPIO UNCLAIMED)
pin 5 (gpio5): (MUX UNCLAIMED) (GPIO UNCLAIMED)
pin 6 (gpio6): (MUX UNCLAIMED) (GPIO UNCLAIMED)
pin 7 (gpio7): (MUX UNCLAIMED) (GPIO UNCLAIMED)
pin 8 (gpio8): (MUX UNCLAIMED) (GPIO UNCLAIMED)
pin 9 (gpio9): (MUX UNCLAIMED) (GPIO UNCLAIMED)
pin 10 (gpio10): (MUX UNCLAIMED) (GPIO UNCLAIMED)
pin 11 (gpio11): (MUX UNCLAIMED) (GPIO UNCLAIMED)
pin 12 (gpio12): (MUX UNCLAIMED) (GPIO UNCLAIMED)
pin 13 (gpio13): (MUX UNCLAIMED) (GPIO UNCLAIMED)
pin 14 (gpio14): (MUX UNCLAIMED) (GPIO UNCLAIMED)
pin 15 (gpio15): (MUX UNCLAIMED) (GPIO UNCLAIMED)
pin 16 (gpio16): (MUX UNCLAIMED) (GPIO UNCLAIMED)
pin 17 (gpio17): (MUX UNCLAIMED) (GPIO UNCLAIMED)
pin 18 (gpio18): (MUX UNCLAIMED) (GPIO UNCLAIMED)
pin 19 (gpio19): (MUX UNCLAIMED) (GPIO UNCLAIMED)
pin 20 (gpio20): (MUX UNCLAIMED) (GPIO UNCLAIMED)
pin 21 (gpio21): (MUX UNCLAIMED) (GPIO UNCLAIMED)
pin 22 (gpio22): (MUX UNCLAIMED) (GPIO UNCLAIMED)
pin 23 (gpio23): (MUX UNCLAIMED) (GPIO UNCLAIMED)
pin 24 (gpio24): (MUX UNCLAIMED) (GPIO UNCLAIMED)
pin 25 (gpio25): (MUX UNCLAIMED) (GPIO UNCLAIMED)
pin 26 (gpio26): (MUX UNCLAIMED) (GPIO UNCLAIMED)
pin 27 (gpio27): (MUX UNCLAIMED) (GPIO UNCLAIMED)
**pin 28 (gpio28): (MUX UNCLAIMED) pinctrl-bcm2835:28
pin 29 (gpio29): (MUX UNCLAIMED) pinctrl-bcm2835:29**
pin 30 (gpio30): (MUX UNCLAIMED) (GPIO UNCLAIMED)
pin 31 (gpio31): (MUX UNCLAIMED) (GPIO UNCLAIMED)
pin 32 (gpio32): 3f201000.serial (GPIO UNCLAIMED) function alt3 group gpio32
pin 33 (gpio33): 3f201000.serial (GPIO UNCLAIMED) function alt3 group gpio33
pin 34 (gpio34): 3f300000.mmc (GPIO UNCLAIMED) function alt3 group gpio34
pin 35 (gpio35): 3f300000.mmc (GPIO UNCLAIMED) function alt3 group gpio35
pin 36 (gpio36): 3f300000.mmc (GPIO UNCLAIMED) function alt3 group gpio36
pin 37 (gpio37): 3f300000.mmc (GPIO UNCLAIMED) function alt3 group gpio37
pin 38 (gpio38): 3f300000.mmc (GPIO UNCLAIMED) function alt3 group gpio38
pin 39 (gpio39): 3f300000.mmc (GPIO UNCLAIMED) function alt3 group gpio39
pin 40 (gpio40): (MUX UNCLAIMED) (GPIO UNCLAIMED)
pin 41 (gpio41): (MUX UNCLAIMED) (GPIO UNCLAIMED)
pin 42 (gpio42): (MUX UNCLAIMED) (GPIO UNCLAIMED)
pin 43 (gpio43): 3f201000.serial (GPIO UNCLAIMED) function alt0 group gpio43
pin 44 (gpio44): (MUX UNCLAIMED) (GPIO UNCLAIMED)
pin 45 (gpio45): (MUX UNCLAIMED) (GPIO UNCLAIMED)
pin 46 (gpio46): (MUX UNCLAIMED) (GPIO UNCLAIMED)
pin 47 (gpio47): (MUX UNCLAIMED) (GPIO UNCLAIMED)
pin 48 (gpio48): 3f202000.sdhost (GPIO UNCLAIMED) function alt0 group gpio48
pin 49 (gpio49): 3f202000.sdhost (GPIO UNCLAIMED) function alt0 group gpio49
pin 50 (gpio50): 3f202000.sdhost (GPIO UNCLAIMED) function alt0 group gpio50
pin 51 (gpio51): 3f202000.sdhost (GPIO UNCLAIMED) function alt0 group gpio51
pin 52 (gpio52): 3f202000.sdhost (GPIO UNCLAIMED) function alt0 group gpio52
pin 53 (gpio53): 3f202000.sdhost (GPIO UNCLAIMED) function alt0 group gpio53

anholt:
Yet another SD card died today, so I've swapped my main dev system from a pi2 with u-boot on the sd card to a pi3 network-booting without u-boot. This should make downstream dev a lot easier for me.

In the process I reproduced the error, which was that we were using the wrong pins for I2C on pi3 and early pi1.  Fixed now.

Electron752:
Cool.  I'm rebuilding your branch.

Electron752:
OK, it works but the display is upside down.  Also, I'm seeing some flickering in the display.  
Any idea what that could be?

I see the pins have been changed and I don't see the firmware error anymore.
i2c-gpio soc:i2cdsi: using pins 44 (SDA) and 45 (SCL)

ktb83:
Oh, is it ignoring lcd_rotate=2?

Electron752:
No.

And the coordinates are backward compared to the firmware touchscreen input driver.

Electron752:
Backward meaning upside down.  I touch at the top and it thinks I touched the bottom.

Electron752:
Also, I'm wondering if this is related to the flickering:
[    3.389870] [drm] Supports vblank timestamp caching Rev 2 (21.10.2013).
[    3.389928] [drm] Driver supports precise vblank timestamp query.

ktb83:
Interesting. I see the flickering too. I have epilepsy (I'll be careful :) )
The antialiasing is rough (non-existent?) in general, but it seems to be working OK for me.

@Electron752 I don't notice a reversed/flipped image or reversed/flipped touch interface on my panel. That seems to be working correctly for me at least in LightDM+Xfce. I am using lcd_rotate=2.

Electron752:
I'm also using LightDM+xfce.   I tried adding lcd_rotate=2 and it didn't change anything.  I know the display can be mounted in both directions so maybe I just have it mounted differently.

I suspect something may be wrong with the vertical hold, but I really don't know.  I'll leave that up to the experts.

I just stopped in because I noticed the I2C issue and that's something I know about.  Just thought I could help.  I'm also looking for a fix for weird GPU Resets/Hangs and CMA issues.   I was maybe hoping these got fixed.

popcornmix:
With no config.txt settings the LCD is upside down compared to firmware driver.
However the touchscreen is not oriented the same as screen (it matches firmware orientation).

If I add lcd_rotate=2 to config.txt then the orientation of firmware and kms driver matches (including touchscreen).

I do see some sparkling orange pixels with the kms driver that aren't present in the firmware driver (visible with the default PIXEL "road" desktop wallpaper).

anholt:
config.txt is configuration for the firmware, not Linux, so the Linux driver has no idea what you put there.

pelwell:
> config.txt is configuration for the firmware, not Linux, so the Linux driver has no idea what you put there.

Of course - I think @popcornmix was just trying to establish a common frame of reference, i.e. which way is up.

popcornmix:
> config.txt is configuration for the firmware, not Linux, so the Linux driver has no idea what you put there

Indeed. But with blank config.txt the display is upside down compared to touchscreen. With ``lcd_rotate=2`` the two are aligned which suggests some of the firmware setup is persisting.

I'm quite happy for ``lcd_rotate`` to be unsupported with the kms driver, but it appears necessary at the moment to get a consistent orientation of display and touchscreen.

ktb83:
I added in the snd_bcm2835 and vchiq arm64 compatibility fixes @Electron752 got commited to the 4.10 branch. Those seem to work fine.

I have it behaving pretty well on most boots now:
Image of ARM64 Pi3B Xfce glxgears

anholt:
@popcornmix do you have thoughts about the appropriate default orientation of the panel?  I chose this orientation because that's what the display mounting kit I have is built for, and it seemed from searching that most people with the touchscreen were setting the config.txt flip setting as well.  Could we go with this as default for vc4 and have the firmware automatically set the input flipping when vc4 is present?

On a related note, I had a chat with Peter Hutterer about input/screen orientation matching, and in the Linux world it's up to the desktop environment to keep the two matched up when the desktop flips a touchscreen.  This is done with some heuristics in gnome-settings-daemon, and I don't know if XFCE has similar heuristics or not.

This means that people using xrandr to manually rotate their panels will also need to manually set the CTM in xinput.

popcornmix:
Personally I think the default panel orientation should match the default orientation of the firmware driver. Anything else just adds confusion.

The firmware orientation was chosen due to the viewing angle of the display (it is better from above than below). It is not an option to change the default firmware orientation at this point when so many users would be disrupted.

Electron752:
@anholt - Arbitrary issues aside, I was wondering if this pull request could possibly contain a fix for the weird GPU reset/hangs on 64 bit. @ktb83 seems to be getting some high number on glxgears, so I'm hopeing that issue has been resolved.

I'm really looking for that 64 bit fix.

ktb83:
Hi @Electron752,

That screenshot was from what turned out to be a rare boot where the image on the screen wasn't dancing/jumping up and down/flickering and the text/antialiasing was fairly sharp. Unfortunately, that only happened a few more times out of many boots to Xfce. :( glxgears was run with vblank_mode=0 and v3d_freq was 500, I think.

I should mention that I've been using LXQT and that doesn't seem to exhibit the same frequent lock ups (frozen image, spinning unmovable cursor on the Pi all while SSH is completely responsive) as Xfce4. It does however seem to exhibit issues with the recommended Compton config (backend = "glx"; unredir-if-possible = true; glx-swap-method = "buffer-age"; glx-no-stencil = true; paint-on-overlay = true; vsync = "opengl";). I think I've narrowed the cause down to the unredir-if-possible = true; option. That seems to cause the Panel in LXQT to be invisible.

I posted some other info here: https://www.raspberrypi.org/forums/viewtopic.php?f=71&t=156557

Electron752:
Very interesting on the vchiq + other drivers.  I've had vchiq 32 bit compat working for awhile now in many different implementations.  I just can't get it through the check-in process.   Since you have the changes for the other drivers, maybe you should submit a pull request and get the changes in the downstream tree?

Although, I'm a bit confused how vcdbg helps with the vc4-kms driver?  I thought vcdbg was for the firmware, and the kms used the GPU registers directly?

ktb83:
I'm not sure if a pull request from me is a good idea for a few reasons. I've never submitted a pull request myself and I don't have much Git experience in general. I know enough to clone a repo, merge in a pull request to a new local branch and modify the code manually, but that's about it. I didn't document the changes I made (and IIRC there were plenty of changes to make your 4.10 commits work), so I imagine it would have to be reviewed carefully. I also haven't really programmed in C/C++ since high school so it required a lot of Googling. I'm just a web developer.

With that said, I see no reason why I can't upload my local branch to one of my GitHub accounts if anybody wants to see/try the code. I might need instructions to make that happen. :)

I'm mostly hoping to use vcdbg for device tree stuff. Perhaps it doesn't help with that any longer?

Electron752:
Uploading to a github tree makes sense. ATM, I'm mostly interested in the 32 bit compat layer for vchiq.  I especially like the idea of using a separate file for compat since it makes the diff easier to review and makes the review process easier in general.  

I would be certainly willing to review it for you and submit it through the process.   It would be reviewed by others as well.

I'm just wondering if it would make sense to put in in the downstream kernel first, or if it better to try to send it to the main kernel source first.  I would tend to think check it into downstream first since it has a smaller distribution scope until it gets tested.  And nobody would really want it except to run raspbian I would think.

Electron752:
BTW, I've been kicking around an idea without much success of splitting the userland API layer of vchiq into an independent driver.  I wonder how that would fit in with your ideas?

For example, the main kernel source is mostly interested in the camera module.  I would like to think a bit a the ALSA sound system as well.  Both of which need absolutely no user interface except the standard ALSA and video for linux API layers.

ktb83:
That sounds like a very good idea.

Electron752:
Which part?

ktb83:
It sort of sounds like one idea to me -- separating specialized userland related code out from the more generic code such as for the camera and audio.

Electron752:
Cool. I think I need to pound this out some more and maybe I can send an RFC to the rpi-linux mailing list.

Can you share out your work? All you would need to do is something like this if you already have a github account.

git push https://www.github.com/[ACOUNTNAME]/[REPONAME] +[LOCAL BRANCH]:[REMOTE BRANCH]
ACOUNTNAME        = Your github account name
REPONAME             = Name of the repository on github
LOCAL BRANCH      = Name of the branch on your local tree(get this from git status)
REMOTE BRANCH   = Branch name on github.

ktb83:
Sure. I think I'm going to build it once more right now after turning the vchiq debug level down (mm7 set that high). It kind of floods the logs an annoying amount. Then I'll upload it. Thanks for the instructions.

ktb83:
That command helped. I also had to figure out how to create a remote repo first. The command you gave me pushed rpi-4.9.y+PR#1813 and then I had to figure out how to commit my changes and push them too. Hopefully it worked. https://github.com/ktb83/linux

anholt:
Testing the commit to default to the same flip mode as the closed stack.  In the meantime, people with flickering bugs: Can you please attach (not paste) the output of `cat /debug/clk/clk_summary` after mounting debugfs?

Electron752:
I've attached the file.  
BTW, I've noticed this doesn't happen on every reboot. Just most of the time.  Some times it works fine.

[clk_summary.txt](https://github.com/raspberrypi/linux/files/740655/clk_summary.txt)

anholt:
Nothing obviously wrong with the clocks there.  A next step might be to try smashing in the same timing parameters as the closed firmware.  I should have that patch laying around somewhere...

Electron752:
Let me know when you find the patch.

anholt:
OK, pushed my branch with the rotation fix without having specifically tested it downstream.  Right now dwc_otg is blowing up when network-manager comes up.

Electron752:
Does this include the parameters patch for the flicking?  If so, I'm willing to see if I can help debug dwc_otg.

anholt:
No, I wouldn't intentionally put debug code like that in a PR.

Electron752:
OK, I'll build and take a look at dwc_otg and see if I can see anything obvious...

Electron752:
So I built it for arm64 and it boots just fine. I can use the network and ssh into the RPI 3 just fine.
I have a static IP in interfaces and no keyboard attached.  Perhaps that's the issue.  I'll try a few things on my end, but otherwise I'll have to rebuild for 32 bit.

I still see the flickering on the LCD panel.

Electron752:
OK, I boot with network manager instead of having a static IP in /etc/network/interfaces.   I don't see the problem on my end.  Networking is fine and the keyboard and touchscreen both work...

I'm still seeing very bad GPU hangs and the flicker.

What exactly do you mean by dwc_otg blowing up?  Is everything unresponsive? Does lsusb still work?

Electron752:
Yet, it works on my end.

Do any of your drivers use FIQ?  Because that's probably the only difference I can think of between 32 bit and 64 bit.  The 64 bit version doesn't use FIQ and it round robins the interrupts.   The 32 bit version does use FIQ and all the interrupts are on core 1.

anholt:
My USB problem is not worth debugging here.  It's probably just power troubles, since I've got a cheap cable, some rando USB power adapter, and the DSI panel attached.

Electron752:
So I'm looking at:
vc4_irq(int irq, void *arg)

I see some assumptions that IRQ are all on the same CPU.   And interrupts are ACKed before they are handled at times.

Would it be possible to fix VC4 to handle IRQs from different cores?

anholt:
@Electron752 this is a pull request for the 4.9 backport.  If you have a bug to report on vc4, please report it in a separate issue.

Electron752:
OK, where should I open the issue at and what should I put in it?  I don't want to hold up the pull request by any means, but it looks like this could potentially get fixed rather easily.

Perhaps instead of using the default work queue, just create a new work queue that's bound to a single cpu.  Or always schedule the work to the same cpu.

Electron752:
Yes, I think all that would be required would be in 
vc4_irq to replace the call to schedule_work(...) with schedule_work_on(0, ...)

anholt:
https://github.com/anholt/linux/issues would be the place for vc4 kernel bugs.  But please be sure to explain there what the bug is that you think you've found.

Electron752:
OK. Will do.  I tried the schedule_work_on(0,..) and it isn't sufficient to fix this.   Oh well.

Electron752:
@anholt: Sorry to bother you and I really don't want to take over your pull request.  But I was wondering what the purpose of using the wait queue is in vc4.

Have you seen this entry in menuconfig?  This is appears to be a relatively recent change so it's no longer safe to assume that work is always going to run on the scheduling cpu.  This doesn't seem to just apply to the change I made for 64 bit.

config DEBUG_WQ_FORCE_RR_CPU
	bool "Force round-robin CPU selection for unbound work items"
	depends on DEBUG_KERNEL
	default n
	help
	  Workqueue used to implicitly guarantee that work items queued
	  without explicit CPU specified are put on the local CPU.  This
	  guarantee is no longer true and while local CPU is still
	  preferred work items may be put on foreign CPUs.  Kernel
	  parameter "workqueue.debug_force_rr_cpu" is added to force
	  round-robin CPU selection to flush out usages which depend on the
	  now broken guarantee.  This config option enables the debug
	  feature by default.  When enabled, memory and cache locality will
	  be impacted.

Also, it looks like another issue is breaking the dwc_otg driver.  Apparently, now we have hard_irqs and threaded_irqs.   The concept of threaded irqs make alot of sense, but I can't find much documentation on it.   If these threaded irqs have preemption, then it would eliminate much of the need for FIQ.

Electron752:
Sometimes, one has to pick battles and ARM64 is still experimental and unsupported. 

For what it's worth, I think this pull request should be accepted and if it breaks ARM64 at this time.  So be it.  I've fixed dwc_otg a few times, and I'll figure out another way to make it work.

Sorry.

anholt:
@popcornmix the branch now has a fix for compton/weston/etc. in firmwarekms mode.  The DSI orientation also matches firmware.  Anything blocking merge?

ktb83:
Confirmed. I say ship it. 
[raspbian_arm_rpi3_rpi-4.9.y-pr1813_notes.txt](https://github.com/raspberrypi/linux/files/751055/raspbian_arm_rpi3_rpi-4.9.y-pr1813_notes.txt)

Electron752:
For what it's worth, I tried arm64 on fkms.  Things seems to work quite well.  Even youtube works on my DSI. dwc_otg works just fine, even with the round robin hack.

Is there any reason this pull request is being held?

Electron752:
fkms also works great through HDMI.  I'm running Raspbian BTW. It's just too bad this doesn't have audio.

ktb83:
I've been using analog out. It works pretty well with both fkms and kms. With kms, it seems I have to boot with HDMI connected in addition to DSI in order to get sound. Booting kms with DSI connected alone seems to result in no analog audio out. Maybe the driver doesn't load, I didn't check.

Electron752:
If we're talking about 64 bit, I can cherry-pick the vchiq changes + bcm2835-audio changes from 4.10 into Erick's pull request and see what the issue is.  It's just too bad vchiq's io32 wrapper isn't quite ready as well.  I started cleaning it up, but didn't finish.

I also noticed the kernel alternatives on Raspbian, it seems to me 64 bit could be an alternative for the kernel.  Especially if the deprecated instruction logging issue was disabled.

Either way, I'm not sure 64 bit should hold up progress on this pull request.   It's all just a thought.

Electron752:
So, I was able to get HDMI to work with both fkms and kms.  I was able to get the DSI panel to work with both kms and fkms and the headphones without any HDMI connected at all.

One issue is that the headphones have a very large amount of static. Especially when nothing else is being played.  I remember seeing a if (silence) statement in bcm2835-vchiq.c so maybe that is broken.

I will say though that as much as I would like to get arm64 added to this pull request, I don't think I'm willing to do it with the state that Eric left it in.  It's kind of a mess.

Perhaps this pull request should just be accepted or rejected, your choice and I can open a new request later.  Or perhaps I can just add the changes to a different version.

ktb83:
> One issue is that the headphones have a very large amount of static

True, but audio_pwm_mode=2 seems to work well for me.

popcornmix:
With 4.9 kernel without this PR (e.g. BRANCH=next rpi-update) and compton disabled I can use fkms driver and run neverball. Here is dmesg output
http://paste.debian.net/912752/

If I add this PR and run the same case I get complaints in dmesg log but it appears to run:
http://paste.debian.net/912753/

If I now enable compton which I believe should work I just a hang with desktop wallpaper and blank taskbar.
http://paste.debian.net/912754/

ktb83:
@popcornmix Have you tried playing around with the default Compton config?
https://github.com/raspberrypi/linux/pull/1813#issuecomment-275981625

popcornmix:
> @popcornmix Have you tried playing around with the default Compton config?

No, but it seems to work fine with 4.4 kernel, so I'd expect 4.9 to work the same config.

ktb83:
Hmm. I don't see that error in dmesg (even if I kill -9 the compton process), but I'm still on 4.9.6-v7+. neverball seems to work very well for me on DSI with rpi-4.9.y+pr1813, Compton enabled (or disabled) with default config:
```
pi        1872  2.1  2.7  63348 24708 ?        S    07:19   0:51 compton --backend glx --unredir-if-possible --glx-swap-method buffer-age --glx-no-stencil --paint-on-overlay --vsync opengl
```

 and this config.txt: 
[raspbian_arm_rpi3_rpi-4.9.y-pr1813_vc4-fkms-v3d_config.txt](https://github.com/raspberrypi/linux/files/752334/raspbian_arm_rpi3_rpi-4.9.y-pr1813_vc4-fkms-v3d_config.txt)

OpenArena appears to be playable too.

Electron752:
For what it's worth, I think this pull request should be applied.  I've been testing, and fkms seems to work quite well with debian sid arm64 with absolutely no Raspbian userland components installed.   It works quite well with the open source Mesa 13, and I'm seeing that glxgears is getting 100+ fps now.  Everything is awsome(except for audio and camera).

I think fkms strikes a very good balance between open source and closed source.  Most of the interesting stuff is still happening in Mesa which is open source, just a small portion of the firmware is being used. I'm not sure regular kms will ever work great...

I was able to cleanup Anholt's branch locally.  I would open a cleaned up pull request, but I don't want my name mixed with Anholt's.

To fix all I did was:
git init linux-temp
git fetch https://www.github.com/Anholt/linux rpi-4.9.y:rpi-4.9.y-anholt
git fetch https://www.github.com/raspberrypi/linux rpi-4.9.y:rpi-4.9.y-offical
git checkout rpi-4.9.y-anholt
git pull --rebase ./ rpi-4.9.y-offical

This cleans up his branch and everything is awsome(except for camera and audio).

Now I'm locally cherry-picking the 4.10 changes to make vchiq and audio work with 64 bit.  I can provide the list of changes if needed.  I also know that the know how is available to make vchiq work with 32 bit userland.  But I'm wondering if perhaps a better approach would be to modify the userland tree to use proper opengl APIs rather then calling vchiq directly.

Either way, Raspbian has never been a very high priority for me.

Electron752:
@ktb83 

Interestingly, removing "audio_pwm_mode=2" and just using the default works quite well with my $10 cheap Sony headphones.  Maybe you are connecting the analog to a big amplifier?

ktb83:
I agree about fkms. It's been running solid for me with uptime of over a day now.

I'd be interested in your changes. Or I can just figure out how to work them in again.

I like this idea -- "modify the userland tree to use proper opengl APIs rather then calling vchiq directly." That sounds like it could be a lot of work, but I feel like it is probably worth it in the long run.

I just have the Pi hooked up to a simple pair of JBL speakers which came with an old Compaq desktop. There is a small amp in at least one of them. They're pretty good sounding for what they are. It's possible I haven't even tried to not use that setting yet on 32-bit Jessie. I'll give it a shot.

Electron752:
Here are the changes:
7a29a391ec8e615fab4d37add5708a4cc76421a7 - staging: vc04_services: fix debugfs write functions
35b7ebda57affcfd3616d39d5a727a4495b31123 - staging: vc04_services: Convert dsb() to dsb(sy)
e64568b8ea6c04e747e432c17ce2452652075216 -  staging: vc04_services: Change fields of page list structure to fixed length
24a4262afb10907fce3cdbc3ae336fcf4cdaece5 - staging: vc04_services: Convert pointers in shared state to offsets
8e8e2e36f77ab074016f79ac63b7255f645da3ed - staging: vc04_services: Handle conversion from VCHIQ_SERVICE_HANDLE_T to VCHI_SERVICE_HANDLE_T
f9bee6dd24addfa00c2c8d50c25b73efbfbb28ba - staging: vc04_services: Add casts to remove warnings in vchiq_core.c
cf9caf1929882b66922aee698e99e6c8f357bee5 - staging: vc04_services: Replace dmac_map_area with dmac_map_sg
dc5424ae0be4409019bafc303720d8a33f5b23fe - staging: vc04_services: call sg_init_table to init scatterlist
c01cc53d4eb81a626ca289d784143f22c2fe63ab - staging: vc04_services: setup DMA and coherent mask
2ea15699281e81960cc4daadff4d325a1c94bf09 - staging: vc04_services: parse_rx_slots() - Fix compiler warning
4807f2c0e684e907c501cb96049809d7a957dbc2 - staging: vc04_services: add vchiq_pagelist_info structure
696b182bc513a7c74e02a9bf9dcde2a64f9d3470 - ARM64/SND_BCM2835: Port it to arm64
4957f20f307b38c730400497985bf1e8b78d22a9 - ARM64: Enable HDMI audio and vc04_services in bcmrpi3_defconfig

Electron752:
Looking back, I think some of the vchiq/vc04_services could have been done differently.  Say perhaps converting to the DMA/SG APIs may have been overkill.  Especially when I hear they aren't really implemented very well in core linux.

Also, for the sound card change the callback pointer could have been removed since it always points to the same function rather then adding all the fancy casts.

Electron752:
BTW, I was looking at the issues from pastbin.  Here is an interesting snipit which may be causing some of the issues.

Why is this interrupt firing without the SMICS_INTERRUPTS flags being set?  Is this IRQ shared with something?

static irqreturn_t vc4_crtc_irq_handler(int irq, void *data)
{
	struct vc4_crtc *vc4_crtc = data;
	u32 stat = readl(vc4_crtc->regs + SMICS);
	irqreturn_t ret = IRQ_NONE;

	if (stat & SMICS_INTERRUPTS) {
		writel(0, vc4_crtc->regs + SMICS);
		drm_crtc_handle_vblank(&vc4_crtc->base);
		vc4_crtc_handle_page_flip(vc4_crtc);
		ret = IRQ_HANDLED;
	}

	return ret;
}

Electron752:
I do now have some hints as to what is causing the weird hangs.  At least with fkms.   I haven't isolated it completely, but it appears to be related to very complex issues with the position of the hardware cursor on the screen.

I don't totally understand the issue or how to fix it, but I think it may be better to fix it in the firmware rather then in the driver.

Anyway, that's all I really have to add to this issue or pull request...

Electron752:
For what it worth, I tracked down more of the issue. It's "helpers get in the way.".  Here is a temp fix I tested on my local machine.  The driver simply needs to count vblanks instead of fancy timestamps and disable a workaround for a helper in waiting for the vblank.

When I do this, I don't see anymore weird hangs and everything is alot faster now.  glxgears gets a solid 60fps.

[diff.txt](https://github.com/raspberrypi/linux/files/755228/diff.txt)

schnitzeltony:
In a previous post you wrote

> glxgears is getting 100+ fps

I am getting confused a bit. 

Why this offtopic nitpick? I build kernel with yocto and common benchmark values give me hints if something goes wrong. What values do you accept as good?

anholt:
@popcornmix huh, that sounds like the behavior I'd expect without the most recent commit in the branch.  I'll try out compton today.

anholt:
Updated with the three commits I sent to 4.4.y, and rebased on current 4.9.  Confirmed that without the branch compton fails in fkms, and with the branch it's fine.

ktb83:
It works well for me. Same report as last time.

Kodi 18 works well for me with this config: 
[raspbian_arm_rpi3_rpi-4.9.y-pr1813_vc4-fkms-v3d_kodi_config.txt](https://github.com/raspberrypi/linux/files/765268/raspbian_arm_rpi3_rpi-4.9.y-pr1813_vc4-fkms-v3d_kodi_config.txt)

popcornmix:
I think I've found the cause of the kernel panic/neverball crash.

With fkms driver if HDMI cable is unplugged LCD display works fine. No dmesg complaints, neverball runs.
With fkms driver if HDMI cable is plugged in LCD display works fine for desktop. HDMI is blank as expected. But we get the dmesg complaints and neverball crashes on starting.

Does that help reproduce?

(I'm testing with latest version of this PR).

ktb83:
All my tests have been performed with HDMI and DSI plugged in. However, the HDMI is connected through an HDMI switcher which is switched to a different input (Mac Mini) while using fkms. I'll see if I can reproduce what you're seeing with both DSI and HDMI actually connected.

ktb83:
I tested vc4-fkms-v3d two other ways with my setup.

1. DSI connected, HDMI connected through HDMI switcher (RadioShack 2-In/1-Out Mechanical HDMI Selector Switch) to a Samsung LCD TV (LN22B460B2DXZA, HDMI 1.3) and Pi input selected on the HDMI switcher. This leads to some interesting Neverball behavior in that the intro Neverball screen loads but I'm unable to click or select any of the buttons in the neverball GUI by mouse or keyboard. Neverball is completely unresponsive and requires killing it.

2. DSI connected, HDMI directly connected to the Samsung LCD TV. My experience in this situation is that Neverball works perfectly well for me. I don't experience any issues just like when I only have DSI actually connected.

I see no unusual complaints in dmesg in either scenario. Perhaps I need to set a debug flag/variable some place to see what you're seeing?

I don't know what to think.

Electron752:
For whatever it's worth, I have a simple fix for the weird spinning cursor display hang issue with KMS and FKMS.   I was wondering if anybody would be interested in testing that as well.

https://github.com/raspberrypi/linux/pull/1840

BTW, I don't really understand what is holding up this pull request.  It is progress, perhaps it can be improved with additional requests?

Electron752:
I'm wondering if these topics are related to this pull request?

https://github.com/raspberrypi/userland/issues/375
https://github.com/raspberrypi/linux/pull/1840
https://github.com/raspberrypi/linux/issues/1843

popcornmix:
> BTW, I don't really understand what is holding up this pull request. It is progress, perhaps it can be improved with additional requests?

Lets see what @anholt says about the fkms with DSI display and HDMI attached issue.
If it is something he can reproduce and resolve quickly we'll wait for the fix before merging.

If it is hard to reproduce or tricky to fix then we can accept this PR and get the fix in a subsequent PR, but there is a risk of wasting users' time who run into the issue.

anholt:
@popcornmix Is your bug a regression in this PR, though?  If it's not a regression, it would be really nice to get merged so that we can have it as a baseline (and avoid more rebases of the series).  I may or may not get to testing for your bug today.

popcornmix:
No problem. A rebase was needed so I've merged the patches from your tree directly.
I'll push out a BRANCH=next rpi-update kernel later today for testing.

anholt:
@popcornmix the dmesg warn in drm_irq.c at boot is a race condition of DRM telling us that a vblank interrupt happened before vblank interrupts should have been enabled, which happens because we don't currently have a way to mask out the hacky interrupt we're using in fkms mode.

neverball displays fine for me, though it doesn't process touches as input in its menus or mouse locations in any useful way in game.  input shows up slightly weird (but fine afaict) in xev, and this doesn't seem vc4-related at all.

popcornmix:
@anholt are you running on DSI LCD using fkms driver with HDMI cable connected?
That was the setup that caused neverball to hang after a few seconds of running.
It works fine without that combination.

anholt:
Yes, that's what I was using.

popcornmix:
@ktb83 seemed to have a similar issue but only when going through a hdmi switch.
I'm not quite sure what a mechanical HDMI switch would actually change - I would have thought it would look just like a direct connection, but maybe it alters hotplug or (less likely) edid behaviour.

My setup was using a DELL monitor connected through HDMI. The firmware does choose a CEA mode:
```
pi@domnfs:~ $ tvservice  -s
state 0x12000a [HDMI CEA (16) RGB lim 16:9], 1920x1080 @ 60.00Hz, progressive
```
Possibly something hotplug or edid related causes the issue?

ktb83:
FYI, I am no longer able to reproduce any Neverball hanging issues on BRANCH=next rpi-update.

[rpi3_rpi-4.9.y_vc4-fkms-v3d_2017-02-15_test-notes.txt](https://github.com/raspberrypi/linux/files/777542/rpi3_rpi-4.9.y_vc4-fkms-v3d_2017-02-15_test-notes.txt)

PR #1826, opened 2017-02-02
rpi-4.4.y DRM/clock fixes

anholt:
Here are a few cherry-picked commits for 4.4.y.  The most important is the firmwarekms pageflipping fix, which @spl237 and others had reported bugs on.  The updated "don't rate change PLLs" patch should fix some HDMI modesetting.  I also picked a couple of changes for VEC's clocking, but haven't backported VEC yet.


popcornmix:
Seems fine on a quick test.
vc4-kms-v3d, vc4-fkms-v3d and normal firmware desktop appear to work.
We can now run fkms with or without compton.

PR #1830, opened 2017-02-06
char: broadcom : Remove unnecessary dev_info output

JamesH65:
The open function was spamming syslog every time
called, so have removed call completely.


popcornmix:
I'm happy with this.
It could be changed to dev_dbg which should be silent by default, but can be enabled with ``#define DEBUG 1``. But as there aren't any other useful debugging messages (e.g. in the release function) then it wouldn't be very useful.

@pelwell okay to merge?

PR #1835, opened 2017-02-09
rpi-4.4.y: VC4 crash fix, performance improvement, and sanity checking

anholt:
Three little commits for 4.4.  One should make the system reliable when CMA OOM happens -- this has plagued us forever, and I just found a testcase for reliably reproducing it.  Also a little performance fix, and a sanity check that users don't mess up the CMA settings in a way that would cause GPU hangs/system memory corruption.


popcornmix:
Thanks.

PR #1837, opened 2017-02-11
ARM64: Get most hello_pi examples to work on Raspbian

Electron752:
This change gets most of the hello_pi examples to work on Raspbian with a 64 bit kernel.  The hello_pi examples where built with the compiler that is included in Raspbian.

Some of the examples I tried where:
hello_teapot
hello_tiger
hello_videocube

I plan to submit this to upstream eventually, but ARM 64 support in upstream is getting too controversial at this time and maybe things need to cool a bit.  And perhaps it's best if these changes get tested here before sending them upstream.

I do sometimes see thermal and stability issues when running these demos, but setting the max arm freq to 1000MHz seems to make them go away.


popcornmix:
Ping @pelwell @6by9

Electron752:
BTW, I also have a patch to force deprecated arm32 instructions to hardware emulation.  Normallly when those instructions are enabled, the kernel traps these instructions just to log a message in dmesg.   This really slows down Raspbian which is bad since the A53 in the RPI3 is perfectly capable of running those instructions in hardware.

I didn't include the patch, but when it is included Raspbian zips along with the 64 bit kernel.  

The standard upstream source also has a sysctl option to enable the instructions in hardware, but it would require modify the boot sequence of Raspbian.

pelwell:
Although conceptually simple, a set of thunks like that on top of the 64-bit ioctls is going to penalise the 32-bit userspace with another kernel->user->kernel translation. Since we have no plan to switch to 64-bit userspace, is that a penalty we want to pay?

The alternative is to refactor the ioctl methods into user->kernel translation plus implementation, which should have less of a performance impact.

Electron752:
No, this patch doesn't do a kernel->user->kernel translation.  It simply builds the 64 bit version of the data structures on the user mode stack and then calls the native ioctl directly.

I would personally like to see the code refactored, but getting a change like that in upstream even in stagging is very, very difficult.  You have to have a gazillion patches to be taken seriously.

If you want I can run numbers with vchiq_test and see if I can tell a significant different between a 32 bit kernel running Raspbian and a 64 bit kernel running Raspbian.

Either way, I would still keep a 64 bit kernel experimental and not for general use.

pelwell:
> No, this patch doesn't do a kernel->user->kernel translation. I simply builds the 64 bit version of the data structures on the user mode stack and then calls the native ioctl directly.

That's what I mean. For each call you have to allocate some userspace memory for the new args, read from the old args and reformat into the new, then call into the 64-bit ioctl which goes through the user->kernel translation again.

Electron752:
Yes, but the data itself doesn't go through that translation.  Just the informational structures that describe the data.  Like I said, I can run some benchmarks with vchiq_test if it would help.

On a related topic, I know what it needed to get Mathematica to run.  It simply takes faking the reported processor version to userland when running a 32 bit binary.   I have no idea why this isn't done by default in upstream and nobody could give me a good answer why it isn't done when most of the other platforms do that.

Of course, you guys would have to verify that faking the architecture doesn't violate any license agreements for Mathematica.

drivers/staging/vc04_services/interface/vchiq_arm/vchiq_arm.c
```diff
@@ -47,6 +47,7 @@
 #include <linux/list.h>
 #include <linux/of.h>
 #include <linux/platform_device.h>
+#include <linux/compat.h>
```
6by9, line 50: Make the include conditional on CONFIG_COMPAT

6by9:
> The alternative is to refactor the ioctl methods into user->kernel translation plus implementation, which should have less of a performance impact.

I agree that that is the better approach, even if it is more work to get it upstreamed.
Applying it downstream to "test it" is just going to mean that downstream diverges from upstream.

drivers/staging/vc04_services/interface/vchiq_arm/vchiq_arm.c
Electron752: I did do that in upstream, then got screamed at that I shouldn't have the CONFIG_COMPAT condition.

Electron752:
I was planning to submit the exact same patch to upstream soon.  Things just need to cool down a bit.

The wrapper approach is what anholt recommended in the first place.  Apparently, it's fast enough for the DRM graphics core.

Electron752:
As a side note, I'm thinking that perhaps upstream doesn't really want 32 bit compatibility for some reason.  I have a feeling that no matter what I submit regardless if it's a gazillion patches moving things a line at a time, it will still get rejected.

The general issue with upstream is also that if you fix ANYTHING in a section of code, you have to then fix EVERYTHING.

drivers/staging/vc04_services/interface/vchiq_arm/vchiq_arm.c
notro: Look inside compat.h: ```#ifdef CONFIG_COMPAT```

Electron752:
I'm going to be away from the computer for a bit.  If anybody wants to actually test this on their own to see if it works,  It might be good to test with the deprecated instruction patch as well.  That makes a huge difference in Raspbian performance.

I'm attaching the patch file for deprecated instructions.
[0001-ARM64-Force-hardware-emulation-of-deprecated-instruc.patch.txt](https://github.com/raspberrypi/linux/files/768818/0001-ARM64-Force-hardware-emulation-of-deprecated-instruc.patch.txt)

popcornmix:
I'd be okay with the patch to disable sw emulation of deprecated instructions.
I understand upstream may say running armv6 code on an armv8 doesn't make much sense generally, but I think with raspbian it is actually a use case people might want to run with.
A simple benchmark that shows the difference would be useful.

Electron752:
Graphics performance seems to be affected the most, so I ran the gtkperf from Raspbian.

So without the patch gtkperf -a took 73.69s.
With the patch, gtkperf -a took 15.47s

So forcing hw instructions is a 4.76 times performance improvement.

Number Without Deprecated Ins Patch:
GtkPerf 0.40 - Starting testing: Sat Feb 11 19:57:43 2017

GtkEntry - time:  0.02
GtkComboBox - time: 11.99
GtkComboBoxEntry - time:  7.37
GtkSpinButton - time:  1.46
GtkProgressBar - time:  0.78
GtkToggleButton - time:  1.75

GtkCheckButton - time:  1.53
GtkRadioButton - time:  1.89
GtkTextView - Add text - time:  3.01
GtkTextView - Scroll - time:  3.35
GtkDrawingArea - Lines - time: 16.75
GtkDrawingArea - Circles - time: 16.05
GtkDrawingArea - Text - time:  7.32
GtkDrawingArea - Pixbufs - time:  0.34
 --- 
Total time: 73.69

Quitting..

Number With Deprecated Ins Patch:
GtkPerf 0.40 - Starting testing: Sat Feb 11 20:02:40 2017

GtkEntry - time:  0.00
GtkComboBox - time:  2.60
GtkComboBoxEntry - time:  1.67
GtkSpinButton - time:  0.43
GtkProgressBar - time:  0.28
GtkToggleButton - time:  0.45
GtkCheckButton - time:  0.22
GtkRadioButton - time:  0.26
GtkTextView - Add text - time:  1.26
GtkTextView - Scroll - time:  0.63
GtkDrawingArea - Lines - time:  1.88
GtkDrawingArea - Circles - time:  4.24
GtkDrawingArea - Text - time:  1.28
GtkDrawingArea - Pixbufs - time:  0.25
 --- 
Total time: 15.47

Quitting..

drivers/staging/vc04_services/interface/vchiq_arm/vchiq_arm.c
6by9: Fair enough.

Electron752:
Here are some numbers from vchiq_test.  As you can see, the functional test is significantly slower for 64 bit on Raspbian, but the ping test is almost not affected.  I suspect what is going on is with the compatibility wrapper, the await_completion ioctl is only returning 1 message at a time.  That's increasing the number of user to kernel transitions.

Let me know if you want this as is+ins patch, if I need to improve the await_completion ioctl, or if I'm completely barking up the wrong tree.

32bit(default kernel7.img from Raspbian) - time vchiq_test -f 15
real	0m7.686s
user	0m1.750s
sys	0m2.560s

32bit(default kernel7.img from Raspbian) - time vchiq_test -p 10
real	0m12.908s
user	0m0.060s
sys	0m0.300s

64bit(current bcmrpi3_defconfig 1000Hz timer, forced preemption) - time vchiq_test -f 15
real	0m21.270s
user	0m1.571s
sys	0m14.906s

64bit(current bcmrpi3_defconfig 1000Hz timer, forced preemption) - time vchiq_test -p 10
real	0m13.074s
user	0m0.035s
sys	0m0.556s

64bit(alt .config 250Hz timer, voluntary preemption) - time vchiq_test -f 15
real	0m14.423s
user	0m1.452s
sys	0m9.516s

64bit(alt .config 250Hz timer, voluntary preemption)) - time vchiq_test -p 10
real	0m12.940s
user	0m0.028s
sys	0m0.388s

Electron752:
I rebuilt 4.10.y in arm32 mode and it is slower in general then 4.4.  Here are the timing numbers from 4.10 with a 32 bit kernel for comparison.

32 bit(bcm2709_defconfig) vchiq_test -f 15
real	0m11.767s
user	0m1.410s
sys	0m6.580s

32 bit(bcm2709_defconfig) vchiq_test -p 10
real	0m13.121s
user	0m0.020s
sys	0m0.510s

Electron752:
I'm thinking that maybe these vchiq based wrappers could be a short term solution. I noticed that the hello_pi examples don't work with kms or fkms.  I'm thinking that longer term, what needs to happen is make the old broadcom libraries a selectable thin wrapper on top of the mesa based opengl library if kms or fkms is used.  Otherwise use vchiq.

https://github.com/raspberrypi/userland/issues/375

Electron752:
I noticed that the deprecated instruction patch has been applied, but the other part hasn't.

Even though the on top of solution is not the optimal performance, I went ahead and submitted this patch for inclusion in upstream.  I have no idea if it will be accepted.  I don't think it would hurt anything to have it.

Electron752:
This has been merged upstream now, so I have no need to have it pulled here.

Thanks.

PR #36, opened 2012-06-12
sdhci-bcm2708 performance, compatibility and reliability improvements

grigorig:
This branch fixes compatibility with many cards, and also improves read/write performance, in most cases 4-5x. Many cards that did not work reliably before work much better with those changes. As already discussed in length at http://www.raspberrypi.org/phpBB3/viewtopic.php?f=63&t=5057.

I know bootc's 3.2 kernel is going to become the Pi mainline soon-ish, but in the meantime I think it's still useful to have this in 3.1.

PR #1840, opened 2017-02-11
VC4: KMS/FKMS Fix weird hang with frozen display and spining cursor

Electron752:
The issue with the spinning cursor is that most distros(Debian Gnome/XFCE4) can flood the driver with cursor updates.  The userland software never expects the cursor updates to be synchronized with the vblank.  If they are synchronized, then the display will hang for very long periods of time.

The workaround on both kms and fkms is to update the cursor without any vblank synchronization.

I tested both KMS and FKMS on Debian sid with both the XFCE4 desktop and the Gnome desktop.


popcornmix:
As @anholt maintains these files, I'll only accept patches if he accepts them. I don't want to diverge from his versions of files.
Probably better to report issues to https://github.com/anholt/linux/issues and we can then get fixes from @anholt.

Electron752:
I understand your point of view and it makes sense.

I reported the issue of the hangs awhile back on his page and didn't get much response.   I also directly e-mailed him the KMS fix yesterday and he said it was my version of linux that's broken although I'm not sure what the recommended alternative to Raspbian is since both XFCE4 and Gnome are very broken.

I think their is a school of thought that downstream is only for Raspbian and upstream is for other distributions but only when booting with u-boot.  That's a bit unfortunate since some of the enhancements in downstream work great with other distributions.

Perhaps we can keep the issue active for a few days or so and maybe someone else can chime in to confirm the issue.  Or perhaps @anholt can respond as to why I'm seeing the weird hangs and why the patch I've provided won't work.

ED6E0F17:
I may have hit this issue on Debian Stretch (arm v7), but I do not intend to test the patch until the Anholt backports have been merged.

Electron752:
I would like to see the backports merged as well.  Running KMS on the LCD panel is cool.  I really have no idea why that merge is being delayed.  I don't understand why having a few issues is holding it up, especially since a few similar version was merged into 4.4.

ktb83:
@Electron752  Wow, very interesting. I'm definitely taking out the 64 bit build again to give this a shot. You really got Gnome 3.x running (even if it's rough)?

Electron752:
With the patch, Gnome isn't really rough at all.  Alot of things seem to run faster with the patch.  I think it's because the user land software will sometimes flood the driver with cursor updates.  Waiting for the vblank on every cursor update is way too slow.

I have no idea why this issue isn't being seen on Raspbian.

Electron752:
As a side point, the bottle neck with running that kind of software appears to be more disk access speed and amount of RAM in the system.  I see things will start filling up the swap file even with casual use.

ktb83:
Confirmed. This makes Xfce usable for me. I do get a warning from line 198 in vc4_firmware_kms.c:

lines 198-199:
```
	WARN_ON_ONCE(fb->pitches[0] != state->crtc_w * 4);
	WARN_ON_ONCE(fb->bits_per_pixel != 32);
```

ED6E0F17:
With a very quick test of the Anholt backports and this, everything seems to be working well. I do get:
````
[  119.211359] ------------[ cut here ]------------
[  119.211439] WARNING: CPU: 2 PID: 454 at /linux/drivers/gpu/drm/vc4/vc4_plane.c:721 vc4_plane_async_set_fb+0xa4/0xa8
[  119.211456] Modules linked in:
[  119.211491] CPU: 2 PID: 454 Comm: Xorg Not tainted 4.9.9-armv7+ #36
[  119.211507] Hardware name: BCM2835
[  119.211552] [<80110a48>] (unwind_backtrace) from [<8010cacc>] (show_stack+0x20/0x24)
[  119.211584] [<8010cacc>] (show_stack) from [<804bee6c>] (dump_stack+0xc4/0x108)
[  119.211622] [<804bee6c>] (dump_stack) from [<80121438>] (__warn+0xf8/0x110)
[  119.211653] [<80121438>] (__warn) from [<80121520>] (warn_slowpath_null+0x30/0x38)
[  119.211683] [<80121520>] (warn_slowpath_null) from [<805ad628>] (vc4_plane_async_set_fb+0xa4/0xa8)
[  119.211714] [<805ad628>] (vc4_plane_async_set_fb) from [<805ad794>] (vc4_update_plane+0x168/0x170)
[  119.211745] [<805ad794>] (vc4_update_plane) from [<8059bd8c>] (__setplane_internal+0x190/0x238)
[  119.211775] [<8059bd8c>] (__setplane_internal) from [<8059bf54>] (drm_mode_cursor_universal+0x120/0x1b4)
[  119.211805] [<8059bf54>] (drm_mode_cursor_universal) from [<8059c074>] (drm_mode_cursor_common+0x8c/0x194)
[  119.211836] [<8059c074>] (drm_mode_cursor_common) from [<8059c740>] (drm_mode_cursor2_ioctl+0x18/0x1c)
[  119.211868] [<8059c740>] (drm_mode_cursor2_ioctl) from [<8057e718>] (drm_ioctl+0x214/0x478)
[  119.211905] [<8057e718>] (drm_ioctl) from [<802ae0f8>] (do_vfs_ioctl+0xac/0x810)
[  119.211936] [<802ae0f8>] (do_vfs_ioctl) from [<802ae8d8>] (SyS_ioctl+0x7c/0x8c)
[  119.211965] [<802ae8d8>] (SyS_ioctl) from [<80108480>] (ret_fast_syscall+0x0/0x1c)
[  119.211984] ---[ end trace 002118488d65a68a ]---

anholt (commented):
While I think currently we never have a single update that does both cursor BO and position, I think I'd like us to be protected from that, so:

- only check fb  != plane_fb, rather than position
- Drop the return 0.

Also, we need to make sure the new FB gets referenced by the plane state, so it sticks on future plane updates.  drm_atomic_set_fb_for_plane(plane->state, fb); just before the vc4_plane_async_set_fb() call should do that. 

With those changes, could you submit the patch to the DRI mailing list?  I'd love to have you get credit for this patch upstream.

Electron752:
Sure, I can probably submit a patch.  It's probably better that it goes upstream anyway.

It might take me a day or two to have it ready.

popcornmix:
Ping me here once upstream has accepted patch and I'll merge/cherry-pick it to our tree.

Electron752:
@anholt / @popcornmix : The KMS fix has been accepted in upstream, but it's going to be awhile for it to propagate through all the branches.  It's currently in drm-misc-next, so I'm guessing it won't be available until 4.12 at least.

Any thoughts on what to do about fkms since it has the same issue?  fkms is downstream only so I don't have anything to submit to upstream. I've been told it's the xorg/xserver that's animating the spinning cursor, so I don't think this is an issue with Raspbian at this time.  But I'm not sure what will happen when/if things move to a newer debian base.

Issue #903, opened 2015-03-20
USB isosynchronous - zeroed input buffer returned 


TheTechnobear:
Hi,
Ive got an issue with a USB device (1.1) that uses iso transfers.
all completions back from the device are filled with zeros, even though they are marked as successful completions. It appears the data is not being copied back into the buffer by the kernel.

some background:
Im using libusb,  and have exactly the same code, running for 2 devices, a usb 1.1device and a usb 2.0 device. both devices work on a debian/pc fine, and the usb 2.0 device works fine on the PI2, its only the 1.1 device that has this issue of a 'zeroed buffer'

observations:
- when libusb debug is on, everything is shown as successful there are no errors
- Ive done a usbmon trace on the 1.1 device on the PI2, and can see that the data is being sent from the device correctly.

usually no flags are present in cmdline.txt for this, but Ive tried the various dwc_otg flags to disable/enable fiq/fiq_fsm etc etc, and none seem to make any difference. 

Ive run out to 'things to try', but i need to get this device working :( 

Given the above, it seems likely its a kernel issue, is there something I can do to get more information to the likely cause, or possible fix?
(Id be happy to test any ideas anyone has... )

platform: PI2: 3.18.9-v7
I'm using a separate powered MTT hub, and Ive tried 3 different hubs, also the same hubs work fine on the debian PC installation.

P33M:
Hi.

Some more logs would be useful.
- Please post a `lsusb -v` with the device attached.
- Please post the output of dmesg after running your program for communicating with the device (with default settings of `/boot/cmdline.txt`)?

By what metric are you measuring the fact that the completions are "successful"?

Note that historically it has been virtually impossible to have working/reliable Isochronous transports from a USB1.1 device. For each frame (1ms) of data, you have to precisely and correctly schedule the start-split and complete-split transfers (of which there are multiple complete-split transfers) or the translating hub simply drops the data. The fiq_fsm code substantially resolved most of the issues surrounding this set of devices but there are certain edge cases that may still result in failure.

TheTechnobear:
completion success... libusb reports status complete on both the overall transfer and on the individual packets. also usbmon status field also shows the iso completion as been 'successful'. 

Im not quite sure what you mean by "its been virtually impossible to have reliable iso transports" on 1.1 devices,this device and code is working fine on Windows/Mac OSX and also other Linux distributions (the last of which is using exactly the same code as on the PI).

[...]

PR #1845, opened 2017-02-13
dwc_otg: fix summarize urb->actual_length for isochronous transfers

mcerveny:
Kernel does not copy input data of ISO transfers to userspace
if actual_length is set only in ISO transfers and not summarized
in urb->actual_length. Fixes raspberrypi/linux#903


popcornmix:
Many thanks @mcerveny for getting to the bottom of this issue.
Based on discussion in #903 and a quick build/boot test I'll merge this and get an rpi-update kernel out.

PR #1846, opened 2017-02-13
clk-bcm2835: Mark used PLLs and dividers CRITICAL

pelwell:
PiZero has no ethernet controller, and therefore doesn't initialise GP1. As a result the Linux clock infrastructure thinks that PLLD and the PLLD_PER divider can be disabled when the current user (in this case an I2S audio card) is closed, causing the display to switch off and, eventually, the peripheral bus to lock up.

This patch looks for enabled PLL dividers at boot time and marks them and their PLL parents as CRITICAL, which fixes the immediate problem. I had thought it might be nicer to put the critical clock information into Device Tree, but use of the `clock-critical` property is deprecated.

@anholt @msperl @clivem Are you aware of better solutions to this problem?


popcornmix:
Original issue was reported here: https://forum.libreelec.tv/thread-3848.html
Ping @HiassofT @shawaj

msperl:
I am just wondering: what is then still using PLLD besides I2S?
If the display needs it then there should be a driver claiming the relevant clock.
The critical is just a workarround.

popcornmix:
The firmware uses PLLD for almost all peripherals.
GPCLK (used for ethernet clock), DSI, DPI, SDRAM (when changing frequencies), HDMI state machine, camera, uart, PWM audio.
Turning it off or messing with it is fatal to the firmware.

shawaj:
Thanks for the heads up. If you need any more tests performing myself and @francesco-vannini would be happy to help :-)

Got plenty of Pi Zero kicking around here!

popcornmix:
@shawaj Tonight's @MilhouseVH build should contain this patch, so would be useful if you can confirm that works with JustBoom on Pi0.
Assuming testing is positive and we don't get a suggestion for a better fix then we'll get a fix into next LibreELEC release.

shawaj:
OK cool - where do we find the Milhouse build?

Will that be posted [here](http://forum.kodi.tv/showthread.php?tid=298461)?

popcornmix:
> Will that be posted here?

Yes.

pelwell:
Thank you @HiassofT for the detailed instructions to reproduce the lock-up - it saved a lot of time.

shawaj:
OK great, we'll test it first thing tomorrow

anholt:
We don't want PLLH_AUX/PIX or PLLD_DSI* to be critical when vc4 is enabled -- we should be able to power them off and on at runtime.  Could you throw this under a check for a DT property that only the vc4-kms-v3d overlay sets?

For vc4-fkms-v3d or non-vc4 mode, I agree that we need to mark PLLH_AUX, PIX and PLLD_DSI* as critical.  I don't think it will end up keeping any other dividers forced on (it's not like we'd ever be able to power down PLLD_PER)

shawaj:
OK I've spoken to @francesco-vannini who has done some more testing today.

Tested with both 7.0.3 and 7.95.2 as the base OS and reproduced the problem. Updated these with Milhouse build 213 and the problem is fixed. 

So all looks good :-)

shawaj:
@pelwell @popcornmix let us know if you want us to test any other permutations of the latest build or anything like that.

pelwell:
I've rebased and included a second commit that adds a claim-clocks property to the vc4-kms-v3d overlay:
```
	fragment@19 {
		target = <&clocks>;
		__overlay__  {
			claim-clocks = <
				BCM2835_PLLD_DSI0
				BCM2835_PLLD_DSI1
				BCM2835_PLLH_AUX
				BCM2835_PLLH_PIX
			>;
		};
	};
```
This claims the named PLLs and dividers for the clk-bcm2835 driver, preventing them from being marked as critical.

FYI, the debug code I had in there during development gave the following output:
```
pll plla is critical
pll pllb is critical
pll pllc is critical
pll plld is critical
pll pllh is critical
pll divider plla_core is critical
pll divider pllb_arm is critical
pll divider pllc_core0 is critical
pll divider pllc_per is critical
pll divider plld_core is critical
pll divider plld_per is critical
pll divider pllh_rcal is critical
pll divider pllh_aux is critical***
pll divider pllh_pix is critical***
```
The asterisks mark lines that were omitted with the vc4-kms-v3d overlay.

@anholt I'm happy that this implementation works, but I wouldn't be surprised if there were a cleaner implementation.

Electron752:
Just wondering if this issue has been resolved.  It's been sitting around for awhile.

I'm like @msperl, why have a hack like this.   Perhaps a better solution is to have a real driver simply claim the clock, even if it's a dummy placeholder driver that doesn't actual do anything. The real driver could be the Raspberry PI firmware driver that talks to mbox, it could be the video driver, or perhaps it should be something new.

I'm still batting around the idea of having virtual buses and subsystems instead of platform drivers for some of the RPI driver that don't really talk to real hardware like the vchiq, bcm2835-camera, and the firmware based audio driver.  Perhaps this fits in with this?

Finally, I'm going through my head what RPI driver haven't been upstreamed yet and one of the big ones left is the frame buffer based video driver.  There may be cases where people don't want to use VC4 because it's an embedded system and don't need fancy graphics.  They just want the ability to get video for occasional debugging.     Is it true that absolutely no more fbdev driver are being accepted in upstream?  What about the TinyDRM framework that's in upstream now?

Electron752:
Actually spinning it another way, another alternative for the video would be to have VC4 itself run in a stripped down mode to act as a simple display driver.

@anholt or anybody, I'm curious why exactly does the VC4 driver load so late in the boot sequence and why does it aways seem to require alot of RAM?

And if VC4 can't be made to load early or without much RAM, why couldn't a VC4-Lite driver be built out of the same source directory?

shawaj:
@pelwell - did this get fixed somewhere else in the end then?

popcornmix:
The commits form here have been merged. See:
https://github.com/raspberrypi/linux/commits/rpi-4.9.y/drivers/clk/bcm/clk-bcm2835.c

pelwell:
I ended up merging essentially the same code as the two commits above but via the command line. See  https://github.com/raspberrypi/linux/commit/5c083e5fbcb48d9c50414650ac2f7ba5cdb25824 and https://github.com/raspberrypi/linux/commit/204050d0eafb565b68abf512710036c10ef1bd23 .

shawaj:
Oh awesome. Thanks :-)

PR #1854, opened 2017-02-23
Add ads1015 driver to config

scottellis:
Add the kernel module so the ads1015-overlay is usable.


pelwell:
I rewrote the configuration settings using savedefconfig (they weren't in quite the right place), rebased and pushed it by hand.

PR #1858, opened 2017-02-24
dmaengine: bcm2835: Fix cyclic DMA period splitting

HiassofT:
This patch got lost during upstreaming. I've resubmitted it and it should appear in future upstream kernels:
https://lkml.org/lkml/2017/2/20/506

For the time being it would be good to have this in the 4.9 and 4.10 trees - it was included in 4.4 as well.

Without that patch we get repeated clicks when playing audio on I2S cards (eg at 192 kHz)


pelwell:
I have no objection. @popcornmix ?

popcornmix:
Also cherry-picked to 4.10.

PR #1864, opened 2017-02-28
Add overlay for ads1115 ADCs

scottellis:
The ads1015 kernel driver supports ads1015 and ads1115 devices but different data rate tables are used internally based on the device. The data rate tables are used to calculate timeouts.

### DIFF
--- a/arch/arm/boot/dts/overlays/README
+++ b/arch/arm/boot/dts/overlays/README
@@ -196,17 +196,17 @@ Params: addr                    I2C bus address of device. Set based on how the
         http://www.ti.com/lit/ds/symlink/ads1015.pdf
 
 
-Name:  ads1115
-Info:  Texas Instruments ADS1115 ADC
-Load:  dtoverlay=ads1115,<param>[=<val>]
+Name:   ads1115
+Info:   Texas Instruments ADS1115 ADC
+Load:   dtoverlay=ads1115,<param>[=<val>]
 Params: addr                    I2C bus address of device. Set based on how the
                                 addr pin is wired. (default=0x48 assumes addr
                                 is pulled to GND)
         cha_enable              Enable virtual channel a.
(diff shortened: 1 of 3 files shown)


arch/arm/boot/dts/overlays/README
```diff
@@ -196,6 +196,28 @@ Params: addr                    I2C bus address of device. Set based on how the
         http://www.ti.com/lit/ds/symlink/ads1015.pdf
 
 
+Name:  ads1115
+Info:  Texas Instruments ADS1115 ADC
+Load:  dtoverlay=ads1115,<param>[=<val>]
```
pelwell: The preceding three lines need one extra space each to get the indent right.

arch/arm/boot/dts/overlays/README
```diff
@@ -196,6 +196,28 @@ Params: addr                    I2C bus address of device. Set based on how the
...
+Params: addr                    I2C bus address of device. Set based on how the
+                                addr pin is wired. (default=0x48 assumes addr
+                                is pulled to GND)
+        cha_enable              Enable virtual channel a.
+        cha_mux                 Set the input mux for virtual channel a.
```
pelwell: Is this what the overlay calls cha_cfg? They need to match.

arch/arm/boot/dts/overlays/README
```diff
@@ -196,6 +196,28 @@ Params: addr                    I2C bus address of device. Set based on how the
...
+                                (default=4 configures this channel for the
+                                voltage at A0 with respect to GND)
+        cha_rate                Set the datarate (samples/sec) for this channel.
```
pelwell: Another mismatch - rate<->datarate.

pelwell (changes requested):
Apart from the README nitpicks, this looks good.

pelwell (approved):

PR #1867, opened 2017-02-28
Add support for Fe-Pi audio sound card.

Fe-Pi:
Fe-Pi Audio Sound Card is based on NXP SGTL5000 codec.
Mechanical specification of the board is the same the Raspberry Pi Zero.
3.5mm jacks for Headphone/Mic, Line In, and Line Out.


inspector71:
Oddly enough, I first read about various Pi hardware in github entires like this. Would be handy to have a link to the device's webpage :) Not necessary by any means, but helpful. Kind of a odd but useful promotion place for hardware manufacturers :) I mean usually if it's mentioned here, the hardware is now supported in the Raspbian kernel, so that's always a less-annoying time to take a first look at hardware because you know if you like it, buy one, it's usable soon as it arrives :)

http://fe-pi.com/

P.S I realize this comment is both appropriate and inappropriate in terms of being out of usual context. I toolk a chance.
Disclaimer: I've no relationship with this OEM.

PR #1879, opened 2017-03-07
BCM270X_DT: Invert Pi3 power LED to match fw change

6by9:
Firmware expgpio driver reworked due to complaint over
hotplug detect.
Requires power LED to change sense as firmware is no longer
inverting the read value.


6by9:
@pelwell if gpio expander is reading the raw value for #1869, then the power LED has the wrong sense. 
Switch to not inverting.
This requires the associated firmware change.

pelwell:
Looks OK to me.

The sense of the LED only makes a difference to an output. Since the standard trigger is "input", getting 
out-of-sync with the firmware will only affect those people who override the trigger.

pelwell:
This is now in the latest rpi-update firmware release.

PR #1883, opened 2017-03-08
Add support for the AudioInjector Octo soundcard

flatmax:
This patch set consists of two patches.
The first adds support for the Octo soundcard.
The second allows the second i2s channel register to be used for either 2 or 8 channel audio.

These patches have been checked using the checkpatch.pl script to remove formatting errors.


pelwell:
How is the upstreaming of the I2S driver patch going?

HiassofT:
As commented upstream you shouldn't touch bcm2835-i2s. Channel count has to be kept at 2, bcm2835-i2s doesn't support more channels.

flatmax:
People aren't interested in breaking specs on I2S and allowing DSP mode 
for non TDM compliant silicon (such as the bcm2835). For that reason I 
can't see anything more then an uphill battle in trying to add DSP mode 
to the bcm2835 driver.

For that reason I am doing everything in my machine driver.

There is one item which needs addressing in the bcm2835_i2s.c driver 
which is the control of setting the ch2 registers. I have chosen to add 
"case 8" to this guard in that driver - you can see it in the second patch.

Another way to do the same thing is to ensure that only even channel 
counts activate the second I2S channel register.


Matt

flatmax:
I will make a change to I2S to check that we aren't using 1 channel to 
fix that problem.

Matt

HiassofT:
Sorry, as I have explained in detail upstream this is wrong. People are interested in having a standard conforming i2s driver, not one which is hacked up to solve issues created somewhere else.

The thing you are trying to solve (tunnelling 8 channels through a 2 channel I2S link) belongs into the alsa plugin. If you do it this way you don't need to modify any I2S driver.

pelwell:
I'm keen to get support for this innovative product into the standard images, but we are trying to avoid downstream patches to what is otherwise pure upstream code. @flatmax @HiassofT Could you work together to arrive at a mutually acceptable solution?

pelwell:
Other from that, just swap the order of -addons and -wm8731-audio in the Makefile and README.

flatmax:
I am changing the patch set to use a completely standard I2S 
bcm2835_i2s.c driver, see the new commit.

Matt

HiassofT:
@pelwell we discussed this in detail upstream. By using the tunneling approach, userspace alsa plugin packing multichannel stream into 2-channels, codec driver to model the 2-channel I2S to multichannel depacking it can be implemented without having to modify any existing I2S driver and the hardware plus large parts of the machine driver can be used with other I2S drivers / platforms as well.

http://lists.infradead.org/pipermail/linux-rpi-kernel/2017-February/005322.html
http://lists.infradead.org/pipermail/linux-rpi-kernel/2017-February/005589.html

PR #1887, opened 2017-03-09
Fix gadget mode for bcm2835.

ED6E0F17:
Part Revert "Revert "usb: dwc2: gadget: fix TX FIFO size and address initialization""

This part reverts commit 3fa9538539ac737096c81f3315a14670b1609092.

### DIFF
--- a/arch/arm/boot/dts/overlays/dwc2-overlay.dts
+++ b/arch/arm/boot/dts/overlays/dwc2-overlay.dts
@@ -15,7 +15,7 @@
 			dr_mode = "otg";
 			g-np-tx-fifo-size = <32>;
 			g-rx-fifo-size = <256>;
-			g-tx-fifo-size = <256 128 128 64 64 64 32>;
+			g-tx-fifo-size = <512 512 512 512 512 768>;
 			status = "okay";
 		};
 	};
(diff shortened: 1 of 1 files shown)


pelwell:
This commit modifies an upstream file, which we try to avoid. If the commit is also upstream then you need to reference it. You also need to say what it does and why.

Electron752:
Somewhat related to this, I've been sending patches for 64 bit to upstream for awhile now and I know it takes awhile to get through all the branches.  

What are the requirements for getting an upstream change cherrypicked?  Does it need to be in mainline or can it still be in linux-next? If mainline is a requirement, does it need to be in a branch marked stable, or would a RC version be good enough.  I'm specifically talking about drivers that have an upstream version but are RPI specific.

ED6E0F17:
I think this might be a device-tree issue, so it may be possible to avoid changing the upstream file.

ED6E0F17:
Fixed in downstream device-tree using the values read from the hardware by the patch that was reverted upstream. ( First mentioned at https://www.spinics.net/lists/linux-usb/msg134539.html )

pelwell:
Does that mean you will close this PR?

ED6E0F17:
Sorry for the poor wording, what I meant was:

    (I have) Fixed (this) in the Downstream device-tree.

The original patch to re-land the upstream revert has been replaced with one that only touches Downstream commits.I am not convinced that the firmware is loading arrays of integers from config.txt, so it is necessary to fix this in the kernel source.

pelwell:
@P33M This has morphed into a PR to modify the dwc2 overlay, changing the FIFO sizes. Does this change look correct to you?
```
-			g-tx-fifo-size = <256 128 128 64 64 64 32>;
+			g-tx-fifo-size = <512 512 512 512 512 768>;
```

popcornmix:
@ED6E0F17 is this a possible fix for https://www.raspberrypi.org/forums/viewtopic.php?f=29&t=167934&start=175#p1123465 ?

P33M:
Our total FIFO RAM is 4k, so the size allocations look reasonable.

PR #1896, opened 2017-03-13
Audio driver for Google's voiceHAT for raspberry pi 3.

PeterMalkin:
The audio driver rebased for 4.9.y linux version for raspberry pi to support the voiceHAT for raspberry pi.


pelwell:
I have a number of problems with this PR:

1. It includes a second driver for the ICS43432 I2S microphone. Is the standard Linux driver not good enough?

2. This driver also appears to implement a playback interface, even though the 43432 is just a microphone.

3. I think the codec/soundcard boundary is in the wrong place - if you moved the GPIO handling into the soundcard driver I suspect you could use the standard codec driver.

4. The name is a bit cumbersome - you can drop the "soundcard", and use "audio" in the name of the overlay.

5. I see you've fixed the alphabetical ordering in the makefile (thanks), but now github is complaining about conflicts. When you've addressed the issues above, squash your changes down to single commit, rebase against the current rpi-4.9.y, force push to update the PR, then ping me a comment otherwise I won't be notified.

pelwell:
6. Include an entry in overlays/README, being careful to follow the arcane formatting rules (no tabs, strict indents, two blank lines between overlays, alphabetical ordering of overlays).

PR #1899, opened 2017-03-15
bcm2835-camera: Correct port_parameter_get return value

6by9:
The API for port_parameter_get() requires that the
filled length is returned, or if insufficient space
that the required space is returned.


pelwell:
Looks OK to me.

6by9:
It should seeing as that is almost exactly the patch that you reviewed before for upstream :-)

PR #1925, opened 2017-03-28
Add support for Allo Piano DAC 2.1 plus add-on board for Raspberry Pi.

raashidmuhammed:
The Piano DAC 2.1 has support for 4 channels with subwoofer.


pelwell:
It's nice to see a soundcard doing something different - I'm thinking of the sub support -  but it's still frustrating to see so much copied code.

PR #1927, opened 2017-03-29
mm: Fix boot-time CMA alloc MEM_RECLAIM warning

pelwell:
[ I'm putting this here for review, given that I wrote it by analogy rather than with a full understanding of memory management in the kernel. It certainly gets rid of the warning, but at the very least it needs a lot of testing. ]

The first CMA allocation, currently by the bcm2835-sdhost driver,
triggers a warning that a thread marked as allocating memory is
flushing a workqueue that doesn't have a MEM_RECLAIM rescuer to
guarantee forward progress. This warning is correct - the flushing
that occurs before CMA allocations (via drain_all_pages) uses
system_wq, which is created without the WQ_MEM_RECLAIM flag.

Assuming that the thinking behind the warning is correct, the solution
is to use a workqueue with WQ_MEM_RECLAIM set. Setting that flag on
system_wq without a thorough grasp on the implications could have
far-reaching consequences; creating or locating a different queue
with the flag set to use in drain_all_pages ought to be safer.

The lru_add_drain_wq, currently static to mm/swap.c looks like a
likely candidate. Modify mm/swap.c and mm/page_alloc.c to export
lru_add_drain_wq and use it within drain_all_pages.

N.B. This patch was written without a deep understanding of memory
management in Linux, and there may be unintended consequences.


mm/page_alloc.c
```diff
@@ -2422,10 +2424,11 @@ void drain_all_pages(struct zone *zone)
 	for_each_cpu(cpu, &cpus_with_pcps) {
 		struct work_struct *work = per_cpu_ptr(&pcpu_drain, cpu);
 		INIT_WORK(work, drain_local_pages_wq);
-		schedule_work_on(cpu, work);
+		queue_work_on(cpu, lru_add_drain_wq, work);
+	}
+	for_each_cpu(cpu, &cpus_with_pcps) {
```
popcornmix: The extra braces seem redundant and make the diff look bigger than it really is.

pelwell: Yes, they can be deleted now.

lategoodbye:
An alternative could be to allocate a separate workqueue with WQ_MEM_RECLAIM within the sdhost driver?

PR #1960, opened 2017-04-19
vc4_fkms: Apply firmware overscan offset to hardware cursor

popcornmix:


popcornmix:
@anholt can you review this change?
It fixes an issue with fkms driver when overscan has not been disabled where mouse pointer doesn't line up.

Note: overscan settings from config.txt only apply to framebuffer, not to other dispmanx layers (including hw cursor).

PR #1973, opened 2017-04-23
New driver for the G2 labs TDA1543 v2 DAC

McDaisy:


arch/arm/boot/dts/overlays/README
```diff
@@ -458,6 +458,10 @@ Info:   Configures the Fe-Pi Audio Sound Card
 Load:   dtoverlay=fe-pi-audio
 Params: <None>
```
pelwell, line 460: Two blank lines between entries please, here...

arch/arm/boot/dts/overlays/README
```diff
@@ -458,6 +458,10 @@ Info:   Configures the Fe-Pi Audio Sound Card
...
 
+Name:   g2labs-dac
+Info:   Configures the G2 labs TDA1543 v2 DAC
+Load:   dtoverlay=g2labs-dac
+Params: <None>
```
pelwell: ...and here.

arch/arm/boot/dts/overlays/g2labs-dac.dts
```diff
@@ -0,0 +1,34 @@
+// Definitions for G2LABS DAC
```
pelwell: This file name causes a build error because it doesn't end "-overlay.dts". Rename it to `g2labs-dac-overlay.dts`.

sound/soc/bcm/snd-soc-g2labs.c
```diff
@@ -0,0 +1,118 @@
+/*
```
pelwell: This is another filename that breaks the build. The "snd-soc-" prefix is added by the makefiles - call it g2labs-dac.c (it's better to add the -dac in case of future products).

sound/soc/bcm/snd-soc-tda1543.c
```diff
@@ -0,0 +1,69 @@
+/*
```
pelwell: I'm not convinced you need this file at all. You should be able to use the pcm5102a driver, the only difference being that this driver doesn't advertise 32-bit audio. Looking at the datasheet, the TDA1543 is only capable of 16-bit output, so I'm not convinced this is correct.

pelwell (changes requested):
If you fix all the issues I've mentioned you would probably have the drivers and overlay needed for a working system, but:
1) The codec driver isn't Pi-specific and should really be upstreamed.
2) There are no active controls, so I think you don't need a board driver - simple-audio-card should be up to the job. See https://github.com/raspberrypi/linux/blob/rpi-4.9.y/Documentation/devicetree/bindings/sound/simple-card.txt for more information. There are plenty of examples of its use in https://github.com/raspberrypi/linux/tree/rpi-4.9.y/arch/arm/boot/dts .

arch/arm/boot/dts/overlays/README
McDaisy: done

sound/soc/bcm/snd-soc-tda1543.c
McDaisy: I noticed that I uploaded wrong file. I updated it. Please look again. This driver is suited for our new module which is a bit unique and I dont think this driver will work with any other module as it is right now.

pelwell:
You still need to fix the Makefile so that g2labs-dac.c compiles.

sound/soc/bcm/snd-soc-tda1543.c
pelwell: The file looks unchanged to me, and still has the wrong name (tda1543.c would be better).

McDaisy:
Makefile is done.

Regarding codec file "which is not changed" we must have our own tda1543 file with our own inside to referr to our module. Our module is NOS based and differs form all the other modules.

BR
George

pelwell:
Good, at least that all compiles now.

tda1543.c does not query or modify the hardware in any way - it can't, since the codec has no connections except the I2S signals. The only purpose of the "driver" is to advertise to ALSA that the device exists and what its capabilities are.

Please explain what "NOS based" means and why you need a dedicated driver.

McDaisy:
NOS means non-oversampling.
We need a specific driver to our module because our dac is a bit unique when it come to master/slave settings. For example you can see that we use CBM_CFS. This combínation is quite unique. All other dacs uses CBM_CFM or CBS_CFS.

Regarding tda1543 it is more or less to get our neame in the codec driver and to get a "complete" G2 labs driver/codec system.

pelwell:
You are mixing up board/card driver and codec driver.

For the codec driver (tda1543.c) I appreciate your honesty, but codec drivers should not be application specific - they are common items and should really go in sound/soc/codecs after being submitted upstream. If you are unwilling to upstream your driver then you may want to consider using a similar driver which has already been upstreamed, such as pcm5102a.

For the card driver (g2lab-dac.c), simple-audio-card has independent `frame-master` and `bitclock-master` properties, so CBM_CFS should be no problem (set `bitclock-master` but not `frame-master`).

McDaisy:
Dear Phil,

Please check it now. Hope we are there soon =)

pelwell:
This set of commits is not building the codec. And it also doesn't address the problem that generic codecs should be submitted upstream. And you haven't explained why you need a card driver.

McDaisy:
Dear Phil,
Thanks for your patience. Ill try reply on all you questions.

Just to clarify things since this is our first PR.
*sound/soc/codecs/tda1543.c and sound/soc/codecs/Makefile should be made PR from github.com/torvalds/linux/tree/master/sound/soc/codecs (instead of raspberrypi/linux/tree/master/sound/soc/codecs)

*Should I leave this PR open or reopen when I get the codec in torvalds linux?

*You are asking why we want/need our specific driver. Anwser is easy. We have developed a comersial dac module which we want to release but we want to make sure that appropriate driver is out there easy for our future customers. Since there is no other driver compatible with our new dac module we definately need our own ready driver for convinience.
I see that most of available dac modules have their own drivers approved as well.

pelwell:
The Linux devs don't accept PRs - instead they use kernel mailing lists to circulate and review submissions. The process is described here: https://github.com/raspberrypi/linux/blob/rpi-4.9.y/Documentation/SubmittingDrivers
You will also need to have read and followed this: https://github.com/raspberrypi/linux/blob/rpi-4.9.y/Documentation/SubmittingPatches
and this: https://github.com/raspberrypi/linux/blob/rpi-4.9.y/Documentation/SubmitChecklist

I think you can use the simple-audio-card driver instead for your board instead of creating a new one. The adau7002-simple overlay (https://github.com/raspberrypi/linux/blob/rpi-4.9.y/arch/arm/boot/dts/overlays/adau7002-simple-overlay.dts) uses this approach. Yes there are already a lot of soundcard drivers in our tree, but when each is submitted we check to see if an existing driver (particularly simple-audio-card) could also be used. Your driver is so simple that it is a prime candidate.

McDaisy:
Hi again Phil,

I see what you aiming at but honestly I dont quite understand how we can use simple audio driver like adau7002 when our driver not only is different in bitclock-slave and frame-slave options but also bclk_ratio is different which is controlled from the driver.
Maybe our driver is simple but it is unique combination of settings/options in the driver.

bclk_ratio 32 will not work with our module. The design and combination of hardware is made for bclk_ratio 128.

I will do my best to use mailing list to get the codec in kernel. Thanks for that. Is that the only way to get the codec in there ? Thanks!

BR
George

pelwell:
> bclk_ratio 32 will not work with our module. The design and combination of hardware is made for bclk_ratio 128.

I thinking you are correct. simple-audio-card doesn't allow bclk_ratio to be set, so I'll accept your g2labs-dac driver.

clivem:
Ping @msperl ...... Did you ever submit the simple framework patches upstream, which allowed setting the bclk_ratio from dt for simple-audio-card? I still use it for a couple of my cards that require custom ratios, but I'm not sure if you abandoned it or it got submitted to alsa-dev?

msperl:
I tried, but it was turned down - I would need to dig into the thread to find it.

clivem:
@msperl.... That's a shame. For a while I was distributing a couple of overlays that depended on that patch set, and was being used by one/two of the rpi audio specific distributions, specifically to support the guys using uncommon things, like first gen (obsolete) DAC chips in NOS config, but more commonly by people that wanted to end up with integer clock divs rather than fractional. eg. 
[simple-bclk-int-div-50-100-overlay.txt](https://github.com/raspberrypi/linux/files/955906/simple-bclk-int-div-50-100-overlay.txt)

pelwell:
If it would mean we could reduce the number of soundcard drivers then I think we'd be happy to take patches that upstream reject.

msperl:
Note: the main upstream concern was that the DT should not describe linux methods that need to get used - it is supposed to be abstract.
Hence it was rejected and the "one driver per device" recommended.
If it helps you, then I am fine with merging it downstream...

pelwell:
That reason for rejection sounds a bit bogus - surely the bclk ratio is a property of the hardware?

Anyway, give us a PR and I'll merge it - anything to make simple-card more useful.

msperl:
here the link to the RFC patchset that allows rules in the DT:
http://mailman.alsa-project.org/pipermail/alsa-devel/2016-May/108296.html

you should be able to pick up those 2 patches easily:
* http://mailman.alsa-project.org/pipermail/alsa-devel/2016-May/108298.html
* http://mailman.alsa-project.org/pipermail/alsa-devel/2016-May/108297.html

If additional actions (besides asoc_generic_hw_params_set_fixed_bclk_size) or matchers (e.g: asoc_generic_hw_params_match_rate) are needed, then these can get implemented on top of it...

pelwell:
Thanks, Martin. That's more complexity than I was hoping for, but as we know, if it's too simple it isn't useful. I'm still torn between a few options:

1. Use the data-as-code approach to create a complex-card driver (as started by your patches). Both the overlays and the driver would get larger. We'd have to document the new driver and be prepared to extend it to cope with edge cases.
2. Create a kitchen-sink superset-driver with everything in it. Less code to maintain, but all the drivers end up larger than they need to be.
3. Group the existing drivers according to which original driver they've cloned, and put the skinning in the overlay.

If any of the cards have HAT EEPROMs then we'd have to take care to maintain compatibility with the embedded overlays (skinning and support code would have to remain in the driver).

clivem:
You want to do 1&3 IMHO. (That's what I did for the "audio" kernel. Put in a new driver for the ES9023 stuff that was currently piggy-backing off of the original HB, so splitting the PCM5102 and ES9023 stuff, which needed to be done to be able to add the 384k support in a way that wasn't breaking anything that couldn't support it. A generic for the TI 5122 stuff, (of which there is more than anything else), using dt "compatible" for board specific config within that single machine driver. You still need option "1", a usable and highly configurable simple card driver to support any of the corner case stuff, (not limited to weird clocking and other stuff like cards that are not actually based on the PCM5122 chipset, currently piggybacking off of a machine driver that is really intended for a specific 5122 board), and not need a new machine driver for any additional boards that are not covered with a manufacturer, board, or chipset,  specific machine driver, going forward......)

HiassofT:
I've had a look at the simple-card RFC patch and, like the reviewers on the ALSA mailing list, I don't like it too.

I'd say, just extend the simple-card DT/driver with properties to (optionally) configure the bclk ratio, eg something like
```
i2s-bits-16=16;
i2s-bits-24=32;
i2s-bits-32=32;
```
or an array of pairs, or 2 separate arrays with the transformation or something like that. This should allow you to replace a bunch of current drivers.

Cards that need more complex configuration should do that via separate drivers. Trying to handle that via code or complex rules embedded into DT is the wrong approach IMO.

HiassofT:
As a side node: I'm not 100% sure why so many driver set bclk_ratio at all - most chips (like PCM5122) operate perfectly well in 24bit-mode clocked at 48x fs.

I've got the impression the hw_params function with the set_bclk call (which only affects transmitting 24bit  audio at 64 x fs instead of 48 x fs) was copy-and-pasted over and over without any real reason...

clivem:
S24_LE != S24_3LE

HiassofT:
I'm well aware of the difference. And also that S24_LE is a rather odd format - using S32_LE instead would make a lot more sense, especially when the driver is transmitting S24_LE as 32 bits.

HiassofT:
I had a quick glance thorugh the downstream drivers, did some tests with my iqaudio dac (pcm512x) in 24bit mode and now I'm pretty sure that most of the set_bclk_ratio code which uses snd_pcm_format_physical_width instead of snd_pcm_format_width are copy-and-paste errors.

Using a different bclk ratio than the default (sample_bits * channel, like snd_soc_calc_bclk() does) can be needed in some cases, but most of the time deviating from that is a bug. That bug (padding 24bit samples to 32bit) probably went unnoticed for a long time as quite a lot of I2S codecs accept additional bclk cycles after the LSB of a sample.

The good news is: This means we can most certainly ignore the special bclk treatment of several drivers and replace them with simple-card as it is now.

Edit: in case optimization has been added so that the clock can be derived from the 19.2MHz OSC there's an easier solution: just request S32_LE sample format from userspace. Alsalib (the plughw plugin) will automatically do sample-format conversion so disabling S24_LE in the kernel driver (could be doable with simple-card, but haven't looked into it) should achieve the same result - 24bit audio transmitted as 32bit.

clivem:
@HiassofT With the greatest of respect, not only are there "quite a lot of I2S codecs accept additional bclk cycles after the LSB of a sample", there are some that won't function without it. While the original HB DAC driver might on the surface appear to only be in use with the original HB 'B' PCM5102 board, there are a lot of instances of that out in the wild being used to drive Sabre hardware where 64fs is reqd for < 32 bit word lengths. Most of the guys I know using that config are depending on the current S24 behavior from userspace resulting in a 64fs BCLK. My point being, you also need to take into consideration how things are being used out in the wild, not just from reading the code. Or you could just argue that anyone using code that was intended to drive a PCM5102 board with something other than that, deserves what they get, I suppose.

HiassofT:
Sure, as I wrote there are situations where some additional logic is needed - eg when some device only accepts 64fs bclk, regardless of sample depth. Such things will then need to be handled in a driver.

But my point is that always auto-converting 24bit to 32bit in a driver, while keeping 16bit as is, is rather poiintless - the same can be achieved by using 32bit in the first place. Not all things need to be handled by the kernel And if a device can't cope with 24bit / 48fs it's easier to just disable that format in the driver and let userspace alsa deal with it.

Audiophools will then also be happy that they're now using 32bit mode where they previously only had 24bit mode :)

clivem:
> Audiophools will then also be happy that they're now using 32bit mode where they previously only had 24bit mode 

I'm glad you stuck a smiley on the end of that sentence. While supposedly "high-res" 24 bit media is readily available for purchase by the a'phools, I'm unaware of any 32 bit, even before we get into arguments of whether there would actually be any audible improvement from those extra 8 bits of data/noise. ;)

Anyway, I wish you all good luck with this. Not something I'm going to get involved in. I tried once and learned a lesson...... Although to be fair, the old guard, (Daniel, Gordon, etc.) were happy as long as backward compatibility was going to be maintained with their hardware, whether we were talking about a move to generic machine drivers, or new dt files.......

HiassofT:
My intention is not to break a ton of drivers/cards but to look into alternatives how we can deal with the ever increasing amont of RPi soundcards - while preserving current functionality and avoiding to create an overly complex monster driver.

I don't say we can replace all drivers with simple-card and of course we'll have to carefully test. But even if we only manage to replace 30-50% of the copy-and-paste drivers containing only boilerplate code with simple-card that'd be an improvement.

clivem:
> But even if we only manage to replace 30-50% of the copy-and-paste drivers containing only boilerplate code with simple-card that'd be an improvement.

Most of the cut and paste is with the PCM5122 machine drivers. It's not the generic setup stuff that has required the cut and paste to start with, rather than using what was already there, it's the GPIO muting, master clock, and more recently the Allo stuff and the combining of the 2 DAC chips for subwoofer and firmware upload. You can't replace them with simple card, anyway.

HiassofT:
> Most of the cut and paste is with the PCM5122 machine drivers.

Good point and I think it's best to start with that as a first step.
i.e. refactor the existing driver code so that we have a single driver
for each family (pcm5122, wm8804, ...).

Card selection inside the driver can be done via the compatible
string, so DT overlays can stay the same and we shouldn't have
any issues with HATs.

@clivem IIRC you did this before, do you still have your code trees
and can care about that?

HiassofT:
FYI: Today I've implemented tdm-slot support for bcm2835-i2s, still need to do some testing and cleanup but things are looking very good so far.

@pelwell @McDaisy with tdm-slot support added cards that have a fixed bclk ratio (eg always 32 bit per sample) - like this one - can easily be supported via simple-card. We just need to add the tdm-slot config to the cpu node:
```
simple-audio-card,cpu {
    sound-dai = <&i2s>;
    dai-tdm-slot-num = <2>;
    dai-tdm-slot-width = <32>;
};
```

I'll open a PR when I've finished testing and then post a sample DT overlay here.

HiassofT:
PR with support for set_tdm_slot is here: https://github.com/raspberrypi/linux/pull/1982

@McDaisy here's an example DT overlay of how you could get your card working when the above PR is merged. CPU is bit-clock master, codec is frame clock master and each sample consists of 16 bits (bclk is 32fs.
https://gist.github.com/HiassofT/20b18111591ff25149238e2738107e44

arch/arm/boot/dts/overlays/g2labs-dac-overlay.dts
```diff
@@ -0,0 +1,34 @@
...
+			status = "okay";
+		};
+	};
+
+	fragment@1 {
+		target-path = "/";
+		__overlay__ {
+			tda1543-codec {
+				#sound-dai-cells = <0>;
+				compatible = "ti,tda1543";
```
clivem, line 20: compatible=\"\<manufacturer\>,\<model\>\" Texas Instruments manufacture/manufactured the TDA1543? You learn something new every day! ;)

arch/arm/boot/dts/overlays/g2labs-dac-overlay.dts
```diff
@@ -0,0 +1,34 @@
...
+				status = "okay";
+			};
+		};
+	};
+
+	fragment@2 {
+		target = <&sound>;
+		__overlay__ {
+			compatible = "rpi,g2labs-dac";
```
clivem, line 29: compatible = \"rpi,g2labs-dac\"? Wouldn't \"g2labs,g2labs-dac\" be better for compatible string in dt and driver?

McDaisy:
Nice to see so many dedicated people and thanks for all input and new ideas @pelwell @clivem @msperl @HiassofT 

We tried your suggested modified version of bcm2835-i2s as PR #1982 
It compiled and seems to start properly and play properly.


Edit: I wrote earlier that it didnt work with B+ but there was another issue which was resolved and it seems to work fine!

pelwell:
> We tried your suggested modified version of bcm2835-i2s as PR #1982
  It compiled and seems to start properly and play properly.

Once #1982 is merged (I'm waiting for @flatmax to get a chance to do some testing) are you happy to redo this PR using @HiassofT's overlay (and an updated README and Makefile) instead?

clivem:
> Once #1982 is merged (I'm waiting for @flatmax to get a chance to do some testing) 

I think you might want to cast the net a little wider, regarding testing of that patch..... I did a quick 4.9 build with it included at the weekend, and the only sound I got out of the audio-player I tested it with, was "static". (Didn't have time to investigate, before I rolled-back to what was previously working. Sorry, not much help, I know.)

mi-hol:
@McDaisy it has been a bit silent on this PR. Is it still active or should it be closed?

JamesH65:
No activity for over a year. 
This issue will be closed within 30 days unless further interactions are posted. If you wish this issue to remain open, please add a comment. A closed issue may be reopened if requested.

JamesH65:
Closing due to lack of activity. Please request to be reopened if you feel this issue is still relevant.

PR #1982, opened 2017-04-30
bcm2835-i2s: add support for set_tdm_slot and 384kHz sample rates

HiassofT:
set_tdm_slot support allows us to use codecs with fixed bclks per sample with simple-card. This means we will be able to remove a bunch of downstream card drivers and replace them with DT overlays using the simple-card driver. All that's needed then is to set the tdm slot config in the cpu node of simple-card. eg:
```
cpu_dai: simple-audio-card,cpu {
        sound-dai = <&i2s>;
        dai-tdm-slot-num = <2>;
        dai-tdm-slot-width = <32>;
};
```

The old set_bclk_ratio method is still supported, so there's no pressing need to change the current downstream drivers.

384kHz support has been requested for quite a long time. Testing with an iqaudio-dac (pcm5122 codec) went fine so I think it is safe to finally add that.

ping @pelwell @clivem

### DIFF
--- a/sound/soc/bcm/bcm2835-i2s.c
+++ b/sound/soc/bcm/bcm2835-i2s.c
@@ -344,6 +344,9 @@ static int bcm2835_i2s_hw_params(struct snd_pcm_substream *substream,
 	unsigned int rx_mask, tx_mask;
 	unsigned int rx_ch1_pos, rx_ch2_pos, tx_ch1_pos, tx_ch2_pos;
 	unsigned int mode, format;
+	bool bit_clock_master = false;
+	bool frame_sync_master = false;
+	bool frame_start_falling_edge = false;
 	uint32_t csreg;
 	int ret = 0;
 
@@ -387,16 +390,39 @@ static int bcm2835_i2s_hw_params(struct snd_pcm_substream *substream,
 	if (data_length > slot_width)
 		return -EINVAL;
 
-	/* Clock should only be set up here if CPU is clock master */
+	/* Check if CPU is bit clock master */
 	switch (dev->fmt & SND_SOC_DAIFMT_MASTER_MASK) {
 	case SND_SOC_DAIFMT_CBS_CFS:
 	case SND_SOC_DAIFMT_CBS_CFM:
+		bit_clock_master = true;
+		break;
+	case SND_SOC_DAIFMT_CBM_CFS:
+	case SND_SOC_DAIFMT_CBM_CFM:
+		bit_clock_master = false;
+		break;
+	default:
+		return -EINVAL;
+	}
+
+	/* Check if CPU is frame sync master */
+	switch (dev->fmt & SND_SOC_DAIFMT_MASTER_MASK) {
+	case SND_SOC_DAIFMT_CBS_CFS:
+	case SND_SOC_DAIFMT_CBM_CFS:
+		frame_sync_master = true;
+		break;
+	case SND_SOC_DAIFMT_CBS_CFM:
+	case SND_SOC_DAIFMT_CBM_CFM:
+		frame_sync_master = false;
+		break;
+	default:
+		return -EINVAL;
+	}
+
+	/* Clock should only be set up here if CPU is clock master */
+	if (bit_clock_master) {
 		ret = clk_set_rate(dev->clk, bclk_rate);
 		if (ret)
 			return ret;
-		break;
-	default:
-		break;
 	}
 
 	/* Setup the frame format */
@@ -427,13 +453,41 @@ static int bcm2835_i2s_hw_params(struct snd_pcm_substream *substream,
 
 		/* Setup frame sync signal for 50% duty cycle */
 		framesync_length = frame_length / 2;
+		frame_start_falling_edge = true;
+		break;
+	case SND_SOC_DAIFMT_LEFT_J:
+		if (slots & 1)
+			return -EINVAL;
+
+		odd_slot_offset = slots >> 1;
+		data_delay = 0;
+		framesync_length = frame_length / 2;
+		frame_start_falling_edge = false;
+		break;
+	case SND_SOC_DAIFMT_RIGHT_J:
+		if (slots & 1)
+			return -EINVAL;
+
+		/* Odd frame lengths aren't supported */
+		if (frame_length & 1)
+			return -EINVAL;
+
+		odd_slot_offset = slots >> 1;
+		data_delay = slot_width - data_length;
+		framesync_length = frame_length / 2;
+		frame_start_falling_edge = false;
+		break;
+	case SND_SOC_DAIFMT_DSP_A:
+		data_delay = 1;
+		framesync_length = 1;
+		frame_start_falling_edge = false;
+		break;
+	case SND_SOC_DAIFMT_DSP_B:
+		data_delay = 0;
+		framesync_length = 1;
+		frame_start_falling_edge = false;
 		break;
 	default:
-		/*
-		 * TODO
-		 * Others are possible but are not implemented at the moment.
-		 */
-		dev_err(dev->dev, "%s:bad format\n", __func__);
 		return -EINVAL;
 	}
 
@@ -441,6 +495,15 @@ static int bcm2835_i2s_hw_params(struct snd_pcm_substream *substream,
 		rx_mask, slot_width, data_delay, odd_slot_offset);
 	bcm2835_i2s_calc_channel_pos(&tx_ch1_pos, &tx_ch2_pos,
 		tx_mask, slot_width, data_delay, odd_slot_offset);
+
+	/*
+	 * Transmitting data immediately after frame start, eg
+	 * in left-justified or DSP mode A, only works stable
+	 * if bcm2835 is the frame clock master.
+	 */
+	if ((!rx_ch1_pos || !tx_ch1_pos) && !frame_sync_master)
+		dev_warn(dev->dev,
+			"Unstable slave config detected, L/R may be swapped");
 
 	/*
 	 * Set format for both streams.
@@ -472,62 +535,38 @@ static int bcm2835_i2s_hw_params(struct snd_pcm_substream *substream,
 	mode |= BCM2835_I2S_FLEN(frame_length - 1);
 	mode |= BCM2835_I2S_FSLEN(framesync_length);
 
-	/* Master or slave? */
-	switch (dev->fmt & SND_SOC_DAIFMT_MASTER_MASK) {
-	case SND_SOC_DAIFMT_CBS_CFS:
-		/* CPU is master */
-		break;
-	case SND_SOC_DAIFMT_CBM_CFS:
-		/*
-		 * CODEC is bit clock master
-		 * CPU is frame master
-		 */
+	/* CLKM selects bcm2835 clock slave mode */
+	if (!bit_clock_master)
 		mode |= BCM2835_I2S_CLKM;
-		break;
-	case SND_SOC_DAIFMT_CBS_CFM:
-		/*
-		 * CODEC is frame master
-		 * CPU is bit clock master
-		 */
-		mode |= BCM2835_I2S_FSM;
-		break;
-	case SND_SOC_DAIFMT_CBM_CFM:
-		/* CODEC is master */
-		mode |= BCM2835_I2S_CLKM;
-		mode |= BCM2835_I2S_FSM;
-		break;
-	default:
-		dev_err(dev->dev, "%s:bad master\n", __func__);
-		return -EINVAL;
-	}
 
-	/*
-	 * Invert clocks?
-	 *
-	 * The BCM approach seems to be inverted to the classical I2S approach.
-	 */
-	switch (dev->fmt & SND_SOC_DAIFMT_INV_MASK) {
+	/* FSM selects bcm2835 frame sync slave mode */
+	if (!frame_sync_master)
+		mode |= BCM2835_I2S_FSM;
+
+	/* CLKI selects normal clocking mode, sampling on rising edge */
+        switch (dev->fmt & SND_SOC_DAIFMT_INV_MASK) {
 	case SND_SOC_DAIFMT_NB_NF:
-		/* None. Therefore, both for BCM */
-		mode |= BCM2835_I2S_CLKI;
-		mode |= BCM2835_I2S_FSI;
-		break;
-	case SND_SOC_DAIFMT_IB_IF:
-		/* Both. Therefore, none for BCM */
-		break;
 	case SND_SOC_DAIFMT_NB_IF:
-		/*
-		 * Invert only frame sync. Therefore,
-		 * invert only bit clock for BCM
-		 */
 		mode |= BCM2835_I2S_CLKI;
 		break;
 	case SND_SOC_DAIFMT_IB_NF:
-		/*
-		 * Invert only bit clock. Therefore,
-		 * invert only frame sync for BCM
-		 */
-		mode |= BCM2835_I2S_FSI;
+	case SND_SOC_DAIFMT_IB_IF:
+		break;
+	default:
+		return -EINVAL;
+	}
+
+	/* FSI selects frame start on falling edge */
+	switch (dev->fmt & SND_SOC_DAIFMT_INV_MASK) {
+	case SND_SOC_DAIFMT_NB_NF:
+	case SND_SOC_DAIFMT_IB_NF:
+		if (frame_start_falling_edge)
+			mode |= BCM2835_I2S_FSI;
+		break;
+	case SND_SOC_DAIFMT_NB_IF:
+	case SND_SOC_DAIFMT_IB_IF:
+		if (!frame_start_falling_edge)
+			mode |= BCM2835_I2S_FSI;
 		break;
 	default:
 		return -EINVAL;
@@ -562,6 +601,13 @@ static int bcm2835_i2s_hw_params(struct snd_pcm_substream *substream,
 
 	dev_dbg(dev->dev, "sampling rate: %d bclk rate: %d\n",
 		params_rate(params), bclk_rate);
+
+	dev_dbg(dev->dev, "CLKM: %d CLKI: %d FSM: %d FSI: %d frame start: %s edge\n",
+		!!(mode & BCM2835_I2S_CLKM),
+		!!(mode & BCM2835_I2S_CLKI),
+		!!(mode & BCM2835_I2S_FSM),
+		!!(mode & BCM2835_I2S_FSI),
+		(mode & BCM2835_I2S_FSI) ? "falling" : "rising");
 
 	return ret;
 }
--- a/sound/soc/bcm/bcm2835-i2s.c
+++ b/sound/soc/bcm/bcm2835-i2s.c
@@ -783,7 +783,8 @@ static struct snd_soc_dai_driver bcm2835_i2s_dai = {
 				| SNDRV_PCM_FMTBIT_S32_LE
 		},
 	.ops = &bcm2835_i2s_dai_ops,
-	.symmetric_rates = 1
+	.symmetric_rates = 1,
+	.symmetric_samplebits = 1,
 };
 
 static bool bcm2835_i2s_volatile_reg(struct device *dev, unsigned int reg)
--- a/sound/soc/bcm/bcm2835-i2s.c
+++ b/sound/soc/bcm/bcm2835-i2s.c
@@ -765,7 +765,9 @@ static struct snd_soc_dai_driver bcm2835_i2s_dai = {
 	.playback = {
 		.channels_min = 2,
 		.channels_max = 2,
-		.rates =	SNDRV_PCM_RATE_8000_192000,
+		.rates =	SNDRV_PCM_RATE_CONTINUOUS,
+		.rate_min =	8000,
+		.rate_max =	384000,
 		.formats =	SNDRV_PCM_FMTBIT_S16_LE
 				| SNDRV_PCM_FMTBIT_S24_LE
 				| SNDRV_PCM_FMTBIT_S32_LE
@@ -773,7 +775,9 @@ static struct snd_soc_dai_driver bcm2835_i2s_dai = {
 	.capture = {
 		.channels_min = 2,
 		.channels_max = 2,
-		.rates =	SNDRV_PCM_RATE_8000_192000,
+		.rates =	SNDRV_PCM_RATE_CONTINUOUS,
+		.rate_min =	8000,
+		.rate_max =	384000,
 		.formats =	SNDRV_PCM_FMTBIT_S16_LE
 				| SNDRV_PCM_FMTBIT_S24_LE
 				| SNDRV_PCM_FMTBIT_S32_LE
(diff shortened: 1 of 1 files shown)


clivem:
> 384kHz support has been requested for quite a long time. Testing with an iqaudio-dac (pcm5122 codec) went fine so I think it is safe to finally add that.

OK, cool! It's safe to finally add that, now that it has finally been tested! ;)

sound/soc/bcm/bcm2835-i2s.c
```diff
@@ -1,3 +1,4 @@
+#define DEBUG
```
pelwell: Remove this.

sound/soc/bcm/bcm2835-i2s.c
```diff
@@ -221,25 +231,86 @@ static int bcm2835_i2s_set_dai_fmt(struct snd_soc_dai *dai,
...
+	unsigned int* chpos)
+{
+	int i, slot;
+
+	/* mask out unavailable slots */
+	mask &= GENMASK(slots - 1, 0);
+
+	/*
+	 * bcm2835-i2s only supports stereo configurations, so also
+	 * check that excactly 2 bits are set in the mask.
```
pelwell: Typo "excactly"

sound/soc/bcm/bcm2835-i2s.c
```diff
@@ -249,39 +320,29 @@ static int bcm2835_i2s_hw_params(struct snd_pcm_substream *substream,
...
-	case SNDRV_PCM_FORMAT_S24_LE:
-		data_length = 24;
-		break;
-	case SNDRV_PCM_FORMAT_S32_LE:
-		data_length = 32;
-		break;
-	default:
-		return -EINVAL;
-	}
+	data_length = params_width(params);
```
pelwell: Have we lost an error for unsupported formats?

sound/soc/bcm/bcm2835-i2s.c
```diff
@@ -308,28 +369,24 @@ static int bcm2835_i2s_hw_params(struct snd_pcm_substream *substream,
...
-	case 2:
-	case 8:
-		format = BCM2835_I2S_CH1(format) | BCM2835_I2S_CH2(format);
-		format |= BCM2835_I2S_CH1(BCM2835_I2S_CHPOS(ch1pos));
-		format |= BCM2835_I2S_CH2(BCM2835_I2S_CHPOS(ch2pos));
-		break;
-	default:
-		return -EINVAL;
-	}
+	format = BCM2835_I2S_CH1(format) | BCM2835_I2S_CH2(format);
```
pelwell: Do we care that there is no error case here?

@FlatMax How do these changes work for you?

HiassofT:
@pelwell thanks a lot for your feedback! I've removed the debug leftover and fixed the typo.

As for the removed checks: hw_params is already guarded by the (stream) constraints setup by the driver in snd_soc_dai_driver. ALSA/ checks for these constraints and requests for other channels/samplerates/formats won't make it to hw_params. So these checks in hw_params were redundant.

But I noticed another missing check, hw_params missed errors from clk_set_rate. I've added that check to the samplerate patch, using SNDRV_PCM_RATE_CONTINUOUS should now really be safe.

As for Matt's ugly hack: messing with the constraints setup by a driver by directly modifying it's private data structures was never allowed. ALSA drivers should only use the snd_pcm_hw_constraint* methods to add additional constrains. Other than that I'm pretty sure that this hack still works as the checks in hw_params are now removed... But I'd really like to see his driver implemented properly, as Mark Brown and I told him on the ALSA mailing list.

flatmax:
On 01/05/17 16:47, Phil Elwell wrote:
>
> *@pelwell* commented on this pull request.
>
> ------------------------------------------------------------------------
>
> In sound/soc/bcm/bcm2835-i2s.c 
> <https://github.com/raspberrypi/linux/pull/1982#discussion_r114103195>:
>
> > @@ -1,3 +1,4 @@
> +#define DEBUG
>
> Remove this.
>
> ------------------------------------------------------------------------
>
> In sound/soc/bcm/bcm2835-i2s.c 
> <https://github.com/raspberrypi/linux/pull/1982#discussion_r114103218>:
>
> > @@ -221,25 +231,86 @@ static int bcm2835_i2s_set_dai_fmt(struct snd_soc_dai *dai,
>   	return 0;
>   }
>   
> +static int bcm2835_i2s_set_channel_pos(unsigned int mask, int slots, int width,
> +	unsigned int* chpos)
> +{
> +	int i, slot;
> +
> +	/* mask out unavailable slots */
> +	mask &= GENMASK(slots - 1, 0);
> +
> +	/*
> +	 * bcm2835-i2s only supports stereo configurations, so also
> +	 * check that excactly 2 bits are set in the mask.
>
> Typo "excactly"
>
> ------------------------------------------------------------------------
>
> In sound/soc/bcm/bcm2835-i2s.c 
> <https://github.com/raspberrypi/linux/pull/1982#discussion_r114103305>:
>
> > -	 * does not apply.
> -	 */
> -	switch (params_format(params)) {
> -	case SNDRV_PCM_FORMAT_S16_LE:
> -		data_length = 16;
> -		break;
> -	case SNDRV_PCM_FORMAT_S24_LE:
> -		data_length = 24;
> -		break;
> -	case SNDRV_PCM_FORMAT_S32_LE:
> -		data_length = 32;
> -		break;
> -	default:
> -		return -EINVAL;
> -	}
> +	data_length = params_width(params);
>
> Have we lost an error for unsupported formats?
>
> ------------------------------------------------------------------------
>
> In sound/soc/bcm/bcm2835-i2s.c 
> <https://github.com/raspberrypi/linux/pull/1982#discussion_r114103497>:
>
> > @@ -308,28 +369,24 @@ static int bcm2835_i2s_hw_params(struct snd_pcm_substream *substream,
>   		return -EINVAL;
>   	}
>   
> -	ch1pos = data_delay;
> -	ch2pos = bclk_ratio / 2 + data_delay;
> -
> -	switch (params_channels(params)) {
> -	case 2:
> -	case 8:
> -		format = BCM2835_I2S_CH1(format) | BCM2835_I2S_CH2(format);
> -		format |= BCM2835_I2S_CH1(BCM2835_I2S_CHPOS(ch1pos));
> -		format |= BCM2835_I2S_CH2(BCM2835_I2S_CHPOS(ch2pos));
> -		break;
> -	default:
> -		return -EINVAL;
> -	}
> +	format = BCM2835_I2S_CH1(format) | BCM2835_I2S_CH2(format);
>
> Do we care that there is no error case here?
>
> @FlatMax <https://github.com/FlatMax> How do these changes work for you?
>

People have received their Octo 8 channel sound cards and they are 
reporting good stability and great functioning.

I want to check that this patch doesn't break some features in both the 
stereo card and the 8 channel card.

In the case of the stereo card it depends heavily on the 
bcm2835_i2s_set_dai_bclk_ratio function to setup the clocking offsets 
correctly for the wm8731 chip. I want to check that this doesn't effect 
wm8731 based cards.

In the case of the Octo 8 channel card, it relies heavily on the DMA 
subsystem of ALSA. I would want to check that using the TDM ALSA 
subsystem doesn't effect the DMA mapping.

thanks
Matt

flatmax:
On 01/05/17 18:59, Matthias Reichl wrote:
>
> @pelwell <https://github.com/pelwell> thanks a lot for your feedback! 
> I've removed the debug leftover and fixed the typo.
>
> As for the removed checks: hw_params is already guarded by the 
> (stream) constraints setup by the driver in snd_soc_dai_driver. ALSA/ 
> checks for these constraints and requests for other 
> channels/samplerates/formats won't make it to hw_params. So these 
> checks in hw_params were redundant.
>
> But I noticed another missing check, hw_params missed errors from 
> clk_set_rate. I've added that check to the samplerate patch, using 
> SNDRV_PCM_RATE_CONTINUOUS should now really be safe.
>
> As for Matt's ugly hack: messing with the constraints setup by a 
> driver by directly modifying it's private data structures was never 
> allowed. ALSA drivers should only use the snd_pcm_hw_constraint* 
> methods to add additional constrains. Other than that I'm pretty sure 
> that this hack still works as the checks in hw_params are now 
> removed... But I'd really like to see his driver implemented properly, 
> as Mark Brown and I told him on the ALSA mailing list.
>

I understand that this patch is trying to pretend that the BCM chip is a 
time division multiple access chip (TDM) which is used in multichannel 
sound cards.

I can't imagine it would float with ALSA devs, because the BCM chipset 
is strictly stereo, not multichannel TDM.

If you we want to keep the bcm i2s driver closely in line with the 
upstream system, then this very big modification would swing us far away 
from their current point. I am flagging that it may make future cherry 
picks from the mainline kernel difficult to integrate.

thanks
Matt

HiassofT:
@flatmax I plan to upstream these patches, just wanted to add it here first to gain some feedback and testing.

bcm2835 will work perfectly well in TDM setups and can be extended to DSP mode as well. I didn't have the time yet to implement and test DSP mode clocking but once I finished that it'll be possible to eg use bcm2835 as a DSP mode slave in an 8 slot setup. The only limitation is that bcm2835 is restricted to exactly 2 active slots.

That limitation isn't too bad, the whole idea behind TDM is to hook up multiple (slave) devices, each handling a different part of the audio data stream - eg have 4 stereo DACs connected to a DSP with 8-channel (DSP mode) output.

TDM support for bcm2835 means you'll be able to do this with RPis as well, eg connect 4 RPis tp an 8-channel source.

flatmax:
On 01/05/17 19:58, Matthias Reichl wrote:
>
> @flatmax <https://github.com/flatmax> I plan to upstream these 
> patches, just wanted to add it here first to gain some feedback and 
> testing.
>
> bcm2835 will work perfectly well in TDM setups and can be extended to 
> DSP mode as well. I didn't have the time yet to implement and test DSP 
> mode clocking but once I finished that it'll be possible to eg use 
> bcm2835 as a DSP mode slave in an 8 slot setup. The only limitation is 
> that bcm2835 is restricted to exactly 2 active slots.
>
> That limitation isn't too bad, the whole idea behind TDM is to hook up 
> multiple (slave) devices, each handling a different part of the audio 
> data stream - eg have 4 stereo DACs connected to a DSP with 8-channel 
> (DSP mode) output.
>
> TDM support for bcm2835 means you'll be able to do this with RPis as 
> well, eg connect 4 RPis tp an 8-channel source.
>
This is a good idea.

Have you confirmed that these patches don't effect the bit clock offsets 
for L and R channels ? I would like to make sure that the wm8731 machine 
driver with a 12 MHz crystal doesn't get effected. You can see that it 
heavily relies on setting bclk ratios in the 
snd_audioinjector_pi_soundcard_hw_params functionhere :
https://github.com/raspberrypi/linux/blob/rpi-4.9.y/sound/soc/bcm/audioinjector-pi-soundcard.c

Secondly, I would like to do a test of this code against the audio 
injector Octo system which relies on DMA mapping consistency to make 
sure the 8 channels of audio don't jump around.

thanks
Matt

HiassofT:
@flatmax thanks a lot for your feedback! I'd missed that there are actually codecs out there that require an odd bclk ratio. I've updated the commit so that these will now work exactly as before. Please retest with the updated PR.

flatmax:
On 02/05/17 03:35, Matthias Reichl wrote:
>
> @flatmax <https://github.com/flatmax> thanks a lot for your feedback! 
> I'd missed that there are actually codecs out there that require an 
> odd bclk ratio. I've updated the commit so that these will now work 
> exactly as before. Please retest with the updated PR.
>
I am overseas at the moment, but will be back in action and able to test 
these patches next week.

Matt

HiassofT:
Just noticed that for supporting right-justified mode I'll have to store the original slot/channel lengths. I'll refactor my code and push a new version when I'm ready. Please wait with merging until then.

HiassofT:
I've completely refactored my TDM code, position calculation is now done in hw_params as before (with the use of helper functions).

This was necessary so that logical TDM slot numbering could be mapped in the expected way (0 1 2 3... for DSP modes, 0 2 ... 1 3 ... for I2S and left/right justified modes).

I've also implemented support for left/right justified and DSP A and B modes.

I found out why I sometimes got channel swap (or shift) before: this happens when bcm2835 is configured as a frame slave and data has to be transmitted immediately after frame start (eg in DSP mode B with active slot 0). The driver detects such configurations and warns about them.

BTW: it's not unusual for this configuration to fail, wm5102 datasheet also mentions that this is unstable and the driver errors out in such conditions.

Finally, I added a patch that adds the missing symmetry constraint. Looks like I forgot to PR/upstream it before.

I've tested TDM mode (4 slots of 32 bit each) in left-justified, I2S, DSP-A and DSP-B modes, in master and slave configs (but not all combinations of these) with the Cirrus logic audio card and it worked as expected.

I've also included debug printks in the driver and verified that channel positioning and calling set_bclk_ratio with odd numbers works as before.

Please review and test again.

flatmax:
Hi there,

I have begun the test process for this. Cloned your repo Matthias, 
checked out the branch :

git clone --depth=1 https://github.com/HiassofT/rpi-linux.git -b 
i2s-tdm-samplerate-4.9

Matt

flatmax:
I have done a listening test with the WM8731 codec - which has strange 
bit offsets - and it sounds good.

I would like to also test the Octo multichannel sound card (early next 
week or on the weekend) and do a second sanity check.

But so far so good.

Matt

pelwell:
When we have enough positive feedback we'll push the patch to 4.11 to get some testing from Millhouse's LibreElec builds.

@clivem @hifiberry @iqaudio @msperl Any comments?

hifiberry:
Having a generic I2S driver that can be activated using device trees is a really good thing. It would be cool to have a method to add ALSA controls that can run in user space. That would simplify development of drivers that might just add a simple volume control and otherwise run completely using the generic driver. Does anybody know if this has been done before?

HiassofT:
@hifiberry when using simple-card all codec controls (eg "Digital Playback Volume" on pcm5122) are available in userspace - like with separate machine drivers.

If you need some special setup with your card (eg switching between separate clocks for 44.1/48kHz rate families) or need/want to add controls that are not present in codec drivers (eg "IEC958 Playback Default" for controlling the AES bits on wm8804) this has to be done in a machine driver - simple-card can't handle that.

Software volume control, iec958 pcms etc are added via alsa card configuration files. These are looked up by driver name (with a fallback to ASoC card name if driver_name isn't explicitly set) in /usr/share/alsa/cards so that eg all intel HDA based cards automatically get the softvol+dmix plugins activated for their default PCM.

Similarly it would make sense to enable softvol for the default pcm plus the iec958 pcm eg on all wm8804 based cards.

And this is currently a bit of a mess: instead of having single machine drivers for each of the card families (like in upstream code where eg the intel hda driver handles all intel hda variants) separate ones had been added - of course all with different card and driver names and a lot of duplicated code.

This needs to be refactored, so that there's a single driver for all pcm5122 based cards, another one for all wm8804 based cards etc. Then it's easy to just add a single ALSA card config for the (eg) rpi-wm8804 machine driver and all current and future wm8804 based cards can benefit from that default config.

With simple-card unifying the config is a little bit trickier. As it only has a single name property this means you have to set this to the common card (driver) name if you want to have some default card config applied.

If you want to choose this name freely you either have to add a separate card config (matching the name) for it or add an alias for it to aliases.conf in the ALSA cards directory.

Pillar1989:
@HiassofT I still not understand how your PR works ? can you share your  iqaudio-dac (pcm5122 codec) DT source code ?

Pillar1989:
Refer to https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=8b5e2e396b589119bcc9c6a382a999e0202bae18.
You can set channel_max > 2, why your code still keep 2 ?

HiassofT:
@Pillar1989 here's the pcm5122 overlay I used for testing set_tdm_slot
https://gist.github.com/HiassofT/80fdb744f77dfc58c4cb2db876bb0e9a

As for channels: no, we can't set channels_max to > 2. The bcm2835 hardware only has 2 channel position registers so we can't support more than that. The blackfin hardware you referred to supports 8 channels, so this is quite different.

Pillar1989:
So we cannot set dai-tdm-slot-num > 2 ？

HiassofT:
You can set dai-tdm-slot-num in DT to values greater than 2 but then you also have to  set dai-tdm-slot-tx-mask and dai-tdm-slot-rx-mask to exactly 2 active channels - this is what the hardware can support.

eg:
```
dai-tdm-slot-num = <4>;
dai-tdm-slot-width = <32>;
dai-tdm-slot-tx-mask = <0 0 1 1>;
dai-tdm-slot-rx-mask = <1 0 0 1>;
```
Using more than 2 tdm slots and selecting which slots should be active will be more useful if you setup the config in a machine driver. Then you can implement runtime slot selection via alsa controls calling snd_soc_dai_set_tdm_slot.

Pillar1989:
dai-tdm-slot-num and dai-tdm-slot-width are easy to understand. What's the dai-tdm-slot-tx-mask means ? how to set  the value ??

HiassofT:
These are bitmasks that configure the time-slot allocation for the device - 1 bit per slot, bit 0 is slot 0.

eg if you have a setup with 4 time slots and 2 stereo devices and want to configure one of that devices to use slots 0+1 and the other one to use slots 2+3 you set mask of the first device to 0011b/0x03 and the mask of the second one to 1100b/0x0c

Pillar1989:
@HiassofT  I have two 4 channel ADC codec(support I2S TDM), I want to record  8 channel data. But I see your code:
```
		/*
		 * The driver is limited to 2-channel setups.
		 * Check that exactly 2 bits are set in the masks.
		 */
		if (hweight_long((unsigned long) rx_mask) != 2
		    || hweight_long((unsigned long) tx_mask) != 2)
			return -EINVAL;

```
How should I set  the dai-tdm-slot-rx-mask ??

Pillar1989:
image
My DT setting
```
                dailink0_slave: seeed-voicecard,cpu {
                    sound-dai = <&i2s>;
                    dai-tdm-slot-num = <8>;
                    dai-tdm-slot-width = <16>;
                    dai-tdm-slot-tx-mask = <0 0 1 1 0 0 0 0>;
                    dai-tdm-slot-rx-mask = <0 0 0 0 0 0 1 1>;                    
                };
                codec_dai: seeed-voicecard,codec {
                    sound-dai = <&ac108_a>;
                    clocks =  <&ac108_mclk>;
                    clock-id = <1>;
                };
```

How can I use arecord tool record 8 channel  data together ??
You can see:
pi@raspberrypi:~/seeed-voicecard $ arecord --dump-hw-params
Recording WAVE 'stdin' : Unsigned 8 bit, Rate 8000 Hz, Mono
HW Params of device "default":
ACCESS:  MMAP_INTERLEAVED RW_INTERLEAVED
FORMAT:  S16_LE S24_LE S32_LE
SUBFORMAT:  STD
SAMPLE_BITS: [16 32]
FRAME_BITS: [32 64]
CHANNELS: 2
RATE: [8000 96000]
PERIOD_TIME: (333 8192000]
PERIOD_SIZE: [32 65536]
PERIOD_BYTES: [256 524288]
PERIODS: [2 4096]
BUFFER_TIME: (666 16384000]
BUFFER_SIZE: [64 131072]
BUFFER_BYTES: [256 524288]
TICK_TIME: ALL
arecord: set_params:1233: Sample format non available
Available formats:
- S16_LE
- S24_LE
- S32_LE

Pillar1989:
Is it possible to relax the ch2 register setting for 8 channels?? @HiassofT  @flatmax

hifiberry:
The Raspberry Pi SOC only supports 2 channels.

HiassofT:
@Pillar1989 relaxing channel constraints is not the purpose of this PR. The purpose of this PR is to implement modes and functions that are documented in the BCM2835 datasheet and have been verified to work correctly.

I've been thinking a while if and if yes how the channel constraint could be relaxed (which @flatmax is currently doing with a hack). But I haven't fully decided about this yet as there are several issues to take care of:

- First of all I didn't have the time yet to do in-depth tests with more than 2 active channels - I only did a few quick tests some 2 or 3 months ago.
- It looks like despite what's written in the datasheet the frame length register is used in frame slave mode as well and bcm2835 keeps transmitting additional data instead of waiting for the next frame sync edge. So, basically, Matt's implementation seems to rely on undocumented behaviour.
- Matt's implementation also needs special hardware and software (to get the clocking right), so just hooking up some multi-channel codec won't work.
- Simply relaxing the channel constraint is going to break several implementations as non-working multi-channel-capability will be exposed to userspace for all sound cards that use a multi-channel-capable codec. So we'd have to make the relaxation strictly opt-in, eg by installing specific constraints in .startup depending on tdm slot numbers and explicitly checking for allowed/supported configurations.

As these are rather severe issues and not easy to solve the best solution IMHO for now is to make sure Matt's hack keeps working (which it still should with this PR). If you want multi-channel support and can accept the cost/restrictions they come at, use the same approach as Matt.

HiassofT:
@flatmax did you have the time yet to test this PR with your octo card? I _think_ the channel restriction hack in your machine driver should still work but it would be good if you could verify that.

flatmax:
Yes I agree.

I will make time this coming week.

Matt

Pillar1989:
@flatmax what's your special hardware and software (to get the clocking right).  Maybe my hardware clocking is same as yours

flatmax:
@HiassofT just tested tonight, looking good !
I played/recorded a tune in loopback with 8 channels and it sounded good.

Matt

HiassofT:
@flatmax thanks a lot for testting!

@pelwell if you prefer, pick the changes into 4.11 and 4.12 before merging to 4.9 to get some additional test coverage via milhouse LibreELEC builds. I'm rather confident the changes won't break anything so merging to 4.9/11/12 now should be safe as well.

pelwell:
These commits are now in rpi-4.11.y and rpi-4.12.y, so they'll be picked up by Milhouse's nightly builds.

HiassofT:
Testing in Milhouse builds worked fine, no issues reported during the last 3.5 weeks.

@pelwell care to merge?

pelwell:
Thanks, @HiassofT.

HiassofT:
@pelwell @popcornmix just got the notification that all changes were accepted upstream, so bcm2835-i2s in 4.15 should be on par with our downstream version
http://mailman.alsa-project.org/pipermail/alsa-devel/2017-November/127115.html
https://git.kernel.org/pub/scm/linux/kernel/git/broonie/sound.git/tree/sound/soc/bcm/bcm2835-i2s.c?h=for-next

git history is slightly different, as 8a6923d528cb2adbcda5e06d5d6ebdbb2a49cf0c removes the code block added by 8911a632c3cdee700d43d71dd391d5ae81b1dbcb I've dropped the latter that in the upstream commit. If git complains when rebasing, just drop all local commits and use the upstream version

pelwell:
Am I right in thinking that the upstream code won't support the 8-channel usage permitted by https://github.com/raspberrypi/linux/commit/8911a632c3cdee700d43d71dd391d5ae81b1dbcb? If so, we'll need to find a workaround because this was a useful feature.

HiassofT:
The 8-channel hack will still work, upstream bcm2835-i2s.c will be 100% identical to our current downstream version

In case the hack should break some time in the future I'd like to see a proper implementation as I outlined here https://www.raspberrypi.org/forums/viewtopic.php?f=107&t=193550&start=25#p1219382

pelwell:
That's good to know - thanks.

Issue #1709, opened 2016-11-03
USB (VCP) communication issues


carmarri:
Hello,

I've noticed some issues in the USB communication. I'll try to give us much information as possible
More information (data, pictures, etc) on previous discussion on Raspberry PI forum: https://www.raspberrypi.org/forums/viewtopic.php?f=33&t=161922&p=1062581#p1062581

**Introduction**
I'm developing a SO library (C/C++) for one of my company products, using codeblocks and cross-compiling with GCC from Ubuntu Mate. This hardware product uses FTDI and is connected to the Raspberry PI using a USB cable. At the software layer, the communication is performed using the Virtual Comm Port, since I want to allow the Library to detect any port, no matter if its VCP or a physical port. In order to handle the communications I use the kernel library commands open(), close(), read(), write(), ioctl(), termios structures, etc.
The Library has been tested from the raspberry PI through Lazarus (free-pascal) and Python, having both the same behaviour/issue.

**Issue**
If I open the port, send and receive data to/from my device, the communication runs perfect until I close the port. So what I've noticed is that, randomly (maybe once every 5 times), when I try to open the port, the port is succesfully opened (all functions return succeed) but no communication can be performed. I mean, Once the port is open, I send a fixed ASCII command to our product and I expect a fixed response (product detection through an ID). The command is sent sucessfully, in fact, no errors are reported by the write() function, then I've a timeout in the response and I don't get any data. The command should be received within 20-30ms and I've a timeout of 1 second, which I've experimentally increased during testing up to 10 seconds. The bytes available in the buffer/ports are zero. Then, no matter if I send more commands or not, the communication still fails.
So if I open the port and I'm able to communicate, everything goes fine until I close the port; but if the first command fails, then the communication keeps failing until I close the port.
I've monitored the TTL serial lines after the FTDI in my device (see link) and I've noticed that when the communication fails at the software layer, it is actually being performed at the hardware layer.
I'm using a Raspbian Lite Image to which I've installed LXDE and dependencies, since the application is targeting an embedded touch-screen system.

**Testing/Debug**
- The Raspberry PI is up to date, all packages are up to date.
- Tested on a dirty and clean Debian/Raspbian update.
- My device has both USB and RS232 ports, so I've tryed connecting to the physical serial port of the Raspberry PI and the SO library works fine 100% of times, so it always connects and communicates fine.
- I've tested other USB to RS232 chips and converters and the same error is repeated.
[...]

PR #1986, opened 2017-05-02
dwc_otg: fix split transaction data toggle handling around dequeues

P33M:
See https://github.com/raspberrypi/linux/issues/1709

Fix several issues regarding endpoint state when URBs are dequeued
- If the HCD is disconnected, flush FIQ-enabled channels properly
- Save the data toggle state for bulk endpoints if the last transfer
  from an endpoint where URBs were dequeued returned a data packet
- Reset hc->start_pkt_count properly in assign_and_init_hc()


popcornmix:
This was in last night's milhouse LE build without complaint on 4.10 kernel.

PR #1987, opened 2017-05-03
staging: vc04_services: Fix bulk cache maintenance

pelwell:
vchiq_arm supports transfers less than one page and at arbitrary
alignment, using the dma-mapping API to perform its cache maintenance
(even though the VPU drives the DMA hardware). Read (DMA_FROM_DEVICE)
operations use cache invalidation for speed, falling back to
clean+invalidate on partial cache lines, with writes (DMA_TO_DEVICE)
using flushes.

If a read transfer has ends which aren't page-aligned, performing cache
maintenance as if they were whole pages can lead to memory corruption
since the partial cache lines at the ends (and any cache lines before or
after the transfer area) will be invalidated. This bug was masked until
the disabling of the cache flush in flush_dcache_page().

Honouring the requested transfer start- and end-points prevents the
corruption.

See: #1977

Signed-off-by: Phil Elwell <phil@raspberrypi.org>

PR #1990, opened 2017-05-04
SQUASH: staging: vc04_services: Fix warnings

pelwell:
create_pagelist uses a number of WARNs to sanity-check check the input
user pages, verifying that only the first and last entries in the
array are not integral pages. These WARNs use the output array index
when it should be input array index. Although the difference is only
significant in one instance, change all the indices for consistency.

See: https://github.com/raspberrypi/linux/pull/1987


popcornmix:
Fixes warn in log for me with bulk test.

PR #1994, opened 2017-05-07
irq-bcm2836: option to repark offline CPUs

### LINKED ISSUE #1989: maxcpus=1 stops Pi2 and Pi3 from booting
As inadvertently discovered by @P33M, booting a Pi2 or Pi3 with `maxcpus=1` on the kernel cmdline.txt stops it from booting with abnormally large timestamps in the kernel log and SD timeouts (presumably because the timer is running too fast). This smells like an issue with the secondary CPU startup and/or the local interrupt controller, but I haven't identified the specific cause yet.

monkeyman79:
Option to repark offline CPUs on Raspberry Pi 2 and 3. Fixes problems
with "maxcpus" kernel command line switch and allows CPU_HOTPLUG working.

This will fix #1989, but also allows putting CPUs offline

### DIFF
--- a/arch/arm/mach-bcm/Kconfig
+++ b/arch/arm/mach-bcm/Kconfig
@@ -163,6 +163,7 @@ config ARCH_BCM2835
 config BCM2836_CPU_REPARK
 	bool "Repark offline CPUs on BCM2836"
 	depends on ARCH_BCM2835
+	depends on SMP
 	default n
 	help
 	  This config option will tell BCM2836 driver to repark CPUs not started
--- a/arch/arm/mach-bcm/Kconfig
+++ b/arch/arm/mach-bcm/Kconfig
@@ -160,6 +160,25 @@ config ARCH_BCM2835
 	  This enables support for the Broadcom BCM2835 and BCM2836 SoCs.
 	  This SoC is used in the Raspberry Pi and Roku 2 devices.
 
+config BCM2836_CPU_REPARK
+	bool "Repark offline CPUs on BCM2836"
+	depends on ARCH_BCM2835
+	default n
+	help
+	  This config option will tell BCM2836 driver to repark CPUs not started
+	  on boot to a location inside kernel image.
+
+	  Enabling this option prevents problems occurring if original spinning
+	  location gets overwritten and allows safely activating those CPUs
+	  later. It also allows putting CPUs offline again by placing them in
+	  spinning loop in the same location.
+
+	  Say Y here if you experience problems after adding "maxcpus=" to
+	  to the kernel command line or if you want to experiment with turning
+	  CPUs on and off on a Raspberry Pi 2 or 3.
+
+	  You may also want to enable HOTPLUG_CPU.
+
 config ARCH_BCM_53573
 	bool "Broadcom BCM53573 SoC series support"
 	depends on ARCH_MULTI_V7
--- a/drivers/irqchip/irq-bcm2836.c
+++ b/drivers/irqchip/irq-bcm2836.c
@@ -2,6 +2,7 @@
  * Root interrupt controller for the BCM2836 (Raspberry Pi 2).
  *
  * Copyright 2015 Broadcom
+ * Repark CPU modifications copyright 2017 Tadeusz Kijkowski
  *
  * This program is free software; you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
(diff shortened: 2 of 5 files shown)


lategoodbye:
This looks very interesting and this would be nice to have it in mainline, too.

Suggestions for further improvements:

- try to avoid assembler as much as possible
- try to use existing drivers like mbox or RPI firmware
- smp_ops should be defined in board code (arch/arm/mach-bcm) not in irqchip

monkeyman79:
@lategoodbye thanks for your feedback!

>* try to avoid assembler as much as possible

The `bcm2836_repark_loop` function is entered directly from firmware code, potentially with no stack and with MMU disabled. I don't think using assembler is avoidable at this point.

>* try to use existing drivers like mbox or RPI firmware

We are talking about bcm2836 mailboxes located at 0x40000000 memory region, not the bcm2835 mailbox.
There are 4 mailboxes per CPU. Mailbox 0 is used for IPI, mailbox 3 is used to start secondary CPU, rermaining two are not used.
As far as I can tell irq-bcm2836 is the only driver that is accessing those mailboxes.

>* smp_ops should be defined in board code (arch/arm/mach-bcm) not in irqchip

Just to make it clear: I didn't put smp_ops in irq-bcm2836, I just added new handlers.

I agree that it would be good to extract smp_ops to separate file (say `arch/arm/mach-bcm/bcm2836_smp.c`), but I'm not sure how to do it at this moment.
* irq-bcm2836.c would have to expose function to write to mailbox (e.g. `void bcm2836_write_mailbox(uint cpu, uint mbox)`, and possibly `bcm2836_read_mailbox` for completeness) which `bcm2836_smp.c` would use to wake offline CPUs
* `irq-bcm2836.c` would have to somehow pass virtual and physical base address (0x40000000) to `bcm2836_smp.c` to use in `bcm2836_repark_loop`. Alternatively, `bcm2836_smp.c` could use `of_find_compatible_node` to locate address on its own.
* how and when to call `smp_set_ops`? at this moment `irq-bcm2836.c` calls this function in irqchip driver initialization, but only of CONFIG_ARM is set. for arm64 some other method is used. we could either keep it this way - `irq-bcm2836.c` would call a function in `bcm2836_smp.c` which would install smp_ops, or instead use `CPU_METHOD_OF_DECLARE(...,"brcm,bcm2836",...)` - this would autimatically install smp_ops if compatible hardware is found in device tree, right?
* make sure not to break arm64, or maybe even write similar stuff for arm64

lategoodbye:
Sorry, forget about the first 2 points.

Marc Zyngier addressed the 3rd one:
https://www.spinics.net/lists/arm-kernel/msg580431.html

arch/arm/mach-bcm/Kconfig
```diff
@@ -160,6 +160,25 @@ config ARCH_BCM2835
 	  This enables support for the Broadcom BCM2835 and BCM2836 SoCs.
 	  This SoC is used in the Raspberry Pi and Roku 2 devices.
 
+config BCM2836_CPU_REPARK
+	bool "Repark offline CPUs on BCM2836"
+	depends on ARCH_BCM2835
```
zougloub, line 165: How about adding SMP?

drivers/irqchip/irq-bcm2836.c
```diff
@@ -316,6 +494,10 @@ static int __init bcm2836_arm_irqchip_l1_intc_of_init(struct device_node *node,
 	if (!intc.domain)
 		panic("%s: unable to create IRQ domain\n", node->full_name);
 
+#ifdef CONFIG_BCM2836_CPU_REPARK
+	bcm2836_smp_init_repark(node);
```
zougloub, line 498: WRT above comment, building without SMP fails here.

arch/arm/mach-bcm/Kconfig
monkeyman79: @zougloub right

lategoodbye:
Eric noticed me that Marc Zyngier's patch break SMP for RPI 3.

So here is my attempt to move SMP out of irqchip driver, which recently has been accepted (upstream):

http://lists.infradead.org/pipermail/linux-rpi-kernel/2017-August/006790.html

JamesH65:
No activity for over a year.  This issue will be closed within 30 days unless further interactions are posted. If you wish this issue to remain open, please add a comment. A closed issue may be reopened if requested.

JamesH65:
Closing due to lack of activity. Please request to be reopened if you feel this issue is still relevant.

Issue #1971, opened 2017-04-21
build failure: drivers/built-in.o: In function `hcd_init_fiq'


clandmeter:
Running make with [config](http://tpaste.us/JMjg) on;

```
gcc -v
Using built-in specs.
COLLECT_GCC=gcc
COLLECT_LTO_WRAPPER=/usr/libexec/gcc/armv6-alpine-linux-musleabihf/6.3.0/lto-wrapper
Target: armv6-alpine-linux-musleabihf
Configured with: /home/buildozer/aports/main/gcc/src/gcc-6.3.0/configure --prefix=/usr --mandir=/usr/share/man --infodir=/usr/share/info --build=armv6-alpine-linux-musleabihf --host=armv6-alpine-linux-musleabihf --target=armv6-alpine-linux-musleabihf --with-pkgversion='Alpine 6.3.0' --enable-checking=release --disable-fixed-point --disable-libstdcxx-pch --disable-multilib --disable-nls --disable-werror --disable-symvers --enable-__cxa_atexit --enable-default-pie --enable-cloog-backend --enable-languages=c,c++,objc,java,fortran,ada --with-arch=armv6zk --with-tune=arm1176jzf-s --with-fpu=vfp --with-float=hard --with-abi=aapcs-linux --disable-libquadmath --disable-libssp --disable-libmpx --disable-libmudflap --disable-libsanitizer --enable-shared --enable-threads --enable-tls --disable-libitm --with-system-zlib --with-linker-hash-style=gnu
Thread model: posix
gcc version 6.3.0 (Alpine 6.3.0)
```

Results in:

```
  LD [M]  drivers/usb/serial/usbserial.o
  LD      drivers/usb/built-in.o
  LD      drivers/built-in.o
  LD      vmlinux.o
  MODPOST vmlinux.o
  GEN     .version
  CHK     include/generated/compile.h
drivers/built-in.o: In function `hcd_init_fiq':
arm_pmu.c:(.text+0xd657c): undefined reference to `claim_fiq'
arm_pmu.c:(.text+0xd6638): undefined reference to `set_fiq_handler'
arm_pmu.c:(.text+0xd6690): undefined reference to `__set_fiq_regs'
arm_pmu.c:(.text+0xd6778): undefined reference to `enable_fiq'
make[2]: *** [/home/clandmeter/aports/main/linux-rpi/src/linux-4.9/Makefile:969: vmlinux] Error 1
[...]

PR #2001, opened 2017-05-10
Revert "dwc_otg: Add the dependency on FIQ"

johannes-krude:
This reverts commit 8b68f431a6168c65f67f0c6b645e9c925f90a7c7.

### DIFF
--- a/drivers/usb/host/Kconfig
+++ b/drivers/usb/host/Kconfig
@@ -764,7 +764,7 @@ config USB_HWA_HCD
 
 config USB_DWCOTG
 	tristate "Synopsis DWC host support"
-	depends on USB && FIQ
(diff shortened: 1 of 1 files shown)


popcornmix:
This was added for a reason (https://github.com/raspberrypi/linux/issues/1971) so we won't revert it.
But we can probably come up with more accurate dependency.
Does changing it to:
```
depends on USB && (FIQ || ARM64)
```
work for you?

johannes-krude:
Tested it: `depends on USB && (FIQ || ARM64)` leads to a arm64 kernel with working usb.

popcornmix:
Okay - I've pushed that change.

johannes-krude:
Thank you.

PR #2011, opened 2017-05-15
Add Goodix touch controller module and overlay support

xdsopl:
Made two patches, one that enables the module and the other adds the Goodix overlay that I use at the moment to make it work on Artista-IoT.
Would be of course very nice to make it generic, so we could use config.txt parameters to the overlay.
Suggestions welcome.
Ahmet


arch/arm/configs/bcm2709_defconfig
```diff
@@ -570,6 +570,7 @@ CONFIG_INPUT_TOUCHSCREEN=y
 CONFIG_TOUCHSCREEN_ADS7846=m
 CONFIG_TOUCHSCREEN_EGALAX=m
 CONFIG_TOUCHSCREEN_EKTF2127=m
+CONFIG_TOUCHSCREEN_GOODIX=m
```
pelwell: I can see why you put the new addition here, but the Kconfig file entries aren't in alphabetical order. If you use the `savedefconfig` make target you'll find that Kconfig puts it before the EKTF2127 (which is mysteriously absent from bcmrpi_defconfig - not your problem). Can you move it?

arch/arm/boot/dts/overlays/goodix-overlay.dts
```diff
@@ -0,0 +1,38 @@
...
+/plugin/;
+
+/ {
+	compatible = "brcm,bcm2708";
+
+	fragment@0 {
+		target = <&gpio>;
+		__overlay__ {
+			goodix_pins: goodix_pins {
+				brcm,pins = <34 35>; // interrupt and reset on i2c touch connector
```
pelwell: You must be on Compute Module. Could make the pin usage parameters so it is usable by others? See [the gpiopin parameter in the w1-gpio overlay](https://github.com/raspberrypi/linux/blob/rpi-4.9.y/arch/arm/boot/dts/overlays/w1-gpio-overlay.dts#L35) for an example.  You would need two parameters - `interrupt` and `reset`, say. `interrupt` would need to override `<&gpio>,"brcm,pins:0",
 <&gt9271>,"interrupts:0", <&gt9271>,"irq-gpios:4";`. You can work out the other by analogy.

You can leave your values as defaults.

pelwell (commented):
This looks good, except for where I've added line comments, but it does need an entry in the README. Please observe the strict formatting rules - no TABs, two blank lines between entries, same indentation.

xdsopl:
Thank you for the suggestions.
Made the fixes and forced an update.
Also removed mentioning of Artista-IoT, as it is now generic and works on the GPIO header pins (1, 3, 5, 7, 9 and 11) out of the box.
Please let me know if there still some room for improvement.
Ahmet

PR #2012, opened 2017-05-15
drm/vc4: Mark the device as active when enabling runtime PM.

anholt:
Failing to do so meant that we got a resume() callback on first use of
the device, so we would leak the bin BO that we allocated during
probe.

Fixes: 553c942f8b2c ("drm/vc4: Allow using more than 256MB of CMA memory.")


anholt:
This saves 16MB of memory.

anholt:
Slipped another commit in the branch.  I haven't boot tested it yet, but I've been using cma=384M with no alignment value in my development systems for a while now.

popcornmix:
Working okay for me. Saving memory is always good.

PR #2031, opened 2017-05-22
mpu6050 device tree overlay

T0ny0:
Allows to write in `/boot/config.txt`
`dtoverlay=mpu6050`
and use mpu6050 connected on i2c interface from userspace
Also add ability to choose interrupt pin with
`dtparam=mpu6050_interrupt_pin=<gpio_num>`


arch/arm/boot/dts/overlays/Makefile
```diff
@@ -109,7 +109,8 @@ dtbo-$(CONFIG_ARCH_BCM2835) += \
 	vga666.dtbo \
 	w1-gpio.dtbo \
 	w1-gpio-pullup.dtbo \
-	wittypi.dtbo
+	wittypi.dtbo \
+	mpu6050.dtbo
```
pelwell: All the other entries are in strict alphabetical order - please do the same. As well as being tidy and logical it helps to avoid merge conflicts.

arch/arm/boot/dts/overlays/mpu6050-overlay.dts
```diff
@@ -13,12 +13,16 @@
...
                                 compatible = "invensense,mpu6050";
                                 reg = <0x68>;
                                 interrupt-parent = <&gpio>;
                                 interrupts = <4 1>;
                         };
                 };
         };
+
+        __overrides__ {
+                mpu6050_interrupt_pin = <&mpu6050>,"interrupts:0";
```
pelwell: You don't need to qualify the parameter name with the device name - the parameters exposed by an overlay are only in scope until the next overlay is loaded, so it is safe to have multiple overlays all using the same parameter names. Other overlays have used `interrupt` or `int_pin`.

arch/arm/boot/dts/overlays/mpu6050-overlay.dts
```diff
@@ -0,0 +1,24 @@
...
+
+        fragment@0 {
+                target = <&i2c1>;
+                __overlay__ {
+                        #address-cells = <1>;
+                        #size-cells = <0>;
+                        status = "okay";
+                        clock-frequency = <400000>;
+
+                        inv-mpu6050@68 {
```
pelwell: I don't think the `inv-` prefix adds anything - the dtbindings documentation just uses `mpu6050@68`.

arch/arm/configs/bcm2709_defconfig
```diff
@@ -1171,6 +1171,8 @@ CONFIG_IIO=m
 CONFIG_IIO_BUFFER=y
 CONFIG_IIO_BUFFER_CB=m
 CONFIG_IIO_KFIFO_BUF=m
+CONFIG_INV_MPU6050_IIO=m
```
pelwell: This isn't what savedefconfig generates - I get:
```
index 9164668..575cf99 100644
--- a/arch/arm/configs/bcm2709_defconfig
+++ b/arch/arm/configs/bcm2709_defconfig
@@ -1171,15 +1171,12 @@ CONFIG_RASPBERRYPI_POWER=y
 CONFIG_EXTCON=m
 CONFIG_EXTCON_ARIZONA=m
 CONFIG_IIO=m
-CONFIG_IIO_BUFFER=y
 CONFIG_IIO_BUFFER_CB=m
-CONFIG_IIO_KFIFO_BUF=m
-CONFIG_INV_MPU6050_IIO=m
-CONFIG_INV_MPU6050_I2C=m
 CONFIG_MCP320X=m
 CONFIG_MCP3422=m
 CONFIG_DHT11=m
 CONFIG_HTU21=m
+CONFIG_INV_MPU6050_I2C=m
 CONFIG_BMP280=m
 CONFIG_PWM_BCM2835=m
```
Setting INV_MPU6050_I2C is enough to imply INV_MPU6050_IIO, IIO_BUFFER and IIO_KFIFO_BUF. The order in the Kconfig files determines where they should go.

Using `make ARCH=arm savedefconfig` minimises the number of changes, and prevents others encountering spurious differences when they subsequently use the same mechanism themselves.

pelwell (approved):
Apart from the inline comments, I'd like an entry in the README (alphabetical order, max 80 columns, no TABS, strict indentation and 2 blank lines between entries) and a patch to bcmrpi_defconfig as well.

Otherwise it looks fine.

pelwell (changes requested):
* clicked the wrong option *

T0ny0:
Add fixes according to your comments

pelwell:
Somehow I think you missed out bcmrpi_defconfig in your commits. If you could also change the node name from `inv-mpu6050@68` to `mpu6050@68` I would be happier, but I'm not going to demand it because it is only cosmetic.

pelwell (approved):

pelwell:
Merged.

FYI you can also put parameters on the same line, e.g.
```
dtoverlay=mpu6050,interrupt=<gpio_num>
```

Issue #2020, opened 2017-05-17
USB Audio Issue with Raspberry Pi Zero W


byarmis:
When I play music (either `mpg123` on an mp3 file or pink noise via `speaker-test`), there is popping and crackling in the audio output.  If I let `speaker-test` run for a bit (20 seconds or so), the popping and crackling will go away.  The crackling does not go away after playing a song for an even longer amount of time (still crackling after several minutes).

I have tentatively narrowed it down to the USB frequency.  If I run `watch -n0 "cat /proc/asound/card1/stream0"`, I can see that the "Momentary freq" is reported to be 50 Hz higher or lower than it should be (it varies whether it's higher or lower) when there's crackling:

```
AudioQuest AudioQuest DragonFly Black v1.5 at usb-20980000.usb-1, full speed : U
SB Audio

Playback:
  Status: Running
    Interface = 1
    Altset = 1
    Packet Size = 360
    Momentary freq = 48050 Hz (0x30.0ccc)
    Feedback Format = 10.14
  Interface 1
    Altset 1
    Format: S24_3LE
    Channels: 2
    Endpoint: 1 OUT (ASYNC)
    Rates: 44100, 48000, 88200, 96000
```
When the crackling goes away (when running a `speaker-test`), the momentary frequency drops down (or increases) to where it should be:

```
AudioQuest AudioQuest DragonFly Black v1.5 at usb-20980000.usb-1, full speed : U
SB Audio

Playback:
  Status: Running
    Interface = 1
    Altset = 1
    Packet Size = 360
    Momentary freq = 48000 Hz (0x30.0000)
    Feedback Format = 10.14
  Interface 1
    Altset 1
    Format: S24_3LE
    Channels: 2
    Endpoint: 1 OUT (ASYNC)
    Rates: 44100, 48000, 88200, 96000
```
In [Issue #888](https://github.com/raspberrypi/linux/issues/888), I found a suggestion to try adding either `dwc_otg.fiq_fsm_enable=0` or `dwc_otg.fiq_fsm_mask=0x3` to `/boot/cmdline.txt`; neither fixed the issue.
 
## Hardware

* Raspberry Pi Zero W
* USB OTG Adapter
* [Audioquest Dragonfly Black USB DAC](http://www.audioquest.com/dragonfly-series/)

### uname -a

```Linux raspberrypi 4.9.28+ #999 Tue May 16 13:59:16 BST 2017 armv6l GNU/Linux```

### lsusb -v

```
[...]

PR #2038, opened 2017-05-25
dwc_otg: make periodic scheduling behave properly for FS buses

P33M:
see #2020 

This is a fix for the default parameters (microframe scheduling) not behaving well when the root port is in full-speed mode.

As a semi-bandaid fix, transfers in FS mode are not nominated for execution by FIQ code - there's still a bug in there relating to isochronous transfers misbehaving when the endpoint is being shutdown.


popcornmix:
No complaints from last night's Milhouse LibreELEC build.

PR #2040, opened 2017-05-27
Add sysfs entries to control RJ45 LEDs on RPi2.

JoeSchubert:
0 turns the LED off, 1 turns the LED on. All nodes default to 1 so they will maintain normal behavior unless overridden.

/sys/class/smsc95xx_leds/eth_fdx
/sys/class/smsc95xx_leds/eth_lnk
/sys/class/smsc95xx_leds/eth_spd

This commit was found here: https://www.anginf.de/?p=433 and I am only submitting it so that all RPi2 owners can benefit.

There are a number of usecases where this is useful. For instance, HTPC usage of the RPi2... Currently users can disable the LEDs on the front of the board. However, the RJ45 LEDs flashing can be a bit of a nuisance.

### DIFF
--- a/drivers/net/usb/smsc95xx.c
+++ b/drivers/net/usb/smsc95xx.c
@@ -1244,6 +1244,9 @@ static int smsc95xx_reset(struct usbnet *dev)
 	if (ret < 0)
 		return ret;
 
+	// Init sysfs led controller
+	smsc95xx_ledctl(dev);
+
 	/* Init Tx */
 	ret = smsc95xx_write_reg(dev, FLOW, 0);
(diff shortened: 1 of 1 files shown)


popcornmix:
I believe this can be achieved without kernel modifications.
See https://www.raspberrypi.org/forums/viewtopic.php?f=63&t=72070 and https://forum.libreelec.tv/thread/6332-ability-to-disable-ethernet-leds/?postID=50829#post50829

JoeSchubert:
Thanks. I don't know why neither of those came up when I originally was searching for a way to do this.

PR #2041, opened 2017-05-29
Add device tree config for htu21

rumpelsepp:


pelwell:
That's a good start. Can you:
1. Keep the sensors in alphabetical order.
2. Add htu21 to the README (try `dtoverlay -h i2c-sensor` in a shell). Stick the the strict whitespace rules - no TABs, same indentation.

If you fix those in one or more follow-up commits we can squash them when merging.

rumpelsepp:
Updated.

pelwell:
Your README entry says the _default_ address is 0x40. Can the device have other addresses? If so, we should make the addresses selectable through a parameter. If not, you should remove the word "default".

rumpelsepp:
> Your README entry says the default address is 0x40. Can the device have other addresses? If so, we should make the addresses selectable through a parameter. If not, you should remove the word "default".

I will check this and report back, can't remember it atm.

rumpelsepp:
It seems, that the address is hardcoded: https://cdn-shop.adafruit.com/datasheets/1899_HTU21D.pdf

I will remove the address configuration opportunity.

pelwell:
You need the parameter, otherwise there is no way to select your sensor. Put the parameter back, leave everything else as it is and I'll merge it.

pelwell:
The i2c-sensor overlay has now been restructured internally, breaking your PR. I'll patch it up and merge by hand.

rumpelsepp:
Oh nice, thanks! Then I can close this?

pelwell:
I'll do it when I've finished - give me a few minutes.

rumpelsepp:
No need to hurry. :) thanks!

pelwell:
Merged manually.

PR #2045, opened 2017-05-30
Add device tree overlay for bme280 via spi

rumpelsepp:
Maybe not quite there, but enough to get feedback. This overlay works, but I am not sure how you want to include it in the project.

### DIFF
--- a/arch/arm/boot/dts/overlays/bme280-spi.dts
+++ b/arch/arm/boot/dts/overlays/bme280-spi.dts
@@ -1,0 +1,32 @@
+/dts-v1/;
+/plugin/;
+
+/ {
+	compatible = "brcm,bcm2708";
+
+	fragment@0 {
+		target = <&spidev0>;
+
+		__overlay__ {
+			status = "disabled";
+		};
+	};
+
+	fragment@1 {
+		target = <&spi0>;
+
+		__overlay__ {
+			status = "okay";
+			#address-cells = <1>;
+			#size-cells = <0>;
+
+			bme280@0 {
+				compatible = "bosch,bme280";
+				reg = <0x0>;
+				spi-max-frequency = <500000>;
+				default-oversampling = <3>;
+				status = "okay";
+			};
+		};
+	};
+};


pelwell:
That's a reasonable start. However, I'd like to see:
1. Turn this into a generic SPI sensor overlay (equivalent to `i2c-sensor`) with `bme280` as the only parameter (for now) - call the file `spi-sensor-overlay.dts`. See below for a suggested implementation.
2. The overlay needs to be added to the Makefile in the overlays directory, otherwise it won't be built. Please keep the additional entry in alphabetical order.
3. All overlays must be documented in the README file, so that users can type `dtoverlay -h spi-sensor` and find out how to use it. Copy the formatting of entries exactly, including indentation, blank line spacing and no TABS.

Here's a suggested implementation. It is currently limited to spi0.0, but in future we can add support for other buses and chip selects.
```
/dts-v1/;
/plugin/;

/ {
	compatible = "brcm,bcm2708";

	fragment@0 {
		target = <&spidev0>;

		__overlay__ {
			status = "disabled";
		};
	};

	fragment@1 {
		target = <&spi0>;

		__dormant__ {
			status = "okay";
			#address-cells = <1>;
			#size-cells = <0>;

			bme280@0 {
				compatible = "bosch,bme280";
				reg = <0x0>;
				spi-max-frequency = <500000>;
				default-oversampling = <3>;
				status = "okay";
			};
		};
	};

	__overrides__ {
		bme280 = <0>, "+1";
	};
};
```

rumpelsepp:
Will do this.

rumpelsepp:
Since I am on vacation for one week, starting with today, I have to delay this a bit. I hope this is ok. :)

pelwell:
Take your time - just post a reply when you've got something for me to look at.

pelwell:
This PR has been awaiting input from the originator for 18 months, so closing until it gets picked up again.

rumpelsepp:
Oh. thanks for the pointer, I completely forgot about this.

PR #2048, opened 2017-06-01
Allo Digione Driver

sandeepal:
Allo Digione is a high-quality S/PDIF output board for the Raspberry Pi. The S/PDIF interface chip supports up to 192kHz/24bit resolution with dual low-jitter oscillators.


pelwell:
Sigh - another clone.

pelwell:
Don't you want this to work on Pi 1s and Zeroes? You've not updated bcmrpi_defconfig.

allocom:
Added to bcmrpi_defconfig. Thanks.

clivem:
> Sigh - another clone.

From a driver POV, yes. From a simple functional standpoint, (I2S in, SPDIF out), yes. But from a hardware perspective, if I understand what they have done on this board, (not that I've got one), the voltage regs/filtering, isolation, and flip flop re-clock on the output are what make it an "audiophool" absolute minimal jitter solution, rather than just a rip-off of Daniel's HB Digi Pro WM8804 board with the 2x OSC.

Still, there needs to be just one "unified" machine driver for the WM8804 solutions rather than continuing the cut'n'paste merry-go-round.........

pelwell:
Thanks for the insight. And yes, I'm also looking forward to the grand unification.

sandeepal:
Functionality of the digione card is same as other digi cards, so the code is similar, but we have some change in DTS file for the card to work. Also using different company's machine driver for our card, questions may be raised from other companies. Is there a way to reuse the code & proceed.

pelwell:
Other companies can't raise questions about others using their drivers - the code is public and reuse is to be encouraged - but your customers would expect the name of your card to appear in their sound settings.

I'll merge this now, but don't be surprised when the drivers get rationalised.

Issue #2056, opened 2017-06-05
CONFIG_DMA_BCM2708 doesn't support module compilation


agherzan:
CONFIG_DMA_BCM2708 is defined as a tristate config in https://github.com/raspberrypi/linux/blob/204050d0eafb565b68abf512710036c10ef1bd23/drivers/dma/Kconfig .

The code for the above config includes https://github.com/raspberrypi/linux/blob/rpi-4.9.y/include/linux/platform_data/dma-bcm2708.h which defines (for example) bcm_dmaman_probe based on CONFIG_DMA_BCM2708 definition. In the case of CONFIG_DMA_BCM2708 set on module, CONFIG_DMA_BCM2708_MODULE ends up defined and not CONFIG_DMA_BCM2708 which breaks compilation while hitting a redefinition error.

One solution would be to restrict CONFIG_DMA_BCM2708 to bool. A second option would be to take into consideration CONFIG_DMA_BCM2708_MODULE too in dma-bcm2708.h . I can provide patch for either of these options.

popcornmix:
The amount of code in bcm2708-dmaengine.c is tiny, so no big loss if it is forced to be boolean.
But if fixing it up to build as a module is trivial then I'm fine with that.
@pelwell any preference?

agherzan:
I was able to compile it with a simple change in the dma-bcm2708.h which doesn't only checks for CONFIG_DMA_BCM2708 but for CONFIG_DMA_BCM2708_MODULE too. So both solutions seem trivial.

pelwell:
If it really is that simple then let's keep the option of a module build - submit a Pull Request.

Thanks in advance.

PR #2057, opened 2017-06-05
dma-bcm2708: Fix module compilation of CONFIG_DMA_BCM2708

agherzan:
bcm2708-dmaengine.c defines functions like bcm_dma_start which are
defined as well in dma-bcm2708.h as inline versions when
CONFIG_DMA_BCM2708 is not defined. This works fine when
CONFIG_DMA_BCM2708 is built in, but when it is selected as module build
fails with redefinition errors because in the build system when
CONFIG_DMA_BCM2708 is selected as module, the macro becomes
CONFIG_DMA_BCM2708_MODULE.

This patch makes the header use CONFIG_DMA_BCM2708_MODULE too when
available.

Fixes https://github.com/raspberrypi/linux/issues/2056

Signed-off-by: Andrei Gherzan <andrei@gherzan.com>

### DIFF
--- a/include/linux/platform_data/dma-bcm2708.h
+++ b/include/linux/platform_data/dma-bcm2708.h
@@ -75,7 +75,7 @@ struct bcm2708_dma_cb {
 struct scatterlist;
 struct platform_device;
 
-#ifdef CONFIG_DMA_BCM2708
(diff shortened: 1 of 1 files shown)


pelwell:
Thanks.

Issue #1297, opened 2016-02-22
v4l2, why need VIDIOC_DQBUF 2 times to get the latest frame


dodoma:
hello, I have a raspberry pi 2 mode b+

I want to take a still image from camera through v4l2 api,
This is my code http://paste.ubuntu.com/15099128/, 
my question is why do I need ioctl VIDIOC_QBUF and VIDIOC_DQBUF 2 times to get the latest frame?
Thanks.

6by9:
The underlying GPU code is structured for efficient pipelines. If you starve it of buffers, it will have taken the capture(s) and is waiting for a buffer to write that image(s) into. As soon as it has managed that, it will take the next capture and wait for a buffer again.
In the case of H264, there's an ~2MB FIFO in the codec that you will be filling before you actually stall the camera itself.

You're not reducing latency anyway by waiting for the key press before asking for the frame as the sensor will have to deliver the "current" frame, and then capture the new one. You'd be better off constantly circulating buffers, and on key press store the next one to be delivered.

NB I believe the bug still exists where streaming multiple JPEGs like this doesn't allow AE to run, so the exposure settings may be suboptimal.

dodoma:
en, you means try circulating buffers?
I am afraid it's too complex and I can't handle it correctly, would you point it more clear, or, some skeleton code for this will be very helpfull.

The demand is very clear, I want capture a still jpeg image as soon as possible when 'c' pressed(or when target ready notify by GPIO on produce system)

6by9:
Totally hacked together pseudocode.

```
//Allocate buf
ioctl(QBUF, &buf);
while(1) {
  ioctl(DQBUF, &buf);
  if(kbhit()) {
    char c=getch();
    if (c== 'c')
      //do stuff with image
  }
  ioctl(QBUF, &buf);
}
```

So you're constantly asking for buffers, and when you actually want one, you grab the next one available.
(kbhit is a Windows function, but you'll find people asking for equivalent code all over the internet, eg http://cboard.cprogramming.com/c-programming/63166-kbhit-linux.html It's actually cleaner if you're polling the state of a GPIO instead)

If you want better efficiency, switch to using several MMAPed buffers from V4L2 instead of V4L2_MEMORY_USERPTR. It just saves a chunk of work under the bonnet.

dodoma:
ok,very clear, I will try it tomorrow.

dodoma:
hi, 6by9

[...]

PR #2064, opened 2017-06-09
v4l2 - ioctl 2 times bugfix on still mode

dodoma:
Avoid 2 times VIDIOC_QBUF, VIDIOC_DQBUF call on JPEG format capture.
#1297


JamesH65:
@6by9 Any thoughts?

JamesH65:
@6by9 Ping

Issue #2023, opened 2017-05-18
Unable to use more than 2 'Atheros AR9271' WiFi Dongles


pi-resource:
More than two WiFi Dongel with the Atheros AR9271 chipsets are not identified. In this example I used a TP-LINK TL-WN722N. I can confirm there were NO power issues. See dmesg outputs below.

First device plugged in: (all normal)
 [ 3626.648706] usb 1-1.2.1: new high-speed USB device number 6 using dwc_otg
 [ 3626.799647] usb 1-1.2.1: New USB device found, idVendor=0cf3, idProduct=9271
 [ 3626.799661] usb 1-1.2.1: New USB device strings: Mfr=16, Product=32, SerialNumber=48
 [ 3626.799669] usb 1-1.2.1: Product: USB2.0 WLAN
 [ 3626.799677] usb 1-1.2.1: Manufacturer: ATHEROS
 [ 3626.799684] usb 1-1.2.1: SerialNumber: 12345
 [ 3627.944779] usb 1-1.2.1: ath9k_htc: Firmware ath9k_htc/htc_9271-1.4.0.fw requested
 [ 3627.944907] usb 1-1.2.1: Direct firmware load for ath9k_htc/htc_9271-1.4.0.fw failed with error -2
 [ 3627.944922] usb 1-1.2.1: ath9k_htc: Firmware htc_9271.fw requested
 [ 3627.945481] usbcore: registered new interface driver ath9k_htc
 [ 3628.241745] usb 1-1.2.1: ath9k_htc: Transferred FW: htc_9271.fw, size: 51272
 [ 3628.487600] ath9k_htc 1-1.2.1:1.0: ath9k_htc: HTC initialized with 33 credits
 [ 3633.007736] ath9k_htc 1-1.2.1:1.0: ath9k_htc: FW Version: 1.3
 [ 3633.007748] ath9k_htc 1-1.2.1:1.0: FW RMW support: Off
 [ 3633.007755] ath: EEPROM regdomain: 0x809c
 [ 3633.007759] ath: EEPROM indicates we should expect a country code
 [ 3633.007764] ath: doing EEPROM country->regdmn map search
 [ 3633.007769] ath: country maps to regdmn code: 0x52
 [ 3633.007775] ath: Country alpha2 being used: CN
 [ 3633.007779] ath: Regpair used: 0x52
 [ 3633.034162] ieee80211 phy1: Atheros AR9271 Rev:1
[...]

PR #2067, opened 2017-06-12
dwc_otg: add module parameter int_ep_interval_min

P33M:
Add a module parameter (defaulting to ignored) that clamps the polling rate
of high-speed Interrupt endpoints to a minimum microframe interval.

The parameter is modifiable at runtime as it is used when activating new
endpoints (such as on device connect).

Related: #2023


popcornmix:
Looks fine. If you want to enforce power of two then you could use (1<<int_ep_interval_min), but I'm happy to merge it as is.

P33M:
Technically using any integer interval would work, so it seems unnecessarily restrictive to enforce it. The USB spec requires servicing intervals in descriptors to be a power of two, but I doubt there would be any practical effect as we're servicing endpoints slower than requested anyway.

Issue #2024, opened 2017-05-19
Invalid CRC in USB packets on Pi Zero 


yoctopuce:
We have found a bug in the USB stack of the Raspberry PI image.

Using an USB analyser we have found that sometime the Raspberry Pi Zero send USB packets with Invalid CRC. The strange thing is that this issue affects only Raspberry Pi Zero with Raspbian image after March 2016. The same image work perfectly on a Raspberry Pi 2 or 3.

We have publish on GitHub a complete description of the problem and a program that exhibits this issue:
https://github.com/yoctopuce-examples/raspberry_pi_zero_bug

JamesH65:
@P33M One for you I think?

@yoctopuce Can you confirm March **2016** as the date it appears to start going wrong.

P33M:
There's two ways to interpret the statement. Has the Pi Zero ever been reliable on any version of firmware? If so, what was the last good firmware revision/distro?

lategoodbye:
What is the behavior on a Raspberry Pi 1?
Is it possible to reproduce the issue without a Yoctopuce device?

P33M:
To narrow down if a specific firmware version broke functionality, can you test firmware revisions retrieved via rpi-update:

https://github.com/Hexxeh/rpi-firmware/commits/master

Running `sudo rpi-update <hash>` will get the firmware revision corresponding to the hash in the repo above.  A rudimentary pseudo-bisect should get to the first broken revision relatively quickly.

yoctopuce:
Hi all,

Sorry to have took so long to answers, but we are quite busy these days.

@JamesH65 Official image of 18 March 2016 is the last one that *IS WORKGING". Wee log bellow:
         
```
pi@raspberrypi:~/raspberry_pi_zero_bug $ sudo ./bug_pizero
this is a simple program that exhibit a bug in the USB stack
of the Raspberry PI Zero and any Yoctopuce device
Use libUSB v1.0.19.10903
List all Yoctopuce devices present on this host:
 - USB dev: RELAYHI1-85D28 (24E0:D:0)
1 Yoctopuce devices are present
\ Test device RELAYHI1-85D28 (24E0:D)
+ Send USB pkt no 0
- rd_callback for RELAYHI1-85D28 : response 0 is valid (len=64)
+ Send USB pkt no 1
- rd_callback for RELAYHI1-85D28 : response 1 is valid (len=64)
+ Send USB pkt no 2
[...]

PR #2077, opened 2017-06-20
dwc_otg: fiq_fsm: Add non-periodic TT exclusivity constraints

P33M:
Certain hub types do not discriminate between pipe direction (IN or OUT)
when considering non-periodic transfers. Therefore these hubs get confused
if multiple transfers are issued in different directions with the same
device address and endpoint number.

Constrain queuing non-periodic split transactions so they are performed
serially in such cases.

Related: https://github.com/raspberrypi/linux/issues/2024


popcornmix:
No complaints from last night's LibreELEC build which included this.

PR #2085, opened 2017-06-24
Adding IPVLAN module to config

### LINKED ISSUE #2083: Has ipvlan been removed?
Hi, I was doing some testing with folks in the Docker community and we found the `ipvlan` Kernel module   disappears when we do an `rpi-update`. Apparently Justin Cormack from Docker Inc raised an issue here last year asking for the module to be included.

Please could it be reintroduced to the Kernel? At the moment Docker Swarm will break if we do an `dpi-update`.

Thanks,

Alex

hairyhenderson:
Fixes #2083

Signed-off-by: Dave Henderson <dhenderson@gmail.com>


pelwell:
Your patch doesn't work for me. It applies, but using the new defconfigs doesn't result in CONFIG_IPVLAN being defined, and running savedefconfig removes your additions.

It looks like there is an unmet dependency on NET_L3_MASTER_DEV. Do you want to try again?

alexellis:
This is still broken for Stretch 👎 - that means Docker Swarm is broken on RPi. @hairyhenderson @pelwell can either of you help? https://github.com/alexellis/docker-arm/issues/16#issuecomment-323573334

@DieterReuter do you still work with the Raspberry Pi?

alexellis:
Ref: https://github.com/raspberrypi/linux/issues/2083

alexellis:
hi @popcornmix can you help us with this? It blocks Docker Swarm from working properly on RPi.

hairyhenderson:
sorry - I'm going to have to give up on this issue... I thought it was going to be simple, and I really don't have the time or head-space to dig much more deeply into it 😖

remonlam:
@hairyhenderson I think this one can be close, because it has been solved in #2083

PR #2092, opened 2017-06-29
New i2c-rtc-gpio device overlay

ConnerLabs:
Made by combining i2c-rtc and i2c-gpio overlays. Tested with PCF2127 RTC chip on CM3.

### DIFF
--- a/arch/arm/boot/dts/overlays/Makefile
+++ b/arch/arm/boot/dts/overlays/Makefile
@@ -42,6 +42,7 @@ dtbo-$(CONFIG_ARCH_BCM2835) += \
 	i2c-mux.dtbo \
 	i2c-pwm-pca9685a.dtbo \
 	i2c-rtc.dtbo \
+	i2c-rtc-gpio.dtbo \
 	i2c-sensor.dtbo \
 	i2c0-bcm2708.dtbo \
 	i2c1-bcm2708.dtbo \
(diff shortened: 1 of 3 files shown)


pelwell:
I've made a few largely cosmetic changes, but reading it now I think we should modify it just a bit so that it doesn't preclude another i2c-gpio overlay being used.

I suggest you:
1. Delete fragments 1 & 2, renumbering the rest (sorry). This is because we don't want to share this bus with other overlays.
2. Rename the node in fragment 0 to `i2c-gpio-rtc@0` to avoid the clash.

Test the result, and if it works submit the changes - I can squash them all into one neat commit at merge time.

ConnerLabs:
Fragment 0 contains the following:
__overlay__ { i2c_gpio: i2c@0 {
Should I also change the i2c_gpio: part, and change target=<&i2c_gpio> to match in the other fragments I pasted from i2c-rtc-overlay.dts? Or is it enough just to change the i2c@0 to i2c-gpio-rtc@0?

pelwell:
Labels (like `i2c_gpio`) are local to the overlay so you can leave them as they are - only the node name (`i2c@0`) has to change. If you would prefer to change the label, make sure you change all occurrences.

ConnerLabs:
OK, I'll only change the node name then.

ConnerLabs:
Done and tested.

PR #2100, opened 2017-07-06
config: Add CONFIG_W1_SLAVE_DS2438

manio:
Hi guys!
Please also include this 1wire slave module in the config. It is a smart battery monitor but used not only in battery monitoring, but in other applications, such as humidity and weather monitoring.

More info about it here in the module doc:
https://github.com/raspberrypi/linux/blob/rpi-4.12.y/Documentation/w1/slaves/w1_ds2438


pelwell:
Rebasing mangled your commit, but the config setting is now in rpi-4.12.y.

PR #2101, opened 2017-07-06
[4.12] config: add missing arizona regulator modules

HiassofT:
In kernel 4.12 CONFIG_REGULATOR_ARIZONA was replaced by 2 separate
options for LDO1 and MICSUPP regulators. Enable these, they are
needed by the Cirrus Logic Audio Card.

Also regenerate configs with make savedefconfig to get rid of
the duplicated CONFIG_TOUCHSCREEN_EDT_FT5X06 entry.


popcornmix:
Looks reasonable to me. @pelwell okay with you?

pelwell:
No objections from me.

PR #2104, opened 2017-07-07
config: enable generic S/PDIF codec drivers

HiassofT:
These drivers can be used as dummy ADC/DAC drivers for
attaching general codecs that don't need to be configured.

This option will build 2 additional drivers, spdif_receiver
and spdif_transmitter.

Since these drivers have DT bindings they are handy for quick
testing of I2S peripherals with simple-audio-card.

eg:

```
fragment@0 {
    target-path = "/";
    __overlay__ {
        dummy_receiver: spdif-receiver {
            #address-cells = <0>;
            #size-cells = <0>;
            #sound-dai-cells = <0>;
            compatible = "linux,spdif-dir";
            status = "okay";
        };
    };
};
```


HiassofT:
Thanks for merging so quickly!

Could you pick that into 4.11 and 4.12 as well?

pelwell:
Of course - there are a few others to port across as well.

PR #2108, opened 2017-07-09
rtl8192: switch to netdev->priv_destructor()

bamarni:
When trying to build from the rpi-4.11.y branch, I'm getting the
following error :

    drivers/net/wireless/realtek/rtl8192cu/os_dep/linux/ioctl_cfg80211.c:3464:10: error: 'struct net_device' has no member named 'destructor'

It seems to occur since this upstream commit : https://github.com/torvalds/linux/commit/cf124db566e6b036b8bcbe8decbed740bdfac8c6

> [...]
>
> netdev->priv_destructor() performs all actions to free up the private
resources that used to be freed by netdev->destructor(), except for
free_netdev().
>
> netdev->needs_free_netdev is a boolean that indicates whether
free_netdev() should be done at the end of unregister_netdevice().

Signed-off-by: Bilal Amarni <bilal.amarni@gmail.com>

### DIFF
--- a/drivers/net/wireless/realtek/rtl8192cu/os_dep/linux/ioctl_cfg80211.c
+++ b/drivers/net/wireless/realtek/rtl8192cu/os_dep/linux/ioctl_cfg80211.c
@@ -3461,7 +3461,8 @@ static int rtw_cfg80211_add_monitor_if(_adapter *padapter, char *name, struct ne
 #if (LINUX_VERSION_CODE >= KERNEL_VERSION(4,1,0))
 	mon_ndev->name_assign_type = name_assign_type;
 #endif
-	mon_ndev->destructor = rtw_ndev_destructor;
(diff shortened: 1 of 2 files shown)


pelwell:
This looks OK by analogy to the changes to other drivers. We normally expect changes to upstream code to go upstream first, but since the 4.11 tree is essentially broken now and the fix is likely to look just like this I'm happy to take this patch for now.

Any objections, @popcornmix ?

popcornmix:
I'm just compile testing. I did the ``mon_ndev->needs_free_netdev = true;`` for 4.12.
4.11 didn't initially need it, but it looks like the last minor bump has made it required on 4.11 too.
The ``priv_destructor`` part isn't on 4.12 but that looks valid.
I'll merge when build succeeds.

popcornmix:
Thanks!

PR #2118, opened 2017-07-16
Update i2c1-bcm2708-overlay.dts

soodvarun78:
Removing fragment@1 as it is not required. 
The fragment 1 creates another device node  with name i2c1_pins which is already created by dts files bcm2710-rpi-3-b.dts, bcm2708-rpi-b-plus.dts and bcm2709-rpi-2-b.dts.

### DIFF
--- a/arch/arm/boot/dts/overlays/i2c1-bcm2708-overlay.dts
+++ b/arch/arm/boot/dts/overlays/i2c1-bcm2708-overlay.dts
@@ -19,16 +19,6 @@
       };
    };
 
-   fragment@1 {
-      target = <&gpio>;
-      __overlay__ {
-         i2c1_pins: i2c1 {
-            brcm,pins = <2 3>;
-            brcm,function = <4>; /* alt0 */
-         };
-      };
-   };
-
    __overrides__ {
       sda1_pin = <&i2c1_pins>,"brcm,pins:0";
       scl1_pin = <&i2c1_pins>,"brcm,pins:4";


pelwell:
Have you tested the modified overlay? Although you've spotted an undesirable anomaly, with your patch the overlay no longer applies.

There was a time, over a year ago, when the overlay was merged with the base tree before the parameters were applied, but that was changed so that parameters modify the overlay before merging with the base tree in order to fit with the way the kernel applied overlays. Although this sounds like an implementation detail, a consequence is that parameters/overrides can only refer to symbols within the overlay. The workaround is to create a fragment that targets the symbol and populate it with the properties that the parameters have to change (*).

A working version that removes the duplication looks like this:
```
/*
 * Device tree overlay for i2c_bcm2708, i2c1 bus
 *
 * Compile:
 * dtc -@ -I dts -O dtb -o i2c1-bcm2708-overlay.dtb i2c1-bcm2708-overlay.dts
 */

/dts-v1/;
/plugin/;

/{
   compatible = "brcm,bcm2708";

   fragment@0 {
      target = <&i2c1>;
      __overlay__ {
         pinctrl-0 = <&i2c1_pins>;
         status = "okay";
      };
   };

   fragment@1 {
      target = <&i2c1_pins>;
         pins: __overlay__ {
         brcm,pins = <2 3>;
         brcm,function = <4>; /* alt 0 */
      };
   };

   __overrides__ {
      sda1_pin = <&pins>,"brcm,pins:0";
      scl1_pin = <&pins>,"brcm,pins:4";
      pin_func = <&pins>,"brcm,function:0";
   };
};
```

(*) N.B. Although it is permissible for a parameter to create a property that doesn't exist, if the parameter is only modifying a part of the property then the remainder will be filled with zeroes to make it long enough (unless other parameters overwrite the rest), which probably isn't what you want. Therefore it is safer (but less flexible) to provide a default property value.

yaode:
plz unsubscribe the thread. thx.

2017-07-17 16:52 GMT+08:00 Phil Elwell <notifications@github.com>:

pelwell:
https://help.github.com/articles/unwatching-repositories/

soodvarun78:
Hi Phil ,

Thanks for the input. 
Yes I have tested my commit. It was working.
The reason for  commit , is that if we run the command 
sudo dtoverlay i2c1-bcm2708 
from command line , the probe function of i2c  was failing because of an error generated in PinControl subsystem with dmesg 'invalid phandle'. 
So I removed the fragment@1 and probe was successful for the  reason I have mentioned in commit comment.

I verified your patch and it is working... It is much more cleaner than me.

Regards
Varun Sood

pelwell:
The modified patch is now in the 4.9 tree.

Issue #10, opened 2012-04-26
wmii colour misrepresentation


colons:
When running [wmii-hg](http://aur.archlinux.org/packages.php?ID=3497) on Arch Linux Arm on the Raspberry Pi with [my favourite wmiirc](https://bitbucket.org/colons/dotfiles/src/3f4ee6a307e6/wmiirc), the window decoration colours are modified in a bizarre way I have yet to identify.

Screenshots:
distortion: [pi](http://dump.musicfortheblind.co.uk/gallery/photos/rpiwmii/rudiment.png) ([photo](http://dump.musicfortheblind.co.uk/gallery/photos/rpiwmii/IMG_0771.JPG))
expected: [arch x86 box](http://dump.musicfortheblind.co.uk/gallery/photos/rpiwmii/banana.png) ([photo](http://dump.musicfortheblind.co.uk/gallery/photos/rpiwmii/IMG_0772.JPG))

What's weird about this (and what leads me to believe it's a low-level issue) is that when run in a tightvnc x server, the colours load and render fine. Additionally, everything else under the broken wmii instances works fine — images, webpages, etc. render perfectly. I assume this means it's a bug in whatever system wmii uses to initially select its colours.

If this doesn't belong here, I apologise and will be willing to resubmit elsewehere.

popcornmix:
There's an update to kernel source that might fix this. Either rebuild kernel, or wait for an Arch udpate and report back.

PR #2119, opened 2017-07-16
tinydrm backport

notro:
This is a backport of drivers/gpu/tinydrm which was added in 4.11, a couple of patches from 4.12 and a driver from (to be) 4.14.

tinydrm is the future of drivers/staging/fbtft. fbtft can't move out of staging since fbdev is closed for new drivers.

This PR has 2 drivers:

mi0283qt
--------

Works with rpi-display by Watterott and pitft28 by Adafruit.
This driver provides one main benefit over fbtft and that is [PRIME](https://en.wikipedia.org/wiki/Direct_Rendering_Manager#DMA_Buffer_Sharing_and_PRIME) support which makes it possible to do rendering on vc4 and scanout on these SPI connected displays. However, I haven't been able to make it work with real life applications yet (X windows): [vc4 issue #10](https://github.com/anholt/linux/issues/10). The driver might also perform better depending on use case.

Here are overlays if someone wants to try:
- [rpi-display-overlay.dts](https://github.com/notro/tinydrm/blob/master/rpi-overlays/rpi-display-overlay.dts)
- [pitft28-resistive-overlay.dts](https://github.com/notro/tinydrm/blob/master/rpi-overlays/pitft28-resistive-overlay.dts)

repaper
-------

Works with PaPiRus ePaper Screens by [Pi Supply](https://www.pi-supply.com/). fbtft doesn't support these and this the main motivation for doing this backport.

- [PaPiRus ePaper / eInk Screen HAT for Raspberry Pi](https://www.pi-supply.com/product/papirus-epaper-eink-screen-hat-for-raspberry-pi/)
- [PaPiRus Zero – ePaper / eInk Screen pHAT for Pi Zero](https://www.pi-supply.com/product/papirus-zero-epaper-screen-phat-pi-zero/)

DRM core patches
----------------

During rebasing I've tried to avoid touching core drm files to keep this as contained as possible.
I was left with 2 patches:

drm: debugfs: Remove all files automatically on cleanup
For vc4 this patch doesn't change anything.

drm: Create DEFINE_DRM_GEM_CMA_FOPS and roll it out to drivers
I have stripped this down to just adding the macro, not touching drivers.

I get this warning that I don't know how to get rid of:
```
  DTCO    arch/arm/boot/dts/overlays/papirus.dtbo
Warning (model_is_string): "model" property in /__overrides__ is not a string
```


arch/arm/boot/dts/overlays/README
```diff
@@ -1001,6 +1001,17 @@ Params: speed                   Display SPI bus speed
         xohms                   Touchpanel sensitivity (X-plate resistance)
 
 
+Name:   papirus
+Info:   PaPiRus ePaper Screen by Pi Supply (both HAT and pHAT)
+Load:   dtoverlay=papirus,model=<val>
```
pelwell: Nitpick: Can you make this say `<param>=<val>` to match the others?

arch/arm/boot/dts/overlays/papirus-overlay.dts
```diff
@@ -0,0 +1,89 @@
...
+				reset-gpios = <&gpio 24 0>;
+				busy-gpios = <&gpio 25 0>;
+
+				repaper-thermal-zone = "display";
+			};
+		};
+	};
+
+	__overrides__ {
+		model = <&repaper>, "compatible";
```
pelwell: Can you find a different name for this parameter - `panel`, perhaps? `model` is a standard property name and dtc warns about the declaration because it isn't just a string (it starts with a symbol reference).

pelwell (commented):
Apart from the two minor points it looks OK, i.e. it compiles, and anyway it's basically a module so how bad could it be?

notro:
I've adressed your comments and now the warning went away.

pelwell:
Thanks - I'm happy now. @popcornmix ?

popcornmix:
Yes, it looks like a useful PR - thanks!

popcornmix:
rpi-update now includes this if anyone wants to try it.

notro:
Thank you, rpi-update worked fine!

PR #2121, opened 2017-07-17
config: enhance DualShock3 controller support

psyke83:
Enable rumble support in Sony HID & HID battery strength.


pelwell:
This adds about 2KB to the static kernel and ~500 bytes to the hid-sony module. No objections from me.

PR #2135, opened 2017-07-27
Bug fix: unmuted

allocom:
In Dual mono mode, on changing volume, left speaker of 2nd DAC was unmuted. Also in 2.0 mode, 2nd DAC speakers have unmuted.


popcornmix:
A description of the problem would be useful.

allocom:
Updated description, Hope its fine.

pelwell:
The code looks plausible by analogy - the description helps - so I have no objection to this commit being merged.

PR #2137, opened 2017-07-29
rtc: add support for ST M41T62

ecc1:


pelwell:
I have no problem with the overlay part of this commit (if you put the overlay parameter in alphabetical order in the README), but I'm not happy with the driver patch. The driver is a generic upstream driver (i.e. not Pi-specific), so the patch will have to go upstream. In the meantime, if you revert the driver patch, remove the wake-up source reference and sort the README I'll merge it.

PR #2138, opened 2017-07-30
overlays: i2c-rtc: add m41t62

ecc1:
Add support for the ST M41T62 real-time clock chip.


popcornmix:
@pelwell are you around? Any objections?

pelwell:
It still has the wake-up source support, but that isn't harmful - just a bit misleading.

PR #2143, opened 2017-08-02
drm/vc4: Backport more of drm-misc-next

anholt:
This brings in the memory debugging support, a little fix for dual display, a dmesg cleanup, and some prep for CEC support.  It looks like 4.9 is just way too old for CEC, and it's getting harder and harder to backport vc4 work at this point.


anholt:
Pushed another backported fix, which should be a bit stability win under memory pressure.

pelwell:
Some quick testing didn't reveal any obvious problems with the patches applied, so I'm happy to merge. @popcornmix ?

Issue #2151, opened 2017-08-07
Raspberry pi3 fails to boot 64bit kernel 4.12.3 with bcm2837 dts


VT-Github-Raspberrypi:
Raspberry pi3 fails to boot 64bit kernel 4.12.3 with bcm2837 dts

VT-Github-Raspberrypi:
Fix for raspberrypi#2151
https://github.com/VT-Github-Raspberrypi/linux/commit/05d903273846b420c7284af19ca0b4a28e86cb7f

VT-Github-Raspberrypi:
Follow these steps for RPI-3 64bit boot up.
https://github.com/raspberrypi/linux/wiki/config.txt

lategoodbye:
In short mixing a downstream defconfig and a upstream devicetree will usually cause problems.

Good combinations:
arm64/configs/bcmrpi3_defconfig +bcm2710-rpi-3-b.dts
arm64/configs/defconfig + bcm2837-rpi-3-b.dts

In the end you are trying to backport this [commit](https://github.com/raspberrypi/linux/commit/738f859f520862e51a5662b03fb162524e09307b), which is available since 4.13.

pelwell:
We are happy to back-port a commit from a newer kernel if it is proven to solve a problem.

@VT-Github-Raspberrypi On a top-of-tree rpi-4.12.y branch, run:
```
git cherry-pick 738f859f
```
Rebuild the DTBs and confirm that they work as expected.

JamesH65:
@VT-Github-Raspberrypi Did you have a chance to test this out?

JamesH65:
Closing due to lack of activity. Reopen if you feel this issue is still relevant.

PR #2153, opened 2017-08-07
Rpi 4.12.y

VT-Github-Raspberrypi:
ARM64: bcm2837: enable sdhost in device tree for the
Raspberry Pi 3

Fix for https://github.com/raspberrypi/linux/issues/2151

Signed-off-by: VT-Github-Raspberrypi <github.raspberrypi@votarytech.com>


VT-Github-Raspberrypi (commented):
It looks good to me.

pelwell:
screenshot
Really?

pelwell:
And when you fix it, please give your commit a proper name - Rpi 4.12.y isn't very descriptive.

VT-Github-Raspberrypi:
Updated commit message as per review comment by pelwell.

VT-Github-Raspberrypi:
Follow these steps for RPI-3 64bit boot up.
https://github.com/raspberrypi/linux/wiki/config.txt

VT-Github-Raspberrypi (commented):
It looks good to me.

VT-Github-Raspberrypi (commented):
It looks good to me.

arch/arm/boot/dts/bcm2835-rpi.dtsi
```diff
@@ -27,6 +27,13 @@
 			firmware = <&firmware>;
 			#power-domain-cells = <1>;
 		};
+
```
pelwell: This is an upstream file - please don't modify it, except by trying to push changes upstream.

Downstream DT has a framebuffer node in bcm2708-rpi.dtsi.

arch/arm64/boot/dts/broadcom/bcm2837-rpi-3-b.dts
```diff
@@ -22,3 +22,14 @@
 &uart1 {
 	status = "okay";
 };
+
```
pelwell: This is also an upstream file. Have you considered using bcm2710-rpi-3-b.dts instead?

pelwell (changes requested):

VT-Github-Raspberrypi:
Now reverted frambuffer changes.

VT-Github-Raspberrypi (commented):
Now seems all fine.

pelwell:
bcm2837-rpi-3-b.dts is also an upstream file. Have you considered using bcm2710-rpi-3-b.dts instead?

VT-Github-Raspberrypi:
Yes. We only used bcm2837-rpi-3-b.dts.

pelwell:
And I'm saying that I won't accept changes to that file. Try using bcm2710-rpi-3-b.dts instead.

VT-Github-Raspberrypi:
Without this changes Pi-3 board doesn't boot up using bcm2837 dts. 
As per your suggestion like ,"use bcm2710 dts" then What is the use for bcm2837 dts ?.

We have been using bcm2837 dts and making changes according to considering SOC(BCM2837)  raspberry pi-3 64bit.

pelwell:
bcm2837-rpi-3-b.dts is the official upstream Linux support for Pi3 (64-bit, in this case). Upstream changes happen slowly, and they don't support overlays etc., but you can get the source from kernel.org.

bcm2710-rpi-3-b.dts is the downstream Pi3 support, although we don't officially support 64-bit ourselves. Downstream changes can be very quick, but there is likely to be more churn than you get with upstream-only builds.

VT-Github-Raspberrypi:
My changes are on upstream file bcm2837 then what is the reason for not accepting my changes ?.

Let me clear that , What should i do for my changes to be get merged on https://github.com/raspberrypi/linux/tree/rpi-4.12.y ?

pelwell:
Changes to upstream files should be submitted as patches via email to the Linux maintainers. There is a guide to the process here: https://www.kernel.org/doc/html/v4.12/process/submitting-patches.html

VT-Github-Raspberrypi:
What should i do for my changes to be get merged on https://github.com/raspberrypi/linux/tree/rpi-4.12.y ? . 
same process ?.

pelwell:
Once it is accepted upstream (which could be to a 4.13 release candidate, but any branch will do) we can cherry-pick it back to rpi-4.12.y.

lategoodbye:
Please look at my comments in #2151 . This change can't be accepted in upstream because it's already there in 4.13.

pelwell:
Closing as the next RPi kernel version will be 4.14.

PR #2178, opened 2017-08-27
Arm fiq stat

oghorbel:
Hello,

Last year I've added stat support to arm fiq and usb driver.
If this is interesting I can rebase it on a new branch.
    
This patch allows drivers that uses fiq to have a stat on the
execution number of the fiq handler.
For that three APIs has been defined:
- fiq_kstat_enable: this function enables fiq stat and allocates required
memory for it
- fiq_kstat_disable: this function disable fiq stat and free its allocated
memory
- fiq_kstat_this_cpu_inc: This function increments the fiq stat counter of
the current CPU running the fiq handler

A driver may call fiq_kstat_enable at its initialization to enable fiq
stat and then call fiq_kstat_this_cpu_inc from the fiq handler

When the fiq stat is enabled by a driver, then /proc/interrupts shows the
fiq entry as the following example:
FIQ:          0   21642080          0          0  usb_fiq
If the fiq stat is not enabled, the content will be similar to the old one
as the following example:
FIQ:                                              usb_fiq
The fiq name will be always written on the first column after the last CPU
  column

Thanks

### DIFF
--- a/arch/arm/include/asm/fiq.h
+++ b/arch/arm/include/asm/fiq.h
@@ -31,6 +31,8 @@ struct fiq_handler {
 	/* data for the relinquish/reacquire functions
 	 */
 	void *dev_id;
+	/* fiq stats percpu */
+	unsigned int __percpu *fiq_kstat;
 };
 
 extern int claim_fiq(struct fiq_handler *f);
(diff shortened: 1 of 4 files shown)


popcornmix:
@P33M I guess this is your area

P33M:
1. This PR modifies upstream files. We don't pull in changes to upstream files as this causes merge conflicts and they haven't been through the maintainer review process. If the changes are accepted upstream, then we will backport if necessary.

2. What benefit does the FIQ stat offer besides an execution counter? The entire purpose of the FIQ on Raspberry Pi is to *hide* the fact that there's thousands of interrupts per second occurring. Touching more discontiguous variables (the fiq percpu pointer/counter) results is more round-trips to SDRAM in a very performance-critical piece of code.

3. The implementation races against kernel-context threads. What happens if the FIQ is trying to do a fiq_kstat_this_cpu_inc while another CPU just disabled fiq_kstat?

oghorbel:
I agree with with you. The patch was developed to help debugging fiq and usb. I close the PR.

PR #2180, opened 2017-08-28
Fix DTB overlay compilation for arm64 broadcom

olivierschonken:
The dts-dirs variable was overwritten by the assignment of the
stingray directory after the overlays directory, thus no overlays
were being built

Signed-off-by: Olivier Schonken <olivier.schonken@gmail.com>


pelwell:
Thanks!

Issue #396, opened 2013-10-10
PL011: no DMA platform data


jansendup:
in drivers/tty/serial/amba-pl011.c uap->port.dev->platform_data is NULL on the bcm2835 arch causing DMA probe to fail. thus no DMA on the uart module.

popcornmix:
Is there an upstream commit I can pull in?

jansendup:
Not quite sure where to look. Maybe you can check.
In arch/arm/mach-bcm2708/bcm2708.c line 242: "AMBA_DEVICE(uart0, "dev:f1", UART0, NULL);"
The last parameter of AMBA_DEVICE should be a pointer to a "struct amba_pl011_data" that is defined in include/linux/amba/serial.h line 201.

popcornmix:
It is not possible to dma from/to the uart, so I think a DMA probe failure is correct.

P33M:
In the BCM2835 peripherals datasheet, UART1 is explicitly defined as having no DMA support. UART0 has a DREQ line entry for tx/rx in the DMA controller section - DREQ 12/14 respectively.

Are there free DMA channels for use by the UART?

jansendup:
I quickly tried setting filter_fn in struct dma_chan *dma_request_channel(dma_cap_mask_t mask,dma_filter_fn filter_fn,void *filter_param); to NULL. If I understand it correctly, dma_request_channel should just then return any available channel. It returned NULL so it might be that there aren't any available channels. dma_request_slave_channel(dev, "tx") also returned NULL.

jansendup:
It seems that there is no dmaengine driver for the BCM2835 so thats probably why the last parameter in arch/arm/mach-bcm2708/bcm2708.c line 242: "AMBA_DEVICE(uart0, "dev:f1", UART0, NULL);" is NULL.

P33M:
It appears you are correct; there are no dma engine drivers for either the bcm2708 port (this repo) or the upstream bcm2835 port. It looks like several people have had a go but not upstreamed it.

It would be necessary to get a platform DMA driver implemented before adding support to the peripheral drivers (i2c/spi/uart).

popcornmix:
@jansendup 
I assume there is nothing to change here - okay to close?

PR #2181, opened 2017-08-30
VCSM dmabuf import

6by9:
Firmware and userland PRs to come.

### DIFF
--- a/drivers/char/broadcom/vc_sm/vc_sm_defs.h
+++ b/drivers/char/broadcom/vc_sm/vc_sm_defs.h
@@ -26,8 +26,9 @@
 /* Resource name maximum size */
 #define VC_SM_RESOURCE_NAME 32
 
-/* All message types supported for HOST->VC direction */
 typedef enum {
+	/* Message types supported for HOST->VC direction */
+
 	/* Allocate shared memory block */
 	VC_SM_MSG_TYPE_ALLOC,
 	/* Lock allocated shared memory block */
@@ -45,6 +46,21 @@ typedef enum {
 
 	/* A previously applied action will need to be reverted */
 	VC_SM_MSG_TYPE_ACTION_CLEAN,
+
+	/*
+	 * Import a physical address and wrap into a MEM_HANDLE_T.
+	 * Release with VC_SM_MSG_TYPE_FREE.
+	 */
+	VC_SM_MSG_TYPE_IMPORT,
+
+	/* Message types supported for VC->HOST direction */
+
+	/*
+	 * VC has finished with an imported memory allocation.
+	 * Release any Linux reference counts on the underlying block.
+	 */
+	VC_SM_MSG_TYPE_RELEASED,
+
 	VC_SM_MSG_TYPE_MAX
 } VC_SM_MSG_TYPE;
 
@@ -165,6 +181,41 @@ typedef struct {
 
 } VC_SM_FREE_ALL_T;
 
+/* Request to import memory (HOST->VC) */
+struct vc_sm_import {
+	/* type of memory to allocate */
+	VC_SM_ALLOC_TYPE_T type;
+	/* pointer to the VC (ie physical) address of the allocated memory */
+	uint32_t addr;
+	/* size of buffer */
+	uint32_t size;
+	/* opaque handle returned in RELEASED messages */
+	int32_t  kernel_id;
+	/* Allocator identifier */
+	uint32_t allocator;
+	/* resource name (for easier tracking on vc side) */
+	char     name[VC_SM_RESOURCE_NAME];
+};
+
+/* Result of a requested memory import (VC->HOST) */
+struct vc_sm_import_result {
+	/* Transaction identifier */
+	uint32_t trans_id;
+
+	/* Resource handle */
+	uint32_t res_handle;
+};
+
+/* Notification that VC has finished with an allocation (VC->HOST) */
+struct vc_sm_released {
+	/* pointer to the VC (ie physical) address of the allocated memory */
+	uint32_t addr;
+	/* size of buffer */
+	uint32_t size;
+	/* opaque handle returned in RELEASED messages */
+	int32_t  kernel_id;
+};
+
 /* Union of ALL messages */
 typedef union {
 	VC_SM_ALLOC_T alloc;
@@ -175,7 +226,9 @@ typedef union {
 	VC_SM_LOCK_RESULT_T lock_result;
 	VC_SM_RESULT_T result;
 	VC_SM_FREE_ALL_T free_all;
-
+	struct vc_sm_import import;
+	struct vc_sm_import_result import_result;
+	struct vc_sm_released released;
 } VC_SM_MSG_UNION_T;
 
 #endif /* __VC_SM_DEFS_H__INCLUDED__ */
--- a/drivers/char/broadcom/vc_sm/vc_sm_defs.h
+++ b/drivers/char/broadcom/vc_sm/vc_sm_defs.h
@@ -151,8 +151,10 @@ struct vc_sm_lock_result_t {
 	uint32_t res_handle;
 	/* Pointer to resource buffer */
 	uint32_t res_mem;
-	/* Pointer to former resource buffer if the memory
-	 * was reallocated */
+	/*
+	 * Pointer to former resource buffer if the memory
+	 * was reallocated
+	 */
 	uint32_t res_old_mem;
 
 };
--- a/drivers/char/broadcom/vc_sm/vc_sm_defs.h
+++ b/drivers/char/broadcom/vc_sm/vc_sm_defs.h
@@ -88,7 +88,7 @@ typedef struct {
 	/* Resource handle */
 	uint32_t res_handle;
 	/* Pointer to resource buffer */
-	void *res_mem;
+	uint32_t res_mem;
 	/* Resource base size (bytes) */
 	uint32_t res_base_size;
 	/* Resource number */
@@ -101,7 +101,7 @@ typedef struct {
 	/* Resource handle (returned from alloc) */
 	uint32_t res_handle;
 	/* Resource buffer (returned from alloc) */
-	void *res_mem;
+	uint32_t res_mem;
 
 } VC_SM_FREE_T;
 
@@ -110,7 +110,7 @@ typedef struct {
 	/* Resource handle (returned from alloc) */
 	uint32_t res_handle;
 	/* Resource buffer (returned from alloc) */
-	void *res_mem;
+	uint32_t res_mem;
 
 } VC_SM_LOCK_UNLOCK_T;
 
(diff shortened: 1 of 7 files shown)


popcornmix:
No objections here. A quick test builds and runs existing code that uses vcsm.
@pelwell?

drivers/char/broadcom/vc_sm/vmcs_sm.c
```diff
@@ -1749,8 +1751,8 @@ static int vc_sm_ioctl_lock(struct SM_PRIV_DATA_T *private,
 			private->restart_sys = -EINTR;
 			private->int_action = VC_SM_MSG_TYPE_LOCK;
 			goto error;
-		} else if (status != 0 ||
-			   (status == 0 && result.res_mem == NULL)) {
+		} else if (status ||
+			   (!status && !(void *)result.res_mem)) {
```
pelwell, line 1781: Isn't this cast pointless?

drivers/char/broadcom/vc_sm/vc_vchi_sm.h
```diff
@@ -19,63 +19,73 @@
...
+ * Initialize the shared memory service, opens up vchi connection to talk to it.
+ */
 VC_VCHI_SM_HANDLE_T vc_vchi_sm_init(VCHI_INSTANCE_T vchi_instance,
 				    VCHI_CONNECTION_T **vchi_connections,
 				    uint32_t num_connections);
 
-/* Terminates the shared memory service.
-*/
+/*
+ *Terminates the shared memory service.
```
pelwell: Missing space.

drivers/char/broadcom/vc_sm/vmcs_sm.c
```diff
@@ -3104,15 +3096,24 @@ static void vc_sm_connected_init(void)
...
 /* Driver loading. */
-static int __init vc_sm_init(void)
+static int bcm2835_vcsm_probe(struct platform_device *pdev)
 {
 	pr_info("vc-sm: Videocore shared memory driver\n");
+
+	sm_state = kzalloc(sizeof(*sm_state), GFP_KERNEL);
+	if (!sm_state)
+		return -ENOMEM;
+	pr_err("vc-sm: pdev is %p. dev is %p\n", pdev, &pdev->dev);
```
pelwell: pr_info at most.

drivers/char/broadcom/vc_sm/vc_sm_defs.h
```diff
@@ -165,6 +181,41 @@ typedef struct {
 
 } VC_SM_FREE_ALL_T;
 
+/* Request to import memory (HOST->VC) */
+struct VC_SM_IMPORT {
```
pelwell: Linux doesn't use CAPS for structure names.

drivers/char/broadcom/vc_sm/vc_vchi_sm.h
```diff
@@ -89,4 +89,11 @@ int vc_vchi_sm_walk_alloc(VC_VCHI_SM_HANDLE_T handle);
 int vc_vchi_sm_clean_up(VC_VCHI_SM_HANDLE_T handle,
 			VC_SM_ACTION_CLEAN_T *action_clean);
 
+/*
+ * Import a contiguous block of memory and wrap into in a GPU MEM_HANDLE_T.
```
pelwell: "it in a"?

drivers/char/broadcom/vc_sm/vmcs_sm.c
```diff
@@ -500,7 +510,9 @@ static void vmcs_sm_remove_map(struct SM_STATE_T *state,
 static int vc_sm_global_state_show(struct seq_file *s, void *v)
 {
 	struct sm_mmap *map = NULL;
+	struct SM_RESOURCE_T *resource = NULL;
```
pelwell: See CAPS comment.

drivers/char/broadcom/vc_sm/vmcs_sm.c
```diff
@@ -2056,6 +2110,138 @@ static int vc_sm_ioctl_unlock(struct SM_PRIV_DATA_T *private,
...
+	}
+
+	import.type = ((ioparam->cached == VMCS_SM_CACHE_VC) ||
+		       (ioparam->cached == VMCS_SM_CACHE_BOTH)) ?
+				VC_SM_ALLOC_CACHED : VC_SM_ALLOC_NON_CACHED;
+	import.addr = (uint32_t)sg_dma_address(sgt->sgl);
+	import.size = sg_dma_len(sgt->sgl);
+	import.allocator = current->tgid;
+
+	if (*ioparam->name) {
```
pelwell: Lose the braces.

drivers/char/broadcom/vc_sm/vmcs_sm.c
```diff
@@ -2056,6 +2110,138 @@ static int vc_sm_ioctl_unlock(struct SM_PRIV_DATA_T *private,
...
+	status = vc_vchi_sm_import(sm_state->sm_handle, &import, &result,
+				   &private->int_trans_id);
+	if (status == -EINTR) {
+		pr_debug("[%s]: requesting import memory action restart (trans_id: %u)\n",
+			 __func__, private->int_trans_id);
+		ret = -ERESTARTSYS;
+		private->restart_sys = -EINTR;
+		private->int_action = VC_SM_MSG_TYPE_IMPORT;
+		goto error;
+	} else if (status != 0 || (status == 0 && result.res_handle == 0)) {
```
pelwell: `status || !result.res_handle`

drivers/char/broadcom/vc_sm/vc_sm_defs.h
6by9: I retained caps for consistency with the rest of the file. This file is also shared with the firmware.
So either we go through all the code on both sides and adopt the Linux scheme (and ignore the firmware scheme), or we retain the current style.

Then again I didn't retain full consistency as I dropped the typedefs in favour of named structs.

What's the preference here? A clean up set of patches to go the Linux way?

drivers/char/broadcom/vc_sm/vc_vchi_sm.h
6by9: Done

drivers/char/broadcom/vc_sm/vc_vchi_sm.h
6by9: Done

drivers/char/broadcom/vc_sm/vmcs_sm.c
6by9: Done

drivers/char/broadcom/vc_sm/vmcs_sm.c
6by9: Checkpatch disagrees due to braces on the else clause.
CHECK: braces {} should be used on all arms of this statement
#396: FILE: drivers/char/broadcom/vc_sm/vmcs_sm.c:2164:
+	if (*ioparam->name)
[...]
+	else {
[...]

drivers/char/broadcom/vc_sm/vmcs_sm.c
6by9: Removed. Left over debugging.

drivers/char/broadcom/vc_sm/vc_sm_defs.h
pelwell: I would choose to not introduce new rule violations rather than maintain consistency.

drivers/char/broadcom/vc_sm/vmcs_sm.c
pelwell: But the else clause doesn't need them either.

drivers/char/broadcom/vc_sm/vmcs_sm.c
6by9: Yes. Done.
Too many passes through checkpatch.
res_mem being converted to u32 meant that comparison of u32 to NULL (ptr) became a warning. Then it prefers !(ptr) to ptr == NULL hence why we ended up here.

Just looking through and there are a bundle of ```if (X || (!X && Y)```, which either have redundant casts but can also be reduced to ```if (X || Y)```. I'll nuke those too.

drivers/char/broadcom/vc_sm/vmcs_sm.c
6by9: Ack. Think I'd looked at a different instance of almost identical syntax.

drivers/char/broadcom/vc_sm/vc_sm_defs.h
6by9: OK, will do.
I'll do a follow up clean up set of patches to try and make vc-sm as a whole friendlier to mainline.

6by9:
Original 4 commits amended as requested.

16 followups fixing various things that checkpatch complains about. It still leaves a load of line > 80 chars / too many tab indents in vmcs_sm.c, but those can wait.

6by9:
Please don't merge yet - I've still to fully test all of these! It compiles, but hasn't been run, and I've got to pull the IPC header over and make the firmware mods still.

pelwell:
That was close. ;-)

6by9:
Oops, sorry.
Dmabuf import needs the firmware merge request to be able to do anything useful anyway.

6by9:
OK, I'm happy. 
Branch rebased and 6 extra patches added, but they're just switching structure names from all caps to lower case to follow the kernel coding style.
checkpatch with --strict still has 55 warnings and 228 checks, but those really will wait.

One thing that may want addressing is the definition of enum vc_sm_alloc_type_t
```
enum vc_sm_alloc_type_t {
	VC_SM_ALLOC_CACHED,
	VC_SM_ALLOC_NON_CACHED,
};
```
This is used in the IPC message structures, but doesn't guarantee the size of the enum. It's not shown up in practice, but I think that is only because the following fields need 4 byte alignment. Do you agree that needs fixing, or not care?

@popcornmix - do you want to rerun your tests before this is merged? Mine have all been fine, and inspection shows no change in behaviour beyond the new feature.

popcornmix:
@6by9 might be worth adding a dummy enum entry
```VC_SM_ALLOC_FORCE_ENUM_32BIT = 0x80000000```
to make sure a smaller enum type is not used.
(or change to uint32_t).

6by9:
@popcornmix That was my thought too. I was a little surprised that it wasn't there already.

I suspect I'll do another round of vcsm cleanups some time soonish (1-2 weeks) as it will be required upstream if they're wanting to use MMAL efficiently. At least if we've done the first round of cleanups then they might whinge less about it. I'll roll the enum update into the next set of cleanups as it isn't a blocking issue here due to where vc_sm_alloc_type_t is placed in the structures.

PR #2204, opened 2017-09-20
[4.14] enable upstream infrared remote transmitter drivers

HiassofT:
Upstream kernel 4.14 added 2 drivers for supporting infrared remote transmitters (IR LED connected to a GPIO pin).

gpio-ir-tx is a bit-banging implementation, identically to what lirc-rpi is doing, pwm-ir-tx is a more efficient implementation that makes use of PWM0 to generate the IR bursts.

With these 2 drivers added all of the functionality of the downstream lirc-rpi driver is now covered by upstream drivers so you can consider retiring the downstream driver at some point in the future (the old lirc-only drivers from the staging directory are being phased out in upstream kernel and all except lirc_zilog have been converted to rc-core drivers).


popcornmix:
Looks fine. @pelwell can comment on the DT overlays.
So what is the effect on users if we drop the downstream lirc driver?
They need to edit their dtoverlay lines in config.txt I assume?
Any other differences? Any software you know of that will no longer work the same?

pelwell:
The overlays are fine. I was thinking that, provided these drivers work as expected, the lirc_rpi overlay can be repurposed to use the gpio-ir-tx driver instead prior to eventual retirement.

HiassofT:
An important difference is that lirc-rpi uses a single /dev/lirc0 device that supports both send and transmit functionality. With gpio-ir (reciever) plus gpio-ir-tx/pwm-ir-tx (transmitter) enabled we have 2 separate /dev/lircX devices.

This will affect people who used both receive and transmit functionality of lirc-rpi. Probably only very few people did this, but the new drivers aren't a drop-in replacement for that case.

Another thing to consider is GPIO assignment. Unfortunately lirc-rpi used GPIO18 for input and 17 for output - but PWM output is only available on GPIO18 so existing hardware using these GPIOs can only use the bit-banging gpio-ir-tx driver (which then needs to be configured to GPIO17 instead of default 18). So pwm-ir-tx is mainly useful for newer hardware or people that hook up the IR LED via jumper wires.

For receive-only (probalby the one used mostly) and transmit-only setups it's a lot easier. The upstream drivers can be used as drop-in replacements, and the gpio-ir receiver driver also supports in-kernel decoding - so people can use ir-keytable to configure their remotes instead of running userspace lircd.

As a side note: an easy way to use the IR transmitter functionality is via the "ir-ctl" program from current v4l-utils (the versions in Raspbian are probably too old and I'm not sure ir-ctl is available at all) - you can specify the remote protocol and scancode on the command line. eg `ir-ctl -S rc6_mce:0x800f0420` to send a KEY_RIGHT (scancodes are the same as in the rc_keymaps).

popcornmix:
Switching to upstream drivers is clearly desirable so we'll merge this and that allows users to manually switch.

Getting people to stop using the old driver is trickier. If you can make the old lirc_rpi overlay use the new driver (in non-pwm mode) in a transparent way that may be a useful step to removing the old driver. In addition we can encourage users to update config.txt to use the new overlay.

I suspect almost all lirc users are running in receive only mode as a remote to media player.
There will be a much smaller number running in transmit only mode and a tiny number running in both.
If we (eventually) force users to move to new driver then hopefully the fake lirc_rpi overlay will allow the majority to keep running without issue.

HiassofT:
When planning for a transition keep in mind that having two lirc devices (/dev/lirc0 and /dev/lirc1) will break most users' setups.

It's not really guaranteed which one of these will be the receive and which will be the transmit device and standard lircd setup (which also the majority of users will be using) just uses /dev/lirc0.

So people may need to change their configs, add udev rules to assign static device nodes, maybe run a separate lircd instance to transmit etc - quite a PITA.

A hard switch to the upstream drivers or having lirc-rpi available as well (as long as it's easy to keep it working) for a transition period may be better options.

In LibreELEC 9.0 (which will be released probably early 2018) we'll be using gpio-ir exclusively for GPIO IR receivers, current milhouse testbuilds ship without lirc-rpi for a while now. So far it has been rather easy to tell people to change "lirc-rpi" to "gpio-ir" in their config.txt, but the big rush is still to come. Our plan is to officially announce the switch after the 8.2.0 release (which should hopefully happen really soon now) so we and users have some time to prepare for it (gpio-ir already works fine in 8.x builds).

PR #2206, opened 2017-09-21
fix vcsm for multiplatform

invisiblek:


invisiblek:
the commit i pushed yesterday didn't actually address the multi platform problem, it simply fixed a compiler error

popcornmix:
This seems to stop ``bcm2709_defconfig`` from building:
```
arch/arm/mm/built-in.o:(___ksymtab+dmac_clean_range+0x0): undefined reference to `dmac_clean_range'
arch/arm/mm/built-in.o:(___ksymtab+dmac_inv_range+0x0): undefined reference to `dmac_inv_range'
drivers/built-in.o: In function `.LANCHOR1':
arm_pmu.c:(.rodata+0x93c4): undefined reference to `dmac_inv_range'
arm_pmu.c:(.rodata+0x93cc): undefined reference to `dmac_clean_range'
Makefile:972: recipe for target 'vmlinux' failed
```

invisiblek:
Oh, I think I see what's up. I'll have something up shortly. I was only building multi-platform and overlooked non.

invisiblek:
Ok, should be good now. Squashed that fix in to the first commit (as it really does belong there). The diff: https://gist.github.com/invisiblek/31df3a271854f38b04021b51f33e9d64

popcornmix:
This looks okay to me and builds with bcm2709_defconfig and bcmrpi_defconfig.
@pelwell okay with you?

pelwell:
Considering that we're changing upstream code it's as clean as it can be. I can live with this.

PR #2222, opened 2017-10-11
AXI performance monitor driver

JamesH65:
Uses the debugfs I/F to provide access to the AXI
bus performance monitors.

Requires the new mailbox peripheral access for access
to the VPU performance registers, system bus access
is done using direct register reads.


drivers/perf/raspberrypi_axi_monitor.c
```diff
@@ -0,0 +1,633 @@
...
+static struct platform_driver rpi_axiperf_driver  = {
+	.probe =	rpi_axiperf_probe,
+	.remove =	rpi_axiperf_remove,
+	.driver = {
+		.name   = "rpi-axiperf",
+		.of_match_table = of_match_ptr(rpi_axiperf__match),
+	},
+};
+
+static int __init rpi_axiperf_init(void)
```
pelwell: Replace from this line...

drivers/perf/raspberrypi_axi_monitor.c
```diff
@@ -0,0 +1,633 @@
...
+	return platform_driver_register(&rpi_axiperf_driver);
+}
+
+static void __exit rpi_axiperf_exit(void)
+{
+	platform_driver_unregister(&rpi_axiperf_driver);
+}
+
+module_init(rpi_axiperf_init);
+module_exit(rpi_axiperf_exit);
```
pelwell: ...to this one, inclusive, with:
```
module_platform_driver(rpi_axiperf_driver);
```

drivers/perf/raspberrypi_axi_monitor.c
```diff
@@ -0,0 +1,633 @@
...
+
+	struct rpi_firmware *firmware;
+
+	// Sample time spent on for each bus
+	int sample_time;
+
+	// Now storage for the per monitor settings and the resulting
+	// performance figures
+	struct
+	{
```
pelwell: Don't forget to stick to the ugly kernel brace placement rules.

drivers/perf/raspberrypi_axi_monitor.c
```diff
@@ -0,0 +1,633 @@
...
+
+static struct rpi_axiperf *state = NULL;
+
+// Two monitors, System and VPU, each with the following register sets
+// Each monitor can only monitor one bus at a time, so we time share them, giving
+// each bus 100ms of time on its associated monitor
+// Record results from the three Bus watchers per monitor and push to the sysfs
+
+// General reg
+const int GEN_CTRL = 0x00;
```
pelwell: The kernel devs prefer #defines to const ints (go figure), but if you choose to stick with consts you'll have to change the symbols to lower case.

drivers/perf/raspberrypi_axi_monitor.c
```diff
@@ -0,0 +1,633 @@
...
+			dev_err(&state->dev->dev, "Failed to read bus watcher");
+		else
+			memcpy(results, &tmp[2], NUM_BUS_WATCHER_RESULTS *sizeof(u32));
+	}
+	else
+	{
+		int i;
+		void __iomem *addr = state->monitor[monitor].base_address
+				+ watcher + BW_ATRANS_OFFSET;
+		for (i=0;i<NUM_BUS_WATCHER_RESULTS ;i++, addr+=4)
```
pelwell: Spaces after semicolons here, none before.

drivers/perf/raspberrypi_axi_monitor.c
```diff
@@ -0,0 +1,633 @@
...
+			// No, disable all monitoring for this monitor
+			set_monitor_control(monitor, GEN_CTL_RESET_BIT);
+		} else {
+		int i;
+
+		/* Find out how many busses we want to monitor, and
+		 * spread our 3 actual monitors over them
+		 */
+		num_buses[monitor] = hweight32(state->monitor[monitor].bus_enabled);
+		num_buses[monitor] = min(num_buses[monitor],
```
pelwell: The double assignment gave me a momentary WTF... It looks like the line is already over-long, so why not split it like this instead?
```
		num_buses[monitor] = min(
			hweight32(state->monitor[monitor].bus_enabled),
			NUM_BUS_WATCHERS_PER_MONITOR);
```

drivers/perf/raspberrypi_axi_monitor.c
```diff
@@ -0,0 +1,633 @@
...
+
+	mutex_lock(&state->lock);
+
+	for (monitor = 0; monitor < NUM_MONITORS; monitor++) {
+		// Anything enabled?
+		if (state->monitor[monitor].bus_enabled == 0) {
+			// No, disable all monitoring for this monitor
+			set_monitor_control(monitor, GEN_CTL_RESET_BIT);
+		} else {
+		int i;
```
pelwell: The indentation for this section is wrong. You've got enough logic in here that you might want to split out a helper function.

drivers/perf/raspberrypi_axi_monitor.c
```diff
@@ -0,0 +1,633 @@
...
+
+static void monitor(struct rpi_axiperf *state)
+{
+	int monitor, num_buses[NUM_MONITORS];
+
+	mutex_lock(&state->lock);
+
+	for (monitor = 0; monitor < NUM_MONITORS; monitor++) {
+		// Anything enabled?
+		if (state->monitor[monitor].bus_enabled == 0) {
```
pelwell: This loop is crying out for a local pointer to replace "state->monitor[monitor]".

drivers/perf/raspberrypi_axi_monitor.c
```diff
@@ -0,0 +1,633 @@
...
+		num_buses[monitor] = hweight32(state->monitor[monitor].bus_enabled);
+		num_buses[monitor] = min(num_buses[monitor],
+					 NUM_BUS_WATCHERS_PER_MONITOR);
+
+		for (i = 0; i < num_buses[monitor]; i++) {
+			int bus_control;
+			do {
+				state->monitor[monitor].last_monitored++;
+				state->monitor[monitor].last_monitored &= 0xf; //wrap around
+			} while ((state->monitor[monitor].bus_enabled &
```
pelwell: Kernel devs prefer "!x" to "x == 0".

drivers/perf/raspberrypi_axi_monitor.c
pelwell: But see previous comment about a local pointer to `state->monitor[monitor]`.

pelwell (dismissed):
Device tree is supposed to describe the hardware, not specify a driver. I don't see any reason not call this bcm2835-axiperf, even though we wrote it.

I've hit my comments limit. Try running checkpatch and I'll have another go on the result.

JamesH65:
I've pushed a much cleaned up version. Please be gentle.

JamesH65:
@pelwell @6by9 Should be mostly checkpatch compliant....couple of long lines I don't know what to do with - splitting would make it look fairly horrible.

JamesH65:
@popcornmix @pelwell @6by9 I have updated to latest tree so this can now be merged if it is regarded as OK. COuld probably still do with some good 'ole revoooing.

drivers/perf/raspberrypi_axi_monitor.c
```diff
@@ -0,0 +1,635 @@
...
+
+	kfree(string_buffer);
+
+	return cnt;
+}
+
+static ssize_t mywriter(struct file *fp, const char __user *user_buffer,
+			size_t count, loff_t *position)
+{
+	// TODO: this really needs to zero the registers, not this stuff.
```
pelwell: Are you going to implement this?

drivers/perf/raspberrypi_axi_monitor.c
```diff
@@ -0,0 +1,635 @@
...
+
+	}  monitor[NUM_MONITORS];
+
+};
+
+static struct rpi_axiperf *state;
+
+/* Two monitors, System and VPU, each with the following register sets
+ * Each monitor can only monitor one bus at a time, so we time share them,
+ * giving each bus 100ms (default, settable via debufs) of time on its
```
6by9: s/debufs/debugfs

drivers/perf/raspberrypi_axi_monitor.c
```diff
@@ -0,0 +1,635 @@
...
+ * Copyright (C) 2017 Raspberry Pi Trading Ltd.
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License version 2 as
+ * published by the Free Software Foundation.
+ */
+
+#include <linux/device.h>
+#include <linux/debugfs.h>
+#include <linux/netdevice.h>
```
6by9: Upstream would whinge about alphabetical order.

drivers/perf/raspberrypi_axi_monitor.c
```diff
@@ -0,0 +1,635 @@
...
+struct rpi_axiperf {
+	struct platform_device *dev;
+	struct dentry *root_folder;
+
+	struct task_struct *monitor_thread;
+	struct mutex lock;
+
+	struct rpi_firmware *firmware;
+
+	// Sample time spent on for each bus
```
6by9: Phil: Are we being fussy about /* */ commenting over C99 // ?

drivers/perf/raspberrypi_axi_monitor.c
```diff
@@ -0,0 +1,635 @@
+/*
+ * raspberrypi_axi_monitor.c
+ *
+ *  Created on: 26 Sep 2017
+ *      Author: jamesh
```
6by9: Full name?

drivers/perf/raspberrypi_axi_monitor.c
```diff
@@ -0,0 +1,635 @@
...
+		struct dentry *debugfs_entry;
+		void __iomem *base_address;
+
+	}  monitor[NUM_MONITORS];
+
+};
+
+static struct rpi_axiperf *state;
+
+/* Two monitors, System and VPU, each with the following register sets
```
6by9: Fullstop on the end of the line.
Upstream typically want:
```
/*
 * Comments here
 */
```
for multi line comments.

drivers/perf/raspberrypi_axi_monitor.c
```diff
@@ -0,0 +1,635 @@
...
+ * Each monitor can only monitor one bus at a time, so we time share them,
+ * giving each bus 100ms (default, settable via debufs) of time on its
+ * associated monitor
+ * Record results from the three Bus watchers per monitor and push to the sysfs
+ */
+
+// general registers
+const int GEN_CTRL;
+
+const int GEN_CTL_ENABLE_BIT	= 0x01;
```
6by9: Use the ```BIT(x)``` macros throughout?
```GENMASK(<top_bit>, <bottom_bit>)``` for masks.

File: drivers/perf/raspberrypi_axi_monitor.c [hunk 17]
```diff
@@ -0,0 +1,635 @@
...
+const int BW_CTRL_ENABLE_ID_FILTER_BIT	= 1 << 29;
+const int BW_CTRL_LIMIT_HALT_BIT	= 1 << 28;
+
+const int BW_CTRL_SOURCE_SHIFT	= 8;
+const int BW_CTRL_SOURCE_MASK	= 0x1f << 8; // 5 bits
+const int BW_CTRL_BUS_WATCH_SHIFT;
+const int BW_CTRL_BUS_WATCH_MASK	= 0x3d << 0; // 6 bits
+const int BW_CTRL_BUS_FILTER_SHIFT = 8;
+
+static char *bus_filter_strings[] = {
```
6by9: const for all these arrays and sizes?

drivers/perf/raspberrypi_axi_monitor.c
```diff
@@ -0,0 +1,635 @@
...
+{
+	struct rpi_axiperf *state  = data;
+
+	while (1) {
+		monitor(state);
+
+		if (kthread_should_stop())
+			return 0;
+
+		msleep(100);
```
6by9: Comments said default was 100ms settable in debugfs, or is this a different delay?

drivers/perf/raspberrypi_axi_monitor.c
```diff
@@ -0,0 +1,635 @@
...
+				(void __iomem *)0x7ee08000;
+		else
+			state->monitor[i].base_address =
+				devm_ioremap_resource(&pdev->dev, resource);
+
+		if (IS_ERR(state->monitor[i].base_address))
+			return PTR_ERR(state->monitor[i].base_address);
+
+		// Enable all buses by default
+		state->monitor[i].bus_enabled = 0xffff;
```
6by9: All enabled by default, and driver enabled in DT. Is that the desired default? Does it have any real overhead?

drivers/perf/raspberrypi_axi_monitor.c
```diff
@@ -0,0 +1,635 @@
...
+	},
+	{},
+};
+
+
+static struct platform_driver rpi_axiperf_driver  = {
+	.probe =	rpi_axiperf_probe,
+	.remove =	rpi_axiperf_remove,
+	.driver = {
+		.name   = "bcm2835-axiperf",
```
6by9: Same discussion as with the GPIO expander driver. This isn't generic for the BCM283x chips as it relies on the firmware.
 "raspberrypi,axiperf" instead? (It is meant to be "vendor,driver", and you had no vendor)

pelwell: We haven't always followed the principle,  but device tree is supposed to describe the hardware, not the driver. The performance monitors are Broadcom parts, even if the driver is pi-specific.

JamesH65: So, change the function names to bcm-blah-blah instead of rpi-blah-blah as well as the driver name?

pelwell: No, I'm agreeing with you. I think the DT compatible string should remain "bcm2835,axiperf", and I'm happy with the downstream "rpi" function names. I'd go further and change the driver name to include "rpi" or "raspberrypi".

drivers/perf/raspberrypi_axi_monitor.c
pelwell: I'm not fussed provided checkpatch is happy.

JamesH65: I'll change them all anyway.

drivers/perf/raspberrypi_axi_monitor.c
JamesH65: OOOh, bug, that shouldn't be there I don't think.

drivers/perf/raspberrypi_axi_monitor.c
JamesH65: Intention is for this module only to be built in if specifically requested for debugging purposes, overhead in the not built case is clearly 0, but there is overhead when it is installed and running.

drivers/perf/raspberrypi_axi_monitor.c
JamesH65: Will remove the comment - its not really necessary. I'd like to leave the function so it can be used at a later date although I suspect upstream wouldn't like that.

drivers/perf/raspberrypi_axi_monitor.c
6by9: I'll defer to Phil on this, but the usage is a hybrid where half of it is talking to the firmware (compatible string "raspberrypi,bcm2835-firmware") and not directly to the hardware.

drivers/perf/raspberrypi_axi_monitor.c
6by9: In which case shouldn't the default status in DT be "disabled"? It's currently "okay" in bcm270x.dtsi.

(Or have it in the build but disabled in DT by default? Still zero overhead for most, but easier to enable when desired).

drivers/perf/raspberrypi_axi_monitor.c
```diff
@@ -0,0 +1,635 @@
...
+		return -ENOENT;
+	}
+
+	state->firmware = rpi_firmware_get(fw_node);
+	if (!state->firmware)
+		return -EPROBE_DEFER;
+
+	// Special case for the VPU monitor, we must use the mailbox interface
+	// as it is not accessible from the ARM address space.
+	state->monitor[VPU_MONITOR].use_mailbox_interface = 1;
```
6by9: Sorry, an extra one.
This is a nasty dependency between driver and DT that the addresses for the VPU_MONITOR must be second in the DT list of regs. Is there a better way of reflecting that?

You've also hard coded the address as 0x7ee08000 at line 551, so the DT value doesn't actually matter.

JamesH65: It's hardcoded, because the values I was getting back were not the ones I had in the DT entry, and I never figured out why. Probably should look at that but not got too much time for polishing.

pelwell: There are precedents for defining an order for items under `reg`. The alternative is to have a `reg-names` property containing an array of names, and use platform_get_resource_byname().

The reason the address has changed is that the bus address has been converted into an ARM address. You can retrieve the original value using `of_get_property_read_u32_index()` or `of_get_property_read_u32_array()`.

JamesH65:
Ah ha, as Alan Partridge would say. Will fix that up.

JamesH65:
@pelwell @6by9 Have just pushed new commit which hopefully covers the points above.

drivers/perf/raspberrypi_axi_monitor.c
```diff
@@ -0,0 +1,637 @@
+/*
+ * raspberrypi_axi_monitor.c
+ *
+ *  Created on: 26 Sep 2017
```
6by9: Odd line to have in the header.

drivers/perf/raspberrypi_axi_monitor.c
6by9: Am I not looking at the latest, or are we ignoring them (somehow checkpatch is letting that through).
Ah, I have ```--strict --codespell``` set as arguments to checkpatch by default (although that is still allowing C99 commenting!)

arch/arm/boot/dts/bcm270x.dtsi
```diff
@@ -130,6 +130,14 @@
 			/* Add alias */
 			status = "disabled";
 		};
+
+		axiperf: axiperf {
+			compatible = "brcm,bcm2835-axiperf";
+			reg = <0x7e009800 0x100>,
+			      <0x7ee08000 0x100>;
+			firmware = <&firmware>;
+			status = "okay";
```
6by9: @pelwell Are you happy for this to be enabled by default even though the driver won't be built?

6by9 (commented):
Generally happy.

drivers/perf/raspberrypi_axi_monitor.c
JamesH65: Ah, added by Eclipse when I created the file.  Can remove.

arch/arm/boot/dts/bcm270x.dtsi
pelwell: I think I'd rather have it disabled at this level then enabled higher up, e.g. by bcm2708-rpi.dtsi (even though the two are equivalent).

In general it seems to be frowned on to introduce nodes with an explicit status of "okay", since that is the default.

JamesH65: I have no idea what to put here, I copied that section from another section, so do not know what the status parameter actually means.

drivers/perf/raspberrypi_axi_monitor.c
```diff
@@ -0,0 +1,642 @@
...
+	fw_node = of_parse_phandle(np, "firmware", 0);
+	if (!fw_node) {
+		dev_err(dev, "Missing firmware node\n");
+		return -ENOENT;
+	}
+
+	state->firmware = rpi_firmware_get(fw_node);
+	if (!state->firmware)
+	{
+		dev_err(dev, "Probe deferred\n");
```
6by9: dev_err is a touch high for a probe deferral.

drivers/perf/raspberrypi_axi_monitor.c
```diff
@@ -0,0 +1,642 @@
...
+	"DMA_UC",
+	"SDRAM",
+	"L2_IN"
+};
+
+const int num_vpu_buses = ARRAY_SIZE(vpu_bus_string);
+
+const static char *monitor_name[] = {
+	"System Monitor",
+	"VPU Monitor"
```
6by9: These are also the directory names and they have spaces in :-(
We know these are going to be monitoring things, so is the word "monitor" really required in either case?

6by9 (commented):
Really nit-picking.

drivers/perf/raspberrypi_axi_monitor.c
JamesH65: Bugger, put those in this morning for debugging, forgot to take them out. Inclined to remove both.

drivers/perf/raspberrypi_axi_monitor.c
JamesH65: Sorry, see previous comment.

drivers/perf/raspberrypi_axi_monitor.c
JamesH65: Oop, wrong comment. I'll remove spaces, been meaning to for ages anyway...

6by9 (approved):
I'm happy (I'm assuming it still works!)

JamesH65:
I don't believe working is that important, as long as it adheres to the coding standard.

JamesH65:
Can this be merged now?

pelwell:
Can you rebase one last time (promise)?

JamesH65:
@pelwell Done.

PR #2224, opened 2017-10-12
Bug fix: initial max volume level & default mode value.

allocom:
Initial max volume level & default mode value.


pelwell:
Modifying the mixer control's platform_max looks ugly - please explain why you are no longer using  snd_soc_limit_volume (which looks designed for the job).

allocom:
Corrected it. Pls check

pelwell:
That's better. Can you give me a Signed-off-by line (a comment here will do) and I'll squash and merge.

PR #2227, opened 2017-10-13
Upstream Kernel

limjh16:
I have tested this upstream to 4.9.56 and so far no problems have surfaced.


pelwell:
As you may have noticed, these merges usually occur within a day or two of the upstream release - if they don't, then that is for a reason. I would expect us to go to 4.9.56 in the next few days, moving swiftly over 4.9.55...

popcornmix:
linux tree is now 4.9.56 with rpi-update kernel matching it.

PR #2233, opened 2017-10-18
vc-cma: Remove the driver as non-functional on Pi

6by9:
vc-cma was pulled in from previous platforms. It
supposedly allowed the VPU relocatable heap to be
released to the kernel as CMA, but it was always
a little fragile.

Remove it to avoid complications.


6by9:
Reference #2114 where vc-cma was the cause of most of the reloc heap to be marked as offline, resulting in the camera failing.

pelwell:
No objections from me.

Issue #2235, opened 2017-10-20
RPi3: kernel temperature reading ~5°C too low in 4.12-4.14


HiassofT:
I've noticed that the SoC temperature reported by the kernel is about 5°C lower than what vcgencmd shows. It looks like this only happens on RPi3 in kernel versions 4.12-4.14. 4.9 and 4.11 seem to be fine on RPi3.

Test on RPi3 with latest 4.14 kernel on Raspbian (compiled with bcm2709_defconfig):
```
# dmesg | grep Raspberry ; vcgencmd version ; uname -a ; cat /sys/class/thermal/thermal_zone0/temp ; vcgencmd measure_temp
[    0.000000] OF: fdt: Machine model: Raspberry Pi 3 Model B Rev 1.2
Oct 11 2017 23:23:59
Copyright (c) 2012 Broadcom
version e5ad4a8e9f183413eb4f7fc4a7687886c715f739 (clean) (release)
Linux rpi3 4.14.0-rc5-v7+ #1 SMP Fri Oct 20 10:58:59 CEST 2017 armv7l GNU/Linux
34704
temp=39.7'C
```

Same setup on RPi2:
```
# dmesg | grep Raspberry ; vcgencmd version ; uname -a ; cat /sys/class/thermal/thermal_zone0/temp ; vcgencmd measure_temp
[    0.000000] OF: fdt: Machine model: Raspberry Pi 2 Model B Rev 1.1
Oct 11 2017 23:23:59
Copyright (c) 2012 Broadcom
version e5ad4a8e9f183413eb4f7fc4a7687886c715f739 (clean) (release)
Linux rpi3 4.14.0-rc5-v7+ #1 SMP Fri Oct 20 10:58:59 CEST 2017 armv7l GNU/Linux
36856
temp=36.9'C
```

A test with a milhouse LE build and kernel 4.14-rc5 on RPi B is fine, too:
```
# dmesg | grep Raspberry ; vcgencmd version ; uname -a ; cat /sys/cl
ass/thermal/thermal_zone0/temp ; vcgencmd measure_temp
[    0.000000] OF: fdt: Machine model: Raspberry Pi Model B Rev 2
Oct 11 2017 23:27:47
Copyright (c) 2012 Broadcom
version e5ad4a8e9f183413eb4f7fc4a7687886c715f739 (clean) (release)
[...]

PR #2236, opened 2017-10-21
bcm2710: Refactor bcm2710.dtsi using bcm2837.dtsi

pelwell:
The 4.14 tree includes bcm2837.dtsi - use it as a basis for
the downstream bcm2710.dtsi, which only contains a few tweaks.

See: https://github.com/raspberrypi/linux/issues/2235

### DIFF
--- a/arch/arm/boot/dts/bcm2708-rpi.dtsi
+++ b/arch/arm/boot/dts/bcm2708-rpi.dtsi
@@ -156,7 +156,5 @@ sdhost_pins: &sdhost_gpio48 {
 };
 
 &cpu_thermal {
-	coefficients = <(-538)	407000>;
-
 	/delete-node/ trips;
 };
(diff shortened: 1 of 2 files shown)


pelwell:
This patch still needs testing, but it's here for convenience.

HiassofT:
Thanks a lot, quick tests on RPi2 and RPi3 are looking good!

RPi3 now reports correct temperature via sysfs and the rpi-cirrus overlay loads/works still fine as well.

I'll run some more tests the next days.

MilhouseVH:
Posted this reply on the wrong PR...

LibreELEC with 4.14-rc7 + c7114a2 is looking good - temperatures from sysfs and vcgencmd now match.

pelwell:
Thanks, Milhouse. As this is still a bleeding edge branch I'll take the risk of some breakage in exchange for more testing.

Issue #2258, opened 2017-11-05
Include CONFIG_USB_DWCOTG as module fails


malikolivier:
I included CONFIG_USB_DWCOTG as a module.

However, I get a compile error as below:

```
ERROR: "DWC_TASK_HI_SCHEDULE" [drivers/usb/host/dwc_otg/dwc_otg.ko] undefined!
```

I included by defconfig here: https://gist.github.com/malikolivier/abd6d96bd559476c39a12efd86ada63e

I don't know yet if it's something unique to the linux kernel for raspberrypi. Does anyone have some idea?
I suspect that DWC_TASK_HI_SCHEDULE, which is defined in `drivers/usb/host/dwc_common_port/dwc_common_linux.c` is not included when CONFIG_USB_DWCOTG is built as a module.

pelwell:
The standard Raspberry Pi configurations build dwc_otg as an in-kernel module, since it is required for an NFS root filing system. If you (or anyone) have a patch to make it work as a module then we would probably merge it, but until then if you want a module-based USB driver for Raspberry Pi then I would suggest CONFIG_USB_DWC2=m.

malikolivier:
Thanks for the feedback.
 
I was trying to reduce the boot time as much as possible, and it happens that dwc_otg_driver_init  is taking around 0.5ms on my config (compute module). That's why I thought about including USB support as a module.

I'll try your suggestions!

malikolivier:
It seems that USB fails to initialize with CONFIG_USB_DWC2 included (whether it is built in or as a module).

```
root@pi:~# lsusb --verbose
unable to initialize libusb: -99
```

```
root@pi:~# cat /lib/modules/$(uname -r)/modules.builtin | grep usb
kernel/drivers/usb/common/usb-common.ko
kernel/drivers/usb/core/usbcore.ko
kernel/drivers/usb/dwc2/dwc2.ko
```

pelwell:
You need to select the alternate driver with the dwc2 overlay:
```
pi@raspberrypi:~ $ dtoverlay -h dwc2
Name:   dwc2

Info:   Selects the dwc2 USB controller driver

Usage:  dtoverlay=dwc2,<param>=<val>

Params: dr_mode                 Dual role mode: "host", "peripheral" or "otg"

        g-rx-fifo-size          Size of rx fifo size in gadget mode

        g-np-tx-fifo-size       Size of non-periodic tx fifo size in gadget
                                mode
```
[...]

PR #2265, opened 2017-11-13
USB_DWCOTG: Disable building dwc_otg as a module

pelwell:
When dwc_otg is built as a module, build will fail with the following
error:

ERROR: "DWC_TASK_HI_SCHEDULE" [drivers/usb/host/dwc_otg/dwc_otg.ko] undefined!
scripts/Makefile.modpost:91: recipe for target '__modpost' failed
make[1]: *** [__modpost] Error 1
Makefile:1199: recipe for target 'modules' failed
make: *** [modules] Error 2

Even if the error is solved by including the missing
DWC_TASK_HI_SCHEDULE function, the kernel will panic when loading
dwc_otg.

As a workaround, simply prevent user from building dwc_otg as a module
as the current kernel does not support it.

See details in https://github.com/raspberrypi/linux/issues/2258.

Signed-off-by: Malik Olivier Boussejra <malik@boussejra.com>

### DIFF
--- a/drivers/usb/host/Kconfig
+++ b/drivers/usb/host/Kconfig
@@ -763,7 +763,7 @@ config USB_HWA_HCD
 	  will be called "hwa-hc".
 
 config USB_DWCOTG
-	tristate "Synopsis DWC host support"
+	bool "Synopsis DWC host support"
 	depends on USB && (FIQ || ARM64)
 	help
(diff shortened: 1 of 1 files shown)


pelwell:
@P33M @popcornmix Any objections?

popcornmix:
Makes sense to me.

P33M:
No objections here.

PR #2266, opened 2017-11-13
vc4 backport

anholt:
This branch fixes an occasional oops in the OOM work handler for V3D, some bits of the DSI support, a memory leak, and adds Dave's new formats support for the video decode work he's been doing.


anholt:
Added another stability fix today.

pelwell:
Commit https://github.com/raspberrypi/linux/pull/2266/commits/8e809a71ff2f4f028900efb543dfb3ea6175b58a ("drm/vc4: Set up the DSI host at pdev probe time, not component bind.") breaks the full KMS driver for me - the screen remains blank and the monitor switches off. With that single commit reverted it seems to work OK.

I also get a (harmless?) warning from the DRM subsystem during boot because during initialisation the IRQ handler calls into drm_handle_vblank at a time when no CRTCs have been added.

anholt:
OK, I've pulled out the probing changes for DSI so that shouldn't regress, but with those removed, the transactions change failed so we're still stuck with sparklies.  Discussion on dri-devel is continuing, hoping to have something soon.  This branch still gets us some useful changes: the stability fix, some dsi fixes, memory leak fixes, and Dave's new formats support for the work he's been doing.

mi-hol:
@pelwell @anholt I'm wondering if this PR got abandoned?

popcornmix:
I think this got overlooked. KMS and FKMS tested ok with this PR.

Issue #2263, opened 2017-11-11
Using IRQF_TRIGGER_LOW results in spi->irq = 0 after module reload


msperl:
Hi!

Strange observation while writing a new driver:

When using the flags: IRQF_ONESHOT | IRQF_TRIGGER_LOW
with request_threaded_irq I see that after I unload and reload
the module that the spi->irq value is 0.

When instrumenting my code dumping spi->irq in the remove and probe code, then I get the following messages in demsg:
```
[ 1282.222746] mcp2517fd spi0.0: remove: IRQ: 176
[ 1282.311991] irq: type mismatch, failed to map hwirq-16 for /soc/gpio@7e200000!
[ 1282.312031] mcp2517fd spi0.0: probe: IRQ: 0
```
Only solution is rebooting the rpi - then it works 

This does not happen when using IRQF_ONESHOT | IRQF_TRIGGER_FALLING.
I can go thru the rmmod, insmod sequence 100 times without requiring a reboot.

Any Ideas why this could be and how to solve it?

Not using TRIGGER_LOW gives me a race-condition in my interrupt handler...

Any ideas?

Thanks, Martin

P.s: I am using is: 4.9.57 - commit: 2dddb0415295. I have not tried using the upstream kernel yet...
Sidenote: I use the default devicetree for the CM1 (bcm2708-rpi-cm.dtb) with a simple dt-overlay based on mcp2515-can0.

msperl:
Note that after applying this (debugging) patch:
https://www.spinics.net/lists/arm-kernel/msg528469.html

I get:
```
[   64.049763] mcp2517fd spi0.0: remove: IRQ: 176
[   64.165360] irq: type mismatch (2/8), failed to map hwirq-16 for /soc/gpio@7e200000!
[   64.165404] mcp2517fd spi0.0: probe: IRQ: 0
[   64.201945] mcp2517fd spi0.0: probe2: IRQ: 0
```
type 2 = IRQ_TYPE_EDGE_FALLING
type 8 = IRQ_TYPE_LEVEL_LOW

So I wonder how this type gets set to IRQ_TYPE_EDGE_FALLING.

msperl:
For completeness here the request_irq portion of the code:

```
	ret = request_threaded_irq(spi->irq, NULL,
				   mcp2517fd_can_ist,
				   IRQF_ONESHOT | IRQF_TRIGGER_LOW,
//				   IRQF_ONESHOT | IRQF_TRIGGER_FALLING,
[...]

Issue #2175, opened 2017-08-24
MCP2515 Hangs When Set to IRQF_TRIGGER_FALLING


Chris-W-Weir:
https://github.com/raspberrypi/linux/blob/dbe9b03bef2b949093b4393d5f4efd8ef5cdb772/drivers/net/can/spi/mcp251x.c#L939

I'm working with a Raspberry Pi 3 and a PiCAN Duo, and my can0/can1 interfaces will hang after a period of time ranging from 30 seconds to 15 minutes. 

I hooked up a logic analyzer to the MCP2515 and saw that the interrupt line was getting stuck low (active) and the can interface would resume if the line was pulled high again. I suspect that the falling edge interrupt is being missed, though I don't understand why.

I changed line 939 to be
```C
unsigned long flags = IRQF_ONESHOT | IRQF_TRIGGER_LOW;
``` 
 (level triggered low) and the issue has gone away. I haven't had a chance to test if this causes the CAN interface to miss messages or not, but it seems to be working as far as I can tell.

If the falling edge trigger is required for some reason, or I'm doing something foolish, please let me know.

Chris

pelwell:
This is a duplicate of https://github.com/raspberrypi/linux/issues/1490, a long-standing issue for which up to now we don't have a fix. I'd like to read up a bit to verify that your change is safe, but I would always prefer a level-sensitive interrupt to an edge-sensitive one for this reason, so I'm hopeful that we may be able to close both issues soon thanks to your suggestion.

Chris-W-Weir:
I have a 4 channel MCP2515 board I built using SPI 0.0, 0.1, 1.0, and 1.1, which stresses the Pi further, let me know if I can be of any assistance with research into the issue.

mypiandrew:
Hi

I'm busy debugging my dual can adapter and seeing some weird initialization bug ~~side effects of using the flip-flip gate array that albertodesouza provided~~

I much prefer this approach 

If we're not sure about making a sweeping change why don't we add a module option to this to specify whether you want a edge or level trigger and default it to the edge as standard?

I'm not the greatest of C coders, but I can hack together a patch if that helps move things forward 

It would be nice to clear this up before the holiday season.

Chris-W-Weir:
I've seen on other overlays the trigger type as an option, but I'm not sufficiently familiar with how they work to implement it myself.

mypiandrew:
Hey Chris,

[...]

PR #2267, opened 2017-11-14
mcp2515: Change interrupt to TRIGGER_LOW

pelwell:
[ If this is effective we should reconsider all usage of edge-triggered interrupts. ]

The MCP2515 datasheet clearly describes a level-triggered interrupt
pin. Therefore the receiving interrupt controller must also be
configured for level-triggered operation otherwise there is a danger
of a missed interrupt condition blocking all subsequent interrupts.

The ONESHOT flag ensures that the interrupt is masked until the
threaded interrupt handler exits.

See: https://github.com/raspberrypi/linux/issues/2175
     https://github.com/raspberrypi/linux/issues/2263

### DIFF
--- a/arch/arm/boot/dts/overlays/mcp2515-can0-overlay.dts
+++ b/arch/arm/boot/dts/overlays/mcp2515-can0-overlay.dts
@@ -60,7 +60,7 @@
                 pinctrl-0 = <&can0_pins>;
                 spi-max-frequency = <10000000>;
                 interrupt-parent = <&gpio>;
-                interrupts = <25 0x2>;
+                interrupts = <25 8>; /* IRQ_TYPE_LEVEL_LOW */
                 clocks = <&can0_osc>;
             };
         };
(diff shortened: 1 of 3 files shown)


msperl (commented):
For all practical purposes do not specify the trigger type in the driver itself - use 0.

Then  the settings from the Device-Tree will get used. That means you will be "backwards-compatible" with earlier DTs...

pelwell:
Thanks. I've reworked it to simply zero the flags if the device has a DT node.

msperl:
you could also change the DT to:
`#include <dt-bindings/interrupt-controller/irq.h>`
and then you can use:
`IRQ_TYPE_LEVEL_LOW` instead of 8.

But I am not sure if this would really work with the overlays the way these are built...

pelwell:
I considered that, and it does work, but people often build the overlays outside the kernel tree and I didn't want to add an unnecessary external dependency. I will change the comment to IRQ_TYPE_* though.

PR #2275, opened 2017-11-17
Fixed style issues for ac.c

ghost:
This fixes the following style issues for ac.c:
- ac.c:88: ERROR: "foo * bar" should be "foo *bar"
- ac.c:105: ERROR: code indent should use tabs where possible
- ac.c:127: ERROR: code indent should use tabs where possible
- ac.c:157: ERROR: code indent should use tabs where possible
- ac.c:233: ERROR: code indent should use tabs where possible


pelwell:
This is an unmodified upstream file. Any changes should be submitted upstream in the usual way.

PR #2149, opened 2017-08-05
Master volume added 

allocom:
On allo-piano-dac-plus.c: Master volume added, which controls both DACs volume.
On allo-digione.c: 192kHz sampling rate clicking sound fix.

PR #2279, opened 2017-11-18
RFC: Commits lost between rpi-4.9.y and rpi-4.14.y

pelwell:
These are commits that I've found to have been omitted or lost between rpi-4.9.y and rpi-4.14.y. I'm fairly sure some are no longer needed, but I've listed them all here anyway.

@popcornmix, @6by9, @JamesH65, @anholt - can you review (as will I - I've not read them all to sanity check)?


HiassofT:
"BCM270x DT: expose 3.3V and 5V system rails" isn't needed, that's already present in commit 7070ee4707d19d947a043b2febd98c337bfa70c4

pelwell:
Agreed - I've removed it from the PR. Thanks.

The old downstream firmware-based thermal driver has been replaced by the upstream equivalent that accesses the hardware directly, so that commit has also been removed, as has the patch enabling the AUX interrupt controller in DT (now moved to bcm270x.dtsi).

The two BRCMDBG-enabling commits have now been squashed into one, but do we need it at all @JamesH65?

The remaining commits are now grouped thematically for easier review and squashing.

6by9:
Certainly yes to "BCM_VC_SM: select DMA_SHARED_BUFFER".

I thought I had seen about 3 V4L2 driver patches originally, but not now. Did they disappear? (Yes they ought to be upstreamed - just another job)

pelwell:
Here are the V4L2 patches - have they been replaced by the upstream driver?

popcornmix:
All look correct except the alsa one. There is now an upstream/staging alsa driver in:
``drivers/staging/vc04_services/bcm2835-audio``
which I believe works. Now it's possible there are missing features or bug fixes, but if so we should probably patch the upstream/staging driver rather than adding a second.

pelwell:
Thanks - I've dropped the downstream ALSA driver.

6by9:
> Here are the V4L2 patches - have they been replaced by the upstream driver?

Partially.
The upstream driver is missing the last 4 fixes, hence my comment on needing to upstream them. It's also had a couple of things removed, eg the GStreamer workaround. It's true that GStreamer has been fixed but WebRTC fails due to the same issue so really we still want the workaround to be available. I'll have a think about it tomorrow.

pelwell:
I've rounded up a few more stray commits.

@6by9 The downstream V4L2 driver uses the same config symbol as the upstream driver - CONFIG_VIDEO_BCM2835. The redefinition triggers warnings, as does the fact that downstream it's a boolean while the 4.14 defconfigs set it to "m". If it's alright with you I propose to merge all the commits except those for the downstream V4L2 driver, and you can upstream the patches at your leisure.

anholt:
Ack on the vc4/panel changes.

allocom:
Hello Pelwell, I see 1 commit of files allo-piano-dac-plus & allo-digione are not included. Any specific reason? Please let me know what should be done to add it back.

pelwell:
I've squashed three piano-dac-plus commits together, but if you know of some missing commits from 4.14 which aren't in this PR then list them and I'll make sure they are included.

pelwell:
With this PR applied, all of the 4.14 allo source files now match the 4.9 versions, so I'm satisfied there are no omissions.

6by9:
>  If it's alright with you I propose to merge all the commits except those for the downstream V4L2 driver, and you can upstream the patches at your leisure.

Fine by me.

allocom:
3 Commits are missing,  #2135, #2224 & #2149

pelwell:
No - they aren't missing, they are squashed into https://github.com/raspberrypi/linux/pull/2279/commits/55da3a6c60c1cb3decd3ede6556d4b7a60760cf9.

JamesH65:
commit 169139323fe81d05f2fb7805e930c1b04ee2048e seems to be missing?

allocom:
Ok, but change are not visible on 4.14.y tree.

PeterPablo:
That is because this is a pull request that is not yet merged to the
branch. Its diff shows you, what the branch will look like when it is
applied (=merged) .

Am 21.11.2017 11:34 schrieb "Allo" <notifications@github.com>:

PR #2281, opened 2017-11-20
RFC: Proper way to define BCM2836 interrupt polarity

lategoodbye:
Based on the discussion about [irq-bcm2836: Avoid "Invalid trigger warning"](https://marc.info/?l=linux-arm-kernel&m=151081530017240&w=2) i prepared this pull request for discussion, but not merging.

Since i don't know the right interrupt polarity, i will be happy about any corrections.

@popcornmix @pelwell @anholt @msperl


msperl:
One note: as of now I see some issues with level interrupts - at least on the RPI code!

My code contains a counter to see how often the irqhandler gets called.

On initialization of the device I get a single interrupt.

Here the counter inside the driver:
```
root@raspcm3:~# cat /sys/kernel/debug/mcp2517fd-can0/irq_calls
1
```
and here what /proc/interrupts shows:
```
root@raspcm3:~grep mcp251 /proc/interrupts 
182:       3732          0          0          0  pinctrl-bcm2835  16 Level     mcp2517fd
```

I wonder why this is the case - it is obviously consuming CPU unnecessarily - as if the level IRQ is not masked/disabled on the HW while the IRQ handler itself is running...

Should this handle this issue as well?

lategoodbye:
@msperl 
Yes, we should handle this, but not in this patch series. This applies only to the first level interrupt controller on BCM2836/2837.

pelwell:
@lategoodbye The whole concept of interrupt polarity for the BCM2836 (and BCM2835 for that matter) seems to be of academic interest only. None of the interrupt lines are exposed directly to the outside world, so the polarity is of no consequence. Any externally visible interrupts go through an intermediate hardware block (such as the GPIO controller), which is where the polarity is specified (unless the interface standard dictates a specific polarity).

I have no problem with the other changes - the use of _map to avoid the pre-registration hack - but let's 
be clear that the only reason for specifying an interrupt polarity is to keep that arch timer support quiet.

@msperl Are you using threaded IRQs (either explicitly or by deferring the interrupt processing to a workqueue)? If so, are you marking the interrupt as ONESHOT? That should result in the interrupt being masked until the handler has completed.

lategoodbye:
@pelwell Did you notice that my polarity values are different from your approach (set all to level low)? Looking at the other ARMv7 DTS shows that the PMU usually have a high level polarity. Where can we find these information?

pelwell:
Ask yourself this question: What changes in the hardware when I change the polarity in the DT?

msperl:
@pelwell: All I use is:
```
	ret = request_threaded_irq(spi->irq, NULL,
				   mcp2517fd_can_ist,
				   IRQF_ONESHOT | IRQF_TRIGGER_LOW,
				   DEVICE_NAME, priv);
```
but for all practical purposes I should be using IRQF_ONESHOT for flags only, as
a) the IRQ level is defined in the DT and if there is a mismatch, then it produces errors (see http://elixir.free-electrons.com/linux/latest/source/kernel/irq/manage.c#L1152)
b) even then in the same place ONE_SHOT may be cleared if the driver is declared ONE_SHOT_SAFE

So for all practical purposes I should be doing the "right" thing...

pelwell:
@msperl Your code is as I would expect it to be, so if you are seeing excessive interrupts then there is almost certainly a kernel problem.

There have been a number of interrupt-related issues recently so I'm keen to understand this behaviour. Can you create a new issue to track this, since we've drifted off-topic?

msperl:
Note that this observation is made against 4.9.35 right now, so going further up it may be slightly different... I just want the driver to work with the most common kernel in use right now ;)

msperl:
@pelwell: sorry, but that is an artifact - I forgot that interrupt counters did not get reset when the irq was requested (after module unload/reload)...
After a clean reboot I see expected behavior!

lategoodbye:
The main intension behind this patch series is to describe the hardware like the other ARM boards (even it seems to be academic interest) and avoid situation like a driver XY needs to know that a irq is edge or level triggered, but the DT don't provide these information. Fixing these arch timer warning is a nice side effect.

In case nobody wants to provide these super secret information, i will drop these series and fallback to Marc Zygnier's suggestion.

pelwell:
It is not super secret information, but I don't know the answer and can't know without asking somebody at Broadcom to read the Verilog.

Toss a coin and choose a value - it really makes no difference. Just like knowing whether audio phono leads are connected with red or white for the right channel, it's immaterial as long as both ends agree.

anholt:
All the interrupts between the blocks in the SOC that I've seen are high polarity.  Let's use that in general, if the DT value doesn't actually affect behavior.

lategoodbye:
So i would change the behavior of irq-bcm2836 in case there is no 2nd interrupt cell back to no polarity as before this patch. This would simplify things because in could use irq_domain_xlate_onetwocell() instead of open-code it.

PR #2282, opened 2017-11-21
enable sht3x module in kernel, add sht3x definition to i2c-sensor dts definition

ubergesundheit:
This pull request enables the `sht3x` kernel module which in turn can be enabled by specifying
```
dtoverlay=i2c-sensor,sht3x
```
in `config.txt`. The `dtoverlay` instruction also supports changing the i2c address which is used to read the sensor.


pelwell:
That looks fine as far as it goes, but I don't understand your comment:

>  The dtoverlay instruction also supports changing the i2c address which is used to read the sensor.

You can make the address selectable using the `addr` parameter by adding `,<&sht3x>,"reg:0"` to the `addr` `__overrides__` declaration and documenting the fact that the valid addresses are 0x44 and 0x45.

pelwell:
Also, the CONFIG_CRC8 config option is automatically selected when you select CONFIG_SENSORS_SHT3x, so don't include it in the defconfigs.

ubergesundheit:
It seems I had the address changes staged but forgot to commit and push ;)

pelwell:
One final nitpick - can you add SHT3x to the `addr` documentation?

ubergesundheit:
Ah my apologies. I've done this kind of pull request before but forgot everything 😅

ubergesundheit:
Thank you very much!

PR #2287, opened 2017-11-24
Rpi 4.7.y

srishtik:


pelwell:
I assume that was accidental - we won't be updating the 4.7 branch.

PR #2296, opened 2017-12-02
Updates for Pisound module code:

gtrainavicius:
- Merged 'Fix a warning in DEBUG builds' (1c8b82b).
- Updating some strings and copyright information.
- Fix for handling high load of MIDI input and output.
- Use dual rate oversampling ratio for 96kHz instead of single rate one.


pelwell:
That looks OK to me - the automated tests are happy, and I'm assuming you've tested it - but do you really want to change the Copyright date in this way? An older Copyright date is potentially better than a newer one if somebody else tries to also claim copyright over the code - "2016-2017" is safer.

gtrainavicius:
Makes sense, updated to 2016-2017, thank you for the feedback! :)

Yes, I have tested the changes on our side, so it should be good to go.

PR #2310, opened 2017-12-21
Add Raspberry Pi firmware driver to the dependencies of RPi backlight driver

ar-cetitec:
Otherwise the backlight driver fails to build if the firmware
loading driver is not in the kernel

Signed-off-by: Alex Riesen <alexander.riesen@cetitec.com>


pelwell:
Thanks!

PR #2313, opened 2017-12-23
add media center hat overlay

shawaj:
add overlay for Media Center HAT by Pi Supply

### DIFF
--- a/arch/arm/boot/dts/overlays/Makefile
+++ b/arch/arm/boot/dts/overlays/Makefile
@@ -60,6 +60,7 @@ dtbo-$(CONFIG_ARCH_BCM2835) += \
 	mcp2515-can0.dtbo \
 	mcp2515-can1.dtbo \
 	mcp3008.dtbo \
+	media-center.dtbo \
 	midi-uart0.dtbo \
 	midi-uart1.dtbo \
 	mmc.dtbo \
--- a/arch/arm/boot/dts/overlays/README
+++ b/arch/arm/boot/dts/overlays/README
@@ -983,6 +983,30 @@ Params: spi<n>-<m>-present      boolean, configure device at spi<n>, cs<m>
         spi<n>-<m>-speed        integer, set the spi bus speed for this device
 
 
+Name:   media-center
+Info:   Media Center HAT - 2.83" Touch Display + extras by Pi Supply
+Load:   dtoverlay=media-center,<param>=<val>
+Params: speed                   Display SPI bus speed
+        rotate                  Display rotation {0,90,180,270}
+        fps                     Delay between frame updates
+        xohms                   Touchpanel sensitivity (X-plate resistance)
+        swapxy                  Swap x and y axis
+        gpio_out_pin            GPIO for output (default "17")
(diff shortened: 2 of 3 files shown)


pelwell:
1. This doesn't compile because the build rule for .dtbos is looking for "*-overlay.dts", so you will have to rename your dts file.
2. The overlay appears to be two glued back-to-back. This may work, but it is confusing.
3. The parameters for the second embedded overlay are undocumented.
4. You used spaces in the Makefile where all other overlays use a hard tab.

shawaj:
1. sorry about that! updated name now 
2. it is tested and working on the HAT EEPROM
3. added further parameter documentation
4. not sure what happened there. fixed now though.

all squashed into a single commit

pelwell:
I reformatted the overlay to have just one __overrides__ node, and in doing so discovered that there are two `debug` parameters. This wouldn't have worked - the second instance would have replaced the first - so they are now concatenated properly.

shawaj:
@pelwell thanks very much! Wasn't expecting such a speedy response over Christmas :-)

One question for you - is there a way we can make the led backlight parameter configurable with an override?

Hope you have a great Christmas and New Year

pelwell:
> is there a way we can make the led backlight parameter configurable with an override?

If the device driver provides a way to control the backlight via a DT property then it shouldn't be hard to add a parameter to control it. How is the backlight connected?

shawaj:
Backlight is, on our board, by default connected on gpio12 but can also optionally be changed to gpio18 via a jumper.

Gpio18 is what most other displays use (such as rpi-display) but that clashes with the i2s audio pin so we moved it. 

The line `led-gpios = <&gpio 12 1>;` in the overlay determines the backlight setting in the fbtft driver so in essence would just need that configurable.

So busy an override in the overlay?

shawaj:
*just

shawaj:
I guess it would also need to change this line too:

```
brcm,pins = <12 23 24 25>;
```

Would an override like this work:

```
backlight =  <&rpidisplay>,"led-gpios:0";
                      <&rpi_display_pins>,"brcm,pins:0";
```

shawaj:
Or maybe it should be:

```
backlight = <&rpidisplay>,"led-gpios:4",
            <&rpi_display_pins>,"brcm,pins:0";
```

shawaj:
@pelwell - ok - I have tested the backlight override and it works.

See attached .dts and .dtbo files I tested with
[media-center-overlays.zip](https://github.com/raspberrypi/linux/files/1589390/media-center-overlays.zip)

If you are happy with these, I will submit a PR changing them to add this feature?

Perhaps this would also be useful in the [rpi-display overlay](https://github.com/raspberrypi/linux/blob/rpi-4.9.y/arch/arm/boot/dts/overlays/rpi-display-overlay.dts) also, which I also tested successfully, and files attached for reference.

[rpi-display-overlays.zip](https://github.com/raspberrypi/linux/files/1589391/rpi-display-overlays.zip)

There are also other drivers for different displays - hy28a, hy28b and the various pitft models that could do similar. But maybe that is overkill...

PR #2324, opened 2018-01-01
Add overlay for mcp3202 12 bit ADC

penfold42:


pelwell:
Thanks. I wish there was a neater way of handling the multiple bus and chip select combinations, but for now this is the best one can do.

penfold42:
I also wish there were a better way to handle the 10+ ADC chips the module supports without blindly cloning the existing overlay.

This was literally a sed -e s/3008/3202/g from the mcp3008 overlay

pelwell:
I'm hoping to get some time to improve this area, but it won't be soon.

Dloranger:
I think there is a problem with the MCP3202 (aka 12 bit ADC's) overlay. I think it inherited the 10 bit output instead of the 12 bit output as intended.  I have been using the 3008 overlay for some time and someone pointed out there is only 10 bit resolution so I loaded up the 3202 version to get the enhanced resolution, but the ADC is still reporting the same amplitude signals. 

Can you double check on this for me

12bit overlay
root@openrepeater:~# cat /sys/bus/iio/devices/iio\:device0/in_voltage1_raw
590

10bit overlay
root@openrepeater:~# cat /sys/bus/iio/devices/iio\:device0/in_voltage1_raw
581

here is how I am instating the part

#Enable mcp3208 (10 bit) adc overlay, MCP3202 is the only offered
dtoverlay=mcp3208:spi0-0-present,spi0-0-speed=3600000
#12bit dac, but works the same for the purposes of the overlay
#and drivers
#dtoverlay=mcp3202:spi0-0-present,spi0-0-speed=3600000

pelwell:
Both overlays are implemented by a single driver that understands the differences between the devices. The ADC conversion reading code includes this section:
```
	switch (device_index) {
	case mcp3001:
		*val = (adc->rx_buf[0] << 5 | adc->rx_buf[1] >> 3);
		return 0;
	case mcp3002:
	case mcp3004:
	case mcp3008:
		*val = (adc->rx_buf[0] << 2 | adc->rx_buf[1] >> 6);
		return 0;
	case mcp3201:
		*val = (adc->rx_buf[0] << 7 | adc->rx_buf[1] >> 1);
		return 0;
	case mcp3202:
	case mcp3204:
	case mcp3208:
		*val = (adc->rx_buf[0] << 4 | adc->rx_buf[1] >> 4);
		return 0;
```
Some added debug shows that the different overlays are activating the correct personalities of the driver:
```
[    7.538275] mcp320x_probe: 'mcp3008', 3
```
vs
```
[    7.495489] mcp320x_probe: 'mcp3202', 5
```
The number `3` corresponds to `mcp3008` in the switch statement  above, and `5` is `mcp3202`. In other words I think the overlay is working as it should, and the driver is treating the two devices differently.

Doh! Your overlay usage doesn't match your intentions - you are comparing `mcp3202` with `mcp3208`, another 12-bit device, while I think you wanted to compare it to `mcp3008`, which is indeed a 10-bit device that will show different results in the raw output.

Dloranger:
I just realized I had multiple copies of the dtoverlay in the config.txt file, the change I thought I was doing was getting stomped on in the other copy. I got that cleaned up and running the test again and things are behaving as expected now.  Sorry about the confusion here

root@openrepeater:~# cat /boot/config.txt | grep mcp3
#Enable mcp3008 adc overlay
dtoverlay=mcp3008:spi0-0-present,spi0-0-speed=3600000
root@openrepeater:~# cat /sys/bus/iio/devices/iio\:device0/in_voltage1_raw
600
root@openrepeater:~#


~~~~~~~

root@openrepeater:~# cat /boot/config.txt | grep mcp3
#Enable mcp3208 adc overlay
dtoverlay=mcp3202:spi0-0-present,spi0-0-speed=3600000
root@openrepeater:~# cat /sys/bus/iio/devices/iio\:device0/in_voltage1_raw
2681
root@openrepeater:~#

PR #2333, opened 2018-01-04
vcsm: Unify cache manipulating functions

Terminus-IMRC:
`vcsm` has four ioctl calls for cache flushing/cleaning/invalidating: `FLUSH`, `INVALID`, `CLEAN_INVALID`, `CLEAN_INVALID2`.
The operations are finally done by `dmac_{flush,clean,inv}_range()`, but the ioctl handlers used individual wrapper functions. This PR unifies them.

### DIFF
--- a/drivers/char/broadcom/vc_sm/vmcs_sm.c
+++ b/drivers/char/broadcom/vc_sm/vmcs_sm.c
@@ -331,8 +331,8 @@ static unsigned int vmcs_sm_usr_handle_from_pid_and_address(unsigned int pid,
 		list_for_each_entry(map, &sm_state->map_list, map_list) {
 			if (map->res_pid != pid)
 				continue;
-			if (!(map->res_addr <= addr &&
-						addr < map->res_addr + map->resource->res_size))
+			if (addr < map->res_addr ||
+						addr >= (map->res_addr + map->resource->res_size))
 				continue;
 
 			pr_debug("[%s]: global map %p (pid %u, addr %lx) -> usr-hdl %x (vc-hdl %x)\n",
--- a/drivers/char/broadcom/vc_sm/vmcs_sm.c
+++ b/drivers/char/broadcom/vc_sm/vmcs_sm.c
@@ -2990,7 +2990,7 @@ static long vc_sm_ioctl(struct file *file, unsigned int cmd, unsigned long arg)
 							op->block_count, op->block_size,
 							op->inter_block_stride, op->invalidate_mode);
 					if (ret)
-						goto out;
+						break;
 				}
 				kfree(block);
 			}
--- a/drivers/char/broadcom/vc_sm/vmcs_sm.c
+++ b/drivers/char/broadcom/vc_sm/vmcs_sm.c
@@ -1269,6 +1269,8 @@ static int clean_invalid_mem_2d(const void __user *addr,
 	}
 
 	switch (cache_op) {
+	case VCSM_CACHE_OP_NOP:
+		return 0;
 	case VCSM_CACHE_OP_INV:
 		op_fn = dmac_inv_range;
 		break;
@@ -1312,6 +1314,8 @@ static int clean_invalid_resource(const void __user *addr, const size_t size,
 		return 0;
 
 	switch (cache_op) {
+	case VCSM_CACHE_OP_NOP:
+		return 0;
 	case VCSM_CACHE_OP_INV:
 		stat_attempt = INVALID;
 		stat_failure = INVALID_FAIL;
@@ -2936,6 +2940,9 @@ static long vc_sm_ioctl(struct file *file, unsigned int cmd, unsigned long arg)
 				goto out;
 			}
 			for (i = 0; i < sizeof(ioparam.s) / sizeof(*ioparam.s); i++) {
+				if (ioparam.s[i].cmd == VCSM_CACHE_OP_NOP)
+					break;
+
 				/* Locate resource from GUID. */
 				resource =
 					vmcs_sm_acquire_resource(file_data, ioparam.s[i].handle);
@@ -2989,6 +2996,9 @@ static long vc_sm_ioctl(struct file *file, unsigned int cmd, unsigned long arg)
 					ret = clean_invalid_mem_2d((void __user*) op->start_address,
 							op->block_count, op->block_size,
 							op->inter_block_stride, op->invalidate_mode);
+					if (op->invalidate_mode == VCSM_CACHE_OP_NOP)
+						continue;
+
 					if (ret)
 						break;
 				}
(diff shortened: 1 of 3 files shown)


Terminus-IMRC:
Oh, should I propose this PR to `rpi-4.14.y` branch?

drivers/char/broadcom/vc_sm/vmcs_sm.c
```diff
@@ -276,7 +276,10 @@ static unsigned int vmcs_sm_vc_handle_from_pid_and_address(unsigned int pid,
 	/* Lookup the resource. */
 	if (!list_empty(&sm_state->map_list)) {
 		list_for_each_entry(map, &sm_state->map_list, map_list) {
-			if (map->res_pid != pid || map->res_addr != addr)
+			if (map->res_pid != pid)
+				continue;
+			if (!(map->res_addr <= addr &&
```
6by9: Seems a little obfuscated.
To my mind 
```
if (addr < map->res_addr || addr > (map->res_addr + map->resource->res_size))
   continue;
```
is more obvious.
(In a percentage of cases it is likely to be faster too, as it can avoid computing ```addr > (map->res_addr + map->resource->res_size)``` if ```addr < map->res_addr``` was true)

include/linux/broadcom/vmcs_sm_ioctl.h
```diff
@@ -79,6 +79,11 @@ enum vmcs_sm_cache_e {
 	VMCS_SM_CACHE_BOTH,
 };
 
+/* Cache functions */
```
6by9: Could you add either a comment that this relates to struct vmcs_sm_ioctl_clean_invalid cmd and struct vmcs_sm_ioctl_clean_invalid2 invalidate_mode, or move it to just before those structs?

drivers/char/broadcom/vc_sm/vmcs_sm.c
```diff
@@ -1256,61 +1256,106 @@ static const struct vm_operations_struct vcsm_vm_ops = {
...
-	pud_t *pud;
-	pmd_t *pmd;
-	pte_t *pte;
-	unsigned long pgd_next, pud_next, pmd_next;
+	size_t i;
+	void (*op_fn)(const void*, const void*);
 
-	if (addr >= end)
-		return;
+	if (block_size <= 0) {
```
6by9: Nit pick:
```if (!block_size)```
block_size is size_t = unsigned int, therefore it can never be < 0. Your comment already reflects this.

File: drivers/char/broadcom/vc_sm/vmcs_sm.c [hunk 4]
```diff
@@ -3006,36 +2985,12 @@ static long vc_sm_ioctl(struct file *file, unsigned int cmd, unsigned long arg)
...
-					for (j = 0; j < op->block_count; ++j) {
-						const char * const base = (const char *)op->start_address + j * op->inter_block_stride;
-						const char * const end = base + op->block_size;
-						op_fn(base, end);
-					}
+					ret = clean_invalid_mem_2d((void __user*) op->start_address,
+							op->block_count, op->block_size,
+							op->inter_block_stride, op->invalidate_mode);
+					if (ret)
+						goto out;
```
6by9: You leak ```block``` if you goto out here. break is fine.

(I have a cleanup pending to this file that removes a load of the ```goto out``` calls, as they're only doing the equivalent of running on, or break within the switch statement. I'll let your PR go through first, as long as it is functionally correct).

6by9 (changes requested):
One memory leak, and a couple of minor nit picks, otherwise it looks OK on a first pass.
@popcornmix You're more aware of where the vc-sm cache handling ioctl's are used. Any comments?

@pelwell 4.9 or 4.14? I guess it needs to be picked across to the other one anyway.

pelwell:
As you say, it needs cherry-picking either way, so you might as well keep this PR against 4.9.

Terminus-IMRC:
Thank you for reviewing!!

drivers/char/broadcom/vc_sm/vmcs_sm.c
Terminus-IMRC: Yeah, mine was obscure and not efficient. Fixed in https://github.com/raspberrypi/linux/pull/2333/commits/872f3ca2c4dd268bc6d537151435e45f16224b0e.

drivers/char/broadcom/vc_sm/vmcs_sm.c
Terminus-IMRC: I was writing the condition so that it can detect negative size if the size was unintentionally signed, but it's obviously unsigned here, so fixed in https://github.com/raspberrypi/linux/pull/2333/commits/872f3ca2c4dd268bc6d537151435e45f16224b0e.

drivers/char/broadcom/vc_sm/vmcs_sm.c
Terminus-IMRC: Thanks. It really leaks memory. Fixed in https://github.com/raspberrypi/linux/pull/2333/commits/4fd87fbe66417a4c44cf879617e3d1e9abf857f0.

include/linux/broadcom/vmcs_sm_ioctl.h
Terminus-IMRC: Thanks. I did both: https://github.com/raspberrypi/linux/pull/2333/commits/6e3cd4cb50ec8f19287a5b00ddd4563c6a6e255c

drivers/char/broadcom/vc_sm/vmcs_sm.c
Terminus-IMRC: (I changed the second condition of yours from `addr > (map->res_addr + map->resource->res_size)` to `addr >= (map->res_addr + map->resource->res_size)` because yours may match an neighboring map area.)

popcornmix:
I temporarily added this PR to 4.14 last night to get some testing in the Milhouse LibreELEC nightly builds. Unfortunately a number of people have reported a regression of video playback randomly hanging on startup. These commits were isolated as introducing the problem:
https://forum.kodi.tv/showthread.php?tid=298461&pid=2689109#pid2689109

Terminus-IMRC:
Thanks. It may be because the func is flushing a page which is not yet mapped by fault handler, according to [this reply](https://forum.kodi.tv/showthread.php?tid=298461&pid=2689209#pid2689209). I will implement page table walk which may solve the problem.

popcornmix:
Yes, CLEAN_INVALID2 was an optimised API for framebuffers that we know are contiguous and pinned on allocation.
Other buffers may only have pages pinned when first accessed.
Flushing a non-pinned page does fault, and the page table walk code does avoid that (but with a certain performance cost).

Terminus-IMRC:
So my changes are to flush/clean/inv correctly on CLEAN_INVALID and other minor stuffs, rather than unifying them... (Anyway, I keep going)

Terminus-IMRC:
Reverted to operate cache by walking page table for FLUSH, INVALID, CLEAN_INVALID ioctls. This should work with LibreELEC......

Terminus-IMRC:
I confirmed it's now fixed with LibreELEC!

popcornmix:
Last night's LibreELEC build included update patch and seems fine.

Terminus-IMRC:
Thanks for merging! Closing.

Issue #2331, opened 2018-01-03
rotary encoder - additional overrides?


shawaj:
According to this doc - https://elixir.free-electrons.com/linux/latest/source/Documentation/devicetree/bindings/input/rotary-encoder.txt

There are more configurable parameters for the rotary-encoder driver.

Would it be acceptable to add these all in as additional overrides to the driver?

Maybe something like this:
```
	__overrides__ {
		rotary0_pin_a =     <&rotary0>,"gpios:4",
		                    <&rotary0_pins>,"brcm,pins:0";
		rotary0_pin_b =     <&rotary0>,"gpios:16",
		                    <&rotary0_pins>,"brcm,pins:4";
		relative_axis =     <&knob>,"rotary-encoder,relative-axis";
		linux_axis =        <&knob>,"linux,axis";
		rollover =          <&knob>,"rotary-encoder,rollover";
		steps-per-period =  <&knob>,"rotary-encoder,steps-per-period";
		steps =             <&knob>,"rotary-encoder,steps";
		wakeup =            <&knob>,"wakeup-source";
		encoding =          <&knob>,"rotary-encoder,encoding";
    };
```

pelwell:
Yes, of course - go ahead.

As a general point, each PR is a both a request and a topic for discussion, so I don't consider it necessary to ask in this way, polite as it is. Going straight to a PR helps to reduce the clutter and keep all relevant discussion in the same place.

shawaj:
Understood! I'm new to these parts so don't know the etiquette yet and didn't want to step on any toes :-P

See your point though... Better if all in one place

shawaj:
Closing this issue now and moving over to #2334

PR #2334, opened 2018-01-05
add additional overrides to rotary-encoder overlay

shawaj:
Updates to the rotary-encoder overlay to add additional overrides that are possible according to this doc - https://elixir.free-electrons.com/linux/latest/source/Documentation/devicetree/bindings/input/rotary-encoder.txt

See #2331 for more details also

### DIFF
--- a/arch/arm/boot/dts/overlays/README
+++ b/arch/arm/boot/dts/overlays/README
@@ -1307,6 +1307,29 @@ Params: rotary0_pin_a           GPIO connected to rotary encoder channel A
                                 (default 4).
         rotary0_pin_b           GPIO connected to rotary encoder channel B
                                 (default 17).
+        relative_axis           register a relative axis rather than an
+                                absolute one. Relative axis will only
+                                generate +1/-1 events on the input device,
+                                hence no steps need to be passed.
+        linux_axis              the input subsystem axis to map to this
+                                rotary encoder. Defaults to 0 (ABS_X / REL_X)
+        rollover                Automatic rollover when the rotary value
+                                becomes greater than the specified steps or
+                                smaller than 0. For absolute axis only.
+        steps-per-period        Number of steps (stable states) per period.
+                                The values have the following meaning:
+                                1: Full-period mode (default)
+                                2: Half-period mode
+                                4: Quarter-period mode
+        steps                   Number of steps in a full turnaround of the
+                                encoder. Only relevant for absolute axis.
+                                Defaults to 24 which is a typical value for
+                                such devices.
+        wakeup                  Boolean, rotary encoder can wake up the
+                                system.
+        encoding                String, the method used to encode steps.
+                                Supported are "gray" (the default and more
+                                common) and "binary".
 
 
 Name:   rpi-backlight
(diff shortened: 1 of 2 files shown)


pelwell:
I've removed the trailing whitespace, but now your overlay seems to be knobless.

shawaj:
Darn! I guess I should change "knob" to "rotary0"?

pelwell:
Or vice versa - as long as everything agrees it doesn't matter.

pelwell:
You also need to think about the types of the properties you are applying the overrides to - currently all the new parameters are declared to target strings.

shawaj:
@pelwell ok fixed that issue. 

about the other thing you have mentioned - i am not entirely sure what you mean. do i need to set the default values up in `fragment@1`?

pelwell:
Although Device Tree contains no real type information - once compiled it is all just arrays of bytes - properties can be written as strings or arrays of integers of various sizes. Device drivers reading the properties then place an interpretation on those arrays of bytes, usually one that matches the way the property was written in the first place but not necessarily so.

The [DT parameter documentation](https://raspberrypi.org/documentation/configuration/device-tree.md#part2.2) describes the various parameter types and how to declare them. The new properties are mainly integers, but there is a boolean property which is true if it exists and false if it doesn't.

shawaj:
i guess linux,axis, steps, steps-per-period need to have `:0` after them as they would be numbers?

For rotary-encoder,relative-axis - that one doesn't appear to have anything with it? It seems as if the driver "defaults" to an absolute axis but then changes to relative if you declare it? So should I remove that from the `fragment@0` part and only declare it in overrides so you can use either?

and there is the wakeup-source one which is boolean - does that work the same that if declared it is true and if not declared anywhere it is false? The rotary-encoder,rollover one seems to be the same also

pelwell:
`rotary-encoder,relative-axis`, `rotary-encoder,rollover` and `wakeup-source` are booleans, which you declare with a `?` at the end. You create a boolean property which defaults to true by just writing the name without the assignment in the overlay - `wakeup-source;`. If it defaults to false - as all of these probably should - then omit the property from the fragment.

I think it is good practice to put the default property values explicitly into the fragments, for ease of understanding and to protect against a change in the driver's defaults - except for false booleans...

shawaj:
is there a way to check if the booleans default to false? it seems from the txt file I linked to above that they are all false, but I just wondered for future?

pelwell:
I was going to say that there's no easy way, and that I'd read the source (which I had), but it's obvious that booleans have to default (in the driver at least) to false, otherwise you can't distinguish an explicit override from leaving it at the default. What we choose to make the overlay default value is up to us.

shawaj:
ok updated - how does that look?

shawaj:
Hold on - spotted some errors already....one minute and I will fix!!

shawaj:
ok - take a look now

pelwell:
I think you forgot to push - the errors I'm expecting you to fix are that boolean properties are referred to as "wakeup-source?" not "wakeup-source:?".

pelwell:
While PRs are under development it is often better to keep updates as separate commits so it is clear what has changed at each step. GitHub allows changes to be viewed by commit or by file, which effectively squashes them to see the sum of all changes, and multiple commits can be squashed at merge time.

shawaj:
ok, have removed the colons now

pelwell:
Merged - thanks.

shawaj:
great, thanks :-)

PR #2338, opened 2018-01-09
dwc_otg: don't unconditionally force host mode in dwc_otg_cil_init()

P33M:
cc @ghollingworth


pelwell:
It looks like a no-op in the case that the parameter isn't specified, so it's OK by me.

ghollingworth:
Looks good to me

PR #2343, opened 2018-01-15
RFC: NVMEM driver for customer OTP

lategoodbye:
Hi,
this series is currently not intended for merging. I only want to get some feedback about the NVMEM driver for the customer OTP. It provides read-only access, but could be easily extended with write support. After this driver has been merged in mainline, it could be backported to 4.9.

More information about [NVMEM](https://github.com/lategoodbye/linux/blob/rpi-4.9.y-otp/Documentation/nvmem/nvmem.txt)


lategoodbye:
I sent out a newer version to linux-arm-kernel:
https://marc.info/?l=linux-arm-kernel&m=152060248226508&w=2

So we can close this.

PR #2345, opened 2018-01-17
bcm2835-i2s: fix hw_params error when device is in prepared state

HiassofT:
If bcm2835 is configured as bitclock master calling hw_params()
after prepare() fails with EBUSY. This also makes it impossible to
use bcm2835 in full duplex mode.

The error is caused by the split clock setup: clk_set_rate
is called in hw_params, clk_prepare_enable in prepare. As hw_params
doesn't check if the clock was already enabled clk_set_rate
fails with EBUSY.

Fix this by moving clock startup from prepare to hw_params and
let hw_params properly deal with an already set up or enabled
clock.

### DIFF
--- a/sound/soc/bcm/bcm2835-i2s.c
+++ b/sound/soc/bcm/bcm2835-i2s.c
@@ -130,6 +130,7 @@ struct bcm2835_i2s_dev {
 	struct regmap				*i2s_regmap;
 	struct clk				*clk;
 	bool					clk_prepared;
+	int					clk_rate;
 };
 
 static void bcm2835_i2s_start_clock(struct bcm2835_i2s_dev *dev)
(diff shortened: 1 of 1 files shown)


pelwell:
That change looks sensible to me. Are you planning to push this patch upstream?

HiassofT:
Yes, just sent out the mail and put you in CC

pelwell:
Cool. I'll wait until I see an Ack before merging, unless it is particularly urgent?

HiassofT:
Sure, just wait until we get some more feedback.

It's not urgent, that bug has been there forever without anyone noticing it. I tested full duplex mode before, but with bcm2835-i2s in slave mode (where the bug doesn't occur). I also tested with "aplay | arecord" both in slave and master mode, but the bug doesn't occur there either (because hw_params bails out early if a stream is already running so it doesn't hit the problematic code).

I became aware of the issue via this forum post https://www.raspberrypi.org/forums/viewtopic.php?p=1254351#p1254351 and also asked the poster to test and report back

HiassofT:
rebased PR on top of current rpi-4.14.y

pelwell:
```
The patch

   ASoC: bcm2835: fix hw_params error when device is in prepared state

has been applied to the asoc tree at

   https://git.kernel.org/pub/scm/linux/kernel/git/broonie/sound.git 
```
That counts as an Ack.

HiassofT:
wow, that was really quick!

@pelwell could you pick that into the 4.15 tree as well?

pelwell:
It would have been picked up in one of the regular cherry-picks, but it's done now.

HiassofT:
Thanks a lot!

PR #2346, opened 2018-01-18
vc4 backport (4.14.y)

anholt:
- MADVISE lets userspace cooperate with the kernel for CMA out-of-memory handling.
- Fixes an IRQ race during probe
- Adds HVS format support for Dave's work
- Fixes misrendering in 3DMMES
- Fixes display on 4K HDMI monitors


popcornmix:
@anholt I've tested this and seems fine with full kms driver enabled.
With fkms driver enabled I see no framebuffer on boot and cannot log in through ssh.
The fkms driver is working on latest 4.9 kernel. Does it work for you?

anholt:
I'm on my way to Sydney for LCA, and don't have a Pi with me.  I wouldn't expect my merge to have changed anything in fkms, since it doesn't touch that file.  Something probably just didn't get brought over in the rebase.

popcornmix:
I've merged this for further testing - included in latest BRANCH=next rpi-update.
The FKMS issue still needs to be resolved, but is not related to this PR.

PR #2347, opened 2018-01-18
Add I2C attached EETI EXC3000 multi touch driver

xdsopl:
The EXC3000 driver is already in the 4.15 kernel but it needs the following trivial patch to work without change on the 4.9 kernel:
```
timer: Prepare to change timer callback argument type
```
I cherry picked the patch and the driver from the 4.15 kernel and it cleanly applied without problems.
Tested successfully on an Artista-IoT with EXC3000 touchscreen.
I am doing this on behalf of Distec GmbH and would be very pleased if we could give their customers an out-of-box experience without them having to patch / recompile the kernel on Raspbian.
Just putting this in config.txt is enough to make an EXC3000 work on the Artista-IoT:
```
dtoverlay=exc3000,interrupt=34
```

Ahmet


pelwell:
We do accept back-ports from newer kernels, but we'd like the first line of each commit message (after the subject) to say (e.g.):
```
commit 686fef928bba6be13cabe639f154af7d72b63120 upstream.

Modern kernel callback systems pass the structure associated with a
...
```
I also notice that "timer: Prepare to change..." has a followup "Fixes:" patch (52f737c2da40259ac9962170ce608b6fb1b55ee4), so you should include that as well, or explain why it isn't necessary.

Otherwise, nice job on the overlay and associated build files.

xdsopl:
> I also notice that "timer: Prepare to change..." has a followup "Fixes:" patch (52f737c), so you should include that as well, or explain why it isn't necessary.

Okay, that's a fair question and I already proved to myself that this one commit was indeed the only thing needed and did not cause any funny behavior.

Before the timer api change we had the following situation:

```
#define setup_timer(timer, fn, data) \
	__setup_timer((timer), (fn), (data), 0)

static void exc3000_timer(unsigned long data)
{
	struct input_dev *input = (void *)data;
...

static int exc3000_probe(struct i2c_client *client,
			 const struct i2c_device_id *id)
{
...
	setup_timer(&data->timer, exc3000_timer, (unsigned long)input);
...
```

And to prepare for the timer api change we needed some helpers:

```
#define TIMER_DATA_TYPE unsigned long
#define TIMER_FUNC_TYPE void (*)(TIMER_DATA_TYPE)

static inline void timer_setup(struct timer_list *timer,
			       void (*callback)(struct timer_list *),
			       unsigned int flags)
{
	__setup_timer(timer, (TIMER_FUNC_TYPE)callback,
		      (TIMER_DATA_TYPE)timer, flags);
}

#define container_of(ptr, type, member) ({			\
	const typeof( ((type *)0)->member ) *__mptr = (ptr);	\
	(type *)( (char *)__mptr - offsetof(type,member) );})

#define from_timer(var, callback_timer, timer_fieldname) \
	container_of(callback_timer, typeof(*var), timer_fieldname)

static void exc3000_timer(struct timer_list *t)
{
	struct exc3000_data *data = from_timer(data, t, timer);
...

static int exc3000_probe(struct i2c_client *client,
			 const struct i2c_device_id *id)
{
...
timer_setup(&data->timer, exc3000_timer, 0);
...
```
So as you can see, besides some pointer magic, there is nothing special happening and it doesn't change the previous behavior as far as I can see it.

I added the upstream commit hashes you asked for and forced an update.

Ahmet

xdsopl:
Damn, I think you're right for being concerned.
There is another difference: The new one is inline.
Did not see any problems in my testings though but if we want the followup patch to apply cleanly, we need a couple of more patches before ..
I could either change my driver to use the old timer API or I could change the timer API preparation patch to use a macro instead of an inline. Suggestions?

Ahmet

pelwell:
I would prefer you to use the old timer API because that is guaranteed not to affect any other driver.

xdsopl:
Okay, I removed the preparation patch and changed my driver to use the old timer API.
Also added the following lines just below the subject to the driver:
```
commit 7e577a17f2eefeef32f1106ebf91e7cd143ba654 upstream.
beware: code adapted to use the old timer API.
```
Ahmet

pelwell:
Thanks.

PR #2356, opened 2018-01-25
[4.15] defconfigs: enable RC_CORE, disable LIRC_RPI

HiassofT:
Since commit 5573d124292a01afaecb7612773c21321ffcd6d2
RC_CORE defaults to n. So we now have to explicitly set it to y
in the defconfigs to get the same config as with kernel 4.14

Also disable LIRC_RPI and the then no longer needed LIRC_STAGING
since lirc_rpi no longer compiles.

Not sure if it's worth to fix lirc_rpi in 4.15, lirc devices
are already removed in linux-next and gpio-ir-recv and
gpio-ir-tx/pwm-ir-tx exist as upstream alternatives.

ping @popcornmix @pelwell


pelwell:
I'm happy to merge this while we decide what to do with LIRC. @popcornmix?

popcornmix:
I believe @HiassofT has been suggesting users switch from dtoverlay=lirc-rpi to gpio-ir for a while and that seems to be working fine, so no problem with phasing out lirc-rpi.

If there are problems with this we'll send them @HiassofT's way!

PR #2361, opened 2018-01-30
rpi 4.14.y: fkms support

anholt:
The forward port of the fkms code missed that enable/disable_vblank moved to the CRTC funcs, so we had fkms code in vc4_crtc.c that never got called because our enable/disable were entirely missing.  Also brings forward the T tiling patch that got dropped.

### DIFF
--- a/drivers/gpu/drm/vc4/vc4_crtc.c
+++ b/drivers/gpu/drm/vc4/vc4_crtc.c
@@ -685,14 +685,7 @@ static void vc4_crtc_atomic_flush(struct drm_crtc *crtc,
 
 static int vc4_enable_vblank(struct drm_crtc *crtc)
 {
-	struct drm_device *dev = crtc->dev;
-	struct vc4_dev *vc4 = to_vc4_dev(dev);
 	struct vc4_crtc *vc4_crtc = to_vc4_crtc(crtc);
-
-	if (vc4->firmware_kms) {
-		/* XXX: Can we mask the SMI interrupt? */
-		return 0;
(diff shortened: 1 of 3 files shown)


pelwell:
With these commits I do now get a display, but there are still a few problems:
1. The fkms driver has terrible mouse performance - there's far too much buffering in the system.
2. It still gets the warning at line 1550 in drm_handle_vblank.
3. Both versions of the driver (or the overlay?) kill `sudo vcdbg log msg`, causing it to print a slew of:
```
debug_sym: vc_mem_copy: Failed to get memory size via ioctl: Device or resource busy(16)

debug_sym: vc_mem_copy: Failed to get memory size via ioctl: Device or resource busy(16)
```

popcornmix:
I've tried too. 3D performance is fine - neverball runs smoothly with fkms.
But the mouse is unusable - one to two seconds of lag on it.
This doesn't occur on 4.9 kernel. Same bad behaviour on both 4.14 and 4.15.

anholt:
mouse lag fixed, even without the HID throttling (~10% CPU in Xorg when moving the mouse quickly).  no idea on vcdbg log, which I think should be unrelated to this series..

pelwell:
If the rest is OK them I'm happy to shelve the vcdbg issue pending further investigation.

popcornmix:
I think the vcdbg issue is related to using dma from the fb driver to read the top 16M of sdram (as arm cannot see it). I think that gets disabled when fkms is enabled.

PR #2364, opened 2018-02-02
Revert downstream wm8804 changes

HiassofT:
It looks like the changes to the wm8804 codec that we have here in the rpi tree come from a time where the i2s driver didn't support mmap and hacking in 32bit support (which isn't supported by the codec and breaks 24-bit support) was used as a workaround. Not exactly sure why idle_bias_off was changed, but that smells rather suspicously like a workaround, too.

I've successfully tested the Hifiberry Digi Pro, Justboom Digi and Cirrus Logic Audio Card with the local changes reverted.

ping @hifiberry @shawaj could you test with this PR as well?

### DIFF
--- a/sound/soc/codecs/wm8804.c
+++ b/sound/soc/codecs/wm8804.c
@@ -304,7 +304,6 @@ static int wm8804_hw_params(struct snd_pcm_substream *substream,
 		blen = 0x1;
 		break;
 	case 24:
-	case 32:
 		blen = 0x2;
 		break;
 	default:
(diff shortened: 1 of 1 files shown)


shawaj:
@francesco-vannini any chance you could test this? Or @mmilann ?

francesco-vannini:
not until I am back I'm afraid but I can mid next week if that's OK

popcornmix:
@HiassofT we're certainly keen to drop downstream patches if they are no longer required.
Lets plan to merge this in a week if no objections.

HiassofT:
rebased onto current rpi-4.14.y tree and added Signed-off-by

francesco-vannini:
@HiassofT to test this could we do it in the same way as #2365 ?

HiassofT:
@francesco-vannini sure, here's a precompiled RPi2 kernel:
http://www.horus.com/~hias/tmp/rpi/rpi2-4.14-revert-wm8804.tar.gz

pelwell:
Is there any feedback on this PR?

HiassofT:
@francesco-vannini commented here https://github.com/raspberrypi/linux/pull/2365#issuecomment-364189644 that this PR seems to be fine
> Tested a few samples with various samplerate and bitspersample and all seems fine for both #2365 and #2364

francesco-vannini:
Sorry @pelwell , in the excitement of having actually been able to do the compilation and all I forgot to update you with a comment here ;)

Issue #1095, opened 2015-08-04
8192cu WLAN breaks internet connection often, seemingly related to rpi-update to kernel 4.0.


ThomasKorimort:
Hi!
Since some time my WLAN connection which used to work properly apart from speed with EDIMAX WIFI dongle with RTL8188CUS chipset using RPI 2 B firmware driver 8192cu breaks too often to be useful now. I think this is maybe due to a firmware update. There was warning that the update could make some drivers not work anymore (kernel 4.0). Is this such a case? Further: Why do you provide an update, if some drivers don't work anymore afterwards?
I hope this issue can be solved nicely soon.

Many greetings, Thomas Korimort.

ThomasKorimort:
Update: Afte updating another time with Hexxeh script i have looked and saw that suddenly for whatever reaon uname lists again 3.18 trunk of RPi2. I don't know why this is. But still the internet connection problem resides. It seems strange to me as according to the Hexxeh script there should be 4.1 kernel but there is old kernel. For starting i used an image of DEbian 8.0 Jessie not RaspBian. Maybe this is why this Hexxeh script does not work properly? I used the image at https://images.collabora.co.uk/rpi2/ and upgraded Debian Jessie 8.0 from that trunk.

popcornmix:
Are you using NOOBS. How full is your sdcard?
Can you post output of rpi-update.
Note: you can force an rpi-update by
`sudo rm /boot/.firmware_revision && sudo rpi-update`

ThomasKorimort:
I am not using NOOBS. When i installed my RPi2B some months ago, NOOBS was not supporting Debian 8.0 Jessie. Thus i had to use the above-mentioned Image which is a custom-compiled bootstrap image from some knowledgeable person who knows how to bootstrap a running Debian 8.0 Linux system on the RPi2B.

I followed your procedure of deleting .Firmware_revision Prior to executing rpi-update, but still the 3.8.15 trunk is listed by uname -v. Maybe this is because this Image puts the files into a subfolder instead of the Standard one?

My SD Card is 128 GB and after installing the Image of 5 GB Partition size i enlarged the Partition to the full size of 128 GB. What i would like to have is Debian Jessie for my RPi2B. It was suggested that one could use the Debian HardFloat Standard distro for the RPi2B (ARMv7 processor), but i was not clever enough to find out how the Basic Image is put onto the SD Card.

NOOBS i don't want to use as Long as it is not supporting Debian Jessie.

[...]

PR #2365, opened 2018-02-02
properly fix S24_LE mode with pcm512x codec

HiassofT:
This reverts the pcm512x downstream changes introduced in commit 185ea05465aac8bf02a0d2b2f4289d42c72870b7 / PR #1152

The downstream pcm512x changes caused a regression, it broke normal use of the 24bit format with the codec, eg when using simple-audio-card.

The actual bug with 24bit playback is the incorrect usage of physical_width in various drivers in the downstream tree which causes 24bit data to be transmitted with 32 clock cycles. So it's not the pcm512x that needs fixing, it's the soundcard drivers.

I had a go at fixing all the pcm512x based soundcard drivers. Actually using the codec driver as it was intended to be used simplified them quite a bit as I could remove all those set_bclk_ratio calls (and in some cases hw_params / ops as all they did was call set_bclk_ratio).

I've successfully tested 16, 24 and 32bit playback with the IQaudio DAC+ and JustBoom DAC cards. But as I don't have any of the other pcm512x based cards here some more testing and feedback would be welcome - not sure if I got all the soundcard drivers right.

@hifiberry could you check if this works with your DAC+ and DAC pro cards? The non-pro case in your hifiberry_dacplus driver looked rather odd to me - setting bclk ratio to a fixed value of 64 should have broken 16-bit playback IMHO. The allo-boss driver which seems to be a clone of the dacplus driver uses 2 * physical_width in the non-master case (which would be your non-pro case) - which looks more sensible to me.

@iqaudio @shawaj @babuenir @clivem @allocom @miquel83blauw @jgrulich it would be great if you could test this with your pcm512x based cards as well.

### DIFF
--- a/sound/soc/bcm/allo-boss-dac.c
+++ b/sound/soc/bcm/allo-boss-dac.c
@@ -222,14 +222,6 @@ static int snd_allo_boss_update_rate_den(
 	return 0;
 }
 
-static int snd_allo_boss_set_bclk_ratio_pro(
-	struct snd_soc_dai *cpu_dai, struct snd_pcm_hw_params *params)
-{
-	int bratio = snd_pcm_format_physical_width(params_format(params))
-		* params_channels(params);
-	return snd_soc_dai_set_bclk_ratio(cpu_dai, bratio);
-}
-
 static void snd_allo_boss_gpio_mute(struct snd_soc_card *card)
 {
 	if (mute_gpio)
@@ -281,9 +273,6 @@ static int snd_allo_boss_hw_params(
 {
 	int ret = 0;
 	struct snd_soc_pcm_runtime *rtd = substream->private_data;
-	struct snd_soc_dai *cpu_dai = rtd->cpu_dai;
-	unsigned int sample_bits =
-		snd_pcm_format_physical_width(params_format(params));
 
 	if (snd_soc_allo_boss_master) {
 		struct snd_soc_codec *codec = rtd->codec;
@@ -291,13 +280,8 @@ static int snd_allo_boss_hw_params(
 		snd_allo_boss_set_sclk(codec,
 			params_rate(params));
 
-		ret = snd_allo_boss_set_bclk_ratio_pro(cpu_dai,
-			params);
-		if (!ret)
-			ret = snd_allo_boss_update_rate_den(
-				substream, params);
-	} else {
-		ret = snd_soc_dai_set_bclk_ratio(cpu_dai, sample_bits * 2);
+		ret = snd_allo_boss_update_rate_den(
+			substream, params);
 	}
 	return ret;
 }
--- a/sound/soc/bcm/allo-piano-dac-plus.c
+++ b/sound/soc/bcm/allo-piano-dac-plus.c
@@ -795,9 +795,6 @@ static int snd_allo_piano_dac_hw_params(
 		struct snd_pcm_hw_params *params)
 {
 	struct snd_soc_pcm_runtime *rtd = substream->private_data;
-	struct snd_soc_dai *cpu_dai = rtd->cpu_dai;
-	unsigned int sample_bits =
-		snd_pcm_format_physical_width(params_format(params));
 	unsigned int rate = params_rate(params);
 	struct snd_soc_card *card = rtd->card;
 	struct glb_pool *glb_ptr = card->drvdata;
@@ -838,8 +835,6 @@ static int snd_allo_piano_dac_hw_params(
 						glb_ptr->set_lowpass);
 	if (ret < 0)
 		return ret;
-
-	ret = snd_soc_dai_set_bclk_ratio(cpu_dai, sample_bits * 2);
 
 	return ret;
 }
--- a/sound/soc/bcm/allo-piano-dac.c
+++ b/sound/soc/bcm/allo-piano-dac.c
@@ -42,23 +42,6 @@ static int snd_allo_piano_dac_init(struct snd_soc_pcm_runtime *rtd)
 	return 0;
 }
 
-static int snd_allo_piano_dac_hw_params(
-	struct snd_pcm_substream *substream, struct snd_pcm_hw_params *params)
-{
-	struct snd_soc_pcm_runtime *rtd = substream->private_data;
-	struct snd_soc_dai *cpu_dai = rtd->cpu_dai;
-
-	unsigned int sample_bits =
-		snd_pcm_format_physical_width(params_format(params));
-
-	return snd_soc_dai_set_bclk_ratio(cpu_dai, sample_bits * 2);
-}
-
-/* machine stream operations */
-static struct snd_soc_ops snd_allo_piano_dac_ops = {
-	.hw_params = snd_allo_piano_dac_hw_params,
-};
-
 static struct snd_soc_dai_link snd_allo_piano_dac_dai[] = {
 {
 	.name		= "Piano DAC",
@@ -70,7 +53,6 @@ static struct snd_soc_dai_link snd_allo_piano_dac_dai[] = {
 	.dai_fmt	= SND_SOC_DAIFMT_I2S |
 			  SND_SOC_DAIFMT_NB_NF |
 			  SND_SOC_DAIFMT_CBS_CFS,
-	.ops		= &snd_allo_piano_dac_ops,
 	.init		= snd_allo_piano_dac_init,
 },
 };
--- a/sound/soc/bcm/dionaudio_loco-v2.c
+++ b/sound/soc/bcm/dionaudio_loco-v2.c
@@ -41,24 +41,6 @@ static int snd_rpi_dionaudio_loco_v2_init(struct snd_soc_pcm_runtime *rtd)
 	return 0;
 }
 
-static int snd_rpi_dionaudio_loco_v2_hw_params(
-	struct snd_pcm_substream *substream,
-	struct snd_pcm_hw_params *params)
-{
-	struct snd_soc_pcm_runtime *rtd = substream->private_data;
-	struct snd_soc_dai *cpu_dai = rtd->cpu_dai;
-
-	unsigned int sample_bits =
-		snd_pcm_format_physical_width(params_format(params));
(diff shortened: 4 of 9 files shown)


shawaj:
Will do. Copying in @francesco-vannini and @mmilann too

popcornmix:
Looks good to me.

HiassofT:
rebased onto current rpi-4.14.y tree and fixed commit messages (S24_LE format instead of mode)

francesco-vannini:
Hi @HiassofT does this mean that now samples with 16/24/32 can all be played without resampling? Also what do I need to do to test this particular commit on my Raspbian?

HiassofT:
24bit data wasn't actually resampled but padded to 32bit by the I2S controller.

I did some tests and it looks like this padding doesn't simply add zero-bits or repeat the LSB but the MSB of the next channel data is put on the I2S data line. In other words, the samples were padded with some semi-random, fixed value. Although that's more or less an academic issue it's still odd to have it and purists may prefer that their 24bit data is transmitted unmodified as exactly 24bit to the DAC - like it was implemented in the upstream driver.

This is also fixed by this PR, so simply speaking it makes sure all of your 16/24/32 bit data gets transmitted without any "conversion" as 16/24/32bit data to the DAC.

As for testing: it'd be easiest if you just download a tarball of my PR branch via this URL: https://github.com/HiassofT/rpi-linux/archive/4.14-pcm512x-24bit-fix.tar.gz and then compile a kernel from it as usual. This branch is based on the current 4.14 rpi branch, so basically you'll get a `BRANCH=next rpi-update` kernel plus the pcm512x 24bit fix.

francesco-vannini:
@HiassofT thanks for the explanation, I never realised this was the case.

I wouldn't want to be too much of a needy newbie but I don't do kernel compilation so regularly (once every three years maybe :) would you happen to have a couple of notes or a link on how to go about it?

Thanks a lot for this

pelwell:
Allow me: https://www.raspberrypi.org/documentation/linux/kernel/building.md

francesco-vannini:
The hugely famous RTFM or Google it up right? :) Thanks I had found that but wasn't sure. Keep you posted, get everyone off the internet in the meantime I might come up with a new deadly virus by mistake :))

HiassofT:
Ah, Phil was quicker than me :)

To make testing even easier I've uploaded a precompiled RPi2/3 kernel here: http://www.horus.com/~hias/tmp/rpi/rpi2-4.14-pcm512x-24bit-fix.tar.gz

Just download it and then extract it with
```
sudo tar zxf rpi2-4.14-pcm512x-24bit-fix.tar.gz -C /
```

francesco-vannini:
The internet is a safer place, thanks guys! This includes even #2364 ? It doesn't just realised.

miquel83blauw:
Tested it here with the LOCO-V2, everything works flawless!

pelwell:
Thanks. I'll merge this PR once I have a few more confirmations, or 16th February if there are no objections - two weeks is long enough.

iqaudio:
Phil, I will not get a chance to try but as the code was tested using our boards I will have to assume it’s fine. Sorry.

Gordon

francesco-vannini:
Tested a few samples with various samplerate and bitspersample and all seems fine for both #2365 and #2364 
Don't have a good enough DAC to handle 192KHz 24bit samples but at least aplay and omxplayer don't seem to complain.
I even took it upon myself to give it a go with a cross compilation of the branch just to see if I manage :)
Many thanks for your help guys.

francesco-vannini:
Guys I have tried the cross compilation, all seemed fine but when I run lsmod it returns an empty list.
These are the steps I followed:
```wget https://github.com/HiassofT/rpi-linux/archive/4.14-pcm512x-24bit-fix.tar.gz
tar xfvz 4.14-pcm512x-24bit-fix.tar.gz
git clone https://github.com/raspberrypi/tools ~/tools
echo PATH=\$PATH:~/tools/arm-bcm2708/gcc-linaro-arm-linux-gnueabihf-raspbian-x64/bin >> ~/.bashrc
source ~/.bashrc

cd linux
KERNEL=kernel7
make ARCH=arm CROSS_COMPILE=arm-linux-gnueabihf- bcm2709_defconfig
make ARCH=arm CROSS_COMPILE=arm-linux-gnueabihf- zImage modules dtbs
```

then moved the files to the Pi. Am I very far?

pelwell:
Assuming that you have copied the files to the right place, you are just missing the step that builds the aliases database (and a few other things) that allows the kernel to locate a module by compatible string, USB ID etc. You get this for free if you use the modules_install feature of the kernel Makefile, otherwise you must run `depmod` manually.

Unless you are running the same kernel version as the modules you want to re-index, which you aren't, you need to provide the kernel version number, for example:
```
sudo depmod 4.14.17-v7+
```
If the modules are not in `/lib/modules/4.14.17-v7+`, for example if you are preparing a card other than the one you are running from, then you'll also need to specify the base directory:
```
sudo depmod -b /mnt 4.14.17-v7+
```
(assuming the target filesystem is mounted on `/mnt`) which will index the modules in `/mnt/lib/modules/4.14.17-v7+`.

The equivalent command using the kernel Makefiles to also copy the modules would be:
```
sudo make ARCH=arm CROSS_COMPILE=arm-linux-gnueabihf- INSTALL_MOD_PATH=/mnt modules_install
```

francesco-vannini:
@pelwell ... several hours later ... it actually worked! Thanks, I've learnt a bit more today!

pelwell:
In the absence of any negative feedback I'm going to merge this PR. The bleeding-edge LibreELEC community will probably give it a good testing, and we can always revert or patch if there are problems.

adiadore:
hi guys, i've got an issue with S24_LE mode with my hifiberry dac+ pro
small buffering after 3-5 minutes and then followed by sound dropouts with video stuttering
i provided more infos [log](https://forum.kodi.tv/showthread.php?tid=298461&pid=2706220#pid2706220) and [dmesg ++](https://forum.kodi.tv/showthread.php?tid=298461&pid=2706278#pid2706278)
@hifiberry

HiassofT:
It looks like the sound stuttering issue in Kodi is caused by the Hifiberry DAC+ Pro not having suitable oscillators for 24bit mode.

In master mode the clock is derived  directly from the oscillator and with the 24.576MHz oscillator 48kHz 24bit would need a divisor of 10.666. As only integer divisors are possible the driver uses 10 and the effective rate is 51.2kHz instead of 48kHz. So audio runs too fast and Kodi can't sync that with video.

A simple fix for this is to disable 24bit mode, Kodi (and other programs) will then automatically use 32bit mode instead. This is basically the same as before the merge (24bit was transmitted with 32 clock cycles), only difference is that now userspace also knows about it.

I've prepared the changes and asked @MilhouseVH to include them in tomorrow's testbuild. @adiadore please test and report back if it works, then I'll PR the changes to the RPi kernel tree.

Preliminary (completely untested) changes are here if someone else wants to test: https://github.com/HiassofT/rpi-linux/commits/4.14-fix-pcm512x-master

Note: Allo Boss in master mode is quite certainly effected by this issue, too (it uses 49.152/45.158MHz oscillators), the other pcm512x based sound cards run pcm512x in slave mode and are fine.

hifiberry:
@HiassofT Sorry I wasn't able to test this in the last weeks. 
Is there no way to use the S24LE format with 32bit with per channel? Even padding with random data would be ok as they are not really relevant as they will be outside of the dynamic range of the DAC.

pelwell:
Unless we can find a fix for 24-bit data on the Hifiberry DAC+ Pro I'll have to revert this PR.

HiassofT:
@hifiberry could you please test this tree: https://github.com/HiassofT/rpi-linux/tree/4.14-pcm512x-set-tdm-slot

I've implemented .set_tdm_slot in the pcm512x driver so sound card drivers can optionally configure the number of bclk cycles. I also changed the hifiberry_dacplus and allo-boss-dac drivers to use this, so 16, 24 and 32bit should work as before.

@adiadore I'll ask milhouse to include this change in his next build instead of the previous one

hifiberry:
@HiassofT : Thank you very much. I'll have a look, but it might take some days to find time for it. Im just in the process of moving the workplace and it's a bit chaotic at the moment.

PeterPablo:
I can second that the firmware https://github.com/raspberrypi/firmware/commit/9ab68bf218ffc84423a2591658260676d373259a causes issues during playback (radio via kodi) on a raspbian. The previous firmware (https://github.com/Hexxeh/rpi-firmware/commit/98b90b75aa4352c148f2726c84d6f466534a89fc) works fine. I also have a HifiBerry DAC+ Pro. I can test the next firmware being available via `rpi-update`.

pelwell:
We've had some confirmation from @adiadore via the LibreElec forum that these patches fix the issue for him, so they've been cherry-picked into rpi-4.14.y to hit tonight's firmware build.

HiassofT:
@pelwell thanks, I was just about to create a PR for the fix :)

PeterPablo:
I confirm that on https://github.com/Hexxeh/rpi-firmware/commit/ab90bf1971e4b6cd8c75d6ff7c20bf08166b613e the issue is fixed: `Linux rpi2 4.14.21-v7+ #1095 SMP Fri Feb 23 18:21:29 GMT 2018 armv7l GNU/Linux`
Thank you!

pelwell:
Cool.

I've just noticed that I referred to @adiadore as "him" with no evidence to support that assumption - apologies if that is not the case.

PeterPablo:
For what it's worth: adiadore commented in the LibreELEC thread (https://forum.kodi.tv/showthread.php?tid=298461&pid=2706880#pid2706880) that his issue is resolved with the patch of @HiassofT (https://raw.githubusercontent.com/HiassofT/LibreELEC.tv/aa2746c76992c60087706c1bacd0073fd8d24f42/projects/RPi/patches/linux/linux-10-fix-pcm512x-master-mode.patch) applied. So all good it seems, verified by two people independently.

Issue #2366, opened 2018-02-04
Minimal kernel config created from bcm2709_defconfig using allnoconfig fails to compile


simonvanderveldt:
When trying to build a minimal kernel from `bcm2709_defconfig` using `allnoconfig` the resulting configuration doesn't compile.

[`allnoconfig`](https://github.com/raspberrypi/linux/blob/rpi-4.9.y/Documentation/kbuild/kconfig.txt#L75) is the normal way to configure/build the most minimal working kernel. It's used in combination with a defconfig to provide the required defaults.
It can be executed in the following way to generate a minimal config based on `bcm2709_defconfig`:
```
make ARCH=arm KCONFIG_ALLCONFIG=arch/arm/configs/bcm2709_defconfig allnoconfig
```
Now when compiling the generated `.config` I ran into several of these errors:
```
/tmp/ccholjfv.s: Assembler messages:
/tmp/ccholjfv.s:67: Error: selected processor does not support requested special purpose register -- `mrs r3,cpsr'
/tmp/ccholjfv.s:75: Error: selected processor does not support requested special purpose register -- `msr cpsr_c,r3'
```

Doing a `diffconfig` of the configs generated by `bcm2709_defconfig` and `allnoconfig` showed that `ARCH_BCM2835` was missing. Even though [`ARCH_BCM2835` is defined in `bcm2709_defconfig`](https://github.com/raspberrypi/linux/blob/a29fe9d2366b1700bf1015044e561108a5fa29df/arch/arm/configs/bcm2709_defconfig#L41) it doesn't actually get activated because it's prerequisites are not met.

I was able to fix these issues by adding the following two config options to `bcm2709_defconfig`:
```
CONFIG_MMU=y
CONFIG_ARCH_MULTI_V7=y
```
Shall I create a PR to add them? Or is there a specific reason these two config option aren't included in `bcm2709_defconfig`?

Once that was fixed I ran into the following issue, for which I've not yet figured out how to fix it. Any and all suggestions are welcome :)
```
drivers/mmc/host/bcm2835-sdhost.c: In function 'bcm2835_sdhost_dma_complete':
drivers/mmc/host/bcm2835-sdhost.c:541:10: error: implicit declaration of function 'kmap_atomic'; did you mean 'in_atomic'? [-Werror=implicit-function-declaration]
   page = kmap_atomic(host->drain_page);
          ^~~~~~~~~~~
          in_atomic
[...]

PR #2369, opened 2018-02-06
Add SND_RAWMIDI Kconfig selects dep to SND_PISOUND

simonvanderveldt:
Fixes part of #2366


pelwell:
Can you add a slightly-less-minimal commit message? Even just:
```
See: https://github.com/raspberrypi/linux/issues/2366
```
would be better. Include a "Signed-off-by: " line as well.

simonvanderveldt:
@pelwell Sure :) I've amended the commit.

simonvanderveldt:
@pelwell Thanks!
Small question: Do the changes in the 4.9.x branch get pulled into the other branches automatically?

pelwell:
No, but we try not to let them escape.

simonvanderveldt:
@pelwell Ok, thanks for the info! I'll keep an eye on them then :)

PR #2370, opened 2018-02-06
config: enable Audio Graph Card module

HiassofT:
Audio Graph Card is the modern, OF-graph based, successor of Simple-Card.

Besides the somewhat more readable and structured devicetree bindings one advantage for users is that audio graph card allows for automatically enabling an external amplifier via a GPIO whenever the card is used.

An example overlay with the pa-gpios property and the necessary DAPM routes is available here: https://gist.github.com/HiassofT/19546a6fa9bd146652211405438e3ff0

As the core functions (DT parsing etc, in simple-card-utils module) are shared between simple-card and audio-graph-card enabling audio graph card adds only a ~12k module to the trees.


pelwell:
Out of curiosity, apart from the new endpoint mechanism is `pa-gpio` the only new feature?

pelwell:
The bindings document seems to say that there are no other features, but it does look more flexible and I can see why some people might prefer it.

HiassofT:
Yes, pa-gpios is the only new feature. In 4.14 audio-graph-card doesn't support the mclk-fs property, that was added in 4.15, but that's no big deal as IIRC RPi doesn't support that anyways.

Upstream encourages people to use audio graph card for new implementations and I have to add that I personally prefer the port/endpoint design over the old one.

BTW: Thanks a lot for merging!

PR #2378, opened 2018-02-13
Fix added to solve CPU stall while calling  set frontend of Hauppauge WinTV-HVR-950Q dvb tuner.

Haseenamol:
This issue is occurring if we do the section filtering and tuning in parallel.
After changing the transport stopping sequence, issue is not happening.

Signed-off-by: Haseenamol <haseenamol@tataelxsi.co.in>

### DIFF
--- a/drivers/media/usb/au0828/au0828-dvb.c
+++ b/drivers/media/usb/au0828/au0828-dvb.c
@@ -353,8 +353,6 @@ static int au0828_set_frontend(struct dvb_frontend *fe)
 	mutex_lock(&dvb->lock);
 	was_streaming = dev->urb_streaming;
 	if (was_streaming) {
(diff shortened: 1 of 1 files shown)


popcornmix:
As this is an untouched upstream linux file, we won't be merging it here.
Please submit this upstream: https://www.kernel.org/doc/html/v4.14/process/submitting-patches.html

If it is accepted upstream, then we're happy to cherry-pick it early to get it into our kernel builds. But we are not the right people to peer review the fix.

Haseenamol:
Thank you for your comments. Will try to submit this in upstream

popcornmix:
I'll close this here. Let us know if there's progress in getting this accepted upstream and we'll try to get it into our kernel tree.

Issue #2363, opened 2018-02-02
Possible FIQ high speed isochronous transfer issue


antonblanchard:
I have two USB webcams (a Logitech  C270 and a Logitech QuickCam Pro 9000) and neither of them work on my Raspberry Pi 3 running either the 4.9 or 4.14 kernels.

I start up guvciew and the application prints an endless stream of:

```
V4L2_CORE: (get_v4l2_frame) video stream must be started first
V4L2_CORE: (get_v4l2_frame) video stream must be started first
V4L2_CORE: (get_v4l2_frame) video stream must be started first
V4L2_CORE: (get_v4l2_frame) video stream must be started first
```

and in the dmesg:

```
[  184.869874] usb 1-1.2: reset high-speed USB device number 4 using dwc_otg
[  187.657763] uvcvideo: Failed to set UVC probe control : -71 (exp. 26).
[  187.658183] uvcvideo: Failed to set UVC commit control : -71 (exp. 26).
```

If I kill and restart guvcview, both the webcam and the keyboard get reset:

```
[  208.147954] usb 1-1.2: USB disconnect, device number 4
[  208.247784] usb 1-1.2: new full-speed USB device number 8 using dwc_otg
[  208.347737] usb 1-1.2: device descriptor read/64, error -32
[  208.567739] usb 1-1.2: device descriptor read/64, error -32
[  208.787697] usb 1-1.2: new full-speed USB device number 9 using dwc_otg
[  208.887672] usb 1-1.2: device descriptor read/64, error -32
[  209.107689] usb 1-1.2: device descriptor read/64, error -32
[  209.227731] usb 1-1-port2: attempt power cycle
[  209.887624] usb 1-1.2: new high-speed USB device number 10 using dwc_otg
[  210.135972] usb 1-1.2: New USB device found, idVendor=046d, idProduct=0825
[  210.135983] usb 1-1.2: New USB device strings: Mfr=0, Product=0, SerialNumber=2
[  210.135988] usb 1-1.2: SerialNumber: B834B130
[  210.136520] uvcvideo: Found UVC 1.00 device <unnamed> (046d:0825)
[  210.226177] uvcvideo 1-1.2:1.0: Entity type for entity Extension 4 was not initialized!
[  210.226190] uvcvideo 1-1.2:1.0: Entity type for entity Extension 6 was not initialized!
[  210.226196] uvcvideo 1-1.2:1.0: Entity type for entity Extension 7 was not initialized!
[...]

PR #2380, opened 2018-02-13
dwc_otg: Fix a regression when dequeueing isochronous transfers

P33M:
In 282bed95 (dwc_otg: make nak_holdoff work as intended with empty queues)
the dequeue mechanism was changed to leave FIQ-enabled transfers to run
to completion - to avoid leaving hub TT buffers with stale packets lying
around.

This broke FIQ-accelerated isochronous transfers, as this then meant that
dozens of transfers were performed after the dequeue function returned.

Restore the state machine fence for isochronous transfers.


P33M:
c.f. https://github.com/raspberrypi/linux/issues/2363

PR #2386, opened 2018-02-16
Add support for SuperAudioBoard sound card

whollender:
Setup support for [SuperAudioBoard](https://hackaday.io/project/5912-teensy-super-audio-board) sound card.

Based on simple-audio-card generic driver on suggestion of @pelwell


pelwell:
The overlay and codec driver are in the current rpi-update release.

PR #2388, opened 2018-02-17
Generic Rotary Encoder overlay for multiple instances

ismailof:
A slightly modified `rotary-encoder-overlay` to allow multiple instances.
It make use of the `<reg>` property, similar to the `gpio-keys-overlay`, bounded to the `pin_a` parameter.

Then, it is possible to attach multiple rotary encoders to different pins, like the following example:
```
dtoverlay=rotary-encoder,pin_a=13,pin_b=19,relative_axis
dtoverlay=rotary-encoder,pin_a=22,pin_b=27,relative_axis,linux_axis=1
```

Parameters `rotary0_pin_a` and `rotary0_pin_b` are renamed to just `pin_a` and `pin_b` since they do not only refer to one rotary node only.

### DIFF
--- a/arch/arm/boot/dts/overlays/README
+++ b/arch/arm/boot/dts/overlays/README
@@ -1321,9 +1321,9 @@ Params: <None>
 Name:   rotary-encoder
 Info:   Overlay for GPIO connected rotary encoder.
 Load:   dtoverlay=rotary-encoder,<param>=<val>
-Params: rotary0_pin_a           GPIO connected to rotary encoder channel A
+Params: pin_a                   GPIO connected to rotary encoder channel A
                                 (default 4).
-        rotary0_pin_b           GPIO connected to rotary encoder channel B
+        pin_b                   GPIO connected to rotary encoder channel B
                                 (default 17).
         relative_axis           register a relative axis rather than an
                                 absolute one. Relative axis will only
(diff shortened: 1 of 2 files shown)


pelwell:
I think what you are doing is great, but I'm not so happy that it may stop working for existing users unless they rename their parameters to match. Can you keep rotary0_pin_a and _b as aliases, and make sure the README matches (I prefer the shorter pin_a and _b)?

ismailof:
Sure! I don't like to break compatibility for previous users either. I'm quite new to dts overlay format so I don't know the best way to achieve this.

My first simple attemp has been using the aliases section, but compiling fails sind `pin_a` and `pin_b` are not labels but parameters
```
        aliases {    /* for compatibilty with previous properties */
            rotary0_pin_a = &pin_a;
            rotary0_pin_b = &pin_b;
        };
```
I can think of duplicating the parameters inside the `__overrides__` section, and even don't modify the `reg` property here to have same behaviour as previous code:
```
                /* legacy parameters*/
                rotary0_pin_a =   <&rotary>,"gpios:4",
                               <&rotary_pins>,"brcm,pins:0",
                rotary0_pin_b =   <&rotary>,"gpios:16",
                               <&rotary_pins>,"brcm,pins:4";

 ```

Is there a better alternative to this?

pelwell:
No, there is no clever way, just duplicate the parameter definition with a different name like your second example.

ismailof:
Ok. Commited this way

pelwell:
The updated overlay is in the current rpi-update release.

PR #2391, opened 2018-02-20
overlays: Add overlay for missing AUX interrupt controller support

agraf:
Upstream Linux today does not support the AUX interrupt controller
yet. To make sure it can use our device tree, add an overlay that
reverts it to something upstream understands again.

See: raspberrypi/firmware#943


pelwell:
Thanks! (I had to tweak the Makefile and add a README entry).

PR #2394, opened 2018-02-22
remove generated file from git

t-8ch:
The commit introduce this file
(22067e51c6609d9abdac7bf80c8da007fb980913) seems not to come from
upstream as it is missing a real name and SOB line required there.


pelwell:
Can you delete scripts/kconfig/zconf.hash.c as well? We spotted this other day in 4.15 but haven't got round to deleting them yet, so I can back-port your patch.

t-8ch:
@pelwell Done

pelwell:
Thanks!

PR #2395, opened 2018-02-23
bcm2709: enable usb gadget functions

t-8ch:
they are already in bcmrpi_defconfig


pelwell:
Thanks - I don't know what happened there.

pelwell:
Yes I do - DWC2 wasn't enabled in bcm2709_defconfig until recently, and when it was added the gadget support was omitted. I've back-ported the fix to 4.14 and 4.15.

t-8ch:
Thanks!

t-8ch:
Is USB OTG actually supposed to work on a RPI 3 currently/at all?

(Without shields, trying with the power port,
the dcw2 overlay is loaded and the udc detected, but the host does not see anything)

lurch:
The micro-USB power port on the RPi3 is power only, no data lines are connected. The RPi3 has a downstream USB hub chip (LAN9514) connected to the SoC's single USB connection, and this forces the USB connection into master mode.

AFAIK the only way to use USB OTG on bcm2709 (aka BCM2837) is via the Compute Module 3.

t-8ch:
@lurch Thanks for the clarification!

PR #2398, opened 2018-02-24
Fixing memset call in pisound.c

gtrainavicius:
The incorrect call was causing a stack check fail kernel panic on Arch Linux. Strangely it was happening only on 4.14 versions, Arch Linux on 4.9 kernel with the same incorrect code was not triggering the panic.

With the fix included, I can confirm that Arch Linux on 4.14 works correctly.

I will open another pull request for the same change to be integrated to 4.14 branch.

### DIFF
--- a/sound/soc/bcm/pisound.c
+++ b/sound/soc/bcm/pisound.c
@@ -313,7 +313,7 @@ static void spi_transfer(const uint8_t *txbuf, uint8_t *rxbuf, int len)
 	struct spi_transfer transfer;
 	struct spi_message msg;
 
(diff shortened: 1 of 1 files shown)


pelwell:
Good catch.

gtrainavicius:
Thank you for quick turnaround :)

gtrainavicius:
Btw, 4.14 kernels on Raspbian weren't hiting the stack check fail, as on Arch Linux, I wonder what's the difference... I did a quick investigation, it looked like both had the same compilation flags for stack protector checks enabled.

PR #2407, opened 2018-03-01
i2c: bcm2835: Set up the rising/falling edge delays

anholt:
We were leaving them in the power on state (or the state the firmware
had set up for some client, if we were taking over from them).  The
boot state was 30 core clocks, when we actually want to sample some
time after (to make sure that the new input bit has actually arrived).

Signed-off-by: Boris Brezillon <boris.brezillon@bootlin.com>
Signed-off-by: Wolfram Sang <wsa@the-dreams.de>
Cc: stable@kernel.org
(cherry picked from commit fe32a815f05c8568669a062587435e15f9345764)


anholt:
This fixed EDID on at least one of Boris's monitors.

pelwell:
Thanks, Eric. I tweaked the commit message to follow the convention for back-ports.

PR #2413, opened 2018-03-04
[4.14] Fix GCC 6.4.0 and 7.3.0 warnings

nathanchance:
I ran into these warnings on the latest 4.14 branch.

```bash
../drivers/net/wireless/realtek/rtl8192cu/core/rtw_security.c: In function ‘aes_cipher’:
../drivers/net/wireless/realtek/rtl8192cu/core/rtw_security.c:1504:5: warning: this ‘for’ clause does not guard... [-Wmisleading-indentation]
     for (j = 0; j < 8; j++)
     ^~~
../drivers/net/wireless/realtek/rtl8192cu/core/rtw_security.c:1507:2: note: ...this statement, but the latter is misleadingly indented as if it is guarded by the ‘for’
  payload_index = hdrlen + 8;
  ^~~~~~~~~~~~~
../drivers/net/wireless/realtek/rtl8192cu/core/rtw_security.c: In function ‘aes_decipher’:
../drivers/net/wireless/realtek/rtl8192cu/core/rtw_security.c:1878:5: warning: this ‘for’ clause does not guard... [-Wmisleading-indentation]
     for (j = 0; j < 8; j++)
     ^~~
../drivers/net/wireless/realtek/rtl8192cu/core/rtw_security.c:1881:2: note: ...this statement, but the latter is misleadingly indented as if it is guarded by the ‘for’
  payload_index = hdrlen + 8;
  ^~~~~~~~~~~~~
../drivers/staging/vc04_services/interface/vchiq_arm/vchiq_arm.c: In function ‘vchiq_open’:
../drivers/staging/vc04_services/interface/vchiq_arm/vchiq_arm.c:1735:7: warning: unused variable ‘ret’ [-Wunused-variable]
   int ret;
       ^~~
```
Pretty simple to fix and now `make > /dev/null` is clean.

PR #2417, opened 2018-03-05
[4.15] Fix outstanding GCC 6.4.0 and GCC 7.3.0 warnings

nathanchance:
Same ones as #2413, with a couple of extra commits ported forward from older branches since it appears 724f9c495e67696a864cb9b6adcc941d8a88849c overrode the fixes from 39f233ba189c487e861e4a7f8569ac8bf3389711 and 801674ee62a963376caec68ed094657431218feb in 8f1edbe6c5ae2ec547baf976450436a32cae24e2. I will look into submitting these upstream so you don't have to worry about constantly carrying them forward.


pelwell:
Rebased and merged offline - thanks.

PR #2418, opened 2018-03-05
[4.16] Fix outstanding GCC 6.4.0 and GCC 7.3.0 warnings

nathanchance:
Same ones as #2413, with a couple of extra commits ported forward from older branches since it appears d3b1ce08060e64592a184375ad953044eb06f5d4 overrode the fixes from 9db27f31d9476c5629b63d3e0619259a68e0544f and 65290084ac865025a134fc50249579f0d4b7e8ce in 2d806f1f8a37cab9516ef2525e3391be06ca601c. I will look into submitting these upstream so you don't have to worry about constantly carrying them forward.


pelwell:
Thank you.

PR #2423, opened 2018-03-07
Update dts Makefile to include overlays directory

davidw5:
Add "subdir-y += overlays" to arch/arm64/boot/dts/Makefile to enable 'make ARCH=arm64 dtbs' to generate overlay .dtbo files


pelwell:
Thanks.

PR #2425, opened 2018-03-08
arm64: enable thermal / enable mmc

Piraty:


pelwell:
I'm happy to merge these provided you understand that you are enabling two drivers for the same MMC interface, and that one is built in while the other is a module.

Piraty:
Thanks for reacting on this.
`CONFIG_MMC_SDHCI_IPROC=m` is known to enable wifi (from [gentoo](https://forums.gentoo.org/viewtopic-p-8188644.html#8188644)) without increasing the sdcard's index.
`CONFIG_MMC_BCM2835_MMC=y` wasn't tested by me, but suggested by another gentoo user (in the sense "while we are PRing"). I can exclude it from this commit in case of doubt.

pelwell:
The rpi version of the Linux repo includes a patch that allows MMC device numbers to be set by Device Tree aliases, so renumbering doesn't have to be an issue.

In the ARCH=arm defconfigs we just use CONFIG_MMC_BCM2835_MMC=y, but if the attached device doesn't contain the rootfs (which is usually the case) then there is no need for it to be built in.

timofonic:
@Piraty @pelwell Any hopes to have 1st class/1st citizen ARMv8 64bit support with all the rpi3 hardware bells & whistles? VC4, SPI, SD, WiFi, GPIO, video decoding, etc? Will that day happen? Please make it happen, please! :D

JamesH65:
Not in the near future - it's an large amount of work including testing for very little benefit (ie 64 bit has only a small benefit on the Pi).

PR #2429, opened 2018-03-09
vc4: DPI support

anholt:
@6by9 had asked about the state of DPI support in vc4-kms-v3d, so I went ahead and pulled the remaining bits for the Adafruit 7" panel out of one of my branches and made an overlay for it.


timofonic:
@anholt Will changes like this be merged to newer kernel branches? Are there a waiting policy? I just wish there would be some kind of more "homogeneity" between different branches. I would love the -rt branch not become a separate branch be part of upstream/vanilla too, a man can dream...

6by9:
Generally any changes made to the current branch will get merged into the later branches.

RT has already been discussed at https://www.raspberrypi.org/forums/viewtopic.php?f=29&t=206750. There are significant issues with USB and FIQ, and doesn't benefit 99% of users, so those have to be resolved before even considering it.

6by9:
And thank you @anholt. I'll check them out next week.

mi-hol:
@6by9 did the checking progress?

6by9:
No, other things took priority following the launch of the 3B+. It's on the list.

anholt:
Now updated with the SAND display support, since the patches made it upstream.

anholt:
This PR still needs to be merged.

sigmaris:
I've been using a custom kernel compiled with this patch for several weeks to successfully get DPI video output with the vc4 driver - just confirming that it works for me, at least.

popcornmix:
@6by9 did you still want to run some tests on this?

6by9:
@popcornmix I was going to, unless you've beaten me to it. PC issues at the moment though :-(

lrusak:
I'd like to see this get merged also :+1: as it fixes a problem I have with drm modifiers

PR #2434, opened 2018-03-13
4.14 Hauppauge DVB drivers

6by9:
This is the combination of the upstream patches from Hauppauge to fully support all their DVB tuners. They're all merged to the git://linuxtv.org/media_tree.git  master branch, so expect to be in mainline in the next release. I've picked up a couple of other minor patches that mean the patches have been cherry-picked without any conflicts or editing required.

Tested with the WinTV-dualHD, and both tuners work happily for DVB-T/T2 using w_scan, and I'll set tvheadend scanning in a moment to check it is really happy with it.


6by9:
@popcornmix @pelwell Any comments on these cherry-picks? They've all been merged upstream, and I have a WinTV dualHD here if you want to verify the results.

pelwell:
We like cherry-picks from upstream to have `commit <hash> upstream.` as the first line after the subject, but otherwise if they've been accepted (as they have) and they compile and work (as they do) then I'm happy to merge them.

6by9:
OK, I'll amend the commit texts as requested. Thanks.

6by9:
Hmm, just checking and they aren't in 4.16-rc7 as they only merged to linuxtv.org/media_tree.git at the start of March.
I'm assuming you're really after a hash from torvalds/linux master, which won't happen until 4.17. I can add a media_tree commit hash though if sufficient.

pelwell:
I think you'll find that once they're in one of the linux trees the hashes remain the same due to the use of merge commits.

6by9:
I'll take your word for it :-)
Editing the patches now.

6by9:
Please ignore changes for a while - trying to get my trees in order :-/

6by9:
OK, done.
I've reworked 94f60e2 as it seemed to be missing other dependencies.
I've built it, and will try to find a moment to test it tonight when I have an aerial feed available.

6by9:
No, something seems squiffy. wscan's detected a signal on 474.167MHz (ch21), but not found any services. That's one of the DVB-T2 muxes on Sandy Heath, so it may be something off there. I'll check with tvheadend later.

6by9:
Very squiffy. tvheadend can't find anything and leaves the device in a mode that then w_scan can't find anything either. It requires a disconnect/reconnect to reset it and let w_scan work.
I'll do a bisect tonight and update.

6by9:
2f2244c is the commit that appears to have an issue. I can't see an obvious difference to upstream though.

6by9:
I've checked with the latest media_build tree on a Pi, and that works happily using dvbv5scan and tvheadend on both adapters and all muxes from Sandy Heath.
I'll have to do a comparison between the branches. I did notice a few more relevant patches floating past on the mailing list.

6by9:
Hopefully found. I'd missed https://github.com/raspberrypi/linux/commit/1844f49861cbd5dd8c289d3c4c8b4a76a4fe2d37#diff-78c5decbfe7d1f796841ff3c9b61d207

```
[mchehab@s-opensource.com: I ended by applying the first version,
 instead of the right one. So, this patch contains the diff and
 the v2 changelog instead]
```
So "media: si2168: Add ts bus coontrol, turn off bus on sleep" would have broken things. 
Recompiling and will test tonight. I've got a rebased branch that will also resolve the conflicts being flagged.

6by9:
Bad news. Whilst both tuners show up, and both will tune independently using dvbv5-scan, trying to use both in tvheadend results in the first one to be used stalling when the second starts up.
Time to give up on this branch and cherry-pick again from scratch.

6by9:
Closing this. I haven't had a chance to look into why there are issues, and the changes should all be merged in 4.19 which we are switching to.

Issue #2437, opened 2018-03-15
Kernel oops adding eth0 to bridge on Pi3+


burtyb:
When adding eth0 to a bridge with a Pi3+ (Raspbian 2018-03-13) I get the following Oops on boot.

> [    5.451470] Unable to handle kernel NULL pointer dereference at virtual address 000001f0
[    5.464412] pgd = b72e4000
[    5.469282] [000001f0] *pgd=372e8835, *pte=00000000, *ppte=00000000
[    5.478303] Internal error: Oops: 17 [#1] SMP ARM
[    5.485414] Modules linked in: bridge stp llc brcmfmac brcmutil cfg80211 rfkill snd_bcm2835 snd_pcm snd_timer snd bcm2835_gpiomem uio_pdrv_genirq uio fixed ip_tables x_tables ipv6
[    5.505733] CPU: 2 PID: 286 Comm: brctl Not tainted 4.9.80-v7+ #1098
[    5.505737] Hardware name: BCM2835
[    5.505745] task: b97ae740 task.stack: b6578000
[    5.505763] PC is at phy_ethtool_gset+0x14/0x88
[    5.505774] LR is at lan78xx_get_settings+0x3c/0xcc
[    5.505783] pc : [<8054a3c8>]    lr : [<80550df8>]    psr: 60000013
[    5.505783] sp : b6579d18  ip : b6579d28  fp : b6579d24
[    5.505787] r10: 00000080  r9 : b9057500  r8 : b6579d54
[    5.505793] r7 : 00000001  r6 : b9050000  r5 : b6579d98  r4 : 00000000
[    5.505800] r3 : 80550dbc  r2 : 00000001  r1 : b6579d54  r0 : 00000000
[    5.505807] Flags: nZCv  IRQs on  FIQs on  Mode SVC_32  ISA ARM  Segment user
[    5.505813] Control: 10c5383d  Table: 372e406a  DAC: 00000055
[    5.505818] Process brctl (pid: 286, stack limit = 0xb6578210)
[    5.505825] Stack: (0xb6579d18 to 0xb657a000)
[    5.505835] 9d00:                                                       b6579d4c b6579d28
[    5.505845] 9d20: 80550df8 8054a3c0 80550dbc b9050000 b6579d98 b716ef80 00000001 80d04850
[    5.505856] 9d40: b6579d94 b6579d50 8063d4ac 80550dc8 b9b39590 00000001 00000000 00000000
[...]

PR #2439, opened 2018-03-16
[4.14] Fix for Pisound's MIDI Input getting blocked for a while in rare cases.

gtrainavicius:
```
There was a possible race condition which could lead to Input's FIFO queue
to be underflown, causing high amount of processing in the worker thread for
some period of time.

Signed-off-by: Giedrius Trainavicius <giedrius@blokas.io>
```

Link to 4.9 pull request: https://github.com/raspberrypi/linux/pull/2440

### DIFF
--- a/sound/soc/bcm/pisound.c
+++ b/sound/soc/bcm/pisound.c
@@ -56,6 +56,12 @@ static const char *pisnd_spi_get_version(void);
 static int pisnd_midi_init(struct snd_card *card);
 static void pisnd_midi_uninit(void);
 
+enum task_e {
+	TASK_PROCESS = 0,
+};
+
+static void pisnd_schedule_process(enum task_e task);
+
 #define PISOUND_LOG_PREFIX "pisound: "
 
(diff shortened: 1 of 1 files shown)


gtrainavicius:
Thank you for quick turnaround! :)

By the way, what is the policy for 4.9 and 4.14 branches? Should I continue pushing fixes to both branches at the moment? For how long will 4.9 continue to be supported?

pelwell:
We could have launched the 3B+ with 4.14 but the decision was taken to avoid creating two sets of new bugs - 4.14 going mainstream and 3B+-specific problems. With the launch out of the way and, so far, only a relatively niche networking bug (#2437), we'll probably make the jump to 4.14 within the next week or two.

pelwell:
4.9 is an LTS kernel, so we'll probably keep it around as the `stable` branch for a while - we're certainly not blocking submissions for it yet.

gtrainavicius:
Thank you for the information! 👍

PR #2455, opened 2018-03-20
Use all seven dwc2 gadget fifos.

ED6E0F17:
Linux 4.9 needed fifos to be set to their default values,
 leaving the last one (silently) set to zero size.
From 4.12 Linux allows fifos to be set to any size EXCEPT zero.

Resolves https://github.com/raspberrypi/linux/issues/2390

### DIFF
--- a/arch/arm/boot/dts/overlays/dwc2-overlay.dts
+++ b/arch/arm/boot/dts/overlays/dwc2-overlay.dts
@@ -15,7 +15,7 @@
 			dr_mode = "otg";
 			g-np-tx-fifo-size = <32>;
 			g-rx-fifo-size = <256>;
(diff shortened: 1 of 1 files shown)


popcornmix:
Seems reasonable. @pelwell ok with you?

pelwell:
By assuming this patch was written by somebody with more knowledge of the subject than me I can accept the fact that 256 + 256 != 768 for a reason.

ED6E0F17:
"256 + 256 != 768"

Yes, the extra memory is available for the rx_fifo, which can be set from config.txt - I do not want to suggest a value for that until someone has done some performance testing with different values.

pelwell:
Thanks for the explanation.

PR #2464, opened 2018-03-24
This commit adds support for RP3-B-Plus in in arch arm64

derpeter:
I hope this is the right place for the pull request. This probably also needs to be done in newer kernel version. Please let me know if i should submit it somewhere else.


pelwell (commented):
The lan7515 file would be better as a symbolic link to the arch/arm tree, like the smsc9514 equivalent, but otherwise this looks good.

derpeter:
@pelwell thx. here is the change to symlink

PR #2483, opened 2018-03-31
[4.16] rpi-cirrus: change codec to component

HiassofT:
Upstream is replacing snd_soc_codec with snd_soc_component. So better fix the driver before it breaks.


popcornmix:
Thanks.

Issue #2458, opened 2018-03-21
4.14.27-v7+ / 3+  VLAN hw csum failure


sinistermidget:
Adding a VLAN to eth0 and then putting any traffic over it results in the following error regularly repeating:
```
[ 1349.736843] eth0.20: hw csum failure
[ 1349.736865] CPU: 0 PID: 0 Comm: swapper/0 Tainted: G         C      4.14.27-v7+ #1100
[ 1349.736870] Hardware name: BCM2835
[ 1349.736904] [<8010fff8>] (unwind_backtrace) from [<8010c260>] (show_stack+0x20/0x24)
[ 1349.736922] [<8010c260>] (show_stack) from [<8076dd84>] (dump_stack+0xd4/0x118)
[ 1349.736943] [<8076dd84>] (dump_stack) from [<8067665c>] (netdev_rx_csum_fault+0x44/0x48)
[ 1349.736962] [<8067665c>] (netdev_rx_csum_fault) from [<80668f7c>] (__skb_checksum_complete+0xb4/0xb8)
[ 1349.736980] [<80668f7c>] (__skb_checksum_complete) from [<806f997c>] (icmp_rcv+0xd0/0x388)
[ 1349.736997] [<806f997c>] (icmp_rcv) from [<806bfaf4>] (ip_local_deliver_finish+0xe4/0x330)
[ 1349.737012] [<806bfaf4>] (ip_local_deliver_finish) from [<806c0360>] (ip_local_deliver+0x54/0xdc)
[ 1349.737026] [<806c0360>] (ip_local_deliver) from [<806bff7c>] (ip_rcv_finish+0x23c/0x494)
[ 1349.737038] [<806bff7c>] (ip_rcv_finish) from [<806c0704>] (ip_rcv+0x31c/0x554)
[ 1349.737053] [<806c0704>] (ip_rcv) from [<80673c24>] (__netif_receive_skb_core+0x340/0xc84)
[ 1349.737068] [<80673c24>] (__netif_receive_skb_core) from [<806767d0>] (__netif_receive_skb+0x20/0x7c)
[ 1349.737083] [<806767d0>] (__netif_receive_skb) from [<806768c4>] (process_backlog+0x98/0x148)
[ 1349.737100] [<806768c4>] (process_backlog) from [<8067aba0>] (net_rx_action+0x2e8/0x45c)
[...]

PR #2487, opened 2018-04-04
net: lan78xx: Allow for VLAN headers in timeout.

6by9:
The frame abort timeout being set by lan78xx_set_rx_max_frame_length
didn't account for any VLAN headers, resulting in very low
throughput if used with tagged VLANs.
Use VLAN_ETH_HLEN instead of ETH_HLEN to correct for this.

See https://github.com/raspberrypi/linux/issues/2458

### DIFF
--- a/drivers/net/usb/lan78xx.c
+++ b/drivers/net/usb/lan78xx.c
@@ -2221,7 +2221,7 @@ static int lan78xx_change_mtu(struct net_device *netdev, int new_mtu)
 	if ((ll_mtu % dev->maxpacket) == 0)
 		return -EDOM;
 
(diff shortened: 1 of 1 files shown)


6by9:
I guess this probably ought to be merged upstream, or at least notified to Microchip.
Tested on a baby VLAN setup here. between a Pi3B+ and Pi3B. Without the patch iperf3 fails to transfer anything. With the patch I was getting 94Mbits/s.

popcornmix:
Agreed this should go upstream.
But I'm ok with merging this now as it looks pretty sane and fixes a current issue.

PR #2501, opened 2018-04-09
net: lan78xx: Request s/w csum check on VLAN tagged packets.

6by9:
There appears to be some issue in the LAN78xx where the checksum
computed on a VLAN tagged packet is incorrect, or at least not
in the form that the kernel is after. This is most easily shown
by pinging a device via a VLAN tagged interface and it will dump
out the error message and stack trace from netdev_rx_csum_fault.
It has also been seen with standard TCP and UDP packets.

Until this is fully understood, request that the network stack
computes the checksum on packets signalled as having a VLAN tag
applied.

See https://github.com/raspberrypi/linux/issues/2458

### DIFF
--- a/drivers/net/usb/lan78xx.c
+++ b/drivers/net/usb/lan78xx.c
@@ -2954,8 +2954,12 @@ static void lan78xx_rx_csum_offload(struct lan78xx_net *dev,
 				    struct sk_buff *skb,
 				    u32 rx_cmd_a, u32 rx_cmd_b)
 {
+	/* Checksum offload appears to be flawed if used with VLANs.
+	 * Elect for sw checksum check instead.
+	 */
 	if (!(dev->net->features & NETIF_F_RXCSUM) ||
-	    unlikely(rx_cmd_a & RX_CMD_A_ICSM_)) {
+	    unlikely(rx_cmd_a & RX_CMD_A_ICSM_) ||
+	    (rx_cmd_a & RX_CMD_A_FVTG_)) {
 		skb->ip_summed = CHECKSUM_NONE;
 	} else {
 		skb->csum = ntohs((u16)(rx_cmd_b >> RX_CMD_B_CSUM_SHIFT_));


popcornmix:
I'd prefer a better understanding of this (e.g. what is special about VLAN packets? Just the length of something else? Any other classes of network packets that may have the same issue?)

But no objections to this as a temporary fix.

6by9:
Yes, I'm still investigating so it is only (hopefully) a temporary workaround.

As the first guess it is just that the VLAN tag is 4 bytes inserted before the payload, so any assumptions about the offset for parts of the IP header will be invalid. 

Can we get Microchip involved too? It's not totally clear how the RFE (Request Filtering Engine) is breaking the packet down. The datasheet claims that it works on TCP, UDP, ICMP, and IGMP, and should remove the VLAN tag. Something in that lot doesn't seem to be holding true, and yet the status flags claim they've found and computed the checksums.
We know they had checksum issues with IPv6 on 9514, so I'm not trusting the hardware yet.

Current investigations seem to indicate that ICMP will always report 0xFFFF, independent of size, address, or anything else. The odd other packet is also failing.
My observations say that when this is logged the packet is NOT dropped (see net/core/datagram.c __skb_checksum_complete), so this by itself should not be causing any errors.

It also appears that the RFE is correctly extracting the VLAN tag info, but the driver isn't then passing that up the stack so software is then reprocessing the VLAN header. Is hw acceleration of VLAN tags something MicroChip are planning on adding, and might that give us a better combination of options that means we get correct checksums? I'll do a few quick experiments with that bit to see if it comes out in the wash.
(The offload gain is going to be minimal if it doesn't also set up the VLAN filtering table, although it would be unusual to configure your switch to be throwing VLANs down a port to a device that didn't want them).

pelwell:
Microchip are aware of #2458 - it would be good to keep discussion there.

6by9:
Thanks. Will do.

PR #2502, opened 2018-04-10
LAN78xx VLAN checksum fix

6by9:
Implementing VLAN header stripping in the LAN78xx appears to make the checksum issues go away in my test system.

This patch also exposes VLAN filtering - I'm not sure that really adds that much on a sensibly configured switch, but the functionality was 90% there anyway so it seemed worth completing. (That could be split out into a second patch if you want it that way).

With these patches it is still possible to execute ```ethtool -K eth0 rx-vlan-hw-parse off``` and switch into the old bad state of incorrect checksums. Do we want to try and cover that one off, or leave it that the defaults work? It's a couple of extra conditionals in lan78xx_rx_csum_offload to keep part of the workaround in place if VLAN stripping isn't enabled, but that code is going to be executed for every received frame therefore it wants to be as efficient as possible.

### DIFF
--- a/drivers/net/usb/lan78xx.c
+++ b/drivers/net/usb/lan78xx.c
@@ -2295,7 +2295,7 @@ static int lan78xx_set_features(struct net_device *netdev,
 		pdata->rfe_ctl &= ~(RFE_CTL_ICMP_COE_ | RFE_CTL_IGMP_COE_);
 	}
 
-	if (features & NETIF_F_HW_VLAN_CTAG_RX)
+	if (features & NETIF_F_HW_VLAN_CTAG_FILTER)
 		pdata->rfe_ctl |= RFE_CTL_VLAN_FILTER_;
 	else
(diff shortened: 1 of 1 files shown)


6by9:
Tag to #2458

popcornmix:
Looks okay to me.
I'm not too concerned about being able to enable the buggy behaviour with ethtool - perhaps upstream will have a preference?

pelwell:
I suggested falling back to SW checksum in the case that the HW parsing is disabled.

6by9:
Updated, but it's a cherry-pick from another branch so I'm currently building it to ensure it is good.

6by9:
OK, looking fine to me.
Sorry, forgot to tag the original issue in the commit text when I rewrote it :-(

pelwell:
I can't rewrite the commit message without squashing, so you'll have to fix it on your end.

6by9:
OK, patches reworded and pushed again. No changes to the code on the latest push.

PR #2513, opened 2018-04-16
Unicam driver

6by9:
Gordon had asked me to get this merged into our tree.
I know that several of these files are upstream, and I will be merging stuff upstream as I have the energy. Are there any that you want to enforce that they're upstreamed first, the patches modified until they are upstream, or currently dropped?

"tc358743: fix connected/active CSI-2 lane reporting" was submitted upstream, but has sat unloved since September - https://patchwork.linuxtv.org/patch/44397/. I'll nudge them when I'm upstreaming the other stuff.

My testing had shown that the normal ```dtparam=i2c_vc=on``` command worked, and i2c0 then appeared on GPIO 0&1 (or appropriate), but I will double check it as it was a while ago.


pelwell:
I've not been through the whole PR but I have a few early comments:
1. My `overlaycheck` script is complaining that i2c0_pins isn't defined (it's referenced by the `i2c0-bcm2708` script). How do you square using the i2c muxer with users who are used to also having i2c0 to themselves (usually on Compute Modules) with the ability to choose the pin usage?
2. Early Pis (Rev 1 Model B and Model A) had the I2C buses swapped (see CAMERA_0_I2C_PORT in dt-blob.dts). This means that the mux would have to be on i2c1 for those Pis - how are you going to do that?

pelwell:
To partly answer 1 myself, could you leave i2c0_pins in place and use it instead of i2c0_gpio in the muxer?

6by9:
1. It's a tricky one. ```i2c0-bcm2708``` probably needs deprecating for normal variants - the only time it really got used was when you wanted to talk to the display or camera from Linux, and that now isn't necessary if you go via the mux. CM would potentially make sense to leave the mux out to avoid it claiming pins that they may not want.
2. That also came up when I had tried getting this upstream. The intent was that the mux wasn't defined for the early Pi variants. I need to remind myself whether we have that differentiation available in kernel DT between rev1 and rev2 B.

pelwell:
1. What about my suggestion?
2. Upstream has dedicated rev1 and rev2 dtbs, we adjust at boot time by defining aliases accordingly.

6by9:
1. Yes we could leave i2c0_pins there. (It's a shame you can't set an alias to point at an alias seeing as upstream have defined all the GPIO combinations. Duplication is likely to cause confusion.)
Do we still have a problem if people do use ```i2c0-bcm2708``` on a Pi 0/B+/2/3 where they will potentially remap the mux port 0 to be the same as port 1. The mux won't load as the second port won't be able to claim the GPIO pins. Do we rework ```i2c0-bcm2708``` to disable the mux and revert to direct access?
2. OK, that's a pain. I think the only solution is to drop the mux for both then, and not support camera/display access on them without custom DT changes. I can live with that.

pelwell:
1. Yes - I thought the overlay could disable the mux in any way you choose.
2. I'm not averse to a splitting bcm2708-rpi-b.dts - we are trying to move closer to upstream, and the firmware has a shiny new way of specifying fallback DTB files it would be nice to make more use of.

6by9:
1. OK, that's a plan. ```i2c0-bcm2708``` will kill the mux and give you i2c0 on the specified node. I suspect it will have issues with dynamic loading as it would need to rebind the i2c nodes in /dev to alternate numbers.
2. ~~As just discussed, those early models don't bring the alternative pin mux options out to a useful header. One BSC always goes to header pins 3&5, and the other BSC goes to the camera.~~ Blast, B rev2 exposes GPIOs 28 & 29 on the P5 header. OK, I can look at splitting rev1 & 2 so that we can fully describe that. I'm still tempted to leave the mux out though for those boards - if I load the mux by default then it'll claim 28 & 29, and P5 was generally used for I2S so we'd end up with a conflict.

6by9:
I'm trying to push just the driver changes here, leaving DT changes for a new PR. Looks like I've left the CSI DT changes in here though, so updates incoming.

6by9:
OK, I think that's the driver side all sorted. New PR incoming for DT.

6by9:
@pelwell Can this lot get a review please?
I can then sort overlays for the devices that just remux i2c0 rather than needing all the changes of #2578.

drivers/media/i2c/ov5647.c
```diff
@@ -442,8 +442,30 @@ static int ov5647_enum_mbus_code(struct v4l2_subdev *sd,
...
+	fmt->width = 640;
+	fmt->height = 480;
+
+	return 0;
+}
+
 static const struct v4l2_subdev_pad_ops ov5647_subdev_pad_ops = {
 	.enum_mbus_code = ov5647_enum_mbus_code,
+	.set_fmt =	  ov5647_set_get_fmt,
+	.get_fmt =	  ov5647_set_get_fmt,
```
pelwell, line 488: I find it strange that this shared set_get handler is acceptable, but I see at least one other driver using it.

drivers/media/i2c/tc358743.c
```diff
@@ -1458,28 +1458,29 @@ static int tc358743_g_mbus_config(struct v4l2_subdev *sd,
...
-	/* Support for non-continuous CSI-2 clock is missing in the driver */
-	cfg->flags = V4L2_MBUS_CSI2_CONTINUOUS_CLOCK;
+	/* In DT mode, only report the number of active lanes */
+	if (sd->dev->of_node)
+		return 0;
 
-	switch (state->csi_lanes_in_use) {
-	case 1:
+	/* Support for non-continuous CSI-2 clock is missing in pdata mode */
+	cfg->flags |= V4L2_MBUS_CSI2_CONTINUOUS_CLOCK;
```
pelwell, line 1491: Why is this assignment unconditional, and not copying (some of) `state->bus.flags`?

drivers/media/i2c/tc358743.c
```diff
@@ -219,7 +223,7 @@ static noinline void i2c_wrreg(struct v4l2_subdev *sd, u16 reg, u32 val, u32 n)
 
 static u8 i2c_rd8(struct v4l2_subdev *sd, u16 reg)
 {
-	return i2c_rdreg(sd, reg, 1);
+	return i2c_rdreg(sd, reg, 1, NULL);
```
pelwell: This is a bit ugly - why not make i2c_rdreg a call to a new function - i2c_rdreg_err, say, passing in a NULL error pointer? The callers that care about the error would then call the new function.

drivers/media/i2c/adv7180.c
```diff
@@ -1266,6 +1267,18 @@ static int init_device(struct adv7180_state *state)
 			goto out_unlock;
 	}
 
+	/* Select first valid input */
+	for (i = 0; i < 31; i++) {
```
pelwell: Is it not possible to have 32 inputs? Is 31 the best maximum?

pelwell (commented):
I didn't review the driver internals in detail - that's a job for V4L2 experts upstream - but apart from the questions and comments it looks OK.

drivers/media/i2c/ov5647.c
6by9: set_fmt is obliged to do the best it can to match the requested format/resolution.
In this case there is only one format/resolution, therefore the answer is always the same, and is the same as getting the format.

drivers/media/i2c/tc358743.c
6by9: Will do.

drivers/media/i2c/ov5647.c
pelwell: So it isn't strange that the parameters supplied to the `set` get overwritten?

drivers/media/i2c/adv7180.c
6by9: The valid values are defined by ADV7180_INPUT_* and ADV7182_INPUT_* at line 156ish. They max out at 0x11.
Checking the datasheet, whilst INSEL is 5 bits, 10010 to 11111 are reserved, so there is no valid use case for it to be set to greater than 0x11.

Seeing as I need to update tc358743, I'll amend this to check the top bit anyway.

drivers/media/i2c/ov5647.c
6by9: No, not in this case.
Should the driver support multiple modes then it'd get updated to choose the best match, but often in sensor driver such as this where only certain modes are supported, that will still result in an overwrite of the input.

Docs at https://www.kernel.org/doc/html/v4.15/media/uapi/v4l/vidioc-g-fmt.html and https://www.kernel.org/doc/html/v4.15/media/uapi/v4l/vidioc-subdev-g-fmt.html#vidioc-subdev-g-fmt (Main V4L2 and subdev APIs respectively.)
V4L2 docs section 7.31.4 para 3:
> When the application calls the VIDIOC_S_FMT ioctl with a pointer to a struct v4l2_format structure the driver checks and adjusts the parameters against hardware abilities. Drivers should not return an error code unless the type field is invalid, this is a mechanism to fathom device capabilities and to approach parameters acceptable for both the application and driver. On success the driver may program the hardware, allocate resources and generally prepare for data exchange. Finally the VIDIOC_S_FMT ioctl returns the current format parameters as VIDIOC_G_FMT does. **Very simple, inflexible devices may even ignore all input and always return the default parameter**

drivers/media/i2c/tc358743.c
6by9: Sorry, missed this one earlier. That's part of an upstream commit by Philipp Zabel that needs to be pinged. https://patchwork.linuxtv.org/patch/44397/

Much discussion on the linux-media over g_mbus_config fullstop. Hans Verkuil was fighting that it was unneeded and deprecated as DT conveys all the info, but that doesn't handle the case where the device (eg TC358743) can dynamically choose to use fewer lanes than are physically present (DT only describes those physically present).
This line is part of the original mode of operation, but is now only valid for platform data driven systems (Cisco use this chip and driver in one of their video conferencing products via platform data). DT systems have already aborted by this point.

6by9:
I think that addresses your review comments - thank you :-)

pelwell:
Ship it!

6by9:
:-o It's finally merged (even if not upstream).

I will kick all those driver patches upstream (except the debug hack on adv7180) when I have the energy.

PR #2519, opened 2018-04-19
Driver for Allo Katana DAC

allocom:
Allo Katana DAC uses Sabre ESS codec.


clivem:
Have ESS changed their stance? AFAIK, there is still the issue for any of their programmable Sabre chips, (ie. I2C software controlled), that you sign a NDA to get the register info/address/values. 
Unless they have changed their stance, they have forbidden any Open Source driver where any register address or programming value, that is obtained from a NDA'd datasheet, being exposed in an Open Source driver! Maybe you know this, maybe you don't. Just a friendly warning before you potentially drag the RPi guys into a legal nightmare with ESS, for "publishing" such a driver......

HiassofT:
The soundcard driver (allo-katana-dac.c) doesn't seem to do much, you can drop it and use the simple card or audio graph card driver instead.

Here's a (completely untested) dt overlay, please give that a try: https://gist.github.com/HiassofT/5f7aec2407f854e184aecdc7d502779e

pelwell:
[ Comment 1 has been largely preempted before submission, but still stands. ]
1. I believe you may be violating your NDA by publishing this driver - do you really want to do this?
2. The codec is a standard component and its driver should be upstreamed.

AlloKatana:
Hi,

 I am one of the designers of the Katana Dac (using ess9038q2m). We are very well aware of the NDA from ess sabre (that we signed). So we used the only solution possible. Katana Dac has an MCU on board with the actual protected information (nda) of Ess Sabre. The driver that we are submitting is the driver of the MCU (and the translation is done securely in the MCU). Let me know if you have any other questions.

pelwell:
So is the codec a standard component or have you modified the MCU firmware for your application?

Is the API exposed by the MCU in the public domain?

HiassofT:
If the codec driver (sabre-ess-i2c) isn't talking to the ESS chip but your MCU then you should name it differently - allo-ess-bridge-i2c or sth like that

allocom:
Hi Pelwell, Codec is not a standard component. MCU firmware is custom written & will not be made public. Register addresses & its definition are not directly mapped to the ESS chip registers.

pelwell:
In that case you should:
1. delete the sound card driver and use the overlay suggested by HiassofT, and
2. move the codec driver into sound/soc/bcm, giving it a product-specific name - allo_katana_codec.c would be fine.

allocom:
I am checking the overlay suggested by HiassofT (Thanks HiassofT) & will move the codec driver.

allocom:
Hello Pelwell, I modified to have only one file (allo_katana_codec.c) under bcm. Pls check if it is fine.

arch/arm/boot/dts/overlays/README
```diff
@@ -314,6 +314,12 @@ Load:   dtoverlay=allo-digione
 Params: <None>
 
 
+Name:   allo-katana
```
pelwell: This should be `allo-katana-dac-audio`

arch/arm/boot/dts/overlays/README
```diff
@@ -314,6 +314,12 @@ Load:   dtoverlay=allo-digione
...
+Info:   Configures the Allo Katana audio card
```
pelwell: `Allo Katana DAC audio card` for consistency? You can choose.

pelwell (commented):
Apart from my README comments, that looks good.

allocom:
Modified README file.

AlloKatana:
Thx you Pelwell for merging, HiassofT for the suggestions and CliveM for the comment (valid).

pelwell:
I used the wrong merge option - I intended to squash them - but I'll fix it up in the other kernel versions.

allocom:
Thanks all

Issue #2512, opened 2018-04-16
 Under-voltage detected! (0x00050005) spams dmesg on new kernel 4.14.30-v7+


E3V3A:
After upgrade of kernel to 4.14. dmesg is now spammed by `Under-voltage detected! (0x00050005)` messages where no problem was shown previously. I've ran this device non-stop for months, without any problem until after update, so under-voltage level settings or other config must have changed. 
Spamming dmesg or `journlctl --system` buffer certainly is not helping anyone. 

```
kern  :crit  : [ 1701.464833 <    2.116656>] Under-voltage detected! (0x00050005)
kern  :info  : [ 1707.668180 <    6.203347>] Voltage normalised (0x00000000)
```

Also related to #2367

pelwell:
The kernel under voltage notification is new, but the threshold and detection mechanism is unchanged. You are now being made aware of the fact that your Pi is insufficiently powered for the load placed upon it. This is bad for performance and potentially harmful to system stability.

rodizio1:
I can tell from lots of testing and measuring with a multimeter, that 99% of the "typical" USB power supplies (phone chargers, gadget chargers, power banks, etc.) either don't supply the advertised amps, or they drop voltage too much. Or both. Even if the power supply itself delivers stable 5V at 2-2.5amps, often the voltage at the end of the cable has dropped too much because of too thin wires.

E3V3A:
I'm well aware of the working of the RPi power supply. But the fact of the matter is that:

* It didn't happen with the earlier kernel, so what make you think this is an improvement?  
   (The flash icon was good and annoying enough!)
* Spamming kernel messages (and all other related buffers) with repeated messages on a device where which has proven to be working just fine under those conditions before, is now risking excessive SD card wear and harder debugging because older and more relevant kernel messages get FIFO'd out eventually.
* The current `crit` level doesn't even respect the `printk` settings and keep spamming even after setting `dmesg -n 1` or using `sysctl -w '1 1 1 1'`. So AFAICT, this is neither *critial*, nor compliant to standard *nix behavior, and does not provide any improvement whatsoever.

P33M:
>It didn't happen with the earlier kernel, so what make you think this is an improvement?
>(The flash icon was good and annoying enough!)

[...]

PR #2520, opened 2018-04-20
Prevent voltage low warnings from filling log

JamesH65:
Although the correct fix for low voltage warnings is to
improve the power supply, the current implementation
of the detection can fill the log if the warning
happens freqently. This replaces the logging with
slightly custom ratelimited logging.

### DIFF
--- a/drivers/firmware/raspberrypi.c
+++ b/drivers/firmware/raspberrypi.c
@@ -24,6 +24,38 @@
 
 #define UNDERVOLTAGE_BIT		BIT(0)
 
+
+/*
+ * This section defines some rate limited logging that prevent
+ * repeated messages at much lower Hz than the default kernel settings.
+ * It's usually 5s, this is 5 minutes.
+ * Burst 3 means you may get three messages 'quickly', before
+ * the ratelimiting kicks in.
+ */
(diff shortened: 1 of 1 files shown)


JamesH65:
Just realised my comment isn't to kernel coding standard. Will fix that.

jakemagee:
Does adding this code / complexity actually accomplish anything useful?  I've been following #2512 ... It seems pretty clear that the problem is a hardware issue.  It seems backwards to add code / complexity that affects everyone in-order to try and mask one person's issue.  It might be debatable if the change fixed a bug, but the original code seems correct and logical... if a power issue is detected, log it... period... no extra complexity necessary.  Virtually everyone would agree power issues need to be addressed immediately... the potential of "spamming" the logs is irrelevant.

I believe there have been suggestions in #2512 as to how the person could mask any warnings they desire... it could even be suggested that they modify and rebuild with the warning messages completely removed if they like (the great thing about open source code).

JamesH65:
@jakemagee This commit ratelimits the logging, which is still present, rather than turning it off completely. Very little extra complexity, I've used the kernel ratelimiting functions so no new code. Although I do agree that the underlying problem should be fixed!!

The problems is that some people seem to think ignoring the symptoms is a sensible thing to allow. It isn't, of course, which is why this ratelimits rather than ignoring.

PR #2534, opened 2018-04-27
Reduce log spam when mailbox call not implemented

JamesH65:
This changes the logging message when a mailbox call
fails to the dev_dbg level. In addition, it fixes the
low voltage detection logging code so that if the
mailbox call doies fails, it logs at error level
and flags so the call is no longer attempted.

### DIFF
--- a/drivers/firmware/raspberrypi.c
+++ b/drivers/firmware/raspberrypi.c
@@ -151,7 +151,7 @@ int rpi_firmware_property_list(struct rpi_firmware *fw,
 		 * error, if there were multiple tags in the request.
 		 * But single-tag is the most common, so go with it.
 		 */
-		dev_err(fw->cl.dev, "Request 0x%08x returned status 0x%08x\n",
+		dev_dbg(fw->cl.dev, "Request 0x%08x returned status 0x%08x\n",
 			buf[2], buf[1]);
 		ret = -EINVAL;
 	}
@@ -204,6 +204,7 @@ EXPORT_SYMBOL_GPL(rpi_firmware_property);
 
 static int rpi_firmware_get_throttled(struct rpi_firmware *fw, u32 *value)
 {
+	static int old_firmware;
 	static ktime_t old_timestamp;
 	static u32 old_value;
 	u32 new_sticky, old_sticky, new_uv, old_uv;
@@ -213,6 +214,9 @@ static int rpi_firmware_get_throttled(struct rpi_firmware *fw, u32 *value)
 
 	if (!fw)
 		return -EBUSY;
+
+	if (old_firmware)
+		return -EINVAL;
 
 	/*
 	 * We can't run faster than the sticky shift (100ms) since we get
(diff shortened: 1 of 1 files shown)


pelwell:
checkpatch has a few issues, but otherwise I'm OK with this.

JamesH65:
BLOODY CHECKPATCH. I always forget to use it.

drivers/firmware/raspberrypi.c
```diff
@@ -232,8 +236,18 @@ static int rpi_firmware_get_throttled(struct rpi_firmware *fw, u32 *value)
 
 	ret = rpi_firmware_property(fw, RPI_FIRMWARE_GET_THROTTLED,
 				    value, sizeof(*value));
+
 	if (ret)
+	{
+		/* If the mailbox call fails once, then it will continue to
```
procount: Is having old firmware the only reason why this call shall fail? In which case I'm ok. 
Or are there other error returns apart from 0x80000001 that could be returned and should be considered?

pelwell: There is no nuance to the error returns, it's just a case of "Computer says no".

JamesH65: Comment should probably say "usually". There are other reasons, but all probably fairly fatal.

6by9:
It seems a little wasteful that we're still polling away even when we're going to abort in ```get_throttled```.
Why not amend ```get_throttled_poll``` to
```
 	ret = rpi_firmware_get_throttled(fw, &dummy);
 	if (ret)
 		pr_debug("%s: Failed to read value (%d)", __func__, ret);
 	else
 	 	schedule_delayed_work(&fw->get_throttled_poll_work, 2 * HZ);
```
So if we fail we never reschedule the work job? That also means we don't need to keep a flag of old_firmware.

JamesH65:
Yes, that is a better approach. I'll update.

6by9:
Ah, ```rpi_firmware_get_throttled``` is called from ```get_throttled_show``` too.
~~Although is that used now? It looks like it's dead code. (static, but not apparently referenced)~~

edit: macro hell! ```static DEVICE_ATTR_RO(get_throttled);``` pulls it in.

JamesH65:
Yes, the whole macro thing in the mailbox stuff is a nightmare.

JamesH65:
Just pushed a combination of 6by9 suggestion and original. Because the function is called from multiple places, the check still need to be in the function itself, but have taken in to account the workqueue reschedule as well.

6by9:
Looks good to me, although I haven't looked as to what checkpatch says to it (no Signed-off-by would be its first whinge).

JamesH65:
Added signoff.   checkpatch -f doesn't flag up signoffs since it works on the file rather than a patch file. Is there an easy way of  checking a commit as a file but as if it were a patch?

pelwell:
```
$ git format-patch --stdout HEAD^ | scripts/checkpatch.pl
```
or
```
$ git format-patch --stdout -1 HEAD | scripts/checkpatch.pl
```

6by9:
Add
```
#!/bin/sh
```
as .git/hooks/post-commit for your Linux tree and it'll automagically run checkpatch after every commit.

pelwell:
s/--output/--stdout/

PR #2536, opened 2018-04-30
overlays: Add support for Balena Fin board

floion:
Signed-off-by: Florin Sarbu <florin@resin.io>

### DIFF
--- a/arch/arm/boot/dts/overlays/Makefile
+++ b/arch/arm/boot/dts/overlays/Makefile
@@ -17,6 +17,7 @@ dtbo-$(CONFIG_ARCH_BCM2835) += \
 	audioinjector-addons.dtbo \
 	audioinjector-wm8731-audio.dtbo \
 	audremap.dtbo \
+	balena-fin.dtbo \
 	bmp085_i2c-sensor.dtbo \
 	dht11.dtbo \
 	dionaudio-loco.dtbo \
--- a/arch/arm/boot/dts/overlays/README
+++ b/arch/arm/boot/dts/overlays/README
@@ -404,6 +404,13 @@ Params: swap_lr                 Reverse the channel allocation, which will also
                                 (default off)
 
 
+Name:   balena-fin
+Info:   Overlay that enables WiFi, Bluetooth and the GPIO expander on the
+        Balena Fin board.
+Load:   dtoverlay=balena-fin
+Params: <None>
+
+
 Name:   bmp085_i2c-sensor
 Info:   This overlay is now deprecated - see i2c-sensor
 Load:   dtoverlay=bmp085_i2c-sensor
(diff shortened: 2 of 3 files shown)


floion:
Hi @pelwell ,
this is a pull request for adding support for a CM3 lite based board:

https://resin.io/blog/introducing-project-fin-a-board-for-fleet-owners/

Thank you

pelwell:
1. Your overlay source should not contain `__fixups__` or `__local_fixups__` - it looks like you've written this overlay by decompiling an existing one and modifying it. Although the end result may work the result is hard to understand and fragile - a change to the overlay could break it. Please rewrite it so I can review it properly.
2. Please add an entry to the README file in the overlays directory, observing the strict indentation rules (max 80 columns, no TABs, 2 blank lines between entries).

pelwell:
3. Can you explain the reduced SDIO speed?

floion:
Hi @pelwell , thanks for the reply.
Sorry for the initial commit, I did indeed commit a decompiled version of the overlay. I have now amended the commit with the requested changes.

As for limiting the reduced SDIO speed, having anything higher than 35 MHz resulted in errors such as:
> mmc1: error -110 whilst initialising SDIO card

pelwell:
>  I have now amended the commit with the requested changes.

That looks good - I guess you know you could apply fragment 3 directly into the i2c@0 node in fragment two, but this way is quite readable and easy to compare with other overlays.

> mmc1: error -110 whilst initialising SDIO card

If reducing the bus speed makes it work then it smells like a signal integrity issue, assuming the device you are driving is supposed to be OK at 50MHz. But that's a matter between you and your users.

pelwell:
Ah, there is a problem (although you must be getting away with it) - a frustrating limitation of overlays is that they can't delete a property. Most of this time this isn't a problem, but boolean properties like `non-removable` are true if present and false if absent, so if the base DTB includes a true boolean there is no way to delete it.  [ I'd like to fix this, but it will mean extending the FDT encoding and getting it accepted upstream. ]

There is an ugly workaround - creating a clone of the mmc node in the overlay without the property (or with a parameter to delete it before the overlay is applied) with another fragment to disable the existing node. Duplication aside, this is ugly because if the mmc node content ever changes then the cloning overlays will have to be updated accordingly, but it does work.

floion:
@pelwell if you are okay with leaving the 3rd fragment, I would do that for readability as you said.

As for the property deletion problem, I would rather not do the workaround of redefining the target node and would just delete it from here entirely (I amended the PR and removed that entry altogether as well as I have redefined the fragment 1 overlay node to say "sdio_wifi" instead of "sdio_mmc" for better readability).

Thank you

floion:
Thanks a lot @pelwell

PR #2546, opened 2018-05-08
Added support to MAX3421E USB Host in Device Tree

cezarsantanna-zz:


pelwell:
I'm happy with the intent behind this commit but there are a few problems with it:

1. 4.14 is the current shipping kernel version - it would make more sense to submit the PR for rpi-4.15.y, but I see that the 4.14 version of the driver doesn't have DT support. Is that the reason you chose 4.15?
2. The MAX3421 driver isn't enabled by the standard defconfigs, so this overlay won't work for people unless they customise the configuration.
3. There is no entry in .../overlays/Makefile for the overlay, so it won't get built.
4. There is no entry in .../overlays/README for the overlay.
5. The overlay source file should be called `<name>-overlay.dts`, which is then compiled into `<name>,dtbo`.
6. The use of `usb0` in the name is confusing - is that name chosen because the overlay uses SPI0?
7. The overlay forces the use of GPIO 3 on the device to drive VBUS - should this really have a parameter?
8. By default, GPIO 25 pulls down, so at best it will be fighting against the external pull-up (or drive of the MAX3421), but at worst it won't work. `brcm,pull = <2>;` would be a sensible addition.

cezarsantanna-zz:
After this post I discovery some issues about the driver or the overlay file, I am getting this info on dmesg:
` 7886.729387] max3421_handle_irqs: BUSEVENT Bus Resume Done
[ 7946.916769] max3421_handle_irqs: BUSEVENT Bus Resume Done
[ 8006.719030] max3421_handle_irqs: BUSEVENT Bus Resume Done
[ 8185.987975] max3421_handle_irqs: BUSEVENT Bus Resume Done`

Then I lost the connection on USB guest, my usb gest is a uBlox SARA-U201.

cezarsantanna-zz:
I am closing this PR because de dts file was not working write, I will make the corrections and then make a new PR again.

PR #2552, opened 2018-05-12
Add ability to export the pin used by the gpio-poweroff driver

nick-bulleid:
If the GPIO line is configured to be an input, or for diagnostic purposes, it would be useful to examine the state of the GPIO line from userspace. This change allows this by exporting the GPIO line to sysfs.


pelwell:
That's neat. Could you:
1. fix the indentation of the `export` parameter (use TABs not spaces to match the rest of it),
2. add the `export` parameter to the overlays README,
3. arrange the changes into three commits - driver, bindings document, and overlay+README - including a "Signed-off-by:" line (`git commit -s`, and `git push -f` to update the PR)?

The first two of those commits (not the overlay) could then be submitted upstream, but I'll accept it here in the meantime.

nick-bulleid:
> That's neat.

Thanks!

I've made the changes you suggested.

PR #2569, opened 2018-05-31
Enable bbr module for arm64

xunzhaocnm:


pelwell:
These options are present in the other Pi defconfigs, so happy to merge.

PR #2578, opened 2018-06-08
Unicam driver DT changes

6by9:
Picking this up again after a moderate break, so I'm having to check the exact state.
I still need to double check that this addresses all the issues raised from #2513 review.

i2c0-bcm2708 overlay should do "the right thing" and disable the mux.
As discussed, I was intending to not support the mux (and hence any of the overlays) at all on model A and Bs where it varies on whether it is BSC0 or BSC1. It looks like I haven't done that bit.


6by9:
Long since obsolete. Closing.

PR #2582, opened 2018-06-12
[4.17] Justboom Dac+Digi: change codec to component

HiassofT:
struct snd_soc_codec has been replaced with snd_soc_component.

In 4.17 the struct is still present, but for example codec in snd_soc_runtime isn't initiallized anymore and accessing it results in a null pointer oops.

In 4.18 snd_soc_codec is removed completely and code using it will no longer compile.

This PR fixes the Justboom Dac and Digi drivers, compile and runtime test on RPi3 worked fine.

@popcornmix the other soundcard drivers need to be changed following the same pattern


popcornmix:
Thanks!

PR #2583, opened 2018-06-13
net: lan78xx: Disable TCP Segmentation Offload (TSO)

6by9:
TSO seems to be having issues when packets are dropped and the
remote end uses Selective Acknowledge (SACK) to denote that
data is missing. The missing data is never resent, so the
connection eventually stalls.

There is a module parameter of enable_tso added to allow
further debugging without forcing a rebuild of the kernel.

https://github.com/raspberrypi/linux/issues/2449
https://github.com/raspberrypi/linux/issues/2482


JamesH65:
Was going to say "LGTM", but already merged! The Speed!

PR #2589, opened 2018-06-24
Add device tree overlay for HD44780

jboomer:
Added a device tree overlay for an HD44780 compatible LCD


arch/arm/boot/dts/overlays/hd44780-lcd-overlay.dts
```diff
@@ -0,0 +1,46 @@
+/dts-v1/;
+/plugin/;
+
+/ {
+    compatible = "brcm, bcm2708";
```
pelwell: Although in practice this string isn't checked, it should at least be one of the recognised strings. These days I would just use: "brcm,bcm2835" (note the lack of a space after the comma).

pelwell (changes requested):
Sorry for the review delay - this one slipped under the radar.

Apart from the compatible string, I think you could shorten the parameter names - they don't need the `lcd_` prefix - but I leave the decision to you.

Also note that this overlay won't be much use until the driver module is enabled in the defconfigs.

Otherwise this looks good.

jboomer:
> Sorry for the review delay - this one slipped under the radar.

No problem!

> Also note that this overlay won't be much use until the driver module is enabled in the defconfigs.

I know, it will be useful for people compiling their own kernel I guess. If it is included, I guess the dtbo will also be added to the firmware repo?

> I think you could shorten the parameter names

Won't this potentially clash with other parameter names? Or if there are two overlays with a 'pin' parameter and config.txt contains:

```
dtoverlay=first-overlay`
dtparam=pin=1
dtoverlay=second-overlay
dtparam=pin=2

```
, will the kernel know which one applies to which overlay?

I will add the fixes later this weekend

pelwell:
> I know, it will be useful for people compiling their own kernel I guess.

That's fine, but if you did want those options (CONFIG_AUXDISPLAY=y and CONFIG_HD44780=m) setting as standard then they would only add two modules of 15kB and 10kB...)

> If it is included, I guess the dtbo will also be added to the firmware repo?

Yes, of course.

> Won't this potentially clash with other parameter names?

No. The parameters declared by an overlay go out of scope at the next `dtoverlay` directive. The worst that can happen is that they mask a parameter of the base .dtb, in which case you can force the scope to close with an empty `dtoverlay=` line. The corollary of this is that until the next `dtoverlay` directive the parameters remain in scope, meaning that the parameters to an overlay can be split across multiple lines as you have above, rather than the more concise version which is also supported:
```
dtoverlay=first-overlay,pin=1
dtoverlay=second-overlay,pin=2
```

jboomer:
Added requested changes, would you want the changes (enable hd44780 module) to defconfig in a separate PR or should I put it here as well?

pelwell:
A separate commit would make it easier in the future - when we jump kernel versions the config commits get squashed into a single commit, and the DT commits into another.

pelwell:
P.S. check the indentation in the README - it looks like it's changed and the parser is very picky and dumb.

jboomer:
Fixed the indentation and added changes to defconfig in a separate commit.

pelwell:
Thanks. I've added the options to the bcm2709_defconfig as well, squashed it all down to two commits and applied them via the command line tools.

PR #2591, opened 2018-06-26
Updated default value

allocom:
Small change in the initial value for DoP.


pelwell:
A more descriptive commit message - at least mention Allo Katana DAC or sound - and a Signed-off-by would be nice, otherwise I have no objections.

allocom:
Hope its fine, Pelwell. Thanks.

popcornmix:
Can you say what the effect of the change is?

allocom:
DoP is set ON by default, so even if players do not access this setting, DoP can be played.

pelwell:
That's the opposite of what I thought you would say. It now sounds like the old default value was more correct, even if turning it off wouldn't make any difference.

allocom:
Old default value was Off (1) (invert is set). Now it changed to On (0). 
I want it to keep it On be default, as in players separate code must be added to change this setting. If player code is not added also they should be able to play DoP easily without opening alsamixer.

PR #2596, opened 2018-06-27
Rpi 4.18.y , Problems encountered by cloning a single branch

desstiony:
I user a command ''git clone -b rpi 4.18.y XXX.git'' to find that the size was 6118895kb
I wonder why it is so big?


pelwell:
You have created a Pull Request where you wanted to create an Issue, but really this is a question that doesn't belong here at all. Try googling "git clone single branch only" and "git clone shallow" for a better understanding of your options.

PR #2600, opened 2018-06-29
arm: dts: bcm2710-rpi-3-b-plus: fix hpd gpio pin

lrusak:
hpd gpio is listed as 28 in https://github.com/raspberrypi/firmware/blob/master/extra/dt-blob.dts#L1335

This fixes hotplug detection when using vc4

### DIFF
--- a/arch/arm/boot/dts/bcm2710-rpi-3-b-plus.dts
+++ b/arch/arm/boot/dts/bcm2710-rpi-3-b-plus.dts
@@ -162,7 +162,7 @@
 };
 
 &hdmi {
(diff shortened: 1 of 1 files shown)


popcornmix:
This looks right to me. @pelwell ?

pelwell:
Yes - I've double-checked on the schematic and it is correct. The exp-gpio 4 value is a hang-over from the 3B.

pelwell:
We normally ask for a "Signed-off-by:" line in commits - can you provide one, @lrusak ?

lrusak:
@pelwell updated :+1:

pelwell:
Thanks!

PR #2604, opened 2018-07-02
Remove duplicated code in sound/soc/bcm soundcard drivers.

timg236:
There's quite a lot of duplicated code in the Raspberry Pi soundcard drivers and since they aren't upstreamed it's a bit of a maintenance burden. 

This series of commits firstly does some general refactoring to make the code less verbose by using devm for card registration and moves static bclk configuration to init instead of implementing a hw_params callback.

The second and larger change introduces a generic driver for simple cards. The simple card drivers don't really do much more than set bclk but unfortunately there's no way to do that via the soc/generic cards preventing them from becoming "DT only" drivers.

This version adds support for ADAU1977 ADC, Google VoiceHat, HifiBerry AMP, HifiBerry DAC and RPI DAC.

TODO
There's a middle ground for drivers which are slightly more complex. It might be possible to create utils functions for common features e.g. muting via gpio, volume limiting. However, I think it's better to get the simple fixes in first.


popcornmix:
Ping @HiassofT - interested in your comments.

sound/soc/bcm/Kconfig
```diff
@@ -196,3 +196,16 @@ config SND_PISOUND
...
 	  Say Y or M if you want to add support for Blokas Labs pisound.
+
+config SND_RPI_SIMPLE_SOUNDCARD
+	tristate "Support for Raspberry Pi simple soundcard driver"
+	depends on SND_BCM2708_SOC_I2S || SND_BCM2835_SOC_I2S
+	depends on I2C
+	select SND_SOC_ADAU1977_I2C
+	select SND_SOC_PCM1794A
+	select SND_SOC_PCM5102A
+	select SND_SOC_PCM512x_I2C
```
HiassofT: PCM512x doesn't seem to be used by the drivers and can be dropped

HiassofT:
thanks, looks fine, only thing that stuck out was the PCM512x_I2S selection in Kconfig which can be dropped (unless I missed something)

sound/soc/bcm/Kconfig
timg236: Will do. That's some old config from when I was looking at the iqaudio-dac which isn't covered by RPi simple

pelwell:
I was wondering if we could keep the old per-soundcard config settings but make them select the rpi-simple driver (which then wouldn't need a menu entry). The Kconfig entry for each card would select its own codec and our driver, and the rpi-simple driver could have codec support predicated on the soundcard config settings.

I think this scheme would allow our umbrella driver to support a single soundcard without adding any significant bloat over the dedicated driver, and the full-fat Raspbian build would be identical to this PR.

HiassofT:
Setting up Kconfig that way is a good idea. If we keep settings as is (and only make them select the generic drivers) we also get a seamless upgrade path via make oldconfig

sound/soc/bcm/Makefile
```diff
@@ -1,4 +1,4 @@
-# BCM2835 Platform Support
+# BCM2836 Platform Support
```
timg236: Remove vim ctrl+a typo :)

timg236:
I've added an extra commit which adds a generic WM8804 driver replacing JustBoom Digi, IQAudIO Digi, HifiBerry Digi and Allo DigiOne cards.

As with the rpi-simple driver this should not be visible to end users or require any defconfig changes.

Looking for feedback from the driver authors before posting  this on the forums.

I have tested with the JustBoom digi soundcard. Unfortunately, I don't have the other cards to test with at the moment.

pelwell:
@Hifiberry @iqaudio @shawaj @allocom @PeterMalkin Any feedback?

hifiberry:
This won't work as the whole clock-selection code has been removed.
 Both the DAC+ Pro and the Digi+ Pro are not "generic" cards, but use their own clock generators that are controlled by the existing drivers. 

While I like the idea to have as much code as possible moved to a generic driver, there is a need for adoptions for a specific hardware.

hifiberry:
@timg236 : Please contact me at info@hifiberry.com if you need samples for development.

timg236:
@hifiberry the DAC+Pro driver should not be changed by these patches. 

The HifiBerry Digi+ and Allo Digione should both have the correct clock selection from snd_rpi_wm8804_probe unless I've missed something.

hifiberry:
@timg236 : Right, the DAC+ driver isn't affected, but the Digi+ is. There is hardware-specific clock control code (check lines 43 and later in the Digi+ driver) that is not part of a generic WM8804 driver. As far as I see this code is simply remove by this patch.

timg236:
@hifiberry snd_rpi_hifiberry_digi_enable_clock and moved to snd_rpi_wm8804_enable_clock at line 112.

It should be following the same logic of checking for "clock44" and "clock48" from devm and using that to distinguish between the Digi and the Digi+. The Allo DigiOne seemed to have almost identical logic so I basically renamed the functions to be generic.

Although on the Allo DigiOne there is only one variant so the absence of clock44 and clock48 are an error.

For the IQAudIO and JustBoom Digi there's no special clock configuration so the code just ignores the absence of clock44/clock48.

The design was to use the snd_rpi_wm8804_drvdata to encapsulate configuration data or optional callbacks for hardware specific customisation so there's a fair amount of wiggle room there if we need to change things.

hifiberry:
@timg236 I can't find this. What file did you include this function (snd_rpi_wm8804_enable_clock) ?

hifiberry:
Ok, now I found the file. It was hidden in the Github UI :(
I'll make sure that you get sample boards soon.

allocom:
Hi Tim,  I will test and update the result with our Allo digione.
For further, development & testing on our side, you can write to Andre (Andre@allo.com) for a sample.

HiassofT:
A quick test with a Hifiberry Digi Pro looks fine - 44.1, 48, 88.2 and 96kHz work nicely. So clock switching seems to be OK

Tested on RPi3 with current rpi-4.18.y tree plus this PR, using bcm2709_defconfig

allocom:
On quick testing with Allo Digione, it is working fine.

timg236:
Rebased - no changes to the patches

pelwell:
Thanks for the feedback. As 4.18 is hardly mainstream yet I'm happy to merge and let a few early adopters kick the tyres.

PeterMalkin:
Thank you! We have tested on VoiceHAT - works great!

PR #2605, opened 2018-07-02
vc4 backports

lrusak:
Just some backports from mainline

I'm not sure what proper etiquette is for fixing commits so I just left the small fixup as a seperate commit for now. Please advise what I should do for that.

I need these commits otherwise I have some odd behaviour when playing video in the primary plane using YUV formats and the size doesn't fill the resolution (ie. 1920x800 in a 1920x1080 screen)

This also fixes some other alpha issues.

Even if you don't want to merge this I just wanted to PR it as a useful resource incase anyone else has the same issue.

This is the issues I'm talking about:
vc4 issue

vc4 ghosting


pelwell:
> I'm not sure what proper etiquette is for fixing commits so I just left the small fixup as a seperate commit for now. Please advise what I should do for that.

Thanks for asking. I'd prefer each commit to be correct in itself and to reflect the spirit of the original commit, even if the details have to change slightly. If the commit titles match exactly the upstream versions then it is easier to skip them when forward-porting patches to a newer kernel.

> Even if you don't want to merge this I just wanted to PR it as a useful resource incase anyone else has the same issue.

We're always happy to take back-ports if they fix problems (and don't cause new ones).

lrusak:
@pelwell thanks, I pushed the two changes as fixups. If you are ok with this then I will squash them down into their relevant commits.

pelwell:
Replacing `fb->format` with `vc4_get_hvs_format(fb->format->format)` looks wrong, but I'm sure you've tested it.

One thing I should have mentioned before is that we like the first line of the commit message of a back-port (after the title and the mandatory blank line) to be:
```
commit <long hash> upstream.

<original message>
```

If you patch up the message and squash in your fixups, I'll merge.

lrusak:
>Replacing fb->format with vc4_get_hvs_format(fb->format->format) looks wrong, but I'm sure you've tested it.

I'm not sure where you see this?

pelwell:
You are replacing `fb->format` with `format`:
```diff
-	vc4_state->needs_bg_fill = fb->format->has_alpha || !covers_screen;
+	vc4_state->needs_bg_fill = format->has_alpha || !covers_screen;
```
where `format` is defined thus:
```
 	const struct hvs_format *format = vc4_get_hvs_format(fb->format->format);
```

lrusak:
Ah! Sorry, the reason it's like that is because we do not have this upstream change https://github.com/torvalds/linux/commit/124e5dac9a596aa9e8533e46ae6957edbe6758eb

pelwell:
OK. I'll apply the patches once you've squashed in the fixups and added the "commit ... upstream." commit comments.

lrusak:
updated

pelwell:
Thanks - rebased and applied.

PR #2607, opened 2018-07-05
Unicam simplified DT overlays

6by9:
This is the simplified version of #2578 which drops all the I2C pinctrl mux stuff but provides options to adjust the pinctrl setup as required.
I'll come back to #2578 when I have more enthusiasm to fight DT.

### DIFF
--- a/arch/arm/boot/dts/overlays/README
+++ b/arch/arm/boot/dts/overlays/README
@@ -267,6 +267,15 @@ Params: cs                      SPI bus Chip Select (default 1)
         www.kernel.org/doc/Documentation/devicetree/bindings/input/ads7846.txt
 
 
+Name:   adv7282m
+Info:   Analog Devices ADV7282M analogue video to CSI2 bridge.
+        Uses Unicam1, which is the standard camera connector on most Pi
+        variants.
+Load:   dtoverlay=adv7282m,<param>=<val>
+Params: i2c_pins_28_29          Use pins 28&29 for the I2C instead of 44&45.
+                                This is required for Pi B+, 2, 0, and 0W.
+
+
 Name:   akkordion-iqdacplus
 Info:   Configures the Digital Dreamtime Akkordion Music Player (based on the
         OEM IQAudIO DAC+ or DAC Zero module).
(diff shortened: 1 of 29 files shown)


arch/arm/boot/dts/overlays/README
```diff
@@ -267,6 +267,15 @@ Params: cs                      SPI bus Chip Select (default 1)
         www.kernel.org/doc/Documentation/devicetree/bindings/input/ads7846.txt
 
 
+Name:   adv7282m
+Info:   Analog Devices ADV7282M analogue video to CSI2 bridge.
+        Uses Unicam1, which is the standard camera connector on most Pi
+        variants.
+Load:   dtoverlay=adv7282m
```
pelwell: Sorry, but you'll need `,<param>=<val>` here.

arch/arm/boot/dts/overlays/tc358743-overlay.dts
```diff
@@ -0,0 +1,111 @@
...
+		target = <&i2c_vc>;
+		__overlay__ {
+			status = "okay";
+		};
+	};
+
+	__overrides__ {
+		i2c_pins_28_29 = <0>,"+4-5";
+		4lane = <0>, "-2+3";
+	        link-frequency = <&tc358743>,"link-frequencies:0";
```
pelwell: Does this work? The target property is a 64-bit value and FDT is a big-endian format, so you must either use a hack and set the byte offset to 4 (and hope the frequency is never more than 4GHz (!)) or declare it as a 64-bit property using a '#':
```
	        link-frequency = <&tc358743>,"link-frequencies#0";
```

pelwell (changes requested):
See line comments.

arch/arm/boot/dts/overlays/tc358743-overlay.dts
6by9: I thought I'd tried it, but possibly not.
972Mbit/s (486MHz frequency) per lane is the speed we want to use to achieve anything sensible over 2 lanes, but the upstream driver previously only allowed 594Mbit/s (297MHz) so I thought it worth giving the option.
I'll go for the # option.

pelwell (approved):

Issue #2608, opened 2018-07-05
Raspberry Pi 3B+ transferring large amounts of data with the ethernet interface locks up system 


magore:
By locks up I mean - no console response - blank console - no logs written - CPU lan chip and power supply chip stay at  just over 50C as if they are doing something - . red power light no green light
Transfer tests via rsync of 500G of data.
I tried this on several RPI 3B+ bought over the last few month starting in march - all have the problem
Does not happen on RPI3B
Note: Using an external USB Ethernet 1000T adapter or the wifi interface fixes the lockups
Used latest RASPBIAN

Steps to reproduce
1) Install the latest version of RASPBIAN and apply all updates
    If you also try rpi-update the results will be the same regardless of options
2) install openssh server
2) Attach a known working power supply 
    ( I used a 5.1V 3A supply tested with a HP Agilent 34401A meter at the GPIO headers while the PI was under very heavy loads) measure noise with a RIGOL 1054 scope.
3)  Attach a powered USB hard drive with an EXT4 file system
     It does not matter if you used a powered external drive or a power hub and USB drive as the results are the same - tested.
4) mount the USB drive partition on a folder - say /backup
5) Connect PI to 1000T network switch and computer - or directly to the computer with 1000T interface - the resulting crash will still be the same - tested
6)  Use rsync to copy a large amount of data
    I used an Ubuntu 16.04 desktop and rsynced a copy of the complete system to the USB drive on the PI "pi-desktop" - about 500G. Example: ionice -c 3  rsync  --delete -a -H -x -S --numeric-ids  --info=progress2 -delete  --exclude ".gvfs"  /  root@pi-desktop:/backup/ 
Wait for the crash - typically less then 60G of data for this to happen

There were only two conditions where this worked
1) Not using the builtin Ethernet interface and instead using an external USB 1000T Ethernet adapter on the PI for the connection - this finished without errors at about 20 mega bytes per second average
2) using the WIFI connection - this worked without errors at about 5.7 mega bytes per second average
Test was repeated many times over several days

What Always failed -  using the internal interface
   Latest version RASPIAN as of 1 July 2018   - any update method using rpi-update NEXT or defaults 
Settings that did not impact the crashing with internal adapter
   1) ethtool --offload eth0 rx off tx off
        Note - this really slows down the copy process but still dies
   2) ethtool -K eth0 tx-tcp-segmentation off
   3) /boot/config.txt
              dtparam=eee=off
              sdram_freq=450 
[...]

PR #2614, opened 2018-07-10
net: lan78xx: Fix race in tx skb size calculation

lategoodbye:
The skb size calculation in lan78xx_tx_bh is in race with the start_xmit,
which could lead to kernel oopses. So protect the whole skb walk with
a spin lock. As a benefit we can unlink the skb directly.

https://github.com/raspberrypi/linux/issues/2608

### DIFF
--- a/drivers/net/usb/lan78xx.c
+++ b/drivers/net/usb/lan78xx.c
@@ -3292,6 +3292,7 @@ static void lan78xx_tx_bh(struct lan78xx_net *dev)
 	pkt_cnt = 0;
 	count = 0;
 	length = 0;
+	spin_lock_irqsave(&tqp->lock, flags);
 	for (skb = tqp->next; pkt_cnt < tqp->qlen; skb = skb->next) {
 		if (skb_is_gso(skb)) {
 			if (pkt_cnt) {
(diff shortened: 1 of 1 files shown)


JamesH65:
@lategoodbye Good work. Are you going to send this to netdev for further discussion and upstreaming?

lategoodbye:
Yes, but this will happen next week.

pelwell:
Thanks, gents.

lategoodbye:
Yesterday, i sent the [patch](https://patchwork.ozlabs.org/patch/944086/) to netdev.

JamesH65:
Following on from discussion in linked thread, I seem to be seeing odd behaviour with this patch, with the unplugging of the network cable not being detected. Which makes no sense. Investigating.

JamesH65:
Some minor logging changes indicate that the lan78xx_status function is never called  in the failing case., which would indicate that the interrupt over USB (intr_urb) from the lan78xx is being lost somewhere. Looking at logging in the working case, I get that 'interrupt' every time the cable is removed. However, I cannot think of a sequence invvolving that queue walk that could result in this sort of behaviour.

lategoodbye:
I only want to inform you that this patch has been merged upstream and also go to stable.

pelwell:
Thanks, Stefan. If GitHub had a Super Contributor badge you would get one.

Issue #2617, opened 2018-07-12
Ethernet not working until ping is sent from the Pi (Only happens on Pi3b+)


vanfanel:
Hi there, 

I am not using dhcpcd. I am manually setting up the wired networking as I have been doing for years, but I have just moved from Pi3b to Pi3b+ and I found that I can't ping the Pi3b+ from the PC unless I ping the PC from the pi3b+ first. Strange!

Just in case, this is the simple sequence of commands I am using on my personal networking setup script:
```
sudo ip link set eth0 up
sudo ip addr add 192.168.5.8/24 dev eth0
sudo ip route add default via 192.168.5.7
```

Any ideas on what's going on?

maxnet:
Probably because the driver doesn't start polling for incoming packets until an outbound one has been sent.
If there isn't anything you want to do that generates an outbound packet (e.g. synchronize your time with ntp), just arping your gateway IP in your networking setup script as workaround.

Think driver needs a `tasklet_schedule(&dev->bh);` added somewhere after link goes up.
Otherwise the tasklet (that handles both sends and receives) will not be run until something gets sent.

lategoodbye:
@vanfanel  I'm not able to reproduce your issue with Rasbian and Kernel 4.14.54 (downstream) and 4.18rc3 (upstream). 

I added the following to /etc/dhcpcd.conf:
`denyinterfaces eth0`

And this to /etc/network/interfaces:
`iface eth0 inet manual`

Then i bring up the interface manual with the commands like in your script.

So please provide your changes more in detail. Also the output of `ifconfig eth0` while the RPi 3B+ is unreachable would be helpful.

vanfanel:
@maxnet : Thanks to your comment, I may have an idea on what's going on: since most people run Raspbian with services that are sending packets from the Pi3b+ at startup and I don't, this problem may have gone unnoticed to most people.

@lategoodbye : dhcpcd is NOT running on my system. There are no other services sending packets from the Pi. This is my list of services run at startup:

```
pi@raspberrypi:~ $ sudo systemd-analyze blame
          1.060s dev-mmcblk0p2.device
           724ms rc-local.service
           314ms systemd-udevd.service
           301ms systemd-udev-trigger.service
           288ms systemd-remount-fs.service
           177ms fake-hwclock.service
            89ms alsa-restore.service
            65ms sys-kernel-config.mount
            63ms dropbear.service
            40ms systemd-modules-load.service
            37ms boot.mount
[...]

PR #2620, opened 2018-07-16
net: lan78xx: fix rx handling before first packet is send

lategoodbye:
As long the bh tasklet isn't scheduled once, no packet from the rx path
will be handled. Since the tx path also schedule the same tasklet
this situation only persits until the first packet transmission.
So fix this issue by scheduling the tasklet during ndo_open like in usbnet.

https://github.com/raspberrypi/linux/issues/2617


JamesH65:
LGTM

PR #2639, opened 2018-08-07
End log messages in one newline, not two.

gson1703:
Signed-off-by: Andreas Gustafsson <gson@gson.org>


popcornmix:
Looks good to me. I think there may be another one here:
https://github.com/raspberrypi/linux/blob/33b3d2113e42f5a3c98fa05e462bfde786c14173/drivers/staging/vc04_services/interface/vchiq_arm/vchiq_core.c#L3155

popcornmix:
@pelwell okay with you?

pelwell:
Ideally these patches would go upstream, but there are still a few downstream patches (one of which is mine) so it would be hypocritical to refuse it here.

popcornmix:
@gson1703 if you want to help more, then https://www.kernel.org/doc/html/v4.17/process/submitting-patches.html describes the procedure to get these fixed at the source.

But I'll merge these here for now. Thanks for contributing.

PR #2653, opened 2018-08-20
[4.18] add upstream fix for arizona-ldo1, used by Cirrus Logic Audio Card

HiassofT:
An important fix for arizona-ldo1 hasn't made it into upstream 4.18 (yet), I've pinged stable to add it to the 4.18 stable queue
https://lore.kernel.org/lkml/20180820101617.5zzqizi6g7ivnqge@camel2.lan/T/#u

Pick the commit locally until it's processed upstream, without the fix the Wolfson/Cirrus cards won't work

### DIFF
--- a/drivers/regulator/arizona-ldo1.c
+++ b/drivers/regulator/arizona-ldo1.c
@@ -36,6 +36,8 @@ struct arizona_ldo1 {
 
 	struct regulator_consumer_supply supply;
 	struct regulator_init_data init_data;
+
+	struct gpio_desc *ena_gpiod;
 };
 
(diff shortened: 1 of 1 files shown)


HiassofT:
Just a quick heads-up: GregKH has added that commit to his stable queue (regulator-arizona-ldo1-use-correct-device-to-get-enable-gpio.patch) and it should appear in the next stable update

https://git.kernel.org/pub/scm/linux/kernel/git/stable/stable-queue.git/tree/queue-4.18

pelwell:
Thanks, HiassofT.

PR #2656, opened 2018-08-23
dtoverlays: Add support for ADV7280-M and ADV7281-M chips.

6by9:
The driver that supports the ADV7282-M also supports the ADV7280-M and
ADV7281-M.
The 7280 exposes 2 extra analogue inputs. The 7281 doesn't have the
I2P deinterlacing block. Otherwise they are the same as ADV7282-M.

Rename the existing overlay to adv728x-m.
Add a backwards compatibility overlay for adv7282m that includes adv728x-m.
Add parameters to alter the compatible string for the 7280 and 7281.

### DIFF
--- a/arch/arm/boot/dts/overlays/adv728x-m-overlay.dts
+++ b/arch/arm/boot/dts/overlays/adv728x-m-overlay.dts
@@ -1,0 +1,36 @@
+// Definitions for Analog Devices ADV728[0|1|2]-M video to CSI2 bridges on VC
+// I2C bus
+
+#include "adv7282m-overlay.dts"
+
+/{
+	compatible = "brcm,bcm2708";
+
+	// Fragment numbers deliberately high to avoid conflicts with the
+	// included adv7282m overlay file.
+
+	fragment@101 {
+		target = <&adv728x>;
+		__dormant__ {
+			compatible = "adi,adv7280-m";
+		};
+	};
+	fragment@102 {
+		target = <&adv728x>;
+		__dormant__ {
+			compatible = "adi,adv7281-m";
+		};
+	};
+	fragment@103 {
+		target = <&adv728x>;
+		__dormant__ {
+			compatible = "adi,adv7281-ma";
+		};
+	};
+
+	__overrides__ {
+		adv7280m = <0>, "+101";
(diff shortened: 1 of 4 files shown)


pelwell:
The overlay checking script has a few objections:
* The new "Load:" example has the wrong overlay name, which should be fixed.
* The old overlay doesn't include the documentation for the new parameters.

I could tell the checker to ignore those parameters for that overlay, or you could refactor so that adv728x-m is a specialisation of adv7282m with the additional parameters and associated __dormant__ fragments (I don't think it is necessary to put the default value in its own fragment).

6by9:
OK, so invert it so "adv728x-m" includes "7282m", and adds the dormant params to override the compatible string - sounds like a plan.

Do fragment numbers have to be contiguous? I'm just thinking if 7282 has fragments 0-4, does 728x-m have to start at 5?
If yes, then should 7282 ever gain an extra fragment 728x-m would need to be renumbered as well.
If no, then 728x-m can have fragments 101 & 102 to avoid collisions.

pelwell:
> Do fragment numbers have to be contiguous?

Although I've encouraged people to keep them contiguous they really only have to be unique.

6by9:
> Although I've encouraged people to keep them contiguous they really only have to be unique.

I'll make a comment in adv748x-m as to why they are odd.

6by9:
I think I'm done.
I've run your tool and it seems happy enough, and it seems to have the right effect on the device.

Whilst tinkering I've also added a parameter to set the I2C address (until I checked the datasheet for the differences between the chips I had forgotten that it could be configured).

arch/arm/boot/dts/overlays/adv728x-m-overlay.dts
```diff
@@ -0,0 +1,38 @@
+// Definitions for Analog Devices ADV728[0|1|2]-M video to CSI2 bridges on VC
+// I2C bus
+/dts-v1/;
+/plugin/;
```
pelwell: Can you omit these two tags? They shouldn't really be duplicated, which is what happens when you #include.

6by9: Done.

PR #2663, opened 2018-08-26
Adding audioinjector.net ultra sound card support

flatmax:
Contains the sound/soc/bcm ALSA machine driver and necessary alterations to
the Kconfig and Makefile.
Adds the dts overlay and updates the Makefile and README.
Updates the relevant defconfig files to enable building for the Raspberry Pi.


HiassofT:
The sound card driver contains just boilerplate code and can be replaced by the simple-audio-card driver.
See https://github.com/raspberrypi/linux/blob/rpi-4.14.y/Documentation/devicetree/bindings/sound/simple-card.txt for more info and look at the various other DT overlays using simple-audio-card.

Hints: use the tdm-slot-num/width DT properties as a replacement for the set_bclk_ratio call in init and set the codec sysclock with the system-clock-frequency DT property

flatmax:
OK,

I can implement this.

The codec also has to be compiled as part of this. What is the best way 
to ensure the codec is built ?

Currently this is done with the "select SND_SOC_CS4265" in the Kconfig file.

Matt

pelwell:
Add SND_SOC_CS4265 to the defconfigs instead.

flatmax:
Closing this pull request to switch over to the overlay simple card approach.

Issue #2643, opened 2018-08-10
[4.18] corrupted tvservice -m mode list


HiassofT:
With the current rpi-4.18.y kernel and latest rpi-update firmware (version 7e7f762267acd3e85aa3b72724f969b40fa4179c, Aug  8 2018 16:02:15) tvservice -m displays a corrupted list of modes. Several modes, resolutions and clocks show up as 0. This is especially bad as the preferred / current mode is missing.

eg when connected to a HP1755 (1280x1024, only DMT, no CEA modes)

on kernel 4.18
```
Group DMT has 12 modes:
           mode 4: 640x480 @ 60Hz 4:3, clock:25MHz progressive
           mode 5: 640x480 @ 72Hz 4:3, clock:31MHz progressive
           mode 6: 640x480 @ 75Hz 4:3, clock:31MHz progressive
           mode 9: 800x600 @ 60Hz 4:3, clock:40MHz progressive
           mode 10: 800x600 @ 72Hz 4:3, clock:50MHz progressive
           mode 11: 800x600 @ 75Hz 4:3, clock:49MHz progressive
           mode 16: 1024x768 @ 60Hz 4:3, clock:65MHz progressive
           mode 17: 1024x768 @ 70Hz 4:3, clock:75MHz progressive
           mode 18: 1024x768 @ 75Hz 4:3, clock:78MHz progressive
           mode 21: 1152x864 @ 75Hz 4:3, clock:0MHz progressive
           mode 0: 0x0 @ 0Hz unknown AR, clock:0MHz progressive
           mode 0: 0x0 @ 0Hz unknown AR, clock:0MHz progressive
```

on kernel 4.14  (current rpi-update kernel, same fw version as above) the output is fine
```
Group DMT has 12 modes:
           mode 4: 640x480 @ 60Hz 4:3, clock:25MHz progressive
           mode 5: 640x480 @ 72Hz 4:3, clock:31MHz progressive
           mode 6: 640x480 @ 75Hz 4:3, clock:31MHz progressive
           mode 9: 800x600 @ 60Hz 4:3, clock:40MHz progressive
           mode 10: 800x600 @ 72Hz 4:3, clock:50MHz progressive
           mode 11: 800x600 @ 75Hz 4:3, clock:49MHz progressive
           mode 16: 1024x768 @ 60Hz 4:3, clock:65MHz progressive
           mode 17: 1024x768 @ 70Hz 4:3, clock:75MHz progressive
[...]

PR #2666, opened 2018-08-28
Vchiq cache

pelwell:
This series of commits gives the vchiq driver the ability to discern the
correct cache_line_size value in the absence of the cache-line-size
property. The correction of the reg size value doubles as a hint to
the firmware that the kernel includes the updated value.

See: https://github.com/raspberrypi/linux/issues/2643


pelwell:
A preview firmware release that implements the new logic can be downloaded here: https://drive.google.com/file/d/13c66cJkNmCQFMd7_f3GNA_tk5usoK0ZW/view?usp=sharing

HiassofT:
Thanks a lot!

Testing on RPi3B+ with the preview firmware worked fine - tvservice -m showed correct lists of modes.

When using the current official firmware I got a corrupted mode list with this PR - but I guess this is to be expected?

pelwell:
Yes it is. The current firmware assumes that a lack of `cache-line-size` property (strictly, a lack of `cache_line_size` parameter) means it is a very old BCM2835 kernel and therefore requires a 32 byte cache line. I could improve backwards compatibility on a downstream kernel/DTB by retaining a `cache_line_size` parameter that does nothing...

pelwell:
The DT commit has been updated to retain a vestigial `cache_line_size` parameter, which should keep old firmware happy.

HiassofT:
thanks, the updated version works fine with the August 24 firmware

mkreisl:
Confirmed, 4.18.5 works with older firmware (Jul 23)

PR #2674, opened 2018-09-06
rpi-simple-soundcard: Use nicer driver name "RPi-simple"

timg236:
Rename the driver from "RPI simple soundcard" to "RPi-simple" so that
the driver name won't be mangled allowing to be used unaltered as the
card conf filename.


timg236:
@HiassofT

HiassofT:
looks good to me, thanks a lot!

PR #2685, opened 2018-09-18
backport bcm2835-audio

lategoodbye:
This pull request contains all upstream bcm2835-audio patches incl. the recent patch series by Takashi Iwai (currently only in linux-next).

I reverted commit f6da082a790f857d5163ac99357beae6b59f1eb5 ("bcm2835: interpolate audio delay") in order to apply the patches without rebase and it seems to me that "staging: bcm2835-audio: Add 10ms period constraint" is the proper one. If i'm wrong i will reapply it.


pelwell:
Please explain why reverting the audio delay interpolation patch doesn't cause a regression in playback position determination.

lategoodbye:
As i wrote above i don't know. This commit was never send upstream and i don't know the test scenario.

pelwell:
What you wrote was ambiguous, but I can see how that may have been the message you were intending to convey.

The onus is on the submitter of the PR to ensure that it doesn't break anything, and saying "I didn't understand it so I reverted it" is not sufficient. To my eyes the patch it is fairly clear what the patch is doing, so rebase and fix it up as necessary so that it looks the same, then the original issue reporter (@mikebrady) can test that his use case still works as expected.

lategoodbye:
Only to make it clear, i didn't revert the patch because i didn't understand it. I revert it because of two reason:
- i wanted to avoid any changes on the upstream patches
- i had the hope the mentioned patch is now obsolete

In case this patch is important than it should be upstreamed first.

pelwell:
I think that if you don't know why reverting it will cause a regression then you don't understand it.

You can decide which is more important to you - upstreaming the interpolation patch or getting the other patches into rpi-4.14.y.

lategoodbye:
Okay

mikebrady:
Hi there. Could you explain what the situation with that patch is now please? I really hope is will be included, as it makes it possible to get an accurate estimate of the "time_to_hear" using [Safe ALSA](http://0pointer.de/blog/projects/guide-to-sound-apis), bringing it into line with USB and I2S DACs. Reverting to an estimate with a 10 ms granularity would be retrograde.

pelwell:
The original patch is still in place. @lategoodbye has presumably closed this PR while the interpolation patch is upstreamed, at which point it will be included in a new PR.

mikebrady:
FYI, [this](https://github.com/raspberrypi/firmware/issues/1026) refers.

mikebrady:
Great, thanks.

mikebrady:
Okay, thanks.

lategoodbye:
@mikebrady Thanks for your quick response. Takashi made this upstream changes without knowing about your fix. I would be really happy, if you can upstream your changes.

I already forwarded your downstream patch to Takashi, but he has some [comments](https://lkml.org/lkml/2018/9/19/384). Do you want to join in (i need your email address)?

popcornmix:
The "bcm2835: interpolate audio delay" patch was originally written by @wm4 a long time ago (but not merged then).
@mikebrady tested it as he was having an issue that looked to be caused by the coarse latency values returned by driver and I remembered the interpolate commit.

mikebrady:
Thanks @lategoodbye, I've emailed you directly.

PR #2686, opened 2018-09-18
OV5647 kernel driver/overlay tweaks

6by9:
It seems that I asked for an old set of DT changes to be merged for setting up the camera GPIOs resulting in it not working with the OV5647 where we need to power up the sensor (TC358743 and ADV7282 don't bother).
I've amended the DT patches appropriately and confirmed it works on a 3B+. On doing that I found that the kernel objected because the driver was calling gpiod_set_value on a GPIO controller that can sleep. There's therefore a driver patch to call gpiod_set_value_cansleep instead (which probably ought to be the default for all media drivers in this use case).

https://www.raspberrypi.org/forums/viewtopic.php?f=29&t=222851

### DIFF
--- a/arch/arm/boot/dts/bcm270x.dtsi
+++ b/arch/arm/boot/dts/bcm270x.dtsi
@@ -152,6 +152,13 @@
 		regulator-max-microvolt = <3300000>;
 		regulator-always-on;
 	};
+
+	__overrides__ {
+		cam0-pwdn-ctrl;
+		cam0-pwdn;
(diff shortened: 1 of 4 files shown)


pelwell:
Since `cam0-led` and `cam0-pwdn` are no longer user-accessible parameters, can you remove them from the README?

6by9:
> Since cam0-led and cam0-pwdn are no longer user-accessible parameters, can you remove them from the README?

Done.

PR #2695, opened 2018-09-26
Read without -r will mangle backslashes. (Fix)

mrmajik45:
By default, read will interpret backslashes before spaces and line feeds, and otherwise strip them. This is rarely expected or desired.

Normally you just want to read data, which is what read -r does. You should always use -r unless you have a good reason not to.   ( I like Raspberry Pi's so I want to help :P hello! )


pelwell:
These are all unmodified upstream files. If you feel strongly about this, take it up with the upstream devs, but I don't fancy your chances of persuading them.

PR #2700, opened 2018-10-02
thermal: add hysteresis support and adjust PoE HAT trip points

XECDesign:
Currently, the step_wise thermal scaling governor ping pongs around the trip point and ignores the hysteresis value.

This PR steals step_wise hysteresis support from https://patchwork.kernel.org/patch/10384571/. This still results in some bounce at the hysteresis temperature if the 'trend' is rising, which is addressed by the second commit.

In the PoE HAT overlay, a cooling level has been removed and hysteresis values were increased to 5C.

### DIFF
--- a/arch/arm/boot/dts/overlays/README
+++ b/arch/arm/boot/dts/overlays/README
@@ -1596,9 +1596,16 @@ Params: touchscreen-size-x      Touchscreen X resolution (default 800)
 
 
 Name:   rpi-poe
-Info:   Raspberry Pi POE HAT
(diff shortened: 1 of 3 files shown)


pelwell:
For the curious, this overlay is adding its parameters to the base DTB; without this trick, they would go out of scope when the overlay for a stacked HAT is loaded, meaning they could never be used.

pelwell:
The README entry now requires  `,<param>=<val>` in the `Load:` line, but otherwise it looks ready to merge.

PR #2397, opened 2018-02-24
firmware/raspberrypi: Add a get_throttled sysfs file

notro:
This is a result of the discussion in #2367.

Kernel log:
```
[  530.474085] Under-voltage detected! (0x00050005)
[  534.634106] Voltage normalised (0x00000000)
```

A simple script that polls on the sysfs file and shows the changes:
```python
import select
import datetime

epoll = select.epoll()

file = open("/sys/devices/platform/soc/soc:firmware/get_throttled")
epoll.register(file.fileno(), select.EPOLLPRI | select.EPOLLERR)
status = file.read()

while(True):
    epoll.poll()
    file.seek(0)
    status = file.read()
    get_throttled = int(status, 16)
    print("%s 0x%08x" % (datetime.datetime.now().strftime('%H:%M:%S'), get_throttled))

epoll.unregister(file.fileno())
file.close()

```

A simple proof of concept for desktop notification:
```python
import dbus
import select
import datetime
import signal

msg = """Under-voltage is blah blah not good.
See: http://some-explanation.org
"""

def quit(signo, _frame):
    exit(0)

for sig in ('TERM', 'HUP', 'INT'):
    signal.signal(getattr(signal, 'SIG'+sig), quit);

bus = dbus.SessionBus()
notifier = bus.get_object('org.freedesktop.Notifications', '/org/freedesktop/Notifications')
iface = dbus.Interface(notifier, 'org.freedesktop.Notifications')
nid = None

epoll = select.epoll()
file = open("/sys/devices/platform/soc/soc:firmware/get_throttled")
epoll.register(file.fileno(), select.EPOLLPRI | select.EPOLLERR)
status = file.read()

while(True):
    epoll.poll()
    file.seek(0)
    status = file.read()
    get_throttled = int(status, 16)

    if get_throttled & (1 << 16):
        try:
            nid = iface.Notify('get_throttled', 0, 'dialog-warning', 'Under-voltage!', msg, [], {}, 0)
        except dbus.exceptions.DBusException as e:
            sys.stderr.write("Error sending notification: %s\n" % e); sys.stderr.flush()
    elif nid:
        try: # The user might have closed it
            iface.CloseNotification(nid)
        except dbus.exceptions.DBusException:
            pass
        nid = None

epoll.unregister(file.fileno())
file.close()

```

dbus notification daemon (supports cancelling notifications):
```
sudo apt install dunst
```

PR #2706, opened 2018-10-06
Rpi 4.19.y RFC Enable Raspberry Pi voltage monitor

lategoodbye:
@notro [pull request](https://github.com/raspberrypi/linux/pull/2397) has been upstreamed as the Raspberry Pi voltage monitor within the hwmon subsystem and is available since 4.19.

In order avoid having two driver doing the same thing, this pull requests tries to push the Raspberry Kernel more into upstream direction.

"Old" get_trottled sysfs entry:
`/sys/devices/platform/soc/soc:firmware/get_trottled`

"New" hwmon entry:
`/sys/class/hwmon/hwmon0/in0_lcrit_alarm`

Documentation:
https://github.com/raspberrypi/linux/blob/rpi-4.19.y/Documentation/hwmon/raspberrypi-hwmon


notro:
@lategoodbye thanks for taking this upstream.

This PR needs rebasing, so in the meanwhile here's the link to the commits:
https://github.com/lategoodbye/linux/commits/rpi-4.19.y-hwmon

lategoodbye:
@notro The reason for this pull request is to discuss if the userspace ABI break (different path and different value) is acceptable for Raspbian. I would rebase after final 4.19 is out.

popcornmix:
Is it feasible to switch to the upstream driver and provide a downstream patch that additionally creates ``/sys/devices/platform/soc/soc:firmware/get_throttled``
We warn users that the old node is deprecated and will be removed in a future update.

lategoodbye:
From my understanding the hwmon driver cannot simply create such a file, because the sysfs entries are always relative.

Edit: Okay, i think of a hacky solution with a global variable or to extract the response for the get_throttled within the firmware driver.

notro:
I see now that there's a difference between downstream and mainline in that downstream returns the whole throttled value, wereas mainline only returns the undervoltage bit.

Another difference is that ->last_throttled is set during probe. AFAIU this means that the undervoltage warning won't be reported if undervoltage happened before hwmon probe. The voltage normalise message will show though.

lategoodbye:
> I see now that there's a difference between downstream and mainline in that downstream returns the whole throttled value, wereas mainline only returns the undervoltage bit.

Per hwmon sysfs `in0_lcrit_alarm` must be boolean. In case we need the other bits, we need to look for a matching candidate(s).

> Another difference is that ->last_throttled is set during probe. AFAIU this means that the undervoltage warning won't be reported if undervoltage happened before hwmon probe. The voltage normalise message will show though.

Nice catch. I could add a warning into the probe function.

lategoodbye:
Okay, i added a patch which "restore" the old get_throttled.

popcornmix:
@lategoodbye thanks for this. The backward compatible node seems to work well (the WARN backtrace will probably generate some worried reports, but should identify any significant users of the old node).

I've picked it to rpi-4.19 (dues to conflict after the bump to v4.19) for further exposure (should be an rpi-update kernel with this in very soon).

Issue #2713, opened 2018-10-12
 4.14.74-v7+ 3b+ VLAN udp IPv4/6 - hw csum failure 


Wireheadbe:
Hi All - seems like issue https://github.com/raspberrypi/linux/issues/2458 is popping up again.

Setup is eth0 on local lan; eth0.30 with tagged data. Running IPv6 as well. (both over eth0.30 as tun0 for openvpn)
Having OpenVPN traffic comming in over eth0.30 cause lots of these errors:
```
[  862.694409] eth0: hw csum failure
[  862.694452] CPU: 0 PID: 0 Comm: swapper/0 Tainted: G         C      4.14.74-v7+ #1149
[  862.694458] Hardware name: BCM2835
[  862.694495] [<8010ffd4>] (unwind_backtrace) from [<8010c240>] (show_stack+0x20/0x24)
[  862.694514] [<8010c240>] (show_stack) from [<80788604>] (dump_stack+0xd4/0x118)
[  862.694537] [<80788604>] (dump_stack) from [<8068e870>] (netdev_rx_csum_fault+0x44/0x48)
[  862.694567] [<8068e870>] (netdev_rx_csum_fault) from [<80681144>] (__skb_checksum_complete+0xbc/0xc0)
[  862.694587] [<80681144>] (__skb_checksum_complete) from [<807331c8>] (nf_ip_checksum+0xd4/0x130)
[  862.694704] [<807331c8>] (nf_ip_checksum) from [<7f558a4c>] (udp_error+0x138/0x1c8 [nf_conntrack])
[  862.694894] [<7f558a4c>] (udp_error [nf_conntrack]) from [<7f551994>] (nf_conntrack_in+0xec/0x560 [nf_conntrack])
[  862.695004] [<7f551994>] (nf_conntrack_in [nf_conntrack]) from [<7f59f2dc>] (ipv4_conntrack_in+0x28/0x2c [nf_conntrack_ipv4])
[  862.695031] [<7f59f2dc>] (ipv4_conntrack_in [nf_conntrack_ipv4]) from [<806d04dc>] (nf_hook_slow+0x4c/0xd0)
[  862.695056] [<806d04dc>] (nf_hook_slow) from [<806d8df4>] (ip_rcv+0x460/0x514)
[  862.695075] [<806d8df4>] (ip_rcv) from [<8068be4c>] (__netif_receive_skb_core+0x340/0xc84)
[...]

Issue #2712, opened 2018-10-11
SMSC9514 eth0: hw csum failure - NOT Pi3b+


6by9:
Raised to avoid confusion with #2659.

Report from @ricardosalveti of seeing ```eth0: hw csum failure``` messages from the kernel on 3B (ie SMSC9514) with 4.18 kernel.

> lso started to see this issue lately, but I'm using the 3 model B, which uses smsc95xx.
This is with the latest 4.18 kernel tree (rpi patches + stable).

https://github.com/raspberrypi/linux/issues/2659#issuecomment-428716193
https://github.com/raspberrypi/linux/issues/2659#issuecomment-428735551
are the relevant comments.

New issue raised to avoid confusing lan78xx issues vs smsc95xx issues.

6by9:
> Looking at my local network, the service generating this issue was iceccd (running on ubuntu 16.04).
> After disabling iceccd on the ubuntu machine I'm not not getting the hw csum errors anymore.
> Logs from tcpdump:

```
21:12:03.771275 IP (tos 0x0, ttl 64, id 57986, offset 0, flags [DF], proto UDP (17), length 29)
    192.168.1.44.58909 > 192.168.1.255.8765: UDP, length 1
E.....@.@......,......"=.       S. ............../..
```
> Unclear if this is really an issue at the sender or if an issue at the receiver side.

A UDP packet of length 29 would fall into the category of the UDP checksum being in the last 4 words of the packet, however the patch that had been posted was for checksum offload failure on transmit, not receive.
I'll see if I can find the relevant reference docs for SMSC9514 as to the details on that offload issue.

6by9:
Closing as this does seem to be a common issue in the main IP core and not a driver related thing.

There is a check in the SMSC9514 driver that s/w checks buffers under 45 bytes, so I'd expect this to be trapped there.

Issue #2659, opened 2018-08-24
kernel 4.18.4 eth0: hw csum failure - Pi3b+ only


polojoe:
I get following errors
`eth0: hw csum failure`
since upgrade to 4.18 from 4.14

kernel log http://ix.io/1l8g

rpi is connected directly to my isp router

popcornmix:
How repeatable is the error? Do you see it every boot? Always after playing a video?
Can you try running:
```ethtool -K eth0 rx off ; ethtool -K eth0 tx off```

and let us know if problem still occurs?
Ping @6by9

**Comment by polojoe** [comment]
errors start repeatable several minutes after boot.
`ethtool -K eth0 rx off ; ethtool -K eth0 tx off`
solves errors in kernel log

**Comment by popcornmix** [comment]
Can you test just ``ethtool -K eth0 rx off `` and just ``ethtool -K eth0 tx off `` to see which is fixing the error?

**Comment by polojoe** [comment]
`ethtool -K eth0 rx off`
does the trick

**Comment by 6by9** [comment]
I assume this is on a 3B+.
Are you running VLANs? That was the only time we were previously seeing these errors, and there was a patch that swapped to s/w checksum offload in those conditions.

**Comment by polojoe** [comment]
I'm using rpi3+, no vlan configured.
Direct lan connection to router with integrated 4 port 100 Mbit switch.

**Comment by maxnet** [comment]
>I'm using rpi3+, no vlan configured.

```
[    4.827651] 8021q: 802.1Q VLAN Support v1.8
[    4.827687] 8021q: adding VLAN 0 to HW filter on device eth0
```

Module normally only gets loaded when you do configure VLAN tagging.
Or do you have a custom kernel configuration in which it is statically built into kernel?

**Comment by MilhouseVH** [comment]
>  Or do you have a custom kernel configuration in which it is statically built into kernel?

In LibreELEC 9 with rpi-4.18.y, the config from https://github.com/LibreELEC/LibreELEC.tv/pull/2913 is being used.

A quick grep of the config for VLAN shows the following:
```
neil@nm-linux:~/projects/pullrequest_repos/LibreELEC.tv/projects/RPi/devices/RPi2$ git grep VLAN
linux/linux.arm.conf:# CONFIG_BRIDGE_VLAN_FILTERING is not set
linux/linux.arm.conf:CONFIG_VLAN_8021Q=m
linux/linux.arm.conf:# CONFIG_VLAN_8021Q_GVRP is not set
[...]

PR #2716, opened 2018-10-15
Revert "net: pskb_trim_rcsum() and CHECKSUM_COMPLETE are friends"

6by9:
This reverts commit 88078d98d1bb085d72af8437707279e203524fa5.

Various people have been reporting seeing "eth0: hw csum failure"
and callstacks dumped in the kernel log on 4.18, and since 4.14.71,
on both SMSC9514 and LAN7800 adapters.
This commit appears to be the reason, but potentially due to an
issue further down the stack. Revert whilst investigating the
trigger.

https://github.com/raspberrypi/linux/issues/2713
https://github.com/raspberrypi/linux/issues/2659
https://github.com/raspberrypi/linux/issues/2712


6by9:
@pelwell revert as discussed.

Do we want to defer merging this one until we have a way to reproduce? Fixing 4.14 needs to be a priority as that is our supported kernel, but 4.18 isn't so may lead to someone giving us guaranteed steps to reproduce the issue and investigate.

pelwell:
If we're realistically expecting 4.18 to be used by a significant number of users it should be released to the `next` branch. What do you think, @popcornmix ?

6by9:
James has found https://patchwork.ozlabs.org/patch/982642/ which might be a candidate for a potential fix on 4.18 instead of revert.

6by9:
#2718 is a cherry-pick of the mailing list patch applied to 4.18 as an alternative to this revert.
Building/testing it now to see if it works generally even if we can't reproduce the failure.

6by9:
Seeing as popcornmix has merged the net-dev patch, I'll close this for now.

PR #2717, opened 2018-10-15
Revert "net: pskb_trim_rcsum() and CHECKSUM_COMPLETE are friends"

6by9:
This reverts commit 6bf32cda46ebfbaf13da3c48a0a009adae925703.

Various people have been reporting seeing "eth0: hw csum failure"
and callstacks dumped in the kernel log on 4.18, and since 4.14.71,
on both SMSC9514 and LAN7800 adapters.
This commit appears to be the reason, but potentially due to an
issue further down the stack. Revert whilst investigating the
trigger.

https://github.com/raspberrypi/linux/issues/2713
https://github.com/raspberrypi/linux/issues/2659
https://github.com/raspberrypi/linux/issues/2712

### DIFF
--- a/include/linux/skbuff.h
+++ b/include/linux/skbuff.h
@@ -3139,7 +3139,6 @@ static inline void *skb_push_rcsum(struct sk_buff *skb, unsigned int len)
 	return skb->data;
 }
 
-int pskb_trim_rcsum_slow(struct sk_buff *skb, unsigned int len);
 /**
  *	pskb_trim_rcsum - trim received skb and update checksum
  *	@skb: buffer to trim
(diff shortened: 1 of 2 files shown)


6by9:
@pelwell revert for 4.14 as discussed.

jens-maus:
@pelwell Now that other projects (see https://github.com/jens-maus/RaspberryMatic/issues/449) are facing the same issue, when will you release a new official `raspberrypi-kernel_1.2018XXXX-1` version (including corresponding Pi-firmware) these projects can pick up?

popcornmix:
ping @XECDesign who deals with stable apt releases.
The upstream fix is on 4.19 kernel tree but hasn't reached 4.14. I suspect it will appear on next minor kernel bump (probably this week), and that may be a good point to bump the apt kernel.

popcornmix:
Latest rpi-update kernel removes this revert and makes use of the upstream fix.

jens-maus:
@popcornmix Then where are the proper tagged version in raspberrypi/linux and raspberrypi/firmware? I can only see the 20180924 versions there.

popcornmix:
@jens-maus Tagged versions won't appear until kernel/firmware is updated in apt.
This is still in rpi-update testing. If we get positive reports the apt version will be updated.

Issue #932, opened 2015-04-17
watchdog: Report reason for last reset


julianscheel:
Would it be possible to properly report if a watchdog timeout caused the last reset or a normal powercycle? Or maybe even detect resets due to voltage dropout, etc?

I had hoped the PM_RSTS register described in the watchdog section of `bcm2708/include/mach/platform.h` would be helpful. Reading return `0x0` only for me, though. And I couldn't find any documentation on that register.

Maybe someone can provide details?

popcornmix:
PM_RSTS is reset when read. But we do cache the original version away on GPU side.

In one of my scripts I so this:

```
 # check if we rebooted through good or bad path
 RSTS=`vcgencmd get_rsts`
 RSTS=${RSTS: -3}
 # 1000 is power on reset, so treat an run again
 if [ "$RSTS" == "1000" ]; then
  POWERON=1
 elif [ "$RSTS" == "574" ]; then
  PASS=1
 else
  PASS=0
 fi
```

julianscheel:
Ah nice, then one more question: Is there some documentation on the possible RSTS values?
Some are understandable from your snippet and the platform.h (like HADPOR), but not all...

nbrachet:
Wouldn't it be better to use:

```
RSTS=`vcgencmd get_rsts`
RSTS=${RSTS#*=}
```

`${RSTS: -3}` will only select the last 3 charaters won't it? And so RSTS wouldn't be equal to 1000 but to 000.

popcornmix:
PM_RSTC

```
31-24 PASSWD Power Manager password “5a” W
23-13 - Unused R
21-20 HRCFG Hard reset configuration R/W
Not used on BCM2708A0.
19-18 - Unused R
17-16 FRCFG Full reset configuration R/W
Not used on BCM2708A0.
15-14 - Unused R
13-12 QRCFG Quick reset configuration R/W
0 = do not reset PLLs
1 = reset PLLs
2 = do not reset PLLs
3 = reset PLLs
11-10 - Unused R
9-8 SRCFG Software reset configuration R/W
0 = no reset
1 = quick reset
2 = full reset
3 = hard reset
7-6 - Unused R
5-4 WRCFG Watchdog reset configuration R/W
0 = no reset
1 = quick reset
2 = full reset
3 = hard reset
3-2 - Unused R
1-0 DRCFG Debugger reset configuration R/W
0 = no reset
1 = quick reset
2 = full reset
3 = hard reset
```

[...]

PR #2719, opened 2018-10-15
watchdog: bcm2835: make normal restart detectable

AmsSoftwareDept:
Currently PM_RSTS gives the same value for a normal restart as it does for a watchdog lapse (as the restart routine appears to use a watchdog lapse). If we have a watchdog monitored process that crashes, we would like to be able to log the incident at next startup. PM_RSTS Bit 9 is described [in issue 932 as "9 HADSRF Had a software full reset R/W"](https://github.com/raspberrypi/linux/issues/932#issuecomment-93989581) which seems appropriate, and it does not appear to be currently used. The behavior of currently used bits have not been affected, so it should be backwards compatible with anything that might be checking for those bit positions. Any comments, feedback, alternatives, or suggestions are welcome. Thank you for your time.

Example of checking PM_RSTS via script:
```bash
#!/bin/bash
RSTS="$(vcgencmd get_rsts)"
RSTS=${RSTS#*=}
RSTS=$(("0x$RSTS"))
STR="?"
# order is important as multiple bits can be set
if ((RSTS & 0x200)); then
    # software reset
    STR="SoftwareReset"
elif ((RSTS & 0x20)); then
    # dog trip (or software reset if before this patch)
    STR="DogReset"
elif ((RSTS & 0x1000)); then
    # power cycle
    STR="PowerCycle"
fi
printf "RSTS=0x%08X CAUSE=%s\n" $RSTS $STR
```
(This was tested with a yocto (poky) core-image-base build on the rocko branch. We have not tested it in any other distros. Test hardware was the Pi3 B+.)

### DIFF
--- a/drivers/watchdog/bcm2835_wdt.c
+++ b/drivers/watchdog/bcm2835_wdt.c
@@ -115,6 +115,14 @@ static void bcm2835_restart(struct bcm2835_wdt *wdt, u8 partition)
 {
 	u32 val, rsts;
 
+	/* Allow code to distinguish a normal reboot from a watchdog reboot
+	 * by setting a magic partition number.
+	 * A partition number higher than the number of partitions on sdcard
+	 * will result in a normal reboot.
+	 */
+	if (partition == 0)
+		partition = 62;
+
 	rsts = (partition & BIT(0)) | ((partition & BIT(1)) << 1) |
 	       ((partition & BIT(2)) << 2) | ((partition & BIT(3)) << 3) |
 	       ((partition & BIT(4)) << 4) | ((partition & BIT(5)) << 5);


lategoodbye:
Here my comments to the commit:

- change to the subject to something like "watchdog: bcm2835: make normal restart detectable"
- add commit log which explain the reason for this commit especially why manipulating PM_RSTS instead of PM_RSTC
- add Signed-off-by
- try to fix linux/scripts/checkpatch.pl issues

Thanks

popcornmix:
It looks like 6 bits (bits 0,2,4,6,8) are used for the partition to reboot off, with partition 63 treated as HALT (don't continue booting from bootcode.bin).
Other high partition numbers could also be used for passing information as a normal boot will occur if the partition number conveyed does not exist.
``5 HADWRF Had a watchdog full reset`` will be set by both a watchdog reboot or kernel reboot (which also uses the watchdog to enable the reset).
That leaves ``1 HADDRF Had a debugger full reset`` and ``9 HADSRF Had a software full reset`` as currently unused. 
Now these nine bits are the only peripheral registers in the whole chip that can convey information over a reboot, so they are a very scarce resource.

I wonder if we couldn't achieve the same effect without sacrificing another bit by using a new magic partition number. e.g. something like:
```
diff --git a/drivers/watchdog/bcm2835_wdt.c b/drivers/watchdog/bcm2835_wdt.c
index 9699e07119a7..705bf2c35007 100644
--- a/drivers/watchdog/bcm2835_wdt.c
+++ b/drivers/watchdog/bcm2835_wdt.c
@@ -98,6 +98,13 @@ static void __bcm2835_restart(struct bcm2835_wdt *wdt, u8 partition)
 {
        u32 val, rsts;
 
+       /* Allow code to distinguish a normal reboot from a watchdog reboot
+        * by setting a magic partition number.
+        * A partition number higher than the number of partitions on sdcard
+        * will result in a normal reboot */
+       if (partition == 0)
+               partition = 62;
+
        rsts = (partition & BIT(0)) | ((partition & BIT(1)) << 1) |
               ((partition & BIT(2)) << 2) | ((partition & BIT(3)) << 3) |
               ((partition & BIT(4)) << 4) | ((partition & BIT(5)) << 5);
```

pelwell:
Yes - I prefer that approach. An unexpected watchdog reset would never set the partition number, and nobody has 62 partitions, so this is a better way of distinguishing a reboot from a watchdog reset.

lategoodbye:
@AmsSoftwareDept Sorry, i was a little bit unclear about the checkpatch.pl. Please don't fix old style issues because there are not relevant to the patch. Only styles issues of your already commited patch:

```
git format-patch -1
scripts/checkpatch.pl your.patch
```

Definition of [Signed-off-by](https://github.com/wking/signed-off-by/blob/master/Documentation/SubmittingPatches)

AmsSoftwareDept:
So far quick tests of patch proposed above https://github.com/raspberrypi/linux/pull/2719#issuecomment-430267340 seem to work fine tested with the modified script below. I assume that any non-zero partition number would have to come from a normal restart (or at least not from a watchdog lapse caused by a stalled process).

```bash
#!/bin/bash
RSTS="$(vcgencmd get_rsts)"
RSTS=${RSTS#*=}
RSTS=$(("0x$RSTS"))
STR="?"
# order is important as multiple bits can be set
if ((RSTS & 0x555)); then
    # reset (partition number of 62 (or ANY non-zero number) encoded in bits: 0,2,4,6,8,10)
    STR="NormalReset"
elif ((RSTS & 0x20)); then
    # dog trip
    STR="DogReset"
elif ((RSTS & 0x1000)); then
    # power cycle
    STR="PowerCycle"
fi
printf "RSTS=0x%08X CAUSE=%s\n" $RSTS $STR
```
PR commit has been modified to reflect the proposed change along with checkpatch formatting adjustment. Thank you kindly.

PR #2720, opened 2018-10-16
ASoC: Fix compatible string for Allo Piano DAC

timg236:
The RPi PCM512x had the wrong compatible for the Allo Piano DAC which it
replaced. Change this to the correct value.


timg236:
@allo-com @pelwell @HiassofT

PR #2721, opened 2018-10-16
Add hy28b 2017 model device tree overlay

warlo:
The HY28B is shipped with a newer ILI9325 D chip, which uses another initialisation code as provided [here](http://knoppix.net/store/Hy28b-2-8-Touch-Screen-Tft-Lcd-With-Spi-8-bit-16-bit-Arduino-2017-Version_183082258048.html).

I added a new dt-overlay which seemingly works – I do atleast get text on the screen using 
`dtoverlay=hy28b-2017,rotate=90` in `boot/config.txt` after building it according to the instructions from fbtft [here](https://github.com/notro/fbtft/wiki/FBTFT-RPI-overlays).

### DIFF
--- a/arch/arm/boot/dts/overlays/Makefile
+++ b/arch/arm/boot/dts/overlays/Makefile
@@ -50,6 +50,7 @@ dtbo-$(CONFIG_ARCH_BCM2835) += \
 	hifiberry-digi-pro.dtbo \
 	hy28a.dtbo \
 	hy28b.dtbo \
+	hy28b-2017.dtbo \
 	i2c-bcm2708.dtbo \
 	i2c-gpio.dtbo \
 	i2c-mux.dtbo \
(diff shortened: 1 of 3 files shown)


pelwell:
Every overlay requires an entry in the README (which you can largely copy from the original hy28b) and a line in the Makefile, without which it won't be built. You should probably update the comments in the source file as well.

It would be nice if you had done a bit more testing, but as this device is currently unsupported I can live with a best-guess first attempt.

warlo:
Right, I have done some changes now!

Maybe I was a bit vague, I have built (outside of the kernel) and verified that the overlay works using the instructions from the fbtft archive. Kind of felt that more excessive testing wasn't critical, since this is only initialisation changes. If the pre-2017 hy28b overlay works – this overlay _should_ work as well. 🙂

Let me know if there is anything else

pelwell:
That looks good. Finally, do you have a name and email address for the Signed-off-by: line? I can apply it in the merge provided I know what it is. For example, mine would be:
```
```

warlo:
```
Signed-off-by: Hans-Wilhelm Warlo <hw@warlo.no>
```

PR #2727, opened 2018-10-23
Google VoiceHAT fixups

6by9:
Fixed issue flagged from https://www.raspberrypi.org/forums/viewtopic.php?f=29&t=224931 of not reading the sdmode GPIO from DT.

Cleaned up all checkpatch issues whilst I was there.

### DIFF
--- a/sound/soc/bcm/googlevoicehat-codec.c
+++ b/sound/soc/bcm/googlevoicehat-codec.c
@@ -50,9 +50,9 @@ static void voicehat_enable_sdmode_work(struct work_struct *work) {
 static int voicehat_component_probe(struct snd_soc_component *component) {
   struct voicehat_priv *voicehat = snd_soc_component_get_drvdata(component);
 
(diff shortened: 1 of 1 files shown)


popcornmix:
Looks plausible to me.
@HiassofT does it look okay to you?

6by9:
I have a VoiceHAT sitting here if you wanted to try it.

HiassofT:
Yes, looks fine, looking at DT overlay gpio was definitely grabbed from the wrong device / node. Puzzling how this could have worked before...

6by9:
I don't think it did work under 4.18 after the conversion from codec to component https://github.com/raspberrypi/linux/commit/dd71d79267b30682132b7602f399ae76c07f14f6 
The forum report initially said it was failing on 4.18 and there are no other changes in that area that I could see based on a quick look.

popcornmix:
I cherry picked this to 4.18 as it looks like it is needed there too.
I'll bump the branch=next kernel and see if it resolved the reported issue.

HiassofT:
Ah, yes, that makes sense. It was codec before in 4.14 and should then be component in 4.17+. "card" seems to have accidentally sneaked in

Issue #2728, opened 2018-10-25
sdhost: command MMC_SEND_EXT_CSD can unmount rootfs


lategoodbye:
During my investigation about this [issue](https://bugzilla.redhat.com/show_bug.cgi?id=1607872) with the upstream version of sdhost, i've found a problem which affects downstream (bcm2835-sdhost.c) and upstream (bcm2835.c).

Preparation:
- Install mmc-utils on Raspbian SD card ( mmc-utils_0+git20180327.b4fe0c8c-1_armhf.deb )

Steps to reproduce on Raspberry Pi 3B:
- run the following command a few times `sudo mmc extcsd read /dev/mmcblock0` (since this is timing critical it need some attempts)

Expected results:
- timeout of command MMC_SEND_EXT_CSD since this is a SD card not eMMC

Experienced results:
- at least after 10 attempts the root partition is unmounted which is really bad
 
Affected kernel versions:
- rpi-4.14.78
- rpi-4.19
- 4.19 (upstream)

Current investigation results:
Debugging this issue was a real pain because most of the commands depends on the rootfs. But i think i narrowed it down to that point that the command MMC_SEND_EXT_CSD confuses the driver / controller. The command MMC_SEND_EXT_CSD fails with an timeout BEFORE the controller FSM has finished. So the next valid command to the SD card could fail with 

`previous command (8) not complete (EDM 00010004)`

after this strange IO errors occur, the ext4 journal is aborted and the rootfs isn't available. The sdhost isn't able to break free from this situation.

As a band-aid for the upstream driver i made the following [patch](https://github.com/lategoodbye/rpi-zero/commit/74eb7a8341cde51d260f603aa9d0ec6308390041). This avoids the issue but is probably not the best solution. I assume you have (more) documentation about the sdhost controller to identify what's wrong here.

Edit: I apologize but the patch wasn't the right one. Fixed

lategoodbye:
Additionally here is a screenshot after the issue occured on a Raspberry Pi 3B:
![img_20181024_152618](https://user-images.githubusercontent.com/1618443/47495158-b677f480-d853-11e8-8b6d-add4f90872c5.jpg)

pelwell:
Thanks for the debug effort on this issue. Using an NFS root partition takes most of the pain out of SD driver errors, and network booting further streamlines the process.

[...]

PR #2729, opened 2018-10-26
mmc/bcm2835: Recover from MMC_SEND_EXT_CSD

pelwell:
See commit messages, and discussion in #2728.


lategoodbye:
I successfully tested the mmc/bcm2835 patch against 4.19 with a Raspberry Pi 3B (bad case) and a Compute Module 1 (good case).

Thanks

PR #2733, opened 2018-10-28
arm64: dts: broadcom: Enable fixups to support overlays on BCM2835 pl…

WallaceIT:
…atforms

Signed-off-by: Francesco Valla <valla.francesco@gmail.com>


ED6E0F17:
````
# Enable fixups to support overlays on BCM2835 platforms
ifeq ($(CONFIG_ARCH_BCM2835),y)
	DTC_FLAGS ?= -@
endif
````

I am assuming that this was dropped from arm64 kernels after 4.14 because of a merge conflict, but the "-H epapr" option has been dropped from 32bit kernels after 4.14. Is this now correct for aarch64, or was the 32bit change another merge conflict?


 https://github.com/raspberrypi/linux/blob/rpi-4.15.y/arch/arm/boot/dts/Makefile#L1127

pelwell:
As far as I know, the `-Hepapr` flag is essentially cosmetic for Raspberry Pi - because we don't make use of the `linux,phandle` properties there is no point in generating them, but failing to omit them will only result in some bloating of the outputs. However, I was the one who added the option in the first place, and its omission in recent kernels is an oversight.

The arm64 breakage is more serious, and probably does result from a merge conflict after the support for overlay generation was centralised.

pelwell:
`-H epapr` is the default for DTC in the 4.19 kernel and later, therefore the option isn't needed. I've merged another implementation of this PR to 4.19 (currently https://github.com/raspberrypi/linux/commit/05b4ba877df5a1f9a814f6f89ad375f3222f91bd, but rebasing...).

PR #2742, opened 2018-11-08
lan78xx: Increase default int urb interval

jemele:
The bInterval is set to 4 (i.e. 8 microframes => 1ms) and the only bit
that the driver pays attention to is "link was reset". If there's a
flapping status bit in that endpoint data, (such as if PHY negotiation
needs a few tries to get a stable link) then polling at a slower rate
would act as a de-bounce.

See: https://github.com/raspberrypi/linux/issues/2447

### DIFF
--- a/drivers/net/usb/lan78xx.c
+++ b/drivers/net/usb/lan78xx.c
@@ -359,6 +359,11 @@ static int msg_level = -1;
 module_param(msg_level, int, 0);
 MODULE_PARM_DESC(msg_level, "Override default message level");
 
+#define INT_URB_MICROFRAMES_PER_MS	8
+static int int_urb_interval_ms = 8;
+module_param(int_urb_interval_ms, int, 0);
+MODULE_PARM_DESC(int_urb_interval_ms, "Override usb interrupt urb interval");
+
 static int lan78xx_read_reg(struct lan78xx_net *dev, u32 index, u32 *data)
 {
 	u32 *buf = kmalloc(sizeof(u32), GFP_KERNEL);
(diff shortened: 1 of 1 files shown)


JamesH65:
This makes sense to me, but this should really be sent to the upstream netdev kernel mailing list to get some expert eyes on it, since it is unchanged upstream code. Hmm.  Are you familair with the upstreaming process?

JamesH65:
Althogh ISTR that upstream are a little reluctant to take on new module parameters, I wonder if it should be done via device tree?

pelwell:
Because that's far less contentious...

jemele:
>  but this should really be sent to the upstream netdev kernel mailing list to get some expert eyes on it

I can try this.

>  Are you familair with the upstreaming process?

Not really, but I can learn. Have you any constructive suggestions?

jemele:
> Because that's far less contentious...

I can also migrate the change to get the polling period from device tree, the module parameter was convenient for testing behavior quickly across a wide range of values. I'll post what I have to the upstream netdev list, and see what feedback I get there.

pelwell:
The reason for my comment is that, as upstream see it, Device Tree is meant to describe hardware in a platform-agnostic way, not to say how it should be configured - that's software's job. Here in downstream land we have a slightly more flexible attitude.

I think the module parameter approach is more likely to succeed.

lategoodbye:
I'm not the maintainer of this driver, but for a bugfix i would expect this one liner
`period = 8 * INT_URB_MICROFRAMES_PER_MS;`
and an explaining comment. Please no module parameter or device tree property.

@jemele If you want to upstream this patch, you should try to build net-next Kernel (4.20-rc1) on top of Raspbian and verify the issue is still reproducible.

jemele:
> Please no module parameter or device tree property.

Understood, thanks for the suggestion!

> @jemele If you want to upstream this patch, you should try to build net-next Kernel (4.20-rc1) on top of Raspbian and verify the issue is still reproducible.

I'll try that out tonight/tomorrow morning as time allows. Thanks for the constructive feedback!

JamesH65:
@jemele Did you have an progress with upstreaming this?

pelwell:
FYI An almost identical patch (5fa970b0) was merged as part of #2447, so I think we can close this.

JamesH65:
Please update to the latest kernel which may contain a fix for this issue.

PR #2739, opened 2018-10-31
Rpi 4.19.y unicam driver

6by9:
Cherry-picking all the Unicam driver and "supported" source subdevice driver fixes from 4.14 to 4.19, with minor fixups due to API changes.
TC358743 and ADV7282 both tested and working.

PR #2746, opened 2018-11-13
Commits lost since rpi-4.14.y (and fix a few warnings)

pelwell:
@popcormix @6by9 @JamesH65 Can you give these a quick glance before I merge them?


6by9:
The ov5647/tc358743/unicam patches duplicate #2739.
Are they a straight cherry-pick from 4.14 as I thought there were a couple of things I had to patch?

The bcm2835-camera/mmal-vchiq driver changes I've also patched up with the intent of upstreaming (Stefan and Eric want it bumped, plus the codec driver). From my experimentation it doesn't work at all without Stefan's fixup patches from https://github.com/anholt/linux/commits/bcm2835-audio, or at least the platform driver loading stuff doesn't.

pelwell:
Yes - they're straight cherry-picks. I have a script that almost automates it. I've dropped those 8 "staging" commits.

JamesH65:
The only one from me is the template stuff, so that's fine. The reports we have been getting are already better!

6by9:
The Unicam stuff is missing at least
media: tc358743: Check I2C succeeded during probe.  https://github.com/raspberrypi/linux/commit/66de8dc6c1db17cdae7d17ed33327f70ba10745c
ov5647: Add support for PWDN GPIO.  https://github.com/raspberrypi/linux/commit/490d95124d9e1c0b640468195a7970631b816e9a
And the little one bcm2835-unicam: Driver for CCP2/CSI2 camera interface https://github.com/raspberrypi/linux/commit/71cfc95113843fa0c34b03a8f6d8e6d5fac8eb0d

Do you want to drop all drivers/media ones and pick up #2739 instead? I'm rebasing it, will test and repush.

pelwell:
That's all of the media and camera stuff removed.

6by9:
> That's all of the media and camera stuff removed.

Not quite - you've still got "BCM283x DT: Add CSI nodes to the device tree. " and "BCM270X_DT: Add CSI defines for all the downstream Pi platforms ", which are going to cause some oddness without the corresponding driver.

pelwell:
Agg/ack - nixed.

6by9:
I'm happy.
The Unicam update should be today (build just completing). 
V4L2 driver changes just need finalising, and working out what to do with regard loading as a platform driver (no more needing to ```modprobe bcm2835-v4l2```, but slightly more resources used on the VPU constantly. Probably needs Stefan's patches which aren't merged upstream yet)

arch/arm/boot/dts/bcm270x.dtsi
```diff
@@ -150,3 +150,7 @@
 &vc4 {
 	status = "disabled";
 };
+
+&vc4 {
+	status = "disabled";
+};
```
popcornmix: This seems to be adding a duplicate node...

drivers/net/wireless/realtek/rtl8192cu/core/rtw_security.c
```diff
@@ -1878,8 +1878,8 @@ _func_enter_;
     for (j = 0; j < 8; j++)
     	message[payload_index+j] = mic[j];
 
-	payload_index = hdrlen + 8;
-	for (i=0; i< num_blocks; i++)
+    payload_index = hdrlen + 8;
+    for (i=0; i< num_blocks; i++)
```
popcornmix: I'd suggest dropping this commit. We are now using unmodified code from:  https://github.com/pvaret/rtl8192cu-fixes
If this is wanted, it could be submitted there and we'll pick it up.

pelwell: Done.

popcornmix:
Apart from two comments, all looks fine.

pelwell:
Thanks everyone.

PR #2749, opened 2018-11-13
Mailbox firmware calls now use kmalloc

JamesH65:
A previous change moved away from variable stack
allocation of a data buffer to a fixed maximum size.
However, some mailbox calls use larger data buffers
than the maximum allowed. This change moves from
stack storage to kmalloc to ensure all sizes are
catered for.

### DIFF
--- a/drivers/firmware/raspberrypi.c
+++ b/drivers/firmware/raspberrypi.c
@@ -15,14 +15,13 @@
 #include <linux/of_platform.h>
 #include <linux/platform_device.h>
 #include <linux/reboot.h>
+#include <linux/slab.h>
 #include <soc/bcm2835/raspberrypi-firmware.h>
 
 #define MBOX_MSG(chan, data28)		(((data28) & ~0xf) | ((chan) & 0xf))
 #define MBOX_CHAN(msg)			((msg) & 0xf)
 #define MBOX_DATA28(msg)		((msg) & ~0xf)
 #define MBOX_CHAN_PROPERTY		8
-
-#define MAX_RPI_FW_PROP_BUF_SIZE	48
 
 static struct platform_device *rpi_hwmon;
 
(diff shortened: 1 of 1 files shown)


pelwell (commented):
I think you can get rid of some casts if you make `data` a `void *`.

JamesH65:
I only see one cast, to struct rpi_firmware_property_tag_header for the header variable, and I think that is required?

pelwell:
That's the one - any pointer can be treated as a void *, and any void * can be treated as any pointer, without the need for a cast.

JamesH65:
Still need the cast in order to write the contents of the header, so needs to be a struct rpi_firmware_property_tag_header pointer. Could assign that directly from the kmalloc of course, but that would break the later pointer arithmetic which assumes bytes.

pelwell:
No cast needed - the header pointer has the correct type.

GCC considers sizeof(*(void*)), i.e. sizeof(void), to be 1, so pointer arithmetic works as you would want. This usage is common in the kernel.

JamesH65:
Ah, finally got what you mean. So you can assign to the header point from a void without the cast if the original ptr is a void.

pelwell:
By jove, etc.

JamesH65:
Have pushed fix. It was the sizeof(void) bit that I wasn't aware of. Aged brain. Must have seen it loads of time in the kernel!

pelwell:
Looks good to me.

JamesH65:
@lategoodbye  This should probably be sent upstream (ref. #2738), is that something you would be happy to do?

lategoodbye:
In general i prefer that the patch author submit its own patch upstream, because this makes things much easier in communication. I'm fine with these changes, but the commit log need some improvements.

JamesH65:
OK. Which mailing list does it need to go to?

lategoodbye:
Suggestions for the commit log:
- subject should be something like: firmware: raspberrypi: Fix firmware calls with large buffers
- also please refer to the first 12 digits of the commit hash which introduces the regression
- add a reference to the current mailbox interface description in github
- add a fixes tag

You will get the relevant recipients by calling scripts/get_maintainers.pl against your patch. But the more important question would be: Against which branch should i rebase my patch?
Answer: https://github.com/anholt/linux/commits/bcm2835-soc-next

lategoodbye:
Sorry, i mixed up the branch names and needed to rebase them. Now the proper one would be the following:
https://github.com/anholt/linux/commits/bcm2835-drivers-next

PR #2750, opened 2018-11-13
Update README

JamesH65:
Small update to the DT blob docs to include the axiperf option.


pelwell:
I've removed the trailing whitespace. Is it OK to add a Signed-off-by: when squashing?

JamesH65:
Indeed. I did the edit via github editor, and not sure you can add an automatic signed off line in that - woudl have to edit it in I suspect.

PR #2755, opened 2018-11-14
V4L2 codec interface

6by9:
Pushing for comment.

The first 11 patches were submitted by Stefan to the list (25th Oct - not in the archives!) but a couple of changes were requested. We could wait for those changes or take as is. Where I've based stuff on his I've taken on the comments made on his patches.
I need ```vchiq_register_child``` to be able to add the codec driver and vcsm-cma correctly with dma masks (I do get given the wrong physical address alias, so something is wrong that I haven't got to the bottom of).

The last patch is being resurrected on linux-media with an updated patch being pushed today (not yet archived). It makes life significantly easier for me as the buffer is exported within the driver to pass to vcsm-cma and be mem_wrapped, and otherwise a standard shutdown can trigger warnings as the VPU hasn't released the buffer in time. All buffers are ref counted via dmabufs, so it is safe.

The rest is all up for review.


popcornmix:
> We could wait for those changes or take as is.

As long as upstream versions have (approximately) the same commit message and functionality, then no problem with merging these now. When commits appear upstream we can just drop the downstream commits on the rebase.

6by9:
Decoder works quite happily via GStreamer 1.14. Not tested on 1.10(?) which is shipped with Stretch.
```
gst-launch-1.0 -e -vvv filesrc location=usbhdd/big_buck_bunny_1080p_h264.mov ! qtdemux name=demux ! queue ! h264parse ! v4l2h264dec capture-io-mode=4 ! kmssink demux. ! queue ! faad ! autoaudiosink
```
plays back Big Buck Bunny via the full KMS driver pretty well (audio and video, but a few jerks in the video).

Transcoding is working
```
gst-launch-1.0 -e -vvv filesrc location=usbhdd/big_buck_bunny_1080p_h264.mov ! qtdemux name=demux ! queue ! h264parse ! v4l2h264dec ! v4l2h264enc ! filesink location=foo.h264
```
You should be able to alter the encoding parameters too although I've not tested it.

Plain encoding from v4l2src is giving some errors on UYVY - need to investigate further.

6by9:
> As long as upstream versions have (approximately) the same commit message and functionality, then no problem with merging these now. When commits appear upstream we can just drop the downstream commits on the rebase.

The request was a simple one to check for an error from ```vchiq_register_child``` during probe and set the handle to NULL on error, rather than having to check ```IS_ERR``` during ```vchiq_remove```. That's why mine do that whilst Stefan's don't.
I'd expect the commit text to remain the same.

popcornmix:
I think stretch gstreamer is too old (it failed for me), but I could play video with ffmpeg-drm test app.
The kernel didn't obviously break anything (kodi with the legacy kms driver still plays hevc for example).
I didn't spot anything on a quick scan through the commits so no objections here.

6by9:
I'm not in tomorrow morning, but will have a look at what level of working I can achieve with GStreamer in Stretch. Buster has GStreamer 1.14.4, so that should support it.

Next job is to get Chromium built and trying to use it in the GL world. Not quite so trivial as they use custom kernel headers which aren't then shipped in the Chromium source :-/ It also uses some hardcoded node names - /dev/video-dec and /dev/video-enc.
I'll also be aiming to get these patches upstreamed - fun, fun, fun!

6by9:
Rebased to 4.19.2.
Github appears to be doing odd things with the order it is listing the commits, but the hashes are correct.

6by9:
Rebased again to 4.19.4 with minor fix to the encode path for YUYV or RGB formats.

Tested with GStreamer 1.15 again. GST1.10 appears to be just too old to be of any real use.
Also tested with FFmpeg and seems to do sensible things on at least decoding.

I'm trying to get Chromium running with Buster as it has the latest libDRM and Mesa.

What else do you want to be done before this is merged?

6by9:
@pelwell Any chance of a once over?

Chromium is proving to be a pain as it pulls an alternate huge lump of UI code, which seems to not work at the moment (segfaults on starting). The current build doesn't use any of that, so John hasn't any pearls of wisdom to impart. I'll keep poking at it, but it'd be nice to get this merged (rebasing the PR with each kernel bump gets boring).

drivers/staging/vc04_services/bcm2835-camera/bcm2835-camera.c
```diff
@@ -358,68 +360,72 @@ static void buffer_cb(struct vchiq_mmal_instance *instance,
...
+			v4l2_dbg(1, bcm2835_v4l2_debug, &dev->v4l2_dev,
+				 "Buffer time set as start timestamp - %lld",
+				 buf->vb.vb2_buf.timestamp);
 		}
 	}
+	dev->capture.last_timestamp = buf->vb.vb2_buf.timestamp;
+
+	vb2_set_plane_payload(&buf->vb.vb2_buf, 0, length);
+	if (mmal_flags & MMAL_BUFFER_HEADER_FLAG_KEYFRAME)
+		buf->vb.flags |= V4L2_BUF_FLAG_KEYFRAME;
```
pelwell, line 433: Where did this pair of lines appear from?

drivers/staging/vc04_services/bcm2835-camera/bcm2835-camera.c
```diff
@@ -799,27 +805,27 @@ static int vidioc_overlay(struct file *file, void *f, unsigned int on)
...
+		return ret;
 
 	ret = set_overlay_params(dev, dst);
 	if (ret < 0)
-		goto error;
+		return ret;
 
 	if (enable_camera(dev) < 0)
-		goto error;
+		return ret;
```
pelwell, line 851: This doesn't actually set the return value, so the function will appear to succeed.

drivers/staging/vc04_services/bcm2835-camera/mmal-common.h
```diff
@@ -13,6 +13,8 @@
  * MMAL structures
  *
  */
+#ifndef __MMAL_COMMON_H
+#define __MMAL_COMMON_H
```
pelwell: Do you need the extra underscores here and not in mmal_msg.h?

drivers/staging/vc04_services/vchiq-mmal/mmal-vchiq.h
```diff
@@ -82,6 +82,7 @@ struct vchiq_mmal_port {
 };
 
 struct vchiq_mmal_component {
+	u32 in_use:1;
```
pelwell, line 90: Is this preferable to a bool (as featured on the line below)?

drivers/staging/vc04_services/vchiq-mmal/mmal-vchiq.c
```diff
@@ -44,6 +44,12 @@ MODULE_VERSION("0.0.1");
  */
 #define VCHIQ_MMAL_MAX_COMPONENTS 64
 
+/*
+ * timeout for synchronous msg responses in seconds.
```
pelwell: Capitalise `Timeout`, since it starts a sentence? I only noticed because of the line below.

drivers/staging/vc04_services/bcm2835-camera/bcm2835-camera.c
```diff
@@ -267,14 +273,17 @@ static int buffer_init(struct vb2_buffer *vb)
...
+				container_of(vb2, struct vb2_mmal_buffer, vb);
 
 	v4l2_dbg(1, bcm2835_v4l2_debug, &dev->v4l2_dev, "%s: dev:%p, vb %p\n",
 		 __func__, dev, vb);
-	buf->buffer = vb2_plane_vaddr(&buf->vb.vb2_buf, 0);
-	buf->buffer_size = vb2_plane_size(&buf->vb.vb2_buf, 0);
+	buf->mmal.buffer = vb2_plane_vaddr(&buf->vb.vb2_buf, 0);
+	buf->mmal.buffer_size = vb2_plane_size(&buf->vb.vb2_buf, 0);
+
+	mmal_vchi_buffer_init(dev->instance, &buf->mmal);
```
pelwell: Is this init function guaranteed never to fail? You don't mention this change,

drivers/staging/vc04_services/bcm2835-camera/bcm2835-camera.c
```diff
@@ -672,6 +690,14 @@ static void stop_streaming(struct vb2_queue *vq)
 		}
 	}
 
+	/* Release the VCHIQ buffers */
+	for (i = 0; i < vq->num_buffers; i++) {
+		vb2 = to_vb2_v4l2_buffer(vq->bufs[i]);
+		buf = container_of(vb2, struct vb2_mmal_buffer, vb);
+
+		mmal_vchi_buffer_cleanup(&buf->mmal);
+	}
```
pelwell: Why has this loop suddenly become necessary?

pelwell (commented):
Batch 1 - up to "staging: mmal-vchiq: Make a mmal_buf struct for passing parameters"

drivers/staging/vc04_services/bcm2835-camera/controls.c
```diff
@@ -1109,7 +1109,7 @@ static const struct bm2835_mmal_v4l2_ctrl v4l2_ctrls[V4L2_CTRL_COUNT] = {
 	{
 		V4L2_CID_POWER_LINE_FREQUENCY, MMAL_CONTROL_TYPE_STD_MENU,
 		0, ARRAY_SIZE(mains_freq_qmenu) - 1,
-		1, 1, NULL,
+		1, 1, mains_freq_qmenu,
```
pelwell: It seems odd to include this patch when the next patch deletes mains_freq_qmenu.

drivers/staging/vc04_services/bcm2835-camera/bcm2835-camera.c
```diff
@@ -360,8 +360,13 @@ static void buffer_cb(struct vchiq_mmal_instance *instance,
 		}
```
pelwell, line 377: Should you change the commit message from "H264 header come off VC" to "H264 headers come from VC"?

drivers/staging/vc04_services/bcm2835-camera/bcm2835-camera.c
```diff
@@ -1062,12 +1062,15 @@ static int mmal_setup_components(struct bm2835_mmal_dev *dev,
...
 		/* Make a further decision on port based on resolution */
 		if (f->fmt.pix.width <= max_video_width &&
-		    f->fmt.pix.height <= max_video_height)
-			camera_port = port =
+		    f->fmt.pix.height <= max_video_height) {
+			camera_port =
 			    &dev->component[COMP_CAMERA]->output[CAM_PORT_VIDEO];
-		else
-			camera_port = port =
+			port = camera_port;
```
pelwell: Move the duplicated assignment to port out outside the conditional and avoid the extra braces.

drivers/staging/vc04_services/interface/vchiq_arm/vchiq_arm.c
```diff
@@ -3656,6 +3657,9 @@ static int vchiq_probe(struct platform_device *pdev)
 
 	bcm2835_camera = vchiq_register_child(pdev, "bcm2835-camera");
 	bcm2835_audio = vchiq_register_child(pdev, "bcm2835_audio");
+	bcm2835_codec = vchiq_register_child(pdev, "bcm2835-codec");
+	if (IS_ERR(bcm2835_codec))
```
pelwell, line 3669: Why the inconsistent error handling? If you change the others (some of which were added in this patch set) to match in a later commit, is it not better to keep them consistent here then change them all?

drivers/staging/vc04_services/interface/vchiq_arm/vchiq_arm.c
```diff
@@ -3655,6 +3656,9 @@ static int vchiq_probe(struct platform_device *pdev)
 		VCHIQ_VERSION, VCHIQ_VERSION_MIN,
 		MAJOR(vchiq_devid), MINOR(vchiq_devid));
 
+	vcsm_cma = vchiq_register_child(pdev, "vcsm-cma");
+	if (IS_ERR(vcsm_cma))
+		vcsm_cma = NULL;
```
pelwell, line 3661: Inconsistent error handling again.

pelwell (commented):
2nd batch - the rest of the commits. Just a few minor quibbles.

pelwell:
GitHub is demanding another manual rebase, which I don't understand since there were no conflicts when I rebased the PR manually offline.

drivers/staging/vc04_services/bcm2835-camera/bcm2835-camera.c
6by9: Should have been a cherry-pick of 5b0d431bd87468f13f7ba8c96eb791c4c4bee467 "BCM2835-V4L2: Correctly denote key frames in encoded data" from 4.14.
I'll find out what happened there.

drivers/staging/vc04_services/bcm2835-camera/bcm2835-camera.c
6by9: Ack.

drivers/staging/vc04_services/bcm2835-camera/mmal-common.h
6by9: We've a slight mismash with mmal_parameters.h having underscores with most of the others not. I'll unify them.

drivers/staging/vc04_services/vchiq-mmal/mmal-vchiq.h
6by9: IIRC checkpatch whinges at bool. Looking at my ToT a later patch changes the line below to ```u32 enabled:1```

drivers/staging/vc04_services/vchiq-mmal/mmal-vchiq.c
6by9: Ack

drivers/staging/vc04_services/bcm2835-camera/bcm2835-camera.c
6by9: Github loses the commit the reviewer was looking at when they made the comment, so it's possible this has become misplaced in the series.
A later patch does check the return value, but this one ought to too.

The vchi context is now attached to the buffer at allocation, not allocated on demand. I'll review to check that bits got added at the correct point.

drivers/staging/vc04_services/bcm2835-camera/bcm2835-camera.c
6by9: As above, the vchi context is now attached to the buffer at allocation, not allocated on demand. I'll review to check that bits got added at the correct point.

drivers/staging/vc04_services/bcm2835-camera/controls.c
6by9: The 1st patch got merged upstream, so I brought us into line with upstream before fixing the actual commit.

drivers/staging/vc04_services/bcm2835-camera/bcm2835-camera.c
6by9: Ack

drivers/staging/vc04_services/interface/vchiq_arm/vchiq_arm.c
6by9: Stefan's patchset has not been merged upstream as they wanted changes. Shall I make a fixup to his patch first include the error handling, and then add the new ones to match?

6by9:
> GitHub is demanding another manual rebase, which I don't understand since there were no conflicts when I rebased the PR manually offline.

Each force push bumping the Linux release is wanting the manual rebase even with no conflicts. It's getting a little annoying, but not something we can fix.

drivers/staging/vc04_services/interface/vchiq_arm/vchiq_arm.c
6by9: As above, Stefan's patches aren't merged. I'll fix them up as a separate patch.

drivers/staging/vc04_services/interface/vchiq_arm/vchiq_arm.c
pelwell: As you see fit, but that sounds preferable.

pelwell:
I'll merge when you're happy with it.

drivers/staging/vc04_services/interface/vchiq_arm/vchiq_arm.c
6by9: OK, will do.

6by9:
Thanks for all your efforts in reviewing. I'll update as noted.

drivers/staging/vc04_services/bcm2835-camera/bcm2835-camera.c
6by9: (That makes more sense now that I've found your comment against a commit rather than in the summary. Blasted Github!)
Yes, the change was incorrect. I suspect it got tripped up in the port from 4.14 to 4.19 (buffer_init is already upstream in 4.19).

drivers/staging/vc04_services/bcm2835-camera/bcm2835-camera.c
6by9: You're right - this is handled from buffer_cleanup. I'm not sure why it got duplicated here.

6by9 (commented):
Hopefully the latest push should address all the issues you raised. Thanks again for the review.

pelwell:
I'm going to take your word for it and merge before another rebase is required

Issue #2754, opened 2018-11-14
Race-condition in spi-bcm2835.c when performing an SPI transfer in interrupt mode


niklasekstrom:
(I initially reported this bug to the Raspberry pi forum, https://www.raspberrypi.org/forums/viewtopic.php?f=44&t=214648&p=1321017#p1321017, but repeat it here for completeness.)

The bug is present in the spi-bcm2835.c SPI driver for Raspberry PI. Occasionally when the driver is reading data using an interrupt transfer, the last byte, or even more rarely the last couple of bytes, are not transferred from the hardware RX FIFO buffer to the receive buffer in memory by the driver.

To find the problem I did the following, using a Raspberry PI 3 (should work on other models that use the same driver). I have an SPI memory connected to the Raspberry, and I wrote a small test program that repeatedly does the following cycle: write a sequence of bytes of random length to a random address, and then read back that sequence of bytes and verify that the bytes read back are the same as those written. Almost all of the time the byte string is read back correctly, but occasionally the last byte was not correctly read back.

My analysis is as follows. I noticed a pattern in the length of the sequence of bytes that the failing write/read cycle had. The length of a failing transfer was always between 67 bytes and 95 bytes. I looked in the source code for the driver (https://github.com/raspberrypi/linux/blob/rpi-4.14.y/drivers/spi/spi-bcm2835.c) and found that a transfer can be done in one of three modes, depending on the length of the transfer: (1) polling, (2) interrupt, and (3) DMA. The code to determine which mode is used is in bcm2835_spi_transfer_one(); note that the lower boundary of 67 bytes depends on the SPI speed which for me is 20 MHz on a RPi3 (400 MHz core clock). All the transfers that were failing were running in the interrupt mode.

I then looked in the bcm2835_spi_interrupt() interrupt handler, and there is a race-condition. At line 154 (https://github.com/raspberrypi/linux/blob/rpi-4.14.y/drivers/spi/spi-bcm2835.c#L154) bytes are drained from the RX FIFO. Then on line 159 there is a check to see if the transfer is complete (the DONE bit is set in the CS register). However, if the transfer is found to be completed on line 159 there is still a (small) chance that there is data in the RX FIFO that arrived at a time in between when lines 154 and 159 were executed.

Fix: By making the following change to the driver, it works correctly after that:

line 159:  if (bcm2835_rd(bs, BCM2835_SPI_CS) & BCM2835_SPI_CS_DONE) { // this line is unchanged
[...]

PR #2756, opened 2018-11-14
Fix to avoid race condition bug.

niklasekstrom:
A pull request for the fix to avoid the race condition bug, re. issue #2754.


pelwell:
Closing this PR while we wait for Lukas Wunner's patch set to be accepted upstream.

PR #2757, opened 2018-11-15
Delayed shutdown operation inhibits resetting the soundcard if it is recaptured before timeout

drestebon:
Some applications release an ALSA device whenever the stream stops as, for example, between tracks in a playlist. In some setups this causes associated hardware to reset: in a DAC, connected to the sound card through S/PDIF, mechanical relays will click after every track, delaying playback and hindering hardware service life.

Apparently the clicks are produced by setting up wm8804's PLL.

This issue has been reported in the context of applications:

https://github.com/librespot-org/librespot/issues/68

Proposed Solution
=================

In this patch, the issue is tackled using a timer to control a shutdown flag. The flag is active for up to 5 seconds after shutdown() has been called. When the flag is active, it inhibits setting up wm8804's PLL in hw_params(), this is, if it gets called before the timeout. If after 5 seconds the sound card is not captured again, DAPM will shut it down. After this has happened, setting up the PLL in hw_params() doesn't produce any further hiccups.

### DIFF
--- a/sound/soc/bcm/hifiberry_digi.c
+++ b/sound/soc/bcm/hifiberry_digi.c
@@ -18,6 +18,9 @@
 #include <linux/module.h>
 #include <linux/platform_device.h>
 
+#include <linux/timer.h>
+#include <linux/jiffies.h>
+
 #include <sound/core.h>
 #include <sound/pcm.h>
 #include <sound/pcm_params.h>
@@ -27,10 +30,6 @@
 
 #include "../codecs/wm8804.h"
 
-static short int auto_shutdown_output = 0;
-module_param(auto_shutdown_output, short, S_IRUSR | S_IWUSR | S_IRGRP | S_IWGRP);
-MODULE_PARM_DESC(auto_shutdown_output, "Shutdown SP/DIF output if playback is stopped");
-
 #define CLK_44EN_RATE 22579200UL
 #define CLK_48EN_RATE 24576000UL
 
@@ -39,6 +38,11 @@ static struct gpio_desc *snd_rpi_hifiberry_clk44gpio;
 static struct gpio_desc *snd_rpi_hifiberry_clk48gpio;
 
 static int samplerate=44100;
+
+#define FLAG_TIMEOUT_IN_SECONDS 5
+static bool shutdown_flag;
+void drop_shutdown_flag(unsigned long data);
+static struct timer_list shutdown_flag_timer = TIMER_INITIALIZER(drop_shutdown_flag, FLAG_TIMEOUT_IN_SECONDS*HZ, 0);
 
 static uint32_t snd_rpi_hifiberry_digi_enable_clock(int sample_rate)
 {
@@ -85,16 +89,17 @@ static int snd_rpi_hifiberry_digi_startup(struct snd_pcm_substream *substream) {
 	return 0;
 }
 
-static void snd_rpi_hifiberry_digi_shutdown(struct snd_pcm_substream *substream) {
-	/* turn off output */
-	if (auto_shutdown_output) {
-		/* turn off output */
-		struct snd_soc_pcm_runtime *rtd = substream->private_data;
-		struct snd_soc_codec *codec = rtd->codec;
-		snd_soc_update_bits(codec, WM8804_PWRDN, 0x3c, 0x3c);
-	}
+void drop_shutdown_flag(unsigned long data)
+{
+	shutdown_flag = 0;
 }
 
+static void snd_rpi_hifiberry_digi_shutdown(struct snd_pcm_substream *substream)
+{
+	/* raise shutdown flag */
+	shutdown_flag = 1;
+	mod_timer(&shutdown_flag_timer, jiffies + FLAG_TIMEOUT_IN_SECONDS*HZ);
+}
 
 static int snd_rpi_hifiberry_digi_hw_params(struct snd_pcm_substream *substream,
 				       struct snd_pcm_hw_params *params)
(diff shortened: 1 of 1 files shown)


pelwell:
I have a few problems with this PR:
1. It's a fairly large patch, mainly as a result of the indentation change. In fact, indentation as a whole is inconsistent - please follow kernel coding style and run your patch through scripts/checkpatch.pl before submission.
2. Although the commit message explains the aim of the patch, it needs an explanation of the strategy - why it is implemented in that way. For example, why is startup also delayed?  Is that really desirable?
3. Is the use of atomic operations really necessary?
4. The correct spelling is "shutting".

drestebon:
Hi Phil,

Thank you very much for your comments. I'll try to address them here:

1. I corrected the commit, to include only the new code. I did the best I could regarding the style. I wouldn't mind correcting the style of the whole text. Should I make a new commit with such changes?
2. I included an explanation of the strategy in the PR text. Startup is **not** delayed: the startup timer only handles the startup flag, which inhibits the delayed shutdown. You are right: a delayed startup is not desirable.
3. I assumed that atomic operations would be required with such timed operations relaying on shared flags. Indeed, atomic operations are not needed if every call to the driver functions (included those fired by the timers) block the rest. If you think this is the case, static booleans would do.
4. Corrected the spelling.

HiassofT:
I'd recommend you take a different approach:

The problem is caused by the driver's shutdown code. It shouldn't change the wm8804's PWRDN register there.

Actually I'm not sure why the PWRDN register is touched in startup/shutdown in the first place as this interferes with the wm8804 code - most of the bits set/cleared by the soundcard driver are under DAPM control (see the wm8804 code).

From a quick glance only bit 5 (0x20) isn't set in the wm8804 driver - try changing the snd_soc_update_bits calls so that only 0x20 is touched.

DAPM already implements a delayed shutdown feature, to avoid exactly the stop/start issue, by default it waits 5 seconds after (playback) stop before the audio interface is brought down - see eg the comments on the pmdown_time module parameter here https://github.com/raspberrypi/linux/blob/rpi-4.14.y/sound/soc/soc-core.c#L63-L70
```
/*
 * This is a timeout to do a DAPM powerdown after a stream is closed().
 * It can be used to eliminate pops between different playback streams, e.g.
 * between two audio tracks.
 */
static int pmdown_time = 5000;
module_param(pmdown_time, int, 0);
MODULE_PARM_DESC(pmdown_time, "DAPM stream powerdown time (msecs)");
```

HiassofT:
I just had a look at the WM8804 datasheet, bit 5 of PWRDN register is "tristate all outputs". The wm8804 driver sets that bit only during suspend (and clears it after resume).

So I'd say just try dropping the snd_soc_update_bits call - everything should already be set up properly by the wm8804 driver

drestebon:
Hi HiassofT,

I made an attempt at this. I managed to get rid of the clicks without any fancy flags or delays. Actually, the drivers almost doesn't do anything in startup, hw_params or shutdown. Still, I believe that something must be missing.

I got rid of all writes to WM8804_PWRDN, but that didn't get rid of the clicks. I had to get rid of snd_soc_dai_set_clkdiv(), snd_soc_dai_set_pll() and snd_soc_dai_set_sysclk(). With this, the card works alright in my setup, except I get a complain from bcm2835:

`bcm2835-i2s 3f203000.i2s: I2S SYNC error!`

Also, if I understand things correctly, the pro version of the interface attempts to use external 48 and 44.1[kHz] clock sources. I don't have that card to make it fail, unfortunately. Should the clock and pll calls be made in probe or init instead? All interface drivers for the wm8804 indicate otherwise.

I am not sure how to proceed. Any ideas?

**UPDATE**

The PLL must be setup (at least once) in the life of snd_soc_wm8804 otherwise the i2s just fails. In my previous attempts I didn't reload the wm8804 driver, so the pll configuration persisted and I could get away with loading a hifiberry_digi module that didn't call set_pll() and still get sound output.

Now I am under the impression that the clicks are caused by clock changes in set_pll() here
https://github.com/raspberrypi/linux/blob/f6878de4d2818c2256b2a340eaeeabfb0b24c71b/sound/soc/codecs/wm8804.c#L427-L445

I came up with a simpler _compromise_ fix, where shutdown is used to control a flag (active for up to 5 seconds after shutdown is called) that inhibits setting up the pll in hw_params, if it gets called before the flag timeout. Also, shutdown doesn't bother modifying PWRDN. That is left to DAPM.

HiassofT:
Instead of using an additional timer it's better to perform changes only when they are needed.

The driver already keeps track of the current samplerate so you can just check in hwparams if the samplerate changed and bail out immediately if it's the same.

I did a very quick test on a hifiberry digi pro with that change, and all PWRDN stuff removed and it seemed to work fine. Here's the tree https://github.com/HiassofT/rpi-linux/commits/rpi-4.14.y-hifiberry-digi-test

drestebon:
It works perfectly in my setup as well. I agree, the best solution is to do changes only when necessary. I'm closing this PR now. Thanks for the work!

drestebon:
One note, getting rid of the PRDN write in startup produces this error:

       `bcm2835-i2s 3f203000.i2s: I2S SYNC error!`

This should probably be fixed in DAPM instead?

pelwell:
@HiassofT I'll happily take a PR (or grab from your branch above) when you're ready.

drestebon:
This includes your code, plus the write to WM8804_PWRDN in startup(). It works in my setup without any issues:

ce01a4157dd2a2730ae89606415e824affb6609c

I think it will take me a long while to dig into a fix in DAPM. I am not convinced that is really necessary. It will probably cause more problems that it will fix anyway.

HiassofT:
The I2S SYNC error is harmless (at least if you get only one such message on playback start).

It's caused by the bcm2835_i2s_clear_fifos call from bcm2835_i2s_hw_params. At this point the wm8804 (or most other codecs acting as a clock master) hasn't been started up so no clocks are present and the code will run into a timeout - and log that error.

A second call to bcm2835_i2s_clear_fifos is made just before stream startup (in bcm2835_i2s_prepare), at this point clocks should be present (unless something's really broken), the stream's FIFO will be cleared and the sync bit check will succeed.

@pelwell I think it'd be best to test this in the 4.19 branch first. I only did a quick test with the Hifiberry DigiPro card and can do another test on a Justboom Digi - but I don't have any other cards.

Although I'm pretty sure we can drop all the PWRDN stuff (it was needed before DAPM and runtime PM support was added to the wm8804 driver in kernel 4.0) I'd rather not break your current stable kernel in case I missed something.

I'll prepare a patch series for 4.19 and PR it later if local tests are fine.

HiassofT:
4.19 tests looked good here and I've opened PR #2759 - which should benefit all WM8804 based soundcards.

Getting some feedback on that PR would be nice :-)

PR #2758, opened 2018-11-17
ASoC: add driver for 3Dlab Nano soundcard

dev-3Dlab:
3Dlab Nano Player is a high-end audio player supporting high resolution PCM and DSD audio.


pelwell:
Sorry for the delay - it looks like you've stopped developing now. This driver is so simple it feels as though it could use simple-card, but since it effectively embeds the codec driver it is probably as small as it's going to get. 

It would be nice to have a `Signed-off-by: Name <email>` line to go in the commit message - just reply here with something suitable and I'll incorporate it when squashing - but otherwise I have no objections.

Any comments, @timg236?

dev-3Dlab:
Hi pelwell,

Thank you for your comment! I took this opportunity to push a very last minor code simplification, before merging.

The sign off line would be:
`Signed-off-by: GT <dev@3d-lab-av.com>`

PR #2759, opened 2018-11-18
[4.19] RPi wm8804 soundcard driver cleanup and improvements

HiassofT:
This is a followup to #2757 to prevent clicks on stop/start operations

Since kernel 4.0 the DAPM code in the wm8804 driver handles setting up it's PWRDN register and the code in the soundcard driver to do that is no longer needed. Actually that code interfers with DAPM and disables delayed DAPM shutdown handling (which tries to prevent clicks from stop+start operations).

Changing the WM8804 clocks/PLLs can also lead to clicks, so avoid doing that in hwparams when they are already setup properly.

Another change in this PR is the removal of the idle_bias change in the wm8804 driver which seems to have accidentally sneaked back in 4.19 - in 4.14 this change was already dropped and the wm8804 driver was at upstream level.

I tested the change with the Hifiberry Digi+ Pro and Justboom Digi HATs but some more tests would be welcome.

Note: with this change you may see a harmless "I2S SYNC error" message in dmesg on stream startup. This is caused by the clear fifo call in hwparams which is performed before the wm8804 clock has been started. The second clear fifo call from .prepare will succeed though and everything should be fine.

@drestebon can you test if this change works for you, too?

@timg236 a second pair of eyes glancing over these changes would be nice, too

@MilhouseVH could you include this PR in your LE testbuilds?

### DIFF
--- a/sound/soc/bcm/rpi-wm8804-soundcard.c
+++ b/sound/soc/bcm/rpi-wm8804-soundcard.c
@@ -54,8 +54,6 @@ struct snd_rpi_wm8804_drvdata {
 	struct snd_soc_dai_link *dai;
 	/* Required - snd_soc_card name */
 	const char *card_name;
-	/* Optional- Overrides the module paramter */
-	unsigned short auto_shutdown_output;
 	/* Optional DT node names if card info is defined in DT */
 	const char *card_name_dt;
 	const char *dai_name_dt;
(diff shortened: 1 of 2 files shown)


MilhouseVH:
Will do, thanks.

drestebon:
Works like a charm on the hifiberry's digi+ for both outputs (optical and coaxial). No hiccups.
Thanks for the great job!

pelwell:
@timg236 Are you happy with these changes?

timg236:
Yep, they look fine. I haven't had a chance to test it yet but am happy for it to be merged.

PR #2761, opened 2018-11-20
dtoverlays: Fix pi3-disable overlays

mbgg:
The root node definition of pi3-disable-wifi and
pi3-disable-bt is not correct. This breaks the
overlays. This patch fixes the root note definition.

Signed-off-by: Matthias Brugger <mbrugger@suse.com>

### DIFF
--- a/arch/arm/boot/dts/overlays/pi3-disable-bt-overlay.dts
+++ b/arch/arm/boot/dts/overlays/pi3-disable-bt-overlay.dts
@@ -8,7 +8,7 @@
        sudo systemctl disable hciuart
 */
 
-/{
+/ {
 	compatible = "brcm,bcm2708";
 
 	fragment@0 {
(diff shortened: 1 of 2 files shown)


mbgg:
Problem also present in rpi-4.20.y

pelwell:
dtc compiles it and the firmware applies it, so in what way are the overlays broken?

mbgg:
I generate the dtbo using
`dtc -R 0 -p 0 -@ -H epapr -I dts -O dtb [...]`

When I apply the overlay, it doesn't work. When I fix the root node definition to comply with the devicetree spec [1] it gets correctly applied. Apart, all other overlays have the rootnode defined like this, which gives a hint that something is fishy with this files.

[1] https://github.com/devicetree-org/devicetree-specification/blob/master/source/source-language.rst

pelwell:
Using two different versions of dtc, with your parameters, both are happy without the space between the `/` and the `{` - they produce identical output. From what I can see, the grammar does not demand whitespace except to separate elements that might otherwise be treated together as one. Since the brace is not part of any other syntax element, the presence of whitespace between it and the preceding path is completely optional.

Please upload a "bad" compiled overlay (.dtbo) and a "good" one, so I can see the difference.

mbgg:
error 40, 40 centimeters in front of the display. These overlays are actually working for me as well. Sorry for the noise.

PR #2763, opened 2018-11-21
[4.19 RFT] Alternative solution for reboot notify

lategoodbye:
The current solution for the reboot notify acc64abcc54e62726c47035d3f0e33ad56427dd6 is a little bit heavyweight. Using the shutdown callback is much smaller and would be a better candidate for upstreaming.

This pull request is only intended for testing, which was offered by @pelwell 

Thanks


pelwell:
I can confirm that using this PR allows a 3B+ to network reboot - with just the reversion it fails.

lategoodbye:
Thanks. Are you okay with adding your name as Tested-by ?

pelwell:
Sure - go ahead.
```
Tested-by: Phil Elwell <phil@raspberrypi.org>
```

PR #2768, opened 2018-11-26
Unicam/TC358743 fixups for colo(u)rspace handling.

6by9:
Discovered whilst testing GStreamer, it was trying UYVY whilst the current format was RGB3 and failing with unsupported colorimetry due to the tc358743 issue, although a subsequent attempt would work as we would then already be in UYVY mode.


6by9:
@pelwell I will send this upstream (along with my other patch for the tc358743), but this gets the chip working correctly with GStreamer and the V4L2 codec driver.

PR #2769, opened 2018-11-26
dtoverlays: Add i2c on 0&1 option to TC358743, ADV7282 and OV5647

6by9:
Adds the option of configuring i2c0 to be on GPIOs 0&1 as
this is of use on the Compute Module.

Also fixes the ov5647 overlay where the override enabled the wrong
fragments.


pelwell:
Looks good to me.

Issue #2475, opened 2018-03-27
lan78xx on 3B+ does not detect network link when ethernet cable has only 2 pairs wired up (100Mbps)


gtrainavicius:
I have a lengthy 100Mbps cable at my home (30+ meters) and it works well with Raspberry Pi 3B, as well as other network devices like laptops, however, Raspberry Pi 3B+ is unable to automatically recognize and negotiate the network link.

`dmesg` on 3B gives:

```
 IPv6: ADDRCONF(NETDEV_UP): eth0: link is not ready
 IPv6: ADDRCONF(NETDEV_CHANGE): wlan0: link becomes ready
 IPv6: ADDRCONF(NETDEV_CHANGE): eth0: link becomes ready
 smsc95xx 1-1.1:1.0 eth0: link up, 100Mbps, full-duplex, lpa 0xC3E1
```

the green and amber LEDs blink.

and on 3B+:

```
 libphy: lan78xx-mdiobus: probed
...
 IPv6: ADDRCONF(NETDEV_UP): wlan0: link is not ready
 IPv6: ADDRCONF(NETDEV_UP): eth0: link is not ready
 IPv6: ADDRCONF(NETDEV_CHANGE): wlan0: link becomes ready
```

If 100Mbps full duplex mode is forced on 3B+, the network connection starts working and the green LED starts blinking (note the 'Link detected'):

```
modep@modep:~ $ sudo ethtool eth0
Settings for eth0:
        Supported ports: [ TP MII ]
        Supported link modes:   10baseT/Half 10baseT/Full
                                100baseT/Half 100baseT/Full
                                1000baseT/Full
        Supported pause frame use: Symmetric Receive-only
        Supports auto-negotiation: Yes
        Advertised link modes:  10baseT/Half 10baseT/Full
                                100baseT/Half 100baseT/Full
                                1000baseT/Full
        Advertised pause frame use: Symmetric
        Advertised auto-negotiation: Yes
        Link partner advertised link modes:  10baseT/Half 10baseT/Full
                                             100baseT/Half 100baseT/Full
        Link partner advertised pause frame use: No
        Link partner advertised auto-negotiation: Yes
        Speed: 100Mb/s
        Duplex: Full
        Port: MII
        PHYAD: 1
        Transceiver: external
        Auto-negotiation: on
        Supports Wake-on: pumbag
        Wake-on: g
        Current message level: 0x00000007 (7)
                               drv probe link
        Link detected: no
[...]

PR #2771, opened 2018-11-27
Enable the LAN7800 PHY auto-downshift feature

pelwell:
Ethernet cables with faulty or missing pairs (specifically pairs C and
D) allow auto-negotiation to 1000Mbs, but do not support the successful
establishment of a link. Add a DT property, "microchip,downshift-after",
to configure the number of auto-negotiation failures after which it
falls back to 100Mbs. Valid values are 2, 3, 4, 5 and 0, where 0 means
never downshift.

Also add two new DT parameters so this behaviour can be controlled by
the user.

See: https://github.com/raspberrypi/linux/issues/2475


arch/arm/boot/dts/bcm283x-rpi-lan7515.dtsi
```diff
@@ -36,9 +46,11 @@
...
-		eee = <&ethernet>,"microchip,eee-enabled?";
-		tx_lpi_timer = <&ethernet>,"microchip,tx-lpi-timer:0";
-		eth_led0 = <&ethernet>,"microchip,led-modes:0";
-		eth_led1 = <&ethernet>,"microchip,led-modes:4";
+		eee = <&eth_phy>,"microchip,eee-enabled?";
+		tx_lpi_timer = <&eth_phy>,"microchip,tx-lpi-timer:0";
+		eth_led0 = <&eth_phy>,"microchip,led-modes:0";
+		eth_led1 = <&eth_phy>,"microchip,led-modes:4";
+		eth_downshift_after = <&eth_phy>,"microchip,downshift-after";
+		eth_max_speed = <&eth_phy>,"max-speed";
```
6by9: Have I missed this eth_max_speed parameter somewhere? I don't see it in the binding nor bcm283x-rpi-lan7515.dtsi

6by9 (commented):
Your commit text on "net: lan78xx: Support auto-downshift to 100Mb/s " will get comments should you try upstreaming.

10baseT and 100baseT require 2 pairs, so comments on single pair cat3 are erroneous.

10baseT and 100baseT require 2 pairs, whilst 1000baseT requires 4. Some installations have repurposed the spare pairs from 10 and 100baseT for other uses (eg passive PoE), which results in 1000baseT being negotiated but unable to fully establish.

Also in "BCM270X_DT: Add new Ethernet DT parameters" you've got a typo of "activatess" in the commit text.

Did you find cables to test this on? If not I'll grab my hacked cable from home and test tomorrow.

arch/arm/boot/dts/bcm283x-rpi-lan7515.dtsi
pelwell: The eth_max_speed parameter is defined above, but the `max-speed` property it sets is a common to all(?) PHYs, hence the lack of a vendor prefix.

pelwell:
I've fixed the pairs reference and the typo.

arch/arm/boot/dts/bcm283x-rpi-lan7515.dtsi
6by9: Probably my misunderstanding then.
I thought the overrides section could only change existing parameters. I can't see max-speed having been declared in the eth-phy node, so was confused as to what it was going to be changing. Can overrides add nodes?

6by9:
Still a reference to cat3 in "ARM: dts: bcm283x: Set downshift-after to 2" commit text if we really care.

arch/arm/boot/dts/bcm283x-rpi-lan7515.dtsi
pelwell: Overrides can add properties, and by enabling fragments they can add nodes.

6by9: Ah, I learn something new then :-)

pelwell:
I've reworded the 2nd commit message.

6by9:
Yes that seems to work pretty well.

I had minor reservations as my system uses an NFS root file system, but as lan7800 is builtin to kernel7.img on the SD card it works fine. I guess full network boot is going to fail in these conditions as the bootrom won't set this up, but not much that can be done there.
NFS booting at 10Mbit/s is a touch slow :-)

Are we worrying about upstreaming? Checkpatch complains about patch 2.
```
WARNING: DT binding docs and includes should be a separate patch. See: Documentation/devicetree/bindings/submitting-patches.txt

ERROR: spaces required around that '=' (ctx:VxW)
#49: FILE: drivers/net/phy/microchip.c:265:
+		u32 val= LAN78XX_PHY_CTRL3_AUTO_DOWNSHIFT;
 		       ^

ERROR: trailing statements should be on next line
#52: FILE: drivers/net/phy/microchip.c:268:
+		case 2:	val |= LAN78XX_PHY_CTRL3_DOWNSHIFT_CTRL_2;

ERROR: trailing statements should be on next line
#54: FILE: drivers/net/phy/microchip.c:270:
+		case 3:	val |= LAN78XX_PHY_CTRL3_DOWNSHIFT_CTRL_3;

ERROR: trailing statements should be on next line
#56: FILE: drivers/net/phy/microchip.c:272:
+		case 4:	val |= LAN78XX_PHY_CTRL3_DOWNSHIFT_CTRL_4;

ERROR: trailing statements should be on next line
#58: FILE: drivers/net/phy/microchip.c:274:
+		case 5:	val |= LAN78XX_PHY_CTRL3_DOWNSHIFT_CTRL_5;

total: 5 errors, 1 warnings, 0 checks, 62 lines checked
```
And patch 3 still references cat3.
I'm not fussed, so up to you whether you amend or not. Functionally it's good to go.

pelwell:
Patched up and merged - thanks for the review.

ghost:
@pelwell you might want to edit the original description on this bug, since AIUI the problem this patch was developed for is when the cable only has 2 pairs. This is not the same as a cat 3 cable. Just for super-completeness so there's no potential for misunderstanding in future.

pelwell:
Hmm - where have I heard that before? ;-)

Updated.

PR #2773, opened 2018-11-30
staging: bcm2835-camera: Fix stride on RGB3/BGR3 formats

6by9:
RGB3/BGR3 end up being 3 bytes per pixel, which meant that
the alignment code ended up trying to align using bitmasking
with a mask of 96.
That doesn't work, so align switch to an arithmetic alignment
for those formats.

Issues had been noted by Gordon with OpenCV, and the Foundation with Scratch, that QVGA wasn't working as the buffer size and stride calcs all went squiffy.

### DIFF
--- a/drivers/staging/vc04_services/bcm2835-camera/bcm2835-camera.c
+++ b/drivers/staging/vc04_services/bcm2835-camera/bcm2835-camera.c
@@ -1020,13 +1020,27 @@ static int vidioc_try_fmt_vid_cap(struct file *file, void *priv,
 			      1, 0);
 	f->fmt.pix.bytesperline = f->fmt.pix.width * mfmt->ybbp;
 	if (!mfmt->remove_padding) {
-		int align_mask = ((32 * mfmt->depth) >> 3) - 1;
-		/* GPU isn't removing padding, so stride is aligned to 32 */
-		f->fmt.pix.bytesperline =
-			(f->fmt.pix.bytesperline + align_mask) & ~align_mask;
+		if (mfmt->depth == 24) {
+			/*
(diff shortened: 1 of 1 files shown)


drivers/staging/vc04_services/bcm2835-camera/bcm2835-camera.c
```diff
@@ -1020,13 +1020,27 @@ static int vidioc_try_fmt_vid_cap(struct file *file, void *priv,
...
-		/* GPU isn't removing padding, so stride is aligned to 32 */
-		f->fmt.pix.bytesperline =
-			(f->fmt.pix.bytesperline + align_mask) & ~align_mask;
+		if (mfmt->depth == 24) {
+			/*
+			 * 24bpp is a pain as we can't use simple masking.
+			 * Min stride is width aligned to 16, times 24bpp.
+			 */
+			f->fmt.pix.bytesperline =
+				((f->fmt.pix.width + 15) & ~15) * mfmt->ybbp;
```
pelwell: Is there any reason not to replace mfmt->ybbp to 3? I find it clearer when read with the comment.

pelwell:
Apart from the inline comment/question, there's a spurious "align" in the commit message, but functionally it looks fine.

drivers/staging/vc04_services/bcm2835-camera/bcm2835-camera.c
6by9: James had made the same comment verbally.
mfmt->ybpp in this case will always be 3, therefore it can be replaced.

6by9:
Done both changes.

PR #2777, opened 2018-12-04
Revert "staging: vchiq_arm: Register a platform device for the audio …

6by9:
…driver"

This reverts commit ab59590ed562b89db51fe46cee5db96b9bc5abd8.

Issues have been observed in LibreElec as this was unconditionally
loading the audio driver instead of having the DT parameter to
enable it.

Includes a partial revert of 2147700eb7a1b9e55e0684f0749114ce35d61571
which fixed up the error handling.

### DIFF
--- a/drivers/staging/vc04_services/bcm2835-audio/bcm2835.c
+++ b/drivers/staging/vc04_services/bcm2835-audio/bcm2835.c
@@ -4,17 +4,15 @@
 #include <linux/platform_device.h>
 
 #include <linux/init.h>
-#include <linux/dma-mapping.h>
-#include <linux/of_device.h>
 #include <linux/slab.h>
 #include <linux/module.h>
+#include <linux/of.h>
 
 #include "bcm2835.h"
 
 static bool enable_hdmi;
 static bool enable_headphones;
 static bool enable_compat_alsa = true;
-static int num_channels = MAX_SUBSTREAMS;
 
 module_param(enable_hdmi, bool, 0444);
 MODULE_PARM_DESC(enable_hdmi, "Enables HDMI virtual audio device");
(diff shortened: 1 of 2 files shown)


6by9:
@popcornmix Fixing the issues being discussed on LibreElec Slack.
Need to discuss on the list (and in the office) what the correct mechanism is for loading this driver.

HiassofT:
I think you also need to (partially) revert https://github.com/raspberrypi/linux/commit/933bc853bb764e476b0b0f633588f46d20f1f76a, enabling bcm2835-audio via dtparam=audio=on no longer works as the of match table is gone

HiassofT:
I did a quick test, reverting 933bc85 restores the old behaviour, bcm2835-audio can be enabled via dtparam=audio=on as before

6by9:
Updated to revert 933bc85.

6by9:
@pelwell Reverting part of Stefan's patches from the mailing list as they make loading bcm2835-audio unconditional. LibreElec folk were seeing ALSA weirdness in last night's build.

lategoodbye:
Does it make sense to send my patch series to staging-next? The error description here is a little bit vague.

HiassofT:
@lategoodbye the current bcm2835-audio driver offers the feature to be optionally enabled via device tree, and in LibreELEC we didn't enable it because it's not needed (Kodi uses the userspace vc interface to do audio output) and it works around a Kodi bug dealing with multiple ALSA cards.

The Kodi bug is something separate to solve (I'll try to look into it the next weeks/months when I have time), but for now it would be good to keep the current state to be able to disable bcm2835-audio.

@popcornmix should be able to comment on the future plans for kodi and if the vc "Pi Sink" driver is going to be dropped and replaced by the ALSA interface or not.

Using the ALSA driver would certainly make sense as it'll be the same interface that all other platforms are using but AFAIK there are still some issues with bcm2835-audio and multichannel ALSA output.

lategoodbye:
@HiassofT So the only "regression" is the unexpected probing of bcm2835-audio?

HiassofT:
@lategoodbye yes, exactly

PR #2781, opened 2018-12-09
dtoverlays: fe-pi-audio: fix sgtl5000 compatible string

lopsided98:
The compatible string was set to "fepi,sgtl5000", which worked for some reason in 4.14, but does not work in 4.19, presumably due to some change in the kernel matching logic. The correct string is "fsl,sgtl5000". This overlay still works correctly with 4.14 with this change.


pelwell:
Thanks!

Issue #2690, opened 2018-09-24
Please include hid-bigbenff.c driver


hzulla:
Hi there,

my driver submission has just been accepted for the upcoming 4.20 mainline kernel by the input maintainers:
https://git.kernel.org/pub/scm/linux/kernel/git/jikos/hid.git/commit/?h=for-4.20/bigbenff&id=256a90ed9e46b270bbc4e15ef05216ff049c3721

The BigBen Interactive PS3OFMINIPAD kids' gamepad is a Playstation3 accessory and was designed for smaller hands. Linux didn't support its rumble effects, so the device's HID protocol needed to be reverse-engineered to write this driver.

I wrote this driver for the Pi. As Retropie is one of the projects my kids use with the Raspberry Pi, please include this driver for the Pi's kernel. The patch applies cleanly on 4.14 with no changes, so there's no backporting to do.

Thank you.

pelwell:
The corresponding config change (`CONFIG_HID_BIGBEN_FF=m`) just builds the 9KB module.

Any objection to me merging the driver and config patch, @popcornmix?

pelwell:
Cool - no objection. Merged.

hzulla:
Thank you!

PR #2785, opened 2018-12-17
Add hid-bigbenff to list of "have_special_driver" for rpi-4.14 kernel.

hzulla:
Hello,

thanks for [adding hid-bigbenff.c to the raspberry pi kernel](https://github.com/raspberrypi/linux/issues/2690). The driver was written for 4.20.

I now tested the driver with the current 4.14 release kernel and the "NEXT" 4.19 kernel on my rpi.

The driver works fine without any change for 4.19, but on the current 4.14 kernel it needs this patch.

Again, this patch is not needed for 4.19 any more, as the handling of special drivers had been changed with `hid-quirks.c`.

Thank you.


pelwell:
I'd normally expect patches to unmodified upstream files such as this to go upstream first, but since this clearly has a very short lifespan (possibly only weeks) that would be a pointless waste of time.

hzulla:
Thank you.

PR #2789, opened 2018-12-27
USB Audio: generic DSD detection for XMOS-based implemtations

gkkpch:
The purpose of this backport is to implement the generic method to detect many DSD-capable implementations from XMOS/ Thesycon and Mytek as found in kernel 4.19 LTS.
This simple patch makes it available for kernel version 4.14.y

Additionally the patch implements support for the following DSD-capable devices:

AURALiC VEGA
Wyred 4 Sound DAC-2 DSD
Matrix Audio X-SPDIF 2
OPPO Sonica
OPPO UDP-205
OPPO HA-2
LH Labs VI DAC Infinity
SMSL D1 DAC
Encore mDSD
Hegel HD12 DSD
NuPrime Audio DAC-9
NuPrime DAC-10
Furutech ADL Stratos
Bryston BDA3
Denafrips Ares DAC
Eastern El. MiniMax Tube DAC Supreme
Unison Research Unico CD Due
CH Precision C1 DAC
Singxer F-1 converter board
HDTA Serenade DSD
M2Tech Young MkIII
PS Audio NuWave DAC
Audiolab M-DAC+
W4S DAC-2v2SE
Mutec MC3+ USB
MSB Technology

PR #2790, opened 2018-12-27
pr2789 remix

pelwell:
Refactor #2789 as a straight back-port commit and a few additional USB IDs in some switch statements.


popcornmix:
@gkkpch happy with this version?

gkkpch:
Yes, this looks fine to me, thanks.
I will use that for the other platforms as well, so the patch will be the same

PR #2801, opened 2019-01-09
gpio-ir: change default pull configuration to up

HiassofT:
IR receivers like the TSOP series from Vishay and compatible ones
have active-low open collector outputs with an internal pull up of
about 30k (according to the TSOP datasheets).

Activating a pull-down resistor on the GPIO will make it work against
the pull-up in the IR receiver and brings the idle input voltage down
to about 1.9V (measured on a RPi3B+ with a TSOP4438). While that's
usually enough to make the RPi see a high signal it's certainly not
optimal and may even fail when using an IR receiver with a weaker pull-up.

Switching the default GPIO pull to "up" results in an input voltage
level of about 3.3V and ensures that an idle state (high signal) will
be detected if no IR receiver is attached.


pelwell:
I'd normally be reluctant to change the default, but it looks like this was just wrong.

PR #2803, opened 2019-01-10
dtoverlays: Added 1-bit SDIO overlay on GPIO 34-37.

RICCIARDI-Adrien:


pelwell:
Rather than add a separate overlay, would you be happy with a parameter on the original to switch to GPIOs 34-37?:
```
...
        fragment@2 {
                target = <&gpio>;
                __overlay__ {
                        sdio_1bit_pins: sdio_1bit_pins {
                                brcm,pins = <22 23 24 25>;
                                brcm,function = <7>; /* ALT3 = SD1 */
                                brcm,pull = <0 2 2 2>;
                        };
                };
        };

        fragment@3 {
                target = <&sdio_1bit_pins>;
                __dormant__ {
                        brcm,pins = <34 35 36 37>;
                };
        };

        fragment@4 {
                target-path = "/aliases";
                __overlay__ {
                        mmc1 = "/soc/sdio@7e300000";
                };
        };


        __overrides__ {
                poll_once = <&sdio_1bit>,"non-removable?";
                sdio_overclock = <&sdio_1bit>,"brcm,overclock-50:0";
                gpios_34_37 = <0>,"=3";
        };
};
```
You would apply it with:
```
dtoverlay=sdio-1bit,gpios_34_37
```

pelwell:
I've noticed that both 1-bit overlays seem to be allocating pins for two data lines when only one is used. Can you try your overlay (or mine, if you can figure it out) with:
```
        brcm,pins = <34 35 36>;
        brcm,function = <7>; /* ALT3 = SD1 */
        brcm,pull = <0 2 2>;
```

It also occurs to me that with enough parameters we could do away with thie sdio-1bit overlay altogether, replacing it with:
```
dtoverlay=sdio,gpios_22_25
```
or
```
dtoverlay=sdio,gpios_22_24
```
Let me know what you think.

pelwell:
1-bit mode for SDIO uses SD pin 8  (DAT1, GPIO 25/37/51) as a dedicated IRQ line, so you can ignore the `brcm,pins = <34 35 36>;` suggestion above.

RICCIARDI-Adrien:
Hi,
The DT parameter is far more better, but I didn't know how to do it !
The second pin (DAT1) is mandatory in 1-bit SDIO (it reports interrupts).

Yes, it could be possible to use sdio-overlay for all cases thanks to `bus_width` parameter. It would be great to merge both overlays in a second time, but I don't know if it's possible as people are using both files.

pelwell:
I can make the appropriate changes to the sdio overlay, then close this PR when it's don. The existing sdio-1bit overlay can be marked as deprecated in the README, but kept around for the time being.

RICCIARDI-Adrien:
Ok, looks fine to me.

RICCIARDI-Adrien:
When modifying sdio-overlay, make sure that only DAT0 and DAT1 pins are configured as SDIO when using 1-bit mode (for instance, I'm using pins 38-39 as UART0 hardware flow control, so I need them free for another DT overlay).

pelwell:
See https://github.com/raspberrypi/linux/pull/2804.

PR #2805, opened 2019-01-10
Update rpi-5.0-y to support upstream touchscreen

vianpl:
The upstream driver that was accepted in 5.0-rc1 changed some of the bindings and configuration flags. This small series updates everything.

I changed some of the node names to better match the upstream bindings. In an ideal world I would have also like to rename the overlay to something like _rpi-fw-touchscreen-overlay.dtso_ as _rpi-ft5406_ feels too low level. I think it would imply a fw change so I left it for another occasion. 

This was tested with the latest Raspbian image available.


popcornmix:
Thanks!

pelwell:
The other defconfigs will also need updating.

PR #2806, opened 2019-01-11
raspberrypi firmware driver to log build variant and hash during probe.

6by9:
Requires the associated firmware change to work, but will behave sensibly if it is absent.

### DIFF
--- a/drivers/firmware/raspberrypi.c
+++ b/drivers/firmware/raspberrypi.c
@@ -227,21 +227,39 @@ static const struct attribute_group rpi_firmware_dev_group = {
 static void
 rpi_firmware_print_firmware_revision(struct rpi_firmware *fw)
 {
+	static const char * const variant_strs[] = {
+		"unknown",
+		"start",
+		"start_x",
+		"start_db",
+		"start_cd",
+	};
+	const char *variant_str = "cmd unsupported";
 	u32 packet;
+	u32 variant;
+	struct tm tm;
 	int ret = rpi_firmware_property(fw,
 					RPI_FIRMWARE_GET_FIRMWARE_REVISION,
 					&packet, sizeof(packet));
 
-	if (ret == 0) {
-		struct tm tm;
+	if (ret)
+		return;
 
-		time64_to_tm(packet, 0, &tm);
+	ret = rpi_firmware_property(fw, RPI_FIRMWARE_GET_FIRMWARE_VARIANT,
+				    &variant, sizeof(variant));
 
-		dev_info(fw->cl.dev,
-			 "Attached to firmware from %04ld-%02d-%02d %02d:%02d\n",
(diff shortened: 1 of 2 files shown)


drivers/firmware/raspberrypi.c
```diff
@@ -227,21 +227,54 @@ static const struct attribute_group rpi_firmware_dev_group = {
...
 
-		dev_info(fw->cl.dev,
-			 "Attached to firmware from %04ld-%02d-%02d %02d:%02d\n",
-			 tm.tm_year + 1900, tm.tm_mon + 1, tm.tm_mday,
-			 tm.tm_hour, tm.tm_min);
-	}
+	if (ret)
+		variant = ARRAY_SIZE(variant_strs) - 1;
+	if (variant >= ARRAY_SIZE(variant_strs))
+		variant = 0;
```
pelwell: I think the "not supported" case could be handled better. As it stands, as well as having to remember to keep the "cmd unsupported" option at the end of the array, if the firmware returns a value which is one larger than we expect it will be reported as "cmd unsupported".

6by9: Preferred alternative?
```
const char *variant_str = "cmd unsupported";
if (!ret) {
   if (variant >= ARRAY_SIZE(variant_strs))
      variant = 0;
   variant_str = variant_str[variant];
}
dev_info(....., variant_str);
```
Removes the need for the extra magic entry.

pelwell: Much better.

6by9:
Sorry, introduced blank line errors there. Repushing in a mo.

6by9:
Fixed those. Hopefully good to go now.

pelwell:
I just noticed the & on the `hash` array - not wrong, but not strictly necessary - I don't know what the kernel devs would say.

Other than that, LGTM.

6by9:
> I just noticed the & on the hash array - not wrong, but not strictly necessary - I don't know what the kernel devs would say.

Happy to update.

nullr0ute:
Can we also get these sent upstream, they'll be universally useful for debug :)

6by9:
> Can we also get these sent upstream, they'll be universally useful for debug :)

Upstreaming stuff isn't a hugely high priority for us, but it'll get added to the list.
I mainly picked the task up so that the firmware side got done, because that is the bit that no one outside of RPT can do.

lategoodbye:
@6by9 Any plans to upstream this month? I only need to know.

6by9:
@lategoodbye Sorry, snowed under with other stuff at the moment, so very unlikely to be sorting patches to send this upstream.

Assuming vc04_services hasn't moved too far, I will try to sort out my patches for the camera stack as those were pretty much sorted.

lategoodbye:
No problem, i try to take care at least of the variant patch.

vc04_services has been very quiet since beginning of the new year. So good luck.

PR #2808, opened 2019-01-14
overlays: add better target to rpi-ft5406

vianpl:
The current overlay's target is set to "/firmware". This is worng as the
real firmware node is located at "/soc/firmware". This is not ideal as
the touchcreen didn't appear in the final device tree as a child of the
firmware node.

the overlay still worked fine as the driver uses 'of_get_parent()' to
get the firmware node which in the end pointed to the right place.

Signed-off-by: Nicolas Saenz Julienne <nsaenzjulienne@suse.de>

### DIFF
--- a/arch/arm/boot/dts/overlays/rpi-ft5406-overlay.dts
+++ b/arch/arm/boot/dts/overlays/rpi-ft5406-overlay.dts
@@ -5,14 +5,12 @@
 	compatible = "brcm,bcm2708";
 
 	fragment@0 {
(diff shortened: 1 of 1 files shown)


vianpl:
I did test this with the latest Raspbian.

pelwell:
I've not tested this update, but it looks plausible, it's a bleeding edge kernel, and I trust you.

vianpl:
Thanks!

PR #2816, opened 2019-01-18
bcm2835-mmc: Fix DMA channel leak

l1k:
The BCM2835 MMC host driver requests a DMA channel on probe but neglects
to release the channel in the probe error path and on driver unbind.

I'm seeing this happen on every boot of the Compute Module 3: On first
driver probe, DMA channel 2 is allocated and then leaked with a "could
not get clk, deferring probe" message. On second driver probe, channel 4
is allocated.

Fix it.

Signed-off-by: Lukas Wunner <lukas@wunner.de>
Cc: @pavlic244

### DIFF
--- a/drivers/mmc/host/bcm2835-mmc.c
+++ b/drivers/mmc/host/bcm2835-mmc.c
@@ -1441,7 +1441,8 @@ static int bcm2835_mmc_probe(struct platform_device *pdev)
 	addr = of_get_address(node, 0, NULL, NULL);
 	if (!addr) {
 		dev_err(dev, "could not get DMA-register address\n");
-		return -ENODEV;
+		ret = -ENODEV;
+		goto err;
 	}
 	host->bus_addr = be32_to_cpup(addr);
 	pr_debug(" - ioaddr %lx, iomem->start %lx, bus_addr %lx\n",
--- a/drivers/mmc/host/bcm2835-mmc.c
+++ b/drivers/mmc/host/bcm2835-mmc.c
@@ -1400,10 +1400,16 @@ static int bcm2835_mmc_add_host(struct bcm2835_host *host)
 	}
 
 	mmiowb();
-	mmc_add_host(mmc);
+	ret = mmc_add_host(mmc);
+	if (ret) {
+		dev_err(dev, "could not add MMC host\n");
+		goto free_irq;
+	}
 
 	return 0;
 
+free_irq:
+	free_irq(host->irq, host);
 untasklet:
 	tasklet_kill(&host->finish_tasklet);
 
--- a/drivers/mmc/host/bcm2835-mmc.c
+++ b/drivers/mmc/host/bcm2835-mmc.c
@@ -1391,9 +1391,9 @@ static int bcm2835_mmc_add_host(struct bcm2835_host *host)
 	init_waitqueue_head(&host->buf_ready_int);
 
 	bcm2835_mmc_init(host, 0);
-	ret = devm_request_threaded_irq(dev, host->irq, bcm2835_mmc_irq,
-					bcm2835_mmc_thread_irq, IRQF_SHARED,
-					mmc_hostname(mmc), host);
+	ret = request_threaded_irq(host->irq, bcm2835_mmc_irq,
+				   bcm2835_mmc_thread_irq, IRQF_SHARED,
+				   mmc_hostname(mmc), host);
 	if (ret) {
 		dev_err(dev, "Failed to request IRQ %d: %d\n", host->irq, ret);
 		goto untasklet;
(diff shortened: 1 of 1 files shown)


lategoodbye:
Nice catch. It looks like the that irq isn't freed in case bcm2835_mmc_add_host() fails.

Btw upstream bcm2835.c is also affected :-(

l1k:
Ok I can see to it that I fix it in the upstream driver as well. That one is actually for the other interface (the non-Arasan one), curious that it has the same issue.

lategoodbye:
After this pull request has been merge, i will prepare patches to fix the missing return code handling of mmc_of_parse().

Similiar to this:
https://patchwork.kernel.org/patch/10741805/

l1k:
Further scrutiny has revealed more issues in the probe/remove routines of the downstream Arasan driver and I've updated this pull to address them. I've also posted the DMA channel leak for the upstream non-Arasan driver to the list, together with a bunch of cleanups. So far this is all compile-tested only, I won't be able to test on real hardware before returning to the office on Tuesday. Stefan, would you be able to give the two series a whirl? Thanks!

lategoodbye:
@l1k I'm busy with the pull requests for Linux 5.1, but i try to take a look at them. Please always add a note (e.g. in the cover letter) in case a patch is only compile tested. Because as a maintainer i assume all patches has been tested with at least 1 platform before submission.

lategoodbye (changes requested):
I prefer to drop free_irq() instead. Or do you see a problem with the device managed interrupt?

l1k:
I don't think dropping free_irq() is an option. As I've explained in the commit message, invocation of the interrupt handler needs to be avoided after driver structures have been torn down. If the interrupt is device-managed, it won't be freed until `bcm2835_mmc_remove()` has finished and that function frees the `struct mmc_host` and `struct bcm2835_host`, which are accessed in the hardirq and threaded interrupt handler, potentially leading to a use-after-free. Note that it's not sufficient to quiesce the interrupt at the hardware level because the interrupt is requested with `IRQF_SHARED`, so it's possible that the handler runs if an interrupt for another device is signaled which shares the interrupt. (It's not clear to me from the datasheet which other devices share the interrupt. According to `Documentation/devicetree/bindings/interrupt-controller/brcm,bcm2835-armctrl-ic.txt`, bank 2 interrupt 30 is used by `VC_ARASANSDIO`, no word about devices sharing the interrupt. Maybe `IRQF_SHARED` isn't really needed here, but independently, freeing the interrupt after driver structures have been torn down seems unnecessarily risky. Big no-no.)

lategoodbye:
Thanks for the explanation. Yes, the IRQF_SHARED is also wrong to me, because no other driver should use it. But changing from devm_request_threaded_irq to request_threaded_irq isn't sufficient in your commit. You will need to call free_irq in all error cases after the irq has been requested.

l1k:
Huh? There *are* no error cases after the IRQ has been requested: `bcm2835_mmc_add_host()` contains no further error checks after the call to `request_threaded_irq()` and `bcm2835_mmc_add_host()` is the last function called by `bcm2835_mmc_probe()` which might fail.

The call to `mmc_add_host()` is not error checked at the moment, it should be and *then* would a `free_irq()` be necessary in an error path. However, the question is how much time should be wasted fixing this driver's bugs. There's also deadlocks in the reset code. Upstream uses `sdhci-iproc.c` but doesn't support DMA. What's the plan? Add DMA to `sdhci-iproc.c` and scrap this driver? Why is the upstream SD host driver so similar to the downstream Arasan driver, what's the history here?

lategoodbye:
Sorry, i confuse it with bcm2835 sdhost.

AFAIK the Arasan core of the BCM2835 doesn't support DMA according to SDHCI spec. So DMA implementation could be a hassle. I didn't invest any time to estimate the possible effort. But it was my proposal to drop the old bcm2835-mmc (which also didn't support DMA) in upstream and to use sdhci-iproc instead.

The upstream bcm2835 sdhost was derived from an older version of the downstream bcm2835 sdhost. But didn't know the history of the downstream ones.

l1k:
I've amended this pull with a commit to handle errors occurring in `bcm2835_mmc_add_host()` and successfully tested both the series contained herein as well as the one for the upstream sdhost driver (which I've submitted to the list), so this pull is ready to go in from my POV. Thanks.

lategoodbye:
Thank you, i'm fine with this pull request.

lategoodbye:
In the meantime the upstream patches has been merged.

l1k:
Right, however those were for the upstream sdhost driver. The present pull is for the downstream Arasan driver (which isn't present in upstream). (Just in case this wasn't clear already.)

l1k:
@pelwell: These patches are missing on the rpi-4.19.y branch. Should I send another pull or can you cherry-pick them? Thanks!

pelwell:
They're cherry-picked now.

PR #2823, opened 2019-01-23
V4L2 compliance fixes, and 64 bit kernel improvements

6by9:
Two V4L2 compliance test fixes for the camera driver.

Hopefully fixes mmal-vchiq to support 64 bit kernels. (Fails at present due to dma_alloc_coherent failing, so unable to confirm).

Fixes up vc_mem for 32 bit userspace/64 bit kernel, so vcdbg now works under those conditions. Cleans up checkpatch issues with it too.

PR #2824, opened 2019-01-24
vcio 64bit kernel/32bit user compat ioctl handling.

6by9:
Equivalent tweaks to vcio as #2823 provides for vc-mem.

Adds compat_ioctl handling as IOCTL_MBOX_PROPERTY changes due to char* being of a different size.
Tested with https://github.com/6by9/rpi3-gpiovirtbuf, with ```rpi3-gpiovertbuf s 130 [0|1]``` to toggle the power LED on/off on a 3B+.


popcornmix:
Looks fine. @pelwell ?

PR #2825, opened 2019-01-24
clk: clk-bcm2835: Use %zd when printing size_t

6by9:
The debug text for how many clocks have been registered
uses "%d" with a size_t. Correct it to "%zd".


popcornmix:
Looks fine. Was going to mention upstream, but this line is from downstream so no problems.

6by9:
> Was going to mention upstream, but this line is from downstream so no problems.

Yup, I'd already checked that :-)

Sorry, more of these to come, but I'll try to batch them up rather than have lots of individual commits.

PR #2828, opened 2019-01-25
64bit cleanups

6by9:
Cleans up/fixes the 64bit build of bcm2708_fb and dwc_otg.
Fixes a couple of minor issues in the V4L2 drivers.

### DIFF
--- a/drivers/usb/host/dwc_otg/dwc_otg_hcd.c
+++ b/drivers/usb/host/dwc_otg/dwc_otg_hcd.c
@@ -929,6 +929,8 @@ static void dwc_otg_hcd_free(dwc_otg_hcd_t * dwc_otg_hcd)
 	DWC_TIMER_FREE(dwc_otg_hcd->conn_timer);
 	DWC_TASK_FREE(dwc_otg_hcd->reset_tasklet);
 	DWC_TASK_FREE(dwc_otg_hcd->completion_tasklet);
+	DWC_DMA_FREE(dev, 16, dwc_otg_hcd->fiq_state->dummy_send,
+		     dwc_otg_hcd->fiq_state->dummy_send_dma);
 	DWC_FREE(dwc_otg_hcd->fiq_state);
 
 #ifdef DWC_DEV_SRPCAP
@@ -1021,7 +1023,8 @@ int dwc_otg_hcd_init(dwc_otg_hcd_t * hcd, dwc_otg_core_if_t * core_if)
 		for (i = 0; i < num_channels; i++) {
 			hcd->fiq_state->channel[i].fsm = FIQ_PASSTHROUGH;
 		}
-		hcd->fiq_state->dummy_send = DWC_ALLOC_ATOMIC(16);
+		hcd->fiq_state->dummy_send = DWC_DMA_ALLOC_ATOMIC(dev, 16,
+							 &hcd->fiq_state->dummy_send_dma);
 
 		hcd->fiq_stack = DWC_ALLOC(sizeof(struct fiq_stack));
 		if (!hcd->fiq_stack) {
--- a/drivers/usb/host/dwc_otg/dwc_otg_hcd.c
+++ b/drivers/usb/host/dwc_otg/dwc_otg_hcd.c
@@ -1041,8 +1041,8 @@ int dwc_otg_hcd_init(dwc_otg_hcd_t * hcd, dwc_otg_core_if_t * core_if)
 		 * moderately readable array casts.
 		 */
 		hcd->fiq_dmab = DWC_DMA_ALLOC(dev, (sizeof(struct fiq_dma_channel) * num_channels), &hcd->fiq_state->dma_base);
-		DWC_WARN("FIQ DMA bounce buffers: virt = 0x%08x dma = 0x%08x len=%d",
-				(unsigned int)hcd->fiq_dmab, (unsigned int)hcd->fiq_state->dma_base,
+		DWC_WARN("FIQ DMA bounce buffers: virt = %px dma = %pad len=%zu",
+				hcd->fiq_dmab, &hcd->fiq_state->dma_base,
 				sizeof(struct fiq_dma_channel) * num_channels);
 
 		DWC_MEMSET(hcd->fiq_dmab, 0x6b, 9024);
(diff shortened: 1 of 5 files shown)


6by9:
These are currently cherry-picks onto a fresh branch - I'm waiting for my build to complete before I'd declare them as definitely good.

drivers/usb/host/dwc_otg/dwc_otg_hcd.c
```diff
@@ -1524,7 +1524,7 @@ int fiq_fsm_setup_periodic_dma(dwc_otg_hcd_t *hcd, struct fiq_channel_state *st,
 		 * state depending on transaction progress.
 		 */
 		blob = (struct fiq_dma_blob *) hcd->fiq_state->dma_base;
-		st->hcdma_copy.d32 = (uint32_t) &blob->channel[hc->hc_num].index[0].buf[0];
+		st->hcdma_copy.d32 = (dma_addr_t) &blob->channel[hc->hc_num].index[0].buf[0];
```
pelwell: This just looks wrong. It at least needs a comment.

drivers/usb/host/dwc_otg/dwc_otg_hcd.c
```diff
@@ -1573,7 +1573,7 @@ int fiq_fsm_setup_periodic_dma(dwc_otg_hcd_t *hcd, struct fiq_channel_state *st,
 			ptr = qtd->urb->buf + frame_desc->offset;
 			/* Point the HC at the DMA address of the bounce buffers */
 			blob = (struct fiq_dma_blob *) hcd->fiq_state->dma_base;
-			st->hcdma_copy.d32 = (uint32_t) &blob->channel[hc->hc_num].index[0].buf[0];
+			st->hcdma_copy.d32 = (dma_addr_t) &blob->channel[hc->hc_num].index[0].buf[0];
```
pelwell: Ditto here.

6by9:
Sorry, buried in looking at other drivers. Will address your comments tomorrow and repush.

It is rather tortuous in there working out what is being stored, but memory says that it was programming the DMA address of the buffer into the peripheral.
However having seen that ```&``` at the start you have got me wondering too.

drivers/usb/host/dwc_otg/dwc_otg_hcd.c
6by9: Looking at it, I can't see how the existing code is correct. ```&blob->....buf[0]``` is going to be a kernel virtual address.

```st->hcdma_copy``` is a ```union hcdma_data```, description
```
/**
 * This union represents the bit fields in the Host DMA Address
 * Register used in Descriptor DMA mode.
 */
..
		/** Start Address of Descriptor List */
		unsigned dma_addr:21;
```

Other places where it is assigned it is given a dma_addr_t
Line 1675:
```
	st->hcdma_copy.d32 = qtd->urb->dma;
```
and ```qtd->urb->dma``` is a ```dwc_dma_t``` (```typedef dma_addr_t dwc_dma_t```)
Line 1854
```
st->hcdma_copy.d32 = hc->align_buff;
```
```hc->align_buff``` is a ```dwc_dma_t```
Except line 1856
```
				st->hcdma_copy.d32 = ((unsigned long) hc->xfer_buff & 0xFFFFFFFF);
```
where ```uint8_t *xfer_buff;```

So ```st->hcdma_copy.d32``` is being treated as a dma_addr_t in most cases, but this one and a couple of others appear to be getting a kernel virtual address. On 64bits that is going to be too big for the 32bit register, so is that ever going to work at the hardware level?

@P33M could you advise?

6by9:
As this USB one seems to be a tad more involved than it at first seemed, I'll push move the other commits to a new PR and then reduce the scope of this PR to just the one USB patch when we have got to the bottom of what is happening.

drivers/usb/host/dwc_otg/dwc_otg_hcd.c
P33M: We're casting a 32-bit DMA address (hcd->fiq_state->dma_base) into a "struct fiq_dma_blob \*" to do pointer arithmetic. The "fiq_dma_blob" munging is because we're stitching together up to 6 188-byte separate DMA transfers, so instead of porting the DMA pool API into the FIQ we just use array addressing on the DMA base address within a 9kByte contiguous buffer. The buffer is allocated via dma_alloc_coherent() so dma_sync_* calls aren't required.

hc->align_buf is required because of the crippled DMA engine inside the OTG core - it can only DMA to addresses that start at a 4-byte boundary so if we get a URB with a urb->dma address with bottom 2 bits nonzero, a separate buffer is allocated for the core to use.

The DMA address registers are 32 bits wide but the hardware can only address 1G of SDRAM total. If dma_addr_t ends up being 64 bits then we need to guarantee the top 66 bits are 0, or we're going to have a bad time.

6by9: Thanks. In which case I believe the cast to dma_addr_t is correct, and it wants a comment to explain what the heck is going on.

I think line 1856 ```st->hcdma_copy.d32 = ((unsigned long) hc->xfer_buff & 0xFFFFFFFF);``` should also be a cast to dma_addr_t.
Then again there appears to be some nasty stuff around line 1243 where ```hc->xfer_buff``` being defined as a ```void*``` but gets assigned from ```urb->dma``` (a ```dma_addr_t```), or ```urb->buf``` (a ```void*```) dependent on ```hcd->core_if->dma_enable```. That looks like a module parameter that should have a large warning notice on it.
I think at this point I want to stop looking in case I find further nastiness! It compiles cleanly on both architectures and works in the default configuration, so that's good enough.

dma_addr_t is 64 bits on arm64, but the top 66 bits will be 0 (or is it a 0xc0000000 cache alias?) on BCM283[5|6|7], so all is good.

6by9:
Cleaned up, but not tested. Will do so in the morning.
We gained a patch too as I found another nasty hack with the dummy_send buffer.

6by9:
Tested as far as I can.
All seems fine on 32bit.
There's no regression on 64bit. Mouse & keyboard work fine. Both with and without these patches it struggles to stream from a UVC webcam, dropping more and more frames as you increase the resolution. From comments by P33M over FIQ support on aarch64, that is probably to be expected.

@P33M could you review too please?

6by9:
Rebased to 4.19.19.

PR #2837, opened 2019-01-29
mfd: Add rpi_sense_core of compatible string

XECDesign:


pelwell:
This patch is necessary because of commit af503716ac1444db61d80cb6d17cfe62929c21df. In particular:
```
NOTE: This patch may break out-of-tree drivers that were relying on this
      behavior, and only had an I2C device ID table even when the device
      was registered via OF. There are no remaining drivers in mainline
      that do this, but out-of-tree drivers have to be fixed and define
      a proper OF device ID table to have module auto-loading working.
```
That means we'll need to trawl the code looking for `i2c:` modaliases that don't have a corresponding `of:` modalias.

PR #2838, opened 2019-01-29
Unicam/OV5647 fixes

6by9:
First patch was a 4.14 commit that hadn't been cherry-picked forward.

Second is so the GPIO calls don't throw a wobbler on the 3B/3B+ GPIO expander that needs to sleep.

Third is a change to the timing of when the sensor gets powered up/down within the Unicam driver, as discussed on https://www.raspberrypi.org/forums/viewtopic.php?f=43&t=232437

### DIFF
--- a/arch/arm/boot/dts/bcm270x.dtsi
+++ b/arch/arm/boot/dts/bcm270x.dtsi
@@ -152,6 +152,13 @@
 		regulator-max-microvolt = <3300000>;
 		regulator-always-on;
 	};
+
+	__overrides__ {
+		cam0-pwdn-ctrl;
+		cam0-pwdn;
+		cam0-led-ctrl;
+		cam0-led;
+	};
 };
 
 &vc4 {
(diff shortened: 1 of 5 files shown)


6by9:
Sorry, just noticed that the compatible string being requested is wrong. It should be ```ovti,ov5647```, not just ```ov5647```. Updating....

6by9:
OK, done.

pelwell:
The first two commits are fine. The third - the unicam stream start sequencing patch - looks plausible, and I assume that it works. The only thing I would question is the naming of the open & close functions. I know they are static, but wouldn't a name that made it clear it wasn't a framework function be more appropriate?

6by9:
What are you suggesting? I'd based it on the other sensor receiver drivers:
- Atmel ISC ```isc_open``` in https://github.com/raspberrypi/linux/blob/rpi-4.19.y/drivers/media/platform/atmel/atmel-isc.c#L1526
- TI ```vpe_open``` in https://github.com/raspberrypi/linux/blob/rpi-4.19.y/drivers/media/platform/ti-vpe/vpe.c#L2240
- TI VPFE ```vpfe_open``` in https://github.com/raspberrypi/linux/blob/rpi-4.19.y/drivers/media/platform/am437x/am437x-vpfe.c#L1219

This change makes the behaviour almost identical to the Atmel driver.

pelwell:
The behaviour looks fine. I thought the name made it look a bit generic, but unicam is your common prefix (not bcm2835_unicam) and it is a unique prefix so I have no objection.

PR #2839, opened 2019-01-29
More 64bit build cleanups

6by9:
Debug messages using the wrong format specifiers again.

### DIFF
--- a/drivers/input/touchscreen/rpi-ft5406.c
+++ b/drivers/input/touchscreen/rpi-ft5406.c
@@ -218,8 +218,8 @@ static int ft5406_probe(struct platform_device *pdev)
 
 	if (!ts->ts_base) {
 		dev_warn(dev,
-		"set failed, trying get (err:%d touchbuf:%x virt:%p bus:%x)\n",
-		err, touchbuf, ts->ts_base, ts->bus_addr);
+		"set failed, trying get (err:%d touchbuf:%x virt:%p bus:%pad)\n",
+		err, touchbuf, ts->ts_base, &ts->bus_addr);
 
 		err = rpi_firmware_property(
 				fw,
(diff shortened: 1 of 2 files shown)


drivers/input/touchscreen/rpi-ft5406.c
```diff
@@ -218,8 +218,8 @@ static int ft5406_probe(struct platform_device *pdev)
 
 	if (!ts->ts_base) {
 		dev_warn(dev,
-		"set failed, trying get (err:%d touchbuf:%x virt:%p bus:%x)\n",
-		err, touchbuf, ts->ts_base, ts->bus_addr);
+		"set failed, trying get (err:%d touchbuf:%x virt:%p bus:%pad)\n",
+		err, touchbuf, ts->ts_base, &ts->bus_addr);
```
popcornmix, line 222: Isn't the contents of this more useful than the address?

pelwell: The kernel printk physical address support expects a pointer to the address.

popcornmix: Ok

6by9: As Phil says, it wants a reference, which is a right pain as you can't then print "base+offset" without assigning it to a variable first. Hence ```bo->paddr + fb->offsets[0]``` being changed to ```&fbinfo->base``` (which has the same value) in a couple of places.

Docs at https://github.com/raspberrypi/linux/blob/rpi-4.19.y/Documentation/core-api/printk-formats.rst#dma-address-types-dma_addr_t
```
DMA address types dma_addr_t
----------------------------
	%pad	0x01234567 or 0x0123456789abcdef

For printing a dma_addr_t type which can vary based on build options,
regardless of the width of the CPU data path.

Passed by reference.
```

PR #2845, opened 2019-02-01
Platform driver DMA config, and vcsm-cma fixes for 64 bit

6by9:
The dma config stuff is so that we get the correct cache aliases from vc-sm-cma and similar. I do need to bounce it off upstream at some point to get their view.

The 64 bit fixups are:
- logging format specifiers.
- to use an IDR for the kernel id
- changing the in-kernel vc-sm handle to being a void* instead of an int, otherwise we needed yet another IDR to convert from handle to vc-sm structure. The codec driver/mmal-vchiq is the only thing that uses the handle currently and has been updated. Perhaps better would be to declare it as a ```struct dma_buf *``` (which is what it is), but I wouldn't like to guarantee that it will behave sensibly to all dma_buf operations. Thoughts? This isn't an issue as and when we get to userspace being able to make allocations/import with this driver, as that will be giving out genuine dmabuf fds.

This should get the V4L2 codec driver working on 64 bit. I have got a follow up that cleans up some of the dmabuf code ready for VPU allocations.

### DIFF
--- a/drivers/staging/vc04_services/vc-sm-cma/vc_sm.c
+++ b/drivers/staging/vc04_services/vc-sm-cma/vc_sm.c
@@ -745,7 +745,7 @@ static int bcm2835_vc_sm_cma_remove(struct platform_device *pdev)
 }
 
 /* Get an internal resource handle mapped from the external one. */
-int vc_sm_cma_int_handle(int handle)
+int vc_sm_cma_int_handle(void *handle)
 {
 	struct dma_buf *dma_buf = (struct dma_buf *)handle;
 	struct vc_sm_buffer *res;
@@ -762,7 +762,7 @@ int vc_sm_cma_int_handle(int handle)
 EXPORT_SYMBOL_GPL(vc_sm_cma_int_handle);
 
 /* Free a previously allocated shared memory handle and block. */
-int vc_sm_cma_free(int handle)
+int vc_sm_cma_free(void *handle)
 {
 	struct dma_buf *dma_buf = (struct dma_buf *)handle;
 
@@ -772,7 +772,7 @@ int vc_sm_cma_free(int handle)
 		return -EPERM;
 	}
 
-	pr_debug("%s: handle %08x/dmabuf %p\n", __func__, handle, dma_buf);
+	pr_debug("%s: handle %p/dmabuf %p\n", __func__, handle, dma_buf);
 
 	dma_buf_put(dma_buf);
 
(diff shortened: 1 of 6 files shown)


6by9:
Rebased to 4.19.19

popcornmix:
@6by9 I assume you are happy for this to be merged now? (rather than waiting for upstream comments).
No objections here. @pelwell okay with you?

drivers/staging/vc04_services/vc-sm-cma/vc_sm.c
```diff
@@ -714,13 +711,11 @@ static int bcm2835_vc_sm_cma_probe(struct platform_device *pdev)
...
-		return err;
-	}
+	pdev->dev.dma_parms = devm_kzalloc(&pdev->dev,
+					   sizeof(*pdev->dev.dma_parms),
+					   GFP_KERNEL);
+	/* dma_set_max_seg_size checks if dma_parms is NULL. */
+	dma_set_max_seg_size(&pdev->dev, 0x3FFFFFFF);
 
 	vchiq_add_connected_callback(vc_sm_connected_init);
 	return 0;
```
pelwell, line 753: Should you return success when the dma_parms allocation could have failed?

drivers/staging/vc04_services/vc-sm-cma/vc_sm.c
```diff
@@ -586,7 +614,7 @@ vc_sm_vpu_event(struct sm_instance *instance, struct vc_sm_result_t *reply,
 	{
 		struct vc_sm_released *release = (struct vc_sm_released *)reply;
 		struct vc_sm_buffer *buffer =
-				(struct vc_sm_buffer *)release->kernel_id;
+					lookup_kernel_id(release->kernel_id);
```
pelwell, line 617: I can see this is no less safe than it was before, but shouldn't you acknowledge that the lookup could fail?

drivers/staging/vc04_services/vc-sm-cma/vc_sm.c
6by9: All that will happen in reality is that the err_printk at https://elixir.bootlin.com/linux/latest/source/kernel/dma/debug.c#L1300 will get printed out.
I've actually been seeing that error message from the vc4 graphics driver in some situations, so that needs persuing.

The handling in both cases is to shrug shoulders and continue regardless!

drivers/staging/vc04_services/vc-sm-cma/vc_sm.c
6by9: I've got further patches that add in additional checking (and nearly VPU reloc heap allocations from CMA), and cleans up some of the sequencing on releasing buffers.

There is also the comment immediately below this:
```
/*
		 * FIXME: Need to check buffer is still valid and allocated
		 * before continuing
		 */
```
so I have acknowledged it in comment if not code.

PR #2848, opened 2019-02-04
bcmrpi3_defconfig additions

6by9:
The last patch is dependent on #2845 being merged to actually work, but all the other bits are in and working.


6by9:
Rebased to 4.19.19.

Also added a patch to enable IPVLAN which was enabled for 32bit builds but not 64bit.

popcornmix:
This seems fine to me. @pelwell ?

PR #2852, opened 2019-02-08
balenaFin v1.1.0 carrier board update

majorz:
Hi @pelwell, kindly asking for review on this PR, which includes two commits targeting our balenaFin carrier board. The changes are described in more detail in the commit messages.


pelwell:
It all looks good, but I think it would be less confusing for folks if you add CONFIG_LEDS_PCA963X=m to bcmrpi_defconfig as well.

majorz:
Right, I was wondering about that as well. I will make that change now.

majorz:
I added it to bcmrpi_defconfig as well.

PR #2857, opened 2019-02-13
Rpi 4.19.y V4L2 M2M updates

6by9:
- Queries the VPU for the supported encodings rather than having a big list.
- Adds a wrapper around the ISP for a format converter. It needs a load of controls to be added, but is a first step.


6by9:
Sorry, hold fire on this just now.
I found out last night that GStreamer has a v4l2convert component that I ought to make sure the ISP device works with first. Prior to that I wasn't aware of any tools that would actually try to use this.

6by9:
OK, tested with GST 1.15.1.
One error in the driver found and PR updated (we tried setting v4l2 controls when there were none). Otherwise works fine.

dmabuf import seems to not work for either codecs or ISP, but that's a separate issue.

6by9:
Added two more commits.

1) Disables support for Bayer formats as GStreamers v4l2convert component considers them not to be raw, and therefore doesn't consider the V4L2 device to be compatible - http://gstreamer-devel.966125.n4.nabble.com/Gstreamer-and-vim2m-with-bayer-capture-formats-td4689564.html

2) Fixes a misimplementation when importing dmabufs from elsewhere.

6by9:
And now removed an errant logging line.

6by9:
Rebased, and a further two patches to enable SPS timing information in the H264 headers for encode.

popcornmix:
@pelwell any objections?

6by9:
Hold fire 10 minutes - I've just spotted a resource leak.
A MMAL component is created to check the supported formats, but never then destroyed.

6by9:
Change made to "staging: bcm2835_codec: Query supported formats from the component" at the end of ```bcm2835_codec_get_supported_fmts```.
I rebased whilst I was at it.

Just doing a sanity test build, but not expecting any failures.

pelwell:
Do we not care about resource leaks in error paths? I can accept a "no".

6by9:
> Do we not care about resource leaks in error paths? I can accept a "no".

Error paths of ```bcm2835_codec_get_supported_fmts```, or somewhere else?
You're right that I've got a couple of straight ```return```, not destroying the component. Fixing them.
Please tell me where else you're thinking of (if anywhere).

drivers/staging/vc04_services/bcm2835-codec/bcm2835-v4l2-codec.c
```diff
@@ -1906,20 +1919,48 @@ static int bcm2835_codec_buf_prepare(struct vb2_buffer *vb)
...
+			 * changed.
+			 */
+			if (buf->mmal.dma_buf) {
+				v4l2_err(&ctx->dev->v4l2_dev,
+					 "%s Buffer changed - why did the core not call cleanup?\n",
+					 __func__);
+				bcm2835_codec_mmal_buf_cleanup(&buf->mmal);
+			}
+
+			buf->mmal.dma_buf = dma_buf_get(vb->planes[0].m.fd);
```
pelwell: This second dma_buf_get looks like it leaks a reference on the file descriptor.

pelwell:
Just in bcm2835_codec_get_supported_fmts, although I've just spotted what might be another leak.

drivers/staging/vc04_services/bcm2835-codec/bcm2835-v4l2-codec.c
```diff
@@ -1785,6 +1785,9 @@ static int bcm2835_codec_create_component(struct bcm2835_codec_ctx *ctx)
 		goto destroy_component;
 
 	if (dev->role == ENCODE) {
+		u32 param = 1;
+		u32 param_size = sizeof(param);
```
pelwell: Why create a variable for this, rather than putting the sizeof at the point of use (c.f. line 1769)?

drivers/staging/vc04_services/bcm2835-codec/bcm2835-v4l2-codec.c
6by9: Ack. Too many rounds of revisions trying to work out what the heck V4L2 is trying to do for you.

L1937 should be ```buf->mmal.dma_buf = dma_buf```, which is then released by ```bcm2835_codec_buffer_cleanup``` calling ```bcm2835_codec_mmal_buf_cleanup``` to ```dma_buf_put```

drivers/staging/vc04_services/bcm2835-codec/bcm2835-v4l2-codec.c
6by9: Because my mind was thinking of ```vchiq_mmal_port_parameter_get``` where you have to pass a pointer to the size.
Ack.

pelwell:
That's it for my comments. BTW, I can't help reading your Acks as an expression of dismay.

6by9:
> That's it for my comments. 

Thank you.

> BTW, I can't help reading your Acks as an expression of dismay.

Possible slight dismay in myself for not spotting it, but otherwise not really. Only acknowledging that it needs fixing and I'm doing it.

drivers/staging/vc04_services/bcm2835-codec/bcm2835-v4l2-codec.c
```diff
@@ -2209,13 +2286,128 @@ static const struct v4l2_m2m_ops m2m_ops = {
...
+	for (i = 0, j = 0; i < num_encodings; i++) {
+		const struct bcm2835_codec_fmt *fmt = get_fmt(fourccs[i]);
+
+		if (fmt) {
+			list[j] = *fmt;
+			j++;
+		}
+	}
+	dev->supported_fmts[1].num_entries = j;
+
```
pelwell, line 2497: I think you need a ret = 0 here, otherwise there's a danger of returning MMAL_MSG_STATUS_ENOSPC.

6by9: Yup. More haste, less speed.

popcornmix:
@pelwell is this good now?

PR #2861, opened 2019-02-17
overlays: TPM SLB9670 overlay compatible updated for RPi 3

alexmurray:
List bcm2837 as compatible for the TPM SLB9670 so this works as expected
on the newer Raspberry Pi 3 models


pelwell:
Nothing checks the compatible string, so if the overlay isn't working then it is for another reason.

alexmurray:
@pelwell thanks - sorry for the noise (I had conflicting overlays both using spi when testing).

PR #2864, opened 2019-02-19
vc4-fkms fixes on 4.19

6by9:
fkms wasn't booting on 4.19 as it was trying to write to HVS registers when it didn't "own" the HVS.
Make the call to vc4_ctm_commit conditional.

Also fixes a WARN_ON where upstream have moved crtc info from the drm_plane to drm_plane_state for atomic drivers.

### DIFF
--- a/drivers/gpu/drm/vc4/vc4_firmware_kms.c
+++ b/drivers/gpu/drm/vc4/vc4_firmware_kms.c
@@ -193,8 +193,8 @@ static void vc4_cursor_plane_atomic_update(struct drm_plane *plane,
 					   struct drm_plane_state *old_state)
 {
 	struct vc4_dev *vc4 = to_vc4_dev(plane->dev);
-	struct vc4_crtc *vc4_crtc = to_vc4_crtc(plane->crtc);
 	struct drm_plane_state *state = plane->state;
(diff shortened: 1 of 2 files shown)


6by9:
Further investigations ongoing.
Booting to a console works fine, but starting X doesn't. I suspect it is something up with using the cursor place, but that is a pure guess at the moment.

6by9:
The issue starting X isn't related to these changes at all.
For some reason both fake and full KMS drivers are failing to start X with the logging reporting
```
Fatal server error:
[    42.571] (EE) Caught signal 6 (Aborted). Server aborting
```
Both boot fine to a console. This is on Buster, so I'll drop back to Stretch and see if that is related. The legacy driver works fine running X.

6by9:
Stretch is happy with both fake kms and full kms, so I'm happy with the patches.
@popcornmix @pelwell care for a review?

I wonder if we have a Mesa or similar breakage with a very recent Buster update - I'm sure full KMS started X about a week ago because it was the only way I was successfully running X on 64bit 4.19.

popcornmix:
Looks perfectly sane to me. Let me try a build.

pelwell:
I was on the verge of merging.

popcornmix:
Thanks - fkms booted to desktop and could run neverball.

PR #2873, opened 2019-02-27
Fix for Pisound kernel module in Real Time kernel configuration.

gtrainavicius:
When handler of data_available interrupt is fired, queue_work ends up getting called and it can block on a spin lock which is not allowed in interrupt context. The fix was to run the handler from a thread context instead.

Exactly the same fix should be applied to rpi-4.14.y-rt.


gtrainavicius:
Thank you! Here's the pull request for RT branch: https://github.com/raspberrypi/linux/pull/2874

PR #2874, opened 2019-02-27
Fix for Pisound kernel module in Real Time kernel configuration.

gtrainavicius:
When handler of data_available interrupt is fired, queue_work ends up getting called and it can block on a spin lock which is not allowed in interrupt context. The fix was to run the handler from a thread context instead.

Same changes as in https://github.com/raspberrypi/linux/pull/2873, this one is targeted at rpi-4.14.y-rt branch.

### DIFF
--- a/sound/soc/bcm/pisound.c
+++ b/sound/soc/bcm/pisound.c
@@ -1,6 +1,6 @@
 /*
  * Pisound Linux kernel module.
- * Copyright (C) 2016-2017  Vilniaus Blokas UAB, https://blokas.io/pisound
+ * Copyright (C) 2016-2019  Vilniaus Blokas UAB, https://blokas.io/pisound
  *
  * This program is free software; you can redistribute it and/or
  * modify it under the terms of the GNU General Public License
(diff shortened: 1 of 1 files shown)


gtrainavicius:
Can this be merged to the rpi-4.14.y-rt branch? Or is the procedure different to get code into that branch? :)

pelwell:
The -rt branch is maintained by @TiejunChina with merges from the mainline rpi branches, and your commit would normally have been pulled in automatically, but these updates naturally get batched together with a variable delay.

I can merge this now, then Tiejun can revert it (or forcibly remove it) before the next merge.

gtrainavicius:
Ok, thank you! We had already merged the fix into rpi-4.19.y, should I then make another pull request for rpi-4.14.y, so the fix won't disappear from rpi-4.14.y-rt branch during updates? Or will the -rt branch move forward to 4.19?

Anyway, just let me know what'd be the most convenient for you. :)

pelwell:
If the patch is never in 4.14 then there's no reason to revert this rt-specific patch. I would expect Tiejun to create rpi-4.19.y-rt the next time he gets to spend some time on the kernel.

Issue #2885, opened 2019-03-07
4.19.y: bcm2835_codec/v4l2 module unload oops


jjsuwa:
When vc_sm allocation is met on boot, unloading both `bcm2835_codec` and `bcm2835_v4l2` (in no particular order) will occur kernel oops in latter module.

```
[    1.633192] [vc_sm_connected_init]: start
[    1.640403] [vc_sm_connected_init]: end - returning 0
[    5.524657] bcm2835_vc_sm_cma_probe: Videocore shared memory driver
[    5.524670] [vc_sm_connected_init]: start
[    5.531295] [vc_sm_connected_init]: installed successfully
.....
$ modprobe -r bcm2835_v4l2
$ modprobe -r bcm2835_codec
.....
[ 1045.482001] bcm2835-codec bcm2835-codec: Removing bcm2835-codec
[ 1045.482419] bcm2835-codec bcm2835-codec: Removing bcm2835-codec
[ 1045.482714] bcm2835-codec bcm2835-codec: Removing bcm2835-codec
[ 1045.507402] Unable to handle kernel NULL pointer dereference at virtual address 00000004
[ 1045.517725] pgd = 49c43cec
[ 1045.521742] [00000004] *pgd=36545835, *pte=00000000, *ppte=00000000
[ 1045.529224] Internal error: Oops: 817 [#1] PREEMPT SMP ARM
[ 1045.535831] Modules linked in: arc4 ecb md4 md5 sha512_generic cmac hmac nls_utf8 cifs ccm brcmfmac brcmutil sha256_generic spidev cfg80211 v4l2_mem2mem evdev rfkill videobuf2_v4l2 videobuf2_common videodev snd_soc_pcm5102a snd_soc_rpi_simple_soundcard snd_soc_bcm2835_i2s media vc_sm_cma(C-) snd_soc_core gpio_keys spi_bcm2835 raspberrypi_hwmon snd_compress hwmon snd_pcm_dmaengine i2c_bcm2835 snd_pcm snd_timer snd uio_pdrv_genirq uio fixed i2c_dev ip_tables x_tables ipv6 [last unloaded: bcm2835_mmal_vchiq]
[ 1045.589342] CPU: 2 PID: 8530 Comm: modprobe Tainted: G         C        4.19.27-v7+ #1
[ 1045.599972] Hardware name: BCM2835
[ 1045.604738] PC is at misc_deregister+0x54/0xb0
[ 1045.610556] LR is at 0x100
[...]

PR #2888, opened 2019-03-08
staging: vc_sm_cma: Remove erroneous misc_deregister

6by9:
Code from the misc /dev node was still present in
bcm2835_vc_sm_cma_remove, which caused a NULL deref.
Remove it.

See #2885.


6by9:
You're too quick for me - there are other cleanups that ought to be done. Oh well, a new PR it is.

pelwell:
The idea is that you create the PR when you are ready, or label it WIP, RFC etc.

6by9:
I know, I'd fixed the one bug and created the PR, then noticed the next thing wrong and was thinking I'd tuck it on the back. You got there before I did :-)

pelwell:
I'll resume my usual slow service.

PR #2889, opened 2019-03-08
vc-sm-cma tidy ups, and one for bcm2835_codec

6by9:
Error handling fixes.

### DIFF
--- a/drivers/staging/vc04_services/vc-sm-cma/vc_sm.c
+++ b/drivers/staging/vc04_services/vc-sm-cma/vc_sm.c
@@ -757,7 +757,7 @@ static int bcm2835_vc_sm_cma_remove(struct platform_device *pdev)
 	pr_debug("[%s]: start\n", __func__);
 	if (sm_inited) {
 		/* Remove all proc entries. */
-		//debugfs_remove_recursive(sm_state->dir_root);
+		debugfs_remove_recursive(sm_state->dir_root);
 
 		/* Stop the videocore shared memory service. */
 		vc_sm_cma_vchi_stop(&sm_state->sm_handle);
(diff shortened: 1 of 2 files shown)


6by9:
I can't see any more this time, and I think I'll stop looking  :-)

6by9:
@pelwell I'm good with this one now.

drivers/staging/vc04_services/vc-sm-cma/vc_sm.c
```diff
@@ -757,16 +748,17 @@ static int bcm2835_vc_sm_cma_remove(struct platform_device *pdev)
...
 	if (sm_inited) {
 		/* Remove all proc entries. */
-		//debugfs_remove_recursive(sm_state->dir_root);
+		debugfs_remove_recursive(sm_state->dir_root);
 
 		/* Stop the videocore shared memory service. */
 		vc_sm_cma_vchi_stop(&sm_state->sm_handle);
+	}
 
+	if (sm_state) {
```
JamesH65, line 757: Isn't this guaranteed to be non-null?

JamesH65: Hold on, might have misread the diff.

6by9: No, I think you're right, but I'm now rereading it again.

6by9: You're right.
Should the kzalloc in probe fail then we'll bail out and remove will never be called. There should therefore never be a path where remove gets called with sm_state non-NULL.

6by9:
Force pushed an update to address James' comment.

6by9:
Rebased to 4.19.29

PR #2893, opened 2019-03-13
arch/arm/Makefile: Added dtbo rule

mickey-happygolucky:
Device Tree related rules have been moved to the Top of tree Makefile from 4.19 to 5.0.
At that time, dtbo's rule seems to be removed, but I think it is necessary at least in arch/arm, such as when using with yocto.

### DIFF
--- a/arch/arm/Makefile
+++ b/arch/arm/Makefile
@@ -343,6 +343,9 @@ $(BOOT_TARGETS): vmlinux
 $(INSTALL_TARGETS):
 	$(Q)$(MAKE) $(build)=$(boot) MACHINE=$(MACHINE) $@
 
+%.dtbo: | scripts
+	$(Q)$(MAKE) $(build)=$(boot)/dts MACHINE=$(MACHINE) $(boot)/dts/$@
+
 PHONY += vdso_install
 vdso_install:
 ifeq ($(CONFIG_VDSO),y)


pelwell:
The required .dtbo rule is in scripts/Makefile.lib. This PR is unnecessary.

mickey-happygolucky:
Thanks for reply.

However, when this kernel was actually built with yocto, the following error occurred.

```txt
| make[2]: *** No rule to make target 'overlays/at86rf233.dtbo'.  Stop.
```

Same results were obtained for other dts under overlays.

I know there are dtbo rules in scripts/Makefile.lib,
But it seems to works not fine.

If you have another approach, please let us know what you think.

pelwell:
If you follow the standard Raspberry Pi building instructions (https://www.raspberrypi.org/documentation/linux/kernel/building.md) with the current rpi-5.0.y tree and it doesn't build then report back here, but it works for me.

mickey-happygolucky:
Okay. I will investigate.
My goal is to build this kernel with the yoctoproject build system.(but it is unofficial)

We know in 4.19 that we have confirmed that we can build without problems, so we think that there is something that causes this failure at this change point.

Thanks.

mickey-happygolucky:
I have confirmed that there are no problems with the standard Raspberry Pi building instructions.
We know that the build with yocto will succeed in the workaround, so we will withdraw this PR.

PR #2895, opened 2019-03-18
vcsm: rpi-5.0.y - gcc-v8 fixes

kbingham:
Three fix up commits for when building the 5.0.y tree with a GCC v8 compiler
arm-linux-gnueabihf-gcc (Ubuntu 8.2.0-7ubuntu1) 8.2.0

I suspect the warnings were enabled by default in recent compilers, and the VC_SM module enables -Werror which fails the build.

These issues are on the 4.19.y branch too - so could likely be cherry-picked there as well.


6by9:
Looks fine to me.

pelwell:
Thanks!

Issue #2894, opened 2019-03-14
F2FS-fs (mmcblk0p3): access invalid blkaddr:259


f00b4r0:
Writting a special inode to the F2FS filesystem (via e.g. mknod or copying a device node from /dev) will corrupt the filesystem.

If I read the commit message correctly this bug should already be fixed by dda9f4b9cac6bdd2a96253b4444d7a6ce5132edb, however I am running:
```
$ uname -a
Linux rPi2 4.14.98-v7+ #1200 SMP Tue Feb 12 20:27:48 GMT 2019 armv7l GNU/Linux
```

And I've been bitten by this issue. I am unable to remove any of the created special files which appear in `ls` as:

```
 ls -l chroot/dev/
ls: cannot access 'chroot/dev/null': Bad address
ls: cannot access 'chroot/dev/random': Bad address
total 0
c????????? ? ? ? ?            ? null
c????????? ? ? ? ?            ? random
```

`dmesg` output is then filled with:
```
[321687.954285] F2FS-fs (mmcblk0p3): access invalid blkaddr:264
[321687.954296] ------------[ cut here ]------------
[321687.954316] WARNING: CPU: 3 PID: 31161 at fs/f2fs/checkpoint.c:150 f2fs_is_valid_blkaddr+0x19c/0x1c0
[321687.954322] Modules linked in: ip_set_hash_ip xt_set ip_set nfnetlink xt_CT iptable_raw iptable_nat nf_nat_ipv4 nf_nat xt_limit nf_conntrack_ipv4 nf_defrag_ipv4 xt_tcpudp xt_conntrack nf_conntrack iptable_filter uio_pdrv_genirq uio fixed i2c_dev ip_tables x_tables
[321687.954437] CPU: 3 PID: 31161 Comm: ls Tainted: G        W       4.14.98-v7+ #1200
[321687.954444] Hardware name: BCM2835
[321687.954469] [<8010ff30>] (unwind_backtrace) from [<8010c174>] (show_stack+0x20/0x24)
[321687.954489] [<8010c174>] (show_stack) from [<8078cf04>] (dump_stack+0xd4/0x118)
[321687.954507] [<8078cf04>] (dump_stack) from [<8011dd70>] (__warn+0xf8/0x110)
[321687.954523] [<8011dd70>] (__warn) from [<8011de58>] (warn_slowpath_null+0x30/0x38)
[...]

PR #2896, opened 2019-03-18
f2fs: fix to skip verifying block address for non-regular inode

pelwell:
commit dda9f4b9cac6bdd2a96253b4444d7a6ce5132edb upstream.

generic/184 1s ... [failed, exit status 1]- output mismatch
    --- tests/generic/184.out	2015-01-11 16:52:27.643681072 +0800
     QA output created by 184 - silence is golden
    +rm: cannot remove '/mnt/f2fs/null': Bad address
    +mknod: '/mnt/f2fs/null': Bad address
    +chmod: cannot access '/mnt/f2fs/null': Bad address
    +./tests/generic/184: line 36: /mnt/f2fs/null: Bad address
    ...

F2FS-fs (zram0): access invalid blkaddr:259
EIP: f2fs_is_valid_blkaddr+0x14b/0x1b0 [f2fs]
 f2fs_iget+0x927/0x1010 [f2fs]
 f2fs_lookup+0x26e/0x630 [f2fs]
 __lookup_slow+0xb3/0x140
 lookup_slow+0x31/0x50
 walk_component+0x185/0x1f0
 path_lookupat+0x51/0x190
 filename_lookup+0x7f/0x140
 user_path_at_empty+0x36/0x40
 vfs_statx+0x61/0xc0
 __do_sys_stat64+0x29/0x40
 sys_stat64+0x13/0x20
 do_fast_syscall_32+0xaa/0x22c
 entry_SYSENTER_32+0x53/0x86

In f2fs_iget(), we will check inode's first block address, if it is valid,
we will set FI_FIRST_BLOCK_WRITTEN flag in inode.

But we should only do this for regular inode, otherwise, like special
inode, i_addr[0] is used for storing device info instead of block address,
it will fail checking flow obviously.

So for non-regular inode, let's skip verifying address and setting flag.

Signed-off-by: Chao Yu <yuchao0@huawei.com>
Signed-off-by: Jaegeuk Kim <jaegeuk@kernel.org>

### DIFF
--- a/fs/f2fs/inode.c
+++ b/fs/f2fs/inode.c
@@ -310,13 +310,15 @@ static int do_read_inode(struct inode *inode)
 	/* get rdev by using inline_info */
 	__get_inode_rdev(inode, ri);
 
-	err = __written_first_block(sbi, ri);
-	if (err < 0) {
-		f2fs_put_page(node_page, 1);
-		return err;
+	if (S_ISREG(inode->i_mode)) {
+		err = __written_first_block(sbi, ri);
+		if (err < 0) {
+			f2fs_put_page(node_page, 1);
+			return err;
+		}
+		if (!err)
+			set_inode_flag(inode, FI_FIRST_BLOCK_WRITTEN);
 	}
-	if (!err)
-		set_inode_flag(inode, FI_FIRST_BLOCK_WRITTEN);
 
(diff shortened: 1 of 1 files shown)


pelwell:
See #2894.

PR #2900, opened 2019-03-20
bcm2835_codec: Add role to video device name

kbingham:
Hi  @6by9 ,

The first patch fixes up the roles[] array declaration, and then the second patch is more of a proposal. I've added the role to the video node name to distinguish them, producing:

video10: bcm2835-codec-decode
video11: bcm2835-codec-encode
video12: bcm2835-codec-isp

Perhaps as ISP is not directly a 'codec' the codec reference could be removed, but i thought that was more of a point for discussion so I've simply appended the role.

### DIFF
--- a/drivers/staging/vc04_services/bcm2835-codec/bcm2835-v4l2-codec.c
+++ b/drivers/staging/vc04_services/bcm2835-codec/bcm2835-v4l2-codec.c
@@ -77,7 +77,7 @@ enum bcm2835_codec_role {
 	ISP,
 };
 
-const static char *roles[] = {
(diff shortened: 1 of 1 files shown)


6by9:
Thanks @kbingham. Both look fine to me.

As per email thread, is this the only string you are searching on, or is there gain in updating what QUERYCAP returns too?

kbingham:
QUERYCAP should be certainly be updated too I think. I'll respin the patches when I get that far :)

kbingham:
Closing this PR with changes required

PR #2902, opened 2019-03-21
RFC: sound: Fixes for audioinjector-octo under 4.19

pelwell:
1. Move the DT alias declaration to the I2C shim in the cases
where the shim is enabled. This works around a problem caused by a
4.19 commit [1] that generates DT/OF uevents for I2C drivers.

2. Fix the diagnostics in an error path of the soundcard driver to
correctly identify the reason for the failure to load.

3. Move the declaration of the clock node in the overlay outside
the I2C node to avoid warnings.

4. Sort the overlay nodes so that dependencies are only to earlier
fragments, in an attempt to get runtime dtoverlay application to
work (it still doesn't...)

See: https://github.com/Audio-Injector/Octo/issues/14

[1] af503716ac14 ("i2c: core: report OF style module alias for devices registered via OF")

### DIFF
--- a/arch/arm/boot/dts/overlays/audioinjector-addons-overlay.dts
+++ b/arch/arm/boot/dts/overlays/audioinjector-addons-overlay.dts
@@ -13,6 +13,17 @@
 	};
 
 	fragment@1 {
+		target-path = "/";
+		__overlay__ {
+			cs42448_mclk: codec-mclk {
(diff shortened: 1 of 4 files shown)


pelwell:
Do you have any feedback for this patch?

pelwell:
Rebased.

PR #2915, opened 2019-03-30
Enable Dynamic Clock Calculation on TLV320AIC3204

mcdermj:
This code has been accepted upstream for Linux 5.2, but we would like to see if we can get it back ported to the Pi Foundation so that it starts being included in the default distribution.  This is to support the NW Digital Radio DRAWS and UDRC HATs (see http://nwdigitalradio.com).  All changes have been tested to compile and function on the Raspberry Pi with the DRAWS board.  Future PRs will fix a couple of more bugs (after they've been submitted upstream) and provide the DT overlay files for the NWDR HATs.


pelwell:
Applied offline - we like "commit hash upstream." comments to be included, which is easier when automated.

PR #2919, opened 2019-04-03
[Update] Update the ili210x support for ilitek ili251x touch device

samhsuism:
The upstream's ili210x touchscreen driver is updated since v5.1-rc1, this latest version of ili210x driver has been compiled and tested with the new changes of the ili210.c source and the relative necessary edits such as the device tree, the config files ( in bcmrpi_defconfig, bcm2709_defconfig ) and the Makefile. It works as functional and we would like these new patches could be updated to the next released Raspberry Pi's image for the supports of the Ilitek touch devices ili210x and ili251x.


pelwell:
If some of these commits are (or contain) upstream changes as you suggest, please include those commits without modification (except for adding `commit ... upstream.` as the first line of the comment, where `...` is replaced with the full commit hash), followed by a blank line.

PR #2920, opened 2019-04-05
Add Support for the NW Digital Radio DRAWS/UDRC

mcdermj:
This series adds support for the Northwest Digital Radio UDRC and DRAWS HATs.  The tlv320aic32x4 driver changes have already been committed upstream and are marked as such.  The dtoverlay pieces are particular to this project and are included here.  The board is capable of using the ASoC "simplecard" driver and the DTs reflect that.

### DIFF
--- a/arch/arm/boot/dts/overlays/Makefile
+++ b/arch/arm/boot/dts/overlays/Makefile
@@ -29,6 +29,7 @@ dtbo-$(CONFIG_ARCH_BCM2835) += \
 	dionaudio-loco-v2.dtbo \
 	dpi18.dtbo \
 	dpi24.dtbo \
+	draws.dtbo \
 	dwc-otg.dtbo \
 	dwc2.dtbo \
 	enc28j60.dtbo \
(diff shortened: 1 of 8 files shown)


pelwell:
The `_gain` and `_datarate` parameters have been declared as targeting strings when they're actually targeting 32-bit integers; you need to but `:0` (4-byte integer at byte offset 0) after the property name in the parameter declarations:
```
        draws_adc_ch4_gain = <&adc_ch4>,"ti,gain:0";
```
I'd also like an entry in the README in the overlays directory, with a brief description of the overlay and its parameters - beware the strict formatting rules: no tabs, fixed indents and two blank lines between overlays.

Otherwise, nice job.

mcdermj:
Thanks so much for your patience and commentary, Phil.  I have fixed your concerns and pushed  a couple of new commits that should solve the problems.

pelwell:
A single TAB crept into the README (editors can be sneaky), and the extra commits were really fixups to the overlay commit, so they've been squashed and merged offline.

PR #2921, opened 2019-04-05
[Update] pull upstream ili210x to support ilitek ili251x touch devices

samhsuism:
The upstream's ili210x touchscreen driver is updated since v5.1-rc1. With this latest version of ili210x driver has been compiled and tested with the new changes of the ili210.c source and the relative necessary edits such as the device tree, the config files ( in bcmrpi_defconfig, bcm2709_defconfig ) and the Makefile. It works as functional and we would like these new patches could be updated to the next released Raspberry Pi's image for the supports of the Ilitek touch devices ili210x and ili251x.

commit f67cc3e927d8414ad3872e046764534ea1f5db0d upstream.


pelwell:
That's looking better, but there are a few things to improve:
1. Is the dt-bindings commit really not present upstream? I'm surprised they'd accept the driver change without it. If it is upstream, please cherry-pick the commit and put the upstream commits first. Also, it's nitpicky, but the "commit ... upstream." (note the period) would ideally be the first line of the comment.

2. The README describes (well, lists - some more detail would be nice) the sizex and sizey parameters, but they are missing from the overlay. It also puts the entry in the wrong position alphabetically - it should go before iqaudio-dac.

3. You can squash all the downstream DT commits (the overlay, the README and the Makefile) into one commit.

4. Likewise, the defconfig changes can be in a single commit - please include arch/arm64/configs/bcmrpi3_defconfig as well.

5. Finally, please  include a "Signed-off-by:" line in your downstream commits.

You'll find "git rebase -i" very useful for re-ordering and squashing commits. And when you have made the changes, use "git push -f ..." to update this PR - you don't need to close it and start again.

PR #2922, opened 2019-04-05
Audiophonics I-Sabre 9038Q2M DAC driver

audiophonics:
This patch add support for the Audiophonics 9038 Q2M Dac board.

Thanks to Klaus Schulz aka soundcheck for the kernel patch and integration support.


pelwell:
Merged offline with some minor fix-ups.

Issue #2907, opened 2019-03-23
USB Broken for CM3 post 4.14.18 through 4.19.y


grimepoch:
There is some major problem with the USB driver starting with the fix that went into 4.14.20.  I have verified that the problem comes from the changes of these two files:

drivers/usb/host/dwc_otg/dwc_otg_hcd.c
drivers/usb/host/dwc_otg/dwc_otg_hcd_linux.c

I replaced the files in 4.14.20 with those 2 files in 4.14.18 and the USB began working properly.

Specifically the problem is on hot swapping.  With the changes introduced in 4.14.20, if I removed a USB device while the system was running, there was a good chance of the whole system freezing.  (Nothing was written out to logs either).

Using a hub, the problem would happen less often, but still quite a bit.

The biggest problem I had, if I removed a WiFi USB adapter, it would NEVER recover from that.  With this roll back to the older code, if I reattach the WiFi adapter, it will reconnect and everything.

I am not sure the implications of what changed in the driver, it's possible it fixed a different problem and introduced this new one, but after a day of testing, the USB works much better with the older source on those two files.

4.19 has the same problem.  I did not test using the older code with 4.19 as we are not ready to move up to that yet, but just wanted to identify the problem still exists.

Hardware: CM3

6by9:
https://www.raspberrypi.org/forums/viewtopic.php?f=98&t=230161

Specfically what Wifi adapter/devices are you using? ```lsusb -vvv```.

grimepoch:
What I have observed is it isn't only the WiFi adapter, if I have a MIDI device attached and connected, and I have it open, if I remove the device, with 4.14.20, it will lock the whole system as well (completely unresponsive pi).

If I use a small hub like what is listed below, the WiFi remove will almost always lock it.  The MIDI becomes more sporadic, but it still happens.

The way I was testing the MIDI removal, any MIDI device attached, then doing
`while true; do cat /dev/snd/midiC0D0; done`

In the console.  When you remove the MIDI device, it just loops saying device not found, when you insert one, it will stop and echo any messages you send with the MIDI device.

On 4.14.18, I can add and remove MIDI devices, never locks up the system.

On 4.14.20, it will lock up.


`
Bus 001 Device 003: ID 148f:5370 Ralink Technology, Corp. RT5370 Wireless Adapter
Device Descriptor:
  bLength                18
  bDescriptorType         1
  bcdUSB               2.00
  bDeviceClass            0 (Defined at Interface level)
[...]

PR #2927, opened 2019-04-09
dwc_otg: fix locking around dequeueing and killing URBs

P33M:
kill_urbs_in_qh_list() is practically only ever called with the fiq lock
already held, so don't spinlock twice in the case where we need to cancel
an isochronous transfer.

Also fix up a case where the global interrupt register could be read with
the fiq lock not held.

Fixes the deadlock seen in https://github.com/raspberrypi/linux/issues/2907


popcornmix:
Fine by me.

Issue #2829, opened 2019-01-26
4.19: ext4_inode_cache slab memory usage, doesn't get reclaimed


Fiyorin:
i tried out the next branch of https://github.com/raspberrypi/firmware/ with "sudo BRANCH=next rpi-update". this gave me kernel version 4.19.16. i had no issues with it until i noticed that the memory usage was going up and up. 
after executing "sudo slabtop", ext4_inode_cache was using ~170 MB. normally this would get reclaimed when the rest of the memory is getting full, however this was not the case here. even when the memory was fully used up by userspace programs, and the kernel was agressively swapping, the ext4_inode_cache would not go down. increasing "vm.vfs_cache_pressure" and "echo 3 > /proc/sys/vm/drop_caches" did not help. i've since went back to the older 4.14 kernel and the issue doesn't appear there.

* Which model of Raspberry Pi? e.g. Pi3B+, PiZeroW
Pi3B+
* Which OS and version (`cat /etc/rpi-issue`)?
Raspbian Stretch with desktop

**Additional context**
I'm not using a microSD card: the system is on a USB attached SSD drive. Let me know if you need more info to reproduce this issue.

DrJosh9000:
No idea if it's the same bug, but this seems similar to an issue I have been wrangling recently with the 4.19.16 and 4.19.20 Linux kernels on some non-Raspberry Pis. 

In my case it turned out that the kernel was leaking memory associated with cgroups - and over time it grew to a large size because we had some faulty systemd services restarting over and over (creating and removing a cgroup each time, but the kernel leaked). 
Here's an LKML thread discussing backporting the patches: https://lkml.org/lkml/2018/11/2/160 
And a bug in Ubuntu: https://bugs.launchpad.net/ubuntu/+source/linux/+bug/1792349

Pmant:
I think I'm experiencing the same issue on my Pi 3.

```
root@raspberrypi:~# slabtop -s c
 Active / Total Objects (% used)    : 1718215 / 1738017 (98.9%)
 Active / Total Slabs (% used)      : 66822 / 66822 (100.0%)
 Active / Total Caches (% used)     : 88 / 102 (86.3%)
 Active / Total Size (% used)       : 599262.30K / 603247.54K (99.3%)
 Minimum / Average / Maximum Object : 0.02K / 0.35K / 8.00K

  OBJS ACTIVE  USE OBJ SIZE  SLABS OBJ/SLAB CACHE SIZE NAME
[...]

PR #2941, opened 2019-04-24
cgroup mayhem

P33M:
Between 4.14 and 4.19, it became unsafe to have memory cgroups enabled in the kernel config but "disabled" at runtime. Revert the patch that disables memcg by default, then revert the config nobble.

There will be a small static RAM usage penalty until the runtime disable/enable is fixed.

See https://github.com/raspberrypi/linux/issues/2829


pelwell (approved):
We ought to be able to construct a patch that gets us runtime cgroup memory support without the memory overhead for those who don't want it, but until then the fixed memory overhead is more attractive than the leak.

PR #2944, opened 2019-04-26
brcmfmac: added LED triggers for transmit/receive

RussellJoyce:
Add three basic LED triggers to brcmfmac, based on those in mac80211: one for transmit, one for receive, and one for combined transmit/receive.

This allows the Raspberry Pi Wi-Fi activity to be easily displayed on one or more LEDs using standard Linux LED triggers (`phy0rx`, `phy0tx` and `phy0rxtx`).

An example device tree overlay using this can be found at https://github.com/yorkrobotlab/pi-puck/blob/prototype/overlays/pi-puck-v4_0.dts#L24.

Note that this was previously submitted upstream and rejected, mostly due to the maintainers preferring a more generic solution to the lack of LED triggers in a number of wireless drivers (see https://patchwork.kernel.org/patch/9830565/). However, the specific patch here is useful in the Raspberry Pi kernel as it targets the Broadcom driver used by the built-in Wi-Fi, and (at least in my experience) replaces functionality that was lost when switching from using third-party USB Wi-Fi dongles with older Raspberry Pis that used the SoftMAC driver (which _does_ include LED triggers).

### DIFF
--- a/drivers/net/wireless/broadcom/brcm80211/Kconfig
+++ b/drivers/net/wireless/broadcom/brcm80211/Kconfig
@@ -67,6 +67,18 @@ config BRCMFMAC_PCIE
 	  IEEE802.11ac embedded FullMAC WLAN driver. Say Y if you want to
 	  use the driver for an PCIE wireless card.
 
+config BRCMFMAC_LEDS
+	bool "LED trigger support for FullMAC driver"
+	depends on BRCMFMAC
+	depends on LEDS_CLASS
+	select LEDS_TRIGGERS
+	default y
+	---help---
+	  This option enables LED trigger support for Broadcom FullMAC WLAN
+	  driver. Say Y to create LED triggers for receive events (phyXrx),
+	  transmit events (phyXtx), and both events combined (phyXrxtx), on
+	  each adapter (where 'phyX' is the phy name).
+
 config BRCM_TRACING
 	bool "Broadcom device tracing"
 	depends on BRCMSMAC || BRCMFMAC
--- a/drivers/net/wireless/broadcom/brcm80211/brcmfmac/Makefile
+++ b/drivers/net/wireless/broadcom/brcm80211/brcmfmac/Makefile
@@ -48,6 +48,8 @@ brcmfmac-$(CONFIG_BRCMFMAC_USB) += \
 		usb.o
 brcmfmac-$(CONFIG_BRCMFMAC_PCIE) += \
 		pcie.o
+brcmfmac-$(CONFIG_BRCMFMAC_LEDS) += \
+		led.o
 brcmfmac-$(CONFIG_BRCMDBG) += \
 		debug.o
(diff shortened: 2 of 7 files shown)


ghost:
Reading the thread you posted, they haven't actually rejected the patch, just said that it would be better done in cfg80211. If you're wanting upstream approval, you will probably want to put it into cfg80211. That would be the best implementation, since it would allow other wifi drivers to potentially benefit from your work in the future. Is there a reason why you would prefer not the split your implementation between cfg80211 and brcmfmac? Your patch is obviously of potential use to some people. Is this something you need on your Pi?

lategoodbye:
The patch "leds: trigger: Introduce a NETDEV trigger" already provides a general LED trigger for network activity.

What is the benefit of this pull request?

RussellJoyce:
@andrum99 That's true. To be honest, I don't care too much about upstream approval, it would just be nice not to have to keep compiling my own Raspberry Pi kernel to get this feature (especially as it's being released in a product soon). I could take a look at splitting it between those files though, now that I have a bit more time than when I originally submitted that patch a couple of years ago.

However...

@lategoodbye I hadn't seen that patch, thanks for pointing it out. It's not _quite_ the same as it can't be fully set-up from the device tree, as it needs the `device_name`, `rx` and `tx` values to be specified using sysfs attributes (unless I've missed something?). I guess these could be set using a script running at boot though, so it's not the end of the world. Also it's only in v4.16 or newer of the kernel, so unfortunately not currently available on Raspberry Pis by default (although obviously my code isn't either).

Therefore, maybe the best thing to do would be to add device tree or module parameter support into ledtrig-netdev so it can be set-up properly on boot (if this is even possible, I can look into this) and persuade someone to include it in the Raspberry Pi defconfigs once the default release gets past v4.16.

ghost:
4.19 is coming pretty soon - development of the Pi kernel has switched to the 4.19 tree. The next official Pi kernel will be 4.19-based. The generic netdev LED patch is already in the Raspberry Pi 4.19 branch so you just need to 'rpi-update' to get it. The usual warnings about using rpi-update apply - it's a beta quality kernel at best and may break stuff.

RussellJoyce:
Thanks for the information.

If I build my own kernel from the rpi-4.19.y branch here using `bcmrpi_defconfig` then `CONFIG_LEDS_TRIGGER_NETDEV` is not set. Should be be?

ghost:
Ah, good point. The code is there, but it's obviously disabled by the config. My mistake. I suggest raising a new issue to get this enabled by default. If you can provide justification then the Raspberry Pi folk are usually OK with such requests. You will probably be asked to check if the kernel gets any bigger, and what the impact on memory is. If you do that then I can't foresee any problems getting this kernel config change integrated into the default Pi kernel.

RussellJoyce:
Great, I'll give that a go 👍

Thank you both.

PR #2950, opened 2019-05-01
smsc95xx: dynamically fix up TX buffer alignment with padding bytes

P33M:
dwc_otg requires a 32-bit aligned buffer start address, otherwise
expensive bounce buffers are used. The LAN951x hardware can skip up to
3 bytes between the TX header and the start of frame data, which can
be used to force alignment of the URB passed to dwc_otg.

As found in https://github.com/raspberrypi/linux/issues/2924

### DIFF
--- a/drivers/net/usb/smsc95xx.c
+++ b/drivers/net/usb/smsc95xx.c
@@ -2082,7 +2082,9 @@ static struct sk_buff *smsc95xx_tx_fixup(struct usbnet *dev,
 					 struct sk_buff *skb, gfp_t flags)
 {
 	bool csum = skb->ip_summed == CHECKSUM_PARTIAL;
-	int overhead = csum ? SMSC95XX_TX_OVERHEAD_CSUM : SMSC95XX_TX_OVERHEAD;
+	unsigned int align_bytes = -((uintptr_t)skb->data) & 0x3;
+	int overhead = csum ? SMSC95XX_TX_OVERHEAD_CSUM + align_bytes
+				: SMSC95XX_TX_OVERHEAD + align_bytes;
 	u32 tx_cmd_a, tx_cmd_b;
 
 	/* We do not advertise SG, so skbs should be already linearized */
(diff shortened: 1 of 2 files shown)


P33M:
From dwc_otg stats, previously *every* tx packet was misaligned. Adding the pad bytes drops the misalign count to 0.

@popcornmix I'd prefer two separate rpi-update revisions - one with the coherent pool increase, and one with this commit added so we can spot regressions.

drivers/net/usb/smsc95xx.c
```diff
@@ -2082,7 +2082,9 @@ static struct sk_buff *smsc95xx_tx_fixup(struct usbnet *dev,
 					 struct sk_buff *skb, gfp_t flags)
 {
 	bool csum = skb->ip_summed == CHECKSUM_PARTIAL;
-	int overhead = csum ? SMSC95XX_TX_OVERHEAD_CSUM : SMSC95XX_TX_OVERHEAD;
+	unsigned int align_bytes = (4 - ((uintptr_t)skb->data & 0x3)) & 0x3;
```
pelwell: The first `& 0x3` is unnecessary.

drivers/net/usb/smsc95xx.h
```diff
@@ -21,7 +21,7 @@
 #define _SMSC95XX_H
 
 /* Tx command words */
-#define TX_CMD_A_DATA_OFFSET_	(0x001F0000)	/* Data Start Offset */
+#define TX_CMD_A_DATA_OFFSET_	(0x00030000)	/* Data Start Offset */
```
pelwell, line 24: What is the reason behind this change, and does it have any effect?

drivers/net/usb/smsc95xx.c
popcornmix: Pretty sure the 4 is unnecessary too. ``(-x)&3`` should be enough...

pelwell: Indeed.

drivers/net/usb/smsc95xx.h
P33M: It has no effect on code output, but the bitfield specifier is wrong. Edited to prevent (reduce?) confusion.

drivers/net/usb/smsc95xx.c
P33M: GCC agrees, disassembly is the same

pelwell: That comment needs an emoji to disambiguate the snark level.

P33M: Or a push

paul-1:
We use Cloudflare CDN for our site, and since this commit, all of the rpi boards with the smsc95xx ethernet are experiencing stalled downloads when receiving larger files (using wget).  The stalls typically occur after receiving less than 1MB of data.   But every attempt does stall.

I've only experience stalls coming from CloudFlare servers, but  RPI3B+ boards ethernet and RPI Wifi 
 work fine.

We also use a custom kernel/OS, but I validated the testing with a fresh Raspbian Lite image, with only the kernel updated as below.

RPI 4.19.40 (6cb1da772a8c56f44591bec904046b15f1126949) < Doesn't work

sudo rpi-update 18e0a0f9a31e7a3a47d9c4301c7705b980ab0516  < Doesn't work

sudo rpi-update 6cf2788e74b497c109865d052cd6fa68d094cf38 < downloads work fine here.

P33M:
Please open a separate issue detailing your findings. In particular, we need to be able to replicate a problem in order to debug it - can you provide an example script that can be run or Wireshark logs around the stall?

Issue #2599, opened 2018-06-28
Periodical IPv6 alignment issues


lategoodbye:
Hi,
if i connect my Raspberry Pi 3B+ (running Raspbian with 4.14.52-v7+) to my home network via Gigabit-Ethernet, i see a slowly increasing number of alignment issues:

```
# cat /proc/cpu/alignment 
User:		0
System:		172 (__csum_ipv6_magic+0x1c/0xc8)
Skipped:	0
Half:		0
Word:		0
DWord:		0
Multi:		172
User faults:	2 (fixup)
```

As far as i remember this issue is new and didn't exist in the last Raspbian release from April.

Possibly related: https://www.mail-archive.com/netdev@vger.kernel.org/msg232190.html

wizofe:
Similarly here: https://gist.github.com/wizofe/469d8c4f092e5ae7d9adc9147f83e167 from the #2587

cyph84:
Same thing seems to happen on Rpi 3B on kernel 4.14.50-2-ARCH

lategoodbye:
@wizofe 
Please provide some context to trigger the issue.

@cyph84 
Do you refer to wizofe's issue or mine?

cyph84:
@lategoodbye I mean the issue by @wizofe

wizofe:
@lategoodbye There's a detailed description on how to replicate if you click on the issue but apparently, @timg236 already will submit a firmware patch :)

wizofe:
@cyph84 How do you are triggering the alignment mismatch? Any logs?

timg236:
@wizofe The root cause for bug#2587 is a memory trample from the high quality analog audio processing and would only occur if audio was playing. The kernel oops will be pretty random depending on what bit of memory was trampled i.e. I don't think it will fix this issue.

lategoodbye:
Here is a tcpdump when the alignment issue occur. At least one of those 3 ICMPv6 messages seems to trigger this issue. A simple ping6 doesn't trigger it.

[alignment_issue.zip](https://github.com/raspberrypi/linux/files/2160238/alignment_issue.zip)

JamesH65:
No expert, but reading this https://www.kernel.org/doc/Documentation/arm/mem_alignment, can you 
`echo 1 > /proc/cpu/alignment` to turn on a bit more tracking which should narrow down what is causing the traps?

lategoodbye:
[...]

PR #2953, opened 2019-05-02
lan78xx: use default alignment for rx buffers

P33M:
The lan78xx uses a 12-byte hardware rx header, so there is no need
to allocate SKBs with NET_IP_ALIGN set. Removes alignment faults
in both dwc_otg and in ipv6 processing.


P33M:
@JamesH65 just pointed me at https://github.com/raspberrypi/linux/issues/2599
Where a patch apparently went upstream replicating the change made here. This change can be dropped if and when the upstream change filters through to us.

pelwell:
This is probably the patch going upstream: https://www.spinics.net/lists/netdev/msg545832.html
I notice it has no "Signed-off-by:" line, which usually means failure, and Patchwork says "Changes requested" even though I've not seen any real feedback to the patch. Don't hold your breath.

PR #2954, opened 2019-05-02
bcm2835-codec / mmal-vchiq fixes

6by9:
4 minor fixes for the V4L2 drivers and support framework.


pelwell:
In a similar vein, it looks as though `vchiq_mmal_component_init` might leak on error.

6by9:
> In a similar vein, it looks as though vchiq_mmal_component_init might leak on error.

Yes, I think you're right. More patches....

PR #2956, opened 2019-05-05
arm: dts: overlays: rpi-sense: add upstream humidity compatible

nullr0ute:
The upstream humidiity driver uses "st,hts221" for the compatible
string so add that in as well so it will work with an unmodified
upstream kernel driver. We leave the downstream as the priority.

### DIFF
--- a/arch/arm/boot/dts/overlays/rpi-sense-overlay.dts
+++ b/arch/arm/boot/dts/overlays/rpi-sense-overlay.dts
@@ -38,7 +38,7 @@
 			};
 
 			hts221-humid@5f {
-				compatible = "st,hts221-humid";
+				compatible = "st,hts221-humid", "st,hts221";
 				reg = <0x5f>;
 				status = "okay";
 			};


pelwell:
> We leave the downstream as the priority.

If I've understood you correctly, the commit seems to contradict the commit message. I expected the upstream compatibility string to be appended, not prepended.

nullr0ute:
I believe it matches right to left, so the "st,hts221-humid" matches first and then "st,hts221", eg the compatible at the top "brcm,bcm2708", "brcm,bcm2709"

pelwell:
My understanding is that it matches left to right: https://elinux.org/Device_Tree_Usage#Understanding_the_compatible_Property

The Linux code agrees: https://github.com/raspberrypi/linux/blob/rpi-4.19.y/drivers/of/base.c#L482
and: https://github.com/raspberrypi/linux/blob/rpi-4.19.y/drivers/of/base.c#L513

pelwell:
The compatible strings at the outer level of the overlays are ignored - they should probably all just say "brcm,bcm2835".

nullr0ute:
Updated

XECDesign:
IIRC, `st,hts221-humid` isn't actually used by anything, it was just a placeholder, so it could just be replaced.

The included sense hat libraries access the sensor via /dev/i2c-1, so if this option is enabled along with the hts221 driver, it will break astropi code unless everything is updated to use IIO instead.

In other words, it would be a bad idea to add `CONFIG_HTS221=m` with this change and dtoverlay=rpi-sense or an updated EEPROM.

PR #2957, opened 2019-05-07
[5.1] switch pcm512x soundcard drivers to set_bclk_ration

HiassofT:
since kernel 5.1 (commit ccc8d6c7b6d2) the upstream pcm512x driver implements the set_bclk_ratio interface which we should use in the soundcard drivers instead of the set_tdm_slot interface (which I added in downstream RPi kernel 4.14 and has now been removed from RPi kernel 5.1).

Without this PR the Allo Boss DAC and Hifiberry DAC Plus / DAC Plus ADC cards are probably broken as calling snd_soc_dai_set_tdm_slot for the pcm512x DAI returns an error.

@allocom @hifiberry could you do a runtime test with this PR? I don't have any of these cards so can't check myself

### DIFF
--- a/sound/soc/bcm/allo-boss-dac.c
+++ b/sound/soc/bcm/allo-boss-dac.c
@@ -288,12 +288,10 @@ static int snd_allo_boss_hw_params(
 			return ret;
 	}
 
-	ret = snd_soc_dai_set_tdm_slot(rtd->cpu_dai, 0x03, 0x03,
-		channels, width);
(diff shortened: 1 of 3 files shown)


pelwell:
Any feedback, @allocom @hifiberry?

hifiberry:
Sorry, didn't have time yet to look into this. I'll try to test this next week.

allocom:
@pelwell  we completed the first level testing and it is working fine.

pelwell:
Thanks - that's good enough for me.

PR #2961, opened 2019-05-10
V4L2 codec frame rate/timestamp fixes

6by9:
Should sort the issue reported in https://www.raspberrypi.org/forums/viewtopic.php?f=38&t=240170&p=1465986

Still compile testing in case of bad cherry-pick, hence WIP tag.

### DIFF
--- a/drivers/staging/vc04_services/bcm2835-codec/bcm2835-v4l2-codec.c
+++ b/drivers/staging/vc04_services/bcm2835-codec/bcm2835-v4l2-codec.c
@@ -823,7 +823,8 @@ static void op_buffer_cb(struct vchiq_mmal_instance *instance,
 		vb2->flags |= V4L2_BUF_FLAG_LAST;
 	}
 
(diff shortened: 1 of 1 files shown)


6by9:
Doh, that serves me right for adding a "trivial" cast that breaks the build.

6by9:
Build issue fixed.
Good to review now.

popcornmix:
Looks sane to me.
The numerator/denominator swap looked surprising to me, but I guess one uses a frequency base and one uses time base, so it's probably deliberate.

6by9:
> The numerator/denominator swap looked surprising to me, but I guess one uses a frequency base and one uses time base, so it's probably deliberate.

Correct. V4L2 uses frame interval. MMAL uses frame rate.

6by9:
Rebased.
Any further comments?

6by9:
@pelwell Any comments, or can we merge?

PR #2968, opened 2019-05-13
Enable PIDs cgroup

hgontijo:
My use case to is to allow Kubernetes master to run on Raspberry Pi 3.
Kubernetes introduced [Pid limiting](https://github.com/kubernetes/enhancements/blob/master/keps/sig-node/20190129-pid-limiting.md),
on v.1.14.0 which includes [Pod Pids Limit](https://github.com/kubernetes/kubernetes/commit/bce9d5f2043bd86964c9fec80d466e47776071bc)
and [Node Pids Limit](https://github.com/kubernetes/kubernetes/commit/2597a1d97ef4d8f54b1ca661453e32794b756909#diff-fa76bb6ae2d9b4bb3d023737fe5e6029R333)
that requires pids cgroup.

Kernel compiled with this option and running on Raspberry 3 Model B+:
```
$ uname -ar; cat /proc/cgroups
Linux kube-master 4.14.114-v7+ #2 SMP Fri May 10 05:27:17 UTC 2019 armv7l GNU/Linux
#subsys_name	hierarchy	num_cgroups	enabled
cpuset	6	30	1
cpu	3	113	1
cpuacct	3	113	1
blkio	8	113	1
memory	2	116	1
devices	4	113	1
freezer	7	30	1
net_cls	9	30	1
pids	5	118	1
```

Fix https://github.com/teamserverless/k8s-on-raspbian/issues/16.


pelwell:
rpi-4.14.y is our stable branch, only being updated with fixes. Is support in rpi-4.19.y acceptable?

hgontijo:
Ideally in rpi-4.14.y but it's your call. Is correct to say that releasing to 4.14 in here would make it  faster to a Raspbian release? Another question, when is 4.19 expected to be out?

popcornmix:
There won't be any more 4.14 kernels built for raspbian.
4.19 will be the next stable kernel (probably reaching apt within the next week).
See: https://www.raspberrypi.org/forums/viewtopic.php?f=29&t=224931

pelwell:
The PIDS cgroup is now enabled in rpi-4.19.y (see https://github.com/raspberrypi/linux/commit/ac0945c841b352614d21429afcf0b953423aa99e). Expect an rpi-update build in the next few days.

hgontijo:
Thank you!

Issue #2966, opened 2019-05-12
cma_alloc: alloc failed, req-size: 2025 pages, ret: -12


StuartIanNaylor:
cma_alloc: alloc failed, req-size: 2025 pages, ret: -12

4.19.40-v7+ Pi3b

pelwell:
A bit more context would be helpful:

1. Is there any additional hardware attached?

2. What was the Pi doing at the time?

3. Was there any additional indication of what failed?

StuartIanNaylor:
I presumed it would be common after an rpi-update its just vanilla pi3 and displayed on boot but part of the kernel log. Fresh April img with rpi-update.
```
May 12 12:30:24 raspberrypi kernel: [    0.000000] Kernel command line: coherent_pool=1M 8250.nr_uarts=0 bcm2708_fb.fbwidth=1920 bcm2708_fb.fbheight=1080 bcm2708_fb.fbswap=1 vc_mem.mem_base=0x3ec00000 vc_mem.mem_size=0x40000000  dwc_otg.lpm_enable=0 console=ttyS0,115200 console=tty1 root=PARTUUID=bdedf1a2-02 rootfstype=ext4 elevator=deadline fsck.repair=yes rootwait quiet splash plymouth.ignore-serial-consoles
May 12 12:30:24 raspberrypi kernel: [    0.000000] Dentry cache hash table entries: 131072 (order: 7, 524288 bytes)
May 12 12:30:24 raspberrypi kernel: [    0.000000] Inode-cache hash table entries: 65536 (order: 6, 262144 bytes)
May 12 12:30:24 raspberrypi kernel: [    0.000000] Memory: 939088K/970752K available (8192K kernel code, 627K rwdata, 2164K rodata, 1024K init, 821K bss, 23472K reserved, 8192K cma-reserved)
May 12 12:30:24 raspberrypi kernel: [    0.000000] Virtual kernel memory layout:
May 12 12:30:24 raspberrypi kernel: [    0.000000]     vector  : 0xffff0000 - 0xffff1000   (   4 kB)
May 12 12:30:24 raspberrypi kernel: [    0.000000]     fixmap  : 0xffc00000 - 0xfff00000   (3072 kB)
May 12 12:30:24 raspberrypi kernel: [    0.000000]     vmalloc : 0xbb800000 - 0xff800000   (1088 MB)
[...]

PR #2970, opened 2019-05-13
Revert "video: bcm2708_fb: Try allocating on the ARM and passing to VPU"

6by9:
This reverts commit ca36c709fce57e8023d2b8b354376bf161601a49.

The driver tries a cma_alloc to avoid using gpu_mem, but should
that fail the core code is logging an error with no easy way to
test whether it will succeed or fail first.

Revert until we either totally give up on gpu_mem and increase
CMA always, or find a way to try an allocation.


6by9:
Likely cause of #2966

PR #2975, opened 2019-05-16
sound: pcm512x-codec: Adding 352.8kHz samplerate support

klslz:
352.8kHz support can be added to the pcm512c-codec.

This patch is around for years and usually gets applied in custom kernels (e.g. piCoreplayer). 
It allows to convert DoP-128 straight to 352.8kHz PCM.
And it also allows synchronous upsampling from x*44.1, mainly to bypass the internal DSP.

The datasheet doesn't show that 352.8kHz is supported. That's why it is probably missing in the codec.

### DIFF
--- a/sound/soc/codecs/pcm512x.c
+++ b/sound/soc/codecs/pcm512x.c
@@ -542,7 +542,7 @@ static unsigned long pcm512x_ncp_target(struct pcm512x_priv *pcm512x,
 
 static const u32 pcm512x_dai_rates[] = {
 	8000, 11025, 16000, 22050, 32000, 44100, 48000, 64000,
-	88200, 96000, 176400, 192000, 384000,
+	88200, 96000, 176400, 192000, 352800, 384000,
 };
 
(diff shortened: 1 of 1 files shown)


pelwell:
@clivem You tried to upstream this patch years ago. What happened to it?

clivem:
@pelwell Not sure about that specific patch. I gave up submitting any audio patches to upstream. That codec one might not have even been submitted. I have a vague recollection of trying to get the sample rate increased on the Pi I2S driver first..... IIRC, submitted that in '16. It was ignored at the time, even though Eric Anholt added a Reviewed-by, and it wasn't until over a year later that Matthias Reichl re-submitted it with his name on it and Mark King accepted it. In the meantime, for other reasons I lost all interest in audio on Pi.

pelwell:
Thanks, Clive. I'll have another go at pushing it, but until then we can accept it here.

klslz:
Great. Thx.

PR #2991, opened 2019-05-28
Update ethoc.c

Dharma-B:


pelwell:
I think you've created this Pull Request by mistake. The ethoc driver is an unmodified upstream driver, so we won't normally accepted downstream changes, and all you are doing is adding some kernel log spam.

PR #3000, opened 2019-06-04
spi-bcm2835: Always initialize local variable in bcm2835_wr_fifo_count

mbuesch:
If count is less than 4 and memcpy copies less than 4 bytes into var,
some bytes remain uninitialized. These uninitialized bytes are taken
from var and written to the BCM2835_SPI_FIFO register.

While the hardware probably is able to ignore these extra bytes, it's
Undefined Behavior to read uninitialized bytes. So don't do that.

### DIFF
--- a/drivers/spi/spi-bcm2835.c
+++ b/drivers/spi/spi-bcm2835.c
@@ -183,12 +183,11 @@ static inline void bcm2835_wr_fifo_count(struct bcm2835_spi *bs, int count)
 	bs->tx_len -= count;
 
 	while (count > 0) {
+		val = 0;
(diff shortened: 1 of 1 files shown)


popcornmix:
The fix looks correct to me. But this is an upstream kernel file, so it really needs to be fixed there (we will get the fix once it's accepted there).

See: https://kernelnewbies.org/UpstreamMerge/SubmittingPatches

mbuesch:
Thanks @popcornmix.
I was unsure how modifications to these upstream files work. Thanks for explaining it to me.
I'll submit this patch upstream and CC stable, if required.

PR #3009, opened 2019-06-11
Fix two Makefiles for building to separate directory.

jabjoe:
Include paths added using relative paths, instead of using srcpath.
This would mean when built out of tree headers were not found.

Signed-off-by: Joe Burmeister <joe.burmeister@devtank.co.uk>


pelwell:
No objections from me. @popcornmix?

JamesH65:
Shouldn't this go upstream first? Seems like it's not part of the tree we change.

pelwell:
You won't find either driver upstream - dwc_otg is Pi-only, and the rtl8192cu driver is usually out-of-tree.

JamesH65:
Ah, OK, in which case LGTM.

popcornmix:
Would be nice if the rtl8192cu patch is submitted here:
https://github.com/pvaret/rtl8192cu-fixes

as that is where we update from.

jabjoe:
I didn't know where it was from. I did look at mainline, but it wasn't there. I can try the rtl8192cu repo, but it does say "!! This driver is DEPRECATED !!" at the top of the README, which makes me think they might not care about merge requests.

popcornmix:
Repo has said deprecated for a number of months, but they are accepting pull requests (e.g. one was merged 19 days ago).

kbingham:
@pelwell This looks related too, ... but somehow the PR/branches seem to have got a bit messed up.

PR #3011, opened 2019-06-13
i2c: bcm2835: Model Divider in CCF

mcdermj:
Commit bebff81fb8b9216eb4fba22cf910553621ae3477 upstream.

Model the I2C bus clock divider as a part of the Core Clock Framework.
Primarily this removes the clk_get_rate() call from each transfer.
This call causes problems for slave drivers that themselves have
internal clock components that are controlled by an I2C interface.
When the slave's internal clock component is prepared, the prepare
lock is obtained, and it makes calls to the I2C subsystem to
command the hardware to activate the clock.  In order to perform
the I2C transfer, this driver sets the divider, which requires
it to get the parent clock rate, which it does with clk_get_rate().
Unfortunately, this function will try to take the clock prepare
lock, which is already held by the slave's internal clock calls
creating a deadlock.

Modeling the divider in the CCF natively removes this dependency
and the divider value is only set upon changing the bus clock
frequency or changes in the parent clock that cascade down to this
divisor.  This obviates the need to set the divider with every
transfer and avoids the deadlock described above.  It also should
provide better clock debugging and save a few cycles on each
transfer due to not having to recalcuate the divider value.

Signed-off-by: Annaliese McDermond <nh6z@nh6z.net>
Acked-by: Stefan Wahren <stefan.wahren@i2se.com>
Reviewed-by: Eric Anholt <eric@anholt.net>
Signed-off-by: Wolfram Sang <wsa@the-dreams.de>

### DIFF
--- a/drivers/i2c/busses/i2c-bcm2835.c
+++ b/drivers/i2c/busses/i2c-bcm2835.c
@@ -12,6 +12,8 @@
  */
 
 #include <linux/clk.h>
+#include <linux/clkdev.h>
+#include <linux/clk-provider.h>
 #include <linux/completion.h>
 #include <linux/err.h>
 #include <linux/i2c.h>
@@ -71,9 +73,7 @@ struct bcm2835_debug {
 struct bcm2835_i2c_dev {
 	struct device *dev;
 	void __iomem *regs;
-	struct clk *clk;
 	int irq;
-	u32 bus_clk_rate;
 	struct i2c_adapter adapter;
 	struct completion completion;
 	struct i2c_msg *curr_msg;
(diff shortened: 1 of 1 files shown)


kkoopa:
This (likely) broke the functionality of every single i2c device I have. The 4.19.49 release worked fine.
Built from source, modules and drivers built fine and loaded without any error messages but no i2c/iio devices appear.
Unfortunately I have no time to investigate further as vacation is coming up.

pelwell:
The problem appears to be a startup ordering issue. For now, building as a module (as the standard defconfigs do) avoids the problem, but I can see how that might not be an option for some.

kkoopa:
Alright, thanks for a (very) quick response. Good that you are aware and looking into it. I am reverting to 4.19.49 in the meantime.

pelwell:
Or you could just revert this commit.

drivers/i2c/busses/i2c-bcm2835.c
```diff
@@ -485,6 +535,35 @@ static int bcm2835_i2c_probe(struct platform_device *pdev)
 		return -ENODEV;
 	}
 
+	mclk_name = of_clk_get_parent_name(pdev->dev.of_node, 0);
+
```
pelwell, line 539: A minimal fix for the case where i2c_bcm2835 is in-kernel is to add:
```
	if (!mclk_name) {
		free_irq(i2c_dev->irq, i2c_dev);
		return -EPROBE_DEFER;
	}
```
This equates a lack of a clock name with a lack of a clock. There may be a better function to use to determine if the clock exists, but the lack of a name would still cause a failure later on so this may be the best solution.
However, systematic error handling is generally lacking in this function, so a more thorough overhaul is probably called for.

mcdermj:
This is actually where I thought the problem was upon thinking about it in bed last night.  I'll see if I can get a patch together later today.

lategoodbye:
I think we should check the result of of_clk_get_parent_count() before.

pelwell:
From what I remember it will still return 1 even though the parent isn't ready, but I could be mistaken.

mcdermj:
I have proposed a solution in #3018.  Please let me know if that looks acceptable to folks and I'll run it through upstream and resubmit the PR.

PR #3012, opened 2019-06-13
RFC: Fix DMA channel and memory leak in vc4

boatbodger:
Adding a non-Raspberry DSI display.  Even with panel driver and DTB apparently correct, startup was frequently failling with a dmesg entry:
 [ 4.690030] [drm:vc4_dsi_bind [vc4]] *ERROR* Failed to get DMA channel: mask:0001 chan:-19

The vc4/panel startup 'pas a doble' means vc4 requests EPROBE_DEFER before the panel is ready, sometimes several times.

It was found that each iteration was requesting a new DMA channel, and the system was running out.
It was also noticed that small amounts of memory were being allocated each time.
The attached patch seems to fix the second memory leak and the DMA leak.
The patch file contains a proposal for a third patch to fix the first memory leak, but when applied, breaks something, resulting in a kernel oops.
The patch as offered also includes additional debugging messages which could be removed for production.
I have tested this patch in conjunction with the very recent fixes from pelwell which also address a DMA leak.
PS: this is the first time I've done a pull request, so please forgive me if the etiquette isn't quite right...

### DIFF
--- a/drivers/gpu/drm/vc4/vc4_dsi.c
+++ b/drivers/gpu/drm/vc4/vc4_dsi.c
@@ -1674,8 +1674,8 @@ static int vc4_dsi_bind(struct device *dev, struct device *master, void *data)
 
 rel_dma_exit:
 	dma_release_channel(dsi->reg_dma_chan);
-	return ret;
 
+	return ret;
 }
 
 static void vc4_dsi_unbind(struct device *dev, struct device *master,
--- a/drivers/gpu/drm/vc4/vc4_dsi.c
+++ b/drivers/gpu/drm/vc4/vc4_dsi.c
@@ -1496,6 +1496,12 @@ vc4_dsi_init_phy_clocks(struct vc4_dsi *dsi)
 				      dsi->clk_onecell);
 }
 
+static int release_return(struct dma_chan *dmachan, int retval){
+	if (dmachan)
+		dma_release_channel(dmachan);
+	return retval;
+}
+
 static int vc4_dsi_bind(struct device *dev, struct device *master, void *data)
 {
 	struct platform_device *pdev = to_platform_device(dev);
@@ -1507,22 +1513,21 @@ static int vc4_dsi_bind(struct device *dev, struct device *master, void *data)
 	const struct of_device_id *match;
 	dma_cap_mask_t dma_mask;
 	int ret;
-	static int callcount = 0;
+
 	match = of_match_device(vc4_dsi_dt_match, dev);
 	if (!match)
 		return -ENODEV;
-	++callcount;
+
 	dsi->port = (uintptr_t)match->data;
 
-//	if (!dsi->encoder){		// Consider this conditional to prevent memory leak
-		vc4_dsi_encoder = devm_kzalloc(dev, sizeof(*vc4_dsi_encoder),
-				       GFP_KERNEL);
-		if (!vc4_dsi_encoder)
-			return -ENOMEM;
-		vc4_dsi_encoder->base.type = VC4_ENCODER_TYPE_DSI1;
-		vc4_dsi_encoder->dsi = dsi;
-		dsi->encoder = &vc4_dsi_encoder->base.base;
-//	}
+	vc4_dsi_encoder = devm_kzalloc(dev, sizeof(*vc4_dsi_encoder),
+			       GFP_KERNEL);
+	if (!vc4_dsi_encoder)
+		return -ENOMEM;
+	vc4_dsi_encoder->base.type = VC4_ENCODER_TYPE_DSI1;
+	vc4_dsi_encoder->dsi = dsi;
+	dsi->encoder = &vc4_dsi_encoder->base.base;
+
 	dsi->regs = vc4_ioremap_regs(pdev, 0);
 	if (IS_ERR(dsi->regs))
 		return PTR_ERR(dsi->regs);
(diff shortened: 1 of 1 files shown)


drivers/gpu/drm/vc4/vc4_dsi.c
```diff
@@ -1508,21 +1507,22 @@ static int vc4_dsi_bind(struct device *dev, struct device *master, void *data)
 	const struct of_device_id *match;
 	dma_cap_mask_t dma_mask;
 	int ret;
-
+	static int callcount = 0;
```
pelwell: Please leave a blank line after local variable declarations. But this is only used for debug output which doesn't really belong here, in a PR/upstream patch, so just delete it.

drivers/gpu/drm/vc4/vc4_dsi.c
```diff
@@ -1508,21 +1507,22 @@ static int vc4_dsi_bind(struct device *dev, struct device *master, void *data)
...
 	match = of_match_device(vc4_dsi_dt_match, dev);
 	if (!match)
 		return -ENODEV;
-
+	++callcount;
```
pelwell: Delete.

drivers/gpu/drm/vc4/vc4_dsi.c
```diff
@@ -1508,21 +1507,22 @@ static int vc4_dsi_bind(struct device *dev, struct device *master, void *data)
...
 	dsi->port = (uintptr_t)match->data;
 
-	vc4_dsi_encoder = devm_kzalloc(dev, sizeof(*vc4_dsi_encoder),
+//	if (!dsi->encoder){		// Consider this conditional to prevent memory leak
```
pelwell: Delete.

drivers/gpu/drm/vc4/vc4_dsi.c
```diff
@@ -1508,21 +1507,22 @@ static int vc4_dsi_bind(struct device *dev, struct device *master, void *data)
...
-	vc4_dsi_encoder->base.type = VC4_ENCODER_TYPE_DSI1;
-	vc4_dsi_encoder->dsi = dsi;
-	dsi->encoder = &vc4_dsi_encoder->base.base;
-
+		if (!vc4_dsi_encoder)
+			return -ENOMEM;
+		vc4_dsi_encoder->base.type = VC4_ENCODER_TYPE_DSI1;
+		vc4_dsi_encoder->dsi = dsi;
+		dsi->encoder = &vc4_dsi_encoder->base.base;
+//	}
```
pelwell: Delete.

drivers/gpu/drm/vc4/vc4_dsi.c
```diff
@@ -1538,7 +1538,9 @@ static int vc4_dsi_bind(struct device *dev, struct device *master, void *data)
 	 * so set up a channel for talking to it.
 	 */
 	if (dsi->port == 1) {
-		dsi->reg_dma_mem = dma_alloc_coherent(dev, 4,
+		dev_warn(dev,"Count %d: dsi->r.d.mem=0x%08x, dsu->r.d.chan=%d\n",callcount,(unsigned int)dsi->reg_dma_mem, (unsigned int)dsi->reg_dma_chan);
+		if (!dsi->reg_dma_mem)			//Consider this conditionial to prevent memory leak
```
pelwell: I think this is the wrong approach:
1. An unsuccessful bind should probably have no effect on the system state. No additional action should be required to clean up.
2. An unbind should reverse the effects of the bind.

1) requires that error paths result in the releasing of any resources previously claimed during the bind attempt. This may end up duplicating code that appears in unbind, or it may be possible to factor it out into a "clean-up" function that is called from both places.

2) is not satisfied here because nowhere in the code is a call to e.g. `dma_release_channel`.

pelwell (changes requested):
Thank you for trying to improve this code, but I think your general approach is the wrong one. If you follow the pattern I outline in the line comments you will probably find that it doesn't leak and doesn't crash.

If you have changes to submit you have two options:
1. Append a followup patch to your forked branch and it will appear here. The commits can be squashed into one at merge time. This is helpful if the patch is extensive, the changes are small and the end result will be a single commit.
2. Fix up your commits locally to look how you want them to, then force push (`git push -f`) to your forked branch and the whole PR will be rewritten.

drivers/gpu/drm/vc4/vc4_dsi.c
boatbodger: Thank you for the feedback and helpful setting out of the principles to which one one should adhere - they make absolute sense.
I will re-do the changes in the way you suggest - probably early next week.

boatbodger:
I have reworked the patch on the basis of releasing any required resources if it is necessary to error out of the bind.  For most memory allocs that had already been ensured by using managed variants, but there was a one which wasn't.
As no managed calls seem to exist for DMA channels, the code has been modified to ensure its explicit release on error exits.

pelwell:
That's looking much better from a functionality point of view. The remaining changes I would like are essentially cosmetic/stylistic:

* Don't make unnecessary whitespace changes.
* The added DRM_INFO is unnecessary/undesirable in production code.
* You only ever call the release_return function in situations where the DMA channel has been claimed, so the release doesn't need to be conditional.
* Despite the old "goto considered harmful" article, using `goto` to branch to common error handling cleanup code at the end of a function is standard practice in Linux drivers. In the case of multiple resources which may need to be released, multiple labels are used:
```
        ....

        if (!claim_resource_1()) {
                ret = -EMUMBLE;
                goto err_exit;
        }

        ptr = claim_resource_2();
        if (IS_ERR(ptr)) {
                ret = PTR_ERR(ptr);
                goto err_release_1;
        }

        ret = claim_resource_3();
        if (ret)
                goto err_release_2;
...

        return 0;

err_release_2:
        free_resource_2(ptr);
err_release_1:
        free_resource_1();
err_exit:
        return ret;
}
```
* Although Linux is fond of using helper functions, trusting the compiler to inline what it needs to, now that the conditionality and the error-code passing is removed `release_return` is just `dma_release_channel`.

boatbodger:
Thanks for the helpful feedback, which I believe I've reflected in this new P/R.
Happy to take on board any further points you may have, and thank you for your patience.

boatbodger:
@pelwell I wonder if you had a chance to review the changes I made following your feedback? (I may not quite have the hang of the pull request process - I'm hoping my most recent 'commit' would have 'refreshed' the pull request for you, but I'm not totally sure...)

pelwell:
Sorry for the delayed response - I'm temporarily reassigned and unable to test the latest patch, but it looks OK so will probably get merged early next week.

pelwell (approved):

pelwell:
[ Sorry for the abnormally slow responses - the normally slow responses should be in effect now. ]

That looks good (with a minor whitespace change). Do you have a "Signed-off-by: <email>" line to add? If you reply with one I can add it when squashing down to a single commit.

boatbodger:
>Do you have a "Signed-off-by: " line to add? If you reply with one I can add it when squashing down to a single commit.
Signed-off-by: Chris G Miller <chris@creative-electronics.net>

lategoodbye:
Would be nice to get this upstream

PR #3015, opened 2019-06-16
[bcm2835-codec] bytesperline was wrong if width is aligned.

zillevdr:
bytesperline was wrong if width is aligned. bytesperline is defined: Distance in bytes between the leftmost pixels in two adjacent lines.

### DIFF
--- a/drivers/staging/vc04_services/bcm2835-codec/bcm2835-v4l2-codec.c
+++ b/drivers/staging/vc04_services/bcm2835-codec/bcm2835-v4l2-codec.c
@@ -755,7 +755,7 @@ static void handle_fmt_changed(struct bcm2835_codec_ctx *ctx,
 	q_data = get_q_data(ctx, V4L2_BUF_TYPE_VIDEO_CAPTURE);
 	q_data->crop_width = format->es.video.crop.width;
 	q_data->crop_height = format->es.video.crop.height;
-	q_data->bytesperline = format->es.video.crop.width;
+	q_data->bytesperline = format->es.video.width;
 	q_data->height = format->es.video.height;
 	q_data->sizeimage = format->buffer_size_min;
 	if (format->es.video.color_space)
--- a/drivers/staging/vc04_services/bcm2835-codec/bcm2835-v4l2-codec.c
+++ b/drivers/staging/vc04_services/bcm2835-codec/bcm2835-v4l2-codec.c
@@ -1106,9 +1106,9 @@ static int vidioc_s_fmt(struct bcm2835_codec_ctx *ctx, struct v4l2_format *f,
 	bool update_capture_port = false;
 	int ret;
 
-	v4l2_dbg(1, debug, &ctx->dev->v4l2_dev,	"Setting format for type %d, wxh: %dx%d, fmt: %08x, size %u\n",
+	v4l2_dbg(1, debug, &ctx->dev->v4l2_dev,	"Setting format for type %d, wxh: %dx%d, fmt: %.4s, size %u\n",
 		 f->type, f->fmt.pix.width, f->fmt.pix.height,
-		 f->fmt.pix.pixelformat, f->fmt.pix.sizeimage);
+		 (char*)&f->fmt.pix.pixelformat, f->fmt.pix.sizeimage);
 
 	vq = v4l2_m2m_get_vq(ctx->fh.m2m_ctx, f->type);
 	if (!vq)


popcornmix:
Looks correct to me. @6by9 ?

drivers/staging/vc04_services/bcm2835-codec/bcm2835-v4l2-codec.c
```diff
@@ -1106,9 +1106,9 @@ static int vidioc_s_fmt(struct bcm2835_codec_ctx *ctx, struct v4l2_format *f,
 	bool update_capture_port = false;
 	int ret;
 
-	v4l2_dbg(1, debug, &ctx->dev->v4l2_dev,	"Setting format for type %d, wxh: %dx%d, fmt: %08x, size %u\n",
+	v4l2_dbg(1, debug, &ctx->dev->v4l2_dev,	"Setting format for type %d, wxh: %dx%d, fmt: %.4s, size %u\n",
 		 f->type, f->fmt.pix.width, f->fmt.pix.height,
-		 f->fmt.pix.pixelformat, f->fmt.pix.sizeimage);
+		 (char*)&f->fmt.pix.pixelformat, f->fmt.pix.sizeimage);
```
6by9, line 1111: NAK. 
The MSB is used to denote big-endian mode - see https://git.linuxtv.org/v4l-utils.git/tree/utils/common/v4l2-info.cpp#n142 or https://elixir.bootlin.com/linux/latest/source/drivers/media/v4l2-core/v4l2-ioctl.c#L1365

When trying to upstream the Unicam driver I had previously proposed upstream adding utility macros for doing this. Whilst it has never been merged there, we do have the commit here. https://github.com/raspberrypi/linux/commit/c3f4b15fcdca56a4f05eda5419d6d51ffbcb5e8c
So 
```
v4l2_dbg(1, debug, &ctx->dev->v4l2_dev,	"Setting format for type %d, wxh: %dx%d, fmt: " V4L2_FOURCC_CONV  ", size %u\n",
		 f->type, f->fmt.pix.width, f->fmt.pix.height,
		 V4L2_FOURCC_CONV_ARGS(f->fmt.pix.pixelformat), f->fmt.pix.sizeimage);
```
would be acceptable.

drivers/staging/vc04_services/bcm2835-codec/bcm2835-v4l2-codec.c
```diff
@@ -755,7 +755,7 @@ static void handle_fmt_changed(struct bcm2835_codec_ctx *ctx,
 	q_data = get_q_data(ctx, V4L2_BUF_TYPE_VIDEO_CAPTURE);
 	q_data->crop_width = format->es.video.crop.width;
 	q_data->crop_height = format->es.video.crop.height;
-	q_data->bytesperline = format->es.video.crop.width;
+	q_data->bytesperline = format->es.video.width;
```
6by9, line 758: I'm confused by this one, but it may just be the wording.
> bytesperline was wrong if width is aligned

If width is aligned, then ```crop.width``` and ```width``` are the same value.

Reviewing the context, this is totally wrong should the format be anything other than YUV because it doesn't take into account that there could be more than 1 byte per pixel.
What is needed is a call to ```get_bytesperline();```. I don't believe the firmware will ever change the colour format, so q_data->fmt is still valid as the format. If passing ```format->es.video.width``` as the width then it should always be aligned, but get_bytesperline will enforce that anyway.

Thanks for picking this up though - I'm now wondering if this is why ffmpeg was having issues and I had to patch it.

6by9 (changes requested):
Thanks for the patch, but a couple of issues to be addressed.

6by9:
Oh, and can you include "Signed-off-by:" tags in your commit text.
I am intending on upstreaming this driver, therefore need to be fairly strict on ownership, contributors, and distribution rights. 
https://github.com/raspberrypi/linux/blob/rpi-4.19.y/Documentation/process/submitting-patches.rst#11-sign-your-work---the-developers-certificate-of-origin

drivers/staging/vc04_services/bcm2835-codec/bcm2835-v4l2-codec.c
zillevdr: My patch is a quick hack. Your Macro are a cleaner solution.

drivers/staging/vc04_services/bcm2835-codec/bcm2835-v4l2-codec.c
zillevdr: get_bytesperline() is also good. Now are in bytesperline a wrong value. For example a video width 720 have a bytesperline 736. But in bytesperline are 720. Do you make a patch or I shall do it?

zillevdr:
I have still an other suggestion. If the capture buffers correct configured should handle_fmt_changed() test it and only if buffers not correct do queue_res_chg_event(). Now an application must do streamoff ... streamon.

6by9:
If the capture buffers are already correctly configured then the firmware shouldn't generate a MMAL_EVENT_FORMAT_CHANGED event at all. Note that it may generate one for setting the colourspace only (plumbing still to be completed).

If the firmware has detected a format change then it will stop producing frames until the output is disabled and re-enabled.

zillevdr:
I must clear it more. The app use only the v4l2_buffers. This capture buffers are correct configured. The mmal_buffers are not seen on the v4l2 API. This must do the wrapper. I think on the v4l2 API is no streamoff / streamon for that necessary. handle_fmt_changed() can test it and configure the mmal_buffer.

On exynos the equal procedure for MMAL_EVENT_FORMAT_CHANGED is done between STREAMON on OUTPUT and G_FMT on CAPTURE. [There on site 16](https://events.static.linuxfound.org/images/stories/pdf/lceu2012_debski.pdf) is described it. This is simpler for the v4l2 api user.

6by9:
Correctly configured based on what definition? The codec has parsed the headers and come out with a different definition of the image geometry.
The MMAL data buffers are the ones allocated by videobuf2, passed to the VPU for filling, and handed back out to your application, therefore the format of them must match in order for your application to correctly interpret them.

The official documentation for video decoding using V4L2 stateful M2M devices is still being finalised. The latest version posted to the mailing list is https://www.spinics.net/lists/linux-media/msg151893.html. That is the spec that drivers need to adhere to. (I know it is a little backward that they are only just getting around to documenting it).

Initialization step 4
> If the values configured by the client do not match those parsed by the decoder, a `Dynamic resolution change` will be triggered to reconfigure them.

From the  `Dynamic resolution change` section:
> The sequence starts when the decoder detects a coded frame with one or more of
> the following parameters different from those previously established (and
> reflected by corresponding queries):
> * coded resolution (``OUTPUT`` width and height),
> * visible resolution (selection rectangles),
> * the minimum number of buffers needed for decoding.
>
> Whenever that happens, the decoder **must** proceed as follows:

Not implementing that section of the spec, particularly when marked with a must, is reliant on your application always getting it exactly right, which is risky.
I'll also note that GStreamer doesn't implement V4L2_EVENT_SOURCE_CHANGE, but I've yet to see it have issues decoding streams. That may be due to the test set that I've been using so will expand the set.

I will look again at the handling here as colourspace and interlace status reporting does need to be addressed, but I'm not convinced that the port format changing is specifically wrong.

I also see they've updated the "Dynamic resolution change" section to require that the last buffer before the change get marked with ```V4L2_BUF_FLAG_LAST```. It's always fun working with a shifting spec.

zillevdr:
> Capture setup
> =============
> 
> 1.  Call :c:func:`VIDIOC_G_FMT` on the ``CAPTURE`` queue to get format for the
>     destination buffers parsed/decoded from the bitstream.

This sounds more then Exynos do it. In my app V4L2_EVENT_SOURCE_CHANGE is integratet. In the first capture setup get wrong dimensions(640 x 480). So DRM FB wrong configured. With DQBUF the first buffer get the right dimension. All configuration must cleaned and twice configure. This is not smart and need much time. So I parse the stream, get height and width and configure the capture buffer. But just the same. Must do streamoff / streamon.

But back to the issue wrong bytesperline. Do you make a patch or I shall do it?

6by9:
Apologies for the delay - life's been a touch hectic with the Pi4 launch (which obviously I couldn't say anything about when discussing this issue before).

I'll sort out a patch tomorrow to update bytesperline.

For reference, one of the V4L2 maintainers is hosting a copy of the proposed codec API docs at https://hverkuil.home.xs4all.nl/codec-api/uapi/v4l/dev-mem2mem.html

pelwell:
Superceded by #3129.

PR #3017, opened 2019-06-17
WIP: dt: overlays: Correct compatible string for ds1307

6by9:
The overlay was requesting maxim,ds1307, which after
af50371 i2c: core: report OF style module alias for devices registered via OF
means that it doesn't match.
Correct the string to dallas,ds1307

https://www.raspberrypi.org/forums/viewtopic.php?f=44&t=242902


6by9:
Forgot to fix the i2c-rtc-gpio overlay as well, and haven't got time to do it now.

pelwell:
There are a few other instances - I'll add them to my queue.

PR #3018, opened 2019-06-19
Monolithic kernel fixes

mcdermj:
This is a proposed fix to the issues described in #3011.  

The first patch is a cleanup patch because the probe function does not properly free the IRQ upon failure.  This can be relatively easily remedied by having it use the devm_ function to properly do resource management.

The second is the meat of the fix.  The real way to see if the clock is there is to restore the code to try to get it.  We just do this and then get the clock's name in the clock registration function.

This has been tested on both monolithic and module-based kernels to make sure it works correctly for both.  It seems to work fine with the hardware I have to test with here.  Commentary is always welcome.

If this looks acceptable, I'll get this committed upstream and resubmit a PR with the upstream commit ID.

Thanks for your patience and many apologies for breaking things for folks.

### DIFF
--- a/drivers/i2c/busses/i2c-bcm2835.c
+++ b/drivers/i2c/busses/i2c-bcm2835.c
@@ -521,20 +521,6 @@ static int bcm2835_i2c_probe(struct platform_device *pdev)
 	if (IS_ERR(i2c_dev->regs))
 		return PTR_ERR(i2c_dev->regs);
 
-	irq = platform_get_resource(pdev, IORESOURCE_IRQ, 0);
-	if (!irq) {
-		dev_err(&pdev->dev, "No IRQ resource\n");
-		return -ENODEV;
-	}
-	i2c_dev->irq = irq->start;
-
-	ret = request_irq(i2c_dev->irq, bcm2835_i2c_isr, IRQF_SHARED,
-			  dev_name(&pdev->dev), i2c_dev);
-	if (ret) {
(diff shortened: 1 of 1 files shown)


lategoodbye (commented):
In general devm_ functions are preferred, but there are corner cases regarding race conditions (IRQ is shared between 3 interfaces including VPU). So we must be careful to not replace a potential resource leak with a hard to debug race condition during driver unload. An alternative solution would be to move all the clk handling before the irq stuff. Please consider this only as a gut feeling.

lategoodbye (commented):
The clock fix looks good to me.

pelwell:
I'm happy with these patches, but I defer to @lategoodbye on the mysterious potential devm irq race condition.

pelwell:
I'm happy with these patches, but I defer to @lategoodbye on the mysterious potential devm irq race condition.

pelwell:
I'm happy with these patches, but I defer to @lategoodbye on the mysterious potential devm irq race condition.

mcdermj:
I’ll defer to @lategoodbye as well as that’s out of my area of expertise. I’ll respin the second patch with the IRQ registration code at the end and get it submitted for upstream review this evening.

mcdermj:
I submitted a fix upstream this evening.  As I stated before, I'll submit a PR when I have it accepted there.

mcdermj:
@pelwell This has been accepted upstream and seems to fix the issue.  Please feel free to let me know if there's something else I need to take care of.

Issue #3024, opened 2019-06-26
RPi 4 kernel 4.19.56 Compile error


Dark-Sky:
I tried to compile the kernel 4.19.56 kernel for the RPi 4 and and the compile errors out with this. I searched the include/ directory with grep and did not get any return on the undeclared. Possible missing header file?
```
`drivers/staging/vc04_services/vc-sm-cma/vc_sm.c: In function 'cache_op_to_func':
drivers/staging/vc04_services/vc-sm-cma/vc_sm.c:1271:10: error: 'dmac_inv_range' undeclared (first use in this function)
 1271 |   return dmac_inv_range;
      |          ^~~~~~~~~~~~~~
drivers/staging/vc04_services/vc-sm-cma/vc_sm.c:1271:10: note: each undeclared identifier is reported only once for each function it appears in
drivers/staging/vc04_services/vc-sm-cma/vc_sm.c:1274:10: error: 'dmac_clean_range' undeclared (first use in this function)
 1274 |   return dmac_clean_range;
      |          ^~~~~~~~~~~~~~~~
drivers/staging/vc04_services/vc-sm-cma/vc_sm.c:1277:10: error: 'dmac_flush_range' undeclared (first use in this function)
 1277 |   return dmac_flush_range;
      |          ^~~~~~~~~~~~~~~~
drivers/staging/vc04_services/vc-sm-cma/vc_sm.c: At top level:
drivers/staging/vc04_services/vc-sm-cma/vc_sm.c:1473:9: error: redefinition of 'struct vc_sm_cma_ioctl_clean_invalid_block'
 1473 |  struct vc_sm_cma_ioctl_clean_invalid_block {
      |         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from drivers/staging/vc04_services/vc-sm-cma/vc_sm.c:57:
./include/linux/broadcom/vc_sm_cma_ioctl.h:92:9: note: originally defined here
   92 |  struct vc_sm_cma_ioctl_clean_invalid_block {
      |         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
drivers/staging/vc04_services/vc-sm-cma/vc_sm.c: In function 'vc_sm_cma_compat_ioctl':
drivers/staging/vc04_services/vc-sm-cma/vc_sm.c:1497:1: warning: control reaches end of non-void function [-Wreturn-type]
 1497 | }`
```

pelwell:
[...]

PR #3030, opened 2019-06-28
configs: Drop V4L2 camera and codec drivers from bcmrpi3_defconfig

6by9:
They rely on mmal_vchiq, which in turn wants vc-sm-cma.
vc-sm-cma needs some attention for 64 bit, so drop it for now.


6by9:
Temporarily solves #3024 for Pi3.

pelwell:
For some values of "Solves".

6by9:
I'm working on vc-sm-cma still, so will address 64bit support when I get there.

It's not just the cache ioctls - mmal_vchiq can't see some of the functions exported from vc-sm-cma for some reason. I suspect it's related to vc-sm-cma being a bool vs module. Too much black magic in there for a bear of very little brain.

agherzan:
@pelwell should we remove V4L2 for `bcm2711_defconfig` as well?

6by9:
arm64/bcm2711_defconfig hasn't been a supported platform, therefore not building isn't a regression. 
arm64/bcmrpi3_defconfig not working was a regression.

I know what the issue is with vc-sm-cma. It's working but not quite in a shape to push.
Fixing that is the end goal rather, not dropping features.

agherzan:
I agree, but the "solved" comment points to an issue that is raised against RPI4 which uses `bcm2711_defconfig`. I don't think it hurts to just have it in both places just for people who want to have an initial try.

6by9:
Earlier comment updated.

pelwell:
@agherzan Yes - see 9d1deec.

agherzan:
Thanks @pelwell . I really appreciate it.

PR #3041, opened 2019-07-02
Cleanup backport of V4L2 patches

6by9:
I'd pulled in an early version of patch "media: vb2: Allow reqbufs(0) with "in use" MMAP buffers" which was updated before it was accepted. This means it fails v4l2-compliance testing because it is missing a bit of signalling that was added.

Revert the early version and pull in the relevant patches from upstream. Minor conflict resolution required as upstream also included the request API changes.


popcornmix:
Fine by me.

Issue #3032, opened 2019-06-28
aarch64 bcm2711_defconfig fails to get to early console


hvenev:
aarch64 `bcm2711_defconfig` with `CONFIG_BCM_VIDEOCORE` disabled (see https://github.com/raspberrypi/linux/issues/3031) fails to get to early console with the kernel command line `earlycon=pl011,0xfe201000 console=ttyAMA0,115200 root=/dev/mmcblk0p2 rootfstype=f2fs rootwait`. The `kernel7l.img` provided in https://github.com/raspberrypi/firmware with the same kernel command line and firmware boots just fine (and indeed starts printing to the console earlier than when `earlycon` is not set).

hvenev:
This happens both on current rpi-4.19.y (e8a66b4f610b3a20bae8f706256d230135916c26) with firmware from https://github.com/raspberrypi/firmware/commit/bcf40b5c2b94178c7564fb451098d44968e44af5 and raspberrypi-kernel_1.20190620-1 (883a1a5bf14abc9b3a014a3907772cd0b3129c8a) with firmware from 1.20190620+1 (https://github.com/raspberrypi/firmware/commit/88ca9081f5e51cdedd16d5dbc85ed12a25123201).

pelwell:
2711 v8 support is a work in progress. In particular, the v8 ARM stubs need GIC support.

hvenev:
Isn't early console initialized before the GIC?

pelwell:
Perhaps, but the L2 cache latencies need to be set before either of them.

hvenev:
OK... How do I reproduce the results from cdb78ce891f6c6367a69c0a46b5779a58164bd4b? Is it enough to set `total_mem=1024`? Also, where are the L2 cache latencies set for 32-bit arm?

agherzan:
@pelwell I see that https://github.com/raspberrypi/tools/commit/7f4a937e1bacbc111a22552169bc890b4bb26a94#diff-d9d684ceaf48229906f8a534b4cac276 includes gic support for v8 arm stubs. Isn't this included?

pelwell:
It's not built into the firmware yet, but if you build it yourself and add `armstub=...` to config.txt it should work.

agherzan:
That was what I suspected. Thanks. I"ll try that. Are there any other known issues even with this stub added?

pelwell:
[...]

PR #3045, opened 2019-07-03
configs: arm64/bcm2711: Add MMC_SDHCI_IPROC

agherzan:
This driver is used in the device tree for the emmc2 node.

See #3032


pelwell:
Thanks! This must have got lost somewhere along the way.

By the way, the next firmware release should include the ARMv8 + GIC stub.

Strit:
> Thanks! This must have got lost somewhere along the way.
> 
> By the way, the next firmware release should include the ARMv8 + GIC stub.

Awesome. Can't wait for it. Right now we are just hacking it in! :smiley:

agherzan:
Good stuff @pelwell

agherzan:
@Strit if you are using Yocto you will be having the "hack" built in soon. Just waiting on the builds.

PR #3052, opened 2019-07-05
overlays: Rename pi3- overlays to be less model-specific

pelwell:
See: https://github.com/raspberrypi/firmware/issues/1174


popcornmix:
Makes sense to me.

Issue #3060, opened 2019-07-08
Pi 4: rtl_test hangs with RTL2838U chips


P33M:
Doing rtl_test on a Pi 4 invariably results in:
```
pi@raspberrypi:~$ rtl_test
Found 1 device(s):
  0:  Realtek, RTL2838UHIDIR, SN: 000000041

Using device 0: Sweex DVB-T USB
Detached kernel driver
rtlsdr_read_reg failed with -7
rtlsdr_write_reg failed with -7
rtlsdr_read_reg failed with -7
rtlsdr_write_reg failed with -7
^C
```
This is a USB2.0 device, and the issue happens on both 0137a8 and 013701 bridge chip firmwares. A long sequence of short control transfers are used to poke various registers in the chip, and for some reason the VLI chip goes quiet after receiving a stall response from one of these transfers.

On some occasions, the entire sequence completes successfully.

pelwell:
Could this be related to #3054? Can you test with 3GB?

P33M:
The issue persists with total_mem=3072 (or 1024, for that matter).

P33M:
I have wireshark captures of a good trace vs a bad trace. In the bad case, all URBs submitted to the control endpoint after the first STALL is received are returned with status -ENOENT and the requested transfer doesn't go out on the bus. In the functional case, the kernel does not respond to the first transfer after the STALL response with -ENOENT but the actual status (which is another STALL response) and all subsequent requested transfers result in bus activity.

rtl_test isn't submitting any CLEAR_FEATURE ENDPOINT_HALT requests in either of these two situations.

P33M:
I get the same result on a x86 PC with a VLI chipset. Annoyingly, dynamic_debug upsets timing enough that the Pi never completes a sequence successfully. On the PC, here's two traces:
[workingtrace.txt](https://github.com/raspberrypi/linux/files/3368884/workingtrace.txt)
[failtrace.txt](https://github.com/raspberrypi/linux/files/3368885/failtrace.txt)

After the first stall response, the driver then goes and does something different.

mutability:
If you need more data / testers, there are several people seeing the same problem here: https://discussions.flightaware.com/t/new-raspberry-pi-available-pi-4/51584/28

P33M:
Once more, with feeling:
[a.txt](https://github.com/raspberrypi/linux/files/3372353/a.txt)
[b.txt](https://github.com/raspberrypi/linux/files/3372355/b.txt)
[...]

PR #3066, opened 2019-07-11
xhci: add quirk for host controllers that don't update endpoint DCS

P33M:
Seen on a VLI VL805 PCIe to USB controller. For non-stream endpoints
at least, if the xHC halts on a particular TRB due to an error then
the DCS field in the Out Endpoint Context maintained by the hardware
is not updated with the current cycle state.

Using the quirk XHCI_EP_CTX_BROKEN_DCS and instead fetch the DCS bit
from the TRB that the xHC stopped on.

See: https://github.com/raspberrypi/linux/issues/3060

### DIFF
--- a/drivers/usb/host/xhci-ring.c
+++ b/drivers/usb/host/xhci-ring.c
@@ -520,7 +520,10 @@ void xhci_find_new_dequeue_state(struct xhci_hcd *xhci,
 	struct xhci_virt_ep *ep = &dev->eps[ep_index];
 	struct xhci_ring *ep_ring;
 	struct xhci_segment *new_seg;
+	struct xhci_segment *halted_seg = NULL;
 	union xhci_trb *new_deq;
+	union xhci_trb *halted_trb;
+	int index = 0;
 	dma_addr_t addr;
 	u64 hw_dequeue;
 	bool cycle_found = false;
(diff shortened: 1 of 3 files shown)


leshniak:
I've just compiled and installed your branch on my RPi 4 – no issues on tvheadend (rtl2832u dongle) nor weird messages in dmesg so far. Thank you!

drivers/usb/host/xhci-ring.c
```diff
@@ -541,7 +544,27 @@ void xhci_find_new_dequeue_state(struct xhci_hcd *xhci,
 	hw_dequeue = xhci_get_hw_deq(xhci, dev, ep_index, stream_id);
 	new_seg = ep_ring->deq_seg;
 	new_deq = ep_ring->dequeue;
-	state->new_cycle_state = hw_dequeue & 0x1;
+
+	if (!(ep->ep_state & EP_HAS_STREAMS))
+		halted_seg = trb_in_td(xhci, cur_td->start_seg,
```
pelwell: It seems wasteful to calculate this regardless of whether or not the quirk is set.

drivers/usb/host/xhci-ring.c
```diff
@@ -541,7 +544,27 @@ void xhci_find_new_dequeue_state(struct xhci_hcd *xhci,
...
+				       cur_td->first_trb, cur_td->last_trb,
+				       hw_dequeue & ~0xf, false);
+	/*
+	 * Quirk: xHC write-back of the DCS field in the hardware dequeue
+	 * pointer is wrong - use the cycle state of the TRB pointed to by
+	 * the dequeue pointer.
+	 */
+	if (xhci->quirks & XHCI_EP_CTX_BROKEN_DCS && halted_seg) {
+		index = ((dma_addr_t)(hw_dequeue & ~0xf) - halted_seg->dma) /
```
JamesH65, line 559: Does the compiler/checkpatch not complain and tell you to add extra parentheses in here for clarity? I get that one all the time...

JamesH65: In the line above the one I commented on, the if. Sorry.

P33M: Neither checkpatch nor the compiler complained...

drivers/usb/host/xhci-ring.c
P33M: I wanted to avoid a somewhat awkward control flow -  
```
if(quirk && !ep_has_streams) {
    halted_seg = trb_in_td();
    if (halted_seg)
        use_trb_dcs;
    else
        use_hw_dcs;
} else {
    use_hw_dcs;
}
```
but if this has a lighter touch when the quirk is not set then I suppose it's better.

pelwell: I'd be happy with:
```
if(quirk && !ep_has_streams)
    halted_seg = trb_in_td();
if (halted_seg)
    use_trb_dcs;
else
    use_hw_dcs;
```

trejan:
All good here. This quirk patch fixes the RTL2832U problems. Tried all the various SDR tools and no errors.

jbirdlebough:
Cheers anyone know how to fix RPI4 running buster where the /dev/ttyACM0 is missing, AMA0 and USB0 do work.  Have critical RPI3 app that requires /dev/ttyACM0 and no access to source.  Any viable work arounds.   Thanks J

P33M:
This has nothing to do with this commit. Issues unrelated to the linux kernel should be posted on the community forum at https://www.raspberrypi.org/forum

PR #3073, opened 2019-07-13
configs: arm64/bcm2711: Use CONFIG_BRCMSTB_THERMAL instead of CONFIG_BCM2835_THERMAL

### LINKED ISSUE #3071: RPi: arm64: CONFIG_BRCMSTB_THERMAL missing in bcm2711_defconfig
**Describe the bug**
The arm64 bcm2711_defconfig doesn't enable CONFIG_BRCMSTB_THERMAL like it's arm counterpart does. The RPi4 device tree includes a thermal driver with compatible string `brcm,avs-tmon-bcm2838` which is matched by the brcmstb thermal driver, rather than bcm2835.

**To reproduce**
`cat /sys/class/thermal/thermal_zone0/temp`

**Expected behaviour**
Returns a temperature value

**Actual behaviour**
Reading fails with EINVAL
`cat: /sys/class/thermal/thermal_zone0/temp: Invalid argument`

**System**
Raspberry Pi 4 with locally-compiled arm64 kernel.

Enabling CONFIG_BRCMSTB_THERMAL in my config (and disabling CONFIG_BCM2835_THERMAL since it's unneeded) gets the thermal driver working as expected.

aswild:
The Raspberry Pi 4 uses the brcmstb thermal driver rather than brcm2835,
based on the device tree compatible string 'brcm,avs-tmon-bcm2838'. With
CONFIG_BRCMSTB_THERMAL enabled, reading temperature from
/sys/class/thermal/thermal_zone0/temp works as expected instead of
returning EINVAL.

Fixes: https://github.com/raspberrypi/linux/issues/3071
Signed-off-by: Allen Wild <allenwild93@gmail.com>


lategoodbye:
LGTM

PR #3074, opened 2019-07-13
arm64: make CONFIG_ARCH_BCM2835 select CONFIG_MFD_CORE

aswild:
The bcm2835-pm.c driver is compiled as a built-in for
CONFIG_ARCH_BCM2835, and its probe function calls devm_mfd_add_devices,
which comes from the MFD core.

A missing Kconfig dependency/selection means it's possible to select
CONFIG_MFD_CORE=m with CONFIG_ARCH_BCM2835=y, leading to build failures.

Fix by making CONFIG_ARCH_BCM2835 select CONFIG_MFD_CORE, which will
pull in the MFD core as a built-in rather than a module.

Signed-off-by: Allen Wild <allenwild93@gmail.com>

### DIFF
--- a/arch/arm64/Kconfig.platforms
+++ b/arch/arm64/Kconfig.platforms
@@ -25,6 +25,7 @@ config ARCH_BCM2835
 	select ARM_AMBA
 	select ARM_TIMER_SP804
 	select HAVE_ARM_ARCH_TIMER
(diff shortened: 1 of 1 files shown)


aswild:
Fixed in 4cc44111ce00064baf0939a2d21e53964e17eab6

PR #3080, opened 2019-07-16
arm64/mm: Limit the DMA zone for arm64

agherzan:
On RaspberryPi, only the first 1Gb can be used for DMA[1].

[1] http://lists.infradead.org/pipermail/linux-arm-kernel/2019-July/665986.html

Signed-off-by: Andrei Gherzan <andrei@balena.io>

### DIFF
--- a/arch/arm64/mm/init.c
+++ b/arch/arm64/mm/init.c
@@ -224,7 +224,7 @@ static void __init reserve_elfcorehdr(void)
 static phys_addr_t __init max_zone_dma_phys(void)
 {
 	phys_addr_t offset = memblock_start_of_DRAM() & GENMASK_ULL(63, 32);
-	return min(offset + (1ULL << 32), memblock_end_of_DRAM());
+	return min(offset + (1ULL << 30), memblock_end_of_DRAM());
 }
 
 #ifdef CONFIG_NUMA


pelwell:
I have no problem with this, provided it just keeps the coherent DMA buffers in the first 1GB and doesn't prevent a more capable DMA device from addressing any page.

vianpl:
I suggest using this:

```
diff --git a/arch/arm64/mm/init.c b/arch/arm64/mm/init.c
index f3c795278def..cad580bda548 100644
--- a/arch/arm64/mm/init.c
+++ b/arch/arm64/mm/init.c
@@ -170,8 +170,9 @@ static void __init reserve_elfcorehdr(void)
  */
 static phys_addr_t __init max_zone_dma_phys(void)
 {
-       phys_addr_t offset = memblock_start_of_DRAM() & GENMASK_ULL(63, 32);
-       return min(offset + (1ULL << 32), memblock_end_of_DRAM());
+       /*phys_addr_t offset = memblock_start_of_DRAM() & GENMASK_ULL(63, 32);*/
+       /*return min(offset + (1ULL << 32), memblock_end_of_DRAM());*/
+       return 0x40000000;
 }
```

It's pretty much the same,  but you get the proper ZONE_DMA32 areas in the boot log.

agherzan:
@vianpl I like that. I'll push. 
@pelwell I'm not sure of the impact on devices dma capable of accessing any page. I've mostly relayed this change and tested it locally. Even if this is a temporary fix, I reckon we should go forward until we implement something like bounce buffers.

pelwell:
Let me know when you've pushed the update.

agherzan:
@pewell @vianpl I pushed the change. I tweaked it a little.

agherzan:
@pelwell for whatever reason your mention above didn't go through.

PR #3082, opened 2019-07-16
FKMS hdmi_timings settings

6by9:
Queries the firmware for a custom HDMI mode, and adds that as the ONLY supported mode if found.
With the latest firmware PR, the call returns a result if hdmi_mode/hdmi_group are set to non-zero.

It may be possible to combine the EDID and a custom mode, but I'm not sure whether that is worth it or not.

Also a minor cleanup to use drm_mode_vrefresh in places where the refresh rate is actually wanted.

### DIFF
--- a/drivers/gpu/drm/vc4/vc4_firmware_kms.c
+++ b/drivers/gpu/drm/vc4/vc4_firmware_kms.c
@@ -1035,6 +1035,58 @@ vc4_fkms_connector_detect(struct drm_connector *connector, bool force)
 	return connector_status_connected;
 }
 
+/* Queries the firmware to populate a drm_mode structure for this display */
+static int vc4_fkms_get_fw_mode(struct vc4_fkms_connector *fkms_connector,
+				struct drm_display_mode *mode)
+{
+	struct vc4_dev *vc4 = fkms_connector->vc4_dev;
+	struct set_timings timings = { 0 };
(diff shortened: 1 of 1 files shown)


drivers/gpu/drm/vc4/vc4_firmware_kms.c
```diff
@@ -1104,57 +1166,15 @@ static int vc4_fkms_lcd_connector_get_modes(struct drm_connector *connector)
...
-		if (mb.timings.flags & TIMINGS_FLAGS_V_SYNC_POS)
-			fw_mode.flags |= DRM_MODE_FLAG_PVSYNC;
-		else
-			fw_mode.flags |= DRM_MODE_FLAG_NVSYNC;
-		if (mb.timings.flags & TIMINGS_FLAGS_INTERLACE)
-			fw_mode.flags |= DRM_MODE_FLAG_INTERLACE;
-
-		fw_mode.base.type = DRM_MODE_OBJECT_MODE;
 
+	if (vc4_fkms_get_fw_mode(fkms_connector, &fw_mode))
```
6by9: Inverted my logic here. Will fix.
Also needs to check for fw_mode.clock

6by9: Fixed

PR #3089, opened 2019-07-19
ALSA: I2S: add drivers for PCM3060 codec and board

jonronen:
Add I2S-based drivers for TI's PCM3060 codec and for a matching extension
board.

Signed-off-by: Jon Ronen-Drori <jon_ronen@yahoo.com>


jonronen:
Hi, please let me know if you can review this and merge it into the raspberry pi kernel.
These changes were tested successfully with the PCM3060 board on top of a raspberry pi 3a+.
I can submit this to other branches as well if you think it's needed.
Thanks.

arch/arm/boot/dts/overlays/Makefile
```diff
@@ -55,6 +55,7 @@ dtbo-$(CONFIG_ARCH_BCM2835) += \
 	hifiberry-dacplusadc.dtbo \
 	hifiberry-digi.dtbo \
 	hifiberry-digi-pro.dtbo \
+	pcm3060-board.dtbo \
```
pelwell: Makefile entries should be in alphabetical order.

arch/arm/boot/dts/overlays/README
```diff
@@ -893,6 +893,12 @@ Load:   dtoverlay=hifiberry-digi-pro
 Params: <None>
 
 
+Name:   pcm3060-board
```
pelwell: README entries should be in alphabetical order.

sound/soc/bcm/rpi-simple-soundcard.c
```diff
@@ -3,6 +3,7 @@
  * rpi-simple-soundcard.c -- ALSA SoC Raspberry Pi soundcard.
  *
  * Copyright (C) 2018 Raspberry Pi.
+ * Copyright (C) 2019 Jon Ronen-Drori <jon_ronen@yahoo.com>
```
pelwell: It's slightly poor form to claim copyright for such a small contribution to a file, IMHO.

sound/soc/bcm/rpi-simple-soundcard.c
```diff
@@ -16,6 +17,9 @@
  * adau1977-adc.c
  * by Andrey Grodzovsky <andrey2805@gmail.com>
  *
+ * pcm3060_board.c
+ * by Jon Ronen-Drori <jon_ronen@yahoo.com>
```
pelwell, line 20: This is OK, though.

pelwell (changes requested):
There is already an upstream pcm3060 driver introduced in 4.20 (see 6ee47d4a8dacfa484d526c0475730568d979de24) which provides proper I2C and SPI control over things like volume - your "driver" just assumes that the defaults are correct - so I'd rather back-port that into 4.19. There's a cosmetic follow-up commit but there's little point in cherry-picking that since it is already in 4.20.

The overlay and the changes to rpi-simple-card look OK though.

If you are going to resubmit with the cherry-picked codec, please include a "commit ... upstream." line in the commit message, i.e.
```
    ASoC: pcm3060: Add codec driver
    
    commit 6ee47d4a8dacfa484d526c0475730568d979de24 upstream.
    
    This commit adds support for TI PCM3060 CODEC.
    ...
```

sound/soc/bcm/rpi-simple-soundcard.c
jonronen: I'll remove this

jonronen:
Thank you for the review.

I went over the PCM3060 codec changes introduced in 4.20. Indeed they're richer than my simple driver.

However, my simple pcm3060 driver uses a third mode not covered by [6ee47d4](https://github.com/raspberrypi/linux/commit/6ee47d4a8dacfa484d526c0475730568d979de24): The two modes covered are I2C and SPI. The third mode is HW mode (see also [the 3060 spec](https://www.ti.com/lit/ds/symlink/pcm3060.pdf)). HW mode makes all the configuration on the board using pull-up and pull-down resistors, so there's no need to use an additional bus to configure the codec.

My board uses HW mode which requires yet an additional driver, therefore I'd like to resubmit my changes with an additional driver (e.g. pcm3060_hw_mode.c).

Please let me know if it makes sense.

Thanks.

pelwell:
Adding a "hardware mode" to the existing driver seems like a better solution in the long run, but it means modifying an upstream driver which we try to avoid. Let's keep it a separate driver for now - it will be simpler to manage.

What I'm not so keen on is that this is a driver for _your_ pcm3060 board - it's fixed at 48kHz in master mode, while others might want to use it in slave mode for more flexibility (but less fidelity). One could make the driver more flexible and configure it with device tree properties, but that's quite a bit to learn if you haven't done it before.

The smallest change I will accept is a new compatible string so it doesn't clash with the upstream driver in 4.20. (To be honest, I'm not sure how the correct version - I2C or SPI - is loaded if both are enabled). If you were to include the sample rate in the string it would allow for the addition of more rates and slave mode in the future.

HiassofT:
The only purpose of this pcm3060 codec seems to be to specify stream constraints (rates and formats) - there's not really much in it that's pcm3060 specific.

IMO it doesn't make much sense to add hardware mode to the newer, upstream codec driver, instead it's better to use the existing generic codec drivers (spdif transmitter/receiver) and add the rate+format constraints via the machine driver.

arch/arm/boot/dts/overlays/Makefile
jonronen: Fixed

arch/arm/boot/dts/overlays/README
jonronen: Fixed

sound/soc/bcm/rpi-simple-soundcard.c
jonronen: Removed

jonronen:
I uploaded another version. Changed the compatible string from "ti,pcm3060" to "ti,pcm3060hw". I also extended the range of sample rates to a broader range, compatible with the sample rates supported by the codec in hardware mode. Tested again on RPi 3A+.

jonronen:
> The only purpose of this pcm3060 codec seems to be to specify stream constraints (rates and formats) - there's not really much in it that's pcm3060 specific.
> 
> IMO it doesn't make much sense to add hardware mode to the newer, upstream codec driver, instead it's better to use the existing generic codec drivers (spdif transmitter/receiver) and add the rate+format constraints via the machine driver.

I don't see a difference between my pcm3060 codec driver and other codec drivers like pcm5102a, pcm1794a, and wm8782. All those drivers just define stream constraints and don't contain any specific codec information.

@HiassofT, could you please give an example of a driver that implements what you suggest?
Thanks.

HiassofT:
the problem with pcm5102a and pcm1794a is that these codec drivers tend to keep piling up downstream and aren't really necessary if you already have a machine driver.

Have a look eg at the pisound driver how to add various constraints - you just need to define a .startup hook and then use the various snd_pcm_hw_constraint_xxx functions (eg _single for rate and _mask64 for format).

To get both playback and capture devices you have 2 (maybe 3) options:
- create 2 dai links, one for playback using spdif-transmitter codec and one for record using spdif-receiver codec
- create a single dai link with 2 codecs - not 100% sure if that'll work though (never tried it with separate playback/capture codecs)
- create a single dai link with a new generic codec driver that has both playback+capture. Basically your current codec, but with all formats and rates as in the spdif drivers and some generic name. This codec driver can then be reused by other machine drivers without needing to create yet another codec.

PR #3090, opened 2019-07-19
FKMS overscan support.

6by9:
5 patches backported from 5.1, and then implementing margins for FKMS.

You can test with either modetest (from https://gitlab.freedesktop.org/mesa/drm/tree/master/tests/modetest), with 
```modetest -M vc4 -w 51:"top margin":50```
replace "top" with "bottom", "left" or "right" as appropriate.
51 is the DRM connector ID, and can be determined with
```
pi@raspberrypi:~ $ drm/build/tests/modetest/modetest -M vc4 -c
Connectors:
id	encoder	status		name		size (mm)	modes	encoders
51	50	connected	HDMI-A-1       	610x350		43	50
  modes:
	name refresh (Hz) hdisp hss hse htot vdisp vss vse vtot)
  3840x2160 30 3840 4016 4104 4400 2160 2168 2178 2250 297000 flags: phsync, pvsync; type: driver
```

or from xrandr
```
xrandr --output HDMI-1 --set "top margin" 50
```
NB because the primary (background) plane isn't redrawn at this point, your mouse cursor won't now align with things. Change resolution using lxrandr or similar, or otherwise force a full screen redraw, and your new settings should take effect.

### DIFF
--- a/drivers/gpu/drm/drm_connector.c
+++ b/drivers/gpu/drm/drm_connector.c
@@ -1110,6 +1110,70 @@ void drm_hdmi_avi_infoframe_content_type(struct hdmi_avi_infoframe *frame,
 EXPORT_SYMBOL(drm_hdmi_avi_infoframe_content_type);
 
 /**
+ * drm_mode_attach_tv_margin_properties - attach TV connector margin properties
+ * @connector: DRM connector
+ *
+ * Called by a driver when it needs to attach TV margin props to a connector.
+ * Typically used on SDTV and HDMI connectors.
+ */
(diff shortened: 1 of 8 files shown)


6by9:
Odd sized cursor after screen saver resolved.
This patch set moved where the mailbox parameters are set up from vc4_plane_atomic_update to vc4_plane_atomic_check as otherwise we miss some of the plane state required for overscan. This is then in line with the full KMS driver too.
Within check we don't necessarily have a mode set, therefore the code that was trying to clip the cursor plane within the bounds of the mode went nuts.

Dispmanx does that job for us anyway, so ditching it gives the right results.

Branch/PR updated.

PR #3097, opened 2019-07-23
Improvements for bcm2835-codec

wens:
This is a three part improvement series.

The first part switches the driver to use the mplane mem2mem API.
This newer API is used exclusively by some applications. For applications
only supporting the old splane API, the userspace library libv4l2 has a
plugin that can convert splane requests into mplane ones.

Unfortunately it is hard to support both APIs at the same time, due to
having different buffer types, and that screws up the videobuf2 helpers.

The second part is implementing theV4L2_CID_MIN_BUFFERS_FOR_CAPTURE
parameter. Some applications require this (and it is said to be required in the
WiP stateful decoder specification). This just outputs the number given by MMAL.

The third part is just a cleanup of how device_caps is set. Changes match mainline.

With all these, I am able to use v4l2 decoding in a Chromium based application on the
Raspberry Pi 4. Unfortunately the application itself is proprietary, and is based on an
older code-base. But I will try to port the required changes onto Debian buster's
chromium package.

It works most of the time, but occasionally it borks and Chromium falls back to software
decoding. One failure type is the decoder not correctly reporting the image dimension,
then Chromium passes in a buffer that is too small. Why a resolution change event was
not triggered I have not dug in to.

### DIFF
--- a/drivers/staging/vc04_services/bcm2835-codec/bcm2835-v4l2-codec.c
+++ b/drivers/staging/vc04_services/bcm2835-codec/bcm2835-v4l2-codec.c
@@ -2357,6 +2357,18 @@ static int bcm2835_codec_open(struct file *file)
 		}
 		ctx->fh.ctrl_handler = hdl;
 		v4l2_ctrl_handler_setup(hdl);
+	} else if (dev->role == DECODE) {
+		v4l2_ctrl_handler_init(hdl, 1);
+
+		v4l2_ctrl_new_std(hdl, &bcm2835_codec_ctrl_ops,
+				  V4L2_CID_MIN_BUFFERS_FOR_CAPTURE,
+				  1, 1, 1, 1);
+		if (hdl->error) {
+			rc = hdl->error;
+			goto free_ctrl_handler;
+		}
+		ctx->fh.ctrl_handler = hdl;
+		v4l2_ctrl_handler_setup(hdl);
 	}
 
 	ctx->fh.m2m_ctx = v4l2_m2m_ctx_init(dev->m2m_dev, ctx, &queue_init);
--- a/drivers/staging/vc04_services/bcm2835-codec/bcm2835-v4l2-codec.c
+++ b/drivers/staging/vc04_services/bcm2835-codec/bcm2835-v4l2-codec.c
@@ -945,8 +945,6 @@ static int vidioc_querycap(struct file *file, void *priv,
 	strncpy(cap->card, MEM2MEM_NAME, sizeof(cap->card) - 1);
 	snprintf(cap->bus_info, sizeof(cap->bus_info), "platform:%s",
 		 MEM2MEM_NAME);
-	cap->device_caps = V4L2_CAP_VIDEO_M2M_MPLANE | V4L2_CAP_STREAMING;
-	cap->capabilities = cap->device_caps | V4L2_CAP_DEVICE_CAPS;
 	return 0;
 }
 
@@ -2600,6 +2598,7 @@ static int bcm2835_codec_create(struct platform_device *pdev,
 	vfd = &dev->vfd;
 	vfd->lock = &dev->dev_mutex;
 	vfd->v4l2_dev = &dev->v4l2_dev;
+	vfd->device_caps = V4L2_CAP_VIDEO_M2M_MPLANE | V4L2_CAP_STREAMING;
 
 	switch (role) {
 	case DECODE:
--- a/drivers/staging/vc04_services/bcm2835-codec/bcm2835-v4l2-codec.c
+++ b/drivers/staging/vc04_services/bcm2835-codec/bcm2835-v4l2-codec.c
@@ -504,7 +504,7 @@ static struct bcm2835_codec_fmt *find_format(struct v4l2_format *f,
 
 	for (k = 0; k < fmts->num_entries; k++) {
 		fmt = &fmts->list[k];
-		if (fmt->fourcc == f->fmt.pix.pixelformat)
+		if (fmt->fourcc == f->fmt.pix_mp.pixelformat)
 			break;
 	}
 	if (k == fmts->num_entries)
@@ -522,9 +522,9 @@ static struct bcm2835_codec_q_data *get_q_data(struct bcm2835_codec_ctx *ctx,
 					       enum v4l2_buf_type type)
 {
 	switch (type) {
-	case V4L2_BUF_TYPE_VIDEO_OUTPUT:
+	case V4L2_BUF_TYPE_VIDEO_OUTPUT_MPLANE:
 		return &ctx->q_data[V4L2_M2M_SRC];
-	case V4L2_BUF_TYPE_VIDEO_CAPTURE:
+	case V4L2_BUF_TYPE_VIDEO_CAPTURE_MPLANE:
 		return &ctx->q_data[V4L2_M2M_DST];
 	default:
 		v4l2_err(&ctx->dev->v4l2_dev, "%s: Invalid queue type %u\n",
@@ -541,9 +541,9 @@ static struct vchiq_mmal_port *get_port_data(struct bcm2835_codec_ctx *ctx,
 		return NULL;
 
 	switch (type) {
-	case V4L2_BUF_TYPE_VIDEO_OUTPUT:
+	case V4L2_BUF_TYPE_VIDEO_OUTPUT_MPLANE:
 		return &ctx->component->input[0];
-	case V4L2_BUF_TYPE_VIDEO_CAPTURE:
+	case V4L2_BUF_TYPE_VIDEO_CAPTURE_MPLANE:
 		return &ctx->component->output[0];
 	default:
 		v4l2_err(&ctx->dev->v4l2_dev, "%s: Invalid queue type %u\n",
@@ -752,7 +752,7 @@ static void handle_fmt_changed(struct bcm2835_codec_ctx *ctx,
 		 format->es.video.crop.width, format->es.video.crop.height,
 		 format->es.video.color_space);
 
-	q_data = get_q_data(ctx, V4L2_BUF_TYPE_VIDEO_CAPTURE);
+	q_data = get_q_data(ctx, V4L2_BUF_TYPE_VIDEO_CAPTURE_MPLANE);
 	q_data->crop_width = format->es.video.crop.width;
 	q_data->crop_height = format->es.video.crop.height;
 	q_data->bytesperline = format->es.video.crop.width;
@@ -945,7 +945,7 @@ static int vidioc_querycap(struct file *file, void *priv,
 	strncpy(cap->card, MEM2MEM_NAME, sizeof(cap->card) - 1);
 	snprintf(cap->bus_info, sizeof(cap->bus_info), "platform:%s",
 		 MEM2MEM_NAME);
-	cap->device_caps = V4L2_CAP_VIDEO_M2M | V4L2_CAP_STREAMING;
+	cap->device_caps = V4L2_CAP_VIDEO_M2M_MPLANE | V4L2_CAP_STREAMING;
 	cap->capabilities = cap->device_caps | V4L2_CAP_DEVICE_CAPS;
 	return 0;
 }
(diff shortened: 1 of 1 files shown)


6by9 (commented):
The link in your commit text is invalid, but otherwise the change looks OK.

wens:
Looks like the patches in the series were merged, hence no longer visible in patchwork with the default settings. I'll replace the link with one pointing to the mailing list.

drivers/staging/vc04_services/bcm2835-codec/bcm2835-v4l2-codec.c
```diff
@@ -2351,13 +2375,22 @@ static int bcm2835_codec_open(struct file *file)
 					 BIT(V4L2_MPEG_VIDEO_H264_PROFILE_MAIN) |
 					 BIT(V4L2_MPEG_VIDEO_H264_PROFILE_HIGH)),
 					V4L2_MPEG_VIDEO_H264_PROFILE_HIGH);
-		if (hdl->error) {
-			rc = hdl->error;
-			goto free_ctrl_handler;
-		}
-		ctx->fh.ctrl_handler = hdl;
-		v4l2_ctrl_handler_setup(hdl);
+	} else {
```
6by9: Needs to be 
```
else if (dev->role == DECODE) {
```
otherwise you are adding this control for the ISP role as well.

pelwell:
If they've been merged already then we'd prefer to cherry-pick them directly from upstream.

6by9 (changes requested):
bcm2835-codec is a little different from other decoder implementations in that there is an intermediate format conversion between the hardware and V4L2 buffers. The number of capture buffers required is therefore independent of the stream and DPB etc.

There are plans to remove the conversion, but it requires a fair amount of rework within the firmware. Until that is done, you may as well return a static value of 1.

https://hverkuil.home.xs4all.nl/codec-api/uapi/v4l/dev-decoder.html is Hans hosting the latest copy of the decoder spec - no mention of ```V4L2_CID_MIN_BUFFERS_FOR_CAPTURE``` being mandatory.

drivers/staging/vc04_services/bcm2835-codec/bcm2835-v4l2-codec.c
```diff
@@ -1738,15 +1745,15 @@ static int vidioc_encoder_cmd(struct file *file, void *priv,
 static const struct v4l2_ioctl_ops bcm2835_codec_ioctl_ops = {
 	.vidioc_querycap	= vidioc_querycap,
 
-	.vidioc_enum_fmt_vid_cap = vidioc_enum_fmt_vid_cap,
```
6by9, line 1741: Whitespace changes unrelated to the main patch.

wens:
> 
> 
> If they've been merged already then we'd prefer to cherry-pick them directly from upstream.

I meant the patches referenced in the link in the commit log.
They are not required for this pull request.

pelwell:
As long as there is no overlap then that's fine.

6by9 (commented):
> If they've been merged already then we'd prefer to cherry-pick them directly from upstream.

It's a reference to an equivalent change set. The bcm2835-camera equivalent has been merged as b0b48b487de64525b08d03de456333939d6e912d

drivers/staging/vc04_services/bcm2835-codec/bcm2835-v4l2-codec.c
wens: > bcm2835-codec is a little different from other decoder implementations in that there is an intermediate format conversion between the hardware and V4L2 buffers. The number of capture buffers required is therefore independent of the stream and DPB etc.
> 
> There are plans to remove the conversion, but it requires a fair amount of rework within the firmware. Until that is done, you may as well return a static value of 1.

I see. May I quote the above when I rework the commit log, or should I just drop the commit for now?

> https://hverkuil.home.xs4all.nl/codec-api/uapi/v4l/dev-decoder.html is Hans hosting the latest copy of the decoder spec - no mention of `V4L2_CID_MIN_BUFFERS_FOR_CAPTURE` being mandatory.

I must have misread something. I'll drop that part.

6by9 (commented):
Is it really valid to only advertise single planar formats through the multiplanar API?

https://hverkuil.home.xs4all.nl/codec-api/uapi/v4l/pixfmt-v4l2-mplane.html#c.v4l2_pix_format_mplane 
```
__u32 | pixelformat | The pixel format. Both single- and multi-planar four character codes can be used.
```
would imply potentially so.
What I do know for definite is that we can't support the genuine multiplanar formats easily as the lower levels want all planes in a single contiguous buffer, but the multi planar allocator has no options for allocating a single buffer with offsets.

I had looked at the client changes required for multi-planar within Chromium, and they weren't significant.

drivers/staging/vc04_services/bcm2835-codec/bcm2835-v4l2-codec.c
6by9: Doh, me not noticing the addition of the ```_mplane``` to the function name at the same time as the whitespace changes. It's valid.

6by9:
I'm intrguiged as to how you have got the V4L2 codec code in Chromium to actually work, although it may well be down to your environment. 
When I looked, it was all under the huge Ozone switch but the X11 rendering paths in Ozone were fairly impressively broken. Is this using one of the other rendering backends, more akin to how ChromeOS sets things up?

Other than the minor comments, the changes as a whole look reasonable, but they do need to be checked against other clients (mainly GStreamer and FFmpeg).

wens:
AFAIK one specifies different fourcc codes to use the true multi-planar formats, i.e. YUV420M vs YUV420.
So supporting the mplane API without supporting the actual multi-planar formats should be OK.

I asked on the mailing list [1] but so far it was only me and a cedrus fellow talking about it.
[1] https://www.spinics.net/lists/linux-media/msg154053.html

drivers/staging/vc04_services/bcm2835-codec/bcm2835-v4l2-codec.c
6by9: > I see. May I quote the above when I rework the commit log, or should I just drop the commit for now?

Sure.

I'm happy for it to be implemented as a control if it makes more clients magically work.
Checking it in https://hverkuil.home.xs4all.nl/codec-api/uapi/v4l/control.html
> This is a read-only control that can be read by the application and used as a hint to determine the number of CAPTURE buffers to pass to REQBUFS. The value is the minimum number of CAPTURE buffers that is necessary for hardware to work.

So the client is only meant to treat it as a hint, therefore failing should the query fail seems like an error in the client.

wens:
For Chromium, in `GenericV4L2Device::CreateGLImage` I was just bypassing Ozone and calling `base::MakeRefCounted<gfx::NativePixmapDmaBuf>` directly, which is essentially what Ozone does on Linux.

Another part is in `V4L2VideoDecodeAccelerator::ImportBufferForPicture`, which should apply to Linux in general and not just when Ozone is enabled. And since V4L2 is Linux-only, I just removed the check.

As for the X11 rendering path being broken, it only supports EGL, not GLX. I ran Chromium with `--use-gl=egl` so it worked, but I confess this part will need some work.

As for other clients, FFmpeg doesn't work with V4L2 on RPi in general. It complains about pts being zero or something. Unrelated to the changes I made. Gstreamer works before and after my changes.

6by9:
Thanks for the list link. I've half dropped off the mailing list as something in the email at one end or other seems to blackhole me :-(
From Paul's response it sounds like cedrus has a very similar requirement to our own.
The single planar APIs are also every so slightly more memory efficient. When allocating independent planes each plane gets aligned to a page boundary. Generally that is nothing compared to the several MBs for the image data. 

Forcing the use of libv4l is horrid for those who want the single planar API, but may be the way to go.

I'm slightly conflicted over having the bcm2835-v4l2 and bcm2835-unicam drivers both using the single planar API whilst bcm2835-codec goes multi-planar, but switching the other two is more likely to create issues for existing apps, and the libv4l solution isn't automatic.
bcm2835-codec is new enough that I'm happy to make the change as there are unlikely to be that many people dependent on it.
That thought process isn't enough to block this PR, but is a niggle in the back of my mind. If GStreamer and FFmpeg behave, then that will probably cover 99% of users.

> As for other clients, FFmpeg doesn't work with V4L2 on RPi in general. It complains about pts being zero or something. 

Raspbian should have a couple of patches to FFmpeg to make it work. I'll have a look.

wens:
> Thanks for the list link. I've half dropped off the mailing list as something in the email at one end or other seems to blackhole me :-(
> From Paul's response it sounds like cedrus has a very similar requirement to our own.

AFAIK the hardware requires the different planes to be close to each other. Having a contiguous buffer is the easiest way to fulfill that requirement.

> The single planar APIs are also every so slightly more memory efficient. When allocating independent planes each plane gets aligned to a page boundary. Generally that is nothing compared to the several MBs for the image data.

Right, but that only happens if you use multi-planar formats? For the single-planar, a.k.a. contiguous buffer, formats, there's only one plane.

> Forcing the use of libv4l is horrid for those who want the single planar API, but may be the way to go.

What would applications do instead? Direct ioctl calls? That is actually what Chromium does for the decoder. It uses libv4l2, if specified at build time, for the encoder.

> I'm slightly conflicted over having the bcm2835-v4l2 and bcm2835-unicam drivers both using the single planar API whilst bcm2835-codec goes multi-planar, but switching the other two is more likely to create issues for existing apps, and the libv4l solution isn't automatic.
> bcm2835-codec is new enough that I'm happy to make the change as there are unlikely to be that many people dependent on it.

I figure it's easier to have the driver support the newer API, and have libv4l2 for backward-compatibility, than have applications all try to support both APIs. But I suppose the opposite could be argued as well, since maybe only half the drivers in the kernel use mplane.

Either way my main goal was to use V4L2 in Chromium rather than the MMAL library.

> That thought process isn't enough to block this PR, but is a niggle in the back of my mind. If GStreamer and FFmpeg behave, then that will probably cover 99% of users.
> 
> > As for other clients, FFmpeg doesn't work with V4L2 on RPi in general. It complains about pts being zero or something.
> 
> Raspbian should have a couple of patches to FFmpeg to make it work. I'll have a look.

Curious. I was using the newest Raspbian release already.

wens:
I've rebased on top of the latest rpi-4.19.y branch, and updated the patches as discussed.

I realized that for V4L2_CID_MIN_BUFFERS_FOR_CAPTURE, my original idea was to implement the control for all roles. But I just simplified it and limited it to the decoder for now.

drivers/staging/vc04_services/bcm2835-codec/bcm2835-v4l2-codec.c
```diff
@@ -2357,6 +2358,18 @@ static int bcm2835_codec_open(struct file *file)
 		}
 		ctx->fh.ctrl_handler = hdl;
 		v4l2_ctrl_handler_setup(hdl);
+	} else if (dev->role == DECODE) {
+		v4l2_ctrl_handler_init(hdl, 1);
+
+		ctrl = v4l2_ctrl_new_std(hdl, &bcm2835_codec_ctrl_ops,
```
6by9: Apologies if I missed this first time round, but ctrl is never used after this, so no need to define the new variable and assign to it.

6by9:
Sorry, I still haven't had a chance to test this for myself against GStreamer and FFmpeg. I'll try to do so tomorrow, but can't guarantee it.

wens:
My bad. Left over from making V4L2_CID_MIN_BUFFERS_FOR_CAPTURE always return 1.

drivers/staging/vc04_services/bcm2835-codec/bcm2835-v4l2-codec.c
wens: My bad. Left over from making V4L2_CID_MIN_BUFFERS_FOR_CAPTURE always return 1.

wens:
As for FFmpeg not working, someone else had reported the same issue on a different platform [1].

> 14:10 pH5: Chewi: looks to me like the ffmpeg v4l2m2m code doesn't fill the chroma linesizes correctly for single planar formats
> 14:10 pH5: Chewi: I think it is this code: https://github.com/FFmpeg/FFmpeg/blob/master/libavcodec/v4l2_buffers.c#L302
> 14:12 pH5: Chewi: for non-MPLANE queues only plane_info[0].bytesperline is valid, and the chroma strides are half of that. 

[1] https://dri.freedesktop.org/~cbrill/dri-log/?channel=etnaviv&highlight_names=&date=2017-10-30

6by9:
I have patches (which I thought had been merged in Raspbian) that patch https://github.com/FFmpeg/FFmpeg/blob/master/libavcodec/v4l2_buffers.c#L323 to also support NV21 and YUV420P. YV12 isn't a defined format in FFmpeg IIRC.
https://www.raspberrypi.org/forums/viewtopic.php?t=234778#p1436166

NV12 is the standard format used on Samsung SoCs IIRC, so whoever wrote the original support appears to have done the minimum possible, although plausibly because they took the view that it's better to have no code than untested code. An assert for any of the unhandled multiple plane but single planar V4L2  formats would have been nice rather than just falling in a heap.

6by9:
FYI My patch to support the single planar V4L2 API in Chromium:
[0001-v4l2-Initial-support-for-the-single-planar-API.zip](https://github.com/raspberrypi/linux/files/3457876/0001-v4l2-Initial-support-for-the-single-planar-API.zip)
Whilst it built, I'm not sure I ever got Chromium in a state that would actively use it - the Ozone X11 crashes were a major headache which is why we largely dropped it.

Using libv4l adds an unknown into the mix. Should applications be using LD_PRELOAD to insert it or not? Under what conditions? What happens if you needed it and didn't preload it? And I believe there are also issues with it and dmabufs (which I thought Chromium used as it was one of the reasons I was investigating it)

Sorry, I haven't had a chance to do any testing on your patches today. I will make it a priority tomorrow.

6by9 (approved):
I think I'm happy with these.
I've tested with FFmpeg and it works.
I still have slight reservations over going multi-planar, but it'd be an easy one to revert if it really messes things up.

kilograham:
I am independently working on fixing up ffmpeg to fix the pts issues... i can play with h264_v4l2m2m on Pi 4 but something is not quite right along the way as h264 is smoother in software. Not yet sure whether the issues are timestamp related or not (ffplay as an example drops frames without the correct timestamps)

pelwell:
No objections from me, but I'm hardly a qualified observer.

PR #3100, opened 2019-07-23
Add HDMI1 facility to the audio driver.

JamesH65:
For generic ALSA, all you need is the bcm2835.h change, but
have also added structures for IEC958 HDMI. Not sure how to
test those.

Requires VC4 firmware to support the extra device but should be
backward compatible with old firmware.

### DIFF
--- a/drivers/staging/vc04_services/bcm2835-audio/bcm2835.c
+++ b/drivers/staging/vc04_services/bcm2835-audio/bcm2835.c
@@ -79,7 +79,11 @@ static int bcm2835_audio_alsa_newpcm(struct bcm2835_chip *chip,
 	if (err)
 		return err;
 
-	err = snd_bcm2835_new_pcm(chip, "bcm2835 IEC958/HDMI", 1, 0, 1, true);
+	err = snd_bcm2835_new_pcm(chip, "bcm2835 IEC958/HDMI", 1, AUDIO_DEST_HDMI0, 1, true);
+	if (err)
+		return err;
+
+	err = snd_bcm2835_new_pcm(chip, "bcm2835 IEC958/HDMI1", 2, AUDIO_DEST_HDMI1, 1, true);
 	if (err)
 		return err;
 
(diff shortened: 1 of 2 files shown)


JamesH65:
Would be good to get some eyes on this...

drivers/staging/vc04_services/bcm2835-audio/bcm2835.c
```diff
@@ -79,7 +79,11 @@ static int bcm2835_audio_alsa_newpcm(struct bcm2835_chip *chip,
 	if (err)
 		return err;
 
-	err = snd_bcm2835_new_pcm(chip, "bcm2835 IEC958/HDMI", 1, 0, 1, true);
+	err = snd_bcm2835_new_pcm(chip, "bcm2835 IEC958/HDMI0", 1, AUDIO_DEST_HDMI0, 1, true);
```
pelwell: Are there any consequences of this name change, e.g. in ALSA configuration files, the LXPanel volume plugin etc?

drivers/staging/vc04_services/bcm2835-audio/bcm2835.c
```diff
@@ -106,17 +110,30 @@ static struct bcm2835_audio_driver bcm2835_audio_alsa = {
...
 
-static struct bcm2835_audio_driver bcm2835_audio_hdmi = {
+static struct bcm2835_audio_driver bcm2835_audio_hdmi0 = {
 	.driver = {
 		.name = "bcm2835_hdmi",
 		.owner = THIS_MODULE,
 	},
-	.shortname = "bcm2835 HDMI",
-	.longname  = "bcm2835 HDMI",
+	.shortname = "bcm2835 HDMI 0",
```
pelwell: Ditto this change.

drivers/staging/vc04_services/bcm2835-audio/bcm2835.c
JamesH65: Possibly, but I am not sure where or when those particular items are used - not by us AFAIK.

drivers/staging/vc04_services/bcm2835-audio/bcm2835.c
JamesH65: Should I simply rename this back, and assume its zero, leaving 1 as is?

pelwell: That sounds reasonable.

JamesH65: Changed back

PR #3103, opened 2019-07-24
FKMS "Broadcast RGB" property, and KMS command line parsing update

6by9:
"Broadcast RGB" is how Intel have added an override for the full/limited range option for HDMI.

Backports DRM command line patches from drm-misc-next, so should be in 5.3. Allows specifying of margins and rotation from the kernel command line.
eg ```video=HDMI-A-1:1920x1080@60,margin_right=50,rotate=90```

Required a modest amount of reworking as functions have been moved between files.

reflect_x and reflect_y appear not to get actioned. That's not a use case I'm overly concerned about, but I think it's core DRM plumbing missing.

Also enables CONFIG_FRAMEBUFFER_CONSOLE_ROTATION so that the console can be rotated (in software) under DRM/KMS.


6by9:
@pelwell @popcornmix Any chance of a review?

drivers/gpu/drm/vc4/vc4_firmware_kms.c
```diff
@@ -1337,13 +1373,95 @@ static void vc4_fkms_connector_destroy(struct drm_connector *connector)
...
+struct drm_connector_state *
+vc4_connector_duplicate_state(struct drm_connector *connector)
+{
+	struct vc4_fkms_connector_state *state;
+
+	state = kmemdup(connector->state, sizeof(*state), GFP_KERNEL);
+	if (!state)
+		return NULL;
+
+	__drm_atomic_helper_connector_duplicate_state(connector, &state->base);
```
pelwell, line 1394: Either I've missed something or the state gets copied twice.

drivers/gpu/drm/vc4/vc4_firmware_kms.c
```diff
@@ -1370,16 +1516,28 @@ vc4_fkms_connector_init(struct drm_device *dev, struct drm_encoder *encoder,
 	fkms_connector = devm_kzalloc(dev->dev, sizeof(*fkms_connector),
 				      GFP_KERNEL);
 	if (!fkms_connector) {
-		ret = -ENOMEM;
-		goto fail;
+		return ERR_PTR(-ENOMEM);
+	}
```
pelwell, line 1527: Unnecessary braces, although checkpatch seems unconcerned.

pelwell (commented):
One substantive issue/query and a stylistic niggle.

drivers/gpu/drm/vc4/vc4_firmware_kms.c
6by9: kmemdup allocates and copies ```sizeof(struct vc4_fkms_connector_state)```.
```__drm_atomic_helper_connector_duplicate_state``` copies ```sizeof(drm_connector_state)```, which is smaller.

This was reproducing the same functionality as in the Intel i915 driver, and they also copy the base state twice.
https://elixir.bootlin.com/linux/v5.2.6/source/drivers/gpu/drm/i915/intel_atomic.c#L150

To clean it up we'll end up with a funky offsetof(broadcast_rgb) for the generic solution. At the moment it'll only be copying one int. I might just assign the int across with a comment should extra members get added.

drivers/gpu/drm/vc4/vc4_firmware_kms.c
6by9: Checkpatch not being fussy for once. I'm glad I was sitting down!

PR #3104, opened 2019-07-24
[RFC] dts: bcm2838: add missing properties for pmu and gic nodes

P33M:
The GIC has a virtual interface maintenance interrupt and the PMU
interrupts need affinity mappings as they are wired to generic SPIs.

Also, delete incorrect PMU compatible string.

### DIFF
--- a/arch/arm/boot/dts/bcm2838.dtsi
+++ b/arch/arm/boot/dts/bcm2838.dtsi
@@ -35,6 +35,8 @@
 				<0x40042000 0x2000>,
 				<0x40044000 0x2000>,
 				<0x40046000 0x2000>;
+			interrupts = <GIC_PPI 9 (GIC_CPU_MASK_SIMPLE(4) |
+						 IRQ_TYPE_LEVEL_HIGH)>;
 		};
 
 		thermal: thermal@7d5d2200 {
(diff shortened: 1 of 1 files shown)


arch/arm/boot/dts/bcm2838.dtsi
```diff
@@ -35,6 +35,7 @@
 				<0x40042000 0x2000>,
 				<0x40044000 0x2000>,
 				<0x40046000 0x2000>;
+			interrupts = <1 9 0xf04>;
```
pelwell: Wouldn't that be better as `<GIC_PPI 9 (GIC_CPU_MASK_SIMPLE(4) | IRQ_TYPE_LEVEL_HIGH)>`?

arch/arm/boot/dts/bcm2838.dtsi
```diff
@@ -222,15 +223,12 @@
 	};
 
 	arm-pmu {
-		/*
-		 * N.B. the A72 PMU support only exists in arch/arm64, hence
-		 * the fallback to the A53 version.
-		 */
-		compatible = "arm,cortex-a72-pmu", "arm,cortex-a53-pmu";
+		compatible = "arm,cortex-a72-pmu";
```
pelwell, line 227: Marc said:
> You're thus guaranteed to count the wrong events... And you could at
least have "arm,armv8-pmuv3".

Please explain the effect of your change (or why we've ended up with this) in a comment.

P33M: There's no such thing as fall-back from a72 to a53 - the A53 has a unique event type of ARMV8_A53_PERFCTR_PREF_LINEFILL, and all other event types supported by A53 are a subset of events supported by A72.

The the arm64/kernel/perf_event.c driver handles various pmuv3 implementations including the A72 variant. Matching on the generic "arm,armv8-pmuv3" string does not enable optional implementation-dependent event types, of which the A72 has several that are likely to be useful.

Also, there are dts files in the kernel tree that use both the a53 and generic compatible strings, but there are no dts files with more than just the a72 compatible string.

arch/arm/boot/dts/bcm2838.dtsi
P33M: The kernel doc apparently hasn't been updated with helper macros:
Documentation/devicetree/bindings/interrupt-controller/arm,gic.txt

Will add.

PR #3110, opened 2019-07-29
Fixup FKMS interrupt handing for non-existent display

JamesH65:
If an errant interrupt flag was received from a non-existent display,
a NULL pointer access was made. Protect against this by checking if a
second display is present prior to checking the interrupt flags.

### DIFF
--- a/drivers/gpu/drm/vc4/vc4_firmware_kms.c
+++ b/drivers/gpu/drm/vc4/vc4_firmware_kms.c
@@ -1056,14 +1056,17 @@ static irqreturn_t vc4_crtc_irq_handler(int irq, void *data)
 				vc4_crtc_handle_page_flip(crtc_list[0]);
 			}
 
-			/* Check for the secondary display too */
-			chan = readl(crtc_list[0]->regs + SMIDSW1);
+			if (crtc_list[1]) {
+				/* Check for the secondary display too */
+				chan = readl(crtc_list[0]->regs + SMIDSW1);
 
-			if (chan & 1) {
-				writel(SMI_NEW, crtc_list[0]->regs + SMIDSW1);
-				if (crtc_list[1]->vblank_enabled)
-					drm_crtc_handle_vblank(&crtc_list[1]->base);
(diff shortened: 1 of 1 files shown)


6by9:
LGTM

6by9 (approved):

drivers/gpu/drm/vc4/vc4_firmware_kms.c
```diff
@@ -1056,14 +1056,17 @@ static irqreturn_t vc4_crtc_irq_handler(int irq, void *data)
...
+				/* Check for the secondary display too */
+				chan = readl(crtc_list[0]->regs + SMIDSW1);
 
-			if (chan & 1) {
-				writel(SMI_NEW, crtc_list[0]->regs + SMIDSW1);
-				if (crtc_list[1]->vblank_enabled)
-					drm_crtc_handle_vblank(&crtc_list[1]->base);
-				vc4_crtc_handle_page_flip(crtc_list[1]);
+				if (chan & 1) {
+					writel(SMI_NEW, crtc_list[0]->regs + SMIDSW1);
```
popcornmix, line 1064: I assume this clears the interrupt? If we skip this then we'll continually spin in the ISR.
Would it be safer to always clear but just skip the vblank handling for missing display?

pelwell: In the skip case (at least) you might want to read back SMIDSW1 to guarantee the clear has gone through before leaving the interrupt handler.

6by9: The interrupt is cleared at line 1038 by writing to SMICS.
SMIDSW0 and SMIDSW1 are only used as status registers to report what has changed (IIRC they're actually write pointers in the peripheral). Having a value in them won't trigger an interrupt by itself.

pelwell: In which case my comment doesn't apply.

PR #3111, opened 2019-07-29
Correct the name of the Raspberry Pi video decoder

pelwell:
Give a the Raspberry Pi-developed video decoder block in BCM2711
a more self-explanatory name.


JamesH65:
Should the commit also mention the device minor node change? It appears to be the one functional change that might have an impact elsewhere.

pelwell:
You mean that the old name is now temporarily assigned to a new minor number, prior to removal? I'd be very surprised if any code outside the kernel knows or cares about the device numbers, major or minor.

P33M:
The existing ffmpeg "decoder" .so references the device nodes by name, so as long as there's a legacy name match it should still work. I'm unaware of any other software making use of this interface at this time.

ping @popcornmix - any other decoders that we know of?

pelwell:
LibreELEC/OpenELEC uses the names: https://github.com/CoreELEC/CoreELEC/commit/84e70210df

P33M (approved):

popcornmix:
I imagine everything will use names rather than numbers (it will all be based on the ffmpeg code).

VLC will need updating to new names but that hasn't been released so we just need to ping John.

popcornmix:
Note: LibreELEC will need to update their kernel config (they don't use ours).

popcornmix:
Note: the raspbian udev rule 
```
SUBSYSTEM=="argon-*", GROUP="video", MODE="0660"
```
doesn't work with this update.
```
pi@pi4:~ $ ls -l /dev/argon-*
crw------- 1 root root 239, 1 Jul 30 14:02 /dev/argon-h264mem
crw------- 1 root root 241, 1 Jul 30 14:02 /dev/argon-hevcmem
crw------- 1 root root 240, 1 Jul 30 14:02 /dev/argon-intcmem
crw------- 1 root root 238, 1 Jul 30 14:02 /dev/argon-vp9mem
```
Editing it to ``rpivid-*`` does work (and changes both rpivid and the argon aliases).

Perhaps we want to update raspbian to include (ping @XECDesign):
```
SUBSYSTEM=="argon-*", GROUP="video", MODE="0660"
SUBSYSTEM=="rpivid-*", GROUP="video", MODE="0660"
```
before it upates the kernel.

Also - Arch will need a similar update (kodi packages for Arch have recently been made that include udev rules)

XECDesign:
Updated udev rule should show up in the repo shortly

PR #3112, opened 2019-07-29
Backport BCM270X_DT: Rename Pi Zero W DT files

fgiancane8:
This patch needed to be backported to series 4.14.y because it is also supported in meta-raspberrypi for Yocto Project.
Unfortunately, the rename affected the correct building of the 4.14 series kernel, which now results broken.
Applying the same renames on the 4.14.y series would fix the issue for those still using that kernel version.

The downtream Pi Zero W dts file uses the digit 0, whereas upstream
chose to spell it out - "zero-w". The firmware has, for a long time,
looked for bcm2708-rpi-zero-w.dtb first before falling back to the
numerical version. Therefore it is better to follow upstream and
make the switch to "bcm2708-rpi-zero-w".

At the same time, remove some overrides that duplicate values
inherited from the shared .dtsi files.

Signed-off-by: Francesco Giancane <francescogiancane8@gmail.com>

### DIFF
--- a/arch/arm/boot/dts/Makefile
+++ b/arch/arm/boot/dts/Makefile
@@ -5,7 +5,7 @@ dtb-$(CONFIG_ARCH_BCM2835) += \
 	bcm2708-rpi-b.dtb \
 	bcm2708-rpi-b-plus.dtb \
 	bcm2708-rpi-cm.dtb \
-	bcm2708-rpi-0-w.dtb \
+	bcm2708-rpi-zero-w.dtb \
 	bcm2709-rpi-2-b.dtb \
 	bcm2710-rpi-3-b.dtb \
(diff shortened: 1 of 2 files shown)


pelwell:
I feel Yocto ought to fix the problem at their end, so the "needed" in the commit message is open to debate, but I see no harm in accepting this back-port here.

fgiancane8:
@pelwell 
You are right, but I wrote "needed" considering we are using the linux fork from Raspberry Pi and not the mainline one. If you look at device trees sources, they have completely different names and configurations.

In fact, this fix is local to raspberry pi -specific kernel rather than mainline.

fgiancane8:
@pelwell ,
I also already submitted the patch in meta-raspberrypi repository so that it is ready to be used also with Yocto.

Thanks,
F

PR #3121, opened 2019-08-02
net: bcmgenet: Workaround for Pi 4B network issue

pelwell:
Some combinations of Pi 4Bs and Ethernet switches don't reliably get a
DCHP-assigned IP address, leaving the unit with a self=assigned 169.254
address.

Forcing renegotiation has been found to be an effective workaround, so
add an automatic renegotiation after the link comes up for the first
time.

See: https://github.com/raspberrypi/linux/issues/3108

### DIFF
--- a/drivers/net/ethernet/broadcom/genet/bcmgenet.c
+++ b/drivers/net/ethernet/broadcom/genet/bcmgenet.c
@@ -72,6 +72,10 @@
 #define GENET_RDMA_REG_OFF	(priv->hw_params->rdma_offset + \
 				TOTAL_DESC * DMA_DESC_SIZE)
 
+static bool force_reneg = false;
+module_param(force_reneg, bool, 0444);
+MODULE_PARM_DESC(force_reneg, "Force a renegotiation after the initial link-up");
+
 static inline void bcmgenet_writel(u32 value, void __iomem *offset)
 {
 	/* MIPS chips strapped for BE will automagically configure the
@@ -2610,6 +2614,7 @@ static void bcmgenet_irq_task(struct work_struct *work)
 	unsigned int status;
 	struct bcmgenet_priv *priv = container_of(
 			work, struct bcmgenet_priv, bcmgenet_irq_work);
+	static int first_link = 1;
 
 	netif_dbg(priv, intr, priv->dev, "%s\n", __func__);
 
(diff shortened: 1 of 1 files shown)


pelwell:
With this patch applied, a "bad" Pi 4B connected to a "bad" switch successfully received a DHCP IP address and rebooted all night. The only known downside is that it also delays the point at which a "good" pairing gets an IP address by a few seconds. For users and applications wishing to minimise the boot time, I've added a module parameter (`genet.force_reneg`) to enable the feature. Adding `genet.force_reneg=n` to cmdline.txt disables the re-negotiation at boot.

HiassofT:
Can you please make the workaround opt-in?

On managed switches negotiation can take quite some time - eg on my HP 1810-24G it takes about 5 seconds (some Ciscos are said to take even a lot longer) - and this means not only bootup time but also the chance of running into a timeout will increase.

For example in LibreELEC we use a default timeout of 10 seconds to wait before the network is up. With current 4.19 kernel this works fine and still had a lot of slack, even with my HP switch. But if the time to bring the network up is increased I (and users) will need to increase the default timeout (yes, it can be done from the GUI, but it's not nice).

Here's are some timing outputs from raspbian lite, Raspbian kernel 4.19.58 vs current 4.19.63 plus this PR
```
root@raspberrypi:~# cat systemd-analyze-4.19.58 
Startup finished in 1.630s (kernel) + 16.669s (userspace) = 18.300s 
graphical.target reached after 16.626s in userspace
root@raspberrypi:~# cat systemd-analyze-4.19.63 
Startup finished in 1.661s (kernel) + 21.946s (userspace) = 23.607s 
graphical.target reached after 21.903s in userspace
root@raspberrypi:~# grep eth0 dmesg-4.19.58 
[    6.842622] IPv6: ADDRCONF(NETDEV_UP): eth0: link is not ready
[    7.911671] bcmgenet fd580000.genet eth0: Link is Down
[   12.071813] bcmgenet fd580000.genet eth0: Link is Up - 1Gbps/Full - flow control off
[   12.071859] IPv6: ADDRCONF(NETDEV_CHANGE): eth0: link becomes ready
root@raspberrypi:~# grep eth0 dmesg-4.19.63
[    6.902697] IPv6: ADDRCONF(NETDEV_UP): eth0: link is not ready
[    7.911865] bcmgenet fd580000.genet eth0: Link is Down
[   12.071907] bcmgenet fd580000.genet eth0: Link is Up - 1Gbps/Full - flow control off
[   12.071946] IPv6: ADDRCONF(NETDEV_CHANGE): eth0: link becomes ready
[   12.072155] bcmgenet fd580000.genet eth0: Link is Down
[   17.271850] bcmgenet fd580000.genet eth0: Link is Up - 1Gbps/Full - flow control off
```

pelwell:
It's a question of getting the workaround deployed and tested as quickly as possible. As long as it is opt-in we're unlikely to get much in the way of feedback, but if opt-out causes an increase in problem reports then we can easily change the default.

Any thoughts, @popcornmix?

HiassofT:
I understand that you'd like to fix the issue, but as the (hopefully temporary) workaround causes a regression it should not be on by default.

With opt-in the few affected users have a chance to use their RPis until the core issue has been found and fixed.

pelwell:
Alright - the default is now disabled.

radekjuthner:
Can the same problem occur again on the Raspberry Pi 5?

6by9:
> Can the same problem occur again on the Raspberry Pi 5?

No. The ethernet hardware on Pi5 is totally different (part of RP1 rather than the main SoC).

PR #3123, opened 2019-08-03
Rpi 5.2.y: fix compiling in separate directory (O=...)

elkablo:
Some Makefiles in raspberry pi specific drivers use include paths that are relative to compiled objects, not to the sources. This obviously does not work when compiling in separate directory, with O=path variable.

This pull request only fixes this for the 5.2.y branch which I work on.


pelwell:
Thanks!

lategoodbye:
@elkablo Any plans to send a patch against mainline kernel?

lategoodbye:
Sorry, my fault. None of these changes apply upstream.

PR #3130, opened 2019-08-06
[5.3] ASoC: rpi-cirrus: use modern dai_link style

HiassofT:
This fixes the Cirrus Logic Audio Card build issues due to the dai link changes in kernel 5.3

build and runtime tested on RPi4

### DIFF
--- a/sound/soc/bcm/rpi-cirrus.c
+++ b/sound/soc/bcm/rpi-cirrus.c
@@ -887,30 +887,36 @@ static struct snd_soc_ops rpi_cirrus_ops = {
 	.hw_free = rpi_cirrus_hw_free,
 };
 
+SND_SOC_DAILINK_DEFS(wm5102,
+     DAILINK_COMP_ARRAY(COMP_EMPTY()),
+     DAILINK_COMP_ARRAY(COMP_CODEC("wm5102-codec", "wm5102-aif1")),
+     DAILINK_COMP_ARRAY(COMP_EMPTY()));
+
+SND_SOC_DAILINK_DEFS(wm8804,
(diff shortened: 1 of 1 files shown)


pelwell:
What's the semantic difference between a 3-arg SND_SOC_DAILINK_DEFS instantiation and a 4-arg  SND_SOC_DAILINK_DEFS instantiation where the 4th argument is DAILINK_COMP_ARRAY(COMP_EMPTY()))?

HiassofT:
The 3-arg version sets .platforms to null_dailink_component - see https://github.com/raspberrypi/linux/blob/rpi-5.3.y/include/sound/soc.h#L1060

The second DAI link in the driver doesn't need a platform component (it's needed only at the bcm2835-i2s side, in the first DAI link).

The codec-codec link in the rpi-cirrus driver is a bit of a special case, not sure there are some other RPi cards using that (WM8804 is connected to the second audio interface of WM5102 here).

COMP_EMPTY () is just an empty initializer where the driver can fill in the of node for cpu and platform.

pelwell:
Thanks for the explanation. I'm sure one day there will be documentation for all of this,

HiassofT:
Yeah, documentation is quite lacking (and with the removal of the old style variant the comments in the sound_soc_dai_link struct haven't been updated....). At least we got some examples before the new macro definitions in soc.h :-)

But, not much has changed, multi-codecs existed before and now it's even a bit simpler because the single cpu+platform members have been dropped and now everything's in arrays.

Most puzzling thing for me was understanding the helper macros

pelwell:
Why do you think I asked the question.. ;)

PR #3132, opened 2019-08-07
drm/vc4: Add missing NULL check to vc4_crtc_consume_event

6by9:
vc4_crtc_consume_event wasn't checking crtc->state->event was
set before dereferencing it, leading to an OOPS.

Fixes "a5b534b drm/vc4: Resolve the vblank warnings on mode switching"

### DIFF
--- a/drivers/gpu/drm/vc4/vc4_firmware_kms.c
+++ b/drivers/gpu/drm/vc4/vc4_firmware_kms.c
@@ -969,6 +969,9 @@ static void vc4_crtc_consume_event(struct drm_crtc *crtc)
 	struct drm_device *dev = crtc->dev;
 	unsigned long flags;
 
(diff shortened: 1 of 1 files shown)


popcornmix:
Confirmed to fix issue with LibreELEC booting.

JamesH65:
Does this fix the problem with kernel dumps when the screensaver kicks in?

pelwell:
This is the fix to the fix to the screensaver dumps.

6by9:
https://github.com/raspberrypi/linux/pull/3129/commits/06ffdc71459f5f428ce1511ccc8ee66ccbeebe28 was the fix to the kernel warnings when mode changing.
As Phil says, this is a fix to the fix. These lines got merged into a different patch that is still being worked on, hence it was working for me.

PR #3135, opened 2019-08-07
[5.3] justboom-dac, rpi-wm8804-soundcard: use modern dai_link style

HiassofT:
Runtime tested on RPi4 with Justboom DAC, Justboom Digi and HifiBerry Digi-Pro

That's it from my side, I think I'm through all the I2S cards I have here


pelwell:
Thanks, Hias. That's probably enough examples - I'll try and finish the rest.

PR #3137, opened 2019-08-08
According to 5713 pdf doc CLOCK_CTRL is a readonly status register, a…

hlauer0:
…nd it behaves so. Remove useless setting of the register.


pelwell:
It looks like you may be correct (although the datasheet I've found could be clearer on the subject). Any comments, @Hifiberry?

hifiberry:
Good question. Unfortunately I don't have a test setup here right now at the older Amp+ only works with Pi2 or earlier (this has nothing to do with the driver, just power requirements).

hifiberry:
Had a short look at the data sheet and it really seems to bet read-only. Let's remove this.

pelwell:
Thanks.

PR #3140, opened 2019-08-09
v3d_drv: Use firmware mailbox call for clock requests

popcornmix:
Directly modifying clocks means that firmware DVFS cannot be used to reduce voltage (and hence power) when system is idle or we are temperature throttling.

Request firmware adjusts the clocks to make this possible


lategoodbye:
Just my 2 cents from mainline perspective:

An acceptable solution for mainline would be to write a clk driver for the V3D clock. Otherwise the V3D driver will only be useable with the Raspberry Pi.

Please refer to recent patches by @vianpl as inspiration: 
https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?h=v5.3-rc4&id=2256d89333bd17b8b56b42734a7e1046d52f7fc3
https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?h=v5.3-rc4&id=4e85e535e6cc6e8a96350e8ee684d0f22eb8629e

popcornmix:
@lategoodbye I've had a go at following your advice: https://github.com/popcornmix/linux/commits/cpufreq
But I'm finding any attempt at disabling clk-bcm2835's use of V3D causes kernel boot to hang (no panic or messages in log) e.g. https://github.com/popcornmix/linux/commit/557e7b9ea4fccf0cc7f5e2907f8e9a6088fad795

dmesg: https://pastebin.com/0yANDJVu

Is there something simple I've missed you can spot?

popcornmix:
note: the porting of the upstream cpufreq driver and removal of pllb seems to work on that branch

lategoodbye:
This sounds like a clk issue not a driver issue. Maybe some kind of dependency to another clock?
The hang occurs on RPi 3 or 4?

popcornmix:
Just testing on Pi4.

popcornmix:
@lategoodbye @vianpl can you check the top commit of this:
https://github.com/raspberrypi/linux/compare/rpi-4.19.y...popcornmix:avs2_cpufreq?expand=1

to see if it's more suitable for upstreaming. (Still needs checkpatch type checking).

It tries to follow the style of clk-bcm2835 with the behaviour of clk-raspberrypi.

lategoodbye:
General comments:
- try to keep in mind that there other users than Raspberry Pi of v3d which shouldn't break
- please make sure that your patch series is bisectable (compile and working)

PR #3147, opened 2019-08-13
xhci: Use more event ring segment table entries

P33M:
Users have reported log spam created by "Event Ring Full" xHC event
TRBs. These are caused by interrupt latency in conjunction with a very
busy set of devices on the bus. The errors are benign, but throughput
will suffer as the xHC will pause processing of transfers until the
event ring is drained by the kernel. Expand the number of event TRB slots
available by increasing the number of event ring segments in the ERST.

Controllers have a hardware-defined limit as to the number of ERST
entries they can process, so make the actual number in use
min(ERST_MAX_SEGS, hw_max).

### DIFF
--- a/drivers/usb/host/xhci-mem.c
+++ b/drivers/usb/host/xhci-mem.c
@@ -2491,9 +2491,11 @@ int xhci_mem_init(struct xhci_hcd *xhci, gfp_t flags)
 	 * Event ring setup: Allocate a normal ring, but also setup
 	 * the event ring segment table (ERST).  Section 4.9.3.
 	 */
+	val2 = 1 << HCS_ERST_MAX(xhci->hcs_params2);
+	val2 = min_t(unsigned int, ERST_MAX_SEGS, val2);
 	xhci_dbg_trace(xhci, trace_xhci_dbg_init, "// Allocating event ring");
-	xhci->event_ring = xhci_ring_alloc(xhci, ERST_NUM_SEGS, 1, TYPE_EVENT,
-					0, flags);
+	xhci->event_ring = xhci_ring_alloc(xhci, val2, 1, TYPE_EVENT,
+					   0, flags);
 	if (!xhci->event_ring)
 		goto fail;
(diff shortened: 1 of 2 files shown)


drivers/usb/host/xhci-mem.c
```diff
@@ -2504,13 +2506,13 @@ int xhci_mem_init(struct xhci_hcd *xhci, gfp_t flags)
...
 	/* set ERST count with the number of entries in the segment table */
-	val = readl(&xhci->ir_set->erst_size);
-	val &= ERST_SIZE_MASK;
-	val |= ERST_NUM_SEGS;
+	val2 = readl(&xhci->ir_set->erst_size);
+	val2 &= ERST_SIZE_MASK;
+	val2 |= val;
 	xhci_dbg_trace(xhci, trace_xhci_dbg_init,
 			"// Write ERST size = %i to ir_set 0 (some bits preserved)",
 			val);
```
pelwell, line 2514: The swapping of val and val2 (just to use them in order?) is a bit gratuitous, and it breaks/non-obviously-fixes this debug output.

P33M: Bits 31:16 are RsvdP (reserved, preserve anyway) but this now truncates the register if any of the reserved bits are used. Debug should use val2 - will fix

pelwell: And swap val and val2? Plz?

P33M: k

pelwell: Thx

l1k:
Just a heads-up: I've upstreamed this commit, it is now queued for v6.7-rc1 and can hence be omitted when you update your tree to that kernel version.

https://git.kernel.org/mnyman/xhci/c/26a23e0dd061

Thanks!

popcornmix:
@l1k that's good to hear. I'll drop this commit when the upstream version appears.

Issue #3148, opened 2019-08-13
usb_control_msg bugged?


sukoy:
Hi,

Pi3B+
.firmware_revision `4cf061b4500b87c26c721db4d847593452270cff`
4.19.65-v7+ `f1c1b67b26ed3cb789037d1c844d4105deaa3cfd`

I got some troubles with usb, to better understand start from scratch.

I found a lot of cameras with the default param bugged, here is an example:  `v4l2-ctl -d0 --all`

```
brightness 0x00980900 (int)    : min=-64 max=64 step=1 default=-8193 value=0
                       contrast 0x00980901 (int)    : min=0 max=64 step=1 default=57343 value=32
                     saturation 0x00980902 (int)    : min=0 max=128 step=1 default=57343 value=105
                            hue 0x00980903 (int)    : min=-40 max=40 step=1 default=-8193 value=0
 white_balance_temperature_auto 0x0098090c (bool)   : default=1 value=1
                          gamma 0x00980910 (int)    : min=72 max=500 step=1 default=57343 value=100
                           gain 0x00980913 (int)    : min=0 max=100 step=1 default=57343 value=0
           power_line_frequency 0x00980918 (menu)   : min=0 max=2 default=1 value=1
      white_balance_temperature 0x0098091a (int)    : min=2800 max=6500 step=1 default=57343 value=4600 flags=inactive
                      sharpness 0x0098091b (int)    : min=0 max=6 step=1 default=57343 value=3
         backlight_compensation 0x0098091c (int)    : min=0 max=2 step=1 default=57343 value=1
                  exposure_auto 0x009a0901 (menu)   : min=0 max=3 default=0 value=3
              exposure_absolute 0x009a0902 (int)    : min=1 max=5000 step=1 default=93 value=93 flags=inactive
         exposure_auto_priority 0x009a0903 (bool)   : default=0 value=1
```

As you can see some default are corrupted with a common pattern -8193 or 57343 (it depends if default is signed or not ). Going around I found a lot of people with the exactly the same problem, but unfortunately no solution. The same camera works well on other systems with different kernel versions, I suspect the problems is on pi.
During last 2 days I found the right time to start to investigate. Honestly I was quite sure the problem was related to the uvc video driver, but it isn't. After digging in the uvc driver I found a buffer overflow caused maybe by usb stack. 
[...]

PR #3150, opened 2019-08-14
dwc_otg: use align_buf for small IN control transfers

P33M:
The hardware will do a 4-byte write to memory on any IN packet received
that is between 1 and 3 bytes long. This tramples memory in the uvcvideo
driver, as it uses a sequence of 1- and 2-byte control transfers to
query the min/max/range/step of each individual camera control and
gives us buffers that are offsets into a struct.

Catch small control transfers in the data phase and use the align_buf
to bounce the correct number of bytes into the URB's buffer.

In general, short packets on non-control endpoints should be OK as URBs
should have enough buffer space for a wMaxPacket size transfer.

See: https://github.com/raspberrypi/linux/issues/3148

### DIFF
--- a/drivers/usb/host/dwc_otg/dwc_otg_hcd.c
+++ b/drivers/usb/host/dwc_otg/dwc_otg_hcd.c
@@ -1182,6 +1182,7 @@ static void assign_and_init_hc(dwc_otg_hcd_t * hcd, dwc_otg_qh_t * qh)
 	dwc_otg_qtd_t *qtd;
 	dwc_otg_hcd_urb_t *urb;
 	void* ptr = NULL;
+	uint16_t wLength;
 	uint32_t intr_enable;
 	unsigned long flags;
 	gintmsk_data_t gintmsk = { .d32 = 0, };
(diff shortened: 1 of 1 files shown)


LF85:
Unfortunately this commit seems to breaks correct enumeration when using many usb devices. We habe a setup with multiple hubs and about 15+ devices (mostly cp210x based serial-to-usb converters)
and since this patch we regularly encounter  ERROR::assign_and_init_hc:1408: assign_and_init_hc: Failed to allocate memory to handle non-dword aligned buffer case

Could the default buffer size of 4096 be so small in that case (I'm sorry but I'm not yet familiar with the details of the usb-subsystem) ?  

We are running a RPi 3B+ on Raspbian Buster 4.19.69 (but it's not working since .66)

P33M:
One tedious thing about the driver is that align_bufs are allocated atomically. The memory has to come from a dedicated coherent pool which has a size set at boot.

What happens if you add coherent_pool=2M to the end of the parameter list in /boot/cmdline.txt (and reboot)?

LF85:
Thanks a lot. 
Adding  coherent_pool=2M  to the cmdline.txt seems to fix the issue.

floion:
Hi, so the proper fix for issues such as https://github.com/raspberrypi/linux/pull/3150#issuecomment-529796579 is adding the `coherent_pool=2M` param to cmdline.txt?

P33M:
It's a workaround, not a fix. The current allocation method is inefficient (one coherent buffer per endpoint that needs it). This could be changed to an up-front allocation of one align_buf per host channel instead, but this means there would be a large initial coherent allocation (number of channels * channel max transfer size = 8*2^19 = 4MB).

floion:
Thanks for the details @P33M 
Is there a bug on this that we could monitor?

PR #3158, opened 2019-08-17
RFC Restore V3D power domain

lategoodbye:
The commit ARM: dts: bcm283x: Move BCM2835/6/7 specific to
bcm2835-common.dtsi hasn't ported properly to the newer kernel versions 5.1
and newer. It accidentially drops the V3D power domain, which could lead to
firmware hangups.

This has only been compile tested, because i'm too busy with RPi 4 mainlining.


popcornmix:
Thanks.
@pelwell are you still around to okay this?

lategoodbye:
Reminder: This patch should be cherry-picked for rpi-5.2.y too

popcornmix:
Added to rpi-5.2.y

Issue #3157, opened 2019-08-17
More DMA issues on RPI4b in 64-bit mode with >1G RAM


yaroslavros:
It seems that there are still problems with DMA operation on RPI4 in 64-bit mode with > 1G RAM. This time on bcm2835-mmc.

When I try downloading large file via WiFi using scp, I run into reproducible swiotlb buffer issue at any points between 700Mb to 2Gb of downloaded data. Kernel logs are:

`bcm2835-dma fe00700.dma: swiotlb buffer is full (sz: 1536 bytes)`
`bcm2835-dma fe00700.dma: DMA: Out of SW-IOMMU space for 1536 bytes`
`mmc-bcm2835 fe300000.mmcnr: dma_map_sg returned zero length`
`mmc1: Timeout waiting for hardware interrupt.`

I have confirmed this at 2, 3 and 4Gb RAM. It does not occur on 1Gb RAM - I have downloaded 6Gb file via Wifi without any trouble.

This happens on the latest 4.19.y kernel

I'll dig more into this in upcoming days. I suspect there is a leak somewhere in the relevant driver stack that prevents swiotlb bounce buffers from being freed. This would explain normal behaviour on 1G RAM where bounce buffers are not used.

lategoodbye:
Please provide the exact kernel version

yaroslavros:
I see it on the latest 4.19.y 
`# uname -a`
`Linux rpi4a 4.19.64-v8 #34 SMP Sat Aug 17 13:51:51 BST 2019 aarch64 GNU/Linux`

I strongly suspect that this issue is related to the way bcm2835-mmc works with swiotlb and affects all kernel versions.

I am pretty sure root cause is the same as in #3114

yaroslavros:
Closing as the fix is committed

Issue #3114, opened 2019-07-31
[RPi 4 B][4.19.y] WiFi AP makes Kernel panic


kamarya:
**System**
Raspberry Pi 4 B
Linux Kernel Branch: 4.19.y
Linux Kernel Version: 4.19.58
SHA1 : d5dc848c982dff2e020f294e384447efe6ea6617
Distribution: Yocto Poky (meta-raspberrypi)

**Context of the issue**

Running _hostapd_ on the on-board WiFi interface as an access point. The AP works fine in the beginning and the traffic passes through. However after a few minutes RPi freezes and I get kernel panics about hostapd (see logs). Before having the panic about hostapd I get these interesting lines:
```
[  979.419611] bcm2835-dma fe007000.dma: swiotlb buffer is full (sz: 1536 bytes)
[  979.426928] bcm2835-dma fe007000.dma: DMA: Out of SW-IOMMU space for 1536 bytes
[  979.434559] mmc-bcm2835 fe300000.mmcnr: dma_map_sg returned zero length
[  989.661972] mmc1: Timeout waiting for hardware interrupt.
[  995.168392] ***** psta->aid (54) out of bounds
```

I also get kernel messages at the boot time complaining about cfg80211 (line 752 in sme.c)
This might be related to the fix of @agherzan (see SHA1 above) and it might still need more fixes. I have not tested this with other versions.

**Kernel logs**

```
root@raspberrypi4-64:~# [  965.706974] WARNING: CPU: 0 PID: 158 at net/wireless/sme.c:752 __cfg80211_connect_result+0x2fc/0x328 [cfg80211]
[  965.717220] Modules linked in: nfc bnep bluetooth ecdh_generic ipv6 iptable_filter xt_state xt_conntrack iptable_nat ipt_MASQUERADE nf_nat_ipv4 nf_nat nf_conntrack nf_defrag_ipv6 nf_defrag_ipv4 ip_tables x_tables 88XXau(O) brcmfmac brcmutil vc4 sha256_generic bcm2835_v4l2(C) bcm2835_mmal_vchiq(C) v4l2_common cfg80211 videobuf2_vmalloc snd_soc_core videobuf2_memops videobuf2_v4l2 videobuf2_common snd_compress videodev snd_pcm_dmaengine rfkill snd_pcm media snd_timer snd vc_sm_cma(C) vchiq(C) uio_pdrv_genirq uio fixed
[...]

PR #3159, opened 2019-08-17
Make bcm2835-mmc compatible with swiotlb that is used on ARM64

yaroslavros:
This resolves swiotlb buffer full errors on Wi-Fi bursts on
Raspberry PI4 in 64-bit mode.

Signed-off-by: Yaroslav Rosomakho <yaroslavros@gmail.com>

### DIFF
--- a/drivers/mmc/host/bcm2835-mmc.c
+++ b/drivers/mmc/host/bcm2835-mmc.c
@@ -38,6 +38,7 @@
 #include <linux/dmaengine.h>
 #include <linux/dma-mapping.h>
 #include <linux/of_dma.h>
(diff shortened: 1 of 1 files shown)


yaroslavros:
This fixes #3157 for me and should fix #3114 

I guess more efficient solution would be to implement proper IOMMU support instead of relying on swiotlb bouncing but that needs a lot more investigation.
Meanwhile, this patch does the trick of running WiFi reliably with >1G of RAM without sacrificing too much performance.

yaroslavros:
After stress testing wifi for 12+ hours it did eventually fail again. So, closing this PR.

PR #3164, opened 2019-08-18
Fix DMA leaks in bcm2835-mmc

yaroslavros:
DMA leaks caused swiotlb bounce buffer overflow errors under heavy wifi usage reported in #3157 and #3114 

This patch introduces two missing dma_unmap_sg calls for swiotlb to free relevant bounce buffers and adjusts max_req_size to match swiotlb maximum buffer size.

Signed-off-by: Yaroslav Rosomakho <yaroslavros@gmail.com>

### DIFF
--- a/drivers/mmc/host/bcm2835-mmc.c
+++ b/drivers/mmc/host/bcm2835-mmc.c
@@ -345,16 +345,17 @@ static void bcm2835_mmc_dma_complete(void *param)
 
 	host->use_dma = false;
 
-	if (host->data && !(host->data->flags & MMC_DATA_WRITE)) {
-		/* otherwise handled in SDHCI IRQ */
+	if (host->data) {
 		dma_chan = host->dma_chan_rxtx;
-		dir_data = DMA_FROM_DEVICE;
-
+		if (host->data->flags & MMC_DATA_WRITE)
+			dir_data = DMA_TO_DEVICE;
+		else
+			dir_data = DMA_FROM_DEVICE;
 		dma_unmap_sg(dma_chan->device->dev,
 		     host->data->sg, host->data->sg_len,
 		     dir_data);
-
-		bcm2835_mmc_finish_data(host);
+		if (! (host->data->flags & MMC_DATA_WRITE))
+			bcm2835_mmc_finish_data(host);
 	} else if (host->wait_for_dma) {
 		host->wait_for_dma = false;
 		tasklet_schedule(&host->finish_tasklet);
@@ -540,6 +541,8 @@ static void bcm2835_mmc_transfer_dma(struct bcm2835_host *host)
 		spin_unlock_irqrestore(&host->lock, flags);
 		dmaengine_submit(desc);
 		dma_async_issue_pending(dma_chan);
+	} else {
+		dma_unmap_sg(dma_chan->device->dev, host->data->sg, len, dir_data);
 	}
 
 }
--- a/drivers/mmc/host/bcm2835-mmc.c
+++ b/drivers/mmc/host/bcm2835-mmc.c
@@ -38,6 +38,7 @@
 #include <linux/dmaengine.h>
 #include <linux/dma-mapping.h>
 #include <linux/of_dma.h>
+#include <linux/swiotlb.h>
 
 #include "sdhci.h"
 
@@ -1374,7 +1375,10 @@ static int bcm2835_mmc_add_host(struct bcm2835_host *host)
 	}
 #endif
 	mmc->max_segs = 128;
-	mmc->max_req_size = 524288;
+	if (swiotlb_max_segment())
+		mmc->max_req_size = (1 << IO_TLB_SHIFT) * IO_TLB_SEGSIZE;
+	else
+		mmc->max_req_size = 524288;
 	mmc->max_seg_size = mmc->max_req_size;
 	mmc->max_blk_size = 512;
 	mmc->max_blk_count =  65535;


lategoodbye:
@yaroslavros 
I assume that adjusting max_req_size isn't directly related to the DMA leaks, so it would be better to split your commit. It would be nice to explain in the corresponding commit message, why you want to change the max_req_size?

yaroslavros:
swiotlb cannot handle more than 256K (IO_TLB_SHIFT=11, IO_TLB_SEGSIZE=128). This patch limits max_req_size under arm64 (or any other platform that uses swiotlb) to prevent potential buffer overflow due to bouncing.
I thought it would be reasonable to keep all three minor tweaks that resolve potential swiotlb buffer overflows in a single PR.

lategoodbye:
Keeping in a single PR is fine, but it should be different commits since these are different issues. Squashing multiple changes into a single make it harder to understand and port these changes.

agherzan:
Tested. Can't reproduce the crash anymore.

agherzan:
@yaroslavros Just a small reminder on the change for this PR.

yaroslavros:
As requested, split into two separate commits - https://github.com/raspberrypi/linux/pull/3164/commits/9dfbe9f500e85d2f751a240f6144785f237d795a (max_req_size) and https://github.com/raspberrypi/linux/pull/3164/commits/1395119b339e946fb87f4ff9da2c85c15947734d (dma_unmap_sg calls).

lategoodbye:
Thanks, LGTM.

Note: In case someone wants to port the max_req_size patch to Linux 5.3, this changed could be simplified with introduction of dma_max_mapping_size().
https://github.com/torvalds/linux/commit/63624d13b24d00fa3783181efdfdb9d5afaa7363

kamarya:
@yaroslavros I confirm that the issue persists even with the latest changes/fixes as I reported in https://github.com/raspberrypi/linux/issues/3114#issuecomment-524318646

The on-board Broadcom WiFi is exploited in AP mode and as you mentioned as soon as I start a video stream or download it stops working.

yaroslavros:
@kamarya are you sure you actually applying this commits from this PR? It is not merged yet.

lategoodbye:
@yaroslavros Oops, i didn't noticed this revert. Could please rebase this branch, so only the last 2 commits are left?

yaroslavros:
Cleaned up commits as requested.

kamarya:
Tested on RPi 4 B 2GB RAM and the issue could not be reproduced.

agherzan:
@pelwell Can we get this in?

pelwell:
Please explain why the `dma_unmap_sg` call in `bcm2835_mmc_data_irq` is inadequate for the MMC_DATA_READ case, and why this patch set doesn't result in `dma_unmap_sg` being called twice in that case.

yaroslavros:
I'm not sure if I understand your question.
bcm2835_mmc_data_irq is not actually getting called before or after bcm2835_mmc_dma_complete.

If it would and dma_unmag_sg triggered successfully, we would not have DMA leaks and crashes on arm64 that this PR addresses.

What scenario might lead to double unmap with this patch?

pelwell:
Your patch is simpler than the original, making the call to dma_unmap_sg unconditional on the direction, and it deletes a comment saying that SDHCI IRQ handles the other case. Please explain why your change is the _right_ change, as it is not obvious.

yaroslavros:
In my investigation of DMA leaks, I have added debug messages for all map/unmap calls and it was clear that despite the comment `bcm2835_mmc_data_irq` is actually not getting called before or after `bcm2835_mmc_dma_complete` - hence my addition of unconditional dma_unmap_sg. After change suggested by this PR, I have verified that there is one and only one unmap call corresponding to each map call.
I have followed this with multi-hour stress test of this solution. You can also see feedback from @agherzan and others above.

pelwell:
On which platforms did you test it?

yaroslavros:
RPI4 4GB running arm64 kernel

pelwell:
That's what I figured - no 32-bit testing.

yaroslavros:
At the moment I do not have access to any other platform or 32-bit runtime.

Are you suggesting that irq handling on arm32 is different from arm64?

pelwell:
I'm suggesting that the code has been working on 32-bit platforms for years and I'd like it to continue to do so.

yaroslavros:
Since 32-bit platforms do not utilise swiotlb / bounce buffers, there should not be drastic consequences of not performing unmap for each map call as long as actual memory is freed.

I'll be able to test this patch and confirm same map/unmap behaviour on arm32 in a week or two. Would be great if anyone had time and resource to do that sooner to accelerate things.

tkerber:
I should have a working 32-bit build soon, but I have no idea what I'd be testing for. Just kernel messages and WiFi stress tests?

pelwell:
That should be sufficient.

kamarya:
@tkerber I got the crash when the Wi-Fi interface was used as a hotspot. Streaming YouTube for a few minutes should reproduce the issue.

pelwell:
@kamarya Your testing was to show that the Pi 4 issue had been fixed, while I'm wanting some testing to show that the patch doesn't introduce new issues on other Pis.

tkerber:
I'm currently stress-testing the `armv7l` build, with `bcm2711_defconfig` of this patch on my raspberry pi 4 (4G), by repeatedly `rsync`ing my mail directory to the pi, and deleting it. This has been running for an hour and a half now, and the only kernel errors or warnings I've gotten are sd-card timeouts :) I'll leave the test running until this evening. If you'd like me to test another defconfig or target architecture, I can do that. The only other pi I have available to test is a pi 2, which afaik would be moot to test due to it not having the wireless device this patch touches.

pelwell:
I'll be happy with just a Pi4 in 32-bit mode. The architectural differences to other WiFi Pis aren't large enough to demand wider testing.

tkerber:
My testing got cut short by my router kicking the pi offline while I was at work, and reassigning it a new IP. That said, after over 8 hours of testing this 32-bit kernel with constant wireless traffic (alternating between incoming and outgoing: from my main pc: `while true; do rsync -avz mail <rpi>:; rsync -avz <rpi>:mail /tmp/mail; ssh <rpi> 'rm -rf mail'; rm -rf /tmp/mail; done`) my [`dmesg` output](https://github.com/raspberrypi/linux/files/3576097/kernel.log) shows nothing but a bunch of sdcard timeouts. I forgot to increase the loglevel before testing, so this is at the loglevel 4 (warn).

Noltari:
I've already integrated these patches in OpenWrt and I could see no regressions on a Raspberry Pi 3B running a 32 bit kernel:
https://gist.github.com/Noltari/2f53af3351cd09071aaf5e5414ed6956

Tests performed: Several speed tests & youtube (8k) videos over wifi while copying files to the SD card.

pelwell:
OK - I'll merge it and watch the early adopter canaries.

patricerosay:
Gents,
I'd be happy to test this patch as I'm experiencing a Kernel panic on BCM2835 when I plug an USB WIFI dongle + an USB audio card + a GPS-IMU equipment on a Raspberry Pi 4 Model B - 4GB

Could you please direct me to the documentation telling how to update my insallation with your patch ?

pelwell:
`sudo rpi-update` should be all you need.

patricerosay:
I've upgraded the  kernel with the v4.19.71-v71+ and It looks like this version doesn't fix my issue.
Please let me know if and how I could help

yaroslavros:
This PR is most likely irrelevant if you are running 32-bit kernel.

I would recommend opening a separate issue and providing a screenshot of kernel panic you observe with relevant diagnostic information.

PR #3165, opened 2019-08-19
Backport BCM270X_DT: Rename Pi Zero W DT files

fgiancane8:
This patch needed to be backported to series 4.14.y-rt because it is also supported in meta-raspberrypi for Yocto Project.
Unfortunately, the rename affected the correct building of the 4.14 series kernel, which now results broken.
Applying the same renames on the 4.14.y-rt series would fix the issue for those still using that kernel version.

The downtream Pi Zero W dts file uses the digit 0, whereas upstream
chose to spell it out - "zero-w". The firmware has, for a long time,
looked for bcm2708-rpi-zero-w.dtb first before falling back to the
numerical version. Therefore it is better to follow upstream and
make the switch to "bcm2708-rpi-zero-w".

At the same time, remove some overrides that duplicate values
inherited from the shared .dtsi files.

Signed-off-by: Francesco Giancane <francescogiancane8@gmail.com>

### DIFF
--- a/arch/arm/boot/dts/Makefile
+++ b/arch/arm/boot/dts/Makefile
@@ -5,7 +5,7 @@ dtb-$(CONFIG_ARCH_BCM2835) += \
 	bcm2708-rpi-b.dtb \
 	bcm2708-rpi-b-plus.dtb \
 	bcm2708-rpi-cm.dtb \
-	bcm2708-rpi-0-w.dtb \
+	bcm2708-rpi-zero-w.dtb \
 	bcm2709-rpi-2-b.dtb \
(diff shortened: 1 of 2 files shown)


fgiancane8:
@pelwell ,

Sorry for bothering again. In meta-raspberrypi we still support 4.14 series even for -rt builds. Is it doable to merge the back port i previously upstreamed for 4.14 also for this branch?

Many thanks,
F

pelwell:
I have the same comment as previously - this is a Yocto build issue rather than a problem in the repo itself - but again there's no good reason not to back-port.

PR #3166, opened 2019-08-19
vc-sm-cma: Fix compatibility ioctl

popcornmix:
This code path hasn't been used previously.
Fixed up after testing with kodi on 32-bit userland and 64-bit kernel


popcornmix:
Ping @6by9

6by9:
Lgtm

PR #3179, opened 2019-08-23
staging: bcm2835-codec: remove unnecessary padding on encoder input

tmm1:
The ISP and ENCODE roles have the same underlying hardware. Neither requires vertical alignment.

Originally pointed out by @6by9 in https://github.com/tmm1/FFmpeg/commit/55380ac7bd20fc764b14d282f2f3899b44059c13#r34807642

>Actually the encoder has to internally format convert the buffer for the hardware, so I'm not sure that there actually is a height alignment requirement. There isn't for the ISP role, and the format convert is being done by the same hardware.
>
>See https://github.com/raspberrypi/linux/blob/rpi-4.19.y/drivers/staging/vc04_services/bcm2835-codec/bcm2835-v4l2-codec.c#L1052. I'm suspecting that things should still be happy with ```if (ctx->dev->role == DECODE)```

I tested this and it works as expected.

### DIFF
--- a/drivers/staging/vc04_services/bcm2835-codec/bcm2835-v4l2-codec.c
+++ b/drivers/staging/vc04_services/bcm2835-codec/bcm2835-v4l2-codec.c
@@ -1050,12 +1050,12 @@ static int vidioc_try_fmt(struct bcm2835_codec_ctx *ctx, struct v4l2_format *f,
 			f->fmt.pix_mp.height = MIN_H;
 
 		/*
-		 * For codecs the buffer must have a vertical alignment of 16
+		 * For decoders the buffer must have a vertical alignment of 16
 		 * lines.
 		 * The selection will reflect any cropping rectangle when only
 		 * some of the pixels are active.
 		 */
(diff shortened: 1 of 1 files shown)


6by9:
I've not confirmed it for myself, but this was what I was expecting was the case.

Could you just confirm that your tests were with resolutions that were not naturally a multiple of 16 high?

tmm1:
Yes my tests were with height=1080 and I confirmed that before this change I needed to realign my input to generate correct encoded output.

IIUC, without this change you can't take the output from ISP and feed it directly to ENCODE

tmm1:
>IIUC, without this change you can't take the output from ISP and feed it directly to ENCODE

Well, only if the ISP was generating video sizes that were not multiples of 16

Just to be sure I tried another simple test without ISP, encoding 1920x1080 frames directly and confirmed they work with this change.

tmm1:
cc  @popcornmix and @pelwell  for review as well

6by9 (approved):
On the basis of your testing, approved.

PR #3188, opened 2019-08-28
imx219 kernel driver

6by9:
Sony have given their blessing for the included register sets to be released as a kernel driver, so here we go.
Using the dtoverlay should give raw access to the frames via V4L2.


6by9:
Rebased the branch - somehow my branch had been left behind at 4.19.58

6by9:
@pelwell Any chance of a review and merge here?

drivers/media/i2c/imx219.c
```diff
@@ -0,0 +1,1092 @@
...
+			goto err_unlock;
+		}
+
+		/*
+		 * Apply default & customized values
+		 * and then start streaming.
+		 */
+		ret = imx219_start_streaming(imx219);
+		if (ret)
+			goto err_rpm_put;
```
pelwell: Since this is the only use of err_rpm_put, and the pm_runtime_get* was in this block, and to match the one above, I would prefer:
```
		ret = imx219_start_streaming(imx219);
		if (ret) {
			pm_runtime_put(&client->dev);
			goto err_unlock;
		}
```
and the deletion of err_rpm_put below.

drivers/media/i2c/imx219.c
```diff
@@ -0,0 +1,1092 @@
...
+	ret = imx219_read_reg(imx219, IMX219_REG_CHIP_ID,
+			      IMX219_REG_VALUE_16BIT, &val);
+	if (ret) {
+		dev_err(&client->dev, "failed to read chip id %x\n",
+			IMX219_CHIP_ID);
+		goto power_off;
+	}
+
+	if (val != IMX219_CHIP_ID) {
+		dev_err(&client->dev, "chip id mismatch: %x!=%x\n",
```
pelwell, line 870: Do you need to return an error here?

arch/arm/boot/dts/overlays/imx219-overlay.dts
```diff
@@ -0,0 +1,129 @@
+// SPDX-License-Identifier: GPL-2.0-only
+// Definitions for IMX219 camera module on VC I2C bus
+/dts-v1/;
+/plugin/;
+
+#include <dt-bindings/gpio/gpio.h>
+
+/{
+	compatible = "brcm,bcm2708";
```
pelwell: Please use "brcm,bcm2835" - I'm trying to standardise.

kbingham:
for what it's worth, if it helps, I can add a
 Tested-by: Kieran Bingham <kieran.bingham@ideasonboard.com> 
to this series, as I've used these patches with Libcamera to capture images from the RPi Camera V2.

arch/arm/boot/dts/overlays/imx219-overlay.dts
```diff
@@ -0,0 +1,129 @@
...
+			cam0-pwdn-ctrl = <&imx219_vana>,"gpio:0";
+			cam0-pwdn      = <&imx219_vana>,"gpio:4";
+		};
+	};
+
+	__overrides__ {
+		i2c_pins_0_1 = <0>,"-2-3+4";
+		i2c_pins_28_29 = <0>,"+2-3-4";
+	};
+};
```
pelwell, line 129: This looks good, but I'd like a README entry, e.g. a cut-and-paste of ov5647.

pelwell (changes requested):

6by9:
Branch updated with requested changes.

pelwell (approved):

pelwell:
Do you want to add @kbingham's "Tested-by:" to the main commit? And perhaps add a space between the `@` and `30fps` in the message to be sure that the poor GitHub user doesn't get spammed.

6by9:
Commit text updated as requested.

arch/arm64/configs/bcm2711_defconfig
```diff
@@ -784,6 +784,7 @@ CONFIG_VIDEO_TVP5150=m
 CONFIG_VIDEO_TW2804=m
 CONFIG_VIDEO_TW9903=m
 CONFIG_VIDEO_TW9906=m
+CONFIG_VIDEO_IMX219=m
```
pelwell, line 787: FYI, regenerating the defconfigs loses this setting.

6by9: Thanks for the heads up.
I think we're missing CONFIG_VIDEO_V4L2_SUBDEV_API, which will drop out almost all of the sensor drivers :-(

6by9: Suffice to say, I haven't been testing on 64bit kernels as yet.

6by9: Looking at it, lots of options are missing from arm64/bcm2711_defconfig.
Time for a resync between arm and arm64.

pelwell: 9532eb3c84d8d952ef28da3d135683ac01adc9b8 fixes the immediate problem.

6by9: Thanks, but as CONFIG_VIDEO_BCM2835_UNICAM is also missing you get a sensor driver without anything to connect it to.

All the NFTABLES stuff, CGROUP_PIDS, I2C_MUX, LIRC/RC_CORE, USB DVB tuners, some BCM2708_SOC_DACs, and some other stuff is missing. I've just through with Meld and copied across all the obvious stuff. Now seeing if it builds.....

I'll throw it up as a PR so you can see what it proposed. There are still a few I haven't pulled over as they look processor related so I'm unsure if they are still relevant.

pelwell: > Thanks, but as CONFIG_VIDEO_BCM2835_UNICAM is also missing you get a sensor driver without anything to connect it to.

That's a secondary issue as far as I'm concerned. I just didn't want the IMX219 config setting to get lost as a side effect of adding some new ones.

PR #3189, opened 2019-08-28
Rpi 4.19.y libcamera

kbingham:
@6by9 Here's an updated series based on the latest 4.19.y, now with full links created for media controller objects.
A few fixes along the way too ... please review and consider for integration.

Cheers

Kieran

### DIFF
--- a/drivers/media/platform/bcm2835/bcm2835-unicam.c
+++ b/drivers/media/platform/bcm2835/bcm2835-unicam.c
@@ -963,7 +963,7 @@ static void unicam_cfg_image_id(struct unicam_device *dev)
 	}
 }
 
-void unicam_start_rx(struct unicam_device *dev, unsigned long addr)
+static void unicam_start_rx(struct unicam_device *dev, unsigned long addr)
 {
 	struct unicam_cfg *cfg = &dev->cfg;
 	int line_int_freq = dev->v_fmt.fmt.pix.height >> 2;
--- a/drivers/media/platform/bcm2835/bcm2835-unicam.c
+++ b/drivers/media/platform/bcm2835/bcm2835-unicam.c
@@ -314,6 +314,9 @@ struct unicam_device {
 	struct clk *clock;
 	/* V4l2 device */
 	struct v4l2_device v4l2_dev;
+	struct media_device mdev;
+	struct media_pad pad;
+
 	/* parent device */
 	struct platform_device *pdev;
 	/* subdevice async Notifier */
@@ -1912,6 +1915,8 @@ static int unicam_probe_complete(struct unicam_device *unicam)
 		unicam->v4l2_dev.ctrl_handler = NULL;
 
 	video_set_drvdata(vdev, unicam);
+	vdev->entity.flags |= MEDIA_ENT_FL_DEFAULT;
+
 	ret = video_register_device(vdev, VFL_TYPE_GRABBER, -1);
 	if (ret) {
 		unicam_err(unicam, "Unable to register video device.\n");
@@ -1949,6 +1954,16 @@ static int unicam_probe_complete(struct unicam_device *unicam)
 	if (ret) {
 		unicam_err(unicam,
 			   "Unable to register subdev nodes.\n");
+		video_unregister_device(&unicam->video_dev);
+		return ret;
+	}
+
+	ret = media_create_pad_link(&unicam->sensor->entity, 0,
+				    &unicam->video_dev.entity, 0,
+				    MEDIA_LNK_FL_ENABLED |
+				    MEDIA_LNK_FL_IMMUTABLE);
+	if (ret) {
+		unicam_err(unicam, "Unable to create pad links.\n");
 		video_unregister_device(&unicam->video_dev);
 		return ret;
 	}
@@ -2155,18 +2170,38 @@ static int unicam_probe(struct platform_device *pdev)
 		return -EINVAL;
 	}
 
+	unicam->mdev.dev = &pdev->dev;
+	strscpy(unicam->mdev.model, UNICAM_MODULE_NAME,
+		sizeof(unicam->mdev.model));
+	strscpy(unicam->mdev.serial, "", sizeof(unicam->mdev.serial));
+	snprintf(unicam->mdev.bus_info, sizeof(unicam->mdev.bus_info),
+		 "platform:%s", pdev->name);
+	unicam->mdev.hw_revision = 1;
+
+	media_entity_pads_init(&unicam->video_dev.entity, 1, &unicam->pad);
+	media_device_init(&unicam->mdev);
+
+	unicam->v4l2_dev.mdev = &unicam->mdev;
+
 	ret = v4l2_device_register(&pdev->dev, &unicam->v4l2_dev);
 	if (ret) {
 		unicam_err(unicam,
 			   "Unable to register v4l2 device.\n");
-		return ret;
+		goto media_cleanup;
+	}
+
+	ret = media_device_register(&unicam->mdev);
+	if (ret < 0) {
+		unicam_err(unicam,
+			   "Unable to register media-controller device.\n");
+		goto probe_out_v4l2_unregister;
 	}
 
 	/* Reserve space for the controls */
 	hdl = &unicam->ctrl_handler;
 	ret = v4l2_ctrl_handler_init(hdl, 16);
 	if (ret < 0)
-		goto probe_out_v4l2_unregister;
+		goto media_unregister;
 	unicam->v4l2_dev.ctrl_handler = hdl;
 
 	/* set the driver data in platform device */
@@ -2185,8 +2220,13 @@ static int unicam_probe(struct platform_device *pdev)
 
 free_hdl:
 	v4l2_ctrl_handler_free(hdl);
+media_unregister:
+	media_device_unregister(&unicam->mdev);
 probe_out_v4l2_unregister:
 	v4l2_device_unregister(&unicam->v4l2_dev);
+media_cleanup:
+	media_device_cleanup(&unicam->mdev);
+
 	return ret;
 }
 
@@ -2204,6 +2244,8 @@ static int unicam_remove(struct platform_device *pdev)
 	video_unregister_device(&unicam->video_dev);
 	if (unicam->sensor_config)
 		v4l2_subdev_free_pad_config(unicam->sensor_config);
+	media_device_unregister(&unicam->mdev);
+	media_device_cleanup(&unicam->mdev);
 
 	return 0;
 }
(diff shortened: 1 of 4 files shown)


kbingham:
Hi @tmm1 , Here's my PR to introduce role identifiers and media controller to the unicam and codec drivers.

drivers/media/platform/bcm2835/bcm2835-unicam.c
```diff
@@ -2155,11 +2170,31 @@ static int unicam_probe(struct platform_device *pdev)
 		return -EINVAL;
 	}
 
+	unicam->mdev.dev = &pdev->dev;
+	strscpy(unicam->mdev.model, UNICAM_MODULE_NAME,
+			sizeof(unicam->mdev.model));
```
6by9: Our old friend checkpatch whinges "Alignment should match open parenthesis"

drivers/media/platform/bcm2835/bcm2835-unicam.c
```diff
@@ -2155,11 +2170,31 @@ static int unicam_probe(struct platform_device *pdev)
...
+	strscpy(unicam->mdev.serial, "", sizeof(unicam->mdev.serial));
+	snprintf(unicam->mdev.bus_info, sizeof(unicam->mdev.bus_info),
+			"platform:%s", pdev->name);
```
6by9: And here too.

6by9 (commented):
Looks fine other than the two checkpatch warnings.
I'm building it to see what media-ctl makes of the world.

drivers/media/platform/bcm2835/bcm2835-unicam.c
kbingham: Argh, I have a commit hook that should have caught that (all my commits should get run through checkpatch at commit time and rebase time)...
Not sure why it didn't.
I'll fix up.

kbingham: ah - checkpatch runs clean - but checkpatch --strict doesn't :) I'll update my hooks accordingly.

6by9: :-) I know the frustration.
I have the same sort of hook (hence I found this on cherry-pick), and generally swear at it.

6by9 (approved):
Thanks. I do indeed get some /dev/mediaN nodes, and media-ctl says that it does something.

@popcornmix or @pelwell to review and merge.

tmm1:
> Hi @tmm1 , Here's my PR to introduce role identifiers and media controller to the unicam and codec drivers.

Thanks!

What exactly does card report after this change? I can't tell what the role is being appended to.

6by9:
Ignoring the media controller output:
```
pi@raspberrypi:~ $ v4l2-ctl -D -d 10
Driver Info:
	Driver name      : bcm2835-codec
	Card type        : bcm2835-codec-decode
	Bus info         : platform:bcm2835-codec
	Driver version   : 4.19.68
	Capabilities     : 0x84204000
		Video Memory-to-Memory Multiplanar
		Streaming
		Extended Pix Format
		Device Capabilities
	Device Caps      : 0x04204000
		Video Memory-to-Memory Multiplanar
		Streaming
		Extended Pix Format
pi@raspberrypi:~ $ v4l2-ctl -D -d 11
Driver Info:
	Driver name      : bcm2835-codec
	Card type        : bcm2835-codec-encode
	Bus info         : platform:bcm2835-codec
	Driver version   : 4.19.68
	Capabilities     : 0x84204000
		Video Memory-to-Memory Multiplanar
		Streaming
		Extended Pix Format
		Device Capabilities
	Device Caps      : 0x04204000
		Video Memory-to-Memory Multiplanar
		Streaming
		Extended Pix Format
pi@raspberrypi:~ $ v4l2-ctl -D -d 12
Driver Info:
	Driver name      : bcm2835-codec
	Card type        : bcm2835-codec-isp
	Bus info         : platform:bcm2835-codec
	Driver version   : 4.19.68
	Capabilities     : 0x84204000
		Video Memory-to-Memory Multiplanar
		Streaming
		Extended Pix Format
		Device Capabilities
	Device Caps      : 0x04204000
		Video Memory-to-Memory Multiplanar
		Streaming
		Extended Pix Format
```

tmm1:
Thanks. It seems a bit silly to me to repeat "bcm2835-codec" over and over and over. But this is good enough for my purposes.

kbingham:
We could remove the "bcm2835-codec" prefix from the card types and entity names easy enough I believe.
What do you think @6by9 ? I can't think of a reason why I've left the prefix in, except that I simply appended the roles rather than replacing with the role instead.

Would this be preferred?:

Driver Info:
	Driver name      : bcm2835-codec
	Card type        : isp
	Bus info         : platform:bcm2835-codec

If so I'd do the same for the entity names in the media-controller device.

6by9:
I'm happy either way.

AFAIK there is no standardisation on these strings, so it's only for information to the user who wants to try and identify the devices. It certainly shouldn't be used to try and determine the device purpose - use either media controller, or look at the formats supported on OUTPUT and CAPTURE queues.

The string is defined within ```struct video_device``` as being 32 chars in length, therefore making it shorter doesn't save any memory.

kbingham:
@6by9, I agree, especially as it won't save any space anyway...

And  now I see why I kept the prefix in:
cat /sys/class/video4linux/video12/name 
bcm2835-codec-isp

I think it's better that the video name shows the full context, so IMO, it's better with the prefix even though it is an apparent 'duplication of the bcm2835-codec portion of the string.

And it makes my 'lsv4l2' script output much more readable: (I should really post that for v4l2-utils)

cat /usr/bin/lsv4l2 
> for v in /sys/class/video4linux/video*;
>   do
>     vn=`basename $v`
>     name=`cat $v/name`
>     echo $vn: $name
>   done | sort -V;


pi@raspberrypi:~/libcamera $ lsv4l2 

> video0: unicam
> video10: bcm2835-codec-decode
> video11: bcm2835-codec-encode
> video12: bcm2835-codec-isp

So lets leave this series as it is.

6by9:
(I guess we do save 3 bytes for ```-%s``` in the printf string, twice over, but I'm really not fussing on that!)

Thanks for the discussion - agreed to leave it as is.

tmm1:
Appreciate the context. Agreed that it's fine as is and this is a big improvement over what it was before.

6by9:
@pelwell Any chance of a review and merge here? I'm happy with the patches.

drivers/staging/vc04_services/bcm2835-codec/bcm2835-v4l2-codec.c
```diff
@@ -2615,30 +2617,37 @@ static int bcm2835_codec_create(struct bcm2835_codec_driver *drv,
...
+		goto vchiq_finalise;
 
 	switch (role) {
 	case DECODE:
 		v4l2_disable_ioctl(vfd, VIDIOC_ENCODER_CMD);
 		v4l2_disable_ioctl(vfd, VIDIOC_TRY_ENCODER_CMD);
 		v4l2_disable_ioctl(vfd, VIDIOC_S_PARM);
 		v4l2_disable_ioctl(vfd, VIDIOC_G_PARM);
+		function = MEDIA_ENT_F_PROC_VIDEO_DECODER;
+
```
pelwell: This blank line (and the one below) are a bit unnecessary, especially since the ISP case doesn't use one.

drivers/staging/vc04_services/bcm2835-codec/bcm2835-v4l2-codec.c
```diff
@@ -947,8 +947,10 @@ static void device_run(void *priv)
 static int vidioc_querycap(struct file *file, void *priv,
 			   struct v4l2_capability *cap)
 {
+	struct bcm2835_codec_dev *dev = video_drvdata(file);
+
 	strncpy(cap->driver, MEM2MEM_NAME, sizeof(cap->driver) - 1);
-	strncpy(cap->card, MEM2MEM_NAME, sizeof(cap->card) - 1);
+	strncpy(cap->card, dev->vfd.name, sizeof(cap->card) - 1);
```
pelwell, line 956: You use strscpy in a later commit, which seems preferable, but strncpy here is consistent with the previous line. Your call.

drivers/staging/vc04_services/bcm2835-codec/bcm2835-v4l2-codec.c
```diff
@@ -2754,6 +2784,8 @@ static int bcm2835_codec_remove(struct platform_device *pdev)
 
 	bcm2835_codec_destroy(drv->decode);
 
+	media_device_unregister(&drv->mdev);
```
pelwell: On the basis that fitting is the reverse of removal, it seems nice to unregister the media device first, but is it necessary?

drivers/staging/vc04_services/bcm2835-codec/bcm2835-v4l2-codec.c
kbingham: Yes, I can remove those.

drivers/staging/vc04_services/bcm2835-codec/bcm2835-v4l2-codec.c
kbingham: Indeed, it's just that. I'm not adding or changing the strncpy. I'm only changing the MEM2MEM_NAME parameter.

If you'd like the strncpy changed to strscpy, then that should be a separate patch.

drivers/staging/vc04_services/bcm2835-codec/bcm2835-v4l2-codec.c
kbingham: Yes, I think it should happen before in fact, and I thnik I'm also missing calls to v4l2_m2m_unregister_media_controller() and media_device_cleanup().
I'll update this and repost.

kbingham: Updated, and tested with inserting and removing the module. Updated branch to follow

kbingham:
Branch updated with fixes, (is there a way to get github to show the difference between the previous push?)
Tested inserting and removing the modules, and capturing through libcamera on the imx219.

drivers/media/platform/bcm2835/bcm2835-unicam.c
```diff
@@ -2185,8 +2220,13 @@ static int unicam_probe(struct platform_device *pdev)
 
 free_hdl:
 	v4l2_ctrl_handler_free(hdl);
+media_unregister:
+	media_device_unregister(&unicam->mdev);
 probe_out_v4l2_unregister:
 	v4l2_device_unregister(&unicam->v4l2_dev);
+
```
kbingham: Ugh ... this line should probably be below :-|

kbingham: and fixed

PR #3191, opened 2019-08-28
Add support for the Audio Injector Ultra in 64-bit land.

stiltr:
Enables the CS4265 codec for 64-bit kernels.

@flatmax #2664 

Signed-off-by: Trevor Stiles <tstiles@shel.org>


flatmax:
Looks good.

PR #3193, opened 2019-08-29
arm: dts: add missing Raspberry Pi model names

Noltari:
This is needed to identify the different models on distributions like OpenWrt.

Signed-off-by: Álvaro Fernández Rojas <noltari@gmail.com>

### DIFF
--- a/arch/arm/boot/dts/bcm2710-rpi-cm3.dts
+++ b/arch/arm/boot/dts/bcm2710-rpi-cm3.dts
@@ -6,6 +6,7 @@
 #include "bcm283x-rpi-csi1-4lane.dtsi"
 
 / {
+	compatible = "raspberrypi,3-compute-module", "brcm,bcm2837";
 	model = "Raspberry Pi Compute Module 3";
 };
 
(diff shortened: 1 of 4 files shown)


arch/arm/boot/dts/bcm2710-rpi-cm3.dts
```diff
@@ -6,6 +6,7 @@
 #include "bcm283x-rpi-csi1-4lane.dtsi"
 
 / {
+	compatible = "raspberrypi,3-compute-module", "brcm,bcm2837";
 	model = "Raspberry Pi Compute Module 3";
```
JamesH65, line 10: Seems odd to start with the 3, rather than compute-module-3. Any reason for that?

lategoodbye: Looking at Documentation/devicetree/bindings/arm/bcm/brcm,bcm2835.txt shows it's the right compatible.

Noltari: https://github.com/raspberrypi/linux/blob/rpi-4.19.y/arch/arm/boot/dts/bcm2709-rpi-2-b.dts#L9
https://github.com/raspberrypi/linux/blob/rpi-4.19.y/arch/arm/boot/dts/bcm2710-rpi-3-b.dts#L9
https://github.com/raspberrypi/linux/blob/rpi-4.19.y/arch/arm/boot/dts/bcm2710-rpi-3-b-plus.dts#L9

PR #3202, opened 2019-09-03
softirq: Fix compiling errors when choosing schedule policies other than Fully Preemptible

imgaolong:
if choose no-rt schedule policies other than fully preemptible, such as Low Latency Desktop, the kernel will NOT compile. Define struct tasklet_hrtimer and function ksoftirqd_running to avoid compiling errors.

### DIFF
--- a/include/linux/interrupt.h
+++ b/include/linux/interrupt.h
@@ -648,6 +648,31 @@ extern void tasklet_kill_immediate(struct tasklet_struct *t, unsigned int cpu);
 extern void tasklet_init(struct tasklet_struct *t,
 			 void (*func)(unsigned long), unsigned long data);
 
+struct tasklet_hrtimer {
+	struct hrtimer          timer;
+	struct tasklet_struct   tasklet;
+	enum hrtimer_restart    (*function)(struct hrtimer *);
+};
+
+extern void
+tasklet_hrtimer_init(struct tasklet_hrtimer *ttimer,
+			enum hrtimer_restart (*function)(struct hrtimer *),
+			clockid_t which_clock, enum hrtimer_mode mode);
+
+static inline
+void tasklet_hrtimer_start(struct tasklet_hrtimer *ttimer, ktime_t time,
+			const enum hrtimer_mode mode)
+{
+	hrtimer_start(&ttimer->timer, time, mode);
+}
+
+static inline
+void tasklet_hrtimer_cancel(struct tasklet_hrtimer *ttimer)
+{
+	hrtimer_cancel(&ttimer->timer);
+	tasklet_kill(&ttimer->tasklet);
+}
+
 #ifdef CONFIG_PREEMPT_RT_FULL
 extern void softirq_early_init(void);
 #else
--- a/kernel/softirq.c
+++ b/kernel/softirq.c
@@ -176,6 +176,21 @@ static void wakeup_softirqd(void)
 		wake_up_process(tsk);
 }
 
+/*
+ * If ksoftirqd is scheduled, we do not want to process pending softirqs
+ * right now. Let ksoftirqd handle this at its own rate, to get fairness,
+ * unless we're doing some of the synchronous softirqs.
+ */
+#define SOFTIRQ_NOW_MASK ((1 << HI_SOFTIRQ) | (1 << TASKLET_SOFTIRQ))
+static bool ksoftirqd_running(unsigned long pending)
+{
+	struct task_struct *tsk = __this_cpu_read(ksoftirqd);
+
+	if (pending & SOFTIRQ_NOW_MASK)
+		return false;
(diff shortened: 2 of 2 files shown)


pelwell:
Can you explain why you would build no-rt kernels from a -rt branch?

imgaolong:
In fact, I am observing the effect of scheduling policies to xenomai/ipipe, which is a real-time kernel extension. 

Xenomai is a in-kernel module which can handle interrupts captured by ipipe, in spite of other part of the kernel.
As the rt patch enhanced real-time capability of the linux infrastructures, will it enhance the real-time capability of Xenomai, which is built on the base of Linux kernel ?
But when I came to test the schedule policies of rpi-4.14.y-rt branch, even the basic rt schedule policy,  I find it could not compile, so I tried to correct them.
What's more, I find if I directly apply the original rt patch patch-4.14.115-rt59 to branch rpi-4.14.y, which is 4.14.114, 

the result is even better than the original branch of rpi-4.14.y-rt, at least it can be compiled and can boot, except stop after lan78xx init.


Shall I try to pull a new rpi-4.14.y-rt-try branch of applying patch-4.14.115-rt59 directly to branch rpi-4.14.y?

 



 
 
------------------ Original ------------------

pelwell:
Which kernel config options enable this "ipipe" support (or which files does it involve)?

imgaolong:
> Can you explain why you would build no-rt kernels from a -rt branch?
with some certain net and display load, the cyclictest results on raspbery 3 show latency decrease alone with the schedule policies: Server(410 us max) Desktop(357 us max) Low Latency Desktop(342 us max), that is why I guess it could get even little if I use basic rt and full rt.

imgaolong:
> Which kernel config options enable this "ipipe" support (or which files does it involve)?


you have to apply the ipipe patch

imgaolong:
> Can you explain why you would build no-rt kernels from a -rt branch?


with some certain net and display load, the cyclictest results on raspbery 3 show latency decrease alone with the schedule policies: Server(410 us max) Desktop(357 us max) Low Latency Desktop(342 us max), that is why I guess it could get even little if I use basic rt and full rt.

imgaolong:
> Which kernel config options enable this "ipipe" support (or which files does it involve)?


here they are:
https://xenomai.org/
https://xenomai.org/downloads/ipipe/v4.x/

pelwell:
I don't know much about the '-rt' world, we just host the branches for the community. I think it's unlikely @TiejunChina or one of the other RT experts would object to your patch, but I'll leave it to them to decide.

TiejunChina:
@imgaolong You are supposed to compile everything even in -rt branch. But definitely you cannot  take over this right there since you're changing the common codes. They are not specific to rpi codes. Typically, if something is wrong in common place, you should go rt upstream branch instead of rpi-rt. 

In terms of some softirq codes, I recall that was caused one commit to refactor softirq in rpi-4.14.y, so rpi-4.14.y-rt followed that commit as well.  But as I discussed with @pelwell we don't maintain rpi-4.14.y-rt since we have rpi-4.19.y-rt now. @pelwell Do we need to remove rpi-4.14.y-rt?

pelwell:
We tend not to delete any code, and we apply patches to long-term-stable branches as they appear, but 4.19 is the active branch at the moment so unless the patch is a bug fix (and that's questionable) it probably shouldn't appear here.

imgaolong:
> 
> 
> We tend not to delete any code, and we apply patches to long-term-stable branches as they appear, but 4.19 is the active branch at the moment so unless the patch is a bug fix (and that's questionable) it probably shouldn't appear here.


actually，I pull request another branch of 4.14.y-114 in branch 4.14.y patched with 4.14.115 rt-patch,  instead of 4.14.91 in branch rpi-4.14.y-rt

> 
> 
> @imgaolong You are supposed to compile everything even in -rt branch. But definitely you cannot take over this right there since you're changing the common codes. They are not specific to rpi codes. Typically, if something is wrong in common place, you should go rt upstream branch instead of rpi-rt.
> 
> In terms of some softirq codes, I recall that was caused one commit to refactor softirq in rpi-4.14.y, so rpi-4.14.y-rt followed that commit as well. But as I discussed with @pelwell we don't maintain rpi-4.14.y-rt since we have rpi-4.19.y-rt now. @pelwell Do we need to remove rpi-4.14.y-rt?


In fact, I have pulled another branch of rpi-4.14.y-rt-114(rt:patched with rt patch 4.14.115 on 4.14.114 #3217), which is based on the kernel of 4.14.114 of branch rpi-4.14.y.As the rpi-4.14.y-rt branch is based on the 4.14.91, and have compiling errors when choosing basic rt schedule policy, how about switching to rpi-4.14.y-rt-144 instead of rpi-4.14.y-rt?

PR #3209, opened 2019-09-06
Rpi 5.3.y: alsa/soc/bcm: change to use modern dai_link style otherwise there is building error for v5.3 

jason77-wang:
There are some build errors in the sound/soc/bcm/ when building the 5.3 kernel.  In the v5.3, the kernel dropped all legacy dai_link stuff, so those drivers can't be built anymore, let us fix them.


pelwell:
Thanks! That builds and boots for me.

pelwell:
The commit removing the soundcards from the defconfigs has been reverted.

Issue #3207, opened 2019-09-05
Update the MCP23017 overlay to support using multiple devices with repeated calls


Dloranger:
**Describe the bug**
Currently the overlay only supports 1 chipset, but the device is capable of having 8 on the bus with only an address change (pin setting) and using a new gpio pin for the int pin.

Can we upgrade this dtbo to be able to use the address field to make each call unique to be able to load from 2-8 chipsets with multiple calls to the dtbo?

This would greatly simplify my development workflow by allowing me to use stock dtbo calls rather than having to recompile custom dtbo regularly.

**To reproduce**
call the mcp23017 overlay 2x, only the last one is used.

**Expected behaviour**
call the overlay 2-8x and all 2-8 devices would be loaded as unique devices.

**Actual behaviour**
Only 1 device gets loaded

**System**
this is an issue that has persisted due to the implementation of the overlay, not specific to any particular version of code or hardware.

6by9:
I suspect the issue is the collision in the name ```mcp23017_pins``` to describe the interrupt GPIO.
https://github.com/raspberrypi/linux/blob/rpi-4.19.y/arch/arm/boot/dts/overlays/mcp23017-overlay.dts

The interrupt GPIO is optional, so one could remove it and the mcp23017_pins fragment. Making it an overlay option to remove it would require someone with more DT overlay knowledge than me....

6by9:
Having said that, I think that falls out easily.

```
// Definitions for MCP23017 Gpio Extender from Microchip Semiconductor

/dts-v1/;
/plugin/;

/ {
	compatible = "brcm,bcm2835";

	fragment@0 {
		target = <&i2c1>;
		__overlay__ {
			status = "okay";
		};
	};

	fragment@1 {
		target = <&gpio>;
		__overlay__ {
			mcp23017_pins: mcp23017_pins {
				brcm,pins = <4>;
				brcm,function = <0>;
			};
		};
	};

	fragment@2 {
		target = <&i2c1>;
		__overlay__ {
			#address-cells = <1>;
			#size-cells = <0>;

			mcp23017: mcp@20 {
				compatible = "microchip,mcp23017";
				reg = <0x20>;
				gpio-controller;
				#gpio-cells = <2>;

				status = "okay";
			};
		};
	};

	fragment@3 {
		target = <&mcp23017>;
		__dormant__ {
			compatible = "microchip,mcp23008";
		};
	};

	fragment@4 {
		target = <&mcp23017>;
[...]

PR #3211, opened 2019-09-06
mcp23017 overlay updates

6by9:
Renames the mcp23017_pins node so that we don't get collisions if trying to load it multiple times.

Adds an option for disabling the interrupt GPIO line.


6by9:
#3207

PR #3212, opened 2019-09-06
Add Greyworld AWB option to v4l2 and bcm2835-camera driver

6by9:
Needed for the NoIR camera when running the new AWB algorithm.


pelwell:
Looks OK by analogy.

6by9:
Confirmed via https://github.com/raspberrypi/firmware/issues/1225 that this works.
WIP tag removed.

Issue #1866, opened 2017-02-28
Poor wireless performance on 4.9.* kernels


Rand0mJ0e:
Raspberry pi zero was working fine with 4.4.38 kernel. Upgraded to 4.9.13 and observed very poor wireless performance using RTL8192CU wireless adapter.

With 4.9.13 kernel:
wlan0 IEEE 802.11 ESSID:"CatHouse"
Mode:Managed Frequency:2.422 GHz Access Point: 20:AA:4B:90:0F:BE
Bit Rate=72.2 Mb/s Tx-Power=20 dBm
Retry short limit:7 RTS thr=2347 B Fragment thr:off
Power Management:off
**Link Quality=45/70 Signal level=-65 dBm**
Rx invalid nwid:0 Rx invalid crypt:0 Rx invalid frag:0
Tx excessive retries:0 **Invalid misc:45** Missed beacon:0
With 4.4.38 kernel:
wlan0 IEEE 802.11bgn ESSID:"CatHouse" Nickname:"WIFI@REALTEK"
Mode:Managed Frequency:2.422 GHz Access Point: 20:AA:4B:90:0F:BE
Bit Rate:144.4 Mb/s Sensitivity:0/0
Retry:off RTS thr:off Fragment thr:off
Power Management:off
Link Quality=100/100 Signal level=58/100 Noise level=0/100
Rx invalid nwid:0 Rx invalid crypt:0 Rx invalid frag:0
Tx excessive retries:0 Invalid misc:0 Missed beacon:0

Problem also exists on 4.9.11 on Raspberry pi 2.
Problem resolved by downgrading back to 4.4.38 or to any 4.4.x level.

popcornmix:
@MrEngman
I wonder if you've seen any change in performance with RTL8192CU devices on the newer kernel?

MrEngman:
Hi @popcornmix ,

Sorry, I haven't used the RTL8192cu wifi with 4.9 so can't say.

However, I have notice some odd behaviour using out of tree 8812au or 8192eu wifi and various 4.9.xx kernels. With the 8812au driver using a 5GB AP running iperf and 4.4 kernels I get transfer rates of around 95MBits/s but sometimes using kernel 4.9 I am seeing less than half this rate, around 42MBits/s. Powering off the Pi and disconnecting the PSU then re-applying power and rebooting and then iperf may give the 95MBits/s transfer rate, but not always.

[...]

PR #3215, opened 2019-09-08
Fix two building errors when we build the kernel in the non-src folder

jason77-wang:
It is because the building process can't find the needed header files. And this already introduced building errors on the branches of 5.2.y and 5.3.y for RPi4, and I guess all branches need these two fixes, it is safe to merge these two patches to base branch then it will apply to all branches automatically?


pelwell:
@kbingham This reverses [your commit](f415f6d9a61b6caadcd09d3816179ee31e43e835) - any comments?

kbingham:
@pelwell  Ack on this. 
Tested-by: Kieran Bingham <kieran.bingham@ideasonboard.com>

It looks like my earlier fix was squashed in, so perhaps it got squashed or merged with some other change that affected things here.

Anyway, I can confirm Jason's patch compiles VC_SM where as without Jason's patch - there is a failure to build the VC_SM code.

I have not tested the rtl8192cu.

pelwell:
Thanks. I'll take the 8192 change on trust.

kbingham:
Well, to put your mind at ease, as I had the build environment all set up I've added CONFIG_RTL8192CU=y and it built successfully.
 (Although there is a lot of clean-up required in that driver!!!)

Out of interest, why is this driver included in the RPi kernel?
https://github.com/pvaret/rtl8192cu-fixes declares that it is deprecated and that rtl8xxxu available from linux-4.4 should be used first. As this tree is now 5.3....

Is there a specific use case needed to support this driver? If not - I'd perhaps recommend dropping it!

pelwell:
> Is there a specific use case needed to support this driver?

@popcornmix?

popcornmix:
When using 4.9 kernel we switched to mainline driver and got a lot of complaints about unreliable wireless. We currently build both and blacklist one so users can switch.
See: https://github.com/raspberrypi/linux/issues/1866

It is possible the driver has improved in the meantime, but I'm not too optimistic.

PR #3223, opened 2019-09-11
Composite fix

JamesH65:
Margins were incorrectly assumed to be setup in SDTV mode, but were
not actually done, so this make the setup non-conditional on mode.


6by9 (approved):
Other than the commit text (should start "drm/vc4: " and be slightly more descriptive on the first line, and no "Signed-off-by:"), fine by me.

PR #3224, opened 2019-09-11
add Hifiberry DAC+DSP soundcard driver

j-schambacher:
new driver to be used instead of DAC-only driver to allow capture and playback (depending on DSP firmware capabilities).


arch/arm/boot/dts/overlays/hifiberry-dspdac-overlay.dts
```diff
@@ -0,0 +1,35 @@
+// Definitions for hifiberry dspdac soundcard overlay
+/dts-v1/;
+/plugin/;
+
+/ {
+	compatible = "brcm,bcm2708";
```
pelwell: Please use "brcm,bcm2835" from now on, unless it is Pi4-specific in which case use "brcm,bcm2838".

pelwell (changes requested):
@hifiberry Is this an official driver?

Please squash these commits down to 1, generate a patch file (`git format-patch HEAD^` should work after squashing) and run `scripts/checkpatch.pl` on it - you'll find lots of whitespace and other formatting errors. Once you've got it to the point that checkpatch is only complaining about the lack of documentation of compatible strings, use `git push -f` to update your PR.

hifiberry:
@pelwell Yes this is "official".

j-schambacher:
something went completely wrong with my repo.. still not the version I want ....let you know when I've fixed it...sorry

pelwell:
Take your time. By the way, `git rebase -i HEAD~n` is a very useful command for tidying up commits, if you haven't come across it before.

pelwell:
Oh - and a comment when you're done helps because we don't often get notifications of pushes.

j-schambacher:
@pelwell done the requested fixes and changes. I hope it's ok now to merge.

pelwell:
This is a completely passive board - there are no controls, it just adapts to the clock and sample sizes used by the CPU - so I suspect the codec driver is unnecessary and the card driver could be replaced with simple-card, but I don't have time to investigate further.

j-schambacher:
@pelwell you are right, that's why I integrated it into the simple-card drivers. But so far the set does not contain a simple capture AND play without controls. A generically named copy of it may be a solution, but for now we want to keep the hifiberry specific name to allow potential HW tweaks like REST via GPIO.

j-schambacher:
@pelwell and thanks for all!

PR #3228, opened 2019-09-13
bcm2835-codec: Minor fixes

6by9:


drivers/staging/vc04_services/bcm2835-codec/bcm2835-v4l2-codec.c
```diff
@@ -1260,15 +1260,37 @@ static int vidioc_g_selection(struct file *file, void *priv,
...
+		q_data = &ctx->q_data[V4L2_M2M_SRC];
+		break;
+	default:
 		return -EINVAL;
+	}
 
-	q_data = get_q_data(ctx, s->type);
+	v4l2_err(&ctx->dev->v4l2_dev, "%s: queue %u\n", __func__, type);
+
+	q_data = get_q_data(ctx, type);
```
pelwell: q_data is assigned above, and again here without any intervening use, which looks wrong. By analogy with vidioc_s_selection I think this line should be deleted.

6by9: Quite correct, and made me check the 
```
if (!q_data)
		return -EINVAL;
```
below,. Seeing as q_data is assigned the address of an array within ctx, it can never be NULL, so no need to check it. Removed that too.

6by9: Not to mention the debug line left behind, and ```type``` being unused.

PR #3235, opened 2019-09-16
ARM: dts: Add DTS for Pi 2B rev 1.2 with BCM2837

pelwell:
This is an untested DTS for a Pi 2B rev 1.2 (with BCM2837), posted here for comment.

### DIFF
--- a/arch/arm/boot/dts/bcm2710-rpi-2-b.dts
+++ b/arch/arm/boot/dts/bcm2710-rpi-2-b.dts
@@ -1,0 +1,125 @@
+/dts-v1/;
+
+#include "bcm2710.dtsi"
+#include "bcm2709-rpi.dtsi"
+#include "bcm283x-rpi-smsc9514.dtsi"
+#include "bcm283x-rpi-csi1-2lane.dtsi"
+
+/ {
+	compatible = "raspberrypi,2-model-b-rev2", "brcm,bcm2837";
+	model = "Raspberry Pi 2 Model B rev 1.2";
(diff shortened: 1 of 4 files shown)


digitallyelite:
Tested unchanged here, it does boot and give all 4 CPUs, will do further testing soon-ishly.

arch/arm/boot/dts/bcm2710-rpi-2-b.dts
```diff
@@ -0,0 +1,125 @@
+/dts-v1/;
+
+#include "bcm2710.dtsi"
+#include "bcm2709-rpi.dtsi"
+#include "bcm283x-rpi-smsc9514.dtsi"
+#include "bcm283x-rpi-csi1-2lane.dtsi"
+
+/ {
+	compatible = "raspberrypi,2-model-b", "brcm,bcm2836";
+	model = "Raspberry Pi 2 Model B";
```
lategoodbye: This should be changed to represent the real SoC and maybe a different board naming to distinguish the Raspberry Pi 2 boards.

pelwell: Updated.

digitallyelite:
Tested more here, correct model shows up, all 4 cores show up, 1GB of ram shows up, rng-tools start.  I have not tested the GPU at all, but I would assume it would be same status as the rest of the 64bit support.

pelwell:
Thanks for testing.

PR #3239, opened 2019-09-17
configs: Update arm64/bcm2711_defconfig based on the arm version

popcornmix:
The arm64 and arm configs had diverged in the extra modules that were being built.

Resync the two for the obvious modules.


popcornmix:
@pelwell @6by9 can you see any issues with this?

It builds and boots up.

pelwell:
Looks good to me.

PR #3241, opened 2019-09-18
tty: amba-pl011: add rs485 binding for RTS polarity

hase86:
rs485 allows for robust half-duplex serial communication.  It is often
implemented by attaching an rs485 transceiver to a UART.  The UART's
RTS line is wired to the transceiver's Transmit Enable pin and
determines whether the transceiver is sending or receiving.

Examples for such transceivers are Maxim MAX13451E and TI SN65HVD1781A:
https://datasheets.maximintegrated.com/en/ds/MAX13450E-MAX13451E.pdf
http://www.ti.com/lit/ds/symlink/sn65hvd1781a-q1.pdf

In the devicetree, the transceiver itself is not represented, only the
UART is. A few rs485-specific dt-bindings already exist and these go
into the UART's device node.

This commit adds a binding to set the RTS polarity.  Most (if not all)
transceivers require the Transmit Enable pin be driven high for sending,
but in some cases boards may negate the pin and RTS must then be driven
low. Consequently the polarity defaults to active high but can be
inverted with the newly added "rs485-rts-active-low" binding.

### DIFF
--- a/drivers/tty/serial/amba-pl011.c
+++ b/drivers/tty/serial/amba-pl011.c
@@ -1299,6 +1299,29 @@ static void pl011_stop_tx(struct uart_port *port)
 {
 	struct uart_amba_port *uap =
 	    container_of(port, struct uart_amba_port, port);
+	unsigned int cr_old, cr;
+
+	int timeout = 10000;
+	
+	/* Handle RS-485 */
+	if (port->rs485.flags & SER_RS485_ENABLED) {
+		while ((pl011_read(uap, REG_FR) & UART01x_FR_BUSY) && --timeout)
+			/* the UART is busy transmitting data */
+			udelay(1);
+		
+		cr_old = pl011_read(uap, REG_CR);
+		
+		if (port->rs485.flags & SER_RS485_RTS_AFTER_SEND)
+			cr = cr_old & ~UART011_CR_RTS;	/* inverse logic */
+		else
+			cr = cr_old | UART011_CR_RTS;	/* inverse logic */
+		
+		if (cr_old != cr) {
+			if (port->rs485.delay_rts_after_send > 0)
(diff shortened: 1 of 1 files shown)


pelwell:
This patch should be submitted upstream. Once accepted, we'll back-port it.

The upstream devs will want a commit message which isn't just copied from 6abe9ea8.

Something simple like:
```
tty: amba-pl011: Add software emulated RS485 support for pl011

...

Signed-off-by: Your Name <your.email@whatever.com>
```

drivers/tty/serial/amba-pl011.c
```diff
@@ -1320,6 +1343,24 @@ static void pl011_start_tx(struct uart_port *port)
...
+
+		if (port->rs485.flags & SER_RS485_RTS_ON_SEND)
+			cr = cr_old & ~UART011_CR_RTS;	/* inverse logic */
+		else
+			cr = cr_old | UART011_CR_RTS;	/* inverse logic */
+		
+		if (cr_old != cr) {
+			pl011_write(cr, uap, REG_CR);
+			if (port->rs485.delay_rts_before_send > 0) {
+				mdelay(port->rs485.delay_rts_before_send);
```
lategoodbye, line 1360: Please try to find a better solution which doesn't waste the CPU power.

sistik:
I have created new implementation of RS485 for PL011.
Pending patch:
https://patchwork.kernel.org/patch/11320241/

RuhanSA079:
Hello, is there any plans to merge this PR into the base branch and make it available in the newest rev. kernel?

pelwell:
We're all waiting for the patch to be accepted upstream - it's a significant change to a common driver, and we don't want to maintain it downstream.

PR #3254, opened 2019-09-24
config: Update arm64 USB options to match arm

popcornmix:


popcornmix:
This reduces differences in 32 and 64-bit kernels. @P33M thinks they are fine.
It boots and USB works for me.
See: https://www.raspberrypi.org/forums/viewtopic.php?f=29&t=250730&start=100#p1540314

PR #3255, opened 2019-09-24
RFC: Rename HDMI ALSA device names, check for enable state

JamesH65:
HDMI Alsa devices renamed to match names used by DRM, to
HDMI 1 and HDMI 2

Check for which HDMI devices are connected and only create
devices for those that are present.

The rename of the devices might cause some backwards compatibility
issues, but since this particular part of the driver needs to be
specifically enabled, I suspect the number of people who will see
the problem will be very small.

### DIFF
--- a/drivers/staging/vc04_services/bcm2835-audio/bcm2835.c
+++ b/drivers/staging/vc04_services/bcm2835-audio/bcm2835.c
@@ -9,8 +9,9 @@
 #include <linux/of.h>
 
 #include "bcm2835.h"
+#include <soc/bcm2835/raspberrypi-firmware.h>
 
-static bool enable_hdmi;
+static bool enable_hdmi, enable_hdmi0, enable_hdmi1;
 static bool enable_headphones;
 static bool enable_compat_alsa = true;
 
@@ -115,8 +116,8 @@ static struct bcm2835_audio_driver bcm2835_audio_hdmi0 = {
 		.name = "bcm2835_hdmi",
 		.owner = THIS_MODULE,
 	},
-	.shortname = "bcm2835 HDMI",
-	.longname  = "bcm2835 HDMI",
+	.shortname = "bcm2835 HDMI 1",
+	.longname  = "bcm2835 HDMI 1",
 	.minchannels = 1,
 	.newpcm = bcm2835_audio_simple_newpcm,
 	.newctl = snd_bcm2835_new_hdmi_ctl,
@@ -128,8 +129,8 @@ static struct bcm2835_audio_driver bcm2835_audio_hdmi1 = {
 		.name = "bcm2835_hdmi",
 		.owner = THIS_MODULE,
 	},
-	.shortname = "bcm2835 HDMI 1",
-	.longname  = "bcm2835 HDMI 1",
+	.shortname = "bcm2835 HDMI 2",
+	.longname  = "bcm2835 HDMI 2",
 	.minchannels = 1,
 	.newpcm = bcm2835_audio_simple_newpcm,
 	.newctl = snd_bcm2835_new_hdmi_ctl,
@@ -161,11 +162,11 @@ static struct bcm2835_audio_drivers children_devices[] = {
 	},
 	{
 		.audio_driver = &bcm2835_audio_hdmi0,
-		.is_enabled = &enable_hdmi,
+		.is_enabled = &enable_hdmi0,
 	},
 	{
 		.audio_driver = &bcm2835_audio_hdmi1,
-		.is_enabled = &enable_hdmi,
+		.is_enabled = &enable_hdmi1,
 	},
 	{
 		.audio_driver = &bcm2835_audio_headphones,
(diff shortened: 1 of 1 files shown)


drivers/staging/vc04_services/bcm2835-audio/bcm2835.c
```diff
@@ -312,6 +313,45 @@ static int snd_add_child_devices(struct device *device, u32 numchans)
...
+				&display_id, sizeof(display_id));
+		if (!ret) {
+			if (display_id == 2)
+				enable_hdmi0 = true;
+			if (display_id == 7)
+				enable_hdmi1 = true;
+		}
+	}
+
+	return bits;
```
popcornmix: Is bits meant to be something other than 0?

JamesH65: Good spot, left something in from a previous iteration of the code. Will sort when I get in to the office.

JamesH65: Fixed up and renamed the function to be more accurate.

JamesH65:
To enable this code you need snd-bcm2835.enable_compat_alsa=0 in cmdline.txt

I've checked this with two HDMI devices attached, and I get two sound cards enabled and can switch between them in alsamixer, or play back on each device using aplay -D hw:0,0 <blah> or aplay -D hw:1,0 <blah>.

pelwell:
With a single attached monitor I can see the device name changing from "bcm2835 HDMI 1" (using HDMI0) to "bcm2835 HDMI 2" (on HDMI1). However, the volume control is disabled. Do I need an update?

pelwell:
And a cosmetic point - is "bcm2835 HDMI 1" really a short name? Isn't "HDMI 1" sufficient?

JamesH65:
Have fixed up the scanning so that if only one HDMI is connected, it will always be named HDMI 1, irrespective of which port it is actually attached to.  This should match how DRM names the HDMI's. Have also, in this alsa mode, forced the analog output on. A quick test in alsamixer shows all devices appear as expected, and the audio is correctly routed if only one HDMI device is attached.

Spoke to @spl237 re: renaming the names to something shorter. Currently the bcm2835 prefix is used to identify our driver, and he would prefer something along those lines to remain. I do wonder if simply renaming to Pi HDMI1, Pi HDMI 2 and Pi Headphones would be a better and more descriptive option, and still providing easy parsing for the desktop plugin. Another option might be different long and short names, with the plugin checking which one is most appropriate.

drivers/staging/vc04_services/bcm2835-audio/bcm2835.c
```diff
@@ -110,13 +111,16 @@ static struct bcm2835_audio_driver bcm2835_audio_alsa = {
 	.newctl = snd_bcm2835_new_ctl,
 };
 
-static struct bcm2835_audio_driver bcm2835_audio_hdmi0 = {
+/* This struct NOT static as we may wish to reassign the route
```
pelwell: Are you confusing `static` and `const`? static is fine (the structure is only used within this compilation unit/file), but const would be a problem if it were used.

drivers/staging/vc04_services/bcm2835-audio/bcm2835.c
```diff
@@ -312,6 +316,53 @@ static int snd_add_child_devices(struct device *device, u32 numchans)
...
+
+	if (!enable_hdmi0 && enable_hdmi1) {
+		/* Swap them over and reassign route. This means
+		   that if we only have one connected, it is always named
+		   HDMI1, irrespective of if its on port HDMI0 or HDMI1.
+		   This should match with the naming of HDMI ports in DRM
+		*/
+		enable_hdmi0 = true;
+		enable_hdmi1 = false;
+		bcm2835_audio_hdmi0.route = AUDIO_DEST_HDMI1;
```
pelwell, line 359: Is this a one-shot driver, or should we always set route appropriately?

drivers/staging/vc04_services/bcm2835-audio/bcm2835.c
JamesH65: Bugger.

drivers/staging/vc04_services/bcm2835-audio/bcm2835.c
JamesH65: One shot I believe, but might change depending on HDMI hotplug support I suppose.

drivers/staging/vc04_services/bcm2835-audio/bcm2835.c
JamesH65: Fixed. Rookie mistake!

pelwell:
No objections from me.

JamesH65:
I think there still need to be a discussion on whether this should be enabled by default. (currently needs to be enabled on cmdline). My thoughts are this is now the 'right' way of doing it, but changing the default might cause issues with some people (Kodi?). Right now on the Pi4 we get no audio if using HDMI port 1, and this should fix that.

popcornmix:
Most kodi users use the PiSink (openmax) interface so won't be affected.
The non-optimised debian kodi would be using alsa so may be affected, but I don't imagine many users are using that (the Pi-specific build works much better).

No doubt there are other applications that use alsa that could be affected (I assume only apps that allow configuring audio device and device specific settings, rather than just outputting to the default alsa device).

6by9:
When we enable hot plug detection reporting from kms then you can expect both hdmi display devices to be initialised, wityh the connection state advertised via hot plug and kms. (if a dpi or dsi display is configured, then you'll only get hdmi0 enumerated due to the hvs channel usage).

There was discussion on dri-devel over what the correct interaction between alsa and kms is for adding and removing devices, but I don't recall the final conclusion. In theory the hdmi sink advertises the audio capabilities and the alsa device ought to reflect those. 
I'll try to dig out the thread.

JamesH65:
At the moment, if the driver is removed then reinserted it would reinitialise itself to reflect what the mailbox call to enumerate display devices returns. So as long as that system correctly reflects the connections states, that is one possibility. I guess in KMS we do not have mailbox calls, so that will need to change in the long run. I suspect at some point this driver will need some fairly extensive fixups to deal with whatever DRM decides to do.

JamesH65:
Does anyone think we should run a small beta test program on this change? Run from the forum or similar for those people encountering audio issues on Pi4. Would require a new desktop plugin, unless we give instructions for setting the audio destination on the command line instead.

6by9:
First thing is to test ```snd-bcm2835.enable_compat_alsa=0``` with HDMI passthrough. Does it work?

If it does then that really wants to become the default with all devices enabled. The Volume plugin needs to then handle it properly (I believe it currently sees the names as too close to the "special case" compat mode device, gets confused, and presents no devices instead).

This change then becomes an optimisation for only showing the devices that are actually connected.

JamesH65:
I do not have any equipment to test HDMI passthrough, does anyone else?

JamesH65:
In order for @spl237 to start work on the taskbar plugin, this probably needs to be merged. As above, HDMI pass through has not been checked, but since this code path is only exercised if you specifically enable it, I think we should be mostly OK from end users point of view.

pelwell:
Merged under duress.

PR #3261, opened 2019-10-02
irs1125 infineon tof camera support

Coimbra1984:
As discussed very early this year (january) with @6by9 this pull reuquest adds support for the infineon tof sensor irs1125, which is used in our 3D tof frontend for raspberry pi (see https://www.conrad.de/de/p/makerfactory-x-pieye-3d-cmos-farb-kameramodul-passend-fuer-raspberry-pi-2141283.html).

The changes in bcm2835-unicam.c are some kind of hack from my side. I guess some more discussions might be necessary there.


drivers/media/i2c/irs1125_regs.h
```diff
@@ -0,0 +1,4126 @@
+/*
+ * IRS1125x All-in-One firmware v.2126
+ *
+ * Copyright Infineon Technologies AG 2017-2019.
+ * Distributed under the Boost Software License, Version 1.0.
+ * (See accompanying file LICENSE_1_0.txt or copy at
+ * https://www.boost.org/LICENSE_1_0.txt)
+ */
```
6by9: Why the odd licence? I don't believe this is compatible with the Linux kernel licencing.
See https://github.com/raspberrypi/linux/blob/rpi-4.19.y/Documentation/process/license-rules.rst for guidance, and https://github.com/raspberrypi/linux/tree/rpi-4.19.y/LICENSES for the licence texts

arch/arm/boot/dts/overlays/irs1125-overlay.dts
```diff
@@ -0,0 +1,86 @@
+// Definitions for IRS1125 camera module on VC I2C bus
```
6by9, line 2: SPDX licence descriptor missing. See https://github.com/raspberrypi/linux/blob/rpi-4.19.y/Documentation/process/license-rules.rst for guidance.

drivers/media/i2c/irs1125.c
```diff
@@ -0,0 +1,1137 @@
...
+#define IRS1125_REG_EFUSEVAL4      0xB0A1
+#define IRS1125_REG_DMEM_SHADOW    0xC320
+
+#define IRS1125_ROW_START_DEF       0
+#define IRS1125_COLUMN_START_DEF    0
+#define IRS1125_WINDOW_HEIGHT_DEF    288
+#define IRS1125_WINDOW_WIDTH_DEF    352
+
+struct irs1125 {
+    int power_count;
```
6by9: Coding style: Indentation must be with tabs, and tabs should equate to 8 spaces.

drivers/media/i2c/irs1125.c
```diff
@@ -0,0 +1,1137 @@
...
+}
+
+static struct regval_list irs1125_26MHz[] = {
+    {0xB017, 0x0413},
+    {0xB086, 0x3535},
+    {0xB0AE, 0xEF02},
+    {0xA000, 0x0004},
+    {0xFFFF, 100}, //sleep 100 ms
+
+    {0xB062, 0x6383},  // DPHYPLLCFG1 - 26 MHz
```
6by9: Coding style: Comments should use /* */, and not C99 //

drivers/media/i2c/irs1125.c
```diff
@@ -0,0 +1,1137 @@
...
+#include <linux/of_graph.h>
+#include <linux/slab.h>
+#include <linux/videodev2.h>
+#include <media/v4l2-device.h>
+#include <media/v4l2-fwnode.h>
+#include <media/v4l2-image-sizes.h>
+#include <media/v4l2-mediabus.h>
+#include <media/v4l2-ctrls.h>
+
+#define CHECK_BIT(val,pos) ((val) & (1<<(pos)))
```
6by9: The kernel already has a macro for 1<<pos as ```BIT(pos)```.
This macro is also only used once, therefore not really worth defining.

drivers/media/i2c/irs1125.c
```diff
@@ -0,0 +1,1137 @@
...
+    struct v4l2_mbus_framefmt format;
+    unsigned int width;
+    unsigned int height;
+    struct clk *xclk;
+    struct gpio_desc *reset;
+    unsigned int flags;
+    unsigned int wr_addr;
+    unsigned int num_seq;
+    struct v4l2_ctrl *ctrl_numseq, *ctrl_modplls;
+    struct v4l2_rect crop_rect;
```
6by9: Not used.

drivers/media/i2c/irs1125.c
```diff
@@ -0,0 +1,1137 @@
...
+    struct v4l2_ctrl_handler ctrl_handler;
+    struct media_pad pad;
+    struct mutex lock;
+    struct v4l2_mbus_framefmt format;
+    unsigned int width;
+    unsigned int height;
+    struct clk *xclk;
+    struct gpio_desc *reset;
+    unsigned int flags;
+    unsigned int wr_addr;
```
6by9: Not used.

drivers/media/i2c/irs1125.c
```diff
@@ -0,0 +1,1137 @@
...
+        if (CHECK_BIT(rdval, 14) == 0)
+            break;
+            
+        if (cnt >= 5) {
+            dev_err(&client->dev, "timeout waiting for 0xC40F\n");
+            return -EAGAIN;
+        }
+            
+        cnt++;
+        msleep(100);
```
6by9: Magic sleeps and retries feels a little nasty.
What are we actually waiting on here? The sensor has a ready bit that you're polling? Is there not a defined time in which it will initialise within?

drivers/media/i2c/irs1125.c
```diff
@@ -0,0 +1,1137 @@
...
+    return 0;
+}
+
+static int irs1125_set_get_fmt(struct v4l2_subdev *sd,
+                  struct v4l2_subdev_pad_config *cfg,
+                  struct v4l2_subdev_format *format)
+{
+    struct v4l2_mbus_framefmt *fmt = &format->format;
+    struct irs1125 *irs1125 = to_state(sd);
+    //struct i2c_client *client = v4l2_get_subdevdata(sd);
```
6by9: Please remove commented out code.

drivers/media/i2c/irs1125.c
```diff
@@ -0,0 +1,1137 @@
...
+                err = irs1125_write(&(dev->sd), 0xA892, 0x377);
+            else
+                err = irs1125_write(&(dev->sd), 0xA892, 0x355);
+            break;
+        default:
+            break;
+    }
+    
+    if (err < 0)
+    {
```
6by9: Unnecessary braces

drivers/media/i2c/irs1125.c
```diff
@@ -0,0 +1,1137 @@
...
+        dev_err(&client->dev, "Error executing control ID: %d, val %d, err %d", ctrl->id, ctrl->val, err);
+    }
+    else
+        err = 0;
+    
+    return err;
+}
+
+static int irs1125_g_volatile_ctrl(struct v4l2_ctrl *ctrl)
```
6by9: None of the controls are volatile, therefore we don't need g_volatile_ctrl at all.

drivers/media/i2c/irs1125.c
```diff
@@ -0,0 +1,1137 @@
...
+        .ops = &irs1125_ctrl_ops,
+        .id = IRS1125_CID_NUM_SEQS,
+        .name = "Change number of sequences",
+        .type = V4L2_CTRL_TYPE_INTEGER,
+        .flags = V4L2_CTRL_FLAG_MODIFY_LAYOUT,
+        .min = 1,
+        .max = 20,
+        .step = 1,
+        .def = 5,
+    },
```
6by9: mainline prefer
```
}, {
```
all on one line.

drivers/media/i2c/irs1125.c
```diff
@@ -0,0 +1,1137 @@
...
+}
+
+static int irs1125_open(struct v4l2_subdev *sd, struct v4l2_subdev_fh *fh)
+{
+    struct v4l2_mbus_framefmt *format =
+                v4l2_subdev_get_try_format(sd, fh->pad, 0);
+    struct v4l2_rect *crop =
+                v4l2_subdev_get_try_crop(sd, fh->pad, 0);
+
+    crop->left = IRS1125_COLUMN_START_DEF;
```
6by9: You don't do anything with crop, so remove it.

drivers/media/i2c/irs1125.c
```diff
@@ -0,0 +1,1137 @@
...
+    int ret;
+
+    ep = of_graph_get_next_endpoint(np, NULL);
+    if (!ep)
+        return -EINVAL;
+
+    ret = v4l2_fwnode_endpoint_parse(of_fwnode_handle(ep), &bus_cfg);
+
+    if (!ret)
+        sensor->flags = bus_cfg.bus.mipi_csi2.flags;
```
6by9: sensor->flags appears to be never used, so remove it. I suspect that means you can probably remove the whole of this function.

Having said that, you should probably check the number of CSI2 lanes configured in DT and ensure that it matches your register set.

drivers/media/i2c/irs1125.c
```diff
@@ -0,0 +1,1137 @@
...
+    struct v4l2_subdev *sd;
+    struct device_node *np = client->dev.of_node;
+    u32 xclk_freq;
+    int gpio_num;
+
+    sensor = devm_kzalloc(dev, sizeof(*sensor), GFP_KERNEL);
+    if (!sensor)
+        return -ENOMEM;
+
+    if (IS_ENABLED(CONFIG_OF) && np) {
```
6by9: Do you support configuration without device tree? I suspect not, therefore checking CONFIG_OF is selected is redundant.

I think ov5647 has been a bad example in this case. I merged an imx219 driver about a month ago which is hopefully a better example.

drivers/media/i2c/irs1125.c
```diff
@@ -0,0 +1,1137 @@
...
+
+static const struct i2c_device_id irs1125_id[] = {
+    { "irs1125", 0 },
+    { }
+};
+MODULE_DEVICE_TABLE(i2c, irs1125_id);
+
+#if IS_ENABLED(CONFIG_OF)
+static const struct of_device_id irs1125_of_match[] = {
+    { .compatible = "irs1125" },
```
6by9: The compatible string should always be "vendor,model". eg "ovti,ov5647,ov5647", or "sony,imx219".
As defined in https://github.com/raspberrypi/linux/blob/rpi-4.19.y/Documentation/devicetree/bindings/vendor-prefixes.txt#L175, Infineon Technologies is "infineon"

drivers/media/i2c/irs1125.c
```diff
@@ -0,0 +1,1137 @@
...
+        .name    = SENSOR_NAME,
+    },
+    .probe        = irs1125_probe,
+    .remove        = irs1125_remove,
+    .id_table    = irs1125_id,
+};
+
+static int __init irs1125_init(void)
+{
+    return i2c_add_driver(&irs1125_driver);
```
6by9: This looks very wrong. Module init and exit functions shouldn't be required if configured via device tree. The magic of DT should have registered things already.

drivers/media/i2c/irs1125_regs.h
```diff
@@ -0,0 +1,4126 @@
...
+#define _IRS1125_FIRMWARE_H_
+
+#include <linux/types.h>
+
+struct regval_list {
+	u16 addr;
+	u16 data;
+};
+
+struct regval_list irs1125_alternate_fw[] = {
```
6by9: Wowsers, 4kB of firmware as a list of register writes.

There are firmware loading mechanisms that may be better suited to doing this with a binary firmware file that typically lives in /lib/firmware. Your driver loads it via request_firmware(). See the majority of the DVB drivers for examples (one example at random https://github.com/raspberrypi/linux/blob/rpi-4.19.y/drivers/media/dvb-frontends/tda10071.c#L852).

Doing it this way also decouples the firmware from the driver. Admittedly that may be a bad thing should there be a hard dependency between the two.

If this does need to remain here then it definitely needs to gain a ```const``` to allow the compiler to optimise it.

6by9 (changes requested):
This needs to be split into multiple patches for the different bits of the system.
1) bcm2835-unicam, with reasoning for the change.
2) Add IRS1125 driver
3) Add IRS1125 to defconfigs
4) Add IRS1125 overlays

Kernel coding is against the kernel coding style docs (https://github.com/raspberrypi/linux/tree/rpi-4.19.y/Documentation/process). There is a script at scripts/checkpatch.pl that will check your patch for places where the coding style hasn't been followed. When run with the --strict flag it throws up 474 errors and 5032 warnings with your patch. Many of those will be whitespace issues that can be fixed quickly.
You also need to check licence headers on several files, and the patches need a Signed-off-by line (use ```git commit -s``` for git to do it automatically).

The change to bcm2835-unicam.c needs a touch of finessing. I believe you're trying to unpack the CSI2 raw 12 data format into the 16bpp V4L2_PIX_FMT_Y12 format. I think I'd prefer to see a new V4L2_PIX_FMT_Y12P format added in the same format as V4L2_PIX_FMT_Y10P, but that is a load of faffing.

I do need to revisit the driver to determine the correct format for repacked 16bpp data. I think the better way is to add a 2nd fourcc for the repacked format as we will only ever repack to 16bpp. It needs to support having the base fourcc being 0 for your case, and the alternate fourcc being 0 for cases such as RGB and YUYV where repacking is not applicable. Doing that should be relatively straightforward, so I'll see what I can sort quickly so you can drop your changes.

6by9:
FWIW I've pushed a quick set of mods to the Unicam driver that should support advertising all permutations of V4L2 formats based on unpacking. It's not quite right as requesting an unsupported format doesn't do the right thing, but it's close.
https://github.com/6by9/linux/tree/unicam-repacking

And I've fallen into the same trap of doing too many things at once - the same patch removes the caching of active formats and adds support for repacking. They should have been two patches, so the branch will get updated.

6by9:
https://github.com/6by9/linux/tree/unicam-repacking is updated to add what I hope is the required unpacking options for your code. It's taken a couple of attempts to untangle my patches (you have to "love" git), but I think they're all good. 

I'll create a PR with RFC (Request For Comments) tag against it, but will wait until you confirm that it works with your sensor before merging it.

6by9:
PR #3265 created for the Unicam changes.

drivers/media/i2c/irs1125_regs.h
Coimbra1984: That was a requirement from infineon. Of course this doesn't make sense for a kernel source file. I removed this file entirely and replaced it by require_firmware as you suggested

arch/arm/boot/dts/overlays/irs1125-overlay.dts
Coimbra1984: Thanks for the link. Indeed I had no idea of this, should be fixed now

drivers/media/i2c/irs1125.c
Coimbra1984: okay, I run checkpatch --strict and removed all errors/warnings etc, should be fixed now

drivers/media/i2c/irs1125.c
Coimbra1984: done

drivers/media/i2c/irs1125.c
Coimbra1984: I replaced it by BIT(pos)

drivers/media/i2c/irs1125.c
Coimbra1984: Removed, I also added proper configuration of CSI2 data lanes

drivers/media/i2c/irs1125.c
Coimbra1984: Okay, I followed the example of imx219. Should be better now

drivers/media/i2c/irs1125.c
Coimbra1984: Thanks for the hint. I head no idea about the vendor,model string. Fixed now

drivers/media/i2c/irs1125.c
Coimbra1984: Fixed

drivers/media/i2c/irs1125_regs.h
Coimbra1984: Cool, I did not know that. It's replaced now by require_firmware

drivers/media/i2c/irs1125.h
```diff
@@ -0,0 +1,71 @@
+/*
```
Coimbra1984, line 2: I'm not sure about this file. I use this one also from user space (currently I just copy it into my user space project), but as far as I remember there is also a location, maybe include/video to put these header files which are shared between kernel and user space?

Coimbra1984:
> This needs to be split into multiple patches for the different bits of the system.
> 
> 1. bcm2835-unicam, with reasoning for the change.
> 2. Add IRS1125 driver
> 3. Add IRS1125 to defconfigs
> 4. Add IRS1125 overlays

Okay, I'll do that. I'll need a bit for that as I'm only familiar with the basic git commands (I guess I need this one: https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History)

drivers/media/i2c/irs1125.h
6by9: /include/uapi/linux
Once "published" then uapi defines are pretty much set in stone to avoid regressions, so ensure you've got it right before you push it to there.

That's the correct way, but you'll often find that applications end up having to include the latest and greatest headers from the point at which they were built in order to avoid a #ifdef hell trying to remove code that relies on it.

drivers/media/i2c/irs1125.c
Coimbra1984: removed

drivers/media/i2c/irs1125.c
Coimbra1984: removed, there were a few more unused variables

drivers/media/i2c/irs1125.c
Coimbra1984: removed

drivers/media/i2c/irs1125.c
Coimbra1984: Yes, you are right. The datasheet gives a maximum of 390µs between releasing the reset pin and having a "ready" in this register. So I just removed this sleep

drivers/media/i2c/irs1125.c
Coimbra1984: Yes, all this stuff is removed now after running checkpatch --strict (which became my new friend)

drivers/media/i2c/irs1125.c
Coimbra1984: removed

drivers/media/i2c/irs1125.c
Coimbra1984: okay, done

drivers/media/i2c/irs1125.c
Coimbra1984: removed

drivers/media/i2c/irs1125_regs.h
6by9: That's the other advantage of request_firmware - the firmware binary blobs don't need to be under GPLv2.

It does make it more entertaining getting it upstreamed. The firmware and driver go via different paths, and the driver can't be merged until the firmware has been accepted. Coordinating it all can be fun.

drivers/media/i2c/irs1125.h
Coimbra1984: Okay, than I definitely will leave it in drivers/media/i2c
We are only using a subset of the imager's possibilities so far.

6by9:
> 
> 
> > This needs to be split into multiple patches for the different bits of the system.
> > 
> > 1. bcm2835-unicam, with reasoning for the change.
> > 2. Add IRS1125 driver
> > 3. Add IRS1125 to defconfigs
> > 4. Add IRS1125 overlays
> 
> Okay, I'll do that. I'll need a bit for that as I'm only familiar with the basic git commands (I guess I need this one: https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History)

It can be a tricky one, but in this case I suggest you:
- ```git fetch origin```, ```git rebase``` to rebase your branch against the top our tree.
- Note the commit hash for the top of your branch (currently 191466821501b23c168f025da35b0acdfeb49aa9, but you've made changes and will have just rebased).
- Check out a clean branch of rpi-4.19.y so that you retain your existing branch as a backup.
- Use ```git checkout 191466821501b23c168f025da35b0acdfeb49aa9 drivers/media/i2c``` (replace 191466821501b23c168f025da35b0acdfeb49aa9 with the hash you noted earlier). This should pick up and stage all the changes to that directory on your branch. (You may need to repeat with include/uapi too if you've moved your header).
- ```git commit -s``` to commit those staged changes. Use an appropriate commit text. The first line should be something like "media: i2c: Add driver for Infineon IRS1125 Time of flight depth sensor".
- Use ```git checkout 191466821501b23c168f025da35b0acdfeb49aa9 arch/arm/configs``` to pick up your config changes.
- ```git commit -s``` to commit those staged changes. Again use a sensible commit text.

Repeat for the remainig requested patch splits.

If you've committed them in the wrong order, then use ```git rebase -i``` to reorder the commits. (Rearrange the order of the lines in the editor that it opens).

I've missed one patch that will be required for upstream - a device tree binding for how to load and configure the driver. See https://github.com/raspberrypi/linux/commit/fd5109b991494f864c118dac69013cd878bee83a for the IMX219 binding patch (not that I've upstreamed that yet).
Device tree bindings are also considered to be set in stone once published. Changes must always be backwards compatible (which can be a right pain in the neck).

Coimbra1984:
> It can be a tricky one, but in this case I suggest you:

Okay, maybe it's better in this case that you first look whether the content is good, and after that I make clean commits

Documentation/devicetree/bindings/media/i2c/irs1125.txt
```diff
@@ -0,0 +1,49 @@
+* Infineon irs1125 time of flight sensor
+
+The Infineon irs1125 is a time of flight digital image sensor with
+an active array size of 352H x 286V. It is programmable through I2C
+interface. The I2C address defaults to 0x3D, but can be set to 0x3C or
+none standard mipi address 0x41.
```
6by9: non standard.
What do you mean by a mipi address?

arch/arm/boot/dts/overlays/Makefile
```diff
@@ -104,6 +104,7 @@ dtbo-$(CONFIG_ARCH_BCM2835) += \
 	mpu6050.dtbo \
 	mz61581.dtbo \
 	ov5647.dtbo \
+	irs1125.dtbo \
```
6by9: Sorry, you'll get a request from Phil to keep all of these in alphabetical order.

6by9: There should also be an entry in README to describe what the overlay does.

arch/arm/configs/bcm2709_defconfig
```diff
@@ -919,6 +919,7 @@ CONFIG_VIDEO_TW9906=m
 CONFIG_VIDEO_IMX219=m
 CONFIG_VIDEO_OV5647=m
 CONFIG_VIDEO_OV7640=m
+CONFIG_VIDEO_IRS1125=m
```
6by9, line 922: Again, Phil will ask that this has been set in the right place via ```make savedefconfig```. Is this in the right place, or is this a manual edit?

drivers/media/i2c/irs1125.c
```diff
@@ -0,0 +1,1115 @@
...
+
+	int power_count;
+};
+
+static inline struct irs1125 *to_state(struct v4l2_subdev *sd)
+{
+	return container_of(sd, struct irs1125, sd);
+}
+
+static struct regval_list irs1125_26MHz[] = {
```
6by9, line 90: 26MHz? Your dt binding listed 24MHz.

drivers/media/i2c/irs1125.c
```diff
@@ -0,0 +1,1115 @@
...
+
+		ret = __sensor_init(sd);
+		if (ret < 0) {
+			clk_disable_unprepare(irs1125->xclk);
+			dev_err(&client->dev,
+				"Camera not available, check Power\n");
+			goto out;
+		}
+	}
+
```
6by9, line 449: This may be my omission from imx219, but do we not have power down code?

drivers/media/i2c/irs1125.c
```diff
@@ -0,0 +1,1115 @@
...
+	/* get system clock (xclk) */
+	sensor->xclk = devm_clk_get(dev, NULL);
+	if (IS_ERR(sensor->xclk)) {
+		dev_err(dev, "could not get xclk");
+		return PTR_ERR(sensor->xclk);
+	}
+
+	xclk_freq = clk_get_rate(sensor->xclk);
+	if (xclk_freq != 25000000) {
```
6by9: So the binding gives 24MHz in the example, the register array says 26MHz, and this now checks against 25MHz. Which one should it be?

drivers/media/i2c/irs1125.c
```diff
@@ -0,0 +1,1115 @@
...
+	},
+	.probe		= irs1125_probe,
+	.remove		= irs1125_remove,
+	.id_table	= irs1125_id,
+};
+
+module_i2c_driver(irs1125_driver);
+
+MODULE_AUTHOR("Markus Proeller <markus.proeller@pieye.org>");
+MODULE_DESCRIPTION("A low-level driver for Infineon irs1125 sensors");
```
6by9: All drivers are generally low level.

6by9 (commented):
A few minor comments, but looking pretty good.
It still needs to be broken out into independent patches though.

drivers/media/i2c/irs1125.c
Coimbra1984: Okay, sorry my mistake. These are actually copy-paste artifacts. We have a reference clock of 26MHz on our hardware, which is also recommended by infineon.
To be honest, I did not fully understand the device-tree entry. But now having read fixed-clock.txt and clock-bindings.txt these entries describe an external clock "producer".
So in my case it must be a fixed-clock with frequency 26000000 and clock-cells=0 (only one clock output)

6by9: It is a bit obscure, and DT is a bit of a steep learning curve, but yes it is defining the clock frequency of the oscillator on your board.

In theory that clock could be derived from any of a number of places, and the driver could cope with a number of source clock frequencies, adjusting PLLs appropriately.
If Infineon recommend 26MHz, then it is reasonable for the driver to only work with that one value, but it shouldn't be reflected in the binding as that only describes the potential configuration options.

arch/arm/boot/dts/overlays/Makefile
Coimbra1984: okay, I changed that. Having seen the entires in README, I changed my dtparams to contain i2c_pins_0_1 and ic2_pins_28_29 just like imx219 and ov5647

Documentation/devicetree/bindings/media/i2c/irs1125.txt
Coimbra1984: I mean mipi csi-2. I changed this description a bit. Mipi CSI-2 suggests these I2C addresses:

6.2 CCI Slave Addresses

For camera modules having only raw Bayer output the 7-bit slave address should be 011011Xb, where X =
0 or 1. For all other camera modules the 7-bit slave address should be 011110Xb

drivers/media/i2c/irs1125.c
Coimbra1984: okay, changed

drivers/media/i2c/irs1125.c
Coimbra1984: Okay, I can definitly set the reset pin here. This is also the recommended way in the user manual to save power. I'll change and test it.

arch/arm/configs/bcm2709_defconfig
Coimbra1984: Manual edit :) I'll change it...

pelwell: Good, because I'm very strict. ;-)

Coimbra1984: @pelwell I fully understand.

That's strange I end up on the same line. So what I did is:
1. Remove irs1125 entry from bcm2709_defconfig
2. make ARCH=arm CROSS_COMPILE=arm-linux-gnueabihf- bcm2709_defconfig
3. vim .config -> there was a line # CONFIG_VIDEO_IRS1125 is not set
4. changed that line to CONFIG_VIDEO_IRS1125=m
5. make ARCH=arm CROSS_COMPILE=arm-linux-gnueabihf- savedefconfig
6. vim defconfig and search for IRS1125 which is again on line 922:

 920 CONFIG_VIDEO_OV5647=m
 921 CONFIG_VIDEO_OV7640=m
 922 CONFIG_VIDEO_IRS1125=m
 923 CONFIG_VIDEO_MT9V011=m
 924 CONFIG_DRM=m

Is this the workflow I should follow?

pelwell: Yes - there are other ways to manipulate the .config file (see scripts/config), but the savedefconfig is the important part. You either have excellent instincts or it was a lucky guess.

Coimbra1984: I guess it's a sign to play lottery today

Documentation/devicetree/bindings/media/i2c/irs1125.txt
6by9: Wow, never noticed that bit in the CSI spec, and rarely seen it followed! (Having said that, OV5647 is on 0x36, which is 011011Xb)

TBH I'd drop reference to non-standard MIPI addresses. Specify the I2C addresses that the module can be configured to, and leave it at that. 
"The I2C address will be one of 0x3C, 0x3D, or 0x41, selected by <some mechanism>."
You'll get more detailed feedback on the DT bindings should you try and upstream them.

Coimbra1984: Okay, maybe we have another problem here. The default address after power on is 0x3D, you can than change the address to be either 0x3C, 0x3D or 0x41 by writing to a register. Would it be valid to support only 0x3D or how should that one be modeled?

6by9: Joys! It's one of them.
I seem to recall one other sensor driver does support dynamic changing of the I2C address because people were wanting to use it for stereoscopic capture, but it gets very messy in doing so.

Yes, supporting only 0x3D in the driver is valid.  DT describes what the hardware can potentially do. Drivers don't have to implement all of that (eg DT describes a clock source, but the driver can reject it should the clock not be the one it wanted).

"The I2C address defaults to 0x3D, but can be reconfigured to address 0x3C or 0x41 via I2C commands"

drivers/media/i2c/irs1125.c
6by9: line 444 ```} else if (!on && irs1125->power_count == 1) {```
otherwise the first call to turn the sensor off will always turn it off, even if multiple power on calls have been made.

ON/ON/OFF/ON will currently leave the sensor OFF, because the last ON call will have failed line 427 as !irs1125->power_count is false.

drivers/media/i2c/irs1125.c
```diff
@@ -0,0 +1,1117 @@
...
+#endif
+
+static struct i2c_driver irs1125_driver = {
+	.driver = {
+		.of_match_table = of_match_ptr(irs1125_of_match),
+		.name	 = SENSOR_NAME,
+	},
+	.probe		= irs1125_probe,
+	.remove		= irs1125_remove,
+	.id_table	= irs1125_id,
```
6by9: ```id_table``` shouldn't be required as loading is controlled by device tree, and hence ```struct i2c_device_id irs1125_id``` and ```MODULE_DEVICE_TABLE(i2c, irs1125_id);``` above can also be deleted.

6by9 (commented):
One final bit of code left lying around, and I think I'm good other than splitting up the patches. If you want then I'm happy to split them up for you, push to a branch of mine, and then you take ownership of them back (I'll give you some instructions). Having said that, the instructions I gave earlier in the review process should be relatively simple to follow.

If (hopefully when) you try to upstream it there will be a number of really annoying requests to meet their coding style. Things like local variables in a file being declared with the longest line at the top and getting shorter.
I lose track of them all, and they are local to the individual Linux subsystems (in this case linux-media). It's the point I often give up on my efforts because it has near zero real value :-/

6by9:
Thank you.
I think I'm happy with the overall content, so it's just breaking it up into standalone commits, as I commented at https://github.com/raspberrypi/linux/pull/3261#pullrequestreview-296241698. The offer is there if you want me to do it.

@pelwell Did you want to do a review of the overall change now, or once split?

pelwell:
Split, please - it's easier once the changes are separated.

Coimbra1984:
> I think I'm happy with the overall content, so it's just breaking it up into standalone commits, as I commented at [#3261 (review)](https://github.com/raspberrypi/linux/pull/3261#pullrequestreview-296241698). The offer is there if you want me to do it.

Yes please :) I tried to rebase and make them look like these commits: https://github.com/raspberrypi/linux/pull/3188/commits
but somehow I fail at the end when I try to push my rewritten history to my branch. So if you can do it, it will be much faster I guess

6by9:
OK, branch https://github.com/6by9/linux/tree/pieye should have all your changes from this PR, but under my name.
Hopefully the following will rewrite the history on them to show you as the author.

```
git remote add 6by9 https://github.com/6by9/linux
git remote fetch 6by9
git checkout -t -b split 6by9/pieye
```
Please check that the tree builds and works at this point. If it does then we can rewrite the history.
```
git rebase -i HEAD~3
```
This will present you with an editor looking something like
```
pick 3d51849 dt-bindings: Add binding for the Infineon IRS1125 sensor
pick 085fe80 media: i2c: Add a driver for the Infineon IRS1125 depth sensor
pick ffcbb30 configs: Add CONFIG_VIDEO_IRS1125 to the defconfigs
pick ab737c9 dtoverlays: Add an overlay for the Infineon IRS1125
```
Replace the word "pick" at the start of each line with "e" or "edit" (no quotes).
Save and exit the editor.
Git should now reset to each commit in turn and give you a chance to edit them. You want to run
```git commit --amend --reset-author -s```
on each one in turn to rewrite the commiter details. Then use ```git rebase --continue``` to step forward to the next commit. Repeat those two commands 4 times over for the 4 commits.

```git push -f pieye HEAD:rpi-4.19.y``` (replace pieye with whatever you called your remote) to force update your branch on github, and it should update this PR in the same way your other pushes have done.
NB This does destroy your working branch on github, but that shouldn't matter as we really care about the final result rather than how we got there.

**Comment by 6by9** [comment]
Of coure being git there have to be at least 3 ways to do the same thing.

It may be quicker/easier to do
```
git remote add 6by9 https://github.com/6by9/linux
git remote fetch 6by9
git checkout -t -b split origin/rpi-4.19.y
```
to get a clean Pi kernel tree.
Then cherrypick the relevant commits from my branch:
```
git cherry-pick 3d51849 
git commit --amend --reset-author -s
git cherry-pick 085fe80 
git commit --amend --reset-author -s
git cherry-pick ffcbb30 
git commit --amend --reset-author -s
git cherry-pick ab737c9 
git commit --amend --reset-author -s
```
and push  to github
```
git push -f pieye HEAD:rpi-4.19.y
```

**Comment by Coimbra1984** [comment]
Okay I have finally tested everything on a raspberry pi 3B+ and a raspberry pi 4. Works as expected 👍

**Comment by Coimbra1984** [comment]
> Yes please :) I tried to rebase and make them look like these commits: https://github.com/raspberrypi/linux/pull/3188/commits
> but somehow I fail at the end when I try to push my rewritten history to my branch. So if you can do it, it will be much faster I guess

I guess my mistake was a missing -f on git push ...

**Comment by Coimbra1984** [comment]
Hi @6by9 , just to make sure, there is currently no pending todo from my side, right?

**Comment by 6by9** [comment]
Sorry, hadn't checked your push, but it looks fine.
@pelwell All split out into nice commits now.

**Comment by Coimbra1984** [comment]
@pelwell  do you have any review comments?

File: arch/arm/boot/dts/overlays/irs1125-overlay.dts [hunk 28]
```diff
@@ -0,0 +1,97 @@
...
+	compatible = "brcm,bcm2835";
+
+	fragment@0 {
+		target = <&i2c_vc>;
+		__overlay__ {
+			#address-cells = <1>;
+			#size-cells = <0>;
+			status = "okay";
+
+			irs1125: irs1125@36 {
```
pelwell: The address (the part after the @) is supposed to match the `reg` property.

arch/arm/boot/dts/overlays/README
```diff
@@ -1346,6 +1346,17 @@ Params: card_name               Override the default, "IQAudIODigi", card name.
         dai_stream_name         Override the default, "IQAudIO Digi HiFi",
                                 dai stream name.
```
pelwell, line 1348: Sorry, but there must be two blank lines between entries.

arch/arm/boot/dts/overlays/irs1125-overlay.dts
```diff
@@ -0,0 +1,97 @@
...
+				compatible = "infineon,irs1125";
+				reg = <0x3D>;
+				status = "okay";
+
+				pwdn-gpios = <&expgpio 5 0>;
```
pelwell: The gpio values are going to be overwritten by the firmware so what I'm describing isn't a bug, but it would help my verification script if the default version resolves against all Pi DTBs (it currently won't apply for Pis without a GPIO expander), even if it won't work as written - replacing `&expgpio` with `&gpio` fixes it for me.

drivers/media/i2c/irs1125.c
```diff
@@ -0,0 +1,1110 @@
...
+		return -EINVAL;
+
+	code->code = MEDIA_BUS_FMT_Y12_1X12;
+
+	return 0;
+}
+
+static int irs1125_set_get_fmt(struct v4l2_subdev *sd,
+			       struct v4l2_subdev_pad_config *cfg,
+		struct v4l2_subdev_format *format)
```
pelwell: Indentation.

drivers/media/i2c/irs1125.c
```diff
@@ -0,0 +1,1110 @@
...
+static int irs1125_s_stream(struct v4l2_subdev *sd, int enable)
+{
+	if (enable)
+		return irs1125_stream_on(sd);
+	else
+		return irs1125_stream_off(sd);
+}
+
+static const struct v4l2_subdev_video_ops irs1125_subdev_video_ops = {
+	.s_stream =				 irs1125_s_stream,
```
pelwell: With nothing to align with, the extra spaces are just a distraction.

drivers/media/i2c/irs1125.c
```diff
@@ -0,0 +1,1110 @@
...
+		return PTR_ERR(sensor->reset);
+	}
+
+	gpio_num = desc_to_gpio(sensor->reset);
+
+	mutex_init(&sensor->lock);
+
+	ret = irs1125_ctrls_init(sensor, dev);
+	if (ret < 0)
+		goto error;
```
pelwell: Shouldn't this be `goto mutex_remove;`?

drivers/media/i2c/irs1125.c
```diff
@@ -0,0 +1,1110 @@
...
+	int ret;
+	struct i2c_client *client = v4l2_get_subdevdata(sd);
+
+	ret = irs1125_read(sd, IRS1125_REG_DESIGN_STEP, &read);
+	if (ret < 0) {
+		dev_err(&client->dev, "error reading from i2c\n");
+		return ret;
+	}
+
+	if (read != 0x0a12) {
```
pelwell: A macro constant named for what 0x0a12 represents would be nice, especially since the value is used twice.

pelwell (changes requested):
After @6by9 has been through the code I'm usually left with error handling and stylistic issues, and this was no exception. Fix these few minor niggles and I'll merge.

drivers/media/i2c/irs1125.c
Coimbra1984: okay, done

drivers/media/i2c/irs1125.c
Coimbra1984: you are right, I changed it

drivers/media/i2c/irs1125.c
Coimbra1984: Sure, removed.

drivers/media/i2c/irs1125.c
Coimbra1984: Corrected

arch/arm/boot/dts/overlays/irs1125-overlay.dts
Coimbra1984: okay, I changed this

arch/arm/boot/dts/overlays/README
Coimbra1984: ok, done.

arch/arm/boot/dts/overlays/irs1125-overlay.dts
Coimbra1984: changed to @3D

pelwell (approved):

PR #3264, opened 2019-10-03
KMS and FKMS updates for flips and YUV encodings

6by9:


drivers/gpu/drm/vc4/vc4_plane.c
```diff
@@ -760,9 +807,20 @@ static int vc4_plane_mode_set(struct drm_plane *plane,
...
 	/* Colorspace conversion words */
 	if (vc4_state->is_yuv) {
-		vc4_dlist_write(vc4_state, SCALER_CSC0_ITR_R_601_5);
-		vc4_dlist_write(vc4_state, SCALER_CSC1_ITR_R_601_5);
-		vc4_dlist_write(vc4_state, SCALER_CSC2_ITR_R_601_5);
+		enum drm_color_encoding color_encoding = state->color_encoding;
+		enum drm_color_range color_range = state->color_range;
+		const u32 *ccm;
+
+		if (color_encoding < DRM_COLOR_ENCODING_MAX)
```
popcornmix: Are these the right way round? ``>=`` here would make more sense to me.

6by9: Indeed.
DRM checks many values before the driver gets to see them, therefore checking error paths is difficult. The difference in the colourspaces is often minimal, so difficult to see.

6by9:
@pelwell Sorry, another one waiting for review.
popcornmix's observation has been fixed.

PR #3265, opened 2019-10-03
Unicam repacking

6by9:
Adds support to the Unicam driver to allow repacking the data out of the weird CSI2 formats into the bottom of 16bit words where the relevant V4L2 formats are defined.

Also drops the table of "active_formats" enumerated at probe for enumerating the formats from the sensor driver sensor every time. This solves the problem of Bayer sensors often changing Bayer order with H & V flips.

### DIFF
--- a/drivers/media/platform/bcm2835/bcm2835-unicam.c
+++ b/drivers/media/platform/bcm2835/bcm2835-unicam.c
@@ -273,10 +273,31 @@ static const struct unicam_fmt formats[] = {
 		.code		= MEDIA_BUS_FMT_SRGGB12_1X12,
 		.depth		= 12,
 		.csi_dt		= 0x2c,
+	}, {
+		.fourcc		= V4L2_PIX_FMT_SBGGR14P,
+		.code		= MEDIA_BUS_FMT_SBGGR14_1X14,
+		.depth		= 14,
+		.csi_dt		= 0x2d,
+	}, {
+		.fourcc		= V4L2_PIX_FMT_SGBRG14P,
+		.code		= MEDIA_BUS_FMT_SGBRG14_1X14,
+		.depth		= 14,
+		.csi_dt		= 0x2d,
+	}, {
+		.fourcc		= V4L2_PIX_FMT_SGRBG14P,
+		.code		= MEDIA_BUS_FMT_SGRBG14_1X14,
+		.depth		= 14,
+		.csi_dt		= 0x2d,
+	}, {
+		.fourcc		= V4L2_PIX_FMT_SRGGB14P,
+		.code		= MEDIA_BUS_FMT_SRGGB14_1X14,
+		.depth		= 14,
+		.csi_dt		= 0x2d,
 	},
 	/*
-	 * 14 and 16 bit Bayer formats could be supported, but there are no V4L2
-	 * defines for 14bit packed Bayer, and no CSI2 data_type for raw 16.
+	 * 16 bit Bayer formats could be supported, but there is no CSI2
+	 * data_type defined for raw 16, and no sensors that produce it at
+	 * present.
 	 */
 };
 
--- a/drivers/media/platform/bcm2835/bcm2835-unicam.c
+++ b/drivers/media/platform/bcm2835/bcm2835-unicam.c
@@ -307,12 +307,31 @@ static const struct unicam_fmt formats[] = {
 		.code		= MEDIA_BUS_FMT_SRGGB14_1X14,
 		.depth		= 14,
 		.csi_dt		= 0x2d,
-	},
+	}, {
 	/*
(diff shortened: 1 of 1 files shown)


6by9:
RFC tag added awaiting confirmation that this works with the Infineon IRS1125 driver in #3261 .

Coimbra1984:
> RFC tag added awaiting confirmation that this works with the Infineon IRS1125 driver in #3261 .

Sorry for the delay (but fifa20 is out :)
This PR works perfectly for me.

Just to double check. I only built module bcm2835-unicam of your forked kernel (branch unicam-repacking) and replaced the module on my 4.19.75er kernel on a raspberry pi 3.

6by9:
Thanks for the confirmation it works. RFC tag removed.

Yes, the patch only updates bcm835-unicam. As long as that was rebuilt correctly then changing the one module and rebooting should be sufficient.

Coimbra1984:
> Thanks for the confirmation it works. RFC tag removed.
> 
> Yes, the patch only updates bcm835-unicam. As long as that was rebuilt correctly then changing the one module and rebooting should be sufficient.

Yes it was definitely rebuilt and I also rebooted.

6by9:
@pelwell Any chance of a review?

drivers/media/platform/bcm2835/bcm2835-unicam.c
```diff
@@ -455,50 +453,33 @@ static int find_mbus_depth_by_code(u32 code)
...
-			   i, &dev->active_fmts[i], dev->active_fmts[i].code,
-			   V4L2_FOURCC_CONV_ARGS(dev->active_fmts[i].fourcc),
-			   dev->active_fmts[i].depth);
-	}
-}
-
 static inline unsigned int bytes_per_line(u32 width,
-					  const struct unicam_fmt *fmt)
+					  const struct unicam_fmt *fmt,
+					  u32 v4l2_fourcc)
```
pelwell, line 502: Is the extra parameter future-proofing or an abandoned idea? It doesn't seem to belong here.

drivers/media/platform/bcm2835/bcm2835-unicam.c
```diff
@@ -726,14 +708,37 @@ static int unicam_enum_fmt_vid_cap(struct file *file, void  *priv,
...
 	const struct unicam_fmt *fmt = NULL;
+	int index = 0;
+	int ret = 0;
+	int i;
 
-	if (f->index >= dev->num_active_fmt)
-		return -EINVAL;
 
-	fmt = &dev->active_fmts[f->index];
+	for (i = 0; !ret && i < 128 ; i++) {
```
pelwell, line 745: What's the significance of 128?

drivers/media/platform/bcm2835/bcm2835-unicam.c
```diff
@@ -726,14 +708,37 @@ static int unicam_enum_fmt_vid_cap(struct file *file, void  *priv,
 				   struct v4l2_fmtdesc *f)
 {
 	struct unicam_device *dev = video_drvdata(file);
+	struct v4l2_subdev_mbus_code_enum mbus_code;
 	const struct unicam_fmt *fmt = NULL;
+	int index = 0;
+	int ret = 0;
+	int i;
```
pelwell, line 739: Nuke the following blank line.

drivers/media/platform/bcm2835/bcm2835-unicam.c
6by9: In the wrong patch. Should have been in "media: bcm2835-unicam: Support unpacking CSI format to 16bpp" where the clause becomes 
```
if (v4l2_fourcc == fmt->repacked_fourcc)
		/* Repacking always goes to 16bpp */
		return ALIGN(width << 1, BPL_ALIGNMENT);
	else
		return ALIGN((width * fmt->depth) >> 3, BPL_ALIGNMENT);
```
Will fix.

drivers/media/platform/bcm2835/bcm2835-unicam.c
6by9: Hmm, how did checkpatch not complain on that one?!

drivers/media/platform/bcm2835/bcm2835-unicam.c
6by9: Just a failsafe to ensure a dodgy sensor driver doesn't leave us looping forever. I'll add a comment.

6by9:
Branch rebased, fixes applied, an force pushed

6by9:
There are no actual code changes, just moved where v4l2_fourcc was added to bytes_per_line to the right patch, and added a comment where the extra line was.

PR #3270, opened 2019-10-06
fix warnings for 5.3

teknoraver:


lategoodbye:
This pull request touches some upstream files (bcm2835-dma, vchiq).
Would be better to fix them upstream first.

teknoraver:
Hi @lategoodbye 

both patches fixes code which is not yet upstream.
It seems that such code was introduced in 00dfb1e4dcdf4f7160700ebe8b928c3cc952186e
for vchiq and dcee296a50fa9fb29cdfe462f4f1ac53ca578ced for bcm2835-dma by @pelwell

lategoodbye:
Okay, in this case only 66c291b25c8c783b4f5bd692c5bf99791446f8ca is left.

teknoraver:
> Okay, in this case only [66c291b](https://github.com/raspberrypi/linux/commit/66c291b25c8c783b4f5bd692c5bf99791446f8ca) is left.

Right. I've sent that patch upstream, I will rebase the PR without such patch.

thanks

pelwell:
Thanks.

PR #3275, opened 2019-10-09
bcm2711_defconfig: Disable CONFIG_BRCMDBG

MkfsSion:
* When CONFIG_BRCMDBG is set to y,the following messages are treated as
error which should be seen as info
* brcmfmac: brcmf_fw_alloc_request: using brcm/brcmfmac43455-sdio for
chip BCM4345/6
* ...

Signed-off-by: MkfsSion <mkfssion@mkfssion.com>


pelwell:
In other words, "I don't use this thing and it has an unwanted side effect so I'm going to disable it."

BRCMDBG is required to allow users to turn on various runtime logging features. If you don't like the implementation, send a patch upstream.

PR #3277, opened 2019-10-09
bcm2711: Retain support for old DTBs

pelwell:
The recent series switching to bcm2711 as the DT identifier broke Pis
running with old DTBs. Add some bcm2838 compatible strings as a
temporary measure, at least until the next full Raspbian image with
bcm2711 DTBs.


lategoodbye:
Thanks for fixing this mess.

In theory we could to try to fix the other way round by adding the old compatibles to bcm2838.dtsi (at least for clk).

PR #3280, opened 2019-10-11
(Proof of concept) Disable RPI4 Ethernet LEDs (4.19)

satmandu:
As per suggestion here: https://www.raspberrypi.org/forums/viewtopic.php?p=1540845#p1540845

Obviously this shouldn't be merged, but hopefully at some point a device tree entry for the bcmgenet/BCM54213PE could be fleshed out so that a config.txt option might be used to indicate that the LEDs should be turned off.

### DIFF
--- a/drivers/net/phy/broadcom.c
+++ b/drivers/net/phy/broadcom.c
@@ -61,9 +61,14 @@ static void bcm54213pe_config_init(struct phy_device *phydev)
 	val |= BCM54XX_SHD_LEDCTL_ACTLINK_EN;
 	bcm_phy_write_shadow(phydev, BCM54XX_SHD_LEDCTL, val);
 
-	/* Set ACTIVITY on LED "1" output, LINKSPD[1] on LED "3" output */
-	val = BCM5482_SHD_LEDS1_LED1(BCM_LED_SRC_ACTIVITYLED) |
-		BCM5482_SHD_LEDS1_LED3(BCM_LED_SRC_LINKSPD1);
+	/* Some day there may be a DT wired toggle for this. Disable for now.
+	 * Set ACTIVITY on LED "1" output, LINKSPD[1] on LED "3" output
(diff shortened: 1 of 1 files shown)


pelwell:
Yes, that's the place the changes will go once the DT properties are decided upon. We can then arrange to support the `eth_led0` and `eth_led1` parameters.

satmandu:
> Yes, that's the place the changes will go once the DT properties are decided upon.

While this change appears to work for 4.19 kernels, it appears NOT to work for  rpi-5.3.y kernels. Hmm.

KrisztianKende:
> While this change appears to work for 4.19 kernels, it appears NOT to work for rpi-5.3.y kernels. Hmm.

Check [commit 450895d](https://github.com/raspberrypi/linux/commit/450895d04ba13a96886eddfeddb11556ae8624f1). You need more changes in the kernel 5.3.

JamesH65:
In fact the change need to move to the area of code in the PR just above. I'll take a look  at this and see what DT entry would be appropriate - the code change is not onerous AFAICT, but probably need to be sent upstream first which always takes about eleventee hundred years.

satmandu:
Thanks all. I try keep the lights to a minimum from the pi sitting in the bedroom so the kiddo doesn't get woken up by blinkenlights.

JamesH65:
JFYI,   [commit 450895d](https://github.com/raspberrypi/linux/commit/450895d04ba13a96886eddfeddb11556ae8624f1) can be cherry-picked to 4.19 without problems.  Best to work on that if doing anything.

JamesH65:
@pelwell I have taken a look at other drivers, there does not seem to be any standard DT strings for adjusting the LED's. The only obvious use case is simply to turn them off, so there could be a simple string for that. Presumably if we do anything it needs to go upstream to netdev and then ported back down?

JamesH65:
See #3316 for a DT implementation.

JamesH65:
Please update to the latest kernel which may contain a fix for this issue.

PR #3281, opened 2019-10-11
Update README

thekapilkhandelwal:
Updated readme


pelwell:
I appreciate your effort to explain the Linux make system, but this isn't the best place to do it. This README is unmodified from the Linux version, and we try to minimise the changes we make to "upstream" files.

A better option would be to add something to our documentation repo, which hosts the source for some of our website. The kernel build process is covered [here](https://github.com/raspberrypi/documentation/blob/master/linux/kernel/building.md), with a section on configuration [here](https://github.com/raspberrypi/documentation/blob/master/linux/kernel/configuring.md) - we'd be happy to take a Pull Request to either (or both).

PR #3289, opened 2019-10-16
drm:vc4 Added calls for firmware display blank/unblank

JamesH65:
Requires new display power mailbox call to be present.

### DIFF
--- a/drivers/gpu/drm/vc4/vc4_firmware_kms.c
+++ b/drivers/gpu/drm/vc4/vc4_firmware_kms.c
@@ -91,6 +91,12 @@ struct mailbox_blank_display {
 	u32 blank;
 };
 
+struct mailbox_display_pwr {
+	struct rpi_firmware_property_tag_header tag1;
+	u32 display;
+	u32 state;
+};
+
 struct mailbox_get_edid {
 	struct rpi_firmware_property_tag_header tag1;
 	u32 block;
@@ -272,6 +278,7 @@ struct vc4_fkms_encoder {
 	struct drm_encoder base;
 	bool hdmi_monitor;
 	bool rgb_range_selectable;
+	int display_num;
 };
 
 static inline struct vc4_fkms_encoder *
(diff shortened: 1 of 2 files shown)


JamesH65:
Fixes #3050

drivers/gpu/drm/vc4/vc4_firmware_kms.c
```diff
@@ -1613,13 +1620,29 @@ static const struct drm_encoder_funcs vc4_fkms_encoder_funcs = {
...
 
+static void vc4_fkms_display_power(struct drm_encoder *encoder, bool power)
+{
+	struct vc4_fkms_encoder *vc4_encoder = to_vc4_fkms_encoder(encoder);
+	struct vc4_dev *vc4 = to_vc4_dev(encoder->dev);
+
+	struct mailbox_display_pwr pwr = {
+		.tag1 = {RPI_FIRMWARE_SET_DISPLAY_POWER, 8, 0, },
+		.display = vc4_encoder->display_num,
+		.state = power ? 1 : 0
```
pelwell: It's an annoying nit-pick, but adding a comma at the end of the line would match all the other properties and the normal Linux style (the intention being to minimise inter-line dependencies).

drivers/gpu/drm/vc4/vc4_firmware_kms.c
```diff
@@ -272,6 +278,7 @@ struct vc4_fkms_encoder {
 	struct drm_encoder base;
 	bool hdmi_monitor;
 	bool rgb_range_selectable;
+	int display_num;
```
6by9, line 281: I feel that this should be accessible from either the crtc or connector (the bits either side of the encoder).
In reality I get lost with the potential dereferencing between the structures, and this driver is fairly monolithic anyway.

JamesH65: I did spend about 45 minutes trying to find a way of getting this information from the existing structures, and it ended up being so convoluted that adding this parameter was a hugely simpler option - both from an implementation and understanding point of view.

6by9: Fair enough. I know that finding the right linkings can be less than obvious.

JamesH65:
@popcornmix This can be merged now that that required mailbox command in the firmware has been merged, unless you want a short delay to wait for it to go out in rpi-update.

PR #3303, opened 2019-10-28
build,ci: add minimal Travis-CI integration

commodo:
This change adds a minimal build via Travis-CI. Lately, when
upstreaming patches, we get a lot of reports back (from Kernel-CI and
Intel's lkp bot) about [valid] faults in the code we submit. A lot of them
are caught by newer GCC compilers (GCC 8 & 9).

We also have a branch with rpi-4.19.y, on top of which we add our own
patches:
  https://github.com/analogdevicesinc/linux/tree/rpi-4.19.y

When enabling warnings-as-errors, some of the warnings [caught by GCC] are
not our own, and on the rpi-4.x.y branches the code differs from upstream.

This proposal, adds Travis-CI build with the newest Ubuntu (currently
19.10) docker image; but, since it's the rolling version, this should
update as time goes by, and should have a pretty new GCC compiler.
Travis-CI is a popular CI integration on Github, and helps with
build-visibility.

The intent is to have these warnings handled by the RPi project, so that we
don't have to patch too much downstream. Many of the warnings [that GCC
reports] are reasonable.

This change is an adaption from the scripts we use here:
  https://github.com/analogdevicesinc/linux/tree/master/ci/travis
  https://github.com/analogdevicesinc/libiio/blob/master/CI/travis/lib.sh

Note that this change adds the build stuff. It doesn't do any fixes to the
warnings. That is subject to another set of discussions/proposals, should
this be accepted.

Signed-off-by: Alexandru Ardelean <alexandru.ardelean@analog.com>

### DIFF
--- a/ci/travis/run-build.sh
+++ b/ci/travis/run-build.sh
@@ -1,0 +1,68 @@
+#!/bin/bash
+set -e
+
+TRAVIS_BUILD_DIR="${TRAVIS_BUILD_DIR:-.}"
+
+# cd to docker build dir if it exists
+if [ -d /docker_build_dir ] ; then
+	cd /docker_build_dir
+	TRAVIS_BUILD_DIR="/docker_build_dir"
+fi
+
+. ./ci/travis/lib.sh
+
+if [ -f "${TRAVIS_BUILD_DIR}/env" ] ; then
+	echo_blue "Loading environment variables"
+	cat "${TRAVIS_BUILD_DIR}/env"
+	. "${TRAVIS_BUILD_DIR}/env"
+fi
+
+KCFLAGS="-Werror"
+export KCFLAGS
+
+APT_LIST="build-essential bc u-boot-tools flex bison libssl-dev"
+
+if [ "$ARCH" == "arm64" ] ; then
+	APT_LIST="$APT_LIST gcc-aarch64-linux-gnu"
+else
+	APT_LIST="$APT_LIST gcc-arm-linux-gnueabihf"
+fi
+
+adjust_kcflags_against_gcc() {
+	# FIXME: remove this function once kernel gets upgrade and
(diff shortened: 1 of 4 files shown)


commodo:
Builds for these rules are here: 
* https://travis-ci.org/commodo/linux-1/builds/
* https://travis-ci.org/commodo/linux-1/jobs/603811867

commodo:
Changelog v1 -> v2:
* some GCC versions print the entire version (e.g. 8.3.0) while some only print the major (like Ubuntu's); for future-proofing, cut the major version when checking

ci/travis/run-build.sh
```diff
@@ -0,0 +1,68 @@
...
+	APT_LIST="$APT_LIST gcc-arm-linux-gnueabihf"
+fi
+
+adjust_kcflags_against_gcc() {
+	# FIXME: remove this function once kernel gets upgrade and
+	#        these have been fixed upstream. Currently, these are
+	#	 for kernel 4.19
+	GCC="${CROSS_COMPILE}gcc"
+	if [ "$($GCC -dumpversion | cut -d. -f1)" -ge "8" ]; then
+		KCFLAGS="$KCFLAGS -Wno-error=stringop-truncation"
```
commodo, line 37: with respect to these warnings I would like to add that we've been using an older base of rpi-4.19.y, so it's likely that some of them were fixed in upstream 4.19.y or some of them were fixed in the rpi-4.19.y; 

i would probably need to investigate which are still valid and which can be removed; i could just remove all of them and just let `-Werror`

commodo:
Changelog v2 -> v3:
* minor whitespace/indentation correction

commodo:
not gonna do Travis-CI anymore;
they're charging money now for open-source repos AFAICT

PR #3307, opened 2019-10-29
Enable hwmon for thermal zones

wens:
It would be nice to be able to use standard hwmon interfaces to read the temperature of the CPU thermal zone. This is accomplished by enabling `CONFIG_THERMAL_HWMON`, and also having the `brcmstb_thermal` driver add the hwmon sysfs entries. The older bcm2835_thermal driver already has support for this.

The thermal driver patch does not directly apply to upstream, which has seen a change to `devm_thermal_zone_of_sensor_register()`, removing the `remove()` driver callback.


popcornmix:
The change to brcmstb_thermal.c should be submitted upstream first.
Assuming they are happy then we can apply it here (even if that requires a little fixing up).
That way when we bump to a newer kernel version we should be able to have this feature without a downstream patch.

wens:
Ah. Ok. I'll see what I can do, given the current state of RPi 4 upstream.

wens:
Patch submitted upstream: https://lkml.org/lkml/2019/10/30/767
I suggest also backporting `adad7c7d6a0c ("thermal: broadcom: use devm_thermal_zone_of_sensor_register")` with it, once its in of course.

popcornmix:
Thanks. Can you ping here when it is merged upstream.

wens:
So it looks like the upstream brcmstb_thermal driver doesn't actually support BCM2711, as it targets a slightly different hardware block.

@lategoodbye posted a new BCM2711 thermal driver [1] a few days ago. We'll see how that goes, or if it converges on the existing brcmstb_thermal driver.

[1] https://www.spinics.net/lists/devicetree/msg317462.html

wens:
Looks like the new driver is on track to be included in v5.6. See
https://git.kernel.org/pub/scm/linux/kernel/git/thermal/linux.git/log/?h=thermal/linux-next

lategoodbye:
Yes, but i'm not sure it make sense to backport it to 4.19, i suggest to rebase it on Linux 5.4 (https://github.com/raspberrypi/linux/pull/3385) which has a much smaller diff.

wens:
The DT bindings in upstream don't match what is in the 5.4 branch. I assume a backport would include both the device tree stuff as well as the driver?

lategoodbye:
Yes, you will need to backport the devicetree parts & driver from Linux 5.6. But also enable the different driver in the config.

wens:
Done. The branch is now based on the commit  `b66e1e760f4c5f7cd783c1ca4c590765274e2cbe`,
which is the current kernel found at https://github.com/raspberrypi/firmware

lategoodbye:
@wens LGTM

wens:
Rebased onto latest rpi-5.4.y branch and fixed up commit hashes for the revert commits.

lategoodbye:
@popcornmix ping

popcornmix:
@pelwell this looks good to me and compiled and appeared to work correctly. Any concerns?

pelwell:
They're all reverts and back-ports, so if it compiles and appears to work then I have no problem with it.

popcornmix:
@wens @lategoodbye thanks!

PR #3311, opened 2019-10-29
WIP: Pi4 Full KMS support (DPI)

6by9:
PR covering adding full KMS DPI support for Pi4


6by9:
Creating the PR for this so that progress is at least vaguely public.

Tagging @mbgg after having chatted with him earlier today at ELCE regarding this.

Using a VGA666 board and a VGA monitor is the easiest thing, but you do lose UART and most other GPIO pins. Look at https://uk.pi-supply.com/products/gert-vga-666-hardware-vga-raspberry-pi or ebay to buy one. They're all clones of Gert's design at https://github.com/fenlogic/vga666

You'll need to load vc4-kms-v3d-pi4 to enable the KMS driver, and use either vc4-kms-kippah-7inch to get a fixed 800x480 display, or otherwise enable the DPI node and add a ```video=``` entry to the command line to select some other resolution.

As discussed, currently you'll also need to add
```
enable_dpi_lcd=1
dpi_mode=2
dpi_mode=16
```
or similar to config.txt to get the firmware to initialise the DPI interface first. There is some register that this driver isn't initialising, but I haven't nailed down what. I suspect it is due to the change in how pixel valves are set up as we now have 5 of them, with mux settings linking HVS channels to PVs (previously we had 3 HVS channels and 3 PVs, with muxing on the output of the PVs). That's something else to play with later.

arch/arm/boot/dts/overlays/vc4-kms-v3d-pi4-overlay.dts
```diff
@@ -0,0 +1,152 @@
...
+		target-path = "/chosen";
+		__dormant__ {
+			bootargs = "cma=64M";
+		};
+	};
+
+	fragment@5 {
+		target = <&i2c2>;
+		__overlay__  {
+			status = "okay";
```
6by9, line 51: Note to self - this is wrong as the i2c controller is for HDMI which is disabled, and has also changed on Pi4.

popcornmix:
Wasn't something like ``gpio=0-21=a2,np`` required?
I was also testing with ``dtoverlay=vc4-kms-kippah-7inch``

6by9:
> Wasn't something like gpio=0-21=a2,np required?

I'm glad someone has a better memory than me.
In theory vc4-kms-kippah-7inch should set up all the gpios too, but there was an issue on that. I think it was hooking on to the framebuffer DT node with the GPIOs, and that wasn't enabled or something.
```gpio=0-21=a2``` makes it explicit, although done by the firmware instead of kernel.

> I was also testing with dtoverlay=vc4-kms-kippah-7inch

True. ```dtoverlay=vc4-kms-kippah-7inch``` and a ```video=``` line will add the video mode to the LCD driver as an additional mode, so no need to munge DT further to enable DPI.

6by9:
Closing as other work is in progress and has overtaken this.

nullr0ute:
Is there somewhere this work can be followed?

6by9:
> Is there somewhere this work can be followed?

Not currently publicly as there is a fair amount of churn and reworking required.
I'd hope we're talking a few weeks rather than months for a first push, but there are still issues with how the hardware has to be configured.

rtomasa:
Any new on supporting KMS via DPI/GPIO without requiring the vc4-kms-kippah-7inch which makes some apps not detecting any display preventing them from working?

6by9:
There's no way to probe the resolution of a DPI display, therefore there has to be some form of configuration provided, and that is generally along the lines of the vc4-kms-kippah-7inch overlay.

HDMI has the display EDID to be able to automatically configure the output resolution.

The 5.4 branch is still in a state of flux for the vc4-kms-v3d driver on Pi4 as it has an early version of the patchset submitted upstream. An updated version will be backported when we can get it to work. It's non-trivial as the upstream patchset is against 5.8/5.9, and there are a number of framework updates between that and 5.4.

rtomasa:
@6by9 I've recently opened a new issue related to this one: _Raspberry Pi4 GPIO video add-ons don't work using KMS driver_ #3810
Kindly requesting you to have it a look and let me know if it is feasible at all, or which modifications would be required in the dts to make it work as when used in legacy video drivers. Maybe adding a fake EDID somewhere?

PR #3314, opened 2019-10-30
video: bcm2708_fb Add power down to screen blanking

JamesH65:
The FB_BLANK_POWERDOWN operation now calls the new power
down code in the firmware to correctly power down the specified
display. And conversely FB_BLANK_UNBLANK powers back up.

### DIFF
--- a/drivers/video/fbdev/bcm2708_fb.c
+++ b/drivers/video/fbdev/bcm2708_fb.c
@@ -521,6 +521,21 @@ static int bcm2708_fb_blank(int blank_mode, struct fb_info *info)
 	ret = rpi_firmware_property(fb->fbdev->fw, RPI_FIRMWARE_FRAMEBUFFER_BLANK,
 				    &value, sizeof(value));
 
+	if (!ret && (blank_mode == FB_BLANK_POWERDOWN ||
+	    blank_mode == FB_BLANK_UNBLANK)) {
+		u32 buf[2] = {fb->display_settings.display_num,
+			      blank_mode == FB_BLANK_POWERDOWN ? 0 : 1};
+
+		/* Convert display number to a display id in place */
+		ret = rpi_firmware_property(fb->fbdev->fw,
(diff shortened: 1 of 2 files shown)


popcornmix:
I believe the effect of this change will be the same as having ``hdmi_blanking=1``
We never made this a default option because (from [here](https://www.raspberrypi.org/documentation/configuration/config-txt/video.md):

> NOTE: This feature may cause issues when using applications which don't use the framebuffer, such as omxplayer.

So I'm not sure if we want to enable this by default.

It also seems a bit strange to tell the firmware to blank the screen twice through two different APIs (I believe the former does the same as the latter when ``hdmi_blanking=1`` - apart from handling the display id).

popcornmix:
Is it configurable somewhere whether FB_BLANK_POWERDOWN or FB_BLANK_NORMAL is sent for the normal console blanking that occurs after a few minutes of idle (either from command line or from X)?

JamesH65:
In the xscreensaver app, to get the power down you need to use advanced settings and specify you want an actual power down rather than just a blank. I think this simply activates the DPMS stuff. 

So by default, I do not think you get the powerdown, but I think I need to double check that.

popcornmix:
> So by default, I do not think you get the powerdown, but I think I need to double check that.

Check with a clean raspbian both in boot to console and boot to desktop mode. If neither of those trigger FB_BLANK_POWERDOWN then the non-framebuffer (e.g. omxplayer) issue should be okay.

JamesH65:
So, tested this on a Pi4 with fkms turned off.

Desktop mode: Using xscreensaver, unless you specifically turn on `power off on blanking`, omxplayer stays on screen. Once turned on, you do lose omxplayer as the monitor powers down.

Console mode: Using setterm I could set a blanking time, which simply blanks as expected. omxplayer still stays on screen. Unable to set a powerdown time - a no ioctl error, so not sure how to set a powerdown blank in console mode.

PR #3316, opened 2019-10-31
net:broadcom Set Ethernet LED modes via dt

JamesH65:
Pulls in a patch from 5.3, then adds the DT setting of LED modes over the top.

### DIFF
--- a/arch/arm/boot/dts/bcm2711-rpi-4-b.dts
+++ b/arch/arm/boot/dts/bcm2711-rpi-4-b.dts
@@ -334,5 +334,8 @@
 		pwr_led_gpio = <&pwr_led>,"gpios:4";
 		pwr_led_activelow = <&pwr_led>,"gpios:8";
 		pwr_led_trigger = <&pwr_led>,"linux,default-trigger";
+
+		eth_led0 = <&phy1>,"led-modes:0";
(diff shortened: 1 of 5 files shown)


pelwell (changes requested):
This looks generally OK - it's just the README that needs changing.

1. The existing `eth_*` and `eee` dtparams are currently 3B+-specific.
2. The LED defaults are board-specific.

pelwell:
That looks better. There's a minor capitalisation difference on 3B+, but otherwise I'm happy to merge once you've removed the `RFC:` tag.

JamesH65:
Cannot figure out how to flag the changes as all done, but changes are all done.

pelwell:
Except for the capitalisation of "Pi3b+".

JamesH65:
Ah, found it. Done.

pelwell:
The changes look good, but GitHub doesn't allow me to squash some commits and not others - it's all or nothing. Can you do the honours?

JamesH65:
Ah, bugger, I edited that change in using github, so it gone in as an extra commit rather than an amend. Will need to figure out how to fix that.

JamesH65:
Looks like a forced push of the change has sorted it out.

pelwell (approved):

PR #3321, opened 2019-11-06
overlays: Add apds9960 overlay

MikeDK:
Add an overlay for the AVAGO APDS9960 digital proximity, ambient light, rgb and gesture sensor.
Tested on an RPi3 running Raspbian Buster with a Sparkfun APDS-9960 breakout board, INT connected to GPIO 4.


pelwell:
That's a good effort. I've fixed a few README niggles, and there are just a few things left to do:
1. I'd like the defconfig additions to appear in the places where `make ... savedefconfig` would put them.
2. Two separate commits would be great - one for the overlay, and one for the defconfig changes.
3. A `Signed-off-by:` line in the commit messages is a nice-to-have (although I've just missed one on a one-character change - https://github.com/raspberrypi/linux/commit/eef78b714c3fbc635642332a2ce7d66e51f514d9).
If you can do all that, go ahead, otherwise I can merge it behind the scenes - just let me know if you're happy for me to add `Signed-off-by: Mike Kaplan <m.kaplan@evva.com>` for you.

MikeDK:
Ok I will try to make those changes, but have some questions ;)

ad 1.: I guess its enough to do "cp path-to-config/config-in-question.defconfig .config && make savedefconfig" for all of the configs i already created?

ad 2.: Ok - so I have two options: either revert my changes and do a re-commit, or close this pull request and make a new one ... which one is better? ;)

pelwell:
For 1, `make ... bcmrpi_defconfig`, `make ... savedefconfig`, then copy the savedefconfig on top of the original defconfig.

For 2, I would uncommit (but not undo) the changes with `git reset HEAD~3` Iwhere 3 is the number of commits you want to redo), then use `git add -p` to stage the changes for the first commit. Answer 'y' to include a patch, 'n' to skip it. Then `git commit -s` to commit and sign the staged changes, and repeat. If all the remaining changes are for the same commit you can use `git add -u` instead of `-p`.

MikeDK:
Ok done ;)

one thing ... in arch/arm64/configs/bcmrpi3_defconfig my savedefconfig removed two extra lines, which seem to be not needed (because of dependencies) ... I hope that's ok

pelwell:
That looks perfect. I get the same issue w.r.t. the disappearing setting in the 64-bit bcmrpi3_defconfig, it's just a consequence of the dependencies, as you say.

Can I ask for one tiny change? Make the title of the second commit `configs: apds9960: add module to defconfigs` instead of `overlays: apds9960: add module to defconfigs`. It makes it easier for whoever (usually @popcornmix) ends up squashing the commits as we move to newer kernel versions. With that done I'll merge it without delay.

MikeDK:
done ;)

pelwell:
This overlay is now in the latest rpi-update kernel+firmware release. Note that, like all rpi-update releases, it should be treated with caution.

PR #3323, opened 2019-11-07
net:phy:2711 Change the default ethernet LED actions

JamesH65:
This should return default behaviour back to that of previous
releases.

Also removed a now incorrect comment block.


JamesH65:
@timg236

PR #3327, opened 2019-11-11
UAC audio for quectel modem

jackghg:
To complete quectel modems drivers, this enable usb audio UAC, from quectel docs.
NOTES:
- Can be a security problem ? As far as I know, the uac audio mode still need something like alsaloop to recieve and send audio. But can an uac device "enable it self" and listen from microphone ?
- I found the same code to also help QMI drivers along with other code (already?) in the qmi_wwan.c file.

### DIFF
--- a/drivers/usb/serial/option.c
+++ b/drivers/usb/serial/option.c
@@ -2064,6 +2064,10 @@ static int option_probe(struct usb_serial *serial,
 	 */
 	if (device_flags & NUMEP2 && iface_desc->bNumEndpoints != 2)
 		return -ENODEV;
+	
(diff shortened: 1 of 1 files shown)


popcornmix:
This is patching a file that we haven't modified from upstream. As such any patches should be submitted upstream where the right kernel devs can comment.

Assuming it is accepted we will get it when updating the kernel (or can cherry pick it earlier).

See: https://kernelnewbies.org/UpstreamMerge

jackghg:
wow! i was sure that quectel drivers came from raspberry!

PR #3337, opened 2019-11-16
Add Support for simultaneous use of JustBoom DAC and JustBoom Digi based Audio boards

johannesk:
These two commits enable the simultaneous use of a JustBoom DAC and JustBoom Digi board, such that both simultaneously produce the same audio output. The first commit is necessary since the sound/soc layer does not yet support multicodec setups with one codec being the master. All existing multicodec drivers have all codecs as slave and the soc as master. I tested all changes using a pi zero with extra long pin headers, where both an JustBoom DAC Zero and a JustBoom Digi Zero is connected.


pelwell:
I have a few problems with this PR:
1. sound/soc/soc-core.c is an unmodified (*) upstream file. A change such as you are proposing should be sent to the relevant maintainers and mailing lists - use `git format-patch` to produce a patch file, the run `scripts/get_maintainer.pl` on it to found out where to send it.
2. Does it really need 260 lines of code? Quite a bit of that card driver file looks more like codec driver software.
3. How does this work on the mechanical front? The boards don't look very stackable.
4. There are ordering and whitespace issues in the README and Makefile.

shawaj:
Just weighing in about number 3...the zero boards are stackable using one of the through board headers used on the official PoE and the Sense HAT.

The Digi HAT and DAC HAT have non-populated headers which allow you to stack boards on top of each other if you were to solder a header in. 

Last, but not least, we are doing a board revision soon to change the HAT boards to use the same header as the zero boards so they will be more easilly stackable.

pelwell:
@shawaj Is this dual-headed approach something you are going to be promoting?

shawaj:
@pelwell yes we intend to promote it in the future

johannesk:
To answer the raised questions:
1. I submitted the sound/soc/soc-core.c patch to upstream and I am waiting for for approval or discussion there (https://mailman.alsa-project.org/pipermail/alsa-devel/2019-November/158962.html).
2. The driver contains platform specific codec code which is similar to other sound hat drivers to setup e.g. the pi clock. It is twice as much code in comparison to other drivers since it configures not only one but two codecs.
3. In my setup I stack the zero hats by soldering long pin headers (similar to https://www.ebay.de/i/232311298260), but long extension headers might also work (https://www.reichelt.com/de/en/raspberry-pi-40-pin-stacking-header-rm-2-54-rpi-header-40-p223626.html).
4. I fixed the ordering, but I am unsure about the whitespace issues. There already seems to be a mix of spaces and tabs in the Kconfig file. Are there any whitespace issues left?

pelwell:
The overlay and README look fine - the checker has no complaints.

Please run `git format-patch HEAD~2` and `scripts/checkpatch.pl 000*` (assuming no other patch files are lying around), then fix the errors that are reported.

johannesk:
The pull request now follows the Linux kernel coding style.

The [upstream patch](https://mailman.alsa-project.org/pipermail/alsa-devel/2019-November/158962.html) did not receive any response within 12 days. Can you give any hints on how to get upstream to look at this patch?

pelwell:
I'd just post a follow-up message asking politely for a review.

In the meantime I've added Signed-off-by: lines and merged the patches into 4.19.

PR #3338, opened 2019-11-17
arm: dts: overlays: pitft35-resistive: add upstream compatible

nullr0ute:
The upstream hx8357d driver uses "adafruit,yx350hv15" for the compatible
string explicitly for this screen config and not a hx8357d generic for
the controller so add that in as well so it will work with an unmodified
upstream kernel driver. We leave the downstream as the priority.


nullr0ute:
Not 100% sure how the branches are handled, do you want one against 5.4 too

pelwell:
Thanks - I've cherry picked it.

PR #15, opened 2012-05-07
Smsc95xx patches

steveglen:
Hi! I'm Steve Glendinning, the kernel maintainer of the smsc95xx ethernet driver (as used by the Raspberry Pi).

Please apply these two driver patches to your Rasberry Pi kernel tree.  They're included in Linus's tree from 3.4.0-rc5+, and they address two issues that affect the ethernet hardware on Raspberry-Pi:

1 - MTU is only 1488 without these, so full size ethernet frames can't be sent or received properly
2 - The link patch makes it much quicker to properly detect a link at startup and run DHCP.  Without it you may also see some error frames reported before the link actually comes up.

Please also advise anyone else packaging kernels for Raspberry Pi distributions of these two patches.

Thanks!
steve.glendinning@shawell.net

PR #3362, opened 2019-12-01
bcm2708_fb: Disable warning when calling dma_alloc_coherent

jason77-wang:
When we boot the kernel on the rpi4/3/2, we sometimes could see the
warning calltrace like below:

bcm2708_fb soc:fb: FB found 1 display(s)
cma: cma_alloc: alloc failed, req-size: 1753 pages, ret: -12
WARNING: CPU: 2 PID: 1 at mm/page_alloc.c:4702 __alloc_pages_nodemask+0x284/0x2c8
Modules linked in:
CPU: 2 PID: 1 Comm: swapper/0 Not tainted 5.3.0-1013-raspi2 #15-Ubuntu
Hardware name: Raspberry Pi 3 Model B Plus Rev 1.3 (DT)
pstate: 20400005 (nzCv daif +PAN -UAO)
pc : __alloc_pages_nodemask+0x284/0x2c8
lr : __dma_direct_alloc_pages+0x100/0x1e8
sp : ffff0000100238a0
x29: ffff0000100238a0 x28: 0000000000000000
x27: ffffa7c578867580 x26: ffffa7c57786a080
...

After investigation, I found there are two situations which could
introduce this calltrace.
 1) booting the arm64 kernel on rpi4 boards without the monitor
 connected, in this situation, the xres and yres euqal to 0, so the
 dma_alloc_coherent wants to alloc a memory region with 0 size, this
 will result the get_order() returns (BITS_PER_LONG - PAGE_SHIFT),
 then it will introduce the calltrace because
 unlikely(order >= MAX_ORDER) is true (mm/page_alloc.c)
 2) booting the kernel on rpi2/3 boards with a high resolution monitor
 connected, in this case the xres * yres is too big, the cma_alloc
 doens't have enough memory for it, and finally it will also introduce
 the calltrace by unlikely(order >= MAX_ORDER) is true.

The bcm2708_fb could handle the situation of failure on calling
dma_alloc_coherent, and the driver itself will print some log to
notify users that this alloc fails. So we don't need the mm/page_alloc
to print that calltrace anymore, printing a calltrace when booting
usually makes users think the kernel has a critical problem.

Signed-off-by: Hui Wang <hui.wang@canonical.com>

### DIFF
--- a/drivers/video/fbdev/bcm2708_fb.c
+++ b/drivers/video/fbdev/bcm2708_fb.c
@@ -410,7 +410,8 @@ static int bcm2708_fb_set_par(struct fb_info *info)
 		}
 
 		fb->cpuaddr = dma_alloc_coherent(info->device, image_size,
-						 &fb->dma_addr, GFP_KERNEL);
+						 &fb->dma_addr,
+						 GFP_KERNEL | __GFP_NOWARN);
 
 		if (!fb->cpuaddr) {
 			fb->dma_addr = 0;


6by9:
Ooh, magic flags - I wish I'd found that one when I made the original change!

I believe the change you're looking at that calls dma_alloc_coherent was reverted on 4.19 as I couldn't find a way to silence that warning should the CMA heap be too small, and there were concerns that there could be configurations where the CMA heap was small and the FB ended up taking too much and leaving the rest of the system unstable. I guess those reverts didn't happen on the later branches, and we better review the situation.
(8a61a0791ab90cd7d2c1ef270ebc056580940b0e and a4750f850182ab7dbe25d519b385e592e1d5b353).

My only complaint with the patch is that it makes the line > 80 chars, so isn't compliant with the kernel coding standard. Could you drop the ```GFP_KERNEL | __GFP_NOWARN);``` onto the next line please?

jason77-wang:
Done, fixed the problem of "the line > 80 chars".
Thx.

6by9:
The more usual formatting would be 
```
		fb->cpuaddr = dma_alloc_coherent(info->device, image_size,
						 &fb->dma_addr,
						 GFP_KERNEL | __GFP_NOWARN);
```
to keep the terms of parameter 4 on the same line, but checkpatch doesn't complain at splitting it as you have done.

jason77-wang:
push again, fixed the conflict and put the parameter 4 on the same line. Now the patch is changed to:
                fb->cpuaddr = dma_alloc_coherent(info->device, image_size,
-                                                &fb->dma_addr, GFP_KERNEL);
+                                                &fb->dma_addr,
+                                                GFP_KERNEL | __GFP_NOWARN);
 
thx.

6by9 (approved):
Thanks.
Whilst this isn't a file that will ever be upstreamed to mainline, it's nice to keep the coding style aligned with it.

Issue #3363, opened 2019-12-02
Regression: no /dev/dri/renderD12*


msisov:
Hi,

I am using meta-raspberrypi yocto layer (warrior and thud branches) on my Raspberry Pi3 B+ that points to kernel v4.19 now. Since the time they've started to point to that version, I've noticed that there is no /dev/dri/renderD12* created that makes some client applications fail during startup (for example, Chromium with Ozone/Wayland support that relies on drm render nodes). In the kernel v4.14, /dev/dri/renderD12* was created and the client apps (including weston-simple-dmabuf-drm) worked fine.

I am also using vc4 graphics.

Here are some logs that may help to identify the problem:

Kernel v4.14 with working render nodes:

```
root@raspberrypi3:/# dmesg | grep drm
[ 2.810773] vc4-drm soc:gpu: bound 3f902000.hdmi (ops vc4_hdmi_ops [vc4])
[ 2.811031] vc4-drm soc:gpu: bound 3f806000.vec (ops vc4_vec_ops [vc4])
[ 2.811235] vc4-drm soc:gpu: bound 3f400000.hvs (ops vc4_hvs_ops [vc4])
[ 2.811846] vc4-drm soc:gpu: bound 3f206000.pixelvalve (ops vc4_crtc_ops [vc4])
[ 2.812212] vc4-drm soc:gpu: bound 3f207000.pixelvalve (ops vc4_crtc_ops [vc4])
[ 2.812709] vc4-drm soc:gpu: bound 3f807000.pixelvalve (ops vc4_crtc_ops [vc4])
[ 2.834843] vc4-drm soc:gpu: bound 3fc00000.v3d (ops vc4_v3d_ops [vc4])
[ 2.835890] [drm] Initialized vc4 0.0.0 20140616 for soc:gpu on minor 0
[ 2.835928] [drm] Supports vblank timestamp caching Rev 2 (21.10.2013).
[ 2.835953] [drm] Driver supports precise vblank timestamp query.
[ 2.962854] vc4-drm soc:gpu: fb0: frame buffer device
```

Kernel v4.19 where render nodes are not exposed:

```
root@raspberrypi3:~# dmesg | grep drm
[ 3.562379] vc4-drm soc:gpu: bound 3f902000.hdmi (ops vc4_hdmi_ops [vc4])
[ 3.562729] vc4-drm soc:gpu: bound 3f806000.vec (ops vc4_vec_ops [vc4])
[...]

PR #3365, opened 2019-12-02
drm/vc4: Correct disabling of render nodes

6by9:
See #3363 
4.19 had an early version of this patch merged which didn't match all the DT compatible strings. Revert the early version and cherry-pick the mainline version.


6by9:
WIP as I'm still compiling the resulting tree to test.

6by9:
That seems to work. WIP tag removed.

Issue #3332, opened 2019-11-14
dwc_otg: armv7l kernel: rpi2/3: the dwc_otg will report i/o errors when copying a large-size file (300M~600M)


jason77-wang:
**Is this the right place for my bug report?**
Yes

**Describe the bug**
build an arm32 kernel with the default config (arch/arm/configs/bcm2711_defconfig), then booting this kernel on rpi2/3 with 1G physical memory, and plug a usb stick to the usb port, copy a large size file (300M-600M) to usb stick, the usb host driver dwc_otg will print I/O errors.

Please refer to: https://bugs.launchpad.net/ubuntu/+source/linux-raspi2/+bug/1852510

**To reproduce**
follow the steps as above

**Expected behaviour**
usb stick can work correctly on rpi2/3 under armv7l kernel with HIGHMEM and VMSPLIT_3G enabled.

**Actual behaviour**
usb stick can't work on rpi2/3 under the armv7l kernel with HIGHMEM and VMSPLIT_3G.

**System**
RPI2/3 boards with 1G physical ram

**Logs**
[ 176.129299] sd 0:0:0:0: [sda] tag#0 FAILED Result: hostbyte=DID_ERROR driverbyte=DRIVER_OK
[ 176.129326] sd 0:0:0:0: [sda] tag#0 CDB: Write(10) 2a 00 00 2e 24 b0 00 00 f0 00
[ 176.129349] blk_update_request: I/O error, dev sda, sector 3024048 op 0x1:(WRITE) flags 0x4000 phys_seg 15 prio class 0
[ 176.883968] usb 1-1.1.2: reset high-speed USB device number 8 using dwc_otg
[ 177.079960] usb 1-1.1.2: reset high-speed USB device number 8 using dwc_otg

**Additional context**
I found with the arm32 kernel,  if we enable the HIGHMEM and VMSPLIT_3G (like arch/arm/configs/bcm2711_defconfig) in the kernel, some memory will map to highmem region on rpi2/3 with 1G physical memory, and the dwc_otg will not work anymore when copying a large-size file to usb stick.

pelwell:
bcm2711_defconfig is for Pi 4 (which uses the BCM2711 SoC). For Pi 2 and Pi 3 use bcm2709_defconfig.

jason77-wang:
[...]

PR #3371, opened 2019-12-06
dwc_otg: checking the urb->transfer_buffer too early (#3332)

jason77-wang:
After enable the HIGHMEM and VMSPLIT_3G, the dwc_otg driver doesn't
work well on Pi2/3 boards with 1G physical ram. Users experience
the failure when copying a file of 600M size to the USB stick. And
at the same time, the dmesg shows:
usb 1-1.1.2: reset high-speed USB device number 8 using dwc_otg
sd 0:0:0:0: [sda] tag#0 FAILED Result: hostbyte=DID_ERROR driverbyte=DRIVER_OK
blk_update_request: I/O error, dev sda, sector 3024048 op 0x1:(WRITE) flags 0x4000 phys_seg 15 prio class 0

When this happens, the sg_buf sent to the driver is located in the
highmem region, the usb_sg_init() in the core/message.c will leave
transfer_buffer to NULL if the sg_buf is in highmem, but in the
dwc_otg driver, it returns -EINVAL unconditionally if transfer_buffer
is NULL.

The driver can handle the situation of buffer to be NULL, if it is in
DMA mode, it will convert an address from transfer_dma.

But if the conversion fails or it is in the PIO mode, we should check
buffer and return -EINVAL if it is NULL.

BugLink: https://bugs.launchpad.net/bugs/1852510
Signed-off-by: Hui Wang <hui.wang@canonical.com>

### DIFF
--- a/drivers/usb/host/dwc_otg/dwc_otg_hcd_linux.c
+++ b/drivers/usb/host/dwc_otg/dwc_otg_hcd_linux.c
@@ -821,10 +821,6 @@ static int dwc_otg_urb_enqueue(struct usb_hcd *hcd,
 		dump_urb_info(urb, "dwc_otg_urb_enqueue");
 	}
 #endif
-
-	if (!urb->transfer_buffer && urb->transfer_buffer_length)
-		return -EINVAL;
-
 	if ((usb_pipetype(urb->pipe) == PIPE_ISOCHRONOUS)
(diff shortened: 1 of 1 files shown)


pelwell:
Looks good to me. @P33M Feel free to merge if you're happy with it, and I'll cherry-pick to 4.19 and 5.4 (5.3 is essentially dead as far as we're concerned).

P33M:
LGTM. Thanks for tracking this down

PR #3375, opened 2019-12-13
WIP: drm/vc4: Pass an update_id to the fw to return on completion

6by9:
This should allow KMS to return the events correctly for
page flip completions.

ISSUE SEEN: On quiting X there appears to be a mode change
that is timing out because we aren't returning a completion.
I'm wondering if this is due to kicking off the mode change
when the update is still active, resulting in the fw never
giving us the appropriate callback.
DRM has a 10sec timeout which clears the issue, but it isn't
clean at present.

### DIFF
--- a/drivers/media/platform/bcm2835/bcm2835-unicam.c
+++ b/drivers/media/platform/bcm2835/bcm2835-unicam.c
@@ -766,12 +766,6 @@ static int unicam_all_nodes_streaming(struct unicam_device *dev)
 	return ret;
 }
 
-static int unicam_all_nodes_disabled(struct unicam_device *dev)
-{
-	return !dev->node[IMAGE_PAD].streaming &&
-	       !dev->node[METADATA_PAD].streaming;
-}
-
 static void unicam_queue_event_sof(struct unicam_device *unicam)
 {
 	struct v4l2_event event = {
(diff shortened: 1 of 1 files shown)


6by9:
@popcornmix This is the kernel side patch. Ignore the commit text - I think this should all be good now. I'll amend it and repush.

popcornmix:
What's the behaviour if kernel is updated without firmware? And vice versa?

6by9:
Should all be fine.

Old firmware/new kernel: the fw will return a count ID of 0, which FKMS will take as "I don't support the new ID", and have the same behaviour as before.

New firmware/old kernel: the structure size has changed, so the fw will insert the count ID as being 0. On vsync it'll set that 0 into the top bits of the SMI reg passed across. The kernel will then never look at those registers.

popcornmix:
I'm seeing a problem here. With this patch and corresponding firmware, after exiting kodi the system crashes (kernel panic and loss of ssh connection).
Reverting either kernel or firmware to stock and the issue is gone.

You don't have to play a video in kodi. Just launch it and shutdown (S key) is enough.
Uart not working but panic looks roughly like:
```
Kernel panic - not syncing: System is deadlocked on memory
Backtrace:
(panic) form pagefault_out_of_memory
(out of memory)
__alloc_pages_nodemask
copy_process
_do_fork
sys_clone
```
(can also provide a screenshot).

6by9:
Closing as not current work

PR #3385, opened 2019-12-27
BCM2711 - Upstream Linux 5.5 backport

lategoodbye:
This pull requests tries to reduce the delta between downstream and current upstream development. Since Linux 5.4 LTS has been released i don't see any point to backport 5.5 changes into 4.19.

In short the PR contain the following 5.5 changes:
- BCM2711 board support
- timer cleanup
- U-Boot fix
- Backport of Bluetooth, GENET, RNG

Unfortunately there is no complete backward compatibility of GENET, since the PHY mode was wrong before.

This PR doesn't contain all 5.5 DTS changes:
- generic pinctrl (fundamental change which might break some overlays)
- HDMI (might break graphics)
- CMA fix (breaks Kodi, needs to be investigated further)

I tried to keep the changes as minimal as possible.

This PR has been tested with bcm2711_defconfig.


popcornmix:
`ARM: dts: bcm2711: force CMA into first GB of memory` stops kodi (installed from raspberry pi repo) from working.
It fails with:
```
DRM_IOCTL_MODE_CREATE_DUMB failed: Cannot allocate memory
Failed to create scanout resource
DRM_IOCTL_MODE_CREATE_DUMB failed: Cannot allocate memory
Failed to create scanout resource
terminate called after throwing an instance of 'std::runtime_error'
  what():  eglSwapBuffers failed
```

lategoodbye:
@popcornmix Thanks for testing. I can drop this patch, but this needs to be investigated further.

@vianpl Any ideas?

lategoodbye:
I dropped the CMA fix and updated the initial comment.

vianpl:
Here's my guess. By configuring CMA in the device tree the kernel stops reacting to the command line options. I bet the Kodi install is updating the CMA size trough the kernel command line, which is not being taken into account. The only way to do it after this patch is by using overlays.

popcornmix:
With cma not in cmdline.txt and without the cma patch:
```
[    0.000000] cma: Reserved 256 MiB at 0x000000001bc00000
```
With cma=512M in cmdline.txt and without the cma patch:
```
[    0.000000] cma: Reserved 512 MiB at 0x000000000bc00000
```
with cma not in cmdline.txt and the cma patch:
```
[    0.000000] Reserved memory: created CMA memory pool at 0x0000000029c00000, size 32 MiB
```
With cma=512M in cmdline.txt and the cma patch:
```
[    0.000000] Reserved memory: created CMA memory pool at 0x0000000029c00000, size 32 MiB
```
All these have the vc4-fkms-v3d overlay which attempts to increase CMA to 256M (@pelwell can confirm but I think the overlay ends up modifying the cmdline kernel sees).
But we only ever get 32M of CMA with the patch.

So should cma=<> on cmdline still work? Is there a way to make it do so?

vianpl:
> With cma not in cmdline.txt and without the cma patch:
> 
> ```
> [    0.000000] cma: Reserved 256 MiB at 0x000000001bc00000
> ```
> 
> With cma=512M in cmdline.txt and without the cma patch:
> 
> ```
> [    0.000000] cma: Reserved 512 MiB at 0x000000000bc00000
> ```
> 
> with cma not in cmdline.txt and the cma patch:
> 
> ```
> [    0.000000] Reserved memory: created CMA memory pool at 0x0000000029c00000, size 32 MiB
> ```
> 
> With cma=512M in cmdline.txt and the cma patch:
> 
> ```
> [    0.000000] Reserved memory: created CMA memory pool at 0x0000000029c00000, size 32 MiB
> ```
> 

This is the behavior I observed upstream. I double checked the code, command line is explicitly ignored. 

> All these have the vc4-fkms-v3d overlay which attempts to increase CMA to 256M (@pelwell can confirm but I think the overlay ends up modifying the cmdline kernel sees).
> But we only ever get 32M of CMA with the patch.

I think the overlay could be easily updated to work with the new CMA setup. We'd just need to update the size property in the CMA node instead of altering the command line.

> So should cma=<> on cmdline still work? Is there a way to make it do so?

Well I guess this is something we should discuss with people upstream. I'd say the command line should take precedence to the DTB. At least for it's size. But I guess we'd have to investigate further. There are not so many users of DTB defined CMAs, so we might be the first with such needs.

pelwell:
I'd be happy with a DTB-only solution - patching the command line from an overlay always seemed a bit ugly.

vianpl:
@pelwell  I'm still on holidays until the 7th, but if you can wait until then I'll be happy to have a got at it.

popcornmix:
> I'd be happy with a DTB-only solution - patching the command line from an overlay always seemed a bit ugly.

As long as there's an easy way for users to adjust it. The HEVC V4L2 driver for example wants a lot of CMA to play 10bit 4K video (256M is not enough), and we have examples of high CMA settings causing kernel allocation failures, so we can't just unconditionally set it high.

A config.txt setting that causes firmware to adjust cma value in DTB should be okay if cmdline.txt solution isn't possible.

pelwell:
I'd make it a dtparam.

vianpl:
I've been working on the following small series to address the CMA issue. See the first two patches here:  https://github.com/vianpl/linux/commits/bcm283x-cma-unification.

With the DT patch it'll be trivial to update all overlays that change CMA sizes. And if the cmdline/DT precendence patch doesn't make it we can always create an extra overlay to edit CMA sizes trough a dtparam.

Will this overall approach be acceptable to you?

pelwell:
I've commented on one of the patches, but to summarise: I'm happy with it, but I'd like a label on the `linux,cma` node for overlays (or aliases) to target.

lategoodbye:
Since i dropped the CMA patch, this could be a separate pull request. This pull request is rebased now.

popcornmix:
The patches (excluding CMA one) seemed fine in my limited testing, so no objections.

lategoodbye:
Pull request rebased

lategoodbye:
@pelwell Are there any objections with this pull request?

pelwell:
I'm just giving it a quick smoke test prior to merging.

pelwell:
Note to self: If you ever experiment with different armstubs, be sure to delete them before regression testing.

lategoodbye:
Thanks

PR #3394, opened 2020-01-06
pinctrl: bcm2835: Change init order for gpio hogs

pelwell:
pinctrl-bcm2835 is a combined pinctrl/gpio driver. Currently the gpio
side is registered first, but this breaks gpio hogs (which are
configured during gpiochip_add_data). Part of the hog initialisation
is a call to pinctrl_gpio_request, and since the pinctrl driver hasn't
yet been registered this results in an -EPROBE_DEFER from which it can
never recover.

Change the initialisation sequence to register the pinctrl driver
first. This requires slightly more cleanup in the event of a failure,
but the use of devm_gpiochip_add_data takes care of some of that.

See: https://www.raspberrypi.org/forums/viewtopic.php?f=107&t=260600

### DIFF
--- a/drivers/pinctrl/bcm/pinctrl-bcm2835.c
+++ b/drivers/pinctrl/bcm/pinctrl-bcm2835.c
@@ -1140,7 +1140,7 @@ static int bcm2835_pinctrl_probe(struct platform_device *pdev)
 		raw_spin_lock_init(&pc->irq_lock[i]);
 	}
 
-	err = gpiochip_add_data(&pc->gpio_chip, pc);
+	err = devm_gpiochip_add_data(dev, &pc->gpio_chip, pc);
 	if (err) {
 		dev_err(dev, "could not add GPIO chip\n");
 		return err;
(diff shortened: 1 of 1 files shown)


pelwell:
@6by9 Can you give this the once-over?

6by9:
Based on your description, yes this appears to be correct.

Does the use of devm_gpiochip_add_data  really help the cleanup? I don't see any significant changes in error paths after devm_gpiochip_add_data compared to the previous version. 
Did you mean devm_pinctrl_register instead?

pelwell:
No, I did mean `devm_gpiochip_add_data` - without it, a failure in `gpiochip_irqchip_add` would require a call to gpiochip_remove. `devm_pinctrl_register` does help, but I didn't mention it because it wasn't a change in this commit.

6by9:
OK, so actually you have two changes in the patch
- a fix that on gpiochip_irqchip_add failing it previously didn't release all resources
- reordering of the pinctrl and gpiochip registration.

You know what upstream would say to two things changed in one patch, but I'm not so fussed. Your comment had just confused me slightly.

pelwell:
That's a fair point - it's now two separate commits.

6by9:
I'm happy!

PR #3396, opened 2020-01-07
Pisound: MIDI communication fixes for scaled down CPU.

gtrainavicius:
* Increased maximum SPI communication speed to avoid running too slow when the CPU is scaled down and losing MIDI data.
* Keep track of buffer usage in millibytes for higher precision.

Signed-off-by: Giedrius Trainavičius <giedrius@blokas.io>


gtrainavicius:
Thank you!

PR #3418, opened 2020-01-21
adds the HiFiBerry DAC+HD version

j-schambacher:
This adds the driver for the DAC+HD version supporting HiFiBerry's
PCM179x based DACs. It also adds PLL control for clock generation.

Signed-off-by: Joerg Schambacher joerg@i2audio.com


pelwell:
I had to fix up the Signed-off-by: line and Makefile order, and add the 64-bit configs, but otherwise it looks OK. The huge PLL configurability feels like overkill though.

j-schambacher:
The PLL config has been requested by some customers who want to run the board at 'strange' sample frequencies...

Thanks Phil for your help

pelwell:
I've pushed a forward port to rpi-5.4.y. It required a few changes, so please take a look and confirm it works as expected.

j-schambacher:
From what I see it should be ok, but I can’t get any of the (more complex) Hifiberry drivers to work (setup is P3 B, just migrated from the 4.19 driver development). They get loaded and seem to probe ok, but the sound cards do not show up. I just want to ask for any advice before spending time and dig deeper. Thanks!

j-schambacher:
Your changes are OK. I've found a problem due to the different order of loading/probing the board and the clock driver. It requires a small fix to work with 5.4. Will do a PR.
Didn't check the consequences/necessity for 4.19

PR #3419, opened 2020-01-21
Fixup rpimem to cleanly unload and reload, and make checkpatch clean.

6by9:
Issue noted when JC was trying to unload and reload an overlay for the V4L2 HEVC codec.
Made checkpatch clean (other than DT compatible string) whilst I was at it.

### DIFF
--- a/drivers/char/broadcom/rpivid-mem.c
+++ b/drivers/char/broadcom/rpivid-mem.c
@@ -66,6 +66,7 @@ static int rpivid_mem_open(struct inode *inode, struct file *file)
 	int dev = iminor(inode);
 	int ret = 0;
 	struct rpivid_mem_priv *priv;
+
 	if (dev != DEVICE_MINOR && dev != DEVICE_MINOR + 1)
 		ret = -ENXIO;
 
@@ -134,7 +135,6 @@ static int rpivid_mem_probe(struct platform_device *pdev)
 	struct device *dev = &pdev->dev;
 	struct resource *ioresource;
 	struct rpivid_mem_priv *priv;
-
 
 	/* Allocate buffers and instance data */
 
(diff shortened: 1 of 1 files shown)


drivers/char/broadcom/rpivid-mem.c
```diff
@@ -197,22 +197,29 @@ static int rpivid_mem_probe(struct platform_device *pdev)
...
 		oldname[4] = 'o';
 		oldname[5] = 'n';
-		(void)device_create(priv->class, NULL, priv->devid + 1, NULL,
-				       oldname + 1);
+		old_mem_dev = device_create(priv->class, NULL, priv->devid + 1,
+					    NULL, oldname + 1);
 		kfree(oldname);
+
+		ptr_err = old_mem_dev;
+		if (IS_ERR(old_mem_dev))
```
pelwell: IS_ERR(ptr_err) would seem more logical here. Or just drop the old_mem_dev and assign directly to ptr_err.

drivers/char/broadcom/rpivid-mem.c
```diff
@@ -1,3 +1,4 @@
+// SPDX-License-Identifier: GPL-2.0-only
```
pelwell: The SPDX declaration doesn't appear to match the licence text, which looks like a dual 3-clause BSD and GPL2.

drivers/char/broadcom/rpivid-mem.c
6by9: I'd copied and pasted from the other device_create failure case. Having ptr_err seems redundant anyway, and the error case may as well do ```err = PTR_ERR(rpivid_mem_dev)``` or ```err = PTR_ERR(rpivid_mem_dev)```.

Actually the handling of ```dev``` as being ```&pdev->dev```, and is then assigned to ```priv-dev``` leaves ```dev``` unused through the rest of the function. I'll do a few more lines of cleanup.

drivers/char/broadcom/rpivid-mem.c
6by9: Oops, missed that!

6by9:
Updated. Reworked the conversion from PTR to ERR overall. 
Added an extra patch to remove a deref of ```priv``` when we've just freed it too!

PR #3424, opened 2020-01-23
fix master mode settings of HiFiBerry DAC+ADC PRO card

j-schambacher:
This patch fixes the board DAI setting when in master-mode.
Wrong setting could have caused random pop noise.

Signed-off-by: Joerg Schambacher joerg@i2audio.com


pelwell:
Try and remember to use `<>`s around your email address  in the commit message.

PR #3425, opened 2020-01-24
pisound: Added reading Pisound board hardware revision and exposing it

gtrainavicius:
pisound: Added reading Pisound board hardware revision and exposing it in kernel log and sysfs file:

/sys/kernel/pisound/hw_version

Signed-off-by: Giedrius <giedrius@blokas.io>

### DIFF
--- a/sound/soc/bcm/pisound.c
+++ b/sound/soc/bcm/pisound.c
@@ -591,6 +591,8 @@ static int spi_read_info(void)
 				buffer[0],
 				buffer[1]
 				);
+
+			g_fw_version[MAX_VERSION_STR_LEN-1] = '\0';
 			break;
 		case 3:
 			if (n != 2)
@@ -603,6 +605,8 @@ static int spi_read_info(void)
 				buffer[0],
 				buffer[1]
 			);
+
+			g_hw_version[MAX_VERSION_STR_LEN-1] = '\0';
 			break;
 		case 1:
 			if (n >= sizeof(g_serial_num))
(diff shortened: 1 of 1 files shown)


gtrainavicius:
ping :)

sound/soc/bcm/pisound.c
```diff
@@ -581,13 +585,25 @@ static int spi_read_info(void)
 				return -EINVAL;
 
 			snprintf(
-				g_version,
-				sizeof(g_version),
+				g_fw_version,
```
pelwell, line 588: The output string will _not_ be terminated in the event that the buffer isn't large enough - you need to check the return value, or forceably terminate if you don't mind the truncation.

sound/soc/bcm/pisound.c
```diff
@@ -743,10 +760,18 @@ static const char *pisnd_spi_get_id(void)
 	return "";
 }
 
-static const char *pisnd_spi_get_version(void)
+static const char *pisnd_spi_get_fw_version(void)
```
pelwell, line 763: This function makes no sense. If the saved version is non-zero length then then it returns that version string, otherwise it returns another empty string. Why not just return `g_fw_version` in both cases and drop the strlen?

sound/soc/bcm/pisound.c
gtrainavicius: Thank you for thoroughly looking into the changes. Actually it will always succeed - the buffer is 6 bytes long, the fmt string is "%x.%02x" and another similar one is "%x.%x", that means the longest possible string where the arguments are a single byte each would be for example "ff.ff" (and implicit \0 byte). That fits exactly into the 6 bytes available. It would become a problem if the arguments would be larger than 'char' however, but this location is not likely to change element size.

sound/soc/bcm/pisound.c
pelwell: It's still safer to always explicitly set the last byte of the buffer to a NUL, just in case.

sound/soc/bcm/pisound.c
pelwell: It's still safer to always set the last byte of the buffer to NUL just in case.

gtrainavicius:
Ok, I have simplified the getters and added code to ensure the strings get the NULL terminations. Thank you for the suggestions!

PR #3431, opened 2020-01-26
Fix i2c-pwm-pca9685a overlay

MikeDK:
Overlay was using wrong compatible string - now using the one defined in the driver source.


pelwell:
Thanks!

Issue #2649, opened 2018-08-16
Add zswap support to the kernel to improve swap performance


Syonyk:
In personal experimentation working towards "Using the Raspberry Pi 3B/3B+ as a light duty desktop," I've discovered that fronting my swap file with zswap makes a _huge_ difference in system capability, most notably in how Chromium functions.  With stock settings, Chromium on the Pi 3B cannot load Google Inbox (https://inbox.google.com, assuming one's account is enabled) or Google Docs properly.  With zswap enabled, I can load both, simultaneously, and still have a usable system.

Under normal operation, with Chromium running, I have a fully responsive system with 300-500MB of memory swapped out - this being memory that, while not able to be discarded, is not actively in use.

I'm aware of the concerns about thrashing the SD card (and the glacial performance of said SD card under swap use), which is why zswap works so well.

zswap, in a nutshell, is a compressed frontend for swap.  It's quite configurable, with multiple compression options (lzo and lz4 being the most useful), several ways of storing compressed pages in memory (two and three "slots" per 4k page for compressed data), configurable in terms of percent of total system memory it will cache, etc.

It also includes a LRU (Least Recently Used) algorithm for evicting pages from compressed swap to physical disk swap when the cache is full, which prevents the priority inversion issues one can run into when using zram and physical swap with priorities set (zram fills up with the first stuff swapped out, which is typically least important, leaving physical disk to handle the later, higher priority swap that you'd like to keep in RAM).

Enabling zswap in the kernel requires the following changes to .config:
CONFIG_ZSWAP=y
CONFIG_ZPOOL=y
CONFIG_ZBUD=y
CONFIG_Z3FOLD=y

And, optionally, if you want lzo compression (somewhat better than lz4, but somewhat slower):
CONFIG_CRYPTO_LZO=y

I believe these can be built as modules as well, with no loss of functionality.  If the defaults are not zswap, they should be modules, but if the decision is made to use zswap on all installs, these should be built in.

To enable zswap, there needs to be a backing swapfile (already the case, though 100MB is a bit small in 2018), and the kernel needs to have zswap configured.  I've done this in /boot/cmdline.txt, though other locations would probably work as well.

At a minimum, this requires: zswap.enabled=1

I've also set the following on my install, though a smaller value may be a better default initially.  With Chromium being as memory hungry as it is, I normally raise this at runtime.
zswap.max_pool_percent=15

One can also set:
zswap.zpool=z3fold

[...]

PR #3432, opened 2020-01-27
Add zswap support to Raspberry Pi Kernels

Syonyk:
Context: https://github.com/raspberrypi/linux/issues/2649

Short context: zswap is a compressed/in-memory front end to disk based swap.  When a page is to be swapped out, if zswap is enabled, it tries to compress the page (using a configurable compression algorithm) and store it in memory.  If the page does not compress well, it is sent to disk, and if the configurable-sized in-memory cache fills up, pages are evicted with a LRU algorithm to disk.  On modern kernels with same filled page compression, compression ratios of 2-3x are seen in practice in typical desktop use.

This is distinct from zram, in that it properly evicts old pages to disk - zswap and zram are different features, with different behaviors.  The inversion present in zram (where old pages are stored perpetually in memory and fresh pages are evicted to disk) is not an issue in zswap.

Experimentally, on the Raspberry Pi hardware, zswap adds a substantial amount of "desktop usability" when given a properly sized swap file.  The Pi3 series goes from "unable to handle more than trivial use" to "usable as a light desktop," and the Pi4 benefits as well, depending somewhat on the amount of RAM present.

This patch does /not/ enable zswap by default, but simply adds the features to the kernel.  It adds zswap, zbud (a "two compressed page per page of RAM" storage algorithm), and z3fold (a "three compressed page per page of RAM" storage algorithm).  It is up to the user to properly configure the system to use these features, and other than the disk space used (examined below), this does not, in any way, change the system behavior.

====================

Cross compiling on an 64-bit x86 host with the reference Raspberry Pi cross compiler tools, size changes in the kernels/modules across the platforms:

bcmrpi_defconfig:
zImage: 4916kb to 4924kb, +8kb
Image: 10792kb to 10808kb, +16kb
Modules directory: 58632kb to 58644kb, +12kb

bcm2709_defconfig:
zImage: 5188kb to 5192kb, +4kb
Image: 11360kb to 11376kb, +16kb
Modules directory: 58596kb to 58608kb, +12kb

bcm2711_defconfig:
zImage: 5508kb to 5516kb: +8kb
Image: 12064kb to 12080kb: +16kb
Modules directory: 58796kb to 58804kb, +8kb

====================

For a full Raspbian install, this patch adds 52kb storage to the SD card image (zImage + modules), and offers users additional flexibility in how they can configure the system for various uses.

The intent is that these changes propagate forward and are the new defaults for Raspberry Pi kernels.


Syonyk:
Is there anything required from me to get this merged, or are config changes like this merged at a single point for future kernel releases?

Syonyk:
Is there anything else I need to do to have this merged, or is there no interest in it at this point?

zertyz:
I noticed the Raspberry Pi Foundation Engineer's arguments about **NOT ENABLING** zswap stating that they "didn't find a compelling reason to use zram or zswap", for some tests show improvements and others show degradation in performance on any particular combination of use case & media of their choice.

What I don't understand is that their conclusion must also mean "It is better that no user even try to use zswap" -- implied by the fact that the modules are marked as "do not even compile" -- despite the cost of the modules is just ~52k of storage?

The alternative -- forcing very determined users to have to build the modules on every kernel update -- seems to be unfair for further research as it restricts (or even impedes) the possibility of counter arguments as this "do not even include it" decision imposes a limit on the number of users who would care enough to test and report use cases where zswap are a better choice than no zswap.

Anyway... using this rationale to convince just myself, I'll try to build it as a module and report the improvements here.

zertyz:
I saw your PR adds swap built into the kernel and the other bits as modules.

Maybe it would be better to add all of them as modules? So that the only cost of this would be a tiny storage space...

Syonyk:
Well, nobody from the Pi foundation has offered an opinion on it, so I'm pretty sure they don't care.

zertyz:
Here I share my results, which favor zswap, in speed, by 21%:

The test: on as Raspberry Pi 1 512MiB of RAM, two instances of the following command have been issued at the same time:
        time tar -c /rw/ | dd bs=$((16 * 1024 * 1024)) iflag=fullblock count=2 | xz -7vv >/dev/null
        (this command brings both compressible and uncompressible pages into play, with the amount of memory fine tuned by the parameters to dd and xz)

Using the standard swap (zswap disabled), these commands finished -- each -- with:            
33554432 bytes (34 MB, 32 MiB) copied, 689.966 s, 48.6 kB/s
  100 %      9,022.5 KiB / 32.0 MiB = 0.275    47 KiB/s      11:30

real    11m31.509s
user    3m9.630s
sys     0m13.116s


With zswap enabled with the following parameters:
/sys/module/zswap/parameters/enabled:Y
/sys/module/zswap/parameters/max_pool_percent:50
/sys/module/zswap/parameters/zpool:zsmalloc
/sys/module/zswap/parameters/same_filled_pages_enabled:Y
/sys/module/zswap/parameters/compressor:zstd
    
I got:
33554432 bytes (34 MB, 32 MiB) copied, 573.753 s, 58.5 kB/s
  100 %      9,014.9 KiB / 32.0 MiB = 0.275    57 KiB/s       9:35

real    9m36.542s
user    3m5.735s
sys     0m29.307s


In any order I run the tests, the results are consistent -- after stopping/starting the swap devices and dropping the caches.

As a side note, I've also ran the tests with 5% max pool. It was just marginally better than no zswap:
33554432 bytes (34 MB, 32 MiB) copied, 683.776 s, 49.1 kB/s
  100 %      9,014.9 KiB / 32.0 MiB = 0.275    48 KiB/s      11:25

real    11m25.577s
user    3m6.389s
sys     0m28.182s


About the disk swap usage:
        no zswap -- 30MiB of swap were used
        zswap, 50% -- 70MiB of swap were used
        zswap, 5%  -- 60MiB of swap were used.        

My Raspberry Pi 1, 512MiB of RAM, was idle and had 447168k available for the tests (the rest is for the system and gpu 51MiB GPU RAM).

Conclusions:
  1) There are use cases in which zswap brings advantages -- just like on any PC. Raspberry Pi 1 is no exception here;
  2) Since activating zswap (without the proposed PR) requires each user to compile the kernel -- which takes ~40 hours -- and the cost of making it available is 52k of storage and 0k of RAM?, this PR should be accepted (defaulting to zswap being disabled, but making it available for whoever is interested on it).
  3) The advantages becomes even more obvious when more compressible pages come into play and if the count parameter on dd is higher, since the command takes some time before swapping starts

JamesH65:
@popcornmix @pelwell The use of ZSwap has been bubbling around for while now. If its a module with no impact unless enabled (if that is indeed the case), any reason not to have it?

popcornmix:
ZSWAP is a boolean not a module, so is likely to affect all users, not just those who enable it.
The memory cost may be small - but it needs to be measured and considered.

popcornmix:
Before:
```
$ ls -l kernel.img 
-rwxr-xr-x 1 dom dom 5968256 May  4 13:33 kernel.img

pi@pi4d:~ $ free
Mem:        3919816       51880     3794528        8636       73408     3748268
Swap:             0           0           0

pi@pi4d:~ $ du /lib/modules/5.4.35-v7l+/
58840	/lib/modules/5.4.35-v7l+/
```
After PR
```
$ ls -l kernel.img 
-rwxr-xr-x 1 dom dom 5977880 May  4 13:57 kernel.img

pi@pi4d:~ $ free
Mem:        3919812       50524     3795864        8628       73424     3749612
Swap:             0           0           0

pi@pi4d:~ $ du /lib/modules/5.4.35-v7l+/
58876	/lib/modules/5.4.35-v7l+/
```

So kernel.img up by ~10K, modules up by 36K, free mem down by 1344K without enabling zswap.

The loss of free memory for everyone is concerning.

pelwell:
I'll try and find out where the RAM is going.

Syonyk:
As noted, you can't build zswap as a module, only the various options it has.  I didn't think to check the RAM configs, though can try to replicate that.

My solution to a minor increase in memory use would be to enable zswap with the stock 100MB backing file, as it'll easily cover the delta as soon as there's any swap pressure, but I understand there's hesitation to do that based on the Pi1 behavior.

In my experience, zswap is pretty much a pure win for small values of "maximum memory" - I'd be fine with 10% as the default limit, though I'm not sure how to set that outside config files or the kernel command line.  I've found it to be worth doing on any Pi - the same-filled-page compression alone saves quite a bit of memory as swap pressure goes up.

How repeatable are the "free" results on boot?  I've not actually tested that, but can certainly spend some time on it.

pelwell:
There's a PR at https://github.com/raspberrypi/linux/pull/3600 that is an attempt at reducing the cost of enabling the ZSWAP config option but not enabling it at runtime. If it gets some testing, and the overhead for the disabled case isn't too bad, and if we can stand the thought of maintaining yet another downstream patch, then it might get merged.

Syonyk:
Please don't confuse zram and zswap - they're two entirely different features, entirely different code paths, etc.

Syonyk:
How are you getting the free values?  Just running it repeatedly on a freshly booted Pi with buster-lite, no network, at the console, I'm seeing over a megabyte difference in values over time.

I'll write a script to poll it and average it, but just random sample in time values doesn't seem like the right answer here.

pelwell:
I tried to make my measurements as soon after boot as possible, taking a sequence of 5 or 6 readings and picking the middle value (or the one that looked the most representative) - formal average sounds fine. A figure that excluding filesystem caching would be more stable and representative, but that might need some maths on the contents of /proc/meminfo.

swetoast:
Been testing this patch on my raspberry pi and it runs just fine same as on my cubieboard2, better then regular swapping 

```
Zswap:
.  accept_threshold_percent   90
.  compressor                 lz4
.  enabled                    Y
.  max_pool_percent           99
.  same_filled_pages_enabled  Y
.  zpool                      zbud
.  .  duplicate_entry         0                    
.  .  pool_limit_hit          0                    
.  .  pool_total_size         76943360             
.  .  reject_alloc_fail       0                   
.  .  reject_compress_poor    1151                
.  .  reject_kmemcache_fail   0                    
.  .  reject_reclaim_fail     0                    
.  .  same_filled_pages       10917                
.  .  stored_pages            34456                
.  .  written_back_pages      0                    
.  .  compress_ratio          54%                  
.  .  zswap_store/swap_store  141131776/148242432  95%

``

pelwell:
Did you test this patch (which won't get merged) or #3600 (which might if it gets enough positive feedback)?

swetoast:
tested this one but i can give the other one a whirl also isnt a big deal to compile

pelwell:
Thanks - it would be appreciated.

pelwell:
[ I don't understand why Linux's free memory after boot is so variable, even with no network. It makes serious A/B testing of memory usage almost impossible. ]

See #3626 for a new version of the deferred initialisation patches. The initialisation has been split so that some can still be done during the init phase, but pool creation can still be deferred. For some reason, building in zbud makes virtually no difference to the kernel size, as measured by the symbol `_end` - it literally seems to increase by only 38 bytes, which I can't explain - but for now it is left as a module.

pelwell:
#3626 was merged on the command line, and the latest rpi-update kernel includes it.

Syonyk:
Well, I'll close this one out - thanks!

PR #3440, opened 2020-01-30
overlays: Add ssd1306-spi overlay

MikeDK:
Add an overlay for SSD1306 based OLED boards using SPI.
This will load the staging fbtft driver.

Signed-off-by: Michael Kaplan <m.kaplan@evva.com>

### DIFF
--- a/arch/arm/boot/dts/overlays/README
+++ b/arch/arm/boot/dts/overlays/README
@@ -2427,6 +2427,17 @@ Params: address                 Location in display memory of first character.
         For more information refer to the device datasheet at:
         https://cdn-shop.adafruit.com/datasheets/SSD1306.pdf
 
+        
+Name:   ssd1306-spi
+Info:   Overlay for SSD1306 OLED via SPI using fbtft staging driver.
+Load:   dtoverlay=ssd1306-spi,<param>=<val>
+Params: speed                   SPI bus speed (default 10000000)
+        rotate                  Display rotation (0, 90, 180 or 270; default 0)
+        fps                     Delay between frame updates (default 25)
+        debug                   Debug output level (0-7; default 0)
+        dc_pin                  GPIO pin for D/C (default 24)
+        reset_pin               GPIO pin for RESET (default 25)
+
 
 Name:   superaudioboard
 Info:   Configures the SuperAudioBoard sound card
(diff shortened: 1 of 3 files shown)


MikeDK:
I just realized I forgot to provide overrides for width and height (as there are also 128x32 displays around on aliexpress etc)
But I am not sure how this looks like for a vendor parameter ... will the following work?
```
height = <&ssd1306>,"solomon,height:0";
```

I only have 128x64 displays, so I can't test the other ones :/

pelwell:
`height = <&ssd1306>,"solomon,height:0";` is correct. I'm not in a position to test anything beyond the overlay syntax, but for optional features like overlays best-effort is acceptable.

arch/arm/boot/dts/overlays/README
```diff
@@ -2427,6 +2427,17 @@ Params: address                 Location in display memory of first character.
         For more information refer to the device datasheet at:
         https://cdn-shop.adafruit.com/datasheets/SSD1306.pdf
 
+
```
pelwell: There's some stray whitespace on this line - please remove it.

pelwell (changes requested):
Apart from the additional parameters and the whitespace issue, you are missing an entry in the Makefile. Otherwise it looks good.

MikeDK:
Thanks for the hint with the Makefile ... I always forget about that ;)

Fixed the issues, and added a parameter for the height (only height because I never saw an SSD1306 OLED with a width other than 128)

arch/arm/boot/dts/overlays/ssd1306-spi-overlay.dts
```diff
@@ -73,9 +73,10 @@
 	__overrides__ {
 		speed		= <&ssd1306>,"spi-max-frequency:0";
 		rotate		= <&ssd1306>,"rotate:0";
-		fps		= <&ssd1306>,"fps:0";
+		fps			= <&ssd1306>,"fps:0";
```
pelwell: It looks like the stray whitespace found another home.

MikeDK: gnaaa I hate tabs ;)

PR #3442, opened 2020-01-31
overlays: Add sh1106-spi and ssd1351-spi overlays

MikeDK:
Add overlays for SH1106 and SSD1351 based OLED displays.
SH1106 is present in many 1.3 inch OLEDs and SSD1351 is present in 1.5 inch RGB OLEDs from AliExpress.

This will load the staging fbtft drivers.

### DIFF
--- a/arch/arm/boot/dts/overlays/Makefile
+++ b/arch/arm/boot/dts/overlays/Makefile
@@ -170,7 +170,7 @@ dtbo-$(CONFIG_ARCH_BCM2835) += \
 	spi6-2cs.dtbo \
 	ssd1306.dtbo \
 	ssd1306-spi.dtbo \
-	ssh1351-spi.dtbo \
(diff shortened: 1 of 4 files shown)


arch/arm/boot/dts/overlays/Makefile
```diff
@@ -169,6 +170,7 @@ dtbo-$(CONFIG_ARCH_BCM2835) += \
 	spi6-2cs.dtbo \
 	ssd1306.dtbo \
 	ssd1306-spi.dtbo \
+	ssh1351-spi.dtbo \
```
pelwell: This should be ssd1306-spi.dtbo

MikeDK: Thanks, but in fact it shall be ssd1351-spi.dtbo ;)

PR #3448, opened 2020-02-04
VCHIQ audio Device Tree support

pelwell:
This is an attempt to solve the problem of the BCM2835 audio driver - a client of VCHIQ - which in its downstream incarnation makes use of Device Tree properties. Until a recent reversion the audio device was being instantiated twice - once via DT and once directly by the VCHIQ parent.

VCHIQ kernel client drivers have been transformed into platform drivers instantiated explicitly by the parent. This solves the issue of device creation order, but it poses a problem for downstream use - the clients lose their Device Tree nodes. I understand that use of DT by "devices" that don't directly address hardware is against some of the DT principles, but the VCHIQ clients do control hardware - they just present proxies for services running on the RPi firmware. I therefore don't think it's unreasonable to use DT to describe them, allowing them to benefit from automatic pin configuration etc.

The solution is to give each child device an optional DT node, identified by name, and initialise the platform device object with a pointer to it. In this way the device creation is still controlled by the vchiq "bus" driver, but the children are free to use DT.

### DIFF
--- a/arch/arm/boot/dts/bcm270x-rpi.dtsi
+++ b/arch/arm/boot/dts/bcm270x-rpi.dtsi
@@ -70,13 +70,6 @@
 			status = "okay";
 		};
 
-		/* Onboard audio */
-		audio: audio {
-			compatible = "brcm,bcm2835-audio";
-			brcm,pwm-channels = <8>;
-			status = "disabled";
-		};
-
 		/* External sound card */
 		sound: sound {
 			status = "disabled";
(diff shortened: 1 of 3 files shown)


pelwell:
@popcornmix This is my proposal to resolve the audio problems in a way which is more closely aligned to upstream.
@vianpl Is this less unacceptable than the previous mess we had? I'm not expecting to upstream it, but it would be nice if it wasn't likely to get broken soon by another major upstream change.

pelwell:
Are there any objections to me merging this?

vianpl:
@pelwell Sadly, I wasn't here when you fixed the previous mess so my perspective is limited. But I don't think this part of the driver is likely to change. We've been using this device registering pattern all over the place and seems to work fine. So no objections.

pelwell:
Thank you for the feedback - I'm trying not to diverge too far from the upstream way.

PR #3453, opened 2020-02-05
overlays: add hdmi-backlight-hwhack-gpio-overlay

MikeDK:
This is a Devicetree overlay for GPIO based backlight on/off capability.

Use this if you have one of those HDMI displays whose backlight cannot be controlled via DPMS over HDMI and plan to do a little soldering to use an RPi gpio pin for on/off switching.

See: https://www.waveshare.com/wiki/7inch_HDMI_LCD_(C)#Backlight_Control

This was tested with a clone of the Waveshare "7 inch HDMI Touch LCD C" where I soldered two mosfets to override the backlight dip-switch.
When the overlay is loaded, a sysfs backlight node appears which can be used to modify the brightness value (0 or 1), and is even used by DPMS to switch the display backlight off after the configured timeout.
(On current Raspbian Buster Desktop, it's also possible to wakeup the display via a tap on the touch display :-) )

### DIFF
--- a/arch/arm/boot/dts/overlays/Makefile
+++ b/arch/arm/boot/dts/overlays/Makefile
@@ -51,6 +51,7 @@ dtbo-$(CONFIG_ARCH_BCM2835) += \
 	gpio-poweroff.dtbo \
 	gpio-shutdown.dtbo \
 	hd44780-lcd.dtbo \
+	hdmi-backlight-hwhack-gpio.dtbo \
 	hifiberry-amp.dtbo \
 	hifiberry-dac.dtbo \
 	hifiberry-dacplus.dtbo \
--- a/arch/arm/boot/dts/overlays/README
+++ b/arch/arm/boot/dts/overlays/README
@@ -883,6 +883,20 @@ Params: pin_d4                  GPIO pin for data pin D4 (default 6)
         display_width           Width of the display in characters
 
 
+Name:   hdmi-backlight-hwhack-gpio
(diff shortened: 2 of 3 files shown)


pelwell (approved):
This commit has received more than the usual amount of checking, for obvious reasons, but it looks fine. The name of the overlay is quite long, which is OK. However, the labels used within the overlay are also long which, although correct, is unnecessary because labels defined within overlays are effectively local and therefore don't have to be globally unique.

MikeDK:
@pelwell yes the labels are long, i know ... my first intention was to make it smaller, but then I realized that when the overlay is loaded, there appears a sysfs node link /sys/class/backlight/hdmi_backlight_hwhack_gpio ... so this is either the first or the second part in line 30 of the .dts file ("hdmi_backlight_hwhack_gpio: hdmi_backlight_hwhack_gpio {")

I liked the fact that this is a rather unique node name, so i left it that long...

Another thing is that i will add another overlay in the next few days, which will be called hdmi-backlight-hwhack-pwm, which does essentially the same, but with a pwm for brightness control instead of on/off switching...

pelwell:
In case you're in any doubt:
```
hdmi_backlight_hwhack_gpio: hdmi_backlight_hwhack_gpio {
```
The token before the `:` is a label, and the token after is a node name. Once the overlay is applied, the node name persists into the base DTB and the label disappears (it gets replaced by a unique `phandle` value within the node). It's the node name that ends up being the name of the backlight.

PR #3456, opened 2020-02-06
configs: Add KVM support to arm64 bcm2711_defconfig

pelwell:
This is an attempt to enable KVM support in the 64-bit Pi 4 kernel. I've done nothing beyond setting some obvious config options and verifying that the kernel doesn't grow too much (it doesn't), and that I/O performance doesn't suffer significantly (also negative). I would appreciate some testing by people who actually want to use KVM.

The patch is limited to 64-bit Pi 4 because BCM2711 has enough performance that a small slowdown shouldn't be a problem to most people, and KVM benefits from a 64-bit kernel.


pelwell:
Has anybody managed to test this? It's not going to get merged without some feedback.

macromorgan:
I can’t speak to your request specifically, but I’ve been running Sakaki’s BIS kernel for the past 4 months and it’s been fine for me.

https://github.com/sakaki-/bcm2711-kernel-bis

I can confirm the config options match your pull request, along with a few additional modules.

pelwell:
Merged manually.

PR #3461, opened 2020-02-10
defconfigs: enable CONFIG_BPF_LIRC_MODE2

HiassofT:
This enables support for IR decoding via BPF programs.

That feature is rather handy for home automation and mediacenter users as it allows decoding of a lot more IR remote protocols in the kernel, without having to use userspace lircd.

ir-keytable in buster already comes with a couple of generic BPF decoders (manchester, pulse distance and pulse length), later v4l-utils/ir-keytable versions add more decoders, including the very handy raw decoder which should support almost any protocol.

Kernel image size (tested with current 4.19.102 on RPi4) went up by about 3k, reported memory usage from the kernel was unchanged (probably because it's less than a 4k page).

4.19.102 with bcm2711_defconfig:
```
root@raspberrypi:~# ls -l /boot/kernel7l.img
-rwxr-xr-x 1 root root 5739688 Feb 10 16:17 /boot/kernel7l.img
root@raspberrypi:~# dmesg | grep Memory
[    0.000000] Memory policy: Data cache writealloc
[    0.000000] Memory: 3932696K/4050944K available (8192K kernel code, 688K rwdata, 2412K rodata, 2048K init, 848K bss, 52712K reserved, 65536K cma-reserved, 3198976K highmem)
```

4.19.102 with CONFIG_BPF_LIRC_MODE2 enabled:
```
root@raspberrypi:~# ls -l /boot/kernel7l.img
-rwxr-xr-x 1 root root 5742832 Feb 10 16:57 /boot/kernel7l.img
root@raspberrypi:~# dmesg | grep Memory
[    0.000000] Memory policy: Data cache writealloc
[    0.000000] Memory: 3932696K/4050944K available (8192K kernel code, 688K rwdata, 2412K rodata, 2048K init, 848K bss, 52712K reserved, 65536K cma-reserved, 3198976K highmem)
```


pelwell:
Could you bring the arm64 bcmrpi3_defconfig up to the same level as well?

HiassofT:
ah, completely missed that.

I noticed bcmrpi3_defconfig didn't enable any RC stuff at all, brought that in line with bcm2711_defconfig

pelwell:
Thanks!

PR #3472, opened 2020-02-19
Add SSD2543 touchscreen driver

MihailKuzhbanov:


pelwell:
Is this finished, in your mind? I have several problems with it:

1. The commits don't pass the scripts/checkpatch tests because they don't conform to kernel coding guidelines.
2. The author of this driver should submit it upstream.
3. Users would probably want to enable it with an overlay, which isn't part of the PR.
4. config.txt is not part of the kernel.
5. Don't include merge commits - just a simple flat list of commits rebased onto the current top-of-tree (which should really be rpi-5.4.y now - we'll be making that the default branch very soon).

MihailKuzhbanov:
I bad know git and it my miss. I want merge branches in my repo.

PR #3474, opened 2020-02-20
staging: vc04_services: Fix vcsm overflow bug when counting transactions

naushir:
The response block and local state were using u16 and u32 respectively
to represent transaction id.  When the former would wrap, there is a
mismatch and subsequent transactions will be marked as failures.


6by9:
LGTM. @pelwell?

PR #3476, opened 2020-02-21
OV5647 sensor driver updates

naushir:
Add a bunch of missing functionality:

- New 10-bit modes.
- Gain/exposure control
- Better defaults.


popcornmix:
No objections. I'll leave it to @6by9 to merge when happy.

drivers/media/i2c/ov5647.c
```diff
@@ -269,12 +299,21 @@ static int ov5647_set_virtual_channel(struct v4l2_subdev *sd, int channel)
 	return ov5647_write(sd, OV5647_REG_MIPI_CTRL14, channel_id | (channel << 6));
 }
 
+static int __sensor_init(struct v4l2_subdev *sd);
```
6by9: Generally the function implementations are put first rather than adding a prototype. Move the function.

drivers/media/i2c/ov5647.c
```diff
@@ -269,12 +299,21 @@ static int ov5647_set_virtual_channel(struct v4l2_subdev *sd, int channel)
...
+
 static int ov5647_stream_on(struct v4l2_subdev *sd)
 {
+	struct i2c_client *client = v4l2_get_subdevdata(sd);
 	struct ov5647 *ov5647 = to_state(sd);
 	u8 val = MIPI_CTRL00_BUS_IDLE;
 	int ret;
 
+	ret = __sensor_init(sd);
```
6by9, line 807: __sensor_init for 89 registers at 100kHz is going to take around 36ms, and hence every start streaming will be delayed by that much.
Do we delay here at the point that we've been asked to go, or at power on and each set_format to a new mode? I'm thinking shutter lag on captures and the like.

drivers/media/i2c/ov5647.c
```diff
@@ -392,7 +429,10 @@ static int ov5647_sensor_power(struct v4l2_subdev *sd, int on)
 			goto out;
 		}
 
-		ret = __sensor_init(sd);
+		/*
+		 * Ensure streaming off to make clock lane go into LP-11 state.
+		 */
+		ret = ov5647_stream_off(sd);
```
6by9, line 900: So we've just powered up, not set PLLs or anything, and yet only mess with CSI2 pad registers. Seems a little odd.

drivers/media/i2c/ov5647.c
```diff
@@ -479,38 +528,126 @@ static int ov5647_enum_mbus_code(struct v4l2_subdev *sd,
 				struct v4l2_subdev_pad_config *cfg,
 				struct v4l2_subdev_mbus_code_enum *code)
 {
-	if (code->index > 0)
+	if (code->index == 0 && ARRAY_SIZE(supported_modes_8bit))
+		code->code = MEDIA_BUS_FMT_SBGGR8_1X8;
+	else if (code->index == 0 && ARRAY_SIZE(supported_modes_8bit) == 0 &&
```
6by9, line 1001: Does checkpatch.pl not want this to be ```!code->index && !ARRAY_SIZE(support_modes_8bit)```? Similarly in the other lines.

drivers/media/i2c/ov5647.c
```diff
@@ -730,6 +743,119 @@ static int ov5647_parse_dt(struct device_node *np, struct ov5647 *sensor)
...
+
+	return ret;
+}
+
+static int ov5647_s_exposure_auto(struct v4l2_subdev *sd, u32 val)
+{
+	int ret;
+	u8 reg;
+
+	/* Everything except V4L2_EXPOSURE_MANUAL turns on AEC by
```
6by9, line 1233: Another checkpatch issue? Opening ```/*``` should be on a different line to the first line of the comment? Some subsystems have relaxed that rule.

drivers/media/i2c/ov5647.c
```diff
@@ -730,6 +743,119 @@ static int ov5647_parse_dt(struct device_node *np, struct ov5647 *sensor)
...
+		break;
+	case V4L2_CID_EXPOSURE_AUTO:
+		ov5647_s_exposure_auto(sd, ctrl->val);
+		break;
+	case V4L2_CID_ANALOGUE_GAIN:
+		ov5647_s_analogue_gain(sd, ctrl->val);
+		break;
+	case V4L2_CID_EXPOSURE:
+		ov5647_s_exposure(sd, ctrl->val);
+		break;
```
6by9: All these functions return an error value, which is then ignored.

drivers/media/i2c/ov5647.c
```diff
@@ -213,15 +213,384 @@ static struct regval_list ov5647_640x480[] = {
...
+	{0x3a0f, 0x58},
+	{0x3a10, 0x50},
+	{0x3a1b, 0x58},
+	{0x3a1e, 0x50},
+	{0x3a11, 0x60},
+	{0x3a1f, 0x28},
+	{0x4001, 0x02},
+	{0x4004, 0x04},
+	{0x4000, 0x09},
+	{0x4837, 0x19}, /* 0x16 doesn't work on Raspberry Pi, 0x19 does */
```
6by9: This is a new mode, so where did the settings come from that don't work?

drivers/media/i2c/ov5647.c
```diff
@@ -213,15 +213,384 @@ static struct regval_list ov5647_640x480[] = {
...
+	{0x3805, 0x3F},
+	{0x3806, 0x07},
+	{0x3807, 0xA3},
+	{0x3808, 0x05},
+	{0x3809, 0x10},
+	{0x380A, 0x03},
+	{0x380B, 0xCC},
+	{0x380C, 0x07},
+	{0x380D, 0x68},
+	{0x3811, 0x0c}, /* 0x0c (not 0x10) to remove the RHS white stripe */
```
6by9: So what was using 0x10? Redundant comment?

drivers/media/i2c/ov5647.c
```diff
@@ -232,14 +601,71 @@ static struct ov5647_mode supported_modes_8bit[] = {
 			.width = 640,
 			.height = 480
 		},
-		ov5647_640x480,
-		ARRAY_SIZE(ov5647_640x480)
+		ov5647_640x480_8bit,
+		ARRAY_SIZE(ov5647_640x480_8bit)
 	},
-	/* more modes below here... */
```
6by9: Why remove it here? it's still true that you could add more 8 bit modes here. The comment as a whole is pretty redundant though.

drivers/media/i2c/ov5647.c
```diff
@@ -232,14 +601,71 @@ static struct ov5647_mode supported_modes_8bit[] = {
...
+			.code = MEDIA_BUS_FMT_SBGGR10_1X10,
+			.colorspace = V4L2_COLORSPACE_SRGB,
+			.field = V4L2_FIELD_NONE,
+			.width = 640,
+			.height = 480
+		},
+		ov5647_640x480_10bit,
+		ARRAY_SIZE(ov5647_640x480_10bit)
+	},
+	/* more modes below here... */
```
6by9: Pretty redundant.

drivers/media/i2c/ov5647.c
naushir: This is a tricky one.  In normal usage, we are going to be calling S_FMT then S_STREAM straight after - meaning the delay just get shifted from one function to another and has no bearing on the capture lag.  Unless userland is doing something very clever/complicated by running s_fmt in a separate thread to some other initialisation, I don't think it makes any difference.  

Moving to s_fmt also means we need to test if the sensor is powered on, as this may not always be the case.

drivers/media/i2c/ov5647.c
naushir: This looks to be what the original code was doing, so I don't think @davidplowman wanted to change the behaviour.

naushir:
Fixes applied (including checkpatch --strict) in the latest push.

drivers/media/i2c/ov5647.c
6by9: S_FMT
STREAMON
STREAMOFF
STREAMON
STREAMOFF
is now slower on the second stream on. Do we care enough to worry about it?

drivers/media/i2c/ov5647.c
6by9: Having seen a similar patch for IMX219, it seems to be Renesas that has to see LP-11 state on the CSI2 lines.

If this is effectively how it was, then fine. It just looks odd.

drivers/media/i2c/ov5647.c
6by9: Is it only pointers that checkpatch complains over then? I'm suspecting so having looked at imx219 further.

6by9 (commented):
I started rereviewing, but a number of my previous comments are still outstanding. In particular those redundant comments in the register tables.

naushir:
> I started rereviewing, but a number of my previous comments are still outstanding. In particular those redundant comments in the register tables.

Oh no.  Looks like my attempt to rebase on to pf the latest rpi-5.4-y branch lost my updates to fix those issues :(

Please hold off until my next update!

naushir:
Latest patchset has all the fixes correctly applied.  All pass checkpatch --strict.

> S_FMT
> STREAMON
> STREAMOFF
> STREAMON
> STREAMOFF
> is now slower on the second stream on. Do we care enough to worry about it?

This is now addressed by having a bool set when we need to write the regs (i.e. power on, s_fmt).

drivers/media/i2c/ov5647.c
```diff
@@ -33,6 +33,7 @@
 #include <media/v4l2-fwnode.h>
 #include <media/v4l2-image-sizes.h>
 #include <media/v4l2-mediabus.h>
+#include <media/v4l2-ctrls.h>
```
6by9: Headers should be in alphabetical order.

6by9 (commented):
1st patch ought to have 
```Fixes: 3c2472a [media] media: i2c: Add support for OV5647 sensor```
in the commit log. Otherwise fine.

One minor nit otherwise.

drivers/media/i2c/ov5647.c
naushir: Fixed.

naushir:
@popcornmix or @pelwell, could you merge this when you can please?

PR #3475, opened 2020-02-20
bcm2835-unicam driver for unicam

naushir:
This PR adds the existing (in the 4.19.y kernel tree) bcm2835-unicam as well as adding support for an embedded data node in the driver.

Modify the imx219 driver to expose embedded data on a new pad.

PR #3497, opened 2020-03-11
rpi-5.4.y Update imx219 driver to the version accepted upstream.

6by9:
Reverts the currently merged version first, then adds the new version.
One small fixup required to the DT overlay which is done as a separate patch.


6by9:
@naushir FYI Merging the newer version of this driver.

popcornmix:
Looks like it a needs a rebase.
Always happy to switch to upstream versions of files, so merge if you've seen it working.

6by9:
Needs a rebase? I see 5 commits and Github reporting no conflicts.

Yes, I've proved it works (although that needs the unicam driver which we haven't merged here yet).

naushir:
Unicam PR is at https://github.com/raspberrypi/linux/pull/3475.  Still needs reviewing and possibly some updates pending my staggered write investigations.

PR #3498, opened 2020-03-11
Backport a load of cmdline "video=" token parsing changes

6by9:
Backporting a set of patches from 5.6 that improve allow better parsing of the video= strings from the kernel command line.
The base code required the resolution to be specified under all circumstances, whereas with these we can now omit the resolution and only add the overscan margins or rotations.


popcornmix:
Seems fine to me. @pelwell any objections?

PR #3500, opened 2020-03-12
Kbuild: Allow .dtbo overlays to be built, adjust.

malus-brandywine:
This is adjustment to commit
d368ceaacdccd7732dc97d1d7987bdf7149d62e3 "kbuild: Allow .dtbo overlays to be built piecemeal"

prepare3 target has gone from mainline tree in branch 5.4.y

Signed-off-by: Nataliya Korovkina <malus.brandywine@gmail.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1237,7 +1237,7 @@ ifneq ($(dtstree),)
 %.dtb: include/config/kernel.release scripts_dtc
 	$(Q)$(MAKE) $(build)=$(dtstree) $(dtstree)/$@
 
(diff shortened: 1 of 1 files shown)


pelwell:
Thanks - this isn't a facility we use, so we didn't notice the breakage.

malus-brandywine:
Welcome! I probably have to repeat the drill for 5.5 and 5.6

pelwell:
The ports to 5.5 and 5.6 are already there.

malus-brandywine:
Great, thank you!

PR #3505, opened 2020-03-13
V4L2 HEVC driver

6by9:
This is @jc-kynesim 's driver with a load of checkpatch cleanups done. It's not quite checkpatch clean, but nearly.
- Some long lines (mainly as the V4L2 control defines are so long to start with)
- Uncommented spinlocks
- Uncommented memory barrier (possibly redundant as I've switched to readl/writel instead of John's volatile mapped accesses)
- No signed off on the main patch.
- MAINTAINERS - need to decide who to add to that.
- DT binding doc.

I've checked that it works with FFmpeg in the previous test cases.
John is still working through an issue with @popcornmix that Kodi stops displaying or something. Flagged as WIP until that is resolved.


popcornmix:
I think there are a couple of additional commits (either on my kodi/rpi4-kms tree or on buildbot).
"Stash: add threaded callback" "Dynamically size all buffers" which fixes an issue when hardware signals buffer sizes are insufficient and needs a realloc/retry.

I think the current belief is the kodi black screen issue only occurs with the hardest streams and is a bad interaction when running behind. It may be something that can be improved, but otherwise isn't a blocker.

drivers/media/common/v4l2-tpg/v4l2-tpg-core.c
```diff
@@ -309,6 +309,7 @@ bool tpg_s_fourcc(struct tpg_data *tpg, u32 fourcc)
 	case V4L2_PIX_FMT_NV21M:
 		tpg->buffers = 2;
 		/* fall through */
+	case V4L2_PIX_FMT_SAND8:
```
6by9: Do we actually need to use the test pattern generator?

Documentation/media/uapi/v4l/pixfmt-nv12.rst
```diff
@@ -38,6 +38,14 @@ with a Cr byte.
...
 many pad bytes after its rows.
 
+``V4L2_PIX_FMT_NV12_COL128`` is the tiled version of
+``V4L2_PIX_FMT_NV12`` with the image broken down into 128 pixel wide columns of
+Y followed by the associated combined CbCr plane.
+The normal bytesperline is effectively fixed at 128. However the format
+requires knowledge of the stride between columns, therefore the bytesperline
+value has been repurposed to denote the number of 128 byte long lines between
+the start of each column.
+
```
6by9, line 48: Docs need finishing, and 10bit version adding.

include/uapi/linux/videodev2.h
```diff
@@ -738,6 +738,8 @@ struct v4l2_pix_format {
 #define V4L2_PIX_FMT_INZI     v4l2_fourcc('I', 'N', 'Z', 'I') /* Intel Planar Greyscale 10-bit and Depth 16-bit */
 #define V4L2_PIX_FMT_SUNXI_TILED_NV12 v4l2_fourcc('S', 'T', '1', '2') /* Sunxi Tiled NV12 Format */
 #define V4L2_PIX_FMT_CNF4     v4l2_fourcc('C', 'N', 'F', '4') /* Intel 4-bit packed depth confidence information */
+#define V4L2_PIX_FMT_SAND8    v4l2_fourcc('S', 'A', 'N', 'D') /* RPI 8 bit sand */
```
6by9: What was wrong with the V4L2_PIX_FMT_NV12_COL128 format I'd defined in the earlier patch?

drivers/media/v4l2-core/v4l2-ctrls.c
```diff
@@ -974,6 +974,7 @@ const char *v4l2_ctrl_get_name(u32 id)
 	case V4L2_CID_MPEG_VIDEO_HEVC_SPS:			return "HEVC Sequence Parameter Set";
 	case V4L2_CID_MPEG_VIDEO_HEVC_PPS:			return "HEVC Picture Parameter Set";
 	case V4L2_CID_MPEG_VIDEO_HEVC_SLICE_PARAMS:		return "HEVC Slice Parameters";
+	case V4L2_CID_MPEG_VIDEO_HEVC_SCALING_MATRIX:	return "HEVC Scaling Matrix";
```
6by9: Need to expand out the commit text, and add a Signed-off-by:

6by9 (commented):
Sorry, commenting on my own PR as things are more obvious as soon as you've created it.
Still a couple more jobs to do.

6by9:
John had told me 91187ff10c0447762b641cc7757fadf6914b3696 on his branch test/5.4/v4l2_5 of buildbot was the latest. I'm happy to squash other commits in, but wanted/needed to get it checkpatch clean before considering submitting.
John was happy that redoing subsequent patches on top of a cleaned up base shouldn't be too onerous.

popcornmix:
Sure. Should be functional (for vast majority of streams) without the additional patches.
No problem with getting this merged soon and adding to it later.

include/uapi/linux/videodev2.h
jc-kynesim: > What was wrong with the V4L2_PIX_FMT_NV12_COL128 format I'd defined in the earlier patch?

Probably nothing - either I wrote that line before I found your patch or I failed to notice it.

drivers/media/common/v4l2-tpg/v4l2-tpg-core.c
jc-kynesim: Probably not - I think I added that line when I was adding SAND cases everywhere I could find NV12 used to make sure I got frames through

jc-kynesim:
> I think there are a couple of additional commits (either on my kodi/rpi4-kms tree or on buildbot).
> "Stash: add threaded callback" "Dynamically size all buffers" which fixes an issue when hardware signals buffer sizes are insufficient and needs a realloc/retry.
> I think the current belief is the kodi black screen issue only occurs with the hardest streams and is a bad interaction when running behind. It may be something that can be improved, but otherwise isn't a blocker.

So is your current belief that my dev/5.4/fix_pu_alloc_1 branch actually works? 'cos if so then I would like to merge that in now

6by9:
Tell me which patches are required and I'll merge them in here, along with sorting out the docs and other tasks. Ideally we really do want rpivid all as one squashed commit if we're looking to upstream. 

(Need to get a mainline 5.6 tree booting on a Pi4 before upstreaming to be able to test it).

popcornmix:
> So is your current belief that my dev/5.4/fix_pu_alloc_1 branch actually works? 'cos if so then I would like to merge that in now

It clearly fixes an issue when buffer required by hardware exceeds default size. 
ffmpeg can now decode the whole file rather than stalling at the big frame.

It allows the problematic file to get further in kodi but we see other issues caused by performance being on the edge.

Overall it seems beneficial so happy to include it.

jc-kynesim:
> Uncommented spinlocks

What sort of comments are wanted?  - I'm happy to add them

> Uncommented memory barrier (possibly redundant as I've switched to readl/writel instead of John's volatile mapped accesses)

Not actually the best fix I think - it adds many more memory barrier ops than needed - the bulk of them just want to be writel_relaxed() with only the last in the set being a writel()

jc-kynesim:
> Tell me which patches are required and I'll merge them in here, along with sorting out the docs and other tasks. Ideally we really do want rpivid all as one squashed commit if we're looking to upstream.
> (Need to get a mainline 5.6 tree booting on a Pi4 before upstreaming to be able to test it).

As the branch is currently set up for debug it is probably easiest (though non-trivial) for me to recreate it on your patch and remove the unwanted debug as I go rather than get you to do it - I'll also fix up spinlock comments etc.

6by9:
>> Uncommented spinlocks

> What sort of comments are wanted? - I'm happy to add them

What's being protected by them. Either explicitly commented, or blocked together, eg
https://elixir.bootlin.com/linux/latest/source/include/media/v4l2-dev.h#L289
```
	/* V4L2 file handles */
	spinlock_t		fh_lock;
	struct list_head	fh_list;

```
If you provide guidance, I can make the edits.
>> Uncommented memory barrier (possibly redundant as I've switched to readl/writel instead of John's volatile mapped accesses)

> Not actually the best fix I think - it adds many more memory barrier ops than needed - the bulk of them just want to be writel_relaxed() with only the last in the set being a writel()

If so then we can make a new function the final [write|read]l, but raw accesses to volatile mapped registers isn't going to be acceptable upstream.

drivers/staging/media/rpivid/Kconfig
```diff
@@ -0,0 +1,17 @@
+# SPDX-License-Identifier: GPL-2.0
+
+# SPDX-License-Identifier: GPL-2.0
```
lategoodbye: I think the second one isn't necessary

drivers/staging/media/rpivid/Kconfig
```diff
@@ -0,0 +1,17 @@
...
+	depends on MEDIA_CONTROLLER
+	depends on OF
+	depends on MEDIA_CONTROLLER_REQUEST_API
+	select VIDEOBUF2_DMA_CONTIG
+	select V4L2_MEM2MEM_DEV
+	help
+	  Support for the Rpi H265 h/w decoder.
+
+	  To compile this driver as a module, choose M here: the module
+	  will be called sunxi-cedrus.
```
lategoodbye: Oops, copy & paste

drivers/staging/media/rpivid/rpivid_hw.c
```diff
@@ -0,0 +1,259 @@
...
+	}
+
+#if !OPT_DEBUG_POLL_IRQ
+	irq_dec = platform_get_irq(dev->pdev, 0);
+	if (irq_dec <= 0)
+		return irq_dec;
+	ret = devm_request_irq(dev->dev, irq_dec, rpivid_irq,
+			       0, dev_name(dev->dev), dev);
+	if (ret) {
+		dev_err(dev->dev, "Failed to request IRQ\n");
```
lategoodbye: The return code might be interesting here

drivers/staging/media/rpivid/rpivid_video.c
```diff
@@ -0,0 +1,572 @@
...
+#define RPIVID_MIN_HEIGHT	16U
+#define RPIVID_MAX_WIDTH	4096U
+#define RPIVID_MAX_HEIGHT	4096U
+
+static inline struct rpivid_ctx *rpivid_file2ctx(struct file *file)
+{
+	return container_of(file->private_data, struct rpivid_ctx, fh);
+}
+
+/* constrian x to y,y*2 */
```
lategoodbye: constrain?

drivers/staging/media/rpivid/rpivid_video.c
```diff
@@ -0,0 +1,572 @@
...
+static inline unsigned int constrain2x(unsigned int x, unsigned int y)
+{
+	return x < y ? y : x > y * 2 ? y : x;
```
lategoodbye: this is very hard to read

drivers/staging/media/rpivid/rpivid_video.c
```diff
@@ -0,0 +1,572 @@
...
+	    ctb_log2_size_y > 6 ||
+	    max_tb_log2_size_y > 5 ||
+	    max_tb_log2_size_y > ctb_log2_size_y ||
+	    sps->max_transform_hierarchy_depth_inter >
+				(ctb_log2_size_y - min_tb_log2_size_y) ||
+	    sps->max_transform_hierarchy_depth_intra >
+				(ctb_log2_size_y - min_tb_log2_size_y) ||
+	    /* Check pcm stuff */
+	    sps->num_short_term_ref_pic_sets > 64 ||
+	    sps->num_long_term_ref_pics_sps > 32)
```
lategoodbye: Please check for single error cases and return with 0

drivers/staging/media/rpivid/rpivid_video.c
```diff
@@ -0,0 +1,572 @@
...
+static int rpivid_g_fmt_vid_out(struct file *file, void *priv,
+				struct v4l2_format *f)
+{
+	struct rpivid_ctx *ctx = rpivid_file2ctx(file);
+
+	f->fmt.pix = ctx->src_fmt;
+	return 0;
+}
+
+static inline void cpy_colour(struct v4l2_pix_format *d,
```
lategoodbye: copy_color ?

drivers/staging/media/rpivid/rpivid_video.c
```diff
@@ -0,0 +1,572 @@
...
+	.vidioc_dqbuf			= v4l2_m2m_ioctl_dqbuf,
+	.vidioc_prepare_buf		= v4l2_m2m_ioctl_prepare_buf,
+	.vidioc_create_bufs		= v4l2_m2m_ioctl_create_bufs,
+	.vidioc_expbuf			= v4l2_m2m_ioctl_expbuf,
+
+	.vidioc_streamon		= v4l2_m2m_ioctl_streamon,
+	.vidioc_streamoff		= v4l2_m2m_ioctl_streamoff,
+
+	.vidioc_try_decoder_cmd		=
+				v4l2_m2m_ioctl_stateless_try_decoder_cmd,
```
lategoodbye: In this case it's okay to violate the 80 char limit

drivers/staging/media/rpivid/rpivid_video.c
```diff
@@ -0,0 +1,572 @@
...
+
+	return pf;
+}
+
+static struct v4l2_pix_format
+rpivid_hevc_default_dst_fmt(struct rpivid_ctx * const ctx)
+{
+	const struct v4l2_ctrl_hevc_sps * const sps =
+		rpivid_find_control_data(ctx, V4L2_CID_MPEG_VIDEO_HEVC_SPS);
+	struct v4l2_pix_format pix_fmt = {
```
lategoodbye, line 230: According to the definition rpivid_find_control_data could return NULL, so this impossible in this and some other cases?

drivers/staging/media/rpivid/rpivid.c
```diff
@@ -0,0 +1,430 @@
...
+	{ /* sentinel */ }
+};
+MODULE_DEVICE_TABLE(of, rpivid_dt_match);
+
+static struct platform_driver rpivid_driver = {
+	.probe		= rpivid_probe,
+	.remove		= rpivid_remove,
+	.driver		= {
+		.name		= RPIVID_NAME,
+		.owner = THIS_MODULE,
```
lategoodbye: Please drop this, it's already done by platform_driver

drivers/staging/media/rpivid/rpivid_h265.c
```diff
@@ -0,0 +1,2075 @@
...
+#define RPI_FRAMESIZE 0x802C
+#define RPI_MVBASE 0x8030
+#define RPI_MVSTRIDE 0x8034
+#define RPI_COLBASE 0x8038
+#define RPI_COLSTRIDE 0x803C
+#define RPI_CURRPOC 0x8040
+
+//////////////////////////////////////////////////////////////////////////////
+
+struct RPI_CMD {
```
lategoodbye: i think this should be lower case

drivers/staging/media/rpivid/rpivid_h265.c
```diff
@@ -0,0 +1,2075 @@
...
+#define RPI_OUTYSTRIDE 0x801C
+#define RPI_OUTCBASE 0x8020
+#define RPI_OUTCSTRIDE 0x8024
+#define RPI_STATUS2 0x8028
+#define RPI_FRAMESIZE 0x802C
+#define RPI_MVBASE 0x8030
+#define RPI_MVSTRIDE 0x8034
+#define RPI_COLBASE 0x8038
+#define RPI_COLSTRIDE 0x803C
+#define RPI_CURRPOC 0x8040
```
lategoodbye: use tabs for indenting would improve readability

drivers/staging/media/rpivid/rpivid_h265.c
```diff
@@ -0,0 +1,2075 @@
...
+
+	return de->cmd_len++;
+}
+
+static int ctb_to_tile(unsigned int ctb, unsigned int *bd, int num)
+{
+	int i;
+
+	for (i = 1; ctb >= bd[i]; i++)
+		; // bd[] has num+1 elements; bd[0]=0; see hevc_ps.c
```
lategoodbye: i assume this source reference doesn't apply anymore

drivers/staging/media/rpivid/rpivid_h265.c
```diff
@@ -0,0 +1,2075 @@
...
+		153, 121, 136, 122, 169, 208, 166, 167, 154, 152, 167, 182, 125,
+		110, 124, 110, 95,  94,  125, 111, 111, 79,  125, 126, 111, 111,
+		79,  108, 123, 93,  125, 110, 124, 110, 95,  94,  125, 111, 111,
+		79,  125, 126, 111, 111, 79,  108, 123, 93,  121, 140, 61,  154,
+		107, 167, 91,  107, 107, 167, 139, 139, 170, 154, 139, 153, 139,
+		123, 123, 63,  124, 166, 183, 140, 136, 153, 154, 166, 183, 140,
+		136, 153, 154, 166, 183, 140, 136, 153, 154, 170, 153, 138, 138,
+		122, 121, 122, 121, 167, 151, 183, 140, 151, 183, 140, 0,   0,
+	},
+};
```
lategoodbye, line 479: a comment about these values would be nice

drivers/staging/media/rpivid/rpivid_h265.c
```diff
@@ -0,0 +1,2075 @@
...
+
+static void WriteProb(struct rpivid_dec_env *const de,
```
lategoodbye: Please avoid camel case

drivers/staging/media/rpivid/rpivid_h265.c
```diff
@@ -0,0 +1,2075 @@
...
+				*d++ = *s;
+				*d++ = *s;
+				*d++ = *s;
+				*d++ = *s++;
+			}
+		}
+		dst0[0] = dc;
+		break;
+	}
+	}
```
lategoodbye: the indentation looks broken

drivers/staging/media/rpivid/rpivid_h265.c
```diff
@@ -0,0 +1,2075 @@
...
+
+	pred_weight_table = &sh->pred_weight_table;
+
+	//    v4l2_info(&dev->v4l2_dev, "rpivid_h265_setup: ts=%d:%d:%d.%d\n",
+	//              run->src->timecode.hours,
+	//              run->src->timecode.minutes,
+	//              run->src->timecode.seconds,
+	//              run->src->timecode.frames);
+	//    v4l2_info(&dev->v4l2_dev, "vptr=%p\n",
+	//    vb2_plane_vaddr(&run->src->vb2_buf, 0));
```
lategoodbye: Please drop this or make it a debug message

drivers/staging/media/rpivid/rpivid_h265.c
```diff
@@ -0,0 +1,2075 @@
...
+	unsigned int decode_order;
+
+	int phase_no; // Current phase (i.e. the last one we waited for)
+	struct rpivid_dec_env *phase_wait_q_next;
+
+	struct RPI_CMD *cmd_fifo;
+	unsigned int cmd_len, cmd_max;
+	unsigned int num_slice_msgs;
+	unsigned int PicWidthInCtbsY;
+	unsigned int PicHeightInCtbsY;
```
lategoodbye: please no camel case

drivers/staging/media/rpivid/rpivid_h265.c
```diff
@@ -0,0 +1,2075 @@
...
+	for (i = 0; i != HEVC_MAX_REFS; ++i)
+		aux_q_release(ctx, &s->ref_aux[i]);
+	aux_q_release(ctx, &s->frame_aux);
+
+	kfree(s);
+}
+
+static struct rpivid_dec_state *dec_state_new(void)
+{
+	struct rpivid_dec_state *s = kzalloc(sizeof(*s), GFP_KERNEL);
```
lategoodbye: i think it's preferred to use kzalloc directly

drivers/staging/media/rpivid/rpivid_h265.c
```diff
@@ -0,0 +1,2075 @@
...
+		de->state = s->frame_end ? RPIVID_DECODE_ERROR_DONE :
+					   RPIVID_DECODE_ERROR_CONTINUE;
+}
+
+//////////////////////////////////////////////////////////////////////////////
+// Handle PU and COEFF stream overflow
+
+// Returns:
+// -2 Other error
+// -1 Out of coeff space
```
lategoodbye: Maybe we can use standard error code like -ENOMEM

lategoodbye (changes requested):

drivers/staging/media/rpivid/rpivid_video.c
popcornmix: Based on comment I'd have expected ``x < y ? y : x > y * 2 ? y * 2 : x;``

drivers/staging/media/rpivid/Kconfig
6by9: Thanks for the review. I thought this was worth saying twice ;-)

drivers/staging/media/rpivid/Kconfig
6by9: I'm also renaming the module to rpivid-hevc. rpivid-v4l2 has the potential to get confused against another codec block.

drivers/staging/media/rpivid/rpivid_video.c
6by9: I'll break it up onto multiple lines once someone confirms whether it is functionally correct or not.

drivers/staging/media/rpivid/rpivid_h265.c
6by9: @jc-kynesim Please advise as to what these are.

drivers/staging/media/rpivid/rpivid_h265.c
6by9: I thought I'd killed it all as checkpatch had complained in places :-( Done

drivers/staging/media/rpivid/rpivid_h265.c
6by9: It's not actually as it's the end of the case which was declaring variables in that scope, and the end of the overall switch statement, but I'll agree it looks strange.

Seeing as d, x, and y were used in 2 of the 4 cases I've switched to defining them at the top of the function.

drivers/staging/media/rpivid/rpivid_video.c
6by9: Where checking eg sps->pic_width_in_luma_samples is between 32 & 4096 I've kept it as one ```if```, but split the rest up.

drivers/staging/media/rpivid/rpivid_video.c
jc-kynesim: The original code was, in fact, correct - the comment was misleading.  The intention was to constrain buffer sizes - if they were in the range min to min*2 then accept them, otherwise choose min as if I'm going to reject the users suggestion I might as well choose the "correct" size.

drivers/staging/media/rpivid/rpivid_h265.c
6by9: I'm not sure if it's what you mean, but the function is actually giving us no gain.
```if (!s) return NULL``` is equivalent to ```return s```, and that is what happens in the success case. I'll drop the function.

File: drivers/staging/media/rpivid/rpivid_h265.c (hunk 24 above)
Comment (jc-kynesim) [reply drivers/staging/media/rpivid/rpivid_h265.c]: Internal status not communicated back to user and anyway rewritten in my current patch

Comment (6by9) [reply drivers/staging/media/rpivid/rpivid_h265.c]: The caller does nothing other than ```if (check_status != 0)```, so I've amended it to -EINVAL and -ENOMEM.

**Comment by 6by9** [comment]
Branch updated based on lategoodbye's review comments. Now to fix up my own.

File: drivers/staging/media/rpivid/rpivid_h265.c (hunk 23 above)
Comment (lategoodbye) [reply drivers/staging/media/rpivid/rpivid_h265.c]: Yes, this is what i tried to say

**Comment by 6by9** [comment]
Branch https://github.com/6by9/linux/tree/rpi-5.4.y-hevc-wip pushed with fixups, early DT binding, and John's latest improvements. Now going to squash it down to update this PR.
(Still need to complete the V4L2 pixel format docs)

Note that https://github.com/6by9/linux/commit/1c73ca4e799f03217cea879eaf40d6c601a2b0e6 needs to be dropped. SAND as a fourcc means nothing to anyone outside Broadcom/Pi Towers, but it needs FFmpeg to be updated to match, and I haven't rebuilt that at present.

File: drivers/staging/media/rpivid/rpivid_video.c [hunk 25]
```diff
@@ -0,0 +1,588 @@
...
+
+static int rpivid_start_streaming(struct vb2_queue *vq, unsigned int count)
+{
+	struct rpivid_ctx *ctx = vb2_get_drv_priv(vq);
+	struct rpivid_dev *dev = ctx->dev;
+	int ret = 0;
+
+	switch (ctx->src_fmt.pixelformat) {
+	case V4L2_PIX_FMT_HEVC_SLICE:
+		ctx->current_codec = RPIVID_CODEC_H265;
```
6by9: @jc-kynesim This only ever gets set to RPIVID_CODEC_H265.

Based on other recent conversations, do we really see this driver supporting multiple codecs? I read the conclusion as "no", therefore we can strip out a load of indirection.

drivers/staging/media/rpivid/rpivid_h265.c
jc-kynesim: CABAC probability init values - commented usefully in my current patch

drivers/staging/media/rpivid/rpivid_video.c
jc-kynesim: I've stripped some out from the original but not all so go ahead - though you probably want to keep the checks that we are, in fact, being asked to do what we expect.

6by9: I have kept the check that we are getting V4L2_PIX_FMT_HEVC_SLICE, but anything else should be impossible to get through rpivid_s_fmt_vid_cap anyway.

I have also kept dev->dec_ops as a function pointer, so actually you could change the function pointer on change of format to achieve the same end result.

popcornmix:
Any reason not to add this to arm64 defconfig?

6by9:
> Any reason not to add this to arm64 defconfig?

Probably not, but I haven't tried it.
Allocations are from the CMA heap, so those should still be under the same restrictions as 32-bit world.

lategoodbye:
> Branch https://github.com/6by9/linux/tree/rpi-5.4.y-hevc-wip pushed with fixups, early DT binding, and John's latest improvements. Now going to squash it down to update this PR.
> (Still need to complete the V4L2 pixel format docs)

New DT binding should be in YAML format:
https://www.kernel.org/doc/Documentation/devicetree/writing-schema.md

Unfortunately i don't know a good tutorial. Copy & paste from media is a good starting point.

Finally you need to make sure that this step succeed:
make dtbs_check DT_SCHEMA_FILES=Documentation/devicetree/bindings/media/rpivid_hevc.yaml

6by9:
> New DT binding should be in YAML format:

Thank you. As I was backporting to 5.4 they are all still text files. Yes, copy/paste is always a good starting point.

6by9:
Docs completed and look reasonable.
DT binding updated to YAML.

We have 3 patches from @jc-kynesim without Signed-Off-By: lines.
John - please review the patches and confirm if you're happy for me to attach them. Refer to https://www.kernel.org/doc/html/latest/process/submitting-patches.html#developer-s-certificate-of-origin-1-1 as to the meaning of SoB.

The one bit that may be wanted is clock control. The firmware appears to have clock control for the HEVC block, but this driver isn't controlling it. Once Maxime's updated clock driver is merged then it becomes easy, so I'm tempted to skip it for now. It can be added later as an optional clock property,
Thoughts?

popcornmix:
Clock control is needed (I believe it was in @jc-kynesim 's earlier ffmpeg hevc driver).
It should read the max hevc clock and request that while it's active.
Otherwise we'll run at half speed when the arm is <50% busy (which it typically is) and fail to keep up.

6by9:
As above, clock control is going to need to wait for Maxime's driver to be merged, or at least the discussions completed.
I don't fancy adding it into clk-bcm2835 unless really needed.

popcornmix:
It can only safely use the firmware clock driver (or mailbox directly, but that would be frowned upon upstream).

We can merge this (with note that performance isn't full without force_turbo=1 or performance governor) and add the clock request when available.

6by9:
```vcgencmd measure_clock hevc``` is reporting 250MHz for me on an idle system. I can't find the max permitted quickly.

It looks like there are ```hevc_freq_min``` and ```hevc_freq_max``` config.txt options, so that would be one way around it.
Memory says there is a way of allocating a NULL clock (code review saying NULL was a valid return from ```devm_clk_get```). If I can quickly find the relevant bit for that then I'll add it to the current configuration, and we can replace it with the firmware clock driver at a suitable point in the future.

Do we just always ask for max, or is there any better logic around?

**Comment by popcornmix** [comment]
This was logic for ffmpeg based hevc driver https://github.com/popcornmix/FFmpeg/commit/d9d273b9a642565807bf63e55457312cccd23e37

**Comment by jc-kynesim** [comment]
> We have 3 patches from @jc-kynesim without Signed-Off-By: lines.
> John - please review the patches and confirm if you're happy for me to attach them. Refer to https://www.kernel.org/doc/html/latest/process/submitting-patches.html#developer-s-certificate-of-origin-1-1 as to the meaning of SoB.

I'm certainly happy with the decoder patch, at one point I'm pretty sure the mem2mem patch had a signed off line when seems to have become lost so I'm happy with that, the "scaling matrix" patch is mostly jernejs work (I've added a flag) & has, at least in part I think, been submitted to upstream once where it faded into oblivion.

**Comment by popcornmix** [comment]
@6by9 needs a rebase after 5.4 update.
What else does this need before merging?

**Comment by 6by9** [comment]
> What else does this need before merging?

Clock driver hooks.
I have what I hope is the correct code, and DT is advertising a fixed clock for now. 
I'm trying to get it tested and pushed today, but also need to head into the office to collect some kit.

**Comment by 6by9** [comment]
Updated and with clock changes (fixed clock defined in DT overlay at present).

@jc-kynesim I've added your SoB to the driver, cherry-picked jernej's patch from https://patchwork.linuxtv.org/patch/60728/ and added an extra one that adds your one define, and hopefully now have a suitable tree!

**Comment by 6by9** [comment]
Except I've messed something up in the last fixup! I've obviously lost part of John's patch as slice_segment_addr and num_entry_point_offsets are missing.

**Comment by popcornmix** [comment]
What effect does the 500MHz hevc clock request in DT have?

**Comment by 6by9** [comment]
@jc-kynesim So you appear to have merged 
- https://patchwork.linuxtv.org/patch/60728/
- https://patchwork.linuxtv.org/patch/60725/
- Adding define V4L2_HEVC_SLICE_PARAMS_FLAG_DEPENDENT_SLICE_SEGMENT
- And added num_entry_point_offsets and entry_point_offset_minus1[256] from somewhere. Do you have any idea as where?

With these being uapi headers that clients will be using, heritage is extremely important.

**Comment by 6by9** [comment]
> What effect does the 500MHz hevc clock request in DT have?

Absolutely none.
Until we have Maxime's clock driver this is just a place holder so that the driver can ask a clock to run at 500MHz without error. There is currently no clock control happening at all.

**Comment by pelwell** [comment]
Yes - it's just a way to pass a clock frequency into the driver.

**Comment by popcornmix** [comment]
Okay just checking it didn't magically drive the upstream clock driver.

Ideally the 500MHz shouldn't be hard coded, but queried from firmware (so it respects changes in firmware clocking or config.txt ``hevc_clock`` settings).
I believe the firmware clock driver does squirrel away the max/min of each clock queried from firmware.
But that can be discussed later. Would be good to get this merged soon.

**Comment by 6by9** [comment]
https://github.com/raspberrypi/linux/pull/3505/commits/5d7d08109a2c6972ae1e68cf40915f66548ef1a3 needs explanation as to where it's come from, otherwise I think this is good to go.

Clock stuff needs to be revisited, but there's no point fussing over it until we have a working fw clock driver merged. At present it doesn't query the frequency range on general clocks, and I haven't found a suitable API for extracting that from a clock anyway.

**Comment by jc-kynesim** [comment]
> @jc-kynesim So you appear to have merged
> 
> * https://patchwork.linuxtv.org/patch/60728/
> * https://patchwork.linuxtv.org/patch/60725/
> * Adding define V4L2_HEVC_SLICE_PARAMS_FLAG_DEPENDENT_SLICE_SEGMENT
> * And added num_entry_point_offsets and entry_point_offset_minus1[256] from somewhere. Do you have any idea as where?
> 
> With these being uapi headers that clients will be using, heritage is extremely important.

Apart from V4L2_HEVC_SLICE_PARAMS_FLAG_DEPENDENT_SLICE_SEGMENT I got all of that by simply copying jernejs (at least I think its his) hevc-ctrl.h from the current Kodi tree.

**Comment by 6by9** [comment]
> I got all of that by simply copying jernejs (at least I think its his) hevc-ctrl.h from the current Kodi tree.

Being a pedant, Kodi is a userspace app and this is a kernel header. Checking https://github.com/xbmc/xbmc there is no file hevc-ctrls.h in the tree.

Please can you provide a link to the tree that you used so I can cherry-pick the commit.

**Comment by jc-kynesim** [comment]
> > I got all of that by simply copying jernejs (at least I think its his) hevc-ctrl.h from the current Kodi tree.
> 
> Being a pedant, Kodi is a userspace app and this is a kernel header. Checking https://github.com/xbmc/xbmc there is no file hevc-ctrls.h in the tree.
> 
> Please can you provide a link to the tree that you used so I can cherry-pick the commit.

Mea culpa - I misremembered - it came from the LibreELEC distribution - specifically projects/Allwinner/patches/linux/0005-cedrus-improvements.patch gives num_entry_point_offsets and entry_point_offset_minus1 - which came from  https://github.com/lrusak/LibreELEC.tv.git branch ffmpeg-update - commit 203d54d9e8f376a57b643fe3de19d4ff70a35c3c

Sorry about that

**Comment by jc-kynesim** [comment]
>> I got all of that by simply copying jernejs (at least I think its his) hevc-ctrl.h from the current Kodi tree.
>
>Being a pedant, Kodi is a userspace app and this is a kernel header. Checking https://github.com/xbmc/xbmc there is no file hevc-ctrls.h in the tree.
>
>Please can you provide a link to the tree that you used so I can cherry-pick the commit.

OK - Sorry about the confusion here - in particular depending on how
github works you may have seen that I created a comment which I then
deleted ('cos I thought I was talking rubbish) which I now find was
correct.  I'll try harder this time.

When I said Kodi I actually meant LibreELEC.tv

The patch for entry_point_offset_minus1 cames from

repo: https://github.com/lrusak/LibreELEC.tv.git
branch: ffmpeg-update
commit: 203d54d9e8f376a57b643fe3de19d4ff70a35c3c
file: /projects/Allwinner/patches/linux/0005-cedrus-improvements.patch

but I will have applied all the paches in that dir - many will not
matter.

Regards

JC

**Comment by 6by9** [comment]
So it looks like you only use num_entry_point_offsets and not the entry_point_offset_minus1 array in wpp_decode_slice. How does that work?
I've added logging for num_entry_point_offsets and it doesn't seem to get triggered. Is this some specific optional extension that only some clips will use?

**Comment by jc-kynesim** [comment]
>So it looks like you only use num_entry_point_offsets and not the entry_point_offset_minus1 array in wpp_decode_slice. How does that work?
>I've added logging for num_entry_point_offsets and it doesn't seem to get triggered. Is this some specific optional extension that only some clips will use?

Yes - WPP allows you to process the "MB" (not called that anymore in
H.265) lines of a slice in parallel - the info (and appropriate
encoding) will only be there for some clips (it is quite common but by
no means universal).  You only really need the offsets if you are
actually going to do stuff in parallel and we don't.

Does that help?  I didn't add the offset array but it might clearly be
needed - I guess the Cedrus codec may need them.

**Comment by jc-kynesim** [comment]
And as a FYI to my previous comment - the Pi3 H265 decode does use the offsets to parallelise decode and it does help.

**Comment by 6by9** [comment]
Thanks. I've pinged jernej on Slack asking if there is a plan to upstream it.
If not then I'll just amend commit text of my current patch and we'll run with it for now. I can foresee pushback again because num_entry_point_offsets has been dropped into the middle of the structure, which means a breakage for existing clients. It's not in the released uapi yet, so it doesn't really make any difference.

**Comment by 6by9** [comment]
WIP tag removed.

I've reworded the commit text for the WPP properties patch, and reordered it so that all steps should compile (I haven't tested that).

If someone else could do a test and confirm that this is still functional, then I think it's good to go.
Please note that the V4L2 pixel formats have changed in both four cc and define. FFmpeg will need updating to match.

**Comment by popcornmix** [comment]
@6by9 I updated ffmpeg: `V4L2_PIX_FMT_SAND8`->`V4L2_PIX_FMT_NV12_COL128` and `V4L2_PIX_FMT_SAND30`->`V4L2_PIX_FMT_NV12_10_COL128`
and I could play both 8-bit hevc and 10-bit through fkms with hw decode speed.

Also worked with kodi (with fkms)

So I think this is good to go.

**Comment by 6by9** [comment]
@pelwell Any other requests from you?
Otherwise can we merge before popcornmix next rebases to a new upstream release?

**Comment by pelwell** [comment]
Just a few minor overlay niggles:
1. The overlay is in the wrong position in the Makefile (off-by-one).
2. The compatible string should be "brcm,bcm2711" because it is strictly 2711-only (otherwise the test script will try to apply it to other platforms and fail).
3. It needs an entry in the README file.

**Comment by 6by9** [comment]
I've made those 3 changes, so hopefully your overlay test scripts will now be happy.

**Comment by pelwell** [comment]
I'm seeing a warning:
```
drivers/staging/media/rpivid/rpivid_h265.c: In function 'write_bitstream':
drivers/staging/media/rpivid/rpivid_h265.c:534:2: warning: #warning "EMU set - probably shouldn't be wanted" [-Wcpp]
 #warning "EMU set - probably shouldn't be wanted"
```

**Comment by 6by9** [comment]
@jc-kynesim I meant to raise that warning with you before.
```
drivers/staging/media/rpivid/rpivid_h265.c: In function 'write_bitstream':
drivers/staging/media/rpivid/rpivid_h265.c:534:2: warning: #warning "EMU set - probably shouldn't be wanted" [-Wcpp]
 #warning "EMU set - probably shouldn't be wanted"
```

**Comment by jc-kynesim** [comment]
> I'm seeing a warning:
> 
> ```
> drivers/staging/media/rpivid/rpivid_h265.c: In function 'write_bitstream':
> drivers/staging/media/rpivid/rpivid_h265.c:534:2: warning: #warning "EMU set - probably shouldn't be wanted" [-Wcpp]
>  #warning "EMU set - probably shouldn't be wanted"
> ```

Yeah - note from me to me - the current ffmpeg code passes bitstreams with emulation prevention left in. My reading of the "standard" suggests that it shouldn't. The warning could be removed - I left it there as a hard note that I thought I was doing something wrong.

I could also try and "fix" FFmpeg and remove the warning and h/w setup.  There are some tedious implications to doing so though.

**Comment by pelwell** [comment]
If it's expected or unavoidable then I'd prefer it to be removed.

**Comment by jc-kynesim** [comment]
My actual guess is that the current behaviour will become the standard for this stream type and another stream type may be invented if anyone ends up caring.  So it is probably safe / sensible to remove the warning - replace it with a comment.

**Comment by 6by9** [comment]
> So it is probably safe / sensible to remove the warning - replace it with a comment.

OK, will do.

**Comment by 6by9** [comment]
Done.

**Comment by popcornmix** [comment]
@pelwell okay to merge?

**Comment by hvenev** [comment]
> @6by9 I updated ffmpeg: `V4L2_PIX_FMT_SAND8`->`V4L2_PIX_FMT_NV12_COL128` and `V4L2_PIX_FMT_SAND30`->`V4L2_PIX_FMT_NV12_10_COL128`
> and I could play both 8-bit hevc and 10-bit through fkms with hw decode speed.
> 
> Also worked with kodi (with fkms)

Where can I find this version of ffmpeg?

**Comment by 6by9** [comment]
@hvenev https://github.com/popcornmix/FFmpeg/tree/kodi_v4l2_1

**Comment by hvenev** [comment]
> @hvenev https://github.com/popcornmix/FFmpeg/tree/kodi_v4l2_1

That seems to mention `V4L2_PIX_FMT_SAND8` and `V4L2_PIX_FMT_SAND30` but not `V4L2_PIX_FMT_NV12_COL128` and `V4L2_PIX_FMT_NV12_10_COL128`.

Is this patch more recent: https://github.com/popcornmix/xbmc/blob/gbm/tools/depends/target/ffmpeg/0001-rpi-Add-hevc-acceleration.patch ?

**Comment by 6by9** [comment]
I took that branch and replaced ```V4L2_PIX_FMT_SAND8```->```V4L2_PIX_FMT_NV12_COL128``` and ```V4L2_PIX_FMT_SAND30```->```V4L2_PIX_FMT_NV12_10_COL128```.

I don't know what has been released/latest version - I leave that to @popcornmix and @jc-kynesim.

**Comment by jc-kynesim** [comment]
The Kodi branch is the only one which currently has the V4L2 work in it.  Annoyingly it is a different base version to the standard Debian one.  I'm guessing Kodi must still be using an older header file to have the older #defines.  I guess that the Debian port should be updated to have the V4L2 work in it too now we have the possibility of using it.

**Comment by popcornmix** [comment]
https://github.com/popcornmix/xbmc/blob/gbm/tools/depends/target/ffmpeg/0001-rpi-Add-hevc-acceleration.patch is the source being used by LibreELEC kodi builds and works with kernels including this PR.
The tree this came from is here: https://github.com/popcornmix/FFmpeg/tree/kodi_v4l2_4_2_2

**Comment by 6by9** [comment]
> The tree this came from is here: https://github.com/popcornmix/FFmpeg/tree/kodi_v4l2_4_2_2

Can we get that merged into https://github.com/Rpi-distro/FFmpeg, or are we waiting on something else?

**Comment by popcornmix** [comment]
It's complicated by kodi using ffmpeg 4.2.2 and debian buster using 4.1.4.

@jc-kynesim is https://github.com/jc-kynesim/rpi-ffmpeg/tree/test/4.1.4/rpi_main the right tree for debian? Can we get a version that supports 5.4 v4l2 and still works with existing VLC?

**Comment by jc-kynesim** [comment]
Yes that should be the right tree.  I'll work on the port once I'm done with my current VAAPI work which may impact our FFmpeg port as well.  We will need to get rid of the horrid config kludge that enables/disables drm_prime passthrough too and replace it either with a command line switch if I can work out how to get one passed to where I need it or preferably some sort of automagic.

**Comment by hvenev** [comment]
Please note that ffmpeg with that patch fails to build:
 - without `--enable-rpi` because various SAND methods are undefined (undefined references while libavutil is linked; `CONFIG_SAND` is disabled so `rpi_sand_fns.c` is not built)
 - with `--enable-rpi` because `MMAL_ES_FORMAT_FLAG_COL_FMTS_WIDTH_IS_COL_STRIDE` is not defined (and that macro is not mentioned in https://github.com/raspberrypi/userland)

**Comment by jc-kynesim** [comment]
MMAL_ES_FORMAT_FLAG_COL_FMTS_WIDTH_IS_COL_STRIDE is defined in /opt/vc/include/interface/mmal/mmal_format.h on my pi4 and I don't think that is 'cos I've edited it.

**Comment by 6by9** [comment]
> MMAL_ES_FORMAT_FLAG_COL_FMTS_WIDTH_IS_COL_STRIDE is defined in /opt/vc/include/interface/mmal/mmal_format.h on my pi4 and I don't think that is 'cos I've edited it.

I suspect we've got another patch that hasn't been merged from firmware to userland :-(

**Comment by hvenev** [comment]
It appears that it is only present in the repo with the 32-bit blobs of unknown origin:

 - https://github.com/raspberrypi/firmware/search?q=MMAL_ES_FORMAT_FLAG_COL_FMTS_WIDTH_IS_COL_STRIDE&type=
 - https://github.com/raspberrypi/userland/search?q=MMAL_ES_FORMAT_FLAG_COL_FMTS_WIDTH_IS_COL_STRIDE&type=

**Comment by hvenev** [comment]
> > MMAL_ES_FORMAT_FLAG_COL_FMTS_WIDTH_IS_COL_STRIDE is defined in /opt/vc/include/interface/mmal/mmal_format.h on my pi4 and I don't think that is 'cos I've edited it.
> 
> I suspect we've got another patch that hasn't been merged from firmware to userland :-(

Can't you just rsync them or something?

**Comment by 6by9** [comment]
> Can't you just rsync them or something?

Two git trees, userland being a subset of firmware. Applying with history isn't totally trivial, and someone needs to pick up on the fact that it's needed.
https://github.com/raspberrypi/userland/pull/624

**Comment by hvenev** [comment]
Just out of curiosity, are there any other patches for which the source code isn't available in [raspberrypi/userland](https://github.com/raspberrypi/userland) but are a part of the 32-bit blobs of questionable origin at [raspberrypi/firmware](https://github.com/raspberrypi/firmware)?

**Comment by hvenev** [comment]
Also note that even with the userland pull request, ffmpeg still fails to build with `--enable-rpi`:
```
libavcodec/rpi_mailbox.h:47:13: error: size of array 'vc_image_t_size_check' is negative
   47 | typedef int vc_image_t_size_check[(sizeof(VC_IMAGE_T) == 64) * 2 - 1];
      |             ^~~~~~~~~~~~~~~~~~~~~
```

**Comment by hvenev** [comment]
This time I tried `--disable-rpi` and I fixed the SAND mess.

Another build of ffmpeg later:
```
/usr/bin/ld: libavutil/libavutil.so: undefined reference to `rpi_sand128b_stripe_to_8_10'
```

It looks like the function is used when `HAVE_NEON`, but it's really only implemented when `defined(__arm__) && HAVE_NEON`.

I will send a pull request when I get ffmpeg to build (and hopefully work).

**Comment by jc-kynesim** [comment]
I don't know if it will help you but if you have https://github.com/popcornmix/FFmpeg/tree/kodi_v4l2_4_2_2 checked out then you will find a shell script pi-util/conf_native.sh which contains the set of config options I use for compiling that code on a pi4 (and it does compile - I've just checked).  The cross-compile scripts don't currently work on this codebase.

I normally spot the userland/firmware issues you are having as they show up in the cross compiles but whilst all the kernel & v4l2 stuff that are needed for V4L2 were still up in the air I found it too hard to keep sysroots in any sort of sync and resorted to compiling on the pi and so failed to spot  the fact that stuff was out of sync (I have no power to fix this up - only to winge).

Anyway - good luck

**Comment by jc-kynesim** [comment]
> It looks like the function is used when `HAVE_NEON`, but it's really only implemented when `defined(__arm__) && HAVE_NEON`.

OK - I'll bite - how do you have neon and not arm?

**Comment by popcornmix** [comment]
https://github.com/popcornmix/FFmpeg/tree/kodi_v4l2_4_2_2 is more recent and should match the kodi patch.

**Comment by nkichukov** [comment]
I still fail to compile kodi-gbm with ffmpeg kodi_v4l2_4_2_2 both from popcornmix repo on rpi4 with 32-bit userland, the failure is:

```
src/libavcodec/v4l2_request.c:165:5: error: ‘V4L2_PIX_FMT_NV12_COL128’ undeclared here (not in a function); did you mean ‘V4L2_PIX_FMT_NV21M’?
  165 |     V4L2_PIX_FMT_NV12_COL128,
      |     ^~~~~~~~~~~~~~~~~~~~~~~~
      |     V4L2_PIX_FMT_NV21M
src/libavcodec/v4l2_request.c:166:5: error: ‘V4L2_PIX_FMT_NV12_10_COL128’ undeclared here (not in a function); did you mean ‘V4L2_PIX_FMT_NV12MT_16X16’?
  166 |     V4L2_PIX_FMT_NV12_10_COL128,
      |     ^~~~~~~~~~~~~~~~~~~~~~~~~~~
      |     V4L2_PIX_FMT_NV12MT_16X16
src/libavcodec/v4l2_request.c: In function ‘v4l2_request_set_drm_descriptor’:
src/libavcodec/v4l2_request.c:189:25: error: ‘DRM_FORMAT_P030’ undeclared (first use in this function); did you mean ‘DRM_FORMAT_P010’?
  189 |         layer->format = DRM_FORMAT_P030;
      |                         ^~~~~~~~~~~~~~~
      |                         DRM_FORMAT_P010
```

seems like those are not defined...

Kernel: 5.4.28

**Comment by jc-kynesim** [comment]
Yeah - the situation with the v4l2 and the drm header files is quite frustrating. They don't seem to be exported to userspace in the way one would like.  I've had to grab them from the linux tree.  I'm not sure what is meant to have occurred.

**Comment by nkichukov** [comment]
Thanks @jc-kynesim 

Those seem to be defined in the kernel includes:
include/uapi/drm/drm_fourcc.h and include/uapi/linux/videodev2.h, so wonder why do not get included?

I've just defined them for the time being as per your suggestion in the libavcodec/v4l2_request.c:

```
#define V4L2_PIX_FMT_NV12_COL128 v4l2_fourcc('N', 'C', '1', '2') /* 12  Y/CbCr 4:2:0 128 pixel wide column */
#define V4L2_PIX_FMT_NV12_10_COL128 v4l2_fourcc('N', 'C', '3', '0')
                                                                /* Y/CbCr 4:2:0 10bpc, 3x10 packed as 4 bytes in
                                                                 * a 128 bytes / 96 pixel wide column */

#define DRM_FORMAT_P030         fourcc_code('P', '0', '3', '0') /* 2x2 subsampled Cr:Cb plane 10 bits per channel packed */
#define DRM_FORMAT_P010         fourcc_code('P', '0', '1', '0') /* 2x2 subsampled Cr:Cb plane 10 bits per channel */
```

and it managed to move on.

hvenev:
Please keep in mind the fact that you might also need to rebuild software that links against ffmpeg. The version provided at https://github.com/popcornmix/FFmpeg/tree/kodi_v4l2_4_2_2 is not ABI-compatible with upstream ffmpeg 4.2.2. I'm trying to get things to work in  https://github.com/popcornmix/ffmpeg/pull/1.

PR #3506, opened 2020-03-13
Add 10bit YUV support to FKMS

6by9:
Also includes a cleanup patch from Maxime.

The definition in drivers/gpu/drm/drm_fourcc.c ought to be sanity checked at some point, but I don't believe any of our current code actually uses it.


popcornmix:
Looks fine to me.

JamesH65:
LGTM.

6by9:
Still compiling on this clean tree, but not expecting any surprises....

6by9:
Built fine.
I'm happy if someone wants to merge.

Issue #3460, opened 2020-02-10
Dispatching compute shader costs us at least 30ms


notogawa:
**Describe the bug**

When we dispatch a shader program using `ioctl(SUBMIT_CSD)` on Raspberry Pi 4, it costs us 30ms. Calling `ioctl(SUBMIT_CSD)` 10 times before `ioctl(WAIT_BO)` means 300ms to complete all jobs. Even if a shader program is short enough and do-nothing, this cost is not discounted at all.

We checked trace and confirmed that `v3d_clean_caches`, called after each dispatched job, has this 30ms. https://github.com/raspberrypi/linux/commit/c1a40a21b2637d2ce7f71a71cc9c7fe101a14fef says this `v3d_clean_caches` is required to avoid MMU errors. Is this a performance problem of `v3d_clean_caches` ?

**To reproduce**

[This is an example program](https://gist.github.com/notogawa/36d0cc9168ae3236902729f26064281d#file-main-c) to reproduce. In this example, a shader is 8 `nop` instructions.

~~~console
git clone https://gist.github.com/notogawa/36d0cc9168ae3236902729f26064281d
cd 36d0cc9168ae3236902729f26064281d
gcc -o main main.c
./main
~~~

**Expected behaviour**

Shader jobs should be completed at a suitable time. I think 30ms is too long.

**Actual behaviour**

at least 30ms / dispatch job

[This is an example program output.](https://gist.github.com/notogawa/36d0cc9168ae3236902729f26064281d#file-main-log)

**System**

* Which model of Raspberry Pi? e.g. Pi3B+, PiZeroW

~~~
Hardware        : BCM2835
Revision        : c03112
Serial          : 100000006499218e
Model           : Raspberry Pi 4 Model B Rev 1.2
~~~

* Which OS and version (`cat /etc/rpi-issue`)?

~~~
Raspberry Pi reference 2019-09-26
Generated using pi-gen, https://github.com/RPi-Distro/pi-gen, 80d486687ea77d31fc3fc13cf3a2f8b464e129be, stage2
~~~

* Which firmware version (`vcgencmd version`)?

~~~
Sep 24 2019 17:34:30
Copyright (c) 2012 Broadcom
version cd3add54955f8fa065b414d8fc07c525e7ddffc8 (clean) (release) (start)
~~~

* Which kernel version (`uname -a`)?

~~~
[...]

PR #3510, opened 2020-03-20
Backport: drm/v3d: Replace wait_for macros to remove use of msleep

Terminus-IMRC:
This pull request is just a backport from [drm-misc-next](https://cgit.freedesktop.org/drm/drm-misc/commit/?id=9daee6141cc9c75b09659b02b1cb9eeb2f5e16cc), which was discussed in https://github.com/raspberrypi/linux/issues/3460.
To measure the precise running time of QPU programs, we want this commit to be also in rpi-4.19.y tree.
If we are moving to rpi-5.4.y, then I will post another pull request if needed.

### DIFF
--- a/drivers/gpu/drm/v3d/v3d_drv.h
+++ b/drivers/gpu/drm/v3d/v3d_drv.h
@@ -267,26 +267,41 @@ struct v3d_csd_job {
 };
 
 /**
- * _wait_for - magic (register) wait macro
+ * __wait_for - magic wait macro
  *
- * Does the right thing for modeset paths when run under kdgb or similar atomic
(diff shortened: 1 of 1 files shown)


pelwell:
I'm surprised that the body of this change is a macro and not a function, but I see it's been accepted already so who am I to argue.

Don't worry about the ports to later kernels unless something makes it non-trivial - it will get picked up with the other downstream changes.

JamesH65:
The original was a macro, this was just making it use nano sleep instead of msleep. Basically just a copy of a better version from an upstream x86 driver IIRC.

Terminus-IMRC:
Thanks for the developing and merging!

PR #3511, opened 2020-03-20
[5.4] pick bcmgenet upstream fix for non-functional network when netbooting

HiassofT:
As reported on the forum https://www.raspberrypi.org/forums/viewtopic.php?p=1629200#p1629200 upstream commit 402482a6a78e5c61d8a2ec6311fc5b4aca392cd6 is needed to have working network when netbooting RPi4.

The upstream commit contains a Fixes tag so should eventually find it's way into stable kernels


timg236:
LGTM There's probably scope for improving the bootloader setup and considering how much start.elf should reset before starting Linux

PR #3515, opened 2020-03-25
Pi4 KMS driver

6by9:
Merging Maxime's patches to support full KMS on Pi4 (whilst hopefully not breaking Pi3!)

I've included the content of popcornmix's patches to change the driver to match ALSA's defaults on audio B frame marker, and cleaned up the audio support patches.

This is a work in progress, and it is known that mode changes are not always reliable, and we get the one pixel wrap around on mode changes due to a pixel getting stuck in a FIFO. CEC is also currently disabled on Pi4.
These will be addressed as and when they are understood!


pelwell:
Is this supposed to work with `dtoverlay=vc4-kms-v3d-pi4`? I get errors in the kernel log:
```
[    1.061069] raspberrypi-clk raspberrypi-clk: Missing firmware node
[    1.341070] brcmstb-i2c fef04500.i2c: IRQ index 0 not found
[    1.346271] brcmstb-i2c fef09500.i2c: IRQ index 0 not found
[    7.274793] debugfs: Directory 'fef00700.hdmi' with parent 'vc4-hdmi' already present!
[    7.316239] debugfs: Directory 'vc4-hdmi' with parent 'asoc' already present!
```
And judging by the speed of neverball it's using software rendering.

arch/arm/boot/dts/overlays/vc4-kms-v3d-pi4-overlay.dts
```diff
@@ -67,9 +67,11 @@
 	status = "okay";
```
pelwell: The Subject of this commit message is misleading - I was misled - it should be vc4-kms-pi4-v3d.

6by9: You want the overlay renamed, or the commit message adjusted?
Sorry, I'm not seeing where you've been misled. The Pi4 version of vc4-kms-v3d is surely vc4-kms-v3d-pi4, and it is.

The commit text does want to be reworded as it enables more than it declares as Maxime has made further bits work (eg HDMI) since I created the overlay.

pelwell: The Subject is `overlays: Fix up vc4-kms-pi4 overlay to not enable the VEC.`, which is clearly wrong - the overlay is called `vc4-kms-v3d-pi4`.

6by9: Your comment is showing up on Github against "[DOWNSTREAM] dtoverlays: Add Pi4 version of vc4-kms-v3d ", hence the confusion.
I'll squash the two together anyway.

arch/arm/boot/dts/overlays/vc4-kms-v3d-pi4-overlay.dts
```diff
@@ -0,0 +1,76 @@
+/*
+ * vc4-kms-v3d-pi4-overlay.dts
+ */
+
+/dts-v1/;
+/plugin/;
+
+#include <dt-bindings/clock/bcm2835.h>
+
+&{/chosen} {
```
pelwell: The overlay checking tools don't yet understand the new overlay syntax, and this overlay is missing a compatible string. You'll need to use the old syntax for now if you want it accepted.

arch/arm/boot/dts/overlays/vc4-kms-v3d-pi4-overlay.dts
popcornmix: @pelwell the world would be simpler if `dtoverlay=vc4-kms-v3d` worked on pi3 and pi4 (as `dtoverlay=vc4-fkms-v3d` does). Can you imagine any DT magic that would make a single overlay support both? (firmware side adjustment clearly possible but there may be a nicer option).

6by9: We had this conversation last week as to whether overlays could have compatible strings that match. The simple answer is not at present.
A firmware side overlay name amendment was what was proposed at least in the medium term.

pelwell: @popcornmix I can't think of a way to make it work within an overlay that didn't require a lot of work figuring out which symbols are actually needed before trying to resolve them.

What would be simpler and safer is an overlay name indirection system - using a text file (e..g. `vc4-kms-v3d.ovl`) containing a mapping from platforms to `.dtbo` files; an absent `.ovl` file would cause it fall back to loading `vc4-kms-v3d.dtbo`.

arch/arm/boot/dts/bcm2711.dtsi
```diff
@@ -348,6 +348,8 @@
 			clock-names = "hdmi";
 			resets = <&dvp 0>;
 			ddc = <&ddc0>;
+			dmas = <&dma 10>;
```
6by9, line 352: This patch "drm/vc4: HDMI audio support on Pi4. " and the revert will be removed. It was an intermediate step in cleaning up all the audio changes.

6by9:
Rebased. Overlay fixed up and squashed. Audio patch patch&revert removed.
One test patch from Maxime added which kicks the core clock during mode changes. That seems to make mode changes more reliable. I've yet to give it a real bashing though.

popcornmix:
I couldn't boot on Pi4 with fkms overlay. Seemed to hang with black screen on modeset towards end of linux boot.
Using kms overlay did boot to prompt.

popcornmix:
@6by9 needs a rebase

6by9:
Rebased, and FKMS is working for me (both console and X - there was an issue for each)

popcornmix:
Had a LibreELEC build with this PR (in kms mode) running for 3 days without issues, so looks promising.

pelwell:
FKMS looks good now, but I get software 3D in KMS mode. There are a number of warnings/errors in the kernel log that could be relevant:
```
pi@raspberrypi:~$ dmesg -l warn,err
[    1.081444] raspberrypi-clk raspberrypi-clk: Missing firmware node
[    1.084150] raspberrypi-clk: probe of raspberrypi-clk failed with error -2
[    1.361293] brcmstb-i2c fef04500.i2c: IRQ index 0 not found
[    1.366454] brcmstb-i2c fef09500.i2c: IRQ index 0 not found
[    7.268400] vc4_hdmi fef00700.hdmi: Unknown register ID 46
[    7.268819] debugfs: Directory 'fef00700.hdmi' with parent 'vc4-hdmi' already present!
[    7.301977] vc4_hdmi fef05700.hdmi: Unknown register ID 46
[    7.302235] debugfs: Directory 'vc4-hdmi' with parent 'asoc' already present!
```

popcornmix:
@pelwell is fkms is okay for 3d?
You are using the vc4-kms-v3d-pi4 overlay? (I think the non-pi4 version does boot with video support but no 3d).
LE/Kodi was using harwdare 3d for me with this PR.

6by9:
```brcmstb-i2c fef04500.i2c: IRQ index 0 not found``` isn't an issue - it falls back to polling. It's actually ```platform_get_irq``` objecting, which probably ought to be a ```platform_get_irq_optional```

```vc4_hdmi fef00700.hdmi: Unknown register ID 46``` is a dev_warn as HDMI_M_CTL is only applicable for vc4 (not vc5). I don't see it, so do you have a non-default log level enabled?

```debugfs: Directory 'fef00700.hdmi' with parent 'vc4-hdmi' already present!``` is debugfs getting upset so has no operational effect. It's trying to register both HDMI ports under 'vc4-hdmi' which leads to the collision. Initial guess is that is for the audio side under /sys/kernel/debug/asoc/vc4-hdmi/. I'm just trying out changing the asoc card name between the two instances to see if that clears it.

I'm certainly getting glxinfo reporting "Broadcom V3D 4.2" for rendering.

pelwell:
I'm getting `llvmpipe`.

pelwell:
Ah - after an apt upgrade the driver is working. Presumably the mesa libraries were out of date.

I'm happy to merge this after a rebase.

6by9:
I have patches to deal with the i2c-brcmstb IRQ whinge, and to name the HDMI audio instances differently. I'll push them in a mo along with a rebase.

It looks like the debugfs whinge is inherent in calling ```devm_snd_soc_register_component``` to register a sound card. Either that or something along those lines with regard exactly the calls we are making to set up the cards.

Please remember that these are not the final patches that will be merged upstream. This is v1 with a bundle of extra stuff. How we handle future revisions is a matter for discussion.

6by9:
Rebased and 3 extra minor patches.
- Give the audio interfaces different card name
- Avoid the IRQ error message from i2c-brcmstb during probe
- Drop the I2C frequency for the HDMI ports to be within spec (<100kHz)

6by9:
As others are happy I've dropped the WIP tag.

pelwell:
GItHub claims it still needs a rebase, or I can merge it via the back door.

6by9:
> GItHub claims it still needs a rebase, or I can merge it via the back door.

It does? I don't get to see that bit, only "This branch has no conflicts with the base branch"
git status reports "Your branch is ahead of 'rpi/rpi-5.4.y' by 114 commits."

Call me confused.

pelwell:
It's cleared now - as you say, weird.

pelwell:
Not cleared after all - the mobile site appeared happy but failed the rebase. I've merged it manually.

PR #3517, opened 2020-03-26
rpi-5.4.y: add Sensirion SPS30 to i2c-sensor overlay

ynezz:


ynezz:
Should I create patches/PRs for 5.5 and 5.6 as well, or this could be cherry-picked?

pelwell:
This looks OK, but it needs an entry in the README (watch the formatting - no TABs, same indents).

It will be cherry-picked to other branches

ynezz:
@pelwell Done.

pelwell:
The address appears to be modifiable, but you don't list the legal values or default.

ynezz:
> The address appears to be modifiable, but you don't list the legal values or default.

@pelwell AFAIK it's fixed address, added that info.

arch/arm/boot/dts/overlays/i2c-sensor-overlay.dts
```diff
@@ -231,11 +231,25 @@
...
+				status = "okay";
+			};
+		};
+	};
 
 	__overrides__ {
 		addr =  <&bme280>,"reg:0", <&bmp280>,"reg:0", <&tmp102>,"reg:0",
 			<&lm75>,"reg:0", <&hdc100x>,"reg:0", <&sht3x>,"reg:0",
-			<&ds1621>,"reg:0", <&bme680>,"reg:0";
+			<&ds1621>,"reg:0", <&bme680>,"reg:0", <&sps30>,"reg:0";
```
pelwell: If addresses are fixed, you don't need this change.

ynezz: Ok, fixed.

PR #3527, opened 2020-04-02
overlays: Add an overlay map

pelwell:
With a suitable firmware, an overlay map file can:
* allow a single overlay name to be mapped to platform-specific implementations,
* reject overlay names on some platforms,
* rename overlays, and
* mark overlays as deprecated,

all with helpful error messages.

These commits add a base map file (suitable for 4.19 if necessary), tidy up some old overlays, and add a remapping for vc4-kms-v3d on Pi 4.


arch/arm/boot/dts/overlays/overlay_map.dts
```diff
@@ -108,4 +108,13 @@
...
 		deprecated = "no longer necessary";
 	};
+
+	vc4-kms-v3d {
+		bcm2835;
+		bcm2711 = "vc4-kms-v3d-pi4";
+	};
+
+	vc4-kms-v3d-pi4 {
+		bcm2711;
```
vianpl, line 127: I guess this could be implied by the property above. Although I see how it can make the implementation complicated.

pelwell: Yes it is implied, but I don't really fancy having to search the whole map for string properties that match the current name every time an overlay is applied. However, I can add a test that confirms that mappings like this are consistent.

vianpl: Fair enough :)

popcornmix:
Looks good to me.

pelwell:
Thanks. It will need the firmware side pushing to be generally useful, but it's harmless until then,

PR #3528, opened 2020-04-02
vc4_hdmi_phy: Fix offset calculation

popcornmix:
The original firmware code worked with float and did
   offset = ((vco_freq / fref * 2) * (1 << 22));
   offset >>= 2;

In this code it's all integer so doing the integer divide before the shift loses lots of precision

This fixes the issue of 1080p59.94 mode having 59.64 fps


6by9:
LGTM, although I haven't checked for overflows.

@mripard For version 2.

popcornmix:
I thought about overflows and it looks safe.
vco is a few gigs, so maybe 33 bits. 34 with the *2. 56 with the << 22.
Still around 8 bits of headroom.

I chose to leave the diffs minimal rather than fixing it up.
I'm sure this commit will be squashed and tidied when upstreaming.

Issue #3520, opened 2020-03-27
RFC: Multi-platform overlays - a proposal


pelwell:
The introduction of the Pi 4, built around the BCM2711 SoC, has caused a split in our code base; for the first time, some overlays won't work on all (can't even be applied) - on all Pis. Some of these failures come under the general heading of "don't do that", e.g. trying to use `uart5` on a 3B+, but for others it is clear that something different needs to happen on each platform - a change of implementation, fewer parameters, etc.

There is therefore a need for a method of tailoring an overlay to multiple platforms with differing hardware. Supporting them all in a single `.dtbo` file would require heavy use of hidden ("dormant") fragments and a switch to an on-demand symbol resolution mechanism so that a missing symbol that isn't needed doesn't cause a failure. A simpler solution is to add a facility to map an overlay name to one of several implementation files depending on the current platform.

For the purposes of this discussion, `<overlay>` is the name of an overlay, where `<overlay>.dtbo` is the name of the default implementation file. The first plan was to use an overlay-specific proxy file as a matchmaker for several implementing `.dtbo` files - `<overlay>.ovl`, say. It would contain a list of mappings from platform names to overlay file names - a textual format would probably suffice. The absence of a platform from a `.ovl` means the overlay isn't supported on that platform. The absence of a `<overlay>.ovl` file would indicate that all platforms should use the common default implementation. The advantage of this scheme is that each overlay remains self contained, but it does require an extra file test/load for each overlay, something which is particularly slow when network booting.

Instead, I favour the addition of a single map file that gets loaded once at start of day. It can be written in `.dts` format with a node per overlay:

```
/ {
    vc4-kms-v3d {
        bcm2711 = "vc4-kms-v3d-pi4";
        bcm2835;
    };
};
```

Notes:
1. The platform names are compatible strings (minus the vendor prefix for brevity).
2. A platform name with no value (an empty property) indicates that the default overlay file (`vc4-kms-v3d.dtbo` in this case) should be used.
3. As with the previous plan, this only exceptions need to be listed - the absence of a node for an overlay means that the default file should be used.

Once the map has been compiled, use is straightforward - locate the node for an overlay and, if found, retrieve the property with the current platform name.

Although simple, this mechanism allows some nice extensions in the event that the platform property doesn't exist:
```
/ {
    vc4-kms-v3d {
        bcm2711 = "vc4-kms-v3d-pi4";
        bcm2835;
    };

    pi3-disable-bt {
        renamed = "disable-bt";
[...]

PR #3529, opened 2020-04-02
CMA overlay centalization

vianpl:
This is still a WIP, as I haven't properly tested it. Yet, the overall idea is pretty clear at this point and I'd appreciate your input.

### DIFF
--- a/arch/arm/boot/dts/bcm2711-rpi-4-b.dts
+++ b/arch/arm/boot/dts/bcm2711-rpi-4-b.dts
@@ -150,7 +150,7 @@
 
 / {
 	chosen {
-		bootargs = "coherent_pool=1M 8250.nr_uarts=1 cma=64M";
+		bootargs = "coherent_pool=1M 8250.nr_uarts=1";
 	};
 
 	aliases {
@@ -170,10 +170,6 @@
 	};
 
 	/delete-node/ wifi-pwrseq;
-
-	reserved-memory {
-		/delete-node/ linux,cma;
-	};
 };
 
 &mmcnr {
--- a/arch/arm/boot/dts/overlays/Makefile
+++ b/arch/arm/boot/dts/overlays/Makefile
@@ -26,6 +26,7 @@ dtbo-$(CONFIG_ARCH_BCM2835) += \
 	audremap.dtbo \
 	balena-fin.dtbo \
 	bmp085_i2c-sensor.dtbo \
+	cma.dtbo \
 	dht11.dtbo \
 	dionaudio-loco.dtbo \
 	dionaudio-loco-v2.dtbo \
--- a/arch/arm/boot/dts/overlays/README
+++ b/arch/arm/boot/dts/overlays/README
@@ -548,6 +548,19 @@ Info:   This overlay is now deprecated - see i2c-sensor
 Load:   <Deprecated>
 
 
+Name:   cma
+Info:   Set custom CMA sizes, only use if you know what you are doing, might
+        clash with other overlays like vc4-fkms-v3d and vc4-kms-v3d.
+Load:   dtoverlay=cma,<param>=<val>
+Params: cma-256                 CMA is 256MB (needs 1GB)
+        cma-192                 CMA is 192MB (needs 1GB)
+        cma-128                 CMA is 128MB
+        cma-96                  CMA is 96MB
+        cma-64                  CMA is 64MB
+        cma-size                CMA size in bytes, 4MB aligned
+        cma-default             Use upstream's default value
+
+
 Name:   dht11
 Info:   Overlay for the DHT11/DHT21/DHT22 humidity/temperature sensors
         Also sometimes found with the part number(s) AM230x.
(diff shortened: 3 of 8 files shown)


pelwell:
I like it - it looks like the kind of thing I would have ended up with (sorting out CMA in DT was next-but-one on the list of things to do).

There are a few things I would tweak:
1. cma has to come after balena-fin in the README.
2. If cma is going to exist as a standalone overlay then it needs an entry in the Makefile; if not then it should be a dtsi.
3. When I regenerate the upstream overlay with ovmerge it includes the cma fragment, even though it is hidden. Since the checker script regenerates it every time to test for changes that fragment is going to get put back, so its better to include it now (at least until I change ovmerge to drop hidden segments).

There is a related item that will affect the end result. The Pi 4 KMS driver we merged into rpi-5.4.y this week needs a different version of the vc4-kms-v3d overlay - vc4-kms-v3d-pi4. This means that there will have to be two versions of the upstream overlay as well. #3520 describes a way of loading a platform-specific implementation of an overlay, and #3527 is a candidate for what it would look like in the kernel repo (but lacking the Pi 4 version of the upstream overlay).

vianpl:
Hi @pelwell, thanks for reviewing.

I don't fully get point 3:

I changed the command to: `ovmerge -c vc4-kms-v3d-overlay.dts,cma-default dwc2-overlay.dts,dr_mode=otg`

Then, by pulling in vc4-kms-v3d-overlay.dts, I'm getting cma-overlay.dts as a side effect. And by adding the cma-default option the CMA fragment is left dormant. That leaves the `&cma` node unaltered, which I presume it's what we want.

pelwell:
Hi @vianpl,

What I was trying to say in point 3 is that the version of upstream-overlay.dts as contained in the PR does not match the one that I get when I run the same ovmerge command on the other files in this PR. A diff looks like this:
```diff
diff --git a/arch/arm/boot/dts/overlays/upstream-overlay.dts b/arch/arm/boot/dts/overlays/upstream-overlay.dts
index e634387d64fb..4d46790d81c9 100644
--- a/arch/arm/boot/dts/overlays/upstream-overlay.dts
+++ b/arch/arm/boot/dts/overlays/upstream-overlay.dts
@@ -8,84 +8,90 @@
 / {
        compatible = "brcm,bcm2835";
        fragment@0 {
+               target = <&cma>;
+               __dormant__ {
+                       size = <0x10000000>;
+               };
+       };
+       fragment@1 {
                target = <&i2c2>;
                __overlay__ {
                        status = "okay";
                };
        };
-       fragment@1 {
+       fragment@2 {
                target = <&fb>;
                __overlay__ {
                        status = "disabled";
                };
        };
-       fragment@2 {
+       fragment@3 {
                target = <&pixelvalve0>;
                __overlay__ {
                        status = "okay";
                };
        };
-       fragment@3 {
+       fragment@4 {
                target = <&pixelvalve1>;
                __overlay__ {
                        status = "okay";
                };
        };
...
```
The version that I generated include the first fragment, albeit a dormant/hidden one, whereas the one uploaded doesn't. The two overlays have the same end result, but it means that next time I run the command to see if it has changed it will find a difference, even though nothing else as changed.

vianpl:
understood

vianpl:
I've performed my share of testing on both 64bit and 32bit builds and multiple devices (RPi3b+ and RPi4b). I double-checked the kernel command line still behaves as expected and the new CMA overlay property modifiers.

Note that, in order to configure CMA trough the kernel command line on RPi4 you also have to provide correct placement constraints. For example `cma=256M@0x0-0x40000000`. Otherwise it'll be placed in the higher parts of memory, which we can't really use.

I can't do much for `vc4-kms-v3d-pi4` as it's only available on v5.4.y and back-porting this series is not trivial. On the other hand forward-porting it will be fairly easy and can take care of it if needed. As for #3520, I welcome it, and think it'll play alright with this series.

Also, I addressed all @pelwell's comments. So overall I'd say it's ready for proper review.

pelwell:
There's just one hard TAB on line 553 of the README, otherwise I'm happy to merge.

I'll sort out the "upstream" handling on 5.4, then apply the necessary changes to 5.6.

vianpl:
Fixed!

pelwell:
Thanks!

HiassofT:
@pelwell do you plan to backport this change to 5.4?

It's a nice cleanup and would help us in LibreELEC, too - I'm currently testing with cma=512M on cmdline.txt (some 4k HEVC files seem to need more memory to decode properly) and being able to select this via dt instead of cmdline would make things a lot easier for us

pelwell:
I was planning to, but I can add it to the list - or take a PR.

HiassofT:
Thanks!

And no need to hurry, take your time, we don't have a pressing need to have this immediately in 5.4.

pelwell:
If we're going to switch then it would be nice if it was there from launch.

pelwell:
@vianpl - the DT work is easy enough, but how many infrastructure changes need to be back-ported to 5.4 for this to work? Do you think it's sensible?

vianpl:
@pelwell It's totally doable, I've actually done it for our (SUSE) v5.3 LTS kernel. I can have a go at it if you're patient enough.

pelwell:
That would be great - or just give me any helpful tips and I'll do the donkey work.

pelwell:
This is as far as I've got: https://github.com/raspberrypi/linux/pull/3536
It's working on 32-bit, but 64-bit is broken.

PR #3531, opened 2020-04-06
clk-raspberrypi: Allow cpufreq driver to also adjust gpu clocks

popcornmix:
Cherry-pick from rpi-4.19.y.
Without this we don't boost core and other gpu clocks when arm requests turbo mode.


pelwell:
@mripard Can you explain why you reverted this on rpi-5.4.y?

pelwell:
@popcornmix I don't mind merging this now until @mripard responds.

mripard:
That was a mistake, sorry :/

pelwell:
No problem - it's better than having to start a lengthy investigation!

PR #3533, opened 2020-04-08
Rpi 5.4.y add the audioinjector.net isolated sound card

flatmax:


flatmax:
Thanks for merging.
About this patch https://github.com/raspberrypi/linux/pull/3533/commits/626d6ada3333664ffcc92416c528899bbf1d971f
It removes the cs4271 codec from the build. The audio injector isolated sound card uses that codec !
Does the commit need to be dropped ?

pelwell:
Are you sure? Check the .config.

All I did was regenerate the defconfigs, and because your soundcard config selects the codec (as it should) the additional explicit enabling of the codec is unnecessary.

flatmax:
Great, what you said makes sense.
I'll pull those changes and do a test. I'll report back if there is a hiccup.

PR #3483, opened 2020-02-24
Add support for merus-amp soundcard and ma120x0p codec

AMuszkat:
This patch add support for [merus-amp soundcard](https://www.infineon.com/cms/en/product/evaluation-boards/kit_40w_amp_hat_zw/)
and its codec ma120x0p. The code has been tested on a Raspberry pi ZW, pi 3B+ and Raspberry pi 4.
The soundcard consists in a Class-D amplifier HAT with a built-in boost converter.

PR #3534, opened 2020-04-08
sound:soc:codecs:ma120x0p.c update and added to defconfigs

AMuszkat:
This patch add the MA120X0P codec to the defconfigs files as it was not being compiled in [first patch](https://github.com/raspberrypi/linux/pull/3483) when generating the config file.

Also, the patch add 96KHz rate support and convert the mute and enable gpio lines as optional. (This will be required in future sound cards using the same codec).


pelwell:
Merged offline (including 64-bit platforms).

PR #3535, opened 2020-04-08
KMS patches - RGB range

6by9:
3 minor patches from @popcornmix (I've dropped the ones that limit the pixel rate as they shouldn't be needed)
1 from me adding "Broadcast RGB" as a DRM property on HDMI connectors.


popcornmix:
Let me improve the channel mapping one before merging.
It really needs a lookup table from number of channels to channel map.

It fixes the main cause of 8 channel audio, but currently choosing, say, 5.1 has the speaker in wrong locations.

popcornmix:
How does a user make use of the broadcast RGB override?

6by9:
Dropping the audio patches if popcornmix is working on them.

You get a DRM property on the HDMI connector. For me from modetest I get
```
Connectors:
id      encoder status          name            size (mm)       modes   encoders
32      31      connected       HDMI-A-1        520x320         40      31
  modes:
        index name refresh (Hz) hdisp hss hse htot vdisp vss vse vtot)
  #0 1920x1200 59.95 1920 1968 2000 2080 1200 1203 1209 1235 154000 flags: phsync, pvsync; type: preferred, driver
...
  props:
        1 EDID:
                flags: immutable blob
...
        37 Broadcast RGB:
                flags: enum
                enums: Automatic=0 Full=1 Limited 16:235=2
                value: 1
```
and the same for HDMI-A-2.
```~/drm/build/tests/modetest/modetest -M vc4 -w 32:"Broadcast RGB":1``` will set it to Full, or ```~/drm/build/tests/modetest/modetest -M vc4 -w 32:"Broadcast RGB":2``` will set it to limited.
If you're implementing a DRM app (which you are), then you can query for the property and change it as you wish.

popcornmix:
What about X? Presumably the same users will have wrong colours in Chromium, VLC etc?

6by9:
> What about X? Presumably the same users will have wrong colours in Chromium, VLC etc?

Once set the property persists.

You can also set it via xrandr - ```xrandr --output HDMI1 --set "Broadcast RGB" "Full"``` or similar.

I seem to recall a patch that allowed drm properties to be set from the kernel command line too, but I don't recall whether those have been merged as yet (I suspect not).

popcornmix:
Command line seems a convenient method for users (as they'll presumably want it for all applications).

6by9:
> Command line seems a convenient method for users (as they'll presumably want it for all applications).

I seem to remember discussion, not sure about patches implementing it. I'll have a hunt for patches.

6by9:
Hmm, nothing on the lists, and all the parsing stuff at present is tied in to setting the mode.

I'll rework this to move broadcast_rgb out into the common code, and then we'll be in a better place to add it as a cmdline thing. It'll be extra baggage to carry around within the core code though, unless we can upstream it.

6by9:
Actually looking at it, trying to make this a common property is a bad idea, so the PR is complete as it is.

I have asked Maxime if he has any good ideas about how to support it on the command line, so we'll see what he comes back with. That can be addressed in a later patch.

6by9:
Alternate implementation pushed by popcornmix.

PR #3536, opened 2020-04-08
Configure CMA using DT

pelwell:
This is an attempt at a back-port of @vianpl's PR (#3529), which targets rpi-5.6.y.

It appears to work on 32-bit builds, but 64-bit looks like DMA address mapping is wrong.

### DIFF
--- a/arch/arm/boot/dts/bcm2711-rpi-4-b.dts
+++ b/arch/arm/boot/dts/bcm2711-rpi-4-b.dts
@@ -195,7 +195,7 @@
 
 / {
 	chosen {
-		bootargs = "coherent_pool=1M 8250.nr_uarts=1 cma=64M";
+		bootargs = "coherent_pool=1M 8250.nr_uarts=1";
 	};
 
 	aliases {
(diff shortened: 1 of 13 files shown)


pelwell:
@HiassofT I think I've got this working now - can you give it a try?

HiassofT:
@pelwell thanks a lot! testing on RPi4 worked fine, tested the cma-size parameter both of cma and c4-kms-v3d-pi4 overlays (32bit kernel only though, haven't checked 64bit)

pelwell:
FYI, I had to revert this patch set for a while because of an issue causing 32-bit boot failures with the CMA allocation set to 192MB - it was being created at a location that overlapped the HIGHMEM space at 0x30000000, which killed the CMA activation. The downstream file bcm2711-rpi.dtsi now patches the alloc-ranges to restrict CMA to the first 768MB, and I've not had any problems since.

vianpl:
@pelwell, are you forward porting that change to rpi-5.6.y?

pelwell:
Yes - I'm just (gasp) testing it first!

pelwell:
Done.

PR #3537, opened 2020-04-09
BSC0 config for I2C_MUX_PINCTRL to drive camera/display and GPIO

6by9:
Trying to ressurrect these patches to add the pinctrl i2c mux onto BSC0 so we can access both camera/display and HAT ID pins simultaneously.

Tested on Pi3B+ & Pi4 with bcm27xx dt, and Pi3B+ with bcm2837-rpi-3-b-plus (downstream kernel image, but with ```device_tree=bcm2837-rpi-3-b-plus.dtb``` override).
Need to test other variants, particularly the old versions where i2c1 was used for the camera, or it wasn't present on the header. I suspect we need to use delete-node to remove i2c10mux totally.


6by9:
@pelwell for comment.

Also need to decide what to do with CM and CM3 where normally we don't assign GPIOs for resources unless really needed.

arch/arm64/configs/bcm2711_defconfig
```diff
@@ -663,6 +663,7 @@ CONFIG_I2C_CHARDEV=m
 CONFIG_I2C_MUX=m
 CONFIG_I2C_MUX_GPMUX=m
 CONFIG_I2C_MUX_PCA954x=m
+CONFIG_I2C_MUX_PINCTRL=m
```
pelwell, line 665: We should include arch/arm64/configs/bcmrpi3_defconfig as well.

arch/arm/boot/dts/bcm283x.dtsi
```diff
@@ -327,7 +327,7 @@
 			status = "disabled";
 		};
 
-		i2c0: i2c@7e205000 {
+		i2c10: i2c@7e205000 {
```
pelwell: I'm not keen on the name `i2c10` - it doesn't convey much information - but I'm guessing you want to change it because the firmware expects the `i2c0` label to identify the header-accessible I2C0 pins (you don't say). `i2c0hw`, `i2c0if` or even `_i2c0` would be more obvious. But you should consider changing the firmware, since you will no longer be using `i2c_vc` from overlays.

arch/arm/boot/dts/bcm283x.dtsi
```diff
@@ -337,6 +337,30 @@
 			status = "disabled";
 		};
 
+		i2c10mux: i2c10mux {
```
pelwell: `i2c0mux`?

arch/arm/boot/dts/bcm2835-rpi-a-plus.dts
```diff
@@ -126,3 +126,8 @@
 	pinctrl-0 = <&uart0_gpio14>;
 	status = "okay";
 };
+
```
pelwell, line 129: [ I can't leave a comment against the commit message, so... ]

This commit modified the platform dts files. not the overlays.

arch/arm/boot/dts/bcm283x.dtsi
pelwell: And on the subject of the firmware and it's arm<->vc remapping, what is your proposal for the rev1 model B (&A?) boards that use I2C1 for the camera? Are we orphaning them?

pelwell (changes requested):
These comments and questions need addressing before scrutinising the individual overlay patches.

arch/arm/boot/dts/bcm283x.dtsi
6by9: Technically it's bsc0.
Yes, i2c0 needs to point at the GPIOs 0&1.

It's the weirdness that this base node gets registered as an /dev/i2cN node, but it wants to be up out the way and not used because it will just adopt the pinctrl of the last muxed access. i2c10 puts it out the way. AFAIK you can't create it as /dev/i2c0hw - it has to be a pure numeric suffix.

Hopefully ```dtparam=i2c_vc``` from the firmware should still enable the complete infrastructure of bsc0 and the mux (I think I updated both status values).
Use of ```i2c_vc``` from an overlay should be equivalent to selecting i2c0 on GPIOs 0&1 (ie the HAT ID pins).

I did make a comment about needing to refine rev1 model B's. There I think we have to remove the mux node totally and revert back to the old scheme. I believe we can do that via a ```/delete-node/```, as is done for pixelvalves on 2711 https://github.com/raspberrypi/linux/blob/rpi-5.4.y/arch/arm/boot/dts/bcm2711-rpi-4-b.dts#L188, and i2c_vc is inherently the CSI port as there were no HAT ID pins.
I'm not sure if we can get the aliases to all point to the right places.

6by9: Minor addendum to the above as I hit "comment" too soon.
Currently the Pi4 HDMI I2C interfaces are not fixed, so become i2c11 and i2c12. I don't know if we want to lock those down, and if so to what numbers.

arch/arm/boot/dts/bcm283x.dtsi
6by9: ```bsc0mux``` if you wish.
i2c0 is a physical device, but this mux is also controlling /dev/i2c<something else> connected to the camera and display port.

arch/arm/boot/dts/bcm2835-rpi-a-plus.dts
6by9: Ack. It was the mainline patch that was the one really still in progress when I did the push last thing Thursday.

arch/arm64/configs/bcm2711_defconfig
6by9: OK

arch/arm/boot/dts/bcm283x.dtsi
pelwell: So the motivation is to arrange that the /dev device name is still i2c0, while the kernel is trying to match the DT names to the device numbers? That makes sense - I'll have a ponder.

6by9:
Hmm, I need to do some double checking. Comparing against a Tegra board that uses i2c-mux-pinctrl, that needs to clear down the pinmuxing as well as setting up the new ones.

I currently haven't done that (just setting up the new pinctrl in each case), and that appeared to work. I'm a little confused as to how it did.

https://github.com/raspberrypi/linux/blob/rpi-5.4.y/arch/arm/boot/dts/tegra20-seaboard.dts#L288

arch/arm/boot/dts/bcm283x.dtsi
6by9: Yes, I wanted to leave any existing users with the same behaviour and functionality as they currently get, hopefully without them having to make any changes.

The exception will be for those who used the i2c0-bcm2708 overlay (or similar) to deliberately remap i2c0 to 28&29 or 44&45 for some other reason. That's not possible whilst also exposing i2c on both sets of gpios.

The camera/display I2C is currently going to end up as i2c9 due to the DT alias.

So it's what the additional (redundant) parent i2c device gets called that is the issue.

The i2c core does look through all devices to work out the highest statically allocated bus number, and all dynamically assigned busses start from that. We could leave this parent node as a dynamically assigned bus number.

pelwell: [ Back on this after an adventure in CMA-related non-booting ]

> Currently the Pi4 HDMI I2C interfaces are not fixed, so become i2c11 and i2c12. I don't know if we want to lock those down, and if so to what numbers.

Do we want users accessing the HDMI I2C interface other than via the HDMI driver?

arch/arm/boot/dts/overlays/Makefile
```diff
@@ -111,6 +111,7 @@ dtbo-$(CONFIG_ARCH_BCM2835) += \
 	mpu6050.dtbo \
 	mz61581.dtbo \
 	ov5647.dtbo \
+	pac193x.dtbo \
```
pelwell: Is this line spurious?

arch/arm/boot/dts/bcm283x.dtsi
6by9: We had the ```i2c2_iknowwhatimdoing``` dtparam for Pi0-3. There's no reason not to have an equivalent for Pi4.
Some people want to do DDC directly over I2C to their monitors as some allow control of brightness/contrast type parameters over.

arch/arm/boot/dts/overlays/Makefile
6by9: Yes. Compile testing for https://www.raspberrypi.org/forums/viewtopic.php?f=44&t=243562. Sorry.

pelwell (approved):

pelwell:
I've pushed a rebase and a whole bunch of changes in a fixups commit to get your opinion.

6by9:
I'm surprised you can push to a branch in my repo, but that's Github for you.

So you're:
-  renaming the parent node to i2c0if, and leaving it as a dynamically assigned i2c bus number
- renaming the mux to i2c0mux.
- renumbering i2c_csi_dsi to i2c10 (was i2c9).
- cleaning up a couple of overlay errors.

Yes, I'm happy on that lot. I'll squash them back into the parent commits and retest. Seeing as several of these files are upstream ones, and mainline should soonish be getting CSI and DSI support, it'd be nice to have them in a shape that can be upstreamed.

I've gone through my stash of devices and have a 0W, B+, 2B, 3B, 3B+, 4B, and CM3 here, so can check something sensible happens on all of them. (I think my 256MB B is sitting in the office for the awkward case). So apart from the hassle of building the 3 variants of kernels, I should be able to test out the majority that we care about.

6by9:
On further checking of the pinmuxing, something is clearing down each selection.
Checking /sys/kernel/debug/pinctrl/fe200000.gpio-pinctrl-bcm2835/pinmux-pins, the pins are going from ```MUX UNCLAIMED``` to ```soc:i2c10mux```, and then returning to ```MUX UNCLAIMED``` when selecting the other muxing again.

So whilst the bindings imply that you need to reset all the pin-ctrl groups, reality seems different. I guess specifying them all stops some other user coming along and claiming the mux.

What's your view? Explicitly state them in all states, or leave it as it is?

pelwell:
My guess is that the Tegra DTS is explicitly setting functions for the unused pins to avoid the default UNCLAIMED state, which seems to be "input". I think the current behaviour without explicitly reserving the other pins is fine.

pelwell:
That still leaves us with the thorny issue of the rev1 model B, which would ideally have a mux on i2c1. Instead of the firmware creating device-specific aliases/labels (i2c_vc, i2c_arm etc.) on all boards, it might be cleaner and more flexible to build the "normal" values into the base DTBs and apply a magic "modelb-rev1" overlay to patch things up. That does depend on us being able to undo/disable any unwanted settings of the base DTB from an overlay.

6by9:
> That still leaves us with the thorny issue of the rev1 model B, which would ideally have a mux on i2c1. 

There is no requirement for a mux on rev1 B - i2c1 **only** goes to the camera (GPIOs 2&3), and i2c0 **only** goes to the 26 pin header as pins 3&5. The alternate muxings via 28&29 and 44&45 aren't used. The mux needs removing, and pinctrl setup restored to the current setup.
I believe that can be done with a ```/delete-node/ i2c0mux```, adding the pin-ctrl to ```i2c0if```, and altering the alias for ```i2c0``` to pointing at ```i2c0if```.
The alias for i2c_vc would already be pointing at i2c1, but I guess the alias for i2c_arm is pointing at i2c0mux port0 (named ```i2c0```) which is wrong.  Or does the dtoverlay_set_synonym look for an alias instead of a label? You may know the answer, but I'd have to try it.

Unless having a mux with only one port for consistency is nicer. I think in this case it's uglier as it'd be an i2c1mux so you still need to do the above and then add a second mux.

pelwell:
I think that the only sane way to get the right end result is to fork the Model B .dts as upstream have done, but using `bcm2708-model-b-rev1` and `bcm2708-model-b` (as opposed to `bcm2708-model-b` and `bcm2708-model-b-rev2`) for backwards-compatibility. If we bake in the `i2c_vc` and `i2c_arm` aliases that's one less thing for the firmware to do.

pelwell:
And I can add `i2c_cam`, to distinguish it from the `i2c_vc` on the header.

6by9:
OK, that sounds like a plan.

I'm being hassled to get DSI stuff working, so I'm going to shelve this temporarily whilst using it myself.

```i2c_cam``` is also ```i2c_dsi_display```, hence the current name ```i2c_csi_dsi```. ```i2c_cam_display``` if you want to rename it?
(That is of course on all except the Model B rev 1 that forgot to route I2C to the display connector)

**Comment by pelwell** [comment]
I'll do the necessary rework and push it as clean patches, if that's OK with you.

You're right about the label/alias - i2c_csi_dsi is accurate, and descriptive enough.

**Comment by 6by9** [comment]
Hang fire then as I've merged all your earlier changes into the base commits, but I don't think I've pushed that. No point in duplicating that work.

**Comment by 6by9** [comment]
Hang fire then as I've merged all your earlier changes into the base commits, but I don't think I've pushed that. No point in duplicating that work.

**Comment by pelwell** [comment]
That's fine - I've not squashed anything yet.

**Comment by pelwell** [comment]
That's fine - I've not squashed anything yet.

**Comment by pelwell** [comment]
That's fine - I've not squashed anything yet.

**Comment by 6by9** [comment]
Rebased and repushed with the updates I'd applied from your fixups.
I've pushed the branch that you had fixed up to  rpi-5.4.y-unicam-pe as a backup.

I haven't touched the B rev1 config yet, not run exhaustive tests on the various boards.

File: arch/arm/boot/dts/bcm2708-rpi-zero-w.dts [hunk 6]
```diff
@@ -3,6 +3,7 @@
 #include "bcm2708.dtsi"
 #include "bcm2708-rpi.dtsi"
 #include "bcm283x-rpi-csi1-2lane.dtsi"
+#include "bcm283x-rpi-i2c0mux_0_28.dtsi"
```
pelwell, line 6: So I don't leap to the wrong conclusion, is there a reason why bcm2708-rpi-zero.dts doesn't get the same treatment? Rev 1.2 hardware didn't have the connector, but rev 1.3 does.

6by9: I suspect I just missed it.
Having the mux does no harm on the 1.2, and it has benefit on the 1.3, so better to have it in there.

arch/arm/boot/dts/bcm2708-rpi-b-plus.dts
```diff
@@ -22,6 +22,11 @@
 		brcm,function = <1>; /* output */
 	};
 
+	i2c0_pins: i2c0 {
+		brcm,pins = <0 1>;
+		brcm,function = <4>;
+	};
+
```
6by9: We have
```
			i2c0_gpio0: i2c0_gpio0 {
				brcm,pins = <0 1>;
				brcm,function = <BCM2835_FSEL_ALT0>;
			};
```
in bcm283x.dtsi. Any reason to duplicate?

arch/arm/boot/dts/bcm2708-rpi-b-plus.dts
```diff
@@ -68,6 +73,10 @@
 	clock-frequency = <100000>;
 };
 
+&i2c0mux {
+	pinctrl-0 = <&i2c0_pins>;
+};
+
```
6by9, line 79: bcm283x-rpi-i2c0mux_0_28.dtsi is included which sets
```
&i2c0mux {
	pinctrl-0 = <&i2c0_gpio0>;
	pinctrl-1 = <&i2c0_gpio28>;
};
```
Do we need to duplicate? Or is there a commit missing here?

6by9 (commented):
I'm confused by some of the extra changes

arch/arm/boot/dts/bcm2708-rpi-b-plus.dts
pelwell: Because existing overlays target `i2c0_pins`.

arch/arm/boot/dts/bcm2708-rpi-b-plus.dts
pelwell: i2c0_pins is a downstream entity so I don't want to refer to it from an upstream file, but I do want to retain the name, hence the apparent duplication.

pelwell:
Force-pushed to update the commit message for the rev1 addition. If you have no objections I'll squash the fixups with their respective targets and prepare to merge.

6by9:
If you're happy then I'm happy :-)

Issue #3513, opened 2020-03-21
[drm:vc4_hdmi_bind] *ERROR* Failed to get pixel clock


jameshilliard:
**Describe the bug**
vc4_hdmi driver does not appear to appropriately handle logging for `-EPROBE_DEFER` errors when trying to get the pixel clock.

**To reproduce**
Build and run the kernel with built in(non-module) vc4 drm driver.

**Expected behaviour**
Kernel driver should ignore non-fatal `-EPROBE_DEFER` errors or load the vc4 driver only after the pixel clock is available.

**Actual behaviour**
The vc4 kernel driver spams dmesg with `-EPROBE_DEFER` errors.

**System**
* Which model of Raspberry Pi? 
Raspberry Pi 3B
* Which OS and version (`cat /etc/rpi-issue`)?
Buildroot master
* Which firmware version (`vcgencmd version`)?
commit: [01ecfd2ba2b7cf3a2f4aa75ada895ee4a3e729f5](https://github.com/raspberrypi/firmware/tree/01ecfd2ba2b7cf3a2f4aa75ada895ee4a3e729f5)
* Which kernel version (`uname -a`)?
`Linux buildroot 4.19.97-v8 #1 SMP PREEMPT Fri Mar 20 17:43:53 MDT 2020 aarch64 GNU/Linux`
commit: [64d0a9870ac14d5eb5253f67d984ae348eec1393](https://github.com/raspberrypi/linux/tree/64d0a9870ac14d5eb5253f67d984ae348eec1393)

**Logs**
```
[    0.348120] [drm:vc4_hdmi_bind] *ERROR* Failed to get pixel clock
[    0.348204] vc4-drm soc:gpu: failed to bind 3f902000.hdmi (ops vc4_hdmi_ops): -517
[    0.348392] vc4-drm soc:gpu: master bind failed: -517
[    0.846746] vc4-drm soc:gpu: failed to bind 3f902000.hdmi (ops vc4_hdmi_ops): -517
[    0.846950] vc4-drm soc:gpu: master bind failed: -517
[    0.928787] vc4-drm soc:gpu: failed to bind 3f902000.hdmi (ops vc4_hdmi_ops): -517
[    0.929026] vc4-drm soc:gpu: master bind failed: -517
[    0.932427] vc4-drm soc:gpu: failed to bind 3f902000.hdmi (ops vc4_hdmi_ops): -517
[    0.932588] vc4-drm soc:gpu: master bind failed: -517
[...]

PR #3538, opened 2020-04-11
drm/vc4: hdmi: Silence pixel clock error on -EPROBE_DEFER

jameshilliard:
If the vc4 hdmi driver loads before the pixel clock is available we see a spurious "*ERROR* Failed to get pixel clock" error.

Partial fix for #3513.


6by9:
Acked by me. Will conflict against rpi-5.4.y with Maxime's reworking. The equivalent change needs to be made there too.

jameshilliard:
Upstream submission for reference https://patchwork.kernel.org/patch/11483961/.

PR #3539, opened 2020-04-11
component: Silence bind error on -EPROBE_DEFER

jameshilliard:
If a component fails to bind due to -EPROBE_DEFER we should not log an error as this is not a real failure.

Fixes:
```
vc4-drm soc:gpu: failed to bind 3f902000.hdmi (ops vc4_hdmi_ops): -517
vc4-drm soc:gpu: master bind failed: -517
```

Partial fix for #3513.

### DIFF
--- a/drivers/base/component.c
+++ b/drivers/base/component.c
@@ -235,7 +235,8 @@ static int try_to_bring_up_master(struct master *master,
 	ret = master->ops->bind(master->dev);
 	if (ret < 0) {
 		devres_release_group(master->dev, NULL);
-		dev_info(master->dev, "master bind failed: %d\n", ret);
+		if (ret != -EPROBE_DEFER)
+			dev_info(master->dev, "master bind failed: %d\n", ret);
 		return ret;
 	}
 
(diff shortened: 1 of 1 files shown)


6by9:
Acked by me.

drivers/base/component.c
```diff
@@ -506,8 +507,10 @@ static int component_bind(struct component *component, struct master *master,
 		devres_release_group(component->dev, NULL);
 		devres_release_group(master->dev, NULL);
 
-		dev_err(master->dev, "failed to bind %s (ops %ps): %d\n",
-			dev_name(component->dev), component->ops, ret);
+		if (ret != -EPROBE_DEFER) {
```
pelwell: Please remove the extra braces.

pelwell (changes requested):

drivers/base/component.c
jameshilliard: aren't the braces required due to the `dev_err` going across 2 lines?

jameshilliard:
Upstream submission for reference https://patchwork.kernel.org/patch/11483959/.

drivers/base/component.c
pelwell: C doesn't need them - delete them, unless you are saying that the Linux style requires them.

jameshilliard: ah, thought they were required by C(I don't really do a lot of C coding day to day), updated just now

pelwell:
Thanks.

pelwell:
C doesn't distinguish between spaces, TABs and newlines, so splitting a statement across multiple lines makes no difference. Braces would be needed if there were two statements within the body of the if.

jameshilliard:
Ah, probably mixed that up since I mostly write Python code day to day.

jameshilliard:
Upstream v2 patch with braces removed https://patchwork.kernel.org/patch/11484225/.

jameshilliard:
This has now [landed upstream](https://github.com/torvalds/linux/commit/7706b0a76a9697021e2bf395f3f065c18f51043d) FYI.

pelwell:
We noticed, thanks. rpi-5.7.y doesn't include your downstream patch.

PR #3544, opened 2020-04-14
IRS1125 sensor driver updates

Coimbra1984:
Changed a few functions:

- i2c_transfer used for reading
- minor refactoring including renaming
- V4L2 control for safe reconfiguration changed to allow individual access for each sequence
- Reconfiguration of modulation PLLs after imager reset fixed


drivers/media/i2c/irs1125.h
```diff
@@ -21,18 +21,57 @@
...
+#define IRS1125_CID_CONTINUOUS_TRIG	(IRS1125_CID_CUSTOM_BASE + 40)
+#define IRS1125_CID_TRIGGER		(IRS1125_CID_CUSTOM_BASE + 41)
+#define IRS1125_CID_RECONFIG		(IRS1125_CID_CUSTOM_BASE + 42)
+#define IRS1125_CID_ILLU_ON		(IRS1125_CID_CUSTOM_BASE + 43)
+#define IRS1125_CID_NUM_SEQS		(IRS1125_CID_CUSTOM_BASE + 44)
+#define IRS1125_CID_MOD_PLL		(IRS1125_CID_CUSTOM_BASE + 45)
+#define IRS1125_CID_SEQ_CONFIG		(IRS1125_CID_CUSTOM_BASE + 46)
+#define IRS1125_CID_IDENT0		(IRS1125_CID_CUSTOM_BASE + 47)
+#define IRS1125_CID_IDENT1		(IRS1125_CID_CUSTOM_BASE + 48)
+#define IRS1125_CID_IDENT2		(IRS1125_CID_CUSTOM_BASE + 49)
```
6by9: You are the only user of this driver at present, but rewriting all the control CIDs like this means you have totally destroyed any backwards compatibility.
Generally speaking you should retain the same CID where possible, deprectate (and not reuse) those which are obsolete, and add new controls to the end. A user app can then know that the control matches the headers that it has.

drivers/media/i2c/irs1125.c
```diff
@@ -29,6 +29,7 @@
 #include <media/v4l2-image-sizes.h>
 #include <media/v4l2-mediabus.h>
 #include <media/v4l2-ctrls.h>
+#include <asm/unaligned.h>
```
6by9: Includes should normally be in alhpabetical order.

drivers/media/i2c/irs1125.c
```diff
@@ -248,27 +249,31 @@ static int irs1125_write(struct v4l2_subdev *sd, u16 reg, u16 val)
...
-	ret = i2c_master_recv(client, rdval, 2);
-	if (ret < 0)
-		dev_err(&client->dev, "%s: i2c read error, reg: %x\n",
-			__func__, reg);
+	ret = i2c_transfer(client->adapter, msgs, ARRAY_SIZE(msgs));
+	if (ret != ARRAY_SIZE(msgs))
+		return -EIO;
 
-	*val = rdval[1] | (rdval[0] << 8);
+	*val = get_unaligned_be16(data_buf);
```
6by9: Personally I prefer the explicit assignment rather than the use of the be16 functions here as the reassembly isn't down to CPU endianness. Others seem to use them in drivers/media/i2c though, so I'm not going to fuss.

6by9 (commented):
A couple of comments for you to consider. You are the only users of this driver, therefore I'm not going to fuss over them excessively. Should you ever try to upstream this then you will get the same sort of queries there, and breaking backwards compatibility would be a big no-no.

drivers/media/i2c/irs1125.h
Coimbra1984: Sure, you are right, of course.

drivers/media/i2c/irs1125.c
Coimbra1984: Okay, I didn't know that

drivers/media/i2c/irs1125.c
Coimbra1984: No problem, I'll use the previous version.

drivers/media/i2c/irs1125.c
```diff
@@ -248,27 +248,31 @@ static int irs1125_write(struct v4l2_subdev *sd, u16 reg, u16 val)
...
+	msgs[1].flags = I2C_M_RD;
+	msgs[1].len = 2;
+	msgs[1].buf = data_buf;
 
-	ret = i2c_master_recv(client, rdval, 2);
-	if (ret < 0)
-		dev_err(&client->dev, "%s: i2c read error, reg: %x\n",
-			__func__, reg);
+	ret = i2c_transfer(client->adapter, msgs, ARRAY_SIZE(msgs));
+	if (ret != ARRAY_SIZE(msgs))
```
6by9: https://elixir.bootlin.com/linux/latest/source/drivers/i2c/i2c-core-base.c#L1987
"Returns negative errno, else the number of messages executed."
So this masks any error number and will always return -EIO on failure.

```
if (ret != ARRAY_SIZE(msgs)) {
  if (ret >= 0)
    ret = -EIO;
  return ret;
}
```

drivers/media/i2c/irs1125.c
```diff
@@ -560,133 +560,144 @@ static int irs1125_s_ctrl(struct v4l2_ctrl *ctrl)
...
-			if (mod_cur[i].pllcfg8 != mod_new[i].pllcfg8) {
-				addr = 0xC250 + i * 5;
-				val = mod_new[i].pllcfg8;
-				err = irs1125_write(&dev->sd, addr, val);
-				if (err < 0)
-					break;
-			}
+			addr = 0xC3A0 + i * 3;
+			val = mod_new[i].pllcfg1;
+			err = irs1125_write(&dev->sd, addr, val);
```
6by9: Seeing as addr is only used the once, the calculation may as well be in the call to irs1125_write
Likewise you only use val once.
```err = irs1125_write(&dev->sd, 0xC3A0 + i * 3, mod_new[i].pllcfg1);```

Same all through this function.

Comment (6by9) [reply drivers/media/i2c/irs1125.c]: There could be a minor optimisation in computing i*3 (and i*5) once per loop, and then using that as a pll_offset value to be added in each call.

File: drivers/media/i2c/irs1125.c [hunk 6]
```diff
@@ -921,6 +921,38 @@ static int irs1125_ctrls_init(struct irs1125 *sensor, struct device *dev)
...
+		ctrl_cfg.name = name;
+		ctrl = v4l2_ctrl_new_custom(hdl, &ctrl_cfg,
+					    NULL);
+		if (!ctrl)
+			dev_err(dev, "Failed to init custom control %s\n",
+				ctrl_cfg.name);
+
+		snprintf(name, sizeof(name),
+			 "Safe reconfiguration of framerate of sequence %d", i);
+		ctrl_cfg.id += 1;
```
6by9: ```ctrl_cfg.id++;``` would be more normal.

drivers/media/i2c/irs1125.c
```diff
@@ -921,6 +921,38 @@ static int irs1125_ctrls_init(struct irs1125 *sensor, struct device *dev)
...
+		snprintf(name, sizeof(name),
+			 "Safe reconfiguration of exposure of sequence %d", i);
+		ctrl_cfg.name = name;
+		ctrl = v4l2_ctrl_new_custom(hdl, &ctrl_cfg,
+					    NULL);
+		if (!ctrl)
+			dev_err(dev, "Failed to init custom control %s\n",
+				ctrl_cfg.name);
+
+		snprintf(name, sizeof(name),
```
6by9: This has me worried.
```struct v4l2_ctrl``` only has a ```const char *name``` you're passing in a stack allocated pointer, and I don't believe anything in the framework is going to take a persistent copy of that. That means that ```name``` is pointer into random stuff, and is the same for all of your controls.

Does ```v4l2-ctl --list-ctrls``` actually list out the right thing?
I suspect not, and you really need two arrays of const char* names, with IRS1125_NUM_SEQ_ENTRIES entries in each. 

Doing that would also mean that it'd be neater to have two loops, one for adding the exposure ctrls, and one for the framerate ones. Initialise the ```struct v4l2_ctrl_config``` outside the loop, and then insert id and pointer to the name. i+=2 in the loop increment.

File: drivers/media/i2c/irs1125.c [hunk 8]
```diff
@@ -748,33 +759,22 @@ static const struct v4l2_ctrl_config irs1125_custom_ctrls[] = {
 		.id = IRS1125_CID_MOD_PLL,
 		.name = "Reconfigure modulation PLLs",
 		.type = V4L2_CTRL_TYPE_U16,
-		.flags = V4L2_CTRL_FLAG_HAS_PAYLOAD,
+		.flags = V4L2_CTRL_FLAG_HAS_PAYLOAD |
+			V4L2_CTRL_FLAG_EXECUTE_ON_WRITE,
```
6by9: Do you really need to write it every time, even if it hasn't changed? I'm willing to believe it, but it's slightly unusual.

drivers/media/i2c/irs1125.c
```diff
@@ -614,8 +613,23 @@ static int irs1125_s_ctrl(struct v4l2_ctrl *ctrl)
 	}
 	case IRS1125_CID_MOD_PLL: {
 		struct irs1125_mod_pll *mod_new;
+		u8 *pval;
+		bool valid;
+		int s;
 
+		s = sizeof(struct irs1125_mod_pll) * IRS1125_NUM_MOD_PLLS;
```
6by9: Can be done as the one line
```int s = sizeof(struct irs1125_mod_pll) * IRS1125_NUM_MOD_PLLS;```
(line wrap at the * if it's over 80 chars)

File: drivers/media/i2c/irs1125.c [hunk 10]
```diff
@@ -614,8 +613,23 @@ static int irs1125_s_ctrl(struct v4l2_ctrl *ctrl)
...
 		mod_new = (struct irs1125_mod_pll *)ctrl->p_new.p;
+
+		for (pval = (u8 *)mod_new; pval < (u8 *)mod_new + s; pval++) {
+			if (*pval != 0) {
+				valid = true;
+				break;
+			}
+		}
+		if (!valid)
+			break;
```
6by9, line 669: Do we not want to return an error if invalid?

drivers/media/i2c/irs1125.c
```diff
@@ -614,8 +613,23 @@ static int irs1125_s_ctrl(struct v4l2_ctrl *ctrl)
...
+		s = sizeof(struct irs1125_mod_pll) * IRS1125_NUM_MOD_PLLS;
+		valid = false;
 		mod_new = (struct irs1125_mod_pll *)ctrl->p_new.p;
+
+		for (pval = (u8 *)mod_new; pval < (u8 *)mod_new + s; pval++) {
+			if (*pval != 0) {
+				valid = true;
+				break;
+			}
+		}
```
6by9: I'm a little confused by the validity test.
If any one 8bit value within an array of 8 16bit values is non-zero, then the whole thing is valid?

So all of these are all equally valid
0x0000 0x0000 0x0000 0x0001 0x0000  0x0000  0x0000 0x0000 
0xdead 0xbeef 0xdead 0xbeef 0xdead 0xbeef 0xdead 0xbeef
0x0000 0x0000 0x0000 0x000 0x0000  0x0000  0x0000 0x8000 
0x8000 0x0000 0x0000 0x000 0x0000  0x0000  0x0000 0x0000 
I'll believe you. Do you really need to constrain this because the device dies if all 0's are set, or can you just let userspace worry about setting something sane?

And why test as 8bit values? The same comparison can be done on the u16 values, again treated as an array.

6by9 (changes requested):
A few concerns/comments.

The commit texts could be more descriptive too.

drivers/media/i2c/irs1125.c
Coimbra1984: I have a conceptual problem here. I would also like userspace to worry about setting something sane, but IRS1125_CID_MOD_PLL is a compound control. I want to keep controls and hardware in sync which I guess is a key concept of V4L2 (I'll remove all unnecessary V4L2_CTRL_FLAG_EXECUTE_ON_WRITE).

Now the problem is, that when userspace calls a "close" on the video device, the imager goes into deep sleep, which I guess is a good idea to safe power. But in this state, it also looses all its register contents and the V4L2 controls need to be synced with the HW again after the wakeup.
I do that inside __sensor_init with v4l2_ctrl_handler_setup(...). Here the IRS1125_CID_MOD_PLL contains a default value of "all zeros" which kills the imager. I'll look deeper into that problem, I think that the pre-devider value of 0 might be the problem, which would make the code much clearer here.

Anyway I did not find any other solution to that problem. Maybe you have one...? I think a p_def member to v4l2_ctrl_config is added to more recent kernel versions (https://lore.kernel.org/patchwork/patch/1147819/) which would give me the possibility to add a senseful default value for IRS1125_CID_MOD_PLL.

6by9: OK, so you need IRS1125_CID_MOD_PLL initialised to something, or skip it if it hasn't been set.

Do you also need to guard against userspace setting all 0's? This will show up as a standard V4L2 device, so if all zeros is really fatal (hardware damage fatal?) then the driver ought to guard against it.
Minor nitpick - ```if (*pval) {``` is neater than ```if (*pval != 0) {```
It would also be nice if the sensor would just stream with default values, but I realise that this may not be terribly useful.

If you rebase against 5.4, then I'm happy for those patches from 5.5 to be cherry-picked back (probably ignore the vivid ones). We do insert a line 
``` Commit <hash> upstream```
into the commit text when cherry-picking back, just to make maintenance easier.
You can then set a default :-)

File: drivers/media/i2c/irs1125.c (hunk 4 above)
Comment (Coimbra1984) [reply drivers/media/i2c/irs1125.c]: You are right. This should be fixed now

File: drivers/media/i2c/irs1125.c (hunk 5 above)
Comment (Coimbra1984) [reply drivers/media/i2c/irs1125.c]: right, done!

File: drivers/media/i2c/irs1125.c (hunk 6 above)
Comment (Coimbra1984) [reply drivers/media/i2c/irs1125.c]: I changed that now

File: drivers/media/i2c/irs1125.c (hunk 7 above)
Comment (Coimbra1984) [reply drivers/media/i2c/irs1125.c]: You are right, the ctrl did not show up as expected. I changed this to use const char* arrays now.

I tested it again with `v4l2-ctl --lsit-ctrls` and it shows the expected results now.

File: drivers/media/i2c/irs1125.c (hunk 8 above)
Comment (Coimbra1984) [reply drivers/media/i2c/irs1125.c]: No this is not needed, you are right.

File: drivers/media/i2c/irs1125.c (hunk 9 above)
Comment (Coimbra1984) [reply drivers/media/i2c/irs1125.c]: Sure, this is changed now

File: drivers/media/i2c/irs1125.c (hunk 10 above)
Comment (Coimbra1984) [reply drivers/media/i2c/irs1125.c:669]: I now skip only writing this control in case of internal sync.

File: drivers/media/i2c/irs1125.c (hunk 11 above)
Comment (Coimbra1984) [reply drivers/media/i2c/irs1125.c]: You can look at the current change. I think it is a good idea, that the userspace is responsible for writing senseful data in this control. I skip writing this control only in case of the internal sync after HW power up.

File: drivers/media/i2c/irs1125.c [hunk 12]
```diff
@@ -1019,6 +1108,7 @@ static int irs1125_probe(struct i2c_client *client,
 	}
 
 	sensor->num_seq = 5;
+	sensor->ctrl_init = false;
```
6by9: Technically redudnant as you did a kzalloc of sensor.

drivers/media/i2c/irs1125.c
```diff
@@ -552,133 +612,147 @@ static int irs1125_s_ctrl(struct v4l2_ctrl *ctrl)
 					struct irs1125, ctrl_handler);
 	struct i2c_client *client = v4l2_get_subdevdata(&dev->sd);
 	int err, i;
-	struct irs1125_mod_pll *mod_cur, *mod_new;
-	struct irs1125_seq_cfg *cfg_cur, *cfg_new;
-	u16 addr, val;
 
 	err = 0;
```
6by9: Observation in passing, it's more normal to declare an initialise something like this in one step, ie
```
int err = 0, i;
```
at line 614.

drivers/media/i2c/irs1125.c
```diff
@@ -552,133 +612,147 @@ static int irs1125_s_ctrl(struct v4l2_ctrl *ctrl)
...
+				    IRS1125_REG_SAFE_RECONFIG + offset,
+				    ctrl->val);
 		break;
 	}
-	case IRS1125_CID_MOD_PLL:
+	case IRS1125_CID_MOD_PLL: {
+		struct irs1125_mod_pll *mod_new;
+
+		if (dev->ctrl_init)
+			return 0;
```
6by9: Seeing as err is already 0, you could just ```break;```

drivers/media/i2c/irs1125.c
```diff
@@ -82,6 +82,7 @@ struct irs1125 {
 	struct v4l2_ctrl *ctrl_numseq;
 
 	int power_count;
+	bool ctrl_init;
```
6by9: I can't comment on the commit text directly, but the line
> Thus, this ctrl is not written inside the driver into after power up. The userspace has to take care to write senseful data here.

is a little oddly phrased.
>Thus, this ctrl is not written to the hardware until it has been initialised by userspace. Userspace must take care to write meaningful data to the control - it is not validated.

Secondly ```ctrl_init``` is rather generic. It's specifically pll_ctrl_init, or pll_ctrl_set.

6by9 (commented):
Down to the minor nit picks now.

drivers/media/i2c/irs1125.c
Coimbra1984: Right, removed

drivers/media/i2c/irs1125.c
Coimbra1984: Okay, I changed that. Actually I always separate declarations and initializations. A former colleague told me to do so. If I remember correctly, he was a freeBSD developer and they are very strict there, an excerpt of their coding styles:

> Be careful to not obfuscate the code by initializing variables in the
> declarations.  Use this feature only thoughtfully.

drivers/media/i2c/irs1125.c
Coimbra1984: Changed.

drivers/media/i2c/irs1125.c
Coimbra1984: I hope, the comment is better now.

I renamed `ctrl_init` to `mod_pll_init`

Coimbra1984:
@6by9 I think I have reworked everything according to your comments.

6by9:
Thanks. As long as this still works with the hardware then I think I'm happy.

bjajoh:
@Coimbra1984 @6by9  I have tested it and it works just fine.

Coimbra1984:
@6by9 I tested it also. Works fine. @bjajoh is our best employee. He also tested this version with success. I think it is good.

pelwell:
Based on @6by9's review, the fact that it looks sensible, and the limited scope for collateral damage, I'm happy to merge.

PR #3545, opened 2020-04-15
Fixes a problem with clock settings of HiFiBerry DAC+ADC PRO

j-schambacher:
This patch fixes a problem of the re-calculation of
i2s-clock and -parameter settings when only the ADC is activated.

### DIFF
--- a/sound/soc/bcm/hifiberry_dacplusadcpro.c
+++ b/sound/soc/bcm/hifiberry_dacplusadcpro.c
@@ -419,10 +419,8 @@ static int snd_rpi_hifiberry_dacplusadcpro_hw_params(
 	if (ret)
 		return ret;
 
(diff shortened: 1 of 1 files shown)


sound/soc/bcm/hifiberry_dacplusadcpro.c
```diff
@@ -414,6 +416,13 @@ static int snd_rpi_hifiberry_dacplusadcpro_hw_params(
 		return ret;
 	ret = snd_soc_dai_set_tdm_slot(rtd->codec_dais[1], 0x03, 0x03,
 		channels, width);
+	if (ret)
+		return ret;
+
+	if (snd_rpi_hifiberry_is_dacpro) {
```
pelwell: You can turn the nested if into an && and lose the braces. Otherwise I'm OK with this patch.

PR #3551, opened 2020-04-17
Unicam Frame Start event

naushir:
Need to trigger a FS event in order to correctly drive exposure and gain changes from userland.


naushir:
@6by9 and @pelwell, could you eyeball this please?  I've reviewed and verified it does what we want.

6by9 (approved):
LGTM

naushir:
@popcornmix or @pelwell could you merge if you are happy with this please?

PR #3552, opened 2020-04-17
media: imx219: Advertise embedded data node on media pad 1

naushir:
userland.

### DIFF
--- a/drivers/media/i2c/imx219.c
+++ b/drivers/media/i2c/imx219.c
@@ -112,6 +112,16 @@
 #define IMX219_TESTP_BLUE_DEFAULT	0
 #define IMX219_TESTP_GREENB_DEFAULT	0
 
+/* Embedded metadata stream structure */
+#define IMX219_EMBEDDED_LINE_WIDTH 16384
+#define IMX219_NUM_EMBEDDED_LINES 1
+
+enum pad_types {
+	IMAGE_PAD,
+	METADATA_PAD,
+	NUM_PADS
+};
+
 struct imx219_reg {
 	u16 address;
 	u8 val;
@@ -503,7 +513,7 @@ static const struct imx219_mode supported_modes[] = {
 
 struct imx219 {
 	struct v4l2_subdev sd;
-	struct media_pad pad;
+	struct media_pad pad[NUM_PADS];
 
 	struct v4l2_mbus_framefmt fmt;
 
@@ -652,17 +662,25 @@ static void imx219_set_default_format(struct imx219 *imx219)
 static int imx219_open(struct v4l2_subdev *sd, struct v4l2_subdev_fh *fh)
 {
 	struct imx219 *imx219 = to_imx219(sd);
-	struct v4l2_mbus_framefmt *try_fmt =
-		v4l2_subdev_get_try_format(sd, fh->pad, 0);
+	struct v4l2_mbus_framefmt *try_fmt_img =
+		v4l2_subdev_get_try_format(sd, fh->pad, IMAGE_PAD);
+	struct v4l2_mbus_framefmt *try_fmt_meta =
+		v4l2_subdev_get_try_format(sd, fh->pad, METADATA_PAD);
 
 	mutex_lock(&imx219->mutex);
 
-	/* Initialize try_fmt */
-	try_fmt->width = supported_modes[0].width;
-	try_fmt->height = supported_modes[0].height;
-	try_fmt->code = imx219_get_format_code(imx219,
-					       MEDIA_BUS_FMT_SRGGB10_1X10);
-	try_fmt->field = V4L2_FIELD_NONE;
+	/* Initialize try_fmt for the image pad */
+	try_fmt_img->width = supported_modes[0].width;
+	try_fmt_img->height = supported_modes[0].height;
+	try_fmt_img->code = imx219_get_format_code(imx219,
+						   MEDIA_BUS_FMT_SRGGB10_1X10);
+	try_fmt_img->field = V4L2_FIELD_NONE;
+
+	/* Initialize try_fmt for the embedded metadata pad */
+	try_fmt_meta->width = IMX219_EMBEDDED_LINE_WIDTH;
+	try_fmt_meta->height = IMX219_NUM_EMBEDDED_LINES;
+	try_fmt_meta->code = MEDIA_BUS_FMT_SENSOR_DATA;
+	try_fmt_meta->field = V4L2_FIELD_NONE;
 
 	mutex_unlock(&imx219->mutex);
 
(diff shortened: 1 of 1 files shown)


naushir:
@6by9, here's a first pass at adding embedded data pad to imx219.

drivers/media/i2c/imx219.c
```diff
@@ -538,16 +547,23 @@ static int imx219_open(struct v4l2_subdev *sd, struct v4l2_subdev_fh *fh)
...
 	try_fmt->code = imx219_get_format_code(imx219);
 	try_fmt->field = V4L2_FIELD_NONE;
 
+	/* Initialize try_fmt for the embedded metadata pad */
+	try_fmt = v4l2_subdev_get_try_format(sd, fh->pad, METADATA_PAD);
+	try_fmt->width = IMX219_EMBEDDED_LINE_WIDTH;
+	try_fmt->height = IMX219_NUM_EMBEDDED_LINES;
+	try_fmt->code = MEDIA_BUS_FMT_SENSOR_DATA;
+	try_fmt->field = V4L2_FIELD_NONE;
+
```
6by9, line 684: So one v4l2_subdev_get_try_format is called with the mutex held, and one without. If the comment on the mutex is to be believed, then it's only needed for the call to imx219_get_format_code.

6by9 (commented):
One comment on the patch.

However there have been a number of patches submitted (not merged) upstream which would be worth backporting first. I have them on my local branch, so I'll push that now.
If we don't do it now, then the diffs will be more of a pain to handle later.

naushir:
Ok, ill rebase this change off the latest patches once you have pushed them.  Can you ping me the PR when it's ready?

6by9:
Minor correction - 3 of the patches have been merged into mainline. The 4th is a bug fix to the 2nd, so worth having.
#3553.

drivers/media/i2c/imx219.c
naushir: To be safe, i've done both v4l2_subdev_get_try_format calls before the mutex is held.

naushir:
Rebased on top of PR https://github.com/raspberrypi/linux/pull/3553.

drivers/media/i2c/imx219.c
6by9: v4l2_subdev_get_try_format is only getting a pointer to a structure within v4l2_subdev_pad_config (which is fh->pad).

As fh has only just been created, there is no issue over holding the mutex. It was more that one was acquired with it held and one without.

drivers/media/i2c/imx219.c
```diff
@@ -767,7 +784,14 @@ static int imx219_enum_mbus_code(struct v4l2_subdev *sd,
...
 
-	code->code = imx219_get_format_code(imx219, codes[code->index * 4]);
+	if (code->pad > METADATA_PAD)
+		return -EINVAL;
+
+	if (code->pad == IMAGE_PAD)
+		code->code = imx219_get_format_code(imx219,
+						    codes[code->index * 4]);
+	else
+		code->code = MEDIA_BUS_FMT_SENSOR_DATA;
```
6by9, line 798: You'll return two formats of type MEDIA_BUS_FMT_SENSOR_DATA on the metadata node as this is written.
The ```if (code->index >= (ARRAY_SIZE(codes) / 4))``` is range limiting code->index for the image pad. The metadata pad wants to limit as ```code->index > 0``` as it only has one format.

drivers/media/i2c/imx219.c
```diff
@@ -778,16 +802,29 @@ static int imx219_enum_frame_size(struct v4l2_subdev *sd,
...
+
+		if (fse->code != imx219_get_format_code(imx219, fse->code))
+			return -EINVAL;
+
+		fse->min_width = supported_modes[fse->index].width;
+		fse->max_width = fse->min_width;
+		fse->min_height = supported_modes[fse->index].height;
+		fse->max_height = fse->min_height;
+	} else {
+		if (fse->code != MEDIA_BUS_FMT_SENSOR_DATA)
```
6by9: Range check index on the metadata pad. 
```if (code->index > 0) return -EINVAL;```

File: drivers/media/i2c/imx219.c [hunk 4]
```diff
@@ -112,6 +112,15 @@
...
 #define IMX219_TESTP_GREENB_DEFAULT	0
 
+/* Embedded metadata stream structure */
+#define IMX219_EMBEDDED_LINE_WIDTH 16384
+#define IMX219_NUM_EMBEDDED_LINES 1
+
+enum pad_types {
+	IMAGE_PAD,
+	METADATA_PAD,
+};
```
pelwell, line 123: You could add a count/num terminator (I'd call it PAD_COUNT if the names were the other way round) and use it in place of the literal `2`s below. Otherwise it looks reasonable to the untrained eye.

drivers/media/i2c/imx219.c
naushir: Oops, that test (and the one below) actually got lost when I rebased.  Will fix.

drivers/media/i2c/imx219.c
naushir: Will fix.

6by9: If you have a PAD_COUNT, then all the ```if (code->pad > METADATA_PAD)``` checks ought to be ```if (code->pad >= PAD_COUNT```.

drivers/media/i2c/imx219.c
naushir: Done

drivers/media/i2c/imx219.c
naushir: Done

drivers/media/i2c/imx219.c
naushir: Done.

6by9:
Something wrong with your branch - conflicts in odd files.

naushir:
Hmmm, I see that, not entirely sure what happened.  I will rebase and re-push.

naushir:
Really have no idea what went wrong there.  There was no rebase needed, I just force pushed again and it seems to be fine...???

drivers/media/i2c/imx219.c
```diff
@@ -503,7 +513,7 @@ static const struct imx219_mode supported_modes[] = {
 
 struct imx219 {
 	struct v4l2_subdev sd;
-	struct media_pad pad;
+	struct media_pad pad[2];
```
pelwell: This should be another NUM_PADS.

naushir: Missed that one.  Fixed.

pelwell:
@6by9 Are you happy for this to be merged?

drivers/media/i2c/imx219.c
```diff
@@ -855,51 +915,74 @@ static int imx219_set_pad_format(struct v4l2_subdev *sd,
...
+			 */
+			hblank = IMX219_PPL_DEFAULT - mode->width;
+			__v4l2_ctrl_modify_range(imx219->hblank, hblank, hblank,
+						 1, hblank);
+		}
+	} else {
+		if (fmt->which == V4L2_SUBDEV_FORMAT_TRY) {
+			framefmt = v4l2_subdev_get_try_format(sd, cfg,
+							      fmt->pad);
+			*framefmt = fmt->format;
```
6by9, line 981: Seeing as we're only getting the TRY format, which is fixed in imx219_open to be the same numbers as set in imx219_update_metadata_pad_format, I think they are always going to be the same. Might as well leave them as independent just in case it gets changed in future.

6by9:
Yes, I'm happy with this.

PR #3554, opened 2020-04-17
configs: Enable CONFIG_ARM_BCM2835_CPUFREQ

popcornmix:
We actually use the 64-bit 2711 kernel on Pi3 too.
It still uses the downstream cpufreq driver in DT.

See: https://www.raspberrypi.org/forums/viewtopic.php?f=29&t=269769&start=100#p1644752


popcornmix:
TODO: can we use upstream cpufreq driver on Pi3.

pelwell:
> TODO: can we use upstream cpufreq driver on Pi3.

AFAIK the Pi 3 doesn't include that bit of STB hardware, so I don't see how it could work.

pelwell:
Ignore me - crossed wires.

pelwell:
See https://github.com/raspberrypi/linux/pull/3604.

Issue #3555, opened 2020-04-17
Enable Wireguard on 5.6.y by default


n0toose:
After I opened an issue (#3550) out of confusion, due to the fact that I couldn't figure out what was up with Wireguard, I spent my day reading up more about Linux kernel development and experimenting with it, with the help of the folk over at the #raspberrypi IRC channel on Freenode.

I figured that maybe I should perhaps ask whether enabling Wireguard by default on the [5.6.y](https://github.com/raspberrypi/linux/tree/rpi-5.6.y) tree could be considered.

n0toose:
I've prepared [my own tree](https://github.com/AlwaysLivid/linux-rpi-wg/tree/rpi-5.6.y-wg) containing my own changes that are meant for personal use, but I'd like to open a pull request with those rather simple changes if that's alright.

pelwell:
That's very polite of you, but it's less email to just open the pull request (politely, of course). I see you've enabled it on 64-bit only, which ought to make any overhead in size or performance less noticeable.

PR #3557, opened 2020-04-19
Enabled Wireguard on arm64 configs

n0toose:
Continuation of #3555

`I see you've enabled it on 64-bit only, which ought to make any overhead in size or performance less noticeable.`

I really wasn't sure as to whether I should enable it on other configurations, primarily because I'd assume that since Wireguard works significantly better on `arm64` kernels, one would probably opt for `arm64` instead. I'd gladly add the same changes to other configurations.


pelwell:
Merged manually - see https://github.com/raspberrypi/linux/commit/6a69fd5bd5c55dbd468f4aff95eb77b34d8b849e (the settings weren't quite as "make savedefconfig" would leave them).

n0toose:
> Merged manually - see [6a69fd5]

Cool, thanks for taking the time to correct this for me as well!

PR #3561, opened 2020-04-21
Revert "USB: hub: Don't record a connect-change event during reset-resume"

P33M:
The offending commit is also on 4.19, so please cherry-pick there if there's going to be another 4.19 stable release.

I assume the lengthy description in the original commit means it's eventually going to be part of a larger cleanup, but for now this fixes Pi 4 downstream.

### DIFF
--- a/drivers/usb/core/hub.c
+++ b/drivers/usb/core/hub.c
@@ -1222,6 +1222,11 @@ static void hub_activate(struct usb_hub *hub, enum hub_activation_type type)
 #ifdef CONFIG_PM
 			udev->reset_resume = 1;
 #endif
+			/* Don't set the change_bits when the device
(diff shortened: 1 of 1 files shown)


popcornmix:
I think this needs reporting upstream. Maintaining a revert long term tends to be painful.
I didn't read anything from the commit message that suggested this commit was temporary.

pelwell:
I agree, but in the meantime it's in rpi-4.19.y and rpi-5.6.y as well.

P33M:
In the process of digging around, it looks like we're racing with upstream:
https://www.spinics.net/lists/linux-usb/msg193667.html

I suspect that the patch will get backported to upstream stable quite quickly.

PR #3562, opened 2020-04-21
media: bcm2835-unicam: Re-fetch mbus code from subdev on a g_fmt call

naushir:
The sensor subdevice may change the Bayer order if a H/V flip is
requested after a s_fmt call.  Unicam g_fmt must call the subdev get_fmt
in case this has happened and return out the correct format 4cc.

### DIFF
--- a/drivers/media/platform/bcm2835/bcm2835-unicam.c
+++ b/drivers/media/platform/bcm2835/bcm2835-unicam.c
@@ -967,11 +967,30 @@ static int unicam_enum_fmt_vid_cap(struct file *file, void  *priv,
 static int unicam_g_fmt_vid_cap(struct file *file, void *priv,
 				struct v4l2_format *f)
 {
+	struct v4l2_mbus_framefmt mbus_fmt = {0};
 	struct unicam_node *node = video_drvdata(file);
+	struct unicam_device *dev = node->dev;
+	const struct unicam_fmt *fmt = NULL;
+	int ret;
 
-	if (node->pad_id == METADATA_PAD)
+	if (node->pad_id != IMAGE_PAD)
 		return -EINVAL;
 
+	/*
+	 * If a flip has occurred in the sensor, the fmt code might have
+	 * changed. So we will need to re-fetch the format from the subdevice.
+	 */
+	ret = __subdev_get_format(dev, &mbus_fmt, node->pad_id);
+	if (ret)
+		return -EINVAL;
+
+	/* Find the V4L2 format from mbus code. We must match a known format. */
+	fmt = find_format_by_code(mbus_fmt.code);
+	if (!fmt)
+		return -EINVAL;
+
+	node->fmt = fmt;
+	node->v_fmt.fmt.pix.pixelformat = fmt->fourcc;
 	*f = node->v_fmt;
 
 	return 0;


naushir:
@6by9, any other way you can think of for doing this?

naushir:
@6by9 this one needs reviewing as well when you have a chance.

drivers/media/platform/bcm2835/bcm2835-unicam.c
```diff
@@ -967,11 +967,30 @@ static int unicam_enum_fmt_vid_cap(struct file *file, void  *priv,
...
+	if (ret)
 		return -EINVAL;
 
+	/* Find the V4L2 format from mbus code. We must match a known format. */
+	fmt = find_format_by_code(mbus_fmt.code);
+	if (!fmt)
+		return -EINVAL;
+
+	node->fmt = fmt;
+	node->v_fmt.fmt.pix.pixelformat = fmt->fourcc;
```
6by9, line 993: Possibly simpler to call ```unicam_reset_format``` which will update ```node->v_fmt``` with the correct complete format structure, not just update the pixelformat.

6by9: That's what unicam_s_std does, and setting the video standard will generally update the resolution (eg from NTSC 720x480 to PAL 720x576), and unicam_s_dv_timings which does the same thing for digital sources.

naushir: Yes, I did think that I should use `unicam_reset_format()`, but looking at what that function does:

```
	ret = __subdev_get_format(dev, &mbus_fmt, node->pad_id);
	if (ret) {
		unicam_err(dev, "Failed to get_format - ret %d\n", ret);
		return ret;
	}

	if (mbus_fmt.code != node->fmt->code) {
		unicam_err(dev, "code mismatch - fmt->code %08x, mbus_fmt.code %08x\n",
			   node->fmt->code, mbus_fmt.code);
		return ret;
	}
```
So the `unicam_g_fmt_vid_cap` function would still have to call `__subdev_get_format` and update `node->fmt`, otherwise `unicam_reset_format()` bombs out early in the clause above.

If I were to call `unicam_reset_format()`, it would redo the `__subdev_get_format` call and update the `node->v_fmt` struct.  The reason for only updated the pixelformat - that is the only thing that can change (due to a h/v flip) between a call to `s_fmt` and `g_fmt` (although I may be wrong here).

Of course, I'll use your judgement as to what you think is best here.

6by9: I think we drop the 
```
		if (mbus_fmt.code != node->fmt->code) {
			unicam_err(dev, "code mismatch - fmt->code %08x, mbus_fmt.code %08x\n",
				   node->fmt->code, mbus_fmt.code);
			return ret;
		}
``` 
clause from unicam_reset_format. 
The source subdev is the ultimate master of the format that is produced, therefore we ought to just abide by what it says. I don't see the real need for two alternate versions of very nearly the same thing.

I thought we did a call to __subdev_get_format just before streaming too to ensure we had the correct properties, but I can't see it. Then again I'd need to think as to whether that is correct, as we shouldn't change the buffer format that userspace thought it had. If the source is sending something else, then it'll get mis-saved to memory, but the format will be what userspace thought it was.

naushir: Hmmm, this may not be that trivial.

All calls to `unicam_reset_format()` expect `node->fmt ` and `node->v_fmt.fmt.pix.pixelformat` to be set (the former is needed for `unicam_calc_format_size_bpl()` and the latter never gets set in the function.

In order to replace the exising code with `unicam_reset_format()` I am going to have to do a bit more refactoring, and replace the exising calls to it.

6by9: OK, leave it as it is in this patch then. There must be a better solution, but let's not obsess over it now.

6by9:
LGTM

drivers/media/platform/bcm2835/bcm2835-unicam.c
naushir: Agreed.  I think a refactor of `unicam_reset_format()` would help, but leave it for another day.

naushir:
@pelwell  or @popcornmix could you merge if you are happy please?

PR #3566, opened 2020-04-22
High peripheral mode

hvenev:
The first patch just changes the `#size-cells` of `&scb`. The second patch is the one that changes addresses. Both seem to work (ethernet + usb + no errors in dmesg)

I haven't tested it, but I suspect it may be possible to do the second patch as an overlay instead.


arch/arm/boot/dts/bcm2711-rpi.dtsi
```diff
@@ -185,17 +185,15 @@
...
-		 <0x0 0x40000000  0x0 0xff800000  0x00800000>,
-		 <0x6 0x00000000  0x6 0x00000000  0x40000000>,
-		 <0x0 0x00000000  0x0 0x00000000  0xfc000000>;
-	dma-ranges = <0x0 0x00000000  0x0 0x00000000  0xfc000000>,
-		     <0x1 0x00000000  0x1 0x00000000  0x80000000>,
-		     <0x1 0x80000000  0x1 0x80000000  0x80000000>;
+	ranges = <0x0 0x7c000000  0x4 0x7c000000  0x0 0x04000000>,
+		 <0x0 0x40000000  0x4 0xc0000000  0x0 0x00800000>,
+		 <0x6 0x00000000  0x6 0x00000000  0x0 0x40000000>,
+		 <0x0 0x00000000  0x0 0x00000000  0x0 0x80000000>;
```
hvenev: Oops. I'll fix the size here.

hvenev: Actually, what is this supposed to mean? There seems to be overlap in the bus addresses in the original DTS. Should the SDRAM ranges even be here?

pelwell: Which bit are you questioning? I can't see a problem.

hvenev: The one that maps 0x0 -> 0x0. I accidentally changed the size from 0xfc000000 to 0x80000000 (instead of 0x100000000).

It looks like that range overlaps with the first two. Should it even be there? None of the nodes inside mention them, and none of the other nodes (except memory) define those ranges.

pelwell: It's probably unnecessary - delete it.

hvenev:
I added a commit that removes the unnecessary range, but I've only tested high peripheral mode.

hvenev:
The overlay compiles with warnings about missing `#size-cells` and `#address-cells`. Those need to be specified both in the node and its parent to silence the warning.

Other than that, the overlay works.

pelwell:
I was going to apologise about asking for the old syntax because my checker tools don't understand the new format yet, but in this case the old syntax allows you to get rid of the warnings. I've updated the PR with an equivalent overlay in the old style. If you approve, I'll merge the PR.

hvenev:
LGTM

pelwell:
Thanks again.

PR #3571, opened 2020-04-23
Backport of udmabuf and dma-heaps

6by9:
As title. dma-heaps can hopefully simplify vcsm-cma, and they're wanted by the Libreelec folk.

They both seem to work in so far as they create /dev/udmabuf and /dev/dma_heap/linux,cma respectively. I haven't tested them from userspace.

Need to add a commit that adds them to the defconfig.


6by9:
Rebased to top of 5.4.
@popcornmix  Does this actually work for Kodi/Libreelec? I'm about to see if it works for libcamera (to remove the dependency on vcsm-cma doing allocations).

6by9:
Hmm, having cherry-picked to my working branch, I'm getting /dev/dma_heap/system and /dev/dma_heap/reserved, but no longer /dev/dma_heap/linux,cma. It may be a build issue, but the option is enabled
```
CONFIG_DMABUF_HEAPS=y
CONFIG_DMABUF_HEAPS_SYSTEM=y
CONFIG_DMABUF_HEAPS_CMA=y
```
...
Ah, if there is a cma= entry on the command line then it gets names ```/dev/dma_heap/reserved```. If it's configured by DT only, then it's ```/dev/dma_heap/linux,cma```.
You get the messages
```
[    0.000000] Reserved memory: bypass linux,cma node, using cmdline CMA params instead
[    0.000000] OF: reserved mem: node linux,cma compatible matching fail
[    0.000000] cma: Reserved 384 MiB at 0x0000000016c00000
```
in dmesg if cma= exists.
This is going to be a pain to manage from userspace - yet another file to check for :-(

6by9:
Tested with https://github.com/6by9/userland/tree/clean having a modified user-vcsm library, and dma-heaps work.

Not tested udmabuf - it wasn't clear whether they actually help us on the Pi where most dmabuf consumers want contiguous allocations. The backport won't do any harm though.

6by9:
Question to @XECDesign.
This adds a device as /dev/dma_heap/linux,cma, however it defaults to ownership root/root, permissions 600.
What's the correct way to alter the permissions to 660 root/video or similar? I never did get udev rules to work for vcsm-cma.

XECDesign:
I'm guessing copying on of the existing rules and adjusting it in the obvious way didn't work.  I'd need to poke around with `udevadm monitor` a bit to figure and see if I can figure out the right runes.

6by9:
I never managed to get the right runes for vcsm-cma.
```udevadm monitor``` is of limited use as the events have already happened during boot, and there's no way I know of to trigger it at a later point.

**Comment by XECDesign** [comment]
I seem to recall `udevadm` having options to rescan devices or query existing devices based on their dev nodes. I'm a little rusty with all of that but I can take a look tomorrow.

**Comment by pelwell** [comment]
If that doesn't work, there is a kernel function you can instrument to log the events on the way out.

**Comment by 6by9** [comment]
Thanks both. Things to look at.

**Comment by 6by9** [comment]
Found. Thanks for the guidance
```
$ udevadm info -n /dev/dma_heap/linux,cma
P: /devices/virtual/dma_heap/linux,cma
N: dma_heap/linux,cma
L: 0
E: DEVPATH=/devices/virtual/dma_heap/linux,cma
E: DEVNAME=/dev/dma_heap/linux,cma
E: MAJOR=253
E: MINOR=0
E: SUBSYSTEM=dma_heap
```
so adding 
```
SUBSYSTEM=="dma_heap", GROUP="video", MODE="0660"
```
does the magic.

**Comment by pelwell** [comment]
Thanks for the update - that's a useful trick.

**Comment by XECDesign** [comment]
Should I add that for the next image?

**Comment by popcornmix** [comment]
@XECDesign I think it's worth adding. I believe it is working with kodi (from logs it can access /dev/dma_heap once permissions are fixed). I think buffers allocated come from here (although I'll see if I can be sure with debugger) and video plays okay.

**Comment by 6by9** [comment]
@XECDesign Yes please.

Now I know the runes, give me a few minutes and I'll see if I can get the runes right for vcsm-cma as well so that we can restrict it a little more than it currently is.

**Comment by popcornmix** [comment]
I have ``SUBSYSTEM=="vcsm-cma", GROUP="video", MODE="0660"`` in `/etc/udev/rules.d/99-com.rules`
Is that not what you want?

**Comment by 6by9** [comment]
```
pi@raspberrypi:~ $ udevadm info -n /dev/vcsm-cma 
P: /devices/virtual/misc/vcsm-cma
N: vcsm-cma
L: 0
E: DEVPATH=/devices/virtual/misc/vcsm-cma
E: DEVNAME=/dev/vcsm-cma
E: DEVMODE=0666
E: MAJOR=10
E: MINOR=60
E: SUBSYSTEM=misc
```
so that implies that SUBSYSTEM is not vcsm-cma. DRIVER might be, but that hasn't worked for me.
Trying it, your line doesn't appear to set the group to video.

The driver has a nasty hack in it to set permissions to 666 https://github.com/raspberrypi/linux/blob/rpi-5.4.y/drivers/staging/vc04_services/vc-sm-cma/vc_sm.c#L1602

**Comment by popcornmix** [comment]
Ah yes - it works from the 666 permissions not the ownership.

**Comment by 6by9** [comment]
Further magic runes (got to love search engines)
```
pi@raspberrypi:~ $ udevadm info --name=/dev/vcsm-cma --attribute-walk

Udevadm info starts with the device specified by the devpath and then
walks up the chain of parent devices. It prints for every device
found, all possible attributes in the udev rules key format.
A rule to match, can be composed by the attributes of the device
and the attributes from one single parent device.

  looking at device '/devices/virtual/misc/vcsm-cma':
    KERNEL=="vcsm-cma"
    SUBSYSTEM=="misc"
    DRIVER==""
```
So 
```
KERNEL=="vcsm-cma", GROUP="video", MODE="0660"
```
is the correct line for vcsm-cma. Can we add that one to the default rules as well please? I can then revert the kernel hack.

6by9:
Messages from LibreElec folk say that this works for them, and I've had it working with both libcamera and a hacked user-vcsm (need to push that).
Any further comments/objections, or can it be merged before popcornmix rebases the branch again?

pelwell:
@popcornmix seems happy (and he can always revert/delete if necessary) - merging.

PR #3585, opened 2020-04-29
Sensor updates

6by9:
Updates OV5647 fora load of required controls.
Minor fix to imx219.
Adds [G|S]_SELECTION support to Unicam.

### DIFF
--- a/drivers/media/i2c/ov5647.c
+++ b/drivers/media/i2c/ov5647.c
@@ -1146,8 +1146,10 @@ static int ov5647_set_fmt(struct v4l2_subdev *sd,
 	else
 		mode = mode_8bit;
 
-	if (!mode)
+	if (!mode) {
+		mutex_unlock(&state->lock);
 		return -EINVAL;
+	}
 
 	*fmt = mode->format;
 	if (format->which == V4L2_SUBDEV_FORMAT_TRY) {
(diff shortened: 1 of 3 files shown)


6by9:
@naushir Need CID_VBLANK still.

6by9:
OK, I think this may be near complete. Please test and see what you get.

6by9:
Take it back. The exposure control range isn't being set to that defined by VBLANK - sorry but the orriginal patch adding exposure is just plain wrong.
Nearly there though.

drivers/media/i2c/ov5647.c
```diff
@@ -613,65 +618,124 @@ static struct ov5647_mode supported_modes_10bit[] = {
...
 			.height = 1944
 		},
-		ov5647_2592x1944_10bit,
-		ARRAY_SIZE(ov5647_2592x1944_10bit)
+		.crop = {
+			.left = 0,
+			.top = 0,
+			.width = 2624,
+			.height = 1956,
+		},
```
naushir, line 639: Should this not be 2592x1944?

6by9: This was the mess I got into over native array vs active vs TIMING_x_OUTPUT_SIZE vs TIMING_x_ADDR_END-TIMING_x_ADDR_START vs what the heck was wanted by userspace.

Yes, 2592x1944 is probably correct.

6by9:
Updated. I think I'm done. Flips can wait for now.
@naushir If you can review and test, then say if you're happy.

naushir:
LGTM

davidplowman:
I've been testing this with my ov5647 and libcamera - looks good to me too.

naushir:
@pelwell @popcornmix, this is ready to be merged if you are happy with it.

PR #3583, opened 2020-04-29
media: i2c: imx219: Implement get_selection

naushir:
Implement the get_selection pad operation for the IMX219 sensor driver.
The supported targets report the sensor's native size, the crop default
rectangle and the crop rectangle.

Reviewed-by: Laurent Pinchart <laurent.pinchart@ideasonboard.com>
Signed-off-by: Jacopo Mondi <jacopo@jmondi.org>

PR #3587, opened 2020-04-30
media: bcm2835-unicam: Fix file open/close handling.

naushir:
When closing the video node, _vb2_fop_release() is called and handles
stopping the stream. There's thus no need to stop unicam manually.
Remove the call to unicam_stop_streaming().

Additionally, when opening the video node, we were skipping the
reference counting if the file was already open, causing us to stall
when starting to stream.

### DIFF
--- a/drivers/media/platform/bcm2835/bcm2835-unicam.c
+++ b/drivers/media/platform/bcm2835/bcm2835-unicam.c
@@ -2137,13 +2137,13 @@ static int unicam_open(struct file *file)
 		goto unlock;
 	}
 
-	if (!v4l2_fh_is_singular_file(file))
-		goto unlock;
-
-	ret = v4l2_subdev_call(dev->sensor, core, s_power, 1);
-	if (ret < 0 && ret != -ENOIOCTLCMD) {
-		v4l2_fh_release(file);
-		goto unlock;
+	if (v4l2_fh_is_singular_file(file)) {
+		/* Only the first open instance must power the subdev. */
+		ret = v4l2_subdev_call(dev->sensor, core, s_power, 1);
+		if (ret < 0 && ret != -ENOIOCTLCMD) {
+			v4l2_fh_release(file);
+			goto unlock;
+		}
 	}
 
 	node->open++;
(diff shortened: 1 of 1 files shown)


naushir:
@6by9

6by9:
The unicam_release removing stop_streaming is already on #3585

I agree over the open change too - I'll pull it into #3585

drivers/media/platform/bcm2835/bcm2835-unicam.c
```diff
@@ -2137,13 +2137,13 @@ static int unicam_open(struct file *file)
...
-	if (ret < 0 && ret != -ENOIOCTLCMD) {
-		v4l2_fh_release(file);
-		goto unlock;
+	if (v4l2_fh_is_singular_file(file)) {
+		/* Only the first open instance must power the subdev. */
+		ret = v4l2_subdev_call(dev->sensor, core, s_power, 1);
+		if (ret < 0 && ret != -ENOIOCTLCMD) {
+			v4l2_fh_release(file);
+			goto unlock;
+		}
```
6by9, line 2146: I actually disagree with doing it this way. 
The v4l2_fh_open call is the one that makes the node open or not. s_power is then a secondary operation. Just move the node->open++ to before the v4l2_fh_is_singular_file check.

6by9: And I'm now wondering if there is a better way to use the fh_list within struct video_device to determine if there are open handles or not in unicam_all_nodes_streaming, rather than adding our own reference counting. One for later.

naushir: But if we fail to power the subdev, do we leave the node open?

PS, should we abandon this PR and use #3583?

6by9: It needs the node->open--; that we removed under review ;-)

#3583 is merged (imx219 g_selection). I think you meant #3585, and yes.

naushir:
Ok, closing this and continue under #3585

PR #2103, opened 2017-07-07
overlays: Add gpio-shutdown overlay

matthijskooijman:
This overlay facilitates the addition of a powerbutton by converting
GPIO edges into KEY_POWER keypresses, which can be handled by
systemd-logind to shut down the system.

The overlay was separately tested on:

 - Raspberry Pi Zero W running Linux telegraaf 4.9.28+ #998 Mon May 15 16:50:35 BST 2017 armv6l GNU/Linux
 - Raspberry Pi B+ running Linux raspberrypi 4.4.50+ #970 Mon Feb 20 19:12:50 GMT 2017 armv6l GNU/Linux
 - Rasbperry Pi B running Linux raspberrypi 4.4.34+ #930 Wed Nov 23 15:12:30 GMT 2016 armv6l GNU/Linux

I tested manually using `dtc -@ -I dts -O dtb -o gpio-shutdown.dtbo gpio-shutdown-overlay.dts`. I did not test compilation inside the kernel tree, for I have not got the environment set up for that.

For some more background, see the comments in the overlay source file and this blogpost: http://www.stderr.nl/Blog/Hardware/RaspberryPi/PowerButton.html

PR #3589, opened 2020-04-30
overlays: gpio-shutdown: Add information for Raspberry PI 1 rev 1

pali:
Older Raspberry PI 1 rev 1 uses GPIO1 for power-up instead of GPIO3.

See discussion: https://github.com/raspberrypi/linux/pull/2103#issuecomment-502347710
CC: @matthijskooijman

### DIFF
--- a/arch/arm/boot/dts/overlays/README
+++ b/arch/arm/boot/dts/overlays/README
@@ -838,6 +838,7 @@ Params: gpiopin                 GPIO for signalling (default 26)
 Name:   gpio-shutdown
 Info:   Initiates a shutdown when GPIO pin changes. The given GPIO pin
         is configured as an input key that generates KEY_POWER events.
+
         This event is handled by systemd-logind by initiating a
         shutdown. Systemd versions older than 225 need an udev rule
         enable listening to the input device:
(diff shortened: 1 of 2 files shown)


matthijskooijman (approved):
Thanks for picking this up. Changes look perfect to me.

arch/arm/boot/dts/overlays/README
```diff
@@ -874,9 +874,14 @@ Info:   Initiates a shutdown when GPIO pin changes. The given GPIO pin
         can be powered up again by driving GPIO3 low. The default
         configuration uses GPIO3 with a pullup, so if you connect a
         button between GPIO3 and GND (pin 5 and 6 on the 40-pin header),
-        you get a shutdown and power-up button.
+        you get a shutdown and power-up button. Please note that older
+        Raspberry PI 1 rev 1 uses GPIO1 instead of GPIO3.
```
pelwell: We prefer to write it "Raspberry Pi", but the device in question was the "Model B rev 1" (which is long enough on its own without the "Raspberry Pi" at the start), so please replace all occurrences of "Raspberry PI 1" with "Model B".

pelwell (commented):
Nice job on this additional documentation. I have one small request on device naming, but otherwise it looks good.

ghost:
This wording implies that only older Pi B rev 1 boards use GPIO1, while newer Pi B rev 1 boards use GPIO3. Is that correct, or do all Pi B rev 1 boards use GPIO1? If the latter, please remove 'older' from the description.

pelwell:
I didn't read it that way, but you are right that the word older is at least unnecessary.

pali:
To make it clear, see https://raspberrypi.stackexchange.com/a/87594

Could suggest how to rewrite documentation to be unambiguous?

pelwell:
If you remove every occurrence of the word "older" from your additions it should be clear enough and unambiguous.

pali:
Done!

pelwell:
I can live with that.

PR #3590, opened 2020-05-01
staging: vc04_services: ISP: Add enum_framesizes ioctl

naushir:
This is used to enumerate available frame sizes on all nodes
apart from statistics output.

### DIFF
--- a/drivers/staging/vc04_services/bcm2835-isp/bcm2835-v4l2-isp.c
+++ b/drivers/staging/vc04_services/bcm2835-isp/bcm2835-v4l2-isp.c
@@ -227,8 +227,9 @@ static const struct bcm2835_isp_fmt *get_fmt(u32 mmal_fmt)
 	return NULL;
 }
 
-static struct bcm2835_isp_fmt *find_format(struct v4l2_format *f,
-					   struct bcm2835_isp_node *node)
+static const
+struct bcm2835_isp_fmt *find_format_by_fourcc(unsigned int fourcc,
+					      struct bcm2835_isp_node *node)
 {
 	struct bcm2835_isp_fmt_list *fmts = &node->supported_fmts;
 	struct bcm2835_isp_fmt *fmt;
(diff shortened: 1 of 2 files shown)


drivers/staging/vc04_services/bcm2835-isp/bcm2835-v4l2-isp.c
```diff
@@ -1047,6 +1092,8 @@ static const struct v4l2_ioctl_ops bcm2835_isp_node_ioctl_ops = {
 	.vidioc_enum_fmt_vid_out	= bcm2835_isp_node_enum_fmt,
 	.vidioc_enum_fmt_meta_cap	= bcm2835_isp_node_enum_fmt,
```
pelwell: Does it need this blank line? It groups nicely with the previous lines.

naushir: Yes will do.

naushir:
@6by9  - still marked as WIP until I actually build and test!

drivers/staging/vc04_services/bcm2835-isp/bcm2835-v4l2-isp.c
```diff
@@ -229,16 +229,22 @@ static const struct bcm2835_isp_fmt *get_fmt(u32 mmal_fmt)
...
 	unsigned int i;
 
 	for (i = 0; i < fmts->num_entries; i++) {
 		fmt = &fmts->list[i];
-		if (fmt->fourcc == (node_is_stats(node) ?
-					    f->fmt.meta.dataformat :
-					    f->fmt.pix.pixelformat))
+		if (fmt->fourcc == node_is_stats(node) ?
+				   f->fmt.meta.dataformat :
+				   f->fmt.pix.pixelformat)
```
6by9: Comment from the last version:

Haven't we lost the switch between f->fmt.meta.dataformat and f->fmt.pix.pixelformat totally?

I was expecting find_format to be
```return find_format_by_pix(node_is_stats(node) ? f->fmt.meta.dataformat : f->fmt.pix.pixelformat, node);```

Minor nit too that it isn't necessarily a pixelformat in here, it could be a dataformat. ```fourcc``` instead?
I haven't looked at how f->fourcc compares to these two parameters.
End quote.

This version you appear to be referencing f->fmt when f doesn't exist in find_format_by_pix

Comment (naushir) [reply drivers/staging/vc04_services/bcm2835-isp/bcm2835-v4l2-isp.c]: Yeah sorry.  Got way too much going on right now...
Fixed.

**Comment by 6by9** [comment]
That looks better. I'm happy when you are.

**Comment by naushir** [comment]
Ok, that works for me.  v4l2-compliance tests all pass.  LGTM.

@pelwell , if you are happy with the change, please hit the magic button.

**Comment by 6by9** [comment]
I've just noticed a compile warning. It should be 
```
static *const* struct bcm2835_isp_fmt *find_format(struct v4l2_format *f,
					   struct bcm2835_isp_node *node)
```
to avoid a warning over discarding the const in the return from find_format_by_fourcc.
I'll create a followup.

naushir:
Oops, sorry I missed that.

Thanks for fixing!

PR #3591, opened 2020-05-01
Image sensor cherry-picks

6by9:
As discussed. Slightly reordered cf rpi-5.4.y as I did one driver at a time.
(Still compile testing at the moment though)


6by9:
Added two patches to deal with the const'ness of struct bcm2835_isp_fmt as they come from a read only array. I'll create a 5.4 PR for those in a sec.

6by9:
@naushir 5.6 version of the const bcm2835_isp_fmt patches, plus the sensor driver cherry-picks.

PR #3592, opened 2020-05-01
bcm2835_isp fixes for const'ness

6by9:


6by9:
@naushir

drivers/staging/vc04_services/bcm2835-isp/bcm2835-v4l2-isp.c
```diff
@@ -1154,47 +1154,12 @@ static int bcm2835_isp_get_supported_fmts(struct bcm2835_isp_node *node)
 		const struct bcm2835_isp_fmt *fmt = get_fmt(fourccs[i]);
 
 		if (fmt) {
-			list[j] = *fmt;
+			list[j] = fmt;
```
pelwell, line 1159: Is this change covered by the commit message? Doesn't it belong more with the other commit?

drivers/staging/vc04_services/bcm2835-isp/bcm2835-v4l2-isp.c
```diff
@@ -232,20 +232,21 @@ struct bcm2835_isp_fmt *find_format_by_fourcc(unsigned int fourcc,
 					      struct bcm2835_isp_node *node)
 {
 	struct bcm2835_isp_fmt_list *fmts = &node->supported_fmts;
-	struct bcm2835_isp_fmt *fmt;
+	const struct bcm2835_isp_fmt *fmt;
 	unsigned int i;
 
 	for (i = 0; i < fmts->num_entries; i++) {
-		fmt = &fmts->list[i];
+		fmt = fmts->list[i];
```
pelwell, line 239: This change isn't obvious from the commit message.

naushir:
Apart from the above comments, I see nothing wrong.  LGTM

drivers/staging/vc04_services/bcm2835-isp/bcm2835-v4l2-isp.c
6by9: :-(
git add -i split it not quite where I thought. Yes it's part of the other commit.

drivers/staging/vc04_services/bcm2835-isp/bcm2835-v4l2-isp.c
6by9: "The list of supported formats was taking a copy of entries from this table.
This is unnecessary, therefore allocate an array of pointers instead of
an array of entries."

Seeing as it was all messing with struct bcm2835_isp_fmt I wasn't too worried over splitting it into multiple commits.

naushir:
LGTM

PR #3600, opened 2020-05-05
RFC: Test of zswap deferred initialisation

pelwell:
This PR is the result of some experimentation to try and defer the cost
of enabling the ZSWAP configuration option until zswap is actually
enabled. There is no guarantee it will get merged, but some feedback
that zswap still works and that the overhead is acceptable would make
it more likely.

See also the discussion at https://github.com/raspberrypi/linux/pull/3432.

[ From the first commit message ]
```
Enabling zswap support in the kernel configuration costs about 1.5MB
of RAM, even when zswap is not enabled at runtime. This cost can be
reduced significantly by deferring initialisation (including pool
creation) until the "enabled" parameter is set to true. There is a
small cost to this in that some __init functions have to remain in
memory after the init phase, just in case they are needed later,
but the total size is negligible.
```


Syonyk:
My first thought about maintaining a separate difference from upstream is that if this works as advertised, it should be pushed upstream.  It's enabled in most stock x86 kernels, so late initialization if not used to save 1.5MB of memory seems entirely sane to me.

pelwell:
Yes, of course - I think that goes without saying - but this won't go anywhere without some feedback.

popcornmix:
If some zswap users can confirm this behaves as expected then it looks good to me.
I'd hope upstream will be receptive as 1.5MB for a configured but unused feature seems extremely wasteful.

Syonyk:
I'll see if I can get this tested out in the next day or two - just depends on weather, lots going on outside if it's tolerable out there.

swetoast:
compiled gonna run test until tomorrow

swetoast:
Works as intended i would say it ok to merge

pelwell:
Thanks. Did anybody else try it?

pelwell:
@Syonyk Did you find time since the chill hit to try this patch?

Syonyk:
Sorry, took a bit to get time on a bad weather day.  We've had people doing a bunch of property work around here.

It works, in that you can boot without zswap enabled, enable it once the system is up, and have it work, but I've managed to trip it up such that I can't enable it by using the kernel command line flags.

Adding ```zswap.enabled=1``` to /boot/cmdline.txt will lead to a failure in initialization on boot, which is fine, but you can't then enable it later.

In early boot:
```
[    0.792727] zswap: default zpool zbud not available
[    0.792900] zswap: pool creation failed
```

Ok, no module available, that's fine.

But if you try to enable it later, in ```/sys/module/zswap/parameters```, you get the following failure:

```
root@raspberrypi:/sys/module/zswap/parameters# echo Y > enabled
bash: echo: write error: No such device
```

And the corresponding dmesg output:

```
[  401.984101] zswap: can't enable, initialization failed
```

If you want to take a stab at fixing that, I've got a version of your tree up to date with kernel objects in it.  I don't mind that the command line initialization fails, just that it then fails into a state where it can't be later loaded.

EDIT: To be clear, I don't know if this is an issue with your patches, or if the upstream kernel behaves like this as well.  I'll get the Pi set up for remote access and mess with it a bit.

zbud.o isn't even 10kb... may be worth just building that in.

Syonyk:
Stock kernel behaves exactly the same with zbud as a module, so no change there.  I suppose I ought to fix that...

In any case, I'm fine with this patch as submitted.  I think it would be worth the space to include zbud in the kernel image, for the massive increase in desktop usability this sort of thing offers, but as long as it's somehow usable in the stock Raspbian kernel, I'm happy!

pelwell:
Failure to load zbud at init time shouldn't preclude one from trying again later. I think it can work better, I just didn't get round to it yet.

Syonyk:
I agree - I was mostly just verifying that your changes didn't impact that case by testing against the reference 4.19 code.  It's broken upstream.

I would still mildly prefer to see zbud built as part of the kernel, such that zswap can initialize from the command line.  It's a 10kb difference in the kernel, give or take, and you save that on module space.

PR #3601, opened 2020-05-05
vc4_hdmi: Add CEC support for 2711

popcornmix:


popcornmix:
Ping @6by9 
A few open questions:
-  is adding variant functions for each difference in behaviour best, or would conditional code based on an `is_vc5` flag be more concise?
- Shall we strip out `cec_available` now it's always true?
- Hardware retry is mentioned in the docs. I can't be sure it's doing useful stuff but it doesn't appear to break things. It could be dropped with no loss of functionality (but potentially less efficient).

drivers/gpu/drm/vc4/vc4_hdmi.c
```diff
@@ -1264,8 +1296,13 @@ static void vc4_cec_read_msg(struct vc4_hdmi *vc4_hdmi, u32 cntrl1)
...
 	msg->len = 1 + ((cntrl1 & VC4_HDMI_CEC_REC_WRD_CNT_MASK) >>
 					VC4_HDMI_CEC_REC_WRD_CNT_SHIFT);
+
+	if (msg->len > 16) {
+		DRM_ERROR("Attempting to read too much data (%d)\n", msg->len);
+		return;
+	}
 	for (i = 0; i < msg->len; i += 4) {
-		u32 val = HDMI_READ(HDMI_CEC_RX_DATA_1 + i);
+		u32 val = HDMI_READ(HDMI_CEC_RX_DATA_1 + (i>>2));
```
6by9, line 1273: Is this correct for Pi3 as well? It appears to be unconditional on revision.

drivers/gpu/drm/vc4/vc4_hdmi.c
```diff
@@ -1751,6 +1803,7 @@ static int vc4_hdmi_dev_remove(struct platform_device *pdev)
 
 static const struct vc4_hdmi_variant bcm2835_variant = {
 	.max_pixel_clock	= 162000000,
+	.cec_input_clock	= 163682864,
```
6by9: To nit pick, it'd be nice to have this as a define along the lines of VC4_HDMI_HSM_CLOCK_RATE, as the same value is used in vc4_hdmi_calc_hsm_clock.

drivers/gpu/drm/vc4/vc4_hdmi.c
```diff
@@ -1791,13 +1850,18 @@ static const struct vc4_hdmi_variant bcm2711_hdmi0_variant = {
 	.phy_rng_disable	= vc5_hdmi_phy_rng_disable,
 	.get_hsm_clock		= vc5_hdmi_get_hsm_clock,
 	.calc_hsm_clock		= vc5_hdmi_calc_hsm_clock,
+	.set_cec_retries	= vc5_hdmi_set_cec_retries,
+	.get_cec_retries	= vc5_hdmi_get_cec_retries,
 	.channel_map		= vc5_hdmi_channel_map,
+
+	.cec_mask = VC5_HDMI0_CPU_CEC_RX | VC5_HDMI0_CPU_CEC_TX | VC5_HDMI0_CPU_HOTPLUG_CONN | VC5_HDMI0_CPU_HOTPLUG_REM,
```
6by9: Hotplug bits as well? Why?

6by9:
>    is adding variant functions for each difference in behaviour best, or would conditional code based on an is_vc5 flag be more concise?

I'll leave that as one for @mripard to resolve when he rolls this into the upstream patchset. If we can get it working, then he can worry about the preferred way of expressing it.

>    Shall we strip out cec_available now it's always true?

Yes

>    Hardware retry is mentioned in the docs. I can't be sure it's doing useful stuff but it doesn't appear to break things. It could be dropped with no loss of functionality (but potentially less efficient).

Seeing as the CEC framework also appears to have a retry mechanism that we can plumb it into, then why not.

Maxime had mentioned that there is already a driver for the irq-brcmstb-l2 controller at https://github.com/raspberrypi/linux/blob/rpi-5.4.y/drivers/irqchip/irq-brcmstb-l2.c, but he hadn't got it to work when trying to make the HDMI I2C blocks work (they still currently poll).
It'd be nice to use that for the CEC interrupt controller instead of reimplementing the same controller code, but that can be a cleanup. The Pi3 version appears to be the same interrupt controller arrangement too, so it should be common to both platforms.

drivers/gpu/drm/vc4/vc4_hdmi.c
popcornmix: Yes it is a bug on Pi3. Any long messages would trample other registers (and we are getting longer messages in log).

popcornmix:
Updated based on comments and removed WIP.

drivers/gpu/drm/vc4/vc4_hdmi.c
```diff
@@ -1360,9 +1365,12 @@ static int vc4_hdmi_cec_adap_transmit(struct cec_adapter *adap, u8 attempts,
 	struct vc4_hdmi *vc4_hdmi = cec_get_drvdata(adap);
 	u32 val;
 	unsigned int i;
-
```
pelwell, line 1363: This blank line is required by kernel coding standards.

drivers/gpu/drm/vc4/vc4_hdmi_regs.h
```diff
@@ -33,11 +33,12 @@ enum vc4_hdmi_field {
 	HDMI_CEC_CNTRL_3,
```
pelwell, line 34: The commit message is in too long a line (72 char limit), and includes two typos - either go with "Fixes an incorrect register's address" or "Fixes an incorrect register address".

6by9:
LGTM

drivers/gpu/drm/vc4/vc4_hdmi.c
```diff
@@ -1580,6 +1580,14 @@ static int vc5_hdmi_init_resources(struct vc4_hdmi *vc4_hdmi)
 	if (IS_ERR(vc4_hdmi->dvp_regs))
 		return PTR_ERR(vc4_hdmi->dvp_regs);
 
+	res = platform_get_resource_byname(pdev, IORESOURCE_MEM, "intr2");
+	if (!res)
+		return -ENODEV;
+
+	vc4_hdmi->intr2_regs = devm_ioremap(dev, res->start, resource_size(res));
```
pelwell, line 1588: dev_ioremap_resource? One ought to be able to combine the resource lookup and the ioremap, but I can't find such a convenience function.

arch/arm/configs/bcm2711_defconfig
```diff
@@ -890,6 +890,7 @@ CONFIG_DRM_PANEL_SIMPLE=m
 CONFIG_DRM_PANEL_RASPBERRYPI_TOUCHSCREEN=m
 CONFIG_DRM_V3D=m
 CONFIG_DRM_VC4=m
+CONFIG_DRM_VC4_HDMI_CEC=y
```
pelwell, line 893: Is this not also applicable to 64-bit builds?

pelwell (commented):
Mainly niggles, apart from the arm64 support.

drivers/gpu/drm/vc4/vc4_hdmi.c
6by9: I don't believe there is, and it matches all the other mappings from Maxime's code.

6by9:
As Phil's being picky (rightly so), there's also the matter of the author and ```Signed-off-by``` being a pseudonym.

pelwell:
I was going to say that we shouldn't unmask @popcornmix, but there is prior evidence in the logs.

arch/arm/configs/bcm2711_defconfig
popcornmix: Untested but no reason I can think of why 64-bit build would fail.

pelwell: Throw it in there, then.

popcornmix:
Updated.

drivers/gpu/drm/vc4/vc4_hdmi.c
pelwell: You can use devm_ioremap_resource here:
```
	vc4_hdmi->intr2_regs = devm_ioremap(dev, res);
```
I think it was the combined find and remap that @6by9 was agreeing doesn't exists, but for consistency I'll leave this as is.

PR #3604, opened 2020-05-07
Use the upstream cpufreq driver on non-BCM2835 Pis

pelwell:
Inspired by the comment on https://github.com/raspberrypi/linux/pull/3554.


pelwell:
Any objections, @popcornmix?

naushir:
FYI, been running with this change briefly and saw no problems with it.

PR #3607, opened 2020-05-11
overlays: Move "fixed-clock" nodes to the root

pelwell:
Apart from some special cases, device objects are only created for
nodes if they are children of a bus or the root node. "fixed-clock"
is one of the exceptions that will be instantiated wherever it is
found, but only during kernel initialisation - ruling out loading the
overlay at runtime.

Move most of the affected clocks to be children of the root, only
leaving those in overlays that could be multiply instantiated, to avoid
a potential name clash.

See: https://github.com/raspberrypi/linux/issues/3602


6by9 (commented):
Looks fine to me.

PR #3608, opened 2020-05-13
media: bcm2835-unicam: Always service interrupts

6by9:
From when bringing up the driver, there was a check in the isr
to ignore interrupts (claiming them handled) should the driver
not be streaming.

The VPU now will not register a camera driver if it finds a
CSI2 node enabled in device tree, therefore this flawed check is
redundant.

https://github.com/raspberrypi/linux/issues/3602


6by9:
@naushir This change was brought up on the linux-media review as well, but it is actually fixing a problem.

PR #3611, opened 2020-05-14
drm/vc4: Fix VIC usage with Broadcast RGB

6by9:
Adding the Broadcast RGB range selection broke the VIC
field of the AVI infoframes on HDMI, zeroing them for all
modes on an HDMI monitor.

Correct this so that it is only zeroed if the range is
contrary to the standard range of the mode.

### DIFF
--- a/drivers/gpu/drm/vc4/vc4_firmware_kms.c
+++ b/drivers/gpu/drm/vc4/vc4_firmware_kms.c
@@ -936,18 +936,13 @@ static void vc4_crtc_mode_set_nofb(struct drm_crtc *crtc)
 		break;
 	}
 
+	mb.timings.video_id_code = frame.avi.video_code;
+
 	if (!vc4_encoder->hdmi_monitor) {
 		mb.timings.flags |= TIMINGS_FLAGS_DVI;
-		mb.timings.video_id_code = frame.avi.video_code;
 	} else {
 		struct vc4_fkms_connector_state *conn_state =
 			to_vc4_fkms_connector_state(vc4_crtc->connector->state);
-
-		/* Do not provide a VIC as the HDMI spec requires that we do not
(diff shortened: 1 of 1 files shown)


drivers/gpu/drm/vc4/vc4_firmware_kms.c
```diff
@@ -958,6 +953,16 @@ static void vc4_crtc_mode_set_nofb(struct drm_crtc *crtc)
...
 				mb.timings.flags |= TIMINGS_FLAGS_RGB_LIMITED;
+
+			/* If not using the default range, then do not provide
+			 * a VIC as the HDMI spec requires that we do not
+			 * signal the opposite of the defined range in the AVI
+			 * infoframe.
+			 */
+			if (!!(mb.timings.flags & TIMINGS_FLAGS_RGB_LIMITED) ^
+			    (drm_default_rgb_quant_range(mode) ==
+					HDMI_QUANTIZATION_RANGE_LIMITED))
```
JamesH65, line 964: This logic is a bit unclear.

pelwell: If the `^` were a `!=` would that be clearer?

6by9: Is that a request for a change, or an explanation for JamesH65?

6by9: To change, or not to change? That is the question.

pelwell: `^` has no benefit over `!=` here, so change it.

6by9: Done.

PR #3617, opened 2020-05-15
arm: dt: For 2711 move vc4 gpu node under soc to get correct dma-ranges

6by9:
The vc4: gpu / "brcm,bcm2711-vc5" DT node was under the root node,
therefore didn't get the soc dma-ranges to choose the uncached
cache alias. Buffers were therefore allocated and passed to the
HVS in both full and fake KMS modes with incorrect addresses.

Move the node under /soc to get the correct addresses.

### DIFF
--- a/arch/arm/boot/dts/bcm2711.dtsi
+++ b/arch/arm/boot/dts/bcm2711.dtsi
@@ -12,11 +12,6 @@
 
 	interrupt-parent = <&gicv2>;
 
-	vc4: gpu {
-		compatible = "brcm,bcm2711-vc5";
-		status = "disabled";
-	};
-
 	clk_108MHz: clk-108M {
 		#clock-cells = <0>;
 		compatible = "fixed-clock";
@@ -385,6 +380,12 @@
 			clock-frequency = <97500>;
 			status = "disabled";
 		};
+
+		vc4: gpu {
+			compatible = "brcm,bcm2711-vc5";
+			status = "disabled";
+		};
+
 	};
 
 	arm-pmu {


pelwell:
Are you certain that this won't adversely affect any other aspect of the driver? Does it only ever need DMA address for CMA memory? (I'm asking because I don't know).

6by9:
AIUI, we now have ```v3d``` dealing with all the 3D side.
```vc4``` is only dealing with the HVS/PVs/encoders/TXP. Those can only deal with the VideoCore ranged addresses.

I believe that we can remap the buffer again, but I don't believe it is necessary.

I have pinged Maxime to confirm the behaviour.

pelwell:
I'm happy to merge, unless you want to wait for confirmation.

6by9:
Give it until Monday for a response.
Other than the assert within the firmware I haven't seen any bad behaviour due to this even though the HVS can't get to the cached alias. I'm assuming something is remapping the alias within the SDRAM/cache controller so that it doesn't actually fail.
Those of use that the assert is tripping up can grab the patch from this PR.

6by9:
I've had a response from Maxime. The "better" way forward is to have the base drm driver find the HVS node and adopt the dma ranges from there using ```of_dma_configure```.
Allwinner chips have to do a similar thing, although they have a dedicated memory controller node for the display engine accesses, therefore they find that - https://elixir.bootlin.com/linux/v5.7-rc6/source/drivers/gpu/drm/sun4i/sun4i_backend.c#L797

Closing to rework.

PR #3620, opened 2020-05-19
Switch to snd_soc_dai_set_bclk_ratio

j-schambacher:
Replaces obsolete function snd_soc_dai_set_tdm_slot.
Small style fix.


pelwell:
Do you no longer need to say the ratio for codec_dais[1]?

j-schambacher:
No, that's no more necessary as the ADC (=codec_dais[1|) runs always in slave mode and the HW adapts automatically to what is set by DAC/system.

pelwell:
That's the kind of thing to put in a commit message.

PR #3622, opened 2020-05-19
media: bcm2835-unicam: Retain packing information on G_FMT

6by9:
The change to retrieve the pixel format always on g_fmt didn't
check whether the native or unpacked version of the format
had been requested, and always returned the packed one.
Correct this so that the packing setting is retained whereever
possible.

Fixes "9d59e89 media: bcm2835-unicam: Re-fetch mbus code from subdev
on a g_fmt call"


6by9:
@naushir

naushir:
LGTM.

As we said for the offending commit, this really should be rolled into unicam_reset_format at some point.

PR #3629, opened 2020-05-20
Clean up the VCHIQ 2711 DMA support

pelwell:
Following a suggestion from @vianpl, these commits tidy up the way the 2711-specific VCHIQ code obtains DMA addresses, which removes the need to move the DT node and generally looks better.


pelwell:
@pinchartl @6by9 A version of this should probably go in the next patch set.

popcornmix:
Looks nicer.

PR #3630, opened 2020-05-21
media: i2c: imx477: Return correct result on sensor id verification

naushir:
The test should return -EIO if the register read id does not match
the expected sensor id.


naushir:
@6by9

6by9 (approved):
imx219 does 
```
if(val != CHIP_ID) {
   return -EIO;
}
return 0;
```
I seem to recall that was raised upstream in review, but it's not a big deal.

naushir:
Will follow the above to avoid churn later on.

Issue #1221, opened 2015-12-08
No Multicore Scheduler support?


robingroppe:
Just checked the kernel config again and realized that no Multicore Scheduler Support is in use.
Is there a specific reason for this?

P33M:
Are you talking about CONFIG_SCHED_SMT or CONFIG_SCHED_MC?

```
pi@raspberrypi:~$ zgrep SCHED /proc/config.gz
CONFIG_GENERIC_SCHED_CLOCK=y
CONFIG_CGROUP_SCHED=y
CONFIG_FAIR_GROUP_SCHED=y
# CONFIG_RT_GROUP_SCHED is not set
CONFIG_SCHED_AUTOGROUP=y
CONFIG_IOSCHED_NOOP=y
CONFIG_IOSCHED_DEADLINE=y
CONFIG_IOSCHED_CFQ=y
CONFIG_CFQ_GROUP_IOSCHED=y
CONFIG_DEFAULT_IOSCHED="cfq"
# CONFIG_SCHED_MC is not set
# CONFIG_SCHED_SMT is not set
CONFIG_SCHED_HRTICK=y
CONFIG_NET_SCHED=y
CONFIG_SCHED_DEBUG=y
# CONFIG_SCHEDSTATS is not set
# CONFIG_SCHED_STACK_END_CHECK is not set
CONFIG_SCHED_TRACER=y
```

popcornmix:
It's CONFIG_SCHED_MC. The help text says:

> Multi-core scheduler support improves the CPU scheduler's decision making when dealing with multi-core CPU chips at a cost of slightly increased overhead in some places. If unsure say N here.

Someone sees a 10% reduction in performance here:
https://www.redhat.com/archives/dm-devel/2010-August/msg00164.html

[edit: although there may have been other reasons for that report]

So, again I think need careful analysis before enabling.

robingroppe:
The question would be if it is in use at all when there is only one cpu in the system like the Pi 1 or Zero.

P33M:
We have two different configs for 2835 and 2836 (bcmrpi_defconfig and bcm2709_defconfig).

I don't think Sched_MC will have any positive benefit. It makes the kernel aware of CPU locality (cache hierarchy), but all our A7s share an L2 cache.

https://wiki.linaro.org/WorkingGroups/PowerManagement/Archives/SchedMc

robingroppe:
Okay. Thank you for making that clear.

PR #3635, opened 2020-05-26
Raspberry Pi 64bit Kernel Support configurations

IComplainInComments:
**Multi core for RPI4**
Missing configuration for the kernel to support multi core scheduling in its configuration file.

**Desktop Timing for RPI3**
Default RPI 3 config sets ticks to 1000hz which is around a server configuration and not one that would be useful for a desktop (and honestly, I think 1000 ticks is much to high for ARM). Setting to 250hz (Default for Desktop kernel ticks).


JamesH65:
These appear to be two entirely unrelated changes, so should be in two PR's.

I suspect the tick change would be unacceptable anyway, but other would need to comment more fully.

pelwell:
I want to see an explanation of the changes and why they should be made for all users.

IComplainInComments:
> These appear to be two entirely unrelated changes, so should be in two PR's.
> 
> I suspect the tick change would be unacceptable anyway, but other would need to comment more fully.

Sorry for the two-being-one. First time doing a pull-request, and im generally new to using github. So I apologize for that.



> I want to see an explanation of the changes and why they should be made for all users.

The scheduler is used for the kernel to better use the CPU cores in a multi core system. Basically the Kernel is able to better control what-goes-where on the CPU cores, along with have better control of locks/runs/schedules/ticks/etc.

Also all other RPI Configurations has this option enable, while RPI4 did not. So its also for general consistency around the kernel builds, along with performance gains for many threaded tasks.

For reducing the ticks. Having less ticks per kernel task switch basically allows better multi-tasking and general system responsiveness. Unlike a Server which generally is left alone to do its own bidding, a desktop is more chaotic as users do many different things. As such allowing the kernel to switch tasks more often will allow better multi-tasking capabilities for the kernel.

also this is the default kernel option for all other builds for RPI variants

Titaniumtown:
This is very interesting. Should these changes also be applied to the other branches as well?

popcornmix:
CONFIG_SCHED_MC was considered many years ago (https://github.com/raspberrypi/linux/issues/1221) and decided against. @P33M thought it was unlikely to benefit a single cluster of 4 cores with shared L2 cache.

@IComplainInComments you'll need to provide some test shows a benefit from the option. Find a benchmark or workflow that shows a measurable improvement with the change.

P33M:
Things may have changed in scheduler land since then, meaning CONFIG_SCHED_MC may now enable other enhancements - or add more side-effects.

I'd be happy to change my mind if there's evidence that this config option provides a net benefit, though.

IComplainInComments:
> CONFIG_SCHED_MC was considered many years ago (#1221) and decided against. @P33M thought it was unlikely to benefit a single cluster of 4 cores with shared L2 cache.
> 
> @IComplainInComments you'll need to provide some test shows a benefit from the option. Find a benchmark or workflow that shows a measurable improvement with the change.

Searching the internet, I honestly cannot find any sort of numbers or tests to give any indication of performance increase or reduction. 

The texts I have found is multiple articles insisting on using that option so the CPU governors can consolidate threads to manage heat/power better to allow less heat to be generated --  which would indicate better overclocking potential, along with heavy loads less likely to throttle the CPU, and obviously less heat == longer life.

The next benefit from my search is that option also allows the kernel to better manage threads and Work Units on a Multicore CPU. So it allows the kernel to better utilize the multiple cores for programs, increasing processing efficiency.

Here are some of the pages I found talking about the config option (indirectly or directly).
https://developer.ibm.com/technologies/linux/tutorials/l-cpufreq-1/
https://www.geeksforgeeks.org/multiple-processor-scheduling-in-operating-system/
https://elinux.org/images/d/dc/Elc2013_Na.pdf
https://elinux.org/images/4/43/Understanding_And_Using_SMP_Multicore_Processors_Anderson.pdf

P.S. If you would like. Give me a set of tasks you would like me to benchmark. Ill swap between the current kernel and the kernel I built and post the statistics :)



> Things may have changed in scheduler land since then, meaning CONFIG_SCHED_MC may now enable other enhancements - or add more side-effects.
> 
> I'd be happy to change my mind if there's evidence that this config option provides a net benefit, though.

Please look above for context.

The Scheduler in general is very aware of what threads are being ran on the CPUs. The scheduler's goal is set from a user setting that can be modified (You can google for setting the options). The defaults goal is consolidating tasks to free up system resources in the goal of reducing power and heat for the processors. With threads consolidated, the CPU cores will be freed up for more tasks, and with the less demand for power due to the other cores not being as busy, reduces power needs, thus reducing heat, thus allowing all cores on the CPU to stay at their top frequencies during work. This also allows for better overclocking abilities, and better task management as the scheduler also utilizes priority tasking, so high-performance tasks are put in a better position. And obviously less heat == better period.

So in general, having the kernel be Multicore aware gains a lot of very nice benefits for the System's benefit of having fine control of all the cores.

P33M:
All the pages you reference that argue for multicore-aware scheduling are at least 10 years old. Absent a demonstration of a workload that gains a benefit from your proposed change, I remain unconvinced.

IComplainInComments:
> All the pages you reference that argue for multicore-aware scheduling are at least 10 years old. Absent a demonstration of a workload that gains a benefit from your proposed change, I remain unconvinced.

...Newton's work is a few hundred years old, but they're still used in physics. If people claimed benefit then, it probably does now, especially since no one is reporting regressions in its use. 

I don't know how I am supposed to drum up recent documentation when I am a person, and not a company/organization with the tools/resources to completely document a Kernel feature I did not write/contribute/design/plan/implement.

I am more than happy to demonstrate a workload for you, but we are not in Goldy Locks and the Three Bears. Tell me which workloads you would like benchmarks to see if a benefit exists, and I will be happy to conduct and measure them for you, but I am not going to run random benchmarks for you to end up telling me "This doesnt prove anything, xyz benchmark wouldnt benefit from it". Tell what you would like, and I will do it.

P33M:
You're proposing a change to our default kernel config. If you propose such a change, you should give at least one reproducible example of where it is of benefit on the target platform - i.e. a Pi.

pelwell:
I've not looked at the details of the changes caused by CONFIG_SCHED_MC=y, but I do note that it is included in arch/arm64/configs/defconfig. The commit responsible says:
```diff
commit aa644fa64c25ab2231a0fa9464892f5579d4e161
Author: Dietmar Eggemann <dietmar.eggemann@arm.com>
Date:   Mon Oct 19 17:55:49 2015 +0100

    ARM64: Enable multi-core scheduler support by default
    
    Make sure that the task scheduler domain hierarchy is set-up correctly
    on systems with single or multi-cluster topology.
    
    Signed-off-by: Dietmar Eggemann <dietmar.eggemann@arm.com>
    Acked-by: Punit Agrawal <punit.agrawal@arm.com>
    Signed-off-by: Catalin Marinas <catalin.marinas@arm.com>

diff --git a/arch/arm64/configs/defconfig b/arch/arm64/configs/defconfig
index c8ccda16e079..5f760347aee2 100644
--- a/arch/arm64/configs/defconfig
+++ b/arch/arm64/configs/defconfig
@@ -51,6 +51,7 @@ CONFIG_PCI=y
 CONFIG_PCI_MSI=y
 CONFIG_PCI_XGENE=y
 CONFIG_SMP=y
+CONFIG_SCHED_MC=y
 CONFIG_PREEMPT=y
```

IComplainInComments:
> You're proposing a change to our default kernel config. If you propose such a change, you should give at least one reproducible example of where it is of benefit on the target platform - i.e. a Pi.

...im just proposing this changed due to every other configuration for the armv7 kernels has this option enabled; as in its default. So you can stop trying to act like you're the Raspberry Pi Linux Kernel Linus Torvalds, when you clearly aren't.



> I've not looked at the details of the changes caused by CONFIG_SCHED_MC=y, but I do note that it is included in arch/arm64/configs/defconfig. The commit responsible says:
> 
> ```diff
> commit aa644fa64c25ab2231a0fa9464892f5579d4e161
> Author: Dietmar Eggemann <dietmar.eggemann@arm.com>
> Date:   Mon Oct 19 17:55:49 2015 +0100
> 
>     ARM64: Enable multi-core scheduler support by default
>     
>     Make sure that the task scheduler domain hierarchy is set-up correctly
>     on systems with single or multi-cluster topology.
>     
>     Signed-off-by: Dietmar Eggemann <dietmar.eggemann@arm.com>
>     Acked-by: Punit Agrawal <punit.agrawal@arm.com>
>     Signed-off-by: Catalin Marinas <catalin.marinas@arm.com>
> 
> diff --git a/arch/arm64/configs/defconfig b/arch/arm64/configs/defconfig
> index c8ccda16e079..5f760347aee2 100644
> --- a/arch/arm64/configs/defconfig
> +++ b/arch/arm64/configs/defconfig
> @@ -51,6 +51,7 @@ CONFIG_PCI=y
>  CONFIG_PCI_MSI=y
>  CONFIG_PCI_XGENE=y
>  CONFIG_SMP=y
> +CONFIG_SCHED_MC=y
>  CONFIG_PREEMPT=y
> ```

I noticed this as well, but I didn't bother mentioning it at the time of my initial post due to me assuming the defconfig is just some generic profile made by a platform maintainer for the kernel.

JamesH65:
> ...im just proposing this changed due to every other configuration for the armv7 kernels has this option enabled; as in its default. So you can stop trying to act like you're the Raspberry Pi Linux Kernel Linus Torvalds, when you clearly aren't.
> 

Hmm, when asking for changes, probably best not to offend one of the Raspberry Pi engineers who would actually be responsible for making the change.

He is one of our Torvald's btw.

IComplainInComments:
You know what, it’s not worth it.

I’m not dealing with programmers ego for a device targeted at kids.

JamesH65:
I agree, the change was really not worth it.

IComplainInComments:
> I agree, the change was really not worth it.

Yeah, why on earth would you wanna make sure that the configuration settings match the armv7 kernel settings.

Granted with all the warnings that are generated by the compiler for all the BCM drivers you probably think that's a good thing.

PR #3638, opened 2020-05-28
media: bcm2835-unicam: change minimum number of vb2_queue buffers to 1

davidplowman:
Since the unicam driver was modified to write to a dummy buffer when no
user-supplied buffer is available, it can now write to and return a
buffer even when there's only a single one. Enable this by changing the
min_buffers_needed in the vb2_queue; it will be useful for enabling
still captures without allocating more memory than absolutely necessary.


davidplowman:
@naushir @6by9

naushir:
@davidplowman, this is going to need a "Signed-off-by:" tag in the commit message.

6by9:
LGTM

PR #3640, opened 2020-05-28
snd_bcm2835: disable HDMI audio when vc4 is used

hvenev:
Things don't work too well when both the vc4 driver and the firmware
driver are trying to control the same audio output:

[  763.569406] bcm2835_audio bcm2835_audio: vchi message timeout, msg=5

Hence, when the vc4 HDMI driver is used, let it control audio. This is done
by introducing a new device tree property to the audio node.

Signed-off-by: Hristo Venev <hristo@venev.name>


hvenev:
Should I rebase the pull request?

Also, should I add the same change for the Pi <=3 overlay?

popcornmix:
@pelwell what's your preference for this?
Basically:
```
dtparam=audio=on
dtoverlay=vc4-kms-v3d (without ,audio=off)
```
are dangerous if both audio devices are used.
Would this be better handled by firmware's DT mangling? (e.g. if one is detected ensure other is disabled)

hvenev:
Personally I'd like to keep the ability to use the headphone jack at the same time as HDMI.

pelwell:
DT mangling is only useful if there is something to observe the change. If there is an existing way to disable the audio support in the KMS driver then this overlay should use it, but I don't think there is one.

An alternative approach would be for the bcm2835 driver to detect the presence of the enabled kms driver automatically, but this is much simpler. You could hide the disabling property if "audio=off" is specified, if there are situations where one might want the bcm2835 driver to control the audio.

hvenev:
> An alternative approach would be for the bcm2835 driver to detect the presence of the enabled kms driver automatically, but this is much simpler. You could hide the disabling property if "audio=off" is specified, if there are situations where one might want the bcm2835 driver to control the audio.

Yes, that would be better. However, on the Pi4 there are two toggles, one for each HDMI port. We need to somehow detect if both are disabled, and only have snd_bcm2835 handle audio then.

My overlay-fu isn't very strong. Do you think the following will work:

vc4-kms-v3d-pi4-overlay.dts
```
fragment@19 {
	target = <&audio>;
	__overlay__  {
	    brcm,disable-hdmi;
	};
};

__overrides__ {
	audio   = <0>,"!17", <0>,"=19";
	audio1   = <0>,"!18", <0>,"=19";
};
```

Then `&audio` `brcm,disable-hdmi` is not set if both `audio=off` and `audio1=off`.

Edit: I'm not sure this works. How will the overrides behave when `audio=on,audio1=off`? Or `audio1=off,audio=on`?

pelwell:
I think it's safer to add a third parameter:
```
noaudio = <0>,"=17", <0>,"=18",<0>,"!19";
```

hvenev:
Maybe `<0>,"+17", <0>,"+18", <0>,"-19"`, so that it can be used as

```
dtoverlay=vc4-kms-v3d,noaudio
```

instead of

```
dtoverlay=vc4-kms-v3d,noaudio=on
```

?

pelwell:
If no parameter value is supplied then the default value of 1/on/true is assumed, so my version works with or without `=on` - the difference is only apparent if someone tries `...,noaudio=off`. I have no strong feelings either way, so I'll let you choose. You can also include an underscore in the name for readability if you prefer.

hvenev:
I added the `noaudio` parameter to the Pi <=3 and Pi 4 overlays. I haven't actually tested the audio output itself, but `alsamixer` seems to show the right set of sound cards.

On the Pi 4, the old (current) behaviour can be configured with

`dtoverlay=vc4-kms-v3d,noaudio,audio,audio1`

This first disables vc4 HDMI audio and lets snd_bcm2835 handle it using the firmware thing, and then also tells vc4 to handle HDMI audio as well.

pelwell:
I'll give it a spin tomorrow.

pelwell:
I've added commits that make "audio=off" do the sensible thing for the non-Pi4 case, and idocument the noaudio parameter. If you're happy with these changes I'll squash and merge all three.

hvenev:
LGTM

pelwell:
Thanks!

T0MuX:
Hi, I have exactly the same issue, but none of any given solutions works for me.
I tried with just **,noaudio** I tried with the whole line **,noaudio,audio,audio1** ..
Also, I tried to change linux kernel : switched from linux-aarch64 to linux(raspberrypi4, I think it's better (now I can start a compositor which was not the case before).
Another thing to know : with linux-aarch64 kernel, when I type pacmd list-sinks I can see the both audio outputs (hdmi and analog), bit with linux-raspberrypi4 I can only see the hdmi one. I don't understand, and I don't know what I have to do at all.

In dmesg, I can see this line : **[    0.000000] Kernel command line: coherent_pool=1M 8250.nr_uarts=0 snd_bcm2835.enable_compat_alsa=0 snd_bcm2835.enable_hdmi=1 video=HDMI-A-1:720x480M@60,margin_left=4,margin_right=4,margin_top=8,margin_bottom=8 smsc95xx.macaddr=DC:A6:32:9C:F4:15 vc_mem.mem_base=0x3ec00000 vc_mem.mem_size=0x40000000  root=/dev/mmcblk0p2 rw rootwait console=ttyS0,115200 console=tty1 selinux=0 plymouth.enable=0 smsc95xx.turbo_mode=N dwc_otg.lpm_enable=0 kgdboc=ttyS0,115200 quiet**

I tried to add **snd_bcm2835.enable_compat_alsa=1 snd_bcm2835.enable_hdmi=0** to the line in the cmdline.txt file, but this didn't changed anything.

pelwell:
And with **snd_bcm2835.enable_compat_alsa=0 snd_bcm2835.enable_hdmi=0**?

PR #3645, opened 2020-05-29
media: bcm2835-isp: fix bytes per line calculations for some image fo…

davidplowman:
…rmats

The bytes per line numbers calculated by get_bytesperline was not
matching the equivalent calculation being performed by the VideoCore
(mostly by the calculate_pitch function there), resulting in failures
to set the image format with some image width values. This patches up
the RGB24 and YUYV type formats to match the VideoCore calculation.

### DIFF
--- a/drivers/staging/vc04_services/bcm2835-isp/bcm2835-v4l2-isp.c
+++ b/drivers/staging/vc04_services/bcm2835-isp/bcm2835-v4l2-isp.c
@@ -676,7 +676,11 @@ struct bcm2835_isp_fmt *get_default_format(struct bcm2835_isp_node *node)
 static inline unsigned int get_bytesperline(int width,
 					    const struct bcm2835_isp_fmt *fmt)
 {
-	return ALIGN((width * fmt->depth) >> 3, fmt->bytesperline_align);
+	/* GPU aligns 24bpp images to a multiple of 32 pixels (not bytes). */
+	if (fmt->depth == 24)
+		return ALIGN(width, 32) * 3;
+	else
+		return ALIGN((width * fmt->depth) >> 3, fmt->bytesperline_align);
 }
 
 static inline unsigned int get_sizeimage(int bpl, int width, int height,
--- a/drivers/staging/vc04_services/bcm2835-isp/bcm2835_isp_fmts.h
+++ b/drivers/staging/vc04_services/bcm2835-isp/bcm2835_isp_fmts.h
@@ -71,7 +71,7 @@ static const struct bcm2835_isp_fmt supported_formats[] = {
 	}, {
 		.fourcc		    = V4L2_PIX_FMT_YUYV,
 		.depth		    = 16,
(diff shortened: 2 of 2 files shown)


davidplowman:
@naushir @6by9 
(I think the RGB565 format is OK. On the VideoCore side it's aligned to 16 pixels before doubling, so 32 bytes per line on the V4L2 side looks correct.)

drivers/staging/vc04_services/bcm2835-isp/bcm2835-v4l2-isp.c
```diff
@@ -676,7 +676,18 @@ struct bcm2835_isp_fmt *get_default_format(struct bcm2835_isp_node *node)
...
+	if (align & align_minus_1) {
+		/* align is not a power of 2 */
+		unsigned int remainder = bpl % align;
+		if (remainder)
+			bpl += (align - remainder);
+	}
+	else
+		bpl = (bpl + align_minus_1) & ~align_minus_1;
+	return bpl;
 }
```
naushir, line 684: Not entirely sure, but I would guess the upstream folks might object to using align_minus_1 variable over an explicit (align - 1).

davidplowman: No problem. Given that I was using  it several times I thought I'd make it explicit that I expected the value to be re-used, but if we think (a) the compiler would do it anyway and (b) upstream maintainers might object, then I'll happily change it. Is that the general feeling?

pelwell: They would probably prefer you to use the existing ALIGN and IS_ALIGNED macros.

6by9: Personally I don't think we need the complication with manually working out if it's a bitmask or not.
24bpp is the only format that has quirks, and you have a near match with https://github.com/raspberrypi/linux/blob/rpi-5.4.y/drivers/staging/vc04_services/bcm2835-camera/bcm2835-camera.c#L994
```
if (fmt->depth == 24)
   return <whatever the correct align is here, and it doesn't need to be 96>
else
  return ((width * fmt->depth) >> 3, fmt->bytesperline_align);
```
Admittedly it's not a performance critical path, but manually computing whether we need to do the modulo seems nuts.

MMAL sets the stride via a multiple of width * bpp. 32pixels * 24bpp is 96 bytes, which fulfils the ISP requirement for a multiple of 32 bytes. So I think we can do 
```
  return ((ALIGN(width, fmt->bytesperline_align) * fmt->depth) >> 3);
```
and leave the bytesperline_align at 32 (or hardcode it here).

(edited - it helps to actually call the ALIGN macro)

davidplowman:
New version that hardcodes the special case.

6by9:
I'm happy with that.

PR #3651, opened 2020-06-01
Add Micro Crystal RV-1805 to i2c-rtc overlays

kpfleming:
In discussions with one of the maintainers about the rtc-abx80x driver and support of the RV-1805 device, it was made clear that relying on the auto-detection behavior in the driver will result in this device being improperly initialized. The maintainer would actually prefer to remove the auto-detection logic completely, but it at least works properly for the Abracon-branded devices originally supported by the driver.

This patch adds an explicit rv1805 entry to the i2c-rtc overlays so that the user can specify it in config.txt. It has been produced (and tested) against the rpi-5.4.y branch but should apply cleanly to the other branches as well.

Note that there is an existing difference between these overlays: i2c-rtc-overlay has support for the pcf85363, but the i2c-rtc-gpio-overlay does not. This results in numbering differences between some of the fragments.

### DIFF
--- a/arch/arm/boot/dts/overlays/i2c-rtc-gpio-overlay.dts
+++ b/arch/arm/boot/dts/overlays/i2c-rtc-gpio-overlay.dts
@@ -204,6 +204,23 @@
 		};
 	};
 
+	fragment@13 {
+		target = <&i2c_gpio>;
+		__dormant__ {
+			#address-cells = <1>;
+			#size-cells = <0>;
+			status = "okay";
+
+			rv1805: rv1805@69 {
+				compatible = "microcrystal,rv1805";
+				reg = <0x69>;
+				abracon,tc-diode = "standard";
+				abracon,tc-resistor = <0>;
+				status = "okay";
+			};
+		};
+	};
+
 	__overrides__ {
 		abx80x = <0>,"+1";
 		ds1307 = <0>,"+2";
@@ -217,6 +234,7 @@
 		m41t62 = <0>,"+10";
 		rv3028 = <0>,"+11";
 		pcf2129 = <0>,"+12";
+		rv1805 = <0>,"+13";
 
 		addr = <&abx80x>, "reg:0",
 		       <&ds1307>, "reg:0",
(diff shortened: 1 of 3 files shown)


arch/arm/boot/dts/overlays/i2c-rtc-gpio-overlay.dts
```diff
@@ -204,6 +204,23 @@
...
 	};
 
+	fragment@13 {
+		target = <&i2c_gpio>;
+		__dormant__ {
+			#address-cells = <1>;
+			#size-cells = <0>;
+			status = "okay";
+
+			rv1805: abx80x@69 {
```
kpfleming: One open question here, as I wasn't able to find sufficient Device Tree documentation: should the 'abx80x' on this line be 'rv1805'? I'm not sure what this label refers to.

pelwell: The part after the colon and before the left brace - `abx80x@69` - is the name of the node; it has to be unique amongst the subnodes/children of the i2c_gpio node. The bit before the colon is the label, used when something else (another fragment or a parameter) refers to a node; labels have to be unique for the whole dtb or overlay, although normally multiple overlays can define the same label without clashing (labels aren't normally preserved when the overlay is applied).

A node name can be anything you like - they have no significance - but there are conventions. The part after the `@` (if there is one) is referred to as the address, and should match the value of the `reg` property. The remainder should be something meaningful - in this case I would use `rv1805`.

pelwell: Plus, the automated checker is happier if the "name" portion matches the name of the parameter that enables the fragment, so `rv1805` is much preferred.

pelwell:
Apart from the node name discussed above, you'll need to document the new parameters in the README.

arch/arm/boot/dts/overlays/i2c-rtc-gpio-overlay.dts
kpfleming: All done, thanks.

pelwell:
Thanks!

PR #3655, opened 2020-06-03
vc4: cec: Restore cec physical address on reconnect

popcornmix:
Currently we call cec_phys_addr_invalidate on a hotplug deassert.
That may be due to a TV power cycling, or an AVR being switched
on (and switching edid). This makes CEC unusable.

Set it back up again on the hotplug assert.

### DIFF
--- a/drivers/gpu/drm/vc4/vc4_hdmi.c
+++ b/drivers/gpu/drm/vc4/vc4_hdmi.c
@@ -113,20 +113,29 @@ static enum drm_connector_status
 vc4_hdmi_connector_detect(struct drm_connector *connector, bool force)
 {
 	struct vc4_hdmi *vc4_hdmi = connector_to_vc4_hdmi(connector);
+	bool connected = false;
 
 	if (vc4_hdmi->hpd_gpio) {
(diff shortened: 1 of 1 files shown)


drivers/gpu/drm/vc4/vc4_hdmi.c
```diff
@@ -113,20 +113,29 @@ static enum drm_connector_status
...
-		return connector_status_connected;
-
+			connected = true;
+	} else if (drm_probe_ddc(vc4_hdmi->ddc))
+		connected = true;
 	if (HDMI_READ(HDMI_HOTPLUG) & VC4_HDMI_HOTPLUG_CONNECTED)
+		connected = true;
+	if (connected) {
+		if (connector->status != connector_status_connected) {
+			struct edit *edid = drm_get_edid(connector, vc4_hdmi->ddc);
```
HiassofT: there's a typo here, it should be `struct edid` (not edi**t**)

popcornmix: Oops. Fixed.

PR #3656, opened 2020-06-03
vc4: Set driver_name for card

popcornmix:
Allows use of the same alsa conf file for hdmi1

From suggestion by @HiassofT

### DIFF
--- a/drivers/gpu/drm/vc4/vc4_hdmi.c
+++ b/drivers/gpu/drm/vc4/vc4_hdmi.c
@@ -1216,6 +1216,7 @@ static int vc4_hdmi_audio_init(struct vc4_hdmi *vc4_hdmi)
 	card->dai_link = dai_link;
 	card->num_links = 1;
 	card->name = vc4_hdmi->variant->id ? "vc4-hdmi1" : "vc4-hdmi";
+	card->driver_name = "vc4-hdmi";
 	card->dev = dev;
 
 	/*


JamesH65:
Just out of interest, how does it achieve its objective?

pelwell:
I would guess that the ALSA configuration file allows you to select by type (`driver_name`) as well as by instance (`name`).

popcornmix:
Currently we need a file `/usr/share/alsa/cards/vc4-hdmi.conf`
```
# Configuration for the VC4-HDMI sound card using software IEC958
# subframe conversion

<confdir:pcm/hdmi.conf>
vc4-hdmi.pcm.hdmi.0 {
	@args [ CARD AES0 AES1 AES2 AES3 ]
	@args.CARD {
		type string
	}
	@args.AES0 {
		type integer
	}
	@args.AES1 {
		type integer
	}
	@args.AES2 {
		type integer
	}
	@args.AES3 {
		type integer
	}
	type iec958
	slave {
		format IEC958_SUBFRAME_LE
		pcm {
			type hooks
			slave.pcm {
				type hw
				card $CARD
				device 0
			}
			hooks.0 {
				type ctl_elems
				hook_args [
				{
					name "IEC958 Playback Default"
					optional true
					lock true
					preserve true
					value [ $AES0 $AES1 $AES2 $AES3 ]
				}
				]
			}
		}
	}
	status [ $AES0 $AES1 $AES2 $AES3 ]
}

# default with plug
vc4-hdmi.pcm.default {
	@args [ CARD ]
	@args.CARD {
		type string
	}
	type plug
	slave.pcm {
		type softvol
		slave.pcm {
			@func concat
			strings [ "hdmi:" $CARD ]
		}
		control {
			name "PCM Playback Volume"
			card $CARD
		}
	}
}
```
to handle its very quirky nature. That doesn't handle the second hdmi port which has the same quirky behaviour. So you either need to duplicate that file, set an alias in `/usr/share/alsa/cards/aliases.conf`
```
vc4-hdmi1 cards.vc4-hdmi
```
or set the driver name which is also considered for matching the conf file.

This PR seems the option that has the least chance for users to get it wrong.

HiassofT:
@JamesH65 alsa lib uses the driver name to look up the card configuration file. So a single "HDA-Intel.conf" or "USB-Audio.conf" file can be used for a wide variety of sound cards. As the ALSA drivers usually expose a fixed structure of devices, mixer controls etc that works fine.

The conf file lookup happens in the "load_for_all_cards" block in /usr/share/alsa/alsa.conf https://git.alsa-project.org/?p=alsa-lib.git;a=blob;f=src/conf/alsa.conf;h=0e01c887dc28023cd0f7b75f443bde73443afbe1;hb=HEAD#l35 private_string gets set to the driver name - see also the load_for_all_cards docs https://www.alsa-project.org/alsa-doc/alsa-lib/conf_8c.html#ac64a8c73293ee35ed3f3958d47bdc721

On the kernel side the ALSA driver name is stored in the "driver" member of snd_card https://github.com/raspberrypi/linux/blob/rpi-5.4.y/include/sound/core.h#L84 which is exposed to userspace via the card info control.

For ASoC things are a bit more tricky, as completely different cards could be handled by a single generic driver (eg the simple audio card driver which is configured via DT).

By default the ALSA driver name is set to the ASoC card name (which is either set in the ASoC driver or passed in via DT), but if we set driver_name that takes precendence: https://github.com/raspberrypi/linux/blob/rpi-4.19.y/sound/soc/soc-core.c#L2104-L2105

So, if we are sure that our ASoC driver will always expose the same device/control/... structure we can simply set driver_name to use a single ALSA card conf file.

JamesH65:
Thanks for the explanations everyone.

PR #3666, opened 2020-06-11
/dev/vc-mem read() using DMA

hvenev:
I've updated https://git.venev.name/hristo/rpi-vc-log to use `pread()` and it works on my Pi 4.

Note that we rely on the fact that the DMA channel has the same view of memory as the VPU. Using the 40-bit DMA channels will require translating addresses somehow.

The way we map VPU addresses to ARM physical addresses needs some work. Instead of just adding some `phys_addr`, we should probably go through the device tree `ranges` for every page, and fail to map the ones that cannot be translated.

Edit: Can you please explain why something is mangling `/axi/ranges`?


hvenev:
It looks like this also works on Pi 3B+.

pelwell:
That's good to hear. I wasn't to take a while to understand what you've done and why, but I'll give you some feedback in the next few days.

hvenev:
Documented dt binding and rebased.

Issue #3675, opened 2020-06-14
Bluetooth is not autodetected with device tree (bcm2837-rpi-3-b.dtb) on Raspberry Pi 3


gentoo-root:
**Is this the right place for my bug report?**
I hope so, because it's kernel-related and Raspberry Pi-specific.

**Describe the bug**
The upstream device tree (bcm2837-rpi-3-b.dtb) attempts to configure Bluetooth:
```
/* uart0 communicates with the BT module */
&uart0 {
        pinctrl-names = "default";
        pinctrl-0 = <&uart0_gpio32 &gpclk2_gpio43>;
        status = "okay";

        bluetooth {
                compatible = "brcm,bcm43438-bt";
                max-speed = <2000000>;
                shutdown-gpios = <&expgpio 0 GPIO_ACTIVE_HIGH>;
        };
};
```
However, the code in [`drivers/of/platform.c`](https://github.com/raspberrypi/linux/blob/b6690fa9e11f53b017cde139b0bba92b93afecc5/drivers/of/platform.c#L383-L390) skips the `bluetooth` subsection, because `uart0` is defined in `bcm283x.dtsi` as follows:
```
                uart0: serial@7e201000 {
                        compatible = "arm,pl011", "arm,primecell";
                        reg = <0x7e201000 0x200>;
                        interrupts = <2 25>;
                        clocks = <&clocks BCM2835_CLOCK_UART>,
                                 <&clocks BCM2835_CLOCK_VPU>;
                        clock-names = "uartclk", "apb_pclk";
                        arm,primecell-periphid = <0x00241011>;
                };
```
It contains `"arm,primecell"` in `compatible`, which makes the enumeration code in `drivers/of/platform.c` to return early and skip enumerating children of `uart0`, i.e. the `bluetooth` node. As a result, hci_uart is not autoloaded by udev, and when loaded manually, it doesn't autodetect the Bluetooth hardware, so I need to run `btattach` or `hciattach` manually.

Tested on both this kernel (built from the 5.7.y branch) and the upstream kernel (5.7.2).

**To reproduce**
1. Have bluez installed. Disable `hciuart.service` or any other scripts that run `hciattach` or `btattach` on boot.
2. Set `device_tree=dtbs/broadcom/bcm2837-rpi-3-b.dtb` in `/boot/config.txt`.
3. Boot the system.

**Expected behaviour**
[...]

PR #3682, opened 2020-06-21
Add Bluetooth overlays

gentoo-root:
As a follow-up for #3675, this pull request adds device tree overlays to give users an option to configure Bluetooth without hciattach/btattach, and changes the defconfigs to make this way work.


pelwell:
All overlays require documentation in the README file - if you add that now I can review it all together.

macmpi:
Nice.
You may want to add reference to PiZeroW in bt-overlay.dts , and also do similar for bcm2708 (bcm2835 in upstream DTBs).
@pelwell could that eventually be merged into 5.4.y too? That would really be sweet.

pelwell:
Anything merged is eventually ported to all current branches (where possible).

gentoo-root:
>You may want to add reference to PiZeroW in bt-overlay.dts

Sure, I can mention it. Just to clarify (because I don't have one for tests), does Zero W support flow control (so that bt-overlay.dts is the right one)?

>also do similar for bcm2708 (bcm2835 in upstream DTBs).

Sorry, I didn't get it... Do you mean I need to add some different overlay for bcm2708?

pelwell:
I have no problem with what you've done so far, but I think we can do better from a user's perspective. I don't like the fact that they have to know whether their Pi supports BT flow control - it would be better that enabling Bluetooth in a common way (i.e. one overlay, or one parameter) automatically chose the default baud rate, but with an override to change it.

The serdev enumeration that enables the Bluetooth support checks for enabled nodes, which means the `bluetooth` nodes can appear in the base DTBs by default, with the correct baud rates, but disabled, with an overlay or (better) dtparam to enable it. If we ever switch over so that this method is used by default we could just change the default state and keep the parameter as a way to disable it.

What do you think?

gentoo-root:
Sounds great! Basically, I can add the `bluetooth` nodes to the main trees, disabled by default, and with correct baud rates for different boards (i.e. 3000000 for Pi4, 921600 for Pi3), and a dtparam to enable Bluetooth — I think, I can do that.

However, there is this Pi3 rev 1.2/1.3 difference in flow control support, and both of them share the same device tree. Do you have an idea how to handle it with your approach? I can only think of another dtparam to force baud rate, but it means the user will have to set it manually on Pi3 rev 1.3.

pelwell:
Hmm - okay then, in that case it makes sense for the firmware to set a sensible baud rate using the parameter before any user settings are applied. Until the firmware is updated, users of the 3B <= rev 1.2 will have to set the baud rate manually.

pelwell:
Which (I think) means the `bluetooth` node just needs to go in bcm270x-rpi.dtsi.

macmpi:
What's the final take on this?\
`bluetooth` node in _bcm270x-rpi.dtsi_, and `dtparam` to activate & eventually set custom baud rate (particularly for Pi3 rev 1.2)?

(@gentoo-root yes indeed, PiZeroW does support HW flowcontrol)

gentoo-root:
> `bluetooth` node in _bcm270x-rpi.dtsi_

`bcm270x-rpi.dtsi` is too generic, it's included by the boards without Bluetooth, too. I was planning to patch the concrete boards with Bluetooth: `bcm2708-rpi-zero-w.dts`, `bcm2710-rpi-3-b.dts`, `bcm2710-rpi-3-b-plus.dts`, `bcm2711-rpi-4-b.dts` (I hope the list is complete), because `*.dtsi` files are always shared with some non-Bluetooth boards. Probably it makes sense to add a new `bcm27xx-rpi-bt.dtsi` and include it from those four `*.dts` files.

>`dtparam` to activate & eventually set custom baud rate (particularly for Pi3 rev 1.2)?

Yes.

BTW, adding the `bluetooth` node to the main tree will require changes to `disable-bt-overlay.dts` and `miniuart-bt-overlay.dts`.

I'm on it currently.

gentoo-root:
I updated the patch, tested the following combinations on Pi3B rev 1.2 and got the expected results:

```
# Probes at 3 Mbaud.
dtparam=bluetooth=on
```

```
# Probes at 921600 baud.
dtparam=bluetooth=on
dtparam=bluetooth_baudrate=921600
```

```
# Doesn't probe.
dtparam=bluetooth=on
dtparam=bluetooth_baudrate=921600
dtoverlay=disable-bt
```

```
# Doesn't autoprobe; `btattach -B /dev/ttyS0 -P bcm -S 460800 -N` works.
dtparam=bluetooth=on
dtparam=bluetooth_baudrate=921600
dtoverlay=miniuart-bt
```

```
# Probes at 460800 on miniUART, /dev/ttyS0 doesn't appear.
dtoverlay=miniuart-bt,probe=on
```

arch/arm/boot/dts/bcm2708-rpi-zero-w.dts
```diff
@@ -4,6 +4,7 @@
 #include "bcm2708-rpi.dtsi"
 #include "bcm283x-rpi-csi1-2lane.dtsi"
 #include "bcm283x-rpi-i2c0mux_0_28.dtsi"
+#include "bcm27xx-rpi-bt.dtsi"
 
 / {
```
pelwell, line 9: For Zero W you also need to change `shutdown-gpios` - as it is this doesn't compile:
```

&bt {
	shutdown-gpios = <&gpio 45 GPIO_ACTIVE_HIGH>;
};
```

arch/arm/boot/dts/bcm27xx-rpi-bt.dtsi
```diff
@@ -0,0 +1,17 @@
...
+		compatible = "brcm,bcm43438-bt";
+		max-speed = <3000000>;
+		shutdown-gpios = <&expgpio 0 GPIO_ACTIVE_HIGH>;
+		status = "disabled";
+	};
+};
+
+/ {
+	__overrides__ {
+		bluetooth = <&bt>,"status";
```
pelwell: I think we need a different for kernel-configured Bluetooth - explaining to somebody that they need to use Bluetooth with `bluetooth=off` is going to be confusing. I suggest something intentionally technical-sounding like `kernelbt` or even `knlbt` - the people who want to enable this feature will be able to handle an unpronouncable string of 5 characters.

arch/arm/boot/dts/overlays/miniuart-bt-overlay.dts
```diff
@@ -64,11 +80,15 @@
...
+	fragment@6 {
 		target-path = "/aliases";
 		__overlay__ {
 			serial0 = "/soc/serial@7e201000";
 			serial1 = "/soc/serial@7e215040";
 		};
 	};
+
+	__overrides__ {
+		probe = <&minibt>,"status";
```
pelwell: `probe` is a bit vague - I suggest re-using the same name used with the base DTBS, e.g. `knlbt`.

arch/arm/configs/bcm2709_defconfig
```diff
@@ -648,7 +648,8 @@ CONFIG_SERIAL_AMBA_PL011=y
 CONFIG_SERIAL_AMBA_PL011_CONSOLE=y
 CONFIG_SERIAL_SC16IS7XX=m
 CONFIG_SERIAL_SC16IS7XX_SPI=y
-CONFIG_SERIAL_DEV_BUS=m
+CONFIG_SERIAL_DEV_BUS=y
+CONFIG_SERIAL_DEV_CTRL_TTYPORT=y
```
pelwell, line 652: The kernels are going to grow by ~30kB, but in this case I think the eventual end justifies it.

pelwell (changes requested):
Apart from addressing the line comments I'm going to need some entries in the README file for the new parameter, but we're nearly there.

arch/arm/boot/dts/bcm2708-rpi-zero-w.dts
gentoo-root: Oops, missed this one because I was testing build on arm64.

Do you have ideas how to deal with miniuart-bt-overlay best? It also refers to `&expgpio` after my change, and it should work with all Pi models. Is there a way for an overlay to pick one of `&expgpio` or `&gpio`, depending on the main tree? Or any way to add a label alias to point `expgpio` to `&gpio` in the Pi0W tree?

arch/arm/boot/dts/bcm27xx-rpi-bt.dtsi
gentoo-root: Well, technically, the user doesn't have to pass `bluetooth=off` explicitly, as it's the default, but of course, I can rename the parameter.

arch/arm/boot/dts/bcm2708-rpi-zero-w.dts
pelwell: The logical way would be to give uart1 a disabled bluetooth node in the base DTBs, just as we did with uart0.

gentoo-root: Thanks, that sounds like a solution. I'll update the patch shortly.

gentoo-root:
I updated the patch, addressing all the comments above. I'll retest it one more time after we are done with the review.

arch/arm/boot/dts/bcm2708-rpi-bt.dtsi
```diff
@@ -0,0 +1,26 @@
+// SPDX-License-Identifier: GPL-2.0
+
+&uart0 {
+	bt: bluetooth {
+		compatible = "brcm,bcm43438-bt";
+		max-speed = <3000000>;
+		shutdown-gpios = <&gpio 0 GPIO_ACTIVE_HIGH>;
```
pelwell: Shouldn't this be GPIO 45? (I'm on a phone screen so might have missed something)

gentoo-root: Yes, you are absolutely right. Fixing.

pelwell:
That looks good - I'll test it in the morning.

pelwell:
Thanks - that works as expected. It doesn't solve the problem of Pis which haven't had their BDADDR written to OTP, but as this is an opt-in change I can live with that.

pelwell (approved):

macmpi:
Great!Tthanks so much @gentoo-root and @pelwell 
Phil, can merge in 5.4.y too, please?

pelwell:
Your wish is my command.

pelwell:
Just a note to say that the krnbt parameter of the miniuart-bt overlay won't apply as it was written - it was an overlay parameter that tried to directly modify a node in the base DTB, and that's not currently supported. The workaround is to add another fragment, which can be initially empty, targetting the node in the base DTB, and modify the parameter to update the fragment. The fixed version is in rpi-5.4.y now (and will be copied to the others...) but just missed the firmware release.

Having said that, I can't make miniuart-bt,kernbt work at the moment, for reasons I haven't figured out. The kernel claims the uart1 (/dev/ttyS0 disappears) and attempts to bring up the HCI interface, but it times out on the baudrate change:
```
[    5.716118] Bluetooth: HCI device and connection manager initialized
[    5.716143] Bluetooth: HCI socket layer initialized
[    5.751398] Bluetooth: HCI UART driver ver 2.3
[    5.751417] Bluetooth: HCI UART protocol H4 registered
[    5.751507] Bluetooth: HCI UART protocol Three-wire (H5) registered
[    5.751857] Bluetooth: HCI UART protocol Broadcom registered
[    5.753400] hci_uart_bcm serial0-0: serial0-0 supply vbat not found, using dummy regulator
[    5.753529] hci_uart_bcm serial0-0: serial0-0 supply vddio not found, using dummy regulator
[    8.031745] Bluetooth: hci0: command 0xfc18 tx timeout
[   16.095674] Bluetooth: hci0: BCM: failed to write update baudrate (-110)
[   16.095682] Bluetooth: hci0: Failed to set baudrate
[   18.111671] Bluetooth: hci0: command 0x0c03 tx timeout
[   26.335699] Bluetooth: hci0: BCM: Reset failed (-110)
```

pelwell:
This change has caused additional serial ports to no longer appear as character devices - see https://github.com/raspberrypi/firmware/issues/1438 for the report.
Do you know how to solve that problem without reverting this PR?

gearhead:
Thanks for doing this. It is a big help. Are there specific kernel config parameters which need to be set to make this work? Any specific packages which need to be installed? 
I confirm it works for my PiZero and a B3 with the current 5.4.51-1 kernel on Arch Arm for armv6 and armv7. Trying to get it going on aarch64 and have had no joy, yet. Re-building 5.4.51-1 for it right now. To get it to work on my B3, I had to install firmware-brcm43xx. This package is not available for armv6 on arch, though, but it still works on my PiZero.
Update: Re-built kernel with slightly different configs and still no onboard bluetooth.
Do I need to do something with the cmdline.txt?

gearhead:
Fixed it. I meandered around the kernel config and changed a bunch of things but did not change it. I then went and grabbed the raspberry kernel source from Rpi and made a new config and rebuilt and now this config.txt setting works on aarch64 as well.

PR #3684, opened 2020-06-22
Alternative version of PR 3666

pelwell:
@hvenev This is a modified version of #3666 that:
* drops the reduction in DMA channel usage because it's orthogonal and worth a discussion of its own,
* omits the `ranges` changes as they don't seem to achieve anything at the moment,
* enables range checking on the read accesses, and
* fixes a few other things - some cosmetic, some not.

### DIFF
--- a/drivers/char/broadcom/vc_mem.c
+++ b/drivers/char/broadcom/vc_mem.c
@@ -82,23 +82,6 @@ vc_mem_release(struct inode *inode, struct file *file)
 	return 0;
 }
 
-static void
-vc_mem_get_size(void)
-{
-}
-
-static void
-vc_mem_get_base(void)
-{
-}
-
-int
-vc_mem_get_current_size(void)
-{
-	return mm_vc_mem_size;
-}
-EXPORT_SYMBOL_GPL(vc_mem_get_current_size);
-
 static long
 vc_mem_ioctl(struct file *file, unsigned int cmd, unsigned long arg)
 {
@@ -123,9 +106,6 @@ vc_mem_ioctl(struct file *file, unsigned int cmd, unsigned long arg)
 		}
 	case VC_MEM_IOC_MEM_SIZE:
 		{
-			/* Get the videocore memory size first */
-			vc_mem_get_size();
-
 			pr_debug("%s: VC_MEM_IOC_MEM_SIZE=%x\n", __func__,
 				 mm_vc_mem_size);
 
@@ -137,9 +117,6 @@ vc_mem_ioctl(struct file *file, unsigned int cmd, unsigned long arg)
 		}
 	case VC_MEM_IOC_MEM_BASE:
 		{
-			/* Get the videocore memory base */
-			vc_mem_get_base();
-
 			pr_debug("%s: VC_MEM_IOC_MEM_BASE=%x\n", __func__,
 				 mm_vc_mem_base);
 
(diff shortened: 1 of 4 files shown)


drivers/char/broadcom/vc_mem.c
```diff
@@ -216,132 +198,104 @@ static const struct file_operations vc_mem_fops = {
...
+	platform_set_drvdata(pdev, drv);
+
+	// Note that the "reg" property provided by firmware is does not
+	// follow any conventions. It is just a bunch of 32-bit addresses.
+	addrp = of_get_property(pdev->dev.of_node, "reg", &n_addr_bytes);
+	if (!addrp || n_addr_bytes != 12)
+		return -ENODEV;
+	drv->base = be32_to_cpu(addrp[0]);
+	drv->size = be32_to_cpu(addrp[1]);
+	drv->phys_addr = be32_to_cpu(addrp[2]);
```
hvenev, line 347: Are you sure this is correct? On my Pi 4B `addrp[2] == 0xC0000000`. I think that's where the VC sees the start of SDRAM. I would also guess that the purpose served by `addrp[2]` is that on older Pi models it is an offset to add to addresses when doing DMA:
```
dma_memcpy(vc_addr, addrp[2] + host_addr, size);
```
However this is redundant because we have `dma-ranges`, which is the mechanism used by Linux to turn physical addresses into `dma_addr_t`.

drivers/char/broadcom/vc_mem.c
```diff
@@ -157,6 +164,115 @@ vc_mem_compat_ioctl(struct file *file, unsigned int cmd, unsigned long arg)
...
+
+	if (!read_size)
+		return 0;
+
+	// Use the preferred alias for accesses
+	src_begin = init_pos + drv->phys_addr;
+	src_end = src_begin + read_size;
+	src_at = src_begin;
+
+	buf = dma_alloc_coherent(dev, VC_MEM_DMA_SIZE, &buf_paddr, GFP_KERNEL);
```
hvenev, line 222: I think this will be very broken if you don't have `dma-ranges`. From what I remember on my Pi 4 it used to allocate memory anywhere (in the first 4G because that's how much I had), not just the first 1G. The DMA engine ended up overwriting the wrong memory. Funnily enough that killed some pages in the page cache that were then written back so I lost some files.

drivers/char/broadcom/vc_mem.c
```diff
@@ -157,6 +164,115 @@ vc_mem_compat_ioctl(struct file *file, unsigned int cmd, unsigned long arg)
...
+	if ((init_pos & alias_mask) != alias_cached &&
+	    (init_pos & alias_mask) != alias_direct &&
+	    (init_pos & alias_mask) != (drv->phys_addr & alias_mask))
+		return -EINVAL;
+
+	// Discard the cache alias bits for bounds checking
+	init_pos &= ~alias_mask;
+
+	if (init_pos < drv->base || init_pos > drv->size ||
+	    init_pos + read_size > drv->size)
```
hvenev, line 211: On 32-bit we are relying on the fact that `read_size <= MAX_RW_COUNT == 0x7ffff000` and on the `&= ~alias_mask` above. If either were not the case, `init_pos + read_size` could overflow.

I think this should either be changed to something that never overflows or explained with a comment.

drivers/char/broadcom/vc_mem.c
pelwell: Several of these issues stem from putting the vc_mem node under /axi, forcing one to duplicate the ranges and dma-ranges of /soc. We can replace the dev in the dma_* calls with the DMA controller, but it seems more correct to move vc_mem under /soc and locate /axi/reg explicitly.

hvenev: Good idea. Should we add a pointer from `/soc/vc_mem` to `/axi/vc_mem`? Something like `/soc/vc_mem { raspberrypi,legacy-node = <&axi_vc_mem>; };`

PR #3688, opened 2020-06-23
Upstream Unicam driver

6by9:
Whilst not merged, this takes the upstream Unicam driver, backports it to 5.4, and includes any of our fixes since then, and those raised by upstream. With those patches squashed it should be a reasonable v3.
It is dependent on another upstream patchset that hasn't been merged yet for dynamically changing the number of active CSI data lanes being used, so that has been backported as well.

DT has changed slightly, so those are all fixed up as well.


6by9:
@davidplowman @naushir Please test. It's working for me with TC358743, but I haven't tried the camera modules and libcamera.

@pinchartl Hopefully a good next step for v3.
Drop caa8557f427a1f6c18fbbb6764ec827b6a2d4a30 and you could still upstream it without the dependency on Jacopo's patchset.

Now on to the ISP driver and vc-sm-cma....

naushir:
Looks to be working fine for me in libcamera.

drivers/media/platform/bcm2835/bcm2835-unicam.c
```diff
@@ -2428,35 +2411,33 @@ static int register_node(struct unicam_device *unicam, struct unicam_node *node,
...
 	}
-
-	if (node->pad_id == METADATA_PAD) {
-		v4l2_disable_ioctl(vdev, VIDIOC_DQEVENT);
-		v4l2_disable_ioctl(vdev, VIDIOC_SUBSCRIBE_EVENT);
-		v4l2_disable_ioctl(vdev, VIDIOC_UNSUBSCRIBE_EVENT);
-	}
-	if (node->pad_id == METADATA_PAD ||
+	/* TODO: review which of these disables are still needed */
+	if (pad_id == METADATA_PAD ||
```
naushir, line 2430: Is this TODO still needed?

6by9: Yes.
Hans' comment on https://patchwork.linuxtv.org/patch/63531/

> Some of these are probably no longer needed since the v4l2 core will disable them for you.

I've removed the disables of the event ioctls which we discussed the other day, but I didn't look into these.
I believe it is true that the core will disable some ioctls based on this being a metadata type, but it'd need careful checking, and may be different between 5.4 and linux-media-next.

drivers/media/platform/bcm2835/bcm2835-unicam.c
```diff
@@ -0,0 +1,2825 @@
...
+
+static const struct v4l2_async_notifier_operations unicam_async_ops = {
+	.bound = unicam_async_bound,
+	.complete = unicam_async_complete,
+};
+
+static int of_unicam_connect_subdevs(struct unicam_device *dev)
+{
+	struct platform_device *pdev = dev->pdev;
+	struct v4l2_fwnode_endpoint ep = { 0 };
```
pelwell: I get a build warning on this line:
```
drivers/media/platform/bcm2835/bcm2835-unicam.c: In function 'of_unicam_connect_subdevs':
drivers/media/platform/bcm2835/bcm2835-unicam.c:2582:9: warning: missing braces around initializer [-Wmissing-braces]
  struct v4l2_fwnode_endpoint ep = { 0 };
         ^
drivers/media/platform/bcm2835/bcm2835-unicam.c:2582:9: warning: (near initialization for 'ep.base') [-Wmissing-braces]
```

pelwell: It's happy with ` = { { 0 } };`.

popcornmix: Isn't `= { }` the correct initialiser?
Including the `0` means the first element of structure needs to be assignable to zero which isn't the case with many types (including sub-structures)

pelwell: I think the problem is that the first element of a `v4l2_fwnode_endpoint` is another struct - `fwnode_endpoint` - which has a scalar value as the first element.

popcornmix: Yes but doesn't `= {}` avoid the warning?
It's a gcc construct, but is widely used in the kernel.

pelwell: Are you saying you don't get a warning?

popcornmix: I don't with `= {}` I do with `={0}`

pelwell:
```
phil@buildbot:~/pi/linux$ git grep '= { { 0 } };' |  wc
     54     533    5135
phil@buildbot:~/pi/linux$ git grep '= { { { 0 } } };' |  wc
      4      49     396
```

popcornmix:
```
$ git grep '= { };' | wc
    573    4043   41864
```

JamesH65: = {0} is the correct syntax to zero an entire structure to zero, but there is/was a known bug in gcc that meant it failed to compile in some cases. I thought it had been fixed though. I was on the distribution list for the fix because I reported it at some point in the ancient past.

popcornmix:
```
$ git grep '= {};' | wc
   1754   10450  132323
```

pelwell:
```
phil@buildbot:~/pi/linux$ git grep '= {};' |  wc
   1754   10450  132323
```

pelwell: You win. It's not an extension I'd noticed before, but it makes more sense than the alternative.

6by9: It comes from the version of the patch that was being upstreamed, and they tend to use a later gcc than we appear to.
{0} is meant to zero the whole struct.

6by9:
Rebased. Now just checking the disable_ioctl calls for metadata nodes.

drivers/media/platform/bcm2835/bcm2835-unicam.c
pelwell: This isn't going to make for a particularly tricky merge conflict if it does change upstream.

6by9: I'm happy to create an additional downstream patch here to use memset instead to avoid the build warning, but I'm hoping Laurent will otherwise pick up the patches to create the next patchset to send to the list, and they'll want ```= { 0 }```

pelwell:
Were you expecting to lose "media: bcm2835-unicam: Retain packing information on G_FMT" and "media: bcm2835-unicam: change minimum number of vb2_queue buffers to 1" around the time of the rebase?

6by9:
> Were you expecting to lose "media: bcm2835-unicam: Retain packing information on G_FMT" and "media: bcm2835-unicam: change minimum number of vb2_queue buffers to 1" around the time of the rebase?

No :-(
Time to retrieve them.

6by9:
OK, retrieved the two missing commits, and added one to use memset instead of ={0}.
Apologies if I've introduced a build error here - my laptop is creaking in the heat, and probably wants a reboot too.

pelwell:
Did you not want to use ` = {};` (note the lack of a 0), since it is the most common way to initialise a structure to zero in the kernel?
Otherwise this looks good, and there are no build errors (or new warnings).

6by9:
Up to you over ```= {};``` vs ```memset```. This is a downstream only patch so makes little difference in the grand scheme of things.

Give me a chance to check up on those disable_ioctl calls as it'd be nice to merge all changes in this one PR.

Anything else from @naushir?

pelwell:
I'd prefer `= {};` because it's one fewer line to read and one fewer line to contain mistakes.

I'll wait for the nod re: merging.

6by9:
Updated with ``` = { };```

We do need the disable_ioctl calls as both video and metadata nodes are type VFL_TYPE_[VIDEO|GRABBER] (it got renamed), and the core disabling of ioctls is based on that type. We want to disable them based on whether the source subdev supports the relevant subdev API call or not, so our struct of ops always contains the superset of all of them.

6by9:
I'm all happy now.

pelwell:
Just in time for the weekend.

PR #3691, opened 2020-06-26
Sony IMX290/IMX327 support

6by9:
Sony IMX290 and IMX327 are broadly compatible sensors.
Backport the mainline driver and latest fixes (in linuxtv/master, not yet in torvalds linux/master).
Adds additional controls, alternate oscillator frequency support, and an overlay to load the thing.

I still need to compile test this branch, and make sure that the changes in the header comments to the driver are in a sensible patch (I think they may have got mixed up with other stuff).


6by9:
Rebased and removed the WIP tag.
I think this is good for review now.

6by9:
@andrey-konovalov FYI Your/Mani's patches backport, and my couple of extras on top.

Hopefully @davidplowman will be getting it running with libcamera next week when his module arrives.

pelwell:
Is there a preferred merge order between this and #3688?

6by9:
> Is there a preferred merge order between this and #3688?

There shouldn't be. This should work with the older Unicam driver. The imx290 dtoverlay may have a couple of extra bits that the older unicam driver doesn't care about, but those will be ignored.

6by9:
Oh, and the reason I've split the dtoverlay into a imx290 and imx290_327 overlay is so that we can add an imx327 overlay easily if we get to a point where we support the differences between the two (120fps mode on IMX290).

arch/arm/boot/dts/overlays/README
```diff
@@ -1430,6 +1430,21 @@ Load:   dtoverlay=imx219
 Params: <None>
 
 
+Name:   imx290
+Info:   Sony IMX290 camera module.
+        Uses Unicam 1, which is the standard camera connector on most Pi
+        variants. NB This currently uses 4 CSI2 data lanes and therefore will
+        only work on a CM.
+Load:   dtoverlay=imx290,<param>
+Params: 4lanes              Enable 4 CSI2 lanes. This requires a Compute Module
```
pelwell: The overlay uses the name `4lane` (singular). And (annoyingly, I know) parameter descriptions _must_ be indented to column 33, i.e. with 32 leading spaces (this is column 29). e.g.:
```
Params: 4lane                   Enable 4 CSI2 lanes. This requires a Compute
                                Module (1, 3, or 4).
        clock-frequency         Sets the clock frequency to match that used on
                                the board. Modules from Vision Components use
                                37.125MHz (the default), whilst those from
                                Innomaker use 74.25MHz.
        mono                    Denote that the module is a mono sensor.
```

arch/arm/boot/dts/overlays/imx290_327-overlay.dts
```diff
@@ -0,0 +1,145 @@
+// SPDX-License-Identifier: GPL-2.0-only
```
pelwell: Could you make this a `.dtsi` file instead to make it clear to the reader (and the checker) that it isn't a complete overlay that needs documentation and a Makefile entry?

drivers/media/i2c/imx290.c
```diff
@@ -40,6 +40,9 @@ enum imx290_clk_index {
 #define IMX290_GAIN 0x3014
 #define IMX290_HMAX_LOW 0x301c
 #define IMX290_HMAX_HIGH 0x301d
+#define IMX290_HMAX_MIN_2LANE 4400 /* Min of 4400 pixels = 30fps */
+#define IMX290_HMAX_MIN_4LANE 2200 /* Min of 200 pixels = 60fps */
```
pelwell: Typo - `200` should be `2200`.

drivers/media/i2c/imx290.c
```diff
@@ -518,6 +523,26 @@ static int imx290_set_gain(struct imx290 *imx290, u32 value)
 	return ret;
 }
 
+static int imx290_set_hmax(struct imx290 *imx290, u32 val)
+{
+	u32 hmax = val + imx290->current_mode->width;
+	int ret;
+
+	ret = imx290_write_reg(imx290, IMX290_HMAX_LOW, (hmax & 0xff));
```
pelwell: Why are IMX290_HMAX_LOW & _HIGH written explicitly as two 8-bit registers, rather than using imx290_write_buffered_reg(imx290, IMX290_HMAX_LOW, 2, hmax)? Have I missed a subtlety?

drivers/media/i2c/imx290.c
```diff
@@ -618,6 +622,16 @@ static int imx290_set_ctrl(struct v4l2_ctrl *ctrl)
...
 		ret = imx290_set_vmax(imx290, ctrl->val);
 		break;
+	case V4L2_CID_HFLIP:
+	case V4L2_CID_VFLIP:
+		/* WINMODE is in bits [6:4], so need to read-modify-write */
+		ret = imx290_read_reg(imx290, IMX290_FLIP_WINMODE, &val);
+		if (ret)
+			break;
+		val &= ~0x03;
+		val |= imx290->vflip->val | (imx290->hflip->val << 1);
```
pelwell, line 637: Does V4L2 guarantee these are always 0 or 1?

drivers/media/i2c/imx290.c
```diff
@@ -648,7 +648,7 @@ static int imx290_power_off(struct device *dev)
 }
```
pelwell, line 1054: Typo `uptream` in commit message.

drivers/media/i2c/imx290.c
```diff
@@ -312,6 +358,11 @@ static const struct imx290_mode imx290_modes_2lanes[] = {
 		.link_freq_index = FREQ_INDEX_1080P,
 		.data = imx290_1080p_settings,
 		.data_size = ARRAY_SIZE(imx290_1080p_settings),
+		.clk_data = {
+			imx290_37_125mhz_clock_1080p,
```
pelwell: `[CLK_37_125] = ...`? Or a comment about the order? I'll take a no.

drivers/media/i2c/imx290.c
```diff
@@ -1091,6 +1106,12 @@ static s64 imx290_check_link_freqs(const struct imx290 *imx290,
 	return 0;
 }
 
+static const struct of_device_id imx290_of_match[] = {
+	{ .compatible = "sony,imx290", .data = (void *)0 },
+	{ .compatible = "sony,imx290-mono", .data = (void *)1 },
+	{ /* sentinel */ }
```
pelwell, line 1106: Oh look - a initialising a structure to zero without the aid of a zero.

drivers/media/i2c/imx290.c
```diff
@@ -1115,6 +1137,17 @@ static int imx290_probe(struct i2c_client *client)
 		return -ENODEV;
 	}
 
+	match = of_match_device(imx290_of_match, dev);
+	if (!match)
+		return -ENODEV;
+	if (!match->data) {
```
pelwell: It's shame you can't just use the match data directly, but those ARRAY_SIZE()s make it hard to find a neat way to do it.

arch/arm/boot/dts/overlays/README
6by9: Done

arch/arm/boot/dts/overlays/imx290_327-overlay.dts
6by9: Done

drivers/media/i2c/imx290.c
6by9: Done

drivers/media/i2c/imx290.c
6by9: I'd meant to query this on the mailing list, but missed my chance as they got merged just before I sent my comment and Tested-by: tag.

drivers/media/i2c/imx290.c
6by9: Yes.
https://elixir.bootlin.com/linux/latest/source/drivers/media/v4l2-core/v4l2-ctrls.c#L1153
Type V4L2_CTRL_TYPE_BOOLEAN, min 0, max 1

drivers/media/i2c/imx290.c
6by9: Not that my diff here is only moving the function from below as it needs to be called from imx290_set_ctrl

drivers/media/i2c/imx290.c
6by9: Done.

drivers/media/i2c/imx290.c
6by9: :-P

drivers/media/i2c/imx290.c
6by9: There is a comment in the definition of struct imx290_mode
```
	/* Clock setup can vary. Index as enum imx290_clk_index */
	const struct imx290_regval *clk_data[2];
```
but it doesn't hurt to be explicit.
Done.

drivers/media/i2c/imx290.c
6by9: There are only ever two formats, so the arrays should always be the same size.

If I can remember the compile time array size check macros to enforce that then I'd happily switch to passing the format pointer directly from match data.

6by9: Actually
```
#define IMX290_NUM_FORMATS 2

static const struct imx290_pixfmt imx290_colour_formats[IMX290_NUM_FORMATS] = {
	{ MEDIA_BUS_FMT_SRGGB10_1X10, 10 },
	{ MEDIA_BUS_FMT_SRGGB12_1X12, 12 },
};

static const struct imx290_pixfmt imx290_mono_formats[IMX290_NUM_FORMATS] = {
	{ MEDIA_BUS_FMT_Y10_1X10, 10 },
	{ MEDIA_BUS_FMT_Y12_1X12, 12 },
};
```
and then using IMX290_NUM_FORMATS instead of the equivalent ARRAY_SIZE value. If someone were to ever add an extra value into the array then it'd break the compile.

Happy with that solution?

pelwell: Delighted.

6by9: Done then.

drivers/media/i2c/imx290.c
pelwell: OK - it was just an observation. I initially had a comment against the IMX290_VMAX support before realising that it was correct.

drivers/media/i2c/imx290.c
```diff
@@ -312,6 +358,11 @@ static const struct imx290_mode imx290_modes_2lanes[] = {
 		.link_freq_index = FREQ_INDEX_1080P,
 		.data = imx290_1080p_settings,
 		.data_size = ARRAY_SIZE(imx290_1080p_settings),
+		.clk_data = {
+			[CLK_37_125] = imx290_37_125mhz_clock_1080p,
+			[CLK_74_250] = imx290_74_250mhz_clock_1080p,
```
pelwell: This should be CLK_74_25, and below. And the _4lanes modes could also do with the same treatment.

drivers/media/i2c/imx290.c
6by9: I've added an extra patch that converts imx290_set_hmax to use imx290_write_buffered_reg

drivers/media/i2c/imx290.c
6by9: My compile was still running :-(

6by9:
Branch updated based on your review comments, but my test build is still running....

pelwell:
Spoiler - it builds.

pelwell:
Happy for me to merge?

6by9:
> Spoiler - it builds.

I need a faster machine, or to switch branches less.

> Happy for me to merge?

Yes. Thanks for your review.

pelwell:
Use git worktrees, and buildbot.

Issue #2177, opened 2017-08-27
Could nf_tables be enabled in raspbian


github-user-1:
Hello all,
the module for nf_tables (and it's possible dependencies) is not loaded in the latest Raspbian Stretch. I thought it would be (based on the announcement that it will become available with Debian Stretch). Since netfiltering should be moved from iptables etc. over to nftables in the long run I thought this would be a good start now with Stretch being released.

Are there any plans to enable nft (nftables) by standard for Raspbian as well ?

popcornmix:
To be clear, are you requesting ``CONFIG_NF_TABLES=m`` be added to kernel config?

github-user-1:
Probably yes - however I consider myself rather an end-user of linux (gnu cli rather). So I don't know much about the in and outs of the linux kernel used for raspbian but I was simply wondering why nft wasn't working with Raspbian Stretch (as it was officially announced for Debian Stretch to become the replacement of iptables etc.). Looking for the reason why nft didn't work after I manually installed it from the repo, I can see e.g. that the modul is not loaded:
```
$ modinfo nf_tables
modinfo: ERROR: Module nf_tables not found.
```
whereas iptables is for example:
```
 $ modinfo ip_tables
filename:       /lib/modules/4.9.41-v7+/kernel/net/ipv4/netfilter/ip_tables.ko
description:    IPv4 packet filter
author:         Netfilter Core Team <coreteam@netfilter.org>
license:        GPL
srcversion:     073F6A28550B59E4A9A7F74
depends:        x_tables
intree:         Y
vermagic:       4.9.41-v7+ SMP mod_unload modversions ARMv7 p2v8
```

Ferroin:
For nft to actually be usable will require more than just CONFIG_NF_TABLES=m.  Ideally, most of the options under that in the config menu should be enabled as well, otherwise you can't really do much with nft.

github-user-1:
Just as a start and to provide some hints what might be required in the kernel config (copied from https://home.regit.org/netfilter-en/nftables-quick-howto/):

`$ make oldconfig

Netfilter Xtables support (required for ip_tables) (NETFILTER_XTABLES) [M/y/?] m
Netfilter nf_tables support (NF_TABLES) [N/m] (NEW) m
  Netfilter nf_tables payload module (NFT_PAYLOAD) [N/m] (NEW) m
  Netfilter nf_tables IPv6 exthdr module (NFT_EXTHDR) [N/m] (NEW) m
[...]

PR #3696, opened 2020-06-30
Enable nf_tables support for aarch64

ghost:
nftables has already been enabled on armv7 (see https://github.com/raspberrypi/linux/issues/2177). this PR enables nftables for the aarch64 architecture.


pelwell:
In fact those options are already enabled in the Pi 4 defconfig (bcm2711_defconfig), which should work on a Pi 3 as well, but I'm happy to add them to bcmrpi3_defconfig as well.

Issue #3694, opened 2020-06-28
vc4-fkms-v3d: problems with interlaced modes


DarkAyron:
The vc4 driver doesn't handle interlaced modes correctly. When an interlaced video mode is set in config.txt, as soon as the vc4 module is loaded, output switches to progressive.

To reproduce the bug add the following to config.txt:
hdmi_group=1
hdmi_mode=20
hdmi_drive=2
hdmi_pixel_encoding=3
dtoverlay=vc4-fkms-v3d

reboot and watch the output of the HDMI port. The Pi starts with 1080i and as soon as the vc4 kernel module loads, the output changes to 1080p. The same thing happens with mode 5, 6, 7, 21 and 22 but changing to 480p, 576p or 1080p respectively.

Especially when driving a CRT not supporting 1080p or only SD, this causes it to display garbage or nothing at all (and in the worst case it ends with a puff of smoke and a broken horizontal deflection driver).

This was tested on an Raspberry Pi 3B+ and 4B. Both behave the same.
Image used is release 2020-05-27
Kernel version 4.19.118-v7+ #1311

6by9:
Interlacing is a non-trivial one from the configuration perspective as to field order, offsets, and the like. AFAICT the kernel frameworks don't give you that information and rely on the driver working it out for itself by magic. There isn't a generic answer for those settings (eg PAL is upper field first, but NTSC is lower field first), and varies based on output format as well.

Without having an easy solution on that, it was quicker to disable interlaced modes through FKMS until time has been found to resolve the differences. It's a matter of priorities.

Interlaced modes are supported via composite output as there the mode is fixed by the firmware, and the kernel just follows what is set in config.txt.

6by9:
I had a quick look and it turned out simpler than anticipated.
PR up for the kernel change, but there's also a firmware update required.

bellum07:
Unfortunately the patch doesn’t work here with a Raspberry Pi 2B and a Sony TV.
I’ve upgraded my Raspberry Pi OS lite to the latest version with Kernel:
   Linux raspberrypi 5.4.51-v7+ #1326 SMP Fri Jul 17 10:45:25 BST 2020 armv7l GNU/Linux
and added
   hdmi_group=1
   hdmi_mode=20

Without FKMS this works and the TV displays 1080i.
   root@raspberrypi:~# tvservice -sn
[...]

PR #3698, opened 2020-06-30
drm/vc4: Allow interlaced HDMI modes from FKMS.

6by9:
Having checked the firmware handling for interlaced modes,
it appears to be possible to support interlaced modes on
HDMI without adverse side effects, so do so.

https://github.com/raspberrypi/linux/issues/3694


6by9:
Needs the matching firmware change (MR 887)

PR #3704, opened 2020-07-03
FKMS: max refresh rate and blocking 1366x768

6by9:


drivers/gpu/drm/vc4/vc4_firmware_kms.c
```diff
@@ -1097,6 +1099,17 @@ vc4_crtc_mode_valid(struct drm_crtc *crtc, const struct drm_display_mode *mode)
...
+	/* Pi4 can't generate odd horizontal timings on HDMI, so reject modes
+	 * that would set them.
+	 */
+	if (fkms->bcm2711 &&
+	    (vc4_crtc->display_number == 2 || vc4_crtc->display_number == 7) &&
+	    ((mode->hdisplay & 1) ||			    /* active */
+	     ((mode->hsync_start - mode->hdisplay) & 1) ||  /* front porch */
+	     ((mode->hsync_end - mode->hsync_start) & 1) || /* sync pulse */
+	     ((mode->htotal - mode->hsync_end) & 1)))       /* back porch */
+		return MODE_H_ILLEGAL;
```
popcornmix, line 1111: `(A|B|C|D)&1` should generate better code, but it's arguable which is clearer.

6by9: I've just done exactly that in the firmware patch. 

It's not a path taken that often, so optimising a few ARM instructions out seems a little excessive. I'll let @pelwell make the call.

pelwell:
Can you split the DTS change into a separate commit? DT files are regularly squashed down to a single downstream DT commit, and having to split this is just more work for @popcornmix.

drivers/gpu/drm/vc4/vc4_firmware_kms.c
pelwell: You could compromise and just change the `||`s to `|`s, which might well generate equally good code.

6by9: That sounds less readable.
I've gone for the ```(A|B|C|D)&1```

pelwell: Opinions, etc.

6by9:
DT split out, using ```(A|B|C|D)&1```, and cleaned up the includes now as I found out I'd included <linux/module.h> that was already included as "linux/module.h".

drivers/gpu/drm/vc4/vc4_firmware_kms.c
```diff
@@ -1097,6 +1098,17 @@ vc4_crtc_mode_valid(struct drm_crtc *crtc, const struct drm_display_mode *mode)
 		break;
 	}
 
+	/* Pi4 can't generate odd horizontal timings on HDMI, so reject modes
+	 * that would set them.
+	 */
+	if (fkms->bcm2711 &&
+	    (vc4_crtc->display_number == 2 || vc4_crtc->display_number == 7) &&
+	    ((mode->hdisplay |				/* active */
+	     (mode->hsync_start - mode->hdisplay) |	/* front porch */
```
pelwell: Sorry to be a stickler, but this line and the following few would benefit from an extra preceding space - everything else is so beautifully indented.

6by9: You can go off a person you know!

pelwell: Only if you like them in the first place.

PR #3709, opened 2020-07-06
OV9281 driver and overlay

6by9:
Pulls in a driver for OV9281 from Rockchip's kernel tree, and then fixes it up to match mainline.

As it stands the first 3 commits won't build, but also the module isn't in the defconfig. Patch 4 fixes up the driver. The patches could be squashed but I really wanted to leave the history of where it came from.

~Just noted that I've missed adding it to the defconfig, but also the patch for the Vision Components board doesn't work as the ID comes back as 0x72FF, so I'm reworking that.~


6by9:
@naushir  @davidplowman Just FYI.

6by9:
defconfigs and reading the chip ID fixed.
Naush - if I you fancied a quick review it would be appreciated.

drivers/media/i2c/ov9281.c
```diff
@@ -0,0 +1,1053 @@
...
+	{0x5d01, 0x00},
+	{REG_NULL, 0x00},
+};
+
+static const struct ov9281_mode supported_modes[] = {
+	{
+		.width = 1280,
+		.height = 800,
+		.exp_def = 0x0320,
+		.hts_def = 0x05b0,//0x2d8*2
```
naushir: C style comment

drivers/media/i2c/ov9281.c
```diff
@@ -0,0 +1,1053 @@
...
+		*v4l2_subdev_get_try_format(sd, cfg, fmt->pad) = fmt->format;
+	} else {
+		ov9281->cur_mode = mode;
+		h_blank = mode->hts_def - mode->width;
+		__v4l2_ctrl_modify_range(ov9281->hblank, h_blank,
+					 h_blank, 1, h_blank);
+		vblank_def = mode->vts_def - mode->height;
+		__v4l2_ctrl_modify_range(ov9281->vblank, vblank_def,
+					 OV9281_VTS_MAX - mode->height,
+					 1, vblank_def);
```
naushir, line 406: Call __v4l2_ctrl_s_ctrl() needed here to apply the new vblank_def setting if it has changed?

drivers/media/i2c/ov9281.c
```diff
@@ -0,0 +1,1053 @@
...
+
+	return 0;
+}
+
+static int ov9281_enum_mbus_code(struct v4l2_subdev *sd,
+				 struct v4l2_subdev_pad_config *cfg,
+				 struct v4l2_subdev_mbus_code_enum *code)
+{
+	if (code->index != 0)
+		return -EINVAL;
```
naushir, line 450: if (!code->index) following kernel style?

drivers/media/i2c/ov9281.c
```diff
@@ -0,0 +1,1053 @@
...
+			pm_runtime_put(&client->dev);
+			goto unlock_and_return;
+		}
+	} else {
+		__ov9281_stop_stream(ov9281);
+		pm_runtime_put(&client->dev);
+	}
+
+	ov9281->streaming = on;
+
```
naushir, line 573: Better as ov9281->streaming = !!on as streaming is a bool.

drivers/media/i2c/ov9281.c
6by9: Done

6by9: Oddly checkpatch didn't object to it.

naushir (commented):
Given this driver is imported from another tree, you may not want to address these minor things.

6by9:
> Given this driver is imported from another tree, you may not want to address these minor things.

I have fixups for general API usage, so fixups on style can go in there too.

drivers/media/i2c/ov9281.c
6by9: Done (actually as ```if (code->index)```)

drivers/media/i2c/ov9281.c
6by9: Not sure on this one. Is the expectation that the current value should change on a mode change?

I've the same funny on imx290/327 that the ranges change, and if the current value is still valid within the new range then it's not reset to default.
Is the expected behaviour actually defined anywhere? (Silly question, I know).

naushir: My personal opinion is that it should, as we are changing operating modes so this parameter should be reset to the mode default.  imx219 and imx477 do this and (for the case of imx219), upstream folks did not question it.

Having said that, in our libcamera stack, it will not matter as we do (will - when it gets merged) set vblank explicitly before start stream.  So if you feel this should not be changed, I am ok with that.

6by9: Seems reasonable, but it's a mismash upstream as to whether they update or not.

It does raise the question again of whether exposure should be recomputed or reset seeing as changing mode likely changes the line length/time, so the actual exposure time changes. It really is a horrid bit of userspace API!

6by9:
More fixups required - I ran v4l2-compliance against it and got 8 failures. Colorspace mainly, and some streaming ioctl ones.

6by9:
Updated. Hopefully thaat is all the changes required for libcamera to be happy too.

naushir:
LGTM

pelwell:
I've not seen anything wrong with it.

PR #3714, opened 2020-07-07
OV7251 overlay and defconfig

6by9:
Overlay and defconfig change to enable OV7251.
Requested on https://www.raspberrypi.org/forums/viewtopic.php?f=28&t=277885
I don't have any hardware to test with, but hacking the driver not to try any of the I2C reads, it does probe and register with V4L2 properly.


arch/arm/boot/dts/overlays/ov7251-overlay.dts
```diff
@@ -0,0 +1,111 @@
...
+
+	fragment@1 {
+		target = <&csi1>;
+		__overlay__ {
+			status = "okay";
+
+			port {
+				csi1_ep: endpoint {
+					remote-endpoint = <&ov7251_0>;
+					data-lanes = <1>;
```
pelwell, line 55: Some of the camera overlays give the `csi` endpoint a `clock-lanes` property, while others don't. What's the logic behind the choice?

6by9: ```clock-lanes``` ought to be there for completeness, but isn't mandatory in our case.
Some SoCs (and a very few sensors) allow reassigning of CSI pair to function, so you might get
```
clock-lanes = <4>
data-lanes = <0 2 1 3>
```
if someone had been particularly perverse in their wiring.

Unicam doesn't support reordering, so ```clock-lane``` will always be <0> (assumed and never checked), and the only thing that matters in ```data-lanes``` is the number of entries. I did put in ```clock-lane``` for completeness on a couple of overlays, and probably ought to here too.

The other parameter that might be worth specifying is the ```bus-type``` (CSI-2 DPHY vs CCP2 vs parallel, etc). Upstream are moving towards pushing this explicitly into DT instead of deducing it from the other parameters that are specified on the endpoint. I may blitz them all if I get particularly bored.

6by9:
It's been confirmed on https://www.raspberrypi.org/forums/viewtopic.php?f=43&t=277885 that the overlay loads and they can stream images from the sensor, so I'm good with this.

PR #3715, opened 2020-07-08
staging: vc04_services: isp: Rework lens shading to take a dmabuf

6by9:
This removes the need for the client to use vcsm at all.


6by9:
@davidplowman @naushir Needs the corresponding libcamera changes and is an ABI breakage.

drivers/staging/vc04_services/bcm2835-isp/bcm2835-v4l2-isp.c
```diff
@@ -722,10 +724,38 @@ static int bcm2835_isp_s_ctrl(struct v4l2_ctrl *ctrl)
...
+		struct dma_buf *dmabuf;
+		void *vcsm_handle;
+
+		memcpy(&ls, ctrl->p_new.p_u8,
+		       sizeof(struct bcm2835_isp_lens_shading));
+
+		dmabuf = dma_buf_get(ls.dmabuf);
+		if (!dmabuf)
+			return -EINVAL;
+
```
JamesH65, line 738: Should this error return be ENOMEM ? No idea what the correct thingy would be.

JamesH65: Or is it a allocated handle already? Hmm, think it probably is.

6by9: This is taking a dmabuf allocated by something else (most likely a dmaheap), passing it in to the ISP driver, and then the ISP driver doing the vcsm calls to import it into the VPU heap.
As such the handle is a parameter into this control, so EINVAL is a sensible return value if it isn't a suitable dmabuf.

naushir:
Change looks good to me.  However, we must sync with the libcamera changes as this will be an ABI breakage that we must advertise.

pelwell:
Let me know how and when you want this merged.

naushir:
Will have to do it next week I'm afraid.  I'm just finishing off some other things, then will start testing the libcamera changes needed for this.

pelwell:
Rebased and merged offline.

kamomil:
Hi, I see that branch rpi-5.8.y have this change, but branch rpi-5.9.y does not

6by9:
> Hi, I see that branch rpi-5.8.y have this change, but branch rpi-5.9.y does not

Mainline 5.9 release includes a significant rework to VCHI which stops vcsm-cma (a dependency of bcm2835-codec and bcm2835-isp) from working. 
It's being worked on, and I'll be making sure that everything is supported on 5.9 when done.

kamomil:
Hi, thanks for the fast response, I tried to use the driver on rpi-5.8.y but I get the error:
```
bcm2835-isp bcm2835-isp: bcm2835_isp_probe: ril.isp returned 1 i/p (1 expected), 2 o/p (3 expected) ports
```
which originate from this part:
https://github.com/raspberrypi/linux/blob/rpi-5.8.y/drivers/staging/vc04_services/bcm2835-isp/bcm2835-v4l2-isp.c#L1662

You have any idea why does it happen and how can I track it back?
I see that `outputs` is set here:
https://github.com/raspberrypi/linux/blob/rpi-5.8.y/drivers/staging/vc04_services/vchiq-mmal/mmal-vchiq.c#L1099
but how/where is `rmsg`  get filled ?

Thanks

kamomil:
hi, I found out the problem, I had to udpate my firmware to fit the new kernel with `rpi-update`

PR #3724, opened 2020-07-15
vc4: Report channel mapping back to userspace

popcornmix:
This follows logic in hdmi-codec.c to use speaker layout
from ELD to choose a suitable speaker mapping based on
number of channels requested and signal that in audio
infoframe  and report this back to userspace.

This allows apps like speaker-test and kodi to get the
output to the right speakers.

### DIFF
--- a/drivers/gpu/drm/vc4/vc4_hdmi.c
+++ b/drivers/gpu/drm/vc4/vc4_hdmi.c
@@ -48,6 +48,7 @@
 #include <sound/pcm_drm_eld.h>
 #include <sound/pcm_params.h>
 #include <sound/soc.h>
+#include <sound/tlv.h>
 #include "media/cec.h"
 #include "vc4_drv.h"
 #include "vc4_hdmi.h"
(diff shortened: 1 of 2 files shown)


popcornmix:
Currently only briefly tested, so don't merge for a bit.
@HiassofT comments welcome.

drivers/gpu/drm/vc4/vc4_hdmi.c
```diff
@@ -82,6 +83,311 @@
 #define CEC_CLOCK_FREQ 40000
 #define VC4_HSM_CLOCK 163682864
 
+#define HDMI_CODEC_CHMAP_IDX_UNKNOWN  -1
+
+/*
+ * CEA speaker placement for HDMI 1.4:
+ *
+ *  FL  FLC   FC   FRC   FR   FRW
```
ghost, line 91: FRW?

popcornmix: From: https://elixir.bootlin.com/linux/v5.4.51/source/sound/soc/codecs/hdmi-codec.c#L31
I've checked CEA-861 where the other channels come from and it's not there. Also doesn't appear in enum below.

HiassofT:
I think it might be worth to use the Intel HDA driver's channel mapping code as a base https://github.com/raspberrypi/linux/blob/rpi-5.4.y/sound/hda/hdmi_chmap.c

Compared to the hdmi-codec's implementation it's more feature complete, contains HDMI2.0 / CTA-861-F definitions (inluding FRW :-) and implements variable channel mapping.

IIRC vc4-hdmi should support arbitrary (ALSA-)input to (HDMI-)output mappings so that would be a nice feature to have (OFC the vc4-hdmi specific mapping code would need to be implemented).

This would allow applications to also use more exotic configurations (eg CA 0x03 FL/FR/LFE/FC).

popcornmix:
Arbitrary channel mapping is trivial - just jumble the nibbles here:
https://github.com/raspberrypi/linux/blob/fc7fa9b035fcbb75be4b31f24a17609fa1ed272c/drivers/gpu/drm/vc4/vc4_hdmi.c#L1111

Is your suggestion to follow the logic in hdmi_chmap.c methods or actually call into it's code? I feel I should be calling into `snd_hdac_add_chmap_ctls` but I'm not quite sure where (and if) this should be plumbed in.

HiassofT:
calling into the hdmi_chmap code will be a bit tricky, it's coupled rather tightly to the HDA driver structure and we'd need o fake a hdac_device so we can call snd_hdac_register_chmap_ops to get the channel_allocations table initialized...

So it's easier to just follow the logic, copy tables and/or code from there for now.

Channel mapping code may be a candidate for refactoring, as it's duplicated in various incarnations in several kernel drivers. Might be easiest to do that with hdmi-codec's code, add in the feature the HDA code has and decouple it from hdmi-codec but that may need discussion with upstream folks how they'd like to have the structures, APIs etc.

popcornmix:
I think copying code and getting the desired behaviour is fine for now. This will likely be upstreamed by someone far more knowledgeable (@mripard) who can decide if duplication is best, or if refactoring makes sense.
I wouldn't really want to be responsible for breaking many other platforms I don't use...

HiassofT:
Yes, this is also what I think. First get it feature complete, then check if there's some potential for refactoring.

I think the HDA code is pretty close to what's needed for vc4, but we might need or want to do a few things a bit differently.

Could well be that these few bits are enough to make refactoring tricky.

As the channel mapping code basically is just a few tables plus some rather straight forward code it wouldn't be too bad if it's duplicated and adapted where needed.

mripard:
I guess we could make it fairly non-intrusive by just making snd_soc_dai_driver, and then registering the component ourselves instead of our codec.

The other HDMI drivers rely on an external i2s controller and a dedicated sound card, but we can't really do that in our driver, so I guess that would be the easier path forward.

popcornmix:
I've spend a few hours looking at this and I'd prefer to leave this PR as it is. It's had positive reports from forum users and does what 99% of users want.

I suspect @mripard's suggestion is the right approach, but for me, knowing very little about audio drivers I wouldn't know where to start. I did spend a while trying to switch over to hdmi_chmap.c but it's far from trivial and feels like it's not likely to be the proper solution.

mripard:
Yeah, I didn't mean to slow down the merge of that PR, just to point out how we could have something that shares that code eventually.

popcornmix:
@mripard I'd be very happy if you could get a shared version (ideally based on the better featured Intel HDA driver) into upstream so we can drop this version.

This just allows users to get multichannel audio from the right speakers in the meantime.

mripard:
It just got added to my todo list then :)

PR #3725, opened 2020-07-16
drm/vc4: Add DRM_MODE_FLAG_DBLCLK support to vc4-fkms

6by9:
480i and several other modes use DRM_MODE_FLAG_DBLCLK and pixel
replication.

Add in flags for that so that FKMS can select CEA modes 6 & 7.


6by9:
Needs the paired firmware change.

alanbork:
480i (mode 6) still doesn't work right - every pixel is still 2x high, testing rpi-update 70598414a73afab7f8e521c358a7cfd5ffb65d3e

PR #3731, opened 2020-07-17
pi: add skip-init flag to disable-bt overlay

fwjavox:
If this flag is not set, u-boot will fail while trying to (re)init the
uart devices when the disable-bt overlay is used.

Signed-off-by: Florian Wickert <fw@ferncast.de>

### DIFF
--- a/arch/arm/boot/dts/overlays/disable-bt-overlay.dts
+++ b/arch/arm/boot/dts/overlays/disable-bt-overlay.dts
@@ -17,6 +17,7 @@
 		target = <&uart1>;
 		__overlay__ {
 			status = "disabled";
+			skip-init;
 		};
 	};
 
@@ -26,6 +27,7 @@
 			pinctrl-names = "default";
 			pinctrl-0 = <&uart0_pins>;
 			status = "okay";
+			skip-init;
 		};
 	};
 


pelwell:
I'm not against this  PR in principle, but it needs more explanation of how U-boot fails and why this is the right solution. In particular, I have a few questions:
1. Why is the `skip-init` flag needed for a disabled UART?
2. Using `dtoverlay=disable-bt` on a Bluetooth-equipped Pi should leave the UARTs in the same state as they would be on a non-Bluetooth Pi, so why haven't we been asked to add `skip-init` before?

fwjavox:
The `skip-init` Flag is evaluated by u-boot and helps to prevent double initialization of the hardware.
1. `uart0` is the more important part because `uart0` is enabled by default. But the line `dtparam=uart1=on` in `config.txt` is enough to trigger the same problem for `uart1`.
2. This bug only applies to RPi4 because u-boot treats it very differently. There is no u-boot specific device-tree for the RPi4 anymore (while RPi<=3 do have them), so it reads the already existing device-tree from memory. Before this, u-boot did not care about `config.txt` and the overlays at all.

pelwell:
Is there any situation where you would want U-boot to initialise either of the UARTs? If not, it would be cleaner to patch the Pi 4 dtb.

fwjavox:
For the older RPIs a similar fix was applied to the u-boot internal device-tree: https://lists.denx.de/pipermail/u-boot/2017-April/285606.html

fwjavox:
> Is there any situation where you would want U-boot to initialise either of the UARTs? If not, it would be cleaner to patch the Pi 4 dtb.

Good question, I cannot think of any reason.

pelwell:
See #3733 - it would be great if you could test it to confirm it has the desired effect.

pelwell:
#3733 is now merged.

PR #3733, opened 2020-07-17
ARM: dts: Add UART skip-init properties for U-boot

pelwell:
U-boot can get stuck trying to initialise UARTs that aren't mapped
to the pin header. There is no reason for U-boot not to rely on the
initialisation by the firmware, so tag both UARTs with the u-boot
magic boolean property "skip-init".

See: https://github.com/raspberrypi/linux/pull/3731
     https://lists.denx.de/pipermail/u-boot/2017-April/285606.html

### DIFF
--- a/arch/arm/boot/dts/bcm270x-rpi.dtsi
+++ b/arch/arm/boot/dts/bcm270x-rpi.dtsi
@@ -102,6 +102,14 @@
 	};
 };
 
+&uart0 {
+	skip-init;
+};
+
+&uart1 {
+	skip-init;
+};
+
 &txp {
 	status = "disabled";
 };


fwjavox:
Looks good to me. I cannot test right now but will do as soon as possible.

pelwell:
@fwjavox Did you get a chance to test this patch?

fwjavox:
Looks good. The Pi boots and I can see the `skip-init` flags in the device-tree.
```
# find /proc/device-tree/ -name "skip-init"
/proc/device-tree/soc/serial@7e215040/skip-init
/proc/device-tree/soc/serial@7e201000/skip-init
```

pelwell:
Thanks.

PR #3739, opened 2020-07-20
extend table check for PCM rate

spockfish:


pelwell:
This PR changes a Linux file that is currently unmodified in our downstream kernels. The file isn't Pi-specific, and it doesn't change the code generated - there's no reason for us to add this to the collection of downstream patches we maintain.

You can try submitting your patch to the upstream kernel developers, but they will want at least a few words about what you have changed and why.

PR #3744, opened 2020-07-21
rpi-5.8.y forward-ports

pelwell:
These are the head commits from rpi-5.7.y that missed the cut, and a fix of a DTS merge funny.


popcornmix:
Thanks - still builds and boots.

PR #3752, opened 2020-07-24
Revert "drm/vc4: Allocate binner bo when starting to use the V3D"

EwanRoycroft:
This reverts commit 35c8b4b2c0fed637d2b30a8ec8e5d7f4c19c8d9d.

Fixes issues with minigbm on 5.4.

### DIFF
--- a/drivers/gpu/drm/vc4/vc4_bo.c
+++ b/drivers/gpu/drm/vc4/vc4_bo.c
@@ -793,35 +793,12 @@ vc4_prime_import_sg_table(struct drm_device *dev,
 	return obj;
 }
 
-static int vc4_grab_bin_bo(struct vc4_dev *vc4, struct vc4_file *vc4file)
-{
-	int ret;
-
-	if (!vc4->v3d)
-		return -ENODEV;
-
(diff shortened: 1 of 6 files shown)


popcornmix:
This is an upstream commit, so this should be submitted there with evidence as to why it is needed.
If it is agreed to be correct to revert this by the maintainers then we can pick this up.
https://www.kernel.org/doc/html/v5.4/process/submitting-patches.html

6by9:
It's one I'm investigating. This comes from https://www.raspberrypi.org/forums/viewtopic.php?f=67&t=280864 and https://www.raspberrypi.org/forums/viewtopic.php?f=67&t=280639

It looks like mainline have taken DRM_IOCTL_VC4_CREATE_BO to be solely for binner buffer objects, and rejects them if there is no v3d under vc4 (ie not Pi0-3), but I don't believe that is necessarily the case.

One of the X/mesa EGL import paths seems to be trying to allocate the binner BO from vc4, even though it will be consumed by v3d. Not sure if that is a Mesa failing or kernel driver.

Issue #3665, opened 2020-06-11
Garbled screen when running an OpenGLESv2 app on X11  and fkms without compositor with 5.4 kernel


imbens:
**Describe the bug**
This has been discussed in [https://github.com/raspberrypi/firmware/issues/1382](https://github.com/raspberrypi/firmware/issues/1382) in a follow up to an unrelated camera issue. I now have more precise steps to reproduce the problem. It happens only when using kernel 5.4 (it does happen with the latest 32bit image when you rpi-update to the 5.4 kernel) and fkms and X11 without lightdm (or with lightdm but with the compositor disabled) with a fullscreen Xwindow and with vsync enabled (via eglSwapInterval (display, 1)).

**To reproduce**
Start from the 64bit image from https://downloads.raspberrypi.org/raspios_arm64/images/raspios_arm64-2020-05-28/2020-05-27-raspios-buster-arm64.zip
Boot the image, it will boot to the desktop.

In a terminal or via ssh enter (this fixes the raspberrypi-dev package):
```
sudo apt update
sudo apt -q -y full-upgrade
```
At this point you should probably reboot.
In a terminal or via ssh enter:
```
sudo apt -q -y install libegl1-mesa-dev libgles2-mesa-dev libx11-dev xterm
git clone git@github.com:imbens/opengltest
cd opengltest
gcc opengltest.c -lGLESv2 -lEGL -lX11 -o opengltest
export DISPLAY=:0 && ./opengltest
```
EDIT: fixed the link to the git repository above
The application will show a red rectangle on a background that cycles between green and blue.
In a terminal or via ssh enter (this will disable lightdm):
`sudo systemctl set-default multi-user.target`
Reboot. The pi will now boot to a console.
In the console or via ssh enter:
`sudo xinit &`
This will start plain X11 without a desktop manager and it will open an xterm.
In xterm or via ssh enter:
```
cd opengltest
export DISPLAY=:0 && ./opengltest
```
The application will show a garbled red rectangle.
Edit opengltest.c and change `eglSwapInterval (mEGLDisplay, 1);` to `eglSwapInterval (mEGLDisplay, 0);` (this will disable vsync).
In xterm or via ssh enter:
```
gcc opengltest.c -lGLESv2 -lEGL -lX11 -o opengltest
export DISPLAY=:0 && ./opengltest
```
Now the rectangle is ok again.

**Expected behaviour**
I expect a red rectangle on a background that cycles between green and blue.

**Actual behaviour**
[...]

PR #3762, opened 2020-07-28
drm/vc4: Remove UIF from the list of modifiers returned by format_mod…

6by9:
…_supported

FKMS was listing UIF in the supported modifiers from format_mod_supported
when actually the pipeline doesn't support it. X was then choosing to
use it, and that then failed to render.

Remove references to UIF.

https://github.com/raspberrypi/linux/issues/3665


6by9:
@pelwell @popcornmix This one should fix the corrupted screens when not using X's compositor, as noted in #3665

PR #3772, opened 2020-08-02
arm64: bcm2711_defconfig: additional scalar and NEON crypto accelerat…

KrisztianKende:
…ions

ghash-ce and crct10dif-ce modules loads the NEON drivers instead of using Crypto Extensions


popcornmix:
Have you tested this to see what the performance improvement is?

KrisztianKende:
I think the Linux kernel uses the [sha512-armv8.pl](https://github.com/openssl/openssl/blob/master/crypto/sha/asm/sha512-armv8.pl) from OpenSSL ([arch/arm64/crypto/sha512-armv8.pl](https://github.com/raspberrypi/linux/blob/rpi-5.4.y/arch/arm64/crypto/sha512-armv8.pl)), [more precisely an earlier version](https://github.com/openssl/openssl/blob/32bbb62ea634239e7cb91d6450ba23517082bab6/crypto/sha/asm/sha512-armv8.pl). There are some results in this file [and here](https://patchwork.kernel.org/patch/9423059/). We can expect some performance increase compared to the generic variants.

popcornmix:
Have you built a kernel with these options and confirmed a use case that shows the benefits?

KrisztianKende:
Yes, I've built the kernel with these extra options.
I'm not really sure where the benefits will be. But the stock RPi OS loads the _sha256_generic_ and _libsha256_ modules by default. So something uses the kernel's SHA256 crypto capabilities.
With my custom kernel, I see that additional modules are loaded: _sha256_arm64_, _crct10dif_ce_ and _crct10dif_common_. So if there is any benefit, then something takes advantage of it.
The _sha256-arm64-neon_ have priority 150, compared to _sha256-generic_ which is 100.

Note: the [arch/arm64/crypto/sha512-core.S_shipped](https://github.com/raspberrypi/linux/blob/rpi-5.4.y/arch/arm64/crypto/sha512-core.S_shipped) and [sha256-core.S_shipped](https://github.com/raspberrypi/linux/blob/rpi-5.4.y/arch/arm64/crypto/sha256-core.S_shipped) files can be regenerated from the latest [sha512-armv8.pl](https://github.com/openssl/openssl/blob/master/crypto/sha/asm/sha512-armv8.pl). See [the Makefile](https://github.com/raspberrypi/linux/blob/rpi-5.4.y/arch/arm64/crypto/Makefile) _(ifdef REGENERATE_ARM64_CRYPTO_).

JamesH65:
SSH uses crypto and there have been reports of slowdowns in file transfers because of it. So might be worth doing some testing to see if your code improves performance. Checkout the forums for various posts on crypto performance.

KrisztianKende:
Ok, that's understandable.
I've done some file transfer tests with _scp_, but i don't see any performance difference:
Stock kernel:
_testfile1  100% 4058MB  76.8MB/s   00:52
testfile2  100% 4058MB  72.6MB/s   00:55
testfile3  100% 4058MB  77.2MB/s   00:52_
Custom kernel:
_testfile1  100% 4058MB  73.1MB/s   00:55
testfile2  100% 4058MB  76.1MB/s   00:53
testfile3  100% 4058MB  75.5MB/s   00:53_

It's maybe because the server->client and client->server cipher is chacha20-poly1305, and the SHA256 can only be used for Key Exchange.

_aes128-ctr: 35,9712 MB/s
aes192-ctr: 32,0513 MB/s
aes256-ctr: 28,169 MB/s
aes128-gcm@openssh.com: 31,1526 MB/s
aes256-gcm@openssh.com: 24,5098 MB/s
chacha20-poly1305@openssh.com: 81,9672 MB/s_

So the performance advantage is not that significant.

popcornmix:
Perhaps someone will suggest a use case where the benefit is more obvious.
We're happy to consider this assuming we can find a use case where it shows a benefit and the overhead (ram/sdcard) is reasonable.

hvenev:
AFAIK OpenSSH does not use the kernel crypto API, so any difference in performance should be noise.

popcornmix:
Do you know what might show an improvement?

PR #3773, opened 2020-08-02
Fixes for the Northwest Digital Radio DRAWS board on the 5.4 kernel

mcdermj:
The upstream Linux developers decided that the ads1015 hwmon driver was redundant, since there was a largely duplicative driver in the Industrial IO subsystem as well.  They removed this driver advising people to use the iio_hwmon bridge to provide equivalent functionality.

These changes rework things so that the NWDR DRAWS board continues to function under the 5.4 kernel releases.  This requires that the iio_hwmon driver is included in the default kernel, which it is currently not.  Also there is an overlay change for draws to activate the iio_hwmon driver and wire it to the iio device already created in the i2c bus.


pelwell:
Your PRs usually require few changes, and this was no different - I just added the 64-bit platforms and tweaked the whitespace (git shows TABs after spaces in red).

PR #3776, opened 2020-08-06
Rename README to README.md

ashwinrase:
File README renamed with README.md, to get visibility of syntax highlighting markdown


popcornmix:
This is an upstream file untouched by us, so the patch should be directed there.
https://www.kernel.org/doc/html/v5.4/process/submitting-patches.html

ashwinrase:
I tried but could not change it due to following restriction

image

popcornmix:
You need to read the linked submitting patches documentation. They don't use github for merge requests.

ashwinrase:
OK, thank you for the information

PR #3780, opened 2020-08-07
dts: bcm2711: Disable DVP by default

timg236:
The HDMI DVP should be disabled by default as is the case for other
display related drivers. This changes resolves an issue when using
the legacy firmware display driver where the DVP caused the 108 MHz
clock in HDMI TX to be gated off when Linux started. This effectively
stopped the firmware from being able to change the HDMI analog PHY
registers.

### DIFF
--- a/arch/arm/boot/dts/bcm2711-rpi.dtsi
+++ b/arch/arm/boot/dts/bcm2711-rpi.dtsi
@@ -46,6 +46,7 @@
 			clocks = <&clk_108MHz>;
 			#clock-cells = <1>;
 			#reset-cells = <1>;
+			status = "disabled";
 		};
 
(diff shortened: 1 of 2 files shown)


timg236:
@6by9

6by9:
https://github.com/raspberrypi/linux/blob/rpi-5.4.y/arch/arm/boot/dts/overlays/vc4-kms-v3d-pi4-overlay.dts#L141 needs to gain an extra fragment to enable it again.

6by9:
LGTM.
I know @pelwell 's off, but this would benefit from getting his blessing too.

timg236:
It can probably wait for a few days - I can work around the issues locally but we probably want to include this in the next significant rpi-update release

Issue #3325, opened 2019-11-08
Pi 4 - bcm2835-codec - Driver leaks buffers because frame completion might be too aggressive


nemosupremo:
**Describe the bug**
I have a Pi 4 running [WPE Webkit](https://wpewebkit.org/) v0.22.5 on Raspbian Buster. I am playing videos, in a loop, on the HTML5 browser. WPE Webkit uses gstreamer internally which uses the v4l2 stack to play hardware accelerated video. Every time the video stops running, the driver leaks 2 buffers with the message `bcm2835-codec: bcm2835_codec_stop_streaming: Timeout waiting for buffers to be returned - 178 outstanding`. Eventually when there are ~6000 buffers outstanding, the device becomes unstable and refuses to play anymore video. Kernel timeouts also start to happen. 

**To reproduce**
There is no straightforward way to reproduce the issue without building WPEWebkit / Cog / Weston, which takes hours. I tried to reproduce the issue with `gst-play-1.0`, but I wasn't able to reproduce it.

**Expected behaviour**
The buffers are not leaked.

**Actual behaviour**
Buffers are leaked

**System**

* Which model of Raspberry Pi? Pi 4
* Which OS and version (`cat /etc/rpi-issue`)?
```
Raspberry Pi reference 2019-09-26
Generated using pi-gen, https://github.com/RPi-Distro/pi-gen, 80d486687ea77d31fc3fc13cf3a2f8b464e129be, stage2
```
* Which firmware version (`vcgencmd version`)?  
```
Nov  1 2019 14:39:41 
Copyright (c) 2012 Broadcom
version 2faac616a157856a1d205f3a9713a7130defa643 (clean) (release) (start)
```
* Which kernel version (`uname -a`)? `Linux raspberry 4.19.81-v7l+ #1276 SMP Fri Nov 1 14:23:48 GMT 2019 armv7l GNU/Linux`

**Logs**
```
[12231.713321] WARNING*: CPU: 3 PID: 672 at drivers/media/common/videobuf2/videobuf2-core.c:1688 __vb2_queue_cancel+0x158/0x1c8 \[videobuf2_common]*

[...]

PR #3790, opened 2020-08-13
bcm2835_codec fixes

6by9:
These are from investigating seeking not working with gst-play. It does now.

### DIFF
--- a/drivers/staging/vc04_services/bcm2835-codec/bcm2835-v4l2-codec.c
+++ b/drivers/staging/vc04_services/bcm2835-codec/bcm2835-v4l2-codec.c
@@ -2320,10 +2320,7 @@ static void bcm2835_codec_stop_streaming(struct vb2_queue *q)
 	struct bcm2835_codec_q_data *q_data = get_q_data(ctx, q->type);
 	struct vchiq_mmal_port *port = get_port_data(ctx, q->type);
 	struct vb2_v4l2_buffer *vbuf;
-	struct vb2_v4l2_buffer *vb2;
-	struct v4l2_m2m_buffer *m2m;
-	struct m2m_mmal_buffer *buf;
-	int ret, i;
+	int ret;
 
 	v4l2_dbg(1, debug, &ctx->dev->v4l2_dev, "%s: type: %d - return buffers\n",
 		 __func__, q->type);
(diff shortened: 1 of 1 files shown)


6by9:
@popcornmix I don't know whether these will fix FFmpeg/Kodi as well.

The second patch is working around an issue in GStreamer that has been fixed in 1.16, but I believe is still valid within the V4L2 API, so we ought to support.

popcornmix:
I can test with ffmpeg/kodi.
Do you believe header bytes need to be fed in again after a seek? (i.e. have we flushed decoder in such a way it will have forgotten previous ones)

6by9:
Not sure - I didn't analyse the packet that GStreamer fed back in after the seek.

We have done a full codec shutdown on the VPU, however video_decode updates an internal structure which may be enough to restore the codec state after the seek.

6by9:
If FFmpeg doesn't work, then the way I read the [V4L2 M2M spec](https://www.kernel.org/doc/html/latest/userspace-api/media/v4l/dev-decoder.html#seek), as long as it follows the specified procedure then the codec is responsible for remembering the header bytes. That's based on the warning regarding H264 and seeking over a change of SPS/PPS - that wouldn't be necessary if the codec was expected to drop the header bytes.

GStreamer shuts down both OUTPUT and CAPTURE queues on seek, therefore I'd view that as reasonable to forget everything.

6by9:
Still waiting on tenterhooks to know if worked with Kodi!

popcornmix:
Had some build issues and ended up having to do a clean build.
There's a kodi build on Pi4 and an LE cross-compile in progress. Hopefully one will work.

popcornmix:
Got an LE build running, but no joy. H264 video plays but I don't get any video frames out after a seek.

popcornmix:
Not sure if there's any useful info here, but after a seek, there are complaints in dmesg (not immediately but after a few seconds).
http://ix.io/2uh3

timemaster5:
Hmm, unfortunately, the problem persists. I tested on Raspberry pi 3b+ with youtube playing 1080p video on wpewebkit-fdo platform. Tested on kernel 5.4.59.

moritzvieli:
I don't understand enough details in this topic, but this patch fixed the seeking issue for me: https://github.com/raspberrypi/linux/issues/3325#issuecomment-684040830

timemaster5:
I believe this issue is also related to this one https://github.com/raspberrypi/linux/issues/3791, where I posted the full dmesg

PR #3794, opened 2020-08-13
tpreston/pifacedigital overlay

tompreston:
Hi, this mostly works. I can control the PiFace Digital from userland using the new(ish) char dev interface and also monitor interrupts. Just submitting WIP for review while I fix up a few other things.

You can test the device tree on its own - the kernel changes are not required. In fact, I haven't tested them - the only thing blocking me tonight is building and installing a new kernel to test.

Still to do:
- The mcp23s17 driver requires the `interrupt-controller` dt property before it enables interrupts, which I think is wrong. The interrupt feature should be gated on the `interrupts` or `interrupt-parent` properties. The parent gpio-controller is the `interrupt-controller`, not the mcp23s17. Is that right?
- Enable multiple devices. The PiFace Digital has jumper switches to change SPI device. We should be able to configure this in the device tree. I don't like how the mcp23s17 overlay handles this, it's too complicated imo.

Thanks for looking!
Tom

### DIFF
--- a/arch/arm/boot/dts/overlays/Makefile
+++ b/arch/arm/boot/dts/overlays/Makefile
@@ -120,6 +120,7 @@ dtbo-$(CONFIG_ARCH_BCM2835) += \
 	ov9281.dtbo \
 	papirus.dtbo \
 	pibell.dtbo \
+	pifacedigital.dtbo \
 	piglow.dtbo \
 	piscreen.dtbo \
 	piscreen2r.dtbo \
--- a/arch/arm/boot/dts/overlays/README
+++ b/arch/arm/boot/dts/overlays/README
@@ -1913,6 +1913,14 @@ Params: alsaname                Set the name as it appears in ALSA (default
                                 "PiBell")
 
 
+Name:   pifacedigital
+Info:   Configures the PiFace Digital mcp23s17 GPIO port expander.
+Load:   dtoverlay=pifacedigital,<param>=<val>
+Params: spi-present-mask        8-bit integer, bitmap indicating MCP23S17 SPI0
+                                CS0 address. PiFace Digital supports addresses
+                                0-3, which can be configured with JP1 and JP2.
+
+
 Name:   piglow
 Info:   Configures the PiGlow by pimoroni.com
 Load:   dtoverlay=piglow
--- a/arch/arm/boot/dts/overlays/pifacedigital-overlay.dts
+++ b/arch/arm/boot/dts/overlays/pifacedigital-overlay.dts
@@ -1,0 +1,144 @@
+// SPDX-License-Identifier: GPL-2.0-only
+/*
+ * PiFace Digital, Device Tree Overlay.
+ * Copyright (C) 2020 Thomas Preston <thomas.preston@codethink.co.uk>
+ *
+ * The PiFace Digital is a convenient breakout board for the Microchip mcp23s17
+ * SPI GPIO port expander.
+ *
+ * The first eight GPIOs 0..7 (bank A) are connected to eight output terminals
+ * and LEDs, plus two relays on the first two outputs. These output loads are
+ * active-high.
+ *
+ * The next eight GPIOs 8..15 (bank B) are connected to eight input terminals
+ * with four on-board switches connecting them to ground. Inputs devices are
+ * therefore expected to bridge terminals to ground, so the mcp23s17 pullups are
+ * activated for GPIO bank B.
+ *
+ * On PiFace Digital, the mcp23s17 is connected to the Raspberry Pi's SPI0 CS0
+ * bus. Each SPI bus supports up to eight addressable child devices. The PiFace
+ * Digital only supports addresses 0-4, which can be configured by jumpers JP1
+ * and JP2.
+ *
+ * You can tell the driver about these jumper configurations with the
+ * spi-present-mask bitmask:
+ *
+ *     | JP1 | JP2 | dtoverlay line in /boot/config.txt         |
+ *     | --- | --- | ------------------------------------------ |
+ *     |  0  |  0  | dtoverlay=pifacedigital                    |
+ *     |  0  |  0  | dtoverlay=pifacedigital:spi-present-mask=1 |
+ *     |  0  |  1  | dtoverlay=pifacedigital:spi-present-mask=2 |
+ *     |  1  |  0  | dtoverlay=pifacedigital:spi-present-mask=4 |
+ *     |  1  |  1  | dtoverlay=pifacedigital:spi-present-mask=8 |
+ *
+ * # Example
+ * Set the dtoverlay config in /boot/config.txt and power off the Raspberry Pi:
+ *
+ *     $ grep pifacedigital /boot/config.txt
+ *     dtoverlay=pifacedigital
+ *     $ sudo systemctl poweroff
+ *
+ * Attach the PiFace Digital and power on the Raspberry Pi.
+ * Then use the libgpiod tools to query the device:
+ *
+ *     $ sudo apt install gpiod
+ *     $ gpiodetect | grep mcp23s17
+ *     gpiochip2 [mcp23s17.0] (16 lines)
+ *
+ * Set GPIO outputs 0, 2 and 5:
+ *
+ *     $ gpioset gpiochip2 0=1 2=1 5=1
+ *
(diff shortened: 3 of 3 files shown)


tompreston:
I've submitted patches to linux-gpio, fixing up the mcp23s17 driver. Waiting for review.

Next steps:
- Respond to mcp23s17 review comments
- Add "interrupt-controller" back into pifacedigital overlay, with comment explaining that it should be removed after v5.8 (or similar)
- Figure out a way to handle multiple PiFace Digital boards. At least, changing the jumper pins for spi0.1, spi1.0, spi1.1. It's possible there's no nice way to do this and the best approach will be to just document how create a new overlay to do it.

pelwell:
I think your interrupt-controller changes are misguided. The mcp23s17 is a GPIO controller, and like all good GPIO controllers the driver can be a source of interrupts when GPIO levels change, as well as being a consumer of interrupts indicating that something on the device as changed.

tompreston:
Hey @pelwell, thanks for taking a look.

So you mean if another device was connected to the PiFace Digital and it had an interrupt line connected to GPIO 0 - then the pifacedigital node would need to be an interrupt-controller in order to chain the irq? That makes sense, but in that case wouldn't you inject the interrupt-controller property via the new device overlay. Like how we overlay pin 25 on `&gpio` as an interrupt [here](https://github.com/raspberrypi/linux/pull/3794/files#diff-1d21a527d3a5221fea14b807a9a3460bR128).

Either way, I think interrupts on the mcp23s17 driver should be enabled even if we're only a interrupt-client (on presence of `interrupts` property) and not rely on the presence of `interrupt-controller`, per say.

However, I do agree a generic gpio-controller should be an `interrupt-controller`.

What do you think of these next steps:
- Keep the [device_property_present(dev, "interrupts");](https://github.com/raspberrypi/linux/pull/3794/files#diff-d50ca4155847f102cfc23934359ffb01R778) change
- Undo the `interrupt-controller` documentation changes
- Add `interrupt-controller` back into PiFace Digital, with a comment explaining what it's for

tompreston:
Now that I think about it, the [Raspberry Pi gpio-controller](https://github.com/raspberrypi/linux/blob/rpi-5.4.y/arch/arm/boot/dts/bcm283x.dtsi#L113) is an interrupt-controller by default and so should this mcp23s17 device - ready to pass through any interrupts.

It's actually this device as an interrupt-client which is optional. I see it now!

Next steps:
- Revert interrupt-controller changes

tompreston:
Ok I've removed:
- `interrupt-controller` changes
- Cosmetic regmap changes to mcp23s08 driver (I'll go through upstream channels with this)

And added `interrupt-controller` back into pifacedigital-overlay.dts. I'll have a think about multiple devices over the weekend and un-WIP this. Thanks @pelwell!

pelwell:
The new `maxtherm` overlay (https://github.com/raspberrypi/linux/blob/rpi-5.4.y/arch/arm/boot/dts/overlays/maxtherm-overlay.dts) might give you some ideas about how to handle multiple CS lines and SPI interfaces in a neater way than the mcp23s17 overlay (which had no choice because some of the more interesting overlay parameter features weren't available when it was written).

tompreston:
Thanks @pelwell, one problem with that is that it doesn't support multiple simultaneous PiFace Digitals on different addresses. For example with the discontinued [PiFace Rack](https://www.newark.com/piface/pirack/pi-rack-raspberry-pi/dp/97W2503), or maybe a future PiFace Digital which matches the stacking HAT specification (speculating!) For which I'd need four different device nodes. I can't think of the right way to share the identical node properties. :confused:

To be honest, that's such an infrequent use case it's not worth complicating the overlay for. Those sorts of users could write their own overlay for example.

pelwell:
You could always use the magic `name` pseudo-property - assigning a value to it changes the name of the parent node.

tompreston:
Ok I've learned that the PiFace Digital jumpers are actually for setting the internal address on the same SPI0 CS0 bus. The pinctrl-mcp23s08.c driver learns about these addresses from device tree with the `microchip,spi-present-mask` and then the driver probes multiple gpiochips based on these. I've updated overlay to reflect this and tested the different jumper configurations work.

I don't have any way of testing more advanced configuration such as multiple PiFace Digital SPI0 CS0 addresses at the same time, or PiFace Digital on a different chip-select (with something like PiFace Rack). There exist no PiFace Digital configurations for different SPI lines altogether (to my knowledge).

I will un-WIP this PR now, I think it's ready to merge. Thanks for reviewing this @pelwell.

I also have one more, slightly unrelated question. Is it expected behaviour that the `gpioget` tool sets the GPIO pin direction to *input*? This means I cannot read GPIO output pins. For example:
```
pi@raspberrypi:~ $ gpioset gpiochip2 0=1

pi@raspberrypi:~ $ gpioinfo gpiochip2 | head -3
gpiochip2 - 16 lines:
        line   0:      unnamed       unused  output  active-high
        line   1:      unnamed       unused   input  active-high

pi@raspberrypi:~ $ gpioget gpiochip2 0
0

pi@raspberrypi:~ $ gpioinfo gpiochip2 | head -3
gpiochip2 - 16 lines:
        line   0:      unnamed       unused   input  active-high
        line   1:      unnamed       unused   input  active-high
```

I suspect this is just a limitation of the libgpiod example tools.

pelwell:
There are a few problems with the overlay:
1. The GPIO for the irq_pin is declared in the pinctrl node `pifacedigital_irq_pin`, but because it isn't mentioned in a `pinctrl-0` property it has no effect. Fortunately it isn't really needed - the driver will claim it as a GPIO anyway as a result of the `interrupts` property - but it can provide better error messages in the result of a clash. One of the examples in the bindings documentation (https://github.com/raspberrypi/linux/blob/rpi-5.4.y/Documentation/devicetree/bindings/pinctrl/pinctrl-mcp23s08.txt#L137) shows how this is done (although it doesn't include the irq pinctrl node itself).
2. Every overlay in this repo has an entry in the README file - you will need to add one, following the strict formatting rules (no TABs, no trailing spaces, indent to the same columns, two blank lines between entries.

pelwell:
The answer to your question is not straightforward. There is an explicit comment in the internals of the gpiolib kernel support that says:
```
	if (cmd == GPIOHANDLE_GET_LINE_VALUES_IOCTL) {
		/* NOTE: It's ok to read values of output lines. */
```
This would suggest that you are seeing a limitation of the `gpioget` utility, and to a certain extent you are - it's possible to use the libgpio IOCTL API to query a GPIO value without forcing it to be an input. However, the value of a GPIO after `gpioset` exits is undefined. In the case of the mcp23s17 driver the values appear to persist, but that shouldn't be relied on in general (and indeed it doesn't work with the onboard GPIO driver - it returns pins to inputs when they are released).

`gpioset` takes a mode parameter that allows GPIOs to the set for a specified interval or until the Return key is pressed, but doing so hogs the GPIO until `gpioset` exits, making it impossible to query the current value with `gpioget`. You could write a libgpio client application that sets a value and also reads back the value - but that only really makes sense in an open-drain use case.

tompreston:
Hey, updated thanks.

Now I can see a total number of interrupts and individual interrupts in /proc, which I don't think were there before.
```
$ cat /proc/interrupts
[snip]
 58:         37          0          0          0  pinctrl-bcm2835  25 Level     spi0.0
 59:          0          0          0         11    spi0.0   8 Edge
 60:          0          0          0          8    spi0.0   9 Edge
 61:          0          0          0         10    spi0.0  10 Edge
 62:          0          0          0          8    spi0.0  11 Edge
[snip]
```

pelwell:
Merged, after adding `,<param>=<val>` to the `Load:` section (the checker is _very_ picky).

tompreston:
Ah damn, is the checker in `scripts/`? I ran:

```
git format-patch -1
scripts/checkpatch.py 0001-dt-overlays-Add-PiFace-Digital-Device-Tree-Overlay.patch
```

Maybe you could add the checker script into CI somewhere? Or mention it instead of the [note](https://github.com/raspberrypi/linux/blob/rpi-5.4.y/arch/arm/boot/dts/overlays/README#L87).

Thank you for merging! Appreciate the review feedback too.

pelwell:
The script has too many dependencies to be easily distributed - dtmerge, ovmerge, cpp, dtc, dtdiff, ...

PR #3797, opened 2020-08-15
rtc: rv3028: Refresh RAM on EEPROM write

evading:
The active RV3028 settings are in RAM so after modifying the settings in
EEPROM the RAM should be refreshed so that they take effect.

See: #3616

Signed-off-by: Einar Vading <einar.vading@rhimagnesita.com>

### DIFF
--- a/drivers/rtc/rtc-rv3028.c
+++ b/drivers/rtc/rtc-rv3028.c
@@ -18,6 +18,7 @@
 #include <linux/of_device.h>
 #include <linux/regmap.h>
 #include <linux/rtc.h>
+//#include "rtc-core.h"
 
 #define RV3028_SEC			0x00
 #define RV3028_MIN			0x01
@@ -73,7 +74,7 @@
 
 #define RV3028_BACKUP_TCE		BIT(5)
 #define RV3028_BACKUP_TCR_MASK		GENMASK(1,0)
-#define RV3028_BACKUP_BSM_MASK		0x0C
+#define RV3028_BACKUP_BSM_MASK		GENMASK(3,2)
 
 #define OFFSET_STEP_PPT			953674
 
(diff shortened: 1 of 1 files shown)


drivers/rtc/rtc-rv3028.c
```diff
@@ -583,6 +584,59 @@ static int rv3028_eeprom_read(void *priv, unsigned int offset, void *val,
...
+	ret = regmap_read_poll_timeout(priv, RV3028_STATUS, status,
+				       !(status & RV3028_STATUS_EEBUSY),
+				       RV3028_EEBUSY_POLL,
+				       RV3028_EEBUSY_TIMEOUT);
+	if (ret)
+		goto restore_eerd;
+
+restore_eerd:
+	if (!(ctrl1 & RV3028_CTRL1_EERD))
+	{
```
pelwell: Kernel coding standard (which you have mostly stuck to) requires the brace to be on the previous line.

drivers/rtc/rtc-rv3028.c
```diff
@@ -722,6 +776,8 @@ static int rv3028_probe(struct i2c_client *client)
 					(void *)(rv3028->regmap),
 					RV3028_BACKUP, (void *)&backup, 1);
 			}
+			if (!ret)
```
pelwell: Is it necessary to refresh the RAM if the backup bits haven't changed?

drivers/rtc/rtc-rv3028.c
evading: Of course. This was a copy paste error from the `rv3028_eeprom_read` and I didn't run checkpatch. I'll fix.

drivers/rtc/rtc-rv3028.c
evading: Nope. That should go inside the brace above. I'll fix.

pelwell:
That's looking better, but I think you'll have to update your local repo's view of rpi-5.4.y and rebase those changes on top of it.

pelwell:
e.g.:
```
$ git fetch origin
$ git rebase --onto origin/rpi-5.4.y rv3028~2 rv3028
```

pelwell:
Sorry for the delay - GitHub rarely sends notifications about pushes, so it's wise to comment as well.

PR #3800, opened 2020-08-17
dtoverlays: Add overlay for the PCA953x family of GPIO expanders

6by9:
Adds an overlay for configuring all the GPIO expanders supported
by the driver under GPIO_PCA953X.

### DIFF
--- a/arch/arm/boot/dts/overlays/README
+++ b/arch/arm/boot/dts/overlays/README
@@ -1873,6 +1873,42 @@ Params: panel                   Display panel (required):
         speed                   Display SPI bus speed
 
 
+Name:   pca953x
+Info:   TI PCA953x family of I2C GPIO expanders. Default is for NXP PCA9534.
+Load:   dtoverlay=pca953x,<param>=<val>
+Params: addr                    I2C address of expander. Default 0x20.
+        pca6416                 Select the NXP PCA6416 (16 bit)
+        pca9505                 Select the NXP PCA9505 (40 bit)
+        pca9535                 Select the NXP PCA9535 (16 bit)
+        pca9536                 Select the NXP PCA9536 or TI PCA9536 (4 bit)
+        pca9537                 Select the NXP PCA9537 (4 bit)
+        pca9538                 Select the NXP PCA9538 (8 bit)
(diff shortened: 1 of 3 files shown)


arch/arm/boot/dts/overlays/README
```diff
@@ -1873,6 +1873,42 @@ Params: panel                   Display panel (required):
         speed                   Display SPI bus speed
 
 
+Name:   pca953x
+Info:   TI PCA953x family of I2C GPIO expanders. Default is for NXP PCA9534.
+Load:   dtoverlay=pca953x,<param>=<val>
+Params: addr                    I2C address of expander. Default 0x20.
+        pca6416                 Select the NXP PCA6416 (16 bit)
```
6by9, line 1880: @pelwell Is there meant to be a blank line between parameters or not? Some overlays do, some don't.

pelwell: I leave that to the author - longer descriptions would probably warrant a blank line, otherwise it's not necessary.

arch/arm/boot/dts/overlays/pca953x-overlay.dts
```diff
@@ -0,0 +1,240 @@
...
+		pca9575 = <0>, "+13";
+		pca9698 = <0>, "+14";
+		pca16416 = <0>, "+15";
+		pca16524 = <0>, "+16";
+		pca19555a = <0>, "+17";
+		max7310 = <0>, "+18";
+		max7312 = <0>, "+19";
+		max7313 = <0>, "+20";
+		max7315 = <0>, "+21";
+		pca6017 = <0>, "+22";
```
pelwell: pca6107?

arch/arm/boot/dts/overlays/README
6by9: I'll leave it without then.

arch/arm/boot/dts/overlays/pca953x-overlay.dts
6by9: I knew I'd make at least one typo in those stupid lists. Fixed.

PR #3806, opened 2020-08-19
kconfig fixes

hvenev:
Since 016baa59bf9f6c2550480b73f18100285e3a4fd2 (part of 5.8), MEDIA_CONTROLLER_REQUEST_API has to be selected rather than depended upon. Do the same with some other options.


hvenev:
None of these drivers are upstreamed yet, right?

popcornmix:
I don't believe any have been merged. @6by9 - happy with this?

6by9:
Looks reasonable.

I vaguely recall the patchsets changing MEDIA_CONTROLLER etc from depends to selects, but haven't been running 5.8 in anger yet.

popcornmix:
@pelwell okay with you?

PR #3813, opened 2020-08-24
staging: vc04_services: ISP: Fix dmabuf error check in S_CTRL

Rahi374:
In bcm2835_isp_s_ctrl, the error check for dma_buf_get() is incorrect,
and considers ERR_PTR pointers as valid dmabufs. Fix this error check.

Signed-off-by: Paul Elder <paul.elder@ideasonboard.com>


pelwell:
The patch looks good to me. @6by9 Are you OK with it returning `-EINVAL` rather than `PTR_ERR(dmabuf)`?

JamesH65:
There are some other places in our vc04_services staging folder where this same apparent mistake has been made.

6by9:
LGTM.
Yes, -EINVAL is fine as it's reporting that the s_ctrl fails due to invalid data.

It looks like there is a similar one in vc_sm.c.

bcm2835-v4l2-codec.c and bcm2835-v4l2-isp.c in buf_prepare on being given an imported dmabuf should check the return from dma_buf_get for being an error before continuing on to use the buffer.

pelwell:
Thanks!

PR #3815, opened 2020-08-25
Switch to upstream cpufreq driver for Pi 1

popcornmix:
This switches to upstream cpufreq driver for Pi 1. Downstream driver is now unused and can be dropped going forwards.

This also removes the "Only report integer pll divisor frequencies" commit which isn't required after latest rpi-update firmware which relaxes conditions on arm frequency. This is necessary on pi1 and typically the normal and overclock frequencies (e.g. 700 and 1000) are not both supported and get quantized to same frequency.

I believe having this PR without latest firmware is safe, but suboptimal as kernel will ask for frequencies not directly supported by firmware that will be rounded down. Max and Min frequencies should be okay, so probably not a major issue.

Marked as WIP while I test.


popcornmix:
Removed WIP. Tested on Pi1, Pi3 and Pi4.
Pi4 kernel change with older firmware is okay - I seem arm ask for frequencies spaced at 100MHz intervals and firmware rounds down to 1500/1000/750/600. With latest firmware and kernel you get the 100MHz intervals requested.

PR #3837, opened 2020-09-06
trace: add trace events for open(), exec() and uselib()

marchom:
This patch uses TRACE_EVENT to add tracepoints for the open(),
exec() and uselib() syscalls so that ureadahead can cheaply trace
the boot sequence to determine what to read to speed up the next.

It's not upstream because it will need to be rebased onto the syscall
trace events whenever that gets merged, and is a stop-gap.

Signed-off-by: Scott James Remnant <scott@ubuntu.com>

### DIFF
--- a/fs/exec.c
+++ b/fs/exec.c
@@ -68,6 +68,7 @@
 #include <asm/mmu_context.h>
 #include <asm/tlb.h>
 
+#include <trace/events/fs.h>
 #include <trace/events/task.h>
 #include "internal.h"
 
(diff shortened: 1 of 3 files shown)


pelwell:
Explain to me why we should merge these patches when we don't use ureadahead and Ubuntu doesn't even seem to support it on 20.04 and later?

marchom:
By all means, don't.
I'm running on 18.04 still and noticed this behavior in the boot up logs. I imagined I'm not the only one so here you go.
There's also an issue I opened so if people wonder what these errors are about- the info is there.

I understand that it's extra work having to keep track of diversions to upstream so if you see no value, just reject.

PR #3840, opened 2020-09-07
BCM2711 vc4-kms-v3d mega update

6by9:
Just a small PR ;-)

- Takes rpi-5.9.y
- reverts all the early version of the vc4-kms-v3d Pi4 support and FKMS
- applies the latest patch set (which has just been merged to drm-misc with one minor typo corrected)
- fixes up the downstream DT.
- adds the downstream fixes for CEC, HD audio, etc
- adds FKMS squashed into a single commit


popcornmix:
I've been testing this tree for last few days and it seems good. kms works with kodi, including hevc v4l2 and hd audio passthrough.
fkms seemed working too. 
kms desktop use was also working for me (and isn't currently on 5.4). So +1 from me.
It needs a rebase though...

6by9:
> It needs a rebase though...

But I only rebased it about 2 hours ago!!
Rebased now.

6by9:
@pelwell Could you check you're happy with the DT changes predominantly in
https://github.com/raspberrypi/linux/pull/3840/commits/e024e217f145c62f02ff4fd77d3fd911b8108b02
https://github.com/raspberrypi/linux/pull/3840/commits/dff2b5e4e37bfb8cbdde91e7af345a4af0721c56

I was a little surprised that the interrupt setup for some of the components was in the base bcm2711.dtsi as I thought those have previously been in the platform file, but those are part of Maxime's upstream changes.

Has upstream not merged any patches enabling V3D for Pi4? I was surprised to see it in bcm211-rpi.dtsi (our downstream file).

6by9:
Please hold fire on merging just for a day - popcornmix has noted that I've missed at least one patch from vc4_planes.c

6by9:
Updated.
I had missed three downstream patches for 10bit YUV support, selecting BT601/709 on planes, and adding fkms as a dma range.

I'm happy again now. (Please don't make me rebase it too many times!)

pelwell:
I'm looking at it now. @popcornmix - hold fire on the rebasing.

pelwell:
I have a slight reservation about the new hvs node being on the `soc` bus - `scb` seems like a more natural home given that HVS 5 can address more than 1GB.

6by9:
> scb seems like a more natural home given that HVS 5 can address more than 1GB.

It can? We only write 32bit addresses into the display list, so presumably that is still through the VPU cache aliases?

pelwell:
Nick F said it could, but the truth turns out to be more complicated - there are two paging options in the SYSAC block that allow either a single 2GB window or the 1GB window to be moved, which is too painful to consider. So `/soc` is the right home.

I want to change the way some of the downstream includes work, adding `#ifndef BCM2711` around some blocks rather than including and deleting them, but that doesn't belong here.

6by9:
> Nick F said it could,

Foolish me - I read the documentation! ;-)

Does this mean everyone's happy now? Can we merge?

pelwell:
GitHub is claiming there are conflicts when rebasing, which is odd.

pelwell:
Merged behind the scenes.

PR #3845, opened 2020-09-09
Cherry-pick Unicam and IMX477 patches

6by9:
These patches were apparently missed in the rebase. This does mean that the unicam patches are in a slightly different order to earlier branches, but that's not a real problem.

### DIFF
--- a/MAINTAINERS
+++ b/MAINTAINERS
@@ -16165,6 +16165,14 @@ S:	Maintained
 T:	git git://linuxtv.org/media_tree.git
 F:	drivers/media/i2c/imx355.c
 
+SONY IMX477 SENSOR DRIVER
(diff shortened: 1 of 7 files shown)


popcornmix:
I assume unicam is working? Fine by me.

6by9:
Yes, it was largely working anyway but with a couple of quirks due to the missing patches (I couldn't selected the unpacked bayer formats for example).
IMX477 being totally missing of course meant the HQ camera didn't work!

6by9:
We do appear to have lost CONFIG_VIDEO_IMX477=m from bcm2711_defconfig though. It's still in 2709 and rpi, and the 64bit configs.

popcornmix:
bcm2711_defconfig was regenerated with make defconfig, so if that option wasn't in tree at the time it would have been removed.

PR #3859, opened 2020-09-18
rpi-5.9.y - latest KMS patches from Maxime

6by9:
3 patches that have been sent to dri-devel.

With these patches I've been switching dual HDMI monitors around between resolutions without any failures :-)


6by9:
I am sending review comments to the mailing list over "drm/vc4: crtc: Keep the previously assigned HVS FIFO".
As long as we're happy to revert and reapply, the patch (dropping it in the rebase), then I'm happy to merge ahead of the response. 

It's a malloc failure error path and code a coding style issue, so not likely to cause grief.

popcornmix:
On this tree don't worry about reverts and reapplies if a commit changes in review. We can drop them in a rebase.

drivers/gpu/drm/vc4/vc4_crtc.c
```diff
@@ -852,11 +852,18 @@ void vc4_crtc_destroy_state(struct drm_crtc *crtc,
...
+
 	if (crtc->state)
 		vc4_crtc_destroy_state(crtc, crtc->state);
-	crtc->state = kzalloc(sizeof(struct vc4_crtc_state), GFP_KERNEL);
-	if (crtc->state)
-		__drm_atomic_helper_crtc_reset(crtc, crtc->state);
+
+	vc4_crtc_state = kzalloc(sizeof(*vc4_crtc_state), GFP_KERNEL);
+	if (!vc4_crtc_state)
+		return;
```
6by9, line 867: Just FYI, my comment sent to the list on this is as follows:

This error path has me worried, but I may have missed it.

If crtc->state was set then it's been freed via
vc4_crtc_destroy_state. However I don't see anything that has reset
crtc->state to NULL.
If the new alloc fails then I believe we exit with crtc->state still
set to the old freed pointer.

The old code directly set crtc->state, so it got reset to NULL by the failure.

(There are also a couple of checkpatch complaints)

6by9:
> On this tree don't worry about reverts and reapplies if a commit changes in review. We can drop them in a rebase.

I thought as much, but thought I ought to flag it.

popcornmix:
Feel free to merge this if you are happy. I've been running with it locally.

6by9:
Maxime's just pushed the changes for
- 10 & 12 bit output 
- filtering 1366x768
- avoiding wifi frequencies.

I'm testing those and may as well add them to this.

pelwell:
OK - kick me when they're ready.

6by9:
Extra commits pushed to this branch.
I'm rebuilding to test, although I don't believe I have any displays that can take 10 or 12bit input.

drivers/gpu/drm/vc4/vc4_hdmi.c
```diff
@@ -1128,6 +1189,11 @@ static int vc4_hdmi_encoder_atomic_check(struct drm_encoder *encoder,
 	if (mode->flags & DRM_MODE_FLAG_DBLCLK)
 		pixel_rate *= 2;
 
+	if (conn_state->max_bpc == 12)
+		pixel_rate *= 150 / 100;
```
HiassofT: this looks odd - doesn't 150/100 just evaluate to 1, so that line is a no-op?

same thing 2 lines below

6by9: Quite right.
I've just pulled them in off the dri-devel mailing list - they haven't been reviewed yet. Feel free to comment in your own right.

popcornmix: Yep...

6by9: I'm emailing Maxime with a heads up on this one.

HiassofT: thanks!

pelwell:
This PR is just waiting on a fix for the very rounded `*= a/b` scaling.

6by9:
> This PR is just waiting on a fix for the very rounded *= a/b scaling.

Yes, I was giving it a day to see if Maxime came back over it. There's also been a query over the wifi patch on dri-devel.

If I don't hear anything by close of play then I'll drop the 10/12bit and wifi patches and run with that for now.

stevenhoving:
@pelwell is it really the case that this PR is just waiting on a fix for something as trivial as this?

```
if (conn_state->max_bpc == 12)
	pixel_rate = (pixel_rate * 150) / 100;
else if (conn_state->max_bpc == 10)
	pixel_rate = (pixel_rate * 125) / 100;
```

Or this PR is waiting for a upstream fix for this problem; so waiting for Maxime to publish the upstream fix?

popcornmix:
Yes, the fix is trivial, but ideally we'd like to get the same fix as is merged upstream so we don't get conflicts when we rebase.

There's also the slight concern that any testing done didn't include the pixel clock boost so that should really be double checked.

stevenhoving:
> Yes, the fix is trivial, but ideally we'd like to get the same fix as is merged upstream so we don't get conflicts when we rebase.
> 
> There's also the slight concern that any testing done didn't include the pixel clock boost so that should really be double checked.

Thanks for clarifying

6by9:
PR updated to drop those patches. Hopefully the others won't change.

pelwell:
I'll take that chance.

PR #3863, opened 2020-09-21
rpivid_h265: Fix width/height typo

popcornmix:


popcornmix:
@jc-kynesim noticed this when scrolling through code.

pelwell:
Hmm - I think this patch needs a wider review...

PR #3890, opened 2020-10-08
Rpi 5.4.y: Enable Unicam and V4L2 test drivers

kbingham:
The arm64 kernel configs lack unicam driver support.
This means that the 64 bit raspberry-os does not support libcamera out of the box.

Enable the relevant entries in the Raspberry Pi defconfigs.


pelwell:
Thanks!

PR #3894, opened 2020-10-08
V4L2 stateful codec API for rpi-5.9.y

6by9:


6by9:
@popcornmix I believe vc-sm-cma is now doing the right thing after the upstream VCHI(Q) changes. Certainly I can play H264 via GStreamer fine.

I've got a set of patches to test out for @jc-kynesim to allow resolution changes without having to STREAMON on the CAPTURE (output) queue, but I need to be sure that it doesn't mess up encode or ISP. If they just work tomorrow then I'll add them to this, otherwise they can be added as a new PR later on.

Then to look at Kodi's seeking.

HiassofT:
I did a few tests with kodi/LibreELEC and it looks good so far, thanks!

popcornmix:
Looks good to me too. Behaves the same as 5.4 tree.

pelwell:
Do you want niggly formatting comments (because if it works for you and compiles cleanly that's about all you're likely to get from me)?

popcornmix:
Shouldn't you be adding CONFIG_VIDEO_CODEC_BCM2835 to config files?

6by9:
> Do you want niggly formatting comments (because if it works for you and compiles cleanly that's about all you're likely to get from me)?

Happy for you to review. It's effectively the same as on 5.4 but squashed, and vc-sm-cma needed a few tweaks due to the upstream removing of the VCHI shim layer and queue.
Now that it's working I can pass it off to the Ideas On Board folk to upstream it, although they're more interested in the ISP driver that I haven't pulled across yet to get libcamera running.

> Shouldn't you be adding CONFIG_VIDEO_CODEC_BCM2835 to config files?

That can be added as a further patch.

6by9:
So GStreamer and the Kodi seem happy with the STREAMON update, but I'd like it checked against other clients before merging.

@jc-kynesim Could you check to see if this makes Chromium happier?

Also added both codec drivers to the defconfigs

jc-kynesim:
@6by9 Those deltas are to 5.9. Can I just compile that and copy it over my 5.4 and hope that it works?

popcornmix:
In general switching kernel version is straightforward (as long as you remember to update modules and dtb's too).
You can put new kernel and dtb's in a subdirectory (e.g. test) of boot and add "os_prefix=test/" to config.txt if you want to switch back and forth.
https://www.raspberrypi.org/documentation/configuration/config-txt/boot.md

6by9:
Hmm, jc says no change in Chromium with the patch, so I've dropped it for now. We can always squash at a later date.

6by9:
Oh, do we need to include a link back to all the contributors to the codec driver on the rpi-5.4.y branch?

pelwell:
> do we need to include a link back to all the contributors to the codec driver on the rpi-5.4.y branch?

Pardon?

6by9:
https://github.com/raspberrypi/linux/commits/rpi-5.4.y/drivers/staging/vc04_services/bcm2835-codec

We've got 5 authors including myself. To squash them all down removes the detail of additional contributions. Do we need to retain that, or at least provide a link to it?

pelwell:
I think to some extent it depends on the size of the contribution, but it would be polite to at least list them in the squashed commit message.

6by9:
(I'm sure popcornmix always rebases the branch just after I've done a push!)
Rebased again, and codec commit text updated to include contributors.

popcornmix:
@6by9 is this ready to merge form your side?

6by9:
Yes, I'm happy with it.
I'll investigate the improvement for JC separately once I've brought the ISP driver in to get libcamera going.

PR #3895, opened 2020-10-10
bcm2835-dma: only reserve channel 0 if legacy dma driver is enabled

HiassofT:
If CONFIG_DMA_BCM2708 isn't enabled there's no need to mask out
one of the already scarce DMA channels.

### DIFF
--- a/drivers/dma/bcm2835-dma.c
+++ b/drivers/dma/bcm2835-dma.c
@@ -1286,6 +1286,7 @@ static int bcm2835_dma_probe(struct platform_device *pdev)
 		goto err_no_dma;
 	}
 
+#ifdef CONFIG_DMA_BCM2708
 	/* One channel is reserved for the legacy API */
 	if (chans_available & BCM2835_DMA_BULK_MASK) {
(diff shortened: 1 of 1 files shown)


HiassofT:
RPi is really short on DMA channels. Currently it's not possible to load the rpi-cirrus-wm5102 driver in conjunction with vc4-kms on RPi4 - vc4-kms needs 2 DMAs for audio (noaudio or audio(1)=off dtparam don't currently work, they break vc4 loading completely), and the soundcard driver needs both I2S with 2 DMA channels for RX and TX and SPI (which wants 2 DMAs as well).

Depending on the phase of the moon I2S might be probed first and everything's well, as SPI can fallback to non-DMA mode. Or we are not that lucky, SPI gets 2 DMAs and I2S and soundcard driver fail miserably because I2S really needs the 2 DMAs and can only get one of it...

Disabling CONFIG_DMA_BCM2708 and CONFIG_FB_BCM2708 is one possibility to free up a DMA, but probably the legacy DMA reserving code should be revisited so it's possible to configure DMA reserving dynamically

PR #3897, opened 2020-10-12
ISP to Rpi 5.9.y

6by9:
@naushir Please review the small amendments to the driver and ideally test. (I might see if I can get libcamera up quickly if easier)


6by9:
Ping @naushir seeing as I made a typo first time around.

naushir:
Will give it a run tomorrow morning with libcamera.

6by9:
Something funny using ToT qcam
```
pi@raspberrypi:~/libcamera/build/src/qcam $ DISPLAY=:0 ./qcam
[0:01:06.768317347] [1045]  INFO IPAManager ipa_manager.cpp:136 libcamera is not installed. Adding '/home/pi/libcamera/build/src/ipa' to the IPA search path
[0:01:06.833091551] [1045]  INFO Camera camera_manager.cpp:292 libcamera v0.0.0+1914-9183d218
[0:01:07.113385736] [1047]  INFO IPAProxy ipa_proxy.cpp:122 libcamera is not installed. Loading IPA configuration from '/home/pi/libcamera/src/ipa/raspberrypi/data'
[0:01:07.283246977] [1045]  INFO Camera camera.cpp:811 configuring streams: (0) 800x600-BGR888
[0:01:07.283934217] [1047]  INFO RPI raspberrypi.cpp:117 Mode: 1012x760 fmt RG10 Score: 1598.26 (best 1598.26)
[0:01:07.284077847] [1047]  INFO RPI raspberrypi.cpp:117 Mode: 4056x3040 fmt RG12 Score: 1924.33 (best 1598.26)
[0:01:07.284141180] [1047]  INFO RPI raspberrypi.cpp:117 Mode: 2028x1520 fmt RG12 Score: 1037.33 (best 1037.33)
[0:01:07.284202588] [1047]  INFO RPI raspberrypi.cpp:117 Mode: 2028x1080 fmt RG12 Score: 1131.17 (best 1037.33)
[0:01:07.284267847] [1047]  INFO RPI raspberrypi.cpp:117 Mode: 1012x760 fmt pRAA Score: 1098.26 (best 1037.33)
[0:01:07.284334106] [1047]  INFO RPI raspberrypi.cpp:117 Mode: 4056x3040 fmt pRCC Score: 1424.33 (best 1037.33)
[0:01:07.284393217] [1047]  INFO RPI raspberrypi.cpp:117 Mode: 2028x1520 fmt pRCC Score: 537.329 (best 537.329)
[0:01:07.284451347] [1047]  INFO RPI raspberrypi.cpp:117 Mode: 2028x1080 fmt pRCC Score: 631.167 (best 537.329)
[0:01:07.284538884] [1047]  INFO RPI raspberrypi.cpp:581 Sensor: /base/soc/i2c0mux/i2c@1/imx477@1a - Selected mode: 2028x1520-pRCC
[0:01:07.365307736] [1047]  INFO RPI_S_W staggered_ctrl.cpp:36 Init ctrl 0x00980911 with delay 2
[0:01:07.365432236] [1047]  INFO RPI_S_W staggered_ctrl.cpp:36 Init ctrl 0x009e0903 with delay 1
Zero-copy enabled
[0:01:07.400303847] [1047]  INFO RPISTREAM rpi_stream.cpp:120 No buffers available for ISP Output0
[0:01:07.400390199] [1047]  INFO RPISTREAM rpi_stream.cpp:120 No buffers available for ISP Output0
[0:01:07.956144365] [1051] ERROR IPARPI raspberrypi.cpp:906 Can't find CCM control
[0:01:07.956396754] [1051] ERROR IPARPI raspberrypi.cpp:1039 Can't find LS control
[0:01:07.956597865] [1051] ERROR IPARPI raspberrypi.cpp:927 Can't find Gamma control
[0:01:07.956776865] [1051] ERROR IPARPI raspberrypi.cpp:946 Can't find black level control
[0:01:07.956951865] [1051] ERROR IPARPI raspberrypi.cpp:964 Can't find geq control
[0:01:07.957254643] [1051] ERROR IPARPI raspberrypi.cpp:982 Can't find denoise control
[0:01:07.957470198] [1051] ERROR IPARPI raspberrypi.cpp:1002 Can't find sharpen control
[0:01:07.957708180] [1051] ERROR IPARPI raspberrypi.cpp:1023 Can't find DPC control
[0:01:07.959073884] [1047] ERROR V4L2 v4l2_videodevice.cpp:1441 /dev/video13[out]: Failed to queue buffer 0: Invalid argument
[0:01:07.959246087] [1047] ERROR RPISTREAM rpi_stream.cpp:229 Failed to queue buffer for ISP Input
```
```
pi@raspberrypi:~/libcamera/build/src/qcam $ v4l2-ctl -d 13 --list-ctrls

User Controls

                    red_balance 0x0098090e (int)    : min=1 max=65535 step=1 default=1000 value=2522 flags=slider
                   blue_balance 0x0098090f (int)    : min=1 max=65535 step=1 default=1000 value=1897 flags=slider
       colour_correction_matrix 0x009819e1 (u8)     : min=0 max=255 step=1 default=0 [88] flags=has-payload
                   lens_shading 0x009819e2 (u8)     : min=0 max=255 step=1 default=0 [36] flags=has-payload, execute-on-write
                    black_level 0x009819e3 (u8)     : min=0 max=255 step=1 default=0 [12] flags=has-payload
             green_equalisation 0x009819e4 (u8)     : min=0 max=255 step=1 default=0 [16] flags=has-payload
                          gamma 0x009819e5 (u8)     : min=0 max=255 step=1 default=0 [136] flags=has-payload
                        denoise 0x009819e6 (u8)     : min=0 max=255 step=1 default=0 [24] flags=has-payload
                        sharpen 0x009819e7 (u8)     : min=0 max=255 step=1 default=0 [28] flags=has-payload
     defective_pixel_correction 0x009819e8 (u8)     : min=0 max=255 step=1 default=0 [8] flags=has-payload

Image Processing Controls

                   digital_gain 0x009f0905 (int)    : min=1 max=65535 step=1 default=1000 value=1000
```
I've checked with yavta and that is passing images through fine. That's using MMAL for the ISP though, so not a truly fair comparison.
Drop back to 5.4.68 and I get the same issue.

naushir:
Could be an enum value mismatch between the libcamera uapi vs kernel uapi header?

6by9:
I'll leave that for you to determine.
Are you and @davidplowman not running against something based off libcamera master then? If so then I'd expect you to see the same thing.

naushir:
We are both running of libcamera master, so am a bit surprised at this result.

6by9:
Hang on, I think I hadn't switched kernel to 5.4 when I thought I had. That's just worked on 5.4.68.

I wonder if something has changed in media controller and it's not enumerating as things expect. I'll dig a little further.

6by9:
Ah, we also have a couple of slightly ugly dependency issues. include/uapi/linux/bcm2835-isp.h had already been merged when it needs to be part of the ISP commit - time for a couple of reverts and reapplications.

6by9:
> Could be an enum value mismatch between the libcamera uapi vs kernel uapi header?

Good guess for half the problem - someone had updated the patch on 5.9 to amend V4L2_CID_USER_BCM2835_ISP_BASE from ```(V4L2_CID_USER_BASE + 0x10c0)``` to ```(V4L2_CID_USER_BASE + 0x10e0)``` as ```V4L2_CID_USER_ATMEL_ISC_BASE``` takes +0x10c0 for 32.
This is going to be messy to handle!

6by9:
Found it - ```vb2_fill_vb2_v4l2_buffer``` is failing the call to ```__verify_length```
Looking at the logging
```
[   52.479025] video13: VIDIOC_QBUF: error -22: 00:00:52.000438124 index=0, type=vid-out, request_fd=0, flags=0x00000000, field=none, sequence=0, memory=dmabuf, bytesused=4669440, offset/userptr=0x20, length=0
```
We've used 4669440 of a buffer that is of length 0. Oops.

```
diff --git a/src/libcamera/v4l2_videodevice.cpp b/src/libcamera/v4l2_videodevice.cpp
index 508522e..5512f4f 100644
--- a/src/libcamera/v4l2_videodevice.cpp
+++ b/src/libcamera/v4l2_videodevice.cpp
@@ -1427,6 +1427,8 @@ int V4L2VideoDevice::queueBuffer(FrameBuffer *buffer)
                } else {
                        if (metadata.planes.size())
                                buf.bytesused = metadata.planes[0].bytesused;
+
+                       buf.length = buffer->planes()[0].length;
                }
 
                buf.sequence = metadata.sequence;
```
Fix that and the changed ```V4L2_CID_USER_BCM2835_ISP_BASE```, and libcamera runs.

Can I leave it with one of you to get the patch merged to libcamera?

drivers/staging/vc04_services/bcm2835-isp/bcm2835-isp-fmts.h
```diff
@@ -0,0 +1,388 @@
...
+#define BCM2835_ISP_FMTS
+
+#include <linux/videodev2.h>
+#include "vchiq-mmal/mmal-encodings.h"
+
+struct bcm2835_isp_fmt {
+	u32 fourcc;
+	int depth;
+	int bytesperline_align;
+	u32 flags;
```
6by9: flags appears to be always 0. Suspect left over from the COMPRESSED flag in the codec code.

Needs removing from ```bcm2835_isp_node_enum_fmt``` as well.

drivers/staging/vc04_services/bcm2835-isp/bcm2835-isp-fmts.h
```diff
@@ -0,0 +1,388 @@
...
+		.fourcc		    = V4L2_PIX_FMT_BGR24,
+		.depth		    = 24,
+		.bytesperline_align = 32,
+		.flags		    = 0,
+		.mmal_fmt	    = MMAL_ENCODING_BGR24,
+		.size_multiplier_x2 = 2,
+		.colorspace	    = V4L2_COLORSPACE_SRGB,
+		.step_size	    = 1,
+	}, {
+		.fourcc		    = V4L2_PIX_FMT_ABGR32,
```
6by9: I've just had this dilemma with the codec & camera (https://github.com/raspberrypi/linux/commit/076f6ac5e4f5ab7e96e3d9a1a402c0de896eb7ca).

ABGR32 or V4L2_PIX_FMT_XBGR32. We don't have full control over the alpha byte (ISP writes either 0 or 255), so surely it's better to specify it as undefined.

drivers/staging/vc04_services/bcm2835-isp/bcm2835-v4l2-isp.c
```diff
@@ -0,0 +1,1708 @@
...
+#define DEFAULT_DIM 1280U
+/*
+ * Maximum frame dimension of 16384 pixels.  Even though the ISP runs in tiles,
+ * have a sensible limit so that we do not create an excessive number of tiles
+ * to process.
+ */
+#define MAX_DIM 16384U
+/*
+ * Minimum frame dimension of 64 pixels.  Anything lower, and the tiling
+ * algorihtm may not be able to cope when applying filter context.
```
6by9: Typo - algorihtm

drivers/staging/vc04_services/bcm2835-isp/bcm2835-v4l2-isp.c
```diff
@@ -0,0 +1,1708 @@
...
+	struct media_intf_devnode *intf_devnode;
+	struct media_link *intf_link;
+	struct mutex lock; /* top level device node lock */
+	struct mutex queue_lock;
+
+	struct vb2_queue queue;
+	unsigned int sequence;
+
+	/* The list of formats supported on the node. */
+	struct bcm2835_isp_fmt_list supported_fmts;
```
6by9: I think it's probably cleaner to insert the *list and num_entries elements directly in here.

The codec code from which you've inherited this had an array of 2 for the input and output queues. You don't as they're individual nodes.

drivers/staging/vc04_services/bcm2835-isp/bcm2835-v4l2-isp.c
```diff
@@ -0,0 +1,1708 @@
...
+
+	struct bcm2835_isp_q_data q_data;
+
+	/* Parent device structure */
+	struct bcm2835_isp_dev *dev;
+
+	bool registered;
+	bool media_node_registered;
+	bool queue_init;
```
6by9: Having both registered and queue_init feels a bit ugly. If you fail part way through ```register_node```, then clean up before exit.
```media_node_registered``` is something different, so it's odd having it in the middle of the 3.

File: drivers/staging/vc04_services/bcm2835-isp/bcm2835-v4l2-isp.c [hunk 6]
```diff
@@ -0,0 +1,1708 @@
...
+	struct vchiq_mmal_component *component;
+	struct completion frame_cmplt;
+
+	struct bcm2835_isp_node node[BCM2835_ISP_NUM_NODES];
+	struct media_pad pad[BCM2835_ISP_NUM_NODES];
+	atomic_t num_streaming;
+
+	/* Image pipeline controls. */
+	int r_gain;
+	int b_gain;
```
6by9, line 132: I think we already had the discussion about whether it's better to store local copies of these values, handles to the ctrl_handlers, or use the search functions. See what others think.

drivers/staging/vc04_services/bcm2835-isp/bcm2835-v4l2-isp.c
```diff
@@ -0,0 +1,1708 @@
...
+
+	if (!dev->component)
+		return NULL;
+
+	switch (node->queue.type) {
+	case V4L2_BUF_TYPE_VIDEO_OUTPUT:
+		return &dev->component->input[node->id];
+	case V4L2_BUF_TYPE_VIDEO_CAPTURE:
+	case V4L2_BUF_TYPE_META_CAPTURE:
+		return &dev->component->output[node->id];
```
6by9: Are we better off adding a ```*port``` to ```struct bcm2835_isp_node``` that is initialised on component create to point to the correct port? It saves a bit of messing around here on every buffer passed or parameter set.

drivers/staging/vc04_services/bcm2835-isp/bcm2835-v4l2-isp.c
```diff
@@ -0,0 +1,1708 @@
...
+		v4l2_err(&dev->v4l2_dev,
+			 "%s: Failed disabling %s port, ret %d\n", __func__,
+			 node_is_output(node) ? "i/p" : "o/p",
+			 ret);
+
+	while (atomic_read(&port->buffers_with_vpu)) {
+		v4l2_dbg(1, debug, &dev->v4l2_dev,
+			 "%s: Waiting for buffers to be returned - %d outstanding\n",
+			 __func__, atomic_read(&port->buffers_with_vpu));
+		ret = wait_for_completion_timeout(&dev->frame_cmplt, HZ);
```
6by9: Ought to use a define here instead of just ```HZ```.

drivers/staging/vc04_services/bcm2835-isp/bcm2835-v4l2-isp.c
```diff
@@ -0,0 +1,1708 @@
...
+		ret = vc_sm_cma_import_dmabuf(dmabuf,
+					      &vcsm_handle);
+		if (ret) {
+			dma_buf_put(dmabuf);
+			return -EINVAL;
+		}
+
+		ls.dmabuf = vc_sm_cma_int_handle(vcsm_handle);
+		if (ls.dmabuf)
+			ret = set_isp_param(node,
```
6by9, line 736: Probably worth a comment that setting this parameter causes the VPU to hold a reference to the resource, and vcsm won't release the dmabuf until that reference is released.

drivers/staging/vc04_services/bcm2835-isp/bcm2835-v4l2-isp.c
```diff
@@ -0,0 +1,1708 @@
...
+		q_data->height = 0;
+		q_data->bytesperline = 0;
+		q_data->sizeimage = f->fmt.meta.buffersize;
+	}
+
+	v4l2_dbg(1, debug, &dev->v4l2_dev,
+		 "%s: Calculated bpl as %u, size %u\n", __func__,
+		 q_data->bytesperline, q_data->sizeimage);
+
+	/* If we have a component then setup the port as well */
```
6by9: Have we lost a conditional?

I think this is legacy from the codec code where we had a component per instance rather than one total. The comment is therefore obsolete.

drivers/staging/vc04_services/bcm2835-isp/bcm2835-isp-fmts.h
```diff
@@ -0,0 +1,388 @@
...
+		.step_size	    = 2,
+	}, {
+		.fourcc		    = V4L2_PIX_FMT_YUYV,
+		.depth		    = 16,
+		.bytesperline_align = 64,
+		.flags		    = 0,
+		.mmal_fmt	    = MMAL_ENCODING_YUYV,
+		.size_multiplier_x2 = 2,
+		.colorspace	    = V4L2_COLORSPACE_SMPTE170M,
+		.step_size	    = 2,
```
6by9, line 68: Is there any need/gain in having an ```h_step_size = 2, v_step_size = 1``` here? YUV422 is only horizontally subsampled.

drivers/staging/vc04_services/bcm2835-isp/bcm2835-v4l2-isp.c
```diff
@@ -0,0 +1,1708 @@
...
+		fmt = get_default_format(node);
+
+	if (!node_is_stats(node)) {
+		f->fmt.pix.width = max(min(f->fmt.pix.width, MAX_DIM),
+				       MIN_DIM);
+		f->fmt.pix.height = max(min(f->fmt.pix.height, MAX_DIM),
+					MIN_DIM);
+
+		f->fmt.pix.pixelformat = fmt->fourcc;
+		f->fmt.pix.colorspace = fmt->colorspace;
```
6by9, line 972: Do we need to support setting the correct matrices for 601/709 and full/limited range? I know we've had the discussions previously over the output config, but this also potentially applies to the input side if feeding YUV in. (Same discussion therefore applies for bcm2835-codec).

drivers/staging/vc04_services/bcm2835-isp/bcm2835-v4l2-isp.c
```diff
@@ -0,0 +1,1708 @@
...
+	.vidioc_subscribe_event		= bcm3285_isp_subscribe_event,
+	.vidioc_unsubscribe_event	= v4l2_event_unsubscribe,
+};
+
+/*
+ * Size of the array to provide to the VPU when asking for the list of supported
+ * formats.
+ *
+ * The ISP component currently advertises 44 input formats, so add a small
+ * overhead on that.
```
6by9: Extend the comment. "Too many formats will only result in some not being advertised."

drivers/staging/vc04_services/bcm2835-isp/bcm2835-v4l2-isp.c
```diff
@@ -0,0 +1,1708 @@
...
+		v4l2_info(&dev->v4l2_dev, "vb2_queue_init failed\n");
+		return ret;
+	}
+	node->queue_init = true;
+
+	/* Set some controls and defaults, but only on the VIDEO_OUTPUT node. */
+	if (node_is_output(node)) {
+		unsigned int i;
+
+		/* Use this ctrl template to assign all out ISP custom ctrls. */
```
6by9: Typo "all **our** ISP custom ctrls."

drivers/staging/vc04_services/bcm2835-isp/bcm2835-v4l2-isp.c
```diff
@@ -0,0 +1,1708 @@
...
+		struct v4l2_ctrl_config ctrl_template = {
+			.ops		= &bcm2835_isp_ctrl_ops,
+			.type		= V4L2_CTRL_TYPE_U8,
+			.def		= 0,
+			.min		= 0x00,
+			.max		= 0xff,
+			.step		= 1,
+		};
+
+		ret = v4l2_ctrl_handler_init(&dev->ctrl_handler, 12);
```
6by9: I only count 11 controls.

How about ```3 + ARRAY_SIZE(custom_ctrls)``` (3 being the number of standard ctrls).

drivers/staging/vc04_services/bcm2835-isp/bcm2835-v4l2-isp.c
```diff
@@ -0,0 +1,1708 @@
...
+	node->registered = true;
+	video_set_drvdata(vfd, node);
+
+	v4l2_info(&dev->v4l2_dev,
+		  "Device node %s[%d] registered as /dev/video%d\n",
+		  node->name, node->id, vfd->num);
+
+	return 0;
+
+ctrl_cleanup:
```
6by9, line 1389: Further argument for the funny of having ```node->queue_init``` and ```node->registered``` when you have some cleanup code for the ctrl_handler here.

The cleanup code is also wrong as it always cleans up the handler, but it was only created ```if (node_is_output(node))```.

drivers/staging/vc04_services/bcm2835-isp/bcm2835-v4l2-isp.c
```diff
@@ -0,0 +1,1708 @@
...
+	int ret;
+
+	v4l2_info(&dev->v4l2_dev,
+		  "Register %s node %d with media controller\n",
+		  output ? "output" : "capture", num);
+	entity->obj_type = MEDIA_ENTITY_TYPE_VIDEO_DEVICE;
+	entity->function = MEDIA_ENT_F_IO_V4L;
+	entity->info.dev.major = VIDEO_MAJOR;
+	entity->info.dev.minor = node->vfd.minor;
+	name = kmalloc(BCM2835_ISP_ENTITY_NAME_LEN, GFP_KERNEL);
```
6by9, line 1474: Is there a reason for not making this a managed resource?

drivers/staging/vc04_services/bcm2835-isp/bcm2835-v4l2-isp.c
```diff
@@ -0,0 +1,1708 @@
...
+	strscpy(dev->mdev.model, "bcm2835-isp",
+		sizeof(dev->mdev.model));
+	strscpy(dev->mdev.bus_info, "platform:bcm2835-isp",
+		sizeof(dev->mdev.bus_info));
+	media_device_init(&dev->mdev);
+	dev->v4l2_dev.mdev = &dev->mdev;
+
+	v4l2_dbg(2, debug, &dev->v4l2_dev, "Register entity for nodes\n");
+
+	name = kmalloc(BCM2835_ISP_ENTITY_NAME_LEN, GFP_KERNEL);
```
6by9, line 1555: Is there a reason for not making this a managed resource?

drivers/staging/vc04_services/vchiq-mmal/mmal-parameters.h
```diff
@@ -786,7 +842,102 @@ struct mmal_parameter_camera_info {
...
+	u32 enabled;
+	u32 offset;
+	struct mmal_parameter_rational slope;
+};
+
+#define MMAL_NUM_GAMMA_PTS 33
+struct mmal_parameter_gamma {
+	u32 enabled;
+	u16 x[MMAL_NUM_GAMMA_PTS];
+	u16 y[MMAL_NUM_GAMMA_PTS];
```
6by9, line 910: Having 33 points in each array means that y is not aligned and you have a potential slight abiguity in packing.

6by9 (commented):
Reviewing the file as a whole looking for things that may get raised in upstreaming.

Please raise any objections otherwise I'll go ahead and make the changes.

naushir:
> Found it - `vb2_fill_vb2_v4l2_buffer` is failing the call to `__verify_length`
> Looking at the logging
> 
> ```
> [   52.479025] video13: VIDIOC_QBUF: error -22: 00:00:52.000438124 index=0, type=vid-out, request_fd=0, flags=0x00000000, field=none, sequence=0, memory=dmabuf, bytesused=4669440, offset/userptr=0x20, length=0
> ```
> 
> We've used 4669440 of a buffer that is of length 0. Oops.
> 
> ```
> diff --git a/src/libcamera/v4l2_videodevice.cpp b/src/libcamera/v4l2_videodevice.cpp
> index 508522e..5512f4f 100644
> --- a/src/libcamera/v4l2_videodevice.cpp
> +++ b/src/libcamera/v4l2_videodevice.cpp
> @@ -1427,6 +1427,8 @@ int V4L2VideoDevice::queueBuffer(FrameBuffer *buffer)
>                 } else {
>                         if (metadata.planes.size())
>                                 buf.bytesused = metadata.planes[0].bytesused;
> +
> +                       buf.length = buffer->planes()[0].length;
>                 }
>  
>                 buf.sequence = metadata.sequence;
> ```
> 
> Fix that and the changed `V4L2_CID_USER_BCM2835_ISP_BASE`, and libcamera runs.
> 
> Can I leave it with one of you to get the patch merged to libcamera?

Sure, I'll submit this to the mailing list.

naushir:
The V4L2_CID_USER_BCM2835_ISP_BASE is going to be tricky to handle.  Should we port that change to 5.4 as well so that users could continue using kernel 5.4 (with this change) instead of forcing them to use 5.9?

All other changes look reasonable to me.

6by9:
> The V4L2_CID_USER_BCM2835_ISP_BASE is going to be tricky to handle. Should we port that change to 5.4 as well so that users could continue using kernel 5.4 (with this change) instead of forcing them to use 5.9?

Certainly that's the easiest solution, although they will have to update libcamera in tandem. The error messages are all from our pipeline handler/ipa, so we could detect it early and give a nicer early abort if the control enums don't match.

> All other changes look reasonable to me.

OK, I'll push on with them.

```struct mmal_parameter_gamma``` packing is a nasty one. Any thoughts?

BGRX32 vs ABGR32? libcamera appears to handle either.
It needs some testsing actually - bcm2835-camera now uses ```V4L2_PIX_FMT_BGRX32``` to map ```MMAL_ENCODING_BGRA```, and qv4l2 and Cheese give the correct output. bcm2835-isp maps it ```V4L2_PIX_FMT_ABGR32```. So play hunt the alpha/padding.
(Blast - I've just got that one wrong with bcm2835-codec and left the deprecated ```V4L2_PIX_FMT_BGR32``` in there)

**Comment by naushir** [comment]
> `struct mmal_parameter_gamma` packing is a nasty one. Any thoughts?

I suppose we could align to a multiple of 4 easily enough for each array.  Given the CID change is going to break existing users any way (until they update both libcamera + kernel), we could do the same with this?

**Comment by 6by9** [comment]
OK, so we acknowledge an ABI break and make all the changes now. I'm good with that.

@pelwell Could you give a quick glance over the userspace API structures in https://github.com/raspberrypi/linux/blob/b1e9e6094fa21f982cebddb9145d5b63fa67075b/include/uapi/linux/bcm2835-isp.h to make sure we're not missing any other gotchas on alignment?

**Comment by 6by9** [comment]
Oh, ```__u32 gain_format``` in ```struct bcm2835_isp_lens_shading``` really ought to be an ```enum bcm2835_isp_gain_format```, but the enum then needs to be forced to 32bit.

Likewise ```strength``` in ```struct bcm2835_isp_dpc``` being an ```enum bcm2835_isp_dpc_mode```.

Before I go down a blind alley, was there any specific reason not to use the enum, other than needing to force it to 32bits?

**Comment by popcornmix** [comment]
Alignment looked okay to me apart from bcm2835_isp_gamma

You could put x and y in a 32-bit struct and then have array of BCM2835_NUM_GAMMA_PTS of those to avoid odd array sizes breaking alignment.

Breaking ABI at this point makes sense. I'm guessing number of users is low now but the longer it's left the more there'll be.

**Comment by popcornmix** [comment]
I'd have though enums are best avoided is user api structures. I'm certainly seeing instances of:
```
        __u8    er_reason;      /* reason (enum fc_els_rjt_reason below) */
```
in include/uapi where using enums has been specifically avoided.

pelwell:
I would move the `hist` member of `struct bcm2835_isp_stats` to the end, since the size isn't guaranteed to be a multiple of 8 if the value of NUM_HISTOGRAM_BINS is changed. `bcm2835_isp_gamma` isn't a problem because the two u16 arrays of equal length guarantee u32-alignment.

popcornmix:
I'm assuming the concern was that a compiler may choose to align the start of array `y` at a 32-bit boundary even when `BCM2835_NUM_GAMMA_PTS` was odd. And a compiler is perfectly entitled to do that (although I don't believe gcc will).

But a compiler is also allowed to align `{ __u16 x; __u8 z; __u8 padding; }` with each element at 32-bit alignment, so none of the padding tricks are guaranteed. I think the assumption here is the compiler will behave much like gcc.

6by9:
Ack - skip enums in uapi. mediabus and v4l2 do, and we're in their domain 
https://elixir.bootlin.com/linux/latest/source/include/uapi/linux/videodev2.h#L503
https://elixir.bootlin.com/linux/latest/source/include/uapi/linux/v4l2-mediabus.h#L33

Most of these structs are direct copies of the MMAL parameters, therefore any update of the actual content requires a matching firmware change, or repacking in the kernel driver.

With an odd value of ```BCM2835_NUM_GAMMA_PTS``` I was _expecting_ Metaware and gcc to align ```y``` to a 32bit boundary, so was going to add an extra u16 of padding.
Checking with Metaware I see that it doesn't, so actually we have to leave it as is, and hope it never gets padded. As you were then.

We have to have an enforced ABI break now as the value we used for V4L2_CID_USER_BCM2835_ISP_BASE has been taken by another upstream driver and we therefore need to renumber. I was looking to fix up anything else at the same point, hence the questions. It sounds like we're all OK though. Thanks!

drivers/staging/vc04_services/bcm2835-isp/bcm2835-isp-fmts.h
6by9: Discussion says ignore this for now.

drivers/staging/vc04_services/bcm2835-isp/bcm2835-v4l2-isp.c
6by9: This is the equivalent to [```v4l2_m2m_register_entity```](https://elixir.bootlin.com/linux/latest/source/drivers/media/v4l2-core/v4l2-mem2mem.c#L956) which uses kmalloc, so I guess there is a good reason. Not changing.

drivers/staging/vc04_services/bcm2835-isp/bcm2835-v4l2-isp.c
6by9: Kick this one down the road.

I'll look to solve it for the codecs so they give the correct signalling, then we can worry about how to get the correct values into the ISP CCMs.

6by9:
Slight hold up confirming the pixel format mappings.
Who assigned the V4L2 pixel formats?! https://www.kernel.org/doc/html/latest/userspace-api/media/v4l/pixfmt-rgb.html

V4L2_PIX_FMT_BGRX32 | ‘RX24’ bytewise in memory ```X B G R```

V4L2_PIX_FMT_RGBX32 | ‘XB24’ bytewise in memory ```R G B X```

V4L2_PIX_FMT_XBGR32 | ‘XR24’ bytewise in memory ```B G R X```

V4L2_PIX_FMT_XRGB32 | ‘BX24’ bytewise in memory ```X R G B```

so the 4CCs are consistent but backwards, but the V4L2_PIX_FMT_xxx strings don't follow any logic for BGR with X.
(GStreamer's gstv4l2object.c has mappings that do map back to byte order)

6by9:
OK, colour formats hopefully resolved, and I've squashed all the patches into one.

I'm happy with this now. Any further comments from @naushir or @davidplowman, or can we merge before Dom's next rebase?

pelwell:
Can you include the update for the arm64 bcmrpi3_defconfig as well?

6by9:
> Can you include the update for the arm64 bcmrpi3_defconfig as well?

You always miss one! Doing it now.

naushir:
> I'm happy with this now. Any further comments from @naushir or @davidplowman, or can we merge before Dom's next rebase?

LGTM.  After I'm done with my existing tasks, I'll backport the change to 5.4 and sync with the libcamera changes.

6by9:
> LGTM. After I'm done with my existing tasks, I'll backport the change to 5.4 and sync with the libcamera changes.

I'm happy to create a 5.4 PR with the uapi changes if needed.
libcamera I'll definitely leave to you though.

naushir:
> I'm happy to create a 5.4 PR with the uapi changes if needed.

That would be great!  We can't hit the merge button until the libcamera changes are ready though.

naushir:
> > Found it - `vb2_fill_vb2_v4l2_buffer` is failing the call to `__verify_length`
> > Looking at the logging
> > ```
> > [   52.479025] video13: VIDIOC_QBUF: error -22: 00:00:52.000438124 index=0, type=vid-out, request_fd=0, flags=0x00000000, field=none, sequence=0, memory=dmabuf, bytesused=4669440, offset/userptr=0x20, length=0
> > ```
> > 
> > 
> > We've used 4669440 of a buffer that is of length 0. Oops.
> > ```
> > diff --git a/src/libcamera/v4l2_videodevice.cpp b/src/libcamera/v4l2_videodevice.cpp
> > index 508522e..5512f4f 100644
> > --- a/src/libcamera/v4l2_videodevice.cpp
> > +++ b/src/libcamera/v4l2_videodevice.cpp
> > @@ -1427,6 +1427,8 @@ int V4L2VideoDevice::queueBuffer(FrameBuffer *buffer)
> >                 } else {
> >                         if (metadata.planes.size())
> >                                 buf.bytesused = metadata.planes[0].bytesused;
> > +
> > +                       buf.length = buffer->planes()[0].length;
> >                 }
> >  
> >                 buf.sequence = metadata.sequence;
> > ```
> > 
> > 
> > Fix that and the changed `V4L2_CID_USER_BCM2835_ISP_BASE`, and libcamera runs.
> > Can I leave it with one of you to get the patch merged to libcamera?
> 
> Sure, I'll submit this to the mailing list.

Sorry for resurrecting this conversion on a closed PR, but looking at the documentation [here](https://www.kernel.org/doc/html/v4.12/media/uapi/v4l/buffer.html#c.v4l2_buffer), `buf.length` is meant to be set by the driver for single plane mode, and set by the app for multiplanar mode.  So the bcm2835_isp driver should be setting this, correct?

6by9:
> Sorry for resurrecting this conversion on a closed PR, but looking at the documentation here, buf.length is meant to be set by the driver for single plane mode, and set by the app for multiplanar mode. So the bcm2835_isp driver should be setting this, correct?

In which case I'd expect videobuf2 to handle it all for us.
However this is in the case of using DMABUFs, so REQBUFS isn't actually allocating any buffer storage, so how does it know how big the buffer is?

videobuf2 does appear to set ```length``` based on the dmabuf size [if 0 has been passed in](https://elixir.bootlin.com/linux/latest/source/drivers/media/common/videobuf2/videobuf2-core.c#L1205), but I suspect that is called after the relatively new code to verify the lengths.

Ah, I think we may have been knee-capped by @jc-kynesim and https://github.com/raspberrypi/linux/commit/961d3b27a2aff52dda0b97d35085a743a27c2f46 (merged to mainline, and therefore supposedly reviewed by them!)
Before that it would have compared the length against ```vb->planes[0].length```, but now it's compared to b->length.
If the docs alter the requirement between single and multi-planar on who sets length, then it may haave been correct before.

naushir:
Confirmed that reverting https://github.com/raspberrypi/linux/commit/961d3b27a2aff52dda0b97d35085a743a27c2f46 fixes it.

jc-kynesim:
I've now forgotten exactly why I needed that fix - I think it was because I wanted variable length ES buffers as input to the H265 decoder and I had dmabufs to put them in.  I'll also argue that if USERPTRs want b->length then that should apply to  DMABUFs too.

6by9:
Laurent commented on linux-media a week ago. Shame no one beyond Hans looked at it before.
https://www.spinics.net/lists/linux-media/msg178649.html
https://patchwork.linuxtv.org/project/linux-media/patch/gh5kef5bkeel3o6b2dkgc2dfagu9klj4c0@4ax.com/

>  I'll also argue that if USERPTRs want b->length then that should apply to DMABUFs too.

In a sane world I'd agree, but that would be an ABI change. The underlying dmabuf does include the length, so userspace doesn't need to pass it across (as Laurent points out setting to 0 denotes "just ask the dmabuf").

I'm looking through my emails for history. 11 June 2020 "Fix for V4L2 monoplanar dmabuf queueing" looks like a candidate, but not enough detail to remind me.

jc-kynesim:
The issue I had was that I wanted to pass in a larger buffer than the default size and so it was rejected which was daft.  It would appear that upstream agreed with me.
On the other hand if I can get the needed behaviour by simply using a single plane multiplanar dmabuf I guess I could rework stuff to do that.

6by9:
Yes, I think the preferred approach is to set length to 0 for VB2_MEMORY_DMABUF so that it just takes the dmabuf size.

jc-kynesim:
That would be great if bytesused was compared against dmabuf size rather than the irrelevant plane size (what is the plane size of an ES?)

6by9:
Yes, there is the unanswerable question of what size is a buffer of compressed data, because a V4L2 buffer has to be a complete frame as they don't support fragmentation.
What do you set ```sizeimage``` to in the image format? And who is allocating these dmabufs that can sometimes be bigger than that?

For stateful encode/decode I just set the buffersize to 512kB for <=720p, and 768kB for >720p. It is totally hideous, but options are limited.

jc-kynesim:
As it stands in the H265 encoder I accept pretty much whatever sizeimage I'm asked for on the basis that the caller probably knows best with a minimum of 1k.  Stateless/Sliced is even worse than stateful for quessing bitbuffer sizes as you may well get data in a lot of rather small slices and forcing large buffer alloc to deal with that is wasteful especially as cache management on works on whole buffers.

The calling code in question is my VAAPI code - in order to not waste CMA if I don't need to I start with small buffers which I grow if needed.  It is true that I could ask for enormous planes with hopefully no downside unless there is another check in the code that checks if my dmabufs are at least as big as the plane (which I think there might be and may be why I got into this problem in the first place).

I can find no text that suggests that the length field should be treated differently between multi-planar and single plane - can you point me at it? (Obviously I'm comparing the buffer length field in single plane with the plane length field in multi-plane here)

But as I think it is only my VAAPI stuff that cares and that isn't in use feel free to do what ever you want here and I'll fix up my end later given that the multi-planar api should let me do it.

6by9:
Naush linked to it earlier - https://www.kernel.org/doc/html/v4.12/media/uapi/v4l/buffer.html#c.v4l2_buffer description of ```length```

> Size of the buffer (not the payload) in bytes for the single-planar API. This is set by the driver based on the calls to ioctl VIDIOC_REQBUFS and/or ioctl VIDIOC_CREATE_BUFS. For the multi-planar API the application sets this to the number of elements in the planes array. The driver will fill in the actual number of valid elements in that array.

For multi-planar it does say effectively the same thing, but we're looking in v4l2_plane - https://www.kernel.org/doc/html/v4.12/media/uapi/v4l/buffer.html#struct-v4l2-plane

> Size in bytes of the plane (not its payload). This is set by the driver based on the calls to ioctl VIDIOC_REQBUFS and/or ioctl VIDIOC_CREATE_BUFS.

It is however true that libcamera (one of the main users of this) is setting length from the application
https://git.linuxtv.org/libcamera.git/tree/src/libcamera/v4l2_videodevice.cpp#n1424

So your VAAPI code uses something external to V4L2 to do the allocations? (DMA heaps?) And then reallocate those if the buffer proves to be too small? Whilst nicer on the system, I think that isn't the best plan, or at least you need to update the V4L2 format to reflect that increased sizeimage so that it can keep performing sanity checks.

jc-kynesim:
Yes I'm using dmaheaps to do the allocation.  If I want to do a S_FMT then I'm pretty sure have to streamoff which is a poor idea especially if I want to do it mid-frame.  If I want to reject an incoming buffer due to odd size I can discard it and mark its output frame as error (and as in some circumstances I have to accumulate input buffers before processing that is what I do if that overflows the max size).

6by9:
Whilst everything you've written makes sense, and you're being nice by not allocating masses of CMA heap up front, I suspect you'll get pushback from mainline as in their view the V4L2 format sizeimage should be "correct" (for some definition of correct).
As to the difference between single and multi planar checking length in ```__verify_length```, I guess that's just one of those things. 

It'd be a question to ask on the mailing list as to why ```m.planes[plane].length=0``` there isn't permitted to use the size of the dmabuf.

jc-kynesim:
I refer you to the original dmabufs commit that altered the checking:
https://github.com/jc-kynesim/linux/commit/8a75ffb81

jc-kynesim:
That didn't format how I was expecting - how should I have inserted a link?

popcornmix:
It seems to link the url you used. What's the issue?

jc-kynesim:
It displays for me as `jc-kynesim@8a75ffb` rather than the whole link

popcornmix:
Yes, that's what github does. What happens if you click on it?

jc-kynesim:
The link works fine it is just how it presents that was unexpected. If it is an expected github thing then that is great - all other URLs come out in full so that is what I was expecting.

6by9:
Need to go back to the mailing list.

I'd agree that you're doing the same as Hans did there, but neither is what the spec actually says The spec says that the driver sets length. No apparent statement of the difference in behaviour for V4L2_MEMORY_USERPTR either that I can see.

I can't immediately find reference to ```length = 0``` being use the dmabuf size in either single or multi-planar API.

What I can find is struct ```v4l2_buffer``` [```bytesused```](https://www.kernel.org/doc/html/latest/userspace-api/media/v4l/buffer.html#struct-v4l2-buffer)

> If the application sets this to 0 for an output stream, then bytesused will be set to the size of the buffer (see the length field of this struct) by the driver. For multiplanar formats this field is ignored and the planes pointer is used instead.

Same in ```struct v4l2_plane```.

If V4L2_MEMORY_DMABUF ```length=0``` is "use the dmabuf length", then it needs to read that out, not use ```vb->v4l2_planes[0].length```.
If ``length!=0``` then the driver is meant to have set the length. If the application has used CREATE_BUFS, I don't believe there is any requirement for the size of the created buffer to be the same as the plane size.

Some experimentation required to confirm the behaviour, but I don't believe either path is correct at present.

jc-kynesim:
Well, I'm not going to argue with you here.  I'd be more than happy with 0 meaning "use the length of the dmabuf", but when you do argue the point on the list remember that you are going to have to deal with USERPTRs too and if it can be argued that they need the length field then it is no stretch to say that DMABUFs might too.  If you can manage to generate an argument/patch where my code still works then I would be grateful. Would a patch that adds "if dmabuf & b->length=0 then b->length = dmabuf->length" before the existing test do it for you?

6by9:
I'm starting to write the email on your behalf now.

> Would a patch that adds "if dmabuf & b->length=0 then b->length = dmabuf->length" before the existing test do it for you?

Largely, but it really wants to do what it says and check against the length of the dmabuf, not against some length that is internal to the driver.
The dma_buf isn't extracted from the fd until [__prepare_dmabuf](https://elixir.bootlin.com/linux/latest/source/drivers/media/common/videobuf2/videobuf2-core.c#L1196) though, so does length=0 get special cased here and checked later?

jc-kynesim:
> I'm starting to write the email on your behalf now.

Feel free to write it under your name - existing behaviour works fine for me.

> > Would a patch that adds "if dmabuf & b->length=0 then b->length = dmabuf->length" before the existing test do it for you?
> 
> Largely, but it really wants to do what it says and check against the length of the dmabuf, not against some length that is internal to the driver.

That was what my pseudo-code was trying to do.  If you set b->length to dmabuf->length before the test as currently written (with my patch) then surely everything works?

> The dma_buf isn't extracted from the fd until [__prepare_dmabuf](https://elixir.bootlin.com/linux/latest/source/drivers/media/common/videobuf2/videobuf2-core.c#L1196) though, so does length=0 get special cased here and checked later?

If that is the case that may be why Hans wrote his patch the way he did.  It wouldn't be the first time redundant parameters existed in calls.

6by9:
Hmm, there is a hole in the current rpivid should you use V4L2_MEMORY_MMAP and REQBUFS. It'll use the default sizeimage of 1024 which won't be big enough for anything useful.
You can use CREATEBUFS to create larger buffers, but that's not the normal approach taken.

jc-kynesim:
If you've set the size with S_FMT then it will use that size.  Setting the size there is legit.

jc-kynesim:
To be honest, buffer sizing is the least of rpivids worries - the patch I use to get 2 concurrent decodes breaks the media spec guarantee that once the media fd goes ready everything else is ready and that is something that I did have to patch FFmpeg to fix (buffer sizing works without patch), but the alternative is to completely rewrite the stateless scheduler which is something that needs to be done by an established list member.   As far as I can tell everyone on the list agrees that this should be done but no-one wants to do it.

6by9:
> If you've set the size with S_FMT then it will use that size. Setting the size there is legit.

Ah yes, you can override it.
Seeing as the stateless decoder API isn't your typical V4L2 device I can accept some extra config hoops to jump through.

The other potential approach is to increase the plane size in rpivid_queue_setup based on vq->memory, although that is listed as a private field with no accessor :-(

PR #3899, opened 2020-10-12
Add a gpio-fsm driver and some example overlays

pelwell:
The gpio-fsm driver implements simple state machines that allow GPIOs
to be controlled in response to inputs from other GPIOs - real and
soft/virtual - and time delays. It can:
+ create dummy GPIOs for drivers that demand them,
+ drive multiple GPIOs from a single input, with optional delays,
+ add a debounce circuit to an input,
+ drive pattern sequences onto LEDs
etc.

To do:
* Add irqchip support so that soft GPIOs can be waited on.
* Add a sysfs interface to display the current state of each FSM.


pelwell:
Are there any objections to merging this? Apart from some warnings about dt-bindings and documentation it's checkpatch clean, and harmless unless used.

popcornmix:
It looks interesting and I didn't spot anything troubling on a quick read through.
This isn't Pi specific - any thoughts on upstreaming?

But for now, feel free to merge.

pelwell:
The biggest hurdle to upstreaming it is likely to be that the DT devs would object to the fact that I'm using Device Tree to configure some software.

JamesH65:
BURN THE HERETIC!

PR #3904, opened 2020-10-15
Rpi 5.9.y - KMS updates

6by9:
Maxime's latest patches for muxing.
Phil's missing patch for disabling HDMI audio.
A new patch that allows dumping out the dlists for debug purposes.


popcornmix:
Works for me. dlist debug is nice to have (need to get a dlist parser...)

6by9:
>  (need to get a dlist parser...)

Tim G wrote one for both vc4 and vc5 during bringup, and I've tweaked as it didn't handle all the use cases. We ought to push it somewhere (private or public?)

PR #3909, opened 2020-10-16
Fixes a problem when module probes before i2c module is available

j-schambacher:
The driver crashed while a NULL pointer returned by i2c_get_adapter()
has been used to access the i2c bus functions.
The headphone probing function hb_hp_probe() now returns -EPROBE_DEFER
in case the i2c module has not been loaded yet.

### DIFF
--- a/sound/soc/bcm/hifiberry_dacplus.c
+++ b/sound/soc/bcm/hifiberry_dacplus.c
@@ -315,11 +315,13 @@ static int hb_hp_detect(void)
 {
 	struct i2c_adapter *adap = i2c_get_adapter(1);
 	int ret;
-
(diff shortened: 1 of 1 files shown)


pelwell:
`hb_hp_probe` is an unusual function in that it returns a boolean or an error - which is fine. However, the `if (ret) {` that you've added will make most Linux coders think it's the error path, which is then reinforced by you checking for -EPROBE_DEFER.

I think it would be clearer as :
```
        if (ret < 0)
                return ret;
        if (ret) {
                ...
        }
```
Other than that I'm happy with the patch.

j-schambacher:
Agreed. Fixed it. Thanks!

PR #3913, opened 2020-10-22
Allo Boss2 dac driver addition

alloboss2:
I am Sudeepkumar from Allo.com , email id : sudeepkumar@cem-solutions.net
we would like add Allo Boss2 driver to raspberrypi linux Repository.


arch/arm/boot/dts/overlays/README
```diff
@@ -431,6 +431,10 @@ Info:   Configures the Allo Katana DAC audio card
 Load:   dtoverlay=allo-katana-dac-audio
 Params: <None>
```
pelwell: Two blank lines are required between entries.

arch/arm/boot/dts/overlays/README
```diff
@@ -431,6 +431,10 @@ Info:   Configures the Allo Katana DAC audio card
...
 
+Name:   allo-boss2-dac-audio
+Info:   Configures the Allo Boss2 DAC audio card
+Load:   dtoverlay=allo-boss2-dac-audio
+Params: <None>
```
pelwell: And here.

arch/arm/boot/dts/overlays/README
```diff
@@ -431,6 +431,10 @@ Info:   Configures the Allo Katana DAC audio card
 Load:   dtoverlay=allo-katana-dac-audio
 Params: <None>
 
+Name:   allo-boss2-dac-audio
```
pelwell: But actually this whole entry belongs before allo-digione - you got the Makefile order right,

pelwell (commented):
Kudos for using the `audio-graph-card` driver (the new `simple-audio-card`), but in order to do that you've:
1. created a codec driver when one already exists for that codec (cs43130 implements "cirrus,cs43198"), with no explanation why, and
2. included in that codec driver things which really belong at the board level (`clock44-gpio`, `clock48-gpio`).

The end result is a lot more code to maintain than a dedicated board driver that makes use of the existing codec driver.

Please explain why this is a good thing.

pelwell:
And just to list all of the issues upfront:
* It's helpful if you keep code, defconfig and dts/overlay changes in three separate commits - it helps the subsequent squashing.
* Don't forget to also update arm64/bcm2711_defconfig and arm64/bcmrpi3_defconfig.
* Commit messages should have at least a minimal body and a `Signed-off-by:` tag.

alloboss2:
@pelwell  thanks for your reply and suggestions. We have updated the files you suggested. 

Probing with default CS43130 codec is not working with our hardware setup.
Our Boss2 dac does not support hardware based interrupt.
some features with default CS43130 not working and some we are not using.

pelwell:
Using the existing cs43130 header file was a good idea, and that clearly meant you had to change some macro names. Personally I think you've gone a bit far in the renaming, but it's not my driver.

What you need to do now is a `git reset HEAD~2` to undo the commits but keep the changes, then create three commits:
1. The driver and its Makefile.
2. The overlay, README and its Makefile.
3. The config changes.

Finally, `git push -f <remote> HEAD:rpi-5.4.y` (possibly `git push -f alloboss2 HEAD:rpi-5.4.y`).

alloboss2:
@pelwell 

we hope all commits are ok and pull request will be accepted soon.

pelwell:
We don't (often) get notifications of pushes, so it's always a good idea to also add a comment when you think it's ready for testing.

alloboss2:
@pelwell   as you suggested done as 3 commits. please do the needful.

PR #3914, opened 2020-10-22
5.9: ov9281 cherrypick and kms patches

6by9:
Cherry-pick of ff93994f from rpi-5.4.y

One bug fix patch to vc4_hdmi that fixes incorrect hot plug behaviour on Pi3.
One debug patch to dump out all the HDMI registers on PI4 instead of only those that overlap with Pi0-3.

### DIFF
--- a/drivers/gpu/drm/vc4/vc4_hdmi.c
+++ b/drivers/gpu/drm/vc4/vc4_hdmi.c
@@ -442,8 +442,9 @@ vc4_hdmi_connector_detect(struct drm_connector *connector, bool force)
 			connected = true;
 	} else if (drm_probe_ddc(vc4_hdmi->ddc))
 		connected = true;
-	if (HDMI_READ(HDMI_HOTPLUG) & VC4_HDMI_HOTPLUG_CONNECTED)
+	else if (HDMI_READ(HDMI_HOTPLUG) & VC4_HDMI_HOTPLUG_CONNECTED)
 		connected = true;
+
 	if (connected) {
 		if (connector->status != connector_status_connected) {
 			struct edid *edid = drm_get_edid(connector, vc4_hdmi->ddc);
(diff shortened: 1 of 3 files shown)


drivers/gpu/drm/vc4/vc4_hdmi.c
```diff
@@ -2281,6 +2289,42 @@ static int vc5_hdmi_init_resources(struct vc4_hdmi *vc4_hdmi)
 		return PTR_ERR(vc4_hdmi->reset);
 	}
 
+	ret = vc4_hdmi_build_regset(vc4_hdmi, &vc4_hdmi->hdmi_regset, VC4_HDMI);
```
pelwell, line 2292: Were `hdmi_regset` and `hd_regset` previously used uninitialised?

6by9: They would have been zeroed by the kzalloc of ```vc4_hdmi``` in ```vc4_hdmi_bind```. The printing code of ```drm_print_regset32``` loops through from 0 to ```regset->nregs```, which will be 0 unless any registers have been added.

I'm relying on the same behaviour to do nothing for all the new vc5 register blocks when called on vc4.

6by9:
@popcornmix This HDMI hotplug error may link into your CEC failure on Pi3 as it invalidates settings and generally messes around with CEC state when connected/disconnected.

popcornmix:
I'll give it a test.
I had been debugging and found firmware was overwriting some registers (e.g. `HDMI_CEC_CNTRL_1`) but hadn't worked out why this only affected pi0-pi3 and 5.9.

6by9:
The firmware is overwriting registers? Have we not managed a clean handover from VPU to kernel then? Eek.

arm_dt.c masks the VPU away from the interrupts of any DT nodes that start ```v3d hdmi pixelvalve hvs txp firmwarekms csi```, so hopefully that would stop it servicing them, but has HDMI been started before that point?

Issue #3915, opened 2020-10-22
Overlay for MAX31856


dennyfmn:
According to PhilE there is support for the MAX31856 in kernel 5.9 but no overlay yet. Ref: 
[https://www.raspberrypi.org/forums/viewtopic.php?f=107&t=288967](url)
I would like to request this to be created, I would test it if it were available. Thanks!

pelwell:
If PhilE says so then he's probably correct. ;-)

Although I can have a guess at what a suitable overlay might look like, I would prefer to have it tested before adding it to the kernel. Are you running 5.9 at the moment, or more importantly, are you prepared to upgrade so you can test whatever I come up with?

dennyfmn:
I have a brand new current stable 5.4.51+ #1333 system that I'm testing the MAX6675 with. I can easily clone it and update to 5.9 for testing. I would intend to use it with a type K thermocouple. I'd be very pleased to work with you.

pelwell:
I've opened a Pull Request (#3916) with the modifications to the maxtherm overlay to support the MAX31856. You can download a build of the overlay here: https://drive.google.com/file/d/1YSHc3-WQAO4gZHKLbJ1GtHXPOL2RYDad/view?usp=sharing
Just copy it into `/boot/overlays/` (you'll need to use `sudo cp ...`). The updated README gives usage information: https://github.com/raspberrypi/linux/blob/17a22a794a6c8ff350e52d93069eff1378d1b8d6/arch/arm/boot/dts/overlays/README#L1668

dennyfmn:
Cool! I have an Adafruit MAX31856 board on order, it should be here on Saturday. I'll let you know how it goes...

dennyfmn:
Ok I have it connected with this in /boot/config.txt
dtoverlay=maxtherm,spi0-0,max31856,type_k   (also tried without ,type_k same results)
I have this structure at /sys/bus/spi/devices/spi0.0
-rw-r--r-- 1 root root 4096 Oct 24 17:04 driver_override
-r--r--r-- 1 root root 4096 Oct 24 17:04 modalias
lrwxrwxrwx 1 root root    0 Oct 24 17:04 of_node -> ../../../../../../../firmware/devicetree/base/soc/spi@7e204000/maxtherm@0
drwxr-xr-x 2 root root    0 Oct 24 17:04 power
drwxr-xr-x 2 root root    0 Oct 24 17:04 statistics
[...]

PR #3916, opened 2020-10-22
overlays: Add MAX31856 support to maxtherm overlay

pelwell:
Extend the maxtherm overlay with support for the MAX31856.
The driver reads the thermocouple type from a property, which is much
more civilised.

See: https://github.com/raspberrypi/linux/issues/3915


pelwell:
Merged offline long ago.

Issue #3918, opened 2020-10-22
config:  could we enable parallel port support?


garlick:
There's an open hardware/software design for a GPIO-based IEEE 1284 parallel port HAT here:
https://github.com/worlickwerx/pi-parport

This has been used to resurrect some older scientific equipment with parallel port interfaces and other fun things.

Would it be possible to enable parport support in the default raspbian kernels?   It would make it easier for people to run the board, since otherwise they either have to rebuild the kernel or build the parport bits out of tree.  I suppose it wouldn't hurt with CM4 offering a usable PCIe option down the road, since that would be another way to get a parallel port.

I'd be happy to submit a PR for the config changes if given the nod here.

On the off chance that at PR for the driver and device tree overlay could be accepted, I'd happily submit that too, although I assume that upstream is where that should go, and I'm not quite ready to go on that ride.

pelwell:
Enabling modules is usually uncontroversial. Submit what you have as PRs then we can make some decisions.

PR #3923, opened 2020-10-24
add support for GPIO based parallel port

garlick:
Following up on #3918, this PR enables the parport stack in the raspberry pi configs, adds a small device driver that implements a parport driver based on GPIOs, and adds a device tree overlay.

The device tree overlay matches an open hardware design for a HAT here:  http://github.com/worlickwerx/pi-parport.

This approach works for the `lp` and `ppdev` drivers, and several other fun applications including a parallel port astronomy camera http://github.com/garlick/sbig-parport (the original motivation for the project).

If the driver + overlay are not acceptable, I'd still be keen to enable the parport bits that we use, so that people trying use the board don't have to reconfigure a kernel, or build thiose bits out of tree.   Let me know and I'll modify this PR accordingly.   Thanks!


garlick:
This has been sitting a while.  Should I drop the driver and overlay and just submit the config changes for a start?

PR #3927, opened 2020-10-26
dt updates, and kms using bcm2711-hvs for dma range

6by9:


arch/arm64/boot/dts/broadcom/bcm2711-rpi-cm4.dts
```diff
@@ -0,0 +1,3 @@
+#define RPI364
```
pelwell: This macro is no longer used in our kernel branches.

pelwell: Or at least it isn't in rpi-5.4.y, and yet it's still there (or there again) in rpi-5.9.y. Hmm.

6by9: It's present in https://github.com/raspberrypi/linux/blob/rpi-5.9.y/arch/arm64/boot/dts/broadcom/bcm2710-rpi-cm3.dts#L1 and all the other 64bit variants of the DT files.

https://github.com/raspberrypi/linux/blob/rpi-5.9.y/arch/arm/boot/dts/bcm2710.dtsi#L8 also says otherwise.

pelwell: rpi-5.9.y was missing 349815ca97853d846a1d53176ac85a81c7abc08b, which dates back to March and applies without conflicts so its unclear how it got dropped. It's present now, so you can drop the `RPI364`.

6by9: OK, rebased and drops that.

PR #3928, opened 2020-10-26
Allow the headphone output to be disabled

pelwell:
Pi Zeroes and CM4 have no headphone outputs. Disable the corresponding ALSA device so that it doesn't show up in the GUI and sounds utilities.

### DIFF
--- a/drivers/staging/vc04_services/bcm2835-audio/bcm2835.c
+++ b/drivers/staging/vc04_services/bcm2835-audio/bcm2835.c
@@ -381,11 +381,16 @@ static int snd_bcm2835_alsa_probe(struct platform_device *pdev)
 	}
 
 	if (!enable_compat_alsa) {
+		// In this mode, enable analog output by default
+		u32 disable_headphones = 0;
+
 		if (!of_property_read_bool(dev->of_node, "brcm,disable-hdmi"))
 			set_hdmi_enables(dev);
 
-		// In this mode, always enable analog output
-		enable_headphones = true;
+		of_property_read_u32(dev->of_node,
+				     "brcm,disable-headphones",
+				     &disable_headphones);
+		enable_headphones = !disable_headphones;
 	} else {
 		enable_hdmi0 = enable_hdmi;
 	}
--- a/arch/arm/boot/dts/bcm2708-rpi-zero-w.dts
+++ b/arch/arm/boot/dts/bcm2708-rpi-zero-w.dts
@@ -152,6 +152,7 @@
 &audio {
 	pinctrl-names = "default";
 	pinctrl-0 = <&audio_pins>;
+	brcm,disable-headphones = <1>;
 };
 
 / {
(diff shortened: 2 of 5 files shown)


pelwell:
Could you give this a once-over, @6by9?

drivers/staging/vc04_services/bcm2835-audio/bcm2835.c
```diff
@@ -384,8 +385,12 @@ static int snd_bcm2835_alsa_probe(struct platform_device *pdev)
 		if (!of_property_read_bool(dev->of_node, "brcm,disable-hdmi"))
 			set_hdmi_enables(dev);
 
-		// In this mode, always enable analog output
-		enable_headphones = true;
+		// In this mode, enable analog output by default
+		disable_headphones = 0;
```
6by9: nit pick: Just use ```u32 disable_headphones = 0;``` within this scope (defined at the beginning of it), instead of having the variable definition and initialisation separate.

6by9 (commented):
So headphones are disabled on these platforms unless you use audremap to push them onto the 40pin header? Using that overlay automatically re-enables the headphone out.

Sounds reasonable and should do that.

drivers/staging/vc04_services/bcm2835-audio/bcm2835.c
pelwell: This was a conscious choice because I like the fact that the initialisation is closer to where the value is used - it requires less scrolling back to discover the default value - and checkpatch is happy with it.

pelwell: But if you are aware of similar cases where upstream devs have objected then I am willing to change it,

6by9: Is 3 lines really that much to scroll back up?
```
	if (!enable_compat_alsa) {
		// In this mode, enable analog output by default
		u32 disable_headphones = 0;

		if (!of_property_read_bool(dev->of_node, "brcm,disable-hdmi"))
			set_hdmi_enables(dev);

		of_property_read_u32(dev->of_node,
				     "brcm,disable-headphones",
				     &disable_headphones);
		enable_headphones = !disable_headphones;
```
It's not used outside of this clause, so why define it at the start of the function?

pelwell: Because that seems to be the kernel style - old-school C. But checkpatch is happy with either so the updated version goes with the block-scoped `disable_headphones`.

6by9: It's certainly acceptable (and very commonly used) with for loops, including in this file:
https://github.com/raspberrypi/linux/blob/0531eec70c535ad04138b5300231b5bb4c1f5f76/drivers/staging/vc04_services/bcm2835-audio/bcm2835.c#L282-L287

I've never seen anything saying you can't or shouldn't use it with any other block-scoping.

pelwell:
Thanks for the review.

PR #3933, opened 2020-10-27
bcm2835-pcm: Fix up multichannel pcm audio

popcornmix:
Fixes: a9c1660ff5f02d048c5f31abf1fd1108ccf9ef87

### DIFF
--- a/drivers/staging/vc04_services/bcm2835-audio/bcm2835-pcm.c
+++ b/drivers/staging/vc04_services/bcm2835-audio/bcm2835-pcm.c
@@ -14,14 +14,14 @@ static const struct snd_pcm_hardware snd_bcm2835_playback_hw = {
 		 SNDRV_PCM_INFO_MMAP | SNDRV_PCM_INFO_MMAP_VALID |
 		 SNDRV_PCM_INFO_SYNC_APPLPTR),
 	.formats = SNDRV_PCM_FMTBIT_U8 | SNDRV_PCM_FMTBIT_S16_LE,
-	.rates = SNDRV_PCM_RATE_CONTINUOUS |  SNDRV_PCM_RATE_8000_192000,
+	.rates = SNDRV_PCM_RATE_CONTINUOUS | SNDRV_PCM_RATE_8000_192000,
 	.rate_min = 8000,
 	.rate_max = 192000,
 	.channels_min = 1,
-	.channels_max = 2,
-	.buffer_bytes_max = 128 * 1024,
+	.channels_max = 8,
+	.buffer_bytes_max = 512 * 1024,
 	.period_bytes_min = 1 * 1024,
-	.period_bytes_max = 128 * 1024,
+	.period_bytes_max = 512 * 1024,
 	.periods_min = 1,
 	.periods_max = 128,
 };
@@ -31,16 +31,15 @@ static const struct snd_pcm_hardware snd_bcm2835_playback_spdif_hw = {
 		 SNDRV_PCM_INFO_MMAP | SNDRV_PCM_INFO_MMAP_VALID |
 		 SNDRV_PCM_INFO_SYNC_APPLPTR),
 	.formats = SNDRV_PCM_FMTBIT_S16_LE,
-	.rates = SNDRV_PCM_RATE_44100 | SNDRV_PCM_RATE_48000 |
-	SNDRV_PCM_RATE_88200 | SNDRV_PCM_RATE_96000 |
-	SNDRV_PCM_RATE_176400 | SNDRV_PCM_RATE_192000,
+	.rates = SNDRV_PCM_RATE_CONTINUOUS | SNDRV_PCM_RATE_44100 |
+	SNDRV_PCM_RATE_48000,
 	.rate_min = 44100,
-	.rate_max = 192000,
+	.rate_max = 48000,
 	.channels_min = 2,
-	.channels_max = 8,
-	.buffer_bytes_max = 512 * 1024,
+	.channels_max = 2,
+	.buffer_bytes_max = 128 * 1024,
 	.period_bytes_min = 1 * 1024,
-	.period_bytes_max = 512 * 1024,
+	.period_bytes_max = 128 * 1024,
 	.periods_min = 1,
 	.periods_max = 128,
 };


popcornmix:
The previous commit enabled multichannel for the spdif device (as we were experimenting with HBR passthrough which never worked through this interface). The more useful interface to enable multichannel is the pcm one. 

Tested here: https://www.diyaudio.com/forums/pc-based/334443-2-8-dsp-platform-using-raspberry-pi-hats-10.html

pavhofman:
IMO the problem is the HDMI cards created by the xxx_simple_xxx config call the function snd_bcm2835_new_pcm with param spdif=false which causes that the 2ch hw params for the headphones route are chosen, instead of the 8ch spdif hw params.

Perhaps something like this could be a solution (not tested at all):

```
index 092ccbf..e57f5f8 100644
--- a/bcm2835.c
+++ b/bcm2835.c
@@ -91,7 +91,7 @@ static int bcm2835_audio_alsa_newpcm(struct bcm2835_chip *chip,
        return 0;
 }
 
-static int bcm2835_audio_simple_newpcm(struct bcm2835_chip *chip,
+static int bcm2835_audio_headphones_newpcm(struct bcm2835_chip *chip,
                                       const char *name,
                                       enum snd_bcm2835_route route,
                                       u32 numchannels)
@@ -99,6 +99,14 @@ static int bcm2835_audio_simple_newpcm(struct bcm2835_chip *chip,
        return snd_bcm2835_new_pcm(chip, name, 0, route, numchannels, false);
 }
 
+static int bcm2835_audio_hdmi_newpcm(struct bcm2835_chip *chip,
+                                      const char *name,
+                                      enum snd_bcm2835_route route,
+                                      u32 numchannels)
+{
+       return snd_bcm2835_new_pcm(chip, name, 0, route, numchannels, true);
+}
+
 static struct bcm2835_audio_driver bcm2835_audio_alsa = {
        .driver = {
                .name = "bcm2835_alsa",
@@ -119,7 +127,7 @@ static struct bcm2835_audio_driver bcm2835_audio_hdmi0 = {
        .shortname = "bcm2835 HDMI 1",
        .longname  = "bcm2835 HDMI 1",
        .minchannels = 1,
-       .newpcm = bcm2835_audio_simple_newpcm,
+       .newpcm = bcm2835_audio_hdmi_newpcm,
        .newctl = snd_bcm2835_new_hdmi_ctl,
        .route = AUDIO_DEST_HDMI0
 };
@@ -132,7 +140,7 @@ static struct bcm2835_audio_driver bcm2835_audio_hdmi1 = {
        .shortname = "bcm2835 HDMI 2",
        .longname  = "bcm2835 HDMI 2",
        .minchannels = 1,
-       .newpcm = bcm2835_audio_simple_newpcm,
+       .newpcm = bcm2835_audio_hdmi_newpcm,
        .newctl = snd_bcm2835_new_hdmi_ctl,
        .route = AUDIO_DEST_HDMI1
 };
@@ -145,7 +153,7 @@ static struct bcm2835_audio_driver bcm2835_audio_headphones = {
        .shortname = "bcm2835 Headphones",
        .longname  = "bcm2835 Headphones",
        .minchannels = 1,
-       .newpcm = bcm2835_audio_simple_newpcm,
+       .newpcm = bcm2835_audio_headphones_newpcm,
        .newctl = snd_bcm2835_new_headphones_ctl,
        .route = AUDIO_DEST_HEADPHONES
 };
```

drivers/staging/vc04_services/bcm2835-audio/bcm2835-pcm.c
```diff
@@ -35,12 +35,12 @@ static const struct snd_pcm_hardware snd_bcm2835_playback_spdif_hw = {
 	SNDRV_PCM_RATE_88200 | SNDRV_PCM_RATE_96000 |
 	SNDRV_PCM_RATE_176400 | SNDRV_PCM_RATE_192000,
 	.rate_min = 44100,
-	.rate_max = 192000,
+	.rate_max = 48000,
```
pelwell, line 37: Is this correct? If so, don't you also need to change the .rates above?

popcornmix: It is correct (reverting a9c1660) but you are right, the rates should change as well.

pavhofman: IMO the HDMI hardware does support 192kHz. But the channel count is incorrect for HDMI separate cards created by the bcm2835_audio_simple_newpcm call.

popcornmix: The snd_bcm2835_playback_spdif_hw is intended for non-pcm (passthough). I don't believe we usefully support anything beyond 2 channel 48kHz (for dts and ac3).

pavhofman:
Well, multiple users report 8ch HDMI works OK on their RPis https://www.diyaudio.com/forums/pc-based/334443-2-8-dsp-platform-using-raspberry-pi-hats-10.html#post6345442

popcornmix:
@HiassofT what is your view on these settings? 
On Pi0-3 we can support:
8 channel 48kHz PCM
2 channel 192kHz PCM
2 channel 44.1/48kHz passthrough

HiassofT:
@popcornmix I don't think it's possible to express that via standard pcm constraints.

A solution would be to do it programatically. i.e. specify relaxed constraints (max 8ch, 192kHz) and reject invalid combinations with EINVAL in hw_params

popcornmix:
What about the `snd_bcm2835_playback_hw` and `snd_bcm2835_playback_spdif_hw` devices?
Do we treat the spdif as passthrough only and only report what we believe to be useful or report the full range of channels/samplerates?

HiassofT:
TBH I'm not too familiar with that particular driver.

From a quick glance it looks like the spdif device might work like normal HDMI audio devices and PCM/non-audio can be selected via IEC958 ctl. So supportting multi-channel and higher rates on would be fine - difference to the non-spdif device seems to be that it's a single-user non-mixing device.

OTOH it depends on what the hardware+firmware can support and what you want to support - allowing non-working combinations to be used might not be ideal. So if you want the spdif device to be 2ch/48kHz only then that's fine, too.

pavhofman:
That's my experience too. The spdif (or HDMI) is a regular PCM device, without mixer controls provided by the codecs. IMO "passthrough" is just a name of a bit-perfect (i.e. no mixer/volume control) path. Digital (nonPCM and nonDSD) formats have a marker in the stream (and a corresponding flag in the SPDIF preamble, if applicable).

People use RPi to feed 8ch HDMI to a  HDMI extractor to get 8 analog channels.  Until recently they had to recompile kernel to change the channels_max hw param, with the latest patches the snd_bcm2835_playback_spdif_hw  params already specify 8ch but the card-based device (enable_hdmi=1 instead of the device based enable_compat_alsa=1) uses  snd_bcm2835_playback_hw params defined for the headphone audio (PWM-based) hardware, even though it's still the same HDMI hardware.

popcornmix:
Fixed the rates of `snd_bcm2835_playback_spdif_hw` so it now purely a revert of a9c1660.
I believe this device was always intended to be used for passthough so just report what works.
The pcm device is upgraded for multichannel and should be used when that is the use case.

pavhofman:
Please do I understand correctly this is the current result?

module param enable_headphones=1 =>  using snd_bcm2835_playback_hw  => 8ch/192kHz on the stereo analog jack
enable_hdmi=1 =>  using snd_bcm2835_playback_hw  => 8ch/192kHz on HDMI0/1 (depending on which output is currently used)

enable_compat_alsa=1 => using snd_bcm2835_playback_hw for headphones and snd_bcm2835_playback_spdif_hw for HDMI0/1 ( https://github.com/popcornmix/linux/blob/75daef083656bd9d84a388089b5671e1eca8640c/drivers/staging/vc04_services/bcm2835-audio/bcm2835.c#L78 )   => 8ch/192kHz on the stereo analog jack + 2ch/48kHz on HDMI0 + 2ch/48kHz on HDMI1

Please correct me if I am wrong.

gordoste:
Hi @popcornmix , I think it may be necessary to modify lines 83 and 87 of bcm2835.c - the HDMI devices get a maximum channel count of 2 because the spdif parameter to snd_bcm2835_new_pcm is true.
Of course, the workaround is to simply not use the enable_compat_alsa parameter.

PR #3937, opened 2020-10-28
Add support for 'tryboot' OS upgrade fallback mechanism on BCM2711

timg236:
Extend parsing of reboot notifier parameters to check for an additional 'tryboot' parameter. This instructs the GPU firmware to set a bit which is read then cleared and boot that causes the bootloader/firmware to read from tryboot.txt instead of config.txt

This alternate configuration file allows the pending updated to be tested. If a watchdog reset or power cycle occurs then the previous configuration will be loaded. 

This change requires an updated GPU firmware and bootloader.

### DIFF
--- a/drivers/firmware/raspberrypi.c
+++ b/drivers/firmware/raspberrypi.c
@@ -190,6 +190,7 @@ static int rpi_firmware_notify_reboot(struct notifier_block *nb,
 {
 	struct rpi_firmware *fw;
 	struct platform_device *pdev = g_pdev;
+	u32 reboot_flags = 0;
 
 	if (!pdev)
 		return 0;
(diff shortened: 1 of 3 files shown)


drivers/firmware/raspberrypi.c
```diff
@@ -198,8 +199,27 @@ static int rpi_firmware_notify_reboot(struct notifier_block *nb,
 	if (!fw)
 		return 0;
 
-	(void)rpi_firmware_property(fw, RPI_FIRMWARE_NOTIFY_REBOOT,
-				    0, 0);
+	// The partition id is the first parameter followed by zero or
+	// more flags indicating the reason for the reboot.
+	//
+	// 'tryboot': Sets a one-shot flag which is cleared upon reboot but
+	//            causes the config.txt [tryboot] conditional filter
```
pelwell: A lot of stuff has happened since this comment was written...

timg236: Good point!

PR #3946, opened 2020-11-05
staging: vcsm-cma: Fix memory leak from not detaching dmabuf

6by9:
When importing there was a missing call to detach the buffer,
so each import leaked the sg table entry.

Actually the release process for both locally allocated and
imported buffers is identical, so fix them to both use the same
function.

### DIFF
--- a/drivers/staging/vc04_services/vc-sm-cma/vc_sm.c
+++ b/drivers/staging/vc04_services/vc-sm-cma/vc_sm.c
@@ -237,6 +237,7 @@ static void vc_sm_add_resource(struct vc_sm_privdata_t *privdata,
 
 /*
  * Cleans up imported dmabuf.
+ * Should be called with mutex held.
  */
 static void vc_sm_clean_up_dmabuf(struct vc_sm_buffer *buffer)
 {
@@ -244,7 +245,6 @@ static void vc_sm_clean_up_dmabuf(struct vc_sm_buffer *buffer)
 		return;
 
 	/* Handle cleaning up imported dmabufs */
-	mutex_lock(&buffer->lock);
 	if (buffer->import.sgt) {
 		dma_buf_unmap_attachment(buffer->import.attach,
 					 buffer->import.sgt,
@@ -255,7 +255,6 @@ static void vc_sm_clean_up_dmabuf(struct vc_sm_buffer *buffer)
 		dma_buf_detach(buffer->dma_buf, buffer->import.attach);
 		buffer->import.attach = NULL;
 	}
-	mutex_unlock(&buffer->lock);
 }
 
 /*
@@ -687,23 +686,6 @@ int vc_sm_import_dmabuf_mmap(struct dma_buf *dmabuf, struct vm_area_struct *vma)
 }
 
 static
-void vc_sm_import_dma_buf_release(struct dma_buf *dmabuf)
-{
-	struct vc_sm_buffer *buf = dmabuf->priv;
-
-	pr_debug("%s: Relasing dma_buf %p\n", __func__, dmabuf);
-	mutex_lock(&buf->lock);
-	if (!buf->imported)
-		return;
-
-	buf->in_use = 0;
-
-	vc_sm_vpu_free(buf);
-
-	vc_sm_release_resource(buf);
-}
-
-static
 void *vc_sm_import_dma_buf_kmap(struct dma_buf *dmabuf,
 				unsigned long offset)
 {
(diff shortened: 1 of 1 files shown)


6by9:
@naushir  @davidplowman Can you confirm this solves your libcamera memory leak please?

I do note that you will get a lingering ls_grid allocation after the app has been closed.
The VPU ISP component has been given a grid and it will hold a reference to it until:
- given a new one
- told to release it
- the component is destroyed.

Seeing as the V4L2 driver retains the component for the duration whilst it is loaded, the last one never happens. It's only a small allocation at 32kB, but it would be nice if it got released. We therefore need to look at doing one of the others (namely releasing it).
The niggle is that doing so will require LS to always be reinitialised by each new V4L2 client, and we need to invalidate the V4L2 control on release, but that feels reasonable.

drivers/staging/vc04_services/vc-sm-cma/vc_sm.c
```diff
@@ -237,14 +237,14 @@ static void vc_sm_add_resource(struct vc_sm_privdata_t *privdata,
...
  * Cleans up imported dmabuf.
+ * Should be called with mutex held.
  */
 static void vc_sm_clean_up_dmabuf(struct vc_sm_buffer *buffer)
 {
 	if (!buffer->imported)
 		return;
 
 	/* Handle cleaning up imported dmabufs */
-	mutex_lock(&buffer->lock);
```
pelwell, line 247: Was `vc_sm_dma_buf_release` never used before? I can't see how the double mutex_lock could have worked?

6by9: Allocations by vcsm and imported dmabufs created dmabufs with different dma_buf_ops.

Allocs used ```vc_sm_dma_buf_release```, which called ```vc_sm_clean_up_dmabuf```, but as ```buffer->imported``` will have been 0, it immediately returned.

Imported dmabufs used ```vc_sm_import_dma_buf_release```, which omited to call ```vc_sm_clean_up_dmabuf```, hence the memory leak of the sgt due to not unmapping and detaching the buffer, but also never going through this path that would have double locked.

5.9 I'm ditching vcsm doing any allocations and getting userspace to handle the allocs via dma-heaps.

pelwell:
I'm happy to merge after some confirmation that it solves the problem.

davidplowman:
I'm building Dave's rpi-5.4.y branch now and I'll let you know...

davidplowman:
Well, preliminary indications are good. qcam starts and stops happily. That tell-tale "devices attached" number from /sys/kernel/debuf/dma_buf/bufinfo is no longer going up ("ls_grid" seems to report generally as "0 devices attached" - does that sound right?). No lock-ups. Nothing weird in dmesg. Performing the original test - letting it run for 24 hours - might only happen over the weekend now, but it seems to be behaving.

Do we know when this will become available via rpi-update? It would be nice to post a final reply to that thread on libcamera-devel when that happens. Thanks, everyone!

6by9:
> ("ls_grid" seems to report generally as "0 devices attached" - does that sound right?).

See my comment above.
Count is 0001 on that element which is the VPU holding a reference to the buffer still.

To get the VPU to release that reference requires either explicitly telling it to release (in which case we have a slightly odd behaviour in that the V4L2 control still references the dmabuf), or for the component to be destroyed (which is a faff).

naushir:
Just to confirm, all looks good for me so far...

popcornmix:
@davidplowman rpi-update will get this fix.

AradoSKY:
Hello, 

I am having an issue with this package. It will not successfully install.  The following are the product evidences of why it is not able to install. I opened a separate issue for this in GitHUB raspberry firmware repository. 

root@astroberry:/home/astroberry# apt-get dist-upgrade
Reading package lists... Done
Building dependency tree       
Reading state information... Done
Calculating upgrade... Done
0 upgraded, 0 newly installed, 0 to remove and 0 not upgraded.
1 not fully installed or removed.
After this operation, 0 B of additional disk space will be used.
Do you want to continue? [Y/n] y
Setting up flash-kernel (3.98ubuntu12) ...
Using DTB: bcm2710-rpi-3-b-plus.dtb
Installing /etc/flash-kernel/dtbs/bcm2710-rpi-3-b-plus.dtb into /boot/dtbs//./bcm2710-rpi-3-b-plus.dtb
Taking backup of bcm2710-rpi-3-b-plus.dtb.
Installing new bcm2710-rpi-3-b-plus.dtb.
ln: failed to create symbolic link '/boot/dtb-': Operation not permitted
dpkg: error processing package flash-kernel (--configure):
 installed flash-kernel package post-installation script subprocess returned error exit status 1
Errors were encountered while processing:
 flash-kernel
E: Sub-process /usr/bin/dpkg returned an error code (1)
root@astroberry:/home/astroberry#

pelwell:
As @XECDesign said in https://github.com/raspberrypi/firmware/issues/1501, that isn't one of our packages, and asking again in a different repo doesn't make it so.

If you aren't getting support from the Ubuntu Mate maintainers, perhaps you might want to consider a different distribution.

juergh:
@AradoSKY Use the Ubuntu Mate bugtracker: https://bugs.launchpad.net/ubuntu-mate/+filebug

PR #3949, opened 2020-11-09
backport of mcp251xfd (v5.10-rc3)

marckleinebudde:
The mcp251xfd is a CAN-FD SPI controller found on many rpi add-on modules. This pull requests ports back the driver as of v5.10-rc3.


arch/arm/boot/dts/overlays/README
```diff
@@ -1762,6 +1762,21 @@ Params: oscillator              Clock frequency for the CAN controller (Hz)
 
         interrupt               GPIO for interrupt signal
 
+Name:   mcp25xxfd-spiX-Y
+Info:   Configures the MCPXXFD CAN controller family on spiX.Y
+Load:   dtoverlay=mcp25xxfd-spiX-Y,<param>=<val>
+Params: oscillator              Clock frequency for the CAN controller (Hz)
+
+        spimaxfrequency         Maximum SPI frequence (Hz)
```
pelwell: There's a convention to call this property `speed`.

arch/arm/boot/dts/overlays/README
```diff
@@ -1762,6 +1762,21 @@ Params: oscillator              Clock frequency for the CAN controller (Hz)
 
         interrupt               GPIO for interrupt signal
 
+Name:   mcp25xxfd-spiX-Y
```
pelwell: Recent advances in the overlay parameter mechanism have made it possible to include the SPI bus choice as a parameter or parameters. Look at the `maxtherm` overlay to see how it's done.

arch/arm/boot/dts/overlays/README
marckleinebudde: ok

arch/arm/boot/dts/overlays/README
marckleinebudde: It's a pain to debug, but now it's working.

What's the best way to enable and configure the SPI chip selects for spi1 and spi2? I've found this in the README:
```
        For devices on spi1 or spi2, the interfaces should be enabled
        with one of the spi1-1/2/3cs and/or spi2-1/2/3cs overlays.
```

pelwell: Do what maxtherm does, i.e. assume that spi1 and spi2 have been enabled and configured as necessary using one of the spi1-* and spi2-* overlays. All this overlay has to do is disabled any spidev instance on the CS it is using.

You can push what you've got so far if you'd like me to look it over.

arch/arm/boot/dts/overlays/README
```diff
@@ -1762,6 +1762,23 @@ Params: oscillator              Clock frequency for the CAN controller (Hz)
 
         interrupt               GPIO for interrupt signal
```
pelwell, line 1764: There should be two blank lines between entries.

arch/arm/boot/dts/overlays/README
```diff
@@ -1762,6 +1762,23 @@ Params: oscillator              Clock frequency for the CAN controller (Hz)
...
+        speed                   Maximum SPI frequence (Hz)
+
+        interrupt               GPIO for interrupt signal
+
+        rx_interrupt            GPIO for RX interrupt signal (nINT1) (optional)
+
+        xceiver_enable          GPIO for CAN transceiver enable (optional)
+
+        xceiver_active_high     specifiy if CAN transceiver enable pin is
+                                active high (optional, default: active low)
```
pelwell, line 1783: And here as well.

pelwell (commented):
Apart from the missing blank lines, that looks good.

arch/arm/boot/dts/overlays/README
marckleinebudde: done

arch/arm/boot/dts/overlays/README
marckleinebudde: done

pelwell:
Thanks - I added the `spi<m>-<n>` parameter to the README and merged it offline.

PR #3950, opened 2020-11-09
vc4_hdmi: Report that 3d/stereo is allowed

popcornmix:
You can query stereo (3d) modes through DRM, e.g. kodi does here:
https://github.com/xbmc/xbmc/blob/5b06cdf3e64347085484435f2a4963e919edc8ec/xbmc/windowing/gbm/drm/DRMUtils.cpp#L383

currently none are reported as we don't report the capability.

### DIFF
--- a/drivers/gpu/drm/vc4/vc4_hdmi.c
+++ b/drivers/gpu/drm/vc4/vc4_hdmi.c
@@ -564,6 +564,7 @@ static int vc4_hdmi_connector_init(struct drm_device *dev,
 
 	connector->interlace_allowed = 1;
 	connector->doublescan_allowed = 0;
+	connector->stereo_allowed = 1;
 
 	drm_connector_attach_encoder(connector, encoder);
 


6by9:
While this will avoid filtering out 3D modes from the EDID, does DRM really not want to specify anything different to produce 3D output? 

When I'd looked I couldn't find any userspace code that tried to use 3D modes, so I couldn't tell what was wanted there.

pelwell:
I was just about to merge this on the basis that it compiles, and popcornmix wouldn't be requesting it if it wasn't useful.

popcornmix:
Kodi's code isn't working yet, but this appears to be step one.
We enumerate the extra modes with flags for SBS, TAB or FP based on edid (I can see this part working),
and can then request a 3d mode and drm will (I hope) set the infoframe to signal 3d so TV switches over (not tested this yet).

I would assume we end up with a double size framebuffer (or 2*H+45 for FP), but haven't confirmed that yet.

It's possible more is needed (in kernel) but I'd have thought that is all generic code so may have been dealt with.

I believe this change does nothing if you don't call:
`drmSetClientCap(m_fd, DRM_CLIENT_CAP_STEREO_3D, 1);`

so probably won't cause an issue for non-3d users.

6by9:
A grep for ```DRM_MODE_FLAG_3D``` in drivers/gpu/drm will show that there are at least a couple of mode tweaks required to validate stereo modes.

Infoframes are generated by drivers/video/hdmi.c. It looks like [hdmi_vendor_infoframe_*](https://github.com/torvalds/linux/blob/master/drivers/video/hdmi.c#L484) will set up some of the 3D stuff for you. It doesn't look like vc4_hdmi.c sets the vendor infoframe at all, but I could have missed it.

Issue #3777, opened 2020-08-06
official 7-inch touchscreen polling rate too low


ardera:
**Describe the bug**
The rate of input events arriving at userspace when using the Official 7inch touchscreen is too low. Input events arrive at user space at about 33Hz, resulting in perceived lagginess of touch applications, even though the application is running at 60fps.

This bug was present once before, see #3227, so this is kinda a regression. I'm not sure exactly what commit/update caused it to go back down to 33Hz though.

**To reproduce**
See #3227 

**Expected behaviour**
Expected is at least 60Hz, possibly even higher (since touchscreen controller seems to support 100Hz, and higher polling rates result in less input lag. All iPhones past iPhone X for example poll at 120Hz, some android gaming phones at 240Hz)

**Actual behaviour**
The frequency of input events arriving at userspace is 33Hz.

**System**
```
System Information
------------------

Raspberry Pi 4 Model B Rev 1.1
PRETTY_NAME="Raspbian GNU/Linux 10 (buster)"
NAME="Raspbian GNU/Linux"
VERSION_ID="10"
VERSION="10 (buster)"

Raspberry Pi reference 2019-09-26
Generated using pi-gen, https://github.com/RPi-Distro/pi-gen, 80d486687ea77d31fc3fc13cf3a2f8b464e129be, stage2

Linux hpi4 5.4.51-v7l+ #1327 SMP Thu Jul 23 11:04:39 BST 2020 armv7l GNU/Linux
Revision        : b03111
Serial          : 100000001afe70ee
Model           : Raspberry Pi 4 Model B Rev 1.1
Throttled flag  : throttled=0x0
Camera          : supported=0 detected=0

Videocore information
---------------------

Jul 17 2020 10:59:17
Copyright (c) 2012 Broadcom
version 21a15cb094f41c7506ad65d2cb9b29c550693057 (clean) (release) (start)

alloc failures:     0
compactions:        0
legacy block fails: 0

Filesystem information
----------------------
Filesystem     1K-blocks    Used Available Use% Mounted on
/dev/root       30391116 8174312  20946604  29% /
devtmpfs          732824       0    732824   0% /dev
tmpfs             864920       0    864920   0% /dev/shm
tmpfs             864920    8960    855960   2% /run
tmpfs               5120       4      5116   1% /run/lock
tmpfs             864920       0    864920   0% /sys/fs/cgroup
/dev/mmcblk0p1    258095   54461    203634  22% /boot
[...]

PR #3954, opened 2020-11-10
KMS DSI 7" panel support

6by9:
This gets the DSI panel working on Pi3. Pi4 is still a work in progress.

The touchscreen is using the upstream driver for the TS controller chip, but that mandates that the interrupt line be wired up, and it isn't on our panel.
For some reason that driver can't detect the TS controller if loaded during boot (ie ```dtoverlay=edt-ft5406``` in config.txt), but is fine from the command line. The reason is not understood at present.

(I'm just creating the matching PR for 5.10 as well).

### DIFF
--- a/drivers/input/touchscreen/edt-ft5x06.c
+++ b/drivers/input/touchscreen/edt-ft5x06.c
@@ -69,6 +69,8 @@
 #define EDT_RAW_DATA_RETRIES		100
 #define EDT_RAW_DATA_DELAY		1000 /* usec */
 
+#define POLL_INTERVAL_MS		17	/* 17ms = 60fps */
+
 enum edt_pmode {
 	EDT_PMODE_NOT_SUPPORTED,
 	EDT_PMODE_HIBERNATE,
@@ -126,6 +128,9 @@ struct edt_ft5x06_ts_data {
 
 	struct edt_reg_addr reg_addr;
 	enum edt_ver version;
+
+	struct timer_list timer;
+	struct work_struct work_i2c_poll;
 };
 
 struct edt_i2c_chip_data {
@@ -273,6 +278,22 @@ static irqreturn_t edt_ft5x06_ts_isr(int irq, void *dev_id)
 
 out:
 	return IRQ_HANDLED;
+}
+
+static void edt_ft5x06_ts_irq_poll_timer(struct timer_list *t)
+{
+	struct edt_ft5x06_ts_data *tsdata = from_timer(tsdata, t, timer);
+
+	schedule_work(&tsdata->work_i2c_poll);
+	mod_timer(&tsdata->timer, jiffies + msecs_to_jiffies(POLL_INTERVAL_MS));
+}
+
+static void edt_ft5x06_ts_work_i2c_poll(struct work_struct *work)
+{
+	struct edt_ft5x06_ts_data *tsdata = container_of(work,
+			struct edt_ft5x06_ts_data, work_i2c_poll);
+
+	edt_ft5x06_ts_isr(0, tsdata);
 }
 
 static int edt_ft5x06_register_write(struct edt_ft5x06_ts_data *tsdata,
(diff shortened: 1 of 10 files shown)


6by9:
I think I now understand the touchpanel dtoverlay issue.

The Atmel on the daughterboard includes a line to the touchscreen reset line. Default is to hold the TS in reset, and the REG_POWERON command will release that reset.

The touchscreen driver knows nothing about the Atmel, therefore it's a race as to which gets initialised first. Generally the TS is first and hence the TS is still in reset.
Time to add a load more dependencies :-/ I seem to recall there was a thread on dri-devel recently that was looking at something around this lot.

6by9:
Found. It was adding it as a regulator controller, which will probably do the trick.

They're in 5.10
https://github.com/torvalds/linux/commit/9c535960891470eaec5664eed2a56dbb2b6a205b
https://github.com/torvalds/linux/commit/38573472278da49e2ab588062d88bc2f0c11a325


There's also patches to make the TC358762 an independent driver, which are in 5.10.
https://github.com/torvalds/linux/commit/1930d19cb49930b7d5be0379be9f88fcbeba9976
https://github.com/torvalds/linux/commit/97722439eeed2f2e4c1a1d862faa0ee7bc2783c3

Now to rework it all to see if we can get the DT side to match up!
I think that'll be a separate PR to use the newer drivers, as this is doing a reasonable job at present.

pelwell:
For obvious reasons this has the same review feedback as #3955

drivers/input/touchscreen/edt-ft5x06.c
```diff
@@ -1221,17 +1242,27 @@ static int edt_ft5x06_ts_probe(struct i2c_client *client,
...
+			dev_err(&client->dev, "Unable to request touchscreen IRQ.\n");
+			return error;
+		}
+	} else {
+		INIT_WORK(&tsdata->work_i2c_poll,
+			  edt_ft5x06_ts_work_i2c_poll);
+		timer_setup(&tsdata->timer, edt_ft5x06_ts_irq_poll_timer, 0);
+		tsdata->timer.expires = jiffies +
+					msecs_to_jiffies(POLL_INTERVAL_MS);
+		add_timer(&tsdata->timer);
```
vianpl, line 1265: Not that important since it won't affect functionality, but would make the patch simpler. There is an API for polled input devices:

```
input_setup_polling()
input_set_poll_interval()
...
```

vianpl:
@6by9 Any plans on upstreaming this series?

drivers/input/touchscreen/edt-ft5x06.c
6by9: Does it work? #3777 is still under investigation for the firmware based TS driver where it isn't polling at the rate requested.

6by9:
@vianpl 
> Any plans on upstreaming this series?

You generally know our approach in offloading the upstreaming to others.
Maxime is contracted to look after the KMS driver side, so we'll probably ask him to look at upstreaming it when it's all working. 
Getting Pi4 DSI working is now my main priority.

A chunk of this is also likely to be superceded by the TC358762 and Pi display regulator driver from 5.10, but one step at a time.

drivers/input/touchscreen/edt-ft5x06.c
vianpl: Sorry, I wasn't aware of that issue. It's indeed very bad with HZ=100 and only barely acceptable with HZ=250 (upstream arm64's default).  I guess a workaround would be to round down the jiffies count. I wonder if moving the implementation to hrtimers when the polling delay has the same order of magnitude as HZ's delta makes sense.

vianpl:
> You generally know our approach in offloading the upstreaming to others.

Of course, but this being a smaller driver I was unsure whether to add it to my TODO list. :)

> Maxime is contracted to look after the KMS driver side, so we'll probably ask him to look at upstreaming it when it's all working.
> Getting Pi4 DSI working is now my main priority.

That's perfect, thanks!

6by9:
Updated to address Phil's review comments on #3955, and as Pi4 is now at least partially working.

Sending DSI commands seems to be slightly unreliable and sometimes times out, so the TC358762 isn't always configured fully. The same is true on Pi3, so I think the host_transfer part of the DSI driver needs a closer examination.

jamwu:
@6by9 Thanks for your great job! I am trying to used dsi1 on cm4. Which kernel should I choose? 5.9 or 5.10. Is the KMS driver ready for cm4 now?

pelwell:
5.10 is going to be our next "master" kernel - I suggest you try it and report any problems you encounter.

6by9:
I was developing on 5.9, but all patches should now be in 5.10 as well. All future changes will be on 5.10.
I haven't tested on the CM4 as yet, but will be doing so in the next couple of days.

jamwu:
> 5.10 is going to be our next "master" kernel - I suggest you try it and report any problems you encounter.

Thanks for your advice. I will give it a try.

jamwu:
> I was developing on 5.9, but all patches should now be in 5.10 as well. All future changes will be on 5.10.
> I haven't tested on the CM4 as yet, but will be doing so in the next couple of days.

Understand. Hope to hear the good news about CM4 next month.

PR #3955, opened 2020-11-10
KMS DSI 7" support - 5.10

6by9:
Duplicate #3954 for 5.10.

### DIFF
--- a/arch/arm/boot/dts/overlays/README
+++ b/arch/arm/boot/dts/overlays/README
@@ -2935,6 +2935,14 @@ Params: cma-512                 CMA is 512MB (needs 1GB)
         cma-default             Use upstream's default value
 
 
+Name:   vc4-kms-dsi-7inch
+Info:   Enable the Raspberry Pi DSI 7" screen.
+        Use edt-ft5406 for the touchscreen element.
+        Requires vc4-kms-v3d to be loaded.
+Load:   dtoverlay=vc4-kms-dsi-7inch
+Params: <None>
+
+
 Name:   vc4-kms-kippah-7inch
 Info:   Enable the Adafruit DPI Kippah with the 7" Ontat panel attached.
         Requires vc4-kms-v3d to be loaded.
--- a/arch/arm/boot/dts/overlays/README
+++ b/arch/arm/boot/dts/overlays/README
@@ -741,6 +741,20 @@ Params: int_pin                 GPIO used for INT (default 25)
         speed                   SPI bus speed (default 12000000)
 
 
+Name:   edt-ft5406
+Info:   Overlay for the EDT FT5406 touchscreen on the CSI/DSI I2C interface.
+        This works with the Raspberry Pi 7" touchscreen when not being polled
+        by the firmware.
+        You MUST use either "disable_touchscreen=1" or "ignore_lcd=1" in
+        config.txt to stop the firmware polling the touchscreen.
+Load:   dtoverlay=edt-ft5406,<param>=<val>
+Params: sizex                   Touchscreen size x (default 800)
+        sizey                   Touchscreen size y (default 480)
+        invx                    Touchscreen inverted x axis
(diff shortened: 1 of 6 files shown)


arch/arm/boot/dts/overlays/README
```diff
@@ -741,6 +741,20 @@ Params: int_pin                 GPIO used for INT (default 25)
         speed                   SPI bus speed (default 12000000)
 
 
+Name:   edt-ft5406
```
pelwell, line 744: You're off-by-one on the ordering, but otherwise this looks OK.

arch/arm/boot/dts/overlays/edt-ft5406-overlay.dts
```diff
@@ -0,0 +1,44 @@
+/*
+ * Device Tree overlay for RaspberryPi 7" Touchscreen panel
+ *
+ */
+
+/dts-v1/;
+/plugin/;
+
+/ {
+	compatible = "brcm,bcm2835", "brcm,bcm2708", "brcm,bcm2709";
```
pelwell, line 10: Just "brcm,bcm2835", please.

arch/arm/boot/dts/overlays/edt-ft5406-overlay.dts
```diff
@@ -0,0 +1,44 @@
...
+			#address-cells = <1>;
+			#size-cells = <0>;
+			status = "okay";
+			ft5406: ts@38 {
+				compatible = "edt,edt-ft5406";
+				reg = <0x38>;
+
+				touchscreen-size-x = < 800 >;
+				touchscreen-size-y = < 480 >;
+				touchscreen-inverted-x = < 0 >;
```
pelwell, line 24: Are these "inverted" flags booleans or integer properties? The defaults and the parameters disagree.

arch/arm/boot/dts/overlays/vc4-kms-dsi-7inch-overlay.dts
```diff
@@ -0,0 +1,56 @@
+/*
+ * Device Tree overlay for RaspberryPi 7" Touchscreen panel
+ *
+ */
+
+/dts-v1/;
+/plugin/;
+
+/ {
+	compatible = "brcm,bcm2835", "brcm,bcm2708", "brcm,bcm2709";
```
pelwell, line 10: "brcm,bcm2835" again.

arch/arm/boot/dts/overlays/edt-ft5406-overlay.dts
6by9: Bools. That's present vs non-present then, isn't it.

pelwell: It is, annoyingly (if you're an overlay trying to set it to false and failing because you can't delete properties).

6by9: Not an issue here as we just don't create it in this fragment, and let the override create it necessary, but yes, a general annoyance.

6by9:
Sorry 5.9 was my working branch, so I've updated that one first. Please review there and then I'll pull them across to 5.10 and update this one.

pelwell:
Updates to #3954 applied here, and merged offline.

6by9:
@pelwell you said you applied this offline. I see no evidence of the an equivalent change to 6cabe22 or 5c2aeee in the rpi-5.10.y branch.

pelwell:
There are only 5 commits here, so I updated them and applied them. I can merge the other four if you want - they're queued up here to go - but that wasn't the brief.

6by9:
I was going to update this branch once 5.9 was reviewed and merged.
You took that process out of my hands, so I had assumed you had duplicated #3954.
Yes, please merge the others.

pelwell:
Done.

PR #3956, opened 2020-11-12
Fix issue 2551

jboero:
Kernel null dereference occasionally (race condition) when using certain webcams.  Need to check before deref.  Affects mainline Linux kernel too.

### DIFF
--- a/drivers/usb/core/usb.c
+++ b/drivers/usb/core/usb.c
@@ -273,9 +273,13 @@ EXPORT_SYMBOL_GPL(usb_find_alt_setting);
 struct usb_interface *usb_ifnum_to_if(const struct usb_device *dev,
 				      unsigned ifnum)
 {
-	struct usb_host_config *config = dev->actconfig;
+	struct usb_host_config *config = NULL;
 	int i;
 
+	if (!dev)
+		return NULL;
+	
+	config = dev->actconfig;
 	if (!config)
 		return NULL;
 	for (i = 0; i < config->desc.bNumInterfaces; i++)


drivers/usb/core/usb.c
```diff
@@ -273,9 +273,13 @@ EXPORT_SYMBOL_GPL(usb_find_alt_setting);
 struct usb_interface *usb_ifnum_to_if(const struct usb_device *dev,
 				      unsigned ifnum)
 {
-	struct usb_host_config *config = dev->actconfig;
+	struct usb_host_config *config = NULL;
 	int i;
 
+	if (!dev)
+		return NULL;
```
PeterPablo, line 280: I'm really no linux hacker, but wouldn't it make sense to move this check all the way up to the opening curly brace? Initializing the struct and integer is of no use, if the device is NULL.

PeterPablo:
Did you already take your proposal upstream? This is what you'll be asked. Once it is accepted upstream, the commit could be applied here.
edit: sorry, I haven't seen the discussion that happened in #2551

drivers/usb/core/usb.c
popcornmix: The code is fine. You need all declarations to precede code in the C standard kernel targets.

pelwell:
As @popcornmix says, the commit itself looks fine, although we'd rather not have merge requests in PRs. As also said, we'd prefer to see it go upstream first, but if you get no response there come back and give us a nudge.

jboero:
Hi yes I submitted upstream.  Still waiting to hear back from the usb maintainers.  Please hold on this for now. Hoping to get it upstream.

drivers/usb/core/usb.c
jboero: I wanted to do that too but C best practices - declare variables first.  Some compilers may have issue with the order.

jboero:
Upstream PR is here:
https://github.com/torvalds/linux/commit/a40519014549f60969c8e67a2fd91426db05fe04

6by9:
You do realise that all mainline patches go via the mailing lists, and not as PRs to Linus' repo.

The patch subject line should also start with the subsystem that it applies to, ie "usb: core:" in this case.

https://www.kernel.org/doc/html/latest/process/submitting-patches.html

jboero:
Yes sorry I'll let the mailing list deal with it.  At least they can see the code changes necessary and the explanation.

Issue #20, opened 2012-05-12
JIRA SW-5991


nomis:
Could you provide an explanation of the workaround that references JIRA SW-5991?

```
diff --git a/arch/arm/mm/proc-v6.S b/arch/arm/mm/proc-v6.S
index 5900cd5..952736d 100644
--- a/arch/arm/mm/proc-v6.S
+++ b/arch/arm/mm/proc-v6.S
@@ -73,10 +73,19 @@ ENDPROC(cpu_v6_reset)
  *
  * IRQs are already disabled.
  */
+
+/* See jira SW-5991 for details of this workaround */
 ENTRY(cpu_v6_do_idle)
-   mov r1, #0
-   mcr p15, 0, r1, c7, c10, 4      @ DWB - WFI may enter a low-power mode
-   mcr p15, 0, r1, c7, c0, 4       @ wait for interrupt
+   .align 5
+   mov     r1, #2
+1: subs    r1, #1
+   nop
+   mcreq   p15, 0, r1, c7, c10, 4      @ DWB - WFI may enter a low-power mode
+   mcreq   p15, 0, r1, c7, c0, 4       @ wait for interrupt
+   nop
+   nop
+   nop
+   bne 1b
    mov pc, lr

 ENTRY(cpu_v6_dcache_clean_area)
```

raspberrypi:
Entering WFI state with an outstanding instruction cache read in
flight can hang the ARM core, so you have to be very careful to
preload the region following the WFI into the cache before going to
sleep. We do this by going round a loop once with the mcr instructions
conditionally disabled before doing it for real.

On Sat, May 12, 2012 at 8:39 PM, Simon Arlott
reply@reply.github.com
wrote:

swarren:
Is there an ARM bug number that can be referenced when upstreaming this code, or is is somehow specific to Broadcom's implementation?

raspberrypi:
I don't have an ARM bug number for this, I'm afraid, and I can't
actually find the errata list any more on ARM's website. I debugged
this one myself, and I'm certain that it will affect other ARM11
implementations. I believe the code in proc_v6.S was provided to us by
ARM, so I know they're aware of the problem.

Eben

On Fri, May 25, 2012 at 3:27 AM, Stephen Warren
reply@reply.github.com
wrote:

PR #3958, opened 2020-11-12
WIP: vc4: Enable hdmi vendor infoframe

popcornmix:


popcornmix:
Marking as WIP until I work out what else is needed.

With this, we do generate a valid vendor infoframe with 3d parameters.
And Kodi will make the 3D signal on my TV come on.

However the display image is garbled.
Need to work out if this is a clocking type issue or framebuffer geometry issue.

drivers/gpu/drm/vc4/vc4_hdmi.c
```diff
@@ -689,6 +689,30 @@ static void vc4_hdmi_set_spd_infoframe(struct drm_encoder *encoder)
...
+	struct drm_crtc *crtc = encoder->crtc;
+	const struct drm_display_mode *mode = &crtc->state->adjusted_mode;
+	union hdmi_infoframe frame;
+	int ret;
+
+	ret = drm_hdmi_vendor_infoframe_from_display_mode(&frame.vendor.hdmi,
+							  connector,
+							  mode);
+	if (ret < 0) {
+		DRM_ERROR("couldn't fill HDMI infoframe (%d)\n", ret);
```
6by9, line 705: https://elixir.bootlin.com/linux/latest/source/drivers/gpu/drm/drm_edid.c#L5809
>   Note that there's is a need to send HDMI vendor infoframes only when using a
  4k or stereoscopic 3D mode. So when giving any other mode as input this
  function will return -EINVAL, error that can be safely ignored.

Logging an error is therefore a bad move.

drivers/gpu/drm/vc4/vc4_hdmi.c
```diff
@@ -689,6 +689,30 @@ static void vc4_hdmi_set_spd_infoframe(struct drm_encoder *encoder)
...
+		return;
+	}
+	ret = hdmi_vendor_infoframe_check(&frame.vendor.hdmi);
```
6by9, line 708: Seeing as we've only just generated this, is there a possibility that it is wrong? It seems to vary between drivers whether they check this or not.

popcornmix: I was looking at intel_hdmi which does do this (for this infoframe and others).
Happy to skip it if it's believed redundant at this point.

I also noticed we aren't adding drm_hdmi_infoframe_set_hdr_metadata which we may want in future
(although I suspect Maxime may be adding this in 10-bit colour work.).

6by9: i915 is a pretty good reference :-)
Even so, I think it's impossible.
Looking at ```hdmi_vendor_infoframe_check_only``` - https://elixir.bootlin.com/linux/latest/source/drivers/video/hdmi.c#L515, 
```type```, ```version```, and ```oui``` will have been set by ```hdmi_vendor_infoframe_init```
```frame->length``` has been set to ```hdmi_vendor_infoframe_length(frame)```
That only leaves ```frame->vic != 0 && frame->s3d_struct != HDMI_3D_STRUCTURE_INVALID```. 
```frame->vic``` is set by ```drm_mode_hdmi_vic``` which includes ```if (mode->flags & DRM_MODE_FLAG_3D_MASK) return 0;```, so that should ensure it is 0 in any 3D mode.

HDR patches are in progress, but yes that is the right patchset to be adding the infoframe. There's no point in sending it if you don't support HDR.

**Comment by popcornmix** [comment]
I think I'm happy with the half-SBS and half-TAB modes doing the right thing (they trigger the right 3d mode on TV and poking kodi I can get a framebuffer that matches).

I've added an enable 3d flag to modetest. It will switch to the half-SBS/half-TAB modes and display a testcard (not in 3d).
It doesn't like the FP mode and gets an ENOSPC when choosing it. Tracked it to here:
```
[ 1340.665441] [drm:drm_ioctl [drm]] comm="modetest" pid=1823, dev=0xe201, auth=1, DRM_IOCTL_MODE_SETCRTC
[ 1340.665957] [drm:drm_mode_setcrtc [drm]] [CRTC:72:crtc-3]
[ 1340.666664] [drm:drm_framebuffer_check_src_coords [drm]] Invalid source coordinates 1920.000000x2205.000000+0.000000+0.000000 (fb 1920x1080)
[ 1340.666744] [drm:drm_mode_object_put.part.0 [drm]] OBJ ID: 222 (2)
[ 1340.667139] [drm:drm_ioctl [drm]] comm="modetest", pid=1823, ret=-28
```
I like the 1920x2205 (that's a FP framebuffer with 45 lines of padding between left and right). Not sure why the dimensions are not matching. Possibly an issue in modetest? But it can otherwise switch resolutions....

**Comment by 6by9** [comment]
The source rectangle being requested is bigger than the frame buffer, or effectively off into non-framebuffer memory.
https://elixir.bootlin.com/linux/latest/source/drivers/gpu/drm/drm_framebuffer.c#L76

**Comment by popcornmix** [comment]
Yes, but why? 
modetest -3 is reporting:
```
  #20 1920x1080 24.00 1920 2558 2602 2750 1080 1084 1089 1125 74250 flags: phsync, pvsync; 3d: fp; type: driver
```
I'm just passing:
```
modetest -3 -d -s 32:#20
```
so I'd assume modetest is just requesting the dimensions it has been given by drm.
The same procedure with the non-3d or half-SBS/half-TAB behaves sensibly.

6by9:
Presumably the half modes result in a framebuffer that is still 1920x1080 overall. It's asking for the wrong part of the image for each eye, but that's not going to be a failure as such.

Note that both i915 and nouveau double the clock for frame packing mode.
https://elixir.bootlin.com/linux/latest/source/drivers/gpu/drm/i915/display/intel_hdmi.c#L2239
https://elixir.bootlin.com/linux/latest/source/drivers/gpu/drm/nouveau/nouveau_connector.c#L1074
It may just be modetest not reporting it, but I'm assuming we need to do likewise. Ah, except that's only for mode validation of clock rate though, so shouldn't cause failures.

https://elixir.bootlin.com/linux/latest/source/drivers/gpu/drm/drm_modes.c#L861 appears to be where the stereo config gets expanded up vertically. 
I'd suspect that the drm framebuffer emulation doesn't note the 3d packing flag, so has only created a single image sized FB, and it complains when it then tries to display it.

PR #3980, opened 2020-11-30
drm/vc4: Correct DSI register definition

6by9:
The DSI1_PHY_AFEC0_PD_DLANE1 and DSI1_PHY_AFEC0_PD_DLANE3 register
definitions were swapped, so trying to use more than a single data
lane failed as lane 1 would get powered down.
(In theory a 4 lane device would work as all lanes would remain
powered).

Correct the definitions.


drivers/gpu/drm/vc4/vc4_dsi.c
```diff
@@ -306,11 +306,11 @@
 # define DSI0_PHY_AFEC0_RESET			BIT(11)
 # define DSI1_PHY_AFEC0_PD_BG			BIT(11)
 # define DSI0_PHY_AFEC0_PD			BIT(10)
-# define DSI1_PHY_AFEC0_PD_DLANE3		BIT(10)
+# define DSI1_PHY_AFEC0_PD_DLANE3		BIT(8)
```
pelwell: Interleaving the bit positions for two registers is confusing enough, but including some of them out of order makes it worse, even if it hides the inconsistency of the bit assignments.

6by9: It's an upstream file. Whilst it should get upstreamed, how many diffs do you wish to maintain?

pelwell: Your change isn't upstream.

pelwell: In other words, move the renumbered definitions into the correct position in the weird, interleaved sequence.

6by9: OK. Your initial comment seemed to imply you wanted the interleaving removed, which would have made it quite a large downstream patch.
Done.

PR #3982, opened 2020-12-02
KMS 10 & 12 bpc updates (5.10)

6by9:
Maxime's latest patches from mainline.

Works as long as the resulting rate remains below 340MHz, so 4k30 is going to remain at 8bpc for now.


popcornmix:
Tested and doesn't seem to break anything.
Do you know what use settings make use of `drm_connector_attach_max_bpc_property`?
i.e. how do we switch it on (and, say check on analyser)?

6by9:
You can enable via modetest
```~/drm/build/tests/modetest/modetest -M vc4 -s 32:1920x1080-60 -w 32:"max bpc":10```
Whether intentionally or not I believe you do need a mode selection to go with the property change, otherwise you don't get the required disable/reconfigure/enable required.
To the best of my knowledege bpc is not something can be switched dynamically within the pipeline.

Certainly this analyer is quite happy and reports the correct TMDS frequencies, bit depths, and GCP packets for 10 & 12bpc.

**Comment by popcornmix** [comment]
I'm struggling to see a difference with the Astro analyser. 
8, 10 and 12 bits all report as pixel clock of 148.5MHz (and a visible picture).
Adding logging to driver it does look like pixel_rate is 148.5MHz / 185.625MHz / 222.75MHz
for these cases, so it may just be this analyser doesn't show much useful.

But it doesn't break anything and is quite possibly doing something useful, so sounds good to me.

**Comment by 6by9** [comment]
Possible brain fart.
Your display has to declare that it supports greater than 8bpc in the EDID, otherwise the property will get set happily but actually end up doing nothing. It's clipped by ```drm_atomic_connector_check``` https://elixir.bootlin.com/linux/v5.10-rc6/source/drivers/gpu/drm/drm_atomic.c#L410

This analyser is showing GCP packets being sent when in 10 or 12bpc modes (does that analyser allow viewing of those?), and an increased TMDS clock speed.

**Comment by popcornmix** [comment]
I was seeing the increased clock here:
https://github.com/raspberrypi/linux/blob/faa0dcac61271116ddeeaebcb350791876fcad10/drivers/gpu/drm/vc4/vc4_hdmi.c#L989

Astro doesn't appear to show TMDS speed or GCP packets.

This was edid I was using:
```
AP///////wA9y2ILAAAAAAAVAQOAAAB4Ctr/o1hKoikXSUsAAAABAQEBAQEBAQEBAQEBAQEBAjqA
0HI4LUAQLEWAuoghAAAeAjqAGHE4LUBYLEUAuoghAAAeAAAA/ABUWC1OUjYwOQogICAgAAAA/QAX
PQ9EDwAKICAgICAgAV0CA2FyXJ+QFAUgEwQSAxECFgcVBgEeDx0OGgsZCiYkJSM4CX8HD38HFwdQ
PwbATQIAVwYAX34BZ1QAg08AAH4DDAAlALgm7wAAAACAEAYIABYYACAmKAAwNjgAQEbjBR8BAR2A
0HIcFiAQLCWAuoghAACeAAAAAAAAAAAAAAAA2w==
```
what am I looking for in edid-decode?

**Comment by 6by9** [comment]
> what am I looking for in edid-decode?

```
Vendor-Specific Data Block
    DC_36bit
    DC_30bit
```
As extracted by [```drm_parse_hdmi_deep_color_info()```](https://elixir.bootlin.com/linux/v5.10-rc6/source/drivers/gpu/drm/drm_edid.c#L4908) to set up the display's ```info->bpc```.

It does sound like the Astro just doesn't tell you about deep colour :-/

6by9:
v4 just been sent to clean up a couple of review comments.
Added WIP tag so that we pull in what is hopefully the final thing.

HiassofT:
@6by9 could you rebase onto current rpi-5.10.y and maybe update to v5?

6by9:
Rebased on rpi-5.10.y with v5.

6by9:
Annoyingly I've had to make comment on v5 9/9, so there's likely to be a v6 :-(

popcornmix:
If it's something minor (looks like handling kzalloc failure which is unlikely to be happening), then merging now and revert/cherry-pick the commit in question seems reasonable.

6by9:
It is only kzalloc failure handling, but it would be nicer to keep it as a single merge. Depends on whether you're actually close to being able to use 10 & 12bpc modes. Does adding HDR metadata actually end up with something extra working?
Otherwise I'm still trying to get scrambling working.

popcornmix:
I think this and HBR commits are just pieces that get us closer to correct HBR.
I think we also hdmi block outputting YUV.
And we want to be happy that decoded YUV video data going in to HVS is coming out of HDMI unchanged (well no worse that lsb type errors).
Whether we can do that with hvs de-gamma / convert to rgb / gamma / hdmi convert to yuv I'm not sure (I'm concerned about any clamping or truncation that is applied at each step). 
Running in a mode where all gamma and colourspace conversions are unity would be useful and should give correct video (at the expense of incorrect composited GUI).

6by9:
I thought the conclusion was that the HVS was always going to be composing to RGB but without any de-gamma / gamma conversions. The HDMI block may convert back to YUV if required.

So as long as the conversion matrices matched and we signal BT2020 to the display, then all operations are linear and you should get the right output.

Up to you whether you merge early and revert/replace, or just wait
If there's no desperate rush I'd vote for waiting as Maxime generally picks these up fairly quickly.

popcornmix:
The only concern is whether numbers saturate or are clamped in the CSC. Maybe it's not an issue.

6by9:
Updated with the latest patchset which has Reviewed-by tags for all patches now, so should get merged.
Just testing to make sure it still works!

popcornmix:
@6by9 are you happy to merge this now? There's a 5.10.1 rebase coming soon...

6by9:
I haven't had a chance to actually hook up the analyser to confirm everything, but I'm happy with it as is.

PR #3985, opened 2020-12-02
WIP: KMS DSI panel config

6by9:
Creating as an example of how to configure an ILI9881 based panel.

I haven't found a supplier of panel that is actually on the market and available to buy, but have tested a very similar overlay with a panel from a source I can't specify.

The ILI9881C driver was written for the bananapi panel originally, but it appears to be obsolete/unavailable now. This is an overlay and minor driver change to work with it. I don't know whether we actually want to merge it or quite what. Thoughts?


6by9:
Sorry, it was on top of my changes for #3982. Those bits can be ignored.

6by9:
I've messed up the source branch for this, so closing.
We have other examples of using DSI now anyway.

PR #3986, opened 2020-12-02
media: bcm2835-unicam: Clear clock state when stopping streaming

naushir:
Commit 65e08c465020d4c5b51afb452efc2246d80fd66f failed to clear the
clock state when the device stopped streaming. Fix this, as it might
again cause the same problems when doing an unprepare.


naushir:
@pelwell, sorry about that!

PR #3983, opened 2020-12-02
media: i2c: imx219: Selection compliance fixes

jmondi:
To comply with the intended usage of the V4L2 selection target when
used to retrieve a sensor image properties, adjust the rectangles
returned by the imx219 driver.

The top/left crop coordinates of the TGT_CROP rectangle were set to
(0, 0) instead of (8, 8) which is the offset from the larger physical
pixel array rectangle. This was also a mismatch with the default values
crop rectangle value, so this is corrected. Found with v4l2-compliance.

While at it, add V4L2_SEL_TGT_CROP_BOUNDS support: CROP_DEFAULT and
CROP_BOUNDS have the same size as the non-active pixels are not readable
using the selection API. Found with v4l2-compliance.

Fixes: e6d4ef7d58aa7 ("media: i2c: imx219: Implement get_selection")
Signed-off-by: Hans Verkuil <hverkuil-cisco@xs4all.nl>
[reword commit message, use macros for pixel offsets]
Signed-off-by: Jacopo Mondi <jacopo@jmondi.org>

PR #3997, opened 2020-12-09
media: i2c: imx477: Selection compliance fixes

naushir:
To comply with the intended usage of the V4L2 selection target when
used to retrieve a sensor image properties, adjust the rectangles
returned by the imx477 driver.

The top/left crop coordinates of the TGT_CROP rectangle were set to
(0, 0) instead of (8, 16) which is the offset from the larger physical
pixel array rectangle. This was also a mismatch with the default values
crop rectangle value, so this is corrected. Found with v4l2-compliance.

While at it, add V4L2_SEL_TGT_CROP_BOUNDS support: CROP_DEFAULT and
CROP_BOUNDS have the same size as the non-active pixels are not readable
using the selection API. Found with v4l2-compliance.

This commit mirrors 543790f777ba1b3264c168c653db6d415e7c983f done for
the imx219 sensor.

### DIFF
--- a/drivers/media/i2c/imx477.c
+++ b/drivers/media/i2c/imx477.c
@@ -900,8 +900,8 @@ static const struct imx477_mode supported_modes_12bit[] = {
 		.height = 3040,
 		.line_length_pix = 0x5dc0,
 		.crop = {
-			.left = 0,
-			.top = 0,
+			.left = IMX477_PIXEL_ARRAY_LEFT,
+			.top = IMX477_PIXEL_ARRAY_TOP,
 			.width = 4056,
 			.height = 3040,
 		},
@@ -924,8 +924,8 @@ static const struct imx477_mode supported_modes_12bit[] = {
 		.height = 1520,
 		.line_length_pix = 0x31c4,
 		.crop = {
-			.left = 0,
-			.top = 0,
+			.left = IMX477_PIXEL_ARRAY_LEFT,
+			.top = IMX477_PIXEL_ARRAY_TOP,
 			.width = 4056,
 			.height = 3040,
 		},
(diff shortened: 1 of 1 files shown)


naushir:
@davidplowman and @jmondi and @6by9 

This mirrors the change done for the imx219 in https://github.com/raspberrypi/linux/pull/3983

jmondi (approved):
Not sure my 'approval' has any value in this context, but the change matches what has been done for imx219

Reviewed-by: Jacopo Mondi <jacopo@jmondi.org>

drivers/media/i2c/imx477.c
```diff
@@ -983,8 +983,8 @@ static const struct imx477_mode supported_modes_10bit[] = {
 			 * rectangle once the driver is expanded to represent
 			 * its processing blocks with multiple subdevs.
 			 */
-			.left = 4,
-			.top = 0,
+			.left = 12,
```
6by9: I wonder if generally ```IMX477_PIXEL_ARRAY_LEFT + 4``` is more sensible for these values to make it more explicit.
Probably a more general question than worrying over fixing here.

6by9 (commented):
I'm happy with it, but one point for future discussion.

drivers/media/i2c/imx477.c
naushir: Agree, I did want to do it as `IMX477_PIXEL_ARRAY_LEFT + 4`, but I though it best that I match the changes in imx219.

davidplowman:
I've been running with these changes since first thing this morning, and the weird lens shading blob has vanished again!

drivers/media/i2c/imx477.c
6by9: The existing isn't always right :-)

If a sensor was added that only had a cropped mode then you wouldn't know if the offset were correct or not. Making it ```_PIXEL_ARRAY_LEFT + X``` makes it obvious that it has been considered.

naushir: Ok, will post an update.

naushir:
@popcornmix or @pelwell, feel free to merge this if you are happy with the change.

PR #3998, opened 2020-12-09
staging/bcm2835-codec: Ensure OUTPUT timestamps are always forwarded

6by9:
The firmware by default tries to ensure that decoded frame
timestamps always increment. This is counter to the V4L2 API
which wants exactly the OUTPUT queue timestamps passed to the
CAPTURE queue buffers.

Disable the firmware option.


popcornmix:
Merging as this improves behaviour of audio sync and seeking with vlc/kodi.
Needs corresponding updated firmware, but without that should behave as before.

PR #3999, opened 2020-12-09
5.10 hdmi cec rework

popcornmix:
This switches our CEC support to the version @mripard has cleaned up for upstreaming.
In testing it worked for me using kms on Pi3 and Pi4 in a range of HDMI modes.


popcornmix:
@pelwell I've attempted to disable by default and enable in the overlay but it's not working.
I think target should reference `interrupt-controller` but that's not working.

arch/arm64/configs/bcmrpi3_defconfig
```diff
@@ -1216,6 +1216,7 @@ CONFIG_MAXIM_THERMOCOUPLE=m
 CONFIG_MAX31856=m
 CONFIG_PWM_BCM2835=m
 CONFIG_PWM_PCA9685=m
+CONFIG_BRCMSTB_L2_IRQ=y
```
6by9: Why enable it for Pi3 defconfig? That seems odd?

popcornmix: Good point - I'll remove it.

popcornmix: (not entirely sure of the use of bcmrpi_3_defconfig - we use bcm2711_defconfig for official 64-bits on Pi3 and Pi4).

pelwell: We could get rid of it, except that people are using it and might want the slightly smaller kernel on Pi 3.

arch/arm/boot/dts/overlays/vc4-kms-v3d-overlay.dts
```diff
@@ -115,6 +115,20 @@
 		};
 	};
 
+	fragment@15 {
```
pelwell: This is the wrong overlay - you want vc4-kms-v3d-pi4.

popcornmix: Ah...

pelwell:
Does the overlay now successfully enable the interrupt controllers?

mripard (commented):
We should probably not merge 961cd2f98034, or introduce it back.

It was meant to ease the rebasing of my branch, but it still looks useful

popcornmix:
Added "drm/vc4: Reading the hotplug register is only valid if no GPIO defined" back.

Yes, overlay changes appear successful.
We mask CEC interrupt on firmware and interrupts work on arm when vc4 overlay is enabled.
We do not mask CEC interrupt on firmware when vc4 overlay is not enabled and firmware cec works.

popcornmix:
Updated to v2 of patches. cec now works on second hdmi port. Removed WIP.

popcornmix:
[edit] ignore that - I'm on an older firmware that doesn't disable its handling of cec interrupts

popcornmix:
I think this is good. I believe it's been accepted upstream and testing on Pi3 and Pi4 with kms worked.

popcornmix:
@pelwell okay to merge?

6by9:
I rebased my tree this morning (still 5.10.13) and I appear to not be able to enable i2c_vc or i2c_arm successfully
```
[    6.153644] i2c-bcm2835 fe804000.i2c: Could not request IRQ
```
This is the main DT change going, but I can't immediately see any issue with it.

I'm doing a quick roll-back to try and confirm. Whilst that's building I'll do an rpi-update to check 5.10.14 out.

6by9:
Sorry, I should add that it was when vc4-kms-v3d was loaded as well.

6by9:
5.10.14-v7l from rpi-update files also fail with ```Could not request IRQ```.
I get /dev/i2c-11 and /dev/i2c-12 for the HDMI devices, but not /dev/i2c-1.
Remove ```dtoverlay=vc4-kms-v3d``` from config.txt and /dev/i2c-1 returns.

6by9:
98a26692449ad74be15a25c05bb5be5396aaea4f (with a couple of WIP patches on top) is good.

PR #4002, opened 2020-12-14
V4L2 codec updates

6by9:
Whilst decode the edict is that the H264 header bytes should be with the first frame, encode appears to have a control that switches between that and putting them in a separate packet. FFmpeg chooses the latter.

Unobviously, the driver for decode has to manually kick the framework on DEC_CMD_START to clear the EOS flag. Add it.


6by9:
@popcornmix I'm assuming you're wanting to merge this to fix up seeking.

popcornmix:
Yes, been running with it for a few days.

PR #4004, opened 2020-12-15
Revert "media: videobuf2: Fix length check for single plane dmabuf queueing"

naushir:
This reverts commit 961d3b27a2aff52dda0b97d35085a743a27c2f46.

The updated length check for dmabuf types broke existing usage in v4l2
userland clients.

### DIFF
--- a/drivers/media/common/videobuf2/videobuf2-v4l2.c
+++ b/drivers/media/common/videobuf2/videobuf2-v4l2.c
@@ -118,8 +118,7 @@ static int __verify_length(struct vb2_buffer *vb, const struct v4l2_buffer *b)
 				return -EINVAL;
 		}
 	} else {
-		length = (b->memory == VB2_MEMORY_USERPTR ||
-			  b->memory == VB2_MEMORY_DMABUF)
+		length = (b->memory == VB2_MEMORY_USERPTR)
 			? b->length : vb->planes[0].length;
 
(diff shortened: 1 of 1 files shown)


naushir:
@6by9 and @jc-kynesim, let me know if you are ok with this.
Needed for libcamera clients to work correctly with kernels v5.10 and above.

6by9:
Yes, need to upstream that revert too.
I started to sort the replacement but priorities changed.

jc-kynesim:
Hi

>@6by9 and @jc-kynesim, let me know if you are ok with this.
>Needed for libcamera clients to work correctly with kernels v5.10 and above.

Shrug - I can work round the lack of this fix - I needed it for my
V4L2->VAAPI work but we aren't using that anywhere yet, and if it
becomes critical I can fix the H265 decoder to cope with multi-planar
which does have the fix in.

Regards

JC

6by9:
I'll dig out my rework. Validation in both single and multi planar cases then becomes against the underlying dmabuf size, not the size that videobuf2 thinks the buffer will be, nor what userland says.

jc-kynesim:
>I'll dig out my rework. Validation in both single and multi planar cases then becomes against the underlying dmabuf size, not the size that videobuf2 thinks the buffer will be, nor what userland says.

That sounds like what I actually want

Ta

PR #4005, opened 2020-12-15
drm/vc4: Make normalize_zpos conditional on using fkms

popcornmix:
Eric's view was that there was no point in having zpos
support on vc4 as all the planes had the same functionality.

Can be later squashed into (and fixes):
drm/vc4: Add firmware-kms mode


6by9:
LGTM

PR #4013, opened 2020-12-18
WIP: bcm2835-codec updates

6by9:
@popcornmix

Add some support for interlace decode. Needs tidying up to only apply to the decoder. GStreamer is happier with it, but objects to not being able to set colorimetry 2:0:0:0 (don't ask where it got that from). I haven't tried it against FFmpeg.
It really ought to be updated on format changed too, but that requires more plumbing. At least the ```field``` field on all the returned buffers should be correct.

The other patch is just a compliance test thing.


6by9:
Oops, pushed to an incorrect branch. Closing this PR.

PR #4016, opened 2020-12-21
Temporary reversions to fix vchiq

pelwell:
Something in the recent 5.10 upstream changes to vchiq (or possibly the lack of corresponding downstream changes) has broken bulk transfers in rpi-5.10.y. These three reversions and one fixup patch for vc-sm-cma are sufficient to get it working again while the real issue is root-caused.


naushir:
Fixes the raspistill/raspivid seg faults.

LGTM

pelwell:
Any other comments?

6by9:
I haven't gone through the whole thing, but those reverts all seem reasonable.

pelwell:
Merging on the basis that it has been seen to work, unlike the previous state.

PR #4022, opened 2020-12-22
Permanent fix for vchiq bulk transfer failures on 5.10

pelwell:
This fix replaces #4016, reverting the reversions and applying a simple patch to fix the real bug, which was that the `userdata` for a bulk transfer was not preserved in the common case where a non-blocking transfer was being performed.

@naushir,@6by9 If either of you are in a position to easily give this a spin I'd welcome the feedback, but vchiq_test in all its modes is happy (on 32-bit builds - 64-bit doesn't work with either fix...)

### DIFF
--- a/drivers/staging/vc04_services/bcm2835-audio/bcm2835-vchiq.c
+++ b/drivers/staging/vc04_services/bcm2835-audio/bcm2835-vchiq.c
@@ -122,7 +122,7 @@ static int
 vc_vchi_audio_init(struct vchiq_instance *vchiq_instance,
 		   struct bcm2835_audio_instance *instance)
 {
-	struct vchiq_service_params params = {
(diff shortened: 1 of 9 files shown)


6by9 (commented):
Looks reasonable but haven't tested it. Will do so tomorrow.

You could drop the initialisation of userdata to NULL seeing as it is now set to a value under all situations.

popcornmix:
The fix looks sane to me. Tested it fixes issue with vchiq_test -b 1000

pelwell:
I was going to leave the initialisation, but it's been removed now.

6by9:
Looks good. raspistill just captured an image for me.

naushir:
LGTM too. Raspistill and raspivid work as expected.

PR #4025, opened 2020-12-24
media: i2c: ov5647: Selection compliance fixes

Rahi374:
To comply with the intended usage of the V4L2 selection target when
used to retrieve a sensor image properties, adjust the rectangles
returned by the ov5647 driver.

The top/left crop coordinates of the TGT_CROP rectangle were set to
(0, 0) instead of (16, 16) which is the offset from the larger physical
pixel array rectangle. This was also a mismatch with the default values
crop rectangle value, so this is corrected. Found with v4l2-compliance.

While at it, add V4L2_SEL_TGT_CROP_BOUNDS support: CROP_DEFAULT and
CROP_BOUNDS have the same size as the non-active pixels are not readable
using the selection API. Found with v4l2-compliance.

Signed-off-by: Paul Elder <paul.elder@ideasonboard.com>

### DIFF
--- a/drivers/media/i2c/ov5647.c
+++ b/drivers/media/i2c/ov5647.c
@@ -606,8 +606,8 @@ static struct ov5647_mode supported_modes_8bit[] = {
 			.height = 480
 		},
 		.crop = {
-			.left = 0,
-			.top = 0,
+			.left = OV5647_PIXEL_ARRAY_LEFT,
+			.top = OV5647_PIXEL_ARRAY_TOP,
 			.width = 1280,
 			.height = 960,
 		},
@@ -632,8 +632,8 @@ static struct ov5647_mode supported_modes_10bit[] = {
 			.height = 1944
 		},
 		.crop = {
-			.left = 0,
-			.top = 0,
+			.left = OV5647_PIXEL_ARRAY_LEFT,
+			.top = OV5647_PIXEL_ARRAY_TOP,
 			.width = 2592,
 			.height = 1944
 		},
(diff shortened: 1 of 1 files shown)


6by9:
Has this been submitted upstream at all?
There's a set of 30 patches submitted for ov5647 at https://patchwork.linuxtv.org/project/linux-media/list/?series=3908 which mainly pulls from here plus some cleanups (and I really must review), but I don't see any reference for correcting the selection values.

I must confess to having got a bit lost in the upstream discussions over what each of these targets is meant to represent.

6by9:
And I'll add that most development here is now against the rpi-5.10.y branch.

jmondi:
Hi @6by9

The driver version which is on its path to upstream and should land in next release has already the crop rectangle offsets fixed (see the .crop fields in https://patchwork.linuxtv.org/project/linux-media/patch/20201119163238.783142-8-jacopo@jmondi.org/)

Do you plan to integrate the mainline driver into your downstream ? I expect it won't be an easy session of commit cherry-picking and it might take some time before the two versions align. During that time I think Paul's patch is required to have libcamera working on RPi with the camera module v1.

6by9:
Hi @jmondi 
I hadn't noted that your patches had reworked the selection calls as part of adding the new modes.
rpi-5.10.y is our main branch now, so I'll be cherry-picking your series to it once I've reviewed and sent R-B responses.

I'm happy to merge this for 5.4.y - @pelwell any objections from your side?

jmondi:
Thanks!

And happy new year ;)

6by9:
> And happy new year ;)

And a happy new year to you too. Let's hope 2021 returns slightly more to normal.

PR #4027, opened 2020-12-26
Add configuration override for MSEL0 on merus-amp overlay

shawaj:
Currently, on the [merus amp board](https://www.infineon.com/cms/en/product/evaluation-boards/kit_40w_amp_hat_zw/), the MSEL0 pin on the MA12070P is wired in to GPIO8 / SPI CE0 ... this means that if you have the SPI interface enabled on your Pi it stops the card from working.

I am working on a board using the same chip, but will move the MSEL0 pin to a different GPIO. Whilst I could make an entirely different overlay, this seems pointless when we can just provide overrides.

If you look at page 27 of the chip datasheet linked below, you will see the MSEL0 / MSEL1 config table:

image

My main question (probably for @pelwell) is the best way to deal with the MSEL1 pin (which by default on the merus amp board is tied to ground). I guess we provide an empty initial value and then allow it to be overriden?

I will update the [README](https://github.com/raspberrypi/linux/blob/rpi-5.4.y/arch/arm/boot/dts/overlays/README) for the dts updates once we have some consensus on the best thing to do for this overlay.

* Chip datasheet - [Infineon-MA12070P-DS-v01_00-EN.pdf](https://github.com/raspberrypi/linux/files/5742637/Infineon-MA12070P-DS-v01_00-EN.pdf)
* HAT datasheet - [Infineon-KIT_40W_AMP_HAT_ZW-UserManual-v01_00-EN.pdf](https://github.com/raspberrypi/linux/files/5742638/Infineon-KIT_40W_AMP_HAT_ZW-UserManual-v01_00-EN.pdf)
* HAT application notes - [Infineon-KIT_40W_AMP_HAT_ZW-ApplicationNotes-v01_00-EN.pdf](https://github.com/raspberrypi/linux/files/5742639/Infineon-KIT_40W_AMP_HAT_ZW-ApplicationNotes-v01_00-EN.pdf)

### DIFF
--- a/arch/arm/boot/dts/overlays/merus-amp-overlay.dts
+++ b/arch/arm/boot/dts/overlays/merus-amp-overlay.dts
@@ -57,4 +57,14 @@
 			status = "okay";
 		};
 	};
+
+	fragment@4 {
+		target = <&spi0>;
+		frag4: __overlay__ {
+		};
+	};
+
+	__overrides__ {
+		spioff = <&frag4>, "status=disabled";
+	};
 };
--- a/arch/arm/boot/dts/overlays/README
+++ b/arch/arm/boot/dts/overlays/README
@@ -1854,8 +1854,8 @@ Params: speed                   Display SPI bus speed
 
 Name:   merus-amp
 Info:   Configures the merus-amp audio card
-Load:   dtoverlay=merus-amp
-Params: <None>
+Load:   dtoverlay=merus-amp,<param>=<val>
+Params: spioff                  Turn SPI bus off
 
 
 Name:   midi-uart0


shawaj:
Also @pelwell is there a way we can add an override to turn SPI off in the overlay? i.e. we have something such as `dtoverlay=merus-amp,spioff` and that turns the SPI bus off?

pelwell:
It is certainly possible to disable SPI from the overlay, if you think that's a useful option. There are two ways to that:
a. with a dormant fragment that gets enabled, or
b. using the new option to assign literal values.

I think option b is cleaner:
```
    __overrides__ {
        spioff = <&spi0>, "status=disabled";
    };
```
You could use the same technique to abstract away the board changes:
```
    __overrides__ {
        v2 = <&merus_amp_pins>, "brcm,pins:0=22",
                <&other_label>, "other-assignments=whatever";
    };
```
You can extend a property with an assignment, so the MSEL1 pin can be added to the existing MSEL0 pin. If the pin function or pull is different to that of MSEL0 then you would have to also extend the relevent property.

arch/arm/boot/dts/overlays/merus-amp-overlay.dts
```diff
@@ -57,4 +57,7 @@
 			status = "okay";
 		};
 	};
+		__overrides__ {
```
pelwell: I'm not sure where the extra whitespace came from on this line, but it would be nicer without it.

shawaj: Good spot - think it is because I used the online editor, stupidly. Always seems to mess up formatting.

shawaj: Should be fixed in latest push

shawaj:
@pelwell added the SPI override - agreed that option b is cleaner!

Re the board changes, what I was thinking is that it might make sense for this MA12070P chip to have a way of defining the pins for MSEL0 and MSEL1 on the Pi so you can connect both of these pins and configure it to have pull up / down on either of these with the overlay - meaning that you can dynamically change between any of the signal configuration modes shown in the above table in the first comment of mine.

I was thinking maybe it could define merus_msel0_pin and merus_msel1_pin perhaps...but I don't know if this will actually work:
```
	fragment@1 {
		target = <&gpio>;
		__overlay__ {
			merus_msel0_pin: merus_msel0_pin {
				brcm,pins = <23>;
				brcm,function = <0>; /* in */
				brcm,pull = <2>; /* up */
			};
			merus_msel1_pin: merus_msel1_pin {
				brcm,pins = <20>;
				brcm,function = <0>; /* in */
				brcm,pull = <2>; /* up */
			};
		};
	};
```

Also, with the standard merus-amp board they have msel1 tied to GND and not to a GPIO, so not sure if there is a way to have this "off" by standard - i.e. tied to no pin, and then you override it to tie it to a pin if you want to?

Also, in that scenario, would I just add a pinctrl-1 to the following?
```
	fragment@2 {
		target = <&i2c1>;
		__overlay__ {
			#address-cells = <1>;
			#size-cells = <0>;
			status = "okay";

			merus_amp: ma120x0p@20 {
				#sound-dai-cells = <0>;
				compatible = "ma,ma120x0p";
				reg = <0x20>;
				status = "okay";
				pinctrl-names = "default";
				pinctrl-0 = <&merus_msel0_pin>;
				pinctrl-1 = <&merus_msel1_pin>;
				enable_gp-gpios = <&gpio 14 GPIO_ACTIVE_HIGH>;
				mute_gp-gpios = <&gpio 15 GPIO_ACTIVE_HIGH>;
				booster_gp-gpios = <&gpio 17 GPIO_ACTIVE_HIGH>;
				error_gp-gpios = <&gpio 23 GPIO_ACTIVE_HIGH>;
			};
		};
	};
```

Last but not least, confusingly, in the [schematic for their board](https://github.com/Infineon/merus-audio-amp-hat-zw/blob/master/schematics/main_Rev3_1.pdf) they have got MSEL0 tied to pin 23...however, they have the pinout for the Pi 40 pin header the wrong way round - i.e. they are calling 3V3 pin 2 - not pin 1 which I believe would be correct. Which means that this overlay is presumably currently assigning the wrong pin? 

@AMuszkat perhaps you could clarify this as to whether MSEL0 is actually tied to pin 23 or pin 24 on the merus amp board? I have the PCB, and it looks like it is in fact tied to pin 24 not pin 23....but it is hard to tell without the pcb design due to the vias etc.

pelwell:
Be careful not to confuse header pin number and GPIO number. The terms pin and GPIO are used interchangeably in overlays, but they all refer to the "Broadcom" GPIO numbering which then has to be translated to numbered pins on the 40-pin IDC header. The pinctrl nodes use the Broadcom numbering, so 23 means GPIO23 which is on header pin 16.

shawaj:
@pelwell but in this bit does `brcm,pins` not mean the actual physical pin number?

```
			merus_msel0_pin: merus_msel0_pin {
				brcm,pins = <23>;
				brcm,function = <0>; /* in */
				brcm,pull = <2>; /* up */
```

pelwell:
No it doesn't:
> The pinctrl nodes use the Broadcom numbering

shawaj:
@pelwell aha - ok sorry I misunderstood. So the merus_amp_pins in fragment@1 is actually pulling GPIO23 / Pin 16 up. Which is the /ERROR pin not the MSEL0 pin.

So this overlay doesn't, currently, actually do anything with GPIO11 or GPIO8 (pin 23 or 24) one of which is connected to the MSEL0 pin...unless I am missing something?

pelwell:
> So this overlay doesn't, currently, actually do anything with GPIO11 or GPIO8 (pin 23 or 24) one of which is connected to the MSEL0 pin

Yes, that's correct.

pelwell:
Note, however, that GPIO8 pulls high by default (GPIO11 pulls low).

shawaj:
@pelwell ok, for now I have updated this PR to just be for the spioff override, along with the README entry.

Let me know if that looks ok?

Guess I will speak to @AMuszkat about the other questions and how best to handle them over on their github repo and then report back. For reference - https://github.com/Infineon/merus-audio-amp-hat-zw/issues/7

pelwell:
The overlay checker has reminded me that the only references to external/global symbols can be the targets of fragments - I'm sorry, but you'll need to adjust the overlay like this:
```diff
--- a/arch/arm/boot/dts/overlays/merus-amp-overlay.dts
+++ b/arch/arm/boot/dts/overlays/merus-amp-overlay.dts
@@ -57,7 +57,14 @@
                        status = "okay";
                };
        };
+
+       fragment@4 {
+               target = <&spi0>;
+               frag4: __overlay__ {
+               };
+       };
+
        __overrides__ {
-               spioff = <&spi0>, "status=disabled";
+               spioff = <&frag4>, "status=disabled";
        };
 };
```

shawaj:
@pelwell ok updated now :-)

AMuszkat:
Hi @pelwell , @shawaj .

This is actually not working on a pi4. @shawaj have you tried it in a pi 3? For some reason the fragment in merus-amp-overlay.dts:

fragment@4 {
		target = <&spi0>;
		frag4: __overlay__ {
		};
	};

	__overrides__ {
		spioff = <&frag4>, "status=disabled";
	};

is generating an output of 3.3V on GPIO 8 and the merus amp card needs 5V there, this is why we have a pull up resistor on the board to 5V on the MSEL0 pin. If I take that fragment out and simply turn off the SPI with dtparam=spi=off on the /boot/config.txt it works perfectly fine and I'm back to having 5V on the MSEL0 pins. Any clue why these lines are generating this high value in that GPIO if it's an input?

pelwell:
The Pi GPIOs aren't 5V tolerant - you are putting your Pi at risk.

PR #4041, opened 2021-01-02
overlay: update seeed studio CAN BUS FD HAT

marckleinebudde:
There are several versions of the Seeed Studio CAN BUS FD HAT.

Add a `-v2` postfix to the new version (which is based on the mcp2518fd) and add a new overlay for the old version (based on the mcp2517fd).


marckleinebudde:
Thanks!

PR #4049, opened 2021-01-05
staging/mmal-vchiq: Fix incorrect static vchiq_instance.

6by9:
For some reason lost in history function vchiq_mmal_init used
a static variable for storing the vchiq_instance.
This value is retrieved from vchiq per instance, so worked fine
until you try to call vchiq_mmal_init multiple times concurrently
when things then go wrong.

Remove the static so that each caller gets a unique vchiq_instance.


6by9:
I possibly should have put in the commit text that this fixes the crash that seems to occur fairly frequently if running the cutdown firmware (no MMAL or VCSM services on the VPU).

pelwell:
Feel free - I'll wait.

6by9:
Done

PR #4069, opened 2021-01-12
uapi: bcm2835-isp: Add colour denoise mode config

naushir:
Add a field to the bcm2835_isp_denoise config structure to control
colour denoise operating modes.

### DIFF
--- a/drivers/staging/vc04_services/bcm2835-isp/bcm2835-isp-ctrls.h
+++ b/drivers/staging/vc04_services/bcm2835-isp/bcm2835-isp-ctrls.h
@@ -57,6 +57,11 @@ static const struct bcm2835_isp_custom_ctrl custom_ctrls[] = {
 		.size	= sizeof(struct bcm2835_isp_denoise),
 		.flags  = 0
 	}, {
+		.name	= "Colour Denoise",
+		.id	= V4L2_CID_USER_BCM2835_ISP_CDN,
+		.size	= sizeof(struct bcm2835_isp_cdn),
+		.flags  = 0
+	}, {
 		.name	= "Defective Pixel Correction",
 		.id	= V4L2_CID_USER_BCM2835_ISP_DPC,
 		.size	= sizeof(struct bcm2835_isp_dpc),
--- a/drivers/staging/vc04_services/bcm2835-isp/bcm2835-v4l2-isp.c
+++ b/drivers/staging/vc04_services/bcm2835-isp/bcm2835-v4l2-isp.c
@@ -764,6 +764,11 @@ static int bcm2835_isp_s_ctrl(struct v4l2_ctrl *ctrl)
 				    ctrl->p_new.p_u8,
 				    sizeof(struct bcm2835_isp_denoise));
 		break;
+	case V4L2_CID_USER_BCM2835_ISP_CDN:
+		ret = set_isp_param(node, MMAL_PARAMETER_CDN,
+				    ctrl->p_new.p_u8,
+				    sizeof(struct bcm2835_isp_cdn));
+		break;
 	case V4L2_CID_USER_BCM2835_ISP_SHARPEN:
 		ret = set_isp_param(node, MMAL_PARAMETER_SHARPEN,
 				    ctrl->p_new.p_u8,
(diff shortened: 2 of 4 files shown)


naushir:
@davidplowman and @6by9

Not ready to merge just yet though.

6by9:
Admittedly the uapi hasn't been officially released, but adding extra fields like this does cause problems of backwards compatibility.

What's it actually controlling? This generally maps directly onto the hardware, but I don't see any differentiation of modes there.

naushir:
> Admittedly the uapi hasn't been officially released, but adding extra fields like this does cause problems of backwards compatibility.

Yes, this is a bit of a problem.  I feel that we should be able to get away with it in this case given that we have not upstreamed yet, and libcamera is not yet had a stable release.  But if you do feel that this could cause problems, we may have to do alternative things.


> 
> What's it actually controlling? This generally maps directly onto the hardware, but I don't see any differentiation of modes there.

There are accompanying firmware changes in the works that will make things clearer.

naushir:
@6by9 and @davidplowman, can I get a review please?

davidplowman:
Hi Naush, this thing is showing up with 766 files changed and 481 commits, making me a bit scared of trying to review anything. Is there a way to see just the colour denoise changes? (Surely everything else was merged ages ago??)

6by9:
Dom has rebased the rpi-5.10.y parent branch, so naush needs fetch and rebase.
Or just look at the last one or two commits in the history.

naushir:
> Dom has rebased the rpi-5.10.y parent branch, so naush needs fetch and rebase.
> Or just look at the last one or two commits in the history.

Done.

davidplowman:
Hi Naush, I had just a couple of minor comments to consider:

1. If I interpreted the diffs correctly, did MMAL_PARAMETER_CDN seem to come in the middle of the list of enums rather than at the end? That made me slightly nervous as someone updating their firmware would have to be sure to update their linux (I suppose they would) and recompile their application (again, I expect they would). So maybe not a big deal.
2. I wondered ever so slightly about adding a strength parameter into the mmal_parameter_colour_denoise structure, as there is a possibility we might need it in future (we would hard-code a single value there for now). Or do you think it's better to wait and see on that? I'm not sure, perhaps we should leave it until we know.

naushir:
> 1. If I interpreted the diffs correctly, did MMAL_PARAMETER_CDN seem to come in the middle of the list of enums rather than at the end? That made me slightly nervous as someone updating their firmware would have to be sure to update their linux (I suppose they would) and recompile their application (again, I expect they would). So maybe not a big deal.

Yes, this is indeed the case here.  I did expect users would have to simultaneously update FW, kernel, and libcamera once this change went through.  If that does seem too restrictive, I could put this at the end of the enum.

> 2. I wondered ever so slightly about adding a strength parameter into the mmal_parameter_colour_denoise structure, as there is a possibility we might need it in future (we would hard-code a single value there for now). Or do you think it's better to wait and see on that? I'm not sure, perhaps we should leave it until we know.

I could put in a strength field as a placeholder, and set the value to a constant through libcamera for now.

naushir:
`
> > 1. I wondered ever so slightly about adding a strength parameter into the mmal_parameter_colour_denoise structure, as there is a possibility we might need it in future (we would hard-code a single value there for now). Or do you think it's better to wait and see on that? I'm not sure, perhaps we should leave it until we know.
> 
> I could put in a strength field as a placeholder, and set the value to a constant through libcamera for now.

One question on this one - should I make it a rational value like the SDN strength?

davidplowman:
Maybe it should just be a low/medium/strong thing? Which is probably not a bad reason for leaving it until such time as we have good reason for it...!

naushir:
> Maybe it should just be a low/medium/strong thing? Which is probably not a bad reason for leaving it until such time as we have good reason for it...!

Agreed, I won't make this change.

davidplowman:
Have an "LGTM" from me. I'll leave to it you if you want to change the enum order, I agree that in reality it will probably matter little.

naushir:
Latest version has the enum value at the end.

PR #4070, opened 2021-01-12
bcm2835-isp: Default to JFIF colour space for YUV420

davidplowman:
This is a temporary measure until libcamera can signal colour spaces
correctly.

### DIFF
--- a/drivers/staging/vc04_services/bcm2835-isp/bcm2835-v4l2-isp.c
+++ b/drivers/staging/vc04_services/bcm2835-isp/bcm2835-v4l2-isp.c
@@ -330,10 +330,44 @@ static void mmal_buffer_cb(struct vchiq_mmal_instance *instance,
 		complete(&dev->frame_cmplt);
 }
 
+struct colorspace_translation {
+	enum v4l2_colorspace v4l2_value;
+	u32 mmal_value;
+};
+
+u32 translate_color_space(enum v4l2_colorspace color_space)
+{
+	static const struct colorspace_translation translations[] = {
+		{ V4L2_COLORSPACE_DEFAULT, MMAL_COLOR_SPACE_UNKNOWN },
+		{ V4L2_COLORSPACE_SMPTE170M, MMAL_COLOR_SPACE_ITUR_BT601 },
+		{ V4L2_COLORSPACE_SMPTE240M, MMAL_COLOR_SPACE_SMPTE240M },
+		{ V4L2_COLORSPACE_REC709, MMAL_COLOR_SPACE_ITUR_BT709 },
+		// V4L2_COLORSPACE_BT878 unavailable
+		{ V4L2_COLORSPACE_470_SYSTEM_M, MMAL_COLOR_SPACE_BT470_2_M },
+		{ V4L2_COLORSPACE_470_SYSTEM_BG, MMAL_COLOR_SPACE_BT470_2_BG },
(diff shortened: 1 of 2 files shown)


davidplowman:
@naushir @6by9 
Not for merging just yet - needs some more discussion first.

There are 2 changes here. The first (and biggest) is to send the colour space field correctly through to the VideoCore, previously it was just being missed. The catch is that libcamera is not able to signal colour space information at all (you always get "default"), so a by-product is that everything will now end up with the SMPTE170M (BT601 limited range) colour space, even JPEGs.

To work around that the other change is to use the JPEG/JFIF (BT601 full range) colour space for YUV420. This means my libcamera apps can request YUV420 for JPEGs, and NV12 for h264. Not nice, but I'm low on alternatives. Of course HD video still isn't right, though things are less badly broken than they were.

We could slip some more code into try_fmt along the lines of "if (format == NV12 && colourspace == DEFAULT) set the colour space to SMPTE170M or BT709 according to the resolution", but that's starting to look really unpleasant. What do people think?

drivers/staging/vc04_services/bcm2835-isp/bcm2835-v4l2-isp.c
```diff
@@ -330,10 +330,44 @@ static void mmal_buffer_cb(struct vchiq_mmal_instance *instance,
 		complete(&dev->frame_cmplt);
 }
 
+struct colorspace_translation {
+	enum v4l2_colorspace v4l2_value;
+	u32 mmal_value;
+};
+
+u32 translate_color_space(enum v4l2_colorspace color_space)
```
davidplowman, line 338: Should be static, of course.

naushir:
> This means my libcamera apps can request YUV420 for JPEGs, and NV12 for h264. Not nice, but I'm low on alternatives.

Sadly, this will conflict with my change https://github.com/raspberrypi/linux/pull/4069, where I can only use YUV420 for colour denoise in fast mode :(

naushir:
>To work around that the other change is to use the JPEG/JFIF (BT601 full range) colour space for YUV420. This means my libcamera apps can request YUV420 for JPEGs, and NV12 for h264

Could you flip this around and use NV12 for JPEGS and YUV420 for H.264?

6by9:
Hardware JPEG block won't take NV12, not that libcamera is using the hardware JPEG block.

davidplowman:
Hmm yes, colour denoise rather blows this scheme out of the water. I'm not sure libjpeg will take NV12 natively, so there'd be a memcpy which would rather clobber the throughput, especially for MJPEG. I guess I could consider using YVU420... but it all seems to be getting uglier and uglier.

6by9:
I think libjpeg will take NV12/21 as Android used that as it's preferred internal format, and I'd be surprised if that then blocked using libjpeg.
I also remember a colleague of ours who always used to complain when we asked for things that diverged from the libjpeg API way of doing things (until I pointed out that his API diverged from it!)

For now default to full range 601/170M so your JPEGs are right, but that you can set all the parameters, and we'll worry about other use cases later.
The H264 encoder should be able accept full range 601/170M and represent it correctly in the headers.

davidplowman:
Thanks, that's helpful. After sleeping on this I'm inclined to do away with any attempt at subtle NV12/YUV420 hacks, and do things properly here. I'll just put up with H264 videos being in the wrong colour space until such time as libcamera catches up. I think I'll close this merge request and start a new one.

PR #4072, opened 2021-01-13
overlays: seeed-can-fd-hat: clarify how to identify HAT version

marckleinebudde:
It turns out the used CAN SPI chip is not a good way to identify the version of
the CAN HAT.

There are two different board layouts of the Seeed Studio CAN BUS FD HAT. The
v1 board doesn't have a battery holder, while the v2 board has. Update the
overlay README accordingly.

Link: https://github.com/Seeed-Studio/seeed-linux-dtoverlays/issues/13
Cc: Patrick Menschel <menschel.p@posteo.de>
Signed-off-by: Marc Kleine-Budde <mkl@pengutronix.de>

### DIFF
--- a/arch/arm/boot/dts/overlays/README
+++ b/arch/arm/boot/dts/overlays/README
@@ -2483,16 +2483,18 @@ Load:   <Deprecated>
 
 
 Name:   seeed-can-fd-hat-v1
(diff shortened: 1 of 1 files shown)


marckleinebudde:
Wow - that was quick!

PR #4073, opened 2021-01-14
bcm2835-isp: Allow formats with different colour spaces.

davidplowman:
We select from the supported formats using the colour space as well as
the fourcc, and pass the colour space over to the VideoCore.

The list of supported formats is extended to include JPEG/JFIF and
REC709 colour spaces for all the YUV-type fourccs.

### DIFF
--- a/drivers/staging/vc04_services/bcm2835-isp/bcm2835-v4l2-isp.c
+++ b/drivers/staging/vc04_services/bcm2835-isp/bcm2835-v4l2-isp.c
@@ -69,6 +69,7 @@ struct bcm2835_isp_q_data {
 	unsigned int width;
 	unsigned int height;
 	unsigned int sizeimage;
+	enum v4l2_colorspace colorspace;
 	const struct bcm2835_isp_fmt *fmt;
 };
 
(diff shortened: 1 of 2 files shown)


davidplowman:
@naushir @6by9
Here's a version of our v4l2 ISP driver that handles colour spaces properly, in the sense that it will send the colour space that it is given to the VideoCore. Once libcamera is upgraded to request the correct colour spaces (don't hold your breath), this will cause the ISP to be configured appropriately.

I'm not sure if I've implemented this in the best way. I've made the list of supported formats depend on both the fourcc and the colour space, and tweaked a couple of other functions as necessary. I thought this would end up the easiest approach, but I'm open to suggestions if there's something better.

The changes are:

1. The same translate_color_space as I had in the earlier version of this.
2. For non-stats nodes we match the format by fourcc and colour space (V4L2_COLORSPACE_DEFAULT matches the first we find).
3. I count the number of matching formats before allocating the list for each node. I guess this is a bit more expensive but only happens at setup.
4. The enum_fmt function must avoid returning duplicate fourccs, which we now have in the list. This is a bit icky, but again only happens during setup.

I've been testing this and it works as expected, but suggestions are of course welcome!

6by9:
Duplication in that formats list is horrid.

As previously discussed can we switch the colour space field in that table to be a bit ask of permitted colour spaces for that format? Each format therefore only appears once still, but you have the information available to validate the permitted cs.

I know that the mmal api doesn't allow independent setting of limited/full range and primaries, but I think we need to look at quantization and the other fields in the format, not just the colorspace enum.
Aiui V4l2 colorspace jpeg is limited to compressed jpeg only, not yuv formats. It does not mean jfif.

davidplowman:
Here's a better version. Such documentation as I found on V4L2_COLORSPACE_JPEG (e.g. https://www.kernel.org/doc/html/v4.16/media/uapi/v4l/colorspaces-details.html#colorspace-jpeg-v4l2-colorspace-jpeg) seemed to imply it was an acceptable shorthand for "full range BT601", but maybe there's something I didn't uncover?

naushir:
This change looks like a reasonable approach to me.

davidplowman:
I see that rpi-update moves people onto rpi-5.10.y now, so I don't think I care about rpi-5.4.y any more. Therefore I'm going to close this request and I'll repost it on the rpi-5.10.y branch next week. Thanks everyone!

PR #4075, opened 2021-01-14
vc4: Correct lbm size and calculation

popcornmix:
LBM base address is measured in units of pixels per cycle.
That is 4 for 2711 (hvs5) and 2 for 2708.

We are wasting 75% of lbm by indexing without the scaling.
But we were also using too high a size for the lbm resulting
in partial corruption (right hand side) of vertically
scaled images, usually at 4K or lower resolutions with more layers.

The physical RAM of LBM on 2711 is 8 * 1920 * 16 * 12-bit
(pixels are stored 12-bits per component regardless of format).

The LBM adress indexes work in units of pixels per clock,
so for 4 pixels per clock that means we have 32 * 1920 = 60K

### DIFF
--- a/drivers/gpu/drm/vc4/vc4_plane.c
+++ b/drivers/gpu/drm/vc4/vc4_plane.c
@@ -444,6 +444,7 @@ static void vc4_write_ppf(struct vc4_plane_state *vc4_state, u32 src, u32 dst)
 static u32 vc4_lbm_size(struct drm_plane_state *state)
 {
 	struct vc4_plane_state *vc4_state = to_vc4_plane_state(state);
+	struct vc4_dev *vc4 = to_vc4_dev(state->plane->dev);
 	u32 pix_per_line;
 	u32 lbm;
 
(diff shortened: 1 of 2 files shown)


drivers/gpu/drm/vc4/vc4_plane.c
```diff
@@ -479,7 +479,11 @@ static u32 vc4_lbm_size(struct drm_plane_state *state)
 		lbm = pix_per_line * 16;
 	}
 
-	lbm = roundup(lbm, 32);
+	/* Align it to 64 or 128 (hvs5) bytes */
+	lbm = roundup(lbm, vc4->hvs->hvs5 ? 128 : 64);
```
pelwell, line 484: `error: ‘vc4’ undeclared`

popcornmix: Ah - left out when ignoring debug - added back now.

pelwell: But not pushed?

popcornmix: Pushed now.

pelwell:
That now builds and seems to run OK.

6by9:
Looks reasonable to me based on the conversations we've had.

HiassofT:
Tests on LibreELEC look good and this also fixes the 4k video playback issues we had (flickering vertical lines over video).

popcornmix:
Removed draft status of PR. Been working fine on kodi on Pi3 and Pi4 over weekend, so should be good to merge.

PR #4085, opened 2021-01-20
Fixes a problem with the DAC PLL when changing sample rates

j-schambacher:
This fix for Hifiberry DAC+ and DAC+ADC PRO adds RESET of
PLL-dividers for LR- and B-CLK before every change of
sample rate to allow the PLL a clean start up.

### DIFF
--- a/sound/soc/bcm/hifiberry_dacplus.c
+++ b/sound/soc/bcm/hifiberry_dacplus.c
@@ -57,6 +57,10 @@ static bool leds_off;
 static void snd_rpi_hifiberry_dacplus_select_clk(struct snd_soc_component *component,
 	int clk_id)
 {
+	/* reset PLL divider */
+	snd_soc_component_update_bits(component, PCM512x_MASTER_MODE,
+					PCM512x_RLRK | PCM512x_RBCK, 0);
+
 	switch (clk_id) {
 	case HIFIBERRY_DACPRO_NOCLOCK:
 		snd_soc_component_update_bits(component, PCM512x_GPIO_CONTROL_1, 0x24, 0x00);
--- a/sound/soc/bcm/hifiberry_dacplusadcpro.c
+++ b/sound/soc/bcm/hifiberry_dacplusadcpro.c
@@ -176,6 +176,10 @@ static int pcm1863_add_controls(struct snd_soc_component *component)
 static void snd_rpi_hifiberry_dacplusadcpro_select_clk(
 					struct snd_soc_component *component, int clk_id)
 {
+	/* reset PLL divider */
+	snd_soc_component_update_bits(component, PCM512x_MASTER_MODE,
+					PCM512x_RLRK | PCM512x_RBCK, 0);
+
 	switch (clk_id) {
 	case HIFIBERRY_DACPRO_NOCLOCK:
 		snd_soc_component_update_bits(component,


clivem:
I'm not 100% sure what this patch is intended to fix, but in 10 mins testing with this patch applied (that was the only change), I've seen .....
```
Jan 21 19:04:37 BerryPlusPro2B1 kernel: [   25.757062] bcm2835-i2s 3f203000.i2s: I2S SYNC error!
```
.... which results in playback being stuffed until reboot, on a DAC+ Pro.

clivem:
OK, on closer look, that might be the only change to my card driver but my pcm512x codec driver is modified with upstream patches. 

Joerg, IMHO this is not a wise commit. There is a patch in Linus tree to codec driver, which when it makes it's way into the next release will break the behaviour you are depending on, for your patch.

[ASoC: pcm512x: Move format check into set_fmt()](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/sound/soc/codecs/pcm512x.c?id=26b97d95a05d0346e1ad6096deedac3f24a4607b)

In future those bits will only be programmed from the set_fmt() func, which is only going to be called once, when the driver is initialized. For your patch to work, you are depending on them being re-programmed every time the device is opened in the codec hw_params() func, and that code is being moved into set_fmt(). So if this patch is merged, you'll need to remember this needs to be looked at again when those codec driver patches make their way downstream in the future.

pelwell:
Thanks for your timely input, Clive.

j-schambacher:
Clive, thanks for testing and feedback! In principle, you are right. This commit is only a 'fix' and maybe not wise. The DAC+ driver till 4.19 did run perfectly. With moving to 5.4 we have seen problems when users play various titles with different sample rates immediately after each other. I've spent a few days tracing down the root cause and finally came to the conclusion, that switching sample rate family(44.1<>48) and a different divider ratio causes the PLL to just continue resulting in no sound.
In my opinion, the best place to put the PLL divider reset is in the pcm512x_hw_params() before programming the dividers. In the hw_params function there is only after the complete divider setting in line 1290 the reset of the dividers but with no effect if the device is already in the same mode as before. But we decided to touch the PCM driver...
Why is there a difference between the kernel versions: it seems like a different timing. Switching the sample rate family accesses the PCM512x GPIOs to switch on/off the external oscillators. In case, you do not use the PRO version or manually set the PRO to running as 'slave' the problem does not exist. Maybe, PLL programming and switching oscillators now interfere.

Question/comment on the mentioned commit in Linus' kernel:
As far as I've understood this sets the Master/Slave configuration of the whole I2S interface. This can/must be done only once in the module init. What we change with our fix is only a reset of the dividers for LR- and B-CLK every time you start the DAC. I think it addresses a different topic. 

We have released the fix in HifiberryOS early December and did not see any problem related to it. That's why we decided to release the PR.

Do you have any further comments or even a better idea how to fix it, I'm more than happy to test/implement.

clivem:
Hi Joerg, I'm still not sure what you are actually trying to fix....... (Which might just be me being thick!) My gut feeling is that this might be a "fix/workaround" for an issue Roon users noticed, when using the DAC+Pro as an endpoint...... ISTR, that the problem was that when trying to play a track at 192k, after playing a track at a lower sample rate, they got no sound? Is that it?

If it is, I was never able to reproduce that issue. I have a test suite of 10 second clip @ 32k, 44k1, 48k, 88k2, 96k, 176k4, 192k, 352k8, 384k..... Using LMS/squeezelite I couldn't reproduce. I did try to reproduce with Roon and couldn't either. But I suspect there is one important difference. I was using RoonBridge on the Pi. I since became aware that manufacturers (ie. you) with Roon tested/certified/approved hardware use something different. You have an sdk and build the endpoint code (native C ??????) for HiFiBerry OS rather than use the generic Bridge? So if this is the problem you are trying to solve, my question is does it only happen with the SDK endpoint, and not with the Bridge? Have you tested that? I wonder if the bug you are trying to solve has to do with how they are using the alsa device from the native endpoint code? Because, it's not a problem I've seen in the past 5 years of using the DAC+Pro, and I suspect that I have much more music at 192k in my library than most users, and with that DAC+Pro playing random mix 24/7, I'm getting many, many instances of 192k being played after a lower sample rate, during any 24 hour period.

j-schambacher:
Clive, that's what we thought first but we are able to reproduce it with SOX play and a list of generic test files. ROON users were only the first to notice. A simple

play file1.wav file2.wav .....

and some did not play (no sound, no error message etc. -but as far as I remember the play time was wrong -> a file with 176ksps took 4x after a 44.1ksps file -> PLL divider(s) did not update/lock?), the next after it did play again. Daniel and I just used a clean Raspberry OS without any changes. If you use

play file1
play file2
play file3
....

it never happens.

That lead me towards the timing. First I also tried to just increase some of the already available delays in the code, but without success. Then reading through the TI specs, I ended up digging into the PLL settings. BTW, if you use i2cset commands to reset the PLL during a non-playing file, you can 'activate' it.

clivem:
OK, when I get home this evening I'll try to reproduce doing what you are doing... ie. Using play command with multiple files specified at different sample rates. Timing..... I'm not in front of code right now, but could this be a simple case of needing to add a small sleep after you switch clock via GPIO in the card driver. Didn't the card driver do something cludgy already to work-around the dac chip needed time to detect the clock change..... sclk_sleep(). ie after switching clock with GPIO3/6 high in select_clk(), needing to wait a couple of millis before querying the chip to see if sclk is present, in the detection code where the driver figures if the card is a Pro? Have you tried adding something like "udelay(2000)" after the snd_soc_comp_update_bits() in the card select_clk() function, to give the DAC chip a chance to detect the clock change before the codec driver starts calculating the dividers and playback begins. Just a thought.

j-schambacher:
Clive, thanks for your help. And yes, I've played with delays (existing and at new places in the code). Maybe, I only tried the wrong ones and you have more luck. AFA I remember I did not try anything in the CLK driver as it is used by other cards as well. I've placed our audio test files on GitHub.com/j-schambacher/testfiles for your convenience.

j-schambacher:
...just another thought. I did not consider this to be related but it maybe is. According to this patch https://github.com/hoene/linux/commit/3796d961744d44aabf88d4bcf08413dfb8851210
the driver returns wrong sample rates and messes up with the 44.1/48 families. I'll have a look to it.

clivem:
Joerg, have reproduced with your test files ("play sine-24-44100-8000hz--10db.flac sine-16-192000-8000hz--10db.flac") and as I guessed above, just add a 2ms delay. Problem solved. (Might as well get rid of is_sclk_sleep() wrapper function now that the delay is moved to the select_clk() function.) Try this.....

```
--- hifiberry_dacplus.c.orig	2021-01-22 19:24:05.208519356 +0000
+++ hifiberry_dacplus.c	2021-01-22 19:25:25.967430189 +0000
@@ -68,6 +68,8 @@ static void snd_rpi_hifiberry_dacplus_se
 		snd_soc_component_update_bits(component, PCM512x_GPIO_CONTROL_1, 0x24, 0x04);
 		break;
 	}
+
+	usleep_range(2000, 2100);
 }
 
 static void snd_rpi_hifiberry_dacplus_clk_gpio(struct snd_soc_component *component)
@@ -85,13 +87,6 @@ static bool snd_rpi_hifiberry_dacplus_is
 	return (!(sck & 0x40));
 }
 
-static bool snd_rpi_hifiberry_dacplus_is_sclk_sleep(
-	struct snd_soc_component *component)
-{
-	msleep(2);
-	return snd_rpi_hifiberry_dacplus_is_sclk(component);
-}
-
 static bool snd_rpi_hifiberry_dacplus_is_pro_card(struct snd_soc_component *component)
 {
 	bool isClk44EN, isClk48En, isNoClk;
@@ -99,13 +94,13 @@ static bool snd_rpi_hifiberry_dacplus_is
 	snd_rpi_hifiberry_dacplus_clk_gpio(component);
 
 	snd_rpi_hifiberry_dacplus_select_clk(component, HIFIBERRY_DACPRO_CLK44EN);
-	isClk44EN = snd_rpi_hifiberry_dacplus_is_sclk_sleep(component);
+	isClk44EN = snd_rpi_hifiberry_dacplus_is_sclk(component);
 
 	snd_rpi_hifiberry_dacplus_select_clk(component, HIFIBERRY_DACPRO_NOCLOCK);
-	isNoClk = snd_rpi_hifiberry_dacplus_is_sclk_sleep(component);
+	isNoClk = snd_rpi_hifiberry_dacplus_is_sclk(component);
 
 	snd_rpi_hifiberry_dacplus_select_clk(component, HIFIBERRY_DACPRO_CLK48EN);
-	isClk48En = snd_rpi_hifiberry_dacplus_is_sclk_sleep(component);
+	isClk48En = snd_rpi_hifiberry_dacplus_is_sclk(component);
 
 	return (isClk44EN && isClk48En && !isNoClk);
 }
```

j-schambacher:
Clive, thanks! I'll try. I've started my analysis by increasing the msleep() delay without success. I'll also test it with our newer boards.

j-schambacher:
@clivem I've tested it and confirm it to work. I really wonder what was the difference when I tested it. I'm sure it was not successful, but maybe I screwed things up in a different place while debugging.

@pelwell As the changes are quite small and I need to modify this driver (file) again, my suggestion would be to just close this PR and I include the necessary changes into the next PR.

pelwell:
That's fine by me - it's your driver.

clivem:
@j-schambacher OK. Just FYI, it was also successfully tested by a friend of mine on Ubuntu 20.04 running the bridge for Roon with a DAC+Pro board.

PR #4086, opened 2021-01-21
Enable the aqc111 module on armhf and arm64

Omoeba:


pelwell:
A few words of description explaining what the AQC100 is, and why it isn't needed on non-BCM2711 Pis (it's a PCIe device), would have been nice, but a 25K module is a modest footprint so I'll accept it. However, you can't just hack the defconfig file by hand and submit it - you must use `make [<options>] savedefconfig` to regenerate the defconfig file, otherwise the entries are likely to be in the wrong place which makes it harder to isolate the changes resulting from a single commit.

Omoeba:
The aqc111 driver is for the aqc111u, which is a usb to ethernet chip like the ax88179 and has nothing to do with the pcie aqc100. I’m completely new to the whole kernel side of things so thanks for pointing out the correct way to make changes.

pelwell:
That's an unfortunate name collision. In which case you should also enable support in the other Pi defconfigs - bcmrpi_defconfig, bcm2709_defconfig and the arm64 bcmrpi3_defconfig.

Omoeba:
Should I also add support in older kernels like 5.4?

pelwell:
I don't expect us to make any more 5.4 releases, and development effort isn't go to be more than bug fixes, so confining your patch to 5.10 is fine (we can cherry pick to later kernels).

Omoeba:
Alright, thanks again for your help.

pelwell:
Now added to rpi-5.10.y and rpi-5.11.y.

Omoeba:
Thanks, but which options did you use to edit and save the defconfig so portions weren't being removed?

pelwell:
I use the following sequence:
1. `make ARCH=arm bcm2711_defconfig`
2. `scripts/config --set-val USB_NET_AQC111 m`
3. `make ARCH=arm savedefconfig`
4. `cp defconfig arch/arm/configs/bcm2711_defconfig`

This specific example is for 32-bit Pi 4, but it is easily modified for other platforms.

Omoeba:
Thank you

PR #4094, opened 2021-01-24
Rpi 5.10.y: Fixed OV9281 line bug and added flip controls

manhalt:
With the latest update to the OV9281 driver a bug was introduced, that will generate lines on each picture. It can only be seen by magnifying a picture. This happened by setting the register 0x3778 to 0x10 instead of 0x00. 
This PR also includes the missing v4l controls for OV9281 to set HFLIP and VFLIP. 
Tested on CM3 and Pi4 with a live stream and v4l2-ctl (e.g., v4l2-ctl --set-ctrl=horizontal_flip=1).


drivers/media/i2c/ov9281.c
```diff
@@ -177,6 +177,7 @@ static const struct regval ov9281_common_regs[] = {
 	{0x372d, 0x22},
 	{0x3731, 0x80},
 	{0x3732, 0x30},
+	{0x3778, 0x00},
```
6by9: I'll agree that the 0x3778 is incorrect for 1280x800 and 1280x720, but based on the information I have 0x3778 should be 0x10 for 640x480.

Have you tested the 640x480 mode?

https://github.com/ArduCAM/ArduCAM_USB_Camera_Shield/blob/master/Config/USB3.0_UC-425_Rev.C%2BUC-628_Rev.B%2BUC-625_Rev.B/OV9281/OV9281_1Lane/1280x800/8b/OV9281_comb_A_1280x800.cfg#L201
I don't have a full datasheet for the sensor, just a description of 0x3700-0x37AF as SENSOR CONTROL REGISTERS.

drivers/media/i2c/ov9281.c
```diff
@@ -613,6 +618,34 @@ static int ov9281_enable_test_pattern(struct ov9281 *ov9281, u32 pattern)
 				OV9281_REG_VALUE_08BIT, val);
 }
 
+static int ov9281_set_ctrl_hflip(struct ov9281 *ov9281, int value)
+{
+	u32 current_val;
+	int ret = ov9281_read_reg(ov9281->client, OV9281_REG_TIMING_FORMAT_2,
+			      OV9281_REG_VALUE_08BIT, &current_val);
+	if (!ret)
+	{
```
6by9: Kernel coding style is for the brace to be on the end of the preceding line.

drivers/media/i2c/ov9281.c
```diff
@@ -613,6 +618,34 @@ static int ov9281_enable_test_pattern(struct ov9281 *ov9281, u32 pattern)
...
+	return ret;
+}
+
+static int ov9281_set_ctrl_vflip(struct ov9281 *ov9281, int value)
+{
+	u32 current_val;
+	int ret = ov9281_read_reg(ov9281->client, OV9281_REG_TIMING_FORMAT_1,
+			      OV9281_REG_VALUE_08BIT, &current_val);
+	if (!ret)
+	{
```
6by9: Ditto

drivers/media/i2c/ov9281.c
```diff
@@ -979,7 +1018,7 @@ static int ov9281_initialize_controls(struct ov9281 *ov9281)
 
 	handler = &ov9281->ctrl_handler;
 	mode = ov9281->cur_mode;
-	ret = v4l2_ctrl_handler_init(handler, 8);
+	ret = v4l2_ctrl_handler_init(handler, 10);
```
6by9: My mistake previously, but I make it that there are now 9 controls, not 10.

drivers/media/i2c/ov9281.c
```diff
@@ -613,6 +618,34 @@ static int ov9281_enable_test_pattern(struct ov9281 *ov9281, u32 pattern)
...
 }
 
+static int ov9281_set_ctrl_hflip(struct ov9281 *ov9281, int value)
+{
+	u32 current_val;
+	int ret = ov9281_read_reg(ov9281->client, OV9281_REG_TIMING_FORMAT_2,
+			      OV9281_REG_VALUE_08BIT, &current_val);
+	if (!ret)
+	{
+		current_val ^= (-value ^ current_val) & (1 << 2);
```
6by9: I don't really even want to think through this one.

negative signed value XOR with an unsigned value, and then logical AND, all XOR with the number you first thought of.

Simple logic of always clearing the bit (use a ```#define```), and then conditionally setting it please (reuse same ```#define```).

drivers/media/i2c/ov9281.c
```diff
@@ -613,6 +618,34 @@ static int ov9281_enable_test_pattern(struct ov9281 *ov9281, u32 pattern)
...
+}
+
+static int ov9281_set_ctrl_vflip(struct ov9281 *ov9281, int value)
+{
+	u32 current_val;
+	int ret = ov9281_read_reg(ov9281->client, OV9281_REG_TIMING_FORMAT_1,
+			      OV9281_REG_VALUE_08BIT, &current_val);
+	if (!ret)
+	{
+		current_val ^= (-value ^ current_val) & (1 << 2);
```
6by9: Ditto

6by9 (commented):
Your flip patch reminded me that I hadn't merged my near identical one.
https://github.com/6by9/linux/commit/a2442a547a461168b35d742f8fcbac396185b324

I haven't passed your patches through checkpatch.pl. That would have picked up the error in brace placement, and checked indentation. You're also missing a ```Signed-off By:``` line from the patches.

Register 0x3820 and 0x3821 are covered in my datasheet.
0x3821 never has any other bits set, so using 0 and ```BIT(2)``` as the options is fine. 
0x3820 has bit 6 set by default, so reading it back is just about justified, although jsut taking 0x40 as the alternate value would also be reasonable.

drivers/media/i2c/ov9281.c
manhalt: Fixed it to 9, I really didn't count them, just added +2

drivers/media/i2c/ov9281.c
manhalt: Sorry for this one, it is a bit tricky. I saw you are using the BIT() macro - where does it come from and is there more help like this? I added macros to SET and UNSET a bit (not sure if this already exists), because this a typical task in these drivers I would say.

drivers/media/i2c/ov9281.c
manhalt: I checked carefully the effect by taking pictures with this register set to 0x10 and to 0x00. In the 1280x* modes it will create lines (I think every second line is grey if that makes any sense). 

In the 640x400 mode it does not do this, but when set, the overall brightness of the picture increases. It can be left at 0x00, but since you found evidence that it is maybe useful, I readded it to that mode. 

I guess you are looking at the same document as me? It is a leaked / scanned datasheet I would say from Innomaker? If yes, I also found the blank part on this register. It seems to be a black box and we cannot be sure if it is required to set it. 

I also found prof that it is used in 640x400 mode by NVidia: https://gitlab.incom.co/CM-Shield/android_kernel_nvidia_quill/-/blob/5ff2f17e8a9293d9f15b1a1de73139260ead1967/drivers/media/i2c/ov9281_mode_tbls.h#L333

But I have a feeling that most of the implementations are done without having the full datasheet…

manhalt:
First, thank you very much @6by9 for looking into this. As you surely noticed, this is my first attempt to put a PR. I found out about the checkpatch.pl script, sorry for putting something wrongly formatted. It really motivates me that you took the time to investigate my beginner level programming and telling me about the problems.

> Register 0x3820 and 0x3821 are covered in my datasheet.
> 0x3821 never has any other bits set, so using 0 and `BIT(2)` as the options is fine.
> 0x3820 has bit 6 set by default, so reading it back is just about justified, although jsut taking 0x40 as the alternate value would also be reasonable.

I also thought about that. I would choose to read it in both cases, because the datasheet says, that Bit[1] in both registers is used for binning. This is not implemented in the current driver, but I’m interested to see it working. 

> Your flip patch reminded me that I hadn't merged my near identical one.
> [6by9@a2442a5](https://github.com/6by9/linux/commit/a2442a547a461168b35d742f8fcbac396185b324)

Sorry for doing the work double, but I really couldn't find it.

I will check now how to add the "Signed-off By" line. Let me know if this works or if I should change something. Thank you!

manhalt:
Ok I got the "Signed-off-by" added to all 3 commits. Unfortunatly after my rebase 2 commits that got reverted / squashed don't want to go.

6by9:
I'm glad that those who only consider themselves begineers are prepared to attempt a PR. However because we may well look at upstreaming these drivers into mainline we do have to abide by their rules on coding style etc.

PRs are a set of patches that get applied directly - the original two patches need to be complete and correct as they stand, not fixed up by a third patch. ```git rebase -i``` is generally your friend to allow you to edit any patches that are already committed.

The other useful trick is to create/edit ```.git/hooks/post-commit``` in your kernel tree to read
```
#!/bin/sh
```
so that checkpatch gets run on every patch as soon as you commit it. You soon get used to fixing up coding style issues immediately, and then it becomes second nature.

Most kernel devs will prefer to just write out setting and clearing bits rather than defining a macro - it's such a common function that it's instantly recognisable.
```
val = read_val();
if (ctrl->val)
   val |= BIT(2);
else
   val & ~BIT(2);
```
(Ideally define the bit alongside the register definition so it becomes ```val |= OV9281_FLIP_BIT;```)
The compiler will optimise the bit manipulation down to a single line as the BIT macro can be evaluated at compile time (```or r1, 0x04``` or ```and r1, 0xfffffffb````).
```BIT``` is in https://elixir.bootlin.com/linux/latest/source/include/vdso/bits.h#L7, but there are variants for longer variable types in https://elixir.bootlin.com/linux/latest/source/include/linux/bits.h#L9

When dealing with multiple bit fields within a register there are ```GENMASK```, ```FIELD_GET```, ```FIELD_SET```. That's the way the vc4_hdmi driver sets things up - https://elixir.bootlin.com/linux/latest/source/drivers/gpu/drm/vc4/vc4_regs.h#L12

I hadn't created the PR for my OV9281 flips patch. I was writing the driver for OV2311 (very similar) with flip support, and then realised I could backport it to OV9281.

The use of 0x3778 is going to be some magic. All the evidence I have seen is that it should be set in the 640x400 modes, and not in the higher modes. I'm guessing it's switching between binning summation and averaging.
As it is a variable that changes between modes, it'd be nice to see it listed explicitly in each mode, not set to 0x00 in the common table and then changed in the 640x480 mode.

**Comment by 6by9** [comment]
The one commit that remains on this branch is fine, although I thought we'd concluded that 0x3778 should be 0x10 in the 640x400 mode. It can be fixed up later.

Do you want me to merge this and then you create a new PR for flips? Or I'll merge my patch for flips. I don't really mind which.

**Comment by manhalt** [comment]
@6by9 Sorry I was not done yet. I had some rebase struggle to fix.

Thank you for your previous input. I adapted the BIT() macro and I think it is good to go, please have a look at the changed commit. You can merge it if all is good.

I also wanted to ask for your opinion on something else, not directly connected to these changes, but also this driver:

I found evidence that the native pixel array is 16 pixel more in each direction.  (https://www.ovt.com/download/sensorpdf/206/OmniVision_OV9281.pdf, page 2: 1296x816). I saw that this can be easily changed in the defines and I tested that, but I cannot notice any difference.
I first thought about this when I calculated the blanking times, and the real FPS would always a bit higher then calculated (+1 or 2). It still does not add up but gets closer. Do you have any idea how I can check the transmitted pixels easily? Also, I wonder if more pixels would get transferred, should it be possible to notice border pixels? I looked for white or black pixels at the border but could not find them. 

Another thing I thought of is custom controls for syncing cams or other specialised features for this driver. I know it is not the best to add more custom controls, but I do not see any other way to get these features accessible. Would you agree adding them when really necessary?

**Comment by 6by9** [comment]
Yes that datasheet lists the active array as 1296 x 816, but the readout modes give 1280x800. Often sensors have optically black and/or dummy pixels. eg imx219 lists
```
Total number of pixels : 3296(H) × 2512(V) approx. 8.28M pixels
Number of effective pixels : 3296(H) × 2480(V) approx. 8.17M pixels
Number of active pixels : 3280(H) × 2464(V) approx. 8.08M pixels
```
If you then look at Fig 27 on page 52 of https://raw.githubusercontent.com/rellimmot/Sony-IMX219-Raspberry-Pi-V2-CMOS/master/RASPBERRY%20PI%20CAMERA%20V2%20DATASHEET%20IMX219PQH5_7.0.0_Datasheet_XXX.PDF it shows where those extra pixels are.

The regions reported through g_selection are a little theoretical. There are some sensors where you can read out the optically dark pixels, but very few drivers actually bother.
If you use ```v4l2-ctl --log-status``` whilst streaming, then the Unicam driver should log the detected image size.

Custom controls are a real pain. I did intend to raise the question of syncing sensors on the linux-media mailing list as it applies to imx477, ov9281, and a number of others, but haven't done so as yet.

**Comment by manhalt** [comment]
Thanks for the explanation. I guess without the OV9281 datasheet it might be hard to figure out how these extra pixels are located. I tried to find out if they get transferred by looking at the output of the status as you suggested, but the numbers I found there don’t make much sense. The detected width seems to be wrong always:

` pi@pi:~ $ v4l2-ctl --stream-mmap --stream-count=-1 -d /dev/video0 --stream-to=/dev/null --set-fmt-video=width=1280,height=800,pixelformat=Y10` shows this:

   [  622.542127] unicam fe801000.csi: =================  START STATUS  =================
   [  622.542146] unicam fe801000.csi: -----Receiver status-----
   [  622.542164] unicam fe801000.csi: V4L2 width/height:   1280x800
   [  622.542181] unicam fe801000.csi: Mediabus format:     0000200a
   [  622.542197] unicam fe801000.csi: V4L2 format:         50303159
   [  622.542213] unicam fe801000.csi: Unpacking/packing:   0 / 0
   [  622.542228] unicam fe801000.csi: ----Live data----
   [  622.542243] unicam fe801000.csi: Programmed stride:   1600
   [  622.542259] unicam fe801000.csi: Detected resolution: 800x800
   [  622.542275] unicam fe801000.csi: Write pointer:       f082f3a0
   [  622.542289] unicam fe801000.csi: ==================  END STATUS  ==================


`pi@pi:~ $ v4l2-ctl --stream-mmap --stream-count=-1 -d /dev/video0 --stream-to=/dev/null --set-fmt-video=width=1280,height=800,pixelformat=GREY` shows this:

   [  682.305282] unicam fe801000.csi: =================  START STATUS  =================
   [  682.305302] unicam fe801000.csi: -----Receiver status-----
   [  682.305320] unicam fe801000.csi: V4L2 width/height:   1280x800
   [  682.305336] unicam fe801000.csi: Mediabus format:     00002001
   [  682.305352] unicam fe801000.csi: V4L2 format:         59455247
   [  682.305368] unicam fe801000.csi: Unpacking/packing:   0 / 0
   [  682.305383] unicam fe801000.csi: ----Live data----
   [  682.305398] unicam fe801000.csi: Programmed stride:   1280
   [  682.305414] unicam fe801000.csi: Detected resolution: 640x800
   [  682.305430] unicam fe801000.csi: Write pointer:       f03baa70
   [  682.305445] unicam fe801000.csi: ==================  END STATUS  ==================

Do you have any idea why?

**Comment by 6by9** [comment]
I'm relying on the hardware doing what it's documented to do. Docs say CAMIHSTA bits 15:0 are
>Pixels Per Line
> Indicates the total number of pixels detectedwithin a line 

And similar for CAMIVSTA (only updated on FE)

Something looks a wrong though as Y10 should be unpacking to 16bpp (only bottom 10bits used), but that means the stride is too low.
The ```V4L2 format: 50303159``` implies it's switched to Y10P, in which case the stride is fine, but you're not getting the image you asked for. Is it doing the streaming before the set command?

**Comment by manhalt** [comment]
Sorry, I am not sure if I understood your comment correctly: do you mean that in GREY it shows half of the size because it assumes 16bit and it only has 8bit? 

About the Y10 / Y10P issue: you are right, it shows the wrong format whatever I request. But as I found out, this is related to v4l2-ctl – it will never request Y10 and it also shows that it only has Y10P available. If I request it from another application, it works and it will also show the correct output:

[ 1055.815898] unicam fe801000.csi: =================  START STATUS  =================
[ 1055.815909] unicam fe801000.csi: -----Receiver status-----
[ 1055.815914] unicam fe801000.csi: V4L2 width/height:   1280x800
[ 1055.815918] unicam fe801000.csi: Mediabus format:     0000200a
[ 1055.815922] unicam fe801000.csi: V4L2 format:         20303159
[ 1055.815926] unicam fe801000.csi: Unpacking/packing:   4 / 5
[ 1055.815929] unicam fe801000.csi: ----Live data----
[ 1055.815933] unicam fe801000.csi: Programmed stride:   2560
[ 1055.815937] unicam fe801000.csi: Detected resolution: 1280x800
[ 1055.815940] unicam fe801000.csi: Write pointer:       dff4a590
[ 1055.815943] unicam fe801000.csi: ==================  END STATUS  ==================

**Comment by 6by9** [comment]
Ah, it'll be the space at the end of "Y10 " that is making life tricky for v4l2-ctl. You can select the pixel format by index instead.
```
pi@raspberrypi:~ $ v4l2-ctl --list-formats
ioctl: VIDIOC_ENUM_FMT
	Type: Video Capture

	[0]: 'Y10P' (10-bit Greyscale (MIPI Packed))
	[1]: 'Y10 ' (10-bit Greyscale)
	[2]: 'GREY' (8-bit Greyscale)
```
so
```
v4l2-ctl -v pixelformat=1
```
will select Y10, as will adding quotes and including the space.
```
v4l2-ctl -v pixelformat="Y10 "
```

The detected size was only a debug option. It's possible that it gives erroneous information under some situations. Generally I've seen it give sensible numbers though.

**Comment by 6by9** [comment]
Sorry, hadn't reviewed after your final push.
I hadn't checked the commit texts. Generally the first line of kernel commits will identify the subsystem and driver, eg
```
media/i2c/ov9281: Add h & v flip controls
```
but I'm not going to overly fret here.

LGTM now.
@pelwell Any review comment from you, or are you happy to merge?

PR #4095, opened 2021-01-25
bcm2835-isp: Allow formats with different colour spaces.

davidplowman:
Each supported format now includes a mask showing the allowed colour
spaces, as well as a default colour space for when one was not
specified.

Additionally we translate the colour space to mmal format and pass it
over to the VideoCore.

### DIFF
--- a/drivers/staging/vc04_services/bcm2835-isp/bcm2835-isp-fmts.h
+++ b/drivers/staging/vc04_services/bcm2835-isp/bcm2835-isp-fmts.h
@@ -20,9 +20,28 @@ struct bcm2835_isp_fmt {
 	int bytesperline_align;
 	u32 mmal_fmt;
 	int size_multiplier_x2;
-	enum v4l2_colorspace colorspace;
+	u32 colorspace_mask;
+	enum v4l2_colorspace colorspace_default;
 	unsigned int step_size;
 };
+
+#define V4L2_COLORSPACE_MASK(colorspace) BIT(colorspace)
+
+#define V4L2_COLORSPACE_MASK_JPEG V4L2_COLORSPACE_MASK(V4L2_COLORSPACE_JPEG)
+#define V4L2_COLORSPACE_MASK_SMPTE170M V4L2_COLORSPACE_MASK(V4L2_COLORSPACE_SMPTE170M)
+#define V4L2_COLORSPACE_MASK_REC709 V4L2_COLORSPACE_MASK(V4L2_COLORSPACE_REC709)
+#define V4L2_COLORSPACE_MASK_SRGB V4L2_COLORSPACE_MASK(V4L2_COLORSPACE_SRGB)
+#define V4L2_COLORSPACE_MASK_RAW V4L2_COLORSPACE_MASK(V4L2_COLORSPACE_RAW)
+
+/*
+ * The colour spaces we support for YUV outputs. SRGB features here because,
+ * once you assign the default transfer func and so on, it and JPEG effectively
+ * mean the same.
+ */
+#define V4L2_COLORSPACE_MASK_YUV (V4L2_COLORSPACE_MASK_JPEG | \
+				  V4L2_COLORSPACE_MASK_SRGB | \
+				  V4L2_COLORSPACE_MASK_SMPTE170M | \
+				  V4L2_COLORSPACE_MASK_REC709)
 
 static const struct bcm2835_isp_fmt supported_formats[] = {
 	{
@@ -32,7 +51,8 @@ static const struct bcm2835_isp_fmt supported_formats[] = {
 		.bytesperline_align = 32,
 		.mmal_fmt	    = MMAL_ENCODING_I420,
 		.size_multiplier_x2 = 3,
-		.colorspace	    = V4L2_COLORSPACE_SMPTE170M,
+		.colorspace_mask    = V4L2_COLORSPACE_MASK_YUV,
+		.colorspace_default = V4L2_COLORSPACE_JPEG,
 		.step_size	    = 2,
 	}, {
 		.fourcc		    = V4L2_PIX_FMT_YVU420,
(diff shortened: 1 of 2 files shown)


davidplowman:
@naushir @6by9 
This is the same change as I had previously on the 5.4 branch, with only minor amendments for 5.10.

One thing I wonder about - I changed the default colour space in the supported_formats lists to "V4L2_COLORSPACE_JPEG" for only one of the YUV formats. Maybe I should do it for all of them to keep them the same?

drivers/staging/vc04_services/bcm2835-isp/bcm2835-isp-fmts.h
```diff
@@ -20,10 +20,34 @@ struct bcm2835_isp_fmt {
...
+	enum v4l2_colorspace colorspace_default;
 	unsigned int step_size;
 };
 
+struct bcm2835_isp_fmt_list {
+	struct bcm2835_isp_fmt const **list;
+	unsigned int num_entries;
+};
+
+#define V4L2_COLORSPACE_MASK(colorspace) (1<<(colorspace))
```
6by9: Use the ```BIT(x)``` macro.

drivers/staging/vc04_services/bcm2835-isp/bcm2835-v4l2-isp.c
```diff
@@ -227,14 +228,38 @@ struct bcm2835_isp_fmt *find_format_by_fourcc(unsigned int fourcc,
...
+	unsigned int i;
+
+	for (i = 0; i < node->num_supported_fmts; i++) {
+		fmt = node->supported_fmts[i];
+		if (fmt->fourcc == fourcc) {
+			if (cs == V4L2_COLORSPACE_DEFAULT ||
+			    (fmt->colorspace_mask & V4L2_COLORSPACE_MASK(cs)))
+				return fmt;
+			else
+				return NULL;
```
6by9: Assuming I'm understanding this correctly, if you call try_fmt with ```V4L2_PIX_FMT_Y10P``` but colour space ```V4L2_COLORSPACE_MASK_SMPTE170M``` you get back the default format instead, which is most likely going to be ```V4L2_PIX_FMT_YUV420```. That seems wrong.

```find_format``` can return the correct ```struct bcm2835_isp_fmt```, but then try_fmt ought to validate that the colourspace is valid for this format, and correct it if not (I think you've already got that bit in place, but this is going to mess it up).

Comment (davidplowman) [reply drivers/staging/vc04_services/bcm2835-isp/bcm2835-v4l2-isp.c]: Yes, that makes more sense. I'll fix that up, rebase everything, and re-push!

File: drivers/staging/vc04_services/bcm2835-isp/bcm2835-isp-fmts.h [hunk 3]
```diff
@@ -20,10 +20,34 @@ struct bcm2835_isp_fmt {
...
-	enum v4l2_colorspace colorspace;
+	u32 colorspace_mask;
+	enum v4l2_colorspace colorspace_default;
 	unsigned int step_size;
 };
 
+struct bcm2835_isp_fmt_list {
+	struct bcm2835_isp_fmt const **list;
+	unsigned int num_entries;
+};
```
6by9: This appears to be unused. Left over from a previous version?

drivers/staging/vc04_services/bcm2835-isp/bcm2835-v4l2-isp.c
```diff
@@ -311,10 +312,48 @@ static void mmal_buffer_cb(struct vchiq_mmal_instance *instance,
...
 }
 
+struct colorspace_translation {
+	enum v4l2_colorspace v4l2_value;
+	u32 mmal_value;
+};
+
+static u32 translate_color_space(enum v4l2_colorspace color_space)
+{
+	static const struct colorspace_translation translations[] = {
```
6by9, line 322: It probably doesn't matter, but I don't believe the static is going to do anything useful here.

drivers/staging/vc04_services/bcm2835-isp/bcm2835-v4l2-isp.c
```diff
@@ -324,6 +363,15 @@ static void setup_mmal_port_format(struct bcm2835_isp_node *node,
 	port->es.video.crop.height = q_data->height;
 	port->es.video.crop.x = 0;
 	port->es.video.crop.y = 0;
+
+	mmal_colour_space = translate_color_space(q_data->colorspace);
```
6by9: Quicker and easier to just return ```MMAL_COLOR_SPACE_UNKNOWN``` from ```translate_color_space```, and then that can be assigned direct to ```port->es.video.color_space```. Does the error message really add anything? Shouldn't try_fmt have filtered that out already? Or put the error message in ```translate_color_space```

6by9 (commented):

Does q_data->colorspace get initialised to something sensible, so opening the device and G_FMT gives you the correct result? (```v4l2-ctl -V -d 14``` should tell you that).

davidplowman:
Yes, I'd rather forgotten about this too. But indeed, just after start-up, `v4l2-ctl -V -d 14` reports:
```
Format Video Capture:
	Width/Height      : 1280/1280
	Pixel Format      : 'YUYV' (YUYV 4:2:2)
	Field             : None
	Bytes per Line    : 2560
	Size Image        : 3276800
	Colorspace        : Default
	Transfer Function : Default (maps to Rec. 709)
	YCbCr/HSV Encoding: Default (maps to ITU-R 601)
	Quantization      : Default (maps to Limited Range)
	Flags             : 
```
The colour space is just zero initialised, thus "default". Would a value of SMPTE170M make more sense, that being the default colour space for the format?

6by9:
> Would a value of SMPTE170M make more sense, that being the default colour space for the format?

```get_default_format``` returns the default ```struct bcm2835_isp_fmt```, so take the ```colorspace_default``` from that in ```register_node```

Is YUYV really our default? That's a bit ugly seeing as very few other blocks support it! Oh well.

davidplowman:
I think this latest version should tidy up most of that - anything else we need to do?

6by9:
I'll go with that for now. Thanks.

PR #4104, opened 2021-01-29
drm/vc4: hvs: Fix buffer overflow with the dlist handling

mripard:
Commit 0a038c1c29a7 ("drm/vc4: Move LBM creation out of
vc4_plane_mode_set()") changed the LBM allocation logic from first
allocating the LBM memory for the plane to running mode_set,
adding a gap in the LBM, and then running the dlist allocation filling
that gap.

The gap was introduced by incrementing the dlist array index, but was
never checking whether or not we were over the array length, leading
eventually to memory corruptions if we ever crossed this limit.

vc4_dlist_write had that logic though, and was reallocating a larger
dlist array when reaching the end of the buffer. Let's share the logic
between both functions.

Cc: Boris Brezillon <boris.brezillon@collabora.com>
Cc: Eric Anholt <eric@anholt.net>
Fixes: 0a038c1c29a7 ("drm/vc4: Move LBM creation out of vc4_plane_mode_set()")

### DIFF
--- a/drivers/gpu/drm/vc4/vc4_plane.c
+++ b/drivers/gpu/drm/vc4/vc4_plane.c
@@ -227,7 +227,7 @@ static void vc4_plane_reset(struct drm_plane *plane)
 	__drm_atomic_helper_plane_reset(plane, &vc4_state->base);
 }
 
-static void vc4_dlist_write(struct vc4_plane_state *vc4_state, u32 val)
+static void vc4_dlist_counter_increment(struct vc4_plane_state *vc4_state)
 {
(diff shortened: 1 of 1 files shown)


6by9:
@pelwell PR was to rpi-5.11.y. Can we cherry-pick it back to rpi-5.10.y as well please?

pelwell:
I was spending a moment looking at your DMA problems.

mripard:
That bug was introduced in 5.0, so if there's other maintained branches (5.4?) I guess we should backport it there too?

pelwell:
I'd really like 5.4 to be unmaintained, but we're not there yet. It's done.

PR #4109, opened 2021-02-01
configs: Enable CONFIG_MEDIA_CEC_RC

popcornmix:
See: https://www.raspberrypi.org/forums/viewtopic.php?f=44&t=301072


popcornmix:
Note: these options were already enabled for arm64.

popcornmix:
@PaulWebster has tested this.

6by9:
LGTM. 
There is the question as to how useful DRM/KMS is on Pi0/1, and even more so using the GL drivers, but there's nothing fundamental against providing the option of loading them.

popcornmix:
We've always had DRM/KMS enabled (as module) in Pi0/1, so this just enables a trivial amount of cec code.
Does anyone use it on Pi0/1? Not for desktop use I suspect but it could be used for standalone apps.

We did build LibreELEC using kms for Pi1 and the GUI wasn't unusable but it just ends up using too much RAM before long.
Something simpler should be possible. But it's an interesting question whether anyone is actually using kms on Pi0/1.

pelwell:
`MEDIA_CEC_RC` was not set in the 64-bit defconfigs and is not set after this PR is merged.

popcornmix:
My mistake. Added to arm64 now.

PR #4113, opened 2021-02-02
bcm2835-codec tweaks

6by9:


6by9:
@jc-kynesim I did go for the framework tweak instead of totally reworking the buffer feed code.

@popcornmix Hopefully pixel aspect ratio reporting.

drivers/media/v4l2-core/v4l2-mem2mem.c
```diff
@@ -301,8 +301,9 @@ static void __v4l2_m2m_try_queue(struct v4l2_m2m_dev *m2m_dev,
 
 	dprintk("Trying to schedule a job for m2m_ctx: %p\n", m2m_ctx);
 
-	if (!m2m_ctx->out_q_ctx.q.streaming
-	    || !m2m_ctx->cap_q_ctx.q.streaming) {
+	if (!m2m_ctx->out_q_ctx.buffered &&
+	    (!m2m_ctx->out_q_ctx.q.streaming ||
+	     !m2m_ctx->cap_q_ctx.q.streaming)) {
```
pelwell: I may have misunderstood the intention of this commit, but it appears to bypass the streaming check if running in buffered mode.

drivers/staging/vc04_services/bcm2835-codec/bcm2835-v4l2-codec.c
```diff
@@ -1331,6 +1331,7 @@ static int vidioc_s_fmt(struct bcm2835_codec_ctx *ctx, struct v4l2_format *f,
 	struct vb2_queue *vq;
```
pelwell: Typo in the commit message - `aat`.

drivers/staging/vc04_services/bcm2835-codec/bcm2835-v4l2-codec.c
```diff
@@ -2464,22 +2545,25 @@ static int bcm2835_codec_start_streaming(struct vb2_queue *q,
...
 		q_data->eos_buffer_in_use = false;
 
-		port->cb_ctx = ctx;
 		ret = vchiq_mmal_port_enable(dev->instance,
 					     port,
 					     ip_buffer_cb);
 		if (ret)
 			v4l2_err(&ctx->dev->v4l2_dev, "%s: Failed enabling i/p port, ret %d\n",
 				 __func__, ret);
+
```
pelwell: Is this accidental?

drivers/staging/vc04_services/bcm2835-codec/bcm2835-v4l2-codec.c
```diff
@@ -2464,22 +2545,25 @@ static int bcm2835_codec_start_streaming(struct vb2_queue *q,
...
+		if (!port->enabled) {
+			ret = vchiq_mmal_port_enable(dev->instance,
+						     port,
+						     op_buffer_cb);
+			if (ret)
+				v4l2_err(&ctx->dev->v4l2_dev, "%s: Failed enabling o/p port, ret %d\n",
+					 __func__, ret);
+		}
 	}
+	v4l2_err(&ctx->dev->v4l2_dev, "%s: Done, ret %d\n",
```
pelwell: This looks like an unconditional error message.

drivers/staging/vc04_services/bcm2835-codec/bcm2835-v4l2-codec.c
```diff
@@ -1540,6 +1544,14 @@ static int vidioc_g_selection(struct file *file, void *priv,
 			s->r.width = q_data->crop_width;
 			s->r.height = q_data->crop_height;
 			break;
+		case V4L2_SEL_TGT_CROP_BOUNDS:
```
pelwell, line 1525: This change seems unrelated.

pelwell (commented):
You lost me in the second commit, but here's some feedback.

drivers/staging/vc04_services/bcm2835-codec/bcm2835-v4l2-codec.c
6by9: Yes. Fixed.

drivers/staging/vc04_services/bcm2835-codec/bcm2835-v4l2-codec.c
6by9: Corrected.

drivers/staging/vc04_services/bcm2835-codec/bcm2835-v4l2-codec.c
6by9: Sorted.

drivers/staging/vc04_services/bcm2835-codec/bcm2835-v4l2-codec.c
6by9: Yes and no. I'll make it a second patch.

The only route for userspace to query the pixel aspect ratio is via ```VIDIOC_CROPCAP```, which goes through [v4l2_cropcap](https://elixir.bootlin.com/linux/latest/source/drivers/media/v4l2-core/v4l2-ioctl.c#L2451).
That calls our vidioc_g_pixelaspect function, but then follows it up with calls to g_selection for other properties, including these.
Where and how they are relevant for a video decoder is another questsion, but the call fails otherwise.

Oddly the only other decoder I can find which implements vidioc_g_pixelaspect is staging/media/meson/vdec, but it doesn't implement g_selection at all, so I don't believe there is any driver in mainline that actually implements this, but FFmpeg tries to read it.

drivers/media/v4l2-core/v4l2-mem2mem.c
6by9: You're right, but it wasn't quite my intent. Too much inverted logic.

I wanted the existing logic when not buffering, but to not return should either queue be streaming when buffered. One to rework.

6by9:
Updated to hopefully address your comments.

popcornmix:
Kodi doesn't see the aspect ratio, but it might be getting lost in ffmpeg or kodi itself. I'll see if I can find it...

6by9:
Does Kodi / FFmpeg actually make use of the resolution changed event? If it has configured the output correctly then there is no event and we have no chance to report the PAR.

drivers/media/v4l2-core/v4l2-mem2mem.c
```diff
@@ -301,9 +301,10 @@ static void __v4l2_m2m_try_queue(struct v4l2_m2m_dev *m2m_dev,
 
 	dprintk("Trying to schedule a job for m2m_ctx: %p\n", m2m_ctx);
 
-	if (!m2m_ctx->out_q_ctx.q.streaming
-	    || !m2m_ctx->cap_q_ctx.q.streaming) {
-		dprintk("Streaming needs to be on for both queues\n");
+	if ((!m2m_ctx->out_q_ctx.q.streaming ||
+	     !m2m_ctx->cap_q_ctx.q.streaming) &&
+	    !(m2m_ctx->out_q_ctx.buffered && m2m_ctx->out_q_ctx.q.streaming)) {
```
pelwell, line 306: I just spent a minute or two writing out your condition as you described it:
```
	if (!((m2m_ctx->out_q_ctx.q.streaming &&
	       m2m_ctx->cap_q_ctx.q.streaming) ||
	      (m2m_ctx->out_q_ctx.buffered && m2m_ctx->out_q_ctx.q.streaming))) {
```
and then using De Morgan's laws to transform it into what you have written. But I shouldn't have to - what's wrong with the literal version above?

bitbashing: The OP's version is just an evolution of pre-existing (and thus assumed stable to some degree through testing/deployment) code using a minimal set of changes.

The check could also be written as

```
	if (!(m2m_ctx->out_q_ctx.q.streaming && m2m_ctx-cap_q_ctx.q.streaming) && 
	       !(m2m_ctx->out_q_ctx.buffered && m2m_ctx->out_q_ctx.q.streaming)) {
```

They are all logically equivalent but having OR inside the clause and top-level NOT in the proposed transformed version depending on the conditions could make it a slower performing check if no short-circuiting happens inside and the result is not final until after the negation is performed at the end.

The changes suggested in c54a951 still employ a short-circuit, so they look ok to me.

pelwell: They aren't wrong, but in (especially) kernel work clarity trumps pretty much everything, especially when the compiler will likely generate the same code in both cases.

bitbashing: I agree on clarity, thus my suggestion above. Of course, as more conditions get added, the clause may have to be rewritten again.

6by9: I was just retaining the format of the original term. I'll rework as requested.

6by9:
I've dropped the patch for enabling the decoder on STREAMON(OUTPUT) for now, but done the v4l2_m2m rework that was requested.
This has also gained a patch to enable VC1. It needs a firmware change to actually work reliably, but is harmless to merge this side first.

popcornmix:
I'm happy with this. Aspect ratio does work (needs an ffmpeg patch).
And VC1 is working.

popcornmix:
As mentioned on slack the VC1 patch doesn't support WMV9 (and the hardware/omplayer do).
So might be worth checking that before merging.

6by9:
> As mentioned on slack the VC1 patch doesn't support WMV9 (and the hardware/omplayer do).
So might be worth checking that before merging.

Not going to happen in the short term. The codec is getting upset at quite a low level with that.
It needs a fair amount of analysis and bitstream parsing so that RIL can strip off the headers and then still pass the start of the stream into the codec for decoding. WMV is all special cases in RIL for that :-(

popcornmix:
Okay, I'm happy with PR then.

PR #4119, opened 2021-02-04
media: i2c: imx290: Support V4L2_CID_ANALOGUE_GAIN as well as V4L2_CI…

davidplowman:
…D_GAIN

V4L2_CID_GAIN is already controlling the analogue gain, so we just
need the extra control.

### DIFF
--- a/drivers/media/i2c/imx290.c
+++ b/drivers/media/i2c/imx290.c
@@ -1328,7 +1328,7 @@ static int imx290_probe(struct i2c_client *client)
 	 */
 	imx290_entity_init_cfg(&imx290->sd, NULL);
 
-	v4l2_ctrl_handler_init(&imx290->ctrls, 4);
+	v4l2_ctrl_handler_init(&imx290->ctrls, 9);
 
 	v4l2_ctrl_new_std(&imx290->ctrls, &imx290_ctrl_ops,
 			  V4L2_CID_ANALOGUE_GAIN, 0, 100, 1, 0);
--- a/drivers/media/i2c/imx290.c
+++ b/drivers/media/i2c/imx290.c
@@ -647,7 +647,7 @@ static int imx290_set_ctrl(struct v4l2_ctrl *ctrl)
 		return 0;
 
 	switch (ctrl->id) {
-	case V4L2_CID_GAIN:
+	case V4L2_CID_ANALOGUE_GAIN:
 		ret = imx290_set_gain(imx290, ctrl->val);
 		break;
 	case V4L2_CID_EXPOSURE:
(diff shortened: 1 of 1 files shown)


davidplowman:
@naushir @6by9
This change is enough (in this driver, at least) to make the imx290 work with libcamera. I assume imx290_set_gain is twiddling the analogue gain already, so V4L2_CID_ANALOGUE_GAIN just needs to do the same thing?

6by9:
You can't do that. You now have two controls registered within the framework that program the sensor with the same thing, but don't update each other.

I know this is an upstream driver with some downstream tweaks, but replace their use of V4L2_CID_GAIN with V4L2_CID_ANALOGUE_GAIN.

I can see why they've used V4L2_CID_GAIN as the one register allows up to 30dB of analogue gain, and then an additional 42dB of digital gain, so it is a combined control.

Do you have a datasheet for this one? I can forward it to you if not.

davidplowman:
No, I don't have the datasheet. Are you suggesting I should delete V4L2_CID_GAIN completely, I wasn't quite sure what you meant?

I guess the other fix would be to make our libcamera pipeline handler look for V4L2_CID_GAIN if V4L2_CID_ANALOGUE_GAIN doesn't exist, I'm not sure what Naush would make of that...

naushir:
> No, I don't have the datasheet. Are you suggesting I should delete V4L2_CID_GAIN completely, I wasn't quite sure what you meant?
> 
> I guess the other fix would be to make our libcamera pipeline handler look for V4L2_CID_GAIN if V4L2_CID_ANALOGUE_GAIN doesn't exist, I'm not sure what Naush would make of that...

V4L2_CID_ANALOGUE_GAIN is now a mandatory control in libcamera. In the latest builds, libcamera would fail if the sensor doesn't support this control.

6by9:
Yes, delete ```V4L2_CID_GAIN```
I guess they've gone from the text for _GAIN in https://www.kernel.org/doc/html/latest/userspace-api/media/v4l/control.html
>Devices that recognise the difference between digital and analogue gain use controls V4L2_CID_DIGITAL_GAIN and V4L2_CID_ANALOGUE_GAIN.

As this doesn't, they didn't.

6by9:
And as Naush says, without ANALOGUE_GAIN it can't work with libcamera and is therefore largely useless.

I guess we could limit the gain control to 30dB of analogue gain (0-90) to make it actually follow the letter of the spec. Digital gain can be added in the ISP.

6by9:
Commit text doesn't explain why the change of gain range to only expose the analogue portion of the gain

And the change to the number of controls ought to be in a separate commit, really with a Fixes: tag.

davidplowman:
I've split that into two now with some "Fixes:" tags. I note that the number of controls there is only a "hint", so it was never strictly broken - do you think the "Fixes:" tags are still warranted?

6by9:
Seeing as all the patches are downstream ones (by me - oops) it's not a big deal.
At least the hashes should now be consistent as we've stopped rebasing the branch.

6by9:
And I'm happy with the patches overall now.
I'll let Phil give it a once over though.

PR #4121, opened 2021-02-05
drm/vc4 Enable HDR infoframe data

popcornmix:
This is reported to make HDR symbol appear TV when playing HDR content in Kodi.
Colours are reported to be better in The Hobbit.

Still needs more testing and I don't have a HDR display, so I'm relying on feedback from others.

### DIFF
--- a/drivers/gpu/drm/vc4/vc4_hdmi.c
+++ b/drivers/gpu/drm/vc4/vc4_hdmi.c
@@ -527,6 +527,45 @@ static int vc4_hdmi_connector_get_modes(struct drm_connector *connector)
 	return ret;
 }
 
+static bool hdr_metadata_equal(const struct drm_connector_state *old_state,
+			       const struct drm_connector_state *new_state)
+{
+	struct drm_property_blob *old_blob = old_state->hdr_output_metadata;
+	struct drm_property_blob *new_blob = new_state->hdr_output_metadata;
+
+	if (!old_blob || !new_blob)
+		return old_blob == new_blob;
+
+	if (old_blob->length != new_blob->length)
+		return false;
+
+	return !memcmp(old_blob->data, new_blob->data, old_blob->length);
+}
+
+static int vc4_hdmi_connector_atomic_check(struct drm_connector *connector,
+					  struct drm_atomic_state *state)
+{
+	struct drm_connector_state *old_state =
+		drm_atomic_get_old_connector_state(state, connector);
+	struct drm_connector_state *new_state =
+		drm_atomic_get_new_connector_state(state, connector);
+	struct drm_crtc *crtc = new_state->crtc;
+	struct drm_crtc_state *crtc_state;
+
+	if (!crtc)
+		return 0;
+
+	if (!hdr_metadata_equal(old_state, new_state)) {
+		crtc_state = drm_atomic_get_crtc_state(state, crtc);
+		if (IS_ERR(crtc_state))
+			return PTR_ERR(crtc_state);
+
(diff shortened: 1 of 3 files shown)


include/linux/hdmi.h
```diff
@@ -156,7 +156,7 @@ enum hdmi_content_type {
 };
 
 enum hdmi_metadata_type {
-	HDMI_STATIC_METADATA_TYPE1 = 1,
+	HDMI_STATIC_METADATA_TYPE1 = 0,
```
6by9, line 159: Yes this looks like it is confusion between defining the bit masking and how to define it in the headers as an absolute value or bit position.

6 of one, half a dozen of the other between changing the header vs amending ```hdr_metadata_type```
```
static uint8_t hdr_metadata_type(const u8 *edid_ext)
{
	return edid_ext[3] &
		BIT(HDMI_STATIC_METADATA_TYPE1);
}
```
Has this been passed upstream at all?

drivers/gpu/drm/vc4/vc4_hdmi.c
```diff
@@ -527,6 +527,45 @@ static int vc4_hdmi_connector_get_modes(struct drm_connector *connector)
...
+		crtc_state = drm_atomic_get_crtc_state(state, crtc);
+		if (IS_ERR(crtc_state))
+			return PTR_ERR(crtc_state);
+
+		crtc_state->mode_changed = true;
+	}
+
+	return 0;
+}
+
```
6by9, line 568: Imitation is the most sincere form of flattery :-)
https://elixir.bootlin.com/linux/latest/source/drivers/gpu/drm/bridge/synopsys/dw-hdmi.c#L2398

6by9 (commented):
Looks reasonable. I'll try to give it a whirl next week against the analyser and LG TV.

include/linux/hdmi.h
popcornmix: Yes see https://lists.freedesktop.org/archives/dri-devel/2021-January/295187.html and https://github.com/xbmc/xbmc/issues/19122

> Oh, my bad, I thought this was in drm_mode.h, but it's not. Then yeah should be completely fine to fix it.

6by9: That thread's familiar now - I must have read it in passsing.
Still not merged (https://patchwork.freedesktop.org/patch/415578/?series=86212&rev=1), but it looks like it's largely been accepted as required.

pelwell:
That looks like fine plumbing to me.

HiassofT:
Testing with this PR in LibreELEC looks good so far.

I've tested with several HDR samples on my LG 55C8 and feedback from our forum users looks fine, too https://forum.libreelec.tv/thread/23442-rpi4-testbuild-with-hdr-support/

popcornmix:
@mripard any comments on this?

mripard:
I don't, it looks great :)

llamafilm:
This is great.  Are there any userland tools that can be used to toggle HDR mode on/off?  So far the only one I found is LibreElec.  I was hoping to use VLC or `tvservice`.

popcornmix:
It doesn't really make sense for tvservice. tvservice is just a thin wrapper to firmware code that sets hdmi modes.
HDR support is in the kernel.
Also the way DRM works is changes only persist when the application (the DRM master) is running, so anything tvservice did would only last until it exited.
Also enabling DRM infoframe is only useful when content is generated in that colourspace.
HDR is pretty new in the linux world, and not much supports it. Kodi is one application that does.

PR #4135, opened 2021-02-09
Hifiberry DAC+ADC Pro fix for the PLL when changing sample rates

j-schambacher:
Adds 2 msecs delay when switching between oscillators to allow
correct PLL settling.
Same fix as in the DACplus driver.

Thanks to Clive Messer for the support!


popcornmix:
5.10 kernel would be more useful

pelwell:
Already forward ported.

PR #4137, opened 2021-02-10
Another DAC..

pifi-bz:


pelwell:
LGTM. For extra credit, I think you should be able to use a `simple-audio-card` overlay and do away with the need for the driver change, but I'm not going to demand it.

pelwell:
Let me know, one way or the other.

pifi-bz:
Hi Phil -

Sorry for the delay in getting back to you - it took a bit of feeling daft and like a flounder to get the simple-audio-card overlay up and running.  I've now got two overlays for two different bits of hardware - would you like them in one go or as two separate pull requests?

Cheers --

Dave

pelwell:
All in the same PR is fine. Separate commits is nicer to review, but they'll get squashed in with the other overlays eventually.

PR #4141, opened 2021-02-12
KMS VGA666 overlay

6by9:
Adds an overlay for the VGA666 under full KMS.

If you don't have DDC connected through a level shifter then you'll need a ```video=VGA-1....``` entry in cmdline.txt, or add it via xrandr.


arch/arm/boot/dts/overlays/README
```diff
@@ -3120,6 +3120,15 @@ Params: cma-512                 CMA is 512MB
                                 outputs)
 
 
+Name:   vc4-kms-vga666
+Info:   Enable the VGA666 (resistor ladder ADC) for the vc4-kms-v3d driver.
+        Requires vc4-kms-v3d to be loaded.
+Load:   dtoverlay=vc4-kms-vga666,<param>
+Params: ddc                    Enables GPIOs 0&1 as the I2C to read the EDID
```
pelwell: The indentation on the parameter description is off-by-one, but otherwise this looks OK.

pelwell:
Although these fragments aren't listed in application order, it doesn't matter because the intra-first rule is sufficient provided you don't have fragments that target fragments that target fragments.

arch/arm/boot/dts/overlays/README
6by9: Fixed.

6by9:
Updated.
I've also added a note under vga666 (which just sets the GPIOs to alt2) that it is not for use with vc4-kms-v3d as you'll get a clash over the GPIOs.

Issue #4132, opened 2021-02-08
5.10.14 Pi4 vc4-kms-v3d results in i2c-0 and i2c-1 not probing


6by9:
Placeholder for notes rather than posting on the bottom of #3999 (the suspected cause of the issue)

If ```dtoverlay=vc4-kms-v3d``` is present in config.txt, then ```dtparam=i2c_arm=on``` fails to load i2c_bcm2835 with
```[    6.647857] i2c-bcm2835 fe804000.i2c: Could not request IRQ```

5.10.13 at https://github.com/raspberrypi/linux/commit/98a26692449ad74be15a25c05bb5be5396aaea4f works fine.

That leaves the CEC support, or possibly Phil's I2C bug fix in 431ac1d21b5e842c5a564bc0bab31d7dc11611f5 as the likely candidates.

I'm working through it, but I'm building on my laptop so it's not the fastest.

**Comment by 6by9** [comment]
93a3b097c467 ARM: dts: bcm2711: Add the BSC interrupt controller 
looks to be the commit that causes the trouble.

**Comment by pelwell** [comment]
While I'm no stranger to failure, I was testing I2C heavily during the development of https://github.com/raspberrypi/linux/commit/431ac1d21b5e842c5a564bc0bab31d7dc11611f5, so it's hard to see how something as simple as a failure to initialise could happen.

**Comment by pelwell** [comment]
Phew,

**Comment by 6by9** [comment]
It does require vc4-kms-v3d to be loaded, otherwise all is good anyway.

(As you know I'm trying to get the DSI panels working, therefore want the I2C to the touchpanel an regulator. Updated and it all went to pot!)

**Comment by 6by9** [comment]
```
&i2c0if {
	compatible = "brcm,bcm2711-i2c", "brcm,bcm2835-i2c";
	interrupts = <GIC_SPI 117 IRQ_TYPE_LEVEL_HIGH>;
};
```
```
		bsc_intr: interrupt-controller@7ef00040 {
			compatible = "brcm,bcm2711-l2-intc", "brcm,l2-intc";
			reg = <0x7ef00040 0x30>;
			interrupts = <GIC_SPI 117 IRQ_TYPE_LEVEL_HIGH>;
			interrupt-controller;
			#interrupt-cells = <1>;
		};
```
Need to check the docs that the bsc_intr interrupt is correct (it works so probably is), but they'll have to play nicely and share the interrupt.

6by9:
[...]

PR #4143, opened 2021-02-12
Revert bsc_intr addition for HDMI I2C controllers

6by9:
#4132


pelwell:
This is disappointing but acceptable to me. Any objections, @popcornmix ?

6by9:
Upstream irqchip people have said that there isn't a nice solution and it's being reverted there (although I haven't seen the patch from Florian as yet). This is more jumping ahead because it does break a number of things.
https://lists.freedesktop.org/archives/dri-devel/2021-February/297197.html

pelwell:
Taking silence for consent.

PR #4150, opened 2021-02-16
Imx477 long exposure control

naushir:


drivers/media/i2c/imx477.c
```diff
@@ -1316,7 +1352,8 @@ static int imx477_set_ctrl(struct v4l2_ctrl *ctrl)
 		break;
 	case V4L2_CID_EXPOSURE:
 		ret = imx477_write_reg(imx477, IMX477_REG_EXPOSURE,
-				       IMX477_REG_VALUE_16BIT, ctrl->val);
+				       IMX477_REG_VALUE_16BIT, ctrl->val /
+						(1 << imx477->long_exp_shift));
```
6by9: ```(ctrl->val >> imx477->long_exp_shift)```?

drivers/media/i2c/imx477.c
```diff
@@ -1285,20 +1292,49 @@ static void imx477_adjust_exposure_range(struct imx477 *imx477,
...
+	imx477->long_exp_shift = 0;
+
+	while (val > IMX477_FRAME_LENGTH_MAX) {
+		if (++imx477->long_exp_shift > IMX477_LONG_EXP_SHIFT_MAX) {
+			imx477->long_exp_shift = IMX477_LONG_EXP_SHIFT_MAX;
+			val = IMX477_FRAME_LENGTH_MAX;
+			break;
+		}
+
+		val /= 2;
```
6by9: ```val >>= 1```?
You're dealing with shifts so it feels odd mixing shifts and divides (even if they are the same thing).

drivers/media/i2c/imx477.c
```diff
@@ -1285,20 +1292,49 @@ static void imx477_adjust_exposure_range(struct imx477 *imx477,
...
 }
 
+static int imx477_set_frame_length(struct imx477 *imx477, unsigned int val)
+{
+	int ret = 0;
+
+	imx477->long_exp_shift = 0;
+
+	while (val > IMX477_FRAME_LENGTH_MAX) {
+		if (++imx477->long_exp_shift > IMX477_LONG_EXP_SHIFT_MAX) {
```
6by9: Pre/post increments/decrements in logical statements always make me have to think harder than I'd like over the actual logic being checked. Separate them into two lines?

drivers/media/i2c/imx477.c
```diff
@@ -1285,20 +1292,49 @@ static void imx477_adjust_exposure_range(struct imx477 *imx477,
...
+			imx477->long_exp_shift = IMX477_LONG_EXP_SHIFT_MAX;
```
6by9: And a second question, doesn't this mean that the range on VBLANK has been set wrongly if we exceed the limit? Or are there rounding errors?

drivers/media/i2c/imx477.c
```diff
@@ -1927,7 +1967,8 @@ static int imx477_init_controls(struct imx477 *imx477)
 	imx477->exposure = v4l2_ctrl_new_std(ctrl_hdlr, &imx477_ctrl_ops,
 					     V4L2_CID_EXPOSURE,
 					     IMX477_EXPOSURE_MIN,
-					     IMX477_EXPOSURE_MAX,
+					     IMX477_EXPOSURE_MAX *
+					       (1 << IMX477_LONG_EXP_SHIFT_MAX),
```
6by9: I haven't compiled or run this, but doesn't this leave a max exposure that can't be achieved under most conditions. Shouldn't the exposure range (and step?) be updated when a VBLANK change results in a change of the multiplier?

Ah we do adjust it via ```imx477_adjust_exposure_range```, and that is called at the end of this function. Seeing as the default VBLANK is going to have a multiplier of x1 I'd be tempted to leave this alone, even if it is then updated at the end of the function.

drivers/media/i2c/imx477.c
naushir: I purposely did it with a divide, but in hindsight, the intention is not very clear :)
Will fix this an use shifts for all the other calculations.

drivers/media/i2c/imx477.c
naushir: Yes, strictly speaking, we cannot go over SHIFT_MAX based on the VBLANK limits set.  I was just being a bit defensive with this extra check.  I'll remove it.

drivers/media/i2c/imx477.c
naushir: Yes, it will be updated correctly so no harm in leaving this as the orginal.

naushir:
@pelwell and @6by9 this is ready to go in if you are both happy.

PR #4152, opened 2021-02-16
staging: rpivid: Fix crash when CMA alloc fails

jc-kynesim:
If realloc to increase coeff size fails then attempt to re-allocate
the original size.  If that also fails then flag a fatal error to abort
all further decode.


HiassofT:
We've been successfully testing this patch the last week in LibreELEC. It fixed the hard system lockup when playing the "Samsung Wonderland Two" demo from 4kmedia.org with 384MB CMA memory - we've now bumped the default CMA in LE to 512MB.

The fix is probably even more important for RPiOS and other distributions running with the 256MB default CMA configuration as they are more likely to hit the limit with HEVC files.

pelwell:
Thanks, John & Hias.

PR #4153, opened 2021-02-18
Some fixes to make imx290/imx327 exposure and vblank updates work correctly

davidplowman:
@naushir @6by9
Just a few more small tweaks to make the driver work correctly with libcamera. Hopefully the commit messages and the changes themselves are reasonably self-explanatory. Thanks!


6by9:
I haven't checked against the datasheet, but the changes look reasonable.

naushir:
LGTM

PR #4157, opened 2021-02-19
Imx477 fixes

naushir:
Various minor fixes for the imx477 driver + the new 120fps mode to match the firmware.


naushir:
@6by9 and @davidplowman

drivers/media/i2c/imx477.c
```diff
@@ -770,130 +770,133 @@ static const struct imx477_reg mode_2028x1080_regs[] = {
 };
 
 /* 4x4 binned. 120fps */
-static const struct imx477_reg mode_1012x760_regs[] = {
-	{0x420b, 0x01},
-	{0x990c, 0x00},
-	{0x990d, 0x08},
-	{0x9956, 0x8c},
+static const struct imx477_reg mode_1332x990_regs[] = {
+	{0x420B, 0x01},
```
6by9: I thought kernel coding style was lower case for hex values, which would reduce the number of diffs a fair bit.

drivers/media/i2c/imx477.c
```diff
@@ -1235,21 +1235,9 @@ static u32 imx477_get_format_code(struct imx477 *imx477, u32 code)
...
 	/* Set default mode to max resolution */
 	imx477->mode = &supported_modes_12bit[0];
-
-	fmt->code = MEDIA_BUS_FMT_SRGGB12_1X12;
-	fmt->colorspace = V4L2_COLORSPACE_SRGB;
-	fmt->ycbcr_enc = V4L2_MAP_YCBCR_ENC_DEFAULT(fmt->colorspace);
-	fmt->quantization = V4L2_MAP_QUANTIZATION_DEFAULT(true,
-							  fmt->colorspace,
-							  fmt->ycbcr_enc);
-	fmt->xfer_func = V4L2_MAP_XFER_FUNC_DEFAULT(fmt->colorspace);
```
6by9, line 1246: Do these other fields get set to sensible values still?

drivers/media/i2c/imx477.c
naushir: Yes, I should update these to lowercase.

drivers/media/i2c/imx477.c
naushir: fmt->code is the only field that gets referenced anywhere in the code.

The colourspace gets set in `imx477_update_image_pad_fmt` and `imx477_reset_colourspace`

6by9: OK, I'm happy. The fact that the variable has been renamed from ```imx477->fmt``` to ```imx477->fmt_code``` and it still works does imply that it's not used anywhere else.

naushir:
@pelwell, this is ready to go in if you are happy with the change.

6by9:
Yup, looks reasonable to me now. Thanks Naush.

PR #4170, opened 2021-02-23
Support hevc frequency boost from rpivid

popcornmix:


popcornmix:
Requires firmware update.
The boost works, but the unboost fails to work as desired (hevc clock remains boosted), so leaving as PR as draft for now.

popcornmix:
I think this is okay now. Firmware side is merged.
I'm not switching to the clk_set_min_rate as it doesn't behave how we'd desire (reverting clock down when min request is removed).
As only one driver will adjust hevc clock then explicitly setting it is okay.

pelwell:
The changes look fine, and I've tested that Pi 4 still boots with it. I haven't played any H265 videos, but am happy to merge on that basis.

popcornmix:
The trivial test works (using powersave governor). 
```
pi@pi4d:~ $ vcgencmd measure_clock hevc
frequency(76)=249631344
# start video playing
pi@pi4d:~ $ vcgencmd measure_clock hevc
frequency(76)=550006336
# stop video playing
pi@pi4d:~ $ vcgencmd measure_clock hevc
frequency(76)=249631344
```

PR #4173, opened 2021-02-24
Hifiberry-DAC+:Avoids loading of headphone controls if not defined in…

j-schambacher:
… DT-overlay

This fix keeps the driver from loading the headphone controls in case they are
not defined in the selected device tree overlay. This avoids problems if a
device uses the same I2C bus address as the headphone amp. The driver issues a
warning in that case.

Adds the missing embedded newlines in some printk/info statements.

### DIFF
--- a/sound/soc/bcm/hifiberry_dacplus.c
+++ b/sound/soc/bcm/hifiberry_dacplus.c
@@ -406,29 +406,34 @@ static int snd_rpi_hifiberry_dacplus_probe(struct platform_device *pdev)
 	if (ret < 0)
 		return ret;
 	if (ret) {
-		card->aux_dev = hifiberry_dacplus_aux_devs;
-		card->num_aux_devs =
-				ARRAY_SIZE(hifiberry_dacplus_aux_devs);
 		tpa_node = of_find_compatible_node(NULL, NULL, "ti,tpa6130a2");
 		tpa_prop = of_find_property(tpa_node, "status", &len);
-
-		if (strcmp((char *)tpa_prop->value, "okay")) {
-			/* and activate headphone using change_sets */
-			dev_info(&pdev->dev, "activating headphone amplifier");
-			of_changeset_init(&ocs);
-			ret = of_changeset_update_property(&ocs, tpa_node,
-							&tpa_enable_prop);
-			if (ret) {
-				dev_err(&pdev->dev,
-				"cannot activate headphone amplifier\n");
-				return -ENODEV;
+		if (tpa_prop) {
+			card->aux_dev = hifiberry_dacplus_aux_devs;
+			card->num_aux_devs =
+				ARRAY_SIZE(hifiberry_dacplus_aux_devs);
+			if (strcmp((char *)tpa_prop->value, "okay")) {
+				/* and activate headphone using change_sets */
+				dev_info(&pdev->dev,
+					"activating headphone amplifier\n");
+				of_changeset_init(&ocs);
+				ret = of_changeset_update_property(&ocs,
+					tpa_node, &tpa_enable_prop);
+				if (ret) {
+					dev_err(&pdev->dev,
+					"cannot activate headphone amplifier\n");
+					return -ENODEV;
+				}
+				ret = of_changeset_apply(&ocs);
+				if (ret) {
+					dev_err(&pdev->dev,
+					"cannot activate headphone amplifier\n");
+					return -ENODEV;
+				}
 			}
-			ret = of_changeset_apply(&ocs);
-			if (ret) {
-				dev_err(&pdev->dev,
-				"cannot activate headphone amplifier\n");
-				return -ENODEV;
-			}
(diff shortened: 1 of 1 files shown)


j-schambacher:
This is (hopefully) the last fix for 5.4. We will probably start testing with 5.10 soon.

sound/soc/bcm/hifiberry_dacplus.c
```diff
@@ -406,29 +406,34 @@ static int snd_rpi_hifiberry_dacplus_probe(struct platform_device *pdev)
...
-				"cannot activate headphone amplifier\n");
-				return -ENODEV;
-			}
-			ret = of_changeset_apply(&ocs);
-			if (ret) {
-				dev_err(&pdev->dev,
-				"cannot activate headphone amplifier\n");
-				return -ENODEV;
+		if (tpa_prop) {
+			card->aux_dev = hifiberry_dacplus_aux_devs;
```
pelwell, line 412: Is there a reason that these assignments aren't also conditional on the node being enabled?

sound/soc/bcm/hifiberry_dacplus.c
```diff
@@ -406,29 +406,34 @@ static int snd_rpi_hifiberry_dacplus_probe(struct platform_device *pdev)
...
+			card->num_aux_devs =
+				ARRAY_SIZE(hifiberry_dacplus_aux_devs);
+			if (strcmp((char *)tpa_prop->value, "okay")) {
```
pelwell, line 415: Strictly speaking, the lack of a `status` property also causes a node to be enabled, but see above comment about a better way.

sound/soc/bcm/hifiberry_dacplus.c
```diff
@@ -406,29 +406,34 @@ static int snd_rpi_hifiberry_dacplus_probe(struct platform_device *pdev)
 	if (ret < 0)
 		return ret;
 	if (ret) {
-		card->aux_dev = hifiberry_dacplus_aux_devs;
-		card->num_aux_devs =
-				ARRAY_SIZE(hifiberry_dacplus_aux_devs);
 		tpa_node = of_find_compatible_node(NULL, NULL, "ti,tpa6130a2");
 		tpa_prop = of_find_property(tpa_node, "status", &len);
```
pelwell, line 410: Checking the status of a node can be accomplished with the standard function `of_device_is_available`.

sound/soc/bcm/hifiberry_dacplus.c
j-schambacher: This allows the sole definition/activation in the DT-overlay (in case status is set to "okay" already)

sound/soc/bcm/hifiberry_dacplus.c
j-schambacher: Right, but the two variables are needed later when used in of_changeset_update_property(). These statements are still almost the same as in my commit from last July
https://github.com/j-schambacher/linux/commit/690f3fc9e373f69730c568c473c8d80f88eb00b1#diff-71066e9cdae82faaf0504420fc0e0c9ed62a9d7f51a775b74f95dd847feb4045
What I want to do is to change only the value of the status property - in case the sound card can activate the headphone amp and the amp is physically present. This allows us to simply minimize the number of drivers for our zoo of cards w/ and w/o headphone amps.
This commit now only activates the aux-device in case both conditions are true. I hope my explanation helps.

pelwell: I'm saying that you can replace:
```
		tpa_prop = of_find_property(tpa_node, "status", &len);
		if (tpa_prop) {
			card->aux_dev = hifiberry_dacplus_aux_devs;
			card->num_aux_devs =
				ARRAY_SIZE(hifiberry_dacplus_aux_devs);
			if (strcmp((char *)tpa_prop->value, "okay")) {
```
with 
```
		if (!of_device_is_available(tpa_node)) {
			card->aux_dev = hifiberry_dacplus_aux_devs;
			card->num_aux_devs =
				ARRAY_SIZE(hifiberry_dacplus_aux_devs);
```

pelwell: Comment updated (the GitHub GUI gets confusing when there are multiple commented lines close together).

j-schambacher: Ok, understood, smart function. Your proposal will combine the two if-statements into one, which unfortunately does not lead to the same result. The original code distinguishes between the absence and checking the value of the existing property. The dev_warn() should be issued only in case of a missing property. This has then to be checked with of_find_property(), which will return NULL if node is NULL or does not exist. So, overall it will be around the same complexity as the original code. Please let me know, whether I should try modifying the code according your suggestion.

pelwell: I see - there's not much in it.  Leave it as is.

PR #4177, opened 2021-02-25
draft: drivers/char/broadcom: Fix some kconfig related bugs

kcxt:
Fix a warning when `CONFIG_DEBUG_FS` is not enabled.

Fix an error when `CONFIG_ARM_DMA_MEM_BUFFERABLE` is not enabled. The bcm2835-gpiomem driver depends on this option to compile but isn't marked as such in Kconfig. I wasn't able to figure out what config option usually enables this, I've selected it in `ARCH_BCM2835` although I'm not sure if this is the correct approach.


kcxt:
Some further inspection reveals some issues in my defconfig. I think there is still an issue with ARM_DMA_MEM_BUFFERABLE being selected, but this commit doesn't fix it.

PR #4189, opened 2021-03-05
media: bcm2835-unicam: Fix bug in buffer swapping logic

naushir:
If multiple sets of interrupts occur simultaneously, it may be unsafe
to swap buffers, as the hardware may already be re-using the current
buffers. In such cases, avoid swapping buffers, and wait for the next
opportunity at the Frame End interrupt to signal completion.

Additionally, check the packet compare status when watching for frame
end for buffers swaps, as this could also signify a frame end event.

### DIFF
--- a/drivers/media/platform/bcm2835/bcm2835-unicam.c
+++ b/drivers/media/platform/bcm2835/bcm2835-unicam.c
@@ -798,6 +798,7 @@ static irqreturn_t unicam_isr(int irq, void *dev)
 	unsigned int sequence = unicam->sequence;
 	unsigned int i;
 	u32 ista, sta;
+	bool fe;
 	u64 ts;
 
 	sta = reg_read(unicam, UNICAM_STA);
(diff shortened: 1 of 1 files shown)


naushir:
@mdevaev and @JakubVanek here is a candidate fix for https://github.com/raspberrypi/linux/issues/4128.

Please let me know if you have any issues with this fix.

naushir:
Attached as a patch file for your convenience.

[0001-media-bcm2835-unicam-Fix-bug-in-buffer-swapping-logi.zip](https://github.com/raspberrypi/linux/files/6091587/0001-media-bcm2835-unicam-Fix-bug-in-buffer-swapping-logi.zip)

pelwell:
FYI, clicking on any commit and adding `.patch` to the URL gets you the patch.

naushir:
> FYI, clicking on any commit and adding `.patch` to the URL gets you the patch.

Useful shortcut, thanks!

mdevaev:
Testing...

JakubVanek:
It looks good here - with this patch I can't reproduce any of the problems using JPEG or H264 modes of ustreamer. Before applying this, the NULL dereference occurred the second time I plugged in the cable. Thank you!

mdevaev:
I don't see the problem anymore either. Apparently this is fixed. Thanks!

naushir:
Thanks!

@6by9 this looks like it has fixed this problem.

6by9:
Your commit text would be frowned upon as it describes two things.

```fe``` is a touch cryptic to the casual reader.

But I'm happy enough with it.

PR #4201, opened 2021-03-10
BCM2711 KMS YUV output

mripard:
Hi,

This is a PR to support the YUV 444 and 420 output from the KMS driver on the BCM2711.

While it works, it's marked as a draft since it introduces a new KMS property. Indeed, DRM requires for any new property that it's been agreed on and used by an open-source project: https://www.kernel.org/doc/html/latest/gpu/drm-uapi.html#open-source-userspace-requirements

I was thinking Kodi could be that vessel. The process next would be to make sure we have some patches to support it from within Kodi, and that the Kodi maintainers have reviewed it and agreed to it. However, it shouldn't be merged at this point, since this interface isn't stable from the kernel point-of-view, so it could change.

Once there, we can submit it to the kernel, get it merged, and then we will be able to merge it within Kodi.


drivers/gpu/drm/vc4/vc4_hdmi.c
```diff
@@ -1221,8 +1230,7 @@ static void vc4_hdmi_encoder_pre_crtc_enable(struct drm_encoder *encoder,
 	struct vc4_hdmi_encoder *vc4_encoder = to_vc4_hdmi_encoder(encoder);
 	struct vc4_hdmi *vc4_hdmi = encoder_to_vc4_hdmi(encoder);
 
-	if (vc4_encoder->hdmi_monitor &&
-	    drm_default_rgb_quant_range(mode) == HDMI_QUANTIZATION_RANGE_LIMITED) {
+	if (vc4_hdmi_is_full_range_rgb(vc4_hdmi, mode) {
```
6by9: Hasn't this condition been inverted?
The original checked for being limited range, but now we're looking for being full range RGB.

drivers/gpu/drm/vc4/vc4_hdmi.c
```diff
@@ -1229,20 +1228,10 @@ static void vc4_hdmi_encoder_pre_crtc_enable(struct drm_encoder *encoder,
...
-
-		vc4_encoder->limited_rgb_range = true;
-	} else {
-		if (vc4_hdmi->variant->csc_setup)
-			vc4_hdmi->variant->csc_setup(vc4_hdmi, mode);
-
-		vc4_encoder->limited_rgb_range = false;
-	}
+	if (vc4_hdmi->variant->csc_setup)
+		vc4_hdmi->variant->csc_setup(vc4_hdmi, true);
```
6by9: csc_setup had just changed to taking the mode, but now we're back to ```true```. Think that should have been ```mode```.

drivers/gpu/drm/vc4/vc4_hdmi.c
```diff
@@ -917,13 +918,14 @@ static void vc4_hdmi_csc_setup(struct vc4_hdmi *vc4_hdmi, bool enable)
...
-static void vc5_hdmi_csc_setup(struct vc4_hdmi *vc4_hdmi, bool enable)
+static void vc5_hdmi_csc_setup(struct vc4_hdmi *vc4_hdmi,
+			       const struct drm_display_mode *mode)
 {
 	u32 csc_ctl;
 
 	csc_ctl = 0x07;	/* RGB_CONVERT_MODE = custom matrix, || USE_RGB_TO_YCBCR */
 
-	if (enable) {
+	if (vc4_hdmi_is_full_range_rgb(vc4_hdmi, mode)) {
```
6by9: Inverted. CSC is limited range, check is for full range.

drivers/gpu/drm/vc4/vc4_hdmi.c
```diff
@@ -917,6 +917,50 @@ static void vc4_hdmi_csc_setup(struct vc4_hdmi *vc4_hdmi,
...
 
+
+/*
+ * If we need to output Full Range RGB, then use the unity matrix
+ *
+ * [ 1      0      0      0]
+ * [ 0      1      0      0]
+ * [ 0      0      1      0]
+ *
+ * Matrix is signed 2p13 fixed point, with signed 9p6 offsets
```
6by9, line 1045: Need to check those descriptions. At least one of the CCMs had a very strange representation for negative values, but that may have been the ISP.

drivers/gpu/drm/vc4/vc4_hdmi.c
```diff
@@ -951,6 +962,38 @@ static const u16 vc5_hdmi_csc_full_rgb_to_limited_rgb[3][4] = {
...
+};
+
+/*
+ * Conversion between Full Range RGB and Full Range YUV444 using the
+ * BT.709 Colorspace
+ *
+ * [ -0.117208 -0.394207  0.511416  128 ]
+ * [  0.511416 -0.464524 -0.046891  128 ]
+ * [  0.212639  0.715169  0.072192  0   ]
+ *
```
6by9, line 1093: Oof, does it really swap the channels around. How annoying.

Do we ever have to worry about 601? It is in the HDMI spec, and can be advertised through the C1/C0 colorimetry bit in the AVI infoframe data byte 2. Park for later?

6by9 (commented):
I think you pushed again whilst I was reviewing.

mripard:
Following the discussion with upstream, I reworked that branch to support what the current consensus seems to be.
The most important changes at the moment is that there's now two properties: 

 * one to expose which output formats are supported through a bitmask, updated based on the EDID retrieved from the display and the capabilities of the controller, 
 * and one to actually report and select the mode currently being used, that defaults to RGB888

The bugs reported by @6by9 should also be fixed, and it's based on the 4k PR I just sent, with 4k@60 with 12 bpc and in YUV422 that should work fine (unlike what @popcornmix reported previously).

Since it requires the userspace cooperation, we'll still need the work on Kodi to merge this

drivers/gpu/drm/vc4/vc4_hdmi.c
6by9: Just to resolve my own comment, it's the CSC coefficients in the dlist that I was thinking of. HDMI block is straightforward.

popcornmix:
Thanks. I had a patch for kodi that worked (for lower resolutions) with previous patch.
https://github.com/popcornmix/xbmc/commit/26c3aa23d3b9911e681bf33a6178140464a2a201

I'll try to get something similar that queries the available output formats. Assuming it works with this PR and gracefully handles the lack of these properties then we should be able to get it merged into Kodi.

mripard:
We don't really need to have it merged at that point, since it could still change due to the review process. What we need at this point is a statement from Kodi that they are ok with merging it in its current form :)

popcornmix:
@mripard `modetest -c` shows:
```
	38 available output formats:
		flags: immutable bitmask
		values: RGB444=0x1 YCrCb444=0x2 YCrCb422=0x4 YCrCb420=0x8
		value: 3
```
(and I've also seen `1` here)
but edid-decode shows:
```
Supports YCbCr 4:2:2
```
```
$ base64  /sys/devices/platform/gpu/drm/card1/card1-HDMI-A-1/edid
AP///////wAQrMCgTFQ5MAUZAQOANR546uJFqFVNoyYLUFSlSwBxT4GAqcCpQNHA4QABAQEBo2YA
oPBwH4AwIDUADyghAAAaAAAA/wBQMlBDMjUxUjA5VEwKAAAA/ABERUxMIFAyNDE1UQogAAAA/QAd
TB6MHgAKICAgICAgAasCAyrxU5AFBAIHFgEUHxITJyAhIgMGERUjCQcHbQMMABAAMDwgAGADAgEC
OoAYcTgtQFgsJQAPKCEAAB8BHYAYcRwWIFgsJQAPKCEAAJ4EdAAw8nBagLBYigAPKCEAAB5WXgCg
oKApUDAgNQAPKCEAABoAAAAAAAAAAAAAAAAA+Q==
```

popcornmix:
While
```
modetest -M vc4 -w "32:output format:2" -w "32:max bpc:12" -F tiles,gradient -s 32:3840x2160-30 -P95@80:3840x2160+0+0@XR24
```
gives me a picture on monitor is reports 36bits (good) 20Hz (not so good).
Analyser says:
```
Pixel clock 198MHz
H Freq 45kHz
V Freq 20Hz
```
So I don't think the clocks are right. I haven't worked out what effect 4:2:2 + 12bit should have on the pixel clock.

6by9:
Slightly odd one as selecting "max bpc:8" with YCbCr-422 ("output format:2") is making my analyser report 36bpp. GCP packets certainly say that too.

Confusingly
```
modetest -M vc4 -w "32:output format:2" -w "32:max bpc:8" -F tiles,gradient -s 32:1920x1080-60 -P95@80:3840x2160+0+0@XR24
```
which seems to select 36bpp gives correct colours (greyscale gradient)
```
modetest -M vc4 -w "32:output format:2" -w "32:max bpc:12" -F tiles,gradient -s 32:1920x1080-60 -P95@80:3840x2160+0+0@XR24
```
which tries to force 36bpp, gives a serious red tinge to everything.
Both modes report a character rate/TMDS clock of 148.5MHz.

My analyser appears not to give frame rate or timing info on the HDMI2.0 interface, but does on the HDMI1.4 one. In that mode it does confirm that I am getting 45fps instead of my requested 60. for max bpc:12. 
H 2880/132/66/222 (total 3300) V 1080/4/5/36 (total 1125)
cf max bpc:8
H 1920/ 88/44/148 (total 2200) V 1080/4/5/36 (total 1125)

This sounds very familiar to when I was looking at similar things. Now if only I could remember which clock needs to be scaled up to sort the issue!

popcornmix:
@mripard after a long discussion with Broadcom I think https://github.com/popcornmix/linux/commit/735899f44bfe488066bc7fee44f9a97d334e6bac is what we want.
Without this we get YCC422 1080p60 detected as 1080p40.

Basically YCC422 is always 36-bit, and does not require deep colour signalling (from HDMI 1.4 spec).

mripard:
I just pushed a new version with your commit merged (thanks for figuring it out!), and a fix for the missing YCbCr422 in the bitmask

mripard:
This work has been superseded by PR #4754, which has been merged in the RPi tree and upstream. Closing.

PR #4206, opened 2021-03-12
overlays: gpio-led: new overlay

agordon:
Add generic connection between the kernel's LED framework and
RPI's GPIO pins.

Signed-off-by: Assaf Gordon <assafgordon@gmail.com>

### DIFF
--- a/arch/arm/boot/dts/overlays/README
+++ b/arch/arm/boot/dts/overlays/README
@@ -914,6 +914,62 @@ Params: gpio                    GPIO pin to trigger on (default 3)
         keycode                 Set the key code for the button
 
 
+
+Name:   gpio-led
+Info:   This is a generic overlay for activating LEDs (or any other component)
+        by a GPIO pin. Multiple LEDs can be set up using multiple calls to the
+        overlay. While there are many existing methods to activate LEDs on the
+        RPi, this method offers some advantages:
+        1) Does not require any userspace programs.
+        2) LEDs can be connected to the kernel's led-trigger framework,
+           and drive the LED based on triggers such as cpu load, heartbeat,
+           kernel panic, key input, timers and others.
+        3) LED can be tied to the input state of another GPIO pin.
+        4) The LED is setup early during the kernel boot process (useful
+           for cpu/heartbeat/panic triggers).
+
+        Typical electrical connection is:
+           RPI-GPIO.19  ->  LED  -> 300ohm resister  -> RPI-GND
+        The GPIO pin number can be changed with the 'gpio=' parameter.
+
+        To control an LED from userspace, write a 0 or 1 value:
+           echo 1 > /sys/class/leds/myled1/brightness
+        The 'myled1' name can be changed with the 'label=' parameter.
+
+        To connect the LED to a kernel trigger from userspace:
+           echo cpu > /sys/class/leds/myled1/trigger
+           echo heartbeat > /sys/class/leds/myled1/trigger
+           echo none > /sys/class/leds/myled1/trigger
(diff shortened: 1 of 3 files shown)


agordon:
Hello RPI developers,

I'd like to suggest the following small addition to the RPI's overlays: the ability to connect a LED to any output GPIO. Similar to `gpio-key` in nature, with the advantage of using the kernel's LED system, so able to connect a LED to a kernel trigger (e.g. cpu load / heartbeat / panic / etc ).
(I'm not a kernel developer, and I'm happy to adjust if needed to suit the coding style - let me know.)

arch/arm/boot/dts/overlays/README
```diff
@@ -914,6 +914,51 @@ Params: gpio                    GPIO pin to trigger on (default 3)
...
+           echo gpio > /sys/class/leds/myled1/trigger
+           echo 26 > /sys/class/leds/myled1/gpio
+        Available triggers:
+           cat /sys/class/leds/myled1/trigger
+
+        More information about the Linux kernel LED/Trigger system:
+           https://www.kernel.org/doc/Documentation/leds/leds-class.rst
+           https://www.kernel.org/doc/Documentation/leds/ledtrig-oneshot.rst
+Load:   dtoverlay=gpio-led,<param>=<val>
+Params: gpio                    GPIO pin connected to the LED (default 19)
```
pelwell, line 953: The active_low parameter requires an entry here.

arch/arm/boot/dts/overlays/gpio-led-overlay.dts
```diff
@@ -0,0 +1,108 @@
...
+/plugin/;
+
+/ {
+	compatible = "brcm,bcm2835";
+
+	fragment@0 {
+		// Configure the gpio pin controller
+		target = <&gpio>;
+		__overlay__ {
+			pin_state: led_pins@0 {
```
pelwell: I suggest you make the node "address" match the default gpio number, both for practical reasons and for appearances.

arch/arm/boot/dts/overlays/gpio-led-overlay.dts
```diff
@@ -0,0 +1,108 @@
...
+	};
+	fragment@1 {
+		target-path = "/";
+		__overlay__ {
+			leds: leds@0 {
+				compatible = "gpio-leds";
+				pinctrl-names = "default";
+				pinctrl-0 = <&pin_state>;
+				status = "okay";
+
```
pelwell, line 76: Since you'll be making other changes here, I don't think we need two blank lines.

arch/arm/boot/dts/overlays/gpio-led-overlay.dts
```diff
@@ -0,0 +1,108 @@
...
+
+				led: led {
+			                label = "myled1";
+					gpios = <&gpio 19 0>;
+
+			                linux,default-trigger = "none";
```
pelwell, line 80: Wouldn't `default-on` be a more useful default?

pelwell (changes requested):
What a coincidence - I was thinking about adding a similar overlay only yesterday in response to a Forum post: https://www.raspberrypi.org/forums/viewtopic.php?f=107&t=306585

This is basically fine as is - I just have a few suggestions and requests - but it does need an entry in the Makefile otherwise it won't get built (alphabetical order applies there as well).

agordon:
Hi,
Thanks for the feedback, I updated the PR, except changing the default trigger to 'default-on' - I'm not sure about that - do we want to force users to turn on LEDs as default?

pelwell:
It still needs a Makefile entry, but I'll accept the `none` trigger.

PR #4207, opened 2021-03-12
drm/vc4: crtc: Reduce PV fifo threshold on hvs4

popcornmix:
Experimentally have found PV on hv4 reports fifo full
error with expected settings and does not with one less

This appears as:
`[drm:drm_atomic_helper_wait_for_flip_done] *ERROR* [CRTC:82:crtc-3] flip_done timed out`

with bit 10 of PV_STAT set "HVS driving pixels when the PV FIFO is full"

### DIFF
--- a/drivers/gpu/drm/vc4/vc4_crtc.c
+++ b/drivers/gpu/drm/vc4/vc4_crtc.c
@@ -210,6 +210,7 @@ static u32 vc4_get_fifo_full_level(struct vc4_crtc *vc4_crtc, u32 format)
 {
 	const struct vc4_crtc_data *crtc_data = vc4_crtc_to_vc4_crtc_data(vc4_crtc);
 	const struct vc4_pv_data *pv_data = vc4_crtc_to_vc4_pv_data(vc4_crtc);
+	struct vc4_dev *vc4 = to_vc4_dev(vc4_crtc->base.dev);
 	u32 fifo_len_bytes = pv_data->fifo_depth;
 
 	/*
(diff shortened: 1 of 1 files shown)


popcornmix:
@mripard I believe this is the fix for the flip_done issue on Pi3.
Waiting for @HiassofT to confirm.

My test has got to 738 reboots of test where kodi is started and osd is brought up.
Without this patch it typically fails after a few reboots.

It's hard to say if any other conditions in this function would benefit from a threshold one lower.
We've never seen it on Pi4 and it's so hard to provoke reliably (heisenbug) that looking for it in other situations is unlikely to be successful.

Just note here that if flip_done error with bit 10 of PVSTAT set is found again then reducing fifo threshold for that case should be the first thing to try.

HiassofT:
I've been testing with this PR on RPi3B+ this whole afternoon and didn't get a single flip_done timeout - without the PR I had about a 50:50 chance if flip_done/screen cutout would hit me which made kodi testing nearly impossible.

I'd vote for getting this merged, it's a huge improvement for me (and other users as well)

drivers/gpu/drm/vc4/vc4_crtc.c
```diff
@@ -238,6 +241,13 @@ static u32 vc4_get_fifo_full_level(struct vc4_crtc *vc4_crtc, u32 format)
 		if (crtc_data->hvs_output == 5)
 			return 32;
 
+		/*
+		 * Experimentally have found PV on hv4 reports fifo full
```
6by9: s/hv4/hvs4

drivers/gpu/drm/vc4/vc4_crtc.c
```diff
@@ -210,6 +210,9 @@ static u32 vc4_get_fifo_full_level(struct vc4_crtc *vc4_crtc, u32 format)
 {
 	const struct vc4_crtc_data *crtc_data = vc4_crtc_to_vc4_crtc_data(vc4_crtc);
 	const struct vc4_pv_data *pv_data = vc4_crtc_to_vc4_pv_data(vc4_crtc);
+	struct drm_crtc *crtc = &vc4_crtc->base;
+	struct drm_device *dev = crtc->dev;
+	struct vc4_dev *vc4 = to_vc4_dev(dev);
```
6by9: Slight nitpicking but you don't need crtc, so you may as well do dev = vc4_crtc->base.dev (if I've got my indirections right). Or as dev isn't actually needed either, 
```
struct vc4_dev *vc4 = to_vc4_dev(vc4_crtc->base.dev);
```

popcornmix:
Updated to address @6by9's comments.

PR #4208, opened 2021-03-12
vc4/drm: Adjustments to hdmi audio dma to reduce glitches

popcornmix:
Avoid full hdmi audio fifo writes and increase hdmi audio axi priority to avoid lost samples.


popcornmix:
Draft, while waiting for feedback from @HiassofT

HiassofT:
Testing with my local sample files was fine, I didn't get a single dlate error. So everything OK from my side

popcornmix:
Removed draft status.

PR #4215, opened 2021-03-15
vc4/drm: vc4_plane: Remove subpixel positioning check

popcornmix:
There is little harm in ignoring fractional coordinates
(they just get truncated).

Without this:
modetest -M vc4 -F tiles,gradient -s 32:1920x1080-60 -P89@74:1920x1080*.1.1@XR24

is rejected. We have the same issue in Kodi when trying to
use zoom options on video.

Note: even if all coordinates are fully integer. e.g.
src:[0,0,1920,1080] dest:[-10,-10,1940,1100]

it will still get rejected as drm_atomic_helper_check_plane_state
uses drm_rect_clip_scaled which transforms this to fractional src coords


popcornmix:
Ping @mripard @6by9

6by9:
I have no objections over that.
Possibly round to the nearest pixel rather than truncate though?

popcornmix:
Updated. Added rounding.

6by9:
I've not tried it, but it looks fine.

Issue #4197, opened 2021-03-09
TW6869 PCIe capture card DMA allocation trouble


walmis:
I'm trying to get TW6869 TV capture card running, but I'm seem to be stuck at a DMA allocation issue. I think this is related to DMA coherent allocations. I've increased coherent_pool=32M, but that did not help. Increasing CMA also does not help. Kernel config attached. I'd be grateful for any pointers on how to debug this.

```
[    1.550499] brcm-pcie 47d500000.pcie: host bridge /scb/pcie@7d500000 ranges:
[    1.550534] brcm-pcie 47d500000.pcie:   No bus range found for /scb/pcie@7d500000, using [bus 00-ff]
[    1.550608] brcm-pcie 47d500000.pcie:      MEM 0x0600000000..0x063fffffff -> 0x00c0000000
[    1.550688] brcm-pcie 47d500000.pcie:   IB MEM 0x0000000000..0x00ffffffff -> 0x0400000000
[    1.584131] brcm-pcie 47d500000.pcie: link up, 2.5 GT/s PCIe x1 (SSC)
[    1.584319] brcm-pcie 47d500000.pcie: PCI host bridge to bus 0000:00
[    1.584342] pci_bus 0000:00: root bus resource [bus 00-ff]
[    1.584365] pci_bus 0000:00: root bus resource [mem 0x600000000-0x63fffffff] (bus address [0xc0000000-0xffffffff])
[    1.584449] pci 0000:00:00.0: [14e4:2711] type 01 class 0x060400
[    1.584683] pci 0000:00:00.0: PME# supported from D0 D3hot
[    1.588387] pci 0000:00:00.0: bridge configuration invalid ([bus ff-ff]), reconfiguring
[    1.588634] pci 0000:01:00.0: TW686x PCI class overridden (0x000000 -> 0x048001)
[    1.588656] pci 0000:01:00.0: [1797:6869] type 00 class 0x048001
[    1.588713] pci 0000:01:00.0: reg 0x10: [mem 0x00000000-0x00000fff pref]
[    1.588847] pci 0000:01:00.0: enabling Extended Tags
[    1.589049] pci 0000:01:00.0: PME# supported from D0 D3hot D3cold
[    1.601400] pci_bus 0000:01: busn_res: [bus 01-ff] end is updated to 01
[    1.601444] pci 0000:00:00.0: BAR 8: assigned [mem 0x600000000-0x6000fffff]
[...]

PR #4216, opened 2021-03-15
overlays: Add pcie-32bit-dma overlay

vianpl:
In order to accommodate full PCI DMA access to memory on newer BCM2711
revisions, we're forced to map PCIe's view of physical memory with an
offset. This offset makes DMA addressing dependent on having 64bit
support on the PCI device's side. Which isn't always the case.

In order to mitigate this, introduce the pcie-32bit-dma overlay which
will forbid firmware from updating the default inbound memory window.
The default setting, albeit limited to accessing the lower 3GB of
memory, will allow for 32bit DMA addresses at the expense of having to
bounce buffers.

Link: https://github.com/raspberrypi/linux/issues/4197
Signed-off-by: Nicolas Saenz Julienne <nsaenzjulienne@suse.de>

### DIFF
--- a/arch/arm/boot/dts/overlays/README
+++ b/arch/arm/boot/dts/overlays/README
@@ -2079,6 +2079,13 @@ Params: addr                    I2C address of expander. Default 0x20.
         xra1202                 Select the Exar XRA1202 (8 bit)
 
 
+Name:   pcie-32bit-dma
+Info:   Force PCIe config to support 32bit DMA addresses at the expense of
+        having to bounce buffers.
+Load:   dtoverlay=pcie-32bit-dma
+Params: <None>
+
+
 [ The pcf2127-rtc overlay has been deleted. See i2c-rtc. ]
 
 
--- a/arch/arm/boot/dts/overlays/Makefile
+++ b/arch/arm/boot/dts/overlays/Makefile
@@ -127,6 +127,7 @@ dtbo-$(CONFIG_ARCH_BCM2835) += \
 	ov9281.dtbo \
 	papirus.dtbo \
 	pca953x.dtbo \
+	pcie-32bit-dma.dtbo \
 	pibell.dtbo \
 	pifacedigital.dtbo \
 	pifi-40.dtbo \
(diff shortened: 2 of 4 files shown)


arch/arm/boot/dts/overlays/README
```diff
@@ -2078,6 +2078,10 @@ Params: addr                    I2C address of expander. Default 0x20.
         pca9654                 Select the Onnn PCA9654 (8 bit)
         xra1202                 Select the Exar XRA1202 (8 bit)
```
pelwell, line 2080: Two blank  lines between entries, please.

arch/arm/boot/dts/overlays/README
```diff
@@ -2078,6 +2078,10 @@ Params: addr                    I2C address of expander. Default 0x20.
...
 
+Name:   pci-32bit-dma
+Info:   Force PCIe config to support 32bit DMA addresses at the expense of
+	having to bounce buffers.
+Load:   dtoverlay=pci-32bit-dma
```
pelwell, line 2088: Ditto.

arch/arm/boot/dts/overlays/README
```diff
@@ -2078,6 +2078,10 @@ Params: addr                    I2C address of expander. Default 0x20.
         pca9654                 Select the Onnn PCA9654 (8 bit)
         xra1202                 Select the Exar XRA1202 (8 bit)
 
+Name:   pci-32bit-dma
+Info:   Force PCIe config to support 32bit DMA addresses at the expense of
+	having to bounce buffers.
```
pelwell: Replace this TAB with 8 spaces.

arch/arm/boot/dts/overlays/README
pelwell, line 2088: And add `Params: <None>`.

pelwell (commented):
I have a few README niggles, otherwise it looks good.

pelwell:
Doh - and the name of the overlay is wrong in the README.

vianpl:
Fixed! It's amazing how I consistently get this sort of thing wrong...

pelwell:
Thanks - I'll back-port to 5.10.

PR #4220, opened 2021-03-17
staging/bcm2835-camera: Add support for DMABUFs

6by9:
DMABUFs are all handled by videobuf2, so there is no reason not
to enable support for them.

Note that this driver is still using the vmalloc allocator, so
the buffers it allocates will not be compatible with the codec
or ISP driver that require contiguous buffers. However this
driver should be able to import the buffers allocated by them.


6by9:
@pelwell Any chance of a review on this please?

PR #4222, opened 2021-03-18
Make rpi poe fan less noisy in cool environments

ProBackup-nl:
Run the PoE hat fan with reduced noise when possible. The PoE hat fan will spin even at PWM=1 and spins almost silent. Tested at 17ºC, the fan PWM alternates between 1 and 10, which is a lot less noisy then alternating between PWM levels 0 and 31.

### DIFF
--- a/arch/arm/boot/dts/overlays/rpi-poe-overlay.dts
+++ b/arch/arm/boot/dts/overlays/rpi-poe-overlay.dts
@@ -16,7 +16,7 @@
 				cooling-min-state = <0>;
 				cooling-max-state = <4>;
 				#cooling-cells = <2>;
-				cooling-levels = <0 31 63 150 255>;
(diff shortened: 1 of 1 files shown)


pelwell:
For me, with the default thresholds that causes the fan to spin permanently, but near silently. With a light load it speeds up slightly. The jump is now more pronounced, but most of the time one won't hit it.

Any comments, @ghollingworth ?

ghollingworth:
Ah this is PoE fan...  not case fan...

@XECDesign 

For the case fan, we only toggle on/off at a relatively high temperature (80 degrees by default) since it would otherwise be on when it's not required...

pelwell:
@XECDesign is happy to give this new curve a try. There was some concern about older, dusty fans taking a bit more drive to get started, but in that case the temperature would rise and one of the higher levels would kick in.

PR #4226, opened 2021-03-19
WIP: "Dummy" CSI2 sensor driver

6by9:
To replace raspiraw and the MMAL rawcam component, provide a dummy V4L2 sensor driver that will accept almost any resolution and format asked of it by userspace, which it will then use to configure bcm2835-unicam.

Number of lanes and clock setup needs to come from DT.

I'm now converting https://github.com/raspberrypi/raspiraw to use this instead of MMAL.

### DIFF
--- a/drivers/media/i2c/dummy_sensor.c
+++ b/drivers/media/i2c/dummy_sensor.c
@@ -1,0 +1,477 @@
+// SPDX-License-Identifier: GPL-2.0
+/*
+ * A V4L2 driver for CSI2 sources with no control.
+ * This can be of use when interfacing to an FPGA or other source that is
+ * constantly streaming data, and the V4L2 receiver device just needs to be
+ * configured with the image parameters for the incoming stream, or where
+ * userspace has to send the relevant configuration
+ *.
+ * V4L2 controls are created for the base parameters that libcamera insists
+ * exist, but they are all read-only with dummy values.
+ *
+ * Copyright (C) 2021, Raspberry Pi (Trading) Ltd.
+ *
+ * Based on Sony imx219 camera driver
+ * Copyright (C) 2020 Raspberry Pi (Trading) Ltd.
+ *
+ */
+
+#include <linux/delay.h>
+#include <linux/module.h>
+#include <linux/platform_device.h>
+#include <linux/regulator/consumer.h>
+#include <media/v4l2-ctrls.h>
+#include <media/v4l2-device.h>
+#include <media/v4l2-event.h>
+#include <media/v4l2-fwnode.h>
+#include <media/v4l2-mediabus.h>
+#include <asm/unaligned.h>
+
+/* Array of all the mbus formats that we'll accept */
+u32 mbus_codes[] = {
+	MEDIA_BUS_FMT_SBGGR8_1X8,
+	MEDIA_BUS_FMT_SGBRG8_1X8,
+	MEDIA_BUS_FMT_SGRBG8_1X8,
+	MEDIA_BUS_FMT_SRGGB8_1X8,
+	MEDIA_BUS_FMT_SBGGR10_1X10,
+	MEDIA_BUS_FMT_SGBRG10_1X10,
+	MEDIA_BUS_FMT_SGRBG10_1X10,
+	MEDIA_BUS_FMT_SRGGB10_1X10,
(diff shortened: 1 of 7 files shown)


drivers/media/i2c/dummy_sensor.c
```diff
@@ -0,0 +1,477 @@
...
+ * A V4L2 driver for CSI2 sources with no control.
+ * This can be of use when interfacing to an FPGA or other source that is
+ * constantly streaming data, and the V4L2 receiver device just needs to be
+ * configured with the image parameters for the incoming stream, or where
+ * userspace has to send the relevant configuration
+ *.
+ * V4L2 controls are created for the base parameters that libcamera insists
+ * exist, but they are all read-only with dummy values.
+ *
+ * Copyright (C) 2021, Raspberry Pi (Trading) Ltd
```
pelwell: How about a `.` after `Ltd`, to match the line below?

drivers/media/i2c/dummy_sensor.c
```diff
@@ -0,0 +1,477 @@
...
+
+static const struct v4l2_ctrl_ops sensor_ctrl_ops = {
+	.s_ctrl = sensor_set_ctrl,
+};
+
+static int sensor_enum_mbus_code(struct v4l2_subdev *sd,
+				 struct v4l2_subdev_pad_config *cfg,
+				 struct v4l2_subdev_mbus_code_enum *code)
+{
+	if (code->index > NUM_MBUS_CODES)
```
pelwell: `>=`

drivers/media/i2c/dummy_sensor.c
```diff
@@ -0,0 +1,477 @@
...
+			break;
+
+	if (i >= NUM_MBUS_CODES)
+		i = 0;
+
+	fmt->format.code = mbus_codes[i];
+	fmt->format.field = V4L2_FIELD_NONE;
+	fmt->format.colorspace = V4L2_COLORSPACE_SRGB;
+	fmt->format.ycbcr_enc =
+			V4L2_MAP_YCBCR_ENC_DEFAULT(fmt->format.colorspace);
```
pelwell: One fewer TAB, to line up with the others?

drivers/media/i2c/dummy_sensor.c
```diff
@@ -0,0 +1,477 @@
...
+	.remove = sensor_remove,
+	.driver = {
+		.name = "dummy_csi2_sensor",
+		.of_match_table	= sensor_dt_ids,
+	},
+};
+
+module_platform_driver(sensor_driver);
+
+MODULE_AUTHOR("Dave Stevenson <dave.stevenson@raspberrypi.com");
```
pelwell: Non-RFC822-compliant email address - missing `>`.

pelwell (commented):
A few noddy suggestions, but as a non-V4L2-speaker it looks plausible.

6by9:
Thanks for the review comments. Hopefully those are all fixed up now.

I have missed adding it to the defconfig though, so I'll leave the WIP for now.

I've proven it using a hacked raspiraw which does no MMAL or V4L2 stuff (only sensor config over I2C) with ```v4l2-ctl --stream-mmap=3 --stream-to=/dev/null --stream-count=1000``` running in a different terminal, and V4L2/unicam does start revceiving frames when raspiraw starts, and stops when complete. I'm fairly happy it's all good in that regard.

pelwell:
For the benefit of other readers, the overlay works because fragments that target other fragments get applied before other fragments are merged with the base. Had there been two levels of nesting (fragments that target fragments that target fragments) then it would have been necessary to sort the fragments into application order and put the main fragment at the end.

6by9:
> Had there been two levels of nesting (fragments that target fragments that target fragments) then it would have been necessary to sort the fragments into application order and put the main fragment at the end.

Which is the bit that I never get right :-)

pelwell:
I think you're the only person to attempt it, therefore nobody gets it right.

6by9:
This is dead as it's against 5.10.

It'll get rebased against 6.1 at some point (https://github.com/6by9/linux/tree/rpi-5.15.y-dummy_sensor has the equivalent patches against 5.15, but I've noted things to fix there).

PR #4230, opened 2021-03-21
Added Geniatech MyGica T230C2_LITE and T230A

pego149:
Thread about adding T230A: https://forum.libreelec.tv/thread/23142-mygica-t230a/


pelwell:
This is a change to an unmodified upstream file, and should be submitted to the upstream maintainers. The get_maintainer script suggests emailing the following:
```
Mauro Carvalho Chehab <mchehab@kernel.org>
Sean Young <sean@mess.org>
Julia Lawall <Julia.Lawall@inria.fr>
Hans Verkuil <hverkuil-cisco@xs4all.nl>
Jan Pieter van Woerkom <jp@jpvw.nl>
linux-media@vger.kernel.org
linux-kernel@vger.kernel.org
```
In practice I suspect that if you email Jan Pieter he will take on the change himself.

PR #4232, opened 2021-03-22
vc4/drm: Fix source offsets with DRM_FORMAT_P030

popcornmix:
Spec says: bits [31:4] of the given address should point to
the 128-bit word containing the desired starting pixel,
and bits[3:0] should be between 0 and 11, indicating which
of the 12-pixels in that 128-bit word is the first pixel to be used

### DIFF
--- a/drivers/gpu/drm/vc4/vc4_plane.c
+++ b/drivers/gpu/drm/vc4/vc4_plane.c
@@ -824,9 +824,20 @@ static int vc4_plane_mode_set(struct drm_plane *plane,
 		u32 tile_w, tile, x_off, pix_per_tile;
 
 		if (fb->format->format == DRM_FORMAT_P030) {
+			/*
+			 * Spec says: bits [31:4] of the given address should point to
+			 * the 128-bit word containing the desired starting pixel,
+			 * and bits[3:0] should be between 0 and 11, indicating which
+			 * of the 12-pixels in that 128-bit word is the first pixel to be used
+			 */
+			u32 aligned = vc4_state->src_x / 12;
+			u32 last_bits = vc4_state->src_x % 12;
+
+			x_off = aligned * 16 + last_bits;
 			hvs_format = HVS_PIXEL_FORMAT_YCBCR_10BIT;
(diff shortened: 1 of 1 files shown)


HiassofT:
Testing on LibreELEC looks good, zooming 10bit HEVC now works fine without artifacts

drivers/gpu/drm/vc4/vc4_plane.c
```diff
@@ -872,9 +881,7 @@ static int vc4_plane_mode_set(struct drm_plane *plane,
 			vc4_state->offsets[i] += src_y /
 						 (i ? v_subsample : 1) *
 						 tile_w;
-			vc4_state->offsets[i] += x_off /
-						 (i ? h_subsample : 1) *
-						 fb->format->cpp[i];
+			vc4_state->offsets[i] += x_off & ~(i ? 1:0);
```
pelwell: The lack of spaces around the colon is non-standard, but kernel Devs would probably prefer `~!!i`.

popcornmix:
Positive report [here](https://forum.kodi.tv/showthread.php?tid=361164&pid=3025280#pid3025280)
And I'm happy with `~!!i`

drivers/gpu/drm/vc4/vc4_plane.c
```diff
@@ -826,7 +826,17 @@ static int vc4_plane_mode_set(struct drm_plane *plane,
...
-			tile_w = 96;
+			tile_w = 128;
+			pix_per_tile = 96;
+			/*
+			 * Spec says: bits [31:4] of the given address should point to
+			 * the 128-bit word containing the desired starting pixel,
+			 * and bits[3:0] should be between 0 and 11, indicating which
+			 * of the 12-pixels in that 128-bit word is the first pixel to be used
+			 */
+			u32 aligned = vc4_state->src_x / 12;
```
pelwell, line 833: Mid block declarations are not permitted.

popcornmix:
Updated.
Unfortunately `~!!0` produces a warning (`~ used on boolean, did you mean !`).

PR #4240, opened 2021-03-27
Fix for Full-KMS VEC on Pi4

### LINKED ISSUE #4233: Full-KMS not working for composite output on Pi4
There is a couple of things broken in the kernel that prevent composite output from working on Raspberry Pi 4 Model B.

These are actually pretty trivial to fix:
* [VEC's address](https://github.com/raspberrypi/linux/blob/c9226080e513181ffb3909a905e9c23b8a6e8f62/arch/arm/boot/dts/bcm283x.dtsi#L493) is set to `0x7e806000` on all Pis. On Pi4, it is actually located at `0x7ec13000`. This needs to be fixed in the device tree.
* The `VEC_DAC_CONFIG` register is [configured to `0x460C000C`](https://github.com/raspberrypi/linux/blob/c9226080e513181ffb3909a905e9c23b8a6e8f62/drivers/gpu/drm/vc4/vc4_vec.c#L454) by the VEC driver. This is also the value set by the firmware on Pi3 (I'm not sure about the older models). But on Pi4, the value set by the firmware is `0x61800000`, while `0x460C000C` breaks the image. I'm not sure how to interpret that; possibly consulting firmware devs would be a good idea.
  * Note; `0x61800000` produces a usable image on the Pi3 for me as well. But again, I have very little idea what that means, especially about the lower bits - the higher bits seem to configure some DC bias.
* The PixelValve for the VEC on Pi4 is [configured to use clock source 0](https://github.com/raspberrypi/linux/blob/c9226080e513181ffb3909a905e9c23b8a6e8f62/drivers/gpu/drm/vc4/vc4_crtc.c#L994). This should actually be `PV_CONTROL_CLK_SELECT_VEC`, like [for Pi3](https://github.com/raspberrypi/linux/blob/c9226080e513181ffb3909a905e9c23b8a6e8f62/drivers/gpu/drm/vc4/vc4_crtc.c#L940).

I applied this quick and dirty patch onto the current `rpi-5.10.y` branch:
```patch
diff --git a/drivers/gpu/drm/vc4/vc4_crtc.c b/drivers/gpu/drm/vc4/vc4_crtc.c
index 6d71c7e6979a..eb808e111638 100644
--- a/drivers/gpu/drm/vc4/vc4_crtc.c
+++ b/drivers/gpu/drm/vc4/vc4_crtc.c
@@ -999,7 +999,7 @@ static const struct vc4_pv_data bcm2711_pv3_data = {
        .fifo_depth = 64,
        .pixels_per_clock = 1,
        .encoder_types = {
-               [0] = VC4_ENCODER_TYPE_VEC,
+               [PV_CONTROL_CLK_SELECT_VEC] = VC4_ENCODER_TYPE_VEC,
        },
 };
 
diff --git a/drivers/gpu/drm/vc4/vc4_vec.c b/drivers/gpu/drm/vc4/vc4_vec.c
index bd5b8eb58b18..ec96b77a8d80 100644
--- a/drivers/gpu/drm/vc4/vc4_vec.c
+++ b/drivers/gpu/drm/vc4/vc4_vec.c
@@ -451,10 +451,7 @@ static void vc4_vec_encoder_enable(struct drm_encoder *encoder)
        VEC_WRITE(VEC_CONFIG2,
                  VEC_CONFIG2_UV_DIG_DIS | VEC_CONFIG2_RGB_DIG_DIS);
        VEC_WRITE(VEC_CONFIG3, VEC_CONFIG3_HORIZ_LEN_STD);
-       VEC_WRITE(VEC_DAC_CONFIG,
-                 VEC_DAC_CONFIG_DAC_CTRL(0xc) |
-                 VEC_DAC_CONFIG_DRIVER_CTRL(0xc) |
-                 VEC_DAC_CONFIG_LDO_BIAS_CTRL(0x46));
+       VEC_WRITE(VEC_DAC_CONFIG, 0x61800000);
 
        /* Mask all interrupts. */
        VEC_WRITE(VEC_MASK0, 0);
```
And also loaded this quick and dirty device tree overlay:
```dts
/dts-v1/;
/plugin/;

/ {
    compatible = "brcm,brcm2835-vec";

    fragment@0 {
        target = <&vec>;
        __overlay__ {
            reg = <0x7ec13000 0x1000>;
        };
    };
};
```
And I can confirm that then, `enable_tvout=1` and `dtoverlay=vc4-kms-v3d-pi4,composite=1` in `config.txt` result in usable picture in X on Pi4.

I would post a full-blown pull request, but I'm not familiar with device trees and such enough to know how to realize detection of Pi revision in the VEC driver elegantly and in line with best practices in the kernel. There are more problems with the Full-KMS VEC driver, so I might post pull requests for those later, but I don't feel confident enough to mess with device trees. Sorry about that. I could also just set `VEC_DAC_CONFIG` to `0x61800000` regardless of hardware, as that seems to work - but also I think that such change should be consulted with firmware devs (and/or non-public Broadcom documentation), because I really have no idea what I'm setting with that register ;)

kFYatek:
This fixes #4233

I hope that I didn't mess up any formalities, as this is my first PR here.

I tested it on a Pi 3B and a Pi 4B, both happily display a composite output with Full-KMS now.

### DIFF
--- a/Documentation/devicetree/bindings/display/brcm,bcm2835-vec.yaml
+++ b/Documentation/devicetree/bindings/display/brcm,bcm2835-vec.yaml
@@ -11,7 +11,9 @@ maintainers:
 
 properties:
   compatible:
-    const: brcm,bcm2835-vec
+    enum:
+      - brcm,bcm2835-vec
+      - brcm,bcm2711-vec
 
   reg:
     maxItems: 1
--- a/arch/arm/boot/dts/bcm2711.dtsi
+++ b/arch/arm/boot/dts/bcm2711.dtsi
@@ -301,7 +301,7 @@
 		};
 
 		vec: vec@7ec13000 {
-			compatible = "brcm,bcm2835-vec";
+			compatible = "brcm,bcm2711-vec";
 			reg = <0x7ec13000 0x1000>;
 			clocks = <&clocks BCM2835_CLOCK_VEC>;
 			interrupts = <2 27>;
--- a/arch/arm/boot/dts/bcm2711.dtsi
+++ b/arch/arm/boot/dts/bcm2711.dtsi
@@ -300,6 +300,14 @@
 			status = "disabled";
 		};
 
+		vec: vec@7ec13000 {
+			compatible = "brcm,bcm2835-vec";
+			reg = <0x7ec13000 0x1000>;
+			clocks = <&clocks BCM2835_CLOCK_VEC>;
+			interrupts = <2 27>;
+			status = "disabled";
+		};
+
 		dvp: clock@7ef00000 {
 			compatible = "brcm,brcm2711-dvp";
 			reg = <0x7ef00000 0x10>;
--- a/arch/arm/boot/dts/bcm2835-common.dtsi
+++ b/arch/arm/boot/dts/bcm2835-common.dtsi
@@ -106,6 +106,14 @@
 			status = "okay";
 		};
 
+		vec: vec@7e806000 {
+			compatible = "brcm,bcm2835-vec";
+			reg = <0x7e806000 0x1000>;
+			clocks = <&clocks BCM2835_CLOCK_VEC>;
+			interrupts = <2 27>;
+			status = "disabled";
+		};
+
 		pixelvalve@7e807000 {
 			compatible = "brcm,bcm2835-pixelvalve2";
 			reg = <0x7e807000 0x100>;
--- a/arch/arm/boot/dts/bcm283x.dtsi
+++ b/arch/arm/boot/dts/bcm283x.dtsi
@@ -488,14 +488,6 @@
 			status = "disabled";
 		};
 
-		vec: vec@7e806000 {
-			compatible = "brcm,bcm2835-vec";
-			reg = <0x7e806000 0x1000>;
-			clocks = <&clocks BCM2835_CLOCK_VEC>;
-			interrupts = <2 27>;
-			status = "disabled";
-		};
-
 		usb: usb@7e980000 {
 			compatible = "brcm,bcm2835-usb";
 			reg = <0x7e980000 0x10000>;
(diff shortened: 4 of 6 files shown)


popcornmix:
Thanks for this. Tested and composite on Pi4 with kms was not working prior to this PR and is working now.
I can confirm the updated VEC peripheral address is correct.

@pelwell is the DT change okay?

pelwell:
The DT change itself looks good - bonus marks for updating the YAML bindings - except that it needs to go upstream at some point.

6by9:
> except that it needs to go upstream at some point.

It does. @mripard should be able to do that for us though.

The commit descriptions being empty would be frowned upon upstream, likewise the subject line should start "drm/vc4" instead of "vc4/drm".
Otherwise it all looks pretty clean, although I'm not totally enamored by use of ```vc4_vec_get_variant``` every time we go through ```vc4_vec_encoder_enable```. Looking up the variant during ```bind``` and storing it in ```struct vc4_vec``` would be nicer.

kFYatek:
Just updated my branch.

> The commit descriptions being empty would be frowned upon upstream, likewise the subject line should start "drm/vc4" instead of "vc4/drm".

Sure, done.

> I'm not totally enamored by use of `vc4_vec_get_variant` every time we go through `vc4_vec_encoder_enable`. Looking up the variant during `bind` and storing it in `struct vc4_vec` would be nicer.

Well, I found that putting another field in `struct vc4_vec` would introduce unnecessary redundancy, which is my personal pet peeve ;) But sure, I'm not that familiar with the typical style used in the kernel. So I added a `variant` field in the revised version.

popcornmix:
Are we okay with this? Seems to make something work that currently does not.

eugene-s-nesdev:
Ntemis comment:
> Report this to raspberry pi kernel repo so we can fix this in the future

I confirm this problem on RPi3B+ with lastest "Lakka-RPi3.aarch64-nightly-20210724-58dafbe.img" 
- i've installed fresh lastest build and tested it on 1920x1080 display via HDMI - all works good.
- added this lines to config, then plug RPI to CRT-TV via 3.5mm out:
```
hdmi_force_hotplug=0
hdmi_ignore_hotplug=1
sdtv_mode=16
```
Signal is present (i see deep dark-gray backgroung with some noise) but there are no picture.

CRT-test update:
- Lakka 2.2.x works (2.2.x is removed from the server archive: http://le-builds.lakka.tv/ so you can't download it anymore)
- 2.3.x hangs at flower
- 3.x show blackscreen.

PR #4241, opened 2021-03-28
Fix composite output video norms

kFYatek:
This PR makes the Full-KMS video output over VEC use the same timings as the firmware (with one difference regarding 625-line modes, see below), fixes PAL modes, and adds support for additional TV standards.

Some remarks:
* I feel like my changes in `vc4_crtc.c` are somewhat messy, but I made them to align the PixelValve settings (the `HORZA`/`HORZB`/`VERTA`/`VERTB` registers in particular) perfectly with what the firmware does, while keeping the modeline truthful. For example, I needed that `vtotal_fixup` to make the total number of lines per field accurate (263/313 lines) while keeping `vtotal` in mode definition at 525/625 lines. I could set `vtotal` in mode definition to 526/626, but that's not the actual number of lines, plus it messes up the refresh rates in `xrandr` and such.
* In particular, I have very little idea what's going on with that `if (vec_interlaced) ++field_delay;` thing, but that makes the image cleaner. Compare: [before](https://user-images.githubusercontent.com/4499762/112738569-385c3280-8f64-11eb-83c4-d671537af209.png) vs. [after](https://user-images.githubusercontent.com/4499762/112738578-4f028980-8f64-11eb-8250-c726edcdc087.png). Tweaking this value indeed moves the fields relative to each other, but why `htotal / 2 + 1` in particular is the proper value rather than `htotal / 2`, I have no idea.
* I have moved horizontal sync pulse in the 625-line (PAL/SECAM) mode to 7 pixels earlier than the firmware or the original driver configures it. This is because in that mode, only about 704-706 pixels of the image are output anyway. This "masking" seems to be done by the VEC independently of PixelValve timings; it can be controlled in a very limited way through the `VEC_CONFIG3` register - setting the `HORIZ_LEN_MPEG1_SIF` makes it even narrower, while `SHAPE_NON_LINEAR` changes the gradient at the edges a little bit. Anyway, positioning the sync pulse 7 pixels earlier than the default centers the image relative to this mask. Compare: [before](https://user-images.githubusercontent.com/4499762/112738779-5e82d200-8f66-11eb-8fe1-96b7533acfd6.png) vs. [after](https://user-images.githubusercontent.com/4499762/112738782-66427680-8f66-11eb-8f8c-b3f93359352e.png).
* I'm not sure if the `vc4_vec_get_current_mode()` belongs in the source like I put it, but I wrote it so that the `sdtv_mode` setting in `config.txt` is respected to some extent by the Full-KMS driver as the default mode at boot time. If there is a better way to do it, please tell me.
* The `NTSC-443`, `PAL-M`, `PAL60` and `SECAM` modes are configured in the same way as in [TweakVec](https://github.com/kFYatek/tweakvec). The settings have been verified to work for other users in https://github.com/raspberrypi/firmware/issues/811
* Switching TV norm to one that's not compatible with the mode set at boot (e.g. entering PAL mode when `sdtv_mode=0`; see below for how to do that) causes the ability to switch virtual terminals (Ctrl+Alt+F1) to be lost, unless you switch the TV norm back first. This is how this driver has behaved before my changes. I don't know if this is how it _should_ behave - perhaps it should switch back to e.g. NTSC (or whatever the last 525-line norm was) when doing so? I wrote some (non-public for now) code that does just that, so I can post a supplementary PR or update this one if that's the requested behaviour.

I tested norm switching under X with the following one-liner:
```
xrandr --output Composite-1 --off && xrandr --output Composite-1 --set mode {NTSC|NTSC-J|NTSC-443|PAL|PAL-M|PAL-N|PAL60|SECAM} && xrandr --output Composite-1 --mode {720x480|720x576}
```
`720x480` works for NTSC, NTSC-J, NTSC-443, PAL-M and PAL60. `720x576` works for PAL, PAL-N and SECAM.

To switch to a mode incompatible with the one set at boot time, it has to be defined first using:
```
xrandr --newmode 720x480 13.5 720 734 798 858 480 487 493 525 Interlace
```
or
```
xrandr --newmode 720x576 13.5 720 733 797 864 576 580 586 625 Interlace
```
Modelines that are not exactly one of the two above, are intentionally rejected, at least for now, and this behaviour is carried over from the existing code. I will probably post another PR later that relaxes this requirement a bit, and adds support for progressive (240p/288p) modes.

### DIFF
--- a/drivers/gpu/drm/vc4/vc4_crtc.c
+++ b/drivers/gpu/drm/vc4/vc4_crtc.c
@@ -316,8 +316,14 @@ static void vc4_crtc_config_pv(struct drm_crtc *crtc, struct drm_encoder *encode
 	bool is_dsi = (vc4_encoder->type == VC4_ENCODER_TYPE_DSI0 ||
 		       vc4_encoder->type == VC4_ENCODER_TYPE_DSI1);
 	bool is_dsi1 = vc4_encoder->type == VC4_ENCODER_TYPE_DSI1;
+	bool is_vec = vc4_encoder->type == VC4_ENCODER_TYPE_VEC;
 	u32 format = is_dsi1 ? PV_CONTROL_FORMAT_DSIV_24 : PV_CONTROL_FORMAT_24;
 	u8 ppc = pv_data->pixels_per_clock;
+
+	u16 vert_bp = mode->crtc_vtotal - mode->crtc_vsync_end;
+	u16 vert_sync = mode->crtc_vsync_end - mode->crtc_vsync_start;
+	u16 vert_fp = mode->crtc_vsync_start - mode->crtc_vdisplay;
+
 	bool debug_dump_regs = false;
 
 	if (debug_dump_regs) {
@@ -341,48 +347,58 @@ static void vc4_crtc_config_pv(struct drm_crtc *crtc, struct drm_encoder *encode
 		   VC4_SET_FIELD(mode->hdisplay * pixel_rep / ppc,
 				 PV_HORZB_HACTIVE));
 
-	CRTC_WRITE(PV_VERTA,
-		   VC4_SET_FIELD(mode->crtc_vtotal - mode->crtc_vsync_end +
-				 interlace,
-				 PV_VERTA_VBP) |
-		   VC4_SET_FIELD(mode->crtc_vsync_end - mode->crtc_vsync_start,
-				 PV_VERTA_VSYNC));
-	CRTC_WRITE(PV_VERTB,
-		   VC4_SET_FIELD(mode->crtc_vsync_start - mode->crtc_vdisplay,
-				 PV_VERTB_VFP) |
-		   VC4_SET_FIELD(mode->crtc_vdisplay, PV_VERTB_VACTIVE));
-
 	if (interlace) {
+		bool odd_field_first = false;
+		u32 field_delay = mode->htotal * pixel_rep / (2 * ppc);
+		u16 vert_bp_even = vert_bp;
+		u16 vert_fp_even = vert_fp;
+
+		if (is_vec) {
+			/* VEC (composite output) */
+			++field_delay;
+			if (mode->htotal == 858) {
+				/* 525-line mode (NTSC or PAL-M) */
+				odd_field_first = true;
+			}
+		}
+
+		if (odd_field_first)
+			++vert_fp_even;
+		else
+			++vert_bp;
+
 		CRTC_WRITE(PV_VERTA_EVEN,
-			   VC4_SET_FIELD(mode->crtc_vtotal -
-					 mode->crtc_vsync_end,
-					 PV_VERTA_VBP) |
-			   VC4_SET_FIELD(mode->crtc_vsync_end -
-					 mode->crtc_vsync_start,
-					 PV_VERTA_VSYNC));
+			   VC4_SET_FIELD(vert_bp_even, PV_VERTA_VBP) |
+			   VC4_SET_FIELD(vert_sync, PV_VERTA_VSYNC));
 		CRTC_WRITE(PV_VERTB_EVEN,
-			   VC4_SET_FIELD(mode->crtc_vsync_start -
-					 mode->crtc_vdisplay,
-					 PV_VERTB_VFP) |
+			   VC4_SET_FIELD(vert_fp_even, PV_VERTB_VFP) |
 			   VC4_SET_FIELD(mode->crtc_vdisplay, PV_VERTB_VACTIVE));
 
-		/* We set up first field even mode for HDMI.  VEC's
-		 * NTSC mode would want first field odd instead, once
-		 * we support it (to do so, set ODD_FIRST and put the
-		 * delay in VSYNCD_EVEN instead).
+		/* We set up first field even mode for HDMI and VEC's PAL.
+		 * For NTSC, we need first field odd.
 		 */
 		CRTC_WRITE(PV_V_CONTROL,
 			   PV_VCONTROL_CONTINUOUS |
 			   (is_dsi ? PV_VCONTROL_DSI : 0) |
 			   PV_VCONTROL_INTERLACE |
-			   VC4_SET_FIELD(mode->htotal * pixel_rep / (2 * ppc),
-					 PV_VCONTROL_ODD_DELAY));
-		CRTC_WRITE(PV_VSYNCD_EVEN, 0);
+			   (odd_field_first
+				   ? PV_VCONTROL_ODD_FIRST
+				   : VC4_SET_FIELD(field_delay,
+						   PV_VCONTROL_ODD_DELAY)));
+		CRTC_WRITE(PV_VSYNCD_EVEN,
+			   (odd_field_first ? field_delay : 0));
 	} else {
 		CRTC_WRITE(PV_V_CONTROL,
 			   PV_VCONTROL_CONTINUOUS |
 			   (is_dsi ? PV_VCONTROL_DSI : 0));
 	}
+
+	CRTC_WRITE(PV_VERTA,
+		   VC4_SET_FIELD(vert_bp, PV_VERTA_VBP) |
+		   VC4_SET_FIELD(vert_sync, PV_VERTA_VSYNC));
+	CRTC_WRITE(PV_VERTB,
+		   VC4_SET_FIELD(vert_fp, PV_VERTB_VFP) |
+		   VC4_SET_FIELD(mode->crtc_vdisplay, PV_VERTB_VACTIVE));
 
 	if (is_dsi)
 		CRTC_WRITE(PV_HACT_ACT, mode->hdisplay * pixel_rep);
--- a/drivers/gpu/drm/vc4/vc4_vec.c
+++ b/drivers/gpu/drm/vc4/vc4_vec.c
@@ -45,6 +45,7 @@
 #define VEC_CONFIG0_YDEL(x)		((x) << 26)
 #define VEC_CONFIG0_CDEL_MASK		GENMASK(25, 24)
 #define VEC_CONFIG0_CDEL(x)		((x) << 24)
+#define VEC_CONFIG0_SECAM_STD		BIT(21)
 #define VEC_CONFIG0_PBPR_FIL		BIT(18)
 #define VEC_CONFIG0_CHROMA_GAIN_MASK	GENMASK(17, 16)
 #define VEC_CONFIG0_CHROMA_GAIN_UNITY	(0 << 16)
@@ -75,6 +76,27 @@
 #define VEC_SOFT_RESET			0x10c
 #define VEC_CLMP0_START			0x144
 #define VEC_CLMP0_END			0x148
+
+/*
+ * These set the color subcarrier frequency
+ * if VEC_CONFIG1_CUSTOM_FREQ is enabled.
+ *
+ * VEC_FREQ1_0 contains the most significant 16-bit half-word,
+ * VEC_FREQ3_2 contains the least significant 16-bit half-word.
+ * 0x80000000 seems to be equivalent to the pixel clock
+ * (which itself is the VEC clock divided by 8).
+ *
+ * Reference values (with the default pixel clock of 13.5 MHz):
+ *
+ * NTSC  (3579545.[45] Hz)     - 0x21F07C1F
+ * PAL   (4433618.75 Hz)       - 0x2A098ACB
+ * PAL-M (3575611.[888111] Hz) - 0x21E6EFE3
+ * PAL-N (3582056.25 Hz)       - 0x21F69446
+ *
+ * NOTE: For SECAM, it is used as the Dr center frequency,
+ * regardless of whether VEC_CONFIG1_CUSTOM_FREQ is enabled or not;
+ * that is specified as 4406250 Hz, which corresponds to 0x29C71C72.
+ */
 #define VEC_FREQ3_2			0x180
 #define VEC_FREQ1_0			0x184
 
@@ -117,6 +139,14 @@
 
 #define VEC_INTERRUPT_CONTROL		0x190
 #define VEC_INTERRUPT_STATUS		0x194
+
+/*
+ * Db center frequency for SECAM; the clock for this is the same as for
+ * VEC_FREQ3_2/VEC_FREQ1_0, which is used for Dr center frequency.
+ *
+ * This is specified as 4250000 Hz, which corresponds to 0x284BDA13.
+ * That is also the default value, so no need to set it explicitly.
+ */
 #define VEC_FCW_SECAM_B			0x198
 #define VEC_SECAM_GAIN_VAL		0x19c
 
@@ -210,8 +240,12 @@ to_vc4_vec_connector(struct drm_connector *connector)
 enum vc4_vec_tv_mode_id {
 	VC4_VEC_TV_MODE_NTSC,
 	VC4_VEC_TV_MODE_NTSC_J,
+	VC4_VEC_TV_MODE_NTSC_443,
 	VC4_VEC_TV_MODE_PAL,
 	VC4_VEC_TV_MODE_PAL_M,
+	VC4_VEC_TV_MODE_PAL_N,
+	VC4_VEC_TV_MODE_PAL60,
+	VC4_VEC_TV_MODE_SECAM,
 };
 
 struct vc4_vec_tv_mode {
@@ -273,6 +307,13 @@ static const struct vc4_vec_tv_mode vc4_vec_tv_modes[] = {
 		.config0 = VEC_CONFIG0_NTSC_STD,
 		.config1 = VEC_CONFIG1_C_CVBS_CVBS,
 	},
+	[VC4_VEC_TV_MODE_NTSC_443] = {
+		/* NTSC with PAL chroma frequency */
+		.mode = &drm_mode_480i,
+		.config0 = VEC_CONFIG0_NTSC_STD,
+		.config1 = VEC_CONFIG1_C_CVBS_CVBS | VEC_CONFIG1_CUSTOM_FREQ,
+		.custom_freq = 0x2a098acb,
+	},
 	[VC4_VEC_TV_MODE_PAL] = {
 		.mode = &drm_mode_576i,
 		.config0 = VEC_CONFIG0_PAL_BDGHI_STD,
@@ -282,6 +323,24 @@ static const struct vc4_vec_tv_mode vc4_vec_tv_modes[] = {
 		.mode = &drm_mode_480i,
 		.config0 = VEC_CONFIG0_PAL_M_STD,
 		.config1 = VEC_CONFIG1_C_CVBS_CVBS,
+	},
+	[VC4_VEC_TV_MODE_PAL_N] = {
+		.mode = &drm_mode_576i,
+		.config0 = VEC_CONFIG0_PAL_N_STD,
+		.config1 = VEC_CONFIG1_C_CVBS_CVBS,
+	},
+	[VC4_VEC_TV_MODE_PAL60] = {
+		/* PAL-M with chroma frequency of regular PAL */
+		.mode = &drm_mode_480i,
+		.config0 = VEC_CONFIG0_PAL_M_STD,
+		.config1 = VEC_CONFIG1_C_CVBS_CVBS | VEC_CONFIG1_CUSTOM_FREQ,
+		.custom_freq = 0x2a098acb,
+	},
+	[VC4_VEC_TV_MODE_SECAM] = {
+		.mode = &drm_mode_576i,
+		.config0 = VEC_CONFIG0_SECAM_STD,
+		.config1 = VEC_CONFIG1_C_CVBS_CVBS,
+		.custom_freq = 0x29c71c72,
 	},
 };
 
@@ -505,8 +564,12 @@ static const struct of_device_id vc4_vec_dt_match[] = {
 static const char * const tv_mode_names[] = {
 	[VC4_VEC_TV_MODE_NTSC] = "NTSC",
 	[VC4_VEC_TV_MODE_NTSC_J] = "NTSC-J",
+	[VC4_VEC_TV_MODE_NTSC_443] = "NTSC-443",
 	[VC4_VEC_TV_MODE_PAL] = "PAL",
 	[VC4_VEC_TV_MODE_PAL_M] = "PAL-M",
+	[VC4_VEC_TV_MODE_PAL_N] = "PAL-N",
+	[VC4_VEC_TV_MODE_PAL60] = "PAL60",
+	[VC4_VEC_TV_MODE_SECAM] = "SECAM",
 };
 
 static int vc4_vec_bind(struct device *dev, struct device *master, void *data)
--- a/drivers/gpu/drm/vc4/vc4_vec.c
+++ b/drivers/gpu/drm/vc4/vc4_vec.c
@@ -421,6 +421,31 @@ static void vc4_vec_connector_reset(struct drm_connector *connector)
 		connector->state->tv.mode = vc4_vec_get_default_mode(connector);
 }
 
+static int vc4_vec_connector_atomic_check(struct drm_connector *conn,
+					  struct drm_atomic_state *state)
+{
+	struct drm_connector_state *old_state =
+		drm_atomic_get_old_connector_state(state, conn);
+	struct drm_connector_state *new_state =
+		drm_atomic_get_new_connector_state(state, conn);
+
+	const struct vc4_vec_tv_mode *vec_mode =
+		&vc4_vec_tv_modes[new_state->tv.mode];
+
+	if (new_state->crtc) {
+		struct drm_crtc_state *crtc_state =
+			drm_atomic_get_new_crtc_state(state, new_state->crtc);
+
+		if (!drm_mode_equal(vec_mode->mode, &crtc_state->mode))
+			return -EINVAL;
+
+		if (old_state->tv.mode != new_state->tv.mode)
+			crtc_state->mode_changed = true;
+	}
+
+	return 0;
+}
+
 static const struct drm_connector_funcs vc4_vec_connector_funcs = {
 	.detect = vc4_vec_connector_detect,
 	.fill_modes = drm_helper_probe_single_connector_modes,
@@ -432,6 +457,7 @@ static const struct drm_connector_funcs vc4_vec_connector_funcs = {
 
 static const struct drm_connector_helper_funcs vc4_vec_connector_helper_funcs = {
 	.get_modes = vc4_vec_connector_get_modes,
+	.atomic_check = vc4_vec_connector_atomic_check,
 };
 
 static struct drm_connector *vc4_vec_connector_init(struct drm_device *dev,
@@ -569,26 +595,10 @@ static bool vc4_vec_encoder_mode_fixup(struct drm_encoder *encoder,
 	return true;
 }
 
-static int vc4_vec_encoder_atomic_check(struct drm_encoder *encoder,
-					struct drm_crtc_state *crtc_state,
-					struct drm_connector_state *conn_state)
-{
-	const struct vc4_vec_tv_mode *vec_mode;
-
-	vec_mode = &vc4_vec_tv_modes[conn_state->tv.mode];
-
-	if (conn_state->crtc &&
-	    !drm_mode_equal(vec_mode->mode, &crtc_state->adjusted_mode))
-		return -EINVAL;
-
-	return 0;
-}
-
 static const struct drm_encoder_helper_funcs vc4_vec_encoder_helper_funcs = {
 	.disable = vc4_vec_encoder_disable,
 	.enable = vc4_vec_encoder_enable,
 	.mode_fixup = vc4_vec_encoder_mode_fixup,
-	.atomic_check = vc4_vec_encoder_atomic_check,
 };
 
 static const struct vc4_vec_variant bcm2835_vec_variant = {
--- a/drivers/gpu/drm/vc4/vc4_vec.c
+++ b/drivers/gpu/drm/vc4/vc4_vec.c
@@ -262,7 +262,7 @@ static void vc4_vec_ntsc_j_mode_set(struct vc4_vec *vec)
 static const struct drm_display_mode ntsc_mode = {
 	DRM_MODE("720x480", DRM_MODE_TYPE_DRIVER, 13500,
 		 720, 720 + 14, 720 + 14 + 64, 720 + 14 + 64 + 60, 0,
-		 480, 480 + 3, 480 + 3 + 3, 480 + 3 + 3 + 16, 0,
+		 480, 480 + 7, 480 + 7 + 6, 525, 0,
 		 DRM_MODE_FLAG_INTERLACE)
 };
 
@@ -284,7 +284,7 @@ static void vc4_vec_pal_m_mode_set(struct vc4_vec *vec)
 static const struct drm_display_mode pal_mode = {
 	DRM_MODE("720x576", DRM_MODE_TYPE_DRIVER, 13500,
 		 720, 720 + 20, 720 + 20 + 64, 720 + 20 + 64 + 60, 0,
-		 576, 576 + 2, 576 + 2 + 3, 576 + 2 + 3 + 20, 0,
+		 576, 576 + 4, 576 + 4 + 6, 625, 0,
 		 DRM_MODE_FLAG_INTERLACE)
 };
 
--- a/drivers/gpu/drm/vc4/vc4_vec.c
+++ b/drivers/gpu/drm/vc4/vc4_vec.c
@@ -68,6 +68,7 @@
 #define VEC_CONFIG0_STD_MASK		GENMASK(1, 0)
 #define VEC_CONFIG0_NTSC_STD		0
 #define VEC_CONFIG0_PAL_BDGHI_STD	1
+#define VEC_CONFIG0_PAL_M_STD		2
 #define VEC_CONFIG0_PAL_N_STD		3
 
 #define VEC_SCHPH			0x108
@@ -247,14 +248,14 @@ static const struct debugfs_reg32 vec_regs[] = {
 	VC4_REG32(VEC_DAC_MISC),
 };
 
-static const struct drm_display_mode ntsc_mode = {
+static const struct drm_display_mode drm_mode_480i = {
 	DRM_MODE("720x480", DRM_MODE_TYPE_DRIVER, 13500,
 		 720, 720 + 14, 720 + 14 + 64, 720 + 14 + 64 + 60, 0,
 		 480, 480 + 7, 480 + 7 + 6, 525, 0,
 		 DRM_MODE_FLAG_INTERLACE)
 };
 
-static const struct drm_display_mode pal_mode = {
+static const struct drm_display_mode drm_mode_576i = {
 	DRM_MODE("720x576", DRM_MODE_TYPE_DRIVER, 13500,
 		 720, 720 + 20, 720 + 20 + 64, 720 + 20 + 64 + 60, 0,
 		 576, 576 + 4, 576 + 4 + 6, 625, 0,
@@ -263,25 +264,24 @@ static const struct drm_display_mode pal_mode = {
 
 static const struct vc4_vec_tv_mode vc4_vec_tv_modes[] = {
 	[VC4_VEC_TV_MODE_NTSC] = {
-		.mode = &ntsc_mode,
+		.mode = &drm_mode_480i,
 		.config0 = VEC_CONFIG0_NTSC_STD | VEC_CONFIG0_PDEN,
 		.config1 = VEC_CONFIG1_C_CVBS_CVBS,
 	},
 	[VC4_VEC_TV_MODE_NTSC_J] = {
-		.mode = &ntsc_mode,
+		.mode = &drm_mode_480i,
 		.config0 = VEC_CONFIG0_NTSC_STD,
 		.config1 = VEC_CONFIG1_C_CVBS_CVBS,
 	},
 	[VC4_VEC_TV_MODE_PAL] = {
-		.mode = &pal_mode,
+		.mode = &drm_mode_576i,
 		.config0 = VEC_CONFIG0_PAL_BDGHI_STD,
 		.config1 = VEC_CONFIG1_C_CVBS_CVBS,
 	},
 	[VC4_VEC_TV_MODE_PAL_M] = {
-		.mode = &pal_mode,
-		.config0 = VEC_CONFIG0_PAL_BDGHI_STD,
-		.config1 = VEC_CONFIG1_C_CVBS_CVBS | VEC_CONFIG1_CUSTOM_FREQ,
-		.custom_freq = 0x223b61d1,
+		.mode = &drm_mode_480i,
+		.config0 = VEC_CONFIG0_PAL_M_STD,
+		.config1 = VEC_CONFIG1_C_CVBS_CVBS,
 	},
 };
 
--- a/drivers/gpu/drm/vc4/vc4_vec.c
+++ b/drivers/gpu/drm/vc4/vc4_vec.c
@@ -66,7 +66,7 @@
 #define VEC_CONFIG0_YCDELAY		BIT(4)
 #define VEC_CONFIG0_RAMPEN		BIT(2)
 #define VEC_CONFIG0_YCDIS		BIT(2)
-#define VEC_CONFIG0_STD_MASK		GENMASK(1, 0)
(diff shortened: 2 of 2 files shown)


6by9:
Thanks for the patches. I need to spend a little time looking into them and testing.

There will be a request for all the commit texts to be updated as they should start ```drm/vc4:```, reference the VEC in the subject line, and have some more detail in the description.

It's also a little confusing that you add all these *_mode_set functions for the new standards, then rip them all out again and move the config that they add into struct vc4_vec_tv_mode. It would be nicer to make that move first, and then add the extra modes.

With regard ```vc4_vec_get_current_mode```, the firmware should insert a ```video=Composite-1:...``` line into the kernel command line, so that should give you the right mode. The parsing of that isn't perfect though.

kFYatek:
NOTE: There are no changes in the force-push above yet. I just rebased everything onto the current main branch. I will now proceed with addressing your comments.

kFYatek:
> With regard `vc4_vec_get_current_mode`, the firmware should insert a `video=Composite-1:...` line into the kernel command line, so that should give you the right mode. The parsing of that isn't perfect though.

Well, that line is just e.g. `video=Composite-1:720x576@50i`. It's not enough to differentiate between PAL and PAL-N (and SECAM, if that's ever added to firmware), or NTSC (US) vs. NTSC-J vs. PAL-M.

kFYatek:
NOTE: As far as code is concerned, the only changes compared to the previous version are slight wording improvements in some comments. No actual code has been changed, this is all about reorganizing the commits.

> There will be a request for all the commit texts to be updated as they should start `drm/vc4:`, reference the VEC in the subject line, and have some more detail in the description.

Sure, done.

> It's also a little confusing that you add all these *_mode_set functions for the new standards, then rip them all out again and move the config that they add into struct vc4_vec_tv_mode. It would be nicer to make that move first, and then add the extra modes.

Sure, done as well. The original PR contained the commits "raw", mostly how I actually developed these changes. I don't know much about the etiquette of the kernel repo, so sorry about that. I now rewrote the history so that the commits make more sense on their own, and are more concise as well.

drivers/gpu/drm/vc4/vc4_crtc.c
```diff
@@ -377,31 +381,38 @@ static void vc4_crtc_config_pv(struct drm_crtc *crtc)
...
+		u32 field_delay = mode->htotal * pixel_rep / 2;
+		if (vec_interlaced)
+			++field_delay;
+
 		CRTC_WRITE(PV_VERTA_EVEN,
-			   VC4_SET_FIELD(mode->crtc_vtotal -
-					 mode->crtc_vsync_end - 1,
+			   VC4_SET_FIELD(mode->crtc_vtotal + vtotal_fixup -
+					 mode->crtc_vsync_end -
+					 (ntsc ? 0 : 1),
```
6by9: This one feels odd.

```vtotal_fixup = (vec_interlaced && !ntsc) ? 1 : 0```
but here we're doing ```- (ntsc ? 0 : 1)```, and we're in a condition where we're interlaced, so effectively ```(interlaced && ntsc) ? 0 : 1).
Doesn't that end up with the two operations cancelling each other out?

**Review by mripard — COMMENTED** [review]
I'm not sure we need to read-out the hardware state. The command line parser explicitly supports named modes, so the firmware could just set video=TV-1:NTSC-J or whatever and the initial framebuffer will use that mode.

Switching back and forth between modes is more problematic though. The userspace is free to build whatever mode it wants, and / or the framebuffer emulation (and thus the initial setup) can be disabled.

In such a case, you don't really have a way to differentiate between variants that use the same mode (like between NSTC-J and PAL-M). Once again, you can only hope that the name will be filled and compare the name to the modes you support, and fallback on PAL or NTSC

File: drivers/gpu/drm/vc4/vc4_crtc.c [hunk 2]
```diff
@@ -343,6 +343,10 @@ static void vc4_crtc_config_pv(struct drm_crtc *crtc)
 	bool is_dsi1 = vc4_encoder->type == VC4_ENCODER_TYPE_DSI1;
 	u32 format = is_dsi1 ? PV_CONTROL_FORMAT_DSIV_24 : PV_CONTROL_FORMAT_24;
 	u8 ppc = pv_data->pixels_per_clock;
+	bool vec_interlaced = (vc4_encoder->type == VC4_ENCODER_TYPE_VEC &&
+			       interlace);
+	bool ntsc = (vec_interlaced && mode->htotal == 858);
```
6by9: This way of detecting NTSC does feel a little fragile, but I'm not sure that there is a better solution.

drivers/gpu/drm/vc4/vc4_vec.c
```diff
@@ -283,8 +283,8 @@ static void vc4_vec_pal_m_mode_set(struct vc4_vec *vec)
 
 static const struct drm_display_mode pal_mode = {
 	DRM_MODE("720x576", DRM_MODE_TYPE_DRIVER, 13500,
-		 720, 720 + 20, 720 + 20 + 64, 720 + 20 + 64 + 60, 0,
-		 576, 576 + 2, 576 + 2 + 3, 576 + 2 + 3 + 20, 0,
+		 720, 720 + 13, 720 + 13 + 64, 720 + 13 + 64 + 67, 0,
```
6by9: Has this been tested on a Pi4? Some of the PVs on 2711 operate at 2 pixels/clock, which means that odd values in horizontal timings cause issues as it can't signal the sync event between the two pixels.

I'm not convinced by the centering anyway as it will cause different behaviour compared to the firmware, and it also should be a separate patch to that correcting the timings.

drivers/gpu/drm/vc4/vc4_crtc.c
```diff
@@ -377,31 +381,38 @@ static void vc4_crtc_config_pv(struct drm_crtc *crtc)
...
 					 PV_VERTA_VSYNC));
 		CRTC_WRITE(PV_VERTB_EVEN,
 			   VC4_SET_FIELD(mode->crtc_vsync_start -
-					 mode->crtc_vdisplay,
+					 mode->crtc_vdisplay +
+					 (ntsc ? 1 : 0),
 					 PV_VERTB_VFP) |
-			   VC4_SET_FIELD(mode->crtc_vdisplay, PV_VERTB_VACTIVE));
+			   VC4_SET_FIELD(mode->crtc_vdisplay,
+					 PV_VERTB_VACTIVE));
```
6by9: All of these fixups become quite hard to read and understand what exactly is going on.

My gut feeling is that it'd be clearer to add local variables at least for the vertical active, fp, sw, and bp values, and then have explicit clauses for NTSC and PAL when assigning to the registers. At least that way the fixups are obvious.

6by9 (commented):
Sorry it's taken me a while to get around to looking at this.

Generally looks fine, but the PV setup changes in vc4_crtc.c have me a little concerned.

Seeing as I don't believe it holds any secrets, the firmware config code for the PV when driving the VEC is:
```
   if (use_625_line)
   {
       vec_state.pixelvalve_setup.horizontal_sync_width_in_pixels = 64;
       vec_state.pixelvalve_setup.horizontal_frontporch_in_pixels = 20;
       vec_state.pixelvalve_setup.horizontal_backporch_in_pixels  = 60;
       vec_state.pixelvalve_setup.horizontal_active_in_pixels     = 720;
       vec_state.pixelvalve_setup.vertical_sync_width_in_lines    = 3;
       vec_state.pixelvalve_setup.vertical_frontporch_in_lines    = 2;
       vec_state.pixelvalve_setup.vertical_backporch_in_lines     = use_progressive ? 19 : 20;
       vec_state.pixelvalve_setup.vertical_active_in_lines        = 576/2;
       vec_state.pixelvalve_setup.vertical_even_sync_width_in_lines = vec_state.pixelvalve_setup.vertical_sync_width_in_lines;
       vec_state.pixelvalve_setup.vertical_even_frontporch_in_lines = vec_state.pixelvalve_setup.vertical_frontporch_in_lines;
       vec_state.pixelvalve_setup.vertical_even_backporch_in_lines  = vec_state.pixelvalve_setup.vertical_backporch_in_lines - 1;
       vec_state.pixelvalve_setup.vertical_even_active_in_lines     = vec_state.pixelvalve_setup.vertical_active_in_lines;
       vec_state.pixelvalve_setup.first_field_odd = 0;
   }
   else
   {
       vec_state.pixelvalve_setup.horizontal_sync_width_in_pixels = 64;
       vec_state.pixelvalve_setup.horizontal_frontporch_in_pixels = 14;
       vec_state.pixelvalve_setup.horizontal_backporch_in_pixels  = 60;
       vec_state.pixelvalve_setup.horizontal_active_in_pixels     = 720;
       vec_state.pixelvalve_setup.vertical_sync_width_in_lines    = 3;
       vec_state.pixelvalve_setup.vertical_frontporch_in_lines    = 3;
       vec_state.pixelvalve_setup.vertical_backporch_in_lines     = 16;
       vec_state.pixelvalve_setup.vertical_active_in_lines        = 480/2;
       vec_state.pixelvalve_setup.vertical_even_sync_width_in_lines = vec_state.pixelvalve_setup.vertical_sync_width_in_lines;
       vec_state.pixelvalve_setup.vertical_even_frontporch_in_lines = vec_state.pixelvalve_setup.vertical_frontporch_in_lines + 1;
       vec_state.pixelvalve_setup.vertical_even_backporch_in_lines  = vec_state.pixelvalve_setup.vertical_backporch_in_lines;
       vec_state.pixelvalve_setup.vertical_even_active_in_lines     = vec_state.pixelvalve_setup.vertical_active_in_lines;
       vec_state.pixelvalve_setup.first_field_odd = 1;
   }
   uint32_t half_line = (vec_state.pixelvalve_setup.horizontal_sync_width_in_pixels +
                         vec_state.pixelvalve_setup.horizontal_frontporch_in_pixels +
                         vec_state.pixelvalve_setup.horizontal_backporch_in_pixels +
                         vec_state.pixelvalve_setup.horizontal_active_in_pixels) >> 1;
   vec_state.pixelvalve_setup.second_field_offset = half_line + 1;
   if (pal_mode)
      vec_state.pixelvalve_setup.mode = PIXELVALVE_MODE_VEC_PAL;
   else
      vec_state.pixelvalve_setup.mode = PIXELVALVE_MODE_VEC_NTSC;
```
The values are written without modification into the relevant PV registers.

Note that all the vertical timings have already been divided by 2 due to interlacing. That happens in the magic of ```drm_mode_set_crtcinfo``` for DRM, but that does have the potential for rounding errors on odd values.

drivers/gpu/drm/vc4/vc4_crtc.c
kFYatek: Not exactly.
* Progressive or HDMI interlace: `vtotal_fixup == 0`, `ntsc == 0`, total fixup here: `-1`
* VEC 525-line ("NTSC"): `vtotal_fixup == 0`, `ntsc == 1`, total fixup here: `0`
* VEC 625-line ("PAL"): `vtotal_fixup == 1`, `ntsc == 0`, total fixup here: `0`

I guess this `vtotal_fixup` may be confusing after all, and using `vec_interlaced` and `ntsc` directly at those two places these are used will be cleaner.

drivers/gpu/drm/vc4/vc4_crtc.c
kFYatek: I now realize that this may even be confusing, as this actually detects 525/625-line modes, not NTSC per se (i.e. PAL-M is "NTSC" in this sense and that's intended), so renaming these variables is probably the right thing to do.

drivers/gpu/drm/vc4/vc4_vec.c
kFYatek: Fair enough, keeping the timing consistent with the firmware is a fair argument.

But odd values for front and back porch lengths do indeed work on Pi4.

drivers/gpu/drm/vc4/vc4_crtc.c
kFYatek: OK, I'll try to simplify things a bit.

kFYatek:
> 
> 
> I'm not sure we need to read-out the hardware state. The command line parser explicitly supports named modes, so the firmware could just set video=TV-1:NTSC-J or whatever and the initial framebuffer will use that mode.
> 
> Switching back and forth between modes is more problematic though. The userspace is free to build whatever mode it wants, and / or the framebuffer emulation (and thus the initial setup) can be disabled.
> 
> In such a case, you don't really have a way to differentiate between variants that use the same mode (like between NSTC-J and PAL-M). Once again, you can only hope that the name will be filled and compare the name to the modes you support, and fallback on PAL or NTSC

Well, unfortunately the current versions of the firmware only pass information such as `video=Composite-1:720x480@60i` or `video=Composite-1:720x576@50i`. There is no way to differentiate between NTSC vs. NTSC-J vs. PAL-M or PAL vs. PAL-N, even though all of these are accessible via `sdtv_mode`.

I agree that what I wrote is a bit of a kludge, and I agree that if the firmware could pass the colour system settings on the kernel's command line, that would be much better. I'm happy to update my code once it does, but for now, reading the VEC register state found at boot is the only way to get this information.

kFYatek:
Note; the force-push above is just a rebase, with no changes to the actual code. I'll amend the diff with the next update.

6by9:
> Well, unfortunately the current versions of the firmware only pass information such as video=Composite-1:720x480@60i or video=Composite-1:720x576@50i. There is no way to differentiate between NTSC vs. NTSC-J vs. PAL-M or PAL vs. PAL-N, even though all of these are accessible via sdtv_mode.
> 
> I agree that what I wrote is a bit of a kludge, and I agree that if the firmware could pass the colour system settings on the kernel's command line, that would be much better. I'm happy to update my code once it does, but for now, reading the VEC register state found at boot is the only way to get this information.

It looks like the firmware has a struct [```SDTV_DISPLAY_STATE_T```](https://github.com/raspberrypi/userland/blob/master/interface/vmcs_host/vc_sdtv.h#L111) which includes the mode from ```enum SDTV_MODE_T```. That should be fairly simple to map into an appropriate string on the kernel command line. 

I'm currently failing to find the kernel side of the parsing for ```video=TV-1:NTSC-J```, nor documentation for that option. Most of the docs are in https://github.com/torvalds/linux/blob/master/Documentation/fb/modedb.rst, but not this option.
[drm_mode_parse_command_line_for_connector](https://github.com/torvalds/linux/blob/master/drivers/gpu/drm/drm_modes.c#L1714) would appear to have a mode whitelist for NTSC and PAL, but not the variants.

@mripard Are you believing that the parsing for this is already present and will set the tv_mode_property, or is this a further change that is needed?

kFYatek:
Note: the update from 7 minutes ago updates the first commit in this PR according to the review notes.

The second update from just now contains no code changes, it just updates the commit message of that commit to remove information about the shift by 7 pixels, which has been removed. I forgot to update the commit message the first time.

kFYatek:
@6by9 I tried to check how other drivers handle this issue and it seems that there are exactly two that support composite output and support setting its mode at boot - and both [ch7006](https://github.com/torvalds/linux/blob/a96bfed64c8986d6404e553f18203cae1f5ac7e6/drivers/gpu/drm/i2c/ch7006_drv.c#L532) and [nouveau](https://github.com/torvalds/linux/blob/a96bfed64c8986d6404e553f18203cae1f5ac7e6/drivers/gpu/drm/nouveau/dispnv04/tvnv17.c#L43) introduce a module parameter called `tv_norm`. Maybe we should follow suit?

kFYatek:
I updated the PR with the new logic. The new `vc4.tv_norm=` kernel command line parameter is added, so that one can add e.g. `vc4.tv_norm=NTSC-J` or `vc4.tv_norm=SECAM`. Maybe in the future that could be set by the firmware, but for now there's always `/boot/cmdline.txt`.

I added additional logic so if that option is absent, but the `video=` modeline looks like a PAL mode, PAL is used. This at least makes `sdtv_mode=2` work.

One important note is that if the `video=` argument does not match `vc4.tv_norm`, the fbcon terminal will be inaccessible. This may happen if e.g. `sdtv_mode` is set to `0`, but `vc4.tv_norm=PAL` is also specified on the command line. fbcon always prefers the cmdline mode, but that mode does not pass validation in `vc4_vec_encoder_atomic_check()`, and the result is that you can't do Ctrl+Alt+F1 anymore. X works. I don't know if we should address this issue in any way.

kFYatek:
The above force-push rebased this PR onto current `origin/rpi-5.10.y`. 

The conflict resolution was ("old" is the previously pushed revision, "new" is the current one):
```diff
--- a/drivers/gpu/drm/vc4/vc4_crtc.c
+++ b/drivers/gpu/drm/vc4/vc4_crtc.c
@@ -362,7 +349,7 @@ static void vc4_crtc_config_pv(struct drm_crtc *crtc, struct drm_atomic_state *s
 
        if (interlace) {
                bool odd_field_first = false;
-               u32 field_delay = mode->htotal * pixel_rep / 2;
+               u32 field_delay = mode->htotal * pixel_rep / (2 * ppc);
                u16 vert_bp_even = vert_bp;
                u16 vert_fp_even = vert_fp;
 
@@ -372,16 +359,14 @@ static void vc4_crtc_config_pv(struct drm_crtc *crtc, struct drm_atomic_state *s
                        if (mode->htotal == 858) {
                                /* 525-line mode (NTSC or PAL-M) */
                                odd_field_first = true;
-                               ++vert_fp_even;
-                       } else {
-                               /* 625-line mode (PAL or SECAM) */
-                               ++vert_bp;
                        }
-               } else {
-                       /* HDMI */
-                       --vert_bp_even;
                }
 
+               if (odd_field_first)
+                       ++vert_fp_even;
+               else
+                       ++vert_bp;
+
                CRTC_WRITE(PV_VERTA_EVEN,
                           VC4_SET_FIELD(vert_bp_even, PV_VERTA_VBP) |
                           VC4_SET_FIELD(vert_sync, PV_VERTA_VSYNC));
```

drivers/gpu/drm/vc4/vc4_vec.c
```diff
@@ -343,6 +345,41 @@ static const struct vc4_vec_tv_mode vc4_vec_tv_modes[] = {
...
+	[VC4_VEC_TV_MODE_NTSC_443] = "NTSC-443",
+	[VC4_VEC_TV_MODE_PAL] = "PAL",
+	[VC4_VEC_TV_MODE_PAL_M] = "PAL-M",
+	[VC4_VEC_TV_MODE_PAL_N] = "PAL-N",
+	[VC4_VEC_TV_MODE_PAL60] = "PAL60",
+	[VC4_VEC_TV_MODE_SECAM] = "SECAM",
+};
+
+enum vc4_vec_tv_mode_id
+vc4_vec_get_default_mode(struct drm_connector *connector) {
```
pelwell: Function start braces must go on separate lines.

drivers/gpu/drm/vc4/vc4_vec.c
```diff
@@ -343,6 +345,41 @@ static const struct vc4_vec_tv_mode vc4_vec_tv_modes[] = {
...
+			if (strcmp(vc4_vec_tv_norm, tv_mode_names[i]) == 0)
+				return (enum vc4_vec_tv_mode_id) i;
+	} else if (connector->cmdline_mode.specified &&
+		   ((connector->cmdline_mode.refresh_specified &&
+		     (connector->cmdline_mode.refresh == 25 ||
+		      connector->cmdline_mode.refresh == 50)) ||
+		    (!connector->cmdline_mode.refresh_specified &&
+		     (connector->cmdline_mode.yres == 288 ||
+		      connector->cmdline_mode.yres == 576)))) {
+		/* no explicitly specified TV norm; use PAL if a mode that
```
pelwell: Kernel multi-line comments must be of the form:
```
/*
 * blah blah
 * blah blah blah
 */
```

pelwell:
This isn't really my field, so apologies that I can only contribute nit-picky cosmetic feedback, but this looks like an impressive patch set.

drivers/gpu/drm/vc4/vc4_vec.c
```diff
@@ -68,6 +68,7 @@
 #define VEC_CONFIG0_STD_MASK		GENMASK(1, 0)
 #define VEC_CONFIG0_NTSC_STD		0
 #define VEC_CONFIG0_PAL_BDGHI_STD	1
+#define VEC_CONFIG0_PAL_M_STD		2
```
6by9, line 72: Can I ask where you got that define from? The datasheet I have lists the values as
0 - NTSC
1 - PAL_BDGHI
2 - reserved
3 - PAL_N

The firmware only selects from modes 0, 1, and 3, so I'm wondering where your mapping for 2 has come from.

drivers/gpu/drm/vc4/vc4_vec.c
```diff
@@ -75,6 +76,27 @@
...
+ * 0x80000000 seems to be equivalent to the pixel clock
+ * (which itself is the VEC clock divided by 8).
+ *
+ * Reference values (with the default pixel clock of 13.5 MHz):
+ *
+ * NTSC  (3579545.[45] Hz)     - 0x21F07C1F
+ * PAL   (4433618.75 Hz)       - 0x2A098ACB
+ * PAL-M (3575611.[888111] Hz) - 0x21E6EFE3
+ * PAL-N (3582056.25 Hz)       - 0x21F69446
+ *
```
6by9, line 95: Datasheet says colour subcarrier frequency = 13.5MHz * FREQ / 2^31.

However the firmware code for setting up PAL_M (the only option which uses a custom subcarrier) is 0x223b61d1, which differs from yours. Computing the values, I agree with your numbers though.

drivers/gpu/drm/vc4/vc4_vec.c
```diff
@@ -75,6 +76,27 @@
...
+ * NOTE: For SECAM, it is used as the Dr center frequency,
+ * regardless of whether VEC_CONFIG1_CUSTOM_FREQ is enabled or not;
+ * that is specified as 4406250 Hz, which corresponds to 0x29C71C72.
```
6by9, line 98: That 0x29C71C72 corresponds to the default value by the looks of it.

drivers/gpu/drm/vc4/vc4_vec.c
```diff
@@ -117,6 +139,13 @@
 
 #define VEC_INTERRUPT_CONTROL		0x190
 #define VEC_INTERRUPT_STATUS		0x194
+
+/*
+ * Db center frequency for SECAM; the clock for this is the same as for
+ * VEC_FREQ3_2/VEC_FREQ1_0, which is used for Dr center frequency.
+ *
+ * This is specified as 4250000 Hz, which corresponds to 0x284BDA13.
```
6by9, line 147: 0x284BDA13 is the default value for the register. Probably worth mentioning that as otherwise there is no other reference to it, and VEC_FCW_SECAM_B is never set.

drivers/gpu/drm/vc4/vc4_vec.c
```diff
@@ -45,6 +45,7 @@
 #define VEC_CONFIG0_YDEL(x)		((x) << 26)
 #define VEC_CONFIG0_CDEL_MASK		GENMASK(25, 24)
 #define VEC_CONFIG0_CDEL(x)		((x) << 24)
+#define VEC_CONFIG0_SECAM_STD		BIT(21)
```
6by9, line 48: More magic fields. CONFIG0 bits 23:19 are defined a reserved, write as 0. Then again it also doesn't say how to select SECAM mode!

drivers/gpu/drm/vc4/vc4_vec.c
6by9: The drm subsystem is happy with
```
/* blah
 * blah
*/
```
but the closing ```*/``` does have to be on a separate line.

drivers/gpu/drm/vc4/vc4_vec.c
```diff
@@ -372,11 +409,19 @@ static int vc4_vec_connector_get_modes(struct drm_connector *connector)
 	return 1;
 }
 
+static void vc4_vec_connector_reset(struct drm_connector *connector)
+{
+	drm_atomic_helper_connector_reset(connector);
+	/* preserve TV standard */
+	if (connector->state)
+		connector->state->tv.mode = vc4_vec_get_default_mode(connector);
```
6by9, line 421: Do we want to reset to the default (as defined by the module parameter), or retain the old setting?
If you change the module parameter (it is permissions 0600) then on the next reset call don't we have the same scenario as at boot where the mode may not match the standard?

6by9 (commented):
Thanks for the patchset.

I need to test it out to satisfy myself that it actually does the right thing, but other than the minor niggles below it looks good.
I have asked mripard over passing the tv standard via the kernel command line ```video=``` entry. Seeing as he wrote most of that code he's a pretty good person to ask :-) I don't see any way that it sets the tv_mode property, so the module parameter seems a good option at present.

The only other driver I really see handling tv_mode is i915, and it just always defaults to NTSC.

drivers/gpu/drm/vc4/vc4_vec.c
pelwell: You're right about the comment - checkpatch says "WARNING: Block comments use a trailing */ on a separate line".

mripard:
> > I'm not sure we need to read-out the hardware state. The command line parser explicitly supports named modes, so the firmware could just set video=TV-1:NTSC-J or whatever and the initial framebuffer will use that mode.
> > Switching back and forth between modes is more problematic though. The userspace is free to build whatever mode it wants, and / or the framebuffer emulation (and thus the initial setup) can be disabled.
> > In such a case, you don't really have a way to differentiate between variants that use the same mode (like between NSTC-J and PAL-M). Once again, you can only hope that the name will be filled and compare the name to the modes you support, and fallback on PAL or NTSC
> 
> Well, unfortunately the current versions of the firmware only pass information such as `video=Composite-1:720x480@60i` or `video=Composite-1:720x576@50i`. There is no way to differentiate between NTSC vs. NTSC-J vs. PAL-M or PAL vs. PAL-N, even though all of these are accessible via `sdtv_mode`.
> 
> I agree that what I wrote is a bit of a kludge, and I agree that if the firmware could pass the colour system settings on the kernel's command line, that would be much better. I'm happy to update my code once it does, but for now, reading the VEC register state found at boot is the only way to get this information.

My main issue with reading out the hardware state is that it means that users won't be able to change the TV norm without rebooting, and it has weird interactions with other parts of the system, for example:
  * If the user disables the firmware output in the config.txt, they will be forced into NTSC even though the kernel is supposed to be entirely in charge
  * There's a similar issue if we boot with the HDMI output, and then want to use the TV output
  * If the kernel is build without the simplefb driver and the DRM driver is built as a module, chances are the clocks and power domains of the VEC are going to be disabled at some point, and would make it lose the content of its registers.

The tv_mode parameter you introduced addresses most of these concerns, but it's redundant with the same property that is already set. The command line parser can parse arbitrary properties, so it would make more sense to extend this part:
https://github.com/raspberrypi/linux/blob/rpi-5.10.y/drivers/gpu/drm/drm_modes.c#L1638

Adding another switch statement that fills a new field in drm_cmdline_mode that would store the mode, and then init the tv_mode property to that value here:
https://github.com/raspberrypi/linux/blob/rpi-5.10.y/drivers/gpu/drm/drm_atomic_state_helper.c#L439

would work quite nicely without any ad-hoc solution.

drivers/gpu/drm/vc4/vc4_vec.c
mripard: reset is only called once at boot time, it's what creates the first initial commit. Any state after that is the union of the previous state and whatever has been changed in the various commits that occurred.

kFYatek:
> My main issue with reading out the hardware state is that it means that users won't be able to change the TV norm without rebooting

This is actually not true. `xrandr --output Composite-1 --off && xrandr --output Composite-1 --set mode NTSC-J --mode 720x480` does the job, I do this kind of thing all the time when testing this PR. It works both with the current version of this PR and it also worked with the revision that queried hardware state.

Switching between 525-line (NTSC/NTSC-J/NTSC-443/PAL-M/PAL60) and 625-line (PAL/PAL-N/SECAM) modes requires adding the missing mode via `--newmode` and `--addmode`, but otherwise also works. I wonder if maybe that means we should return both modes through `vc4_vec_connector_get_modes()` - but they cannot be mixed up, i.e. e.g. `720x576` does not work (to the point of outputting complete garbage on the cable) when the norm is set to NTSC, so having a mode that cannot be used without specifying `--set mode` could be confusing.

6by9:
None of it feels totally clean.

Adding an extra property to ```video=``` will end up requiring something like ```video=TV-1:NTSC,tv_mode=NTSC-J``` as the parser currently insists on a width/height first. Mix and match and nasty things will happen.
Do I recall you discussing allowing properties to be set without the mode? It'd be potentially of use for setting HDMI limited/full range overrides too.

It's also slightly ugly that the cmdline parsing appears to be called from ```drm_connector_init```, so the string names for each of the TV modes haven't been added to the connector at that point (they've been created via ```drm_mode_create_tv_properties``` but not ```drm_object_attach_property```ed to the connector),  and there are no standardised enums for these values.

drivers/gpu/drm/vc4/vc4_vec.c
kFYatek: It's been a while since I found this, so my memory is not exactly clear, but AFAIR it was mostly experimentation. I was just messing with the registers and seeing what happens. I set what's defined here as `VEC_CONFIG0_STD_MASK` to 2 and poof, a standards-compliant PAL-M mode appeared.

I wrote https://github.com/kFYatek/tweakvec some time earlier to mess with these registers from user space. The effects of that have been discussed a bit in https://github.com/raspberrypi/firmware/issues/811, where @Sobakin76 confirmed that this works on, let's call it, "period-accurate" displays. Here's a video they made that shows it: https://youtu.be/299IORo56lM (note, however, that they have their Pi set to 240p progressive mode). PAL-M support is uncommon on devices from my region, so I couldn't do that personally, but I did check it on modern-ish devices (particularly the Fushicai USBTV007 grabber with Windows drivers, and a Samsung UE40E6800S TV - both are curiously compatible with pretty much every TV norm ever).

6by9: I've now checked the 2711 register definitions, and those values are listed as reserved there too.

I'll see if I can get anyone at Broadcom to confirm your findings.

drivers/gpu/drm/vc4/vc4_vec.c
kFYatek: `0x223b61d1`, which translates to 3610402.169 Hz, is a long-running bug in the firmware. I'm not aware of a TV norm that would use this frequency ever existing. `sdtv_mode=3` indeed configures that frequency (with `VEC_CONFIG0_STD_MASK` set to 0 for NTSC), but that mode is pretty much unusable. [Here's how `sdtv_mode=3` looks on my Fushicai USBTV007 grabber](https://gist.github.com/kFYatek/0b2dc2f85ecc933b34f9df2559390dd4). See also https://github.com/raspberrypi/firmware/issues/756, there are pictures of @Sobakin76's monitors showing the same thing there.

drivers/gpu/drm/vc4/vc4_vec.c
kFYatek: Well, that's another product of me just messing around with this register. I read this header, saw that `VEC_FCW_SECAM_B` and `VEC_SECAM_GAIN_VAL` registers and was intrigued, especially since nothing else referred to SECAM and I don't think support for SECAM on the Pi was ever seen before. I started flipping bits in this register and when I set this one, again, poof, SECAM appeared.

I empirically confirmed that this mode is detected as SECAM and looks right on a couple of modern devices:
* Fushicai USBTV007 grabber
* Samsung UE40E6800S
* Panasonic TX-L42ET5E
* Samsung UE65MU8002 (through an RF modulator, this TV does not have any baseband analog inputs)

Again, @Sobakin76 also confirmed this working on their more period-accurate Sony PVM-14N5MDE CRT: https://github.com/raspberrypi/firmware/issues/811#issuecomment-802360785

EDIT: Note that the artifacts on their photo are a known issue with the SECAM system itself; this is basically what happens on it instead of PAL/NTSC-style "dot crawl" when the image has too much sharp edges across the horizontal axis (exactly like the fine text on that picture screenshot).

6by9:
I've grabbed the Kramer VP719xl that I have access to which can supposedly handle NTSC, NTSC4.43, PAL, PAL-N, PAL-M, SECAM. So it nominally misses NTSC-J and PAL-60, but should be able to tell me about the rest.

Now to find my A/V lead to connect it up....

kFYatek:
@6by9 If it supports NTSC and PAL, it almost certainly supports PAL-60. It will probably just label it as "PAL", while still syncing at 60Hz.

NTSC-J and regular NTSC are also identical except for the black level; i.e., if you set NTSC-J and never go below about 5% gray level, you'll end up with a compliant NTSC(-M) signal.

So I think your display should handle all the modes.

drivers/gpu/drm/vc4/vc4_vec.c
kFYatek: Uh, this is actually one of the places where I don't exactly know what I'm doing. From what I understand, this function is called when a new connector "state" object is created. By default, it is more or less just zeroed out, which means also setting `tv.mode` field to 0, which is NTSC in our case. The default value set through `drm_object_attach_property()` is not respected for some reason, so without this code, we would end up in NTSC mode at boot regardless of what was set there.

Setting this value later in userspace (`xrandr --set mode PAL` etc.) still works.

Actually the only other driver in the entire kernel codebase that uses the `tv.mode` field is i915 - which defaults to NTSC anyway, as pointed out in https://github.com/raspberrypi/linux/pull/4241#pullrequestreview-691627291. nouveau and ch7006 just keep track of it in a separate private variable and use that instead.

I don't know, maybe a cleaner solution would be to patch the DRM subsystem itself so that it would pull the default value from whatever has been set through `drm_object_attach_property()` - because not respecting that even though the `tv_mode_property` is automatically mapped to that field seems like a bug. Maybe that should be done directly in `vc4_vec_connector_reset()` or something, although I'm not sure if that's even possible.

kFYatek:
The newest update addresses those couple of cosmetic/formatting issues from @pelwell's and @6by9's reviews.

drivers/gpu/drm/vc4/vc4_vec.c
kFYatek: Fixed.

drivers/gpu/drm/vc4/vc4_vec.c
kFYatek: Fixed.

drivers/gpu/drm/vc4/vc4_vec.c
kFYatek: Fixed.

6by9:
Tested against my Kramer (it's a [presentation switcher/scaler](https://k.kramerav.com/downloads/manuals/vp-719xl.pdf) rather than a display).
Whilst the OSD only tells me NTSC, PAL, or SECAM, the controls allow me to explicitly set the standard, and they are all seemingly decoded with the appropriate setting.

I had expected changing mode (at least between those with the same resolution) wouldn't require a manual disable/reconfigure/enable. That should achievable by setting ```crtc_state->mode_changed = true;``` from ```vc4_vec_encoder_atomic_check```, but I don't see a way to get to the old ```drm_connector_state``` from an encoder atomic_check to check if it changed or not.
The encoder atomic_check is also seemingly not called if you just change the property.
Ah, vc4_vec also has a ```struct drm_connector_helper_funcs``` which can implement the connector atomic_check hook. Doing that also correctly stops you setting a mode that doesn't match the standard (I appear to be able to do that with the current patches, although my current hack ends up doing a NULL deref in drm_atomic_get_crtc_state at present).

It may be this setup, but it looks like it's working at a very reduced bit depth on the VEC output. A quick hack to the overlay so that HDMI is active at the same time and the images are significantly different (I'll try and attach a photo in a bit). I'd be interested to know if you see the same. The desktop looks OK, but a JPEG is quantised.

kFYatek:
> I had expected changing mode (at least between those with the same resolution) wouldn't require a manual disable/reconfigure/enable [...]

I refactored the `atomic_check` logic to happen in the connector object instead of in the encoder object. That way it is called when just changing properties and this works in the newest revision.

However, [Xorg does not check the ioctl return value when setting properties](https://github.com/freedesktop/xorg-xserver/blob/a50a0e348542bf32fb4239ee05a485772837f2f3/hw/xfree86/drivers/modesetting/drmmode_display.c#L2982), so if you try to set an invalid value (e.g. `xrandr --output Composite-1 --set mode PAL` when the resolution is 480i), it will do nothing but X will report as if it has been changed. I guess that's a bug in Xorg, though.

> It may be this setup, but it looks like it's working at a very reduced bit depth on the VEC output. A quick hack to the overlay so that HDMI is active at the same time and the images are significantly different (I'll try and attach a photo in a bit). I'd be interested to know if you see the same. The desktop looks OK, but a JPEG is quantised.

I haven't noticed anything like that. What JPEG did you use and what program did you use to display it?

In my experience, there is a bit of banding, but nothing too distracting, and importantly it seems to be more or less the same as in the firmware framebuffer. [I did some screenshots with my USB grabber](https://gist.github.com/kFYatek/58f988965287835ebaf0d25865f91b4d), showing the [gradient.png file from rtings.com](https://www.rtings.com/images/gradient.png) rendered in Chromium in fullscreen mode (F11).

6by9:
Thanks for your perseverance on this.

The refactor to use connector_atomic_check is almost exactly what I'd ended up with. Looks fine.
X doing the wrong thing is X's problem, not the kernel's.

I'd just viewed a JPEG (an unfocused capture from an HQ camera module) and there is a significant difference between the HDMI output and VEC. See this image where the front monitor is on HDMI, and the back one is the VEC fed through the Kramer to give VGA out.
DSC_0103
A greyscale ramp may be the best thing to test with to confirm if it's just my setup or something more fundamental.

kFYatek:
Whoah. That looks bad indeed. But as you can see on my screenshots linked above, I saw nothing like this.

I found some time ago, however, that the Pi (Pi 4 especially) outputs composite signals that are not exactly the standard voltage. When I used an RF modulator (just a [cheapo thing from Amazon](https://www.amazon.co.uk/Modulator-Converter-Channel-Compatible-Extender-default/dp/B08H1XKW8V)) to connect the Pi's composite output to my modern Samsung TV that only accepts analog signals via RF, I found that the brightest parts of the image got clipped. I was able to get the picture to look right by putting a potentiometer between the Pi and the modulator and setting it slightly down from "100%". I can make pictures and measure the actual resistance if you want.

I found that modulator is based on the [MC44BS373CA](https://www.nxp.com/docs/en/data-sheet/MC44BS373CA.pdf) chip, with expects 1 V peak-to-peak video input (see page 7 of the datasheet, among others). The Pi's signal is not exactly that, and it has a nonzero DC component (again, this is especially true on Pi4; on Pi4, I could not find any VEC register tweaks that could fix this issue), which confuses that modulator. All other devices I tested employ some kind of AGC and thus have no problems with its output.

My wild guess is that maybe your Kramer also doesn't like the voltages output by the Pi for whatever reason. Although this weird green tint seems to affect neither the highlights or the shadows, but rather the middle of the brightness range, which is particularly odd and not at all consistent with the voltage mismatch theory...

pelwell:
Say the word when you think this is ready for merging.

kFYatek:
@pelwell I'm not sure who you're asking. I considered it ready from the beginning, and updates were just addressing the reviews ;) With the catch that I expanded on ideas from this PR in #4406, which refactors some code refactored here again - so you might want to take a look there as well.

I was under the impression that we're waiting for acceptance from @mripard and, mostly, @6by9.

pelwell:
Understood - I was addressing the reviewers as well.

popcornmix:
I'm in favour of merging - it sounds like it's had more thorough testing that the current composite kms code.
But an okay from @mripard / @6by9 would be good.

6by9:
Sorry, juggling too many things.
I believe I'm good with these. My last few tests weren't conclusive, but that looked to be my setup rather than the code being wrong.

One last checkpatch complaint (sorry) - patch author doesn't match the Signed-Off-By
```
ERROR: Missing Signed-off-by: line by nominal patch author 'kFYatek <4499762+kFYatek@users.noreply.github.com>'
```
If you've updated your standard author email address, then ```git commit --amend --reset-author``` will up the current patch. Doing it during a ```rebase -i``` allows you to update it on each patch (sorry if I'm teaching you to suck eggs).

kFYatek:
No problem. I believe it should be OK now.

6by9:
Thank you for fixing those up.

Checkpatch --strict still has one final whinge of 
```
CHECK: No space is necessary after a cast
#62: FILE: drivers/gpu/drm/vc4/vc4_vec.c:368:
+				return (enum vc4_vec_tv_mode_id) i;

total: 0 errors, 0 warnings, 1 checks, 115 lines checked
```
but I'm happy to ignore that one as otherwise the ```i``` disappears into the bracket.

LGTM.

PR #4242, opened 2021-03-28
Remove spi override from merus-amp overlay

AMuszkat:
Previously added override at ["Add configuration override for MSEL0 on merus-amp overlay"](https://github.com/raspberrypi/linux/pull/4027) generates no audio in Raspberry pi 3 and 4 as due to MA12070P MSEL0 pin is only compatible with 5V. For now the solution is to remove the override and to avoid future problems on this board a hardware update will be launched soon.


pelwell:
Do you have any comments, @shawaj?

AMuszkat:
@pelwell,

Can we merge this? We haven't had any replies from @shawaj.

Best,
Ariel

pelwell:
Sure, but can you call the commit `overlays: merus-amp: remove spi override`, and rebase it? GitHub looks a bit confused.

AMuszkat:
done!

PR #4244, opened 2021-03-29
DSI updates to use the DSI0 and JDI LT070ME05000 panel

6by9:
Still not 100% reliable :-(

Fixes up DSI0 with the mainline patches rather than earlier versions.


drivers/gpu/drm/vc4/vc4_dsi.c
```diff
@@ -1060,12 +1105,9 @@ static void vc4_dsi_encoder_enable(struct drm_encoder *encoder)
...
 	/* Bring AFE out of reset. */
-	if (dsi->variant->port == 0) {
-	} else {
-		DSI_PORT_WRITE(PHY_AFEC0,
-			       DSI_PORT_READ(PHY_AFEC0) &
-			       ~DSI1_PHY_AFEC0_RESET);
-	}
+	DSI_PORT_WRITE(PHY_AFEC0,
+		       DSI_PORT_READ(PHY_AFEC0) &
+		       ~DSI1_PHY_AFEC0_RESET);
```
pelwell: DSI0_PHY_AFEC0_RESET and DSI1_PHY_AFEC0_RESET are different bit positions (in different registers, although the interleaved definitions don't make that obvious). The reset code above uses the correct value for the port in question, but here the DSI1 value is used in both cases. Use `DSI_PORT_BIT(PHY_AFEC0_RESET)` instead?

drivers/gpu/drm/vc4/vc4_dsi.c
```diff
@@ -1184,13 +1226,28 @@ static ssize_t vc4_dsi_host_transfer(struct mipi_dsi_host *host,
 	/* Enable the appropriate interrupt for the transfer completion. */
 	dsi->xfer_result = 0;
 	reinit_completion(&dsi->xfer_completion);
-	DSI_PORT_WRITE(INT_STAT, DSI1_INT_TXPKT1_DONE | DSI1_INT_PHY_DIR_RTF);
-	if (msg->rx_len) {
-		DSI_PORT_WRITE(INT_EN, (DSI1_INTERRUPTS_ALWAYS_ENABLED |
-					DSI1_INT_PHY_DIR_RTF));
+	if (!dsi->port) {
```
pelwell: `dsi->port == 0` is used above (but not added in this commit).

arch/arm/boot/dts/overlays/vc4-kms-dsi-7inch-overlay.dts
```diff
@@ -6,51 +6,97 @@
...
+
 / {
 	compatible = "brcm,bcm2835";
 
 	fragment@0 {
 		target = <&dsi1>;
 		__overlay__ {
-			#address-cells = <1>; size-cells = <0>;
+			#address-cells = <1>;
+			size-cells = <0>;
```
pelwell: `#size-cells`

arch/arm/boot/dts/overlays/edt-ft5406.dtsi
```diff
@@ -0,0 +1,42 @@
+/*
+ * Device Tree overlay for RaspberryPi 7" Touchscreen panel
+ *
+ */
+
+/dts-v1/;
+/plugin/;
```
pelwell: I'm not keen on having these tags in a dtsi file - they should only appear once in a DTS/overlay file, and I feel they're better of in the parent .dts (although there's at least one counter-example).

pelwell: Alternatively, given that the .dtsi here is a complete overlay, I'd be happy with it remaining as a .dts, even though it is included by another overlay (although the including file should then lose it's own header tags).

arch/arm/boot/dts/overlays/README
```diff
@@ -3141,10 +3141,14 @@ Params: cma-512                 CMA is 512MB (needs 1GB)
 
 Name:   vc4-kms-dsi-7inch
 Info:   Enable the Raspberry Pi DSI 7" screen.
-        Use edt-ft5406 for the touchscreen element.
+        Includes the edt-ft5406 for the touchscreen element.
         Requires vc4-kms-v3d to be loaded.
 Load:   dtoverlay=vc4-kms-dsi-7inch
```
pelwell: Sorry, but this now needs `,<param>=<val>`.

arch/arm64/configs/bcmrpi3_defconfig
```diff
@@ -708,6 +708,7 @@ CONFIG_STMPE_SPI=y
 CONFIG_MFD_ARIZONA_I2C=m
 CONFIG_MFD_ARIZONA_SPI=m
 CONFIG_MFD_WM5102=y
+CONFIG_REGULATOR_RASPBERRYPI_TOUCHSCREEN_ATTINY=m
```
pelwell: This setting isn't working - it needs at least CONFIG_REGULATOR=y, but that doesn't seem to be sufficient.

6by9:
Rebased, updated based on your comments.

I've dropped the 7inch panel updates as I haven't had a chance to resolve the issues, but added in the overlay for the JDI LT070ME05000 panel that is now working using the hardware documented at https://github.com/harlab/CM4_LCD_LT070ME05000

6by9:
@harlab Are you happy with the text I've added for your contributions?

harlab:
> @harlab Are you happy with the text I've added for your contributions?

@6by9 overlays were based on work of Pi forum user, so I can’t take full credit. Leaving only repo link should be fine. 
Thanks for your work.

6by9:
OK, will drop your email address (avoids some spam!)

Just so that I can include correct credits, was it gizmomouse at https://www.raspberrypi.org/forums/viewtopic.php?f=98&t=253912 that you're referencing? I'm happy to include a credit to them too.

harlab:
> OK, will drop your email address (avoids some spam!)

Thanks

> Just so that I can include correct credits, was it gizmomouse at https://www.raspberrypi.org/forums/viewtopic.php?f=98&t=253912 that you're referencing? I'm happy to include a credit to them too.

@6by9 yes, that’s what overlays are based on.

6by9:
Updated commit text based on harlab's comments, and I believe addressed all of pelwell's comments on the earlier version.

Hopefully all good now with that display, so now I can go back to the Pi 7" one.

arch/arm/boot/dts/overlays/README
```diff
@@ -3158,6 +3158,27 @@ Load:   dtoverlay=vc4-kms-dsi-7inch
...
 
+Name:   vc4-kms-dsi-lt070me05000
+Info:   Enable a JDI LT070ME05000 DSI display on DSI1.
+        Note that this is a 4 lane DSI device, so it will only work on a Compute
+        Module.
+        Requires vc4-kms-v3d to be loaded.
+Load:   dtoverlay=vc4-kms-dsi-lt070me05000,<param>
+Params: reset                   GPIO for the reset signal (default "17")
+        enable                  GPIO for the enable signal (default "4")
+        dcdc-en                 GPIO for the DC-DC converter enable (default "5")
```
pelwell: This line is one character too long. I would drop the quotes from the defaults.

pelwell (commented):
One last niggle.

arch/arm/boot/dts/overlays/README
6by9: Done.
We obviously need to get you a bigger monitor as Linus has relented on the hard limit of 80 columns - https://lkml.org/lkml/2020/5/29/1038

pelwell: I became aware of the increased line length limit last week. I still prefer the non-string defaults without quotes, but I could relax the 80-char limit in the script a bit (after checking that the dtoverlay/dtparam utils won't die horribly).

pelwell:
```
    char line_buf[82];
```
Some work required there.

PR #4249, opened 2021-03-31
CM4 PCIe/XHCI updates

pelwell:
CM4 (and, to a lesser extent, Pi 400) has fallen behind 4B in DT support. In particular it's missing the reset controller needed for PCIe support.


pelwell:
How does this look, @timg236 ?

timg236:
It looks good to me, I'll cherry-pick the patch and rebuild to confirm

@peterharperuk

timg236:
Thanks @pelwell that works for me

PR #4251, opened 2021-03-31
vc4/firmware_kms: Add a usecnt on scanout buffers

bpeel:
Like `vc4_kms`, when a buffer is used for scanout, it needs to have a usecnt set on it otherwise it is liable to being purged if userspace marks it as `DONTNEED`. Although this would be a weird thing for userspace to do, this is worth preventing because it will make the hardware scan out from random CMA memory used by other buffers.

The two new functions are identical copies of those in [`vc4_plane.c`](https://github.com/raspberrypi/linux/blob/rpi-5.12.y/drivers/gpu/drm/vc4/vc4_plane.c#L1321).

I also made a demonstration of the problem using a hack on top of kmscube [here](https://gitlab.freedesktop.org/nroberts/kmscube/-/tree/vc4-purgeable-scanout-test).

Although this is probably only a minor problem for now, it would be important to fix this if we merge the [CMA pool work](https://github.com/bpeel/linux/tree/nroberts/cma-pool) because the branch also uses the usecnt to decide whether to move buffers out of the pool.


6by9:
I wouldn't claim to be an expert in how KMS handles buffers, but it looks fine to me.

PR #4258, opened 2021-04-02
Update Allo Piano Plus DAC driver

paul-1:
The patch to create unique dac names as required in 5.4 kernels has been updated to apply to the 5.10.y kernel.   The overlay update that was required has already been applied in 5.10

From my testing, everything seems to work fine with the DAC.

@alloboss2  I checked my commit against your code in https://github.com/allocom/linux/commit/b9f4b9e1e991180efb80f52135f7ad8406dd2290 the only thing I see is I left the driver probe changes alone.


pelwell:
This has now reached rpi-update.

PR #4262, opened 2021-04-06
dwc-otg: fix clang warnings

jo-he:
The one in dwc_otg_pcd.c actually fixes a minor bug.


popcornmix:
Looks plausible. Is https://github.com/raspberrypi/linux/blob/d1fd8a5727908bb677c003d2ae977e9d935a6f94/drivers/usb/host/dwc_otg/dwc_otg_pcd_linux.c#L140 also wrong?

jo-he:
No, that one should be fine, as it is comes from `struct usb_endpoint_descriptor` (include/uapi/linux/usb/ch9.h) where is is declared as `__le16` (and checking for 0 works independently of endianess).

While the one my patch fixes comes from `usb_endpoint_descriptor_t` (drivers/usb/host/dwc_common_port/usb.h) where it is declared as `uWord`, which in turn is a two-byte array.

popcornmix:
Okay, seems good to me. @P33M ?

P33M:
LGTM - note that I don't think there is ever a case where we use the dwc_otg PCD (aka gadget driver), but thanks for fixing it.

PR #4267, opened 2021-04-08
ARM: dts: Remove usless sdhci/wifi node from bcm2711-rpi-cm4

iluminat23:
The sdhci node gets deleted in bcm2711.dtsi. The entry in this dts is
usless and makes the false impression it would be used.

Signed-off-by: Philipp Rosenberger <p.rosenberger@kunbus.com>

### DIFF
--- a/arch/arm/boot/dts/bcm2711-rpi-cm4.dts
+++ b/arch/arm/boot/dts/bcm2711-rpi-cm4.dts
@@ -40,11 +40,6 @@
 		};
 	};
 
-	wifi_pwrseq: wifi-pwrseq {
-		compatible = "mmc-pwrseq-simple";
-		reset-gpios = <&expgpio 1 GPIO_ACTIVE_LOW>;
-	};
-
 	sd_io_1v8_reg: sd_io_1v8_reg {
 		compatible = "regulator-gpio";
 		regulator-name = "vdd-sd-io";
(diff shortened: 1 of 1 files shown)


pelwell:
This CM4 dts file was written in isolation, attempting to minimise the differences from the pi4 dts. The reason the sdhci nodes are created, then deleted and rebuilt is that the first version is what appears in the upstream dts, but it lacks some of the flexibility we make use of with overlays.

Until upstream versions of the Pi 400 and CM4 dts files appear I would rather keep them as they are.

iluminat23:
OK, but I'm a bit puzzled on how the wifi driver gets probed. As the devicetree doesn't contain any reference to the driver. Even on the running system I don't see any entry in the loaded devicetree. Could you give me a pointer where the magic happens? I checked the devicetree in /proc and the blob in /sys/firmware/fdt.

pelwell:
SDIO is discoverable - devices have IDs that Linux can query in order to locate drivers, in the same way that it handles USB devices.

iluminat23:
Thank you, didn't know that sdio is discoverable.

pelwell:
In the absence of instructions not to, Linux treats every MMC/eMMC/SD/SDIO interface as if it could be any of those standards, using the failure to respond to particular commands as an indication that the device doesn't support each standard in turn, and stopping when it finds one that is supported. That means that every time you boot a regular Pi Linux first treats the SD card as is if it might be an SDIO card.

PR #4268, opened 2021-04-09
HDR improvements

mripard:
This is a sync with the HDR support sent upstream plus some changes Dom suggested


HiassofT:
Testing with LibreELEC/Kodi was fine so far, my LG TV showed BT.2020 in status info when playing BT.2020 files, and the feedback from our users also indicated that colorspace signalling worked correctly on other devices.

pelwell:
That looks fine to me. Merge if you're happy, @popcornmix.

PR #4272, opened 2021-04-12
sc16is7xx: Defer probe if device read fails

mcdermj:
Commit 158e800e0fde upstream

A test was added to the probe function to ensure the device was
actually connected and working before successfully completing a
probe.  If the device was actually there, but the I2C bus was not
ready yet for whatever reason, the probe fails permanently.

Change the probe so that we defer the probe on a regmap read
failure so that we try the probe again when the dependent drivers
are potentially loaded.  This should not affect the case where the
device truly isn't present because the probe will never successfully
complete.

Fixes: 2aa916e67db3 ("sc16is7xx: Read the LSR register for basic device presence check")
Cc: stable@vger.kernel.org
Signed-off-by: Annaliese McDermond <nh6z@nh6z.net>
Link: https://lore.kernel.org/r/010101787f9c3fd8-c1815c00-2d6b-4c85-a96a-a13e68597fda-000000@us-west-2.amazonses.com
Signed-off-by: Greg Kroah-Hartman <gregkh@linuxfoundation.org>

### DIFF
--- a/drivers/tty/serial/sc16is7xx.c
+++ b/drivers/tty/serial/sc16is7xx.c
@@ -1204,7 +1204,7 @@ static int sc16is7xx_probe(struct device *dev,
 	ret = regmap_read(regmap,
 			  SC16IS7XX_LSR_REG << SC16IS7XX_REG_SHIFT, &val);
 	if (ret < 0)
-		return ret;
+		return -EPROBE_DEFER;
 
 	/* Alloc port structure */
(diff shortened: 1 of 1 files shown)


mcdermj:
@pelwell Here is another fix that we found we needed for our board having to do with the sc16is7xx serial port driver.  I think other folks are using this device as well.  This was accepted upstream into tty-next and should be filtering its way down to the stable kernels too.

pelwell:
Thanks!

Issue #4200, opened 2021-03-10
audiosense-pi overlay not loading since 5.10 kernel


linuxchris12:
**Describe the bug**
the audiosense-pi overlay does not load correctly anymore in the 5.10 kernel (Pi 4 B)

**To reproduce**
Add: 
```
dtparam=i2s=on
dtparam=spi=on
dtparam=audio=on
dtoverlay=audiosense-pi
```
to `/boot/config.txt`. Restart and use `aplay -l`
This will not list the audiosensepi soundcard anymore

**Expected behaviour**
The soundcard is listed like this:
```
card 1: audiosensepi [audiosense-pi], device 0: TLV320AIC3204 Hifi Audio tlv320aic32x4-hifi-0 [TLV320AIC3204 Hifi Audio tlv320aic32x4-hifi-0]
  Subdevices: 1/1
  Subdevice #0: subdevice #0
```

**Actual behaviour**
the audiosensepi soundcard is not listed. It is shown as loaded though using `lsmod`

**System**
 Copy and paste the results of the raspinfo command in to this section. Alternatively, copy and paste a pastebin link, or add answers to the following questions:

* Which model of Raspberry Pi? Pi4B
* Which OS and version (`cat /etc/rpi-issue`)? Raspberry Pi reference 2019-06-20
* Which firmware version (`vcgencmd version`)? Feb 25 2021 12:10:40, version 564e5f9b852b23a330b1764bcf0b2d022a20afd0 (clean) (release) (start)
* Which kernel version (`uname -a`)? Linux raspberrypi 5.10.17-v7l+ #1403 SMP Mon Feb 22 11:33:35 GMT 2021 armv7l GNU/Linux

**Additional context**
Source of the issue was this change on the tlv320aic32x4 codec:
https://github.com/raspberrypi/linux/commit/df44bc16e616809172cda90fd816596ded4ea219

linuxchris12:
Workaround: comment out some of the changes introduced in commit
https://github.com/raspberrypi/linux/commit/df44bc16e616809172cda90fd816596ded4ea219

```
diff --git a/sound/soc/codecs/tlv320aic32x4.c b/sound/soc/codecs/tlv320aic32x4.c
index 9e3de9ded..5ac1418d7 100644
--- a/sound/soc/codecs/tlv320aic32x4.c
+++ b/sound/soc/codecs/tlv320aic32x4.c
[...]

PR #4273, opened 2021-04-12
Codec Fixes for the tlv320aic32x4 codec on 5.10

mcdermj:
These are a pair of fixes to make the tlv320aic32x4 codec driver work on 5.10.  Both have been submitted upstream and accepted for inclusion into -next.  They were also submitted to hopefully be integrated into -stable as well.  This fixes some of the problems in issue #4200 that I had discussed (but had not fixed the original poster's issue).

We've tested these fixes on our boards and they seem to fix the issues with the codec.


pelwell:
Thanks again.

PR #4276, opened 2021-04-12
vc4/kms: vc4_plane: Support 2020 colourspace for yuv planes

popcornmix:
https://gist.github.com/popcornmix/6b3e23103c60170b02b148e0ba5d6ed7

is the script used to generate the 601, 709 and 2020 colourspaces.
I've regenetated the existing ones using script so it is reprocable
but there are lsb dfferences compared to values here (copied from spec)
whose origin is now lost.


popcornmix:
Some positive reports e.g. https://forum.libreelec.tv/thread/23442-rpi4-testbuild-with-hdr-support/?postID=153971#post153971
I've been through the differences in values from my script and spec.

Spec mentions:
```
JPEG JFIF  Y’/Cb/Cr = [0,255]
Component  Y’ (Cb-128) (Cr-128)
R’         1     0        1.402
G’         1    -0.34414 -0.71414
B’         1     1.772    0
```
and:
```
CSC0 0    CSC1 0xEA34A400    CSC2 0x000599C5
```
Using CSC2 9:0, coef_cb_blu = 0x1C5 = 1 + 0xc5/0x100 = 1.76953125 (off by 0.0025)
Using CSC2 19:10, coef_cr_red = 0x167 = 1 + 0x67/0x100 = 1.40234375 (off by 0.000344)

So how did the spec calculate the fractional parts, 0xc5 and 0x67?

The coef_cb_blu fractional part comes from 0.772 * 0x100 = 197.632 which is 0xc5 when truncated and 0xc6 when rounded
The coef_cr_red fractional part comes from 0.402 * 0x100 = 102.912 which is 0x66 when truncated and 0x67 when rounded

So it looks like spec truncated for coef_cb_blu and rounded for coef_cr_red. Or perhaps rounded for both but used a coefficient with less precision.

The script produces the rounded version for both. I think changing them to a reproducible value makes sense.

pelwell:
This is a low risk patch, and if Hias is happy then so am I.

pelwell:
Although:
> The script produces the rounded version for both. I think changing them to a reproducible value makes sense.

It's not clear whether this patch is the "changing" or if another change is pending.

popcornmix:
The script and values in this PR use rounded coefficients:
https://gist.github.com/popcornmix/6b3e23103c60170b02b148e0ba5d6ed7#file-csc-py-L38

The previous values (which came from spec) seem to have inconsistent rounding.
I've asked Broadcom if they know the origin of the spec values, but no one remaining knows.

pelwell:
So for `reproducible` read `consistent`.

popcornmix:
@mripard This is another piece needed for correct HDR output upstream.

HiassofT:
FYI: noggin checked with an HDMI grabber and verified that the colors are correct https://forum.libreelec.tv/thread/23442-rpi4-testbuild-with-hdr-support/?postID=154099#post154099 - thanks a lot @popcornmix, great work!

PR #4277, opened 2021-04-12
staging: vcsm-cma: Fix memory leak from not detaching dmabuf

davidplowman:
When importing there was a missing call to detach the buffer,
so each import leaked the sg table entry.

Actually the release process for both locally allocated and
imported buffers is identical, so fix them to both use the same
function.


davidplowman:
@6by9 @naushir 
This is Dave's patch for some kernel memory leaks from the 5.4 branch but which never made it onto 5.10.

6by9:
I'm deferring to David P's testing here.

The cherry-pick wasn't clean due to kmap and kunmap having been dropped from the dmabuf API, otherwise it's a straight lift of https://github.com/raspberrypi/linux/commit/2ae9829c3da54597a4701285f91985c4d08ee793

PR #4278, opened 2021-04-13
Shared clock improvements

mripard:
Hi,

Here's an implementation of the mechanism we discussed by mail to raise a clock shared between devices and make it return to its initial rate once done.

This has followed a discussion with upstream's maintainer, and has been sent here:
https://lore.kernel.org/linux-clk/20210413101320.321584-3-maxime@cerno.tech/

We're still in the early stage of review, so I'd expect it to change, but feel free to test it across more devices

### DIFF
--- a/drivers/gpu/drm/vc4/vc4_hdmi.c
+++ b/drivers/gpu/drm/vc4/vc4_hdmi.c
@@ -861,7 +861,9 @@ static void vc4_hdmi_encoder_post_crtc_powerdown(struct drm_encoder *encoder,
 		   HDMI_READ(HDMI_VID_CTL) & ~VC4_HD_VID_CTL_ENABLE);
 
 	clk_disable_unprepare(vc4_hdmi->pixel_bvb_clock);
+	clk_request_done(vc4_hdmi->bvb_req);
 	clk_disable_unprepare(vc4_hdmi->hsm_clock);
+	clk_request_done(vc4_hdmi->hsm_req);
 	clk_disable_unprepare(vc4_hdmi->pixel_clock);
 
 	ret = pm_runtime_put(&vc4_hdmi->pdev->dev);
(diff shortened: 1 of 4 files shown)


popcornmix:
With kms driver 1080p60 and without your patch, results in:
```
pi@pi4d:~ $ vcgencmd measure_clock core
frequency(1)=500000992
```
(as core clock was boosted for mode set which never gets removed)

With your patch:
```
pi@pi4d:~ $ vcgencmd measure_clock core
frequency(1)=267231440
```
So it looks good to me on a quick test.


EDIT: Actually I think this issue is still present. The fix was from another commit I'd been using in test branch:
https://github.com/popcornmix/linux/commit/c04984603fdd19051e8882bd487909d9d3cc2efa

I think that call needs to switch to new api too.

popcornmix:
I think this is desired too. Does the same thing for the core clock boost and in testing stops the core clock getting stuck high.
https://github.com/popcornmix/linux/commit/aa5a68d6d4ad1b2742d559ac81913f8d5f9cf586

popcornmix:
Not much response from upstream merge request. But I think this fixes some issues we currently have with clock usage.

drivers/gpu/drm/vc4/vc4_hdmi.c
```diff
@@ -1185,10 +1187,11 @@ static void vc4_hdmi_encoder_pre_crtc_configure(struct drm_encoder *encoder,
 	 * FIXME: When the pixel freq is 594MHz (4k60), this needs to be setup
 	 * at 300MHz.
 	 */
-	ret = clk_set_min_rate(vc4_hdmi->pixel_bvb_clock,
-			       (hsm_rate > VC4_HSM_MID_CLOCK ? 150000000 : 75000000));
-	if (ret) {
-		DRM_ERROR("Failed to set pixel bvb clock rate: %d\n", ret);
+	vc4_hdmi->bvb_req = clk_request_start(vc4_hdmi->pixel_bvb_clock,
+					      (hsm_rate > VC4_HSM_MID_CLOCK ? 150000000 : 75000000));
+	if (IS_ERR(vc4_hdmi->bvb_req)) {
```
popcornmix, line 1192: This breaks Pi3 which doesn't have a vc4_hdmi->pixel_bvb_clock and so this returns `-EINV`
The other clock function (`clk_set_rate`/`clk_set_min_rate`) return `0` (success) when called with a null clock. I wonder if this should?
e.g. `clk_request_start` returns null if `clk` is null, and clk_request_done does nothing with a null `clkreq`. 
Protecting this with a null check would also work (but is an extra thing that may be unexpected when converting from `clk_set_min_rate`).

pelwell:
There is some more clock strangeness observed in this forum thread: https://www.raspberrypi.org/forums/viewtopic.php?f=98&t=310123
The kernel clock driver ends up with an incorrect view of the core clock speed; its version matches the value set by "core_freq=", but the real core clock differs.

popcornmix:
Isn't that from a firmware/kernel that predates this PR?

pelwell:
That problem is present with and without this PR, so my reply is possibly interesting but off-topic.

mripard:
I had a look and it looks like something's off with the firmware interface. In the kernel, the rate returned in raspberrypi_fw_get_rate is (in my case) 200MHz, meanwhile `vcgencmd measure_clock core` would return around 266MHz

pelwell:
Yes - see my "off topic" comment.

PR #4281, opened 2021-04-15
Remove BCM2835 SPI chipselect limit.

jabjoe:
The limit of 4 chipselects for the BCM2835 was not required and also was
not inforced. Without inforcement it was possible to make a device tree
over this limit which would trample memory.

The chipselect count is now obtained from the device tree and expanded
if more devices are added.

Signed-off-by: Joe Burmeister <joe.burmeister@devtank.co.uk>

### DIFF
--- a/drivers/spi/spi-bcm2835.c
+++ b/drivers/spi/spi-bcm2835.c
@@ -28,6 +28,7 @@
 #include <linux/gpio/consumer.h>
 #include <linux/gpio/machine.h> /* FIXME: using chip internals */
 #include <linux/gpio/driver.h> /* FIXME: using chip internals */
+#include <linux/of_gpio.h>
 #include <linux/of_irq.h>
 #include <linux/spi/spi.h>
 
@@ -134,7 +135,8 @@ struct bcm2835_spi {
 	int tx_prologue;
 	int rx_prologue;
 	unsigned int tx_spillover;
-	u32 prepare_cs[BCM2835_SPI_NUM_CS];
+	unsigned int allocated_cs_num;
+	u32 *prepare_cs;
 
 	struct dentry *debugfs_dir;
 	u64 count_transfer_polling;
@@ -147,9 +149,9 @@ struct bcm2835_spi {
 	unsigned int rx_dma_active;
 	struct dma_async_tx_descriptor *fill_tx_desc;
 	dma_addr_t fill_tx_addr;
-	struct dma_async_tx_descriptor *clear_rx_desc[BCM2835_SPI_NUM_CS];
+	struct dma_async_tx_descriptor **clear_rx_desc;
 	dma_addr_t clear_rx_addr;
-	u32 clear_rx_cs[BCM2835_SPI_NUM_CS] ____cacheline_aligned;
+	u32 *clear_rx_cs;
 };
 
 #if defined(CONFIG_DEBUG_FS)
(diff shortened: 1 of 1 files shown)


pelwell:
How does the SPI subsystem support the runtime addition of chip selects?

Regardless of your answer, a significant change like this ought to go upstream. I think @l1k might have some input.

jabjoe:
It does it in two ways. First, on init, instead of just using the hard coded limit, it takes the number of children from the device tree, if that is bigger that is. Second, if a new device is added, it basically reallocs the arrays where this limit is coming from.

This is a solution, I'm not wedded to it. Any is better than none. The other option would be to enforce the limit, but that breaks our hardware so we'd need to patch anyway.

pelwell:
Increasing via DT I understand. What I'm asking is how a user (or anything) declares a random GPIO to be a CS line after the probe function as completed.

jabjoe:
It would have to be a device tree overlay. It would be possible to do one parametrized enough for the same one could be used for N spidev devices. For device tree overlays of specific devices instead of spidev, they would need the CS parametrized similarly.

l1k:
@jabjoe: Enforcing the limit makes sense. But why does that break your hardware? How many slaves do you have attached?

jabjoe:
We have 2 UARTs and 3/6 EEPROMs depending on config. Basically it's a card system and each card could have an EEPROM.

I started going through the other drivers to look for the same issue.
Seams it is a common problem.

There are two places in spi.c that do:

ctlr->num_chipselect = max_t(int, nb, ctlr->num_chipselect);

which changes what driver thinks it's limited to, to the count in the device tree.

The two I saw the same issue as the bcm SPI are:

spi-stm32-qspi.c
spi-tegra20-slink.c

The also have a define for supported chips attached and allocate a fixed array that size.

You can pretty much see the problem ones by grepping for "->chip_select]".

There are two I noticed that deal with the device tree child count at init/probe, but don't catch further dynamic expansion:

spi-omap2-mcspi.c
spi-gpio.c

Hope I'm missing something as I've only tested with the BCM driver.

For our user case, we don't really need dynamic count change support. We could just take bcm2835_spi_get_num_chipselect and allocated_cs_num and instead of doing the realloc, just error out when someone tries to add too many dynamically.

pelwell:
You are going to struggle to create more than about 24 CE lines, so a hybrid scheme that preallocates the arrays to that maximum size but only performs the DMA cyclic prep when bcm2835_spi_setup first uses a CE line is going to have minimal overhead while keeping the code simpler.

jabjoe:
An enforced 24 limit would certainly do us. So who's doing that? You want me to put it together? I'm writing to Mark Brown <broonie@kernel.org> and linux-spi@vger.kernel.org as the same issue is in a number of other drivers.

rwstephens:
A limit of 24 chipselects would work for me. -- This would likely solve problem seeing a kernel panic (bad kernel memory reference) in 5.11.0-1007-raspi after configuring 6 chipselects for for BCM2835 SPI via dtoverlay. Kernel panic is on shutdown/reboot, however, memory corruption for more than 4 chipselects likely has other runtime consequences.

pelwell:
05f8d58 raises the limit to 24 and checks that it isn't exceeded.

pelwell:
Closing, as the underlying issue has been addressed here and upstream.

l1k:
For reference, the upstream commits are 13817d466eb8 (band-aid fix) and ec679bda639f (long-term solution).

PR #4283, opened 2021-04-16
Prevent A4tech FHD 1080p webcams from being suspended automatically

P33M:
A variant of this camera has an annoying bug surrounding suspend/resume. These two commits fix it.


pelwell:
To require one patch might be considered unfortunate, but two is... _very_ unfortunate,

PR #4285, opened 2021-04-16
staging/bcm2835-codec: Do not update crop from S_FMT after res change

6by9:
During decode, setting the CAPTURE queue format was setting the crop
rectangle to the requested height before aligning up the format to
cater for simple clients that weren't expecting to deal with cropping
and the SELECTION API.
This caused problems on some resolution change events if the client
didn't also then use the selection API.

Disable the crop update after a resolution change.

### DIFF
--- a/drivers/staging/vc04_services/bcm2835-codec/bcm2835-v4l2-codec.c
+++ b/drivers/staging/vc04_services/bcm2835-codec/bcm2835-v4l2-codec.c
@@ -979,6 +979,13 @@ static void handle_fmt_changed(struct bcm2835_codec_ctx *ctx,
 
 	q_data->crop_width = format->es.video.crop.width;
 	q_data->crop_height = format->es.video.crop.height;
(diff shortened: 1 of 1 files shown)


6by9:
@popcornmix @jc-kynesim

6by9:
@popcornmix @jc-kynesim Are you happy with this for making the new FFmpeg world good?

popcornmix:
What was the sample file you were testing with?

6by9:
The one that FFmpeg thought was 1920x1080, whilst the codec viewed it as 1920x1088. le10-nightly-pi4-does-not-play-testvid.mp4 IIRC.
(Hmm, VLC views it as 1920x1090, with display res of 1920x1080. Something funny with the headers in that one then).

jc-kynesim:
Yes I am happy with this.  I believe it makes our behaviour closer to spec besides anything else.

PR #4288, opened 2021-04-17
RFC: vchiq_arm: avoid crashing the kernel

lategoodbye:
Hi,
there some TODOs to get vchiq out of staging. One thing complained by checkpatch is the usage of BUG_ON. Using BUG_ON in a non-essential driver isn't recommend. So we better avoid it by using the favorite version of WARN and try to bail out.

Without a deeper insight of the code this could be very naive. So this a RFC for those changes, which are intended for upstream.


lategoodbye:
@pelwell Any comments?

pelwell:
It all looks fine and uncontroversial.

PR #4292, opened 2021-04-19
overlays: spi-rtc: Add ds3232 and ds3234

pelwell:
Extend the spi-rtc overlay to support the ds3232 and ds3234 RTCs, as
well as adding parameters to select difference SPI controllers and
chip selects.

N.B. The default CS is now active-low - use the "cs_high" parameter to
override this.

### DIFF
--- a/arch/arm/boot/dts/overlays/README
+++ b/arch/arm/boot/dts/overlays/README
@@ -2689,7 +2689,17 @@ Params: <None>
 Name:   spi-rtc
 Info:   Adds support for a number of SPI Real Time Clock devices
 Load:   dtoverlay=spi-rtc,<param>=<val>
-Params: pcf2123                 Select the PCF2123 device
+Params: ds3232                  Select the DS3232 device
+        ds3234                  Select the DS3234 device
+        pcf2123                 Select the PCF2123 device
+
+        spi0_0                  Use spi0.0 (default)
+        spi0_1                  Use spi0.1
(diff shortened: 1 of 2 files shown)


pelwell:
@Kriechi I don't know if you still use the `spi-rtc,pcf2123` overlay, but this PR would invert (i.e. standardise) the sense of the CS line to make it active-low (the usual default). I would like to do this in order to regularise the use of this overlay, but I am aware that this would require current user of the overlay to append `,cs_high` to their config.txt lines.

Do you have any comments?

Kriechi:
@pelwell thanks for reaching out - It's been a while since I last used this overlay. I currently have no active project with a pcf2123.

There could potentially be a lot of other people and projects using it... so I don't want to claim that this is a safe change.
Any chance of detecting this during the package upgrade and emit a warning or fail the upgrade?

pelwell:
There are many routes to install our kernels - rpi-update, apt upgrade, self-build, etc. - and no common mechanism to give such a warning. It is possible to set cs_high for pcf2123 by default and add a `cs_low` parameter (which would normally by a no-op), or change the default for all RTCs, but neither of those options is very appealing.

pelwell:
After some discussion we've opted to go with this version of the overlay. It does mean that existing users of the pcf2123 will find that it stops working, but that should be quick and easy to diagnose and resolve.

PR #4294, opened 2021-04-19
Add support for RTL8153C, RTL8153D, RTL8156A, and RTL8156B USB to Ethernet adapters

Omoeba:
Support for the new Realtek USB to Ethernet adapters were merged into net-next and they work without issues on a pi 4. This is a useful addition to the rpi kernel because USB to Ethernet adapters are the only way to add an additional ethernet port and the RTL8156 is one of the only adapters capable of 2.5GbE.


pelwell:
We prefer to back-port commits rather than take wholesale updates of files. Please try #4295 and verify that it works for you.

pelwell:
Superseded by #4295.

PR #4309, opened 2021-04-27
gpio-poweroff: Remember the old poweroff handler

pelwell:
Keeping a copy of the old poweroff handler allows it to be restored
should this module be unloaded, but also provides a fallback if the
power hasn't been removed when the timeout elapses.

See: https://github.com/raspberrypi/rpi-eeprom/issues/330

### DIFF
--- a/drivers/power/reset/gpio-poweroff.c
+++ b/drivers/power/reset/gpio-poweroff.c
@@ -24,6 +24,7 @@ static struct gpio_desc *reset_gpio;
 static u32 timeout = DEFAULT_TIMEOUT_MS;
 static u32 active_delay = 100;
 static u32 inactive_delay = 100;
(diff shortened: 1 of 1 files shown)


timg236:
LGTM

popcornmix:
Fine by me

pelwell:
There is a small wrinkle - the gpio-poweroff driver is a built-in (and Kconfig declares the config setting as a boolean, not a tristate), meaning that rmmod won't work. If the overlay is loaded at runtime ("dtoverlay gpio-poweroff") then removing it is a case of running "dtoverlay -r gpio-poweroff". The only way (that I'm aware of) to disable the built-in gpio-poweroff is to patch the Device Tree on the fly to set its status to "disabled", but that's beyond the scope of this PR.

PR #3717, opened 2020-07-09
WIP: vc4_hdmi: Support HBR audio

popcornmix:
This allows HBR audio passthough on a Pi 4 (e.g. DTS-HD and TrueHD)

A patch to alsa-lib is also required as it doesn't format status bits correctly for HBR.
https://github.com/LibreELEC/LibreELEC.tv/pull/4470/commits/c70f34d7965233a4b65a9c31f5f52431ff1703bf

With this, and a suitable alsa vc4_hdmi.conf file, an app like kodi can output HBR passthrough.

PR #4311, opened 2021-04-28
Improvements to hdmi-codec and conversion of vc4-hdmi

mripard:
Hi,

Here's a rework of the HBR passthrough and channel mapping support for vc4-hdmi to rely on the hdmi-codec driver instead of duplicating some of its code.

I've checked that the reported ELD is still valid (using amixer -c0 cget numid=$ID), and iecset to dump the iec status.

I don't have the hardware to test this (I think?) so it would be great if someone could test and make sure nothing got broken in the process. I'm especially worried about the ordering of the calls that is mentioned here: https://github.com/raspberrypi/linux/pull/3717/commits/06a9ff06ad3435e67a8377bdc2e2985d6764c737

I couldn't find in Kodi, LibreElec or alsa-lib where the status setup was done and the order relative to the hw_params call so it might be broken at the moment.

Thanks!
Maxime

### DIFF
--- a/include/sound/hdmi-codec.h
+++ b/include/sound/hdmi-codec.h
@@ -34,6 +34,11 @@ struct hdmi_codec_daifmt {
 	unsigned int frame_clk_inv:1;
 	unsigned int bit_clk_master:1;
 	unsigned int frame_clk_master:1;
+	/* bit_fmt could be standard PCM format or
+	 * IEC958 encoded format. ALSA IEC958 plugin will pass
+	 * IEC958_SUBFRAME format to the underneath driver.
+	 */
+	snd_pcm_format_t bit_fmt;
 };
 
 /*
--- a/include/sound/hdmi-codec.h
+++ b/include/sound/hdmi-codec.h
@@ -65,11 +65,21 @@ struct hdmi_codec_ops {
 
 	/*
 	 * Configures HDMI-encoder for audio stream.
-	 * Mandatory
+	 * Having either prepare or hw_params is mandatory.
 	 */
 	int (*hw_params)(struct device *dev, void *data,
 			 struct hdmi_codec_daifmt *fmt,
 			 struct hdmi_codec_params *hparms);
+
+	/*
+	 * Configures HDMI-encoder for audio stream. Can be called
+	 * multiple times for each setup.
+	 *
+	 * Having either prepare or hw_params is mandatory.
+	 */
+	int (*prepare)(struct device *dev, void *data,
+		       struct hdmi_codec_daifmt *fmt,
+		       struct hdmi_codec_params *hparms);
 
 	/*
 	 * Shuts down the audio stream.
--- a/sound/soc/codecs/hdmi-codec.c
+++ b/sound/soc/codecs/hdmi-codec.c
@@ -480,6 +480,42 @@ static void hdmi_codec_shutdown(struct snd_pcm_substream *substream,
 	mutex_unlock(&hcp->lock);
 }
 
+static int hdmi_codec_fill_codec_params(struct snd_soc_dai *dai,
+					unsigned int sample_width,
+					unsigned int sample_rate,
+					unsigned int channels,
+					struct hdmi_codec_params *hp)
+{
+	struct hdmi_codec_priv *hcp = snd_soc_dai_get_drvdata(dai);
+	int idx;
+
+	/* Select a channel allocation that matches with ELD and pcm channels */
+	idx = hdmi_codec_get_ch_alloc_table_idx(hcp, channels);
+	if (idx < 0) {
+		dev_err(dai->dev, "Not able to map channels to speakers (%d)\n",
+			idx);
+		hcp->chmap_idx = HDMI_CODEC_CHMAP_IDX_UNKNOWN;
+		return idx;
+	}
+
+	memset(hp, 0, sizeof(*hp));
+
+	hdmi_audio_infoframe_init(&hp->cea);
+	hp->cea.channels = channels;
+	hp->cea.coding_type = HDMI_AUDIO_CODING_TYPE_STREAM;
+	hp->cea.sample_size = HDMI_AUDIO_SAMPLE_SIZE_STREAM;
+	hp->cea.sample_frequency = HDMI_AUDIO_SAMPLE_FREQUENCY_STREAM;
+	hp->cea.channel_allocation = hdmi_codec_channel_alloc[idx].ca_id;
+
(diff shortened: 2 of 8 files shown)


popcornmix:
I've just given this a test and TrueHD passthrough worked but DTS-HD (despite making DTS-HD led on AVR come on) was silent.
Possibly related to the iec_status bits, but @HiassofT knows more about this.

popcornmix:
I'm wondering about the AES bits set here: https://github.com/xbmc/xbmc/blob/0e7f7628cb5fe8453fedf801b12cc659d2eb7bcb/xbmc/cores/AudioEngine/Sinks/AESinkALSA.cpp#L453
I think they get passed to driver and affect whether we are in PCM or passthrough mode.

(getting them wrong tends to be implementation defined - some sinks will make a best effort go at making a sound, others will just go silent is anything seems wrong).

HiassofT:
We can't select between PCM and HBR frames in hw_params, as the alsa hook plugin passes on the IEC status bits after hw_params (and before prepare) - therefore I moved the code to prepare.

See also the commit message of my change https://github.com/raspberrypi/linux/commit/fd45fc3c28ef696d5d55861f5d98ee713a7e2dfb

sound/soc/codecs/hdmi-codec.c
```diff
@@ -278,6 +278,7 @@ struct hdmi_codec_priv {
 	bool busy;
 	struct snd_soc_jack *jack;
 	unsigned int jack_status;
+	u8 iec_status[4];
```
jernejsk: Maybe it would be better to have 5 byte array? Although fifth byte doesn't seem to be used usually, it's nice to have this option in shared codec (DW-HDMI driver uses that fifth byte).

mripard: Indeed, I changed it, thanks!

mripard:
I've just pushed a new version that implements prepare

include/sound/hdmi-codec.h
```diff
@@ -34,11 +34,6 @@ struct hdmi_codec_daifmt {
 	unsigned int frame_clk_inv:1;
 	unsigned int bit_clk_master:1;
 	unsigned int frame_clk_master:1;
-	/* bit_fmt could be standard PCM format or
-	 * IEC958 encoded format. ALSA IEC958 plugin will pass
-	 * IEC958_SUBFRAME format to the underneath driver.
-	 */
-	snd_pcm_format_t bit_fmt;
 };
```
jernejsk, line 42: Please don't do that, I'm using it in currently out of tree changes for DW-HDMI audio, so I was pretty happy to see it's added.

sound/soc/codecs/hdmi-codec.c
```diff
@@ -619,6 +662,20 @@ static const struct snd_soc_dai_ops hdmi_codec_spdif_dai_ops = {
 			 SNDRV_PCM_FMTBIT_IEC958_SUBFRAME_LE)
 
 struct snd_kcontrol_new hdmi_codec_controls[] = {
+	{
+		.access = SNDRV_CTL_ELEM_ACCESS_READ,
+		.iface = SNDRV_CTL_ELEM_IFACE_MIXER,
```
HiassofT: this should be SNDRV_CTL_ELEM_IFACE_PCM like in the IEC958 Playback Default and ELD controls below

mripard: Hi,
The doc seems to suggest otherwise though, from https://www.kernel.org/doc/html/latest/sound/kernel-api/writing-an-alsa-driver.html#iec958-s-pdif

> “IEC958 Playback Con Mask” is used to return the bit-mask for the IEC958 status bits of consumer mode. Similarly, “IEC958 Playback Pro Mask” returns the bitmask for professional mode. They are read-only controls, and are defined as MIXER controls (iface = SNDRV_CTL_ELEM_IFACE_MIXER).

HiassofT: Having mask and default controls on different interfaces would be really odd as these are closely related to each other.

Current kernel code shows mask/default controls are either all defined as mixer controls (eg hda) or pcm controls (eg ice1712 and intel hdmi).

see eg `git grep -B 2 "SNDRV_CTL_NAME_IEC958.*PLAYBACK" sound`

mripard: Some have the doc suggest too (ice1724, oxygen, trident). It's a bit of a mess, I'll try to clear it up while upstreaming :)

mripard:
I just pushed a new version addressing the comments from @jernejsk  and @HiassofT

popcornmix:
Just tested update PR and TrueHD passthrough is still silent for me. Removing this PR and audio is back.

HiassofT:
@popcornmix did you update vc4-hdmi.conf? https://github.com/HiassofT/LibreELEC.tv/commit/bbde777b672fdf84dc6317fc0df2c10f781b7173

popcornmix:
No - let me try that.

popcornmix:
Yes, that seems to be working for a range of passthrough formats and multichannel PCM.
Looks like a nice clean up.

@HiassofT I assume it was fine for you?

HiassofT:
Yes, it's fine for me.

Haven't tested the latest version (and can't test multichannel PCM output / channel mapping) but I glanced through the changes and they looked OK to me.

Might be worth sending out an RFC series and see what upstream ASoC/dri folks think about the hdmi-codec changes

mripard:
It's prepared already, I was waiting for your testing results to send it. If it works as expected for you, I'll send it in the afternoon

mripard:
I just did:
https://lore.kernel.org/dri-devel/20210507140334.204865-1-maxime@cerno.tech/T/#mab1b0e61ab14460321e455b1f28552a88ded3e5d

mripard:
The ASoC parts have been accepted upstream and are on their way to 5.13

popcornmix:
Does this PR still match upstream version? (not sure if it's been reworked since this PR's version).

mripard:
The commits are not identical in the ASoC part but it's only about the commit logs and some additional documentation I wrote in order to upstream. The code itself is identical.

The DRM side is fairly different but it's mostly due to fact that we had to revert the old code in the downstream tree, while we only had to add new code upstream.

popcornmix:
@pelwell I think this should be merged. I've been running with it for a while at home without issue and it's merged upstream.

pelwell:
Some of these don't apply cleanly to rpi-5.12.y (I've merged the ones that do).

Issue #4307, opened 2021-04-23
new kms driver doesn't detect 480i or work when forced.


alanbork:
the kms driver does not detect 480i. Tested with 2ac4de4eaac5c1d1b25acec4a5e0a9fdb16f0c91. Haven't checked if it woks with older versions.

kms: modetest | grep 480i
[nothing]
fkms: modetest | grep 480i
  #31 720x480i 30.00 720 739 801 858 480 488 494 525 13514 flags: nhsync, nvsync, interlace, dblclk; type: driver
  #32 720x480i 29.97 720 739 801 858 480 488 494 525 13500 flags: nhsync, nvsync, interlace, dblclk; type: driver

there's a bug in modetest that shows the refresh rate as half if the mode is interlaced, those really are 60hz/ntsc.

perhaps relatedly, kms fails if I force 480i by using the same modeline detected by fkms:

720x480i-60.00hz(60): 13.514000 720 739 801 858 480 488 494 525 -hsync -vsync interlace dblclk

it initializes kms/drm without errors but swapbuffer in a tight loop always returns after .28ms and nothing is shown on screen. 

it works if I instead I generate a modeline using cvt:

13.50  720 744 808 896  480 483 493 512 interlace -hsync +vsync

so maybe the issue is dblclk?

alanbork:
seems this exact problem was fixed for fkms a while ago: #3725.

6by9:
DRM works in frames not fields. 480i is 30 frames (60 fields) per second, and 33ms would therefore be correct for vsync timings (I'm assuming .28ms was a typo for 28ms).
You can not update the display between fields.

Your listed timing is totally different from the porch timings of the one parsed from the EDID, so I'm not sure of the relevance.

As per the [docs](https://www.raspberrypi.org/documentation/configuration/config-txt/pi4-hdmi.md), due to the 2-pixels/clock architecture, odd values in horizontal timings can not be supported.
```720x480i 30.00 720 739 801 858 480 488 494 525 13514``` and ```720x480i 29.97 720 739 801 858 480 488 494 525 13500``` will have been pruned due to this.  See https://github.com/raspberrypi/linux/blob/rpi-5.10.y/drivers/gpu/drm/vc4/vc4_hdmi.c#L1444

Fkms has a [specific conditional to allow such modes with DBLCLK](https://github.com/raspberrypi/linux/blob/rpi-5.10.y/drivers/gpu/drm/vc4/vc4_firmware_kms.c#L1093).
[...]

PR #4312, opened 2021-04-28
drm/vc4: Allow DBLCLK modes even if horz timing is odd.

6by9:
The 2711 pixel valve can't produce odd horizontal timings, and
checks were added to vc4_hdmi_encoder_atomic_check and
vc4_hdmi_encoder_mode_valid to filter out/block selection of
such modes.

Modes with DRM_MODE_FLAG_DBLCLK double all the horizontal
timing values before programming them into the PV. The PV values,
therefore, can not be odd, and so the modes can be supported.

Amend the filtering appropriately.

See https://github.com/raspberrypi/linux/issues/4307


6by9:
@mripard FYI Further mode tweaking changes.

PR #4326, opened 2021-05-06
usb: dwc2: Fix gadget DMA unmap direction

pelwell:
The dwc2 gadget support maps and unmaps DMA buffers as necessary. When
mapping and unmapping it uses the direction of the endpoint to select
the direction of the DMA transfer, but this fails for USB OUT transfers
because the unmap occurs after the endpoint direction has been reversed
for the status phase.

A possible solution would be to unmap the buffer before the direction
is changed, but a safer, less invasive fix is to remember the buffer
direction independently of the endpoint direction.

See: https://github.com/raspberrypi/Raspberry-Pi-OS-64bit/issues/127

### DIFF
--- a/drivers/usb/dwc2/core.h
+++ b/drivers/usb/dwc2/core.h
@@ -112,6 +112,7 @@ struct dwc2_hsotg_req;
  * @debugfs: File entry for debugfs file for this endpoint.
  * @dir_in: Set to true if this endpoint is of the IN direction, which
  *          means that it is sending data to the Host.
+ * @map_dir: Set to the value of dir_in when the DMA buffer is mapped.
  * @index: The index for the endpoint registers.
  * @mc: Multi Count - number of transactions per microframe
  * @interval: Interval for periodic endpoints, in frames or microframes.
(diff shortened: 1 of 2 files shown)


P33M:
Correct me if I'm wrong, but isn't the impact of the bug limited to Control OUT transfers?

pelwell:
Probably - the damage is being done by `dwc2_hsotg_ep0_zlp`, the comment for which says:
```
 * dwc2_hsotg_ep0_zlp - send/receive zero-length packet on control endpoint
 * @hsotg: The device instance
 * @dir_in: If IN zlp
 *
 * Generate a zero-length IN packet request for terminating a SETUP
 * transaction.
 *
 * Note, since we don't write any data to the TxFIFO, then it is
 * currently believed that we do not need to wait for any space in
 * the TxFIFO.
```

pavhofman:
The problem was really only with Control OUT transfers, the other transfers worked OK.

pelwell:
Merged with a tweaked commit message.

pavhofman:
Phill, please will you send the patch upstream? IIUC the problem is with all dwc2 implementations. Thanks lot!

pavhofman:
Phill, thanks for the post.

pelwell:
For the record: https://marc.info/?l=linux-usb&m=162030011712416&w=2

PR #4313, opened 2021-04-29
Interrupt-based HDMI Hotplug Detection

mripard:
Hi,

So far the HDMI driver was relying on polling every 10s for the HDMI connector status which could lead to issues when the cable was quickly unplugged and then plugged back, or when a device like an HDMI bridge would just create a short hotplug pulse.

Here's a PR that converts that logic to rely on either the controller interrupts (for the BCM2711) or the HPD gpio interrupt (for earlier SoCs) to remove that limitation.

PR #4327, opened 2021-05-06
HDMI Interrupt-Based Hotplug for Devices with a GPIO

mripard:
Hi, 
I'm pushing a draft PR here since Dave reported that the bank1 interrupts were apparently not working on the Pi2B, in PR https://github.com/raspberrypi/linux/pull/4313

The code here works on the Pi3B+, so it might be helpful to someone.

### DIFF
--- a/drivers/gpu/drm/vc4/vc4_hdmi.c
+++ b/drivers/gpu/drm/vc4/vc4_hdmi.c
@@ -484,9 +484,8 @@ vc4_hdmi_connector_detect(struct drm_connector *connector, bool force)
 	WARN_ON(clk_prepare_enable(vc4_hdmi->hsm_clock));
 
 	if (vc4_hdmi->hpd_gpio) {
-		if (gpio_get_value_cansleep(vc4_hdmi->hpd_gpio) ^
-		    vc4_hdmi->hpd_active_low)
-			connected = true;
+	    if (gpiod_get_value_cansleep(vc4_hdmi->hpd_gpio))
+		connected = true;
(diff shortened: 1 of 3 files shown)


mripard:
I just pushed a change that disables the interrupts-based hotplug on devices using a GPIO on a bank other than bank 0 (like the Pi2B). I don't have a Pi2B so I couldn't test on that hardware, but hacking a bit on the Pi3B+ looks like it should be working there.

6by9:
I'm happy with the patches, although I don't see why only bank 0 interrupts work. Or is that more to keep general GPIO interrupt handling working? That's got potential fall-out on the Compute Module 1&3 (not 4) where Bank 1 GPIOs are available for general use.

My concerns were that the firmware appears to be using the edge detection part of the interrupt block, even if it doesn't enable the actual interrupt and install an ISR.

This PR will make HPD work on those platforms with HPD on the GPIO expander (via polling), but I think I need to look at the firmware to make it drop all HPD stuff should KMS be enabled.

mripard:
The first patch was meant to prevent using the interrupt based hotplug on the devices that have a GPIO on bank > 0 since you reported it as broken. After this patch, as far as I know, the Pi with a GPIO Expander or a GPIO on bank 1 would use polling, and the one with a dedicated interrupt line (Pi4, CM4?) or a GPIO on the bank 0 (Pi3B+) would use interrupts

It's true that the side effect is that any driver using an interrupt handler would now get an error, but the way I understood it is that they were always broken anyway. If it's the firmware interfering, then we shouldn't merge this patch indeed.

PR #4334, opened 2021-05-11
Add bh1750 light sensor ccs811 air quality sensor to i2c-sensor overlay

dopefishh:
BH1750 is a commodity i2c light sensor.
CCS811 is a popular i2c air quality sensor.

Both are used a lot on raspberry pi's but not enabled by default in the kernel and there were no overlays available.

e.g.: https://learn.adafruit.com/adafruit-bh1750-ambient-light-sensor/python-circuitpython

and: https://learn.adafruit.com/adafruit-ccs811-air-quality-sensor/python-circuitpython

This MR adds an overlay and enables the modules in the default config for the raspberry pi variants


pelwell:
This PR required some patching up because the defconfigs were manually edited (`make savedefconfig` doesn't always use alphabetical order) and because it got overtaken by another i2c-sensor addition, but it's in rpi-5.10.y now and will be in future kernel builds.

PR #4335, opened 2021-05-11
Ov5647 fixes

davidplowman:


davidplowman:
@naushir @6by9 
The pixel array fix is I believe correct but doesn't actually cause any specific problems that I'm aware of. The VBLANK min value fixes libcamera lock-ups when using the full resolution mode. The value 24 is verified by trial and error, though I notice it's also the difference between the frame length and the vts_def, which is a number we will have obtained from OV many aeons ago.

naushir:
24 does seem like the right number based on the default value - which is typically provided as the fastest fps of the mode.

LGTM.

pelwell:
Any concerns, @6by9?

6by9:
Both seem reasonable to me.
Does it still pass v4l2-compliance (please ensure you build the latest version from source)?

davidplowman:
Sure, I can try that. What's the correct way to run it, just `v4l2-compliance -d 0`? Or is that testing unicam and not the sensor driver?

6by9:
I believe you can run it against a sub-dev. ```-u 0``` appears to be the runes required.

davidplowman:
Ah, thanks. I seem to get one failure:
`		fail: v4l2-test-controls.cpp(871): subscribe event for control 'User Controls' failed
	test VIDIOC_(UN)SUBSCRIBE_EVENT/DQEVENT: FAIL
`
which I'm guessing must have been pre-existing...?

davidplowman:
Hmm, it does seem to be missing the "subscribe_event" and "unsubscribe_event" handlers in the v4l2_subdev_core_ops that our other drivers seem to have. If that sounds like the right fix then I could add them as another commit and re-push this set?

6by9:
As you say, that was a pre-existing issue, so yes it would be a further commit.
No rush in my book for it, but I think you're right that it's close to a two-liner (adding the subscribe_event and unsubscribe_event handlers using the standard core functions).

davidplowman:
Well, let me try that and see if we can iron that out too. It'll be tomorrow now!

davidplowman:
I've added that extra commit. v4l2-compliance happy now, and libcamera still working!

PR #4342, opened 2021-05-12
i2c0 dtoverlay fixes

6by9:
Hopefully the commits are self-explanatory.

### DIFF
--- a/arch/arm/boot/dts/overlays/i2c0-overlay.dts
+++ b/arch/arm/boot/dts/overlays/i2c0-overlay.dts
@@ -8,6 +8,7 @@
 		target = <&i2c0if>;
 		__overlay__ {
 			status = "okay";
+			pinctrl-names = "default";
 			pinctrl-0 = <&i2c0_pins>;
 		};
 	};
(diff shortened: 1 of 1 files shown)


pelwell:
I believe that fragment 8 could be replaced with:
```
        __exports__ {
                i2c0;
        };
```
Although the exports section was intended to add new symbols to the base DTB, I don't think there's anything to stop it replacing an exsting symbol.

Having said that, your version has a symmetry with the aliases patch, and it's clear that something magic is occuring - an export is perhaps less obvious in that respect.

6by9:
You're more knowledgeable on DT than me.

Will ```__exports__``` work when ```i2c0``` already exists as a node from the base DT (pointing at ```soc/i2c0mux/i2c@0```)?

pelwell:
> I don't think there's anything to stop it replacing an existing symbol.

6by9:
Minor annoyance that it seems that ```sudo dtoverlay i2c0``` registers the node as /dev/i2c-11 (or whatever the next unaliased node is), it doesn't get routed to /dev/i2c-0.
Loading from config.txt sends it to /dev/i2c-0.

6by9:
> > I don't think there's anything to stop it replacing an existing symbol.

Apologies - missed that line first time around.

PR #4355, opened 2021-05-20
add sensirion sgp30 to i2c-sensor overlay

mehmetahsen:
Similar to #3517


pelwell:
I'm happy with this as it stands, but it's less useful without the corresponding defconfig change to add `CONFIG_SENSIRION_SGP30=m`. Would you like me to take care of that?

pelwell:
See https://github.com/raspberrypi/linux/pull/4355/commits/fa62b629aeacea19a794f02470ae7a1068ff8be9.

mehmetahsen:
> See [fa62b62](https://github.com/raspberrypi/linux/commit/fa62b629aeacea19a794f02470ae7a1068ff8be9).

Thanks.

Edit: Just realized PR title and commit message refers to the sensor as spg30, whereas it should have been sgp30. I changed the PR title, and commit message could be changed with a squash & merge?

pelwell:
I've patched up the commit message and force-pushed - it's better for us to keep the overlay commits separate from the config commits, because they get squashed separately.

PR #4356, opened 2021-05-21
drm/vc4: FKMS: Change of Broadcast RGB mode needs a mode change

6by9:
The Broadcast RGB (aka HDMI limited/full range) property is only
notified to the firmware on mode change, so this needs to be
signalled when set.

https://github.com/raspberrypi/firmware/issues/1580


popcornmix:
This looks sane to me (based on how colourspace signalling was done for kms).

PR #4366, opened 2021-05-27
Add Raspberry Pi PoE+ HAT support

XECDesign:
Signed-off-by: Serge Schneider <serge@raspberrypi.com>


pelwell:
See #4367 for a slight reworking to split the commits and add documentation for the overlay.

PR #4367, opened 2021-05-27
PoE HAT Plus support

pelwell:
This is a minor rework of #4366, splitting the commits and adding the overlay documentation.


XECDesign:
Many thanks

PR #4372, opened 2021-05-31
Add device tree overlay for ssd1331-spi screen

alexander-kurichenko:
Connected waveshare 0.95inch RGB OLED with ssd1331 controller


pelwell:
Added README entry and merged offline.

PR #4374, opened 2021-06-03
overlays: hifiberry-dac: covert to using simple-audio-card

nullr0ute:
The TI pcm5102a works well with the simple-audio-card which has
the added advantage of working with upstream kernels so migrate
the HiFiBerry DAC over to using it.

### DIFF
--- a/arch/arm/boot/dts/overlays/hifiberry-dac-overlay.dts
+++ b/arch/arm/boot/dts/overlays/hifiberry-dac-overlay.dts
@@ -8,6 +8,7 @@
 	fragment@0 {
 		target = <&i2s>;
 		__overlay__ {
+			#sound-dai-cells = <0>;
 			status = "okay";
 		};
 	};
@@ -15,7 +16,7 @@
 	fragment@1 {
 		target-path = "/";
 		__overlay__ {
-			pcm5102a-codec {
+			codec_pcm: pcm5102a-codec {
 				#sound-dai-cells = <0>;
 				compatible = "ti,pcm5102a";
 				status = "okay";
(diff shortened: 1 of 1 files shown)


nullr0ute:
I've tested this with the Pimoroni Pirate Audio line out HAT which uses this overlay as well.

pelwell:
I'm concerned that this commit changes the card name (currently "HifiBerry DAC") and the stream name (currently "HifiBerry DAC HiFi"), which might break existing ALSA configurations (e.g. `asound.state`).

nullr0ute:
I have no issues updating that, I didn't want to spend a bunch of time upfront if it was going to be out right rejected. The card name should be straight forward, will dig into stream names.

Gadgetoid:
Counter-proposal: rather than change hifiberry-dac, would it be appropriate to create a *new* overlay specifically named for the pcm5102a?

As you mention- we use this overlay for Pimoroni audio products so having something with a more generic name makes at least some sense. Plus it clearly signals what DAC is readily supported should anyone want to take a shot at making an audio board.

nullr0ute:
> Counter-proposal: rather than change hifiberry-dac, would it be appropriate to create a _new_ overlay specifically named for the pcm5102a?

I'm working on a Pirate Audio HAT overlay separate to that to drive the buttons/display, I need to finish the driver pieces for the display.

> As you mention- we use this overlay for Pimoroni audio products so having something with a more generic name makes at least some sense. Plus it clearly signals what DAC is readily supported should anyone want to take a shot at making an audio board.

Quite happy to work on that, overall I figured it would be useful overall to the RPi maintainership, possibly more so with the new 5.15 LTS kernel, to look at moving from non upstream drivers to something that would take you less maintainer time, this works for me but I've not had time to dig into the naming differences in the generic sound options or if it even makes a real difference (I've not seen any in my not very wide use case).

Gadgetoid:
I think the winds are definitely changing - it feels very sudden,  but I haven't really had time to keep my finger on the pulse here - toward using more upstream methods and abandoning the RPi idiosyncrasies. For better or worse.

I'm not informed enough to have a strong opinion either way, but I'm glad to hear you're working on various overlays because the darn things flummox me. You're right regarding upstream, though, and simple-audio-card is probably the way forward.

It's nice to work with too, I managed to cobble together an overlay that would get our microphone/speaker boards working together - https://github.com/pimoroni/pirate-audio/blob/a9d92d452ea08e86f004b9c87cf63c270687431f/speaker-and-mic/adau7002-pcm5102a-io-overlay.dts

I'm now buried in figuring out how to approach upstream support for HyperPixel 4 with panel drivers and init. :grimacing: 

Thank you!

pelwell:
We will always favour upstream code over downstream, especially if all other things are equal. I fear that one day upstream will embrace overlays in a way that breaks (or at least makes awkward) our way of doing things, but until then I'm happy to advise on new overlays and maintain the existing ones.

PR #4385, opened 2021-06-11
media: i2c: imx477: Fixes for long exposure modes

naushir:
Do not scale IMX477_EXPOSURE_OFFSET with the long exposure factor during
the limit calculations. This allows larger exposure times, and does seem to be
what the sensor is doing internally.

When clipping the frame length calculations during long exposure, use the
maximum possible frame length and exposure factor.

### DIFF
--- a/drivers/media/i2c/imx477.c
+++ b/drivers/media/i2c/imx477.c
@@ -1282,7 +1282,7 @@ static void imx477_adjust_exposure_range(struct imx477 *imx477)
 
 	/* Honour the VBLANK limits when setting exposure. */
 	exposure_max = imx477->mode->height + imx477->vblank->val -
-		       (IMX477_EXPOSURE_OFFSET << imx477->long_exp_shift);
+		       IMX477_EXPOSURE_OFFSET;
 	exposure_def = min(exposure_max, imx477->exposure->val);
 	__v4l2_ctrl_modify_range(imx477->exposure, imx477->exposure->minimum,
 				 exposure_max, imx477->exposure->step,


naushir:
@6by9 and @davidplowman

drivers/media/i2c/imx477.c
```diff
@@ -1547,8 +1552,8 @@ static void imx477_set_framing_limits(struct imx477 *imx477)
 
 	/* Update limits and set FPS to default */
 	__v4l2_ctrl_modify_range(imx477->vblank, frm_length_min - mode->height,
-				 ((1 << IMX477_LONG_EXP_SHIFT_MAX) *
-					IMX477_FRAME_LENGTH_MAX) - mode->height,
+				 (IMX477_FRAME_LENGTH_MAX <<
+				      IMX477_LONG_EXP_SHIFT_MAX) - mode->height,
```
6by9: This made me scratch my head, as it's functionally the same.

drivers/media/i2c/imx477.c
```diff
@@ -1300,6 +1300,11 @@ static int imx477_set_frame_length(struct imx477 *imx477, unsigned int val)
 		val >>= 1;
 	}
 
+	if (imx477->long_exp_shift > IMX477_LONG_EXP_SHIFT_MAX) {
+		imx477->long_exp_shift = IMX477_LONG_EXP_SHIFT_MAX;
+		val = IMX477_FRAME_LENGTH_MAX;
+	}
+
```
6by9: Can this happen?

Called with ```val = (imx477->mode->height + ctrl->val)```.
```ctrl->val``` has a max range of 
```(IMX477_FRAME_LENGTH_MAX << IMX477_LONG_EXP_SHIFT_MAX) - mode->height```, so shouldn't it only ever be possible to get a max of ```IMX477_FRAME_LENGTH_MAX << IMX477_LONG_EXP_SHIFT_MAX```?

6by9 (commented):
I think we only actually have one hunk of actual change. That bit looks sane, but the other two ought to be justified, or be independent cleanup patches.

drivers/media/i2c/imx477.c
naushir: Yeah, I was just being ultra defensive.  I suppose i can remove this and let the framework do the verification on limits.

drivers/media/i2c/imx477.c
naushir: It's just a cleanup.  I'll split it to another patch.

davidplowman:
Nothing from me - all good!

6by9:
I'm happy with that change.
I just got confused in trying to review the others in the context of the commit text.

PR #4396, opened 2021-06-17
configs: add CONFIG_IR_TOY=m

davidschillerJKU:
This enables support for the Infrared Toy / Irdroid series of infrared
transceivers.

I think it would be quite useful to have this built by default.


popcornmix:
Looks reasonable.

pelwell:
It's fine, except it omits the 64-bit defconfigs.

davidschillerJKU:
Sorry, I amended the commit now to include the 64 bit configs.

pelwell:
Thanks!

PR #4398, opened 2021-06-17
media: i2c: imx290: Support 60fps in 2 lane operation

6by9:
Commit "97589ad61c73 media: i2c: imx290: Add support for 2 data lanes"
added support for running in two lane mode (instead of 4), but
without changing the link frequency that resulted in a max of 30fps.

Commit "98e0500eadb7 media: i2c: imx290: Add configurable link frequency
and pixel rate" then doubled the link frequency when in 2 lane mode,
but didn't undo the correction for running at only 30fps, just extending
horizontal blanking instead.
It also didn't update the CSI timing registers in accordance with the
datasheet.

Remove the 30fps limit on 2 lane by correcting the register config
in accordance with the datasheet for 60fps operation over 2 lanes.
Frame rate control (via V4L2_CID_VBLANK or HBLANK) can still reduce
the frame rate on 2 lanes back to 30fps.

### DIFF
--- a/drivers/media/i2c/imx290.c
+++ b/drivers/media/i2c/imx290.c
@@ -853,23 +853,9 @@ static inline u8 imx290_get_link_freq_index(struct imx290 *imx290)
 	return imx290->current_mode->link_freq_index;
 }
 
-static s64 imx290_get_link_freq(struct imx290 *imx290)
-{
-	u8 index = imx290_get_link_freq_index(imx290);
-
-	return *(imx290_link_freqs_ptr(imx290) + index);
-}
-
 static u64 imx290_calc_pixel_rate(struct imx290 *imx290)
 {
-	s64 link_freq = imx290_get_link_freq(imx290);
-	u8 nlanes = imx290->nlanes;
-	u64 pixel_rate;
-
-	/* pixel rate = link_freq * 2 * nr_of_lanes / bits_per_sample */
-	pixel_rate = link_freq * 2 * nlanes;
-	do_div(pixel_rate, imx290->bpp);
-	return pixel_rate;
+	return 148500000;
 }
 
 static int imx290_set_fmt(struct v4l2_subdev *sd,
--- a/drivers/media/i2c/imx290.c
+++ b/drivers/media/i2c/imx290.c
@@ -485,10 +485,10 @@ static const struct imx290_mode imx290_modes_2lanes[] = {
 		.lane_data = imx290_720p_2lane_settings,
 		.lane_data_size = ARRAY_SIZE(imx290_720p_2lane_settings),
 		.clk_data = {
-			[CLK_37_125] = imx290_37_125mhz_clock_1080p,
-			[CLK_74_25] = imx290_74_250mhz_clock_1080p,
+			[CLK_37_125] = imx290_37_125mhz_clock_720p,
+			[CLK_74_25] = imx290_74_250mhz_clock_720p,
 		},
-		.clk_size = ARRAY_SIZE(imx290_37_125mhz_clock_1080p),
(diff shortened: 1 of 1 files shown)


6by9:
@davidplowman Could you test this out please? It's common with IMX327 as well as IMX290.
I'm certainly getting sensible images out at 60fps in both modes.

I do need to check that the exposure and frame rate calcs all fall out correctly.

6by9:
wrt exposure and frame rate calcs, I reckon that the pixel rate is fixed at 148.5MPix/s, regardless of mode. The datasheet appears wrong in this regard, and I'm certainly getting not change in frame rate when changing between 10 and 12 bit, even though the link frequency is nominally the same.

6by9:
2nd commit pushed that sets the pixel rate at 148.5MPix/s.
I've not added a SoB tag to the patch, and added a WIP tag to the PR, as I'd like to verify that with a scope on the clock lines before merging.

The whole series of patches for imx290 really ought to be cleaned up and submitted upstream. Such joys.

davidplowman:
Sorry for the delay - been rather held up by setting up a new laptop! Anyway, it's looking rather good in libcamera. Both the 1080p and 720p modes frame at up to 60fps. When you ask for a particular framerate, you get it (up to 60fps, of course). The exposure times max out just below 1/framerate. So a definite thumbs up from me.

6by9:
No problem. I had the fun of a laptop change a couple of months back. At least VMs transfer across fairly easily.

Good news on it working with libcamera correctly.
I'll add a SoB to the patch changing pixel rate then, and look to come into the office probably tomorrow and put a scope on the module. Link frequencies aren't important to us, so them being wrong shouldn't really block this patchset.

Tested on the Vision Components IMX290 (mono) and IMX327 (colour) which have the 37.125MHz ext clock instead of 74.25MHz. Timings are fine there too.
VC's IMX290 also has all 4 data lanes wired up too. 1080p on 4 lanes doesn't stream in either 10 or 12bit modes, but 720p does (at 60fps in both), so it looks like I may have a register issue there. It doesn't work before these changes, so needs tracking back as to whether it is a downstream or mainline issue.

6by9:
Error found for imx290 4 lane 1080p. We've got a swap of clock setups at 
https://github.com/raspberrypi/linux/blob/rpi-5.10.y/drivers/media/i2c/imx290.c#L424 and
https://github.com/raspberrypi/linux/blob/rpi-5.10.y/drivers/media/i2c/imx290.c#L467 when added in https://github.com/raspberrypi/linux/commit/6b0c094a5b585ffa8fdee4dc12e7fd95bd80d87b

6by9:
OK, fixed up that issue, so both 2 & 4 lanes tested at 37.125Hz, and 2 lanes at 74.25MHz. I don't know of a module doing 4 lanes at 74.25. All report running both modes at 60fps in both 10 and 12 bit formats.

I think that makes me happy with the driver. WIP tag removed.

davidplowman:
Don't feel overly qualified to pronounce on all the register setting changes, *however*...

I've tested with both our usual Innomaker imx327 boards (running at 74.25MHz) and my Soho SE327M12 (running at 37.125MHz), both running with 2 lanes, of course. In all cases everything runs at the framerates requested and doesn't give silly exposure times. On both counts a substantial improvement on the previous version of this driver.

So it's a "LGTM" from me!

davidplowman:
Once this is pushed, we might want to reply to this: https://www.raspberrypi.org/forums/viewtopic.php?f=43&t=312967#p1872108

6by9:
> Once this is pushed, we might want to reply to this: https://www.raspberrypi.org/forums/viewtopic.php?f=43&t=312967#p1872108

It's merged, and I've just made a comment there.

PR #4404, opened 2021-06-23
drm/vc4: Fix timings for interlaced modes

### LINKED ISSUE #4331: corrupted display with kms/drm in interlaced modes
I've tested the new KMS driver (1340be471e2764cca8b9f8266e23347d1288919e) on 3 tvs, and on 2 there's no problem with interlaced modes. But on the third, a 42PX75U, it corrupts the image sent to the TV in a very strange way. I'd blame the TV but the same TV is fine with my pi4 when driven by fkms, as well as when driven by a pi0 with dispmanx.

this happens with 480i and 1080i. I'm only including shots from 480i since larger pixels make it easier to see. 

check the text that's rendered in a solid font, which is now missing some long horizontal segments (but not the entire line). it's almost artistic. 

kms distortion

and here's what it is supposed to look like.

fkms no problem

depending on the line this text is updated somewhere between never and once per second. so it's not a deinterlacing issue, I think.

kFYatek:
Fixes #4331 and an additional inconsistency in calculating vertical timing for interlaced modes. See discussion in that issue for details.

### DIFF
--- a/drivers/gpu/drm/vc4/vc4_crtc.c
+++ b/drivers/gpu/drm/vc4/vc4_crtc.c
@@ -355,7 +355,8 @@ static void vc4_crtc_config_pv(struct drm_crtc *crtc, struct drm_atomic_state *s
 				 PV_HORZB_HACTIVE));
 
 	CRTC_WRITE(PV_VERTA,
-		   VC4_SET_FIELD(mode->crtc_vtotal - mode->crtc_vsync_end,
+		   VC4_SET_FIELD(mode->crtc_vtotal - mode->crtc_vsync_end +
+				 interlace,
 				 PV_VERTA_VBP) |
 		   VC4_SET_FIELD(mode->crtc_vsync_end - mode->crtc_vsync_start,
 				 PV_VERTA_VSYNC));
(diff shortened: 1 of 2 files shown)


6by9:
See comments on #4331.

Agreed on the fix for PV_V_CONTROL.
The others need to be tallied to CEA-861 as it looks more like the rounding on the div2 of the sync timings that is incorrect, and adding one instead of fixing the round is likely to break other interlaced modes.

6by9:
As discussed on #4331, I'm satisfied that these changes are correct.

KMS has taken the VBP (which is an odd number) and divided it by 2. We need to add one to the odd field VBP to correct.
The firmware mode tables have the (odd) VBP value as (divide by 2 + 1), so that is subtracting one from the even field VBP. Copying the firmware behaviour resulted in the VBP that was 2 below where it should have been.

I haven't checked with an HDMI analyzer, but it should be good.

@pelwell over to you to review and merge.
I suspect this will raise a couple of conflicts against #4241, but we'll fix those later.

popcornmix:
@mripard This looks useful for upstream.

6by9:
> This looks useful for upstream.

It should be upstreamed once verified on an analyser.
Probably worth waiting for #4241 to be sorted too as it also messes with interlacing timings to add NTSC "odd field first" and other changes.

PR #4406, opened 2021-06-24
drm/vc4: Add support for non-standard modes in VEC

kFYatek:
This PR adds support for non-standard composite output modelines, specifically:
* **Progressive modes**. These are equivalent to firmware's `sdtv_mode=16`/`sdtv_mode=18` in terms of the resulting signal's timing, although the firmware modes employ implicit scaling (480/576 active lines from software scaled down to 240/288 lines on actual output). I added the `720x240` and `720x288` modes that are reported via `vc4_vec_connector_get_modes()`.
* **Custom interlaced modelines**. Some examples of what can be done with it:
  * `"720x486i" 13.5 720 734 798 858 486 491 497 525 Interlace` - NTSC with all the active lines from the original spec, before the convention of cropping the outermost 6
  * `"720x506i" 13.5 720 734 798 858 506 509 515 525 Interlace` - NTSC mode with basically the entire VBI accessible
  * `"720x610i" 13.5 720 740 804 864 610 612 618 625 Interlace` - the same but for PAL; you can e.g. emit teletext by drawing in the uppermost lines
  * `"720x480i" 13.5 720 740 804 864 480 486 492 525 Interlace` - an approximation of the "NTSC" mode while VEC is put in the 625-line mode; allows e.g. putting SECAM color on top of that, which is a really dumb idea but it works ;)
  * Basically any interlaced mode that has horizontal timings compatible with either the 525-line or 625-line mode, and no more lines than the standard, will work and generate a sane-looking signal. I don't know how could that be useful, but if it works, then why not. Unfortunately the progressive modes are much stricter, the VEC is locked to either 262 or 312 total lines, so the only thing you can do is change the proportions between active picture and VBI.
    * `"640x200" 13.5 640 694 758 858 200 223 226 262` is an example CGA-like mode with black borders.
* ~**Support for clocking the VEC at different frequencies than 108 MHz**. This also includes recalculating the color subcarrier frequencies, so that they stay within spec even in that case. I guess that this point will be the most controversial of it all, but it has some uses:~
  * ~For fun, you can try modes such as `"720x378i 8.68725 720 734 798 858 378 381 387 405 Interlace` - that in theory closely approximates the early British 405-line TV system and _may_ be able to drive such a set. Obviously I don't have hardware to check that ;)~
  * ~More importantly though, this makes the composite output somewhat usable even without `enable_tvout=1` on Pi4. Due to the clock being 0.8% slower than the spec, the ratio relationship between the line frequeny and subcarrier frequency is broken, which makes this mode cause a lot more cross-color artifacts, but the picture is still usable.~ Custom clock rate support has been removed, as it is nearly useless now that #4639 is merged.
  * As a side note, by using a [customized DT overlay](https://gist.github.com/kFYatek/7df69a03fd76fcccee41e78ddd9e6497) that enables both VEC and HDMI at the same time, I was able to get a working dual-screen HDMI+Composite setup on Pi4. Curiously, it works both with and without `enable_tvout=1` (although I don't know, maybe the HDMI timings aren't quite standard then?). So maybe having an option for this type of setup in `vc4-kms-v3d-pi4.dtbo` would make sense.
    * Actually, I was kind of hoping that triple screen (2xHDMI + Composite) setup would be possible, but unfortunately I get CRTC activation error if I try enabling composite while both HDMIs are enabled already ;)

### DIFF
--- a/drivers/gpu/drm/vc4/vc4_vec.c
+++ b/drivers/gpu/drm/vc4/vc4_vec.c
@@ -251,7 +251,8 @@ enum vc4_vec_tv_mode_id {
 };
 
 struct vc4_vec_tv_mode {
-	const struct drm_display_mode *mode;
+	const struct drm_display_mode *interlaced_mode;
+	const struct drm_display_mode *progressive_mode;
 	u32 config0;
 	u32 config1;
 	u32 custom_freq;
@@ -285,61 +286,81 @@ static const struct debugfs_reg32 vec_regs[] = {
 };
 
 static const struct drm_display_mode drm_mode_480i = {
-	DRM_MODE("720x480", DRM_MODE_TYPE_DRIVER, 13500,
+	DRM_MODE("720x480i", DRM_MODE_TYPE_DRIVER, 13500,
 		 720, 720 + 14, 720 + 14 + 64, 720 + 14 + 64 + 60, 0,
 		 480, 480 + 7, 480 + 7 + 6, 525, 0,
 		 DRM_MODE_FLAG_INTERLACE)
 };
 
+static const struct drm_display_mode drm_mode_240p = {
+	DRM_MODE("720x240", DRM_MODE_TYPE_DRIVER, 13500,
+		 720, 720 + 14, 720 + 14 + 64, 720 + 14 + 64 + 60, 0,
+		 240, 240 + 3, 240 + 3 + 3, 262, 0, 0)
+};
+
 static const struct drm_display_mode drm_mode_576i = {
-	DRM_MODE("720x576", DRM_MODE_TYPE_DRIVER, 13500,
+	DRM_MODE("720x576i", DRM_MODE_TYPE_DRIVER, 13500,
 		 720, 720 + 20, 720 + 20 + 64, 720 + 20 + 64 + 60, 0,
 		 576, 576 + 4, 576 + 4 + 6, 625, 0,
 		 DRM_MODE_FLAG_INTERLACE)
 };
 
+static const struct drm_display_mode drm_mode_288p = {
+	DRM_MODE("720x288", DRM_MODE_TYPE_DRIVER, 13500,
+		 720, 720 + 20, 720 + 20 + 64, 720 + 20 + 64 + 60, 0,
+		 288, 288 + 2, 288 + 2 + 3, 312, 0, 0)
+};
+
 static const struct vc4_vec_tv_mode vc4_vec_tv_modes[] = {
 	[VC4_VEC_TV_MODE_NTSC] = {
-		.mode = &drm_mode_480i,
+		.interlaced_mode = &drm_mode_480i,
+		.progressive_mode = &drm_mode_240p,
 		.config0 = VEC_CONFIG0_NTSC_STD | VEC_CONFIG0_PDEN,
 		.config1 = VEC_CONFIG1_C_CVBS_CVBS,
 	},
 	[VC4_VEC_TV_MODE_NTSC_J] = {
-		.mode = &drm_mode_480i,
+		.interlaced_mode = &drm_mode_480i,
+		.progressive_mode = &drm_mode_240p,
 		.config0 = VEC_CONFIG0_NTSC_STD,
 		.config1 = VEC_CONFIG1_C_CVBS_CVBS,
 	},
 	[VC4_VEC_TV_MODE_NTSC_443] = {
 		/* NTSC with PAL chroma frequency */
-		.mode = &drm_mode_480i,
+		.interlaced_mode = &drm_mode_480i,
+		.progressive_mode = &drm_mode_240p,
 		.config0 = VEC_CONFIG0_NTSC_STD,
 		.config1 = VEC_CONFIG1_C_CVBS_CVBS | VEC_CONFIG1_CUSTOM_FREQ,
 		.custom_freq = 0x2a098acb,
 	},
 	[VC4_VEC_TV_MODE_PAL] = {
-		.mode = &drm_mode_576i,
+		.interlaced_mode = &drm_mode_576i,
+		.progressive_mode = &drm_mode_288p,
 		.config0 = VEC_CONFIG0_PAL_BDGHI_STD,
 		.config1 = VEC_CONFIG1_C_CVBS_CVBS,
 	},
 	[VC4_VEC_TV_MODE_PAL_M] = {
-		.mode = &drm_mode_480i,
+		.interlaced_mode = &drm_mode_480i,
+		.progressive_mode = &drm_mode_240p,
 		.config0 = VEC_CONFIG0_PAL_M_STD,
 		.config1 = VEC_CONFIG1_C_CVBS_CVBS,
 	},
 	[VC4_VEC_TV_MODE_PAL_N] = {
-		.mode = &drm_mode_576i,
+		.interlaced_mode = &drm_mode_576i,
+		.progressive_mode = &drm_mode_288p,
 		.config0 = VEC_CONFIG0_PAL_N_STD,
 		.config1 = VEC_CONFIG1_C_CVBS_CVBS,
 	},
 	[VC4_VEC_TV_MODE_PAL60] = {
 		/* PAL-M with chroma frequency of regular PAL */
-		.mode = &drm_mode_480i,
+		.interlaced_mode = &drm_mode_480i,
+		.progressive_mode = &drm_mode_240p,
 		.config0 = VEC_CONFIG0_PAL_M_STD,
 		.config1 = VEC_CONFIG1_C_CVBS_CVBS | VEC_CONFIG1_CUSTOM_FREQ,
 		.custom_freq = 0x2a098acb,
 	},
 	[VC4_VEC_TV_MODE_SECAM] = {
-		.mode = &drm_mode_576i,
+		.interlaced_mode = &drm_mode_576i,
+		.progressive_mode = &drm_mode_288p,
 		.config0 = VEC_CONFIG0_SECAM_STD,
 		.config1 = VEC_CONFIG1_C_CVBS_CVBS,
 		.custom_freq = 0x29c71c72,
@@ -399,16 +420,32 @@ static void vc4_vec_connector_destroy(struct drm_connector *connector)
 static int vc4_vec_connector_get_modes(struct drm_connector *connector)
 {
 	struct drm_connector_state *state = connector->state;
-	struct drm_display_mode *mode;
+	struct drm_display_mode *interlaced_mode, *progressive_mode;
 
-	mode = drm_mode_duplicate(connector->dev,
-				  vc4_vec_tv_modes[state->tv.mode].mode);
-	if (!mode) {
+	interlaced_mode =
+		drm_mode_duplicate(connector->dev,
+				   vc4_vec_tv_modes[state->tv.mode].interlaced_mode);
+	progressive_mode =
+		drm_mode_duplicate(connector->dev,
+				   vc4_vec_tv_modes[state->tv.mode].progressive_mode);
+	if (!interlaced_mode || !progressive_mode) {
 		DRM_ERROR("Failed to create a new display mode\n");
+		drm_mode_destroy(connector->dev, interlaced_mode);
+		drm_mode_destroy(connector->dev, progressive_mode);
 		return -ENOMEM;
 	}
 
-	drm_mode_probed_add(connector, mode);
+	if (connector->cmdline_mode.specified &&
+	    connector->cmdline_mode.refresh_specified &&
+	    !connector->cmdline_mode.interlace)
+		/* progressive mode set at boot, let's make it preferred */
+		progressive_mode->type |= DRM_MODE_TYPE_PREFERRED;
+	else
+		/* otherwise, interlaced mode is preferred */
+		interlaced_mode->type |= DRM_MODE_TYPE_PREFERRED;
+
+	drm_mode_probed_add(connector, interlaced_mode);
+	drm_mode_probed_add(connector, progressive_mode);
 
 	return 1;
 }
@@ -589,7 +626,7 @@ static int vc4_vec_encoder_atomic_check(struct drm_encoder *encoder,
 					struct drm_connector_state *conn_state)
 {
 	const struct drm_display_mode *reference_mode =
-		vc4_vec_tv_modes[conn_state->tv.mode].mode;
+		vc4_vec_tv_modes[conn_state->tv.mode].interlaced_mode;
 
 	if (crtc_state->adjusted_mode.crtc_clock != reference_mode->clock ||
 	    crtc_state->adjusted_mode.crtc_htotal != reference_mode->htotal ||
--- a/drivers/gpu/drm/vc4/vc4_vec.c
+++ b/drivers/gpu/drm/vc4/vc4_vec.c
@@ -429,18 +429,11 @@ static int vc4_vec_connector_atomic_check(struct drm_connector *conn,
 	struct drm_connector_state *new_state =
 		drm_atomic_get_new_connector_state(state, conn);
 
-	const struct vc4_vec_tv_mode *vec_mode =
-		&vc4_vec_tv_modes[new_state->tv.mode];
-
-	if (new_state->crtc) {
+	if (new_state->crtc && old_state->tv.mode != new_state->tv.mode) {
 		struct drm_crtc_state *crtc_state =
 			drm_atomic_get_new_crtc_state(state, new_state->crtc);
 
-		if (!drm_mode_equal(vec_mode->mode, &crtc_state->mode))
-			return -EINVAL;
-
-		if (old_state->tv.mode != new_state->tv.mode)
-			crtc_state->mode_changed = true;
+		crtc_state->mode_changed = true;
 	}
 
 	return 0;
@@ -565,7 +558,10 @@ static void vc4_vec_encoder_enable(struct drm_encoder *encoder)
 	VEC_WRITE(VEC_CLMP0_START, 0xac);
 	VEC_WRITE(VEC_CLMP0_END, 0xec);
 	VEC_WRITE(VEC_CONFIG2,
-		  VEC_CONFIG2_UV_DIG_DIS | VEC_CONFIG2_RGB_DIG_DIS);
+		  VEC_CONFIG2_UV_DIG_DIS |
+		  VEC_CONFIG2_RGB_DIG_DIS |
+		  ((encoder->crtc->state->adjusted_mode.flags &
+		    DRM_MODE_FLAG_INTERLACE) ? 0 : VEC_CONFIG2_PROG_SCAN));
 	VEC_WRITE(VEC_CONFIG3, VEC_CONFIG3_HORIZ_LEN_STD);
 	VEC_WRITE(VEC_DAC_CONFIG, vec->variant->dac_config);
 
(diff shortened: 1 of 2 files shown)


kFYatek:
I've rebased new code now that #4241 is merged.

drivers/gpu/drm/vc4/vc4_vec.c
```diff
@@ -645,6 +645,9 @@ static void vc4_vec_encoder_enable(struct drm_encoder *encoder)
 		return;
 	}
 
+	eff_clk_rate = clk_get_rate(vec->clock);
+	DRM_DEBUG_DRIVER("Effective clock rate: %ld\n", eff_clk_rate);
```
6by9: @popcornmix Do you believe we'll have this level of control over the VEC clock? I thought we locked it at 108MHz for the VEC having ensured we were doing an integer divide to avoid jitter.

It's clock ```clocks = <&clocks BCM2835_CLOCK_VEC>```, so it is the bcm2835-cprman driver and not the firmware clocks.

pelwell:
There are some niggly whitespace complaints from checkpatch about the first commit:
```
0001-drm-vc4-Relax-VEC-modeline-requirements-and-add-prog.patch
---------------------------------------------------------------
ERROR: code indent should use tabs where possible
#59: FILE: drivers/gpu/drm/vc4/vc4_vec.c:449:
+^I^I^I            crtc_state->mode.crtc_vdisplay < 1 ||$

ERROR: code indent should use tabs where possible
#61: FILE: drivers/gpu/drm/vc4/vc4_vec.c:451:
+^I^I^I            crtc_state->mode.crtc_vsync_start != 3 ||$

ERROR: code indent should use tabs where possible
#63: FILE: drivers/gpu/drm/vc4/vc4_vec.c:453:
+^I^I^I            crtc_state->mode.crtc_vsync_end < 4 ||$

ERROR: code indent should use tabs where possible
#88: FILE: drivers/gpu/drm/vc4/vc4_vec.c:478:
+^I^I^I            crtc_state->mode.crtc_vdisplay < 1 ||$

ERROR: code indent should use tabs where possible
#90: FILE: drivers/gpu/drm/vc4/vc4_vec.c:480:
+^I^I^I            crtc_state->mode.crtc_vsync_start != 3 ||$

ERROR: code indent should use tabs where possible
#92: FILE: drivers/gpu/drm/vc4/vc4_vec.c:482:
+^I^I^I            crtc_state->mode.crtc_vsync_end < 2 ||$

total: 6 errors, 0 warnings, 101 lines checked

NOTE: For some of the reported defects, checkpatch may be able to
      mechanically convert to the typical style using --fix or --fix-inplace.

NOTE: Whitespace errors detected.
      You may wish to use scripts/cleanpatch or scripts/cleanfile
```
It wants the indentation to consist of as many TABs as possible, with spaces only at the end to pad out the remainder.

6by9 (commented):
2nd and third patches have checkpatch complaints too
```
CHECK: Lines should not end with a '('
#128: FILE: drivers/gpu/drm/vc4/vc4_vec.c:425:
+	interlaced_mode = drm_mode_duplicate(

CHECK: Lines should not end with a '('
#131: FILE: drivers/gpu/drm/vc4/vc4_vec.c:428:
+	progressive_mode = drm_mode_duplicate(

total: 0 errors, 0 warnings, 2 checks, 146 lines checked
```
break the line after the =, and then align the parameters to drm_mode_duplicate
```
CHECK: No space is necessary after a cast
#148: FILE: drivers/gpu/drm/vc4/vc4_vec.c:687:
+	chroma_freq += (125ull * (u64) eff_clk_rate) >> 1; /* proper rounding */
```
That one the space isn't needed between (u64) and eff_clk_rate, and doesn't aid readability.

As long as the standard modes come out to the same numbers as before then I'm happy with the patches.

pelwell:
The kernel submission guidelines do not specify `--strict` checking as a requirement. The netdev FAQ suggests it as an option, and even hwmon - the only subsystem to require it - says `There should be no errors, no warnings, and few if any check messages`, allowing some CHECKs.

drivers/gpu/drm/vc4/vc4_vec.c
popcornmix: The vec will always be 108MHz following https://github.com/raspberrypi/firmware/commit/1e5456a805804c01731bae3face5f151a9ab7e5
It used to be that the vec clock shared a PLL with core clock, which was affected by `enable_tvout` or `hdmi_enable_4kp60`, and the 108MHz was only accurate with `enable_tvout`.
Now it uses a dedicated PLL and will always be 108MHz, so I feel handling it not being 108MHz is redundant.

6by9: Sorry, I quoted a line which was reading back the clock having set it to ```8000 * adjusted_mode->clock``` a few lines earlier, thereby enabling support for pixel clocks other than 13.5MHz.

You say it's on a dedicated PLL, but is that shared with anything else? (Your link to the firmware change isn't working for me to check myself).

popcornmix: I've corrected the firmware commit. VEC clock will always be 108MHz.
It comes from PLLC which is fixed at 2592MHz (so always a multiple of 108MHz).
It's also used by CLK108 (which unsurprisingly also likes 108MHz) used by HDMI.

6by9: OK, thanks. I think that makes the last commit ineffective then.
AIUI the VEC will produce the pixel clock to drive the PV, which drives HVS, so if that only ever gets 108MHz then the pixel clock is fixed.

It is talking to bcm2835-cprman, which believes it knows how to configure the VEC clock. There's a comment of ```Only operating frequency is 108Mhz``` and PLLH_AUX.
That commit is for 2711 only. Is it feasible on earlier chips?

kFYatek: Just as a quick note: my experimental results were that pretty much any clock rate is precisely configurable on Pi3; on Pi4 (with the kernel and firmware from about a month ago, I believe) it was _roughly_ right, and rounded differently depending on the value of `enable_tvout` in `config.txt`.

Do the changes you mention apply to both chips? Maybe the ability to change the clock could be dependent on the compatible variant? Or maybe `vc4_vec_connector_atomic_check()` could check if `clk_round_rate()` returns a value that's reasonably close to the requested one?

But if that locked 108MHz value applies to every Pi version (BTW, how does that relate to `enable_tvout`?), then indeed, the last commit won't make any sense and needs to be scrapped.

Are those changes you're discussing publicly available already? Is it enough to just call `rpi-update` to get them?

popcornmix: You should get exactly 108MHz for the vec on any Pi and with any config.txt settings with recent firmware.
If you've run `rpi-update` since 24th March you will have this behaviour.

kFYatek: First of all, sorry for going silent on this for so long.

Anyway, I rebased this PR and started fresh with testing:
* I flashed freshly downloaded `2021-05-07-raspios-buster-armhf-full` image onto an SD card
* Updated everything via `apt`
* Ran `rpi-update` - `/boot/.firmware_revision` reads `bad5282375a2d1e4947432bd80a8839fbc3504ae` now
* Recompiled the kernel from the head of this PR as pushed just now and installed it

...and the results are the same as always. On Pi3, I can manipulate the pixel clock with quite an impressive precision; on Pi4, it is much more coarse, but I still have quite some control (that "405-line" mode from the example in the original post is more like 49 Hz instead of 50 Hz on Pi4, according to my oscilloscope). Without `enable_tvout=1` in config, the default modes are still about 0.8% slow, which causes unusable colour on the default branch, and this branch makes the colours look acceptable (albeit with more noticeable cross-colour artifacts than with `enable_tvout=1`).

kFYatek:
@pelwell @6by9 I rebased this PR and resolved the style issues reported by `checkpatch`. Sorry that I went silent on this for so long.

pelwell:
You're slightly outside my comfort zone with this PR, but it makes sense to me. I assume the change in `vc4_vec_connector_atomic_check` to use the `interlace_mode.htotal` is an arbitrary decision, made possible because `progressive_mode .htotal` is defined to be the same (the differences being in the vertical dimension)?

kFYatek:
> You're slightly outside my comfort zone with this PR, but it makes sense to me. I assume the change in `vc4_vec_connector_atomic_check` to use the `interlace_mode.htotal` is an arbitrary decision, made possible because `progressive_mode .htotal` is defined to be the same (the differences being in the vertical dimension)?

The VEC actually misbehaves if `htotal` is set to something else than 858 for the 525-line ("NTSC") or 864 for the 625-line ("PAL") mode. It generates the horizontal sync and blanking signals internally once every 858 or 864 pixels regardless of what it's being fed from the PV, and there doesn't seem to be any way of changing that.

Taking the above into account, attempting to set anything else as `htotal` results in an image that is skewed diagonally with a black diagonal bar in the middle (in the place where PV-generated blanking&sync are). I'm not near my development setup, so I can't make my own screenshots right now, but the effect is something like this: https://github.com/raspberrypi/firmware/issues/811#issuecomment-303446426

Conversely, VEC actually responds to changes in the vertical timing settings quite reasonably, at least in the interlaced mode. Manipulating this is bound to result in modes that are non-standard, but they might be useful in one way or another. But we need to ensure that `htotal` is set appropriately for the signal to make any sense at all.

drivers/gpu/drm/vc4/vc4_vec.c
popcornmix: Can you explain what mechanism you are using to manipulate pixel clock?
Note, the difference between Pi3 and Pi4 is the Pi3 will default to driving composite when there is no hdmi connected. Pi4 will only drive composite with `enable_sdtv=1`, so add `hdmi_force_hotplug=1` to both systems for a fair comparison.

The PLLs on Pi3 and Pi4 (with recent firmware) are both multiples of 108MHz, so you should be able to get an accurate 108Mhz clock on either if the VEC clock is configured. But it won't be configured unless someone has requested it.

The linux driver calls `clk_set_rate(vec->clock, 108000000)` but I think that is done using the (deprecated) clk-bcm2835 clock driver which is likely to be using the wrong PLL, so can't get an accurate 108MHz.  It should be switched to clk-raspberrypi, the newer firmware clock driver (used by core, pixel, arm. hvs etc) which does know which PLL to use.

kFYatek: To be honest, the clocking APIs is not something I'm too well versed in. I did not touch the _flow_ of clocking in the driver in any way, I did not attempt to thoroughly understand it; I just changed that hardcoded 108 MHz to a value derived from the modeline's pixel clock, saw it working to a reasonable extent and was satisfied with that.

Note that the official kernel behaves in the same way - on Pi4 without `enable_sdtv=1`, if you enable the KMS driver with composite enabled, the clock is not quite 108 MHz - but without the modifications I made here, it means broken colours.

Trying to analyse the unfamiliar code, it looks that the clock is selected in the DTS (for [bcm2835](https://github.com/raspberrypi/linux/blob/6237d09759ba6c8746cd1e19b16faee50c97bfac/arch/arm/boot/dts/bcm2835-common.dtsi#L112) and [bcm2711](https://github.com/raspberrypi/linux/blob/6237d09759ba6c8746cd1e19b16faee50c97bfac/arch/arm/boot/dts/bcm2711.dtsi#L306)) and indeed configured in [`clk-bcm2835.c`](https://github.com/raspberrypi/linux/blob/6237d09759ba6c8746cd1e19b16faee50c97bfac/drivers/clk/bcm/clk-bcm2835.c#L2218). There doesn't seem to be any difference in configuration between the BCM2711 and BCM2835 variants.

There seems to be the [`clk-raspberrypi.c`](https://github.com/raspberrypi/linux/blob/6237d09759ba6c8746cd1e19b16faee50c97bfac/drivers/clk/bcm/clk-raspberrypi.c) file you're referencing, but to be honest I have no idea what it does beyond acknowledging the existence of firmware clocks and how to make use of it.

I will do testing with `hdmi_force_hotplug=1` probably today in the evening to see whether that affects anything.

kFYatek: Something weird happens on my Pi3 with `hdmi_force_hotplug=1` - X starts fine on the composite output during the initial boot, but the driver freezes if I try to change modes using xrandr. Curiously, even attempting to set a mode that would be rejected by `vc4_vec_connector_atomic_check()` (causing `xrandr: Configure crtc 3 failed` when `hdmi_force_hotplug=0`), causes this freeze. In dmesg, there's only this:

```
[   82.394798] [drm:drm_atomic_helper_wait_for_flip_done [drm_kms_helper]] *ERROR* [CRTC:83:crtc-3] flip_done timed out
[   92.634957] [drm:drm_atomic_helper_wait_for_dependencies [drm_kms_helper]] *ERROR* [CRTC:83:crtc-3] flip_done timed out
[  102.875033] [drm:drm_atomic_helper_wait_for_dependencies [drm_kms_helper]] *ERROR* [CONNECTOR:49:Composite-1] flip_done timed out
[  113.115075] [drm:drm_atomic_helper_wait_for_flip_done [drm_kms_helper]] *ERROR* [CRTC:83:crtc-3] flip_done timed out
[  123.355119] [drm:drm_atomic_helper_wait_for_dependencies [drm_kms_helper]] *ERROR* [CRTC:83:crtc-3] flip_done timed out
[  133.595143] [drm:drm_atomic_helper_wait_for_flip_done [drm_kms_helper]] *ERROR* [CRTC:83:crtc-3] flip_done timed out
```

However, that also happens with the official kernel (both the newest one from `apt` and the newest one from `rpi-update`), so this is most likely orthogonal to what's going on in this PR.

----

Otherwise, there is not much difference regarding what we're talking about here. On Pi3, the clock is still perfectly steady (the colors look fine on the official kernel), and if I change the clock (which due to the bug above I can only do by hardcoding some other value in the module sources), it is respected precisely.

Pi4 boots up with perfect 13.5 MHz pixel clock when both `hdmi_force_hotplug=1` and `enable_tvout=1` are enabled. In fact, `hdmi_force_hotplug=1` seems to be completely ignored in this case, as I still get the rainbow test pattern on composite at boot.

However, I just discovered something curious that I overlooked previously. When I subsequently tried to change the pixel clock to 12.345 MHz, I ended up with something around 11.718 MHz instead, according to my measurements (I didn't check what the kernel reports, just measured the sync signals on output) - but that's to be expected given the coarser clock on Pi4. The curious part is that if I then try to go back to 13.5 MHz, I end up on that 13.393 or so MHz frequency - the same that is the result of `enable_tvout=0` - even though I had perfect 13.5 MHz at boot time. This is reproducible with `enable_tvout=1` regardless of whether `hdmi_force_hotplug` is enabled or not.

When `hdmi_force_hotplug=1` is set while `enable_tvout=0` on Pi4, the composite output comes up with vertical and horizontal sync patterns implying that imperfect 13.393 MHz pixel clock, but the driver freezes on a black screen and I never get to the desktop. `dmesg` shows similar problems as when attempting to switch modes on Pi3:

```
[   70.875413] [drm:drm_atomic_helper_wait_for_dependencies [drm_kms_helper]] *ERROR* [CRTC:65:crtc-1] flip_done timed out
[   81.115403] [drm:drm_atomic_helper_wait_for_dependencies [drm_kms_helper]] *ERROR* [CONNECTOR:45:Composite-1] flip_done timed out
[   91.355389] [drm:drm_atomic_helper_wait_for_dependencies [drm_kms_helper]] *ERROR* [PLANE:59:plane-1] flip_done timed out
[  101.595385] [drm:drm_atomic_helper_wait_for_flip_done [drm_kms_helper]] *ERROR* [CRTC:65:crtc-1] flip_done timed out
[  111.835388] [drm:drm_atomic_helper_wait_for_dependencies [drm_kms_helper]] *ERROR* [CRTC:65:crtc-1] flip_done timed out
[  122.075157] [drm:drm_atomic_helper_wait_for_dependencies [drm_kms_helper]] *ERROR* [CONNECTOR:45:Composite-1] flip_done timed out
[  132.312235] [drm:drm_atomic_helper_wait_for_dependencies [drm_kms_helper]] *ERROR* [PLANE:59:plane-1] flip_done timed out
[  142.550714] [drm:drm_atomic_helper_wait_for_flip_done [drm_kms_helper]] *ERROR* [CRTC:65:crtc-1] flip_done timed out
[  152.789930] [drm:drm_atomic_helper_wait_for_dependencies [drm_kms_helper]] *ERROR* [CRTC:65:crtc-1] flip_done timed out
```

Again, this happens even on the official kernel from `apt`, so it's not related to this diff at all.

And similar to a Pi3, if I hardcode a different clock speed in the kernel, it is respected at boot time, albeit with Pi4's normal lower accuracy (so 11.718 MHz instead of intended 12.345 MHz) - implied by sync signals on that blank screen, that is.

----

So, to sum up: we've discovered an unrelated bug, and a curious relation between `enable_tvout=1` and changing clock rates at runtime, but otherwise it still doesn't line up with that "locked 108MHz" behavior you seem to hope for.

kFYatek: By the way, here's how composite output with `enable_tvout=0` looks like on the current official kernels in [NTSC](https://user-images.githubusercontent.com/4499762/136106805-5881e5fc-36ce-41e8-a583-4f154ddf70d8.png) and [PAL](https://user-images.githubusercontent.com/4499762/136106855-3dfe100b-a360-4da0-9aed-22860b38700b.png) modes, at least through my grabber. The line frequency difference is little enough that the luminance component is clear, but colour is all over the place.

With this diff in place, the chroma subcarrier frequency is compensated in this case, which makes the image look mostly right, just with some more noticeable cross-colour artifacting, especially noticeable on fine details such as text: [`enable-tvout=0`](https://user-images.githubusercontent.com/4499762/136108048-548d98bd-8606-4e5c-8587-435f81f12457.png) vs. [`enable_tvout=1`](https://user-images.githubusercontent.com/4499762/136108299-d3ca7c0d-ccb4-451d-b214-f0f244e71ad4.png) (these are NTSC; there are far less artifacts with PAL anyway). You can note the slight image width difference on these screenshots, too.

kFYatek:
Note: In this newest revision, aside from rebasing onto current `rpi-5.10.y`, I moved the mode validation code from `vc4_vec_connector_atomic_check()` to `vc4_vec_encoder_atomic_check()`, which itself replaced `vc4_vec_encoder_mode_fixup()`. The new version now performs the validation on `adjusted_mode`. This has been done to avoid doing this move in #4636.

Both approaches seem to yield equivalent results from the userland perspective.

popcornmix:
@kFYatek can you test [this firmware](https://drive.google.com/uc?id=18CzpbXRLnEVJ0KvFMppor1qD7UV3W9Xo&export=download) and this [PR](https://github.com/raspberrypi/linux/pull/4639) and check behaviour of VEC clock?

I believe previously you'd get 107.143MHz if `enable_sdtv=1` was missing as firmware hadn't set up the VEC clock and the kernel code to do it was flawed. With updated fw/kernel it should get the correct 108MHz and it shouldn't be necessary to handle an inaccurate VEC clock.

kFYatek:
@popcornmix Yes, this indeed yields clean 108 MHz VEC clock on both Pi3 and Pi4, both with and without `enable_sdtv=1`.

Attempting to set the clock to something else than 108 MHz still does... something, but it looks like only some very simple fractions of that clock are supported - I could get what looked like 2/3, 3/4 and 6/7 of 108 MHz (i.e., 72 MHz, 81 MHz, 92.571... MHz), but nothing in between, so attempting that is pretty much useless.

So if you intend to merge that, I'll remove that one commit that adds custom clock support.

popcornmix:
Yes, that is expected. PLLC is set to 108 * 24, and PLLC_CHAN_CPER is a divide by 4. 
so you can get clean frequencies that are integer dividers of that, but there's not a lot of control there.

I'll get the firmware and kernel patches pushed.

kFYatek:
This newest force-push includes the following changes:

* Rebase onto current state of `rpi-5.10.y`
* Removal of the "drm/vc4: Make the VEC clock adjustable" commit, as it is useless after having #4639 merged
* Instead, the requirement that the custom mode's clock needs to be equal to the reference one has been added in `vc4_vec_encoder_atomic_check()`
* `reference_mode` local variable has been introduced in `vc4_vec_encoder_atomic_check()`, to make the comparisons cleaner
* The `switch` for validating either 525-line-based or 625-line-based mode parameters has been changed to switch over `reference_mode->vtotal` so that the cases are now 525 and 625, which is probably cleaner than switching on `htotal`

drivers/gpu/drm/vc4/vc4_vec.c
kFYatek: Custom clock rate support is now gone, so I think this thread can be resolved.

popcornmix:
Tested and it appears to work. The following modes gave plausible output:
video=Composite-1:720x480@60i
video=Composite-1:720x576@50i
video=Composite-1:720x240@60
video=Composite-1:720x288@50

PR #4402, opened 2021-06-21
HVS Core Clock Request Fix

mripard:
Hi,

Here's a fix for the breakage introduced by #4365. This significantly reworks the current code to drop the new, buggy, accessors and use the load tracker infrastructure that was already in place.

I haven't tested kmstest --flip yet, but the other issues seem to be gone.

PR #4408, opened 2021-06-24
drm/vc4: Don't create hvs_load_tracker on fkm

popcornmix:
fkms doesn't use vc4->hvs so protect against that


popcornmix:
Fixes fkms regression from https://github.com/raspberrypi/linux/pull/4402

@6by9 @timg236 @mripard

pelwell:
A better title would be `drm/vc4: Don't create hvs_load_tracker on fkms`, but otherwise LGTM.

popcornmix:
Updated title/commit message

6by9:
Fine by me.

timg236:
I missed this early but LGTM

PR #4409, opened 2021-06-24
media: rpivid: Fix H265 aux ent reuse of the same slot

jc-kynesim:
It is legitimate, though unusual, for an aux ent associated with a slot
to be selected in phase 0 before a previous selection has been used and
released in phase 2. Fix such that if the slot is found to be in use
that the aux ent associated with it is reused rather than an new aux
ent being created. This fixes a problem where when the first aux ent
was released the second was lost track of.

This bug spotted in Nick's testing. It may explain some other occasional,
unreliable decode error reports where dmesg included "Missing DPB AUX
ent" logging.

### DIFF
--- a/drivers/staging/media/rpivid/rpivid_h265.c
+++ b/drivers/staging/media/rpivid/rpivid_h265.c
@@ -371,7 +371,8 @@ static void aux_q_free(struct rpivid_ctx *const ctx,
 	kfree(aq);
 }
 
-static struct rpivid_q_aux *aux_q_alloc(struct rpivid_ctx *const ctx)
+static struct rpivid_q_aux *aux_q_alloc(struct rpivid_ctx *const ctx,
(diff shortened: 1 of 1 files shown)


pelwell:
I don't know the H265 hardware or driver, but the change makes sense.

jc-kynesim:
I'm not honestly expecting anyone other than myself to understand the why of this.  Its a tricky bit of buffer tracking that only applies to the internals of the driver.

njhollinghurst:
This fixes an issue where frame decode failures were occasionally reported for some 4K streams in an unusual test case (where the output was not format-converted, displayed or dumped; and with ffmpeg using a small number of threads).

jc-kynesim:
I think some of it was also because you were sending output to null which recycled the buffers faster than in most real cases.  The issue was definitely timing dependant.

PR #4410, opened 2021-06-24
drm: vc4: Fix pixel-wrap issue with DVP teardown

timg236:
@popcornmix @mripard @6by9 

Experimental patches to avoid the dreaded pixel wraparound bug (not really a PR!)

Tweak the order of operations to match the VC firmware.
On C0 use dedicate HW assisted fix

Previously, with a Dell 4K monitor (p30) it would fail nearly every reboot. Now it only seems to fail in a mode-test when switching from an interlaced mode back to a regular mode.

Test on Pi4B 1.1

### DIFF
--- a/drivers/gpu/drm/vc4/vc4_hdmi.c
+++ b/drivers/gpu/drm/vc4/vc4_hdmi.c
@@ -628,12 +628,12 @@ static void vc4_hdmi_encoder_post_crtc_disable(struct drm_encoder *encoder,
 
 	HDMI_WRITE(HDMI_RAM_PACKET_CONFIG, 0);
 
-	HDMI_WRITE(HDMI_VID_CTL, HDMI_READ(HDMI_VID_CTL) |
-		   VC4_HD_VID_CTL_CLRRGB | VC4_HD_VID_CTL_CLRSYNC);
+	HDMI_WRITE(HDMI_VID_CTL, HDMI_READ(HDMI_VID_CTL) | VC4_HD_VID_CTL_CLRRGB);
+
+	mdelay(1);
 
 	HDMI_WRITE(HDMI_VID_CTL,
-		   HDMI_READ(HDMI_VID_CTL) | VC4_HD_VID_CTL_BLANKPIX);
-
+		   HDMI_READ(HDMI_VID_CTL) & ~VC4_HD_VID_CTL_ENABLE);
 	vc4_hdmi_disable_scrambling(encoder);
 }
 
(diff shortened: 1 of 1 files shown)


drivers/gpu/drm/vc4/vc4_hdmi.c
```diff
@@ -874,24 +874,27 @@ static void vc4_hdmi_recenter_fifo(struct vc4_hdmi *vc4_hdmi)
...
-	HDMI_WRITE(HDMI_FIFO_CTL,
-		   drift & ~VC4_HDMI_FIFO_CTL_RECENTER);
-	HDMI_WRITE(HDMI_FIFO_CTL,
-		   drift | VC4_HDMI_FIFO_CTL_RECENTER);
-
-	ret = wait_for(HDMI_READ(HDMI_FIFO_CTL) &
-		       VC4_HDMI_FIFO_CTL_RECENTER_DONE, 1);
-	WARN_ONCE(ret, "Timeout waiting for "
-		  "VC4_HDMI_FIFO_CTL_RECENTER_DONE");
+	for (i = 0; i < 0; i++) { // Once again, with feeling
```
6by9: That's an odd looking loop. It looks more like "Really, don't bother"!

timg236: hmm, must have pressed undo before merging. It probably is required but I guess the other changes are the sigificant ones

popcornmix:
In my testing I'm using:
```
modetest -M vc4 -F tiles,gradient -s 32:1920x1080-60 -P91@76:3840x2160+0+0@XR24
```
with a desktop mode of 1920x1200@59.95.
It appears to be reliably failing (1 pixel wrap) with default kernel tree.
It looks like purely taking the diffs that affect `HDMI_VID_CTL` from this PR makes it reliably work (no wrap).

popcornmix:
The mdelay appears to be essential. This patch is fixing my failure case (but there may be other cases)
https://paste.ubuntu.com/p/G5rhN6RGhh/

timg236:
That's interesting. I'm trying this on a Pi400 with the C0 fix nobbled and drift FIFO restored. Brain failure earlier made me save a different experimental hack!

timg236:
I ran the second version of my patch overnight in a mode-test loop and it fixed the wraparound issue. I suspect the HVS changes aren't required (i.e. same as patch from @popcornmix), so we should probably just merge the minimal changes and see how that goes.

[mode-test.zip](https://github.com/raspberrypi/linux/files/6714071/mode-test.zip)

timg236:
Removed the HVS code and retested

popcornmix:
I removed the kodi workaround last night and ran with this form of patch and didn't see the one pixel wrap.
@HiassofT could you test LE with this PR and 1 pixel workaround reverted?

timg236:
N.B. I think tearing down some low-resolution interlaced modes could still fail (specifically 720x576i). However, other higher resolution interlaced modes are fine e.g. 1080i

popcornmix:
Does a smaller (or larger) delay affect the low resolution modes?

timg236:
> Does a smaller (or larger) delay affect the low resolution modes?

No, but this exiting from 720x576i mode triggers the wrap around on FKMS as well

alanbork:
is this only for a single pixel or the whole line? because I'm still seeing a whole line wrap around, also tending to be triggered by use of interlaced modes, with no fix as of f29ab05611eef385d17675c77190df0b87a0d456.

https://www.youtube.com/watch?v=Qc9BDV9Dssk

popcornmix:
Yes, this is the same 1-pixel wrap issue, and was mentioned in @tim236's preceding post as still being present (on kms and fkms) when exiting 720x576i (and also other SD interlaced modes).
We don't have a fix for that.

PR #4412, opened 2021-06-25
drm/vc4: hdmi: Enable the scrambler on reconnection

### LINKED ISSUE #4411: KMS: Blank screen after disconnect/reconnect if HDMI mode requires HDMI2.0 scrambling
**Describe the bug**
If the HDMI cable is disconnected then re-attached to the same display then the display is not restored. This only happens if the HDMI mode requires scrambling e.g. 4Kp60.

**To reproduce**
Add hdmi_enable_4kp60=1 to config.txt
Reboot
Use arandr to select 4Kp60
Disconnect HDMI cable
Reattach HDMI cable

**Expected behaviour**
Reconnecting to the same display should work. The FKMS firmware monitors the hotplug interrupt and attempts to restore HDMI scrambling if it was previously enabled. This does not happen with KMS.

Hotplugging to different display resolutions doesn't work, but that's a more difficult problem and requires interaction with userspace DRM client to pick alternate modes.  However, until userspace supports this we the display driver should support the most frequent use-case which is coping with the HDMI cable being removed e.g. when moving a monitor.

**Actual behaviour**
Display remains blank

**System**
Raspberry Pi OS latest
Pi 4B 1.1
rpi-update 6edb3992f5df1ec1107f4c5e62b6d63b074f24c3

**Additional context**
I think this also fails with LibreElec. Raising a specific bug because this is a blocker for enabling KMS by default on Raspberry Pi OS.

mripard:
If we have a state already and disconnect/reconnect the display, the
SCDC messages won't be sent again since we didn't go through a disable /
enable cycle.

In order to fix this, let's call the vc4_hdmi_enable_scrambling function
in the detect callback if there is a mode and it needs the scrambler to
be enabled.

Fixes #4411

### DIFF
--- a/drivers/gpu/drm/vc4/vc4_hdmi.c
+++ b/drivers/gpu/drm/vc4/vc4_hdmi.c
@@ -168,6 +168,8 @@ static void vc4_hdmi_cec_update_clk_div(struct vc4_hdmi *vc4_hdmi)
 static void vc4_hdmi_cec_update_clk_div(struct vc4_hdmi *vc4_hdmi) {}
 #endif
 
+static void vc4_hdmi_enable_scrambling(struct drm_encoder *encoder);
+
 static enum drm_connector_status
 vc4_hdmi_connector_detect(struct drm_connector *connector, bool force)
 {
@@ -198,6 +200,8 @@ vc4_hdmi_connector_detect(struct drm_connector *connector, bool force)
 				kfree(edid);
 			}
 		}
+
+		vc4_hdmi_enable_scrambling(&vc4_hdmi->encoder.base.base);
 
 		ret = connector_status_connected;
 		goto out;
@@ -560,9 +564,13 @@ static bool vc4_hdmi_supports_scrambling(struct drm_encoder *encoder,
 
 static void vc4_hdmi_enable_scrambling(struct drm_encoder *encoder)
 {
-	struct drm_display_mode *mode = &encoder->crtc->state->adjusted_mode;
+	struct drm_display_mode *mode;
 	struct vc4_hdmi *vc4_hdmi = encoder_to_vc4_hdmi(encoder);
 
+	if (!encoder->crtc || !encoder->crtc->state)
+		return;
+
+	mode = &encoder->crtc->state->adjusted_mode;
 	if (!vc4_hdmi_supports_scrambling(encoder, mode))
 		return;
 


6by9:
Looks to be the wrong solution.
See my comments on #4411 - drm_probe_ddc is succeeding, so vc4_hdmi_connector_detect is reporting connected.

mripard:
I guess we have two issues that are entangled:
 * While the HPD signal might be deasserted, the call to drm_probe_ddc might succeed and therefore misreport the display as connected. This can be solved by #4371. @popcornmix suggestions to try to detect whether the hotplug signal is broken on a cable makes sense too, but it didn't look like a blocker.

 * If an application is asking for a mode and gets disconnected and reconnected, it might not react at all. This is what this PR is about, and can be tested by running modetest for example. The fbdev emulation has hotplug support, so it won't fall in this case.

popcornmix:
> If an application is asking for a mode and gets disconnected and reconnected, it might not react at all. 

Yes, kodi will not attempt to tear down and bring up display on a hotplug deassert/reassert. This can occur for reasons as benign as flicking through AV inputs. If video was playing at 4kp60, I'd expect it to still be playing when I switch back.
If the driver can make scrambling continue to work in this situation, then it should.

6by9:
> Yes, kodi will not attempt to tear down and bring up display on a hotplug deassert/reassert. This can occur for reasons as benign as flicking through AV inputs. If video was playing at 4kp60, I'd expect it to still be playing when I switch back.
If the driver can make scrambling continue to work in this situation, then it should.

It'd be interesting to know if 4k60 is broken on i915 then, as that only appears to send the relevant scdc commands during encoder enable, which by the sounds of it won't be triggered.

6by9:
If the app isn't required to handle hotplug, and the framework doesn't, then I guess it is up to the driver.

Are there any issues over multiple thread contexts here? I haven't gotten my head around threading in DRM as yet.

HiassofT:
I gave this a quick try on LE here and it looks good so far with my LG 55C8 - I get a picture after putting it out of standby when running kodi at 4kp60

timg236:
Works for me

pelwell:
Any objections to merging now and fixing any threading issues as they arise?

mripard:
> Are there any issues over multiple thread contexts here? I haven't gotten my head around threading in DRM as yet.

Not really (or at least, we shouldn't have any, concurrency is hard). DRM makes sure that there's at most one commit per CRTC in flight at any given moment, and we have some extra code ([in 5.10](https://github.com/raspberrypi/linux/blob/rpi-5.10.y/drivers/gpu/drm/vc4/vc4_kms.c#L480) and [in mainline](https://github.com/raspberrypi/linux/blob/rpi-5.12.y/drivers/gpu/drm/vc4/vc4_kms.c#L385)) to make sure we only have a single commit occurring at a given time across the entire pipeline.

popcornmix:
@pelwell okay to merge?

pelwell:
Sure - there have been no objections.

PR #4414, opened 2021-06-28
drm/vc4: Increase the core clock to a minimum of 500MHz

mripard:
The core clock needs to be raised temporarily during a modeset to
500MHz. However, the HVS core clock requirement might be higher than
500MHz. This rate will be enforced at the end of the mode setting,
meaning that might might end up with a core clock rate lower than
planned on the first mode set.

Use the maximum value of 500MHz and the HVS core clock rate for our
temporary boost to fix this issue.


6by9:
LGTM.

popcornmix:
Me too.

PR #4415, opened 2021-06-28
media: i2c: ov9281: Remove override of subdev name

6by9:
From the original Rockchip driver, the subdev was renamed
from the default to being "mov9281 <dev_name>" whereas the
default would have been "ov9281 <dev_name>".

Remove the override to drop back to the default rather than
a vendor custom string.


6by9:
@davidplowman

davidplowman:
Thanks, looks perfect to me!

PR #4416, opened 2021-06-28
media: i2c: imx477: Extend support to include the imx378 sensor

davidplowman:
The Sony imx378 is very nearly identical to the im477 and with a very
small number of extra register writes can share the same driver.

### DIFF
--- a/drivers/media/i2c/imx477.c
+++ b/drivers/media/i2c/imx477.c
@@ -26,6 +26,7 @@
 /* Chip ID */
 #define IMX477_REG_CHIP_ID		0x0016
 #define IMX477_CHIP_ID			0x0477
+#define IMX378_CHIP_ID			0x0378
 
 #define IMX477_REG_MODE_SELECT		0x0100
 #define IMX477_MODE_STANDBY		0x00
@@ -453,6 +454,12 @@ static const struct imx477_reg mode_common_regs[] = {
 	{0x0350, 0x00},
 	{0xbcf1, 0x02},
 	{0x3ff9, 0x01},
+};
+
+static const struct imx477_reg imx378_extra_regs[] = {
+	{0x3e35, 0x01},
+	{0x4421, 0x08},
+	{0x3ff9, 0x00},
 };
 
 /* 12 mpix 10fps */
@@ -1107,6 +1114,9 @@ struct imx477 {
 
 	/* Current long exposure factor in use. Set through V4L2_CID_VBLANK */
 	unsigned int long_exp_shift;
+
+	/* imx378 sensor needs a tiny number of extra register updates */
+	unsigned int chip_id;
 };
 
 static inline struct imx477 *to_imx477(struct v4l2_subdev *_sd)
@@ -1678,6 +1688,9 @@ static int imx477_start_streaming(struct imx477 *imx477)
 	if (!imx477->common_regs_written) {
 		ret = imx477_write_regs(imx477, mode_common_regs,
 					ARRAY_SIZE(mode_common_regs));
+		if (ret == 0 && imx477->chip_id == IMX378_CHIP_ID)
+			ret = imx477_write_regs(imx477, imx378_extra_regs,
+						ARRAY_SIZE(imx378_extra_regs));
 		if (ret) {
 			dev_err(&client->dev, "%s failed to set common settings\n",
 				__func__);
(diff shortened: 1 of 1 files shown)


davidplowman:
@6by9 @naushir 
This resolves the problems with the imx378 sensor over the imx477 - do you think what I've done would be "acceptable"?

6by9:
Do those registers cause issues on IMX477? Could we send them unconditionally?
Does the tuning need to change between them, so is the name important?

If we do need to distinguish between them, then it really ought to be done in device tree in case further deviation is required.
Add an extra compatible string, and add a pointer to a ```struct imx477_variant``` which contains the sensor ID for that variant, and any special registers. The overlay can then include an override on the compatible string, or we have two overlays that include a common back-end

See imx290.c for an example as it handles a mono and colour variant through that route.
imx290-overlay.dts also does almost the above in including imx290_327-overlay.dtsi, but I never wrote the imx327-overlay.dts variant as we don't support any of the functionality that is 290 specific (1080p120 10bit).

naushir:
> Do those registers cause issues on IMX477? Could we send them unconditionally?

I'd be cautious about doing this without Sony's OK.

davidplowman:
Thanks for the comments!

> Do those registers cause issues on IMX477? Could we send them unconditionally?

The 3ff9 register is documented and wouldn't make any difference to us, though it might do to someone else who knew what it was for. But maybe we don't care. The other two are undocumented. I suspect they'd have no effect - I could do the experiment, though that's not a cast-iron guarantee....

> Does the tuning need to change between them, so is the name important?

Really I think it's more of a libcamera problem because we have no way of distinguishing between modules based on the same sensor. So whilst a different name is convenient we could just indulge in the usual clunky workarounds that libcamera currently forces on us.

> 
> If we do need to distinguish between them, then it really ought to be done in device tree in case further deviation is required.
> Add an extra compatible string, and add a pointer to a `struct imx477_variant` which contains the sensor ID for that variant, and any special registers. The overlay can then include an override on the compatible string, or we have two overlays that include a common back-end
> 
> See imx290.c for an example as it handles a mono and colour variant through that route.
> imx290-overlay.dts also does almost the above in including imx290_327-overlay.dtsi, but I never wrote the imx327-overlay.dts variant as we don't support any of the functionality that is 290 specific (1080p120 10bit).

Understood. I was kinda hoping that if you can tell automatically from the device itself what you need to do then forcing people to mess around with device trees seems rather cruel! (In the imx290 case, I'm guessing that it's only through the device tree that you figure out what you you're dealing with?)

Maybe I'll go and perform that experiment on the imx477 and report back.

davidplowman:
> > Do those registers cause issues on IMX477? Could we send them unconditionally?
> 
> I'd be cautious about doing this without Sony's OK.

Well, we could ask... :)

6by9:
It is true that with imx290 you have no other choice than DT as there is no ID register to query.

Configuring imx378 via an imx477 compatible just feels wrong. Even if the register set were identical, it'd be tempting to add both compatibles in case they diverged in future. Those who then took the easy option out and used the other compatible have only themselves to blame.

pelwell:
+1 for adding an IMX378 compatible string and checking that the device matches.

naushir:
I'll look to see if it is feasible to set the Y plane alignment to 64 bytes for YUV420.  This will correctly set the U/V plane alignments to 32.  Will need a simultaneous change to kernel + firmware.

6by9:
> I'll look to see if it is feasible to set the Y plane alignment to 64 bytes for YUV420. This will correctly set the U/V plane alignments to 32. Will need a simultaneous change to kernel + firmware.

Did you mean to comment on https://github.com/raspberrypi/libcamera-apps/issues/42 ?
Changing the bytesperline_align at https://github.com/raspberrypi/linux/blob/rpi-5.10.y/drivers/staging/vc04_services/bcm2835-isp/bcm2835-isp-fmts.h#L51 to 64 should do it with just a kernel change. Firmware will do as it's told then.

naushir:
> Did you mean to comment on [raspberrypi/libcamera-apps#42](https://github.com/raspberrypi/libcamera-apps/issues/42) ?
> Changing the bytesperline_align at https://github.com/raspberrypi/linux/blob/rpi-5.10.y/drivers/staging/vc04_services/bcm2835-isp/bcm2835-isp-fmts.h#L51 to 64 should do it with just a kernel change. Firmware will do as it's told then.

Sorry yes, wrong window!

PR #4417, opened 2021-06-28
drm/vc4: hdmi: Use a fixed rate for the HSM clock on BCM2835

mripard:
Before the introduction of the BCM2711 support, the HSM clock rate was
fixed, and was the CEC and audio clock source on the SoCs previously
supported.

The HSM clock is also the source of the internal state machine of the
controller and needs to run faster than the pixel clock. All these
requirements were met by running at 101% of the maximum pixel rate,
meeting the fixed clock requirement for audio and CEC, while remaining
faster than any pixel clock we might need.

However, the BCM2711 brought support for 4k and therefore increased
significantly the rate needed for the HSM, and new, independant, clocks
to feed the audio and CEC clocks. Since the HSM clock can also run much
higher, we also need to lower its rate if possible to reduce its power
consumption.

The CEC support code changes its clock divider when the HSM clock rate
is changed, but the audio support never had a similar feature and will
glitch out if audio is played back during a mode set.

Since the HSM rate was meant to be fixed on the SoCs prior to the
BCM2711 anyway, let's introduce back a fixed HSM rate and fix audio.

Fixes: cd4cb49dc5bb ("drm/vc4: hdmi: Adjust HSM clock rate depending on pixel rate")

### DIFF
--- a/drivers/gpu/drm/vc4/vc4_hdmi.c
+++ b/drivers/gpu/drm/vc4/vc4_hdmi.c
@@ -926,23 +926,7 @@ static void vc4_hdmi_encoder_pre_crtc_configure(struct drm_encoder *encoder,
 		return;
 	}
 
-	/*
-	 * As stated in RPi's vc4 firmware "HDMI state machine (HSM) clock must
-	 * be faster than pixel clock, infinitesimally faster, tested in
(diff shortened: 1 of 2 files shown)


popcornmix:
Confirmed this fixes audio playback when running modetest to switch hdmi mode while aplay is running.

PR #4418, opened 2021-06-29
Fix CPU hang when using CEC while HDMI output is disabled

### LINKED ISSUE #4319: RPi 2b & RPi 3b crash after a while with kernel 5.10.31+
**Describe the bug**
Since the upgrade to linux-raspberrypi version 5.10.31 and also 5.10.32 and 5.10.33, my Raspberry Pi 3b crashes after a while. I do not have this problem with kernel 5.10.27. I've also tried a Raspberry Pi 2b with the same config and it also crashes.
Sometimes the system crashes after an hour, sometimes after a day. It seems that the crashes occur randomly.

**System**
* Which model of Raspberry Pi? Pi 2b and Pi 3b
* Which OS and version: Arch Linux Arm, latest version
* Which firmware version: version a48d332c35ee1c1c1ab433228e23317f62dcc5fb
* Which kernel version: 5.10.31 and up

The system is connected to a TV (1920x1080) using HDMI. The TV is switched off most of the time.
The system is used as an wireless access point and sometimes Kodi is started.

**Logs**
I do not see anything in the logs. Also, when the system crashes, the TV output gives no signal.

mripard:
Hi,

Here's a PR to fix an issue similar to 755b2c80ac4d but with CEC. It's doing further cleanups and drive-by patches and finally using the same solution we used.

Fixes #4319

### DIFF
--- a/drivers/gpu/drm/vc4/vc4_hdmi.c
+++ b/drivers/gpu/drm/vc4/vc4_hdmi.c
@@ -660,7 +660,6 @@ static void vc4_hdmi_encoder_post_crtc_powerdown(struct drm_encoder *encoder,
 	clk_disable_unprepare(vc4_hdmi->pixel_bvb_clock);
 	if (vc4_hdmi->bvb_req)
 		clk_request_done(vc4_hdmi->bvb_req);
-	clk_disable_unprepare(vc4_hdmi->hsm_clock);
 	clk_request_done(vc4_hdmi->hsm_req);
 	clk_disable_unprepare(vc4_hdmi->pixel_clock);
 
@@ -941,13 +940,6 @@ static void vc4_hdmi_encoder_pre_crtc_configure(struct drm_encoder *encoder,
 		return;
 	}
 
-	ret = clk_prepare_enable(vc4_hdmi->hsm_clock);
-	if (ret) {
-		DRM_ERROR("Failed to turn on HSM clock: %d\n", ret);
-		clk_disable_unprepare(vc4_hdmi->pixel_clock);
-		return;
-	}
-
 	vc4_hdmi_cec_update_clk_div(vc4_hdmi);
 
 	if (pixel_rate > 297000000)
@@ -962,7 +954,6 @@ static void vc4_hdmi_encoder_pre_crtc_configure(struct drm_encoder *encoder,
 	if (IS_ERR(vc4_hdmi->bvb_req)) {
 		DRM_ERROR("Failed to set pixel bvb clock rate: %ld\n", PTR_ERR(vc4_hdmi->bvb_req));
 		clk_request_done(vc4_hdmi->hsm_req);
-		clk_disable_unprepare(vc4_hdmi->hsm_clock);
 		clk_disable_unprepare(vc4_hdmi->pixel_clock);
 		return;
 	}
@@ -973,7 +964,6 @@ static void vc4_hdmi_encoder_pre_crtc_configure(struct drm_encoder *encoder,
 		if (vc4_hdmi->bvb_req)
 			clk_request_done(vc4_hdmi->bvb_req);
 		clk_request_done(vc4_hdmi->hsm_req);
-		clk_disable_unprepare(vc4_hdmi->hsm_clock);
 		clk_disable_unprepare(vc4_hdmi->pixel_clock);
 		return;
 	}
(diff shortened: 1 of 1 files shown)


pelwell:
That all looks sensible. Can anyone confirm that it fixes the problem, or at least that it doesn't make anything worse?

popcornmix:
I can confirm what @henkv1 reported. This PR breaks boot with kms on Pi3+.
Unfortunately it's a hard lock up, so no dmesg log (even with uart).
Screen just goes blank and uart output stops when we reach the point to switch to kms.
It breaks with "drm/vc4: hdmi: Move the HSM clock enable to runtime_pm" commit.

pelwell:
Lazy question (I could probably find the answer if I looked hard enough), but is it guaranteed that the resume method is called at the start of day? In other words, is the clock starting out off when it should start out on?

mripard:
Is it always occurring for you? I had it a couple of times when I started to debug it earlier, but it never occurred if the driver was built-in, and as soon as I started adding a bit of tracing it went away.

My first guess was that, if it's compiled as a module, we would have the HSM clock being disabled by the clock framework at the end of the kernel boot, and then we would access a register without the clock enabled. However, since it seems to be timing sensitive, I'm wondering if it's not a race somewhere ?

> Lazy question (I could probably find the answer if I looked hard enough), but is it guaranteed that the resume method is called at the start of day? In other words, is the clock starting out off when it should start out on?

it will be called when we call pm_runtime_get or pm_runtime_resume_and_get if the refcount was 0. If we're missing one of these calls, then yeah, the clock will be off (but the power domain will also be for example)

mripard:
As soon as I started testing with RaspiOS (as opposed to the buildroot system I was using earlier), I indeed get the issue all the time.

mripard:
So, I'm not really sure why that particular commit made it bad because the issue has been there for a while, but we were having some registers access at bind time without the HSM clock running. I've fixed it up, it should fix the initial CEC issue now and I'm not having any hang anymore while building as a module

drivers/gpu/drm/vc4/vc4_hdmi.c
```diff
@@ -955,6 +955,7 @@ static void vc4_hdmi_encoder_pre_crtc_configure(struct drm_encoder *encoder,
 		DRM_ERROR("Failed to set pixel bvb clock rate: %ld\n", PTR_ERR(vc4_hdmi->bvb_req));
 		clk_request_done(vc4_hdmi->hsm_req);
 		clk_disable_unprepare(vc4_hdmi->pixel_clock);
+		pm_runtime_put(&vc4_hdmi->pdev->dev);
```
pelwell, line 962: I think you may also need to change all the error paths after the successful pm_runtime_resume_and_get

popcornmix:
> So, I'm not really sure why that particular commit made it bad because the issue has been there for a while

That's interesting as when I first tested this PR and found the hang I started bisecting, and spent a couple of hours going down false paths. Eventually I restarted that testing and found "drm/vc4: hdmi: Move the HSM clock enable to runtime_pm" commit seemed bad and the preceding commit was at least sometimes good.

But it doesn't surprise me that there's something racy there.

mripard:
I've updated the branch according to @pelwell reviews

drivers/gpu/drm/vc4/vc4_hdmi.c
```diff
@@ -917,26 +917,30 @@ static void vc4_hdmi_encoder_pre_crtc_configure(struct drm_encoder *encoder,
...
 	if (ret) {
 		DRM_ERROR("Failed to turn on pixel clock: %d\n", ret);
+		pm_runtime_put(&vc4_hdmi->pdev->dev);
 		return;
 	}
 
 	hsm_rate = vc4_hdmi->variant->calc_hsm_clock(vc4_hdmi, pixel_rate);
 	vc4_hdmi->hsm_req = clk_request_start(vc4_hdmi->hsm_clock, hsm_rate);
 	if (IS_ERR(vc4_hdmi->hsm_req)) {
 		DRM_ERROR("Failed to set HSM clock rate: %ld\n", PTR_ERR(vc4_hdmi->hsm_req));
```
pelwell, line 942: Are you not also missing a `clk_disable_unprepare(vc4_hdmi->pixel_clock);` here?

pelwell: I'm OK with merging this now as a fix, and the clk_disable_unprepare can be added as a later improvement.

popcornmix:
A quick test did work.
Booted okay on Pi3+. cec worked. Disabled `/sys/class/drm/card0-HDMI-A-1/status` and cec still worked. Enabled and still working.

henkv1:
It did not work for me. I compiled ecdd08fd9bba06045eb94922149cdad23eb3820a. But as soon as I modprobe vc4, the system crashes.

PR #4419, opened 2021-06-29
staging: vc04_services: isp: Set the YUV420/YVU420 format stride to 64 bytes

naushir:
The bcm2835 ISP requires all input/output planes to have 32 byte alignment.
Using a Y stride of 32 bytes would not guarantee that the V plane would fulfil
this, e.g. a height of 650 lines would mean the V plane buffer is not 32 byte
aligned for YUV420 formats.

Having a Y stride of 64 bytes would ensure both U and V planes have a 32 byte
alignment, provided the height is an even number of lines.


naushir:
@davidplowman and @6by9 

This seems to fix the issue at https://github.com/raspberrypi/libcamera-apps/issues/42

6by9:
Wording of the commit text only

The bcm2835 ISP requires **the base address of** all input/output planes to have 32 byte alignment.

> provided the height is an even number of lines.

That's already enforced, so "as the luma height will always be an even number of lines".

naushir:
Done!

PR #4420, opened 2021-06-29
media: i2c: imx477: Add support for imx378 as a compatible sensor

davidplowman:
The imx378 is almost identical to the imx477 and can be supported by
the same driver with just a few extra register writes.

### DIFF
--- a/arch/arm/boot/dts/overlays/imx378-overlay.dts
+++ b/arch/arm/boot/dts/overlays/imx378-overlay.dts
@@ -1,10 +1,10 @@
 // SPDX-License-Identifier: GPL-2.0-only
-// Definitions for IMX477 camera module on VC I2C bus
+// Definitions for IMX378 camera module on VC I2C bus
 /dts-v1/;
 /plugin/;
 
 #include "imx477_378-overlay.dtsi"
 
 &imx477 {
-	compatible = "sony,imx477";
+	compatible = "sony,imx378";
 };
--- a/arch/arm/boot/dts/overlays/imx477_378-overlay.dtsi
+++ b/arch/arm/boot/dts/overlays/imx477_378-overlay.dtsi
@@ -1,0 +1,108 @@
+// SPDX-License-Identifier: GPL-2.0-only
+// Definitions for IMX477 camera module on VC I2C bus
+
+/{
+	compatible = "brcm,bcm2835";
+
+	fragment@0 {
+		target = <&i2c_csi_dsi>;
+		__overlay__ {
+			#address-cells = <1>;
+			#size-cells = <0>;
+			status = "okay";
+
+			imx477: imx477@1a {
+				reg = <0x1a>;
+				status = "okay";
+
+				clocks = <&imx477_clk>;
+				clock-names = "xclk";
+
+				VANA-supply = <&cam1_reg>;	/* 2.8v */
+				VDIG-supply = <&imx477_vdig>;	/* 1.05v */
+				VDDL-supply = <&imx477_vddl>;	/* 1.8v */
+
+				rotation = <180>;
+
+				port {
+					imx477_0: endpoint {
+						remote-endpoint = <&csi1_ep>;
+						clock-lanes = <0>;
+						data-lanes = <1 2>;
+						clock-noncontinuous;
+						link-frequencies =
+							/bits/ 64 <450000000>;
+					};
+				};
+			};
+		};
+	};
+
+	fragment@1 {
+		target = <&csi1>;
+		__overlay__ {
+			status = "okay";
+
+			port {
+				csi1_ep: endpoint {
+					remote-endpoint = <&imx477_0>;
+					clock-lanes = <0>;
+					data-lanes = <1 2>;
+					clock-noncontinuous;
+				};
+			};
+		};
+	};
+
+	fragment@2 {
+		target = <&i2c0if>;
+		__overlay__ {
+			status = "okay";
+		};
+	};
+
+	fragment@3 {
+		target-path="/";
+		__overlay__ {
+			imx477_vdig: fixedregulator@0 {
+				compatible = "regulator-fixed";
+				regulator-name = "imx477_vdig";
+				regulator-min-microvolt = <1050000>;
+				regulator-max-microvolt = <1050000>;
+			};
+			imx477_vddl: fixedregulator@1 {
+				compatible = "regulator-fixed";
+				regulator-name = "imx477_vddl";
+				regulator-min-microvolt = <1800000>;
+				regulator-max-microvolt = <1800000>;
+			};
+			imx477_clk: camera-clk {
+				compatible = "fixed-clock";
+				#clock-cells = <0>;
+				clock-frequency = <24000000>;
+			};
+		};
+	};
+
+	fragment@4 {
+		target = <&i2c0mux>;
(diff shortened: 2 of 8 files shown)


davidplowman:
@naushir @6by9 
Here's a new version of this. It adds the imx378 as a "compatible" sensor to the imx477. You need to specify the overlay as either "dtoverlay=imx477" or "dtoverlay=imx378", and it has to match the attached sensor correctly.

One thing I was a bit unsure of was how to update the documentation, particularly as these two sensors are both different, yet intertwined. Any advice there?

drivers/media/i2c/imx477.c
```diff
@@ -1673,11 +1683,18 @@ static int imx477_start_streaming(struct imx477 *imx477)
 {
 	struct i2c_client *client = v4l2_get_subdevdata(&imx477->sd);
 	const struct imx477_reg_list *reg_list;
+	const struct imx477_reg_list *extra_regs;
 	int ret;
 
 	if (!imx477->common_regs_written) {
 		ret = imx477_write_regs(imx477, mode_common_regs,
 					ARRAY_SIZE(mode_common_regs));
+		if (ret == 0) {
```
6by9: ```if (!ret)```

drivers/media/i2c/imx477.c
```diff
@@ -2090,6 +2138,12 @@ static int imx477_probe(struct i2c_client *client)
 
 	v4l2_i2c_subdev_init(&imx477->sd, client, &imx477_subdev_ops);
 
+	match = of_match_device(imx477_dt_ids, dev);
+	if (!match)
+		return -ENODEV;
+	memcpy(&imx477->compatible_data, match->data,
+	       sizeof(imx477->compatible_data));
```
6by9: Why memcpy instead of having ```struct imx477_compatible_data *compatible_data;``` in the struct and copy the pointer?

6by9 (commented):
A couple of minor comments on the code.

Ideally dt-bindings, dtoverlays, and driver changes should be split into independent commits.

drivers/media/i2c/imx477.c
davidplowman: Okeydoke!

drivers/media/i2c/imx477.c
davidplowman: True enough, will fix that!

davidplowman:
So just to try and get this right, should that be 2 commits, one with the driver (presumably that's the 2nd of the two) and one with everything else?  Thanks!

6by9:
> So just to try and get this right, should that be 2 commits, one with the driver (presumably that's the 2nd of the two) and one with everything else? Thanks!

3 please.
- Documentation/devicetree
- arch/arm/boot/dts/boot/overlays
- drivers/media/i2c
It's the normal breakdown for mainline as they are different contributing trees.
boot/dts/overlays is only in our downstream tree, but it is generally squashed when switching to a new release, and that's made more awkward if you've mixed overlays with other stuff.

Can I check that you're running ```scripts/checkpatch.pl --strict``` against your patches before submission? I have a .git/hooks/post-commit hook of 
```
```
so that is automatic on every commit.

davidplowman:
So here's a split up version, with those little code fixes. I do get a couple of checkstyle warnings:

- In the first commit, I get "Possible repeated word: 'git'", but it seems like all the entries have that...
- In the second commit, I have "added, moved or deleted file(s), does MAINTAINERS need updating" though I can't really see what file it might be referring to.

The final (big) patch is completely fine. Any suggestions on those warnings or can one live with them? I also remain a bit baffled by exactly what device tree documentation I should really be providing...    Thanks!!

arch/arm/boot/dts/overlays/imx378-overlay.dts
```diff
@@ -0,0 +1,21 @@
+// SPDX-License-Identifier: GPL-2.0-only
+// Definitions for IMX378 camera module on VC I2C bus
+/dts-v1/;
+/plugin/;
+
+#include <dt-bindings/gpio/gpio.h>
```
pelwell: This header isn't required.

arch/arm/boot/dts/overlays/imx477-overlay.dts
```diff
@@ -4,110 +4,18 @@
 /plugin/;
 
 #include <dt-bindings/gpio/gpio.h>
```
pelwell: Nor here.

arch/arm/boot/dts/overlays/imx477_378-overlay.dtsi
```diff
@@ -0,0 +1,112 @@
+// SPDX-License-Identifier: GPL-2.0-only
+// Definitions for IMX477 camera module on VC I2C bus
+/dts-v1/;
```
pelwell: This line and the next should only appear once - at the top of the including .dts files.

arch/arm/boot/dts/overlays/imx477_378-overlay.dtsi
```diff
@@ -0,0 +1,112 @@
...
+/plugin/;
+
+#include <dt-bindings/gpio/gpio.h>
```
pelwell: Still not needed.

6by9:
It tells you the file and line:
```
WARNING: Possible repeated word: 'git'
#143: FILE: MAINTAINERS:16362:
+T:	git git://linuxtv.org/media_tree.git
```
Line 16362 is correct, except that I believe you now have drivers/media/i2c/imx477.c listed twice in MAINTAINERS, which is a bad thing.
imx378.yaml can be listed under the one entry for IMX477 at present.
I'm not sure what upstream will truly want in this case. You could list the one binding with multiple compatibles, but this is more generic should they ever need to diverge.

```WARNING: added, moved or deleted file(s), does MAINTAINERS need updating?```
You added arch/arm/boot/dts/overlays/imx378-overlay.dts and renamed /arch/arm/boot/dts/overlays/imx477_378-overlay.dtsi
None of the overlays are covered by MAINTAINTERS, so no issue.

File: arch/arm/boot/dts/overlays/imx477-overlay.dts [hunk 7]
```diff
@@ -4,110 +4,18 @@
...
+#include "imx477_378-overlay.dtsi"
 
 /{
 	compatible = "brcm,bcm2835";
```
pelwell: As with the imx378 overlay, the remainder of the file can just be:
```
&imx477 {
        compatible = "sony,imx477";
};
```

arch/arm/boot/dts/overlays/imx378-overlay.dts
```diff
@@ -0,0 +1,21 @@
...
+#include "imx477_378-overlay.dtsi"
+
+/{
```
pelwell: There's no need to repeat the compatible string. In fact, because the change of compatible string is unconditional, this whole node (the remainder of the file) can be replaced with:
```
&imx477 {
        compatible = "sony,imx378";
};
```

davidplowman:
Thanks Phil, Dave. Will look again at those various bits and resubmit!

davidplowman:
Latest attempt, also down to just the one warning now... hopefully we're getting a bit closer!  :)

6by9:
LGTM.

PR #4421, opened 2021-06-29
drm/vc4: Notify firmware when use of simple framebuffer is complete

popcornmix:
With latest firmware and this call, the firmware is able to free the memory allocation for the simple framebuffer that was used in initial booting. This can be up to 16MB of gpu memory when running with `hdmi_enable_4kp60=1`.

It also allows the firmware to release the boost to the core clock that was necessary when driving that hdmi mode, and may no longer be required if kms choosing a lower resolution hdmi mode.

### DIFF
--- a/drivers/gpu/drm/vc4/vc4_drv.c
+++ b/drivers/gpu/drm/vc4/vc4_drv.c
@@ -305,6 +305,8 @@ static int vc4_drm_bind(struct device *dev)
 	if (ret)
 		return ret;
 
+	drm_fb_helper_remove_conflicting_framebuffers(NULL, "vc4drmfb", false);
+
 	ret = component_bind_all(dev, drm);
 	if (ret)
 		return ret;
@@ -314,8 +316,6 @@ static int vc4_drm_bind(struct device *dev)
 		if (ret)
 			goto unbind_all;
 	}
-
-	drm_fb_helper_remove_conflicting_framebuffers(NULL, "vc4drmfb", false);
 
 	ret = vc4_kms_load(drm);
 	if (ret < 0)
--- a/drivers/gpu/drm/vc4/vc4_drv.c
+++ b/drivers/gpu/drm/vc4/vc4_drv.c
@@ -36,6 +36,8 @@
 #include <drm/drm_fb_helper.h>
 #include <drm/drm_vblank.h>
 
+#include <soc/bcm2835/raspberrypi-firmware.h>
+
 #include "uapi/drm/vc4_drm.h"
 
 #include "vc4_drv.h"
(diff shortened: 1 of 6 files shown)


popcornmix:
@mripard this tested okay for me with LibreELEC on Pi3 and Pi4 with "ARM FB" allocation removed from `vcdbg reloc`.

include/soc/bcm2835/raspberrypi-firmware.h
```diff
@@ -98,6 +98,7 @@ enum rpi_firmware_property_tag {
 	RPI_FIRMWARE_SET_POE_HAT_VAL_OLD =                    0x00030050,
 	RPI_FIRMWARE_NOTIFY_XHCI_RESET =                      0x00030058,
 	RPI_FIRMWARE_GET_REBOOT_FLAGS =                       0x00030064,
+	RPI_FIRMWARE_NOTIFY_DISPLAY_DONE =                    0x00030066,
```
pelwell: Please put this after `RPI_FIRMWARE_SET_REBOOT_FLAGS`, i.e. in numerical order ignoring bit 15 (SETs and GETs are kept in pairs).

arch/arm/boot/dts/bcm2711.dtsi
```diff
@@ -14,6 +14,7 @@
 
 	vc4: gpu {
 		compatible = "brcm,bcm2711-vc5";
+		raspberrypi,firmware = <&firmware>;
```
pelwell: I think this would be better off in bcm2711-rpi.dtsi, since it isn't necessary for all the other non-rpi uses of BCM2711.

pelwell (commented):
A few minor niggles...

HiassofT:
Testing on LibreELEC / RPi4 was fine here as well, core clock boost and framebuffer allocations were removed as expected

popcornmix:
@pelwell does this update match what you wanted?
@mripard are you happy with the two changes @pelwell requested?

drivers/gpu/drm/vc4/vc4_drv.c
```diff
@@ -305,6 +307,26 @@ static int vc4_drm_bind(struct device *dev)
...
+	if (node) {
+		vc4->firmware = rpi_firmware_get(dev->of_node);
+		of_node_put(node);
+
+		if (!vc4->firmware)
+			return -EPROBE_DEFER;
+	}
+
+	drm_fb_helper_remove_conflicting_framebuffers(NULL, "vc4drmfb", false);
+
```
pelwell, line 320: Spurious extra blank line.

pelwell (commented):
Yes, that looks better. If you can remove that extra blank line I'll merge.

popcornmix:
@mripard are you happy with this PR? There have been trivial edits to your commits based on @pelwell's suggestions.

mripard:
Yep, it looks great, thanks!

PR #4423, opened 2021-06-30
WIP: VC4 DSI fixes

6by9:
Creating PR for reference.
A couple of fixups, and moving PHY initialisation to pm_runtime.

WIP as apparently drm_encoder.crtc is deprecated and only for non-atomic drivers, so the current implementation of the pm_runtime resume functions are "incorrect".

### DIFF
--- a/drivers/gpu/drm/vc4/vc4_dsi.c
+++ b/drivers/gpu/drm/vc4/vc4_dsi.c
@@ -839,11 +839,9 @@ static bool vc4_dsi_encoder_mode_fixup(struct drm_encoder *encoder,
 	/* Find what divider gets us a faster clock than the requested
 	 * pixel clock.
 	 */
-	for (divider = 1; divider < 8; divider++) {
-		if (parent_rate / divider < pll_clock) {
-			divider--;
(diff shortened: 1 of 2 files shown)


drivers/gpu/drm/vc4/vc4_dsi.c
```diff
@@ -798,29 +800,15 @@ static void vc4_dsi_encoder_disable(struct drm_encoder *encoder)
...
-			iter->funcs->post_disable(iter);
-	}
-
-	clk_disable_unprepare(dsi->pll_phy_clock);
-	clk_disable_unprepare(dsi->escape_clock);
-	clk_disable_unprepare(dsi->pixel_clock);
-
 	pm_runtime_mark_last_busy(dev);
 	pm_runtime_put_autosuspend(dev);
+
```
pelwell, line 811: Accidental blank line?

pelwell (commented):
Niggle aside, it looks reasonable. However, GitHub could do with a smarter way to indicate code that is copied or moved in a diff.

6by9:
Closing as this isn't the solution that upstream are happy with.

PR #4431, opened 2021-07-05
drm/vc4: More CPU Stall Fixes

### LINKED ISSUE #4319: RPi 2b & RPi 3b crash after a while with kernel 5.10.31+

mripard:
Hi,

Here's another for one more case of CPU hang related to the vc4 driver.

There's a patch in there too to make sure the access is not silent and we complain loudly if we're about to access a register and the controller is not powered up.

Fixes #4319

### DIFF
--- a/drivers/gpu/drm/vc4/vc4_hdmi.c
+++ b/drivers/gpu/drm/vc4/vc4_hdmi.c
@@ -1780,13 +1780,14 @@ static int vc4_hdmi_cec_enable(struct cec_adapter *adap)
 	struct vc4_hdmi *vc4_hdmi = cec_get_drvdata(adap);
 	/* clock period in microseconds */
 	const u32 usecs = 1000000 / CEC_CLOCK_FREQ;
-	u32 val = HDMI_READ(HDMI_CEC_CNTRL_5);
+	u32 val;
 	int ret;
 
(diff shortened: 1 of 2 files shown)


pelwell:
Fingers crossed...

pelwell:
The patch is in the latest rpi-update firmware release.

Issue #4428, opened 2021-07-03
pi4/fkms: boot fails due to flip_done timed out (now 1080p and lower)


alanbork:
under f29ab05611eef385d17675c77190df0b87a0d456, I can't get the pi4 to fully boot with fkms enabled. that is to say it boots but the screen shuts down mid-boot and the machine becomes so unresponsive that I frequently can't ssh in.

 it's failing just like it used to for 4k resolutions (https://github.com/raspberrypi/linux/issues/4271), with a

 [  798.560610] [drm:drm_atomic_helper_wait_for_flip_done [drm_kms_helper]] *ERROR* [CRTC:87:crtc-0] flip_done timed out 

but now:

* overclocking doesn't help
* using a lower resolution doesn't help (1080p, 480p; I haven't tried that many since it's so hard to recover from this with SSH unresponsive)

revert to 039088f7d929b0d1cc49e8d2012df0652f908518 and it boots with no problems.

config.txt:

```
dtoverlay=vc4-fkms-v3d
max_framebuffers=1
hdmi_force_hotplug=1
hdmi_group=1
hdmi_mode=16
hdmi_enable_4kp60=1
```

WRT overclocking, and unlike the 4k issue, adding these lines don't help:
```
#overclcking option can be toggled on of off, no change 
core_freq=600
core_freq_min=600
```

popcornmix:
Can you try with latest rpi-update kernel?

alanbork:
did rpi-update 9dbcd6dfed16cc7de4a67b9ca445c4847f68c4c2

working now at 1080p, also 3840x2160-60.00hz(60).

PR #4432, opened 2021-07-05
vc4/drv: Only notify firmware of display done with kms

popcornmix:
fkms driver still wants firmware display to be active

### DIFF
--- a/drivers/gpu/drm/vc4/vc4_drv.c
+++ b/drivers/gpu/drm/vc4/vc4_drv.c
@@ -258,6 +258,18 @@ const struct of_device_id vc4_dma_range_matches[] = {
 	{}
 };
 
+/*
+ * we need this helper function for determining presence of fkms
+ * before it's been bound
+ */
+static bool firmware_kms(void)
+{
+	return of_device_is_available(of_find_compatible_node(NULL, NULL,
+	       "raspberrypi,rpi-firmware-kms")) ||
+	       of_device_is_available(of_find_compatible_node(NULL, NULL,
+	       "raspberrypi,rpi-firmware-kms-2711"));
+}
+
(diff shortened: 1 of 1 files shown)


popcornmix:
This is a fix for https://github.com/raspberrypi/linux/issues/4428 but it doesn't currently work as firmware_kms only gets set by the subsequent `component_bind_all` call (which invokes `vc4_fkms_bind`).

@6by9 @pelwell any suggestions for a clean way to make this work?
We don't want to call `component_bind_all` first (as it does write to hvs registers).
I guess we can query DT directly?

pelwell:
I can't find a nice "locate the enabled device node by it's label" function - only the kernel overlay support seems to understand labels. There is no property of the vc4 node that is a refence to the firmwarekms node, so aside from adding one that leaves either absolute paths (awful) or searching by compatible string, which is better.

There are two possible compatible strings - `raspberrypi,rpi-firmware-kms` and `raspberrypi,rpi-firmware-kms-2711`. You can either look for both or, since vc4 knows what platform it is running on, choose the appropriate one. Once you have a device node pointer you can check if it is "available" (enabled):
```
    struct device_node *dn;

    compatible_string = ...;
    dn = of_find_compatible_node(NULL, NULL, compatible_string);
    if (vc4->firmware && !of_device_is_available(dn)) {
        ...
```

pelwell:
Note that calling of_device_is_available() with a NULL pointer returns failure without complaint (as we want), and that I've updated the example code so that it should be a drop in replacement once you've decided how to choose a compatible string.

popcornmix:
Thanks. Updated PR booted on Pi3+ and Pi4 with both kms and fkms. Notify call was working on kms.

PR #4433, opened 2021-07-05
drm/vc4: Fixes for interrupt races

mripard:
Hi,

These are fixes to address issues with various races in the CEC and hotplug interrupt handlers in the HDMI driver.

The most visible fix is the last one, where we could have a null pointer dereference right when the driver was loaded. @6by9 could trigger it 100% of the time using hdmi_ignore_hotplug:0=1. I couldn't but forced interrupts to be generated by the interrupt controller at various point of the driver's life and could see that crash as well.

### DIFF
--- a/drivers/gpu/drm/vc4/vc4_hdmi.c
+++ b/drivers/gpu/drm/vc4/vc4_hdmi.c
@@ -1621,7 +1621,7 @@ static irqreturn_t vc4_hdmi_hpd_irq_thread(int irq, void *priv)
 	struct vc4_hdmi *vc4_hdmi = priv;
 	struct drm_device *dev = vc4_hdmi->connector.dev;
 
-	if (dev)
+	if (dev && dev->registered)
 		drm_kms_helper_hotplug_event(dev);
 
 	return IRQ_HANDLED;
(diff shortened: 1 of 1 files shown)


6by9:
Yup, that fixes it for me. 

Extra logging confirmed that we were still going into the HPD ISR in the early probes that get deferred, it triggered during at least one of them, but the driver didn't blow up.

Sometimes devm_ isn't your friend.

drivers/gpu/drm/vc4/vc4_hdmi.c
```diff
@@ -1630,33 +1630,47 @@ static irqreturn_t vc4_hdmi_hpd_irq_thread(int irq, void *priv)
...
 			return ret;
+		}
 
 		connector->polled = DRM_CONNECTOR_POLL_HPD;
 	}
 
 	return 0;
 }
 
+static int vc4_hdmi_hotplug_exit(struct vc4_hdmi *vc4_hdmi)
```
pelwell: Is there a convention that _exit functions can fail? If not, I think it would be better to lose the return value.

mripard:
I've updated the series according to @pelwell's feedback. @vianpl also reported the same bug some time ago and told me that it was now working for him as well.

pelwell:
Thanks, Maxime.

PR #4435, opened 2021-07-06
HVS Gamma LUT support for the RPi4

mripard:
Hi,

Here's a PR to implement the gamma lut property on the rpi4. This is work that has mostly been done by @6by9 and Juerg Haefliger from Canonical.

Maxime

### DIFF
--- a/drivers/gpu/drm/vc4/vc4_crtc.c
+++ b/drivers/gpu/drm/vc4/vc4_crtc.c
@@ -1189,19 +1189,42 @@ int vc4_crtc_init(struct drm_device *drm, struct vc4_crtc *vc4_crtc,
 
 	if (!vc4->hvs->hvs5) {
 		drm_mode_crtc_set_gamma_size(crtc, ARRAY_SIZE(vc4_crtc->lut_r));
+	} else {
+		/* This is a lie for hvs5 which uses a 16 point PWL, but it
+		 * allows for something smarter than just 16 linearly spaced
+		 * segments. Conversion is done in vc5_hvs_update_gamma_lut.
+		 */
+		drm_mode_crtc_set_gamma_size(crtc, 256);
+	}
 
-		drm_crtc_enable_color_mgmt(crtc, 0, false, crtc->gamma_size);
+	drm_crtc_enable_color_mgmt(crtc, 0, false, crtc->gamma_size);
 
+	if (!vc4->hvs->hvs5) {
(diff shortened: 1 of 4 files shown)


juergh:
Patch #1 and #2 should really be authored by @6by9 since they're mostly his work. Also, we should drop the register write retries since they are no longer necessary thanks to patch #3. I think. I'll do some testing to verify that.

mripard:
I've changed the authorship to have @6by9 and you as the main author, and removed the register write retries.

juergh:
I have tested this patchset successfully on top of the current rpi-5.10.y branch (commit 03ab8875d1fc756bd6d2fd8fdb211532eff33062). I also verified that LUT register rewrites are indeed no longer necessary and writes always seem to stick. What's a little concerning though is that the screen blanks for ~2 secs for each update of the LUT (96 register writes). FWIW, doing 100 subsequent LUT updates blanks the screen for ~7 secs. I don't know how GNOME or other tools implement the nightlight feature and how many LUT updates at what intervals are performed but it could result in a significant disturbance. As I understand it, this is a HW limitation and IMO it will trigger bug reports.

6by9:
Sorry, I totally missed @mripard's update.

Slightly weird wording in commit 3's commit text
```
Since the pixelvalve cannot have its assigned channel without stalling
unless it's disabled as well, in our case that means forcing a full
disable / enable cycle on the pipeline.
```
Missing the word "changed", or "halted"?

Other than that, the patches look fine.

> FWIW, doing 100 subsequent LUT updates blanks the screen for ~7 secs. I don't know how GNOME or other tools implement the nightlight feature and how many LUT updates at what intervals are performed but it could result in a significant disturbance. As I understand it, this is a HW limitation and IMO it will trigger bug reports.

It is a hardware limitation, so it's the choice of either blanking, or not supporting the gamma controls.

juergh:
Ok. Then I'd rather have support and let the user choose (and deal with the consequences :-)).

mripard:
I've just pushed a new version with the commit log changed like you suggested (and rebased on the current branch)

6by9:
LGTM

@mripard can I ask you to upstream that if you're happy with it?

mripard:
Yep, sure

6by9:
So X seems to throw a load of calls through drm_crtc_legacy_gamma_set which always replaces the gamma property blob.

I'm just seeing if there is a quick way to do a memcmp before the replace.

PR #4441, opened 2021-07-08
vc4/drm: Pick up more upstream patches

popcornmix:
Now 5.12/5.13 have most commits from 5.10, it becomes clearer what 5.10 is missing from upstream.
This aims to make the drivers/gpu/drm/vc4 diffs between 5.10 and 5.12/5.13 as small as possible.


pelwell:
I can boot kms and fkms with those commits.

6by9:
"drm/vc4: dsi: Only register our component once a DSI device is attached" has shown up some issues in probing I2C configured DSI displays. See https://patchwork.freedesktop.org/patch/375135/
Discussion ongoing upstream as to the best solution.

popcornmix:
@6by9 what's your preference - drop it from this PR, or fix it up when upstream has a solution?
(I'm under the impression that kms dsi is still a work-in-progress).

popcornmix:
I think with this PR the display side of kms is virtually identical between 5.10 and 5.12.
There are a number of missing v3d commits, but they are not so easy to cherry-pick so I'm inclined to leave them.

There are a smaller number of comits new in 5.13 (which look simple on a quick glance), but I'll leave those to a follow up PR.

pelwell:
I'm happy to merge as is to get some testing, with the option to patch later - just clear the Draft flag when you are ready.

popcornmix:
Removed draft.

6by9:
> what's your preference - drop it from this PR, or fix it up when upstream has a solution?
(I'm under the impression that kms dsi is still a work-in-progress).

DSI is my may task at present. It's working on some displays, and I'm trying to get it working with more hence #4423.

This patch will almost certainly break the Pi 7" driver as it is drivers where the panel is registered via I2C (rather than being a child of the DSI host) that it will fail with. That is true for raspberry-pi-touchscreen.

6by9:
(Hit "comment" too early)
Can we drop that one patch for now? I'm not sure it is going to be a quick answer upstream.

popcornmix:
""drm/vc4: dsi: Only register our component once a DSI device is attached" now dropped

PR #4442, opened 2021-07-08
dtoverlays: Add overlay for cap1106 capacitive touch sensor

Mr-Bossman:
Tested works on rpi3, alert pin being pin 4

### DIFF
--- a/arch/arm/boot/dts/overlays/cap1106-overlay.dts
+++ b/arch/arm/boot/dts/overlays/cap1106-overlay.dts
@@ -1,0 +1,54 @@
+// Overlay for cap1106 from  Microchip Semiconductor
+// add CONFIG_KEYBOARD_CAP11XX=y
+
+/dts-v1/;
+/plugin/;
+
+/ {
+        compatible = "brcm,bcm2835";
+        fragment@0 {
+                target = <&i2c1>;
+                __overlay__{
+                        status = "okay";
+                        cap1106: cap1106@28 {
+                                compatible = "microchip,cap1106";
+                                pinctrl-0 = <&cap1106_pins>;
+                                pinctrl-names = "default";
+                                interrupt-parent = <&gpio>;
+                                interrupts = <4 2>;
+                                reg = <0x28>;
+                                autorepeat;
+                                microchip,sensor-gain = <2>;
+
+                                linux,keycodes = <2>,           /* KEY_1 */
+                                                <3>,            /* KEY_2 */
+                                                <4>,            /* KEY_3 */
+                                                <5>,            /* KEY_4 */
+                                                <6>,            /* KEY_5 */
+                                                <7>;            /* KEY_6 */
+
+                                #address-cells = <1>;
+                                #size-cells = <0>;
+                                status = "okay";
+
+                        };
+                };
+        };
+        fragment@1 {
+                target = <&gpio>;
+                __overlay__ {
+                        cap1106_pins: cap1106_pins {
+                                brcm,pins = <4>;
+                                brcm,function = <0>; /* in */
+                                brcm,pull = <0>; /* none */
+                        };
+                };
+        };
+
+        __overrides__ {
+                int_pin = <&cap1106>, "interrupts:0",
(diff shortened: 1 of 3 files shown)


drivers/input/keyboard/cap11xx.c
```diff
@@ -476,6 +485,7 @@ MODULE_DEVICE_TABLE(of, cap11xx_dt_ids);
 
 static const struct i2c_device_id cap11xx_i2c_ids[] = {
 	{ "cap1106", CAP1106 },
+	{ "cap1106", CAP1206 },
```
pelwell: `"cap1206"`?

drivers/input/keyboard/cap11xx.c
```diff
@@ -378,17 +380,21 @@ static int cap11xx_i2c_probe(struct i2c_client *i2c_client,
 	node = dev->of_node;
 
 	if (!of_property_read_u32(node, "microchip,sensor-gain", &gain32)) {
+		if (cap->product_id != 0x67)
```
pelwell: This test would be better using either the index to the table (`!= CAP1206`) or by adding a flags field (`!(cap->flags & FLAGS_SENSOR_GAIN)`).

arch/arm/boot/dts/overlays/cap1106-overlay.dts
```diff
@@ -0,0 +1,53 @@
...
+                        };
+                };
+        };
+        fragment@1 {
+                target = <&gpio>;
+                __overlay__ {
+                        cap1106_pins: cap1106_pins {
+                                brcm,pins = <4>;
+                                brcm,function = <0>; /* in */
+                                brcm,pull = <1>; /* none */
```
pelwell: This doesn't make sense. Your interrupts flags above are 2 (`IRQ_TYPE_EDGE_FALLING`), a pull of `1` means pull _down_ (a falling-edge IRQ would normally require an pull up), and the comment says `none`.

arch/arm/boot/dts/overlays/cap1106-overlay.dts
```diff
@@ -0,0 +1,53 @@
...
+
+/ {
+        compatible = "brcm,bcm2835";
+        fragment@0 {
+                target = <&i2c1>;
+                __overlay__{
+                        status = "okay";
+                        cap1106: cap1106@28 {
+                                compatible = "microchip,cap1106";
+                                pinctrl-0 = <&cap1106_pins>;
```
pelwell, line 15: This is missing `pinctrl-names = "default";`

pelwell (changes requested):
This is two PRs in one - the overlay PR belongs here, but the driver change should be sent to the Linux kernel devs.

The overlay looks largely OK, but it's lacking entries in the README and Makefile, and the driver is not enabled in our defconfigs so most people would be unable to make use of it.

pelwell (commented):
The overlay itself looks OK now, and the Makefile change. You're just missing a README entry - see #4444 for some guidance on that.

Before I merge this PR I'd also like a name and email address for the Signed-off-by: line in the commit message.

Can I assume that you would like `CONFIG_KEYBOARD_CAP11XX` adding to our standard configurations?

Mr-Bossman:
> Before I merge this PR I'd also like a name and email address for the Signed-off-by: line in the commit message.

How exactly do I do this... I'm not used to using GitHub for PRs. Do I add it to the git-commit-log description like on u-boot and the kernel?

> Can I assume that you would like CONFIG_KEYBOARD_CAP11XX adding to our standard configurations?

exactly.

> The overlay itself looks OK now, and the Makefile change. You're just missing a README entry - see #4444 for some guidance on that.

will do!

pelwell:
For the Signed-off-by, just write it in a comment and I can add it when merging, but normally it would be included at the end of your commit message.

pelwell:
The int_pin parameter wasn't documented and there were a few whitespace problems, but it's now in rpi-5.10.y along with a commit adding the required config setting.

PR #4456, opened 2021-07-15
drm/vc4: Fix typo when getting firmware node

popcornmix:


popcornmix:
We seem to get away with this on 5.10, but on 5.13 where firmware_get calls are reference counted this gets unhappy.

Issue #2916, opened 2019-04-01
Ceph Kernel Module Missing.


ajh34:
modprobe: FATAL: Module ceph not found in directory /lib/modules/4.14.98-v7+

Looks like the Ceph package does not install the required kernel module to mount cephfs filesystems. Same issue on stretch and previous release.

JamesH65:
@ajh34 We do not enable CEPH in our kernel builds. It seems like a fairly niche system, and unless its impact on memory and CPU bandwidth is minimal (close to zero), we would be unlikely to incorporate it by default. If you can provide those figures, it would give us something to consider. It seems likely that as a module it impact would indeed be minimal, but we would need to be sure.

imjustmatthew:
Unexpectedly ran into this same issue. I would never have expected a modern kernel didn't have Cpeh included. The performance impact of this as a module should be extremely small when it's not loaded and laoding of the module will only occur when a user actually uses it to mount a filesystem. There is a Ceph-FUSE option, but that is much lower performing that the kernel module.

JamesH65:
@pelwell @popcornmix Any thoughts?

pelwell:
As a filesystem I would expect no impact on performance if it isn't used, but there could be some overhead on the static (non-module) kernel size. The usual expectation on somebody asking for a feature to be added is for them to do a trial build and show the size of any additional modules, and the size of the kernel .img and free memory when it is used, along with the same information for a kernel that differs only in that the feature is not enabled.

StrayLightning:
I built from  fe915debc1925397f78db8bbfc9f488b0ce4e009, following the instructions here: https://www.raspberrypi.org/documentation/linux/kernel/building.md

The resultant image:
    Linux cluster0 4.19.127-v7l-ceph+ #1 SMP Thu Jul 9 15:18:05 BST 2020 armv7l GNU/Linux

The kernel previously:
   Linux cluster1 4.19.118-v7l+ #1311 SMP Mon Apr 27 14:26:42 BST 2020 armv7l GNU/Linux

(Different hostnames, but the nodes were updated at the same time.)

The resultant kernel images:
    -rwxr-xr-x  1 root root  5801056 Jul  6 17:13 kernel7l.img
    -rwxr-xr-x  1 root root  5777480 Jul  9 18:05 kernel-ceph.img

Of course, this is not an apples-to-apples comparison: If I get a chance I'll run a build from the same point that the released kernel was built from.

Changes to the kernel configuration:

    21c21
[...]

PR #4461, opened 2021-07-19
Enabled CEPH RBD and FS

kristvanbesien:
I ran in to an issue when building a Kubernetes cluster on top of raspberry pis. I want to use rook/ceph for storage, but this requires RBD to be enabled.


pelwell:
Essentially a duplicate of https://github.com/raspberrypi/linux/issues/2916.

kristvanbesien:
Will this be considered?

PR #4464, opened 2021-07-20
media: i2c: imx477: Fix framerates for 1332x990 mode

davidplowman:
The imx477 driver's line length for this mode had not been updated to
the value supplied to us by the sensor manufacturer. With this
correction the sensor delivers the framerates that are expected.


davidplowman:
@naushir @6by9 I've tested this with libcamera and it fixes the incorrect framerate problem for this mode.

6by9:
LGTM if that is the number we've been given.

naushir:
LGTM.

Not sure where I got the original number from, must have been a straight typo on my part :(

PR #4470, opened 2021-07-23
More vc4 fixes

### LINKED ISSUE #4465: RPi4/kms: boot hangs if two monitors are connected
With current rpi-update kernel/firmware https://github.com/Hexxeh/rpi-firmware/commit/dd59f0eaeae7a53be8d193c21c4cd83f7a667da7 boot hangs if vc4-kms-v3d is enabled and two monitors are connected (1920x1200 monitors here). With just one monitor connected boot completes successfully.

```
root@raspberrypi:~# uname -a
Linux raspberrypi 5.10.50-v7l+ #1438 SMP Mon Jul 19 10:46:09 BST 2021 armv7l GNU/Linux
root@raspberrypi:~# vcgencmd version
Jul 21 2021 16:21:46     
Copyright (c) 2012 Broadcom
version 6a796bb0062a6c75191c57cba1c13f9300076d02 (clean) (release) (start)
```
serial console log: http://ix.io/3tDb

I initially noticed that on LibreELEC running kernel e6410af

console log from LibreELEC: http://ix.io/3tDe

Bisecting through rpi-update firmwares the last working version is https://github.com/Hexxeh/rpi-firmware/commit/9dbcd6dfed16cc7de4a67b9ca445c4847f68c4c2, the first bad vesion is https://github.com/Hexxeh/rpi-firmware/commit/6d1196a54e7197c719299f9f896340bc02123d3b

ping @mripard

mripard:
Hi,

Here's a bunch of fixes for issues that we currently have. The biggest one is the last one that fixes #4465, and possibly more.

Maxime

### DIFF
--- a/drivers/gpu/drm/vc4/vc4_crtc.c
+++ b/drivers/gpu/drm/vc4/vc4_crtc.c
@@ -32,6 +32,7 @@
 #include <linux/clk.h>
 #include <linux/component.h>
 #include <linux/of_device.h>
+#include <linux/pm_runtime.h>
 
 #include <drm/drm_atomic.h>
 #include <drm/drm_atomic_helper.h>
@@ -42,6 +43,7 @@
 #include <drm/drm_vblank.h>
 
 #include "vc4_drv.h"
+#include "vc4_hdmi.h"
 #include "vc4_regs.h"
 
 #define HVS_FIFO_LATENCY_PIX	6
@@ -501,8 +503,10 @@ int vc4_crtc_disable_at_boot(struct drm_crtc *crtc)
 	enum vc4_encoder_type encoder_type;
 	const struct vc4_pv_data *pv_data;
 	struct drm_encoder *encoder;
+	struct vc4_hdmi *vc4_hdmi;
 	unsigned encoder_sel;
 	int channel;
+	int ret;
 
 	if (!(of_device_is_compatible(vc4_crtc->pdev->dev.of_node,
 				      "brcm,bcm2711-pixelvalve2") ||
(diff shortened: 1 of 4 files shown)


popcornmix:
This didn't fix my issue (either on 5.14 or 5.10). Here is dmesg output on rpi-5.10.y:
https://paste.ubuntu.com/p/ScgTkMkVc3/

My use case is booting to desktop with two hdmi monitors connected.
The panic occurs before desktop is usable.

mripard:
I'm not sure what was your issue?

#4465 was a stall right when DRM was probing, and it's fixed (at least here) by that PR.

You seem to have the same issue that @timg236 has though? Is that the same situation where you would start X and then start chromium. I can't reproduce it here, I've been able to start chromium just fine.

popcornmix:
Yes, I think I am seeing @timg236 's issue.
Are you booting with two hdmi monitors? That seems to make the issue occur almost immediately.
With one monitor it usually occurs pretty quickly, but sometimes you can launch chromium and use it briefly.

mripard:
I think the main difference then is that I'm starting X by hand after the boot, and I'm booting off NFS. I'll have a look with a pristine image on an SD Card on monday.

popcornmix:
Yes, I've used raspi-config to choose boot to desktop / autologin (which is default on the non-lite install).

kwirk:
Apologies to take this a little off topic, but I only just spotted this, having raised anholt/linux#156 (*vc4 memory leak*) earlier today. I'm not sure if these changes resolve anholt/linux#156 or not, but just wanted to make you aware in case I raised my issue in wrong place (I can open a issue here as well if required).

popcornmix:
This would be the better place for the report.
Are you able to test the kernel with this PR applied to see if it fixes things?
(I'm not sure it will, but it does sound like there is a recent regression somewhere).

mripard:
> Yes, I think I am seeing @timg236 's issue.
> Are you booting with two hdmi monitors? That seems to make the issue occur almost immediately.
> With one monitor it usually occurs pretty quickly, but sometimes you can launch chromium and use it briefly.

I gave it a try this morning. Took the latest image on the RPi website (2021-05-07), updated the firmware, and then updated the kernel to current rpi-5.10.y HEAD (7aa66c086ae0).

Booting with two monitors didn't work since we were affected by #4465, this branch allowed to boot further. However, even after starting chromium or moving the pointer around I couldn't get the clk_request_done issue you mentionned

PR #4473, opened 2021-07-25
Update cx231xx-cards.c

Fubbel:
Added support for second source devices distributed from MEDION. The original device was also distributed from the discounters Lidl and ALDI. Lidl with the name Silvercrest, ALDI with the name MEDION. MEDION changed the device-ID from 0x6109 to 0x610A.
This patch was released in 2014 on this site https://www.mail-archive.com/linux-media@vger.kernel.org/msg69768.html , but the driver_info was not correct.

### DIFF
--- a/drivers/media/usb/cx231xx/cx231xx-cards.c
+++ b/drivers/media/usb/cx231xx/cx231xx-cards.c
@@ -996,6 +996,8 @@ const unsigned int cx231xx_bcount = ARRAY_SIZE(cx231xx_boards);
 struct usb_device_id cx231xx_id_table[] = {
 	{USB_DEVICE(0x1D19, 0x6109),
 	.driver_info = CX231XX_BOARD_PV_XCAPTURE_USB},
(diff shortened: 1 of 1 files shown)


pelwell:
This patch is for an unmodified upstream file - the patch should go upstream. Fortunately a one-line patch like this should be readily accepted.

The process for submitting patches is documented here: https://git.kernel.org/pub/scm/linux/kernel/git/next/linux-next.git/tree/Documentation/process/submitting-patches.rst
In this cases the patch should be emailed to:

    Mauro Carvalho Chehab <mchehab@kernel.org>
    linux-media@vger.kernel.org
    linux-kernel@vger.kernel.org

You should run `scripts/checkpatch.pl` from the kernel tree on your patch to make sure you are following the kernel rules, but if you are lucky the maintainer (Mauro Carvalho Chehab) will take pity on you and correct any minor errors.

Post a message here once the patch is accepted upstream and we can back-port it.

PR #4483, opened 2021-07-28
Fixed dead code

ekjotbilling:
'device_destroy(vc_mem_class, vc_mem_devnum)' is never being executed and
needs to be moved insde the 'goto out_class_destroy' block.

### DIFF
--- a/drivers/char/broadcom/vc_mem.c
+++ b/drivers/char/broadcom/vc_mem.c
@@ -333,9 +333,8 @@ vc_mem_init(void)
 	vc_mem_inited = 1;
 	return 0;
 
-	device_destroy(vc_mem_class, vc_mem_devnum);
-
 out_class_destroy:
+	device_destroy(vc_mem_class, vc_mem_devnum);
 	class_destroy(vc_mem_class);
 	vc_mem_class = NULL;
 


JamesH65:
@popcornmix @6by9 @pelwell  Could this be the memory leak we have been looking for?

pelwell:
No - this is not the cause of the memory leak. The device_destroy call correctly belongs in an error path that would never be used.

The OP is correct in that this is dead code, but this is not the solution. To understand the code as it stands it is necessary to go all the way back to when the code used to look like this:
```
	vc_mem_class = class_create(THIS_MODULE, DRIVER_NAME);
	if (IS_ERR(vc_mem_class)) {
		rc = PTR_ERR(vc_mem_class);
		LOG_ERR("%s: class_create failed (rc=%d)", __func__, rc);
		goto out_cdev_del;
	}

	dev = device_create(vc_mem_class, NULL, vc_mem_devnum, NULL,
			    DRIVER_NAME);
	if (IS_ERR(dev)) {
		rc = PTR_ERR(dev);
		LOG_ERR("%s: device_create failed (rc=%d)", __func__, rc);
		goto out_class_destroy;
	}

	vc_mem_proc_entry = create_proc_entry(DRIVER_NAME, 0444, NULL);
	if (vc_mem_proc_entry == NULL) {
		rc = -EFAULT;
		LOG_ERR("%s: create_proc_entry failed", __func__);
		goto out_device_destroy;
	}
	vc_mem_proc_entry->read_proc = vc_mem_proc_read;
	vc_mem_proc_entry->write_proc = vc_mem_proc_write;

	vc_mem_inited = 1;
	return 0;

      out_device_destroy:
	device_destroy(vc_mem_class, vc_mem_devnum);

      out_class_destroy:
	class_destroy(vc_mem_class);
	vc_mem_class = NULL;
```
The /proc entry was replaced by a debugfs entry, with creation failure being ignored. As a result there are no error exit paths for which the device has been successfully created, so the correct fix for the dead code is to delete it.

pelwell:
BTW there's no need to call the usual suspects out by handle (correct or otherwise) - we'll all be notified anyway.

pelwell:
See https://github.com/raspberrypi/linux/commit/b2cfe749cfc614ef29e196781375f551699504d0.

PR #4503, opened 2021-08-03
Adding driver for ROHM BH1745 color sensor

WoodenNautilus:
I took the sources of this driver from https://github.com/ztemt/NX531J_M_kernel/blob/master/drivers/hwmon/nubia_sensors/rohm_bh1745.c
I modified it to be compiled on the version 5.10.y of the kernel.
I slightly modified it so that the "enable" and the "eanble_als_sensor" values defined in rgb_bh1745_parameter_init represents the state of the sensor at the end of the initialization of the module. To acehive that, I modified the rgb_bh1745_init_client function.
As I am writing this, I am realizing it would have been smarter to first commit the original file, then modify it into a second commit.
I also corrected a mistake inside the definition of the rgb_bh1745_id struct which was preventing the loading of the module when declaring the module inside the device-tree.

I modified the hwmon/Kconfig, the hwmon/Makefil and bcm2711_defconfig to generate the module.

In addition, I created an overlay called bh1745 to enable this module.

If you think I sould create a first commit with the original *.c file from the ztemt repos and then a second commit with my modifications, I will delete this Pull Request and then recreate it.


pelwell:
Apart from Pi-specific devices (mainly just soundcards) we don't accept Pull Requests for whole drivers (or patches to existing drivers) - we're trying to reduce the number of downstream patches, not add new ones. You will need to send this upstream.

PR #4507, opened 2021-08-04
Revert "overlays: Update display GPIO declarations" for Goodix

xdsopl:
This reverts commit b7d685c0b1bd1b98af0e9c1f5d43769982bdbfb2 for Goodix

Reverting the change for the Goodix overlay, introduced by that commit, makes the Goodix touch work here again.

The Goodix driver supports the decision to use GPIO_ACTIVE_HIGH for both the irq and reset pins:

From `include/linux/gpio/consumer.h`:
```
struct acpi_gpio_params {
        unsigned int crs_entry_index;
        unsigned int line_index;
        bool active_low;
};
```
From `drivers/input/touchscreen/goodix.c`:
```
static const struct acpi_gpio_params first_gpio = { 0, 0, false };
static const struct acpi_gpio_params second_gpio = { 1, 0, false };
```

### DIFF
--- a/arch/arm/boot/dts/overlays/goodix-overlay.dts
+++ b/arch/arm/boot/dts/overlays/goodix-overlay.dts
@@ -31,7 +31,7 @@
 				interrupt-parent = <&gpio>;
 				interrupts = <4 2>; // high-to-low edge triggered
 				irq-gpios = <&gpio 4 0>; // Pin7 on GPIO header
-				reset-gpios = <&gpio 17 1>; // Pin11 on GPIO header
+				reset-gpios = <&gpio 17 0>; // Pin11 on GPIO header
 			};
 		};
 	};


pelwell:
Your reasoning doesn't look correct - the ACPI code is X86-specific - but I can believe that this patch is needed. The driver is using gpiod descriptors but still driving them in the old way. The sequence is:
```
	error = gpiod_direction_output(ts->gpiod_rst, 0);
...
	error = gpiod_direction_output(ts->gpiod_rst, 1);
...
	error = gpiod_direction_input(ts->gpiod_rst);
```
i.e. drive it low, drive it high, then release it, which looks like faking an active-low, open-drain GPIO with an active-high GPIO.

pelwell:
Thanks!

xdsopl:
Thank you for clarifying!

PR #4511, opened 2021-08-06
Dev/5.14/hevc ctrls 1

jc-kynesim:
Rpivid now compiles and appears to work.  Matching ffmpeg still to come.
A thing worth mentioning is that whilst debugging I found that if this code attempts to create a "bad" (unknown to v4l2-ctrl) control when opened then we get a hard kernel crash.  I've investigated this a little and so far I can't see how this is my fault - the crash occurs in v4l2_ctrl_handler_free (called as part of the error handling) and is due to the ctrl_ref chain apparently having the same node twice.  How this can occur I have no idea, this is a structure I do not touch explicitly - some sort of locking race, but the bug is remarkably consistent so probably not?


popcornmix:
I think first commit is better replaced by a revert of 1e11edb9d4739036df18064236441f90cda284bd and 9bea367779e0f1c3dc5fe0dfe69be7654ce172c2 as they caused the reintroduction of the deleted file (and are not going to be doing anything useful in an unreferenced file).

"v4l2: Add a Greyworld AWB mode" will need adding to the right split file, but that can be a subsequent fix.

jc-kynesim:
Reverting 1e11edb will get rid of the GREYWORLD define in the include file which will break the compile of stuff that wants it

jc-kynesim:
and reverting 9bea367 gets rid of the scaling matrix defines which are still needed - I think it is simpler to just accept that those patch a redundant file and then get rid of it

jc-kynesim:
Do you want me to rework the GREWORLD ctrl patch for the new split world?

popcornmix:
Yes, that would help. On this branch we'll squash broken commits + partial reverts, so the end result will be the same.

jc-kynesim:
I think that small patch is all that is required for greyworld. I'm not entirely sure how to do the partial reverts that you want so can I leave it to you or would you prefer to walk me through the git sequence you want?

PR #4512, opened 2021-08-06
V4L2 codec updates for interlaced decode

6by9:
This puts all the correct signalling in place for interlaced content.

There is still a tweak needed so that the decoder stops on any change of colourspace or interlacing as required by the V4L2 spec, but FFmpeg doesn't appear to follow the spec in ALWAYS stopping and restarting the output (V4L2 CAPTURE) queue on a source change event, so it would most likely stall.

GStreamer 1.14.4 likewise seems to have an issue if you tell it that the output stream should be interlaced in that it tries to propagate it further up the pipeline, and h264parse seems to object. Telling it that it's progressive works fine. I'll try to sort a 1.19 build at some point, or Bullseye using 1.18, and see how that fairs.

@popcornmix It'd be interesting to know what Kodi makes of this change.

Chromium 88 seems to be happy with it (using BBC-News2017 clip from TestMedia), although it's a touch glitchy (1080i50). Nothing is actually deinterlacing the content of course.

### DIFF
--- a/drivers/staging/vc04_services/bcm2835-codec/bcm2835-v4l2-codec.c
+++ b/drivers/staging/vc04_services/bcm2835-codec/bcm2835-v4l2-codec.c
@@ -1071,7 +1071,7 @@ static void op_buffer_cb(struct vchiq_mmal_instance *instance,
 	struct vb2_v4l2_buffer *vb2;
 
 	v4l2_dbg(2, debug, &ctx->dev->v4l2_dev,
-		 "%s: status:%d, buf:%p, length:%lu, flags %u, pts %lld\n",
+		 "%s: status:%d, buf:%p, length:%lu, flags %04x, pts %lld\n",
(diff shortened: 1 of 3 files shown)


jc-kynesim:
I think last time I tried doing a "stream on" on every event we ended up up in an event loop with cropping, but that may be fixed now.  I'm happy to try again.

popcornmix:
@jc-kynesim are you happy with this?
I think it's okay from my testing. The issues I've seen have been in ffmpeg.

jc-kynesim:
If it works for you then push it.

6by9:
Rebased, and updated to assign to vb2->field instead of ```|=``.

6by9:
Added one extra patch that is currently untested that should allow a client specified bytesperline value for the stride.

popcornmix:
I think this can be merged. It allows interlaced flags to be reported to ffmpeg.
There will be a follow up PR with support for actual deinterlace.

PR #4514, opened 2021-08-07
Add midi-uart{2345}-overlay.dts

pjx3:
As per this thread: https://www.raspberrypi.org/forums/viewtopic.php?f=107&t=317249
Overlays for enabling MIDI baud rate for UARTs 2-5 (based on midi-uart0-overlay.dts).

### DIFF
--- a/arch/arm/boot/dts/overlays/Makefile
+++ b/arch/arm/boot/dts/overlays/Makefile
@@ -123,6 +123,10 @@ dtbo-$(CONFIG_ARCH_BCM2835) += \
 	merus-amp.dtbo \
 	midi-uart0.dtbo \
 	midi-uart1.dtbo \
+	midi-uart2.dtbo \
+	midi-uart3.dtbo \
+	midi-uart4.dtbo \
+	midi-uart5.dtbo \
 	minipitft13.dtbo \
 	miniuart-bt.dtbo \
 	mmc.dtbo \
--- a/arch/arm/boot/dts/overlays/README
+++ b/arch/arm/boot/dts/overlays/README
@@ -2087,6 +2087,29 @@ Info:   Configures UART1 (ttyS0) so that a requested 38.4kbaud actually gets
 Load:   dtoverlay=midi-uart1
 Params: <None>
 
+Name:   midi-uart2
+Info:   Configures UART2 (ttyAMA1) so that a requested 38.4kbaud actually gets
+        31.25kbaud, the frequency required for MIDI
+Load:   dtoverlay=midi-uart2
+Params: <None>
+
+Name:   midi-uart3
+Info:   Configures UART3 (ttyAMA2) so that a requested 38.4kbaud actually gets
+        31.25kbaud, the frequency required for MIDI
+Load:   dtoverlay=midi-uart3
+Params: <None>
+
+Name:   midi-uart4
+Info:   Configures UART4 (ttyAMA3) so that a requested 38.4kbaud actually gets
+        31.25kbaud, the frequency required for MIDI
+Load:   dtoverlay=midi-uart4
+Params: <None>
+
+Name:   midi-uart5
+Info:   Configures UART5 (ttyAMA4) so that a requested 38.4kbaud actually gets
+        31.25kbaud, the frequency required for MIDI
+Load:   dtoverlay=midi-uart5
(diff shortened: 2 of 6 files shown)


pelwell:
Can you add README and Makefile entries for each new overlay? Include a Signed-off-by: line in one of the commit messages then it can all get squashed into one neat commit when merged.

pjx3:
Oops, of course, here you go. Hope it's okay.

pelwell:
Can you add some extra blank lines to the README? Two are required to separate entries.

pjx3:
Sorry, wasn't aware of that, done!

pelwell:
Thanks!

retango:
Hi! Thanks for this.. Commenting here because I can´t seem to make these work on a Raspberry Pi 4B. I send midi messages on serial, but received garbled messages, I belive it{s because the baud rate is off. I see that these dts files are based on the BCM2835, but I read the Raspberry pi 4 ttyAMAxx are based on the BCM2711, may this be an issue? Is there a way to change this in these dts files? thanks a lot!

retango:
Just for clarification, I'm trying to use it in uart3, and I can't make it work, I read in forums that it works on uart0, but I already use uart0 to communicate with a teensy. Since I already made a custom PCB for this, I would like to be able to use uart3. Thanks a lot!

pelwell:
@retango A note here to say you'd solved your problem, as opposed to just deleting your comments (a fact I've just discovered after spending some time confirming that the midi-uart* overlays work as expected), would have been more considerate.

retango:
@pelwell Hi Phill,I'm very sorry for this, and thank you so much for looking into the issue. As you say, the problem (which I had posted here as a comment) was not the overlay. As soon as I found out about this, I came here to leave a comment, and saw the option just to delete my original comments. Since I had not got a response, and the comments were up only less than 36 hours, I thought no one had seen them and believed it was better just to delete them so as to not leave a wrong trace. Clearly I was a wrong, sorry again for this.

I still don't know 100% what the problem was, and am planning on doing some further debugging. I was connecting the pi to a midi cable of a simple Midi-USB interface, with the USB end connected to a PC where I read the messages. I usually use this interface with other controllers, and tested it again before posting my original messages. With the Pi, using the midi-uart3 overlay, this iterface was garbling the midi messages. If I sent one mesagge alone, it read nothing, if I sent 4 it got 2 or 3 but garbled, so I assumed it was an issue with baud rate. However after some many more experiments, I tried using with another interface of better quality, and bingo it read the messages fine. 

I don't know if the problem is somehow in my circuit, that I'm sending the wrong mAmps for midi (although I´m using the recommended resistors according to the Midi specs), of if there is a slight baud rate offset, which bothers the "cheap" Midi interface does not the other one.

Again, thanks again for looking  into this, and sorry for not having been considerate, it was not my intention, just a bad judgement call. Thank you a lot a lot again for this. Regards,

    On Monday, February 13, 2023 at 08:43:42 AM GMT-3, Phil Elwell ***@***.***> wrote:  
 
 


@retango A note here to say you'd solved your problem, as opposed to just deleting your comments (a fact I've just discovered after spending some time confirming that the midi-uart* overlays work as expected), would have been more considerate.

pelwell:
> I don't know if the problem is somehow in my circuit, that I'm sending the wrong mAmps for midi (although I used the recommended resistors according to Midi Spec), of if there is a slight baud rate offset, which bothers the "cheap" Midi interface does not the other one.

I checked the output with a logic analyser and the pulse width was 32us, which corresponds to a baudrate of 31.25KHz, which is what is required for MIDI (a fact I first learned nearly 40 years ago).

retango:
@pelwell, 

> > I don't know if the problem is somehow in my circuit, that I'm sending the wrong mAmps for midi (although I used the recommended resistors according to Midi Spec), of if there is a slight baud rate offset, which bothers the "cheap" Midi interface does not the other one.
> 
> I checked the output with a logic analyser and the pulse width was 32us, which corresponds to a baudrate of 31.25KHz, which is what is required for MIDI (a fact I first learned nearly 40 years ago).

Thanks for this, so I must have a problem with the eletronic circuit then, I will debug it.. I was planning first to use oscilloscope to read the midi signal, so you saved me a debugging step!.. Thanks again!

PR #4525, opened 2021-08-12
usb: xhci: workaround for bogus SET_DEQ_PENDING endpoint state

P33M:
See https://github.com/raspberrypi/linux/issues/3981

An unknown unsafe memory access can result in the ep_state variable
in xhci_virt_ep being trampled with a stuck SET_DEQ_PENDING state
despite successful completion of a Set TR Deq Pointer command.

All URB enqueue/dequeue calls for the endpoint will fail in this state
so no transfers are possible until the device is reconnected.

As a workaround, clear the flag if we see it set and issue a new Set
TR Deq command anyway - this should be harmless, as a prior Set TR Deq
command will only have been issued in the Stopped state, and if the
endpoint is Running then the controller is required to ignore it and
respond with a Context State Error event TRB.


pelwell:
The error condition is terminal without this patch, and as a fix it looks plausible.

PR #4535, opened 2021-08-18
Add support for the MAX30102 heart rate and blood oxygen sensor

jclsn:
I have created a device tree overlay for the above sensor with the help of PhilE from the forum. The sensor is now fully recognized. I updated the build config files for all models, though I am not sure about the correct naming there, as they seem to differ in every config. This should be verified.


pelwell:
As discussed on the Forum thread (https://www.raspberrypi.org/forums/viewtopic.php?f=107&t=317957) I'll take this and merge it into the existing i2c-sensor overlay. In the process I'll change the config settings to make the driver a loadable module (`=m` instead of `=y`).

jclsn:
Yes, that makes sense. Thank you for the support! This is probably what makes the Pi so successful in comparison to many other SBC.

pelwell:
Can you fetch your rpi-5.10.y branch again and test the new i2c-sensor overlay? You'll find that it's been force pushed - `git reset --hard rpi-5.10.y` should put you back on track.

jclsn:
Yes, I will try that later. Maybe I should also add an override for the LED intensity.

jclsn:
I just finished compiling. The solution before was better. I know have this view

I think I am missing some interfaces (or I just can't find them).

pelwell:
There were two errors in the overlay - the sensor was being enabled by default, and the I2C interface wasn't being enabled for max30102. Try now.

jclsn:
The pull is taking quite long this time. I have 3d print a case for this thing. Will be back in three hours approximately and get back to you then!

coldspark29:
No, same issue as before. I can't access the interface of the device. The driver is not being loaded at all. I think the flags for adding the module are wrong. I can also not find the CONFIG_MAX30102=m in the config file. Was it removed again?

When I was compiling before I checked that one by using xconfig. I am not sure about the actual name. I just tried it again with xconfig, but can't find it in bcm2711_deconfig afterwards. Strange...

coldspark29:
Ah yes, should be right. It is missing in the kernel though. Did you want me to push from the main Pi kernel tree? I guess not...

jclsn:
Sorry, my bad. I never figured you could commit to my fork and tried to build the master Pi kernel. I am still a bit new to Github. I am pulling the PR version now...

jclsn:
Yes, it looks good. The iio device is visible again. Before merging I would like to play around with the overrides a bit more. This could take me some time, since I have four different sensors lying around here and I will take care of the analog ones first. So I don't really know if overlay is in usable state yet.

pelwell:
There's no rush - take your time.

jclsn:
I've been trying to compile the kernel again, but the Pi is somehow using 5.10.63 instead of 5.10.74. I then tried to upgrade the Pi kernel, fetch upstream sources to 5.10.92 and add the driver again, but when I booted almost no modules were loaded. Is there any way I can just add this module to the existing kernel without recompiling? I tried copying the max30102.ko into the right folder, but the module doesn't seem to be available.

Or how can I tell the Pi to boot the 5.10.74 kernel I compiled?

jclsn:
Seems there was the kmod binary missing when copying the modules to the rootfs. I can load the module with modprobe now, but it doesn't last until the next boot. I also can't see the device, although I am pretty sure that it is connected right. 

Any idea why modprobe doesn't last until the next boot? I think last time I did not have to load the module manually. Just setting

```
dtoverlay=max30102,int_pin=4
```

was enough.

pelwell:
> Any idea why modprobe doesn't last until the next boot?

Calls to modprobe are not persistent - modules will remain loaded until they are explicitly unloaded or the device is rebooted.

If you are asking why modprobe is necessary then it is probably because you haven't run depmod to update modules.alias. You can confirm whether the module is correctly registered with:
```
$ grep max30102 /lib/modules/`uname -r`/modules.alias
```

jclsn:
Hmm output is 

```
alias i2c:max30105 max30102
alias i2c:max30102 max30102
alias of:N*T*Cmaxim,max30105C* max30102
alias of:N*T*Cmaxim,max30105 max30102
alias of:N*T*Cmaxim,max30102C* max30102
alias of:N*T*Cmaxim,max30102 max30102
```

so I guess it works. The sensor isn't appearing in the sysfs though

I realized that the first time I installed the modules on the SD card there was a notification that depmod was missing. I then installed the kmod package and ran it again without errors. Does it maybe not work if it failed the first time?

jclsn:
I also don't understand where the commits have gone. In my repository there is only the commit listed that I want to merge. I just wanted to fix the line in the readme you added, but I am a bit confused.

jclsn:
Yeah, I don't know. I've recompiled and pushed files again to SD card but it doesn't work. The module is added to module.dep

```
kernel/drivers/iio/health/max30102.ko: kernel/drivers/iio/buffer/kfifo_buf.ko kernel/drivers/iio/industrialio.ko kernel/drivers/base/regmap/regmap-i2c.ko
```
Actually this sensor isn't important to the project anymore. I just wanted to get it to work for the sake of completeness, but it is too time-consuming.

jclsn:
Well, I guess something went wrong with my installation on the Pi. Actually I already tested this successfully, so I guess it is good to merge. If there are some improvements I can make another PR.

pelwell:
I had to fix a last minute cut and paste error - the README referred to the MAX17040 rather than the MAX30102), but it's merged now.

jclsn:
Yeah that was the error I wanted to fix as well ;)

I hopefully going to finish building the Yocto image with the new kernel on the weekend and test it out.

jclsn:
@pelwell So I got the task to add this module at last. I can't get it to pop up under ``/sys/bus/iio/devices`` anymore. The kernel module object is there, but not loaded. Loading it doesn't make a difference though. Is it only being loaded when the sensor is recognized? 

Could also be that the sensor is just broken. I already ordered a new one...

pelwell:
Does `i2cdetect -y 1` show anything at address 0x57?

jclsn:
I am not on Raspbian. Built an image with Yocto. I just compiled i2c-tools into the image. ``i2cdetect -l`` doesn't show any output. ``i2cdetect -y 1`` gives ``Error: Could not open file `/dev/i2c-1' or `/dev/i2c/1': No such file or directory``. So it doesn't detect the whole i2cbus?  I do have bcm2835 under ``/sys/bus/i2c/devices/i2c-1/`` though

jclsn:
Seems I did not auto-load the ``i2c_dev`` module. Here is the output:

```
root@raspberrypi:~# i2cdetect -y 1
     0  1  2  3  4  5  6  7  8  9  a  b  c  d  e  f
00:                         -- -- -- -- -- -- -- -- 
10: -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- 
20: -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- 
30: -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- 
40: -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- 
50: -- -- -- -- -- -- -- 57 -- -- -- -- -- -- -- -- 
60: -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- 
70: -- -- -- -- -- -- -- --  
```

Still nothing at ``/sys/bus/iio`` though. The max30102 module is not autoloaded.

pelwell:
And when you use:
```
echo max30102 0x57 | sudo tee /sys/class/i2c-adapter/i2c-1/new_device
```
? Does that cause the module to be loaded and populate /sys/bus/iio?

jclsn:
It causes the device to pop up under ``/sys/bus/i2c/devices``, but not under ``/sys/bus/iio/devices``

pelwell:
Does `sudo modprobe industrialio` do anything? Or is it already loaded (lsmod)? It shouldn't be necessary to load the module explicitly, but while debugging the problem you could try loading it before the `echo` command posted above.

Are there any relevant errors in the `dmesg` output (try `dmesg -l warn,err`)?

jclsn:
industrialio is definitely loaded. I have been using the MCP3008 ADC the whole time. 

``dmesg -l `` does not work. I Guess because I have the Busybox version

Nothing suspicious when I grep though
```
root@healthpi64:~# dmesg | grep warn
[    3.841747] snd_bcm2835: module is from the staging directory, the quality is unknown, you have been warned.
[    3.850643] vc_sm_cma: module is from the staging directory, the quality is unknown, you have been warned.
[    3.925948] bcm2835_mmal_vchiq: module is from the staging directory, the quality is unknown, you have been warned.
[    3.950770] bcm2835_mmal_vchiq: module is from the staging directory, the quality is unknown, you have been warned.
[    3.971885] bcm2835_isp: module is from the staging directory, the quality is unknown, you have been warned.
[    3.977595] bcm2835_codec: module is from the staging directory, the quality is unknown, you have been warned.
[    3.984145] bcm2835_v4l2: module is from the staging directory, the quality is unknown, you have been warned.
[    5.197118] random: 7 urandom warning(s) missed due to ratelimiting
root@healthpi64:~# dmesg | grep err 
[    0.000000] Machine model: Raspberry Pi 4 Model B Rev 1.4
[    0.000000] CPU features: detected: ARM errata 1165522, 1319367, or 1530923
[    0.112362] raspberrypi-firmware soc:firmware: Attached to firmware from 2021-09-30T19:21:03, variant start
[    0.116371] raspberrypi-firmware soc:firmware: Firmware hash is 51215b4f6e3966401ecd99652a35cf1c25069113
[    1.199870] kvm [1]: vgic interrupt IRQ9
[    1.469804] sdhci: Copyright(c) Pierre Ossman
[    1.470513] mmc-bcm2835 fe300000.mmcnr: could not get clk, deferring probe
```

pelwell:
I'm running out of ideas. This is likely due to some peculiarity of the buildroot environment, and therefore low priority, but I'll reply if I think of something.

jclsn:
Alright, I will see if the new sensor makes a difference when it arrives. I think I had a similar issue on Raspbian as well though. I mentioned that before the merge, but I was having a bad conscience, because I let this PR sit here for so long. I will ask the Yocto guys as well.

timg236:
If you are using buildroot then it might be worth using systemd or at least devtmpfs eudev rather than init scripts

jclsn:
No, I am using Yocto, not Buildroot. I am already using systemd on it. I will flash a Raspbian image today though, to see if that makes a difference.

timg236:
Also check that the .dtbos on the sd card are up to date and match the ones from the kernel. It’s easy to end up with stale files. Just guesses though!

jclsn:
So I just installed a fresh copy of Raspian and added 

```
dtparam=i2c1=on
dtparam=i2c_arm=on
dtoverlay=max30102,int_pin=4
```
to the config.txt

I have exactly the same issues as on the Yocto image. The module is not being populated under i2c or iio...

I also tried hooking up a new sensor in case the old one here was broken. Same thing.

pelwell:
I have one of these on order, but it will take about 3 weeks to arrive.

jclsn:
Guess it wouldn’t help if I send you one of mine due to customs, would it? 

Anyway, three weeks are fine by me.

pelwell:
It probably depends where you are. If you were to mark the item as a gift (which it would be) and of low value (also true) then it shouldn't be held up much, but if 3 weeks doesn't sound awful then it's better to wait.

jclsn:
It's fine. I will take care of other projects meanwhile. Thank you for taking care of it.

pelwell:
The max30102 I ordered has arrived. Once I soldered on the headers and wired it up, and the driver loaded and probed as expected.

The only relevant line in config.txt is:
```
dtoverlay=i2c-sensor,max30102
```
I didn't specify `int_pin=4` because that is the default, but it still works if you do.

I do, however, get a kernel crash when I try to read from the device - the `active_scan_mask` is NULL in the interrupt handler. Even with a patched driver and it never seems to get any measurements, but perhaps that's because I've only connected the SDA, SCL, INT, VIN and GND pins.

pelwell:
Looking at the waveforms on a scope, the INT signal looks like a near clone of the SDA signal, as if there is coupling or even a short between the pins. As long as I keep the INT pin disconnected the driver seems to work. I'm putting this down to bad hardware or wiring - everything else seems to work as intended.

jclsn:
Hmm strange that it doesn't work for me then. Maybe my Pi has a problem. I should try another one. Will see if I can make it this weekend. Schedule is full though

jclsn:
Oh I just realized that I had the wrong line in ``config.txt``

I had

```
dtoverlay=max30102
```

and not

```
dtoverlay=i2c-sensor, max30102
```

I am pretty sure that it used to be like that though. Weird. Anyway, I can now load the sensor and read from it without issues.

Thank your for the assistance! Glad we got it working.

jclsn:
Still no idea where to find the actual data buffer. I can turn the LEDs on and off, but no idea where to find the actual ADC reading...

PR #4539, opened 2021-08-20
overlays: qca7000: Fix URL & README

lategoodbye:
I2SE has been acquired by in-tech smart charging. Now the product URL in the overlay is broken. So fix the URL & README accordingly.


lategoodbye:
Thanks

PR #4547, opened 2021-08-25
drm/vc4: Set a default HSM rate

### LINKED ISSUE #4457: drm/vc4: Initialization hangs when HDMI display is *not* connected
I was working on #4406, and after rebasing the changes onto the current `rpi-5.10.y` head (e6410af), I found that my Pi3 freezes when trying to initialize the `vc4` module. This is not related to my changes, as it happens on e6410af without any changes as well.

After some investigation, I found that everything worked fine after connecting an HDMI display. I was then also able to successfully switch to composite output afterwards through `xrandr`.

This seems to be somehow related to the 94bc403 commit - the best fix I could find is:
```diff
diff --git a/drivers/gpu/drm/vc4/vc4_hdmi.c b/drivers/gpu/drm/vc4/vc4_hdmi.c
index 5401b6a6f440..768c26b08c15 100644
--- a/drivers/gpu/drm/vc4/vc4_hdmi.c
+++ b/drivers/gpu/drm/vc4/vc4_hdmi.c
@@ -2316,14 +2316,6 @@ static int vc4_hdmi_bind(struct device *dev, struct device *master, void *data)
                        vc4_hdmi->disable_4kp60 = true;
        }
 
-       /*
-        * We need to have the device powered up at this point to call
-        * our reset hook and for the CEC init.
-        */
-       ret = vc4_hdmi_runtime_resume(dev);
-       if (ret)
-               goto err_put_ddc;
-
        pm_runtime_get_noresume(dev);
        pm_runtime_set_active(dev);
        pm_runtime_enable(dev);
```
although I don't dare proposing that as a permanent fix, because a) it does not seem to make any sense, b) the Pi still freezes after hot-plugging the HDMI display after having booted without one.

Please excuse me for taking the liberty to tag @mripard in advance - you're the author of that commit, maybe you'll make some sense of it.

Please also note that I did not test any of that behavior on Pi4 (yet).

mripard:
When the firmware doesn't setup the HSM rate (such as when booting
without an HDMI cable plugged in), its rate is 0 and thus any register
access results in a CPU stall, even though HSM is enabled.

Let's enforce a minimum rate at boot to avoid this issue.

Fixes #4457

### DIFF
--- a/drivers/gpu/drm/vc4/vc4_hdmi.c
+++ b/drivers/gpu/drm/vc4/vc4_hdmi.c
@@ -96,6 +96,7 @@
 # define VC4_HD_M_SW_RST			BIT(2)
 # define VC4_HD_M_ENABLE			BIT(0)
 
+#define HSM_MIN_CLOCK_FREQ	120000000
 #define CEC_CLOCK_FREQ 40000
 #define HDMI_14_MAX_TMDS_CLK   (340 * 1000 * 1000)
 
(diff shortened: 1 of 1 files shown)


popcornmix:
Confirmed this fixed linked issue. Pi3 booted without hdmi connected.

PR #4548, opened 2021-08-26
Add IMX519 driver and device tree overlay

ArduCAM:


6by9:
Initial comments before I start reviewing the driver.

A couple of checkpatch warnings (I do have --strict enabled by default)
```
ImportError: No module named ply
WARNING: Possible unwrapped commit description (prefer a maximum 75 chars per line)
#9: 
which can be disabled through the ENABLE_VCM_DUMMY macro (enabled by default).

WARNING: added, moved or deleted file(s), does MAINTAINERS need updating?
#25: 
new file mode 100644

WARNING: Possible repeated word: 'all'
#74: FILE: Documentation/devicetree/bindings/media/i2c/imx519.yaml:45:
+      Must be released (set high) after all all supplies and INCK are applied.

WARNING: DT binding docs and includes should be a separate patch. See: Documentation/devicetree/bindings/submitting-patches.rst

CHECK: Please don't use multiple blank lines
#1060: FILE: drivers/media/i2c/imx519.c:868:
+
+

#1498: FILE: drivers/media/i2c/imx519.c:1306:
+static int ak7371_i2c_write(struct imx519 *imx519,
+	u8 addr, u16 data, u8 size)

CHECK: Please don't use multiple blank lines
#1519: FILE: drivers/media/i2c/imx519.c:1327:
+
+
```

Device tree bindings should always be in a separate patch from the driver - they are handled via different trees in mainline, and DT can describe things about the hardware that the driver doesn't support.

Is there anything different about your version of the module to the standard Sony one? If not then it should use Sony's vendor prefix and not an Arducam one.

As discussed by email, VCM control is supported by the framework by using a ```lens-focus``` phandle link from the sensor to the lens driver, and it will be automatically created as an extra v4l-subdev node. You'll probably want a link between the VCM driver and the cam1_reg regulator that controls the sensor power, similar to how https://elixir.free-electrons.com/linux/latest/source/drivers/media/i2c/dw9768.c handles it.
Is there actually a difference between AK7371 and AK7375?

arch/arm/boot/dts/overlays/imx519-overlay.dts
```diff
@@ -0,0 +1,113 @@
...
+		__overlay__ {
+			status = "okay";
+			regulator-name = "imx519_vana";
+			regulator-min-microvolt = <2800000>;
+			regulator-max-microvolt = <2800000>;
+		};
+	};
+
+	__overrides__ {
+		rotation = <&imx519>,"rotation:0";
```
6by9, line 112: Could you add the orientation parameter as done in https://github.com/raspberrypi/linux/commit/c3c2ca3c775644baf80dde6a9a79f745b18ce108 please?

arch/arm/boot/dts/overlays/imx519-overlay.dts
```diff
@@ -0,0 +1,113 @@
+// SPDX-License-Identifier: GPL-2.0-only
```
6by9, line 1: Please add an entry in README for the new overlay.

drivers/media/i2c/Makefile
```diff
@@ -128,3 +128,4 @@ obj-$(CONFIG_VIDEO_RDACM20)	+= rdacm20-camera_module.o
 obj-$(CONFIG_VIDEO_ST_MIPID02) += st-mipid02.o
 
 obj-$(CONFIG_SDR_MAX2175) += max2175.o
+obj-$(CONFIG_VIDEO_IMX519) += imx519.o
```
6by9: Whilst not in alphabetical order (and actually it's our downstream imx477 driver that is out of place), the drivers are at least grouped reasonably.
Please add your driver after IMX355.

drivers/media/i2c/imx519.c
```diff
@@ -0,0 +1,2321 @@
...
+
+/* Pixel rate is fixed at 686MHz for all the modes */
+#define IMX519_PIXEL_RATE		686000000
+
+/* V_TIMING internal */
+#define IMX519_REG_FRAME_LENGTH		0x0340
+#define IMX519_FRAME_LENGTH_MAX		0xffdc
+
+/* Long exposure multiplier */
+#define IMX519_LONG_EXP_SHIFT_MAX	7
```
6by9, line 49: I don't have a datasheet for IMX519. Does it support a long exposure multiplier register in the same way as IMX477?

drivers/media/i2c/imx519.c
```diff
@@ -0,0 +1,2321 @@
...
+	struct v4l2_fract timeperframe_default;
+
+	/* Default register values */
+	struct imx519_reg_list reg_list;
+};
+
+static const struct imx519_reg mode_common_regs[] = {
+	{0x0136, 0x18},
+	{0x0137, 0x00},
+	{0x3C7E, 0x01},
```
6by9: The normal coding style is to use lower case for hex values. We have a bit of a mixture in this driver at present.

drivers/media/i2c/imx519.c
```diff
@@ -0,0 +1,2321 @@
...
+			.num_of_regs = ARRAY_SIZE(mode_4656x3496_regs),
+			.regs = mode_4656x3496_regs,
+		}
+	},
+	{
+		.width = 3840,
+		.height = 2160,
+		.line_length_pix = 0x3870,
+		.crop = {
+			.left = IMX519_PIXEL_ARRAY_LEFT + 0x0198,
```
6by9: Could we use decimal here please? I realise this comes from a register that is programmed in hex (presumably 0x0344/5), but it just makes it more readable. Ditto the other similar offsets.

drivers/media/i2c/imx519.c
```diff
@@ -0,0 +1,2321 @@
...
+ * can start capture (i.e. can leave software standby), given by T7 in the
+ * datasheet is 8ms.  This does include I2C setup time as well.
+ *
+ * Note, that delay between XCLR low->high and reading the CCI ID register (T6
+ * in the datasheet) is much smaller - 600us.
+ */
+#define IMX519_XCLR_MIN_DELAY_US	8000
+#define IMX519_XCLR_DELAY_RANGE_US	1000
+
+struct imx519_compatible_data {
```
6by9: Not needed for this driver as there is only one compatible string.

drivers/media/i2c/imx519.c
```diff
@@ -0,0 +1,2321 @@
...
+/*
+ * Initialisation delay between XCLR low->high and the moment when the sensor
+ * can start capture (i.e. can leave software standby), given by T7 in the
+ * datasheet is 8ms.  This does include I2C setup time as well.
+ *
+ * Note, that delay between XCLR low->high and reading the CCI ID register (T6
+ * in the datasheet) is much smaller - 600us.
+ */
+#define IMX519_XCLR_MIN_DELAY_US	8000
+#define IMX519_XCLR_DELAY_RANGE_US	1000
```
6by9, line 979: This is a straight copy from IMX477. Does it actually apply to IMX519?

drivers/media/i2c/imx519.c
```diff
@@ -0,0 +1,2321 @@
...
+	struct v4l2_subdev sd;
+	struct media_pad pad[NUM_PADS];
+#if ENABLE_VCM_DUMMY
+	struct i2c_client *vcm_client;
+#endif
+
+	unsigned int fmt_code;
+
+	struct clk *xclk;
+	u32 xclk_freq;
```
6by9: I realise this is copied from imx477, but xclk_freq is only accessed in imx477_probe, so can be a local variable there. There's currently no expectation to need to work off multiple clock frequencies.

drivers/media/i2c/imx519.c
```diff
@@ -0,0 +1,2321 @@
...
+	/* Any extra information related to different compatible sensors */
+	const struct imx519_compatible_data *compatible_data;
+};
+
+static inline struct imx519 *to_imx519(struct v4l2_subdev *_sd)
+{
+	return container_of(_sd, struct imx519, sd);
+}
+
+static inline void get_mode_table(unsigned int code,
```
6by9: Not needed as we only have the one table, so it can be accessed directly.
(IMX477 has a different list of modes for 10 and 12 bit).

drivers/media/i2c/imx519.c
```diff
@@ -0,0 +1,2321 @@
...
+	lockdep_assert_held(&imx519->mutex);
+
+	for (i = 0; i < ARRAY_SIZE(codes); i++)
+		if (codes[i] == code)
+			break;
+
+	if (i >= ARRAY_SIZE(codes))
+		i = 0;
+
+	i = (i & ~3) | (imx519->vflip->val ? 2 : 0) |
```
6by9: ```(i & ~3)``` is redundant as this only supports 10 bit modes.

drivers/media/i2c/imx519.c
```diff
@@ -0,0 +1,2321 @@
...
+	{0x0217, 0x00},
+	{0x020E, 0x01},
+	{0x020F, 0x00},
+	{0x0218, 0x01},
+	{0x0219, 0x00},
+};
+
+
+/* Mode configs */
+static const struct imx519_mode supported_modes_10bit[] = {
```
6by9, line 809: Does the sensor support 8 or 12 bit readout, and do they work? If not then this can be called just ```supported_modes```.

drivers/media/i2c/imx519.c
```diff
@@ -0,0 +1,2321 @@
...
+
+#define IMX519_REG_MODE_SELECT		0x0100
+#define IMX519_MODE_STANDBY		0x00
+#define IMX519_MODE_STREAMING		0x01
+
+#define IMX519_REG_ORIENTATION		0x101
+
+#define IMX519_XCLK_FREQ		24000000
+
+#define IMX519_DEFAULT_LINK_FREQ	648000000
```
6by9: Minor note that if this is actually correct, then it exceeds the spec for CSI-2 which is 1Gbit/s/lane, which equates to 500MHz as it is DDR.
Unicam (the CSI-2 receiver) does normally cope with slightly higher data rates, but technically you're exceeding the spec.

6by9 (commented):
A number of these look to be due to copying IMX477 but not supporting some of the features. This leaves a number of chunks of effectively dead code around.

6by9:
For mainline, patches need to have a named author - https://www.kernel.org/doc/html/latest/process/submitting-patches.html#sign-your-work-the-developer-s-certificate-of-origin

Also MAINTAINERS hasn't been updated to report who is maintaining the DT binding and driver files.

ArduCAM:
Hi 6by9

Thank you for your review and suggestions,
There is no difference in control between AK7371 and AK7375 (at least the functions I use are not).

We will resolve your proposed amendments as soon as possible.

Regarding the VCM driver, since the current framework is still a bit imperfect, we plan to use this temporary solution. Is this acceptable?

6by9:
> Regarding the VCM driver, since the current framework is still a bit imperfect, we plan to use this temporary solution. Is this acceptable?

Which part of the framework do you have an issue with? Yes the control will be under a different subdev node, but that's how it will end up being anyway.

ArduCAM:
When the camera is turned on again, it will write the last control value: https://github.com/raspberrypi/linux/blob/e4cb65cf23c78b4912786aaf8467aa215d4e1d88/drivers/media/i2c/imx477.c#L1715
But at present, when the camera is turned on, other subdevs are not notified, which makes it impossible to restore the control value of VCM.
is it right? Maybe I missed something?

6by9:
> But at present, when the camera is turned on, other subdevs are not notified, which makes it impossible to restore the control value of VCM.
is it right? Maybe I missed something?

As per my comment
> As discussed by email, VCM control is supported by the framework by using a lens-focus phandle link from the sensor to the lens driver, and it will be automatically created as an extra v4l-subdev node. You'll probably want a link between the VCM driver and the cam1_reg regulator that controls the sensor power, similar to how https://elixir.free-electrons.com/linux/latest/source/drivers/media/i2c/dw9768.c handles it.

Both the lens driver and sensor drivers need to make a request to the regulator driver for the sensor/lens to be powered on when required. The regulator driver will reference count the requests, and only power down when neither want it.
Each can therefore guarantee that the power will remain on whilst in use, and know when they need to reinitialise.

There is a need to be careful that a regulator request to power down and up does not necessarily result in an actual reset if there are other users of the regulator. Most sensors use the software reset at the start of their initialisation sequence anyway, so that's not normally an issue.

ArduCAM:
> Both the lens driver and sensor drivers need to make a request to the regulator driver for the sensor/lens to be powered on when required. The regulator driver will reference count the requests, and only power down when neither want it.
Each can therefore guarantee that the power will remain on whilst in use, and know when they need to reinitialise.

I understand.
My original purpose was this:
Open the camera through the video node (using libcamera or other methods), at this time use the v4l2-ctl command to set the focus value, and when the video node is closed and turned on again, the focus value set last time can be restored.
Since the v4l2-ctl command will execute the Open -> Settings Control -> Close process, it seems that the goal we want cannot be achieved.

I noticed that regulator_register_notifier may achieve my purpose, but I am not sure whether this is the correct usage.

drivers/media/i2c/imx519.c
ArduCAM: Yes, it is the same as imx477.

drivers/media/i2c/imx519.c
ArduCAM: imx519 supports RAW8, but I haven't added RAW8 config yet.

drivers/media/i2c/imx519.c
ArduCAM: This should be my calculation error. 
The actual frequency was 525Mhz before, but it has been reduced to 493.5Mhz.

ArduCAM:
> For mainline, patches need to have a named author - https://www.kernel.org/doc/html/latest/process/submitting-patches.html#sign-your-work-the-developer-s-certificate-of-origin
> 
> Also MAINTAINERS hasn't been updated to report who is maintaining the DT binding and driver files.

Not sure if the link in MAINTAINERS is correct...
https://github.com/ArduCAM/linux/blob/0002a4e39a4c062526927d0fc3588edf009033b5/MAINTAINERS#L16379

pelwell:
I'm happy with this PR from an overlay, Kconfig, Makefile and commit partitioning perspective - I'll leave the driver reviewing to @6by9.

The pachyderm in the parlour is that we expect non-Pi-specific drivers to be submitted upstream - we can't commit to supporting them ourselves indefinitely. What are your plans for upstreaming?

drivers/media/i2c/imx519.c
ArduCAM: From the data sheet, they are the same.

ArduCAM:
Hi pelwell,
Thanks for your review.

I’m not sure about the workflow here,
At first I worried that submitting the same thing in two places would cause conflicts. If there is no conflict, I think we can try to submit it to git://linuxtv.org/media_tree.git

6by9:
> My original purpose was this:
Open the camera through the video node (using libcamera or other methods), at this time use the v4l2-ctl command to set the focus value, and when the video node is closed and turned on again, the focus value set last time can be restored.
Since the v4l2-ctl command will execute the Open -> Settings Control -> Close process, it seems that the goal we want cannot be achieved.
I noticed that regulator_register_notifier may achieve my purpose, but I am not sure whether this is the correct usage.

OK, so it's effectively the "default" case where the sensor is being used and the lens not being cared about by a particular app, so it needs to be restored when the sensor is powered up.

In which case I believe you are correct that ```regulator_register_notifier``` looking for event ```REGULATOR_EVENT_ENABLE``` should achieve what you want, and initialise the VCM from that notifier (do not call pm_runtime or enable the regulator from there). 
```ak7375_vcm_resume``` then wants to either enable the regulator (which will call the notifier), or in the case of no regulator being configured it should do the setup there. This is an existing VCM driver, therefore you can't make the regulator mandatory, so it needs to be requested via ```regulator_get_optional``` (slightly annoyingly there appears to be no devm_ version of that - such is life).

Using ```devm_regulator_register_notifier``` to get a managed version may be more useful, but there are potential race conditions on shutdown if you're not careful (state structures may be destroyed whilst the notifier is still registered - see https://github.com/raspberrypi/linux/commit/c6fa884e792822e33789622bbb74831f54e02529, although that was interrupts rather than regulators).

> I’m not sure about the workflow here,
At first I worried that submitting the same thing in two places would cause conflicts. If there is no conflict, I think we can try to submit it to git://linuxtv.org/media_tree.git

I was offering to do a first pass review on the driver before it was sent upstream. They are likely to have their own comments, but knocking out the initial obvious things saves some time and effort there (I find it isn't necessarily the easiest mailing list to deal with).
Embedding the VCM driver into the sensor driver was not going to be acceptable to them, but you could ask the question on linux-media whether the proposed solution is acceptable.

> Not sure if the link in MAINTAINERS is correct...
https://github.com/ArduCAM/linux/blob/0002a4e39a4c062526927d0fc3588edf009033b5/MAINTAINERS#L16379

This should be as if it is being submitted upstream, so ```git git://linuxtv.org/media_tree.git```.
You don't need a ```W``` entry.

File: drivers/media/i2c/imx519.c (hunk 12 above)
Comment (6by9) [reply drivers/media/i2c/imx519.c:809]: Is there a sensible use case for RAW8? Can it actually run faster or similar? If not then I wouldn't worry about it.

You may get others adding support at a later date (as happened for imx219), but that doesn't need to be your concern initially.

File: drivers/media/i2c/imx519.c [hunk 14]
```diff
@@ -0,0 +1,2107 @@
...
+
+	lockdep_assert_held(&imx519->mutex);
+
+	for (i = 0; i < ARRAY_SIZE(codes); i++)
+		if (codes[i] == code)
+			break;
+
+	if (i >= ARRAY_SIZE(codes))
+		i = 0;
+
```
6by9: We don't need this bit either as it's also still related to supporting multiple bit depths.

drivers/media/i2c/imx519.c
```diff
@@ -0,0 +1,2107 @@
...
+	return codes[i];
+}
+
+static void imx519_set_default_format(struct imx519 *imx519)
+{
+	/* Set default mode to max resolution */
+	imx519->mode = &supported_modes_10bit[0];
+	imx519->fmt_code = MEDIA_BUS_FMT_SRGGB10_1X10;
+}
```
6by9: This is only called from one place in imx519_probe, so may as well just be inserted there.

6by9 (commented):
DT bindings needs to come before the driver. MAINTAINERS normally gets updated once with the bindings to add both bindings and the driver files, eg https://github.com/torvalds/linux/commit/8a97a4676f8b1badcd9cfbed2b081342847bb1b1

By copying imx477 I'm afraid you've got caught in the cross-fire slightly between us and mainline over supporting multiple streams (ie image with embedded metadata). Mainline still haven't resolved this, although patches have been kicked around recently. We adopted just advertising 2 pads for the two streams, but that isn't be acceptable to upstream.

If you're looking to upstream this, then the references to METADATA_PAD need to be removed, and all the pad index checks become ```if (thing->pad)``` instead of ```if (things->pad >= NUM_PADS)```. I'm happy to accept an additional patch to our tree which adds the second pad back in, and proper support for the streams can be added when they've decided how to do it.

drivers/media/i2c/imx519.c
ArduCAM: Maybe I can keep this? Modifying the code here needs to change a way to obtain Bayer order, and also to adapt to various calls.

6by9: You're computing a value of ```i``` which is then not used.
If you add support for 8 bit readout from the start then you need to retain this, but otherwise it's redundant. If sent upstream then you'd get a very similar comment as it serves no purpose.

It does make ```code``` unused too, so that could be removed too.

ArduCAM: Sorry, I confuse the method of calculating the Bayer order here with the method used elsewhere.
The `i` here is indeed not used.

ArduCAM:
Hi 6by9
Thank you very much for your review and suggestions,
I realize that I have made a lot of stupid mistakes that have caused you unnecessary trouble, and I am sorry for that.

6by9:
No problems. I'm happy to review drivers before submission upstream as I recognise that it is a fairly steep learning curve.
Interactions on the linux-media mailing list can be slow and tedious, and often disheartening, with a number of seemingly unwritten rules that you have to pick up over time.

My comment still applies over the METADATA_PAD - that isn't the way upstream are going to want it, so simplest to implement just exposing video initially. Add the metadata pad later on if we have code that will parse it.
I need to investigate the upstream multiplexed stream support and see how it is going to work - https://patchwork.linuxtv.org/project/linux-media/list/?series=6197 / https://git.kernel.org/pub/scm/linux/kernel/git/tomba/linux.git/log/?h=multistream/work-v8

ArduCAM:
Yes, I saw your description of METADATA_PAD, I was thinking maybe we have a chance to merge it into the Raspberry Pi kernel first?

ArduCAM:
Hi @6by9 
If we want to merge into the kernel of raspberry pi first, do we need to delete MATEDATA_PAD?

6by9:
Sorry, I'd missed your previous comment.

We try to minimise any differences to mainline, unless a feature or change is totally unacceptable to mainline (eg the abuse of pads for the metadata stream).
Even those changes should be made as an extra patch on top of what should be acceptable to mainline.

So it really would be preferable to have the METADATA_PAD change split out as a separate patch. You need to do that for mainline anyway.

ArduCAM:
Hi @6by9 
Thanks for your suggestion, do you think this patch is okay?

ArduCAM:
Hi @6by9 
_Just in case you missed this._

6by9:
Apologies, I had missed the update again.

That looks good now. There may be a couple of quibbles upstream over minor details of the implementation, but most of it looks fine to me.
I'm trusting that it actually produces images and the controls do what they're supposed to.

pelwell:
These patches required some adjustments for the newer kernels, but they're in rpi-5.14.y and rpi-5.15.y now (along with blank lines after the first lines of the commit messages).

ArduCAM:
Hi @pelwell 

> These patches required some adjustments for the newer kernels, but they're in rpi-5.14.y and rpi-5.15.y now (along with blank lines after the first lines of the commit messages).

Do I need to make any changes to this?

pelwell:
You tell me - it all compiles, but is otherwise untested.

ArduCAM:
Got it, I will test it on the new version of the kernel.

ArduCAM:
Hi @pelwell 
After testing everything looks fine.
2021-10-18-103352_1920x1080_scrot
2021-10-19-031608_1920x1080_scrot

pelwell:
Excellent.

PR #4559, opened 2021-09-02
media: rpivid: Avoid returning EINVAL to a G_FMT ioctl

jc-kynesim:
V4L2 spec says that G/S/TRY_FMT IOCTLs should never return errors for
anything other than wrong buffer types. Improve the capture format
function such that this is so and unsupported values get converted
to supported ones properly.


jc-kynesim:
This is a small step towards getting the decoder ready for upstream.  It should cherry-pick onto 5.14 cleanly.

popcornmix:
Looks sane. @6by9?

drivers/staging/media/rpivid/rpivid_video.c
```diff
@@ -109,18 +113,31 @@ int rpivid_prepare_src_format(struct v4l2_pix_format_mplane *pix_fmt)
...
+		width = RPIVID_MAX_WIDTH;
+	if (!height)
+		height = RPIVID_DEFAULT_HEIGHT;
+	if (height > RPIVID_MAX_HEIGHT)
+		height = RPIVID_MAX_HEIGHT;
+
 	/* For column formats set bytesperline to column height (stride2) */
+	switch (pix_fmt->pixelformat) {
+	default:
+		pix_fmt->pixelformat = RPIVID_DEFAULT_DST_PIXELFORMAT;
```
6by9: I'd actually make this explicitly V4L2_PIX_FMT_NV12_COL128, otherwise if someone changes RPIVID_DEFAULT_DST_PIXELFORMAT you still fall through into the NV12_COL128 case when it isn't.

6by9 (commented):
I haven't checked through all the paths, but that looks reasonable other than my one comment.

drivers/staging/media/rpivid/rpivid_video.c
jc-kynesim: You have a point.  But I really want everything that is doing defaulting to change to the same thing - would it be OK to have the default case set pix_fmt & then goto the switch? It is mucky but it is the cleanest way of handling a changed default I can think of. (or check with #if and #error if not the expected value?)

6by9: I'd be totally happy to validate and default if necessary the pixel format first, and then you handle alignments/widths for the supported formats.

6by9:
I've just noticed you still have 
```
	if (pix_fmt->pixelformat != V4L2_PIX_FMT_HEVC_SLICE)
		return -EINVAL;
```
in ```rpivid_prepare_src_format```. 
That can probably be an unconditional ```pix_fmt->pixelformat = V4L2_PIX_FMT_HEVC_SLICE;``` as there are no other valid pixelformats supported on the source side. I haven't double checked that the ```struct v4l2_pix_format_mplane *pix_fmt``` passed in is the one supplied from userspace though.

jc-kynesim:
Fair point.
I've just found that if I do
`#if RPIVID_DEFAULT_DST_PIXELFORMAT != V4L2_PIX_FMT_NV12_COL128`
then I get compile errors as the __u32 used in the definition of v4l2_fourcc isn't defined when the preproc comes to look at the #if :-(

drivers/staging/media/rpivid/rpivid_video.c
jc-kynesim: After going round a few circles trying to come up with something I believed in I've come back to your original suggestion of just setting it to V4L2_PIX_FMT_NV12_COL128.  I've also removed the definition of RPIVID_DEFAULT_DST_PIXELFORMAT as that was the only place it was used and having it would just add confusion.

jc-kynesim:
I think that addresses @6by9 concerns

6by9:
LGTM. I have no in-depth knowledge of the operation, so was mainly looking at it from a V4L2 and coding style perspective.

jc-kynesim:
Well these changes are only to make v4l2 compliance happy, they don't fix anything in normal use, so that seems like a valid perspective to me.

6by9:
I would have been equally as happy with retaining RPIVID_DEFAULT_DST_PIXELFORMAT and having
```
switch (pix_fmt->pixelformat) {
  case V4L2_PIX_FMT_NV12_COL128:
  case V4L2_PIX_FMT_NV12_10_COL128:
    break;
  default:
   pix_fmt->pixelformat = RPIVID_DEFAULT_DST_PIXELFORMAT;
   break;
}

switch (pix_fmt->pixelformat) {
   //Your existing format validation.
}
```

jc-kynesim:
This is simple, doesn't require matching case statements, and I'll come back to it if I ever need to have two places where the format is defaulted that therefore need to match.

jc-kynesim:
Ping?

drivers/staging/media/rpivid/rpivid_video.c
```diff
@@ -27,6 +27,8 @@
 
 #define RPIVID_MIN_WIDTH	16U
 #define RPIVID_MIN_HEIGHT	16U
+#define RPIVID_DEFAULT_WIDTH	1920U
+#define RPIVID_DEFAULT_HEIGHT	1088U
```
pelwell, line 31: Why 1088? The bare literal in the code was 1080.

jc-kynesim: Because I'm expecting an HD stream to be 1088 high cropped to 1080.  I'm just picking what I hope to be a better number.  In actual use the default will probably be never used and the sizes will be explicitly set by the user or inferred from the SPS.

PR #4565, opened 2021-09-06
7inch DSI panel fixes

6by9:
These patches get the 7" panel working reliably for me on a Pi4.

@ghollingworth you were wanting to test.

This is still on the old panel driver, so the touchscreen side still won't work as it can't hook into the regulator control, and the TS controller can get powered down underneath it.


pelwell:
LGTM.

6by9:
Update incoming - I've found another underlying issue.

https://elixir.bootlin.com/linux/v5.10.1/source/drivers/gpu/drm/panel/panel-raspberrypi-touchscreen.c#L274
```
	/* Wait for nPWRDWN to go low to indicate poweron is done. */
	for (i = 0; i < 100; i++) {
		if (rpi_touchscreen_i2c_read(ts, REG_PORTB) & 1)
			break;
	}
```
That does the opposite to intended (at least in comparison to the firmware and comment), and explains why we have a race condition.

The same issue exists in https://elixir.bootlin.com/linux/v5.10.1/source/drivers/regulator/rpi-panel-attiny-regulator.c#L43, and that looks to help with the updated driver too.

Touchscreen driver support still seems to be intermittent - we seem to have race conditions on accessing the Atmel that cause -121 EREMOTEIO failures, and those throw everything out of whack.

6by9:
Added WIP tag as the touchscreen side still isn't quite right and is reverted in the dtoverlay, but I seem to have progress generally in making it reliable.

6by9:
WIP tag removed as I think I've got to the point where this is stable and fully working with touch!!!

I am seeing a flip_done timed out on the DSI screen if ```ignore_lcd=1``` isn't set. After the timeout (30secs) it generally comes to life though.
I suspect that the DSI block is not being reset back to an idle state on vc4_dsi load, therefore it charges in still running, the HVS and PVs get reset, but DSI is in a weird state and won't prod them in an appropriate way to start getting pixels.

```disable_touchscreen=1``` must be set under all conditions.

**Comment by JamesH65** [comment]
Is there any way to say "If KMS, set ignore_lcd=1; set disable_touchscreen=1", as it seems that you would never want it any other way.

File: drivers/regulator/rpi-panel-attiny-regulator.c [hunk 1]
```diff
@@ -21,63 +22,146 @@
...
-	ret = regmap_read(rdev->regmap, REG_POWERON, &data);
-	if (ret < 0)
-		return ret;
+	mutex_lock(&state->lock);
 
-	if (!(data & BIT(0)))
-		return 0;
+	for (i = 0; i < 10; i++) {
+		ret = regmap_read(rdev->regmap, REG_PORTC, &data);
+		if (!ret)
```
JamesH65, line 154: Bit confused as to the possibility that regmap_read returning 0 will keep looping, but isn't 0 a valid value given the if (ret <0) below?

pelwell:
The firmware interactions will be adjusted to remove the need for explicit settings by the user.

6by9:
Rebased, and now actually includes edt-ft5406.dtsi :-)

Firmware side MR created to handle auto detecting the display.

I believe this is good for review now.

drivers/gpu/drm/vc4/vc4_dsi.c
```diff
@@ -1717,6 +1719,15 @@ static int vc4_dsi_bind(struct device *dev, struct device *master, void *data)
 	pm_runtime_enable(dev);
 
 	return 0;
+
+err_free_dma:
+	if (dsi->reg_dma_chan)
```
pelwell: The predicates here and in unbind bother me a bit - it shouldn't be possible to get here without a valid channel, and we don't zero the pointers after releasing the resources.

drivers/gpu/drm/panel/panel-raspberrypi-touchscreen.c
```diff
@@ -257,7 +257,7 @@ static void rpi_touchscreen_i2c_write(struct rpi_touchscreen *ts,
```
pelwell, line 257: The commit message's reference to `ts->dev` should be either `ts->dsi->dev` or just `it`.

drivers/regulator/rpi-panel-attiny-regulator.c
```diff
@@ -37,11 +37,24 @@ static const struct regmap_config attiny_regmap_config = {
...
 	unsigned int data;
+	int ret, i;
 
 	regmap_write(rdev->regmap, REG_POWERON, 1);
+	msleep(80);
+
 	/* Wait for nPWRDWN to go low to indicate poweron is done. */
-	regmap_read_poll_timeout(rdev->regmap, REG_PORTB, data,
-					data & BIT(0), 10, 1000000);
+	for (i = 0; i < 20; i++) {
```
pelwell: It's not clear where the 20 comes from - 20 * 10,000 is less than 1,000,000, so does the regmap_read take a significant amount of time?

drivers/regulator/rpi-panel-attiny-regulator.c
```diff
@@ -37,11 +37,24 @@ static const struct regmap_config attiny_regmap_config = {
...
+		ret = regmap_read(rdev->regmap, REG_PORTB, &data);
+		if (!ret) {
+			if (data & BIT(0))
+				break;
+		}
+		usleep_range(10000, 12000);
+	}
+	usleep_range(10000, 12000);
```
pelwell: What does the extra delay achieve?

drivers/regulator/rpi-panel-attiny-regulator.c
```diff
@@ -101,22 +125,36 @@ static const struct regulator_desc attiny_regulator = {
 
 static int attiny_update_status(struct backlight_device *bl)
 {
-	struct regmap *regmap = bl_get_data(bl);
+	struct attiny_lcd *state = bl_get_data(bl);
```
pelwell, line 188: This change comes out of nowhere - where is the corresponding change to devm_backlight_device_register?

pelwell: Ah - it's in the next commit. I think this and the next line belong there instead.

drivers/regulator/rpi-panel-attiny-regulator.c
```diff
@@ -101,22 +125,36 @@ static const struct regulator_desc attiny_regulator = {
...
+	}
+
+	return ret;
 }
 
 static int attiny_get_brightness(struct backlight_device *bl)
 {
-	struct regmap *regmap = bl_get_data(bl);
-	int ret, brightness;
+	struct attiny_lcd *state = bl_get_data(bl);
```
pelwell, line 212: Ditto here.

arch/arm/boot/dts/overlays/edt-ft5406.dtsi
```diff
@@ -0,0 +1,54 @@
...
+		};
+	};
+
+	__overrides__ {
+		sizex = <&ft5406>,"touchscreen-size-x:0";
+		sizey = <&ft5406>,"touchscreen-size-y:0";
+		invx = <0>, "-10";
+		invy = <0>, "-11";
+		swapxy = <&ft5406>,"touchscreen-swapped-x-y?";
+		disable_touch = <0>, "-10-11-12-13";
```
pelwell: This parameter needs to be documented.

arch/arm/boot/dts/overlays/vc4-kms-dsi-7inch-overlay.dts
```diff
@@ -6,51 +6,109 @@
 /dts-v1/;
 /plugin/;
 
+#include "edt-ft5406.dtsi"
+
 / {
 	compatible = "brcm,bcm2835";
```
pelwell: The compatible string is inherited from edt-ft5406.dtsi, soi this is superfluous.

arch/arm/boot/dts/overlays/edt-ft5406.dtsi
```diff
@@ -0,0 +1,54 @@
+/*
+ * Device Tree overlay for an EDT FT5406 touchscreen
+ *
+ */
+
+/ {
+	compatible = "brcm,bcm2835";
+
+	fragment@10 {
```
pelwell, line 11: Starting the shared numbering at 10 is an unusual decision, but it should work now (there was a time when it wouldn't have).

arch/arm/boot/dts/overlays/vc4-kms-dsi-7inch-overlay.dts
```diff
@@ -6,51 +6,109 @@
...
+					};
 				};
 			};
 		};
 	};
 
 	fragment@1 {
+		target-path = "/";
+		__overlay__ {
+			panel_disp1: panel_disp1@0 {
```
pelwell, line 54: What do the addresses + `reg`s on these nodes achieve?

6by9: We have 2 DSI interfaces, so 2 instances of this overlay is possible with i2c_csi_dsi -> i2c_vc, and dsi1 -> dsi0. Regulators, panels, bridges, etc therefore all need to be unique.
Whilst we haven't done the override magic as yet, it seemed to make sense to be prepared.

pelwell: Understood.

arch/arm/boot/dts/overlays/edt-ft5406.dtsi
6by9: This was a question from a while back - was there benefit in having edt-ft5406 distinct from vc4-kms-dsi-7inch. There's nothing stopping running one without the other, so sharing the config in a dtsi was preferred.  The fragment numbers therefore need to be unique, but we could shift the ones in vc4-kms-dsi-7inch instead.

drivers/gpu/drm/vc4/vc4_dsi.c
6by9: Allocating a channel is dependent on ```if (dsi->variant->broken_axi_workaround)``` which is only set for a compatible string of ```bcm2835-dsi1``` (not ```bcm2835-dsi0``` or ```bcm2711-dsi1```). So it is totally possible to get here without a channel due to some other failure when using dsi0 or 2711.

I'm happy to add in zeroing the pointers, but I don't believe there is a path where that would cause issues.
- If bind failed, you won't get an unbind. 
- If already called unbind, you won't get another unbind, and they will have been initialised again before this error path can be taken again in a new bind.

drivers/gpu/drm/panel/panel-raspberrypi-touchscreen.c
6by9: Done

arch/arm/boot/dts/overlays/edt-ft5406.dtsi
6by9: Done

arch/arm/boot/dts/overlays/vc4-kms-dsi-7inch-overlay.dts
6by9: Done. Comment inserted to make it obvious as to why.

arch/arm/boot/dts/overlays/edt-ft5406.dtsi
pelwell: > we could shift the ones in vc4-kms-dsi-7inch instead.

That isn't necessary, but it is less surprising.

drivers/gpu/drm/vc4/vc4_dsi.c
pelwell: I wasn't suggesting there was a real bug here, and you've accounted for one of the tests, so I'm OK as it is.

drivers/regulator/rpi-panel-attiny-regulator.c
6by9: It's trying to work around dubious code in the Atmel - if it doesn't have time between commands then it can hard lock up whilst holding SDA low.

The REG_POWERON, 1 command does a load of processing from within the ISR, including delays, and bitbashing an I2C command to the TC358762 bridge chip. Hopefully the msleep covers that, but otherwise we poll a few times. It should never come anywhere near the 1s that it used to delay for.

As it happens I've never seen the PWRDNX line from the Toshiba go low, so a later patch removes this polling totally as part of switching away from REG_POWERON.

pelwell: So the increased delays are between attempts, not the total time for all attempts. That's fine.

6by9: Syntax is ```regmap_read_poll_timeout(map, addr, val, cond, sleep_us, timeout_us)```, so it was polling every 10us up to a max of 1s. That's crazy fast, but also aborted should any of the transactions fail.

We're now sleeping for 80ms, then polling every 10ms for 20 attempts (ie 200ms max), so total time in the failure path actually comes down from 1s to 280ms (plus I2C transaction time).

arch/arm/boot/dts/overlays/edt-ft5406.dtsi
6by9: I've added a comment to the top of edt-ft5406.dtsi to explain the odd fragment numbers.

drivers/regulator/rpi-panel-attiny-regulator.c
6by9: Mea culpa. Fixed.

drivers/gpu/drm/vc4/vc4_dsi.c
6by9: I'd already updated it to zero the pointers, so it is done.

6by9:
Updated the PR to hopefully address all the issues raised.

arch/arm/boot/dts/overlays/README
```diff
@@ -3365,10 +3365,15 @@ Params: <None>
...
         Requires vc4-kms-v3d to be loaded.
-Load:   dtoverlay=vc4-kms-dsi-7inch
-Params: <None>
+Load:   dtoverlay=vc4-kms-dsi-7inch,<param>=<val>
+Params: sizex                   Touchscreen size x (default 800)
+        sizey                   Touchscreen size y (default 480)
+        invx                    Touchscreen inverted x axis
+        invy                    Touchscreen inverted y axis
+        swapxy                  Touchscreen swapped x y axis
+        disable_touch           Disables the touch screen overlay driver
```
pelwell, line 3376: Annoyingly the edt-ft5406 overlay has also acquired this parameter.

6by9: Fixed. Moved that override to vc4-kms-dsi-7inch.

pelwell:
Thanks!

PR #4570, opened 2021-09-08
media: i2c: imx477: Allow control of on-sensor DPC

davidplowman:
A module parameter "dpc_enable" is added to allow the control of the
sensor's on-board DPC (Defective Pixel Correction) function.

This is a global setting to be configured before using the sensor;
there is no intention that this would ever be changed on-the-fly.

### DIFF
--- a/drivers/media/i2c/imx477.c
+++ b/drivers/media/i2c/imx477.c
@@ -21,6 +21,10 @@
 #include <media/v4l2-fwnode.h>
 #include <media/v4l2-mediabus.h>
 
+static int dpc_enable = 1;
+module_param(dpc_enable, int, 0644);
+MODULE_PARM_DESC(dpc_enable, "Enable on-sensor DPC");
+
 #define IMX477_REG_VALUE_08BIT		1
 #define IMX477_REG_VALUE_16BIT		2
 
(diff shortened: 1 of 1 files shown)


davidplowman:
@naushir @6by9 Here's a small but important change that our astrophotography friends will need if they want to use libcamera.

I haven't bothered distinguishing between different kinds correction as we did previously because it turned out not to matter. Also I don't want to give anything away! The use of a module_param seems to mimic the VC implementation more-or-less directly, and I don't see any point in this being (for example) a V4L2 control. Would others agree?

naushir:
A module param seems reasonable and allows users to switch this on/off without rebooting. 
LGTM

pelwell:
Any objections, @6by9?

6by9:
I'm never totally keen on using ```!!```.

You can define a module parameter as being of type ```bool``` instead of ```int```, but even then you really should check whether it is true or not, rather than taking ```true``` to be ```0x01``` when programming it into a sensor register.

pelwell:
Using `!!` to create a boolean is a common kernel idiom.

6by9:
> Using !! to create a boolean is a common kernel idiom.

Yes, but in this case it's then used as a u32 to program a sensor register
```imx477_write_reg(imx477, 0x0b05, IMX477_REG_VALUE_08BIT, !!dpc_enable);```
relying on ```true``` being ```0x01```.

```dpc_enable ? 1 : 0``` would be more explicit.

JamesH65:
On a different subject and more for my own interest,; I thought module parameters were now frowned upon, and upstream kernel dev preferred device tree to set this sort of thing? Or is this not that sort of thing.

6by9:
> I thought module parameters were now frowned upon, and upstream kernel dev preferred device tree to set this sort of thing? Or is this not that sort of thing.

Device tree is meant to define the hardware. Enabling DPC isn't describing the hardware, therefore it doesn't fit directly under DT.

Module parameters aren't the nicest as this means we can't enable it for one IMX477 and disable for a 2nd connected to the same CM, but adding it as a fully-fledged V4L2 control is more hassle than it's worth.

pelwell:
> relying on `true` being `0x01`.

This is C, not C++ - there are no bools, and stdbool.h defines `true` to be `1`.

PR #4571, opened 2021-09-08
drm/vc4: Fix out of order frames during asynchronous page flips

mripard:
When doing an asynchronous page flip (PAGE_FLIP ioctl with the
DRM_MODE_PAGE_FLIP_ASYNC flag set), the current code waits for the
possible GPU buffer being rendered through a call to
vc4_queue_seqno_cb().

On the BCM2835-37, the GPU driver is part of the vc4 driver and that
function is defined in vc4_gem.c to wait for the buffer to be rendered,
and once it's done, call a callback.

However, on the BCM2711 used on the RaspberryPi4, the GPU driver is
separate (v3d) and that function won't do anything. This was working
because we were going into a path, due to uninitialized variables, that
was always scheduling the callback.

However, we were never actually waiting for the buffer to be rendered
which was resulting in frames being displayed out of order.

The generic API to signal those kind of completion in the kernel are the
DMA fences, and fortunately the v3d drivers supports them and signal
when its job is done. That API also provides an equivalent function that
allows to have a callback being executed when the fence is signalled as
done.

Let's change our driver a bit to rely on the previous function for the
older SoCs, and on DMA fences for the BCM2711.

We should probably wait a bit for some more testing before merging this.

### DIFF
--- a/drivers/gpu/drm/vc4/vc4_crtc.c
+++ b/drivers/gpu/drm/vc4/vc4_crtc.c
@@ -769,6 +769,7 @@ struct vc4_async_flip_state {
 	struct drm_pending_vblank_event *event;
 
 	struct vc4_seqno_cb cb;
+	struct dma_fence_cb fence_cb;
 };
 
 /* Called when the V3D execution for the BO being flipped to is done, so that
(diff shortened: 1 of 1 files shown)


popcornmix:
Not seeing any change issues with kodi on a Pi4 (which doesn't use asynchronous flips).
The non pi4 case looks like a no-op from the code, so should be safe.

I'm not sure I have a use case that triggers the original bug, so can't test for that.

pelwell:
What's the position with this PR? Is the force-push a change or just a rebase?

mripard:
The first version had a regression that would get an asynchronous page flip stuck indefinitely if the buffer it switches to had a fence but was already signalled before we would wait for it. This new version fixed it in my setup but I'm still waiting for additional testing. This new version is thus both that fix and a rebase

pelwell:
Thanks - in the absence of any independent confirmation of a fix I'll accept confirmation of Kodi still working.

mripard:
The original reporter just confirmed that the original issue is now fixed and the regression is no longer there. So if Kodi is happy as well, I think this can be merged

popcornmix:
Kodi is still happy with updated version of this (although I don't believe it uses the async code path).

mripard:
Indeed, it doesn't use the async code path

pelwell:
Thanks!

PR #4573, opened 2021-09-09
Update bcm2835_smi_dev.c

madimario:
The read and write functions did not use the correct pointer offset when dealing with an odd number of bytes after a DMA transfer. I also proposed a few checks to only handle the remaining odd bytes if the DMA transfer completed successfully.


popcornmix:
@Wren6991 look okay to you?

pelwell:
I'm happy with this patch (after deleting a few unnecessary parentheses).

PR #4576, opened 2021-09-12
configs: enable eBPF support (to run Cilium)

mazzy89:
In order to run eBPF programs see the case of Cilium project, it is required to have enabled some config parameters. See here https://docs.cilium.io/en/v1.9/operations/system_requirements/#linux-kernel

In addition I'm trying to run k3os as distro and it requires some config. See here https://github.com/rancher/k3os-kernel/blob/master/patches/bpfilter-umh.patch


mazzy89:
I've tested these changes baking a kernel for my RPI 4 arm64 (64bit) and I could bootstrap an eBPF programs correctly.

timg236:
In order to update the default Raspberry Pi kernel config you need to quantify any overhead resulting from that change e.g. code size, performance, compatibility. It's also useful to explain how that feature might benefit end users.

mazzy89:
@timg236 Could you point to any actual examples? What you are saying is achievable in many forms. Is there any standard that you follow?

timg236:
Not my decision but at a minimum the change to kernel code size would be required plus information about whether this consumes any CPU resources when the feature is not enabled.
Others will probably pick this up later next week.
Rule of thumb is that adding modules is normally ok, but large increases in core kernel size are bad

mazzy89:
Sure perfectly understandable.

Let's try to give some answers to those requests

> It's also useful to explain how that feature might benefit end-users.

Enabling the following modules, the end-users can run CNIs that require to run eBPF programs see the case of Cilium.

Nowadays RPIs are used to run Kubernetes clusters at the edge. Cilium is a very demanded CNI.

>  change to kernel code size

Could you point to any resources that could help to measure this change? Not a kernel engineer here so I'm not very familiar with the procedures behind that.

mazzy89:
the image for RPI 4 (64bit) with these changes weights `22MB`

```
ubuntu@rpi-kernel:~/firmware/boot$ ls -lh kernel8.img
-rw-rw-r-- 1 ubuntu ubuntu 22M Sep 12 12:33 kernel8.img
```

against the `7.54 MB` from the latest firmware `1.20210831`

So the size seems quite different. Not sure why such a big difference. I've baked the image starting from the `HEAD` of this repo, making the change to the configs, and then run 

```
make ARCH=arm64 CROSS_COMPILE=aarch64-linux-gnu- Image modules dtbs
```

juergh:
Compressed vs. uncompressed.

mazzy89:
@juergh the one I was referring that is lighter was from the repo and it is not compressed.

pelwell:
It is compressed - it's had GZIP run on it, and the firmware detects that and automatically decompresses it.

mazzy89:
Ah gotcha. Could you advice the parameter to bake a gzipped image? I haven't found in the doc.

pelwell:
I just use `gzip -c Image > kernel8.img`.

pelwell:
Enabling those options causes the static uncompressed 32-bit kernel to grow by over 400kB the uncompressed 64-bit kernel by over 700kB. That's quite a hit for a feature that most people will never use.

You might be better off using a more full-fat distro.

mazzy89:
Could you explain why such increase can be a problem? 

Also how do you know that not many people will use such feature? Do you have data to corroborate that?

pelwell:
Because RAM is limited (1GB is common for these kernels) and we have to justify its use.

You are asking us to increase the size of our kernel builds for all of our users - the onus is on you to justify the RAM tax. As I said, other builds are available.

mazzy89:
I'm trying to justify that by saying that nowadays rpi have been used more and more at the edge to run ebpf programs. The use cases have been around for a while and out there is full of them.

I would prefer not to get into  ebpf because there is already a big literature around that already proves how much rpi and its ARM architecture has been used to run those programs effectively. 

That being said I can understand that bring those in would have an impact on the RAM usage. 

For instance at this moment I'm already running this kernel. Are you interested more on the data? Would you like to see its memory footprint?

I'm completely open to share those.

mazzy89:
Let's assume to not use the raspberrypi kernel for a second. Would be still possible to use other official kernel and configure aspects like PoE Hat fans and so on? 
Because with the official rpi firmware is super easy thanks to the way how they are controlled via the config file. The same does not happen with other kernel distro.

pelwell:
Some distros use the pure upstream kernel which will lose some features, but others track this repo and take all our patches.

mazzy89:
For me it's not a problem of distro. I'm adressing here the kernel only. I need already to use a specific distro

JamesH65:
> I'm trying to justify that by saying that nowadays rpi have been used more and more at the edge to run ebpf programs. The use cases have been around for a while and out there is full of them.
> 
> I would prefer not to get into ebpf because there is already a big literature around that already proves how much rpi and its ARM architecture has been used to run those programs effectively.

The vast majority of Pi's and Pi user's I don't think even know what ebpf is or would have any idea what this request is about. So I'd vote for not including it, the hit seems way too high just to benefit a small minority of users.

mazzy89:
You definitely know guys more the Pi users than me. So I trust you guys.  If you think then that no many will benefit of this, not much sense to include it. 

Despite what the trends nowadays are and ebpf is getting more and more attraction every day even by many big actors see for instance Microsoft acquisitions and interest, I do think unfortunately that the times are still raw. That's a pity though. However I'm sure that the time will come. 

What I can say is that thanks to the good documentation and tools I could bake my own kernel and use it. I will continue with that then. For my use cases such changes do not have any impact.

I run few RPI 4 with 8GB of RAM and the definitely do not need to count the byte in such case.

Maybe does still make sense to include the changes only to the latest PI since has more RAM?

pelwell:
I'm afraid that's a no from us - eBPF does look very cool, and it may allow smaller kernels in the future, but the case isn't yet compelling enough for our broad user base.

mazzy89:
ok @pelwell at least I tried 😛 thanks for taking a look.

pelwell:
I know you said you weren't looking for a new distro, but Ubuntu Mate might be worth investigation.

mazzy89:
@pelwell 

FYI

```
worker01 [~]$ uname -a
Linux worker01.hzk8s.k3s.home 5.10.63-v8.hzk8s+ #1 SMP PREEMPT Tue Sep 14 08:07:05 UTC 2021 aarch64 GNU/Linux
worker01 [~]$ free -m
Mem:           7760         348        5131          12        2281        7320
Swap:             0
```

with this kernel baked and a running k3s cluster using eBPF programs running on top of it consuming only **348MB**. 🙂

Issue #4577, opened 2021-09-13
vc4_hdmi runtime pm is confused, causes kernel warnings at powersave


pelwell:
The vc4_hdmi driver seems to be written in such a way that the "hsm" clock is disabled by default and enabled as needed - which is good from a power perspective. This is at odds with the suspend/resume methods -`vc4_hdmi_runtime_suspend` and `vc4_hdmi_runtime_resume` - that unconditionally disable at suspend and re-enable at resume. This causes warnings about disabling and unpreparing a clock that is already disabled and unprepared when the screensaver blanks the screen, and the enabled and prepared counts do not drop below 0:
```
[  661.773650] ------------[ cut here ]------------
[  661.773660] WARNING: CPU: 3 PID: 1200 at drivers/clk/clk.c:960 clk_core_disable+0x21c/0x288
[  661.773701] fw-clk-m2mc already disabled
[  661.773711] Modules linked in: rfcomm cmac fuse bnep hci_uart btbcm bluetooth ecdh_generic ecc 8021q garp stp llc snd_soc_hdmi_codec brcmfmac brcmutil v3d gpu_sched sha256_generic cfg80211 rfkill raspberrypi_hwmon bcm2835_codec(C) bcm2835_v4l2(C) i2c_brcmstb bcm2835_isp(C) v4l2_mem2mem bcm2835_mmal_vchiq(C) videobuf2_vmalloc videobuf2_dma_contig videobuf2_memops videobuf2_v4l2 videobuf2_common vc4 snd_bcm2835(C) cec videodev vc_sm_cma(C) mc drm_kms_helper drm drm_panel_orientation_quirks snd_soc_core snd_compress snd_pcm_dmaengine snd_pcm snd_timer snd rpivid_mem syscopyarea sysfillrect sysimgblt fb_sys_fops backlight nvmem_rmem uio_pdrv_genirq uio i2c_dev ip_tables x_tables ipv6
[  661.774224] CPU: 3 PID: 1200 Comm: kworker/3:1 Tainted: G         C        5.14.2-v7l+ #448
[  661.774243] Hardware name: BCM2711
[  661.774256] Workqueue: pm pm_runtime_work
[  661.774285] Backtrace: 
[  661.774302] [<c0bd0c7c>] (dump_backtrace) from [<c0bd1008>] (show_stack+0x20/0x24)
[  661.774335]  r7:000003c0 r6:00000080 r5:c0e39334 r4:60000093
[...]

PR #4579, opened 2021-09-14
HDMI Driver Cleanups

mripard:
Hi,

Here are few clean-ups to the HDMI driver that were found out as part of the discussion of #4577


pelwell:
Thanks!

Issue #4534, opened 2021-08-18
additional patch for your 5.13.y branch


laroche:
I am compiling your 5.13.y kernel, but get a compile error:

/home/flaroche/data/kernel/linux-5/drivers/gpu/drm/amd/amdgpu/../display/amdgpu_dm/amdgpu_dm.c: In function ‘amdgpu_dm_atomic_commit_tail’:
/home/flaroche/data/kernel/linux-5/drivers/gpu/drm/amd/amdgpu/../display/amdgpu_dm/amdgpu_dm.c:8854:4: error: implicit declaration of function ‘is_hdr_metadata_different’ [-Werror=implicit-function-declaration]
 8854 |    is_hdr_metadata_different(old_con_state, new_con_state);
      |    ^~~~~~~~~~~~~~~~~~~~~~~~~
cc1: some warnings being treated as errors
make[7]: *** [/home/flaroche/data/kernel/linux-5/scripts/Makefile.build:278: drivers/gpu/drm/amd/amdgpu/../display/amdgpu_dm/amdgpu_dm.o] Fehler 1
make[6]: *** [/home/flaroche/data/kernel/linux-5/scripts/Makefile.build:521: drivers/gpu/drm/amd/amdgpu] Fehler 2

I haven't looked into this in more detail, but a similar patch is part of your raspberry pi patches and
thus the following change looks suitable:

<pre>
--- a/drivers/gpu/drm/amd/display/amdgpu_dm/amdgpu_dm.c
+++ b/drivers/gpu/drm/amd/display/amdgpu_dm/amdgpu_dm.c
@@ -8851,7 +8851,7 @@
                              dm_old_crtc_state->abm_level;
 
                hdr_changed =
-                       is_hdr_metadata_different(old_con_state, new_con_state);
+                       !drm_connector_atomic_hdr_metadata_equal(old_con_state, new_con_state);
 
                if (!scaling_changed && !abm_changed && !hdr_changed)
                        continue;
</pre>

laroche:
This should complete the following patch:

<pre>
From aae90ba49595d413e5b1b361f48da1bcb817ba08 Mon Sep 17 00:00:00 2001
From: Maxime Ripard <maxime@cerno.tech>
Date: Fri, 19 Mar 2021 13:05:53 +0100
Subject: [PATCH 406/564] drm/connector: Add helper to compare HDR metadata
</pre>

pelwell:
Any comments, @mripard?

[...]

PR #4584, opened 2021-09-16
drm/amdgpu: Add missing conversion of is_hdr_metadata_different for amdgpu

geerlingguy:
(For 5.10.y branch, as that's the one I'm currently testing.)

See: #4534

I just did this so I could finish cross-compiling for 5.10.y, let me know if you'd like something changed or want me to submit differently against all the pre-5.14 branches. (I haven't tested on 5.11/5.12/5.13).


popcornmix:
kernels 13, 14 and 15 should be fine.
11 and 12 are EOL upstream, so there's not a lot of point maintaining them. 
This PR is fine. Newer branches reverted and cherry-picked the newer upstream commit (so we can drop old patch+revert in a rebase), but that's not really needed here.

PR #4590, opened 2021-09-19
overlays: add support for the MLX90640 thermal camera

lyakh:
This allows using the video-i2c camera driver with MLX90640 thermal infrared sensors, connected to Raspberry Pi.

### DIFF
--- a/arch/arm/boot/dts/overlays/README
+++ b/arch/arm/boot/dts/overlays/README
@@ -2061,6 +2061,12 @@ Params: krnbt                   Set to "on" to enable autoprobing of Bluetooth
                                 driver without need of hciattach/btattach
 
 
+Name:   mlx90640
+Info:   Overlay for i2c connected mlx90640 thermal camera
+Load:   dtoverlay=mlx90640
+Params: <None>
+
+
 Name:   mmc
 Info:   Selects the bcm2835-mmc SD/MMC driver, optionally with overclock
 Load:   dtoverlay=mmc,<param>=<val>
--- a/arch/arm/boot/dts/overlays/Makefile
+++ b/arch/arm/boot/dts/overlays/Makefile
@@ -121,6 +121,7 @@ dtbo-$(CONFIG_ARCH_BCM2835) += \
 	midi-uart1.dtbo \
 	minipitft13.dtbo \
 	miniuart-bt.dtbo \
+	mlx90640.dtbo \
 	mmc.dtbo \
 	mpu6050.dtbo \
 	mz61581.dtbo \
--- a/arch/arm/boot/dts/overlays/mlx90640-overlay.dts
+++ b/arch/arm/boot/dts/overlays/mlx90640-overlay.dts
@@ -1,0 +1,22 @@
+/dts-v1/;
+/plugin/;
+
+/ {
+	compatible = "brcm,bcm2835";
+
+	fragment@0 {
+		target = <&i2c_arm>;
+		__overlay__ {
+			#address-cells = <1>;
+			#size-cells = <0>;
+			status = "okay";
+			clock-frequency = <400000>;
+
+			mlx90640: mlx90640@33 {
+				compatible = "melexis,mlx90640";
+				reg = <0x33>;
+				status = "okay";
+			};
+		};
+	};
+};


pelwell:
The overlay looks fine, but it needs an entry in the README. Does it also require an extra config setting in order for the driver module to be built?

lyakh:
> The overlay looks fine, but it needs an entry in the README. Does it also require an extra config setting in order for the driver module to be built?

@pelwell yes, it needs `CONFIG_VIDEO_V4L2_I2C`, where should that be specified?

pelwell:
Add it to our defconfig files - there are 5 in total. Put the config changes in a separate commit - https://github.com/raspberrypi/linux/pull/4355 is an example of how it should look (and shows you which defconfig files to change).

arch/arm/boot/dts/overlays/i2c-mlx90640-overlay.dts
```diff
@@ -0,0 +1,22 @@
...
+/ {
+	compatible = "brcm,bcm2835";
+
+	fragment@0 {
+		target = <&i2c_arm>;
+		__overlay__ {
+			#address-cells = <1>;
+			#size-cells = <0>;
+			status = "okay";
+                        clock-frequency = <400000>;
```
pelwell: This line has ended up using spaces for indentation rather than TABs. Consistency is preferred.

arch/arm/boot/dts/overlays/README
```diff
@@ -2061,6 +2061,12 @@ Params: krnbt                   Set to "on" to enable autoprobing of Bluetooth
                                 driver without need of hciattach/btattach
 
 
+Name:   i2c-mlx90640
```
pelwell: README entries must be in alphabetical order, so it should come before i2c-mux (like the Makefile entry).
However, I'm not sure that the `i2c-` adds anything here (there is no SPI option), so would you consider just calling it `mlx90640`?

pelwell (commented):
When I apply your config patch and ran `savedefconfig` again the new setting (`CONFIG_VIDEO_V4L2_I2C=m`) disappears. This is for two reasons - it's a boolean, so the only valid choices being `y` and `n`, and it has a default value of `y`. `make ... savedefconfig` only stores differences from the defaults, so this option (and the entire second commit) wasn't actually necessary.

You should:
1. Drop the second commit.
2. Fix the overlay indentation.
3. Consider renaming the overlay to just `mlx90640`.
4. Either move the just README entry (if you don't rename) or update both the README and Makefile (if you do).

lyakh:
@pelwell thanks for the comments! Agree, the "i2c-" prefix doesn't add much information and most simple one-device overlays don't use such prefixes. Updated.

pelwell:
Looks good to me!

mogli91:
Hi @lyakh ,
I am using a Raspberry Pi and I would like to use the mlx90640 as a camera listed under /dev/video*. Currently, the camera is not listed but it seems that at least one person has managed to make this work; see [this post](https://answers.opencv.org/question/235934/mlx90640-no-video-stream/). How can I verify whether the driver and the overlay are working correctly?

Here is what I have tried:
As specified [here](https://github.com/raspberrypi/firmware/blob/master/boot/overlays/README), I am loading the overlay by adding `dtoverlay=mlx90640` to my `/boot/config.txt`.
Using `dtdebug=on`, rebooting and then running `sudo vcdbg log msg`, I see 
```
005384.807: brfs: File read: /mfs/sd/config.txt
005385.209: dtparam: i2c_arm=on
005393.711: dtparam: i2c_arm_baudrate=400000
005402.127: brfs: File read: 2263 bytes
005409.967: brfs: File read: /mfs/sd/overlays/mlx90640.dtbo
005415.065: Loaded overlay 'mlx90640'
```
This suggests that the overlay has been loaded and the i2c parameters were set as desired. I can detect the mlx90640 camera by `sudo i2cdetect -y 1` and I can also read data from it using [this adafruit example](https://github.com/adafruit/Adafruit_CircuitPython_MLX90640/blob/main/examples/mlx90640_simpletest.py).

It looks to me like the sensor is working and that it is connected correctly. Also the overlay seems to have been loaded without any errors. What else needs to be in place in order for the mlx90640 to be listed under /dev/video*?

For context: I am using the modules version 5.10.63-v7l+ (just flashed my raspberry 4 2gb with the image today)

lyakh:
Hi,

You need the video_i2c driver - is it loaded? Check your kernel modules
(lsmod | grep video). If this quick hint doesn't help, I can boot my Pi
later today for more details.

Thanks
Guennadi

mogli91:
Hi Guennadi,
Thanks for the fast feedback. Indeed `lsmod | grep video` does not show video_i2c for me. The output is
```
$ lsmod | grep video
videobuf2_dma_contig    20480  2 bcm2835_isp,bcm2835_codec
videobuf2_vmalloc      16384  1 bcm2835_v4l2
videobuf2_memops       16384  2 videobuf2_dma_contig,videobuf2_vmalloc
videobuf2_v4l2         32768  4 bcm2835_isp,bcm2835_codec,bcm2835_v4l2,v4l2_mem2mem
videobuf2_common       61440  5 bcm2835_isp,bcm2835_codec,bcm2835_v4l2,v4l2_mem2mem,videobuf2_v4l2
videodev              249856  6 bcm2835_isp,bcm2835_codec,videobuf2_common,bcm2835_v4l2,v4l2_mem2mem,videobuf2_v4l2
mc                     45056  6 bcm2835_isp,bcm2835_codec,videobuf2_common,videodev,v4l2_mem2mem,videobuf2_v4l2
```

Checking the active kernel config, I noticed that there are 2 potential flags for this driver. Can you advise which one is relevant?
```
$ sudo modprobe configs
$ zgrep -e 'CONFIG_VIDEO.\+I2C' /proc/config.gz 
CONFIG_VIDEO_V4L2_I2C=y
CONFIG_VIDEO_IR_I2C=m
# CONFIG_VIDEO_I2C is not set
```

Best regards,
Jakob

6by9:
CONFIG_VIDEO_I2C is the one you're after.

There was [a request](https://github.com/raspberrypi/linux/pull/4590#issuecomment-922503713) for the defconfigs to be updated to include it, but whilst [a commit](https://github.com/lyakh/linux-raspberry_pi/commit/0008d1433d32ab90404037b454df75673a7566fe) appears to have been created, it's not been merged (no PR?).

mogli91:
That's interesting. Are you sure? I am a little confused because comment with commit 6fd719a261ed38469bdf7e61441778f72cb4f309 says otherwise
```
This allows using the video-i2c camera driver with MLX90640 thermal
infrared sensors, connected to Raspberry Pi. CONFIG_VIDEO_V4L2_I2C
has to be selected to use the camera.
```

6by9:
https://github.com/torvalds/linux/blob/master/drivers/media/i2c/video-i2c.c is the driver you're after. 
The Makefile line to compile that is https://github.com/torvalds/linux/blob/master/drivers/media/i2c/Makefile#L117
```
obj-$(CONFIG_VIDEO_I2C)		+= video-i2c.o
```

VIDEO_V4L2_I2C is part of the V4L2 core framework, and defaults to "y".
The comments above about it are therefore wrong. The fact it is a bool (not tristate), and defaults to y explains why it disappeared for pelwell when he did a ```savedefconfig```.

lyakh:
On Wed, 19 Jan 2022, 6by9 wrote:

> https://github.com/torvalds/linux/blob/master/drivers/media/i2c/video-i2c.c is the driver you're after.
> The Makefile line to compile that is https://github.com/torvalds/linux/blob/master/drivers/media/i2c/Makefile#L117
> ```
> obj-$(CONFIG_VIDEO_I2C)		+= video-i2c.o
> ```
>
> VIDEO_V4L2_I2C is part of the V4L2 core framework, and defaults to "y".
> The comments above about it are therefore wrong. The fact it is a bool
> (not tristate), and defaults to y explains why it disappeared for
> pelwell when he did a ```savedefconfig```.

Ouch, yes, indeed, my bad, it's the CONFIG_VIDEO_I2C that you need.

Thanks
Guennadi

lyakh:
BTW, I've also done a couple of gstreamer plugins to apply calibration data from the camera, convert the image to a colour scale and print a temperature range https://github.com/lyakh/gst-mlx90640

mogli91:
> https://github.com/torvalds/linux/blob/master/drivers/media/i2c/video-i2c.c is the driver you're after. The Makefile line to compile that is https://github.com/torvalds/linux/blob/master/drivers/media/i2c/Makefile#L117
> 
> ```
> obj-$(CONFIG_VIDEO_I2C)		+= video-i2c.o
> ```
> 
> VIDEO_V4L2_I2C is part of the V4L2 core framework, and defaults to "y". The comments above about it are therefore wrong. The fact it is a bool (not tristate), and defaults to y explains why it disappeared for pelwell when he did a `savedefconfig`.

Oh that's great! Thank you for pointing this out @6by9. I see that this line already exists in our Makefile https://github.com/raspberrypi/linux/blob/rpi-5.10.y/drivers/media/i2c/Makefile#L112

Do I then understand correctly that a way to get the driver activated would simply be to compile a new kernel with the only change being `CONFIG_VIDEO_I2C=y`? (sorry for asking before trying because I am currently on quite a re-learning curve)

lyakh:
On Wed, 19 Jan 2022, mogli91 wrote:

> > https://github.com/torvalds/linux/blob/master/drivers/media/i2c/video-i2c.c is the driver you're after. The Makefile line to compile that is https://github.com/torvalds/linux/blob/master/drivers/media/i2c/Makefile#L117
> >
> > ```
> > obj-$(CONFIG_VIDEO_I2C)		+= video-i2c.o
> > ```
> >
> > VIDEO_V4L2_I2C is part of the V4L2 core framework, and defaults to "y". The comments above about it are therefore wrong. The fact it is a bool (not tristate), and defaults to y explains why it disappeared for pelwell when he did a `savedefconfig`.
>
> Oh that's great! Thank you for pointing this out @6by9.
> Do I understand correctly that a way to get the driver activated would
> simply be to compile a kernel with the only change being
> `CONFIG_VIDEO_I2C=y`? (sorry for asking before trying because I am
> currently on quite a re-learning curve)

I currently have that driver built as a module:

***@***.***:~ $ grep VIDEO_I2C .config
CONFIG_VIDEO_I2C=m

***@***.***:~ $ lsmod |grep "^video_i2c"
video_i2c              20480  0

and it's working:

***@***.***:~ $ ls -l /sys/class/video4linux/video0
lrwxrwxrwx 1 root root 0 Jan 19 20:42 /sys/class/video4linux/video0 -> ../../devices/platform/soc/fe804000.i2c/i2c-1/1-0033/video4linux/video0

Cheers
Guennadi

mogli91:
Thanks for your help,  both! Flipping the CONFIG_VIDEO_I2C flag to _m_ made it work. I am curious about the gstreamer solution as I eventually want to store short videos from the camera. I will reach out separately (linkedin).

PR #4591, opened 2021-09-20
Second attempt to fix the CPU startup failure

pelwell:
ARM: proc-v7: Retry uncached stmia if necessary
    
A failure of some CPU cores to come online has been traced to the
failure of a stm instruction while the cache is disabled. The symptom
is that the saved values read back as zeroes, a catastrophic error since
one of the values is a return address.
    
This patch forces a readback and retry until the correct value is
returned,
    
Notes:
    
At this stage in the boot process the core is running with its cache
disabled. Before enabling the cache its contents must be explicitly
invalidated, a process that requires quite a few registers that the
caller must preserve. Evidence suggests that something is writing a
block of zeroes over that space at a time when all other cores should
be idle, possibly some kind of write-combiner, and retrying is an
attempt to avoid the problem.

The previous attempted fix (forcing the accesses to only be 4-byte
aligned) appears to have only worked for a while and likely for less
obvious reasons such as a change in code alignment.
    
See: https://github.com/Hexxeh/rpi-firmware/issues/232

### DIFF
--- a/arch/arm/mm/proc-v7.S
+++ b/arch/arm/mm/proc-v7.S
@@ -288,7 +288,11 @@ __v7_ca17mp_setup:
 1:	adr	r0, __v7_setup_stack_ptr
 	ldr	r12, [r0]
 	add	r12, r12, r0			@ the local stack
+1:
 	stmia	r12, {r1-r6, lr}		@ v7_invalidate_l1 touches r0-r6
+	ldr	r0, [r12, #(6 * 4)]		@ read back the return address
+	teq	r0, lr				@ confirm it is correct
+	bne	1b				@ retrying if not
 	bl      v7_invalidate_l1
 	ldmia	r12, {r1-r6, lr}
 #ifdef CONFIG_SMP
@@ -474,7 +478,11 @@ __v7_setup:
 	adr	r0, __v7_setup_stack_ptr
 	ldr	r12, [r0]
 	add	r12, r12, r0			@ the local stack
+1:
 	stmia	r12, {r1-r6, lr}		@ v7_invalidate_l1 touches r0-r6
+	ldr	r0, [r12, #(6 * 4)]		@ read back the return address
+	teq	r0, lr				@ confirm it is correct
+	bne	1b				@ retrying if not
 	bl      v7_invalidate_l1
 	ldmia	r12, {r1-r6, lr}
 
--- a/arch/arm/mm/proc-v7.S
+++ b/arch/arm/mm/proc-v7.S
@@ -287,8 +287,6 @@ __v7_ca17mp_setup:
 	mov	r10, #0
 1:	adr	r0, __v7_setup_stack_ptr
 	ldr	r12, [r0]
-	tst	r12, #0x1f
-	addeq	r12, r12, #4
 	add	r12, r12, r0			@ the local stack
 	stmia	r12, {r1-r6, lr}		@ v7_invalidate_l1 touches r0-r6
 	bl      v7_invalidate_l1
@@ -476,8 +474,6 @@ __v7_setup:
 	adr	r0, __v7_setup_stack_ptr
 	ldr	r12, [r0]
 	add	r12, r12, r0			@ the local stack
-	tst	r12, #0x1f
-	addeq	r12, r12, #4
 	stmia	r12, {r1-r6, lr}		@ v7_invalidate_l1 touches r0-r6
 	bl      v7_invalidate_l1
 	ldmia	r12, {r1-r6, lr}
@@ -561,7 +557,7 @@ ENDPROC(__v7_setup)
 	.bss
 	.align	2
 __v7_setup_stack:
-	.space	4 * 8				@ 7 registers + 1 spare
+	.space	4 * 7				@ 7 registers
 
 	__INITDATA
 


pelwell:
@popcornmix has reminded me that the startup code was modified in 5.13 so as to no longer need to spill the registers to the stack. The changes are covered by this patchset (https://www.spinics.net/lists/arm-kernel/msg874939.html), comprising the following commits:

  c0e50736e826 ARM: 9057/1: cache-v7: add missing ISB after cache level selection
  f9e7a99fb6b8 ARM: 9058/1: cache-v7: refactor v7_invalidate_l1 to avoid clobbering r5/r6
  95731b8ee63e ARM: 9059/1: cache-v7: get rid of mini-stack

A quick reading suggested that applying these patches on top of the reversion of my first patch would provide an alternative fix for the problem, but so far I've been unable to get it to boot.

pelwell:
This patch fixes the problem for me to a reasonable level of confidence; I've seen 5+ CPU1 failures on a rev 1.1 2B running top-of-tree rpi-5.10.y (the failure mode also takes out USB) in about half an hour of reboots, and nothing in over an hour (including some cold boots) with the retry patch.

JamesH65:
The only question I had was whether the retry will always eventually succeed, or is there the vaguest possible of it infinite looping?

pelwell:
Doing nothing is the lowest risk.

JamesH65:
Well, we've seen reports in the field, so anything is better than nothing.

pelwell:
More importantly, since this is code that runs on each of the secondary cores, an infinite loop will only cause the symptoms people are seeing now (`CPUx failed to come online`) but without the horrible side-effects caused by a branch to zero...

pelwell:
> The only question I had was whether the retry will always eventually succeed

I have a partial answer to that, having seen one failure (CPU2) out of 250 reboots - the primary core gives up waiting after a short interval and marks the secondary as offline. The difference with the patched kernel, apart from the reduced incidence, is that USB was functional after the failure.

JamesH65:
Sounds like a win to me. Can we increase the primary core timeout?

pelwell:
Retesting now with a 10s timeout (was 1s).

pelwell:
Just got another CPU2 failure, despite the 10s timeout - it's dead, Jim.

lurch:
I'm afraid I know sod-all about ARM assembler, but just out of curiosity I googled `stmia` and found https://developer.arm.com/documentation/ddi0406/b/Application-Level-Architecture/Instruction-Details/Alphabetical-list-of-instructions/STM---STMIA---STMEA?lang=en which links to https://developer.arm.com/documentation/ddi0406/b/System-Level-Architecture/System-Instructions/Alphabetical-list-of-instructions/STM--user-registers-?lang=en which in turn says "Store Multiple (user registers) is unpredictable in User or System modes."
Could that explain the unpredictability you're seeing here, or is that referring to something else entirely? (I also know nothing about different CPU-modes)

KanjiMonster:
> @popcornmix has reminded me that the startup code was modified in 5.13 so as to no longer need to spill the registers to the stack. The changes are covered by this patchset (https://www.spinics.net/lists/arm-kernel/msg874939.html), comprising the following commits:
> 
> [c0e5073](https://github.com/raspberrypi/linux/commit/c0e50736e826b51ddc437e6cf0dc68f07e4ad16b) ARM: 9057/1: cache-v7: add missing ISB after cache level selection
> [f9e7a99](https://github.com/raspberrypi/linux/commit/f9e7a99fb6b86aa6a00e53b34ee6973840e005aa) ARM: 9058/1: cache-v7: refactor v7_invalidate_l1 to avoid clobbering r5/r6
> [95731b8](https://github.com/raspberrypi/linux/commit/95731b8ee63ec9419822a51cd9878fa32582fdd2) ARM: 9059/1: cache-v7: get rid of mini-stack
> 
> A quick reading suggested that applying these patches on top of the reversion of my first patch would provide an alternative fix for the problem, but so far I've been unable to get it to boot.

For us it boots (applied on top of 5.10.70), but with this the second CPU consistently fails to come up (4 times out of 4 tries):

```
smp: Bringing up secondary CPUs ...
CPU1: failed to come online
smp: Brought up 1 node, 1 CPU
SMP: Total of 1 processors activated (1000.00 BogoMIPS).
```

[95731b8](https://github.com/raspberrypi/linux/commit/95731b8ee63ec9419822a51cd9878fa32582fdd2) ARM: 9059/1: cache-v7: get rid of mini-stack is what seems to be breaking stuff, and not applying this one makes it come up successfully (though totally negating the intention). At a first glance don't see why though.

So maybe there are some additional changes needed.

popcornmix:
Yes, 5.13/5.14/5.15 include "get rid of mini-stack" and boot, and never get the CPU startup failure.
So there definitely is something between 5.10 and 5.13 that makes it work.

The good news is the CPU startup failure will go when we bump to the next LTS kernel (but that will be some way off).

KanjiMonster:
Good to know, then we'll maybe just sit this out until the next LTS kernel.

KanjiMonster:
After booting a lot of kernels, I identified the missing patch series..es:

Applying commits https://github.com/torvalds/linux/commit/4e79f0211b473f8e1eab8211a9fd50cc41a3a061 to https://github.com/torvalds/linux/commit/9443076e4330a14ae2c6114307668b98a8293b77 (11 commits) and https://github.com/torvalds/linux/commit/67e3f828bd4bf5e4eb4214dc4eb227d8f1c8a877 to https://github.com/torvalds/linux/commit/aaac3733171fca948c4fb66b78257620e3885339 (9 commits) on top of 5.10 makes the mini-stack removal actually work as expected on top of 5.10.

A few of these commits have already been backported to 5.10 stable. I did not try to create a minimal patchset that makes it work, but having both series backported definitely works at a first glance.

pelwell:
Thank you very much for doing that - it's been bothering me, since even with this second patch I have seen one or two failures. As you say, some of those patches have already been back-ported, leaving 17 additional commits and a reversion of my workaround, which are all now in the rpi-5.10y as of 462a5e2a994c6bce92f956f076a11ea546b9bc82.

popcornmix:
The upstream form of fix is now in rpi-update kernel. Would be useful if anyone experiencing the issue can test and report back. Hopefully we can then get it into stable releases soon.

KanjiMonster:
We added a check after every reboot in our automated testing, and haven't seen a failure in about ~1.5k reboots.

PR #4593, opened 2021-09-20
V4L2 stateful decode update

6by9:
A couple of fixes for the stateful decode (wants fw 1239).

Also adds support for H&V flips to the ISP wrapper.


popcornmix:
I tested this (along with firmware changes) with kodi, and didn't see any issues (no observable change).

6by9:
Added one further commit that resolves a refcounting issue should you reuse the same MMAL component instance after a timeout returning buffers. That makes Chromium happier.

6by9:
Rebased and tweaked some patches. Added WIP tag as John needs to test under Chromium.
FFmpeg and GStreamer seem happy.

6by9:
Updated again as we've resolved the Chromium issue.

If @popcornmix could check against Kodi that would be useful, otherwise I think we're good to go.

popcornmix:
Kodi appears fine with this PR (and latest firmware) in a quick test.

popcornmix:
Okay to merge?

6by9:
Thank popcornmix.
I've tested against Chromium with both Buster and Bullseye, FFmpeg (predominantly Bullseye), and GStreamer. That covers most of the main clients.

PR #4596, opened 2021-09-21
Update r8152.c

toranm:
When Hot unplug the USB rtl8152 on raspberry pi (whit open connections and traffic), Need handle the error EPROTO on the read_bulk_callback.


pelwell:
This looks like an unmodified upstream file - we haven't made any local changes - that isn't specific to Raspberry Pis, should a change should be sent upstream to the kernel maintainers.

Issue #4393, opened 2021-06-14
no ethernet hotplug on Raspberry Pi 3B with current raspios


jrfromfl:
**Describe the bug**
after booting the newest version of raspios (May 7th,2021) with a Pi3B without ethernet connection I can't connect to a local network via ETH anymore. Plugging in the ethernet cable will leave the LEDs on the plug dark, /sys/class/net/$interface/carrier=0.
Booting the same sd-card with a Pi4B works fine.

**To reproduce**
copy iso to sd card and boot with a Pi3B without plugged ethernet cable, than connect cable (switch/dhcp srv)

**Expected behaviour**
LEDs on eth plug start blinking, ip addr show lists a ip addr on ethernet interface, link is UP

**Actual behaviour**
LEDs stay dark, ip link show lists ethernet interface as DOWN, no ip addr

**System**
System Information
------------------

Raspberry Pi 3 Model B Rev 1.2
PRETTY_NAME="Raspbian GNU/Linux 10 (buster)"
NAME="Raspbian GNU/Linux"
VERSION_ID="10"
VERSION="10 (buster)"

Raspberry Pi reference 2021-05-07
Generated using pi-gen, https://github.com/RPi-Distro/pi-gen, dcfd74d7d1fa293065ac6d565711e9ff891fe2b8, stage2

Linux raspberrypi 5.10.17-v7+ #1414 SMP Fri Apr 30 13:18:35 BST 2021 armv7l GNU/Linux
Revision	: a02082
Serial		: 000000008bd985a3
Model		: Raspberry Pi 3 Model B Rev 1.2
Throttled flag  : throttled=0x0
Camera          : supported=0 detected=0

Videocore information
---------------------

Apr 30 2021 13:47:07 
Copyright (c) 2012 Broadcom
version d7f29d96450abfc77cd6cf011af1faf1e03e5e56 (clean) (release) (start)

alloc failures:     0
compactions:        0
legacy block fails: 0

Filesystem information
----------------------
Filesystem     1K-blocks    Used Available Use% Mounted on
/dev/root        1507280 1289304    123360  92% /
devtmpfs          439916       0    439916   0% /dev
tmpfs             473196       0    473196   0% /dev/shm
tmpfs             473196   12324    460872   3% /run
tmpfs               5120       4      5116   1% /run/lock
tmpfs             473196       0    473196   0% /sys/fs/cgroup
/dev/mmcblk0p1    258095   48781    209314  19% /boot
[...]

PR #4598, opened 2021-09-22
net: phy: lan87xx: Allow more time for link detect

pelwell:
With EDPWRDOWN set in idle, it must be cleared before checking for
ENERGYON going high, indicating that a link is being established.
The existing code allows 640ms for ENERGYON to go high, but on
Raspberry Pis that appears not to be enough, causing link detection
to fail.

Increase the polling timeout to 1500ms - with a polling interval of
10ms it shouldn't cause unnecessary delays.

See: https://github.com/raspberrypi/linux/issues/4393

### DIFF
--- a/drivers/net/phy/smsc.c
+++ b/drivers/net/phy/smsc.c
@@ -195,12 +195,12 @@ static int lan87xx_read_status(struct phy_device *phydev)
 		if (rc < 0)
 			return rc;
 
-		/* Wait max 640 ms to detect energy and the timeout is not
+		/* Wait max 1500 ms to detect energy and the timeout is not
 		 * an actual error.
 		 */
 		read_poll_timeout(phy_read, rc,
 				  rc & MII_LAN83C185_ENERGYON || rc < 0,
-				  10000, 640000, true, phydev,
+				  10000, 1500000, true, phydev,
 				  MII_LAN83C185_CTRL_STATUS);
 		if (rc < 0)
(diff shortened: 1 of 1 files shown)


6by9:
s/lan87xx/lan78xx

6by9:
Oh joys, they've mixed and matched 8700 and 7800. Lovely!

I was just about to note that the LAN7800 is only on the 3B+, not the 3B the linked issue references.

pelwell:
LAN87xx is the PHY.

pelwell:
Hence:
```diff
@@ -195,12 +195,12 @@ static int lan87xx_read_status(struct phy_device *phydev)
```

l1k:
@pelwell Please note that starting with v5.19, commits 72aa729bc1c4085e2a37ee979a542cba741bb575 and 5e73e0793d406132c8c121f96f53bbda21a7cdef will no longer be necessary and can be dropped from the Foundation's downstream kernel. We've moved to interrupt-driven link detection (instead of polling) in upstream with commit 2642cc6c3bbe0900ba15bab078fd15ad8baccbc5.

l1k:
Excuse me, I meant 1ce8b37241ed291af56f7a49bbdbf20c08728e88. The other one is a fixup for that commit.

pelwell:
Thanks, Lukas - we're always happy to drop downstream patches.

PR #4599, opened 2021-09-22
Dev/5.10/fix streamoff 1

jc-kynesim:
Fixes a CMA leak, also worked around by reordering streamoffs in FFmpeg


pelwell:
Looks ok to me. Can someone kick the tyres?

popcornmix:
Tested with kodi. Was able to play/stop hevc a number of times without issue.

PR #4605, opened 2021-09-25
overlays: mcp2515 overlay that works like the mcp251xfd overlay.

GabyPCgeeK:
Can configure mcp2515 on spi0/1/2 without the need for multiple overlays.

### DIFF
--- a/arch/arm/boot/dts/overlays/Makefile
+++ b/arch/arm/boot/dts/overlays/Makefile
@@ -113,6 +113,7 @@ dtbo-$(CONFIG_ARCH_BCM2835) += \
 	mbed-dac.dtbo \
 	mcp23017.dtbo \
 	mcp23s17.dtbo \
+	mcp2515.dtbo \
 	mcp2515-can0.dtbo \
 	mcp2515-can1.dtbo \
 	mcp251xfd.dtbo \
--- a/arch/arm/boot/dts/overlays/README
+++ b/arch/arm/boot/dts/overlays/README
@@ -1977,6 +1977,21 @@ Params: s08-spi<n>-<m>-present  4-bit integer, bitmap indicating MCP23S08
                                 or INTB output of MCP23S17 is connected.
 
 
+Name:   mcp2515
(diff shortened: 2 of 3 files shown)


pelwell:
Nicely done. It would be good to drop the old "mcp2515-can<x>" versions at some point - the upcoming Bullseye release will present DT errors and warnings from the firmware to users, so it may be possible to transition people over.

TinkerWorX:
Would it be possible to make this work with BCM2711?
I'm on a `Raspberry Pi 4 Model B Rev 1.1`, and `spi0-0` and `spi0-1` are working fine, but `spi1-0` reports the following:
> [    4.893280] spi spi1.0: Native CS is not supported - please configure cs-gpio in device-tree
> ...
> [    5.839085] mcp251x spi1.0: Native CS is not supported - please configure cs-gpio in device-tree
> [    6.854446] mcp251x spi1.0: MCP251x didn't enter in conf mode after reset
> [    6.854623] mcp251x spi1.0: Probe failed, err=110
> [    6.854694] mcp251x: probe of spi1.0 failed with error -110

As far as I know, the board only has native chip select for SPI 0, which would seem to be the issue. However, I know it's possible to make more SPI ports work, so could this overlay be configured to handle that?

pelwell:
Have you enabled spi1 using one of the spi1 overlays?
```
Name:   mcp2515
Info:   Configures the MCP2515 CAN controller on spi0/1/2
        For devices on spi1 or spi2, the interfaces should be enabled
        with one of the spi1-1/2/3cs and/or spi2-1/2/3cs overlays.
Load:   dtoverlay=mcp2515,<param>=<val>
Params: spi<n>-<m>              Configure device at spi<n>, cs<m>
                                (boolean, required)

        oscillator              Clock frequency for the CAN controller (Hz)

        speed                   Maximum SPI frequence (Hz)

```

TinkerWorX:
That appears to have done the trick. I was initially confused because I could enable spi0 without any additional overlays. But adding `dtoverlay=spi1-1cs` appears to have helped. Apologies, still learning how all this works.

PR #4611, opened 2021-09-30
add support for Hifiberry AMP3

j-schambacher:
Adds a board driver for the new Hifiberry AMP3. It also adds the
TAS5754m component driver featuring only I2S master mode as the
AMP3 only supports master mode.


arch/arm/boot/dts/overlays/README
```diff
@@ -1170,6 +1170,32 @@ Params: 24db_digital_gain       Allow gain to be applied via the PCM512x codec
                                 Will be overwritten by ALSA user settings.
 
 
+Name:   hifiberry-amp3
+Info:   Configures the HifiBerry AMP3 audio cards
+Load:   dtoverlay=hifiberry-amp3
```
pelwell, line 1340: Add `,<param>=<val>` here.

pelwell (commented):
There's a simple README change I'd like, but my other issues are probably less straightforward:

1. Judging by the cloned header, the TAS5754M shares the API of the PCM5122. If so, why does it need a dedicated driver? Oh no - the register spacing is different. That should be fixable with a tweak to the pcm512x driver, but are there other differences?
2. If it does need a dedicated driver, it should be submitted upstream because it is a standard component.
3. Before I did accept whatever comes of this PR I would want the patches split into three commits - driver+Makefile+Kconfig, defconfigs, and overlay+Makefile+README.

j-schambacher:
Phil, thanks for the fast feedback
to 1. Yes, at first TAS5754M and PCM512x appear to be compatible. In discussions with the TI apps team, they recommended to write a dedicated driver as there are differences between the parts. I2S slave mode works , but especially the I2S master mode does not. Probably there are more differences in the so called 'hybrid flow' - I did not dig into it, as we do not support it.
to 2. Ok, we'll look into it.
to 3. OK

PR #4613, opened 2021-10-01
WIP: Rework DSI driver to be a bridge

6by9:
(Sorry, another WIP PR).

@mripard Patchset converting DSI to being a bridge as discussed. Could you give it a once over at some point? (I know you're away at present)
I suspect I should be looking somewhere else than dsi->encoder->crtc->state->adjusted_mode for the selected mode now that I'm using the atomic hooks.
I want to have a further think over whether pre_enable is actually better than enable for doing all the work. It probably makes a difference as to the state of the HVS and PV compared to the DSI block, but I'm not clear which is "correct". That switch from enable to pre_enable should probably be a separate patch too.

### DIFF
--- a/drivers/gpu/drm/vc4/vc4_dsi.c
+++ b/drivers/gpu/drm/vc4/vc4_dsi.c
@@ -793,7 +793,8 @@ dsi_esc_timing(u32 ns)
 	return DIV_ROUND_UP(ns, ESC_TIME_NS);
 }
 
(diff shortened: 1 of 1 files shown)


6by9:
Ignore this. The concept of splitting the bridge chain when using atomic bridges is broken as the framework doesn't know about the additional bridges to create the required state, and there is no way to chain it from our duplicate_state.

PR #4617, opened 2021-10-06
dwc_otg: pay attention to qh->interval when rescheduling periodic queues

P33M:
A regression introduced in https://github.com/raspberrypi/linux/pull/3887
meant that if the newly scheduled transfer immediately returned data, and
the driver resubmitted a single URB after every transfer, then the effective
polling interval would end up being approx 1ms.

Use the larger of SCHEDULE_SLOP or the configured endpoint interval.

--

This fixes the broken mousepoll behaviour on Pi 1.


pelwell:
That works for me. FYI, I've been testing on a Pi 2 (Pi 3 should also be affected) using a USB mouse and the following command:
```
pi@raspberrypi:~ $ dd if=/dev/input/mice of=/dev/null bs=3
```
Start it without disturbing the mouse, move the mouse in circles for as close to 10 seconds as you can, then hit Ctrl-C. The number of records is the number of mouse events. A correct result is something close to 600, indicating 60Hz polling.

Issue #4440, opened 2021-07-08
libcamera: BCM2835_ISP_BASE different between 5.10.y and libcamera vs 5.12.y and 5.13.y


olivierschonken:
**Describe the bug**
Due to a mismatch in the BCM2835_ISP_BASE between libcamera and kernel 5.12.y and kernel 5.13.y, libcamera will not work with rpi unless its v4l2-controls.h is also modified to match the kernel v4l2-controls.h

kernel 5.12.y and 5.13.y v4l2-controls.h
#define V4L2_CID_USER_BCM2835_ISP_BASE		(V4L2_CID_USER_BASE + 0x10f0)

kernel 5.10.y and libcamera v4l2-controls.h
#define V4L2_CID_USER_BCM2835_ISP_BASE		(V4L2_CID_USER_BASE + 0x10e0)

**To reproduce**
Use kernel 5.12.y or 5.13.y with current libcamera HEAD
cam -c 1 -C 1

**Expected behaviour**
root@raspberrypi4-64:~# cam -c 1 -C 1
[1:54:10.449586656] [622]  INFO Camera camera_manager.cpp:296 libcamera v0.0.0+2747-fb970d9c-dirty (2021-07-08T15:00:29+00:00)
[1:54:10.479843682] [623]  WARN CameraSensor camera_sensor.cpp:197 'ov5647 10-0036': Recommended V4L2 control 0x009a0922 not supported
[1:54:10.479959532] [623]  WARN CameraSensor camera_sensor.cpp:249 'ov5647 10-0036': The sensor kernel driver needs to be fixed
[1:54:10.480060234] [623]  WARN CameraSensor camera_sensor.cpp:251 'ov5647 10-0036': See Documentation/sensor_driver_requirements.rst in the libcamera sources for more information
[1:54:10.481761242] [623]  WARN CameraSensor camera_sensor.cpp:403 'ov5647 10-0036': Failed to retrieve the camera location
Using camera /base/soc/i2c0mux/i2c@1/ov5647@36
[1:54:10.757968853] [622]  WARN V4L2 v4l2_pixelformat.cpp:181 Unsupported V4L2 pixel format XB24
[1:54:10.759028946] [622]  INFO Camera camera.cpp:912 configuring streams: (0) 800x600-NV12
[1:54:10.759326886] [623]  INFO RPI raspberrypi.cpp:122 Mode: 2592x1944 fmt BG10 Score: 2284 (best 2284)
[...]

PR #4621, opened 2021-10-07
Rpi 5.10.y imx477 fstrobe

cnelson-01:
Adds "call for strobe" settings to the v4l2-ctrls. The idea being that the camera can output a strobe signal which then can be used by an illuminator eg V4L2_CID_FLASH...

Implements the camera strobe control signals for the imx477

 Couple things to check:
- What are the limits on width and delay?
- is setting FLASH_TRIG_RS to 0 the correct way to disable the fstrobe?  I did some basic testing and it seems to work.


6by9:
Thank you for creating a PR for this - it is appreciated.

Creating extra V4L2 controls is something that isn't sustainable in a downstream kernel as we'll then get conflicts when mainline add their next one as we have in #4440.
The only option is to get the controls added to the mainline Linux kernel via the linux-media tree, and we'll then backport. In the meantime module parameters are the only solution.

cnelson-01:
Makes sense. I'll modify this pull to remove the v4l2 changes and see if I can get the camera strobe features added to the linux media tree.

jakekake:
What will be the easiest way to get fstrobe working with libcamera on a 5.15.y kernel version?

PR #4622, opened 2021-10-08
Add support for reTerminal, a Raspberry Pi 4 CM carrier board

AIWintermuteAI:
This PR adds a seeed-reterminal-core overlay and enables building necessary drivers as modules to support reTerminal, a Raspberry Pi 4 CM carrier board.

Seeed Studio reTerminal is a Human-Machine Interface (HMI) carrier board
for Raspberry Pi 4 CM with 5-inch LCD screen, accelerometer,
light sensor, GPIO buttons and user LEDs. reTerminal can easily and
efficiently work with IoT and cloud systems to unlock endless scenarios
at the edge.

- LTR light sensor driver was modified to make it compatible with LTR303. 
- NXP PCA953xx I2C Mux/switch support was added
- LIS331DLH accelerometer driver is unchanged, enabled building as a module
- added display/touchscreen driver

Signed-off-by: Maslov Dmitry  <maslovdmitry@seeed.cc>


arch/arm/boot/dts/overlays/README
```diff
@@ -2837,6 +2837,21 @@ Info:   Overlay for Seeed Studio CAN BUS FD HAT with two CAN FD
 Load:   dtoverlay=seeed-can-fd-hat-v2
 Params: <None>
```
pelwell, line 2839: Two blank lines between entries.

arch/arm/boot/dts/overlays/README
```diff
@@ -2837,6 +2837,21 @@ Info:   Overlay for Seeed Studio CAN BUS FD HAT with two CAN FD
...
 
+Name:   seeed-reterminal-core
+Info:   Overlay that enables 5-inch LCD display, touch screen, signal LEDs, user buttons,
+        accelerometer lis331dlh and GPIO expander on reTerminal,
+        a Seeed studio carrier board for the Raspberry Pi Compute Module 4.
+        https://wiki.seeedstudio.com/reTerminal/
+Load:   dtoverlay=seeed-reterminal-core,<param>=<val>
+Params:	int			interrupt pin and it's configuration of mcp23008(default 6,2)
```
pelwell: Use spaces for indent - no TABs.

arch/arm/boot/dts/overlays/README
```diff
@@ -2837,6 +2837,21 @@ Info:   Overlay for Seeed Studio CAN BUS FD HAT with two CAN FD
...
+        addr			i2c slave address of mcp23008(default 0x20)
+        noints			Disable the interrupt GPIO line
+        key0			event code for key0(default 0x30)
+        key1 			event code for key1(default 0x31)
+        key2 			event code for key2(default 0x32)
+        key3 			event code for key3(default 0x33)
+        mcu_auto_reset	        auto reset the MCU when the touch panel is not connected to PCB board(default 0)
+        tp_rotate		rotate the touch panel's point for 90 degrees counterclockwise(default 0)
```
pelwell, line 2860: One more blank line here.

arch/arm/boot/dts/overlays/seeed-reterminal-core-overlay.dts
```diff
@@ -0,0 +1,244 @@
...
+ * Zhangqun Ming <north_sea@qq.com>
+ *
+ * MIT License
+ *
+ */
+/dts-v1/;
+/plugin/;
+
+/ {
+    compatible = "brcm,bcm2835", "brcm,bcm2708", "brcm,bcm2709", "brcm,bcm2711";
```
pelwell: Just "brcm,bcm2711" please - this will only apply to a Pi 4 DTB because of its use of i2c3.

drivers/iio/light/Kconfig
```diff
@@ -318,7 +318,7 @@ config SENSORS_LM3533
 	  changes. The ALS-control output values can be set per zone for the
 	  three current output channels.
 
-config LTR501
+config LTR
```
pelwell: The ltr501 driver and Kconfig setting are unmodified from the upstream versions. If you wish to make a change like this (and I don't think you need to) you'll have to argue your case to the upstream kernel maintainers.

drivers/iio/light/ltr.c
```diff
@@ -1,8 +1,9 @@
 // SPDX-License-Identifier: GPL-2.0-only
 /*
- * ltr501.c - Support for Lite-On LTR501 ambient light and proximity sensor
+ * ltr.c - Support for Lite-On LTR501, LTR509, LTR301, LTR303 ambient light and proximity sensors (only LTR5xx)
```
pelwell: As mentioned on the previous commit, this is a standard Linux file that hasn't been altered in our branches - this change must be sent to the main kernel maintainers.

drivers/i2c/muxes/Kconfig
```diff
@@ -54,6 +54,15 @@ config I2C_MUX_LTC4306
 
 	  This driver can also be built as a module.  If so, the module
 	  will be called i2c-mux-ltc4306.
+
```
pelwell: There is nothing Pi-specific about this driver - send it upstream.

arch/arm/boot/dts/overlays/README
```diff
@@ -2837,6 +2837,21 @@ Info:   Overlay for Seeed Studio CAN BUS FD HAT with two CAN FD
...
+        https://wiki.seeedstudio.com/reTerminal/
+Load:   dtoverlay=seeed-reterminal-core,<param>=<val>
+Params:	int			interrupt pin and it's configuration of mcp23008(default 6,2)
+        addr			i2c slave address of mcp23008(default 0x20)
+        noints			Disable the interrupt GPIO line
+        key0			event code for key0(default 0x30)
+        key1 			event code for key1(default 0x31)
+        key2 			event code for key2(default 0x32)
+        key3 			event code for key3(default 0x33)
+        mcu_auto_reset	        auto reset the MCU when the touch panel is not connected to PCB board(default 0)
```
pelwell: This and several other lines are too long - there is an 80-character limit.

arch/arm/boot/dts/overlays/seeed-reterminal-core-overlay.dts
```diff
@@ -0,0 +1,244 @@
+/*
```
pelwell: Please make this a UNIX-style text file - no CRs, just LFs.

pelwell (changes requested):
There are several problems with this PR:
1. It contains several drivers (and changes to existing drivers) that aren't Pi-specific and should be upstreamed - we won't take on the support burden.
2. The overlay and README have a few stylistic issues - see the inline comments.
3. The defconfig patches can be squashed into a single one - we will do that eventually, so they might as well start out that way.

lategoodbye:
@AIWintermuteAI FWIW [CM4](https://git.kernel.org/pub/scm/linux/kernel/git/next/linux-next.git/commit/?h=next-20211008&id=b65b46c047a9140f3fe44a4a3e00ce5c2caae8cc) and [CM4 IO](https://git.kernel.org/pub/scm/linux/kernel/git/next/linux-next.git/commit/?h=next-20211008&id=c5d283ef9e722b9ee134c85c5291b1676bf68e9f) support will land in Linux 5.16. The DTS approach differs a little bit from this tree, because of missing full overlay support in upstream.

AIWintermuteAI:
@pelwell Thanks for detailed review!
Alright, so from what I see now, it seems the best way is to:
- do the rebase on commit branch and drop commits related to ltr501 and NXP PCA953xx drivers, since these need to be sent to Linux kernel
- fix the empty lines and some issues in overlay
- leave touch panel as-is
- squash defconfig patches to one (namely for now that only would be the touch panel and acceleromter).
Is that correct?

AIWintermuteAI:
@lategoodbye Thanks for the heads-up! Do you mean that the overlay in my PR needs to be modified because of that?

lategoodbye:
@AIWintermuteAI There is no need to modify the overlay in this PR.

pelwell:
I don't see why the touchpanel driver should be treated any differently - it's also generic code that could be useful for other platforms, so it should go upstream.

drivers/i2c/muxes/i2c-mux-pca953xx.c
```diff
@@ -0,0 +1,1413 @@
+// SPDX-License-Identifier: GPL-2.0-only
```
6by9: You've added this to drivers/i2c/muxes, but it's an I2C GPIO expander. It also appears to be a copy of /driver/gpio/gpio-pca953x, including duplicating a large number of the compatible strings.

drivers/gpu/drm/panel/panel-ilitek-ili9881d.c
```diff
@@ -0,0 +1,309 @@
...
+
+	.hdisplay	= 720,
+	.hsync_start= 720 + 10,
+	.hsync_end	= 720 + 10 + 20,
+	.htotal		= 720 + 10 + 20 + 30,
+
+	.vdisplay	= 1280,
+	.vsync_start= 1280 + 10,
+	.vsync_end	= 1280 + 10 + 20,
+	.vtotal		= 1280 + 10 + 20 + 30,
```
6by9: This timing looks identical to lhr050h41_default_mode in drivers/gpu/drm/panel/panel-ilitek-9881c.c
Are they actually any different?

drivers/gpu/drm/panel/panel-ilitek-ili9881d.c
```diff
@@ -0,0 +1,309 @@
...
+/*
+ * ILI9881D panel driver
+ * 
+ * Copyright (c) 2020 Seeed Studio
+ * Zhangqun Ming<north_sea@qq.com>
+ */
+#include "panel-reTerminal.h"
+
+
+static struct mipi_dsi_device *ili9881d_dsi = NULL;
```
6by9: One static variable, so how would I add 2 of these to a device?

drivers/gpu/drm/panel/panel-ilitek-ili9881d.c
```diff
@@ -0,0 +1,309 @@
...
+	drm_mode_probed_add(connector, mode);
+
+	return 1;
+}
+
+#define ILI9881_PAGE(_page)					DSI_DCS_WRITE(dsi,0xff,0x98,0x81,_page)
+#define IILI9881_COMMAND(_cmd,_data...)		DSI_DCS_WRITE(dsi,_cmd,_data)
+static int ili9881d_prepare(struct drm_panel *panel)
+{
+	struct mipi_dsi_device *dsi = ili9881d_dsi;
```
6by9:
```
struct i2c_mipi_dsi *md = panel_to_md(panel);
struct mipi_dsi_device *dsi = md->dsi;
```
Gets rid of ```ili9881d_dsi```

6by9 (commented):
This feels like a bit of a mismash with multiple functions crammed into one driver.

ILI9881D would appear to be a standalone panel, although it does need GPIO control for reset control and/or regulators. 
https://github.com/raspberrypi/linux/blob/rpi-5.10.y/arch/arm/boot/dts/overlays/vc4-kms-dsi-7inch-overlay.dts has been updated to use the newer drivers which split this functionality for the Pi panel into independent modules and stayed within the normal frameworks.

We also seem to have some new custom device tree properties which have no documentation.

6by9:
Looking at it, your commit text says of the touchscreen
```
however it is connected to the carrier board through a STM32G030 MCU,
which makes a separate driver necessary.
```
If it is the vanilla Goodix touch-controller, it'd seem highly plausible to write a very simple wrapper that looked like an I2C controller to take requests from the existing driver and encapsulate them for the MCU. It largely looks like you're splitting the write and read into two transactions with a delay, so that should be pretty trivial to implement.
Touchscreen rotation isn't normally handled in the driver, so your ```x_y_rotate``` function looks dubious.

arch/arm/configs/bcm2711_defconfig
```diff
@@ -514,6 +514,7 @@ CONFIG_BCMGENET=y
 CONFIG_ENC28J60=m
 CONFIG_QCA7000_SPI=m
 CONFIG_QCA7000_UART=m
+CONFIG_R8169=m
```
pelwell: These changes were merged independently yesterday - see #4625.

AIWintermuteAI: Yes, that snuck with fetching latest changes from upstream. I'll drop this commit from PR when I push the changes for README and overlay.

AIWintermuteAI:
@pelwell I dropped all the commits relating to the modules that you said need to go to upstream. What's left now is overlay and enabling building LIS331DLH driver as a module. 
I'll review the code for the rest of devices and prepare the PRs to upstream. That might take a while of course. 
@6by9 Thank you for the comments. Yes, I agree, it does seem the touch-screen panel driver needs some work. Can you point to some examples, where similar driver was implemented in the way that would conform to Raspberry Pi foundation/Linux kernel standards?

arch/arm/boot/dts/overlays/README
```diff
@@ -2838,6 +2838,26 @@ Load:   dtoverlay=seeed-can-fd-hat-v2
 Params: <None>
 
 
+Name:   seeed-reterminal-core
+Info:   Overlay that enables 5-inch LCD display, touch screen, signal LEDs,
+        user buttons, accelerometer lis331dlh and GPIO expander on reTerminal,
+        a Seeed studio carrier board for the Raspberry Pi Compute Module 4.
+        https://wiki.seeedstudio.com/reTerminal/
+Load:   dtoverlay=seeed-reterminal-core,<param>=<val>
+Params:	int                     interrupt pin and it's configuration
```
pelwell: There's a TAB on this line instead of a space.

pelwell (commented):
If merging this now is useful to you then fix the hard TAB, squash down to two commits - overlay and config - and update the PR one more time.

6by9:
> Can you point to some examples, where similar driver was implemented in the way that would conform to Raspberry Pi foundation/Linux kernel standards?

https://github.com/raspberrypi/linux/blob/rpi-5.10.y/arch/arm/boot/dts/overlays/vc4-kms-dsi-7inch-overlay.dts is the closest as it uses an Atmel as backlight, regulator, and GPIO controller. Remove the TC358762 bridge and panel_disp1, moving to a single ILI9881 panel under &dsi1 instead.

ILI9881C already has a driver at https://github.com/raspberrypi/linux/blob/rpi-5.10.y/drivers/gpu/drm/panel/panel-ilitek-ili9881c.c
I don't know how close to ILI9881D that is, but if your device can be supported by just adding an extra compatible string and init sequence, then so much the better.

The I2C wrapper for the MCU should be relatively simple in taking a parent I2C device and creating a new one. The transfer function then takes in the request and issues multiple transfers to the parent, with delays between each.

6by9:
DRM requires all components to be available before it binds them all. I suspect with dsi1 enabled but no panel driver in the build, adding this overlay will lead to all displays going blank as DRM can never bind. Is that actually useful?

arch/arm/boot/dts/overlays/README
AIWintermuteAI: Those pesky tabs. Fixed it now.

pelwell (approved):

pelwell:
Apart from @6by9's question ("Is [all display's blank] actually useful?") I'm ready to merge. I'm also happy to park this until progress has been made on getting the necessary drivers into the kernel.

6by9:
By coincidence a patch has just been submitted to dri-devel that claims to add an ILI9881D panel to the ILI9881C driver. That implies that they are close enough to reuse the same driver (or it's a typo!)
https://lists.freedesktop.org/archives/dri-devel/2021-October/327570.html

AIWintermuteAI:
@pelwell I discussed it with other members on the team and we decided it's better to park PR for a time being, while we work on sending other drivers upstream.
@6by9 Thanks for the information, I'll have a look at the patch!

Issue #4620, opened 2021-10-07
/dev/vcio is not created when using a custom dtb on 64 bit RPiOS


0n0w1c:
**Describe the bug**
When using a custom .dtb file, /dev/vcio is not created on boot.

**To reproduce**
Install RPiOS 64 bit beta
Update it, including the 5.10.63-v8+ kernel
Create a device tree from the running configuration (note: /dev/vcio does exist) with:
`dtc -I fs -O dtb -o base.dtb /proc/device-tree`
modify config.txt with device_tree=base.dtb
reboot
ls -l /dev/vcio

**Expected behaviour**
/dev/vcio is created

**Actual behaviour**
/dev/vcio is not created

**System**
Raspberry Pi 4 Model B Rev 1.2

Raspberry Pi reference 2021-05-07
Generated using pi-gen, https://github.com/RPi-Distro/pi-gen, 071d3956353f2be964d2d337d3b123f1872b7f53, stage 2

Sep 28, 2021 11:27:35
Copyright (c) 2012 Broadcom
version 778b6a4f3c7d8d48bb63c02c47bcfbac79417bea (clean) (release) (start)

Linux raspberrypi 5.10.63-v8+ #1457 SMP PREEMPT Tue Sep 28 11:27:02 BST 2021 aarch64 GNU/Linux

**Additional context**
This same setup works as expected with the 5.10.63-v7 kernel and 32 bit RPiOS.

Note: The system info was typed not copied, so possible typos exist.

0n0w1c:
I believe the issue is here:
/linux/drivers/char/broadcom/vcio.c
```
np = of_find_compatible_node(NULL, NULL,
				     "raspberrypi,bcm2835-firmware");
if (!of_device_is_available(np))
		return -ENODEV;

vcio.fw = rpi_firmware_get(np);
if (!vcio.fw)
		return -ENODEV;
```

pelwell:
Please upload the .dtb somewhere - Google Drive, etc. - so that I can download and analyse it.

0n0w1c:
If you copy the base32.dtb to the 64-bit RPiOS, it will create /dev/vcio.

[base32.dtb](https://drive.google.com/file/d/1wE_OgPZF0YMan0R5CYClTJ-oYfo_Z5zF/view?usp=sharing)
[base64.dtb](https://drive.google.com/file/d/1cQwXD4YRlxJS9ZykzyXupmY465ZBmS5T/view?usp=sharing)

[...]

PR #4624, opened 2021-10-12
vcio rewrite

pelwell:
#4620 identified a problem with kernel startup ordering that can break the vcio device. This PR avoids the problem by making vcio a platform device, instantiated from Device Tree.

Note that using the new driver will require a corresponding udev rule change (backwards-compatible):
```
KERNEL=="vcio", GROUP="video", MODE="0660"
```
(was `SUBSYSTEM=="bcm2708_vcio", ...`)

### DIFF
--- a/arch/arm/boot/dts/bcm270x-rpi.dtsi
+++ b/arch/arm/boot/dts/bcm270x-rpi.dtsi
@@ -152,3 +152,9 @@
 		status = "disabled";
 	};
 };
+
+&firmware {
+	vcio: vcio {
+		compatible = "raspberrypi,vcio";
+	};
+};
--- a/drivers/char/broadcom/vcio.c
+++ b/drivers/char/broadcom/vcio.c
@@ -1,14 +1,13 @@
 /*
  *  Copyright (C) 2010 Broadcom
  *  Copyright (C) 2015 Noralf Trønnes
+ *  Copyright (C) 2021 Raspberry Pi (Trading) Ltd.
  *
  * This program is free software; you can redistribute it and/or modify
  * it under the terms of the GNU General Public License version 2 as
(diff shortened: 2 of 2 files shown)


pelwell:
@vianpl @waveform80 @Dark-Sky @graysky2 I'd like to merge this PR soon but also want to warn you about the possible need to update your udev rules as shown above. If you are make that change, I would recommend also changing the rule for vchiq in the same way:
```
KERNEL=="vcio", GROUP="video", MODE="0660"
KERNEL=="vchiq", GROUP="video", MODE="0660"
```

vianpl:
@mbgg :arrow_up: 
@pelwell Thanks for letting us know! :slightly_smiling_face:

graysky2:
@pelwell - Thanks for the heads-up

pelwell:
I should also have mentioned that it is safe to make the udev changes now - they are backwards compatible.

graysky2:
@pelwell - For Arch ARM, I believe this change only affects a single line.  Can you view the above mentioned commit in my udev rules for kodi for sanity?

pelwell:
That looks fine (although it doesn't need the extra space).

The vchiq change is only in linux-next at the moment, so you have plenty of time to roll out your patch.

graysky2:
Thanks, if it is backwards compatible I will just work it in sooner than later or else I might forgot :smile:

Dark-Sky:
@pelwell Thank you very much for the heads up!

pelwell:
Thank you all - this PR has now been merged. Send any complaints in my direction.

pelwell:
P.S. RPiOS users can install the updated udev rule with:
```
$ sudo apt update
$ sudo apt install libraspberrypi0
```

PR #4628, opened 2021-10-14
Fixed : fixed typo

RuffaloLavoisier:
fixed typo in stable_api_nonsense.


pelwell:
That file is unmodified from the upstream version - you should send a patch to `Jonathan Corbet <corbet@lwn.net>` and `linux-doc@vger.kernel.org`.

PR #4629, opened 2021-10-14
Wrap PWM output as GPIO

6by9:
This needs a dt binding, but I wondered what people thought.

Scenario is for things like DSI screens where we have I2C up the FFC, and want a combination of GPIO outputs for resets and regulators, and PWM for the backlight.
You can get chips such as [NXP PCA9685](https://www.nxp.com/products/power-management/lighting-driver-and-controller-ics/ic-led-controllers/16-channel-12-bit-pwm-fm-plus-ic-bus-led-controller:PCA9685) which gives 16 PWM outputs, or you can get I2C expanders that give GPIOs. To avoid having to put both on the controller board, wrap some PWM outputs of an PCA9685 as GPIO outputs, and drive them as such.

https://forums.raspberrypi.com/viewtopic.php?f=98&t=321245&p=1925007

One thing that would be nice to do would be to read an array of PWM outputs and create a single gpiochip supporting all of them. Currently you need a separate instance for each output. I couldn't see a way of configuring multiple PWM devices without also specifying names for each one, but I may have just missed it.

### DIFF
--- a/drivers/gpio/gpio-pwm.c
+++ b/drivers/gpio/gpio-pwm.c
@@ -1,0 +1,144 @@
+// SPDX-License-Identifier: GPL-2.0+
+/*
+ * GPIO driver wrapping PWM API
+ *
+ * PWM 0% and PWM 100% are equivalent to digital GPIO
+ * outputs, and there are times where it is useful to use
+ * PWM outputs as straight GPIOs (eg outputs of NXP PCA9685
+ * I2C PWM chip). This driver wraps the PWM API as a GPIO
+ * controller.
+ *
+ * Copyright (C) 2021 Raspberry Pi (Trading) Ltd.
+ */
+
+#include <linux/err.h>
+#include <linux/gpio/driver.h>
+#include <linux/module.h>
+#include <linux/platform_device.h>
+#include <linux/pwm.h>
+
+struct pwm_gpio {
+	struct gpio_chip gc;
+	struct pwm_device **pwm;
+};
+
+static int pwm_gpio_get_direction(struct gpio_chip *gc, unsigned int off)
+{
+	return GPIO_LINE_DIRECTION_OUT;
+}
+
+static void pwm_gpio_set(struct gpio_chip *gc, unsigned int off, int val)
+{
+	struct pwm_gpio *pwm_gpio = gpiochip_get_data(gc);
+	struct pwm_state state;
+
+	pwm_get_state(pwm_gpio->pwm[off], &state);
+	state.duty_cycle = val ? state.period : 0;
+	pwm_apply_state(pwm_gpio->pwm[off], &state);
+}
+
+static int pwm_gpio_parse_dt(struct pwm_gpio *pwm_gpio,
+			     struct device *dev)
+{
+	struct device_node *node = dev->of_node;
+	struct pwm_state state;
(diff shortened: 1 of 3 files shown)


drivers/gpio/gpio-pwm.c
```diff
@@ -0,0 +1,132 @@
...
+ * Copyright (C) 2021 Raspberry Pi (Trading) Ltd.
+ */
+
+#include <linux/err.h>
+#include <linux/gpio/driver.h>
+#include <linux/module.h>
+#include <linux/platform_device.h>
+#include <linux/pwm.h>
+
+#define NUM_GPIO 1
```
pelwell: Have you considered using `of_count_phandle_with_args` to remove the need for this constant?

6by9: As per the first comment, I'd love to get rid of the define and read it from the array size of ```pwms```, I just couldn't find the right API calls to do so.

Ideally I want a ```devm_pwm_get``` call that takes an index, but it looks like the PWM API doesn't allow that.
https://elixir.bootlin.com/linux/latest/source/drivers/pwm/core.c#L756
```
 * of_pwm_get() - request a PWM via the PWM framework
 * @dev: device for PWM consumer
 * @np: device node to get the PWM from
 * @con_id: consumer name
 *
 * Returns the PWM device parsed from the phandle and index specified in the
 * "pwms" property of a device tree node or a negative error-code on failure.
 * Values parsed from the device tree are stored in the returned PWM device
 * object.
 *
 * If con_id is NULL, the first PWM device listed in the "pwms" property will
 * be requested. Otherwise the "pwm-names" property is used to do a reverse
 * lookup of the PWM index. This also means that the "pwm-names" property
 * becomes mandatory for devices that look up the PWM device via the con_id
 * parameter.
```
So it's pwm-names, or only one pwm :-(
I can see an array of names "gpio0", "gpio1", "gpio2" in my future.

pelwell: It does sound a bit painful, but with `of_property_count_strings` and `of_property_read_string_index` at least your driver won't need to hard-code any names - just the overlay.

6by9: Thank you - works beautifully.

6by9:
Updated to wrap all PWMs listed in DT as GPIOs. Names are needed due to the way the way the API works.

dtoverlay hack dropped as it doesn't want to be merged. For reference, it was
```
+
+       fragment@1 {
+               target-path = "/";
+               __overlay__ {
+                       pwm_gpio: pwm_gpio@0 {
+                               compatible = "pwm-gpio";
+                               reg = <0x0>;
+                               gpio-controller;
+                               #gpio-cells = <2>;
+
+                               pwms =  <&pca 15 1000000>, //100ns or 1kHz
+                                       <&pca 14 1000000>,
+                                       <&pca 13 1000000>;
+                               pwm-names = "gpio1", "gpio2", "gpio3";
+
+                               status = "okay";
+                       };
+               };
+       };
+
```
Copying the example set by gpio-fsm, no dt binding doc as it won't be acceptable to upstream anyway ;-)

pelwell:
I'm happy with that. Any objections to me ignoring the "RFC" and merging?

6by9:
RFC removed

PR #4639, opened 2021-10-19
clk: Move vec clock to clk-raspberrypi

popcornmix:
clk-2835 is deprecated and gets an innacurate clock for vec (107.143MHz).
Switch to clk-raspberrypi which uses the right PLL to get an accurate 108MHz


pelwell:
I'm happy with the substance of the PR, but for ease of management it should probably be three commits - removing from the old clock (with a comment, as 661ef43ea28a8fc6d921d30ec1329a1a76139f61), adding to the new clock driver, and the DT changes.

popcornmix:
Updated

pelwell:
I guess that will do.

popcornmix:
Removed draft tag after report of it working in https://github.com/raspberrypi/linux/pull/4406

PR #4640, opened 2021-10-19
Add CONFIG_FB_SIMPLE to default configuration (arm64 rpi3) fix no video output

eriktagirov:
Hello,

I was trying to compile the latest kernel from sources for the Rpi3 in ARM64 mode.
So I followed this guide: https://beta.raspberrypi.org/documentation/computers/linux_kernel.html#cross-compiling-the-kernel
Applied the `bcmrpi3_defconfig` (without changing any parameter in menuconfig) the kernel booted but without any video. 

After looking at the output of the console and comparing to the console output of the latest `kernel8.img` binary. The only difference was the lack of the `simple-framebuffer` node.
This patch to the default configuration did the trick for me.

BR


pelwell:
Our 64-bit kernel is shared between Pi 3 and Pi 4, and we build it using bcm2711_defconfig (bcmrpi3_defconfig was contributed by a user before we started creating official 64-bit builds). However, there's no harm in adding FB_SIMPLE=y to the bcmrpi3_defconfig - I've just moved the setting so that it a future `savedefconfig` doesn't cause a change in order.

PR #4641, opened 2021-10-20
Unicam Media Controller support

6by9:
Updates bcm2835-unicam to allow it to use Media Controller fully for the more complex pipelines.

~This does still want to be updated to add a DT switch between MC-centric and video-centric in addition to the module parameter~. There's then the discussion over which we adopt as the default.
@naushir has patches to make libcamera work with the new scheme, and that is the way we want to go, but it may need some coordination in changeover.

### DIFF
--- a/drivers/media/platform/bcm2835/bcm2835-unicam.c
+++ b/drivers/media/platform/bcm2835/bcm2835-unicam.c
@@ -380,6 +380,8 @@ struct unicam_node {
 	int open;
 	bool streaming;
 	unsigned int pad_id;
+	/* Source pad id on the sensor for this node */
+	unsigned int src_pad_id;
 	/* Pointer pointing to current v4l2_buffer */
 	struct unicam_buffer *cur_frm;
 	/* Pointer pointing to next v4l2_buffer */
@@ -590,7 +592,7 @@ static int __subdev_get_format(struct unicam_device *dev,
 {
 	struct v4l2_subdev_format sd_fmt = {
 		.which = V4L2_SUBDEV_FORMAT_ACTIVE,
-		.pad = pad_id
+		.pad = dev->node[pad_id].src_pad_id,
 	};
 	int ret;
 
@@ -612,7 +614,7 @@ static int __subdev_set_format(struct unicam_device *dev,
 {
 	struct v4l2_subdev_format sd_fmt = {
 		.which = V4L2_SUBDEV_FORMAT_ACTIVE,
-		.pad = pad_id
+		.pad = dev->node[pad_id].src_pad_id,
 	};
 	int ret;
 
@@ -1977,7 +1979,7 @@ static int unicam_enum_framesizes(struct file *file, void *priv,
 
 	fse.which = V4L2_SUBDEV_FORMAT_ACTIVE;
 	fse.index = fsize->index;
-	fse.pad = node->pad_id;
+	fse.pad = node->src_pad_id;
 
 	ret = v4l2_subdev_call(dev->sensor, pad, enum_frame_size, NULL, &fse);
 	if (ret)
@@ -2002,6 +2004,7 @@ static int unicam_enum_frameintervals(struct file *file, void *priv,
 	const struct unicam_fmt *fmt;
 	struct v4l2_subdev_frame_interval_enum fie = {
 		.index = fival->index,
+		.pad = node->src_pad_id,
 		.width = fival->width,
 		.height = fival->height,
 		.which = V4L2_SUBDEV_FORMAT_ACTIVE,
@@ -2093,8 +2096,13 @@ static int unicam_enum_dv_timings(struct file *file, void *priv,
 {
 	struct unicam_node *node = video_drvdata(file);
 	struct unicam_device *dev = node->dev;
+	int ret;
 
-	return v4l2_subdev_call(dev->sensor, pad, enum_dv_timings, timings);
+	timings->pad = node->src_pad_id;
+	ret = v4l2_subdev_call(dev->sensor, pad, enum_dv_timings, timings);
+	timings->pad = node->pad_id;
+
+	return ret;
 }
 
 static int unicam_dv_timings_cap(struct file *file, void *priv,
@@ -2102,8 +2110,13 @@ static int unicam_dv_timings_cap(struct file *file, void *priv,
 {
 	struct unicam_node *node = video_drvdata(file);
 	struct unicam_device *dev = node->dev;
+	int ret;
 
-	return v4l2_subdev_call(dev->sensor, pad, dv_timings_cap, cap);
+	cap->pad = node->src_pad_id;
+	ret = v4l2_subdev_call(dev->sensor, pad, dv_timings_cap, cap);
+	cap->pad = node->pad_id;
+
+	return ret;
 }
 
 static int unicam_subscribe_event(struct v4l2_fh *fh,
@@ -2374,14 +2387,12 @@ static int register_node(struct unicam_device *unicam, struct unicam_node *node,
 			 */
 			fmt = get_first_supported_format(unicam);
 
-			if (!fmt)
-				/* No compatible formats */
-				return -EINVAL;
-
-			mbus_fmt.code = fmt->code;
-			ret = __subdev_set_format(unicam, &mbus_fmt, pad_id);
-			if (ret)
-				return -EINVAL;
+			if (fmt) {
+				mbus_fmt.code = fmt->code;
+				ret = __subdev_set_format(unicam, &mbus_fmt, pad_id);
+				if (ret)
+					return -EINVAL;
+			}
 		}
 		if (mbus_fmt.field != V4L2_FIELD_NONE) {
 			/* Interlaced not supported - disable it now. */
@@ -2391,7 +2402,8 @@ static int register_node(struct unicam_device *unicam, struct unicam_node *node,
 				return -EINVAL;
 		}
 
-		node->v_fmt.fmt.pix.pixelformat = fmt->fourcc ? fmt->fourcc
+		if (fmt)
+			node->v_fmt.fmt.pix.pixelformat = fmt->fourcc ? fmt->fourcc
 						: fmt->repacked_fourcc;
 	} else {
 		/* Fix this node format as embedded data. */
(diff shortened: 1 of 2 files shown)


6by9:
Added a couple of extra patches that will be squashed into the "Add support for configuration via MC API", but Naush may have started reviewing.

drivers/media/platform/bcm2835/bcm2835-unicam.c
```diff
@@ -2293,30 +2354,6 @@ static void unicam_return_buffers(struct unicam_node *node,
 	spin_unlock_irqrestore(&node->dma_queue_lock, flags);
 }
 
-static int unicam_check_format(struct unicam_node *node)
```
pelwell: I think this hunk should be squashed with the previous commit - it was added there and not used.

arch/arm64/configs/bcm2711_defconfig
```diff
@@ -1404,6 +1405,7 @@ CONFIG_ANDROID=y
 CONFIG_ANDROID_BINDER_IPC=y
 CONFIG_ANDROID_BINDERFS=y
 CONFIG_NVMEM_RMEM=m
+CONFIG_MUX_GPIO=m
 CONFIG_EXT4_FS=y
 CONFIG_EXT4_FS_POSIX_ACL=y
 CONFIG_EXT4_FS_SECURITY=y
```
pelwell: Some people out there use the arm64 bcmrpi3_defconfig, so let's keep it up-to-date.

arch/arm/boot/dts/overlays/imx219-mux-overlay.dts
```diff
@@ -0,0 +1,235 @@
...
+		__overlay__ {
+			mipi_sw_oe_hog {
+				gpio-hog;
+				gpios = <0x17 GPIO_ACTIVE_LOW>;
+				output-high;
+			};
+		};
+	};
+
+	__overrides__ {
```
pelwell: This isn't necessary, but I won't object if you'd prefer it to be retained as a placeholder.

arch/arm/boot/dts/overlays/imx219-mux-overlay.dts
```diff
@@ -0,0 +1,235 @@
+// SPDX-License-Identifier: GPL-2.0-only
```
pelwell: The overlay looks fine, but it needs a brief description in the dreaded README.

drivers/media/platform/bcm2835/bcm2835-unicam.c
6by9: > "Added a couple of extra patches that will be squashed into the "Add support for configuration via MC API", but Naush may have started reviewing."

There are further changes coming which backport work being done on mainline, but I wanted Naush to look at the basic changes first.

arch/arm64/configs/bcm2711_defconfig
6by9: OK, will do.

arch/arm/boot/dts/overlays/imx219-mux-overlay.dts
6by9: Legacy. This overlay is very rough and ready (hence no README) to prove the concept of two image sensors off a mux such as https://thepihut.com/products/multi-camera-doubleplexer-stereo-module-v2. I'd started expecting to do 2 imx219's (hence the name), but ended up with an imx219 and an imx477.

What we really want is a way of specifying the sensor name attached to each port, and then import the relevant config from the existing overlays (imx219, imx477, ov5647, ov9281, ov7251, tc358743, imx517, or any others). I don't think that is really practical though, but I could be wrong.

The main thing for the PR is the driver changes, and we'll have to think about whether we want overlays supporting the mux as a further change.

drivers/media/platform/bcm2835/bcm2835-unicam.c
pelwell: It would be kinder to squash it now than to have Naush review a new function which is never used.

drivers/media/platform/bcm2835/bcm2835-unicam.c
```diff
@@ -2593,7 +2610,20 @@ static int unicam_probe_complete(struct unicam_device *unicam)
...
 		return -ENOMEM;
 
-	unicam->sensor_embedded_data = (unicam->sensor->entity.num_pads >= 2);
+	for (i = 0; i < unicam->sensor->entity.num_pads; i++) {
+		if (unicam->sensor->entity.pads[i].flags & MEDIA_PAD_FL_SOURCE) {
+			if (source_pads < MAX_NODES) {
+				unicam->node[source_pads].src_pad_id = i;
+				unicam_err(unicam, "source pad %u is index %u\n",
+					   source_pads, i);
+			}
```
naushir, line 3111: Change to unicam_dbg?

drivers/media/platform/bcm2835/bcm2835-unicam.c
```diff
@@ -2411,15 +2673,69 @@ static int register_node(struct unicam_device *unicam, struct unicam_node *node,
...
 	}
 
+	*ret_fmt = fmt;
+
+	return 0;
+}
+
+static void unicam_mc_set_default_format(struct unicam_node *node, int pad_id)
+{
+	if (pad_id == IMAGE_PAD) {
```
naushir, line 2838: I still wonder if this should query the subdev for the format, but it's not something other drivers do...

drivers/media/platform/bcm2835/bcm2835-unicam.c
```diff
@@ -2776,6 +3180,14 @@ static int unicam_probe(struct platform_device *pdev)
...
 	unicam->pdev = pdev;
 
+	/*
+	 * Adopt the current setting of the module parameter, and check if
+	 * device tree requests it.
+	 */
+	unicam->mc_api = media_controller;
+	if (of_property_read_bool(pdev->dev.of_node, "brcm,media-controller"))
+		unicam->mc_api = true;
+
```
naushir, line 3308: Should we be updating imx477 and imx219 overlays to use this?

drivers/media/platform/bcm2835/bcm2835-unicam.c
6by9: Done

drivers/media/platform/bcm2835/bcm2835-unicam.c
6by9: This conversion is based largely on Tomi's conversion of ti-vpe, and that doesn't do it.
AIUI The only interaction between the subdevs should be during link_validate, and we do that now.

drivers/media/platform/bcm2835/bcm2835-unicam.c
6by9: Yes, although with an override to disable it.
ov5647 as well, and presumably imx517. Basically anything that is expected to use libcamera should default to it being on, and other sensors could default to off (eg ov9281 and ov7251).

naushir: Well, in that case, should it default to MC and have a switch to turn it off...?

6by9: Things like ADV728x and TC358743 remain with it off.

It's 3rd party overlays/drivers that I'm thinking of. They currently work, so making them need changes to keep them working is a bit mean.

drivers/media/platform/bcm2835/bcm2835-unicam.c
```diff
@@ -1267,902 +1319,1216 @@ static int unicam_g_fmt_meta_cap(struct file *file, void *priv,
...
+	.vidioc_querycap      = unicam_querycap,
+	.vidioc_enum_fmt_vid_cap  = unicam_mc_enum_fmt_vid_cap,
+	.vidioc_g_fmt_vid_cap     = unicam_mc_g_fmt_vid_cap,
+	.vidioc_try_fmt_vid_cap   = unicam_mc_try_fmt_vid_cap,
+	.vidioc_s_fmt_vid_cap     = unicam_mc_s_fmt_vid_cap,
+
+	.vidioc_enum_fmt_meta_cap	= unicam_mc_enum_fmt_meta_cap,
+	.vidioc_g_fmt_meta_cap		= unicam_mc_g_fmt_meta_cap,
+	.vidioc_s_fmt_meta_cap		= unicam_g_fmt_meta_cap,
+	.vidioc_try_fmt_meta_cap	= unicam_g_fmt_meta_cap,
```
naushir: Typo!

6by9: Fixed

arch/arm/boot/dts/overlays/imx219-mux-overlay.dts
6by9: imx219-mux-overlay removed from the PR.

arch/arm64/configs/bcm2711_defconfig
6by9: Dropped from the PR for now.

6by9:
Rebased, fixed up comments, added updates to all CSI source overlays, and squashed.

Overlays shouldn't change now (expect based on review comments), but v4l2-compliance is complaining so I need to fix that up.

6by9:
OK, hopefully that has everything sorted.

v4l2-compliance is now happy with both nodes. I have also changed the default embedded data buffer size to 16kB - sorry Naush as that sort of makes your latest changes redundant (at least for now).

Removing WIP tag, but want Naush's confirmation that we're now all good from his side.

naushir:
Yup, all looks good to me.

I'm happy to merge all but d49eb3dda85d6f4782262fa761f21b958eefc9e3 straight away. If we can hold off merging the overlays switch to MC, that can be co-ordinated with the userland changes.

6by9:
I'll update PR to drop https://github.com/raspberrypi/linux/commit/d49eb3dda85d6f4782262fa761f21b958eefc9e3 and make a new PR for it (to be merged when libcamera changes are also accepted).

6by9:
Overlays patch dropped. Should be good to go.

PR #4643, opened 2021-10-21
Colorspace fixes

davidplowman:
@6by9 @naushir
This is a set of small changes to fix colour space issues noticed during libcamera development. Most of them involve sensors that were incorrectly reporting sRGB rather than RAW colour spaces. Additionally I've also made our ISP advertise that it wants full, rather than limited, range RAW input.

Nothing was actually failing before, it was a case of the debug warnings showing that colour spaces were wrong, even though they were being over-written later. I've worked through each of the affected camera modules double-checking that the changes don't break anything!


6by9:
The content looks fine, but none of the commit messages contain the name of the driver that they're changing.
Normally they'd be 
```media: i2c: imx219: Blah```
or similar.

davidplowman:
Ah yes indeed, that should be better now!

PR #4644, opened 2021-10-22
firmware: raspberrypi: Use g_pdev pointer in rpi_firmware_get()

iluminat23:
Since the merge a35653e rpi_firmware_get() can't be called with a NULL pointer anymore. The bcm2835-sdhost driver relies on this. Which then leads to a NULL pointer dereference.


pelwell:
Already fixed - see 48c6ccc8f304.

PR #4645, opened 2021-10-22
dtoverlays: Update all image sensor overlays for Media Controller option

6by9:
Add an option to enable configuration via the Media Controller API
(rather than the video-node-centric /dev/videoN) as about to
be used by libcamera as it enables more complex pipelines to be
handled.

Any source that has a libcamera tuning merged has MC enabled by
default.
Sources with no libcamera tuning merged have it disabled by
default.
In either case it can be overridden with the overlay parameter
"media-controller".

Follow up to #4641 to switch over to MC usage. Wants the libcamera updates to be merged first/at the same time.

### DIFF
--- a/arch/arm/boot/dts/overlays/adv7282m-overlay.dts
+++ b/arch/arm/boot/dts/overlays/adv7282m-overlay.dts
@@ -59,7 +59,15 @@
 		};
 	};
 
+	fragment@4 {
+		target = <&csi1>;
+		__dormant__ {
+			brcm,media-controller;
+		};
+	};
+
 	__overrides__ {
 		addr =			<&adv728x>,"reg:0";
+		media-controller = <0>,"=4";
 	};
 };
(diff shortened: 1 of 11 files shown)


arch/arm/boot/dts/overlays/adv7282m-overlay.dts
```diff
@@ -59,7 +59,15 @@
...
+	fragment@4 {
+		target = <&csi1>;
+		__dormant__ {
+			brcm,media-controller;
+		};
+	};
+
 	__overrides__ {
 		addr =			<&adv728x>,"reg:0";
+		media-controller = "=4";
```
pelwell: Overlay-scope parameters like these require a `<0>` phandle before the string of fragment changes.

6by9: Thanks. Not sure how I missed that one.

6by9:
Updated to fix the phandles for the overrides.

pelwell:
That looks OK. Let me know when you'd like it merged, @naushir.

naushir:
Will do, I've published the accompanying userland changes to the mailing list, so once the review cycle is done, I'll let you know.

naushir:
@popcornmix, this can be merged now.

6by9:
WIP tag removed - libcamera changes are merged.

PR #4646, opened 2021-10-22
media: bcm2835-codec: Limit video callbacks

jc-kynesim:
Limit the number of allowed video callbacks. This helps with limiting
the size of the coded input FIFO which in turn helps to control latency.
Choose -5 as the magic number as it translates to DPB+5 buffers which
has been proven to be a good number in the past.

Ideally coded buffers would not be returned to the user until they
had been decoded into the DPB or been discarded as bad, but that grade
of control is unavailable to us.


jc-kynesim:
The helps with (but on its own does not solve) some of my issues with ffmpeg on the command line + hwaccel.  As stated above I picked -5 as its a number Dom has used in the past that seemed to work.  Ideally -1 or -2 would be better but those have stalled for Dom in the past so don't do that.

popcornmix:
I used this setting for kodi with mmal acceleration as otherwise the latency between enqueued and dequeued buffers can grow to several seconds and media players generally don't expect that behaviour.
So seems sane to me.

6by9:
I see no issues with that, although it'd be useful to check it against GStreamer. Give me a mo to do that.

6by9:
All good in GStreamer (1.18.4 anyway), so I'm good with it.

PR #4647, opened 2021-10-22
staging/bcm2835-camera: Add support for H264_MIN_QP, H264_MAX_QP, MPEG_VIDEO_FORCE_KEY_FRAME

soyersoyer:


6by9:
Thanks for the PR.

A max of 40? I thought QP went up to 51.
I'd want to check whether the min was 0 or 1 as well.

soyersoyer:
I didn't find good documentation for MMAL_PARAMETER_VIDEO_ENCODE_MIN_QUANT. The value 40 comes from the picamera docs.
Setting MMAL_PARAMETER_VIDEO_ENCODE_MIN_QUANT to 0 doesn't work. What is the MMAL_PARAMETER_VIDEO_ENCODE_MIN_QUANT's default value?

6by9:
https://vcodex.blogspot.com/2009/04/h264-quantization-parameter.html
 > Quantization in an H.264 encoder is controlled by a quantization parameter, QP, that ranges from 0 to 51. QP is an index used to derive a scaling matrix. It is possible to calculate the equivalent quantizer step size (Qstep) for each value of QP (see my tutorial on H.264 4x4 transform). As QP increases, Qstep increases; in fact, Qstep doubles for every increase of 6 in QP. The logarithmic relationship can be seen in this graph of QP (x-axis, linear) vs. Qstep (y-axis, logarithmic). 

So I would expect it to support the full range. The smaller the QP the larger the output frame.
I'll check the firmware source when I get a chance.

soyersoyer:
It works within the full range. I can't set them to 0, while the camera is streaming, but otherwise it works.

6by9:
Great. That looks fine to me.

PR #4648, opened 2021-10-22
staging/bcm2835-camera: Limit video bitrate according to H264 level

soyersoyer:


drivers/staging/vc04_services/bcm2835-camera/controls.c
```diff
@@ -596,6 +604,10 @@ static int ctrl_set_bitrate(struct bm2835_mmal_dev *dev,
 	int ret;
 	struct vchiq_mmal_port *encoder_out;
 
+	int max_rate = h264_level_max_rates[dev->capture.enc_level];
```
6by9, line 607: We could be encoding to MJPEG, at which point this check isn't valid.

6by9 (commented):
Thanks for the patch.

One minor issue with MJPEG not using H264 level - see inline.

One bigger issue is that the V4L2 control should really change range based on the level configured, therefore the better way to do this is to call ```__v4l2_ctrl_modify_range``` on the bitrate control when level is changed. If you look at https://github.com/raspberrypi/linux/blob/rpi-5.10.y/drivers/media/i2c/imx219.c then you'll see it does this to exposure as VBLANK (effectively frame rate) is changed.
Without that I suspect the v4l2-compliance tests will fail as they typically test each control can be set to min and max.

drivers/staging/vc04_services/bcm2835-camera/controls.c
soyersoyer: Hm, but we don't know whether the user wants to encode h264 or mjpeg in this function. Should we limit the bitrate somewhere later?

6by9: We have ```dev```, so ```dev->capture.fmt->fourcc == V4L2_PIX_FMT_H264``` will only be true for H264.

If the control's range is modified, then the V4L2 framework ensures that the selected bitrate is in range. That's the better approach.

PR #4649, opened 2021-10-23
staging/bcm2835-camera: Add support for V4L2_MPEG_VIDEO_BITRATE_MODE_CQ

soyersoyer:
Add support for constant quality mode. In this mode the quality can be set via V4L2_CID_MPEG_VIDEO_CONSTANT_QUALITY.

### DIFF
--- a/drivers/staging/vc04_services/bcm2835-camera/bcm2835-camera.h
+++ b/drivers/staging/vc04_services/bcm2835-camera/bcm2835-camera.h
@@ -13,7 +13,7 @@
  * core driver device
  */
 
-#define V4L2_CTRL_COUNT 32 /* number of v4l controls */
+#define V4L2_CTRL_COUNT 33 /* number of v4l controls */
 
 enum {
 	COMP_CAMERA = 0,
@@ -41,6 +41,8 @@ struct bm2835_mmal_dev {
 	/* controls */
 	struct v4l2_ctrl_handler ctrl_handler;
 	struct v4l2_ctrl *ctrls[V4L2_CTRL_COUNT];
+	struct v4l2_ctrl *bitrate_ctrl;
+
 	enum v4l2_scene_mode scene_mode;
 	struct mmal_colourfx colourfx;
 	int hflip;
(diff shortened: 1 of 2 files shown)


6by9:
Reality is that the 62.5Mbit/s limit of level 4.2 isn't achievable by the hardware, at least not in realtime. As bcm2835-camera is only fed by the camera and is therefore only realtime, then it makes little sense increasing the limit.

For bcm2835-codec where we may be transcoding in non-realtime, then there would be a case for it.

soyersoyer:
I am playing with this a bit more. If I set the bitrate to 0, setting the MMAL_PARAMETER_VIDEO_ENCODE_MIN_QUANT, MMAL_PARAMETER_VIDEO_ENCODE_MAX_QUANT doesn't have any effect. But setting the MMAL_PARAMETER_VIDEO_ENCODE_INITIAL_QUANT works.

What is the best way to express this behaviour in v4l2?
Set the INITIAL_QUANT and MIN_QUANT in the H264_MIN_QP control too?
Map the INITIAL_QUANT to H264_I_FRAME_QP or other unused control?
If the bitrate is 0, force the H264_MIN_QP to be equal TO H264_MAX_QP (and INITIAL_QUANT)?

6by9:
This is why I hate getting involved in overriding Qp as it gets involved as to what the dependencies are.

I'd need to have a serious dig through the code as to the dependencies between bitrate, min/max qp, initial qp, bitrate mode, and possibly a number of other parameters. There's no guarantee that they match with the expectations of V4L2, and even that may not be clearly specified.

Rereading https://www.kernel.org/doc/html/latest/userspace-api/media/v4l/ext-ctrls-codec.html
```V4L2_CID_MPEG_VIDEO_FRAME_RC_ENABLE```` implies totally fixed if disabled, but aiming at the bitrate but within min/max qp if enabled. What do you do if the bitrate exceeds the target with max qp? Just ignore it and keep going?

And it really needs to set the min and max for I and P frames (we don't support B frames), but I'd need to look again at the firmware code to work out whether that is supported (it looks like it may take one value and have a weighting between I and P frame QP)

soyersoyer:
Is the V4L2_CID_MPEG_VIDEO_FRAME_RC_ENABLE mandatory? I mean if the user can't set it, because the driver doesn't support it, it's just unset, not disabled.

PR #4650, opened 2021-10-24
staging/bcm2835-camera: Add support for MPEG_VIDEO_FORCE_KEY_FRAME

soyersoyer:


6by9:
I think due to https://elixir.bootlin.com/linux/latest/source/drivers/media/v4l2-core/v4l2-ctrls-defs.c#L1285 that min, max, and def will all be set to 0. Could you check with a ```v4l2-ctl --list-ctrls```?

Otherwise it looks like it should work, however can I ask you to add a "Signed-off by" line to the commit? ```git commit --amend -s``` does it in the right format for you based on your configured user information. This is a file that is in the mainline kernel, and changes ought to be pushed upstream, therefore we need to preserve the authorship on it - https://www.kernel.org/doc/html/latest/process/submitting-patches.html#sign-your-work-the-developer-s-certificate-of-origin

soyersoyer:
The v4l2-ctl output:
force_key_frame 0x009909e5 (button) : flags=write-only, execute-on-write
With VIDIOC_QUERYCTRL:
Force Key Frame	( default: 0 min: 0 max: 0 step: 0)
It works, maybe the 1 is not necessary for the MMAL_PARAMETER_VIDEO_REQUEST_I_FRAME.

soyersoyer:
Should I merge the two PR? They are changing the V4L2_CTRL_COUNT too.

6by9:
> Should I merge the two PR? They are changing the V4L2_CTRL_COUNT too.

They do conflict, so either one gets merged and then you update the other, or they need to be merged. Personally I don't mind which approach you take. But please make use of one of your existing PRs by updating the relevant branch instead of closing both and opening a new one.

Issue #4651, opened 2021-10-25
bcm2835-audio HDMI device is not present on Pi 4 running with KMS and VC4 HDMI devices disabled


KonstaT:
I'm experiencing a similar issue to https://github.com/raspberrypi/linux/issues/4634 that wasn't fixed with https://github.com/raspberrypi/linux/commit/214bf6b3eeae500ac13887ce7b2cca576191f38e.

When running on Pi 4 with KMS and disabling VC4 HDMI devices (i.e. setting `dtoverlay=vc4-kms-v3d,noaudio` in /boot/config.txt) bcm2835-audio HDMI device is not present as it should be.

Running 5.10.74 kernel with latest firmware (https://github.com/raspberrypi/firmware/tree/d2ab264b688a223e23029ffbba86820c8c6cb329) I get:
```
[    3.899252] ALSA device list:
[    3.902277]   #0: bcm2835 Headphones
```

Same kernel but downgrade the firmware to https://github.com/raspberrypi/firmware/tree/d06e0ef85c22825a4491497a1d2a4787471f011b I get:
```
[   10.422612] ALSA device list:
[   10.425629]   #0: bcm2835 HDMI 1
[   10.428908]   #1: bcm2835 Headphones
```
This would be the expected behavior. This doesn't work with any newer firmware (bisected this to something that changed between https://github.com/raspberrypi/firmware/tree/d06e0ef85c22825a4491497a1d2a4787471f011b and https://github.com/raspberrypi/firmware/tree/19272ccd69049aaf42c78a235a0bf37dbabd5ea7)

6by9:
When running vc4-kms-v3d ALL configuration of the HDMI block is done by the kernel.
HDMI audio is inserted into the blanking periods of the video signal, and is therefore dependent on the video mode in use. It also requires signalling in the infoframes as to exactly what audio format is being sent.

bcm2835-audio uses the firmware to handle the audio routing. With vc4-kms-v3d and the kernel in charge of the HDMI block, mode selection, and infoframes, it is not possible for the firmware to correctly insert audio, therefore it can't be supported.
The behaviour under 5.10.74 would therefore be correct - you've asked for no audio, and you've got exactly that.

pelwell:
@mripard Interestingly, I get no picture when the `noaudio` parameter is used on Pi 4 with top-of-tree kernel (f814bfc5f4d3):
```
dtoverlay=vc4-kms-v3d,noaudio
```
```
[...]

PR #4652, opened 2021-10-25
ARM: dts: vc4-kms-v3d: Always disable firmware HDMI

pelwell:
Both the firmware audio driver and the vc4-kms-v3d driver are capable
of providing HDMI audio, but only one should be active at any time.
The vc4-kms-v3d overlays disable the firmware audio driver, but they
also have a noaudio parameter that as well as disabling the ARM-side
HDMI audio also re-enables the firmware HDMI audio. This is not
guaranteed to work and has been seen to break the display completely.

Modify the noaudio parameters so that the firmware HDMI audio support
remains disabled.

See: https://github.com/raspberrypi/linux/issues/4651


6by9:
LGTM.

timg236:
LGTM

PR #4655, opened 2021-10-28
firmware: raspberrypi: Fix botched upstream stable merge

juergh:
The merge of upstream stable commit 60831f5ae6c7 ("firmware: raspberrypi:
Keep count of all consumers") accidentially replaced a statement from
downstream commit e4a6adeb157e ("firmware: bcm2835: Support ARCH_BCM270x").
Fix that.

Signed-off-by: Juerg Haefliger <juergh@canonical.com>

### DIFF
--- a/drivers/firmware/raspberrypi.c
+++ b/drivers/firmware/raspberrypi.c
@@ -431,7 +431,7 @@ static int rpi_firmware_remove(struct platform_device *pdev)
  */
 struct rpi_firmware *rpi_firmware_get(struct device_node *firmware_node)
 {
-	struct platform_device *pdev = of_find_device_by_node(firmware_node);
+	struct platform_device *pdev = g_pdev;
 	struct rpi_firmware *fw;
 
 	if (!pdev)


pelwell:
Although this has been reported elsewhere (e.g. https://github.com/raspberrypi/linux/pull/4644), this change is intentional. If you think there are problems caused by the new approach then please highlight them and we'll tackle them on a case-by-case basis.

krzk:
The trouble is that RPi Git tree has now confusing history. The merge commit a35653ed does not show any conflict resolution in that place. Also with `git show --cc`. Only `git show -c` points it, but it's not a default. When repeating the merge, one can clearly see there is a merge conflict and with normal resolution (`git mergetool`) the output will be different than yours Phil. Therefore:
1. The merge misses information that there was a conflict resolution and specific one branch was chosen.
2. The merge chooses non-obvious resolution, e.g. discards the actual change from RPi branch. Silently.
3. The merge does not explain in the commit why the code was discarded, why other hunk (from stable) was chosen.

`git show 48c6ccc8f` shows nothing about this change. While the expected is to see either a commit following up 48c6ccc8f or something in the merge commit -  `+` or `space+` conflict resolution like this one (automatic one):
```
@@@ -405,7 -314,8 +431,8 @@@
   */
  struct rpi_firmware *rpi_firmware_get(struct device_node *firmware_node)
  {
 -      struct platform_device *pdev = of_find_device_by_node(firmware_node);
 +      struct platform_device *pdev = g_pdev;
+       struct rpi_firmware *fw;
  
        if (!pdev)
                return NULL;
```

juergh:
Phil, what we're saying is that, by trying to understand the git history, we fail to understand how this is an intentional change. Is that single-line change supposed to be part of 48c6ccc? In that commit message you mention that a downstream patch can be removed, which one is that?

pelwell:
I'm out of office today, away from keyboard, so bear with me. Something has clearly gone wrong with the merge, and for that I apologise. This was actually my first real merge - I usually stick to rebases and cherry picks - and I hope to learn from the mistake, whatever it was.

The global variable g_pdev does not appear in the upstream firmware driver, and the intention is that the downstream driver should be as close to that as possible. The merge was meant to remove the use of g_pdev from the common functions, leaving the minimum needed to support the reboot notification.

When I spoke of removing a downstream patch I was referring to the g_pdev use in rpi_firmware_get, but although it is added in e4a6adeb, reverting that would remove too much.

juergh:
Thanks for the elaboration, Phil. It seems this is a no-issue so closing the ticket.

PR #4656, opened 2021-10-28
media: i2c: imx477: Add vsync trigger_mode parameter

neocortex-vision:
trigger_mode == 0 (default) => no effect / no registers written
trigger_mode == 1           => source
trigger_mode == 2           => sink

This can be set e.g. in /boot/cmdline.txt as imx477.trigger_mode=N
Setting it to 0 or not setting it at all has the same effect.

Tested on HQ cameras (XVS <-> XVS, GND <-> GND) with libcamera.

Credits to 6by9.

### DIFF
--- a/drivers/media/i2c/imx477.c
+++ b/drivers/media/i2c/imx477.c
@@ -25,6 +25,10 @@ static int dpc_enable = 1;
 module_param(dpc_enable, int, 0644);
 MODULE_PARM_DESC(dpc_enable, "Enable on-sensor DPC");
 
+static int trigger_mode;
+module_param(trigger_mode, int, 0644);
+MODULE_PARM_DESC(trigger_mode, "Set vsync trigger mode: 1=source, 2=sink");
+
 #define IMX477_REG_VALUE_08BIT		1
 #define IMX477_REG_VALUE_16BIT		2
 
@@ -97,6 +101,12 @@ MODULE_PARM_DESC(dpc_enable, "Enable on-sensor DPC");
 #define IMX477_TEST_PATTERN_GR_DEFAULT	0
 #define IMX477_TEST_PATTERN_B_DEFAULT	0
 #define IMX477_TEST_PATTERN_GB_DEFAULT	0
+
+/* Trigger mode */
+#define IMX477_REG_MC_MODE		0x3f0b
+#define IMX477_REG_MS_SEL		0x3041
+#define IMX477_REG_XVS_IO_CTRL		0x3040
+#define IMX477_REG_EXTOUT_EN		0x4b81
 
 /* Embedded metadata stream structure */
 #define IMX477_EMBEDDED_LINE_WIDTH 16384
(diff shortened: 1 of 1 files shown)


6by9:
The content looks OK, although we get a couple of warnings from ```scripts/checkpatch.pl --strict --codespell```
```
No codespell typos will be found - file '/usr/share/codespell/dictionary.txt': No such file or directory
ERROR: do not initialise statics to 0
#25: FILE: drivers/media/i2c/imx477.c:28:
+static int trigger_mode = 0;

#55: FILE: drivers/media/i2c/imx477.c:1738:
+		imx477_write_reg(imx477, IMX477_REG_MC_MODE,
+			IMX477_REG_VALUE_08BIT, 1);

#57: FILE: drivers/media/i2c/imx477.c:1740:
+		imx477_write_reg(imx477, IMX477_REG_MS_SEL,
+			IMX477_REG_VALUE_08BIT, val);

#59: FILE: drivers/media/i2c/imx477.c:1742:
+		imx477_write_reg(imx477, IMX477_REG_XVS_IO_CTRL,
+			IMX477_REG_VALUE_08BIT, val);

#61: FILE: drivers/media/i2c/imx477.c:1744:
+		imx477_write_reg(imx477, IMX477_REG_EXTOUT_EN,
+			IMX477_REG_VALUE_08BIT, val);

total: 2 errors, 0 warnings, 4 checks, 43 lines checked
```
As currently imx477.c is totally clean of checkpatch warnings, it would be nice to fix all of those up, even though ```--strict``` is not strictly required (those on linux-media seem to expect it though).

"Signed-off By" is required as we are going to upstream this driver, so all contributions need to be appropriately tagged - see https://www.kernel.org/doc/html/latest/process/submitting-patches.html#sign-your-work-the-developer-s-certificate-of-origin. Please note "sorry, no pseudonyms or anonymous contributions".

@naushir We were talking about this change being needed earlier in the week - one fewer job for you then.

naushir:
Once the warning are cleared, LGTM.

drivers/media/i2c/imx477.c
```diff
@@ -25,6 +25,10 @@ static int dpc_enable = 1;
 module_param(dpc_enable, int, 0644);
 MODULE_PARM_DESC(dpc_enable, "Enable on-sensor DPC");
 
+static int trigger_mode = -1;
```
6by9: It was fine using a value of 0, but a static variable is automatically zeroed and so didn't need it explicitly.

```static int trigger_mode;```
would have been fine.

neocortex-vision:
Thanks, done. Regarding initialisation, default (ineffective) value is now -1.

6by9 (commented):
Thanks for the update.

Could you also squash your commits together. We apply them exactly as presented in the PR, so don't need your development history.

pelwell:
If the aim is to squash everything into one commit, especially if the commit message on the first commit doesn't require modification, then we can squash during the merge, but for more complicated combinations it's necessary for the submitter to squash before pushing.

6by9:
Upstream may point out that
```if (trigger_mode != 0)```
can be simplified to 
```if (trigger_mode)```
but otherwise that looks fine to me. Thanks for your efforts.

neocortex-vision:
Thanks for the feedback and excuse the explicit initialisation/checking (it's habitual, coming from a MISRA background).
I was planning to add hsync, vsync and frstobe signal parameters but I see this may have already been planned @naushir.

PR #4657, opened 2021-10-28
Add DT-based WLAN firmware location, useful for Zero 2 W

pelwell:
There are two versions of 43436 which should be given different firmwares, and one of which can't readily be distinguished from 43438. Also, the 5.15 kernel adds support for one of the 43436 parts but gives the firmware a different name. This PR addresses all three issues by adding a way to provide board-specific firmware name overrides in Device Tree.

In the event that the load with the overrides fails, there is a fallback to the original map without the overrides.

Note that kernel 5.15 supports the use of board-type-qualified firmware file names, but:
1. that might lead to duplicate firmwares with different names (although symbolic links would help with this), and
2. it doesn't help when we are trying to use firmware files that have already been released with unqualified names.


popcornmix:
Fine by me.

pelwell:
No - I'm seeing kernel splats from brcmfmac on some platforms. Reverting the code changes while I sort it out.

pelwell:
Duh - the `fwnames` pointer is only conditionally initialised, where it should have been initialised to NULL at the point of declaration. The lack of compiler warning is surprising, but no excuse.

Now fixed.

PR #4661, opened 2021-10-31
bcm2835-v4l2-codec: Remove advertised support of VP8

popcornmix:
The support for this format by firmware is very limited
and won't be faster than the arm.


popcornmix:
See: https://forum.kodi.tv/showthread.php?tid=364660

When I tested I didn't get the crash, but video didn't play and dmesg contained:
```
[  397.854503] bcm2835-codec bcm2835-codec: bcm2835_codec_start_streaming: Failed enabling i/p port, ret -3
```

6by9:
> won't be faster than the arm.

Not true on a Pi0/1.

If really unsupported then it'd be better to disable (remove?) in the firmware than in V4L2. The kernel driver checks the formats that the firmware advertises and converts those into the V4L2 types.

popcornmix:
It could be removed firmware side. It's not usable.

6by9:
So it turns out that the VP6 and VP8 actual codec code hasn't been shipped since Jan 2013, so advertising it is a total waste of time.
I'm happy with this PR, and we'll kill it from the firmware too.

PR #4668, opened 2021-11-04
configs: add LEDS_PWM=y

mhei:
Enable support for PWM driven LEDs.

Signed-off-by: Michael Heimpold <mhei@heimpold.de>


popcornmix:
Any reason not to make this a module? (CONFIG_LEDS_PWM=m)

mhei:
Availability during boot process for speed reasons: when you use it on a HAT for example as a status led, then it takes a little bit longer until modules are loaded.

pelwell:
A PWM-driven status LED is a bit of a luxury. If you are wanting to use one of the onboard PWM channels then a regular GPIO-driven LED is available using the same GPIOs. An external (e.g. I2C) PWM controller won't have a built-in driver, so there is probably no advantage to having leds_pwm built-in unless you're trying to use fancy non-binary brightness levels.

Having said that, it only adds 1.1kB to the kernel:
```
led_pwm_set a8
led_pwm_probe 324
led_pwm_driver_init 28
led_pwm_driver_exit 1c
__initcall_led_pwm_driver_init6 4
led_pwm_driver 68
```

mhei:
> A PWM-driven status LED is a bit of a luxury.

Definitely yes :smile:

> If you are wanting to use one of the onboard PWM channels then a regular
> GPIO-driven LED is available using the same GPIOs.

Yes, but also in case of onboard, you have to model it as
GPIO-driven LED in DT to have it controllable early and
later you cannot switch to PWM-driven. Or do I miss something?

> you're trying to use fancy non-binary brightness levels.

During early boot phase... no - that's correct.
Here we really use "traditional" blinking mode.
Later then, increasing/decreasing brightness is used
to signal some states, that's the reason for PWM.

mhei:
How to proceed here? Shall I adjust to `CONFIG_LEDS_PWM=m`, or is the small increase in size acceptable?

pelwell:
I've spent long enough thinking of objections and alternatives that I've run out of excuses - leave it with me.

PR #4672, opened 2021-11-05
Fixes for modetest helpers for 5.15

mripard:
Hi,

Here's a series of a few issues in the KMS commit code that were introduced by moving to generic helpers in (mainline) 5.12.

I'll submit a similar PR for 5.13 at least, since it's what Canonical is using, and can submit it to any other branch if relevant. 5.10 doesn't suffer from those issues since the commits were applied, and then reverted.

Maxime


pelwell:
Thanks!

PR #4673, opened 2021-11-05
Fixes for modetest helpers for 5.13

mripard:
Hi,

This is the same PR than #4672, for 5.13

### DIFF
--- a/drivers/gpu/drm/vc4/vc4_crtc.c
+++ b/drivers/gpu/drm/vc4/vc4_crtc.c
@@ -32,6 +32,7 @@
 #include <linux/clk.h>
 #include <linux/component.h>
 #include <linux/of_device.h>
+#include <linux/pm_runtime.h>
 
 #include <drm/drm_atomic.h>
 #include <drm/drm_atomic_helper.h>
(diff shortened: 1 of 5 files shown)


pelwell:
As of September 2021, 5.13 is EOL upstream - we've stopped updating rpi-5.13.y.

mripard:
Ok, that was one of the thing I wasn't really sure :) This is the kernel used by Canonical for Ubuntu 21.10, what is our stand here? Use either 5.10 or 5.15?

pelwell:
That would be our preference. I'm not completely against merging this and #4674, but that would leave the kernel lacking several commits that have gone into rpi-5.10.y and rpi-5.15.y - _caveat emptor_.

popcornmix:
https://en.wikipedia.org/wiki/Ubuntu_version_history#Table_of_versions
Ubuntu use a number of EOL kernels. They maintain their own kernel tree and cherry-pick security (and other critical bugs).
I suspect they prefer the kernel to match an Ubuntu release on all platforms, so are unlikely to switch the Pi kernel to match what we are using.

I'm happy to merge fixes here, which will make it easier for Ubuntu to cherry-pick.
But as you say, this is a largely untested tree, so it patches may be useful, but no promises.

juergh:
Thanks guys, I really appreciate this. Untested is fine but it does help me understand better what we might be missing.

PR #4684, opened 2021-11-09
config: Set TCA8418 to module

arturo182:
This chip is a popular way to add a matrix keyboard using I2C.

Would love to have it enabled as a module in the stock image :)


popcornmix:
If accepted we'd want the two arm64 defconfigs to be updated too.

pelwell:
Yes - it's a modest module (~10kB), so no objections on size grounds. Add the same setting to arch/arm64/configs/bcm2711_defconfig and .../bcmrpi3_defconfig and I we'll merge it.

arturo182:
Done and done!

PR #4689, opened 2021-11-11
overlays: vc4-kms-v3d: Change composite handling

pelwell:
On a Pi 4, enabling composite video disables the HDMI output. As a
consequence, the composite output is disabled by default. Change the
vc4-kms-v3d overlay used on older Pis to also disable composite by
default, replacing the "nocomposite" parameter with a "composite"
parameter.

### DIFF
--- a/arch/arm/boot/dts/overlays/README
+++ b/arch/arm/boot/dts/overlays/README
@@ -3620,8 +3620,8 @@ Params: cma-512                 CMA is 512MB (needs 1GB)
         cma-default             Use upstream's default value
         audio                   Enable or disable audio over HDMI (default "on")
         noaudio                 Disable all HDMI audio (default "off")
-        nocomposite             Disable the composite video output (default
-                                "off")
+        composite               Enable the composite output (default "off")
+                                N.B. Disables all other outputs on a Pi 4.
 
 
 Name:   vc4-kms-v3d-pi4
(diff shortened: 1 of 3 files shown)


arch/arm/boot/dts/overlays/README
```diff
@@ -3620,8 +3620,8 @@ Params: cma-512                 CMA is 512MB (needs 1GB)
         cma-default             Use upstream's default value
         audio                   Enable or disable audio over HDMI (default "on")
         noaudio                 Disable all HDMI audio (default "off")
-        nocomposite             Disable the composite video output (default
-                                "off")
+        composite               Enable the composite output (default "off")
+                                N.B. Disables all other outputs on a Pi 4.
```
6by9, line 3624: I'll test, but now that we have the correct VEC 108MHz under all circumstances there's probably no need to disable HDMI on Pi4 with composite.

pelwell: Which would make the only reason to change the default the behaviour of the desktop environment...

popcornmix: Doesn't having composite enabled, mean we are outputting a desktop/console picture there, using sdram bandwidth and power. For something 99% of users don't want?

6by9: Just been chasing this through. Composite-1 reports as "unknown", which DRM takes as "use me as a display of last resort". So normally the console will NOT be using it.

X does seem to enable the output regardless of reported connected state.

pelwell: It didn't stop us in Pi 3 days, when there was less SDRAM bandwidth to go around.

pelwell:
On balance, although it may upset the millions of legacy Pi users using KMS on a composite monitor, making composite output an opt-in is the right outcome.

pelwell:
This patch has been reverted temporarily because not enabling the composite output has been seen to break HDMI on older Pis.

PR #4702, opened 2021-11-16
configs : Add CONFIG_BT_HCIBFUSB=m

RuffaloLavoisier:
Include the HCI BlueFRITZ! USB driver.

support for Bluetooth USB devices with AVM.


pelwell:
This setting is already present - look two lines down from your additions - and the module is included in current releases: https://github.com/raspberrypi/rpi-firmware/blob/master/modules/5.10.78%2B/kernel/drivers/bluetooth/bfusb.ko

From what I can tell we've been building this module (on Pi 1 at least) since 2012, so it should be in your current image.

RuffaloLavoisier:
@pelwell Why did i haven't seen it ???...My mistake , Thanks

PR #4703, opened 2021-11-16
drm/vc4: Don't try disabling SCDC on Pi0-3.

6by9:
The code that set the scdc_enabled flag to ensure it was
disabled at boot time also ran on Pi0-3 where there is no
SCDC support. This lead to a warning in vc4_hdmi_encoder_post_crtc_disable
due to vc4_hdmi_disable_scrambling being called and trying to
read (and write) register HDMI_SCRAMBLER_CTL which doesn't
exist on those platforms.

Only set the flag should the interface be configured to support
more than HDMI 1.4.


6by9:
@pelwell Another quick one that would be nice to squeeze into a release to avoid a kernel WARN.

Issue #4704, opened 2021-11-16
H.264 flicker when using the V4L2 M2M encoder


mdevaev:
**Describe the bug**
I'm researching a new m2m interface to migrate our [streaming software](https://github.com/pikvm/ustreamer) to it and found that the parameters that the encoder uses produce a flickering video for a variable bitrate. When using a constant bitrate, everything is fine, but I need a variable one because it allows me to achieve zero-latency encoding. There were two very important parameters in the old MAIL interface: MIN_QUANT and MAX_QUANT, by configuring which I could get rid of the mentioned artifact. But if the quantization from 16 to 32 is configured, then everything will be fine.

Unfortunately, the new API does not provide this and without them I get the following result 
(Look at the area under the cursor in the terminal):
https://user-images.githubusercontent.com/416259/142009115-0f65da9d-306e-439f-a503-519cd56666b8.mp4

And how it should be (MIN_QUANT=16, MAX_QUANT=32):
https://user-images.githubusercontent.com/416259/142009452-bdcb8587-d94f-4f98-aed9-cf61d38d69c5.mp4

**To reproduce**
Use variable bitrate to encode H.264. In a contrasting black-and-white image, you will notice pulsating artifacts when an I-frame arrives. The full list of parameters is as follows:
```
V4L2_CID_MPEG_VIDEO_BITRATE = 5000000
V4L2_CID_MPEG_VIDEO_H264_I_PERIOD = 30
V4L2_MPEG_VIDEO_H264_PROFILE_CONSTRAINED_BASELINE
V4L2_MPEG_VIDEO_H264_LEVEL_4_0
V4L2_CID_MPEG_VIDEO_REPEAT_SEQ_HEADER = 1
output FPS = 1/30
```

**How to solve**
To solve the problem, we just need the ability to throw `MMAL_PARAMETER_VIDEO_ENCODE_MIN_QUANT/MAX_QUANT` into the kernel using V4L2_CID. The problem is that V4L2 does not seem to provide suitable CIDs for this. I suggest making a pair of specific CIDs with a large offset like this:
```patch
diff --git a/include/media/h264-ctrls.h b/include/media/h264-ctrls.h
index ec4799154..d1b045fde 100644
--- a/include/media/h264-ctrls.h
+++ b/include/media/h264-ctrls.h
@@ -37,6 +37,8 @@
[...]

PR #4705, opened 2021-11-17
Pass V4L2_CID_MPEG_VIDEO_H264_MIN_QP/MAX_QP to bcm2835-v4l2-codec

mdevaev:
Following raspberrypi/linux#4704. This is necessary to set up quantization for variable bitrate to avoid video flickering.

### DIFF
--- a/drivers/staging/vc04_services/bcm2835-codec/bcm2835-v4l2-codec.c
+++ b/drivers/staging/vc04_services/bcm2835-codec/bcm2835-v4l2-codec.c
@@ -2187,6 +2187,28 @@ static int bcm2835_codec_s_ctrl(struct v4l2_ctrl *ctrl)
 		ret = bcm2835_codec_set_level_profile(ctx, ctrl);
 		break;
 
+	case V4L2_CID_MPEG_VIDEO_H264_MIN_QP:
+		if (!ctx->component)
+			break;
+
+		ret = vchiq_mmal_port_parameter_set(ctx->dev->instance,
+						    &ctx->component->output[0],
+						    MMAL_PARAMETER_VIDEO_ENCODE_MIN_QUANT,
+						    &ctrl->val,
+						    sizeof(ctrl->val));
+		break;
+
+	case V4L2_CID_MPEG_VIDEO_H264_MAX_QP:
+		if (!ctx->component)
+			break;
+
+		ret = vchiq_mmal_port_parameter_set(ctx->dev->instance,
+						    &ctx->component->output[0],
+						    MMAL_PARAMETER_VIDEO_ENCODE_MAX_QUANT,
+						    &ctrl->val,
+						    sizeof(ctrl->val));
+		break;
+
 	case V4L2_CID_MPEG_VIDEO_FORCE_KEY_FRAME: {
 		u32 mmal_bool = 1;
 
@@ -3111,7 +3133,7 @@ static int bcm2835_codec_open(struct file *file)
 	case ENCODE:
 	{
 		/* Encode controls */
-		v4l2_ctrl_handler_init(hdl, 9);
+		v4l2_ctrl_handler_init(hdl, 11);
 
 		v4l2_ctrl_new_std_menu(hdl, &bcm2835_codec_ctrl_ops,
 				       V4L2_CID_MPEG_VIDEO_BITRATE_MODE,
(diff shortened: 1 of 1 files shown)


drivers/staging/vc04_services/bcm2835-codec/bcm2835-v4l2-codec.c
```diff
@@ -3159,6 +3181,14 @@ static int bcm2835_codec_open(struct file *file)
 					 BIT(V4L2_MPEG_VIDEO_H264_PROFILE_MAIN) |
 					 BIT(V4L2_MPEG_VIDEO_H264_PROFILE_HIGH)),
 					V4L2_MPEG_VIDEO_H264_PROFILE_HIGH);
+		v4l2_ctrl_new_std(hdl, &bcm2835_codec_ctrl_ops,
+				  V4L2_CID_MPEG_VIDEO_H264_MIN_QP,
+				  0, 0x7FFFFFFF,
```
6by9: Max value for QP is 51.

drivers/staging/vc04_services/bcm2835-codec/bcm2835-v4l2-codec.c
```diff
@@ -3159,6 +3181,14 @@ static int bcm2835_codec_open(struct file *file)
...
+				  1, 0);
+		v4l2_ctrl_new_std(hdl, &bcm2835_codec_ctrl_ops,
+				  V4L2_CID_MPEG_VIDEO_H264_MAX_QP,
+				  0, 0x7FFFFFFF,
+				  1, 0);
```
6by9: Max value is 51.
Shouldn't the default (last parameter) be 51 as well to allow rate control to use the full range? The firmware does special cases 0, so it will assume the defaults anyway, but setting 51 should also work.

Firmware defaults are min of 20, max 51, so those would be reasonable values to adopt.

6by9 (changes requested):

drivers/staging/vc04_services/bcm2835-codec/bcm2835-v4l2-codec.c
mdevaev: Got it

drivers/staging/vc04_services/bcm2835-codec/bcm2835-v4l2-codec.c
mdevaev: I changed the maximum and default values, if that's what you meant and I got it right for you.

6by9 (approved):

6by9:
Thanks, that update was exactly what I was meaning.

I would like to give this a quick test, but fundamentally it looks good now.

Potentially a QP of 0 is going to be ignored as it is the default "not set" condition in the firmware, but I'm going to ignore that situation, and we'll change the min to being 1 if it becomes a problem.

mdevaev:
Cool. I did a backport of this patch (with the previous default values) to 5.10.52 and it works great.

I would also like to add a separate patch to the ability to pass `V4L2_CID_MPEG_VIDEO_BITRATE_PEAK` as `MMAL_PARAMETER_VIDEO_ENCODE_PEAK_RATE` and maybe some others to keep the same behavior with my old implementation on MMAL.

There are three other parameters that were used in my implementation, but I took them from someone else's example and do not fully understand their meaning. These parameters are not set in the V4L2 interface. Are they needed? Because everything seems to work without them, but I have to ask to be sure.

```
MMAL_PARAMETER_NALUNITFORMAT = MMAL_VIDEO_NALUNITFORMAT_STARTCODES
MMAL_PARAMETER_MB_ROWS_PER_SLICE = 0
MMAL_PARAMETER_VIDEO_DROPPABLE_PFRAMES = MMAL_FALSE
```

**UPD**: Presumably, these are all default values in MMAL

6by9:
@pelwell I'm good with these. Any comments from you on this PR?

@mdevaev Generally please avoid adding extra discussion to a PR, but as a quick answer.

MMAL_PARAMETER_NALUNITFORMAT = MMAL_VIDEO_NALUNITFORMAT_STARTCODES
default. Memory says only MMAL_VIDEO_NALUNITFORMAT_FOURBYTEINTERLEAVELENGTH is the other format supported, which substitutes the start code for the length of the NAL (32bit value)

MMAL_PARAMETER_MB_ROWS_PER_SLICE = 0
default. Allows splitting the frame into multiple slices which can reduce latency but at the expense of bitrate.

MMAL_PARAMETER_VIDEO_DROPPABLE_PFRAMES = MMAL_FALSE
default. Setting it means that only every other P frame will be used as a reference frame. If a frame isn't used as a reference, there is no obligation to decode it, so it can be totally discarded without decoding if you want to fast forward in a stream.

pelwell:
Looks OK to me.

mdevaev:
@6by9 sorry and thanks for the answers

PR #4709, opened 2021-11-18
drivers: bcm2835_isp: Allow multiple users for the ISP driver.

naushir:
Add a second (identical) set of device nodes to allow concurrent use of the ISP
hardware by another user. This change effectively creates a second state
structure (struct bcm2835_isp_dev) to maintain independent state for the second
user. Node and media entity names are appened with the instance index
appropriately.

Further users can be added by changing the BCM2835_ISP_NUM_INSTANCES define.

### DIFF
--- a/drivers/staging/vc04_services/bcm2835-isp/bcm2835-v4l2-isp.c
+++ b/drivers/staging/vc04_services/bcm2835-isp/bcm2835-v4l2-isp.c
@@ -26,13 +26,19 @@
 #include "bcm2835-isp-ctrls.h"
 #include "bcm2835-isp-fmts.h"
 
+/*
+ * We want to instantiate 2 independent instances allowing 2 simultaneous users
+ * of the ISP hardware.
+ */
+#define BCM2835_ISP_NUM_INSTANCES 2
+
 static unsigned int debug;
 module_param(debug, uint, 0644);
 MODULE_PARM_DESC(debug, "activates debug info");
 
-static unsigned int video_nr = 13;
-module_param(video_nr, uint, 0644);
-MODULE_PARM_DESC(video_nr, "base video device number");
+static unsigned int video_nr[BCM2835_ISP_NUM_INSTANCES] = { 13, 20 };
+module_param_array(video_nr, uint, NULL, 0644);
+MODULE_PARM_DESC(video_nr, "base video device numbers");
 
 #define BCM2835_ISP_NAME "bcm2835-isp"
 #define BCM2835_ISP_ENTITY_NAME_LEN 32
@@ -1279,6 +1285,7 @@ static int bcm2835_isp_get_supported_fmts(struct bcm2835_isp_node *node)
  * or output nodes.
  */
 static int register_node(struct bcm2835_isp_dev *dev,
+			 unsigned int instance,
 			 struct bcm2835_isp_node *node,
 			 int index)
 {
@@ -1439,7 +1446,7 @@ static int register_node(struct bcm2835_isp_dev *dev,
 	snprintf(vfd->name, sizeof(node->vfd.name), "%s-%s%d", BCM2835_ISP_NAME,
 		 node->name, node->id);
 
-	ret = video_register_device(vfd, VFL_TYPE_VIDEO, video_nr + index);
+	ret = video_register_device(vfd, VFL_TYPE_VIDEO, video_nr[instance]);
 	if (ret) {
 		v4l2_err(&dev->v4l2_dev,
 			 "Failed to register video %s[%d] device node\n",
@@ -1660,9 +1667,8 @@ done:
 	return ret;
 }
 
-static int bcm2835_isp_remove(struct platform_device *pdev)
+static void bcm2835_isp_remove_instance(struct bcm2835_isp_dev *dev)
 {
-	struct bcm2835_isp_dev *dev = platform_get_drvdata(pdev);
 	unsigned int i;
 
 	media_controller_unregister(dev);
@@ -1677,11 +1683,11 @@ static int bcm2835_isp_remove(struct platform_device *pdev)
 					      dev->component);
 
 	vchiq_mmal_finalise(dev->mmal_instance);
-
-	return 0;
 }
 
-static int bcm2835_isp_probe(struct platform_device *pdev)
+static int bcm2835_isp_probe_instance(struct platform_device *pdev,
+				      struct bcm2835_isp_dev **dev_int,
+				      unsigned int instance)
 {
 	struct bcm2835_isp_dev *dev;
 	unsigned int i;
@@ -1691,6 +1697,7 @@ static int bcm2835_isp_probe(struct platform_device *pdev)
 	if (!dev)
 		return -ENOMEM;
 
+	*dev_int = dev;
 	dev->dev = &pdev->dev;
 
 	ret = v4l2_device_register(&pdev->dev, &dev->v4l2_dev);
@@ -1708,7 +1715,7 @@ static int bcm2835_isp_probe(struct platform_device *pdev)
 	if (ret) {
 		v4l2_err(&dev->v4l2_dev,
 			 "%s: failed to create ril.isp component\n", __func__);
-		goto error;
+		return ret;
 	}
 
 	if (dev->component->inputs < BCM2835_ISP_NUM_OUTPUTS ||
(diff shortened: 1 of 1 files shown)


naushir:
@6by9 and @davidplowman 
Here are the changes need to allow concurrent usage for the ISP.

drivers/staging/vc04_services/bcm2835-isp/bcm2835-v4l2-isp.c
```diff
@@ -1436,10 +1444,12 @@ static int register_node(struct bcm2835_isp_dev *dev,
...
 	/* Define the device names */
-	snprintf(vfd->name, sizeof(node->vfd.name), "%s-%s%d", BCM2835_ISP_NAME,
-		 node->name, node->id);
+	snprintf(vfd->name, sizeof(node->vfd.name), "%s%d-%s%d",
+		 BCM2835_ISP_NAME, instance, node->name, node->id);
 
-	ret = video_register_device(vfd, VFL_TYPE_VIDEO, video_nr + index);
+	ret = video_register_device(vfd, VFL_TYPE_VIDEO,
+				    video_nr + index +
+				    (BCM2835_ISP_NUM_NODES * instance));
```
6by9: Please can we have an array of base values as the module parameter.

Deinterlace is already registering as 18, and HEVC decode as 19.

drivers/staging/vc04_services/bcm2835-isp/bcm2835-v4l2-isp.c
```diff
@@ -1677,20 +1688,21 @@ static int bcm2835_isp_remove(struct platform_device *pdev)
...
 	unsigned int i;
 	int ret;
 
-	dev = devm_kzalloc(&pdev->dev, sizeof(*dev), GFP_KERNEL);
-	if (!dev)
+	*dev_int = devm_kzalloc(&pdev->dev, sizeof(*dev), GFP_KERNEL);
+	if (!*dev_int)
 		return -ENOMEM;
 
+	dev = *dev_int;
```
6by9: Can we allocate and store as ```dev``` initially, and only assign to ```*dev_int``` as the last thing on success? Then we don't have a stale non-NULL pointer kicking around in the global state.

drivers/staging/vc04_services/bcm2835-isp/bcm2835-v4l2-isp.c
```diff
@@ -1720,25 +1732,51 @@ static int bcm2835_isp_probe(struct platform_device *pdev)
...
+static int bcm2835_isp_probe(struct platform_device *pdev)
+{
+	unsigned int i;
+	int ret;
+
+	for (i = 0; i < ARRAY_SIZE(bcm2835_isp_instance); i++) {
+		ret = bcm2835_isp_probe_instance(pdev, &bcm2835_isp_instance[i],
+						 i);
+		if (ret)
+			goto error;
```
6by9, line 1781: If it fails with i > 0, then we need to iteratively call bcm2835_isp_remove_instance on those that were created successfully. If we haven't updated ```bcm2835_isp_instance[i]```, then calling ```bcm2835_isp_remove``` will do that for us.

drivers/staging/vc04_services/bcm2835-isp/bcm2835-v4l2-isp.c
```diff
@@ -133,6 +133,13 @@ struct bcm2835_isp_dev {
 	int b_gain;
 };
 
+/*
+ * We want to instantiate 2 independent instances allowing 2 simultaneous users
+ * of the ISP hardware.
+ */
+#define BCM2835_ISP_NUM_INSTANCES 2
+struct bcm2835_isp_dev *bcm2835_isp_instance[BCM2835_ISP_NUM_INSTANCES];
```
6by9: Can we devm_kzalloc this in ```bcm2835_isp_probe```, and store that parent pointer via ```platform_set_drvdata(pdev, dev);```?

I wouldn't expect to get two instances of this driver probed, but currently they'll fight over this structure. In a very ugly manner.

6by9 (commented):
A few minor comments mainly around error handling and cleanup.

drivers/staging/vc04_services/bcm2835-isp/bcm2835-v4l2-isp.c
naushir: Ack.

drivers/staging/vc04_services/bcm2835-isp/bcm2835-v4l2-isp.c
naushir: Ack.

drivers/staging/vc04_services/bcm2835-isp/bcm2835-v4l2-isp.c
naushir: Ack

drivers/staging/vc04_services/bcm2835-isp/bcm2835-v4l2-isp.c
```diff
@@ -1627,7 +1636,7 @@ static int media_controller_register(struct bcm2835_isp_dev *dev)
 		ret = -ENOMEM;
 		goto done;
 	}
-	snprintf(name, BCM2835_ISP_ENTITY_NAME_LEN, "bcm2835_isp0");
+	snprintf(name, BCM2835_ISP_ENTITY_NAME_LEN, BCM2835_ISP_NAME);
```
6by9: Is Libcamera happy with 2 media entities having the same name?

drivers/staging/vc04_services/bcm2835-isp/bcm2835-v4l2-isp.c
```diff
@@ -1708,7 +1716,7 @@ static int bcm2835_isp_probe(struct platform_device *pdev)
 	if (ret) {
 		v4l2_err(&dev->v4l2_dev,
 			 "%s: failed to create ril.isp component\n", __func__);
-		goto error;
+		return ret;
```
6by9, line 1718: Sorry my bad in reviewing your last version.

```goto error``` meant that we cleaned up the instance. If you had updated ```*dev_int```, then ```bcm2835_isp_remove``` will do it for you, but I asked you to change that.

drivers/staging/vc04_services/bcm2835-isp/bcm2835-v4l2-isp.c
naushir: Yes, it is fine as the matching occurs on the node entity names.

naushir:
Small update to the node/entity names - we do not need to index over multiple instances any more.

naushir:
@6by9 and @pelwell this should be ready to review and merge now.

pelwell:
It feels wrong that it is necessary to create separate instances like this to support concurrency, but I don't know the intricacies of the V4L2 world and you wouldn't be proposing this if it wasn't needed.

This implementation is much cleaner than the first version, so LGTM.

naushir:
> It feels wrong that it is necessary to create separate instances like this to support concurrency, but I don't know the intricacies of the V4L2 world and you wouldn't be proposing this if it wasn't needed.
> 

Unfortunately that's how V4L2 works with concurrency, we cannot avoid duplicating the device nodes.

6by9:
> It feels wrong that it is necessary to create separate instances like this to support concurrency, but I don't know the intricacies of the V4L2 world and you wouldn't be proposing this if it wasn't needed.

Annoyingly a single V4L2 node can only support one queue of each type (OUTPUT, CAPTURE, etc).
A simple memory to memory device such as a codec (1 input, 1 output) can be exposed as a single node, and each time you open the device you get a new instance.
With the ISP where we have 1 input but 3 output (high res, low res, and stats) we can't use a single node, and if split over multiple nodes then we can't support concurrent usage as you can't tell which instance each request or buffer is associated with.

So yes, ugly as heck, but no way around it.

drivers/staging/vc04_services/bcm2835-isp/bcm2835-v4l2-isp.c
6by9: Unresolving as we will still leak the instance.

Either we do set ```*dev_int``` at the start so that ```bcm2835_isp_remove``` cleans up, or we need an error handler that calls ```bcm2835_isp_remove_instance```. The first is probably the easiest.

naushir: Oops, I thought I did that!
Will set *dev_int on the top as it was in the first revision.

6by9:
Ta, LGTM now.

PR #4710, opened 2021-11-18
Enable interlaced video mode

rtomasa:
As of today, Raspberry Pi4 + KMS allows to use interlaced modes via DPI, so suggesting to remove the restriction from the vc4_dpi.c source file.


6by9:
Thank you for the PR.

What standard should interlaced DPI follow? NTSC Odd field first or PAL even field first?

As it stands the CRTC will follow the PAL paradigm of even field first, and adding the extra line to the vertical back porch of the odd field cf the even field.

Does that fit with absolutely all interlaced displays that the DPI peripheral could ever drive?
Are there no cases where NTSC / PAL-M's odd field first and increasing the even field front porch by one line cf the odd field?

Just enabling interlaced modes without even considering the further repercussions is asking for issues later.

You've also included your notes and gitignore files. We don't want those in the PR thanks.

rtomasa:
Correct me if I'm wrong but the different interlaced modes are defined by the modeline itself. I mean, if the mode is provided by the EDID then no prob; if the mode is a custom one provided by you, then it's up to you to define the proper parameters so no complain either.

As per my understanding, removing the interlaced flag constrain does not affect to anybody currently using only progressive modes. Also do note that there are many types of interlaced modes, not only NTSC/PAL standards. I use github.com/antonioginer/switchres to generate many different timings (including interlaced ones) for many types of monitors (15Khz, 25Khz, 31Khz...) many of them not following any standard and all of them working fine if you work out the correct parameters for the correct monitor. This is an unique feature of the Pi4 over any PC with any educated graphic card.

I have no problem on maintaining my own kernel patched for my own purposes but I thought that many other people would like to benefit from this feature out of the box.

BTW I've removed the 2 files that I'd committed by mistake.

Regards

6by9:
Your modeline is effectively defining a  ```struct drm_display_mode``` - see https://elixir.bootlin.com/linux/latest/source/include/drm/drm_modes.h#L227
That defines the frame only, not the fields. There is no information on how the frame is split between the fields.

eg https://elixir.bootlin.com/linux/latest/source/drivers/gpu/drm/meson/meson_venc_cvbs.c#L44 or https://elixir.bootlin.com/linux/latest/source/drivers/gpu/drm/vc4/vc4_vec.c#L257 defining the various SD resolutions. 
Meson then uses https://elixir.bootlin.com/linux/latest/source/drivers/gpu/drm/meson/meson_venc.c#L69 to look up how to split the frame between the fields, or https://github.com/raspberrypi/linux/blob/rpi-5.10.y/drivers/gpu/drm/vc4/vc4_crtc.c#L378 for vc4.

All interlacing options I've encountered follow either the NTSC or PAL standards for how to split the lines between the two fields.
AIUI HDMI modes all follow PAL, but I have no information as to whether all DPI displays will do the same.
Yes we could just merge this and assume they're all equivalent to PAL, but it would be nice to know.

rtomasa:
Thanks for the detailed explanation and examples. I cannot guarantee that all modes work as PAL, so I can understand you fearing that some unexpected issue can arise from this pull request. Proceeding to close the same.

Issue #3988, opened 2020-12-03
Some V4L2 settings are missing/broken compared to the available MMAL settings


dgsiegel:
**Describe the bug**

I would expect that all raspistill (which uses MMAL directly) would work for the V4L2 interface as well: https://www.raspberrypi.org/documentation/raspbian/applications/camera.md

**To reproduce**
Compare https://www.raspberrypi.org/documentation/raspbian/applications/camera.md with `v4l2-ctl -L`

**Expected behaviour**
Ideally all MMAL settings would be exposed to V4L2 as well

**Actual behaviour**

Compared to MMAL/raspistill/raspivid, v4l2-ctl is missing the following settings:

**Exposure mode (`scene_mode`)**

Only 3 values available:
-  0: None
-  8: Night
-  11: Sports 

Whereas raspistill offers:

- auto: use automatic exposure mode
- night: select setting for night shooting
- nightpreview:
- backlight: select setting for backlit subject
- spotlight:
- sports: select setting for sports (fast shutter etc.)
- snow: select setting optimised for snowy scenery
- beach: select setting optimised for beach
- verylong: select setting for long exposures
- fixedfps: constrain fps to a fixed value
- antishake: antishake mode
- fireworks: select setting optimised for fireworks

**Color effects (`color_effects`)**

The following effects don't work:

- 4: Emboss 
- 5: Sketch
- 6: Sky Blue (pastel)
- 7: Grass Green (watercolour)
- 9: Vivid (saturation)
- 11: Art Freeze (hatch)
- 12: Silhouette 

The following effects are missing:

- posterise: posterise the image
- whiteboard: whiteboard effect
- blackboard: blackboard effect
- denoise: denoise the image
- hatch: hatch sketch effect
- gpen: graphite sketch effect
- watercolour: watercolour effect
- blur: blur the image

**Colour effect (`color_effects_cbcr`)**

Does not seem to have any effect 

**JPEG Quality (`compression_quality`)**

Does not have any effect on MJPEG stream

**Video stabilisation (`image_stabilization`)**

Does not seem to have any effect 

**Metering (`exposure_metering_mode`)**

Matrix metering is missing

**Exposure mode (`auto_exposure`)**

The following setting are missing:

- night: select setting for night shooting
- nightpreview:
- backlight: select setting for backlit subject
- spotlight:
- sports: select setting for sports (fast shutter etc.)
- snow: select setting optimised for snowy scenery
- beach: select setting optimised for beach
[...]

PR #4715, opened 2021-11-19
staging: bcm2835-camera: Add V4L2 control for sensor_mode

Malvineous:
I wonder if I could request this feature be merged into the Raspberry Pi V4L2 camera driver?  It replicates the functionality of raspivid's `-md` option, which allows for example higher quality videos if a reduced framerate is acceptable.

I am using this to set sensor_mode=2 when capturing at 1296x972 with a Pi V1 camera, as this allows the sensor's full resolution to be used to build the 1296x972 image.  Without this, due to the way the sensor works, the video quality is only effectively 640x480 resolution upscaled to 1296x972.  Since I am using this as a security camera, having finer detail visible in the image is worth the cost of the 15 fps limit introduced by this sensor mode.

The only thing I am unsure about with this patch is that I am borrowing the `V4L2_CID_USER_BCM2835_ISP_BASE` constant even though it is only tangentially related to this code.  You may prefer using a different constant for the custom control ID, so happy to discuss options there.

The new V4L2 control added by this code is used by running `v4l2-ctl --set-ctrl=sensor_mode=2` before starting a V4L2 capture.

### DIFF
--- a/drivers/staging/vc04_services/bcm2835-camera/bcm2835-camera.h
+++ b/drivers/staging/vc04_services/bcm2835-camera/bcm2835-camera.h
@@ -13,7 +13,7 @@
  * core driver device
  */
 
-#define V4L2_CTRL_COUNT 32 /* number of v4l controls */
+#define V4L2_CTRL_COUNT 33 /* number of v4l controls */
 
 enum {
 	COMP_CAMERA = 0,
--- a/drivers/staging/vc04_services/bcm2835-camera/controls.c
+++ b/drivers/staging/vc04_services/bcm2835-camera/controls.c
@@ -28,6 +28,9 @@
 #include "mmal-parameters.h"
 #include "bcm2835-camera.h"
 
+#define MMAL_V4L2_CID_CUSTOM_BASE	V4L2_CID_USER_BCM2835_ISP_BASE
+#define MMAL_V4L2_CID_SENSOR_MODE	(MMAL_V4L2_CID_CUSTOM_BASE + 0)
+
 /* The supported V4L2_CID_AUTO_EXPOSURE_BIAS values are from -4.0 to +4.0.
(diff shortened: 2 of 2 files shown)


6by9:
Creating new custom controls in V4L2 is a real maintenance nightmare. Hijacking one from V4L2_CID_USER_BCM2835_ISP_BASE is the least horrid approach, but even there we've tripped over #4440 because that define isn't in the mainline Linux codebase.

Using V4L2_CID_USER_BCM2835_ISP_BASE+0 I think means we have the same CID meaning different things depending on whether it is on the ISP or the camera stack. Userspace therefore can't rely on it (only v4l2-ctl is about the only thing that searches by control name).

Sorry, but I'm therefore very reluctant to accept this sort of patch into the driver.

Malvineous:
It looks like `V4L2_CID_USER_BCM2835_ISP_BASE` is intended to be merged into the mainline kernel, so if that happens, does that reduce the maintenance burden?  I don't mind waiting with this patch until then if it would make things easier.

The `+0` constant isn't used by the ISP stack (it starts defining constants from `+1`) so there is no conflict.  But if we were going to go down that path, I imagine we'd want to define a `+0` version in the ISP stack to effectively mark it as reserved so it never gets assigned a second use.

pelwell:
If @6by9 is reluctant to merge a patch like this, it's not going to happen.

Malvineous:
So that's it then?  No discussion, no alternative ways to get the functionality implemented so that it is acceptable?  Just "get lost we don't want your help"

JamesH65:
This patch is the wrong way of doing it as explained above. Better suggestions would be helpful. Probably best to do that in an issue rather than a PR I would think.

Malvineous:
No worries, I will continue the discussion in #3988 as it is asking for a superset of the functionality here.

PR #4721, opened 2021-11-20
configs: add CONFIG_HID_PLAYSTATION=m and CONFIG_PLAYSTATION_FF=y

Lahvuun:
Enables support for the PlayStation 5 'DualSense' controller and its
force feedback.

Also, Sony is planning to move PlayStation peripheral support from
hid-sony into hid-playstation in the future:
https://lore.kernel.org/all/20201219062336.72568-1-roderick@gaikai.com/

This means that, potentially, in future kernel releases older
PlayStation controllers may stop working without this option. As
Raspberry Pis are a popular choice as emulation consoles, it is probably
a good idea to enable this.

Signed-off-by: Ilya Trukhanov <lahvuun@gmail.com>


pelwell:
The new options would naturally appear between:
```
CONFIG_HID_PICOLCD=m
```
and:
```
CONFIG_HID_ROCCAT=m
```
Adding them in a different place (as you've done) works, but it means that when we use "make ... savedefconfig" in future there will be a spurious change.

Please update the PR, then I'll merge it.

Lahvuun:
Done.

Sorry, I should've thought about savedefconfig.

pelwell:
Thanks!

Lahvuun:
@pelwell I think this change never made it to the rpi-5.16.y branch, while other changes during the same time frame did. Did it get overlooked or will it get merged later? I'm not familiar with your branching policies.

pelwell:
This is the first 5.15 to 5.16 patch introduced since the initial rpi-5.16.y port, and it fell at the first hurdle because 5.16 brings in a new dependency - `LEDS_CLASS_MULTICOLOR`.

It's patched up and merged now as c94d71236bb9d.

Lahvuun:
> 5.16 brings in a new dependency - LEDS_CLASS_MULTICOLOR.

> It's patched up and merged now as c94d712.

Right. Thank you.

I tried `make ARCH=arm bcmrpi_defconfig` at c94d712 but CONFIG_HID_PLAYSTATION and CONFIG_PLAYSTATION_FF don't get enabled unless I also add `LEDS_CLASS_MULTICOLOR` to the defconfigs. Am I doing this wrong?

pelwell:
It's not you, just some unstaged changes - try 9299025206cb3.

Lahvuun:
Tested it with my Pi Zero W and it works perfectly. Thank you again!

PR #4724, opened 2021-11-21
[5.15] fix VIDIOC_DQEVENT compat32 handling

HiassofT:
Running 32bit ffmpeg on a 64bit kernel has issues with v4l2 decoders, as reported here https://forum.libreelec.tv/thread/24847-latest-build-rpi4-suffers-pausing-randomly-and-micro-stutter-video-playback/?postID=163050#post163050

The cause is a kernel bug in compat32 handling of the VIDIOC_DQEVENT ioctl which is already fixed in linux-next.

Pick the upstream fix until it dribbles down into stable


pelwell:
I was reworking to add the "commit ... upstream" line, but this appears to have been merged offline already. @popcornmix ?

popcornmix:
Ah yes - I cherry-picked it when @HiassofT mentioned this fixed an issue on 5.15.
I'd imagine this will appear upstream soon as it's a regression.

PR #4726, opened 2021-11-22
media: i2c: ov5647: Support HFLIP and VFLIP

davidplowman:
Add these missing V4L2 controls. Tested binned and full resolution
modes in all four orientations using Raspberry Pi running libcamera.

### DIFF
--- a/drivers/media/i2c/ov5647.c
+++ b/drivers/media/i2c/ov5647.c
@@ -64,6 +64,8 @@
 #define OV5647_REG_GAIN_LO		0x350B
 #define OV5647_REG_VTS_HI		0x380e
 #define OV5647_REG_VTS_LO		0x380f
+#define OV5647_REG_VFLIP		0x3820
+#define OV5647_REG_HFLIP		0x3821
 #define OV5647_REG_FRAME_OFF_NUMBER	0x4202
 #define OV5647_REG_MIPI_CTRL00		0x4800
 #define OV5647_REG_MIPI_CTRL14		0x4814
@@ -124,6 +126,8 @@ struct ov5647 {
 	struct v4l2_ctrl		*hblank;
 	struct v4l2_ctrl		*vblank;
 	struct v4l2_ctrl		*exposure;
+	struct v4l2_ctrl		*hflip;
+	struct v4l2_ctrl		*vflip;
 	bool				write_mode_regs;
 };
 
@@ -152,7 +156,7 @@ static struct regval_list ov5647_640x480_8bit[] = {
 	{0x3036, 0x46},
 	{0x303c, 0x11},
 	{0x3106, 0xf5},
-	{0x3821, 0x07},
+	{0x3821, 0x01},
 	{0x3820, 0x41},
 	{0x3827, 0xec},
 	{0x370c, 0x0f},
@@ -240,7 +244,7 @@ static struct regval_list ov5647_2592x1944_10bit[] = {
 	{0x3036, 0x69},
 	{0x303c, 0x11},
 	{0x3106, 0xf5},
-	{0x3821, 0x06},
+	{0x3821, 0x00},
 	{0x3820, 0x00},
 	{0x3827, 0xec},
 	{0x370c, 0x03},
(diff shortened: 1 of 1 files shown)


davidplowman:
@naushir @6by9 Well, this lot seems to fix the problems, but not sure if it's the "best" way...

drivers/media/i2c/ov5647.c
```diff
@@ -152,7 +156,7 @@ static struct regval_list ov5647_640x480_8bit[] = {
 	{0x3036, 0x46},
 	{0x303c, 0x11},
 	{0x3106, 0xf5},
-	{0x3821, 0x07},
+	{0x3821, 0x03},
```
6by9: I think my only comment is that we could drop bit 1 from these registers totally, as the control handler will always be called via __v4l2_ctrl_handler_setup and set it as defined by the control. That only leaves the binning bit defined.

davidplowman: My only thought was for the register set on its own to correspond to "no flips at all", which, bizarrely, involves setting the hflip bit to counteract the secret internal flip. But given that we don't document anything about the registers anyway - I don't suppose it makes any difference really!

davidplowman:
New version with those bit 1 values cleared - do we prefer this?

6by9:
I think I'm happy with that. It just seemed odd that we had bits in the registers that were never going to be used as set.

drivers/media/i2c/ov5647.c
```diff
@@ -1140,14 +1171,16 @@ static int ov5647_enum_frame_size(struct v4l2_subdev *sd,
 {
 	struct ov5647_mode *mode = NULL;
 
-	if (fse->code == MEDIA_BUS_FMT_SBGGR8_1X8) {
+	if (fse->code == ov5647_get_mbus_code(sd, 1)) {
 		if (fse->index >= ARRAY_SIZE(supported_modes_8bit))
 			return -EINVAL;
 		mode = &supported_modes_8bit[fse->index];
-	} else if (fse->code == MEDIA_BUS_FMT_SBGGR10_1X10) {
+		fse->code = ov5647_get_mbus_code(sd, 1);
```
pelwell: Is this not a no-op?
```
    if (a == func()) {
        a = func();
```
looks wrong.

davidplowman: Yes actually, that would seem to be pointless! Let me fix it up...

pelwell:
This patch does not apply cleanly to rpi-5.15.y - can someone more au-fait with V4L2 take a look?

davidplowman:
I had a quick look at this. It seems to me like the 5.15 version simply doesn't support 8 bit modes at all, whereas our 5.10 one has had a 640x480 8-bit mode for some time. I'm thinking therefore that the 5.15 version is basically old and needs to be completely overwritten by the new one, but would appreciate Dave's opinion on that.

6by9:
```
commit 38c22308181562d52926b6701327c3fe209f6d84
Author: Jacopo Mondi <jacopo@jmondi.org>
Date:   Thu Nov 19 17:37:53 2020 +0100

    media: ov5647: Remove 640x480 SBGGR8 mode
    
    Capturing in 640x480 SBGGR8_1X8 hangs the system when capturing
    with the unicam driver on RaspberryPi 4 platform.
    
    Remove it and remove the support for multiple media bus codes in
    the driver.
    
    Signed-off-by: Jacopo Mondi <jacopo@jmondi.org>
    Signed-off-by: Sakari Ailus <sakari.ailus@linux.intel.com>
    Signed-off-by: Mauro Carvalho Chehab <mchehab+huawei@kernel.org>
```
So upstream have dropped the 8bit modes. That should simplify things somewhat.

davidplowman:
I don't think I like that. Our users explicitly want to be able to get the 8-bit mode because it frames faster.

6by9:
Worth testing whether it actually works then as the commit text implies it doesn't.
IIRC it was also a mess in that it was the top left quarter of the image because it was binned and then cropped, not binned and skipped.

davidplowman:
Ah yes, I remember that. Maybe that was the broken mode. Perhaps I'm remembering a different sensor where they wanted a mode with a different bit-depth. Maybe the imx477. In which case this I guess this is OK.

PR #4140, opened 2021-02-12
WIP: dtoverlays: Update CSI2 sensor overlays to add a cam0 parameter

6by9:
Add a cam0 parameter to the CSI2 sensor overlays to switch
to the CAM0 connector on Compute Module IO boards.

Signed-off-by: Dave Stevenson <dave.stevenson@raspberrypi.com>

PR #4727, opened 2021-11-22
Add cam0 parameter to image sensor overlays

6by9:
Supersedes #4140.

Only imx477 and imx219 overlays have currently been updated until confirmation that the style is OK. (README has already been updated for most/all of them).
These apply correctly from the firmware, but not for cam0 if loaded dynamically (fails to fixup the clock, and appears to still be writing to i2c_csi_dsi, and cam1_clk).

The ordering of these fragments is critical as they affect each other. The gaps in the fragment numbering were deliberate to try and make this obvious. What's the preference over this?

OV5647 is also converted to use the regulator framework as otherwise we can't share the GPIO on CM4.


6by9:
Just to make notes of the conversation we've just had:
- The intent is to have camN_reg and camN_clk nodes enabled by default to avoid the issue in the regulator framework that a regulator can be unregistered whilst still having active clients. Those clients are then quite likely to cause a NULL deref if they do anything. Enable all these nodes in the base DT.
- CM1/3 are the only real platforms where all GPIOs are available. Disable the regulators again in the CM1/3 platform specific DT.
- Reinstate enabling the regulators again in the overlays so that they will work on CM1/3 without further changes.

arch/arm/boot/dts/overlays/ov5647-overlay.dts
```diff
@@ -94,6 +91,13 @@
 		};
 	};
 
+	fragment@7 {
```
pelwell: What is the purpose of this fragment?

6by9: This is patch 2. At this point OV5647 relies on using the GPIO as a shutdown line via the GPIO API. If the cam1_reg is enabled then it will claim the GPIO and the sensor will fail to probe - regression.

Patch 3 & 4 add support for the regulator framework to OV5647.

I can swap the patches around so that we convert OV5647 to use the regulators first, and then we won't have this interim part.

pelwell: It's not interim - I'm looking at the cumulative effect of the patches and it's still there, disabling `cam1_reg`.

6by9: Drat, I obviously missed removing it again.

pelwell: That's fine. As long as I know it is no longer needed I can add the deletion to my other patches.

pelwell:
I don't seem to be able to push to your branch, but see [pelwell/rpi-5.10.y-cam0 ](https://github.com/pelwell/linux/tree/rpi-5.10.y-cam0) for my patch that resolves everything except the Compute Module behaviour.

We could enable `cam1_reg` on Compute Modules by default, but also add a `cam0` parameter to the base DTBs to allow it to be disabled if users need the GPIO. Or we could invert the sense so that both the base DTB and the overlay need `cam0` specifying to make it work.

6by9:
Followup from our discussion:

Trying to load an overlay with ```cam0``` on a non-CM results in 
```
[    6.914475] unicam fe800000.csi: number of data lanes not set
[    6.914498] unicam fe800000.csi: Failed to connect subdevs
[    6.914955] unicam: probe of fe800000.csi failed with error -22
```
so we don't actually have to care about the state there as it will fail to load.

The reason for the comment above about CM1/3 is over the GPIO being reserved if the regulator is enabled. Seeing  as one of the cameras defaults to using GPIO2, that stops I2C1 working.
Added a ```cam0_reg``` and ```cam1_reg``` override for CM1/3 to set the status of the regulator, and a ```cam0_reg_gpio``` and ```cam1_reg_gpio``` override to allow configuration of the GPIO on those platforms. I haven't added it to CM4 as there is a nominally reserved GPIO to do that, and that's on the GPIO expander to make life more complex.

I'm partly ignoring dynamically loading the overlays, but we are going to have an issue with imx477_378-overlay.dtsi as it sets ```startup-delay-us``` on ```cam1_reg```, which isn't going to work if the regulator is already enabled. We can only do our best I think.

Branch updated.

ov5647 still throws a regulator wobbly on dynamic unload. imx290/327 throws a wobbler calling ```v4l2_ctrl_handler_free``` (fixed, although now logs ```OF: ERROR: memory leak, expected refcount 1 instead of 2, of_node_get()/of_node_put() unbalanced - destroy cset entry: attach overlay node /soc/i2c0mux/i2c@1/imx290@1a/port/endpoint```. Will hold back on pushing that patch to this PR).

The unload paths in many of these drivers look a little iffy, so perhaps we just merge this, and then clean up as dynamic unload issues present themselves?

arch/arm/boot/dts/overlays/tc358743-overlay.dts
```diff
@@ -106,7 +106,7 @@
 	__overrides__ {
 		4lane = <0>, "-2+3-7+8";
 		link-frequency = <&tc358743_0>,"link-frequencies#0";
-		media-controller = <&csi>,"brcm,media-controller?";
+		media-controller = <0>,"=9";
```
pelwell: I think the correct change here is to delete fragment 9, and even then that's only cosmetic.

arch/arm/boot/dts/overlays/ov9281-overlay.dts
```diff
@@ -48,6 +48,7 @@
 		target = <&csi1>;
 		csi: __overlay__ {
 			status = "okay";
+			brcm,media-controller;
```
pelwell, line 51: Good catch.

arch/arm/boot/dts/bcm2710-rpi-cm3.dts
```diff
@@ -139,10 +133,18 @@
 	pinctrl-0 = <&audio_pins>;
 };
 
+&cam1_reg {
```
pelwell: Duplicate - see above.

6by9:
I'm just going through the combinations.
Looking good with
- ov9281
- imx327 (imx290)
- ov5647
- ov7251
- imx219

Minor niggle with imx477 in that we want to set the regulator startup delay to 300ms, and ov5647 sets it to 20ms. If you apply imx477 first, then ov5647 overwrites the setting.

Just need to test tc358743, and ideally adv7282m, but then we're done.

I'll push my updated branch prior to squashing.

6by9:
I'd missed resolving your previous comments. Now addressed.

pelwell:
I'm happy with that. Squash when ready, remove the WIP and we can merge.

6by9:
Squashed. No diffs from the previous version other than the commit history.
About to remove WIP tag.

pelwell:
The regulator commit wasn't a straightforward merge for rpi-5.15.y but the end result looks plausible and builds without errors - would you mind casting an eye over it? 9f5ceabedf30

6by9:
I've not tested the 5.15 branch, but that patch looks correct.

PR #4728, opened 2021-11-22
Use `platform_set_drvdata` to save platform driver data in memory for BCM2711

buismaarten:
Many drivers use the `platform_set_drvdata` function to save platform driver data in memory to use later.


pelwell:
Yes they do, but unless the driver _does_ make use of it later this is literally _a waste of time and code_.

Did you notice how the private data pointer is passed to `devm_thermal_zone_of_sensor_register`, a pointer that later appears as a parameter to `bcm2711_get_temp`? This is common to all  `thermal_zone_of_device_ops` methods, and saves each of the methods in each of the client drivers from having to do the lookup.

PR #4731, opened 2021-11-25
media: i2c: ov5647: Support HFLIP and VFLIP

davidplowman:
Add these missing V4L2 controls. Tested binned and full resolution
modes in all four orientations using Raspberry Pi running libcamera.


davidplowman:
@naushir @6by9 This is the same h/vflip fix as we applied to 5.10 but with the conflicts resolved for 5.15. In fact, seeing as 5.15 does away with the 8 bit mode it's all rather simpler. Again, I've tested all 4 orientations running libcamera with this kernel.

6by9:
LGTM

Issue #4714, opened 2021-11-19
Running libcamera-hello leads to Division by zero in kernel


oniongarlic:
**Describe the bug**
Running libcamera-hello on a Model B with Camera V1 leads to Division by zero in kernel oops.

**To reproduce**
Run libcamera-hello 

**Expected behaviour**
No Division by zero in kernel

**Actual behaviour**
[   95.199368] Division by zero in kernel.

**System**
 Copy and paste the results of the raspinfo command in to this section. Alternatively, copy and paste a pastebin link, or add answers to the following questions:

* Which model of Raspberry Pi? Vintage Model B
* Which OS and version (`cat /etc/rpi-issue`)? 11
* Which firmware version (`vcgencmd version`)?
Oct 29 2021 10:49:08 
Copyright (c) 2012 Broadcom
version b8a114e5a9877e91ca8f26d1a5ce904b2ad3cf13 (clean) (release) (start)

* Which kernel version (`uname -a`)?
Linux pifhdcamera 5.10.63+ # 1459 Wed Oct 6 16:40:27 BST 2021 armv6l GNU/Linux

**Logs**
```
[   95.199368] Division by zero in kernel.
[   95.199425] CPU: 0 PID: 494 Comm: libcamera-hello Tainted: G         C        5.10.63+ #1459
[   95.199442] Hardware name: BCM2835
[   95.199455] Backtrace: 
[   95.199542] [<c08746c0>] (dump_backtrace) from [<c0874a40>] (show_stack+0x20/0x24)
[   95.199571]  r7:c20b8020 r6:bf3bc280 r5:c20b9178 r4:c0a70ae0
[   95.199606] [<c0874a20>] (show_stack) from [<c0877cb8>] (dump_stack+0x28/0x30)
[   95.199636] [<c0877c90>] (dump_stack) from [<c0874840>] (__div0+0x20/0x28)
[   95.199657]  r5:c20b9178 r4:c28004f8
[   95.199692] [<c0874820>] (__div0) from [<c04f5afc>] (Ldiv0+0x8/0x10)
[   95.199768] [<bf3b7000>] (setup_mmal_port_format [bcm2835_isp]) from [<bf3b82d4>] (bcm2835_isp_node_s_fmt+0x148/0x32c [bcm2835_isp])
[   95.199790]  r5:c2521400 r4:c20b9178
[...]

PR #4732, opened 2021-11-25
drivers: bcm2835_isp: Fix div by 0 bug.

naushir:
Fix a possible division by 0 bug when setting up the mmal port for the stats
port.


naushir:
I believe this should fix https://github.com/raspberrypi/linux/issues/4714.
Quite surprised we have not seen this before really!

PR #4733, opened 2021-11-25
drm/vc4: Add support for composite syncs to vc4_dpi

6by9:
The hardware can combine H&V syncs onto the output enable line
as composite syncs, so add the relevant configuration to do that.


popcornmix:
Discussion here: https://forums.raspberrypi.com/viewtopic.php?p=1942948

6by9:
Yes, sorry forgot to put that link in.

Whilst I haven't got anything to test csyncs with, it should be a nop for existing displays, and the new options match the firmware.

We could delay merging until the person on that forum thread has tried it - I'll add a WIP tag.

popcornmix:
I agree it's a no-op if `DRM_MODE_FLAG_CSYNC` is not set, so no objection.

6by9:
Confirmed as working, so WIP removed.

PR #4742, opened 2021-11-27
Add panel driver NWE080 and overlay for CutiePi

penk:
Signed-off-by: Penk Chen <penk@cutiepi.io>


6by9:
You'd said before that the panel is based ILI9881. Your driver is near identical to https://github.com/torvalds/linux/blob/master/drivers/gpu/drm/panel/panel-ilitek-ili9881c.c except for a number of extra (quite large) delays, and an alternate initialisation sequence (which can be handled in the ili9881c driver). Can it really not be handled by extending the existing driver?

There's no need for the panel driver to directly interact with the backlight - the framework will turn it on and off from [drm_panel_enable](https://github.com/torvalds/linux/blob/master/drivers/gpu/drm/drm_panel.c#L151) / [drm_panel_disable](https://github.com/torvalds/linux/blob/master/drivers/gpu/drm/drm_panel.c#L183)

arch/arm/boot/dts/overlays/Makefile
```diff
@@ -34,6 +34,7 @@ dtbo-$(CONFIG_ARCH_BCM2835) += \
 	cap1106.dtbo \
 	chipdip-dac.dtbo \
 	cma.dtbo \
+	cutiepi-panel.dtbo \
```
pelwell, line 37: The overlay will need an entry in the README file (so users can run `dtoverlay -h cutiepi-panel` to find out what it does).

arch/arm/boot/dts/overlays/cutiepi-panel-overlay.dts
```diff
@@ -0,0 +1,124 @@
+/dts-v1/;
+/plugin/;
+
+/ {
+    compatible = "brcm,bcm2711";
```
pelwell, line 5: If this overlay is truly Pi 4-specific then it would be helpful to put an entry in `overlay_map.dts` so that the firmware can refuse to load it on other platforms, e.g.:
```
	cutiepi-panel {
		bcm2711;
	};
```
See https://github.com/raspberrypi/documentation/blob/develop/documentation/asciidoc/computers/configuration/device-tree.adoc#the-overlay-map-file.

arch/arm/boot/dts/overlays/cutiepi-panel-overlay.dts
```diff
@@ -0,0 +1,124 @@
...
+            status = "okay";
+            pinctrl-names = "default";
+            pinctrl-0 = <&i2c6_pins>;
+            clock-frequency = <100000>;
+        };
+    };
+
+    fragment@5 {
+        target = <&i2c6_pins>;
+        __dormant__ {
```
pelwell: This fragment is useless without a parameter to enable it.

pelwell (changes requested):
Before you resubmit the PR, please also address a few minor issues with the overlay.

arch/arm/boot/dts/overlays/Makefile
penk: Thanks, added in last commit.

arch/arm/boot/dts/overlays/cutiepi-panel-overlay.dts
penk: Added in last commit.

arch/arm/boot/dts/overlays/cutiepi-panel-overlay.dts
penk: Removed the fragment@5.

pelwell (commented):
I approve of your changes - the overlay and associated files look good now.

The remaining obstacles to a merge are:
1. 6by9's concerns about code duplication - can you not extend the ili9881c driver with a different compatible string? That would have the side benefit of removing the explicit backlight handling.
2. Is there a technical reason not to add the config settings to the 32-bit 2711 defconfig?
3. When we do come to merge I want the changes in three commits:
    * driver + Makefile + Kconfig, defconfig changes
    * defconfig changes
    * overlay + overlay_map + README + makefile

penk:
@pelwell Hi Phil, thank you very much for the review! I'll get to the driver shortly and test it on a 32bit system. 

As for splitting changes do you have a recommend workflow for it? (sorry I'm not very experienced in this), do I use interactive rebase?

pelwell:
> As for splitting changes do you have a recommend workflow for it? (sorry I'm not very experienced in this), do I use interactive rebase?

Rather than using partial reverts and squashes, I find the easiest way is usually to:
1. Make a note of any nice commit messages you've written so far.
2. If you have _n_ commits that you wish to split or repartition, `git reset HEAD~n`. This will leave all the changes unstaged.
3. Add the changes selectively. Start with whole files that all belong in one commit using `git add <file>`. Be careful not to forget any new files that you are adding, since they won't show up in the list of outstanding changes.
4. If there are files for which you want only some changes to go in the commit you are building, use `git add -p <file>` to be prompted for each change.
5. Finish that commit with the usual `git commit -s`, possibly copying in your saved commit message, then move onto the next.

pelwell:
6. When you've finished, `git push -f <remote> HEAD:<branch>` etc.

penk:
@pelwell Thank you for the wait, I've extended the ili9881c driver per @6by9 requested. And force pushed the 3 commits. Please have a review, thanks!

PR #4744, opened 2021-11-29
KMS fixes and ISP pitch fix

6by9:
- Drop I2C baud rate whilst still investigating 7" display issues
- Avoid an erroneous error log line from a EPROBE_DEFER in HDMI audio
- Fix ISP pitch requirements for 3 plane YUV formats.

### DIFF
--- a/arch/arm/boot/dts/overlays/vc4-kms-dsi-7inch-overlay.dts
+++ b/arch/arm/boot/dts/overlays/vc4-kms-dsi-7inch-overlay.dts
@@ -95,6 +95,7 @@
 		target = <&i2c0if>;
 		__overlay__ {
 			status = "okay";
+			clock-frequency = <50000>;
 		};
 	};
 
--- a/drivers/gpu/drm/vc4/vc4_hdmi.c
+++ b/drivers/gpu/drm/vc4/vc4_hdmi.c
@@ -1848,7 +1848,7 @@ static int vc4_hdmi_audio_init(struct vc4_hdmi *vc4_hdmi)
 	snd_soc_card_set_drvdata(card, vc4_hdmi);
 	ret = devm_snd_soc_register_card(dev, card);
 	if (ret)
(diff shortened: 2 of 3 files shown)


arch/arm/boot/dts/overlays/vc4-kms-dsi-7inch-overlay.dts
```diff
@@ -95,6 +95,7 @@
 		target = <&i2c0if>;
 		__overlay__ {
 			status = "okay";
+			clock-frequency = <50000>;
```
pelwell, line 98: Could this property be set on the reg_display node rather than here? (Asking because I don't know)

6by9: I don't believe so.

AFAIK Nothing follows the tree down to see what the highest speed supported by all the devices added to the bus can run at, so it has to be set explicitly on the host controller.

https://github.com/raspberrypi/linux/blob/rpi-5.10.y/drivers/i2c/busses/i2c-bcm2835.c#L544 reads this clock-frequency value directly.

pelwell: Yes, it seems there's a layer of control missing  from I2C - maximum permissible clock for this device.

popcornmix:
"staging/bcm2835-codec: bytesperline for YUV420/YVU420 needs to be 64" breaks video decode for me in kodi.
(image currupted by wrong pitch).
Using TestMedia/deinterlace/snooker.ts file (720x576 MPEG-2).

6by9:
OK, I'll do a further test.
video decode is less prone to issues as the height is almost always padded out to a multiple of 16 lines, therefore it doesn't get affected by the height only being a multiple of 2.

I appear to have outstanding commits relating to this for the firmware. I'm rebuilding to check they fix things again.

PR #4745, opened 2021-11-29
media: rpivid: remove min_buffers_needed from src queue

jc-kynesim:
Remove min_buffers_needed=1 from src queue init. Src buffers are bound
to media requests therefore this setting is not needed and generates
a WARN in kernel 5.16.


popcornmix:
Confirmed this fixes a WARN in 5.16 and hevc decode still works.

PR #4746, opened 2021-11-29
staging/bcm2835-codec: Allow a different stride alignment per role

6by9:
Deinterlace and decode aren't affected in the same way as encode
and ISP by the alignment requirement on 3 plane YUV420.
Decode would be affected, but it always aligns the height up to
a macroblock, and uses the selection API to reflect that.

Add in the facility to set the bytesperline alignment per role.


6by9:
@popcornmix

6by9:
Attempt 2 - NULL derefs don't work too well.

popcornmix:
Yes, this fixes my regression from "staging/bcm2835-codec: bytesperline for YUV420/YVU420 needs to be 64"

PR #4747, opened 2021-11-30
TM1637 LED driver

rwana1753:
The driver craetes files for TM1637 LED contoller under sysfs
to control TM1637 LED controller with keyscan over gpio.


pelwell:
There are many problems with this PR:
1. The biggest problem is that the driver should be sent upstream to the Linux kernel maintainers. The get_maintainers script suggests sending it to:
    Miguel Ojeda Sandonis <miguel.ojeda.sandonis@gmail.com>
    linux-kernel@vger.kernel.org
2. DEBUG is still enabled on the driver - that won't go down well upstream.
3. This PR is against 4.19 - we're currently on 5.10. and soon to move to 5.15.
4. The overlay and README look as if they have been edited by different people or different editors with different indentation settings. The README rules are no TABs - just spaces - but elsewhere I just expect consistency.

Given all those problems, and the fact that the driver doesn't expose any standard kernel interfaces (only sysfs), wouldn't it be easier to use an external program to control the GPIOs via libgpiod? There's even a Python module for it.

drivers/auxdisplay/tm1637.c
```diff
@@ -475,17 +479,19 @@ static ssize_t tm1637_store_led(struct device *dev,
 	mutex_lock(&tm->lock);
 
 	if (kstrtoul(buf, 0, &data))
-		return -EINVAL;
+		ret = -EINVAL;
+		goto unlock;
```
pelwell, line 483: You've not tested this - this `goto` is unconditional.

But you're wasting your time - as I said previously:
> The biggest problem is that the driver should be sent upstream to the Linux kernel maintainers.

rwana1753: Hello

I also found an error in the code part that returns the status value,
so I corrected it about an hour ago and reflected [it](https://github.com/rwana1753/linux-driver-tm1637/commit/d6586523bf3892abe782d0e373a7e439b3da99f7).

And when controlling using sysfs, high-speed operation is not possible,
so misc is added, and this part is still before the test.

pelwell: No - not sysfs, libgpiod which is ioctl-based and much quicker. How fast does one need to update a 4x7-segment display?

kajflajf:
🍑🍌🐭

PR #4752, opened 2021-12-01
Add support for the AudioInjector.net RAM Platter sound card

flatmax:


arch/arm/boot/dts/overlays/Makefile
```diff
@@ -25,6 +25,7 @@ dtbo-$(CONFIG_ARCH_BCM2835) += \
 	applepi-dac.dtbo \
 	at86rf233.dtbo \
 	audioinjector-addons.dtbo \
+	audioinjector-ram-platter.dtbo \
```
pelwell, line 28: This needs to go after audioinjector-isolated-soundcard.dtbo.

pelwell (commented):
This PR is against 5.16 which will never be the basis of an official release. We're likely to switch to 5.15 in the near future, so that would be a better target if 5.10 isn't suitable (we'll take care of the forward-porting if you choose 5.10).

As is not uncommon with ALSA card drivers, this is tantalisingly close to only needing a simple-card overlay - I think it's just the bclk ratio calculation stopping you. It could almost be added to our rpi-simple driver, but again that would require the addition of a custom bclk ratio calculation, so I'll take this as is with the provisos listed.

pelwell:
Any updates, Matt?

flatmax:
Thanks for chasing this up.

My plan is to migrate to 5.10. Right now however I am buried with tasks 
prior to holidays.

PR #4753, opened 2021-12-02
DPI panel configuration

6by9:
Adds a generic DPI panel configuration overlay

@JamesH65 Could do with some testing, and possibly even a magic script to convert a ```dpi_output_format=``` and ```dpi_timings=``` line into the relevant overlay parameters.

Not all output modes are configured as the relevant MEDIA_BUS_FMT_xxx defines don't exist.

### DIFF
--- a/arch/arm/boot/dts/overlays/README
+++ b/arch/arm/boot/dts/overlays/README
@@ -3563,6 +3563,34 @@ Load:   dtoverlay=vc4-kms-dpi-at056tn53v1
 Params: <None>
 
 
+Name:   vc4-kms-dpi-generic
+Info:   Enable a generic DPI display under KMS. Default timings are for the
+        Adafruit Kippah with 800x480 panel and RGB666 (GPIOs 0-21)
+        Requires vc4-kms-v3d to be loaded.
+Load:   dtoverlay=vc4-kms-dpi-generic,<param>=<val>
+Params: clock-frequency         Display clock frequency (Hz)
+        hactive                 Horizontal active pixels
+        hfp                     Horizontal front porch
+        hsync                   Horizontal sync pulse width
+        hbp                     Horizontal back porch
+        vactive                 Vertical active lines
+        vfp                     Vertical front porch
+        vsync                   Vertical sync pulse width
+        vbp                     Vertical back porch
+        hsync-invert            Horizontal sync active low
+        vsync-invert            Vertical sync active low
+        de-invert               Data Enable active low
+        pixclk-invert           Negative edge pixel clock
+        width-mm                Define the screen width in mm
+        height-mm               Define the screen height in mm
+        rgb565                  Change to RGB565 output on GPIOs 0-19
+        rgb666-padhi            Change to RGB666 output on GPIOs 0-9, 12-17, and
+                                20-25
+        rgb888                  Change to RGB888 output on GPIOs 0-27
+        bus-format              Override the bus format for a MEDIA_BUS_FMT_*
+                                value. NB also overridden by rgbXXX overrides.
+
+
 Name:   vc4-kms-dsi-7inch
 Info:   Enable the Raspberry Pi DSI 7" screen.
         Includes the edt-ft5406 for the touchscreen element.
--- a/arch/arm/boot/dts/overlays/vc4-kms-dpi-generic-overlay.dts
+++ b/arch/arm/boot/dts/overlays/vc4-kms-dpi-generic-overlay.dts
@@ -1,0 +1,92 @@
+/*
+ * vc4-kms-dpi-at056tn53v1-overlay.dts
+ */
+
+/dts-v1/;
+/plugin/;
+
+#include <dt-bindings/gpio/gpio.h>
+#include <dt-bindings/pinctrl/bcm2835.h>
+
+/ {
+	compatible = "brcm,bcm2835";
+
+	fragment@0 {
+		target-path = "/";
+		__overlay__ {
+			panel: panel {
+				compatible = "panel-dpi";
+
+				width-mm = <154>;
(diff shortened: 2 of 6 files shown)


JamesH65:
Oooo shiny. A quick glance though and it looks OK but I have absolutely no way of testing it. This is a real lifesaver for those coming in from using dpi-timings in config.txt.

Chances of being accepted upstream, minus ONE BILLION. But that's their problem, not ours. Reading back through some of the comments on this sort of thing, apparently writing loads of custom code and rebuilding the kernel is the correct way, not some easy to use DT overlay approach.

6by9:
> Chances of being accepted upstream, minus ONE BILLION. 

Actually the only controversial bit is "drm/panel-simple: Allow the bus format to be read from DT for panel-dpi". The main support is in mainline, and this is only a couple of minor tweaks to stop it logging errors.

A very similar thing has been proposed before - https://lkml.org/lkml/2017/10/13/956. And actually there was a solution merged that then got reverted - https://lore.kernel.org/all/20200314153047.2486-2-sam@ravnborg.org/T/ / https://github.com/torvalds/linux/commit/d021d751c14752a0266865700f6f212fab40a18c

arch/arm/boot/dts/overlays/README
```diff
@@ -3563,6 +3563,34 @@ Load:   dtoverlay=vc4-kms-dpi-at056tn53v1
 Params: <None>
 
 
+Name:   vc4-kms-dpi-generic
+Info:   Enable a generic DPI display under KMS. Default timings are for the
+        Adafruit Kippah with 800x480 panel and RGB666 (GPIOs 0-21)
+        Requires vc4-kms-v3d to be loaded.
+Load:   dtoverlay=vc4-kms-dpi-generic
```
pelwell: `,<param>=<val>`

arch/arm/boot/dts/overlays/vc4-kms-dpi-generic-overlay.dts
```diff
@@ -0,0 +1,119 @@
...
+						remote-endpoint = <&dpi_out>;
+					};
+				};
+			};
+		};
+	};
+
+	fragment@1 {
+		target = <&dpi>;
+		__overlay__  {
```
pelwell: Put a `dpi_node:` label here.

arch/arm/boot/dts/overlays/vc4-kms-dpi-generic-overlay.dts
```diff
@@ -0,0 +1,119 @@
...
+			status = "okay";
+
```
pelwell, line 54: Insert `                       pinctrl-0 = <&dpi_18bit_gpio0>;` here.

arch/arm/boot/dts/overlays/vc4-kms-dpi-generic-overlay.dts
```diff
@@ -0,0 +1,119 @@
...
+			port {
+				dpi_out: endpoint {
+					remote-endpoint = <&panel_in>;
+				};
+			};
+		};
+	};
+
+	fragment@2 {
```
pelwell: Delete fragments 2-5.

arch/arm/boot/dts/overlays/vc4-kms-dpi-generic-overlay.dts
```diff
@@ -0,0 +1,119 @@
...
+		hsync-invert = <&timing>, "hsync-active:0=0";
+		vsync-invert = <&timing>, "vsync-active:0=0";
+		de-invert = <&timing>, "de-active:0=0";
+		pixclk-invert = <&timing>, "pixelclk-active:0=0";
+
+		width-mm = <&panel>, "width-mm:0";
+		height-mm = <&panel>, "height-mm:0";
+
+		rgb565 = <&panel>, "bus-format:0=0x1017",
+			<0>, "-2+3";
```
pelwell: Replace this with `<&dpi_node>, "pinctrl-0:0=",<&dpi_16bit_gpio0>;`.

arch/arm/boot/dts/overlays/vc4-kms-dpi-generic-overlay.dts
```diff
@@ -0,0 +1,119 @@
...
+		rgb666-padhi = <&panel>, "bus-format:0=0x1015",
+			<0>, "-2+4";
```
pelwell: Replace this with `&dpi_node>, "pinctrl-0:0=",<&dpi_18bit_cpadhi_gpio0>;`.

arch/arm/boot/dts/overlays/vc4-kms-dpi-generic-overlay.dts
```diff
@@ -0,0 +1,119 @@
...
+		rgb888 = <&panel>, "bus-format:0=0x100a",
+			<0>, "-2+5";
```
pelwell: Replace this with `<&dpi_node>, "pinctrl-0:0=",<&dpi_gpio0>;`.

pelwell (commented):
Apart from the nitpicky README convention, my other comments are a suggestion of how to improve the bus format selection, replacing the selectively enabling fragments model with writing explicit phandle values directly from parameters, similar to what you've already done for `bus-format` itself.

6by9:
Requested changes made.
Selecting rgb888 certainly seems to set the pinctrl for GPIOs 0-27 as Alt2, so it seems to work.

I do need to find my Kippah and screen and check that the output is as expected, but otherwise I think we're good to go.

6by9:
> I do need to find my Kippah and screen and check that the output is as expected, but otherwise I think we're good to go.

Found, and it works fine.

Issue #4685, opened 2021-11-09
IN Token of USB Endpoint 1 stopped occasionally on Pi 4 


JulianGu001:
**Describe the bug**
In our currently BLE project, we use the Pi4 as the BLE host, sometimes, controller's evently buffer exhausted after sent to USB layer(BLE event transfered via endpoint 1).
We captured the USB infer log bug Lecroy sniffer and finding the the IN token of endpoint 1. On host side, we have confirmed that URB has been submitted to endpoint1 successfully, but no IN token sent out.

**To reproduce**
establlish connection with a mass of BLE devices repeatlly(> 80 devices ).

**Expected behaviour**
all success

**Actual behaviour**
sometimes, no HCI events reported to Host, and controller reports "ALLOC_EVENT_NOBUFF",
with Lecroy USB sniffer,  no IN Token sent out on Endpoint 1.

**System**
PRETTY_NAME="Raspbian GNU/Linux 10 (buster)"
NAME="Raspbian GNU/Linux"
VERSION_ID="10"
VERSION="10 (buster)"
VERSION_CODENAME=buster
ID=raspbian
ID_LIKE=debian
HOME_URL="http://www.raspbian.org/"
SUPPORT_URL="http://www.raspbian.org/RaspbianForums"
BUG_REPORT_URL="http://www.raspbian.org/RaspbianBugs"

* Which model of Raspberry Pi? e.g. Pi3B+, PiZeroW
Pi4, reproduced
Pi3, no reproduced.

* Which OS and version (`cat /etc/rpi-issue`)?
Generated using pi-gen, https://github.com/RPi-Distro/pi-gen, dcfd74d7d1fa293065ac6d565711e9ff891fe2b8, stage4

* Which firmware version (`vcgencmd version`)?
version 40787ee5905644f639a2a0f6e00ae12e517a2211 (clean) (release) (start)

* Which kernel version (`uname -a`)?
Linux PiAP1 5.10.771-v7l+ #2 SMP Tue Nov 9 10:17:02 GMT 2021 armv7l GNU/Linux
reproduced on kernel 5.10.52 and kernel 5.10.77

**Logs**
//from the timestamp "487.569599", after the last "xhci_urb_enqueue: ep1in-intr", no IN Token sent out
          <idle>-0       [000] d.s.   487.545584: xhci_queue_trb: INTR: Buffer 000000040405cb40 length 16 TD size 0 intr 0 type 'Normal' flags b:i:I:c:s:I:e:c
[...]

PR #4755, opened 2021-12-03
Fix for #4685

P33M:
I don't have exactly the same hardware setup as the linked issue, but my equivalent testcase (Asix dongle with a modified interrupt endpoint interval) no longer fails with this patch applied. 

While the failure was only seen when a bulk IN and interrupt IN endpoint were both active, there's no proof that the same thing can't occur with combinations of bulk IN and some other periodic endpoint type, so applying the segment restriction globally hedges against other random breakages.


P33M:
#4685

PR #4757, opened 2021-12-08
rpi-5.16.y: media: imx219: Advertise embedded data node on media pad 1

jhautbois:
userland.

[JMH: Adapt to the mainline 5.16 kernel]
Signed-off-by: Jean-Michel Hautbois <jeanmichel.hautbois@ideasonboard.com>

### DIFF
--- a/drivers/media/i2c/imx219.c
+++ b/drivers/media/i2c/imx219.c
@@ -118,6 +118,16 @@
 #define IMX219_PIXEL_ARRAY_WIDTH	3280U
 #define IMX219_PIXEL_ARRAY_HEIGHT	2464U
 
+/* Embedded metadata stream structure */
+#define IMX219_EMBEDDED_LINE_WIDTH 16384
+#define IMX219_NUM_EMBEDDED_LINES 1
+
+enum pad_types {
+	IMAGE_PAD,
+	METADATA_PAD,
+	NUM_PADS
+};
+
 struct imx219_reg {
 	u16 address;
 	u8 val;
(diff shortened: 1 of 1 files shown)


pelwell:
Does this not apply to rpi-5.15.y, which will be our next shipping kernel?

jhautbois:
> Does this not apply to rpi-5.15.y, which will be our next shipping kernel?

Good question, not tested :-)

jhautbois:
It applies fine on rpi-5.15.y too. Do you want me to close this one and open a new one for rpi-5.15-y instead ?

kbingham:
I expect it needs to go into both branches.

pelwell:
In general I'd prefer to have PRs against the earliest viable kernel because we have procedures for forward-porting, but on this occasion I can back-port.

@naushir I take it you're happy with this PR?

kbingham:
We're on the opposite end, trying to port this to upstream so we need the latest kernel version possible....

kbingham:
(Note that we've skipped v5.15.y almost intentionally because unicam does not appear to be fully upported there from 5.10)

pelwell:
Is it picky to want the structure member to be renamed to `pads`?

pelwell:
> unicam does not appear to be fully upported there from 5.10

Do you mean it is worse in 5.15 than 5.10? 5.15 is an LTS, so it would be a real shame if unicam doesn't work properly.

naushir:
> @naushir I take it you're happy with this PR?

Yes, happy with this change.  It matches our 5.10 driver code.

kbingham:
> > unicam does not appear to be fully upported there from 5.10
> 
> Do you mean it is worse in 5.15 than 5.10? 5.15 is an LTS, so it would be a real shame if unicam doesn't work properly.

Hrm... never mind, it's me mis-undertanding something from earlier today.

git diff rpi/rpi-5.10.y..rpi/rpi-5.16.y ./drivers/media/platform/bcm2835/
and
git diff rpi/rpi-5.10.y..rpi/rpi-5.15.y ./drivers/media/platform/bcm2835/

both look fairly sane.

PR #4758, opened 2021-12-08
rpi-5.15.y: media: imx219: Advertise embedded data node on media pad 1

jhautbois:
userland.

[JMH: Adapt to the mainline 5.16 kernel]
Signed-off-by: Jean-Michel Hautbois <jeanmichel.hautbois@ideasonboard.com>


pelwell:
Thanks!

PR #4768, opened 2021-12-14
v2 fix for the VLI TRB cache bug

P33M:
#4685 

Soak tested overnight and seems OK. This is probably a bit unwieldy to keep as a downstream patch for long, though it did apply with a few fixups to 5.15.


popcornmix:
Can it alsa be sent upstream?
A PR for 5.15 too would be useful if it requires fixups.

PR #4759, opened 2021-12-08
vc4/drm: Ignore vc4_hdmi->output_enabled for allowing audio

popcornmix:
Otherwise we reject audio playback when switching hdmi modes

Signed-off-by: Dom Cobley <popcornmix@gmail.com>

PR #4770, opened 2021-12-14
drm/vc4: Skip writes to disabled packet RAM

popcornmix:
This path actually occurs when audio is started during a hdmi mode set.
As the data will be written by vc4_hdmi_set_infoframes when packet RAM
is enabled again, don't treat as an error

### DIFF
--- a/drivers/gpu/drm/vc4/vc4_hdmi.c
+++ b/drivers/gpu/drm/vc4/vc4_hdmi.c
@@ -692,6 +692,7 @@ static void vc4_hdmi_encoder_post_crtc_disable(struct drm_encoder *encoder,
 
 	mutex_lock(&vc4_hdmi->mutex);
 
+	vc4_hdmi->output_enabled = false;
 	spin_lock_irqsave(&vc4_hdmi->hw_lock, flags);
 
 	HDMI_WRITE(HDMI_RAM_PACKET_CONFIG, 0);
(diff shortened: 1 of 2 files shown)


popcornmix:
@HiassofT does this help your issue from https://github.com/raspberrypi/linux/pull/4759#issuecomment-990946526 ?

6by9:
Having the WARN in vc4_hdmi_write_infoframe is useful for ensuring we don't try to write it when powered down.

It seems better to make the ```vc4_hdmi_set_audio_infoframe``` call in ```vc4_hdmi_audio_prepare``` conditional on the video side being enabled (that would be the ```vc4_hdmi->output_enabled``` flag we just removed, but set in ```vc4_hdmi_encoder_post_crtc_enable``` before we call ```vc4_hdmi_set_infoframes```).

Your current patch is also accessing registers without having acquired ```vc4_hdmi->hw_lock``` first.

popcornmix:
@6by9 updated - was that what you were after?

6by9:
Yes, that was what I was thinking of.

Github seems to have messed up the indentation when viewing the patch, but I'll trust that it's GH at fault.

pelwell:
> Github seems to have messed up the indentation when viewing the patch, but I'll trust that it's GH at fault.

Strange - it looks OK here.

6by9:
I'd expanded file before and after the 3rd hunk, and it's differently indented the unmodified bit vs the patch.

popcornmix:
Github is doing something funny when expanding  this part:
```
		vc4_hdmi_set_infoframes(encoder);
	}
	vc4_hdmi_recenter_fifo(vc4_hdmi);
	vc4_hdmi_enable_scrambling(encoder);
```
Seems like they are using different tab widths for the new code and old code,
although it looks fine when I've copied it and in editor it is fine (all initial white space is tabs).

popcornmix:
Removed draft status as tested by @hiassoft (and me for a few days).

popcornmix:
@pelwell okay to merge?

PR #4658, opened 2021-10-28
media: bcm2835-codec: Increase video callbacks limit

popcornmix:
The previous commit limited callbacks to DPB+5.
This has been found insufficient for kodi and ffmpeg
decoding h.264. Increase to DPB+6

Fixes: f814bfc5f4d3005eb266a1556be8b7b8770629bd

Signed-off-by: Dom Cobley <popcornmix@gmail.com>

PR #4773, opened 2021-12-16
Revert "media: bcm2835-codec: Limit video callbacks"

popcornmix:
This reverts commit f814bfc5f4d3005eb266a1556be8b7b8770629bd.

The commit caused media stalls with kodi and stateful
v4l2 video decode.

John is now using a different way of limiting latency
through stateful v4l2 so this is not required.


popcornmix:
@jc-kynesim okay with you?

popcornmix:
Replaces https://github.com/raspberrypi/linux/pull/4658

jc-kynesim:
Fine by me. I think there must be something about that control that doesn't really work. The patch never worked as well as I hoped - hence the need to fix it in userspace.

PR #4775, opened 2021-12-16
WIP: DSI driver updates

6by9:
Converts the vc4 DSI host driver to be a bridge driver using the atomic APIs, and introduces a new flag that allows the initialisation order of the bridge chain to be reversed for particular devices that want their upstream device (eg DSI host) initialised first.


6by9:
Closing as have moved on to 5.15.

PR #4776, opened 2021-12-16
Fix for cursor corruption with overscan

### LINKED ISSUE #4475: Display Corruptions with KMS and Overscan Margins on the RaspberryPi 4
Moving the cursor along the edges of the display creates some visual artifacts when disable_overscan isn't set.

**To reproduce**

Build a kernel based on current the rpi-5.10.y branch (7aa66c086ae0 in my case), start X and move the cursor along the edges. Complete corruption of the display should occur.

Adding disable_overscan=1 to config.txt makes it behave properly.

mripard:
Hi,

Here's a PR that fixes #4475.

The last commit has most of the details, but the basic idea is that when we move the cursor along the edges, we end up reusing and corrupting the active dlist.

Let me know what you think,
Maxime

### DIFF
--- a/drivers/gpu/drm/vc4/vc4_drv.h
+++ b/drivers/gpu/drm/vc4/vc4_drv.h
@@ -341,6 +341,14 @@ struct vc4_hvs {
 
 	/* HVS version 5 flag, therefore requires updated dlist structures */
 	bool hvs5;
+
+	spinlock_t hw_dlist_lock;
+	struct {
+		unsigned int start;
+		unsigned int size;
+		u32 shadow[2048];
+		bool pending;
+	} fifo[3];
 };
 
 struct vc4_plane {
@@ -380,6 +388,11 @@ struct vc4_plane_state {
 	 * hardware at vc4_crtc_atomic_flush() time.
 	 */
 	u32 __iomem *hw_dlist;
+
+	/* Offset where the plane's dlist was last stored in the
+	 * shadow dlist bucket at vc4_crtc_atomic_flush() time.
+	 */
+	unsigned int dlist_offset;
 
 	/* Clipped coordinates of the plane on the display. */
 	int crtc_x, crtc_y, crtc_w, crtc_h;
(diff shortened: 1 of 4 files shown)


pelwell:
I see this is a PR against 5.15 - will it port back to 5.10 or will that require a rewrite?

mripard:
I should have mentioned that indeed. 5.10 seems immune to it as long as we have the reverts @popcornmix did last summer, and especially 222416869a08ca4284f87823a938284ecfc7b297. So we could reintroduce the original commits and apply that on top, but I don't think it's worth the trouble.

pelwell:
Thanks - that fixes it for me.

popcornmix:
While this fixes the cursor close to overscan issue, the other (seemingly related) issue is still present.
I've booted with two monitors (they happen to be 4kp30). I have overscan disabled.
If I move mouse between screens I'll get a corrupt glitch on one or other displays.
It feels like a similar issue of updating a live display list.

Using BRANCH=next rpi-update kernel (which includes this PR)
```
Linux pi4 5.15.10-v7l+ #1503 SMP Fri Dec 17 15:36:35 GMT 2021 armv7l GNU/Linux
```

If I transverse screens with the mouse slowly, no glitch occurs. Rapidly switching mouse between screens makes the glitch very frequent.

popcornmix:
With 5.10 kernel the issue does not occur:
```
Linux pi4 5.10.87-v7l+ #1502 SMP Fri Dec 17 15:13:41 GMT 2021 armv7l GNU/Linux
```

mripard:
I spent some time looking into this. It seems that it only occurs on the edge that is next to the other monitor, but the cursor doesn't have to cross from one display to the other. If you just move the cursor along the edge, some corruptions occur as well.

I'm not quite sure what's going on at that point. The interrupts fire properly, the dlist seem to be copied over properly, and it really only occurs when we're close to the other display

I'm wondering if it's not due to the cursor plane being larger than the actual image displayed in it, and we have it shared between both displays

popcornmix:
There should be no issue with a source element being displayed on two displays from a hardware perspective.
We've certainly had HVS demos in the past rending a single source image many times (on one display) as sprites.

As far as the hardware is concerned it's just the source address being used for reads. It really doesn't care where they come from in memory.

Now if the images are being resized vertically then the context memory used by hardware for this needs to distinct for each element.

PR #4754, opened 2021-12-03
BCM2711 KMS YUV Support

mripard:
Hi,

This is another attempt of supporting YUV output in the HDMI driver following the discussions made on slack recently.

For reference, the other two attempts were #4201 and #4280.

This series tries to mimic what the Intel driver is doing, and will try to use the RGB format first, and if we're above our clock limits will try again using YUV422. If both fails, we'll reject the commit entirely.

This is based on 5.15, let me know if you would prefer a branch based on 5.10. I also plan on sending it today upstream to start the discussion and see if it gets more traction than the other attempts.

Maxime

PR #4777, opened 2021-12-17
BCM2711 KMS YUV Support

popcornmix:
5.10 version of https://github.com/raspberrypi/linux/pull/4754


popcornmix:
Edited "drm/vc4: hdmi: Take the sink maximum TMDS clock into account " to include change from https://github.com/raspberrypi/linux/pull/4754#issuecomment-996718292

PR #4782, opened 2021-12-20
Add V4L2_CID_NOTIFY_GAINS control

davidplowman:
This is a pair of commits from linuxtv-master (where they are a9c80593 and 311a839a) and which I'm going to need soon-ish on our linux branches. There was a slight merge issue because v4l2-ctrls.c has spawned a v4l2-ctrls-defs.c file in more recent versions, so for 5.15 it may be easier to grab them again from linuxtv.

@6by9 @naushir Presumably no problem in merging this...


6by9:
Phil would normally request that the first line of the commit body text read
```
Commit a9c80593ff80ddb7c6496624e5384e1ea3460a72 upstream.
```
or similar so that it is obvious that they can be dropped when that upstream commit filters through.
Otherwise I'm happy with them.

davidplowman:
No prob, I can do that!

PR #4788, opened 2021-12-22
change max i2c transmission

prodriguezsc:


pelwell:
Why? What is your use case?

prodriguezsc:
> Why? What is your use case?

I need send to a tof sensor 32270 bytes at one time from my RPi.

pelwell:
Does it not support reading multiple values in a single access? One value per access is quite inefficient.

prodriguezsc:
> Does it not support reading multiple values in a single access? One value per access is quite inefficient.

Sure, the rpi sends 32770 bytes 3 times to program the firmware of the tof sensor. But it hangs at 8192, which is the maximum.

bou6:
Can you give details on the sensor you are using ?

6by9:
Kernel drivers don't have this limit, only userspace.
I would have expected TOF sensors to be controlled by a kernel driver, with the firmware load mechanism used to get the required firmware file.

PR #4789, opened 2021-12-22
configs: add CONFIG_LEDS_TRIGGER_PATTERN=y

marcone:
Enable the pattern led trigger.

Signed-off-by: Marco Nelissen <marco.nelissen@gmail.com>


pelwell:
You didn't use `make ... savedefconfig`, so the new config options are not in the natural place.

And do you have a good reason not to make the new trigger a loadable module, since it is the largest object file (bigger even than ledtrig-netdev, which is a module)?

marcone:
> You didn't use `make ... savedefconfig`, so the new config options are not in the natural place.

Sorry, first time contributor. Didn't know that was how it was supposed to be done.
 
> And do you have a good reason not to make the new trigger a loadable module, since it is the largest object file (bigger even than ledtrig-netdev, which is a module)?

I looked at the other triggers that were modules, and figured they were modules because the things they were triggering on were either not present "out of the box" on every variant of Raspberry Pi (not everyone connects a camera, Raspberry Pi Zero doesn't have network unless you connect a USB dongle) or obsolete ("transient" seems to have been superseded by "oneshot").
Since the pattern trigger will work on any Raspberry Pi, I figured it was appropriate to include the pattern trigger in the kernel proper.

I didn't consider size to be a concern because there are currently 10 non-module triggers even though a Raspberry Pi has at most two onboard LEDs and so would be using at most 2 different triggers at a time. I figured that if size were a concern, _all_ triggers would be modules.

marcone:
I redid the changes with `make ... savedefconfig`.
I left it as `...=y` for the reasons I mentioned above, but let me know if you want this as a loadable module instead.

pelwell:
No - I'd like it as a module, please. The pattern trigger appears to be unusable without user-space help to set up the pattern, so there's no reason for that user-space code to not also arrange for the module to be loaded.

marcone:
Makes sense. The pattern trigger pattern doesn't have a default pattern like e.g. timer and heartbeat do.
I've updated the change.

PR #4794, opened 2021-12-30
drm/vc4: fix non-working audio on RPi3 and no video with DVI

HiassofT:
Audio output on RPi3 (and other models with polled HPD) was broken because hdmi_monitor got cleared on subsequent polls.

The RGB format check wasn't necessary (the format is mandatory anyways and other HDMI drivers don't check for it either) and broke video output on DVI monitors.

ping @mripard @popcornmix


popcornmix:
These both look good to me.

popcornmix:
@6by9 are you happy with this?

6by9:
Yes, looks fine to me.

PR #4807, opened 2022-01-04
media: i2c: ov9281: Increase diff between VTS and max exposure

6by9:
The driver did allow the exposure to go up to VTS - 4 lines,
but this would produce a visible line on 1280x800, and a stall of
the sensor at 640x480.

Whilst it appears to work with a difference of 5, the datasheet states
there should be at least 25 lines difference between VTS and exposure,
so use that value.

### DIFF
--- a/drivers/media/i2c/ov9281.c
+++ b/drivers/media/i2c/ov9281.c
@@ -52,7 +52,11 @@
 #define OV9281_REG_EXPOSURE		0x3500
 #define	OV9281_EXPOSURE_MIN		4
 #define	OV9281_EXPOSURE_STEP		1
-#define OV9281_VTS_MAX			0x7fff
+/*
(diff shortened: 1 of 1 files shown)


pelwell:
> Whilst it appears to work with a difference of 5, the datasheet states
there should be at least 25 lines difference between VTS and exposure,
so use that value.

This sentence made me expect OV9281_EXPOSURE_OFFSET to be set to 25, not 5 - "although I can get away with A, the datasheet says B, so do that" suggests "that" is B, not A.

6by9:
Oops, I tested with 5 as I was writing that comment, and then forgot to restore it back to 25. Will update.

6by9:
Updated.

naushir:
LGTM

PR #4808, opened 2022-01-04
drm/vc4: Correct interrupt enable bits on hvs4

popcornmix:
The interrupt enable bits are numbered differently on hv4 (pi0-3) and hvs5 (pi4)

### DIFF
--- a/drivers/gpu/drm/vc4/vc4_hvs.c
+++ b/drivers/gpu/drm/vc4/vc4_hvs.c
@@ -621,7 +621,7 @@ static void vc4_hvs_install_dlist(struct vc4_dev *vc4, unsigned int channel)
 	hvs->fifo[channel].pending = false;
 
 	reg = HVS_READ(SCALER_DISPCTRL);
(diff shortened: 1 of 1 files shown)


popcornmix:
We currently have a "flip done" timeout on pi0-3 with 5.15 kernel (this code doesn't exist on 5.10) due to using wrong interrupt enable bits.

I'm not sure if we have a better scheme for hv4/hv5 specific bit fields.
A variant specific accessor function is possible (but seems somewhat overkill).

@mripard

pelwell:
Looks OK, but l would prefer spaces around the colons before merging..

pelwell:
I'm happy for this to be merged now, unless you want to wait for alternative suggestions from Maxime.

popcornmix:
Okay, I can deal with a reworked upstream version if @mripard prefers to adjust it.

mripard:
For the record, it looks good to me

6by9:
Please see #4821 which takes that further as there were still mismatched bitmasks (oh why did they have to move everything around like that!)

PR #4809, opened 2022-01-04
Use GitHubs issue form for bug reports, add config.yml for general questions and problems with the Raspbian distribution packages

andi34:
config

  
  
- modern look for bug reports
- user don't need to mess with given markdown parts while filling the issue template

Screenshot_2022-01-04_21-47-24


pelwell:
Thanks - we've just been discussing adding forms to our repos, and you've given us a good starting point. Can you change "behavior" to the UK English spelling "behaviour" - that's our house style and we try to be consistent - before I merge this PR?

JamesH65:
Might be worth having a drop down for the HW being used to make sure that information is entered. e.g. Pi1, Pi2, Pi3, Pi4, CM4 etc. With an associated one for memory size?

pelwell:
That doesn't really fit with issues of the form "Please enable this kernel feature".

andi34:
> Thanks - we've just been discussing adding forms to our repos, and you've given us a good starting point. Can you change "behavior" to the UK English spelling "behaviour" - that's our house style and we try to be consistent - before I merge this PR?

Sure, I'll update it later.
Anything else should be added?
In Long Term, if interested, a own form for Kernel features could be added. There's a lot of possibilities.
I am using different forms in my project already:
https://github.com/andi34/photobooth/issues

pelwell:
That's a useful example - I see that we would could create a separate `feature_request.yml` that GitHub would add to the list.

andi34:
Right!
This also solves the problem people not filling given markdown parts ;)

andi34:
Should I add a hardware list to choose from?

JamesH65:
I think so, but it's not my repo! Making it as easy as possible for a user to get valid/requried/accurate information in to a bug report is the aim of the game.

pelwell:
If list ordering is preserved, add "Other" at the bottom of the list.

JamesH65:
Was just about to add "Might need `All`", perhaps `other` covers it?

andi34:
> Was just about to add "Might need `All`", perhaps `other` covers it?

Multi select? I think most people won't test on every model .

timg236:
There's also a "CM4 lite" variant

andi34:
Sorry for the different changes, just did it mobile. I'll squash them later and force push before it can be merged.

Please let me know if anything else should be adjusted.

You can take a look at it  inside my fork for a live preview: https://github.com/andi34/linux-1/issues/new?assignees=&labels=&template=bug_report.yml

Screenshot_20220105-154927

Screenshot_20220105-154916

pelwell:
Thanks - there are likely to be things we want to adjust later, but this is a great first attempt. I'll squash the commits in the merge - I don't see any great value in keeping the first two commits separate.

andi34:
Thanks a lot! Also for keeping things up to date on the Pi :-)

I was about to fill a bug report because on Buster based Pi OS the Pi camera is not detected anymore on Chromium and thought it could be kernel related. Thought the issue form might help keeping things cleaner for you on such a big project.
(But now looks like it's an Chromium issue with the camera, works on Chromium v88, while it fails on v90. Need to investigate further.)

I guess you take care about the other kernel branches to get this changes in too (for consistent)?

pelwell:
I think it is only necessary that the forms are present in the default branch - I haven't been able to circumvent the form yet by switching to another branch - but we'll be sure to copy it over to rpi-5.15.y when that becomes the new default.

PR #4813, opened 2022-01-07
Add overlay for Azoteq IQS550

jlabundy:
This series enables the Azoteq IQS550 trackpad/touchscreen controller and adds a corresponding overlay.

The vendor offers an evaluation kit (IQS550EV02) which, when paired with a suitable X driver, makes for a handy multitouch trackpad. The following demo shows the device paired with Raspberry Pi: https://youtu.be/sRNNx4XZBts

The overlay is only pushed against rpi-5.15.y; this is because the mainline driver only recently allowed the reset GPIO to be optional. The reset GPIO is only used to access the IQS550 bootloader, and the IQS550EV02 must be modified to access the pin.


pelwell:
Active-high interrupts aren't that common, but it is used in the dt-bindings document and obviously works.

Nicely done - I can't find a single fault.

JamesH65:
> Nicely done - I can't find a single fault.

Has this ever happened before? Do we need beer?

6by9:
> Do we need beer?

We always need beer! Shame you're not in the office to share it :-)

PR #4815, opened 2022-01-07
dtoverlays: Add backlight-gpio parameter to vc4-kms-dpi-generic

6by9:
To allow for the cases where a simple panel does have a GPIO
controlled backlight. Defaults to having no backlight defined.


pelwell:
You can have a "Nicely done" as well - I'm hoping to establish an equivalent of the Paul Hollywood handshake.

6by9:
:-)

It did take a second push as first time I missed the double blank line between entries in README.

PR #4821, opened 2022-01-11
5.15 KMS fixes

6by9:
Fixes up interrupt handling for HVS5. Bits have moved in DISPSTAT and DISPCTRL, and the driver wasn't taking it into account.
This effectively extends what popcornmix had done with https://github.com/raspberrypi/linux/commit/3474e574a04b91ea20da6bc5809d3aa67e592361.
I was seeing VSTART2 occurring and not being acknowledged when 3 displays were active. This resulted in the interrupt being disabled, which then meant no further dlist updates.

Adds a blend_mode property to allow for non-pre-multiplied alpha (or ignore alpha).

### DIFF
--- a/drivers/gpu/drm/vc4/vc4_plane.c
+++ b/drivers/gpu/drm/vc4/vc4_plane.c
@@ -665,6 +665,48 @@ static const u32 colorspace_coeffs[2][DRM_COLOR_ENCODING_MAX][3] = {
 	}
 };
 
+static u32 vc4_hvs4_get_alpha_blend_mode(struct drm_plane_state *state)
+{
+	if (!state->fb->format->has_alpha)
+		return VC4_SET_FIELD(SCALER_POS2_ALPHA_MODE_FIXED,
+				     SCALER_POS2_ALPHA_MODE);
+
+	switch (state->pixel_blend_mode) {
+	case DRM_MODE_BLEND_PIXEL_NONE:
+		return VC4_SET_FIELD(SCALER_POS2_ALPHA_MODE_FIXED,
+				     SCALER_POS2_ALPHA_MODE);
+	default:
+	case DRM_MODE_BLEND_PREMULTI:
+		return VC4_SET_FIELD(SCALER_POS2_ALPHA_MODE_PIPELINE,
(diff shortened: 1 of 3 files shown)


drivers/gpu/drm/vc4/vc4_regs.h
```diff
@@ -245,6 +245,23 @@
 /* Enables interrupt generation on scaler profiler interrupt. */
 # define SCALER_DISPCTRL_SCLEIRQ		BIT(0)
 
+/* Enables Display 0 short line and underrun contribution to
+ * SCALER_DISPSTAT_IRQDISP0.  Note that short frame contributions are
+ * always enabled.
+ */
+# define SCALER5_DISPCTRL_DSPEISLUR(x)		BIT(13 + ((x) * 4))
```
popcornmix: Isn't this `9 + ((x) * 4)` ?

6by9: Yes - thank you. Fixed.

drivers/gpu/drm/vc4/vc4_plane.c
```diff
@@ -665,6 +665,50 @@ static const u32 colorspace_coeffs[2][DRM_COLOR_ENCODING_MAX][3] = {
...
+{
+	if (!state->fb->format->has_alpha)
+		return VC4_SET_FIELD(SCALER5_CTL2_ALPHA_MODE_FIXED,
+				     SCALER5_CTL2_ALPHA_MODE);
+
+	switch (state->pixel_blend_mode) {
+	case DRM_MODE_BLEND_PIXEL_NONE:
+		return VC4_SET_FIELD(SCALER5_CTL2_ALPHA_MODE_FIXED,
+				     SCALER5_CTL2_ALPHA_MODE);
+		return 0;
```
popcornmix: Spurious `return 0`?

popcornmix:
LGTM (apart from two comments).

drivers/gpu/drm/vc4/vc4_plane.c
6by9: Sorted.

6by9:
Comments fixed and rebased.

PR #4822, opened 2022-01-11
ti-abb-regulator.c : fixed typo

RuffaloLavoisier:
line 487 `Calulate` -> `Calculate`

### DIFF
--- a/drivers/regulator/ti-abb-regulator.c
+++ b/drivers/regulator/ti-abb-regulator.c
@@ -484,7 +484,7 @@ static int ti_abb_init_timings(struct device *dev, struct ti_abb *abb)
 	/* Calculate cycle rate */
 	cycle_rate = DIV_ROUND_CLOSEST(clock_cycles * 10, clk_rate);
 
-	/* Calulate SR2_WTCNT_VALUE */
+	/* Calculate SR2_WTCNT_VALUE */
 	sr2_wt_cnt_val = DIV_ROUND_CLOSEST(abb->settling_time * 10, cycle_rate);
 
 	dev_dbg(dev, "%s: Clk_rate=%ld, sr2_cnt=0x%08x\n", __func__,


pelwell:
That's not a file we've modified, and not even one we build - report it upstream.

RuffaloLavoisier:
@pelwell 
Hi, there ! what should I do if I want to contribute to Raspberry Pi linux kernel? Is there a guide or document for referring to the procedure or method?

pelwell:
There is no documented procedure, but it would say something like:
1. Open an issue or PR here and you will get guidance.
2. Following the usual Linux coding guidelines: https://www.kernel.org/doc/html/latest/process/coding-style.html
3. If the files you are patching are unmodified compared to the upstream Linux branch, or if the patches are additions that aren't specific to Raspberry Pis, expect to be told to go through the usual Linux submission process. Every downstream patch is a maintenance burden, and we are trying to reduce (not increase) their number.

RuffaloLavoisier:
> 2\. Following the usual Linux coding guidelines: https://www.kernel.org/doc/html/latest/process/coding-style.html

@pelwell 
Thank you for your answer. Does most PR have to go through the Linux patch?
Including these changes!

pelwell:
We don't get a huge number of PRs, and many of them come from Raspberry Pi or our associates. Of the rest, the vase majority are to code that is specific to Raspberry Pi products or to our downstream tree (all the overlays, for example, and audio HATs) so we don't often have to refer people upstream - maybe once or twice a week.

RuffaloLavoisier:
@pelwell 
Today I learned, thank you. Have a nice day!

pelwell:
For the record, I don't like sending users upstream, but it isn't sustainable to accept all patches here.

RuffaloLavoisier:
@pelwell 
Ah..so what will happen to the general pull request merge process ( typo correction, code style, code correction)?

pelwell:
I'm not sure I understand the question, but if the patch is something likely to concern Pi users, e.g. fixing a bug that they might encounter, then you may get some feedback and assistance here. In the list of issues that might get support, fixing typos in messages users might see is very low down; typos in code comments don't make it onto the list at all - sorry.

RuffaloLavoisier:
@pelwell 
Thanks for answer 😄

PR #4826, opened 2022-01-13
drm/vc4: hdmi: Fix clock value used for validating hdmi modes

popcornmix:
We are using mode->crt_clock here which is filled by drm_mode_set_crtcinfo()
which is called right after .mode_valid.

Use mode->clock which is valid here.

Fixes: 624d93a4f0 ("drm/vc4: hdmi: Move clock calculation into its own function")


popcornmix:
This is similar to bug reported here: https://github.com/raspberrypi/linux/pull/4754#pullrequestreview-833096025

@mripard can you check this.

This is the cause of the regression in bullseye with latest kernel where 2560x1440 gives "flip done timed out".

mripard:
Yep, it looks good to me

Issue #4828, opened 2022-01-13
5.10.89 crashes when booted from u-boot -> grub


colemickens:
### Describe the bug

If I build and boot the foundation kernel 5.10.89 (corresponding build for the current stable firmware), I get a problem.

I'm booting via u-boot, which then loads grub's efi application which then tries to boot the foundation kernel.

This is the crash I get early on:
```
[    0.000000] CPU: 0 PID: 0 Comm: swapper Not tainted 5.10.89 #1-NixOS                                          
[    0.000000] Hardware name: Raspberry Pi 4 Model B Rev 1.1 (DT)                                                
[    0.000000] pstate: 80000085 (Nzcv daIf -PAN -UAO -TCO BTYPE=--)                                              
[    0.000000] pc : kthread_should_stop+0x20/0x48                                                                
[    0.000000] lr : add_hwgenerator_randomness+0x4c/0x140                                                        
[    0.000000] sp : ffffffe159ca3d90                                                                             
[    0.000000] x29: ffffffe159ca3d90 x28: 0000000014117380                                                       
[    0.000000] x27: 000000002386f040 x26: fffffffefe4bc050                                                       
[    0.000000] x25: 0000000000000018 x24: 0000000000000040                                                       
[    0.000000] x23: fffffffefe4fc044 x22: 0000000000000200                                                       
[    0.000000] x21: ffffffe159cb7c40 x20: fffffffefe4fc040                                                       
[    0.000000] x19: ffffffe159cb3178 x18: 0000000000000020                                                       
[    0.000000] x17: 0000000000001400 x16: 0000000000001c00                                                       
[    0.000000] x15: ffffffe159cb8190 x14: 52455345524d454d                                                       
[    0.000000] x13: 2030343030373833 x12: 3278303d474e5220                                                       
[    0.000000] x11: ffffffe159d473e0 x10: ffffffe159d2f3a0                                                       
[    0.000000] x9 : ffffffe159d2f3f8 x8 : 0000000000017fe8                                                       
[    0.000000] x7 : c0000000ffffefff x6 : 0000000000000001                                                       
[    0.000000] x5 : 0000000000000000 x4 : 0000000000000000                                                       
[    0.000000] x3 : 0000000000000001 x2 : 0000000000000000                                                       
[    0.000000] x1 : 0000000000200000 x0 : 0000000000000000                                                       
[    0.000000] Call trace:                                                                                       
[    0.000000]  kthread_should_stop+0x20/0x48                                                                    
[...]

PR #4829, opened 2022-01-14
random: fix crash on multiple early calls to add_bootloader_randomness()

pelwell:
This is a possible fix for #4828. Hat tip to @P33M for finding the upstream patch.

commit f7e67b8e803185d0aabe7f29d25a35c8be724a78 upstream.

Currently, if CONFIG_RANDOM_TRUST_BOOTLOADER is enabled, multiple calls
to add_bootloader_randomness() are broken and can cause a NULL pointer
dereference, as noted by Ivan T. Ivanov. This is not only a hypothetical
problem, as qemu on arm64 may provide bootloader entropy via EFI and via
devicetree.

On the first call to add_hwgenerator_randomness(), crng_fast_load() is
executed, and if the seed is long enough, crng_init will be set to 1.
On subsequent calls to add_bootloader_randomness() and then to
add_hwgenerator_randomness(), crng_fast_load() will be skipped. Instead,
wait_event_interruptible() and then credit_entropy_bits() will be called.
If the entropy count for that second seed is large enough, that proceeds
to crng_reseed().

However, both wait_event_interruptible() and crng_reseed() depends
(at least in numa_crng_init()) on workqueues. Therefore, test whether
system_wq is already initialized, which is a sufficient indicator that
workqueue_init_early() has progressed far enough.

If we wind up hitting the !system_wq case, we later want to do what
would have been done there when wqs are up, so set a flag, and do that
work later from the rand_initialize() call.

Reported-by: Ivan T. Ivanov <iivanov@suse.de>
Fixes: 18b915ac6b0a ("efi/random: Treat EFI_RNG_PROTOCOL output as bootloader randomness")
Cc: stable@vger.kernel.org
Signed-off-by: Dominik Brodowski <linux@dominikbrodowski.net>
[Jason: added crng_need_done state and related logic.]
Signed-off-by: Jason A. Donenfeld <Jason@zx2c4.com>

### DIFF
--- a/drivers/char/random.c
+++ b/drivers/char/random.c
@@ -461,6 +461,7 @@ static struct crng_state primary_crng = {
  * its value (from 0->1->2).
  */
 static int crng_init = 0;
+static bool crng_need_final_init = false;
 #define crng_ready() (likely(crng_init > 1))
 static int crng_init_cnt = 0;
 static unsigned long crng_global_init_time = 0;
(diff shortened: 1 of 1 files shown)


colemickens:
Again confirming that this patch does resolve #4828.

PR #4830, opened 2022-01-14
drm/vc4: Add alpha_blend_mode property to each plane.

popcornmix:
Move from only supporting the default of pre-multiplied
alpha to supporting user specified blend mode using the
standardised property.


popcornmix:
@jc-kynesim was requesting this for 5.10 kernel, so cherry-picking from 5.15.

6by9:
I was still waiting from confirmation on https://forums.raspberrypi.com/viewtopic.php?t=327210 that it solved their problem, but if @jc-kynesim  can confirm it, that would also be useful.

PR #4832, opened 2022-01-14
Improved discovery of SSD and SMR drives which require TRIM

tmm1:
This is a patchset developed by linux-scsi maintainer @martinkpetersen, slated for inclusion in 5.18. This PR was backported from https://git.kernel.org/pub/scm/linux/kernel/git/mkp/linux.git/log/?h=5.18/discovery

Prior to this patchset, USB attached SSD and HDDs (using either UAS or legacy USB storage) would not reliably have their `provisioning_mode` set. This caused the fstrim operation to fail. In fact, many such drives would be initialized with their `provisioning_mode` set to `writesame_16`, even though the drive firmware didn't support WRITESAME at all.

- https://community.wd.com/t/invalid-command-operation-code-write-same/242463
- https://bugzilla.redhat.com/show_bug.cgi?id=1874092

The result was that drives that required TRIM were not using it, causing increased wear on SSD drives, and reduced I/O performance on SMR HDDs.

This is a well-known and well-documented issue, with workarounds that essentially changed the `provisioning_mode` to `unmap` from userland, using either udev rules or manual `/sys` twiddling.

- https://forums.raspberrypi.com/viewtopic.php?p=1708655#p1708655
- https://www.jeffgeerling.com/blog/2020/enabling-trim-on-external-ssd-on-raspberry-pi
- https://github.com/muhkuh-sys/raspberry_enable_trim
- https://www.glump.net/howto/desktop/enable-trim-on-an-external-ssd-on-linux
- https://wiki.archlinux.org/title/Solid_state_drive#External_SSD_with_TRIM_support
- https://gist.github.com/cathay4t/e80e02a737242a5f3824606543631bfe

I noticed this issue myself when booting a number of RPI4s with consumer USB HDDs. Over time, some drives' performance deteriorated to the point where the system was spending all its time in iowait. Further investigation revealed that many manufacturers are now using SMR technology, and basically require TRIM to be able to garbage collect the platter. Without TRIM commands being issued, deleted space is never reclaimed and the drive spends all its time searching for empty blocks to write to.

- https://superuser.com/questions/1407990/what-does-trim-on-an-hdd-mean/1511552#1511552
- https://www.anandtech.com/show/14849/wd-my-passport-5tb-das-review-a-compact-capacity-play-sans-smr-hassles/3
- https://unix.stackexchange.com/questions/306229/how-do-i-check-trim/339684#339684

Previous approaches on linux-scsi tried to remove the writesame16 fallback, or blacklist specific models of drives:

- https://www.spinics.net/lists/linux-scsi/msg159347.html
- https://www.spinics.net/lists/linux-scsi/msg159977.html
- https://mpdesouza.com/blog/kernel-adventures-enabling-vpd-pages-for-usb-storage-devices-in-sysfs/

This new patchset supersedes all previous hacks, and will reliably detect drives that advertise TRIM support. It has been tested and confirmed working with a variety of USB SSDs and HDDs, in both UAS and legacy usb storage modes.

I am creating this PR for visibility into this issue, and am not advocating for immediate merge. It probably makes more sense to wait until the patches are submitted to LKML and then backport the final commits once they land upstream.

### DIFF
--- a/drivers/scsi/mpt3sas/mpt3sas_scsih.c
+++ b/drivers/scsi/mpt3sas/mpt3sas_scsih.c
@@ -11313,20 +11313,18 @@ scsih_pci_mmio_enabled(struct pci_dev *pdev)
  */
 bool scsih_ncq_prio_supp(struct scsi_device *sdev)
 {
-	unsigned char *buf;
+	struct scsi_vpd *vpd;
 	bool ncq_prio_supp = false;
 
-	if (!scsi_device_supports_vpd(sdev))
-		return ncq_prio_supp;
+	rcu_read_lock();
+	vpd = rcu_dereference(sdev->vpd_pg89);
+	if (!vpd || vpd->len < 214)
+		goto out;
 
-	buf = kmalloc(SCSI_VPD_PG_LEN, GFP_KERNEL);
-	if (!buf)
-		return ncq_prio_supp;
+	ncq_prio_supp = (vpd->data[213] >> 4) & 1;
+out:
+	rcu_read_unlock();
 
-	if (!scsi_get_vpd_page(sdev, 0x89, buf, SCSI_VPD_PG_LEN))
-		ncq_prio_supp = (buf[213] >> 4) & 1;
-
-	kfree(buf);
 	return ncq_prio_supp;
 }
 /*
--- a/drivers/scsi/scsi.c
+++ b/drivers/scsi/scsi.c
@@ -511,21 +511,30 @@ int scsi_report_opcode(struct scsi_device *sdev, unsigned char *buffer,
 {
 	unsigned char cmd[16];
 	struct scsi_sense_hdr sshdr;
-	int result;
+	int result, request_len;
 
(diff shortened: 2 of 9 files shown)


popcornmix:
I suggested @tmm1 post here when he pointed out the patch set.
We're unlikely to merge it until it gets merged upstream, as we don't have the detailed knowledge to evaluate it.

But if other users suspect they may be having TRIM issues, this provides a discussion point, and a patch-set to test.

If any users try it, we'd be interested to hear any feedback.

andrum993:
The only issue I have had with TRIM is that it is not enabled by default on any of my drives which support it, so I need to go in and set a UDEV rule to enable it. That is the same on all platforms.

I don't think there is a particular problem with TRIM not being enabled - if folk need it to work then the UDEV workaround is easily discoverable via Google - the bug has been in Linux for a while. Better to let mainline shake out any issues before it lands on the Pi: changing the behaviour of a key subsystem mid-release would definitely be a Bad Thing&trade;

PR #4834, opened 2022-01-18
drivers: bcm2835_unicam: Disable trigger mode operation

naushir:
On a Pi3 B/B+ platform the imx219 sensor frequently generates a single corrupt
frame when the sensor first starts. This can either be a missing line, or
invalid samples within the line. This only occurrs using the Unicam kernel
driver.

Disabling trigger mode elimiates this corruption. Since trigger mode is a
legacy feature copied from the firmeare driver and not expected to be needed,
remove it. Tested on the Raspberry Pi cameras and shows no ill effects.

### DIFF
--- a/drivers/media/platform/bcm2835/bcm2835-unicam.c
+++ b/drivers/media/platform/bcm2835/bcm2835-unicam.c
@@ -983,11 +983,6 @@ static irqreturn_t unicam_isr(int irq, void *dev)
 		}
 	}
 
-	if (reg_read(unicam, UNICAM_ICTL) & UNICAM_FCM) {
-		/* Switch out of trigger mode if selected */
-		reg_write_field(unicam, UNICAM_ICTL, 1, UNICAM_TFC);
-		reg_write_field(unicam, UNICAM_ICTL, 0, UNICAM_FCM);
-	}
 	return IRQ_HANDLED;
 }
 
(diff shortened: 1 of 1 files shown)


naushir:
@6by9, this does seem to be a valid fix for the pink frames (fingers crossed)!

naushir:
Only question is do we disable this selectively based on a dt config?  Ideally I would hope we can disable this globally.

drivers/media/platform/bcm2835/bcm2835-unicam.c
```diff
@@ -983,11 +983,6 @@ static irqreturn_t unicam_isr(int irq, void *dev)
 		}
 	}
 
-	if (reg_read(unicam, UNICAM_ICTL) & UNICAM_FCM) {
-		/* Switch out of trigger mode if selected */
-		reg_write_field(unicam, UNICAM_ICTL, 1, UNICAM_TFC);
```
pelwell, line 988: To someone who knows nothing about the subject matter, the `1` here is unexpected.

naushir: That's hitting the trigger bit in the ICTL register.  It's using UNICAM_TFC as the bitmask.

6by9:
I see little reason not to disable it globally.

Typo in commit message "firmeare", but otherwise I'm happy.

Possibly worth pinging Laurent et al about this patch as they are upstreaming Unicam.

naushir:
> Possibly worth pinging Laurent et al about this patch as they are upstreaming Unicam.

Yes, I will ping Laurent and co. about this commit.

PR #4835, opened 2022-01-19
PoE HAT from Linux I2C

6by9:
As discussed with pelwell by email, switching the PoE HAT fan control (initially) to Linux I2C so that it can co-exist with libcamera and KMS driving the 7" DSI panel.

I need to do the same to the PoE+ HAT power supply driver (shouldn't be a big job now), but am creating the PR for comments on any glaring structural issues.


arch/arm/boot/dts/overlays/rpi-poe-overlay.dts
```diff
@@ -89,6 +79,52 @@
 		};
 	};
 
+	fragment@3 {
+		target = <&firmware>;
+		__overlay__ {
+			poe_pwm: pwm {
```
pelwell: Including `fw` somewhere in the label name (as used to be the case) would be clearer.

arch/arm/boot/dts/overlays/rpi-poe-overlay.dts
```diff
@@ -89,6 +79,52 @@
...
+			#size-cells = <0>;
+			status = "okay";
+
+			poe@0x51 {
+				compatible = "raspberrypi,poe-core";
+				reg = <0x51>;
+				status = "okay";
+
+				poe_mfd_pwm: poe_pwm@0xf0 {
+					compatible = "raspberrypi,firmware-poe-pwm";
```
pelwell: This compatible string is no longer correct in the mfd case. I'm not saying it has to change, but using two different compatible strings might help (if only for error checking).

arch/arm/boot/dts/overlays/README
```diff
@@ -2858,6 +2858,8 @@ Params: poe_fan_temp0           Temperature (in millicelcius) at which the fan
                                 speeds up (default 55000)
         poe_fan_temp3_hyst      Temperature delta (in millicelcius) at which
                                 the fan slows down (default 5000)
+        i2c                     Control the fan via Linux I2C drivers instead of
+                                the firmware.
 
 
 Name:   rpi-poe-plus
```
pelwell, line 2865: `rpi-poe-plus` inherits the `i2c` parameter, so also needs a README addition.

drivers/pwm/pwm-raspberrypi-poe.c
```diff
@@ -52,34 +57,41 @@ static int raspberrypi_pwm_set_property(struct rpi_firmware *firmware,
...
-	if (msg.ret)
-		return -EIO;
+	if (pwm->firmware) {
+		ret = rpi_firmware_property(pwm->firmware, RPI_FIRMWARE_SET_POE_HAT_VAL,
+					    &msg, sizeof(msg));
+		if (!ret && msg.ret)
+			ret = -EIO;
+	} else {
+		ret = regmap_write(pwm->regmap, pwm->offset + reg, val);
+		if (ret)
```
pelwell: This is superfluous.

arch/arm/boot/dts/overlays/rpi-poe-overlay.dts
```diff
@@ -89,6 +79,52 @@
...
+	};
+
+	fragment@4 {
+		target = <&i2c0>;
+		__dormant__ {
+			#address-cells = <1>;
+			#size-cells = <0>;
+			status = "okay";
+
+			poe@0x51 {
```
pelwell: The preferred name is `poe@51`.

arch/arm/boot/dts/overlays/rpi-poe-overlay.dts
```diff
@@ -89,6 +79,52 @@
...
+				compatible = "raspberrypi,poe-core";
+				reg = <0x51>;
+				status = "okay";
+
+				poe_mfd_pwm: poe_pwm@0xf0 {
```
pelwell: The preferred name is `pwm@f0`.

pelwell (commented):
That's surprisingly compact. Minor comments only.

arch/arm/boot/dts/overlays/rpi-poe-overlay.dts
6by9: OK, will add a second compatible, even though it'll do exactly the same thing.

arch/arm/boot/dts/overlays/README
6by9: I was looking at PoE+ with the power driver as well, but yes.

arch/arm/boot/dts/overlays/rpi-poe-overlay.dts
6by9: Not sure what I was copying that used the 0x. Ho hum.

drivers/pwm/pwm-raspberrypi-poe.c
6by9: I'd already noted it in copying the same basis to the power driver. Done.

arch/arm/boot/dts/overlays/rpi-poe-overlay.dts
6by9: Reverted the name. I'd thought I could change the thing pointed to via __overlay__ / __dormant__ so was looking for one generic name, but that doesn't work.

6by9:
OK, I think this is good for a proper review.
The fw change to add the i2c param automatically is up as a merge request too.

drivers/pwm/pwm-raspberrypi-poe.c
pelwell: You missed this one.

6by9:
Updated to get rid of that duplicated ```return ret;``` (I can't find the comment for some reason).

arch/arm/boot/dts/overlays/rpi-poe-plus-overlay.dts
```diff
@@ -16,6 +16,21 @@
 			};
 		};
 	};
+	fragment@11 {
+		target = <&poe_mfd>;
+		__dormant__ {
+			rpi-poe-power-supply@f2 {
+				compatible = "raspberrypi,rpi-poe-power-supply";
+				reg = <0xf0>;
```
pelwell: Sorry - I just noticed this should be 0xf2.

6by9: Agreed, but then needs a slight rejig as RPI_POE_ADC_REG is currently 2, so that needs to be reset to 0, and FW needs a base offset to be added.

6by9: Done.

PR #4838, opened 2022-01-21
ARM: dts: BCM2711 AON_INTR2 generates IRQ edges

pelwell:
THe AON_INTR2 controller manages the HDMI interrupts, combining them
into a single interrupt passed to the GIC. bcm2711.dtsi declares the
interrupt as being IRQ_TYPE_LEVEL_HIGH, but it should be
IRQ_TYPE_EDGE_RISING. Most of the time the distinction shouldn't
matter, but there is a small possibility of losing interrupts unless
it is corrected.

See: http://lists.infradead.org/pipermail/linux-arm-kernel/2022-January/710292.html


popcornmix:
CEC seems fine after this change (as is general hdmi use).

PR #4841, opened 2022-01-22
Update rpi-display overlay

awatterott:
Hi,
This will update the rpi-display overlay with new pin state definitions needed for kernel 5.10+.


awatterott:
Thank you.

pelwell:
No problem - hopefully this is the last one.

PR #4845, opened 2022-01-25
media: i2c: imx219: Correct the minimum vblanking value

davidplowman:
The datasheet for this sensor documents the minimum vblanking as being
32 lines. It does fix some problems with occasional black lines at the
bottom of images (tested on Raspberry Pi).


davidplowman:
@naushir @6by9 

We should probably consider squeezing this into our very imminent release?

naushir:
That's what it says in the datasheet... LGTM.

naushir:
This probably ought to be submitted upstream as well.

PR #4846, opened 2022-01-25
vc4: Add zpos support

6by9:
Adds support for zpos to the KMS driver.
The frameworks didn't apply the stored default value to the planes when reset, hence a framework patch first.

@mripard Does this look like an acceptable solution that can be upstreamed?

### DIFF
--- a/drivers/gpu/drm/drm_atomic_state_helper.c
+++ b/drivers/gpu/drm/drm_atomic_state_helper.c
@@ -251,6 +251,20 @@ void __drm_atomic_helper_plane_state_reset(struct drm_plane_state *plane_state,
 	plane_state->alpha = DRM_BLEND_ALPHA_OPAQUE;
 	plane_state->pixel_blend_mode = DRM_MODE_BLEND_PREMULTI;
 
+	if (plane->color_encoding_property) {
+		if (!drm_object_property_get_default_value(&plane->base,
+							   plane->color_encoding_property,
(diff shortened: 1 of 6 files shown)


mripard:
Yes, it does :)

I'll rework it a bit to split it into several patches, but it'll work.

6by9:
OK, great. I'll leave it with you to split as required, and we'll backport it when you've posted that.

I didn't go through all the drivers we'd looked at to undo the setting of default values in their plane state_reset functions, but I guess that could follow on too.

6by9:
Rebased and updated to pick up mripard's upstream versions of the patches, plus the patch implementing zpos on vc4.

pelwell:
Does the `normalize_zpos` property guarantee that all zpos values are contiguous starting from 0?

6by9:
> Does the normalize_zpos property guarantee that all zpos values are contiguous starting from 0?

Yes.
https://elixir.bootlin.com/linux/latest/source/drivers/gpu/drm/drm_blend.c#L349
```
 * Drivers that attach a mutable zpos property to any plane should call the
 * drm_atomic_normalize_zpos() helper during their implementation of
 * &drm_mode_config_funcs.atomic_check(), which will update the normalized zpos
 * values and store them in &drm_plane_state.normalized_zpos. Usually min
 * should be set to 0 and max to maximal number of planes for given crtc - 1.
```

PR #4850, opened 2022-01-26
drm/vc4_hdmi: Add Broadcast RGB property to allow override of RGB range

popcornmix:
Copy Intel's "Broadcast RGB" property semantics to add manual override
of the HDMI pixel range for monitors that don't abide by the content
of the AVI Infoframe.

### DIFF
--- a/drivers/gpu/drm/vc4/vc4_hdmi.c
+++ b/drivers/gpu/drm/vc4/vc4_hdmi.c
@@ -58,6 +58,14 @@
 #include "vc4_hdmi_regs.h"
 #include "vc4_regs.h"
 
+/*
+ * "Broadcast RGB" property.
+ * Allows overriding of HDMI full or limited range RGB
+ */
+#define VC4_BROADCAST_RGB_AUTO 0
+#define VC4_BROADCAST_RGB_FULL 1
+#define VC4_BROADCAST_RGB_LIMITED 2
+
 #define VC5_HDMI_HORZA_HFP_SHIFT		16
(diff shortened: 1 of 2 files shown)


popcornmix:
@6by9 I've dug this out of https://github.com/raspberrypi/linux/pull/3535 as I suspect some users with brightness complaints for videos may be suffering from a display that doesn't handle limited or full range correctly.

It's changed shape a bit due to other restructuring.

popcornmix:
I'll leave notes here of what I tested.
```
systemctl stop kodi
modetest -M vc4 -w 32:"Broadcast RGB":1
systemctl start kodi
```
will force full range RGB with kodi (it does persist after kodi changes hdmi mode).
You can change the 1 to a 2 to force limited range.

6by9:
May as well merge it.
Memory says that later kernels have reworked the code that actually sets the range, so it may not cherry-pick cleanly.

popcornmix:
Conflicts are only trivial (debug_fs registers are declared in different place in header).

drivers/gpu/drm/vc4/vc4_hdmi.c
```diff
@@ -58,6 +58,13 @@
 #include "vc4_hdmi_regs.h"
 #include "vc4_regs.h"
 
+/* "Broadcast RGB" property.
```
pelwell: The preferred format for a multiline comments is to have the `/*` and `*/` on lines of their own.

pelwell:
Cherry-picked to 5.15, 5.16 and 5.17.

PR #4851, opened 2022-01-27
drm/vc4: plane: Switch to upstream version of DRM_FORMAT_P030 support

popcornmix:
The existing code could potentially access uninitialised i and tile_w.
This has been fixed in upstreamed version of this file, so revert and pick that.


popcornmix:
Note: 5.15, 5.16 and 5.17 have already switched to upstream version, so this doesn't need cherry-picking.

PR #4852, opened 2022-01-28
vc4/drm: Add support for fractional scaling and chroma siting

popcornmix:
When chroma is subsampled compared to luma, there are a number of choices of where the chroma is centred.
Different codecs make different choices for this. From ffmpeg:
```
 /**
  * Location of chroma samples.
  *
  * Illustration showing the location of the first (top left) chroma sample of the
  * image, the left shows only luma, the right
  * shows the location of the chroma sample, the 2 could be imagined to overlay
  * each other but are drawn separately due to limitations of ASCII
  *
  *                1st 2nd       1st 2nd horizontal luma sample positions
  *                 v   v         v   v
  *                 ______        ______
  *1st luma line > |X   X ...    |3 4 X ...     X are luma samples,
  *                |             |1 2           1-6 are possible chroma positions
  *2nd luma line > |X   X ...    |5 6 X ...     0 is undefined/unknown position
  */
 enum AVChromaLocation {
     AVCHROMA_LOC_UNSPECIFIED = 0,
     AVCHROMA_LOC_LEFT        = 1, ///< MPEG-2/4 4:2:0, H.264 default for 4:2:0
     AVCHROMA_LOC_CENTER      = 2, ///< MPEG-1 4:2:0, JPEG 4:2:0, H.263 4:2:0
     AVCHROMA_LOC_TOPLEFT     = 3, ///< ITU-R 601, SMPTE 274M 296M S314M(DV 4:1:1), mpeg2 4:2:2
     AVCHROMA_LOC_TOP         = 4,
     AVCHROMA_LOC_BOTTOMLEFT  = 5,
     AVCHROMA_LOC_BOTTOM      = 6,
     AVCHROMA_LOC_NB,              ///< Not part of ABI
 };
  ```
To accurately display decoded video form these different sources, we need to handle this at display end.

Also DRM provides fractional source coordinates for displayed planes.

Both of these can be handled using the phase field of HVS.

### CODE REVIEW

**Comment by popcornmix** [comment]
There are alternatives for the CHROMA_SITING DRM property.
I've gone for an enum that matches ffmpeg. You could make this completely generic
e.g. the [0..256] fractional number for each of horizontal and vertical offsets I convert this into.
That matches the parameters `out_h_chr_pos`/`out_v_chr_pos` in ffmpeg's scale video filter.
But settings outside of ffmpeg's enums are likely to be very rare.

File: drivers/gpu/drm/vc4/vc4_plane.c [hunk 1]
```diff
@@ -1589,6 +1589,16 @@ struct drm_plane *vc4_plane_init(struct drm_device *dev,
 					  DRM_COLOR_YCBCR_BT709,
 					  DRM_COLOR_YCBCR_LIMITED_RANGE);
 
+	drm_plane_create_chroma_siting_properties(plane,
```
6by9, line 1622: This block should be in the 4th patch adding it to vc4, not in the 3rd patch introducing the property.

include/drm/drm_color_mgmt.h
```diff
@@ -87,12 +87,42 @@ enum drm_color_range {
...
+ * each other but are drawn separately due to limitations of ASCII
+ *
+ *                 1st 2nd       1st 2nd horizontal luma sample positions
+ *                  v   v         v   v
+ *                 ______        ______
+ * 1st luma line > |X   X ...    |3 4 X ...     X are luma samples,
+ *                 |             |1 2           1-6 are possible chroma positions
+ * 2nd luma line > |X   X ...    |5 6 X ...     0 is undefined/unknown position
+ */
+enum drm_chroma_siting {
```
6by9, line 105: Chroma siting is applicable for YUV422 as well, but only in the horizontal direction. Contrary to what I said on Slack, we probably want 2 properties so that it is clear that vertical is only applicable to 420, and both apply to 422.
This is going to be one to discuss with mainline though, along with the requirement to have an open source project having support patches reviewed but not merged for any new properties.

drivers/gpu/drm/vc4/vc4_plane.c
```diff
@@ -520,6 +521,33 @@ static u32 vc4_lbm_size(struct drm_plane_state *state)
...
+			return is_horizontal ? 128:128;
+		case DRM_CHROMA_SITING_TOPLEFT:
+			return is_horizontal ? 0:0;
+		case DRM_CHROMA_SITING_TOP:
+			return is_horizontal ? 128:0;
+		case DRM_CHROMA_SITING_BOTTOMLEFT:
+			return is_horizontal ? 0:256;
+		case DRM_CHROMA_SITING_BOTTOM:
+			return is_horizontal ? 128:256;
+	}
```
6by9, line 548: 2D array of [DRM_CHROMA_SITING_MAX][2] instead of switch/case and then ternary expression?

6by9 (commented):
I'll trust you on the actual phase calcs.

Signed-of-by tags and commit author need to be real names, and commits should have a full description.

popcornmix:
Superseded by https://github.com/raspberrypi/linux/pull/4937

PR #4854, opened 2022-01-29
i2c: bcm2835: Make clock-stretch timeout configurable

crawford:
The default clock-stretch timeout is 35 mS, which works well for
SMBus, but there are some I2C devices which can stretch the clock even
longer. Rather than trying to prescribe a safe default for everyone,
allow the timeout to be configured.

Signed-off-by: Alex Crawford <raspberrypi/linux@code.acrawford.com>


crawford:
Accidentally opened this against 5.4.y. See https://github.com/raspberrypi/linux/pull/4855 instead.

PR #4855, opened 2022-01-29
i2c: bcm2835: Make clock-stretch timeout configurable

crawford:
The default clock-stretch timeout is 35 mS, which works well for
SMBus, but there are some I2C devices which can stretch the clock even
longer. Rather than trying to prescribe a safe default for everyone,
allow the timeout to be configured.

Signed-off-by: Alex Crawford <raspberrypi/linux@code.acrawford.com>

### DIFF
--- a/drivers/i2c/busses/i2c-bcm2835.c
+++ b/drivers/i2c/busses/i2c-bcm2835.c
@@ -55,6 +55,10 @@ static unsigned int debug;
 module_param(debug, uint, 0644);
 MODULE_PARM_DESC(debug, "1=err, 2=isr, 3=xfer");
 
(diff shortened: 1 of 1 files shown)


pelwell:
Thanks - we should probably add an override from Device Tree as well.

crawford:
I think a misunderstood the branching strategy. How do I ensure this patch continues being applied to newer branches?

pelwell:
It's already been copied over, although it's been squashed: https://github.com/raspberrypi/linux/blob/rpi-5.15.y/drivers/i2c/busses/i2c-bcm2835.c#L58

PR #4856, opened 2022-01-29
Patching lan78xx for SOF_TIMESTAMPING_TX_SOFTWARE support

tiagofreire-pt:
This patch to the Raspberry 3B+ LAN78xx kernel driver enables the support for _SOF_TIMESTAMPING_TX_SOFTWARE_.

Aiming the software-transmit capability for PTP:

```
 ethtool -T eth0
Time stamping parameters for eth0:
Capabilities:
	software-transmit
	software-receive
	software-system-clock
PTP Hardware Clock: none
Hardware Transmit Timestamp Modes: none
Hardware Receive Filter Modes: none
```


pelwell:
Any objections, @P33M ?

P33M:
No, but I would be sceptical of enabling software timestamping on a network device connected via USB - there's no guarantee that bulk packets are sent at any consistent interval from when the URBs are submitted. You'd probably get performance better than NTP, but not by much.

PR #4859, opened 2022-01-31
drm/vc4: Add DRM 210101010 RGB formats for hvs5.

6by9:
HVS5 supports the 210101010 RGB[A|X] formats, but they were
missing from the DRM to HVS mapping list, so weren't available.
Add them in.

### DIFF
--- a/drivers/gpu/drm/vc4/vc4_plane.c
+++ b/drivers/gpu/drm/vc4/vc4_plane.c
@@ -135,6 +135,34 @@ static const struct hvs_format {
 		.pixel_order = HVS_PIXEL_ORDER_XYCBCR,
 		.hvs5_only = true,
 	},
+	{
+		.drm = DRM_FORMAT_XRGB2101010,
+		.hvs = HVS_PIXEL_FORMAT_RGBA1010102,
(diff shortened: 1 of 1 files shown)


6by9:
@jc-kynesim

popcornmix:
What should I run to test this? I tried:
```
modetest -M vc4 -F tiles,gradient -s 32:1920x1080-60 -P99@82:3840x2160+0+0@XR30
```
But the dlist shows:
```
pixel_format (04-00): value: 0x00000007 decimal:          7 mask:(0x00000007) bits(0x0000001f) 'RGBA:8:8:8:8'
```
so I think this format is not getting through.

popcornmix:
Scratch that - there's a second layer that looks right.
```
                            pixel_format (04-00): value: 0x00000010 decimal:         16 mask:(0x00000010) bits(0x0000001f) 'RGB:10:10:2'
```

6by9:
```modetest -M vc4 -P 80@88:1280x800@AB30```
was my test, with also AR30, XR30, and XB30.

jc-kynesim:
Well I can now drive AR30 from my drm test prog and I can confirm that alpha works and colours are where I expect them.
Also works fine if the output monitor can't do 10 bit

PR #4860, opened 2022-01-31
vc4-kms-dpi overlay updates

6by9:
Cherry-picks 2 commits back from rpi-5.15.y.

Splits out all the common parts of the vc4-kms-dpi overlays.


6by9:
Mods to support #4853. 
I need to verify that these still work.

@aBUGSworstnightmare-rpi Could you verify that ```dtoverlay=vc4-kms-dpi-panel,at056tn53v1``` still works for your panel too?

6by9:
Updated to move the Adafruit Kippah 7inch display into vc4-kms-dpi-panel, and deprecate vc4-kms-kippah-7inch.

arch/arm/boot/dts/overlays/README
```diff
@@ -3687,10 +3732,8 @@ Params: <None>
 
 
 Name:   vc4-kms-kippah-7inch
-Info:   Enable the Adafruit DPI Kippah with the 7" Ontat panel attached.
-        Requires vc4-kms-v3d to be loaded.
-Load:   dtoverlay=vc4-kms-kippah-7inch
-Params: <None>
+Info:   This overlay is now deprecated - see i2c-sensor
```
pelwell: See vc4-kms-dpi-panel, surely.

aBUGSworstnightmare-rpi:
Will have a look at it now. Will use your github branch and compile it (as
I'm not sure what's on the jig which has a display mounted atm).
Will report back later this morning.

Joerg

Am Mo., 31. Jan. 2022 um 18:48 Uhr schrieb 6by9 ***@***.***>:

aBUGSworstnightmare-rpi:
@6by9 Tested Innolux 5.6in with GPIO-backlight. Adding

```
dtoverlay=vc4-kms-dpi-panel
dtparam=at056tn53v1,backlight-gpio=19
```

to config.txt gives below —> **GPIO BL confirmed working**.

```
pi@test1:~ $ sudo vcdbg log msg…
006747.398: brfs: File read: /mfs/sd/overlays/vc4-kms-v3d-pi4.dtbo
006811.768: Loaded overlay 'vc4-kms-v3d'
006811.781: dtparam: noaudio=true
006812.900: dtparam: ant2=true
007019.521: brfs: File read: 3823 bytes
007034.703: brfs: File read: /mfs/sd/overlays/vc4-kms-dpi-panel.dtbo
007062.568: Loaded overlay 'vc4-kms-dpi-panel'
007062.599: dtparam: at056tn53v1=true
007063.452: dtparam: backlight-gpio=19
007130.576: brfs: File read: 4287 bytes
…
pi@test1:~ $ 
```

Tested Innolux 5.6in with PWM-backlight. Adding

```
dtoverlay=vc4-kms-dpi-panel,at056tn53v1
dtparam=backlight-pwm=1,backlight-pwm-chan=1,backlight-pwm-gpio=19
```

To config.txt gives below
```
pi@test1:~ $ sudo vcdbg log msg
…
007031.759: brfs: File read: /mfs/sd/overlays/vc4-kms-dpi-panel.dtbo
007059.670: Loaded overlay 'vc4-kms-dpi-panel'
007059.683: dtparam: at056tn53v1=true
007060.564: dtparam: backlight-pwm=1
007061.895: dtparam: backlight-pwm-chan=1
007062.510: dtparam: backlight-pwm-gpio=19
007157.705: brfs: File read: 4287 bytes
…
pi@test1:~ $ raspi-gpio get
BANK0 (GPIO 0 to 27):
…
GPIO 19: level=1 fsel=2 alt=5 func=PWM0_1 pull=DOWN
..
pi@test1:~ $ 
```

GPIO19 is configured for PWM operation with channel1, but has a pull-down active. Shouldn’t that be no-pull for proper operation?
Change of Fragment 107 as shown below
```
	fragment@107 {
		target = <&gpio>;
		__dormant__ {
			pwm_pins: pwm_pins {
				brcm,pins = <18>;
				brcm,function = <2>; /* Alt5 */
				brcm,pull = <0>;
			};
		};
	};
```

Backlight is working though, tested as outlined below.

```
pi@test1:~ $ sudo -i

SSH is enabled and the default password for the 'pi' user has not been changed.
This is a security risk - please login as the 'pi' user and type 'passwd' to set a new password.

root@test1:~# echo 0  > /sys/class/backlight/backlight/brightness
-bash: /sys/class/backlight/backlight/brightness: No such file or directory
root@test1:~# echo 0  > /sys/class/backlight/backlight_pwm/brightness
root@test1:~# echo 3  > /sys/class/backlight/backlight_pwm/brightness
root@test1:~# echo 16  > /sys/class/backlight/backlight_pwm/brightness
-bash: echo: write error: Invalid argument
root@test1:~# echo 15  > /sys/class/backlight/backlight_pwm/brightness
root@test1:~# echo 10  > /sys/class/backlight/backlight_pwm/brightness
root@test1:~# 
```

arch/arm/boot/dts/overlays/README
6by9: :-( Guess where the copy/paste came from! Will fix it up when I'm in the office.

6by9:
@aBUGSworstnightmare-rpi Thanks for testing. That sounds good and working as intended.
If you don't mind, then I'm going to mark vc4-kms-dpi-at056tn53v1 as deprecated, and to use vc4-kms-dpi-panel,at056tn53v1 instead.

Pulls are largely irrelevant for any pin defined as an output or alt function which drives the pin. 
As you say, adding ```brcm,pull = <0>;``` to the ```pwm_pins``` node of vc4-kms-dpi.dtsi should solve it anyway.

I tested with the Adafruit Kippah 7inch panel, and that was fine too, so other than fixing up my copy/paste error, I think we're good to go.

@pelwell Just checking you are happy with this approach of one fragment per panel to set the compatible, rather than setting it explicitly from the override line?

aBUGSworstnightmare-rpi:
Feel free doing so! Makes things more simple.
Although I need to admit that novice users might be struggling with how to configure the overrides for PWM (as one needs to know that PWM on GPIO19 is channel 1 of the BRCM PWM module).

6by9:
There's nothing stopping a panel override also configuring the backlight properties as well. Adafruit's Kippah has none, and vc4-kms-dpi-at056tn53v1 didn't define one either.

Untested, but make your override
```
at056tn53v1 = <0>, "+0+90",
	  <0>, "+104+105+106+107+108",
	  <&backlight_pwm>, "pwms:4=1",
	  <&pwm_pins>, "brcm,pins:0=19";
```
and it should configure panel and backlight in one hit.

pelwell:
>  Just checking you are happy with this approach of one fragment per panel to set the compatible, rather than setting it explicitly from the override line?

This is fine. I'm not sure you could override a multi-string property like the `compatible`s here without going to a hex byte string approach, which is just horrible.

What I'm not _so_ keen on is deprecating the old overlays immediately in a way that means it ceases to work. I'm thinking particularly about the kippah overlay. I've realised that it shouldn't be hard to extend the firmware so that a `rename` in `overlay_map.dtb` could take a few parameters. It wouldn't cope with change parameters, but I imagine that the _specific overlay_ to _parameter of a generic overlay_ is actually relatively common conversion.

aBUGSworstnightmare-rpi:
> There's nothing stopping a panel override also configuring the backlight properties as well. Adafruit's Kippah has none, and vc4-kms-dpi-at056tn53v1 didn't define one either.
> 
When I did the pull-request for 'vc4-kms-dpi-at056tn53v1-overlay.dts' backlight PWM was not enabled in the default kernel. So - in order to avoid issues - no PWM backlight node was added to the overlay.

6by9:
> What I'm not so keen on is deprecating the old overlays immediately in a way that means it ceases to work. I'm thinking particularly about the kippah overlay. I've realised that it shouldn't be hard to extend the firmware so that a rename in overlay_map.dtb could take a few parameters. It wouldn't cope with change parameters, but I imagine that the specific overlay to parameter of a generic overlay is actually relatively common conversion.

OK, I'll drop the deprecation the old Kippah overlay for now. If we get a mechanism to rename in future then that would be great.
As aBUGSworstnightmare-rpi is the only one using at056tn53v1, and he's OK with deprecating it, I'll do that one.

pelwell:
Cool - that's what I was going to suggest.

6by9:
OK, tested again with vc4-kms-kippah-7inch and vc4-kms-dpi-panel,kippah-7inch, and both work.

rotate and backlight for at056tn53v1 can be added at a later date.

pelwell:
The most recent rpi-update firmware adds support for "rename with added parameters", making it possible to soft-deprecate the old `vc4-kms-kippah-7inch` overlay.

Issue #4818, opened 2022-01-10
Request to fix compilation errors in downstream patches reported by clang


rsglobal:
### Describe the bug

Dear RPI community,  

Please see diff below.  
Please enable clang build in CI to avoid such cases in the future.

### Steps to reproduce the behaviour

Build v5.15 branch using clang

### Device (s)

Raspberry Pi 4 Mod. B

### System

Local build

### Additional context

```diff
diff --git a/drivers/gpu/drm/vc4/vc4_plane.c b/drivers/gpu/drm/vc4/vc4_plane.c
index b323d18b51199..4b51130427efa 100644
--- a/drivers/gpu/drm/vc4/vc4_plane.c
+++ b/drivers/gpu/drm/vc4/vc4_plane.c
@@ -686,7 +686,7 @@ static int vc4_plane_mode_set(struct drm_plane *plane,
 	u32 tiling, src_y;
 	u32 hvs_format = format->hvs;
 	unsigned int rotation;
-	int ret, i;
+	int ret, i = 0;
 
 	if (vc4_state->dlist_initialized)
 		return 0;
@@ -811,7 +811,7 @@ static int vc4_plane_mode_set(struct drm_plane *plane,
 	case DRM_FORMAT_MOD_BROADCOM_SAND128:
 	case DRM_FORMAT_MOD_BROADCOM_SAND256: {
 		uint32_t param = fourcc_mod_broadcom_param(fb->modifier);
-		u32 tile_w, tile, x_off, pix_per_tile;
+		u32 tile_w = 1, tile, x_off, pix_per_tile;
 
 		if (fb->format->format == DRM_FORMAT_P030) {
 			/*
diff --git a/drivers/media/i2c/ov9281.c b/drivers/media/i2c/ov9281.c
index 7d939128899b9..0a5f9e0927aec 100644
--- a/drivers/media/i2c/ov9281.c
+++ b/drivers/media/i2c/ov9281.c
@@ -1124,7 +1124,7 @@ static int ov9281_check_sensor_id(struct ov9281 *ov9281,
 				  struct i2c_client *client)
 {
 	struct device *dev = &ov9281->client->dev;
-	u32 id = 0, id_msb;
+	u32 id = 0, id_msb = 0;
 	int ret;
 
 	ret = ov9281_read_reg(client, OV9281_REG_CHIP_ID + 1,
[...]

PR #4861, opened 2022-02-01
[rpi-5.15.y] Fix clang warnings

nathanchance:
This series cleans up the clang warnings that I see with `ARCH=arm bcm2709_defconfig`, `ARCH=arm bcm2711_defconfig`, and `ARCH=arm64 bcm2711_defconfig`. The warnings are in the commit messages for references.

Most of these fixes line up with the ones suggested in https://github.com/raspberrypi/linux/issues/4818 but I only looked at the full issue after resolving them so that I was only influenced by the warning text and current code.

Closes https://github.com/raspberrypi/linux/issues/4818.

### DIFF
--- a/drivers/media/i2c/ov9281.c
+++ b/drivers/media/i2c/ov9281.c
@@ -1124,7 +1124,7 @@ static int ov9281_check_sensor_id(struct ov9281 *ov9281,
 				  struct i2c_client *client)
 {
 	struct device *dev = &ov9281->client->dev;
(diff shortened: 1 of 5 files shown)


popcornmix:
These all look correct to me. First 3 have no functional change in behaviour.
The unicam one looks like a bug fix in error return value.
The ov9281 one only affects a dmesg error (the incorrect id printed).

pelwell:
Thanks!

PR #4862, opened 2022-02-01
DRM: Clean up handling of panel orientation

6by9:
@mripard to review please.

Noted on #4853

### DIFF
--- a/drivers/gpu/drm/drm_panel.c
+++ b/drivers/gpu/drm/drm_panel.c
@@ -61,6 +61,9 @@ void drm_panel_init(struct drm_panel *panel, struct device *dev,
 	panel->dev = dev;
 	panel->funcs = funcs;
 	panel->connector_type = connector_type;
+
+	panel->orientation = DRM_MODE_PANEL_ORIENTATION_UNKNOWN;
+	of_drm_get_panel_orientation(dev->of_node, &panel->orientation);
 }
 EXPORT_SYMBOL(drm_panel_init);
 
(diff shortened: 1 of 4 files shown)


drivers/gpu/drm/drm_panel.c
```diff
@@ -61,6 +61,21 @@ void drm_panel_init(struct drm_panel *panel, struct device *dev,
...
+		ret = of_drm_get_panel_orientation(dev->of_node,
+						   &panel->orientation);
+		if (ret) {
+			DRM_DEV_ERROR(panel->dev, "%pOF: failed to get orientation %d\n",
+				      dev->of_node, ret);
+			panel->orientation = DRM_MODE_PANEL_ORIENTATION_UNKNOWN;
+		}
+	}
+#else
+	panel->orientation = DRM_MODE_PANEL_ORIENTATION_UNKNOWN;
```
pelwell, line 65: I'm not sure what upstream would prefer, but I think the code would look cleaner with `panel->orientation` initialised unconditionally, then possibly changed by `of_drm_get_panel_orientation`. By the way, `of_drm_get_panel_orientation` won't return an error if there is no orientation property, making the error message here likely to only appear for invalid rotation values.

6by9: This is why I've pinged Maxime.

I'm aware that ```of_drm_get_panel_orientation``` doesn't treat the parameter being missing as an error, so yes it is only going to be invalid rotations.
The implementation previously in panel-simple was to totally fail on an invalid rotation, but I can't replicate that here.

mripard: I agree with @pelwell, we can just have 
```C
panel->orientation = DRM_MODE_PANEL_ORIENTATION_UNKNOWN;
of_drm_get_panel_orientation(dev->of_node, &panel->orientation);
```

of_drm_get_panel_orientation will return ENODEV if CONFIG_OF is not set, so I think we can just ignore its return code, in which case we will have an unknown rotatation. The only drawback I can see is that we won't log anything in that case, but maybe we can add that error message in the function itself?

6by9: Done.

PR #4868, opened 2022-02-02
Add Support for Geekworm MZP280 in vc4-dpi-panel-overlay

macromorgan:
This pull request is to add support for the mzp280 panel in the generic vc4-dpi-panel-overlay that was merged yesterday. Note that since this panel is rotated 270 (depending upon your application anyway) rotation support is also added to the vc4-dpi-panel overlay.

### DIFF
--- a/arch/arm/boot/dts/overlays/README
+++ b/arch/arm/boot/dts/overlays/README
@@ -3656,6 +3656,7 @@ Params: clock-frequency         Display clock frequency (Hz)
         backlight-def-brightness
                                 Set the default brightness. Normal range 1-16.
                                 (default 16).
+        rotate                  Display rotation {0,90,180,270} (default 0)
 
 
 Name:   vc4-kms-dpi-panel
@@ -3683,6 +3684,7 @@ Params: at056tn53v1             Enable an Innolux 5.6in VGA TFT
         backlight-def-brightness
                                 Set the default brightness. Normal range 1-16.
                                 (default 16).
+        rotate                  Display rotation {0,90,180,270} (default 0)
 
 
 Name:   vc4-kms-dsi-7inch
--- a/arch/arm/boot/dts/overlays/README
+++ b/arch/arm/boot/dts/overlays/README
@@ -3665,6 +3665,7 @@ Info:   Enable a preconfigured KMS DPI panel.
 Load:   dtoverlay=vc4-kms-dpi-panel,<param>=<val>
 Params: at056tn53v1             Enable an Innolux 5.6in VGA TFT
         kippah-7inch            Enable an Adafruit Kippah with 7inch panel.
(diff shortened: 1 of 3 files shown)


arch/arm/boot/dts/overlays/README
```diff
@@ -3683,6 +3683,7 @@ Params: at056tn53v1             Enable an Innolux 5.6in VGA TFT
         backlight-def-brightness
                                 Set the default brightness. Normal range 1-16.
                                 (default 16).
+        rotate                  Display rotation {0,90,180,270} (default 0)
```
6by9, line 3688: This also needs to go in the README for vc4-kms-dpi-generic and vc4-kms-kippah-7inch because they also include vc4-kms-dpi.dtsi.

Otherwise looks fine to me.

macromorgan: vc4-kms-kippah-7inch seems to be depreciated in favor of vc4-kms-dpi-panel. I'll update vc4-kms-dpi-generic and resubmit though, thank you.

macromorgan: Nevermind, I understand. Even though it's depreciated it's still a valid devicetree (for now). I'll update it too, thank you.

macromorgan: Fixed and force pushed a new update.

pelwell: You don't need to add it for vc4-kms-kippah-7inch - the deprecation (added very recently so 6by9 probably hadn't seen it) means the parameters aren't listed.

6by9: I'd seen your firmware mod to support renaming to a new overlay with params, but hadn't seen you'd updated the README. Sorry, my bad.

macromorgan: Okay, so to confirm I need to 1) add it to the vc4-kms-dpi-generic, but not to the vc4-kms-kippah-7inch, right?

Also, it's not a big deal but I see a warning about using a dummy regulator. Should I just add for this specific device a power-supply (would point it to whatever the main 5v one is, since that what this runs off of)?

6by9: Correct. Add to vc4-kms-dpi-generic, but not to the vc4-kms-kippah-7inch.

Messages about dummy regulators are almost inevitable. You can create a regulator with ```compatible = "regulator-fixed";``` if you really felt like it, but it's not really necessary.

macromorgan: I won't worry about fixed regulators then for now (I tried it and there's no gain really aside from one less thing in dmesg).

I just updated the pull with the README fixes. Let me know if there is anything else that's needed. Thank you.

Issue #4848, opened 2022-01-25
please fix ahci driver when in pcie-32bit-dma mode


FrancescoRestelli:
### Describe the bug

when setting dtoverlay=pcie-32bit-dma (which is necessary for some dvb cards i am playing with) the ahci driver does not work and crashes

### Steps to reproduce the behaviour

install ASM1064 based sata card with a connected ssd
set dtoverlay=pcie-32bit-dma in /boot/config.txt
reboot

dmesg shows errors
lspci crashes

no /dev/sda* exists

### Device (s)

Raspberry Pi CM4

### System

os:
pi@raspberrypi:~ $ cat /etc/rpi-issue
Raspberry Pi reference 2021-10-30
Generated using pi-gen, https://github.com/RPi-Distro/pi-gen, 288b21fc27e128ea6b330777aca68e0061ebf4fe, stage4

firmware:
Jan 20 2022 13:56:48
Copyright (c) 2012 Broadcom
version bd88f66f8952d34e4e0613a85c7a6d3da49e13e2 (clean) (release) (start)

uname:
Linux raspberrypi 5.10.92-v7l+ #1 SMP Tue Jan 25 18:46:39 GMT 2022 armv7l GNU/Linux

### Logs

[   12.771119] ata1.00: qc timeout (cmd 0xec)
[   12.911545] 8<--- cut here ---
[   12.914604] Unhandled fault: asynchronous external abort (0x1211) at 0x00000000
[   12.921915] pgd = 7ed806c6
[   12.924617] [00000000] *pgd=042cc003, *pmd=1fd26d003
[   12.976726] 8<--- cut here ---
[   12.979781] Unhandled fault: asynchronous external abort (0x1211) at 0x00000000
[   12.987083] pgd = bc4626ff
[   12.989783] [00000000] *pgd=80000000004003, *pmd=00000000
[   12.995185] Internal error: : 1211 [#1] SMP ARM
[...]

PR #4870, opened 2022-02-03
dtoverlay: Reduce size of PCIE IB window in pcie-32-dma overlay

6by9:
The PCIE inbound window is rounded up to a power of 2, so the default
of 3GB rounds up to 4GB starting at 0. This prohibits the MSI vector
sitting at 0x0_fffffffc, and causes warnings from some subsystems
(eg ahci) of a 64bit address on a 32bit configuration.

Reduce the window down to 2GB to avoid this issue.

https://github.com/raspberrypi/linux/issues/4848

### DIFF
--- a/arch/arm/boot/dts/overlays/pcie-32bit-dma-overlay.dts
+++ b/arch/arm/boot/dts/overlays/pcie-32bit-dma-overlay.dts
@@ -11,7 +11,25 @@
 	fragment@0 {
 		target-path = "/aliases";
 		__overlay__ {
+			/*
+			 * Removing this alias stops the firmware patching the
+			 * PCIE DT dma-ranges based on the detected chip
+			 * revision.
+			 */
 			pcie0 = "";
+		};
+	};
+
+	fragment@1 {
+		target = <&pcie0>;
+		__overlay__ {
+			/*
+			 * The size of the range is rounded up to a power of 2,
+			 * so the range ends up being 0-4GB, and the MSI vector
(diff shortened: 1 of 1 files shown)


pelwell:
LGTM. While you're in there, could you add a comment to fragment 0 saying that it prevents the firmware from rewriting the dma-ranges property?

6by9:
> While you're in there, could you add a comment to fragment 0 saying that it prevents the firmware from rewriting the dma-ranges property?

Done.

pelwell:
Ta.

6by9:
Blast. That's thrown up a dtc warning that I hadn't noticed.
```
..../linux/arch/arm/boot/dts/overlays/pcie-32bit-dma-overlay.dts:26.4-27.27: Warning (dma_ranges_format): /fragment@1/__overlay__:dma-ranges: "dma-ranges" property has invalid length (28 bytes) (parent #address-cells == 2, child #address-cells == 2, #size-cells == 1)
```

6by9:
Adding
```
			#address-cells = <3>;
			#size-cells = <2>;
```
(copied from the pcie0 node) to the fragment clears the warnings, but is that the right answer?

pelwell:
Yes, that's the right answer because it's the only method I know of. It has the unfortunate side effect of overwriting properties that we shouldn't need to overwrite, but if the values didn't match the base dtb then the dma-ranges would be of the wrong format anyway (which is what the warning is about).

PR #4873, opened 2022-02-04
media: bcm2835-unicam: Handle a repeated frame start with no end

6by9:
In the case of 2 frame starts being received with no frame end
between, the queued buffer held in next_frm was lost as the
pointer was overwritten with the dummy buffer.


6by9:
@naushir As discussed earlier.

6by9:
Hang on, I've just had a kernel panic in ```unicam_isr```, so something is still not right.

6by9:
Updated to check that cur_frm isn't NULL before putting it back on the list.

PR #4877, opened 2022-02-07
overlays: rpi-poe(-plus): Fix parameters

pelwell:
The support of PoE HATs is complicated because there are two methods to
control them - via the firmware, and directly over I2C - the choice
between the two methods depending on which side "owns" the i2c0 bus.

The firmware determines the ownership of i2c0 by analysing the Device
Tree after applying any dtoverlays and dtparams. For this reason the
PoE HAT overlays have up to now been applied by the firmware _after_
config.txt has been processed, but this has prevented any user-supplied
threshold settings from being applied - either because the parameters
are unknown or because (after an explicit dtoverlay=rpi-poe) the
firmware's reload of the overlay has overwritten the user's settings.

Solve the problem by restructuring the rpi-poe overlays to support
a mode switch _after_ the overlay has been merged, thereby allowing
the overlays to be loaded _before_ config.txt is processed.

There is a companion firmware patch that changes the point at which
the overlays are loaded, and the parameter used to switch modes, but
the updated overlays are still compatible with older firmware, albeit
without working parameters.

See: https://github.com/raspberrypi/firmware/issues/1689

### DIFF
--- a/arch/arm/boot/dts/overlays/rpi-poe-overlay.dts
+++ b/arch/arm/boot/dts/overlays/rpi-poe-overlay.dts
@@ -67,7 +67,7 @@
 
 	fragment@2 {
 		target-path = "/__overrides__";
-		__overlay__ {
+		params: __overlay__ {
 			poe_fan_temp0 =		<&trip0>,"temperature:0";
 			poe_fan_temp0_hyst =	<&trip0>,"hysteresis:0";
 			poe_fan_temp1 =		<&trip1>,"temperature:0";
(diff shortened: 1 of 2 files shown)


pelwell:
> For this reason the PoE HAT overlays have up to now been applied by the firmware _after_ config.txt has been processed,

This could have been better worded - the overlays were originally applied _before_ config.txt, but that changed with 29479a7850c1.

pelwell:
Does this look OK to you, @6by9? Note that I avoid having to individually dis/enable the rpi-poe-power-supply node by toggling the status of the poe_mfd parent instead.

PR #4878, opened 2022-02-07
Change vc4 DSI to being a bridge

6by9:
Bridges and the atomic API are the current approach.

Creating this PR as I need to send the framework patches upstream.

### DIFF
--- a/drivers/gpu/drm/drm_bridge.c
+++ b/drivers/gpu/drm/drm_bridge.c
@@ -462,21 +462,58 @@ EXPORT_SYMBOL(drm_bridge_chain_disable);
  * Calls &drm_bridge_funcs.post_disable op for all the bridges in the
  * encoder chain, starting from the first bridge to the last. These are called
  * after completing the encoder's prepare op.
+ * If a bridge sets the DRM_BRIDGE_OP_UPSTREAM_FIRST, then the post_disable for
+ * that bridge will be called before the previous one to reverse the pre_enable
+ * calling direction.
  *
  * Note: the bridge passed should be the one closest to the encoder
  */
 void drm_bridge_chain_post_disable(struct drm_bridge *bridge)
 {
 	struct drm_encoder *encoder;
+	struct drm_bridge *next, *limit;
 
 	if (!bridge)
 		return;
 
 	encoder = bridge->encoder;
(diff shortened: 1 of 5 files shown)


6by9:
These patches seem pretty solid now. Tested on CM4, Pi4, Pi3 with 7" panel, and CM4 with SN65DSI83 DSI->LVDS bridge.
I'll be sending the main framework changes to dri-devel next week, but these should be at a place that is worth a review.

drivers/gpu/drm/drm_bridge.c
```diff
@@ -462,21 +462,58 @@ EXPORT_SYMBOL(drm_bridge_chain_disable);
...
 	list_for_each_entry_from(bridge, &encoder->bridge_chain, chain_node) {
+		limit = NULL;
+
+		if (!list_is_last(&bridge->chain_node, &encoder->bridge_chain)) {
+			next = list_next_entry(bridge, chain_node);
+
+			if (next->ops & DRM_BRIDGE_OP_UPSTREAM_FIRST) {
+				limit = next;
+
+				list_for_each_entry_from(next, &encoder->bridge_chain,
```
pelwell, line 489: This loop would benefit from at least a single explanatory comment.

drivers/gpu/drm/drm_bridge.c
```diff
@@ -617,37 +704,79 @@ EXPORT_SYMBOL(drm_atomic_bridge_chain_disable);
...
 
 	if (!bridge)
 		return;
 
 	encoder = bridge->encoder;
+
 	list_for_each_entry_from(bridge, &encoder->bridge_chain, chain_node) {
-		if (bridge->funcs->atomic_post_disable) {
-			struct drm_bridge_state *old_bridge_state;
+		limit = NULL;
```
pelwell, line 725: This all looks very familiar - it's a shame the non-atomic case can't be a handled as a special case of the atomic API.

pelwell (commented):
This is definitely TL;DR for a topic outside my experience. A few strategic comments wouldn't hurt. And I suppose there is no elegant way to avoid the duplication between the atomic and non-atomic APIs without introducing more framework.

pelwell:
This does not apply cleanly to 5.16 - can you fix it up?

PR #4880, opened 2022-02-09
media: i2c: imx219: Scale the pixel clock rate for the 640x480 mode

naushir:
The 640x480 mode uses a special binning mode for high framerate operation where
the pixel rate is effectively doubled. Account for this when setting up the
pixel clock rate, and applying the vblank and exposure controls.

### DIFF
--- a/drivers/media/i2c/imx219.c
+++ b/drivers/media/i2c/imx219.c
@@ -153,6 +153,9 @@ struct imx219_mode {
 
 	/* Default register values */
 	struct imx219_reg_list reg_list;
+
+	/* Relative pixel clock rate factor for the mode. */
+	unsigned int rate_factor;
 };
 
 /*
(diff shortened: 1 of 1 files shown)


naushir:
@6by9 

This probably wants to be cherry-picked onto 5.10 as well.

pelwell:
This looks uncontroversial. Would you consider it a bug fix, though? 5.10 is effectively in maintenance mode now.

naushir:
I'd class this as a bug fix rather than a new feature, hence fixing up the 5.10 branch. But I have no strong objections if we don't.

6by9:
Looks sensible assuming it ends up with the right numbers.
We had seen exposure values varying frame rate if over half the frame time - did that get resolved as it doesn't appear to be in this patch.

naushir:
That should be resolved with this change.  I did test with the exposure light box to verify.

PR #4889, opened 2022-02-14
vc4_dpi fixes

6by9:
Fixes RGB565 mapping, and adds BGR mappings to the overlay.

### DIFF
--- a/arch/arm/boot/dts/overlays/README
+++ b/arch/arm/boot/dts/overlays/README
@@ -3651,8 +3651,12 @@ Params: clock-frequency         Display clock frequency (Hz)
         rgb565                  Change to RGB565 output on GPIOs 0-19
         rgb565-padhi            Change to RGB565 output on GPIOs 0-8, 12-17, and
                                 20-24
+        bgr666                  Change to BGR666 output on GPIOs 0-21.
+        bgr666-padhi            Change to BGR666 output on GPIOs 0-9, 12-17, and
+                                20-25
         rgb666-padhi            Change to RGB666 output on GPIOs 0-9, 12-17, and
                                 20-25
+        bgr888                  Change to BGR888 output on GPIOs 0-27
         rgb888                  Change to RGB888 output on GPIOs 0-27
         bus-format              Override the bus format for a MEDIA_BUS_FMT_*
                                 value. NB also overridden by rgbXXX overrides.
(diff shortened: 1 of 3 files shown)


grigorig:
I wonder if it makes sense to add overrides to free up GPIO0/1 for general usage since they are not needed for many displays? With the current structure that would need a lot of duplication though.

6by9:
> I wonder if it makes sense to add overrides to free up GPIO0/1 for general usage since they are not needed for many displays? With the current structure that would need a lot of duplication though.

That can be a later change.
Most DPI panels require at least the pixel clock. The only exception I know of is the VGA666 adapter which uses a resistor ladder as a cheap ADC to convert to VGA (which doesn't need a clock).
It'd need to duplicate all the rgb and bgr overrides to choose an alternate pinctrl setting, which isn't so much duplication.

grigorig:
I think you are right - the VGA adapters which I obviously had in mind are the exception, not the rule.

aBUGSworstnightmare-rpi:
@6by9: https://github.com/raspberrypi/linux/pull/4889/commits/cd762b4ff5ac2e3914189e135deb8ae9e3787742 this overlay is missing the rotation parameter 'rotation = <0>;' (should go in line 16). There is also no override for it.
So, it's either the .dts that's wrong or the readme (as rotate is stated there as available dtparam)

6by9:
> this overlay is missing the rotation parameter 'rotation = <0>;' (should go in line 16). There is also no override for it.
So, it's either the .dts that's wrong or the readme (as rotate is stated there as available dtparam)

No, it's correct.
The override (along with the backlight control ones are in [vc4-kms-dpi.dtsi](https://github.com/raspberrypi/linux/blob/rpi-5.15.y/arch/arm/boot/dts/overlays/vc4-kms-dpi.dtsi#L109), with the default at https://github.com/raspberrypi/linux/blob/rpi-5.15.y/arch/arm/boot/dts/overlays/vc4-kms-dpi.dtsi#L13

pelwell:
> So, it's either the .dts that's wrong or the readme (as rotate is stated there as available dtparam)

Actually it's neither - the overlay is correct. The `rotate` parameter is inherited from the included .dtsi file:

https://github.com/raspberrypi/linux/blob/cd762b4ff5ac2e3914189e135deb8ae9e3787742/arch/arm/boot/dts/overlays/vc4-kms-dpi.dtsi#L109

aBUGSworstnightmare-rpi:
omg! Sorry two things:
1. missed the include 
2. had a dtparam=rotation=180 instead of dtparam=rotate=180

So, all fine! Sorry fo inconveniences !

this is working fine for a 1366x768pixels display (where I needed the other color mapping)
dtoverlay=vc4-kms-dpi-generic
dtparam=clock-frequency=84750000
dtparam=bgr888=1
dtparam=hactive=1364,hfp=69,hsync=136,hbp=208
dtparam=vactive=768,vfp=3,vsync=10,vbp=17
dtparam=rotate=180

The only 'issue' that I still see is that screen configuration editor doesn't know if this rotation and shows 'DPI-1' correct (and not upside down as when rotating it via arandr).

PR #4893, opened 2022-02-17
usb: xhci: add a quirk for Superspeed bulk OUT transfers on VL805

P33M:
See #4844 

Do we want this in 5.10 as well or are we close to an apt release in 5.15?


pelwell:
This is very clearly a bug fix, so it will end up in the `oldstable` branch - it's now in rpi-5.10.y.

PR #4895, opened 2022-02-17
Fix for the cursor corruption with KMS (again)

mripard:
Hi,

Here's a fix for the remaining cursor corruption bug when we were moving the cursor along the edge shared by two monitors with overscan enabled, mentioned in the PR #4776 .

The initial solution implemented by that PR doesn't work, since the root cause of the corruption seems to be that we copy the shadow dlist outside of the vblank window, probably due to scheduling delay.

This PR first reverts the patches implementing the shadow dlist, and implement another solution. 

That corruption is due to the driver overwriting a dlist entry being used by the HVS to generate the current frame. Since we can't reliably copy something during vblank, the alternative is to defer the deallocation of the dlist entries to after the next vblank, when we know for sure they won't be used by the hardware anymore. Commit 5f5bb180b43a has more details on the care that needs to be taken in the implementation.

I can't see any corruption now, even after fairly long testing, so it seems that it's working as expected and it should be fixed for good (famous last words).

Maxime


popcornmix:
I've tested this and it seems improved. I've not triggered the blatant glitches I'd seen without this patch (one screen blanking or showing noise pixels near bottom of screen).

I am seeing something funny. Two monitors at 1080p60.

When cursor is close to right hand edge of left screen and moving, I see the cursor flicker on right screen near left hand edge. 
I believe mouse pointer is actually 64x64 pixels, but only the top left 16x16 are not fully transparent.
I believe the issue occurs when this transparent part of plane straddles the two screens.

It seems easier to provoke this towards bottom of screen.
Put mouse pointer at bottom right of left screen and waggle it as close to right hand edge as you can without crossing into next screen. Mouse pointer is occasionally visible on left edge of right screen.

popcornmix:
Just in case this is a higher level issue (e.g. a bug in mutter) I've tested with fkms driver, and I don't see the issue, so it seems to be in kms driver.

popcornmix:
A bit more testing. Mouse glitch affects buster and bullseye.
I can still see the cursor glitch without this PR (and without PR I get the additional glitches).

I tested this PR with kodi watching a number of videos without regressions.
Pi3+ still boots with working desktop using kms. 

Can't see any issue in commits. The HVS register range change and frame number change are correct. 
The logic for deferring deallocation is sane.

So I'm happy with this PR, it fixes an issue. There still exists an unrelated issue related to mouse pointer spanning two screens, but that can be dealt with in a later PR.

Issue #4898, opened 2022-02-17
5.15 kernel break Hifiberry DAC 2 HD Hat


apastuszak:
### Describe the bug

Upgrading to the 5.15 Linux kernel breaks compatibiilty with the Hifiberry DAC 2 HD. I did an rpi-update, rebooted my pi, and aplay no longer sees the HAT.

This is what is in my config.txt

```
# For more options and information see
# http://rpf.io/configtxt
# Some settings may impact device functionality. See link above for details

# uncomment if you get no picture on HDMI for a default "safe" mode
#hdmi_safe=1

# uncomment the following to adjust overscan. Use positive numbers if console
# goes off screen, and negative if there is too much border
#overscan_left=16
#overscan_right=16
#overscan_top=16
#overscan_bottom=16

# uncomment to force a console size. By default it will be display's size minus
# overscan.
#framebuffer_width=1280
#framebuffer_height=720

# uncomment if hdmi display is not detected and composite is being output
#hdmi_force_hotplug=1

# uncomment to force a specific HDMI mode (this will force VGA)
#hdmi_group=1
#hdmi_mode=1

# uncomment to force a HDMI mode rather than DVI. This can make audio work in
# DMT (computer monitor) modes
#hdmi_drive=2

# uncomment to increase signal to HDMI, if you have interference, blanking, or
# no display
#config_hdmi_boost=4

# uncomment for composite PAL
#sdtv_mode=2

#uncomment to overclock the arm. 700 MHz is the default.
#arm_freq=800

# Uncomment some or all of these to enable the optional hardware interfaces
#dtparam=i2c_arm=on
#dtparam=i2s=on
#dtparam=spi=on

# Uncomment this to enable infrared communication.
#dtoverlay=gpio-ir,gpio_pin=17
#dtoverlay=gpio-ir-tx,gpio_pin=18

# Additional overlays and parameters are documented /boot/overlays/README

# Enable audio (loads snd_bcm2835)
#dtparam=audio=on

# Automatically load overlays for detected cameras
camera_auto_detect=1

# Automatically load overlays for detected DSI displays
display_auto_detect=1

# Enable DRM VC4 V3D driver
dtoverlay=vc4-kms-v3d,audio=off
max_framebuffers=2

# Run in 64-bit mode
arm_64bit=1

# Disable compensation for displays with overscan
disable_overscan=1

[cm4]
# Enable host mode on the 2711 built-in XHCI USB controller.
[...]

PR #4904, opened 2022-02-23
overlays:fixes probing of Hifiberry DAC2 HD

j-schambacher:
Removed clocks-declarations in I2C sections of the DT-overlay
which kept the devices from probing.

Solves  issue  #4898 5.15 kernel break Hifiberry DAC 2 HD Hat 

### DIFF
--- a/arch/arm/boot/dts/overlays/hifiberry-dacplushd-overlay.dts
+++ b/arch/arm/boot/dts/overlays/hifiberry-dacplushd-overlay.dts
@@ -8,23 +8,13 @@
 	compatible = "brcm,bcm2835";
 
 	fragment@0 {
-		target-path = "/";
-		__overlay__ {
-			dachd_osc: pll_dachd_osc {
-				compatible = "hifiberry,dachd-clk";
-				#clock-cells = <0>;
-			};
-		};
-	};
-
-	fragment@1 {
 		target = <&i2s>;
 		__overlay__ {
 			status = "okay";
 		};
 	};
 
-	fragment@2 {
+	fragment@1 {
 		target = <&i2c1>;
 		__overlay__ {
 			#address-cells = <1>;
@@ -35,7 +25,6 @@
 				compatible = "ti,pcm1792a";
 				#sound-dai-cells = <0>;
 				#clock-cells = <0>;
-				clocks = <&dachd_osc>;
 				reg = <0x4c>;
 				status = "okay";
 			};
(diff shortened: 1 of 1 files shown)


pelwell:
Does fragment 0 achieve anything? What happens if you delete it?

j-schambacher:
Yes, you're right with your assumption. That part is not necessary.
Thanks for the hint. I've updated the commit.

fps:
Would this be an issue on other hifiberry cards as well? I'm using a DAC+ADC pro and it's not detected on my 5.15.32-rt39 kernel. I'll try hacking together a patch similar to this and see if that helps. Might take a while since I'm using nixos instead of raspberry OS.

j-schambacher:
@fps The DAC HD uses a dedicated PLL IC. And the DAC IC on the DAC HD requires that PLL clock to be present to go through the POR and finally allow its the detection on the I2C bus. So, here the correct probing is important which had been skrewed by the old DT file. The other Hifiberry cards do not use that concept and therefore, I doubt, a similar patch solves your issue. You may have a look to this issue https://github.com/raspberrypi/linux/issues/5012. We use for Hifiberry OS 5.15.24 without such problems. But please let me know what you find out.

fps:
@j-schambacher It seems the failure is specific to the way the kernel is built/booted on nixos-aarch64. I built the same kernel on raspberry os (64 bit) and the overlay loads fine.

princp69:
My Hifiberry Digi+ Pro board is not recognised by 5.15. Could this be the same problem? I've tried the troubleshooting from Hifiberry's site (https://www.hifiberry.com/docs/software/configuring-linux-3-18-x/) to no avail.

pelwell:
What changes have you made to config.txt?
Are there any messages in `/proc/device-tree/chosen/user-warnings`?

princp69:
dtoverlay=
force_eeprom_read=0
dtdebug=1
dtoverlay=hifiberry-digi-pro

No user-warnings file.

j-schambacher:
Are you using a PI 4? Could it be a HW incompatibility between the Digi and the revision 1.5 of the PI 4? Please have a look here https://www.hifiberry.com/blog/digi2-pro-raspberry-pi-4-1-5-incompatibilities/

princp69:
I've managed to get it working. Multiple installs. Re-sitting the board. Checking everything. Something was amiss. Now it works! Thanks for your help.

PR #4908, opened 2022-02-25
Add driver for Omnivision OV2311

6by9:
Pushing for review, although looking at it there appear to be a few things that could be cleaned up (eg dropping s_power hooks for pm_runtime, and gpiod_set_value_cansleep already ensures that the GPIO is valid, so we don't need to handle it).

Pixel rate, frame rate control, and exposure settings have been validated as correct, even though they have been measured as we don't have a datasheet.
8bpp readout modes are slow by about 0.55%, but not fussing about that. 10bpp mode is almost spot on.


arch/arm/boot/dts/overlays/README
```diff
@@ -2421,6 +2421,22 @@ Params: speed                   Display SPI bus speed
         xohms                   Touchpanel sensitivity (X-plate resistance)
 
 
+Name:   ov2311
+Info:   Omnivision OV2311 camera module.
+        Uses Unicam 1, which is the standard camera connector on most Pi
+        variants.
+Load:   dtoverlay=ov2311,<param>=<val>
+Params: <None>
```
pelwell: Gremlin.

drivers/media/i2c/ov2311.c
```diff
@@ -0,0 +1,1205 @@
...
+		__v4l2_ctrl_modify_range(ov2311->exposure,
+					 ov2311->exposure->minimum, max,
+					 ov2311->exposure->step,
+					 ov2311->exposure->default_value);
+		break;
+	}
+
+	if (pm_runtime_get(&client->dev) <= 0)
+		return 0;
+	switch (ctrl->id) {
```
pelwell, line 882: The rest of the code is so nicely formatted that I couldn't help noticing the missing blank line before the switch compared to the one after. Either would be OK, it's the asymmetry that's noticeable.

pelwell (commented):
One cut-and-paste error, one whitespace ultra-niggle.

drivers/media/i2c/ov2311.c
6by9: Blank line added.

6by9:
Tested and it still runs. I'd also got the media-controller property default wrong. Updated branch.

Please don't merge quite yet - it'd be worth me doing one last check through before. I really ought to create a DT bindings doc as well :-/

(OV9281 could also do with a couple of cleanups based on what I've noticed in reviewing this patchset)

6by9:
Further rebase and final review done by me.

Unless @naushir or @davidplowman want to review, then I'm happy for it to be merged. I'll pass the sensor on to @davidplowman as another one needing a basic tune :-)

PR #4910, opened 2022-02-27
Update README

Kaustubh01:
corrected grammar

### DIFF
--- a/README
+++ b/README
@@ -1,18 +1,18 @@
 Linux kernel
 ============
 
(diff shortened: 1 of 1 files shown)


6by9:
Thanks for the PR, however this file is unmodified from mainline Linux - https://github.com/torvalds/linux/blob/master/README
Any changes should therefore be submitted to mainline.

pelwell:
I don't think your edits are necessary, and some are ungrammatical or misspelled.

PR #4916, opened 2022-03-02
staging/vc-sm-cma: Avoid log spamming on Pi0/1 over cache alias.

6by9:
Pi 0/1 use the 0x80000000 cache alias as the ARM also sees the world
through the VPU L2 cache.
vc-sm-cma was trying to ensure it was in an uncached alias (0xc), and
complaining on every allocation if it weren't. Reduce this logging.


popcornmix:
LGTM

PR #4918, opened 2022-03-03
staging: vc04_services: isp: Permit all sRGB colour spaces on ISP out…

davidplowman:
…puts

ISP outputs actually support all colour spaces that are fundamentally
sRGB underneath, regardless of whether an RGB or YUV output format is
actually requested.


davidplowman:
@naushir @6by9 I seem to keep tying myself in knots trying to explain this patch. The basic point is that the ISP's RGB outputs can accept any colour space that is fundamentally sRGB underneath, not just V4L2_COLOURSPACE_SRGB.

6by9:
I think I follow the logic, but I never fully follow what V4L2 wants to advertise in each case.

naushir:
Makes sense, LGTM.

PR #4921, opened 2022-03-03
drivers: staging: bcm2835-isp: Do not cleanup mmal vcsm buffer on sto…

naushir:
…p_streaming

On stop_streaming() the vcsm buffer handle gets released by the buffer cleanup
code.  This will subsequently cause and error if userland re-queues the same
buffer on the next start_streaming() call.

Remove this cleanup code and rely on the vb2_ops->buf_cleanup() call to do the
cleanups instead.

### DIFF
--- a/drivers/staging/vc04_services/bcm2835-isp/bcm2835-v4l2-isp.c
+++ b/drivers/staging/vc04_services/bcm2835-isp/bcm2835-v4l2-isp.c
@@ -622,7 +622,6 @@ static void bcm2835_isp_node_stop_streaming(struct vb2_queue *q)
 {
 	struct bcm2835_isp_node *node = vb2_get_drv_priv(q);
 	struct bcm2835_isp_dev *dev = node_get_dev(node);
-	unsigned int i;
 	int ret;
 
 	v4l2_dbg(1, debug, &dev->v4l2_dev, "%s: node %s[%d], mmal port %p\n",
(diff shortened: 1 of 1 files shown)


naushir:
@6by9 and @davidplowman

6by9:
LGTM

pelwell:
I'm getting a warning that `i` is unused.

naushir:
> I'm getting a warning that `i` is unused.

Sorry, missed that.  Now fixed!

davidplowman:
I don't really understand this well enough to comment, but you could explain a couple of things for me?
- If these handles were being cleaned up, how did stopping and starting the ISP (with no "configure" in between) ever work?
- Who cleans these handles up now, is vb2 invoking a buffer cleanup operation at some point?

naushir:
> I don't really understand this well enough to comment, but you could explain a couple of things for me?
> 
> * If these handles were being cleaned up, how did stopping and starting the ISP (with no "configure" in between) ever work?

It didn't exactly.  We would always re-allocate buffers on restart and never reuse the old ones.  With our proposed user-land changes, we are now re-using the buffer.

> * Who cleans these handles up now, is vb2 invoking a buffer cleanup operation at some point?

The vb2_ops->buf_cleanup() code will do pretty much exactly what this loop does.

davidplowman:
Thanks. I guess this is perhaps a discussion for somewhere else, but what will the libcamera situation be once this change goes in? Will the updated kernel work with existing libcamera or will we need pipeline handler updates to be committed at the same time?

naushir:
We don't need to sync the userland change with this change.

The usecase that fails will continue failing until both changes are merged.  Merging either userland or kernel change in isolation won't break current working cases either.

Issue #4901, opened 2022-02-21
Ugly Screen Tearing on Kernel 5.15.X and up with OGL apps in DRM


bluestang2006:
### Describe the bug

I am seeing very bad screen tearing in OpenGL apps, specifically vkQuake3, when using the app in KMS+DRM aka console.

This behavior is not seen in an X11 session, and it not seen using the Vulkan driver in either session (X11 or KMS+DRM).

As I said before it's strictly being seen in DRM using an OGL context. 

Switching back to the latest stable firmware (5.10) the issue goes away.

### Steps to reproduce the behaviour

This is easily reproducible if you compile/setup [vkQuake3](https://github.com/suijingfeng/vkQuake3) and set the default renderer to opengl1 - `seta cl_renderer "opengl1"`. This is best done by creating an autoexec.cfg in the baseq3 folder and running the app in DRM.

### Device (s)

Raspberry Pi 4 Mod. B

### System

64-bit RPiOS (Bullseye)
vc4-kms-v3d
kernel 5.15 (latest master branch)
NOTE: I am using the latest MESA git for the v3d+v3dv drivers.

mripard:
It seems like it could be the same issue than the one we fixed in #4571 

Interestingly, that patch never got applied to the rpi-5.15.y branch, which might explain why it works on 5.10 but doesn't on 5.15.

@pelwell

pelwell:
Being discussed offline, but this was one of the many patches that doesn't apply when moving from 5.10 to 5.15.

pelwell:
Or rather, it applies but then produces a build error due to `dma_resv_get_excl` no longer being defined.

pelwell:
@mripard Are you going to update this patch for 5.15?

mripard:
Yes, I was planning on sending a new version but got distracted by the writeback issues, I'll send a PR today or on monday

pelwell:
Sure - when you're ready.

mripard:
I've just done it with PR #4925

bluestang2006:
#4925 fixed this issue. Tested with Quake3e and another OpenGL app called RetroFE and the vysnc issue is resolved. As a bonus this patch was tested on the rpi-5.16.y branch and everything looked great.

pelwell:
Thanks - forward-ported to rpi-5.16.y and rpi-5.17y.

PR #4925, opened 2022-03-04
drm/vc4: Fix out of order frames during asynchronous page flips

mripard:
This a port of the PR #4571 for 5.15, with the updated function name that was broken before.

### DIFF
--- a/drivers/gpu/drm/vc4/vc4_crtc.c
+++ b/drivers/gpu/drm/vc4/vc4_crtc.c
@@ -813,6 +813,7 @@ struct vc4_async_flip_state {
 	struct drm_pending_vblank_event *event;
 
 	struct vc4_seqno_cb cb;
+	struct dma_fence_cb fence_cb;
 };
 
 /* Called when the V3D execution for the BO being flipped to is done, so that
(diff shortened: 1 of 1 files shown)


pelwell:
Did you push the wrong patch? That looks like a direct cherry-pick of f694d1cfee63.

pelwell:
It can wait until next week, though.

popcornmix:
Agreed - this doesn't build on 5.15 due to dma fence api changes.

mripard:
> Did you push the wrong patch? That looks like a direct cherry-pick of [f694d1c](https://github.com/raspberrypi/linux/commit/f694d1cfee63afaaa77e7d9e1f621a83e2f98846).

No, I forgot to commit the changes I had locally... It should be ok now, sorry

popcornmix:
This compiles now and kodi still runs. I don't have a test app that shows the out-of-order issue (but if it worked on 5.10 the odds seem good).

bluestang2006:
> This compiles now and kodi still runs. I don't have a test app that shows the out-of-order issue (but if it worked on 5.10 the odds seem good).

Just tested this patch applied to rpi-5.16.y and it fixes the vsync issue with Quake3e and another OpenGL app I saw this issue with. Thanks for the fix. Closed #4901.

PR #4931, opened 2022-03-07
staging: vc04_services: codec: Add support for V4L2_PIX_FMT_RGBA32 fo…

davidplowman:
…rmat

We already support V4L2_PIX_FMT_BGR32 which is the same thing with red
and blue swapped, so it makes sense to include this variant as well.


davidplowman:
@6by9 @naushir This allows us to pass RGBX from Picamera2 to the video codec.

6by9:
LGTM

PR #4932, opened 2022-03-07
drm/vc4: hdmi: Add CSC for BT601/709/2020 limited and full range output

6by9:
The HVS always composes in the RGB domain, but there is a colourspace
conversion block on the output to allow for sending YCbCr over the
HDMI interface.
The colourspace on that link is configurable via the "Colorspace"
property on the connector, and that updates the infoframes. There
is also selection of limited or full range based on the mode selected
or an override.

Add code to update the CSC as well so that the metadata matches the
image data.

### DIFF
--- a/drivers/gpu/drm/vc4/vc4_hdmi.c
+++ b/drivers/gpu/drm/vc4/vc4_hdmi.c
@@ -138,8 +138,8 @@ static bool vc4_hdmi_mode_needs_scrambling(const struct drm_display_mode *mode,
 	return clock > HDMI_14_MAX_TMDS_CLK;
 }
 
-static bool vc4_hdmi_is_full_range_rgb(struct vc4_hdmi *vc4_hdmi,
-				       const struct drm_display_mode *mode)
+static bool vc4_hdmi_is_full_range(struct vc4_hdmi *vc4_hdmi,
+				   const struct drm_display_mode *mode)
 {
 	struct vc4_hdmi_encoder *vc4_encoder = &vc4_hdmi->encoder;
 
@@ -679,7 +679,7 @@ static void vc4_hdmi_set_avi_infoframe(struct drm_encoder *encoder)
 
 	drm_hdmi_avi_infoframe_quant_range(&frame.avi,
 					   connector, mode,
-					   vc4_hdmi_is_full_range_rgb(vc4_hdmi, mode) ?
+					   vc4_hdmi_is_full_range(vc4_hdmi, mode) ?
 					   HDMI_QUANTIZATION_RANGE_FULL :
 					   HDMI_QUANTIZATION_RANGE_LIMITED);
 	drm_hdmi_avi_infoframe_colorimetry(&frame.avi, cstate);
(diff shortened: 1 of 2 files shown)


drivers/gpu/drm/vc4/vc4_hdmi.c
```diff
@@ -970,42 +961,107 @@ static const u16 vc5_hdmi_csc_full_rgb_unity[3][4] = {
...
  *
  * Matrix is signed 2p13 fixed point, with signed 9p6 offsets
  */
-static const u16 vc5_hdmi_csc_full_rgb_to_full_yuv444_bt709[3][4] = {
-	{ 0xfc41, 0xf364, 0x105e, 0x2000 },
-	{ 0x105e, 0xf124, 0xfe81, 0x2000 },
-	{ 0x06ce, 0x16e3, 0x024f, 0x0000 },
+static const u16 vc5_hdmi_csc_full_rgb_to_yuv_bt2020[2][3][4] = {
+	{
+		/* Full range - unity */
```
popcornmix: unity?

6by9: Too much copy/paste.

6by9:
Fixed up.

Just waiting on jc-kynesim to produce a test app that sets the plane colourspace appropriately, and then will confirm values on the analyser.

drivers/gpu/drm/vc4/vc4_hdmi.c
```diff
@@ -970,42 +960,106 @@ static const u16 vc5_hdmi_csc_full_rgb_unity[3][4] = {
...
+		/* Limited range */
+		{ 0x082f, 0x1012, 0x031f, 0x0400 },
+		{ 0xfb48, 0xf6ba, 0x0e00, 0x2000 },
+		{ 0x0e00, 0xf448, 0xfdba, 0x2000 },
+	}
 };
 
 /*
- * Conversion between Full Range RGB and Full Range YUV444 using the
+ * Conversion between Full Range RGB and Full Range YUV using the
```
popcornmix: No longer "Full Range" YUV

6by9: Fixed

6by9:
BT2020 full and limited range and BT709 limited range validated and are accurate within the 10bit range of the input.

BT709 full range is slightly off in Cb channel (0x300 in 10bit input produces 0xC1B in 12bit output instead of 0xC00). Output CSC ought to be validated there.

6by9:
Updated as there were some DRM_MODE_COLORIMETRY_* values that weren't explicitly mapped to any matrices.

6by9:
@jc-kynesim

jc-kynesim:
This branch has a ridiculous number of commits in it - has it come adrift from reality in some way?

popcornmix:
It needs rebasing following rpi-5.15.y rebase. You can always merge from 6by9:rpi-5.15.y-dpi.

6by9:
This is why I asked yesterday if we were ready to merge. There's little point in rebasing on every change to rpi-5.15.y if we aren't looking at merging it. You can always cherry-pick the top commits.

Rebased now.

popcornmix:
I think we should merge this, as it fixes a clear bug. If any further tweaking of matrix is required it can be a follow up. @6by9 can you rebase?

6by9:
Rebased as requested (despite what the Github PR says).

Github seems to have a couple of issues at present as it took several attempts to push the branch due to an "internal server error", and it's not wanting to accept my comment now.

PR #4935, opened 2022-03-11
imx219: implement v4l2 set selection api

vvarma:
Adding cropping functionality to the imx219 driver using the v4l2 selection api. 
The sequence of operations is:
- imx219_probe - sets the crop based on current mode's default_crop
- VIDIOC_S_FMT - this changes the crop to the selected mode's default_crop
- VIDIOC_S_SELECTION - this changes the crop to the user defined crop after validation and resets the range of exposure and other controls
- imx219_start_streaming - applies the selection into camera registers

### DIFF
--- a/drivers/media/i2c/imx219.c
+++ b/drivers/media/i2c/imx219.c
@@ -26,6 +26,7 @@
 #include <media/v4l2-event.h>
 #include <media/v4l2-fwnode.h>
 #include <media/v4l2-mediabus.h>
+#include <media/v4l2-rect.h>
 #include <asm/unaligned.h>
 
 #define IMX219_REG_VALUE_08BIT		1
@@ -110,6 +111,16 @@
 #define IMX219_TESTP_BLUE_DEFAULT	0
 #define IMX219_TESTP_GREENB_DEFAULT	0
 
+/* IMX219 left, top, height, width registers */
+#define IMX219_REG_X_START	0x0164
+#define IMX219_REG_X_END	0x0166
+#define IMX219_REG_Y_START	0x0168
+#define IMX219_REG_Y_END	0x016a
+#define IMX219_REG_WIDTH	0x016c
+#define IMX219_REG_HEIGHT	0x016e
+#define IMX219_REG_TEST_WIDTH 0x0624
+#define IMX219_REG_TEST_HEIGHT 0x0626
+
 /* IMX219 native and active pixel array size. */
 #define IMX219_NATIVE_WIDTH		3296U
 #define IMX219_NATIVE_HEIGHT		2480U
@@ -145,8 +156,14 @@ struct imx219_mode {
 	/* Frame height */
 	unsigned int height;
 
-	/* Analog crop rectangle. */
-	struct v4l2_rect crop;
+	/* Default crop rectangle. */
+	struct v4l2_rect default_crop;
+
+	/* x binning */
+	u8 x_scaler;
+
+	/* y binning */
+	u8 y_scaler;
 
 	/* V-timing */
 	unsigned int vts_def;
@@ -487,12 +504,14 @@ static const struct imx219_mode supported_modes[] = {
 		/* 8MPix 15fps mode */
 		.width = 3280,
 		.height = 2464,
-		.crop = {
+		.default_crop = {
 			.left = IMX219_PIXEL_ARRAY_LEFT,
 			.top = IMX219_PIXEL_ARRAY_TOP,
 			.width = 3280,
 			.height = 2464
 		},
+		.x_scaler = 1,
+		.y_scaler = 1,
 		.vts_def = IMX219_VTS_15FPS,
 		.reg_list = {
 			.num_of_regs = ARRAY_SIZE(mode_3280x2464_regs),
@@ -504,12 +523,14 @@ static const struct imx219_mode supported_modes[] = {
 		/* 1080P 30fps cropped */
 		.width = 1920,
 		.height = 1080,
-		.crop = {
+		.default_crop = {
 			.left = 688,
 			.top = 700,
 			.width = 1920,
 			.height = 1080
 		},
+		.x_scaler = 1,
+		.y_scaler = 1,
 		.vts_def = IMX219_VTS_30FPS_1080P,
 		.reg_list = {
 			.num_of_regs = ARRAY_SIZE(mode_1920_1080_regs),
@@ -521,12 +542,14 @@ static const struct imx219_mode supported_modes[] = {
 		/* 2x2 binned 30fps mode */
 		.width = 1640,
 		.height = 1232,
-		.crop = {
+		.default_crop = {
 			.left = IMX219_PIXEL_ARRAY_LEFT,
 			.top = IMX219_PIXEL_ARRAY_TOP,
 			.width = 3280,
 			.height = 2464
 		},
+		.x_scaler = 2,
+		.y_scaler = 2,
 		.vts_def = IMX219_VTS_30FPS_BINNED,
 		.reg_list = {
 			.num_of_regs = ARRAY_SIZE(mode_1640_1232_regs),
@@ -538,12 +561,14 @@ static const struct imx219_mode supported_modes[] = {
 		/* 640x480 30fps mode */
 		.width = 640,
 		.height = 480,
-		.crop = {
+		.default_crop = {
 			.left = 1008,
 			.top = 760,
 			.width = 1280,
 			.height = 960
 		},
+		.x_scaler = 2,
+		.y_scaler = 2,
 		.vts_def = IMX219_VTS_30FPS_640x480,
 		.reg_list = {
 			.num_of_regs = ARRAY_SIZE(mode_640_480_regs),
@@ -551,7 +576,7 @@ static const struct imx219_mode supported_modes[] = {
 		},
 		/*
 		 * This mode uses a special 2x2 binning that doubles the
-		 * the internal pixel clock rate.
+		 * internal pixel clock rate.
 		 */
 		.rate_factor = 2,
 	},
@@ -582,6 +607,9 @@ struct imx219 {
 	/* Current mode */
 	const struct imx219_mode *mode;
 
+	/* Current crop rectangle. */
+	struct v4l2_rect crop;
+
 	/*
 	 * Mutex for serialized access:
 	 * Protect sensor module set pad format and start/stop streaming safely.
@@ -591,6 +619,13 @@ struct imx219 {
 	/* Streaming on/off */
 	bool streaming;
 };
+
+#define IMX219_ROUND(dim, step, flags)			\
+	((flags) & V4L2_SEL_FLAG_GE			\
+	? roundup((dim), (step))			\
+	: ((flags) & V4L2_SEL_FLAG_LE			\
+		? rounddown((dim), (step))		\
+		: rounddown((dim) + (step) / 2, (step))))
 
 static inline struct imx219 *to_imx219(struct v4l2_subdev *_sd)
 {
@@ -701,8 +736,8 @@ static void imx219_set_default_format(struct imx219 *imx219)
 							  fmt->colorspace,
 							  fmt->ycbcr_enc);
 	fmt->xfer_func = V4L2_MAP_XFER_FUNC_DEFAULT(fmt->colorspace);
-	fmt->width = supported_modes[0].width;
-	fmt->height = supported_modes[0].height;
+	fmt->width = imx219->crop.width / imx219->mode->x_scaler;
+	fmt->height = imx219->crop.height / imx219->mode->y_scaler;
 	fmt->field = V4L2_FIELD_NONE;
 }
 
@@ -718,8 +753,8 @@ static int imx219_open(struct v4l2_subdev *sd, struct v4l2_subdev_fh *fh)
 	mutex_lock(&imx219->mutex);
 
 	/* Initialize try_fmt for the image pad */
-	try_fmt_img->width = supported_modes[0].width;
-	try_fmt_img->height = supported_modes[0].height;
+	try_fmt_img->width = imx219->crop.width / imx219->mode->x_scaler;
+	try_fmt_img->height = imx219->crop.height / imx219->mode->y_scaler;
 	try_fmt_img->code = imx219_get_format_code(imx219,
 						   MEDIA_BUS_FMT_SRGGB10_1X10);
 	try_fmt_img->field = V4L2_FIELD_NONE;
@@ -731,7 +766,7 @@ static int imx219_open(struct v4l2_subdev *sd, struct v4l2_subdev_fh *fh)
 	try_fmt_meta->field = V4L2_FIELD_NONE;
 
 	/* Initialize try_crop rectangle. */
-	try_crop = v4l2_subdev_get_try_crop(sd, fh->state, 0);
+	try_crop = v4l2_subdev_get_try_crop(sd, fh->state, IMAGE_PAD);
 	try_crop->top = IMX219_PIXEL_ARRAY_TOP;
 	try_crop->left = IMX219_PIXEL_ARRAY_LEFT;
 	try_crop->width = IMX219_PIXEL_ARRAY_WIDTH;
@@ -751,9 +786,10 @@ static int imx219_set_ctrl(struct v4l2_ctrl *ctrl)
 
 	if (ctrl->id == V4L2_CID_VBLANK) {
 		int exposure_max, exposure_def;
-
+		u32 height = imx219->crop.height / imx219->mode->y_scaler;
 		/* Update max exposure while meeting expected vblanking */
-		exposure_max = imx219->mode->height + ctrl->val - 4;
+		exposure_max = height + ctrl->val - 4;
+
 		exposure_def = (exposure_max < IMX219_EXPOSURE_DEFAULT) ?
 			exposure_max : IMX219_EXPOSURE_DEFAULT;
 		__v4l2_ctrl_modify_range(imx219->exposure,
@@ -794,12 +830,15 @@ static int imx219_set_ctrl(struct v4l2_ctrl *ctrl)
 				       imx219->hflip->val |
 				       imx219->vflip->val << 1);
 		break;
-	case V4L2_CID_VBLANK:
+	case V4L2_CID_VBLANK: {
+		u32 height = imx219->crop.height / imx219->mode->y_scaler;
+		u32 vblank_val =
+			(height + ctrl->val) / imx219->mode->rate_factor;
+
 		ret = imx219_write_reg(imx219, IMX219_REG_VTS,
-				       IMX219_REG_VALUE_16BIT,
-				       (imx219->mode->height + ctrl->val) /
-						imx219->mode->rate_factor);
+				       IMX219_REG_VALUE_16BIT, vblank_val);
 		break;
+	}
 	case V4L2_CID_TEST_PATTERN_RED:
 		ret = imx219_write_reg(imx219, IMX219_REG_TESTP_RED,
 				       IMX219_REG_VALUE_16BIT, ctrl->val);
@@ -903,10 +942,11 @@ static void imx219_reset_colorspace(struct v4l2_mbus_framefmt *fmt)
 
 static void imx219_update_image_pad_format(struct imx219 *imx219,
 					   const struct imx219_mode *mode,
+					   const struct v4l2_rect *crop,
 					   struct v4l2_subdev_format *fmt)
 {
-	fmt->format.width = mode->width;
-	fmt->format.height = mode->height;
+	fmt->format.width = crop->width / mode->x_scaler;
+	fmt->format.height = crop->height / mode->y_scaler;
 	fmt->format.field = V4L2_FIELD_NONE;
 	imx219_reset_colorspace(&fmt->format);
 }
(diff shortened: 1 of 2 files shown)


drivers/media/i2c/imx219.c
```diff
@@ -742,6 +768,20 @@ static int imx219_open(struct v4l2_subdev *sd, struct v4l2_subdev_fh *fh)
 	return 0;
 }
 
+static __u32 imx219_get_height(const struct imx219_mode *mode, const struct v4l2_rect *crop)
+{
+    return mode->y_scaler == 0 ?
```
6by9: y_scaler (and x_scaler below) are defined for all modes, so add it as a WARN instead to indicate a driver bug.

drivers/media/i2c/imx219.c
```diff
@@ -742,6 +768,20 @@ static int imx219_open(struct v4l2_subdev *sd, struct v4l2_subdev_fh *fh)
 	return 0;
 }
 
+static __u32 imx219_get_height(const struct imx219_mode *mode, const struct v4l2_rect *crop)
```
6by9: Whilst in the kernel it is preferred to use either ```unsigned int``` if the exact size of the value isn't important, or ```u32``` if it is. ```__u32``` is reserved for uapi headers.
Ditto for the other uses of it below.

drivers/media/i2c/imx219.c
```diff
@@ -794,12 +834,13 @@ static int imx219_set_ctrl(struct v4l2_ctrl *ctrl)
...
 				       imx219->vflip->val << 1);
 		break;
-	case V4L2_CID_VBLANK:
+	case V4L2_CID_VBLANK: {
+		__u32 height = imx219_get_height(imx219->mode, &imx219->crop);
 		ret = imx219_write_reg(imx219, IMX219_REG_VTS,
 				       IMX219_REG_VALUE_16BIT,
-				       (imx219->mode->height + ctrl->val) /
-						imx219->mode->rate_factor);
+				       (height + ctrl->val)/imx219->mode->rate_factor);
```
6by9: Spaces required around the ```/```.

Please use ```scripts/checkpatch.pl --strict --codespell``` to ensure patches comply with the kernel coding standard (strict is required for media subsystem).

drivers/media/i2c/imx219.c
```diff
@@ -1002,42 +1079,12 @@ static int imx219_set_pad_format(struct v4l2_subdev *sd,
...
-			__v4l2_ctrl_modify_range(imx219->hblank, hblank, hblank,
-						 1, hblank);
-
-			/* Scale the pixel rate based on the mode specific factor */
-			pixel_rate =
-				IMX219_PIXEL_RATE * imx219->mode->rate_factor;
-			__v4l2_ctrl_modify_range(imx219->pixel_rate, pixel_rate,
-						 pixel_rate, 1, pixel_rate);
-		}
+            imx219->crop = mode->default_crop;
```
6by9: Indentation appears to have gone wrong here - are you using spaces instead of tabs (8 char wide).

You've also got duplication, so it would be nicer to have
```
		if (fmt->which == V4L2_SUBDEV_FORMAT_TRY) {
			framefmt = v4l2_subdev_get_try_format(sd, sd_state,
							      fmt->pad);
			*framefmt = fmt->format;
		} else {
			if (imx219->mode != mode ||
			    imx219->fmt.code != fmt->format.code)
			    imx219->fmt = fmt->format;
				imx219->mode = mode;

			imx219->crop = mode->default_crop;
			imx219_refresh_ctls(imx219);
        	}

File: drivers/media/i2c/imx219.c [hunk 5]
```diff
@@ -1127,6 +1182,82 @@ static int imx219_get_selection(struct v4l2_subdev *sd,
...
+				struct v4l2_subdev_state *state,
+				struct v4l2_subdev_selection *sel)
+{
+	struct v4l2_rect max_bounds;
+	max_bounds.top = IMX219_PIXEL_ARRAY_TOP;
+	max_bounds.left = IMX219_PIXEL_ARRAY_LEFT;
+	max_bounds.width = IMX219_PIXEL_ARRAY_WIDTH;
+	max_bounds.height = IMX219_PIXEL_ARRAY_HEIGHT;
+	switch (sel->target) {
+	case V4L2_SEL_TGT_CROP: {
```
6by9: Indentation

drivers/media/i2c/imx219.c
```diff
@@ -1127,6 +1182,82 @@ static int imx219_get_selection(struct v4l2_subdev *sd,
...
+		{ IMX219_REG_X_END, 		x_end },
+		{ IMX219_REG_Y_START, 		y_start },
+		{ IMX219_REG_Y_END, 		y_end },
+		{ IMX219_REG_WIDTH, 		width },
+		{ IMX219_REG_HEIGHT, 		height },
+	};
+	struct i2c_client *client = v4l2_get_subdevdata(&imx219->sd);
+	unsigned int i;
+	int ret;
+	for (i = 0; i < ARRAY_SIZE(regs); i++) {
```
6by9, line 1378: Blank line between variable declaration and code.

drivers/media/i2c/imx219.c
```diff
@@ -1127,6 +1182,82 @@ static int imx219_get_selection(struct v4l2_subdev *sd,
 	return -EINVAL;
 }
 
+static int imx219_set_selection(struct v4l2_subdev *sd,
+				struct v4l2_subdev_state *state,
+				struct v4l2_subdev_selection *sel)
+{
+	struct v4l2_rect max_bounds;
+	max_bounds.top = IMX219_PIXEL_ARRAY_TOP;
```
6by9: Blank line between variable declaration and code.

drivers/media/i2c/imx219.c
```diff
@@ -1127,6 +1182,82 @@ static int imx219_get_selection(struct v4l2_subdev *sd,
...
+
+			return ret;
+		}
+	}
+	return 0;
+}
+
+static int imx219_apply_selection(struct imx219 *imx219)
+{
+	__u32 width, height;
```
6by9: Blank line between variable declaration and code.

drivers/media/i2c/imx219.c
```diff
@@ -1127,6 +1182,82 @@ static int imx219_get_selection(struct v4l2_subdev *sd,
...
+	struct v4l2_rect max_bounds;
+	max_bounds.top = IMX219_PIXEL_ARRAY_TOP;
+	max_bounds.left = IMX219_PIXEL_ARRAY_LEFT;
+	max_bounds.width = IMX219_PIXEL_ARRAY_WIDTH;
+	max_bounds.height = IMX219_PIXEL_ARRAY_HEIGHT;
+	switch (sel->target) {
+	case V4L2_SEL_TGT_CROP: {
+		struct imx219 *imx219 = to_imx219(sd);
+
+		if (!v4l2_rect_enclosed(&sel->r, &max_bounds)) {
```
6by9: Should you check that you're not streaming at this point?

6by9 (commented):
Thanks for the patch. A number of relatively minor comments.

There is one major problem though - Bayer order.

As with the flips changing the Bayer order, if you request an odd x offset then you'll get one Bayer order, and an even x offset will get you a different one. Odd and even y offsets will change those again as you're moving around the Bayer quad. Add in flips and it becomes the x and y end offset that is of importance in determining the Bayer order.

There is a general requirement for Bayer formats also to be a multiple of 2 pixels in each direction too, otherwise you end up with a partial quad. (Same applies for YUV420 where otherwise you get complications over the chroma subsampling).

Personally I've never much liked the selection API as it interacts with the format too greatly, and not in a clearly defined way.
https://www.kernel.org/doc/html/latest/userspace-api/media/v4l/selection-api-configuration.html

> Each capture device has a default source rectangle, given by the V4L2_SEL_TGT_CROP_DEFAULT target. This rectangle shall cover what the driver writer considers the complete picture. Drivers shall set the active crop rectangle to the default when the driver is first loaded, but not later.

So we shouldn't reset it on S_FMT?! I don't see that working.

> The composing targets refer to a memory buffer. The limits of composing coordinates are obtained using V4L2_SEL_TGT_COMPOSE_BOUNDS. All coordinates are expressed in pixels. The rectangle’s top/left corner must be located at position (0,0). The width and height are equal to the image size set by [VIDIOC_S_FMT](https://www.kernel.org/doc/html/latest/userspace-api/media/v4l/vidioc-g-fmt.html#vidioc-g-fmt).

So should the S_FMT / imx219_set_pad_format value be independent of the V4L2_SEL_TGT_CROP, but the same as V4L2_SEL_TGT_COMPOSE?
The format (or V4L2_SEL_TGT_COMPOSE?) should be the defined mode (eg 1920x1080, or 3280x2464), and then the selection API defines a smaller part of that buffer that will actually be written to? (That's a question not a statement).

Looking at the existing drivers that implement set_selection (and there's not many of them) it's still not clear to me on the correct behaviour. IMX274 is the main one I'd look to as being relatively recent, or CCS (successor to the SMIA "standard" sensor register set). CCS makes life more complicated though as it exposes multiple subdevs for the pixel_array, binner, and scaler.

The docs on [writing sensor drivers](https://www.kernel.org/doc/html/latest/driver-api/media/camera-sensor.html#frame-size) don't really help here either, but does raise the question as to whether the cropping ought to be advertised as an extra subdevice.

Sorry, so many more questions than answers, and for a lot of them I don't know the right answers.

vvarma:
Thanks @6by9 for the review. Those are some great points and thanks additionally for pointing out the some reference implementations.

> There is one major problem though - Bayer order.

To address this i plan to adjust the user's crop such that top, left, height and width are multiples of 2, while additionally checking that the bounds dont extend beyond `V4L2_SEL_TGT_CROP_BOUNDS` / `imx219->mode->default_crop`

> So should the S_FMT / imx219_set_pad_format value be independent of the V4L2_SEL_TGT_CROP, but the same as V4L2_SEL_TGT_COMPOSE?
The format (or V4L2_SEL_TGT_COMPOSE?) should be the defined mode (eg 1920x1080, or 3280x2464), and then the selection API defines a smaller part of that buffer that will actually be written to? (That's a question not a statement).

The `IMX274` driver was an insightful reference to see how to implement the selection api specs. To summarize:
- on `probe`: Set the crop to `V4L2_SEL_TGT_CROP_BOUNDS`, fmt as the crop scaled by the default mode's binning.
- on `S_FMT`: Select and set a mode that is closest to the parameters passed that scales to the current crop. And set the format as the scaled version of the crop.
- on `SET_TGT_CROP`: Adjust the crop based on flags and sets the driver's crop - then set the format and mode of the driver based on the new crop.
- on `SET_TGT_COMPOSE`: Same as `S_FMT`

I'll implement these and the other changes requested, test and update the PR.

drivers/media/i2c/imx219.c
vvarma: Yes my default config was to use tabs instead of space. I have run the checkscript and fixed the formatting issues.

drivers/media/i2c/imx219.c
vvarma: I dont see the check in S_FMT either, and I didnt spot it in `IMX274` driver too.

vvarma:
@6by9 I have made some changes to the PR, following the pattern of `IMX274`. The big problem in doing so is 2 of the supported modes that are part of the driver internally crops the frame. This is in contrast to the role of S_FMT in comparison to `S_SELECTION_CROP`. The change I have made removes those modes and relies on `S_FMT`/`S_SELECTION_COMPOSE` for scaling the frame and `S_SELECTION_CROP` to crop the frame. However, the effective configuration can now be achieved using `S_FMT` and `S_SELECTION_CROP` in combination.

Please let me know if this is an acceptable change.
Affected modes:
`1920x1080  - mode 1`
```
pi@rpi:~ $ v4l2-ctl -V
Format Video Capture:
        Width/Height      : 3280/2464
        Pixel Format      : 'pRAA' (10-bit Bayer RGRG/GBGB Packed)
        Field             : None
        Bytes per Line    : 4128
        Size Image        : 10171392
        Colorspace        : Raw
        Transfer Function : None
        YCbCr/HSV Encoding: ITU-R 601
        Quantization      : Full Range
        Flags             :
pi@rpi:~ $ v4l2-ctl --set-selection target=crop,top=700,left=688,width=1920,height=1080
pi@rpi:~ $ v4l2-ctl -V
Format Video Capture:
        Width/Height      : 1920/1080
        Pixel Format      : 'pRAA' (10-bit Bayer RGRG/GBGB Packed)
        Field             : None
        Bytes per Line    : 2400
        Size Image        : 2592000
        Colorspace        : Raw
        Transfer Function : None
        YCbCr/HSV Encoding: ITU-R 601
        Quantization      : Full Range
        Flags             :
```
`640x480 - mode 7`
```
pi@rpi:~ $ v4l2-ctl --set-selection target=crop,top=760,left=1008,width=1280,height=960
pi@rpi:~ $ v4l2-ctl -v width=640,height=480
pi@rpi:~ $ v4l2-ctl -V
Format Video Capture:
        Width/Height      : 640/480
        Pixel Format      : 'pRAA' (10-bit Bayer RGRG/GBGB Packed)
        Field             : None
        Bytes per Line    : 800
        Size Image        : 384000
        Colorspace        : Raw
        Transfer Function : None
        YCbCr/HSV Encoding: ITU-R 601
        Quantization      : Full Range
        Flags             :
```

6by9:
> I have made some changes to the PR, following the pattern of IMX274. The big problem in doing so is 2 of the supported modes that are part of the driver internally crops the frame. This is in contrast to the role of S_FMT in comparison to S_SELECTION_CROP. The change I have made removes those modes and relies on S_FMT/S_SELECTION_COMPOSE for scaling the frame and S_SELECTION_CROP to crop the frame. However, the effective configuration can now be achieved using S_FMT and S_SELECTION_CROP in combination.
>
> Please let me know if this is an acceptable change.

Losing the direct selection of those modes is really a regression, and this is a mainline driver so regressions ought to be avoided. I'm nominally down as maintainer of imx219 in mainline, so need to try and play by the rules.
None of the userspace libraries are going to make use of the selection API at present either, so libcamera functionality has just failed as we can no longer get 1080p30.

Whilst the docs do state that a driver shouldn't alter the crop rectangle after initialisation, in this case I think we need to retain the full list of modes via imx219_enum_frame_size, and calling imx219_set_pad_format with (eg) 1920x1080 would select the 3280x2464 mode and also set the cropping appropriately.
I'll have a discussion with @naushir over that and see whether he agrees, or should there be a list of modes with selection parameters that should be used in libcamera instead (effectively an extension to the cam_helper functions).

I also note that we've lost the selection of the "x2-analog" binning in favour of the "x2-analog (special) binning" mode which was previously only used for the 640x480 high frame rate modes. (Registers 0x0174 and 0x175 being 0 for no binning, 1 for "x2 binning", 2 for "x4-binning", and 3 for "x2-analog (special) binning").
IIRC That is optimised for high frame rates, so we've most likely got a minor loss of image quality in normal preview mode.

vvarma:
> Losing the direct selection of those modes is really a regression

I agree that this solution would be controversial and break compatibility for users and force them towards the selection API. A solution that would be, i feel a good compromise between not breaking this compatibility and yet implementing the selection api correctly could be:
- revert `supported_modes` to the original 4
- `S_FMT` will behave as it originally did, with the added behaviour of setting the crop to the default_crop of the selected mode. 
- `S_SELECTION_CROP` and `S_SELECTION_COMPOSE` will remain in the current state - allowing user's to crop and scale to arbitrary frame sizes (1:1 or 1:2)
- the `change_compose` function will need to change to test only a subset (mode 2 and (4 or 7) of the original rpi firmware) of the `supported_modes`

> I also note that we've lost the selection of the "x2-analog" binning in favour of the "x2-analog (special) binning" mode

Since both these modes have the same effective scaling ratio, how should the driver select, between the two modes, when composing a target?

6by9:
I need to look through exactly what the implications would be, and how the crop and compose rectangles are representing.

What do we do if say we had full res selected but with a crop of top right quarter, and then you select the VGA mode. Where do the crop & compose selections end up then?
This is one of the bits that seems quirky with the API - the calls are allowed to amend the user's request, but is it required to amend the user's request due to the crop/compose, or alter the crop/compose to facilitate the user's request. How does the user know that the crop/compose has been altered due to the call?

I'm emailing linux-media to ask what the preferred solution is.

> Since both these modes have the same effective scaling ratio, how should the driver select, between the two modes, when composing a target?

Welcome to the wonderful world of trying to condense everything into one common API.
How do we know that the request is for binning and not skipping? If it were 4:1 skipping, then how would you request skipping 1 and then 7, vs 3 and then 5 as they introduce different amounts of distortion to the Bayer pattern? (2:1 skipping there's no real difference between 1/3 and 3/1)

6by9:
Sorry, I got held up with other things and have only just sent the enquiry to linux-media and libcamera-devel. Let's see what they come back with.
https://lists.libcamera.org/pipermail/libcamera-devel/2022-March/029316.html

vvarma:
@6by9 i dont see an update on the above thread. Do you think it is worth pursuing a middle ground along the lines i'd suggested above where we dont break the compatibility of `s_fmt` but implement `s_selection` as close to the spec as possible. This would make sense if backward compatibility is seen as a hard requirement and the specs as a best effort from the drivers perspective. 

The prime breach of spec would be:
- V4L2_SEL_TGT_CROP_DEFAULT will be set on each `s_fmt` and will also affect the current crop.

> How does the user know that the crop/compose has been altered due to the call?

The user would need to do a `g_selection` to know that the crop is still applied if they call `s_fmt` after a `s_selection`.

6by9:
I'll give it a day or two more and ping the thread.
I think my preference is to retain the list of modes and have S_FMT update the crop too.

> >   How does the user know that the crop/compose has been altered due to the call?

> The user would need to do a g_selection to know that the crop is still applied if they call s_fmt after a s_selection.

My problem with that is that things are persistent between different clients, and many do the bare minimum.
Client A set up some weird crop/compose selection and then quits.
Client B is a simpler app and just calls s_fmt. The images it gets are affected by Client A's configuration.
That's far more philosophical than we need to get on this PR though.

6by9:
Still no response from mainline, so time to do our own thing.

I'm happy to do that as long as we haven't changed the behaviour, which means ENUM_FMT and S_FMT have to accept all the existing resolutions and "do the right thing". s_selection on top then allows overriding the cropping.
Are you happy to implement that?

vvarma:
@6by9 have updated the PR. So the behaviour now is:
- S_FMT and ENUM_FMT no change - they can modify/reset the crop and compose behaviour
- S_SELECTION_CROP will change the default crop 
- S_SELECTION_COMPOSE will use the existing crop (set by S_SELECTION_CROP or S_FMT) and use 1x1 or 2x2 binning (Note: here it doesnt prefer the analog special binning mode - it selects the first best "goodness" - which will be the digital binning)
- CROP_BOUNDS are not affected by S_FMT and is set only at driver load
- COMPOSE_BOUNDS are limited by the crop selected

vvarma:
@6by9 do let me know if the above changes are more suitable or if you have any alternate suggestions.

6by9:
Sorry, I've been off sick and then stupidly busy.
I did get a response on linux-media in the end of basically "it's broken and it needs a lot of work to fix properly" - https://www.spinics.net/lists/linux-media/msg210545.html

I'll try to have a look at the patches early next week.

6by9:
Sorry for the delay in getting around to this.

Individual patches shouldn't create checkpatch issues. Making a mess and then adding a "checkpatch cleanup" patch isn't acceptable.
The "refactor by checkpatch" patch also seems to have changed a load of lines that weren't an issue in the first place.

You've resolved a load of my comments on use of ```__u32```, adding ```WARN``` lines, etc, but I don't see any changes based on them.

The reintroduction of the 1080p and VGA modes also isn't the approach I'd envisaged. I was expecting to only retain the 8MP and 1640x1232 register settings, but have a further table of resolutions with crops that they would apply. eg 1080p would be the 8MP register set, but would also set a crop of 
```
			.left = 688,
			.top = 700,
			.width = 1920,
			.height = 1080
```
that could be read back via the selection API. There are a number of approaches that could be adopted for how those tables are structured, but I haven't thought through which is likely to be most desirable.

A quick look would imply that we should have an array of ```base_modes``` which are probably just the ```reg_list``` elements. Please use an enum or define approach for numbering the array.
You then have an array of ```supported_modes``` which include the relevant crop to be applied to the selection API, and the index of which ```base_mode``` should be used for the register set.
```imx219_enum_frame_size``` then enumerates the full list of ```supported_modes```.
Likewise ```imx219_set_pad_format``` will search against that full list of formats, however it will set the format and selection based on which index into ```base_modes``` is used (Does that mean that ```base_modes``` needs the width/height to provide the full information?). If the format is one of the ```base_modes``` without crop, then don't reset the crop unless we need to.

If my brain has run through the behaviour correctly, then I think that also gets rid of the slightly nasty ```imx219_binning_goodness``` guess at what we should be doing, as ```v4l2_find_nearest_size``` will have chosen the mode for us.

vvarma:
thanks @6by9 for taking the time. 
I had kept the commits separate so that it would be easier to review them with the intent of squashing them later. Apologies, if thats not the norm. I have squashed the commits, ran checkpatch and have checked that they dont change any lines that arent part of this PR. I have also checked my usage of `__u32`.

In the approach you've suggested, we will need to also handle the special binning mode of the VGA format? 

Breaking down the approach you suggested modified to handle binning as well:
- Keep only the 8MP reg list
- create an enum for `binning_mode` - (1x1, 2x2 and 2x2_analog)
- Have a list of `supported_modes` with - crop, x_scaler, y_scaler, rate_factor, height, width (these two would be required for getting `v4l2_find_nearest_size` to work)  and `binning_mode`
- `imx219_enum_frame_size` will enumerate over the list of `supported_modes`
- `imx219_set_pad_format` will use `v4l2_find_nearest_size` to find the closest `supported_mode` to the requested frame size and set its crop and `binning_mode`
-  `set_selection` crop and compose will set the crop and the binning_mode if required 
- on start stream after the 8mp reg_list is applied,  `imx219_apply_selection` will apply the crop selected and a new `imx219_apply_binning` will apply the selected binning registers

>gets rid of the slightly nasty imx219_binning_goodness guess at what we should be doing, as v4l2_find_nearest_size will have chosen the mode for us

I do not see how we can use `v4l2_find_nearest_size` for the crop and compose apis. The `imx219_binning_goodness` is used only in the selection APIs and that since we need to handle arbitrary height and width in the crop and compose requests. `v4l2_find_nearest_size` on the other hand works in the `s_fmt` case since we are dealing with a set of discrete resolutions. 
For example consider the scenario:
- `s_fmt` 640x480
- `s_selection_crop` 600x400 - we use `imx219_binning_goodness` to identify that the format/compose values should be 600x400 ( or x_scaler=1, y_scaler=1)
- then if `s_selection_compose` with 350x250, `imx219_binning_goodness` is used to resolve to 300x200 and x_scaler=2, y_scaler=2

vvarma:
@6by9 Do let me know if the above approach seems sensible to you or if i have missed something important.

vvarma:
@6by9 Could you have another look at this PR?

vvarma:
Closing this PR for now.

6by9:
Converting from fixed modes to selection API should be discussed at the Linux Media Summit in September, so this will hopefully be resolved then.

PR #4936, opened 2022-03-14
Update DSI patches to latest sent to dri-devel

6by9:
Reverts the couple of patches that we had applied and applies the new versions. If we're still rebasing then the originals and reverts can be dropped.

I'm still awaiting further comment from upstream - I've had nothing from v2.

### DIFF
--- a/include/drm/drm_panel.h
+++ b/include/drm/drm_panel.h
@@ -179,6 +179,16 @@ struct drm_panel {
 	 * Panel entry in registry.
 	 */
 	struct list_head list;
+
+	/**
+	 * @prepare_upstream_first:
+	 *
+	 * The upstream controller should be prepared first, before the prepare
+	 * for the panel is called. This is largely required for DSI panels
+	 * where the DSI host controller should be initialised to LP-11 before
+	 * the panel is powered up.
+	 */
+	bool prepare_upstream_first;
 };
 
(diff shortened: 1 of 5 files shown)


include/drm/drm_panel.h
```diff
@@ -179,6 +179,16 @@ struct drm_panel {
...
+
+	/**
+	 * @prepare_upstream_first:
+	 *
+	 * The upstream controller should be prepared first, before the prepare
+	 * for the panel is called. This is largely required for DSI panels
+	 * where the DSI host controller should be initialised to LP-11 before
+	 * the panel is powered up.
+	 */
+	bool prepare_upstream_first;
```
pelwell, line 190: Is this defined to default to false/0 because of the way the structures are allocated? A casual read of this patch set makes it appear that prepare_upstream_first is being used uninitialised.

6by9: Yes, struct drm_panel is almost always embedded within the panel driver's state structure, which will have been kzalloc'ed or similar.

It could be added to ```drm_panel_init```, but then the same could be said for the ```struct backlight_device``` pointer. (```orientation``` is a field that is currently only in the downstream tree through 52904afeb0d6b54d06ecf67078b30cfa12598095. I thought it had been upstreamed, but I can't see it at present).

PR #4937, opened 2022-03-14
vc4/drm: Add support for fractional scaling and chroma siting

popcornmix:
When chroma is subsampled compared to luma, there are a number of choices of where the chroma is centred.
Different codecs make different choices for this. From ffmpeg:
```
 /**
  * Location of chroma samples.
  *
  * Illustration showing the location of the first (top left) chroma sample of the
  * image, the left shows only luma, the right
  * shows the location of the chroma sample, the 2 could be imagined to overlay
  * each other but are drawn separately due to limitations of ASCII
  *
  *                1st 2nd       1st 2nd horizontal luma sample positions
  *                 v   v         v   v
  *                 ______        ______
  *1st luma line > |X   X ...    |3 4 X ...     X are luma samples,
  *                |             |1 2           1-6 are possible chroma positions
  *2nd luma line > |X   X ...    |5 6 X ...     0 is undefined/unknown position
  */
 enum AVChromaLocation {
     AVCHROMA_LOC_UNSPECIFIED = 0,
     AVCHROMA_LOC_LEFT        = 1, ///< MPEG-2/4 4:2:0, H.264 default for 4:2:0
     AVCHROMA_LOC_CENTER      = 2, ///< MPEG-1 4:2:0, JPEG 4:2:0, H.263 4:2:0
     AVCHROMA_LOC_TOPLEFT     = 3, ///< ITU-R 601, SMPTE 274M 296M S314M(DV 4:1:1), mpeg2 4:2:2
     AVCHROMA_LOC_TOP         = 4,
     AVCHROMA_LOC_BOTTOMLEFT  = 5,
     AVCHROMA_LOC_BOTTOM      = 6,
     AVCHROMA_LOC_NB,              ///< Not part of ABI
 };
  ```
To accurately display decoded video form these different sources, we need to handle this at display end.

Also DRM provides fractional source coordinates for displayed planes.

Both of these can be handled using the phase field of HVS.

### CODE REVIEW

**Comment by popcornmix** [comment]
This supersedes https://github.com/raspberrypi/linux/pull/4852 which was based on 5.10 kernel, and includes updates based on comments there.

The last commit is useful for testing and should be dropped before merge.

File: drivers/gpu/drm/vc4/vc4_plane.c [hunk 1]
```diff
@@ -528,13 +563,15 @@ static void vc4_write_scaling_parameters(struct drm_plane_state *state,
 	/* Ch0 H-PPF Word 0: Scaling Parameters */
 	if (vc4_state->x_scaling[channel] == VC4_SCALING_PPF) {
 		vc4_write_ppf(vc4_state,
-			      vc4_state->src_w[channel], vc4_state->crtc_w);
+			      vc4_state->src_w[channel], vc4_state->crtc_w, vc4_state->src_x, channel,
+				  siting_h != ~0 ? siting_h : state->chroma_siting_h);
```
pelwell: Has something peculiar happened to the whitespace on this line (and 574 below)?

include/drm/drm_plane.h
```diff
@@ -179,6 +179,24 @@ struct drm_plane_state {
 	 */
 	enum drm_color_range color_range;
 
+	/**
+	 * @chroma_siting_h:
+	 *
+	 * Location of chroma samples horizontally compared to luma
+	 * 0 means chroma is sited with left luma
+	 * 0x8000 is intersitial. 0x10000 is sited with right luma
```
pelwell: `interstitial`?

pelwell (commented):
This is largely outside my comfort zone, but it looks plausible. Assuming it works for you then I just have two cosmetic issues - whitespace and a typo.

include/drm/drm_plane.h
pelwell: Also repeated below.

drivers/gpu/drm/vc4/vc4_plane.c
```diff
@@ -458,14 +458,46 @@ static void vc4_write_tpz(struct vc4_plane_state *vc4_state, u32 src, u32 dst)
 			VC4_SET_FIELD(recip, SCALER_TPZ1_RECIP));
 }
 
-static void vc4_write_ppf(struct vc4_plane_state *vc4_state, u32 src, u32 dst)
+#define PHASE_BITS 6
```
6by9, line 462: Phase magnitude bits. There's a sign bit on top.

I was having to think what lines 493-496 were doing, as they seemed to be duplication. Ensure +ve values don't touch the sign bit, then truncate negative values.

It'd be worth a comment to that effect.

drivers/gpu/drm/vc4/vc4_plane.c
```diff
@@ -458,14 +458,46 @@ static void vc4_write_tpz(struct vc4_plane_state *vc4_state, u32 src, u32 dst)
...
+static void vc4_write_ppf(struct vc4_plane_state *vc4_state, u32 src, u32 dst, u32 xy, int channel)
 {
-	u32 scale = (1 << 16) * src / dst;
+	u32 scale = (src / dst) >> (16 - SCALE_BITS);
+	s32 offset, offset2;
+	s32 phase;
+
+	/* Start the phase at 1/2 pixel from the 1st pixel at src_x.
+	   1/4 pixel for YUV. */
+	if (channel != 0) {
```
6by9: Kernel preference would be for ```if(channel)```. No need for the !=0

drivers/gpu/drm/vc4/vc4_plane.c
```diff
@@ -458,14 +458,46 @@ static void vc4_write_tpz(struct vc4_plane_state *vc4_state, u32 src, u32 dst)
...
+		offset += -(1 << PHASE_BITS >> 2);
+	} else {
+		/* the phase is relative to scale_src->x, so shift it for display list's x value */
+		offset = (xy & 0xffff) >> (16 - PHASE_BITS);
+		offset += -(1 << PHASE_BITS >> 1);
+	}
+
+	/* This is a kludge to make sure the scaling factors are consitent with YUV's luma scaling.
+	   we lose 1bit precision because of this. */
+	if (channel == 0)
```
6by9: ```if (!channel)```

include/drm/drm_plane.h
```diff
@@ -179,6 +179,24 @@ struct drm_plane_state {
...
+	 */
+	int32_t chroma_siting_h;
+
+	/**
+	 * @chroma_siting_v:
+	 *
+	 * Location of chroma samples vertically compared to luma
+	 * 0 means chroma is sited with top luma
+	 * 0x8000 is intersitial. 0x10000 is sited with bottom luma
```
6by9: If these are the expected values, then why is the range -1<<16, 1<<16 when creating the property with ```drm_property_create_signed_range```?

Defining as a signed range, and then putting apparently positive hex values in here is confusing.

include/drm/drm_plane.h
```diff
@@ -750,6 +768,24 @@ struct drm_plane {
 	 * scaling.
 	 */
 	struct drm_property *scaling_filter_property;
+
+	/**
+	 * @chroma_siting_property_h:
+	 *
+	 * Optional "CHROMA_SITING_H" property for specifying
+	 * chroma siting for YUV formats.
+	 * See drm_plane_create_chroma_properties().
```
6by9: See drm_plane_create_chroma_siting_properties

include/drm/drm_plane.h
```diff
@@ -750,6 +768,24 @@ struct drm_plane {
...
+	 */
+	struct drm_property *chroma_siting_h_property;
+
+	/**
+	 * @chroma_siting_property_v:
+	 *
+	 * Optional "CHROMA_SITING_V" property for specifying
+	 * chroma siting for YUV formats.
+	 * See drm_plane_create_chroma_properties().
```
6by9: Ditto

drivers/gpu/drm/vc4/vc4_plane.c
```diff
@@ -1621,6 +1624,8 @@ struct drm_plane *vc4_plane_init(struct drm_device *dev,
 					  DRM_COLOR_YCBCR_BT709,
 					  DRM_COLOR_YCBCR_LIMITED_RANGE);
 
+	drm_plane_create_chroma_siting_properties(plane, 0, 0);
```
6by9, line 1628: Do the default properties get set based on __drm_atomic_helper_plane_state_reset?

Ah, that's not been merged into the tree yet. I thought I'd backported it.
https://cgit.freedesktop.org/drm/drm-misc/log/drivers/gpu/drm/drm_atomic_state_helper.c https://cgit.freedesktop.org/drm/drm-misc/commit/drivers/gpu/drm/drm_atomic_state_helper.c?id=1a7998dab5dd3d11bada7e3921781922082e7fe6

Ah, was part of #4846, and not backported since Maxime took it over and extended it. I ought to do that.

drivers/gpu/drm/vc4/vc4_plane.c
```diff
@@ -28,6 +28,12 @@
 #include "vc4_drv.h"
 #include "vc4_regs.h"
 
+#include <linux/module.h>
+#include <linux/moduleparam.h>
+static int siting_h = ~0, siting_v = ~0;
+module_param(siting_h, int, 0644);
+module_param(siting_v, int, 0644);
```
6by9: Do we really need runtime override of this other than for debugging the original code? Can we just change the default property values instead?

include/drm/drm_plane.h
popcornmix: I'll change range to 0 to 1<<16. I don't think it make sense to exceed those.
You could also treat 0 as interstitial and have range from -1<<15 to 1<<15.

drivers/gpu/drm/vc4/vc4_plane.c
popcornmix: I commented this commit should be dropped prior to merge.
It was useful for me (so I could see effect of tweaking values, say on a paused frame in kodi).
I wasn't sure if you can modify plane properties you haven't created through modetest.

popcornmix:
Updated to address comments.
Note there is a chroma siting test [here](https://github.com/jc-kynesim/drmu/tree/master)

pelwell:
It looks OK to me. @6by9?

drivers/gpu/drm/vc4/vc4_plane.c
```diff
@@ -458,14 +458,48 @@ static void vc4_write_tpz(struct vc4_plane_state *vc4_state, u32 src, u32 dst)
...
+		offset = (xy & 0x1ffff) >> (16 - PHASE_BITS) >> 1;
+		offset += -(1 << PHASE_BITS >> 2);
+	} else {
+		/* the phase is relative to scale_src->x, so shift it for display list's x value */
+		offset = (xy & 0xffff) >> (16 - PHASE_BITS);
+		offset += -(1 << PHASE_BITS >> 1);
+
+		/* This is a kludge to make sure the scaling factors are consitent with YUV's luma scaling.
+		   we lose 1bit precision because of this. */
+		if (!channel)
```
6by9: Has this moved? We're doing a check for ```if (!channel)``` when we're in the ```else``` of an ```if (channel)``` clause.

drivers/gpu/drm/drm_color_mgmt.c
```diff
@@ -586,6 +586,42 @@ int drm_plane_create_color_properties(struct drm_plane *plane,
...
+ * properties to @plane.
+ */
+int drm_plane_create_chroma_siting_properties(struct drm_plane *plane,
+						int32_t default_chroma_siting_h,
+						int32_t default_chroma_siting_v)
+{
+	struct drm_device *dev = plane->dev;
+	struct drm_property *prop;
+
+	prop = drm_property_create_signed_range(dev, 0, "CHROMA_SITING_H",
```
6by9: Upstream may prefer just ```drm_property_create_range``` as the range is now only positive, but I'm not fussed.

drivers/gpu/drm/vc4/vc4_plane.c
6by9: #4846 has been updated and merged. It'd be worth adding the restore to ```__drm_atomic_helper_plane_state_reset``` as part of "drm: Add chroma siting properties" to ensure the default values are loaded. It'll make no difference in this case as the default is 0,0, so can come later.

6by9 (commented):
1 thing to check with that conditional. The other two are minor and I'm not fussed over.

popcornmix:
@6by9 updated

6by9:
All looks good to me now.

PR #4939, opened 2022-03-15
Integrate ancillary links

jhautbois:
Daniel Scally started his cover letter with [1]:
"At present there's no means in the kernel of describing the supporting relationship between subdevices that work together to form an effective single unit - the type example in this case being a camera sensor and its corresponding vcm."

This series makes it possible to have a strong relationship between a sensor and its vcm (to control focus) or flash light. I have adapted it for v5.15 as the mainline version now calls v4l2_async_nf_call_unbind() and not v4l2_async_notifier_call_unbind().

[1]: https://lore.kernel.org/all/20220302220304.1327896-1-djrscally@gmail.com/

### DIFF
--- a/Documentation/userspace-api/media/mediactl/media-controller-model.rst
+++ b/Documentation/userspace-api/media/mediactl/media-controller-model.rst
@@ -33,3 +33,9 @@ are:
 
 -  An **interface link** is a point-to-point bidirectional control
    connection between a Linux Kernel interface and an entity.
+
+- An **ancillary link** is a point-to-point connection denoting that two
+  entities form a single logical unit. For example this could represent the
+  fact that a particular camera sensor and lens controller form a single
+  physical module, meaning this lens controller drives the lens for this
+  camera sensor.
--- a/Documentation/userspace-api/media/mediactl/media-types.rst
+++ b/Documentation/userspace-api/media/mediactl/media-types.rst
@@ -412,14 +412,21 @@ must be set for every pad.
 	  is set by drivers and is read-only for applications.
 
     *  -  ``MEDIA_LNK_FL_LINK_TYPE``
-       -  This is a bitmask that defines the type of the link. Currently,
-	  two types of links are supported:
+       -  This is a bitmask that defines the type of the link. The following
+	  link types are currently supported:
 
 	  .. _MEDIA-LNK-FL-DATA-LINK:
 
-	  ``MEDIA_LNK_FL_DATA_LINK`` if the link is between two pads
+	  ``MEDIA_LNK_FL_DATA_LINK`` for links that represent a data connection
+     between two pads.
 
 	  .. _MEDIA-LNK-FL-INTERFACE-LINK:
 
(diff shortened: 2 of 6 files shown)


6by9:
These are all part of https://patchwork.linuxtv.org/project/linux-media/list/?series=7431 which, whilst it has R-B tags for all patches, appears not to have been merged as yet.

Ideally we'd like the patches to be merged first, and we amend the commit text to list the upstream commit hash to simplify maintenance. Any guesses on the timescale for the merge?

jhautbois:
Yes, it is on its way, we wanted it to be integrated quickly in the linux-rpi tree because we don't really know when it will be merged to be honest.

kbingham:
Of course it's up to you when to merge this series in your kernel, but at least the work required to adapt to the API changes has been done here for you.

The support required to complement this in libcamera has been merged, ahead of mainline integration, so merging this will allow development to work on AF for RPi devices. I suspect this is only of interest to DavidP at the moment on your side, so I don't think there's any particular urgency to merge, but it's here if you need or want it.

6by9:
Thanks guys.
I have no issue with merging it (I have a near identical set of patches on my local tree), more was querying if there was any knowledge of when it will be merged to linux-media. Those merges always seem to get backed up until just before the merge window.
If the libcamera side has been merged, then it makes a stronger case to merge these kernel patches.

I have a https://www.amazon.co.uk/Arducam-Camera-Autofocus-Raspberry-Motorized/dp/B07SN8GYGD module on my desk and DT overlay changes to support it. There are a couple of other 3rd party modules kicking about with focus VCMs (eg Arducam [IMX298](https://www.arducam.com/product/arducam-pivariety-16mp-imx298-color-camera-module-for-rpi-4b-3b-2b-3a-pi-zero-cm3-cm4/), [IMX219](https://www.arducam.com/product/arducam-imx219-auto-focus-camera-module-drop-in-replacement-for-raspberry-pi-v2-and-nvidia-jetson-nano-camera/), and [IMX477](https://www.arducam.com/product/arducam-12mp-imx477-motorized-focus-high-quality-camera-for-raspberry-pi/)), so may see where I get to on those too.
I only worry about getting the kernel side working, @davidplowman gets to worry about the actual algorithms :-)

kbingham:
If for some horrible reason there is a drastic change in kernel API before merge then we'd have to account for that in both libcamera and your kernel - but I think that's really low risk here, and even if it does happen - it's all under our control so I don't think it would be too painful.

And given that it amounts to only a single line addition in the UAPI: 
 https://github.com/raspberrypi/linux/pull/4939/commits/5c7b9fd68c1f31133c8a415c7caeabf55fefa71b

I don't think it will be too problematic.

6by9:
I'm good with it.
@pelwell Any objections from you?

pelwell:
I hop they get merged upstream unmodified...

6by9:
@kbingham @jhautbois  FWIW #4941 adds the relevant overlay hooks for the Arducam OV5647 and IMX219 modules.

PR #4940, opened 2022-03-16
Clock Framework Improvements, and removal of clock requests API

mripard:
Hi,

This is a PR that brings the changes that have been accepted into upstream to bring the same feature provided by the clock request API we have in the downstream tree.

The approach taken has been different though and has instead relied on fixing the clk_set_rate_range / clk_set_min_rate functions instead.

So this PR contains both the new changes aimed for 5.18, and reverts the API we had before to sync the tree with upstream.

This also fixes the cursorA-vs-flipA-toggle subtest of the kms_cursor_legacy test suite in IGT that was producing a null pointer dereference on the clock request pointer.

Let me know what you think,
Maxime


drivers/gpu/drm/vc4/vc4_kms.c
```diff
@@ -436,14 +431,6 @@ static void vc4_atomic_commit_tail(struct drm_atomic_state *state)
 		drm_dbg(dev, "Running the core clock at %lu Hz\n",
 			new_hvs_state->core_clock_rate);
 
-		/*
-		 * Request a clock rate based on the current HVS
-		 * requirements.
-		 */
-		hvs->core_req = clk_request_start(hvs->core_clk,
```
popcornmix: Should this be replaced by a call to `clk_set_min_rate`?

Gido97081214 (approved):

popcornmix:
I think this PR is okay. In testing I found a case where pixel-bvb clock wasn't being set.
Manually setting that through mailbox brought screen back.
But I think the fix for this is probably firmware side and isn't introduced by this PR.

drivers/gpu/drm/vc4/vc4_kms.c
mripard: It was added in the next patch but you're right that it makes more sense to have it there. I've reworked the branch a bit and pushed a new version.

popcornmix:
@mripard after this PR I can't use a 4kp60 display.

A bit of debugging shows we set `vc4_hdmi->disable_4kp60 = true` as `clk_round_rate(vc4->hvs->core_clk, 550000000)` returns 220000000. I have `hdmi_enable_4kp60=1` in config.txt.

Not quite sure why that doesn't give a working non-4kp60 mode (no mode is set at all).

Without this PR, `clk_round_rate(vc4->hvs->core_clk, 550000000)` returns 55000000 and 4kp60 works.

mripard:
> @mripard after this PR I can't use a 4kp60 display.
> 
> A bit of debugging shows we set `vc4_hdmi->disable_4kp60 = true` as `clk_round_rate(vc4->hvs->core_clk, 550000000)` returns 220000000. I have `hdmi_enable_4kp60=1` in config.txt.

I found what's going on, and I'm not sure how to fix it properly. We were using clk_round_rate to infer whether hdmi_enable_4kp60 was enabled or not since the maximum was higher when enabled.

However, the clock changes we did will now try to always run at the minimum frequency, so we'll end up here with 220MHz, which is the minimum allowed by the firmware at that point in time.

> Not quite sure why that doesn't give a working non-4kp60 mode (no mode is set at all).

Even we couldn't detect the parameter properly, this should work though.

mripard:
> Not quite sure why that doesn't give a working non-4kp60 mode (no mode is set at all).

Actually, it makes sense: the firmware will have used 4kp60 and will pass that on the kernel command line. However, the kernel wrongly assumes it can't do that, so the framebuffer emulation will still try to set that mode and the kernel will reject it. We then end up with a black screen.

popcornmix:
> Actually, it makes sense: the firmware will have used 4kp60 and will pass that on the kernel command line. However, the kernel wrongly assumes it can't do that, so the framebuffer emulation will still try to set that mode and the kernel will reject it. We then end up with a black screen.

Ah yes. `disable_fw_kms_setup=1` in config.txt avoids that (and we end up with 4kp30).

> I found what's going on, and I'm not sure how to fix it properly. We were using clk_round_rate to infer whether hdmi_enable_4kp60 was enabled or not since the maximum was higher when enabled.

We still have the number wanted with this PR
```
pi@pi4:~ $ sudo cat /sys/kernel/debug/clk/fw-clk-core/clk_max_rate
550000000
```
is there not a way of getting hold of that from the kms driver?

mripard:
> ```
> pi@pi4:~ $ sudo cat /sys/kernel/debug/clk/fw-clk-core/clk_max_rate
> 550000000
> ```
> 
> is there not a way of getting hold of that from the kms driver?

Not really, there's no consumer API to retrieve the clock boundaries. I've pushed a new version that will try to raise the minimum to 550MHz. If we don't have enable_hdmi_4kp60, the range reported by the firmware (120-500) will be disjoint with the range we're trying to set and it will error out. If it's enable, the range will become 550-550 and will thus work.

popcornmix:
@mripard updated version of this patch is not working for me.
With 4kp60 + 1920x1200@60 display I get http://sprunge.us/DRnXm1
It looks like new code gets a valid `core_clk` and `clk_set_min_rate` returns `0` (and the bind function is called 3 times), but no displays and flip done timeout.
pi@pi4:~ $ clk | grep fw
Looking at the fw clocks:
```
 fw-clk-vec                           0        0        0           0          0     0  50000         Y
 fw-clk-pixel-bvb                     2        2        0   300000000          0     0  50000         Y
 fw-clk-m2mc                          2        2        0   599940000          0     0  50000         Y
 fw-clk-hevc                          0        0        0   250000000          0     0  50000         Y
 fw-clk-pixel                         0        0        0           0          0     0  50000         Y
 fw-clk-v3d                           0        0        0   250000000          0     0  50000         Y
 fw-clk-core                          1        1        0   220000000          0     0  50000         Y
 fw-clk-arm                           0        0        0   600000000          0     0  50000         Y
```
The core is being set to 220MHz which is too low for 4kp60 (+1920x1080p).

Interestingly when using just a 4kp60 display I get:
```
fw-clk-vec                           0        0        0           0          0     0  50000         Y
 fw-clk-pixel-bvb                     1        1        0   300000000          0     0  50000         Y
 fw-clk-m2mc                          1        1        0   599940000          0     0  50000         Y
 fw-clk-hevc                          0        0        0   250000000          0     0  50000         Y
 fw-clk-pixel                         0        0        0           0          0     0  50000         Y
 fw-clk-v3d                           0        0        0   250000000          0     0  50000         Y
 fw-clk-core                          1        1        0   534600000          0     0  50000         Y
 fw-clk-arm                           0        0        0   600000000          0     0  50000         Y
```
so, core is 534MHz. It feels like the second display is replacing the core clock requirement (rather than adding to it).

mripard:
Actually, the issue is that the clk_set_min_rate calls fail in vc4_atomic_commit_tail.

Indeed, we try to set a minimum of 683 MHz, which is our estimate of what the core clock should be. Since it's way higher than the maximum that the firmware report, clk_set_min_rate will fail, and thus we never have our minimum enforced.

Now the interesting part is that if I cap the HVS rate we try to enforce to 550MHz, it seems to work just fine. So it looks like our calculation of the core clock is completely off.

For reference, this is the calculation we have at the moment:
https://github.com/raspberrypi/linux/blob/rpi-5.15.y/drivers/gpu/drm/vc4/vc4_crtc.c#L712

In our case, HDMI0 runs at 3840x2160@60Hz, so it gives a load of 594000000 * 9 / 10 = 534600000.0

HDMI1 will run at 1920x1080@60Hz, so 148500000.

We'll do the sum here:
https://github.com/raspberrypi/linux/blob/rpi-5.15.y/drivers/gpu/drm/vc4/vc4_kms.c#L620
and here:
https://github.com/raspberrypi/linux/blob/rpi-5.15.y/drivers/gpu/drm/vc4/vc4_kms.c#L901

and thus, we end up with 683100000.0 Hz.

popcornmix:
I think the firmware effectively does `max(534600000, 148500000)` rather than sum for the part related to display timings. The part related to input planes is summed across the displays.

Probably due to HVS processing planes sequentially, compared to Pixel Valve driving displays in parallel.
@timg236 may be able to explain this better.

The numbers are somewhat heuristic, and won't be exact, so it would be better to choose the closest frequency (i.e. max) we can, rather than dropping a too high request.

mripard:
Yeah, this would make sense that it would be the max. I'll let @timg236 confirm.

I've also added a warning when clk_set_min_rate fails to at least complain instead of ignoring it. And I'm currently reworking it to max it to what the firmware will allow us to.

jc-kynesim:
Just to check - this new API has no way of asking for whatever the maximum supported clock is? The HEVC clock really wants that. There isn't a "good" rate I can calculate beforehand, and in nearly all cases faster is better.

mripard:
> Just to check - this new API has no way of asking for whatever the maximum supported clock is?

It's not a new API, it was already there in the kernel. We changed its behaviour for some of our clocks though

And no, mostly because it would be racy and thus wouldn't be very useful. Even if you were to retrieve and ask for the maximum, there's nothing that would prevent another user from changing that maximum before you get the chance to ask for that maximum.

> The HEVC clock really wants that. There isn't a "good" rate I can calculate beforehand, and in nearly all cases faster is better.

So you just want to run the clock at the highest frequency all the time? That doesn't sound very power efficient, but we can do that fairly easily by doing something like https://github.com/raspberrypi/linux/pull/4940/commits/1c195333fbc38f30949f0b58cdfe2ffb8a3d1695

jc-kynesim:
> So you just want to run the clock at the highest frequency all the time? That doesn't sound very power efficient, but we can do that fairly easily by doing something like https://github.com/raspberrypi/linux/commit/1c195333fbc38f30949f0b58cdfe2ffb8a3d1695

I want to run the clock at max while I am decoding - i.e. when I get stream start start it at max, when stop reset to 0 - short of exciting (prone to error) code that measures Q lengths and adjusts the clock dynamically I can't think of a better idea - can you?

mripard:
Like I said this would be possible by implementing something like https://github.com/raspberrypi/linux/commit/1c195333fbc38f30949f0b58cdfe2ffb8a3d1695, but I'm not sure it needs to be part of this PR. It wasn't possible before that PR either.

popcornmix:
@mripard it is currently possible (prior to this PR). 
https://github.com/raspberrypi/linux/blob/6dafd553901ea01d8871010488121e21d131fea4/drivers/staging/media/rpivid/rpivid_video.c#L567
gets the max hevc clock frequency (which is 500MHz by default, 550MHz when using `hdmi_enable_4kp60` and may be higher still with a custom config.txt setting).

mripard:
Right, but that still works?

The HEVC clock will just take the boundaries reported by the firmware and clamp ULONG_MAX to those boundaries.

popcornmix:
Ah - I assumed because we couldn't determine the max core clock, we wouldn't be able to determine the max hevc clock.

mripard:
We don't have that option for KMS because the core clock will always return the minimum available so it's not just a clamp.

It's the only clock in that configuration at the moment, so all the others will just return the rate asked for, assuming it's within the boundaries reported by the firmware.

pelwell:
What's the status on this PR? As I understand things, it (or something like it) is needed, but as it stands dual monitor support (at least in some configurations) is broken.

timg236:
> Yeah, this would make sense that it would be the max. I'll let @timg236 confirm.
> 
> I've also added a warning when clk_set_min_rate fails to at least complain instead of ignoring it. And I'm currently reworking it to max it to what the firmware will allow us to.

On 2711 HVS runs off the the core-clk and the minimum core clock frequency must be the maximum of pixel reading input stage and the composition output stage. The composition stage happens in parallel and reads from the Composition Output Buffer. This must be at least as fast as the highest active pixel rate so 550 MHz on HDMI0 for 4Kp60. However, the PV for HDMI1 has a much smaller FIFO so it would need to be at least as fast as the display pixel-clock on HDMI1.

The input stage is roughly the sum of all the input planes but AFAIK there's no scheduling and it's hard to fully utilise the bus for two displays so you have to add some overhead percentage. You also have to take scaling into account. The firmware numbers for the input stage were generated by experimentally finding a stable clock configuration for common use-cases e.g. dual display + camera + OMX video.

jc-kynesim:
What needs to be done to progress this branch? We've got a bug in the existing code that can cause the linked list of clock requests to grow indefinitely that ends up using enough CPU to break timely video decode at high bitrates.  This patch fixes the issue by removing that logic.  I could search for the previous bug but there doesn't seem a lot of point if the fix is almost certainly in code that is going to vanish.

mripard:
> What's the status on this PR? As I understand things, it (or something like it) is needed, but as it stands dual monitor support (at least in some configurations) is broken.

Yes, dual output is broken with that PR and we needed some input from @timg236 to get going. We shouldn't merge this version of the PR.



> > Yeah, this would make sense that it would be the max. I'll let @timg236 confirm.
> > I've also added a warning when clk_set_min_rate fails to at least complain instead of ignoring it. And I'm currently reworking it to max it to what the firmware will allow us to.
> 
> On 2711 HVS runs off the the core-clk and the minimum core clock frequency must be the maximum of pixel reading input stage and the composition output stage. The composition stage happens in parallel and reads from the Composition Output Buffer. This must be at least as fast as the highest active pixel rate so 550 MHz on HDMI0 for 4Kp60. However, the PV for HDMI1 has a much smaller FIFO so it would need to be at least as fast as the display pixel-clock on HDMI1.

I think we cover the difference between HDMI0 and HDMI1 by using:
https://github.com/raspberrypi/linux/blob/rpi-5.15.y/drivers/gpu/drm/vc4/vc4_crtc.c#L712

And we then associate to each FIFO the frequency we generated before:
https://github.com/raspberrypi/linux/blob/rpi-5.15.y/drivers/gpu/drm/vc4/vc4_kms.c#L925

And finally we sum everything here:
https://github.com/raspberrypi/linux/blob/rpi-5.15.y/drivers/gpu/drm/vc4/vc4_kms.c#L954

If the output part all happens in parallel, then we should move that to take the max, not the sum, right? 

> The input stage is roughly the sum of all the input planes but AFAIK there's no scheduling and it's hard to fully utilise the bus for two displays so you have to add some overhead percentage. You also have to take scaling into account. The firmware numbers for the input stage were generated by experimentally finding a stable clock configuration for common use-cases e.g. dual display + camera + OMX video.

Ok, so it's sum of all input, but max of all the output

mripard:
> What needs to be done to progress this branch? We've got a bug in the existing code that can cause the linked list of clock requests to grow indefinitely that ends up using enough CPU to break timely video decode at high bitrates. This patch fixes the issue by removing that logic. I could search for the previous bug but there doesn't seem a lot of point if the fix is almost certainly in code that is going to vanish.

We were waiting for a clarification, and I'm going to push a new version some time this week

timg236:
"If the output part all happens in parallel, then we should move that to take the max, not the sum, right?"

The maximum would be better. It's not really in parallel, I think it interleaves the scaling / composition of the display FIFOs but runs with an internal data rate of up to 4 pixels per-clock unscaled or 2 pixels per clock scaled. The scheduling and memory fetching won't be perfect so you have to allow some overhead for that.

It might be best to allow the display to slightly overshoot the max-clock rate and clamp that before rejecting elements because it's just too complicated to get a perfect model of the HVS clocks on an unconstrained system (i.e. desktop Linux).

mripard:
I've pushed a new version that addresses the issue @popcornmix pointed out, with more fixes submitted upstream that were sent last week.

popcornmix:
I think we want `.minimize = true` on all the firmware clocks.
Without it (I believe), for example the hevc driver requests a hevc clock of 500MHz, when it stops it drops that request, but the clock stays at 500MHz. That clock uses power (and so raises temperature) and stops the ability to reduce voltage (only possible when arm + all fw clocks are low).

jc-kynesim:
If we do .minimize = true then doesn't that remove hevcs ability to set its clock to whatever the configured maximum is?

popcornmix:
> If we do .minimize = true then doesn't that remove hevcs ability to set its clock to whatever the configured maximum is? 

That is my understanding, but we'd like to do both (minimise clocks when they aren't needed and have a way to request the max clock so you can get a job done as quickly as possible).

jc-kynesim:
> > If we do .minimize = true then doesn't that remove hevcs ability to set its clock to whatever the configured maximum is?
> 
> That is my understanding, but we'd like to do both (minimise clocks when they aren't needed and have a way to request the max clock so you can get a job done as quickly as possible).

Good - as long as it is clear that I don't want .minimize on hevc until there is a way to set the clock to maximum on it once you have (I know you know this but it is sufficiently important that I need it in writing here too).

jc-kynesim:
One further problem with this patch and HEVC is that the previous scheme allowed for multiple HEVC decodes to run at the same time and each had a clock request active which meant that the clock was only released when none were active. This scheme means I should need to have a ref-counter on the clock claims (as I did before the previous scheme turned up) though the fact that the clock doesn't drop when released masks the problem.

I have a fix for this but given that it doesn't currently cause a problem I think it should be added as later patch rather than as part of this set. (it will be needed if/when .minimize is set on HEVC)

And to be clear I think the .minimize issues on HEVC clocks could be dealt with by a later patch as I'm keen for this patch to make it.

mripard:
> > If we do .minimize = true then doesn't that remove hevcs ability to set its clock to whatever the configured maximum is?
> 
> That is my understanding, but we'd like to do both (minimise clocks when they aren't needed and have a way to request the max clock so you can get a job done as quickly as possible).

It doesn't prevent to set it to the maximum per se, but it prevents from discovering that maximum. If we have a limited number of possible maximum available like we do for the core clock (basically either 500 or 550MHz), we could figure it out by trial and error.

But then, if the clock isn't needed, can't we just disable it? If we can, then the rate while inactive isn't really important anymore

mripard:
> One further problem with this patch and HEVC is that the previous scheme allowed for multiple HEVC decodes to run at the same time and each had a clock request active which meant that the clock was only released when none were active.

I'm not sure what you mean there by the clock was released? the clk_request API was orthogonal to whether or not the clock was enabled, it was only about its rate.

> This scheme means I should need to have a ref-counter on the clock claims (as I did before the previous scheme turned up) though the fact that the clock doesn't drop when released masks the problem.

Whether the clock is enabled or not is already reference-counted, so it you do multiple clk_enable() calls, it will only get actually disabled if and when you have the same number of clk_disable() calls. 

This is fairly orthogonal to this series though. If I understand you right, what you'd like in the former API was that you could prepare a frame decoding, attach the clock rate request to that submission, and free it whenever it was decoded?

The clock rate ranges are local to the struct clk, so you can do something similar by calling clk_get() each and every time, clk_set_min_rate() before starting to decode that frame, and clk_put when it's done.

jc-kynesim:
OK - I'll admit I really don't understand exactly how this is meant to work and that is most of the problem.

So as you saying that if I have
`start_clock() {
    clk_set_min_rate();
    clk_prepare_enable();    
}

stop_clock() {
  clk_disable_unprepare();
}`

then I can call start_clock multiple times and the clock will only go away when I have a matching number of stop_clocks?

If that is the case then maybe I don't even need the set_min rate as the only rate I ever want is the maximum available and that can be set in the clock definition?

To answer the question you actually asked - the video device can be opened multiple times and each open does its own stream_start/stream_stops and I need the clock active when any stream is in the stream_on state.

mripard:
So, the clock is modelled through two main structures, `clk_hw` and `clk`. `clk_hw` is unique and is handled by the framework and drivers.

Then, "users" are handling `clk`. `clk` instances are created for each user.

The enable refcounting happens at the `clk_hw`, so if any user called clk_enable on their `clk`, it will be kept on.

Ranges are set on both `clk_hw` (for the hardware constraints), and on `clk`. The actual range is thus the maximum of all the `clk` minimums, and the minimum of all the `clk` maximums (plus whatever `clk_hw` has).

You cannot have multiple ranges for a given `clk`, but you can create multiple `clk` in your driver.

So it should work just fine if you have something like

```c
struct rpivid_ctx {
    ....
    struct clk *clk;
};

static int start_clock(struct rpivid_dev *dev, struct rpivid_ctx *ctx)
{
    ctx->clk = clk_get(dev->dev, "hevc");
    if (IS_ERR(ctx->clk))
        return PTR_ERR(ctx->clk);

    ret = clk_set_min_rate(ctx->clk, RATE);
    if (ret)
        goto err_put_clk;

    ret = clk_prepare_enable(ctx->clk);
    if (ret)
        goto err_drop_range;

    return 0;

err_drop_range:
err_clk:
    clk_put(ctx->clk);
    return ret;
}
```

Then, when you start a new request you create a new clock user (and thus a new `clk` instance), setup a range for that clock, and then asks it to be enabled (increasing the reference counting).

Whenever you're done,

```c
static void stop_clock(struct rpivid_dev *dev, struct rpivid_ctx *ctx)
{
    clk_disable_unprepare(ctx->clk);
    clk_put(ctx->clk);
}
```

Will let the clock framework know we don't need to enable the clock anymore (decreasing the reference counting), and drop our `clk` instance. In the process, we drop any range we had so far, removing our constraints on that particular user.

jc-kynesim:
Thanks - that is very helpful.  That just leaves the elephant in the room: RATE.  Do I just binary chop using clk_set_min_rate to find the biggest number I can successfully set it to or is there some other better call?

Or indeed can I just set it to ULONG_MAX and be done with it?

mripard:
As far as I understood, we have several viable strategies here:
* Either we use something like the minimize flag in the driver, that will always try to set the clock rate at the maximum allowed. Then you wouldn't need the clk_set_min_rate call at all, as long as the clock is enabled, it will be at its maximum rate;
* If minimize isn't set on that clock, you can keep doing the clk_round_rate() dance you mentioned earlier, the framework will clamp the rate at the boundaries reported by the firmware and the current users
* If minimize is set on that clock, then clk_round_rate will always return the minimum, so the above doesn't work anymore. In such a case you can use something like https://github.com/raspberrypi/linux/pull/4940/commits/df368502ecbe1de26cf02a9b7837da9e967d64ca

I can't really judge if maximize, minimize or nothing on that clock would be best here though. @popcornmix was hinting that we'd like to minimize all of them, so maybe it would the last one is the one you should go for.

jc-kynesim:
AFAIK strategy (1) would be fine. As long as the clock rate returns to minimum whenever I'm not using it I think he should be happy.  I suspect the critical bit I/we weren't getting was the distinction between clk_hw & clk and where & how they should be claimed.

mripard:
> AFAIK strategy (1) would be fine. As long as the clock rate returns to minimum whenever I'm not using it I think he should be happy.

We wouldn't have that with the first strategy though. It would *always* run at the maximum allowed. With the earlier code excerpt, it wouldn't necessarily be bad because you would disable the clock whenever it's no longer used so the rate wouldn't be relevant. But if there was another driver that needed it at some point, then we would be running at the maximum again, disregarding whether the codec is doing something or not.

jc-kynesim:
OK I've got a linux built off your branch with start/stop logic built as above (and enough logging to be sure its being called as expected).  I don't have .minimise set so we are in strategy (2) as I understand it.  At the start of time debug/clk_summary reports
`fw-clk-hevc                          0        0        0   250000000          0     0  50000         Y`
I run a video decode, start clock and stop clock get called exactly once each and we have
`fw-clk-hevc                          0        0        0   550000000          0     0  50000         Y`
And `vcgencmd measure_clock hevc` confirms that hevc is running @ 550Mhz

I thought that it should now be disabled or slow?

You can see what I did on https://github.com/jc-kynesim/linux/tree/dev/5.15/rpivid_clk_count_2

jc-kynesim:
Strategy (3) does seem to almost do the trick:  https://github.com/jc-kynesim/linux/tree/dev/5.15/rpivid_clk_count_3
However if I have two instances running then the second is unable to find the max clock as clk_set_min_rate always seems to return non-zero (maybe it would be zero if I picked the exact right rate?)

I could determine max_rate at start of time and always use that I guess

mripard:
> OK I've got a linux built off your branch with start/stop logic built as above (and enough logging to be sure its being called as expected). I don't have .minimise set so we are in strategy (2) as I understand it. At the start of time debug/clk_summary reports `fw-clk-hevc 0 0 0 250000000 0 0 50000 Y` I run a video decode, start clock and stop clock get called exactly once each and we have `fw-clk-hevc 0 0 0 550000000 0 0 50000 Y` And `vcgencmd measure_clock hevc` confirms that hevc is running @ 550Mhz
> 
> I thought that it should now be disabled or slow?

No, it seems to work as expected. In that strategy the clock driver won't enforce anything aside from making sure we end up with a rate within the boundaries expressed by the clock users and the firmware. At that point, you don't have any users anymore, and the rate (550MHz) is still within the boundaries, so the driver will be happy with that. The clock is disabled though, so it shouldn't have any effect on the power consumption.

> Strategy (3) does seem to almost do the trick: https://github.com/jc-kynesim/linux/tree/dev/5.15/rpivid_clk_count_3 However if I have two instances running then the second is unable to find the max clock as clk_set_min_rate always seems to return non-zero (maybe it would be zero if I picked the exact right rate?)
> 
> I could determine max_rate at start of time and always use that I guess

It's probably because as soon as the first is done, the range becomes fairly narrow (basically max_rate only), so if you set a range that doesn't include it, it will reject it.

I was a bit reluctant to add a function that would return the boundaries for a given clock, because it's been rejected before, but since we try to come up with some kind of hacks to work around a drivers behaviour, this is already an abstraction violation. I'll add a function and try to make our case, again :)

Expect a PR by tomorrow :)

popcornmix:
> The clock is disabled though, so it shouldn't have any effect on the power consumption.

Is the fact the clock is disabled indicated to firmware? I'm not sure it is, and the clock is left running at 550MHz.

jc-kynesim:
> > Strategy (3) does seem to almost do the trick: https://github.com/jc-kynesim/linux/tree/dev/5.15/rpivid_clk_count_3 However if I have two instances running then the second is unable to find the max clock as clk_set_min_rate always seems to return non-zero (maybe it would be zero if I picked the exact right rate?)
> > I could determine max_rate at start of time and always use that I guess
> 
> It's probably because as soon as the first is done, the range becomes fairly narrow (basically max_rate only), so if you set a range that doesn't include it, it will reject it.

Yeah I thought that might be the case, but surely that is wrong. I'm setting a min_clock for a new clk object and that surely should be independent of the min_clock I've set on any other clk object?  The case here is muddied by it being a single driver but if I had two independent drivers that used the same clock and the 1st one set a minimum of 250MHz then the 2nd should be able to ask for 200MHz if that's all it needs.
 
> I was a bit reluctant to add a function that would return the boundaries for a given clock, because it's been rejected before, but since we try to come up with some kind of hacks to work around a drivers behaviour, this is already an abstraction violation. I'll add a function and try to make our case, again :)
> 
> Expect a PR by tomorrow :)

Good Oh! Thanks.

jc-kynesim:
Could we "simply" have a special value (say ULONG_MAX) that can be requested that means "the biggest number that can be supported that isn't limited by some other maximum"?

mripard:
> > The clock is disabled though, so it shouldn't have any effect on the power consumption.
> 
> Is the fact the clock is disabled indicated to firmware? I'm not sure it is, and the clock is left running at 550MHz.

[We don't.](https://github.com/raspberrypi/linux/blob/rpi-5.15.y/drivers/clk/bcm/clk-raspberrypi.c#L179)

I thought we did, sorry.

mripard:
I've pushed a new version with accessors for the clock boundaries, and a patch for rpivid to convert to that new accessor.

drivers/gpu/drm/vc4/vc4_kms.c
```diff
@@ -354,6 +354,8 @@ static void vc4_atomic_commit_tail(struct drm_atomic_state *state)
 	struct vc4_hvs_state *new_hvs_state;
 	struct drm_crtc *crtc;
 	struct vc4_hvs_state *old_hvs_state;
+	unsigned long max_clock_rate =
+		hvs->vc5_hdmi_enable_scrambling ? 550000000 : 500000000;
```
popcornmix: Can we use `clk_get_max_rate(hvs->core_clk)`
That would also work in the case the user had a custom setting (e.g. `core_freq=600` in config.txt)

jc-kynesim:
The new version seems to do what I want if I add a .minimize to hevc in clk-raspberrypi.c

popcornmix:
I'm testing with the change to use `clk_get_max_rate(hvs->core_clk)` and all fw clocks setting `.minimize` and this seems to be working as desired.
A range of dual monitor resolution (including 4kp60+4kp30) is okay.

However I've noticed v3d is no longer being clocked up and down with use.
e.g. before the PR when running neverball:
```
 fw-clk-v3d                           0        0        0   550000000          0     0  50000         Y
```
and after
```
fw-clk-v3d                           0        0        0   250000000          0     0  50000         Y
```
I've fixed it with top commit from:
https://github.com/popcornmix/linux/commits/5.15-clk-improvements
although I'm not quite sure why calling clk_set_rate wasn't working.

mripard:
> I've fixed it with top commit from: https://github.com/popcornmix/linux/commits/5.15-clk-improvements although I'm not quite sure why calling clk_set_rate wasn't working.

It was. Under the hood, clk_set_rate() calls clk_round_rate() and will then set whatever rate it returned. By setting minimize on that clock, clk_round_rate() will always return the minimum allowed. So clk_set_rate() will always set the rate to the minimum and basically ignore the rate.

popcornmix:
Okay, then using `clk_set_min_rate` is correct for all firmware clocks.

mripard:
Indeed, which is also why I didn't want to minimize all clocks in that PR since I didn't know how to make sure we didn't have any regression on any device.

Do you want me to pick up the patches you did and push them into this PR, or will you send it?

jc-kynesim:
So barring actually putting all the edits together do we think that we are good now?  It would make my life happy if this could be in rpi-update by Friday evening.

pelwell:
My mouse is (metaphorically) hovering over the "Rebase and Merge" button.

popcornmix:
Okay, merge this, and I'll submit a follow on PR.

pelwell:
It's a long list of commits, with conflicts.

popcornmix:
@mripard can you rebase?
There have been upstream commits (by you) to drivers/clk/clk.c since this PR was made.

mripard:
I just pushed a rebased version

drivers/gpu/drm/vc4/vc4_kms.c
mripard: Good catch, I just addressed it.

pelwell:
```
drivers/staging/media/rpivid/rpivid_video.c: In function ‘rpivid_stop_streaming’:
drivers/staging/media/rpivid/rpivid_video.c:623:6: warning: unused variable ‘ret’ [-Wunused-variable]
  623 |  int ret;
      |      ^~~
drivers/staging/media/rpivid/rpivid_video.c:622:7: warning: unused variable ‘min_hevc_clock’ [-Wunused-variable]
  622 |  long min_hevc_clock = clk_round_rate(dev->clock, 0);
      |       ^~~~~~~~~~~~~~
```

pelwell:
The warnings above are introduced by eba927c2800ba3e21c77f940b3f9c463a47244ea, the commit message for which says:
```
    Revert "rpivid: Switch to new clock api"
    
    This reverts commit ec7556e20c2c29c3df9493248a1a4d60ed20ae38.
```
However, it is far from the inverse of ec7556e20c2c29c3df9493248a1a4d60ed20ae38, and looks generally suspect.

jc-kynesim:
Yeah - the bulk of that commit https://github.com/raspberrypi/linux/commit/eba927c2800ba3e21c77f940b3f9c463a47244ea is sound enough, in that it removes stuff that is about to vanish but the 2 extra lines in rpivid_stop_streaming do seem utterly spurious.

mripard:
Yeah, I wasn't entirely sure how to revert it, but it's revert of both ec7556e, while taking  50d2e7e69d28b02a87770c4c767bf111fa2a775a into account. While we needed to remove the old clock API, the new layout of the code seemed sound to me so I wanted to keep it that way.

The two lines in rpivid_stop_streaming are indeed spurious, I'll remove them.

pelwell:
Thanks - that will do.

popcornmix:
I'm getting a hang on boot following this PR on Pi3+ with hdmi disconnected (i.e. composite output).
Okay with this PR reverted.

mripard:
I have the same issue, I'm looking into it.

mripard:
I found what's going on, but I'm not sure how to address or workaround the issue.

The offending commit is 2a52a4b2395a30d62697207b5efb678bc4d17272. The HSM clock seems to be uninitialized by the firmware when no monitor is plugged in, which was pointed out in #4457.

So far, we had a call to clk_set_rate introduced by #4547, later changed to clk_set_min_rate upstream (2d717d82267550260e4a31e6980d42585dc2f745).

On RaspberryPi3, the HSM clock is handled by the clk-bcm2835 driver:
https://github.com/raspberrypi/linux/blob/rpi-5.15.y/arch/arm/boot/dts/bcm2835-common.dtsi#L132

For the HSM clock, the driver uses the REGISTER_PER_CLK macro
https://github.com/raspberrypi/linux/blob/rpi-5.15.y/drivers/clk/bcm/clk-bcm2835.c#L2158

Which then set the parents to the bcm2835_clock_per_parents list
https://github.com/raspberrypi/linux/blob/rpi-5.15.y/drivers/clk/bcm/clk-bcm2835.c#L1604

When the monitor isn't attached, it looks like the parent is the "gnd" clock, which is never registered. The HSM clock is thus always an orphan, and the clk_set_min_rate call skips the rate change due to the commit above. Since the rate is never changed, the rate is still at 0, and we have a lockup.

I think we could just drop the commit 2a52a4b2395a30d62697207b5efb678bc4d17272. Whether or not to skip the rate change could be argued either way, and it's not critical anyway.

However, the fact that the GND clock is a (default?) parent for most peripheral clocks but is never registered seems like a bug to me. I tried to google for some information about what that clock could be but came short, so I'm not quite sure how to address that.

jc-kynesim:
A naive guess would be that GND is just "ground" i.e. a null 0 Hz clk

popcornmix:
Yes, GND is just a null clock.

> On RaspberryPi3, the HSM clock is handled by the clk-bcm2835 driver:

There's no reason for this. I think moving this to fw driver makes sense (especially if it avoids this issue).

mripard:
It works fine if we use the firmware clock, I've pushed a PR to address this.

alanbork:
Does this mean that 4kp60 is supposed to work out of the box now? Because as of 1c56b3a58974f725e1ac63214eeaf914c5908302, I still get hangs/blank screens if I try to use 4096x2160p60 unless I overclock the cpu via config.txt (core_freq=600  core_freq_min=600)


dmesg reports:

vc4-drm gpu: [drm] *ERROR* Timed out waiting for commit
[  273.107442] [drm:drm_atomic_helper_wait_for_flip_done [drm_kms_helper]] *ERROR* [CRTC:94:crtc-3] flip_done timed out

mripard:
It was supposed to work out of the box before this PR, and is still supposed to work out of the box after.

Could you file a new issue?

PR #4941, opened 2022-03-16
Add VCM support to OV5647 and IMX219 overlays

6by9:
Adds a VCM lens driver and configuration.

Works with Arducam [OV5647](https://www.amazon.co.uk/Arducam-Camera-Autofocus-Raspberry-Motorized/dp/B07SN8GYGD) and [IMX219](https://www.arducam.com/product/arducam-imx219-auto-focus-camera-module-drop-in-replacement-for-raspberry-pi-v2-and-nvidia-jetson-nano-camera/) modules. Load via ```dtoverlay=imx219,vcm```.

Upstream aren't going to like the regulator hook in the VCM driver, but without it you must open the lens driver to power it on. With it, starting the camera will also power up and initialise the VCM.


pelwell:
Why does the AD5398 driver need a hook for the regulator? The AD5820 driver doesn't use one. A brief explanation would be helpful.

By the way, the OV5647's new `vcm` parameter also needs a note in the `README`.

6by9:
> Why does the AD5398 driver need a hook for the regulator? The AD5820 driver doesn't use one. A brief explanation would be helpful.

I'll update the commit text, but it comes down to making the sensor usable without jumping through masses of hoops.

The VCM sensor driver implements a V4L2 control for focus_absolute, therefore the position it has been requested to be at should be persistent. The framework provides no hooks to power up the VCM with the sensor, therefore a simple client that only worries about the sensor and Unicam devices will get out of focus images due to the VCM being powered down.
Add the regulator hook so that the VCM knows when the sensor is powered up, and the focus position will be restored when the sensor is powered up.

Discussion on linux-media: https://www.spinics.net/lists/linux-media/msg201312.html / https://www.spinics.net/lists/linux-media/msg202072.html where they want to pass the responsibility to userspace.

Seeing as they view the assumption of a shared regulator as being invalid, perhaps have 2 regulator hooks, one to control the VCM power, and the other to register with so that power is restored with the sensor. There's no need to implement that for our platforms though.

> By the way, the OV5647's new vcm parameter also needs a note in the README.

ARRRGGHHH!! Can we get your dtoverlay checking script sorted as a post-commit hook, or automated within Github?

6by9:
Description added to the start of the AD5398 driver, and README updated.

pelwell:
The checker is a Perl script that uses `ovmerge` (Perl) and `dtmerge` (compiled). A run currently takes 20s on our build server, increasing as we add platforms. There are also situations where it fails on legitimate overlays, requiring a new exclusion to be added, so it isn't a straightforward enable-and-forget check.

juergh:
Huh? There's already a driver named ad5398: drivers/regulator/ad5398.c

juergh:
```
error: the following would cause module name conflict:
  drivers/regulator/ad5398.ko
  drivers/media/i2c/ad5398.ko
make[2]: *** [/build/jammy/Makefile:1794: modules_check] Error 1
```

6by9:
Sorting a patch to rename it.

AD's ad5398 appears to expose it through the regulator framework, which is near useless for use as a lens driver.

noahwilliamsson:
I'm very happy to see that these modules with voice coil motors are getting official support in the Raspberry Pi kernel.  Similarly, I'm also happy to see support for VCMs being discussed on the libcamera mailing list (Mar, 2022).  I've been toying with them on a CNC for rough alignment purposes and for general video streaming (and also hoping that native MJPEG support will be introduced again, as requested by the PiKVM author I think).

JFYI, and I'm not sure this is a very useful comment, but a quick look suggests that `drivers/media/i2c/ad5398_vcm.c` is quite similar to the already upstreamed `drivers/media/i2c/dw9714.c` driver minus the required support for the regulator. Perhaps the two could be merged?

More context: last year, I bought these Arducam [IMX219](https://www.arducam.com/product/arducam-imx219-auto-focus-camera-module-drop-in-replacement-for-raspberry-pi-v2-and-nvidia-jetson-nano-camera/) and [OV5647](https://www.arducam.com/product/b0033c-arducam-5mp-1080p-camera-module-with-case-for-raspberry-pi-3-3-b-and-more/) modules and asked their support what VCM chip they were using.  They claimed it was DW9714, for which I found an already existing upstream driver: `drivers/media/i2c/dw9714.c` (links: [5.10.y](https://github.com/raspberrypi/linux/blob/rpi-5.10.y/drivers/media/i2c/dw9714.c), [5.15.y](https://github.com/raspberrypi/linux/blob/rpi-5.15.y/drivers/media/i2c/dw9714.c)).

I got the VCM working under V4L2 with pretty much the same DT changes as in this MR (but with: `compatible = "dongwoon,dw9714";`) and after enabling `CONFIG_VIDEO_DW9714` (not in Raspberry Pi's defconfig) and adding support for a regulator in the DW9714 driver: [diff against 5.10.y](https://gist.github.com/noahwilliamsson/32a9e985227693ffe856cd109c8359ce) (sorry, first time exploring regulators).

6by9:
Please note that this is not official support for those Arducam products.
Supporting some form of AF is on the radar for libcamera and the Pi IPA, so this is a basic level of support for some modules that happen to be on the market. I also have a couple of other motorised lenses on my desk that it would be nice to support.

It looks like there are a number of manufacturers who have very similar looking 10bit I2C VCM modules
[Analog Devices AD5398](https://www.analog.com/en/products/ad5398.html)
[Fitipower FP5510E](http://www.jifangsz.com/FP5510E-Preliminary%200.2-JAN-2018.pdf)
[Dongwoon DW9714](https://www.arducam.com/downloads/DW9714A-DONGWOON(Autofocus_motor_manual).pdf)

Bit 15 for all of them is powerdown. Position is in bits 4-13. The remaining bits vary slightly. They could be combined into 1 driver using different compatibles to allow differentiation.

Seeing as I have a significant bias to using the regulator callback hook to power the VCM up with the sensor, and mainline have rejected that as an option, keeping it as a separate driver generally reduces the maintenance burden as it avoids merge conflicts.

PR #4945, opened 2022-03-19
CM3 cam1_reg and cam1_reg_gpio fix

surfdude75:
property name is gpio and not gpios


pelwell:
Thanks - the same problem exists for the original Compute Module as well, also now fixed.

PR #4952, opened 2022-03-22
drivers: staging: bcm2835-isp: Clear LS table handle in the firmware

naushir:
When all nodes have stopped streaming, ensure the firmware has released its
handle on the LS table dmabuf. This is done by passing a null handle in the
LS params.


naushir:
@davidplowman @6by9

davidplowman:
I tried it and it seemed to do what it said on the tin.

6by9:
Looks reasonable to me.

As discussed, it does have the issue that the (persist and cross-process) control will still list the old (process specific) fd for the lens shading table, but such is life. I'd need to look and see if any other controls take in dmabufs or other fds, and how they then handle the "get" function from other processes.

naushir:
@pelwell can you merge if you are happy with this change please?

PR #4954, opened 2022-03-25
backport: drm/tiny: Add MIPI DBI compatible SPI driver

notro:
This is a backport of the panel-mipi-dbi driver that just [entered](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/drivers/gpu/drm/tiny/panel-mipi-dbi.c) Linus' tree and will be available in v5.18.
It's a generic display driver that can be used with all MIPI DBI SPI displays.

Wiki: https://github.com/notro/panel-mipi-dbi/wiki

Thanks to @mripard for the idea of using firmware files for the initialisation sequence making it possible to do a generic driver in mainline.


pelwell:
Very nice!

PR #4955, opened 2022-03-25
AudioInjector.net pi soundcard: silence EPROBE_DEFER

flatmax:
Don't print errors when driver probing defers. Remove some spurious whitespaces.

### DIFF
--- a/sound/soc/bcm/audioinjector-pi-soundcard.c
+++ b/sound/soc/bcm/audioinjector-pi-soundcard.c
@@ -138,7 +138,7 @@ static int audioinjector_pi_soundcard_probe(struct platform_device *pdev)
 {
 	struct snd_soc_card *card = &snd_soc_audioinjector;
 	int ret;
-	
+
 	card->dev = &pdev->dev;
 
 	if (pdev->dev.of_node) {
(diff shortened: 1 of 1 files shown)


sound/soc/bcm/audioinjector-pi-soundcard.c
```diff
@@ -158,9 +158,10 @@ static int audioinjector_pi_soundcard_probe(struct platform_device *pdev)
 			}
 	}
 
-	if ((ret = devm_snd_soc_register_card(&pdev->dev, card))) {
-		dev_err(&pdev->dev, "snd_soc_register_card failed (%d)\n", ret);
-	}
+	if (ret = devm_snd_soc_register_card(&pdev->dev, card))
+		if (ret != -EPROBE_DEFER)
```
6by9, line 162: Thanks for the patch.

The preference seems to be for dev_err_probe now, as it also stores the probe reason.
https://elixir.bootlin.com/linux/latest/source/drivers/base/core.c#L4709

sound/soc/bcm/audioinjector-pi-soundcard.c
```diff
@@ -158,9 +158,10 @@ static int audioinjector_pi_soundcard_probe(struct platform_device *pdev)
...
+			dev_err(&pdev->dev, "snd_soc_register_card failed (%d)\n", ret);
```
HiassofT, line 163: hint: just use dev_err_probe instead
https://github.com/raspberrypi/linux/blob/rpi-5.15.y/drivers/base/core.c#L4666-L4692

flatmax:
Much appreciated info. I will make the change.

PR #4959, opened 2022-03-27
overlays/rpi-display: Add support for DRM driver

notro:
I was inspired by how @pelwell made the [pitft35-resistive](https://github.com/raspberrypi/linux/blob/rpi-5.15.y/arch/arm/boot/dts/overlays/pitft35-resistive-overlay.dts) overlay support both the fbtft and DRM drivers. I mostly use the rpi-display when I test the DRM MIPI DBI code in the kernel and have made an overlay for that. Now with this `drm` parameter I can ditch that and use the stock overlay instead.

I had to revert the backlight workaround since fbtft backlight is now [fixed](https://github.com/notro/fbtft/issues/571#issuecomment-983444543) resulting in inverted backlight.
I don't know if rpi-5.10.y will get any more releases, but if it does the revert patch is needed there as well since the fix was backported in 5.10.83 and rpi-5.10.y is at 5.10.95.
I haven't tested it on 5.10.

Cc: @awatterott

### DIFF
--- a/arch/arm/boot/dts/overlays/README
+++ b/arch/arm/boot/dts/overlays/README
@@ -2913,14 +2913,36 @@ Params: <None>
 
 Name:   rpi-display
 Info:   RPi-Display - 2.8" Touch Display by Watterott
+        Linux has 2 drivers that support this display and this overlay supports
+        both.
+
+        Examples:
+          fbtft/fb_ili9341: dtoverlay=rpi-display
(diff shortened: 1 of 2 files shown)


pelwell:
5.10 is in maintenance mode now - we're only taking bug fixes and security patches.

pelwell:
This is now in the latest rpi-update kernel.

PR #4960, opened 2022-03-28
Bugfix/scatternet investigation

david-halls:


pelwell:
This appears to have been pushed in error.

PR #4964, opened 2022-03-29
mm,page_alloc,cma: introduce a customisable threshold for allocating …

davidplowman:
…pages in cma

On some platforms the cma area can be half the entire system memory,
meaning that allocations start happening in the cma area immediately.
This leads to fragmentation and subsequent fatal cma_alloc failures.

We introduce an "alloc_in_cma_threshold" parameter which requires that
this many sixteenths of the free pages must be in cma before it will
try to use them. By default this is set to 12, but the previous
behaviour can be restored by setting it to 8 on startup.

### DIFF
--- a/mm/page_alloc.c
+++ b/mm/page_alloc.c
@@ -190,6 +190,27 @@ EXPORT_SYMBOL(init_on_alloc);
 DEFINE_STATIC_KEY_MAYBE(CONFIG_INIT_ON_FREE_DEFAULT_ON, init_on_free);
 EXPORT_SYMBOL(init_on_free);
 
+#define ALLOC_IN_CMA_THRESHOLD_MAX 16
+#define ALLOC_IN_CMA_THRESHOLD_DEFAULT 12
+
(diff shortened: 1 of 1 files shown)


davidplowman:
Please correct me if I've got anything wrong, or if any alternative nomenclature might be better. When comparing thresholds I've done the multiplication first, figuring that that's still an awful lot of memory before it could overflow, or would we rather do it the other way round?

davidplowman:
@6by9 @popcornmix

6by9:
The logic looks reasonable on that.

pelwell:
Given that you are working in 16ths and we don't care about losing precision at that scale, wouldn't dividing first be better?

davidplowman:
I guess I was wondering what was more unlikely, the number of free pages being less than 16 or the whole thing overflowing. But I suppose it still works like that so yes, maybe that's preferable - I'll swap it round.

popcornmix:
No objections here.

PR #4965, opened 2022-03-30
ARM: dts: Enable PMU on Cortex-A72 in AArch32 state

bavison:
There is no specific AArch32 driver for the Cortex-A72 PMU, but the
Cortex-A7 one works and is much better than no PMU driver at all.

Signed-off-by: Ben Avison <bavison@riscosopen.org>

### DIFF
--- a/arch/arm/boot/dts/bcm2711.dtsi
+++ b/arch/arm/boot/dts/bcm2711.dtsi
@@ -445,7 +445,7 @@
 	};
 
 	arm-pmu {
-		compatible = "arm,cortex-a72-pmu", "arm,armv8-pmuv3";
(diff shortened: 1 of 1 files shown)


pelwell:
Works better how? What difference can be observed?

bavison:
Without a PMU driver, the perf framework can't report cycles, instructions, branch misses etc. This can be seen for example with `perf stat ls` which at present works on Pi 4 in AArch64 state (because `arm,cortex-a72-pmu` matches a driver in the `arch/arm64` directory) but not Pi 4 in AArch32 state.

There's precedent here in `bcm2710.dtsi`, which means that while Pi 3 in AArch64 state uses the Cortex-A53 PMU driver, Pi 3 in AArch32 state already also uses the Cortex-A7 PMU driver.

pelwell:
Thanks,

PR #4970, opened 2022-03-31
drm/vc4: Force trigger of dlist update on margins change

6by9:
When the margins are changed, the dlist needs to be regenerated
with the changed updated dest regions for each of the planes.

Setting the zpos_changed flag is sufficient to trigger that
without doing a full modeset, therefore set it should the
margins be changed.

Requested by @spl237

### DIFF
--- a/drivers/gpu/drm/vc4/vc4_crtc.c
+++ b/drivers/gpu/drm/vc4/vc4_crtc.c
@@ -728,12 +728,10 @@ static int vc4_crtc_atomic_check(struct drm_crtc *crtc,
 		if (conn_state->crtc != crtc)
 			continue;
 
-		if (memcmp(&conn_state->tv.margins, &vc4_state->margins,
+		if (memcmp(&vc4_state->margins, &conn_state->tv.margins,
 			   sizeof(vc4_state->margins))) {
-			vc4_state->margins.left = conn_state->tv.margins.left;
-			vc4_state->margins.right = conn_state->tv.margins.right;
-			vc4_state->margins.top = conn_state->tv.margins.top;
-			vc4_state->margins.bottom = conn_state->tv.margins.bottom;
+			memcpy(&vc4_state->margins, &conn_state->tv.margins,
+			       sizeof(vc4_state->margins));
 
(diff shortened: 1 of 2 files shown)


popcornmix:
LGTM

drivers/gpu/drm/vc4/vc4_crtc.c
```diff
@@ -728,10 +728,18 @@ static int vc4_crtc_atomic_check(struct drm_crtc *crtc,
 		if (conn_state->crtc != crtc)
 			continue;
 
-		vc4_state->margins.left = conn_state->tv.margins.left;
-		vc4_state->margins.right = conn_state->tv.margins.right;
-		vc4_state->margins.top = conn_state->tv.margins.top;
-		vc4_state->margins.bottom = conn_state->tv.margins.bottom;
+		if (memcmp(&conn_state->tv.margins, &vc4_state->margins,
+			   sizeof(vc4_state->margins))) {
+			vc4_state->margins.left = conn_state->tv.margins.left;
```
pelwell: Why don't you use `memcpy` for the update, since `memcmp` is used for comparison? If the margins structure were to include an extra field then you would potentially detect the need for an update but not actually change anything.

pelwell: And if you do, swap the `memcmp` parameters so they match the `memcpy`.

pelwell:
Something like this?

drivers/gpu/drm/vc4/vc4_crtc.c
6by9: Yup, happy to do that.
I only kept the old code to reduce the diff, but memcpy makes more sense.

6by9:
Observation about Github PR handling.
I'm guessing you did a squash merge in accepting the PR. That appears to have reset the author of the patch to ```6by9 <6by9@users.noreply.github.com>```. Not a big deal, but makes things more awkward if upstreaming.

pelwell:
Strange that I've not seen that happen before - perhaps something about your privacy settings?

I'll fix the patch for 5.16 and 5.17.

PR #4971, opened 2022-03-31
Cursor Corruption Fix, Upstream Edition

mripard:
Hi,

Here's a PR with the fix suggested by upstream for the frame corruption we were seeing along the edges of display when the overscan margins were enabled. This was fixed by PR #4895.

The main issue was that we were ending up skipping the wait for vblank due to a hack meant to handle the legacy cursor APIs. The initial fix implemented by the PR was to work around that hack, but upstream suggested to remove it entirely instead.

This PR thus reverts the workaround and removes the legacy cursor hack.


popcornmix:
This seems to work as well as #4895 (and is simpler which is good).
There is still the issue mentioned there: https://github.com/raspberrypi/linux/pull/4895#issuecomment-1044592212,
but that seems unrelated to original "corrupt cursor" issue.

PR #4972, opened 2022-03-31
vc4_hdmi: Remove VC4_HDMI_SCHEDULER_CONTROL_VERT_ALWAYS_KEEPOUT

popcornmix:
This bit ensures data island packets are never generated when disallowed by HDCP.
As no Pi boards support HDCP this is providing an unnecessary restriction


popcornmix:
@timg236 I believe on firmware side you removed this bit for 2711 (but it has the same meaning on pi0-4).

timg236:
Yes, it looks fine. No Pi boards support HDCP.

popcornmix:
@pelwell any objections?

PR #4979, opened 2022-04-05
Add support for IMX258

6by9:
Updates the mainline driver so that it can work on the Pi over 2 lanes (or 4), and with a 24MHz clock.
It also adds all the features that libcamera needs from a sensor driver.


arch/arm/boot/dts/overlays/imx258-overlay.dts
```diff
@@ -0,0 +1,148 @@
...
+		rotation = <&imx258>,"rotation:0";
+		orientation = <&imx258>,"orientation:0";
+		media-controller = <&csi>,"brcm,media-controller?";
+		cam0 = <&i2c_frag>, "target:0=",<&i2c_vc>,
+		       <&csi_frag>, "target:0=",<&csi0>,
+		       <&clk_frag>, "target:0=",<&cam0_clk>,
+		       <&reg_frag>, "target:0=",<&cam0_reg>,
+		       <&imx258>, "clocks:0=",<&cam0_clk>,
+		       <&imx258>, "vdda-supply:0=",<&cam0_reg>;
+		vcm = <&vcm>, "status=okay",
```
pelwell, line 144: `vcm` and `4lane` need documenting.

arch/arm/boot/dts/overlays/imx258-overlay.dts
```diff
@@ -0,0 +1,148 @@
...
+	__overrides__ {
+		rotation = <&imx258>,"rotation:0";
+		orientation = <&imx258>,"orientation:0";
+		media-controller = <&csi>,"brcm,media-controller?";
+		cam0 = <&i2c_frag>, "target:0=",<&i2c_vc>,
+		       <&csi_frag>, "target:0=",<&csi0>,
+		       <&clk_frag>, "target:0=",<&cam0_clk>,
+		       <&reg_frag>, "target:0=",<&cam0_reg>,
+		       <&imx258>, "clocks:0=",<&cam0_clk>,
+		       <&imx258>, "vdda-supply:0=",<&cam0_reg>;
```
pelwell: `vana-supply` would look more logical here - which is correct?

drivers/media/i2c/imx258.c
```diff
@@ -113,8 +118,8 @@ struct imx258_link_freq_config {
 	u64 link_frequency;
 	u32 pixels_per_line;
 
-	/* PLL registers for this link frequency */
-	struct imx258_reg_list reg_list[IMX258_LANE_CONFIGS];
+	/* Configuration for this link frequency / num lanes selection*/
```
pelwell: There's a space missing at the end of the comment.

drivers/media/i2c/imx258.c
```diff
@@ -1020,6 +1026,26 @@ static void imx258_adjust_exposure_range(struct imx258 *imx258)
 				 exposure_def);
 }
 
+static int imx258_set_frame_length(struct imx258 *imx258, unsigned int val)
+{
+	int ret = 0;
```
pelwell: The initialisation looks a bit pointless.

drivers/media/i2c/imx258.c
```diff
@@ -28,7 +28,7 @@
 #define IMX258_VTS_30FPS		0x0c50
 #define IMX258_VTS_30FPS_2K		0x0638
 #define IMX258_VTS_30FPS_VGA		0x034c
-#define IMX258_VTS_MAX			0xffff
+#define IMX258_VTS_MAX			65525
```
pelwell, line 33: In an earlier commit you have:
```diff
+#define IMX258_EXPOSURE_OFFSET		10
...
-#define IMX258_EXPOSURE_MAX		65535
+#define IMX258_EXPOSURE_MAX		(IMX258_VTS_MAX - IMX258_EXPOSURE_OFFSET)
```
giving an IMX258_EXPOSURE_MAX of 65525. This commit changes it to 65515 - is this correct?

drivers/media/i2c/imx258.c
```diff
@@ -700,16 +720,29 @@ static int imx258_write_regs(struct imx258 *imx258,
...
 }
 
+/* Get bayer order based on flip setting. */
+static u32 imx258_get_format_code(struct imx258 *imx258)
+{
+	unsigned int i;
+
+	lockdep_assert_held(&imx258->mutex);
+
+	i = (imx258->vflip->val ? 2 : 0) |
```
pelwell, line 977: `REG_CONFIG_MIRROR_VFLIP` (and `_HFLIP` below)?

drivers/media/i2c/imx258.c
```diff
@@ -110,7 +110,7 @@ struct imx258_mode {
 	struct imx258_reg_list reg_list;
```
pelwell, line 139: There's a space missing from `(Y_ADD_END – Y_ADD_STA+ 1)` in the commit message.

drivers/media/i2c/imx258.c
```diff
@@ -28,12 +28,6 @@
 #define IMX258_VTS_30FPS_VGA		0x034c
```
pelwell, line 32: `ununsed` in the commit message.

drivers/media/i2c/imx258.c
6by9: Yes. The original driver was wrong twice over.

The datasheet lists the range of 0x340/1 FRM_LENGTH_LINES to be <=65525. (FWIW table 4-4)

Section 9.3 then states
> The setting range of exposure time during long exposure mode is as follows:
> ```1 <= COARSE_INTEG_TIME <= FRM_LENGTH_LINES - 10```

Also documented under register 0x0350 FRM_LENGTH_CTL as the alpha parameter in a calculation, "type specific adjustment parameter, and is 10(d) for this type"

arch/arm/boot/dts/overlays/imx258-overlay.dts
6by9: You're right. Fixed.

drivers/media/i2c/imx258.c
6by9: Actually no. The mapping is coincidence.

In this case we're wanting the index into ```codes``` array where the index is defined as 
```
/*
 * The supported formats.
 * This table MUST contain 4 entries per format, to cover the various flip
 * combinations in the order
 * - no flip
 * - h flip
 * - v flip
 * - h&v flips
 */
```

```REG_CONFIG_MIRROR_VFLIP``` and _HFLIP are used in imx258_set_ctrl when setting the register in the sensor.

6by9:
Branch updated to hopefully address all your review comments (although you're meant to be on holiday, not reviewing PRs!)

pelwell:
This patch set gives a compilation error on the new rpi-5.18.y branch - it complains about `V4L2_MBUS_CSI2_CONTINUOUS_CLOCK` being undefined - so it hasn't been ported over yet.

popcornmix:
I fixed other drivers by changing `V4L2_MBUS_CSI2_CONTINUOUS_CLOCK` to `V4L2_MBUS_CSI2_NONCONTINUOUS_CLOCK` with inverted logic.
There was an upstream commit saying `V4L2_MBUS_CSI2_CONTINUOUS_CLOCK` was redundant as it was always the inverse of `V4L2_MBUS_CSI2_NONCONTINUOUS_CLOCK`

6by9:
Thanks Phil/Dom - I was going to fix it up today. I'll sort a PR now.

Yes, https://github.com/torvalds/linux/commit/b9f7caa7753ad185e0dc7afb3ae4bd423d11f5c0 dropped ```V4L2_MBUS_CSI2_CONTINUOUS_CLOCK``` in favour of ```V4L2_MBUS_CSI2_NONCONTINUOUS_CLOCK``` and ```0```.

PR #4981, opened 2022-04-07
5.18.y media fixups

6by9:


popcornmix:
Thanks - builds for me.

PR #4994, opened 2022-04-13
overlays: mipi-dbi-spi: width-mm and height-mm are mandatory

notro:
There was a last minute change to the binding in commit 1ecc0c09f19f
("dt-bindings: display: panel: mipi-dbi-spi: Make width-mm/height-mm mandatory").
This was done so userspace can calculate DPI.

Update the overlay to ensure it will work with future kernels.
The driver changes won't be backported to rpi-5.15.

Signed-off-by: Noralf Trønnes <noralf@tronnes.org>

### DIFF
--- a/arch/arm/boot/dts/overlays/mipi-dbi-spi-overlay.dts
+++ b/arch/arm/boot/dts/overlays/mipi-dbi-spi-overlay.dts
@@ -31,6 +31,9 @@
 				reg = <0>;
 				spi-max-frequency = <32000000>;
 
+				width-mm = <0>;
+				height-mm = <0>;
+
 				timing: panel-timing {
 					hactive = <320>;
 					vactive = <240>;


6by9:
Is 0 actually going to do something sane for width_mm / height_mm? It sounds like a good case for a div by 0.

Then again ```simpledrm_connector_helper_get_modes``` seems to ignore copying the value from drm_display_mode to drm_connector if 0.

notro:
> Is 0 actually going to do something sane for width_mm / height_mm? It sounds like a good case for a div by 0.

I have always assumed that userspace can deal with that, at least mutter does:
https://gitlab.gnome.org/GNOME/mutter/-/blob/main/src/backends/meta-monitor.c#L1712

I also see that [drm_reset_display_info()](https://elixir.bootlin.com/linux/latest/C/ident/drm_reset_display_info) sets them to 0 in [drm_connector_update_edid_property()](https://elixir.bootlin.com/linux/latest/C/ident/drm_connector_update_edid_property) if there's no edid, so it looks like userspace has to support it.

6by9:
Thanks - I was just checking that it had been considered.

PR #5002, opened 2022-04-20
Support for IEEE1588 timestamping for the BCM54210PE PHY using the kernel mii_timestamper interface

lasselj:
Hello,

This PR adds support for the IEEE1588 functionality on the BCM54210PE PHY using the Linux Kernel mii_timestamper interface. The BCM54210PE PHY can be found in the Raspberry PI Compute Module 4 and the work has been undertaken by Timebeat.app on behalf of Raspberry PI with help and support from the nice engineers at Broadcom.

Design:
After the ethernet frames are identified using the ptp_classify_raw functions they are parsed through the mii_timestamper interface to the bcm54210pe_rxtstamp and bcm54210pe_txtstamp functions through the skb_defer_rx_timestamp and skb_clone_tx_timestamp kernel functions. In both cases these functions enqueue the sk_buff ptrs and schedules a work structs function to attach timestamps and forward the skbs upstream via the netif_rx_ni and skb_complete_tx_timestamp in the bcm54210pe_run_rx_timestamp_match_thread and bcm54210pe_run_tx_timestamp_match_thread functions in non-interrupt context. The driver uses "smart" poll behaviour triggered by the rx or tx of frames.  
In addition to gettime, settime and adjtime and adjfine, I've implemented the gettimex64 function to provide the best possible sync of the kernel clock. However, as the PHY is separated from the MAC by the MDIO bus, I cannot lock and prevent the scheduler from interrupting the 3-timestamp process, but performance is nonetheless reasonable and the kernel clock sees offset variations in the 1-2us range.

Features:
In addition the driver add support for perout and extts functionality using the standard PTP API and handling of ptp_clock_request structs.

Test & Performance:
We have tested the features in our lab and are able as a client to run at -7 intervals without significant performance impact. We are able to maintain synchronisation of the PHC within +/-10ns of its grandmaster PTP source and the system clock within 1-2us of the PHC.

All the best,

Lasse


popcornmix:
It looks like your source tree is based on the 5.10 kernel.
That is no longer getting new features.
You need to apply your patches to the 5.15 kernel (which is what bullseye currently uses).

Now I think you only have one commit to submit. You should be able to fix this up with something like:
(assuming you have a remote called rpi set to `https://github.com/raspberrypi/linux.git`)
```
git fetch rpi
git reset --hard rpi/rpi-5.15.y
git cherry-pick bf2ec31
# fix up any conflicts
git add -u
git commit
git push -f
```

lasselj:
Ok, I might have a look in due course. If you like you can also wait for the upstreaming process to complete or just cheery pick yourselves.

/Lasse

popcornmix:
This has been merged from upstream commits.

PR #5004, opened 2022-04-24
overlays: Add "drm" parameter to pitft28-resistive

ghost:
Based on https://github.com/raspberrypi/linux/commit/c8d697df20e19d199b2092afeabba5eaa516d977 which did the same for pitft35-resistive.

Tested on fully up to date Raspberry Pi OS Bullseye, with and without desktop.

See also https://forums.raspberrypi.com/viewtopic.php?t=331594


pelwell:
Thanks. Some trailing whitespace crept in (now zapped), and I squashed on merge as these DT commits will all get squashed into one big downstream DT commit for later branches.

PR #5006, opened 2022-04-25
ARM: dts: bcm2835: Switch HSM clock to firmware

mripard:
This is a PR to fix a stall on the Pi3 when it boots without an HDMI display connected.

It was reported in the PR #4940


pelwell:
The first sentence of the commit message is hard to follow, but the patch looks OK.

popcornmix:
I can confirm this fixes the hang on boot on Pi3+ with hdmi unplugged.

pelwell:
Merged with a slightly tweaked commit message.

Issue #5008, opened 2022-04-25
Kernel 5.18-rc4 64bit FKMS boot issue


Dark-Sky:
### Describe the bug

Ever since kernel 5.18-rc3 I am unable to boot in to a Desktop Environment with FKMS. No problem with 5.18-rc2 or 5.15.y or 5.17.y.

With my VIZIO tv I have always had to do a manual config in config.txt with the pi4 for KMS/FKMS or all I get is a black screen. Everything works good with KMS configured to boot but all I get is a black screen with FKMS configured to boot. It is like my manual config in config.txt is ignored.

I hooked up to my Gateway monitor and I can see the lines going by at firs boot but when it tries to go in to graphics mode the screen immediately goes black. I also tried my GW edid.dat file I use every once in a while I put in cmdline.txt but it had no effect.

With both monitors I am unable to log in with ssh. It is like maybe the system is locked up.

### Steps to reproduce the behaviour

Configure config.txt to boot in to FKMS.

### Device (s)

Raspberry Pi 4 Mod. B

### System

`Linux pi4 5.18.0-rc4-1-MANJARO-ARM-RPI #1 SMP PREEMPT Mon Apr 25 19:08:32 UTC 2022 aarch64 GNU/Linux`

```
[ray@pi4 ~]$ vcgencmd version
Mar 24 2022 13:19:26 
Copyright (c) 2012 Broadcom
version e5a963efa66a1974127860b42e913d2374139ff5 (clean) (release) (start)
```

### Logs

Unable to retrieve any logs.

Dark-Sky:
Today I compiled the latest 5.15.25 and 5.17.4 kernels and they also now have this issue trying to use FKMS. I also installed the latest 5.15.25 64bit kernel on my arch-arm 32bit image and it also has this issue; booting with KMS was still good with it also. It was mentioned a few days ago in another issue here that FKMS was depreciated and not a priority should I just have to forget about FKMS going forward?

popcornmix:
fkms is not actively developed and, is not the default configuration and shouldn't be used as a permanent setting.
If a user has issues with kms (compared to fkms), then fixing that is the priority.
There is no guarantee that fkms will be supported indefinitely.

However fkms can be useful in triaging bugs. i.e. if a configuration works with fkms but not kms then it's likely an issue with the kms driver, rather than, say, an X or application issue.

And it can be used as a temporary means of getting a working display while other remedies are employed (e.g. running long enough to update kernel).

[...]

PR #5009, opened 2022-04-27
drm/vc4_kms: Protect hvs dereference on fkms

popcornmix:
fkms has a null hvs pointer so this crashes. Protect against it


popcornmix:
See: https://github.com/raspberrypi/linux/issues/5008

timg236:
LGTM

PR #5021, opened 2022-05-05
drm/vc4_hdmi: Force a modeset when Broadcast RGB setting changes

popcornmix:
Without this the change is not visible until the next modeset

### DIFF
--- a/drivers/gpu/drm/drm_atomic_uapi.c
+++ b/drivers/gpu/drm/drm_atomic_uapi.c
@@ -715,6 +715,7 @@ static int drm_atomic_connector_set_property(struct drm_connector *connector,
 {
 	struct drm_device *dev = connector->dev;
 	struct drm_mode_config *config = &dev->mode_config;
+	bool margins_updated = false;
 	bool replaced = false;
 	int ret;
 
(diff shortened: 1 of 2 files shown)


popcornmix:
See: https://forums.raspberrypi.com/viewtopic.php?t=329860

You can switch when running with:
```
DISPLAY=:0 xrandr --output HDMI-1 --set "Broadcast RGB" "Limited 16:235"
```
```
DISPLAY=:0 xrandr --output HDMI-1 --set "Broadcast RGB" "Full"
```
and you see a modeset and colours shift slightly.

popcornmix:
@6by9 I've also added your margins commit. 
I've tested it and it behaves as desired:
```
DISPLAY=:0 xrandr --output HDMI-1 --set "top margin" 100
```
results in a black border at top of screen (immediately).

6by9:
LGTM, although I thought you'd said that the atomic_check wasn't being called.

popcornmix:
Too many atomic_checks... I was incorrectly checking for in vc4_hdmi_encoder_atomic_check which isn't triggered by the property.

6by9:
> Too many atomic_checks... I was incorrectly checking for in vc4_hdmi_encoder_atomic_check which isn't triggered by the property.

I know that problem all too well.
Indeed the encoder atomic_check won't be called, but the connector one will be.

popcornmix:
@pelwell @6by9 okay to merge?

pelwell:
Yup.

PR #5022, opened 2022-05-06
Fixes for IGT crashes

mripard:
Hi,

Here's a collection of fixes related to the IGT crashes @popcornmix reported. It turns out many of them were actually due to the fact that even though the BCM2711 doesn't have the v3d handled by the vc4 driver anymore, we were still using some of the code backed by the v3d.

This PR properly splits everything away, and adds a warning to all the v3d-related entrypoints so that we notice if that it ever happens again.

### DIFF
--- a/drivers/gpu/drm/vc4/vc4_kms.c
+++ b/drivers/gpu/drm/vc4/vc4_kms.c
@@ -395,7 +395,7 @@ static void vc4_atomic_commit_tail(struct drm_atomic_state *state)
 		old_hvs_state->fifo_state[channel].pending_commit = NULL;
 	}
 
-	if (vc4->hvs && vc4->hvs->hvs5) {
+	if (vc4->is_vc5 && !vc4->firmware_kms) {
 		unsigned long state_rate = max(old_hvs_state->core_clock_rate,
 					       new_hvs_state->core_clock_rate);
 		unsigned long core_rate = clamp_t(unsigned long, state_rate,
@@ -409,7 +409,7 @@ static void vc4_atomic_commit_tail(struct drm_atomic_state *state)
 	vc4_ctm_commit(vc4, state);
 
 	if (!vc4->firmware_kms) {
-		if (vc4->hvs && vc4->hvs->hvs5)
+		if (vc4->is_vc5)
 			vc5_hvs_pv_muxing_commit(vc4, state);
 		else
 			vc4_hvs_pv_muxing_commit(vc4, state);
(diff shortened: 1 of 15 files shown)


drivers/gpu/drm/vc4/vc4_kms.c
```diff
@@ -395,7 +395,7 @@ static void vc4_atomic_commit_tail(struct drm_atomic_state *state)
 		old_hvs_state->fifo_state[channel].pending_commit = NULL;
 	}
 
-	if (vc4->hvs && vc4->hvs->hvs5) {
+	if (vc4->is_vc5) {
```
popcornmix: This breaks fkms (where `vc4->hvs` would be null). Suggest `if (vc4->is_vc5 && !vc4->firmware_kms)` as fix.

drivers/gpu/drm/vc4/vc4_kms.c
```diff
@@ -427,7 +427,7 @@ static void vc4_atomic_commit_tail(struct drm_atomic_state *state)
 
 	drm_atomic_helper_cleanup_planes(dev, state);
 
-	if (vc4->hvs && vc4->hvs->hvs5) {
+	if (vc4->is_vc5) {
```
popcornmix: This breaks fkms (where `vc4->hvs` would be null). Suggest `if (vc4->is_vc5 && !vc4->firmware_kms)` as fix.

popcornmix:
I've been running with this on Pi4 with kms over weekend without issue.
fkms threw up a couple a null pointer dereferences (annoying this always occurs in the dead time after uart output stops but before you can log in to see dmesg), but I've suggested fixes for them.
Booted with kms and fkms on PI2.
No issues spotted reading commits.

6by9:
>  (annoying this always occurs in the dead time after uart output stops but before you can log in to see dmesg),

Blacklist ```vc4```. Reboot. Wait for the system to come up. SSH in and run ```sudo modprobe vc4```. You may need a second SSH session to be able to view the logs.

mripard:
Thanks for your comments, I'll address them.

> > (annoying this always occurs in the dead time after uart output stops but before you can log in to see dmesg),
> 
> Blacklist `vc4`. Reboot. Wait for the system to come up. SSH in and run `sudo modprobe vc4`. You may need a second SSH session to be able to view the logs.

Alternatively, we can disable the fbdev emulation in Kconfig (CONFIG_DRM_FBDEV_EMULATION), and then try to set a mode by hand.

popcornmix:
> Blacklist vc4. Reboot. Wait for the system to come up. SSH in and run sudo modprobe vc4

Excellent. That works for finding this hang.

> Alternatively, we can disable the fbdev emulation in Kconfig (CONFIG_DRM_FBDEV_EMULATION), and then try to set a mode by hand.

Yes, this also works.

My scheme was adding a `v7_flush_kern_dcache_area` after `prb_final_commit` then using VideoCore debugger to dump memory from address of `__log_buf` (found from `System.map`). But both of these are simpler.

6by9:
Spot those of us who have also been hunting this style of NULL deref or lockup in the past :-)

pelwell:
I think this is just waiting for the fkms-proofing changes that @popcornmix suggested.

mripard:
I just pushed a new version with the changes @popcornmix requested

pelwell:
Thanks!

PR #5023, opened 2022-05-07
Add Arducam Pivariety driver and device tree overlay

ArduCAM:


arch/arm/boot/dts/overlays/arducam-pivariety-overlay.dts
```diff
@@ -0,0 +1,96 @@
+// SPDX-License-Identifier: GPL-2.0-only
+// Definitions for Arducam Pivariety camera module on VC I2C bus
+/dts-v1/;
+/plugin/;
+
+#include <dt-bindings/gpio/gpio.h>
```
pelwell: This include is spurious - no GPIO bindings are used.

pelwell (commented):
Overall, this looks good, but I have a few questions:
* Is this being upstreamed?
* Do the kernel coding guidelines allow a single-source-file driver to put definitions in a header?

ArduCAM:
Hello pelwell,

Thanks for your review,

> Is this being upstreamed?

No, we haven't submitted a patch to upstream yet, we plan to merge it into raspberry's kernel first.

> Do the kernel coding guidelines allow a single-source-file driver to put definitions in a header?

I'm not sure, but there are examples like this under the drivers/media/i2c/ directory:
https://github.com/raspberrypi/linux/blob/rpi-5.15.y/drivers/media/i2c/adv7183_regs.h

arch/arm/boot/dts/overlays/arducam-pivariety-overlay.dts
ArduCAM: It has been fixed.

pelwell:
This driver doesn't compile in rpi-5.18.y - 5.17.y was OK. This wasn't noticed at first because the Kconfig entry had a dependency on a symbol that no longer exists, but now I've corrected that, compilation is failing with these errors:
```
drivers/media/i2c/arducam-pivariety.c:69:35: error: field ‘bus’ has incomplete type
   69 |  struct v4l2_fwnode_bus_mipi_csi2 bus;
      |                                   ^~~
drivers/media/i2c/arducam-pivariety.c: In function ‘pivariety_get_mbus_config’:
drivers/media/i2c/arducam-pivariety.c:949:19: error: ‘V4L2_MBUS_CSI2_LANE_MASK’ undeclared (first use in this function)
  949 |  const u32 mask = V4L2_MBUS_CSI2_LANE_MASK;
      |                   ^~~~~~~~~~~~~~~~~~~~~~~~
drivers/media/i2c/arducam-pivariety.c:949:19: note: each undeclared identifier is reported only once for each function it appears in
drivers/media/i2c/arducam-pivariety.c:955:5: error: ‘struct v4l2_mbus_config’ has no member named ‘flags’
  955 |  cfg->flags = (pivariety->lanes << __ffs(mask)) & mask;
      |     ^~
```
Note that the current rpi-5.18.y includes a commit that disables the driver in the defconfigs to avoid breaking everyone else's builds

6by9:
It needs a fixup equivalent to 6a7bdd89f50d399dd02847e6f398d408b086df50. Give me a few mins.

PR #5026, opened 2022-05-09
EMC2301 driver

6by9:
Largely a driver from mainline (not accepted), but with various tweaks by Phil and myself.

For CM4IO it needs ```dtoverlay=i2c-fan,emc2301,i2c_csi_dsi``` or similar. Can we make an easier line of either ```dtoverlay=cm4io-fan```, or ```dtoverlay=i2c-fan,cm4io```?

I haven't squashed the changes together so Phil can see where I've deviated from his code, but can do before we merge.


arch/arm/boot/dts/overlays/i2c-fan-overlay.dts
```diff
@@ -83,8 +78,8 @@
 		i2c_csi_dsi =	<&frag100>,"target:0=",<&i2c_csi_dsi>,
 				<0>,"+101+102";
 		addr =		<&emc2301>,"reg:0";
-		minrpm =	<&emc2301>,"min-rpm;0";
-		maxrpm =	<&emc2301>,"max-rpm;0";
+		minrpm =	<&emc2301>,"emc2305,pwm-min;0";
```
pelwell: The documentation uses `minpwm` and `maxpwm`, which make more sense. But it highlights that the PWM setting is device-specific in a way that an RPM setting wouldn't be (provided the driver supports RPMs). Perhaps this would be better as a dedicated CM4 fan overlay? I don't know.

6by9: > The documentation uses minpwm and maxpwm, which make more sense

Yes replaced the docs but not the override name.

This is the downside of the emc2305 driver over the emc2301 one - PWM setting vs RPM.

Then again the RPM range is variable between any two fans. At least default PWM settings should guarantee the fan spins (the controller does give a boost if it thinks it is stalled), whereas the default speed settings won't. (eg my fan here has a range of 200 to 1800 rpm, so the defaults you had used of 3500 and 5500 rpm aren't applicable)

Documentation/devicetree/bindings/hwmon/microchip,emc2305.yaml
```diff
@@ -0,0 +1,54 @@
...
+
+$id: http://devicetree.org/schemas/hwmon/emc2305.yaml#
+$schema: http://devicetree.org/meta-schemas/core.yaml#
+
+title: Microchip EMC2305 RPM-based PWM Fan Speed Controller
+
+properties:
+  compatible:
+    enum:
+      - microcip,emc2305
```
pelwell: 1. Typo.
2. The driver uses `smsc,emc2305`.

drivers/hwmon/emc2305.c
```diff
@@ -0,0 +1,640 @@
...
+	{ "emc2305", 0 },
+	{ "emc2303", 0 },
+	{ "emc2302", 0 },
+	{ "emc2301", 0 },
+	{ }
+};
+MODULE_DEVICE_TABLE(i2c, emc2305_ids);
+
+static const struct of_device_id emc2305_dt_ids[] = {
+	{ .compatible = "smsc,emc2305" },
```
pelwell: Should we change this to "microchip,emc2305"? Or add it as a synonym?

Documentation/devicetree/bindings/hwmon/microchip,emc2305.yaml
6by9: Bindings are from the patch submitted upstream which Rob H has already slated. I guess we ought to fix them up if we're applying it downstream.

https://www.microchip.com/investor/Pressrelease/MCHP%20and%20SMSC%20Announce%20the%20Acquisition%20of%20SMSC%20by%20MCHP.050212.pdf
So SMSC is now Microchip.
I guess we add both to binding and driver.

pelwell: Since this is a new driver, I see no reason not to use `microchip` everywhere.

6by9: OK, done.

6by9:
Edits made, our patches squashed, and updated the PR.

drivers/hwmon/emc2305.c
```diff
@@ -0,0 +1,640 @@
...
+		val = i2c_smbus_read_byte_data(client, EMC2305_REG_FAN_CFG(i));
+		/* Set RNGx so minRPM=500 */
+		val &= ~0x60;
+		i2c_smbus_write_byte_data(client, EMC2305_REG_FAN_CFG(i), val);
+	}
+
+	/* Acknowledge any existing faults. Stops the device responding on the
+	 * SMBus alert address.
+	 */
+	i2c_smbus_read_byte_data(client, EMC2305_REG_FAN_STATUS);
```
pelwell, line 613: I get a compile error on this line because `EMC2305_REG_FAN_STATUS` is undefined. Do you have some unsubmitted changes?

6by9: More likely a fluffed ```rebase -i```. I also seem to have lost the line reading REG_FAN_STALL_STATUS (0x25) which is also needed.

Updated.

pelwell:
Thanks!

PR #5028, opened 2022-05-11
dtoverlays: Fix incorrect property name for display rotation

suriyanth:


pelwell:
Thanks - hopefully that's the last overlay for a DRM display driver that needs correcting.

Issue #5032, opened 2022-05-12
5.18 Unicam seems to have issues


6by9:
### Describe the bug

IMX290 native driver should support 1920x1080@60fps, and that works fine on 5.15.38.
It doesn't work on 5.18.0-rc6, but does work if we've incorrectly configured the clock frequency to be twice the actual oscillator (ie halving the sensor speed, and 1080p60 config becomes 1080p30).

We need to test all sensors on 5.18 to see if this is just some odd configuration issue on this module, or something more significant.

### Steps to reproduce the behaviour

5.18.0-rc6 with Arducam IMX462 (updated IMX290/327).

```dtoverlay=imx290,media-controller=0``` for 37.125MHz clock config.
```v4l2-ctl --stream-mmap=3 --stream-count=1000 --stream-to=/dev/null``` doesn't produce frames.

```dtoverlay=imx290,clock-frequency=74250000,media-controller=0``` for (incorrect) 74.25MHz clock config.
```v4l2-ctl --stream-mmap=3 --stream-count=1000 --stream-to=/dev/null``` DOES produce frames, but only at half speed.

Doesn't matter whether 2 or 4 data lanes are used.

### Device (s)

Raspberry Pi CM4

### System

Raspberry Pi OS.
Self-built debug firmware from 29th March 2022.
5.18.0-rc6

### Additional context

@davidplowman @naushir FYI

**Comment by lategoodbye** [comment]
Not sure this is related by there is a pending clock fix for 5.18:
https://lore.kernel.org/lkml/ee39ddd1-bfce-012d-5e04-448d779ed995@i2se.com/T/

**Comment by ArduCAM** [comment]
I'm not sure if this is the same issue, but we had a camera related issue on 5.18,
We tested IMX219, IMX477 and some Arducam Pivariety series and they all have the same problem

**System Info:**
```
pi@raspberrypi:~ $ uname -a
Linux raspberrypi 5.18.0-rc6-v7l+ #11 SMP Fri May 13 08:42:04 CST 2022 armv7l GNU/Linux
```
```
pi@raspberrypi:~ $ sudo vcgencmd version
May 10 2022 17:23:09 
Copyright (c) 2012 Broadcom
version a256c26efd3e41d6ce5c9513bf2a3909a118e1cc (clean) (release) (start)
```
```
pi@raspberrypi:~ $ cat /etc/os-release 
PRETTY_NAME="Raspbian GNU/Linux 11 (bullseye)"
[...]

PR #5033, opened 2022-05-12
Arducam Pivariety 5.18 fixes

6by9:
As noted on #5023 the kernel APIs have changed slightly and PiVariety would no longer build. Fix it up.

Do note #5032, so I haven't been able to test this (then again I haven't actually run a PiVariety module at all).


pelwell:
Ta.

PR #5035, opened 2022-05-16
Add Arducam 64MP driver and device tree overlay

ArduCAM:


pelwell:
Is this your own sensor? If not, it should have the manufacturer's name and model, e.g. "sony,imx477" as used in the driver on which this is based.

And yes, I should have asked the same question of the Pivariety driver.

Documentation/devicetree/bindings/media/i2c/arducam,64mp.yaml
```diff
@@ -0,0 +1,113 @@
...
+      endpoint:
+        type: object
+        properties:
+          data-lanes:
+            description: |-
+              The sensor supports either two-lane, or four-lane operation.
+              For two-lane operation the property must be set to <1 2>.
+            items:
+              - const: 1
+              - const: 2
```
6by9: The sensor supports either 2 or 4 lanes, so the binding should follow suit and list both options.

drivers/media/i2c/arducam_64mp.c
```diff
@@ -0,0 +1,2630 @@
...
+	{0x020f, 0x00},
+	{0x4018, 0x00},
+	{0x4019, 0x00},
+	{0x401a, 0x00},
+	{0x401b, 0x00},
+	{0x3400, 0x01},
+	{0x3092, 0x01},
+	{0x3093, 0x00},
+	{0x0350, 0x00},
+};
```
6by9, line 1115: Just doing a simple diff on these tables seems to imply that there are about 50% of them that don't change between modes. Could those be moved into ```mode_common_regs```?

drivers/media/i2c/arducam_64mp.c
```diff
@@ -0,0 +1,2630 @@
...
+		.width = 9152,
+		.height = 6944,
+		.line_length_pix = 0xb6b2,
+		.crop = {
+			.left = ARDUCAM_64MP_PIXEL_ARRAY_LEFT,
+			.top = ARDUCAM_64MP_PIXEL_ARRAY_TOP,
+			.width = 9248,
+			.height = 6944,
+		},
+		.timeperframe_min = {
```
6by9: ```timeperframe_min ``` is the same as ```timeperframe_default ``` in each of these cases. Do we need both?

drivers/media/i2c/arducam_64mp.c
```diff
@@ -0,0 +1,2630 @@
...
+/*
+ * Initialisation delay between XCLR low->high and the moment when the sensor
+ * can start capture (i.e. can leave software standby), given by T7 in the
+ * datasheet is 8ms.  This does include I2C setup time as well.
+ *
+ * Note, that delay between XCLR low->high and reading the CCI ID register (T6
+ * in the datasheet) is much smaller - 600us.
+ */
+#define ARDUCAM_64MP_XCLR_MIN_DELAY_US	8000
+#define ARDUCAM_64MP_XCLR_DELAY_RANGE_US	1000
```
6by9, line 1282: This text and the numbers are identical to imx477. Is it exactly the same (including the timings being T6, T7, and T8), or is this just copy/paste?

drivers/media/i2c/arducam_64mp.c
```diff
@@ -0,0 +1,2630 @@
...
+	{0x401a, 0x00},
+	{0x401b, 0x00},
+	{0x3400, 0x01},
+	{0x3092, 0x01},
+	{0x3093, 0x00},
+	{0x0350, 0x00},
+};
+
+/* Mode configs */
+static const struct arducam_64mp_mode supported_modes_10bit[] = {
```
6by9: You appear to only have 10bit modes, so this can be ```supported_modes``` without reference to the bit depth.

drivers/media/i2c/arducam_64mp.c
```diff
@@ -0,0 +1,2630 @@
...
+				      ARDUCAM_64MP_REG_VALUE_08BIT,
+				      arducam_64mp->long_exp_shift);
+}
+
+static int arducam_64mp_set_ctrl(struct v4l2_ctrl *ctrl)
+{
+	struct arducam_64mp *arducam_64mp =
+		container_of(ctrl->handler, struct arducam_64mp, ctrl_handler);
+	struct i2c_client *client = v4l2_get_subdevdata(&arducam_64mp->sd);
+	int ret = 0;
```
6by9: I don't see a path through the function which doesn't set ```ret```, so no need to initialise to 0.

drivers/media/i2c/arducam_64mp.c
```diff
@@ -0,0 +1,2630 @@
...
+		return ret;
+	}
+
+	/* Request optional enable pin */
+	arducam_64mp->reset_gpio = devm_gpiod_get_optional(dev, "reset",
+							   GPIOD_OUT_HIGH);
+
+	/*
+	 * The sensor must be powered for arducam_64mp_identify_module()
+	 * to be able to read the CHIP_ID register
```
6by9: You're not reading the CHIP_ID register of the sensor, so does this apply?

Checking on my preproduction module, the EEPROM does appear to be controlled by the same regulator, so presumably yes it is required. The comment could be updated though.

arch/arm/boot/dts/overlays/arducam-64mp-overlay.dts
```diff
@@ -0,0 +1,94 @@
...
+					arducam_64mp_0: endpoint {
+						remote-endpoint = <&csi1_ep>;
+						clock-lanes = <0>;
+						data-lanes = <1 2>;
+						clock-noncontinuous;
+						link-frequencies =
+							/bits/ 64 <456000000>;
+					};
+				};
+			};
```
6by9, line 41: The module includes a VCM for focus control. Presumably there will be an follow up update to add in support for that.

Documentation/devicetree/bindings/media/i2c/arducam,64mp.yaml
```diff
@@ -0,0 +1,113 @@
...
+title: Arducam 1/1.7-Inch 64Mpixel CMOS Digital Image Sensor
+
+maintainers:
+  - Lee Jackson <info@arducam.com>
+
+description: |-
+  The Arducam 1/1.7-Inch 64Mpixel CMOS active pixel digital image sensor
+  with an active array size of 9248 x 6944. It is programmable through
+  I2C interface. The I2C address is fixed to 0x1A as per sensor data sheet.
+  Image data is sent through MIPI CSI-2, which is configured as either 2 or
```
6by9: which can be configured for operation with either 2 or 4 data lanes.

Documentation/devicetree/bindings/media/i2c/arducam,64mp.yaml
```diff
@@ -0,0 +1,113 @@
...
+              For two-lane operation the property must be set to <1 2>.
+            items:
+              - const: 1
+              - const: 2
+
+          clock-noncontinuous:
+            type: boolean
+            description: |-
+              MIPI CSI-2 clock is non-continuous if this property is present,
+              otherwise it's continuous.
```
6by9: I don't see usage of this flag in the driver.

Checking the register set against the data sheet the driver appears to only ever run in non-continuous mode (the default). In which case the driver ought to ensure that flag is set.

Then again the joys of V4L2 means that you can still incorrectly configure the CSI2 receiver to continuous mode regardless of what the sensor driver has been told.

6by9 (commented):
Generally minor comments.

Use of a second pad for metadata isn't acceptable for mainline, but is how embedded data is handled currently on the Pi branches. Assuming it actually works on this sensor, then that is fine by me.

ArduCAM:
Hi 6by9,
Thank you very much for your review,

> Use of a second pad for metadata isn't acceptable for mainline, but is how embedded data is handled currently on the Pi branches. Assuming it actually works on this sensor, then that is fine by me.

Yes, it works. I missed this, I think I'd better put the embedded part of the code into a separate commit.

Documentation/devicetree/bindings/media/i2c/arducam,64mp.yaml
ArduCAM: Added a new items, a little unsure, since the camera driver does not have 4lane configuration, will this affect dt-bindings?

drivers/media/i2c/arducam_64mp.c
ArduCAM: I moved some registers into common.

drivers/media/i2c/arducam_64mp.c
ArduCAM: timeperframe_min is removed, treat the default as the maximum achievable framerate.

drivers/media/i2c/arducam_64mp.c
ArduCAM: Missed it out, it's a little different from imx477 and has been modified.

drivers/media/i2c/arducam_64mp.c
ArduCAM: done.

drivers/media/i2c/arducam_64mp.c
ArduCAM: done.

drivers/media/i2c/arducam_64mp.c
ArduCAM: You are right, I simply modified the comment.

Documentation/devicetree/bindings/media/i2c/arducam,64mp.yaml
ArduCAM: done.

Documentation/devicetree/bindings/media/i2c/arducam,64mp.yaml
ArduCAM: I'm not sure if my change is correct, I simply changed it to `clock-noncontinuous: true`

pelwell:
Do you have an answer for my question?

> Is this your own sensor? If not, it should have the manufacturer's name and model, e.g. "sony,imx477" as used in the driver on which this is based.

> And yes, I should have asked the same question of the Pivariety driver.

ArduCAM:
Pelwell
Sorry, I intended to reply you through email phil.elwell, but seems it doesn't work.

pelwell:
My aim is to avoid having a number of different drivers for the same sensor packaged by different manufactures - it's better to have one generic driver - hence my question about the sensor. Feel free to continue the conversation by email if you prefer - I'm just `phil@raspberrypi.com`.

pelwell:
Apologies - with holidays and other commitments I've dropped the ball on this PR. I don't have any serious objections to this driver, and no technical issues that I can see, so I'm merging.

adminy:
Can this driver be upstreamed to the mainline linux kernel? (I'd love to try this camera on soquartz cm4)

Also quick question, ArduCam B0399 64MB Autofocus is this PR driver or is it 
https://github.com/raspberrypi/linux/pull/5023

Issue #5019, opened 2022-05-04
SDHCI leaks DMA buffers leading to a full swiotlb buffer


P33M:
### Describe the bug

When recovering data off poorly SD cards that have extremely long read access cycle times, eventually I get this splat:

```
[689174.139067] mmc0: Got data interrupt 0x00000002 even though no data operation was in progress.
[689174.139562] swiotlb_tbl_map_single: 53 callbacks suppressed
[689174.139575] sdhci-iproc fe340000.mmc: swiotlb buffer is full (sz: 32768 bytes), total 32768 (slots), used 30824 (slots)
[689174.139591] ------------[ cut here ]------------
[689174.139598] WARNING: CPU: 0 PID: 1804 at drivers/mmc/host/sdhci.c:1154 sdhci_send_command+0xacc/0xf18
[689174.139625] Modules linked in: cdc_acm dm_mod 8021q garp stp llc snd_soc_hdmi_codec sg brcmfmac brcmutil v3d gpu_sched cfg80211 raspberrypi_hwmon rfkill i2c_brcmstb bcm2835_codec(C) vc4 bcm2835_v4l2(C) bcm2835_isp(C) bcm2835_mmal_vchiq(C) v4l2_mem2mem cec videobuf2_dma_contig videobuf2_vmalloc videobuf2_memops vc_sm_cma(C) drm_kms_helper videobuf2_v4l2 videobuf2_common videodev mc snd_bcm2835(C) snd_soc_core snd_compress snd_pcm_dmaengine rpivid_mem snd_pcm snd_timer snd syscopyarea sysfillrect sysimgblt fb_sys_fops uio_pdrv_genirq uio nvmem_rmem i2c_dev fuse drm drm_panel_orientation_quirks backlight ip_tables x_tables ipv6
[689174.139940] CPU: 0 PID: 1804 Comm: kworker/0:2H Tainted: G        WC        5.15.32-v7l+ #1538
[689174.139954] Hardware name: BCM2711
[689174.139962] Workqueue: kblockd blk_mq_run_work_fn
[689174.139980] Backtrace:
[689174.139991] [<c0bcaf20>] (dump_backtrace) from [<c0bcb16c>] (show_stack+0x20/0x24)
[689174.140019]  r7:00000482 r6:c0e3cf44 r5:00000080 r4:60000093
[689174.140026] [<c0bcb14c>] (show_stack) from [<c0bcf7e8>] (dump_stack_lvl+0x70/0x94)
[...]

PR #5036, opened 2022-05-16
mmc: block: Don't do single-sector reads during recovery

P33M:
See https://github.com/raspberrypi/linux/issues/5019

If an SD card has degraded performance such that IO operations time out
then the MMC block layer will leak SG DMA mappings during recovery. It
retries the same SG and this causes the leak, as it is mapped twice -
once in sdhci_pre_req() and again during single-block reads in
sdhci_prepare_data().

Resetting the card (including power-cycling if a regulator for vmmc is
present) ought to be enough to recover a stuck state, so for now remove
the call to mmc_blk_read_single().

### DIFF
--- a/drivers/mmc/core/block.c
+++ b/drivers/mmc/core/block.c
@@ -1871,7 +1871,11 @@ static void mmc_blk_mq_rw_recovery(struct mmc_queue *mq, struct request *req)
 	}
 
 	/* FIXME: Missing single sector read for large sector size */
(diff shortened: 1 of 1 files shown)


pelwell:
Using `if (0 && !mmc_large_sector(card) ...` instead avoids a compiler warning and makes it clear that something non-standard is happening.

ghost:
> makes it clear that something non-standard is happening.

You probably want a comment as well, particularly if this is something that is intended to be fixed/improved in the future.

PR #5037, opened 2022-05-16
bcm2835-codec: Support SAND 8bit

6by9:
Wanted by @jc-kynesim for HEVC to H264 transcode.

I think I have the plumbing of the relevant data correct, but it would be nice for him to confirm it.

I have not verified the output (V4L2 CAPTURE) side from video_decode/isp is represented correctly.


6by9:
John has said it works for him on the encoder, although he's still fighting getting a non-aligned to 16 height encoded. It's not currently clear where that issue is.

Merge now, or wait for all the ducks to be in a row?

jc-kynesim:
I seem to have crop issues with I420 too which I'm still debugging, so that is probably a separate issue and the SAND support itself seems good so I'd go with merge now.

PR #5038, opened 2022-05-17
kms/vc4_hdmi: Refuse 4096x2160@60 hdmi modes

popcornmix:
These are no reliable without overclocking.
See: https://github.com/raspberrypi/linux/issues/5034

### DIFF
--- a/drivers/gpu/drm/vc4/vc4_drv.h
+++ b/drivers/gpu/drm/vc4/vc4_drv.h
@@ -351,6 +351,12 @@ struct vc4_hvs {
 	 * available.
 	 */
 	bool vc5_hdmi_enable_scrambling;
+
+	/*
+	 * 4096x2160@60 requires a core overclock to work, so register
+	 * whether that is sufficient.
+	 */
+	bool vc5_hdmi_enable_4096by2160;
 };
 
 struct vc4_plane {
--- a/drivers/gpu/drm/vc4/vc4_hdmi.c
+++ b/drivers/gpu/drm/vc4/vc4_hdmi.c
@@ -1822,6 +1822,7 @@ static int vc4_hdmi_encoder_atomic_check(struct drm_encoder *encoder,
 	struct drm_connector *connector = &vc4_hdmi->connector;
 	struct drm_connector_state *old_conn_state = drm_atomic_get_old_connector_state(conn_state->state, connector);
 	struct vc4_hdmi_connector_state *old_vc4_state = conn_state_to_vc4_hdmi_conn_state(old_conn_state);
+	struct vc4_dev *vc4 = to_vc4_dev(connector->dev);
 	unsigned long long pixel_rate = mode->clock * 1000;
 	unsigned long long tmds_rate;
 	int ret;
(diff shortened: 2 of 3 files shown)


popcornmix:
Submitted as draft to hear comments.
We could make this conditional on core_freq being overclocked, but I don't think that is sufficient.
You may also need core_freq_min set (need to check if that is something driver can discover).

popcornmix:
Updated to also check if core_freq_min is boosted and only then allow these modes.

6by9:
Looks reasonable to me as we know that 4096x2160 does require a higher core clock than 3840x2160 to allow the HVS time to generate all the pixels, even though they use the same pixel clock.

alanbork:
As long as they are still accessible if the proper lines are added to config.txt this seems like a reasonable thing to do.

popcornmix:
In my testing the 4096x2160@50/60 modes were omitted by default and were present when
```
core_freq_min=600
core_freq=600
```
were added.

ghost:
This should perhaps be documented somewhere, since the Pi 4 is described as supporting 4K. The obvious location for documentation would be https://github.com/raspberrypi/documentation. Would you like me to submit a PR to have this added? 

An alternative would be to have the software print out a statement that the requested mode requires overclocking.

popcornmix:
I think the common definition of 4k is 3840x2160, but yes, feel to submit a PR that clarifies that 4096x2160@60 is not supported without a core frequency overclock to 600MHz.

ghost:
> I think the common definition of 4k is 3840x2160...

Just for completeness here, 4096 x 2160 @ 60Hz (and 50Hz) is listed as supported by Pi 4 currently in the documentation (at https://www.raspberrypi.com/documentation/computers/config_txt.html#hdmi_mode).

ghost:
Can I just check - if core_freq_min is set to 600, without setting core_freq, then is that not the same as setting both core_freq_min and core_freq to 600? Just trying to nail down the simplest config.

alanbork:
There's enough diversity of usage both amature and profesional that the
term 4k should probably be retired in favor of quoting the actual
horizontal resolution.

rofehr:
4K is 4.096 x 2.160 and 
UHD is 3.840 x 2.160.

alanbork:
> 4K is 4.096 x 2.160 and UHD is 3.840 x 2.160.

plenty of TVs proudly proclaim 4k on the box but are really 3840x

even *if* the terms are well defined the usage doesn't reflect that

popcornmix:
@pelwell okay to merge?

drivers/gpu/drm/vc4/vc4_drv.h
```diff
@@ -351,6 +351,12 @@ struct vc4_hvs {
 	 * available.
 	 */
 	bool vc5_hdmi_enable_scrambling;
+
+	/*
+	 * 4096x2160@60 requires a core overclock to work, so register
+	 * whether that is sufficient.
+	 */
+	bool vc5_hdmi_enable_4096by2160
```
pelwell: Missing semi-colon.

popcornmix: Indeed. gcc only gives a warning for this, which seems surprising.

popcornmix:
Updated.

PR #5039, opened 2022-05-17
Sony imx296 sensors support

naushir:
This driver is not yet fully upstreamed (https://patchwork.linuxtv.org/project/linux-media/list/?series=6989), but I don't anticipate any major changes from this revision.


pelwell:
The overlay is lacking the necessary README and Makefile entries, but otherwise it looks OK (as far as my reading went, which is not very far).

6by9:
I'm reading through it at present.

Presumably the main driver commit is unmodified from that submitted to mainline? I'm just looking at things like exposure being clipped when set, rather than the range being reduced based on VBLANK.

pelwell:
I will need to remember to change the Kconfig entry when forward-porting to 5.18.

arch/arm/boot/dts/overlays/imx296-overlay.dts
```diff
@@ -0,0 +1,103 @@
...
+		target = <&i2c0mux>;
+		__overlay__ {
+			status = "okay";
+		};
+	};
+
+	reg_frag: fragment@5 {
+		target = <&cam1_reg>;
+		cam_reg: __overlay__ {
+			startup-delay-us = <300000>;
```
6by9: Do you validate that this delay needs to be this high?

arch/arm/boot/dts/overlays/imx296-overlay.dts
```diff
@@ -0,0 +1,103 @@
...
+		__overlay__ {
+			status = "okay";
+		};
+	};
+
+	clk_frag: fragment@1 {
+		target = <&cam1_clk>;
+		__overlay__ {
+			status = "okay";
+			clock-frequency = <37125000>;
```
6by9, line 22: Add an override for the clock frequency?

naushir:
> Presumably the main driver commit is unmodified from that submitted to mainline? I'm just looking at things like exposure being clipped when set, rather than the range being reduced based on VBLANK.

The main driver commit is unchanged from the mainline patch. Any additions ought to be done on top.

arch/arm/boot/dts/overlays/imx296-overlay.dts
```diff
@@ -0,0 +1,103 @@
...
+
+				rotation = <180>;
+				orientation = <2>;
+
+				port {
+					imx296_0: endpoint {
+						remote-endpoint = <&csi_ep>;
+						clock-lanes = <0>;
+						data-lanes = <1>;
+						clock-noncontinuous;
```
6by9, line 67: I can't find reference in the datasheet as to whether the clock lane does drop to LP-11 during blanking or not. Ought to validate it.

6by9 (commented):
Looks reasonable.

It'd be nice to fix up the exposure range to follow VBLANK, and to add flips.

6by9:
V4L2_CID_LINK_FREQUENCY isn't advertised either, nor is the link frequency validated from DT.

6by9:
We're also missing an update to the defconfigs.

pelwell:
True - and the overlay commit should be called `overlays: ...`, not `dt-bindings: ...`.

naushir:
I'll update with all the fixups and post an update soon.

arch/arm/boot/dts/overlays/imx296-overlay.dts
naushir: Yes, it does seem to need 300ms.  Not surprising, as it probably has the same reset circuitry as the imx477.

arch/arm/boot/dts/overlays/imx296-overlay.dts
naushir: Yes, need clock-noncontinuous otherwise streaming does not behave.

arch/arm/boot/dts/overlays/imx296-overlay.dts
naushir: In fact, this does not seem enough - I have had to up it to 500ms for glitch free operation!

naushir:
New revision:
- Fixed the "overlay:" comment
- Added h/v flip and link frequency ctrls.
- Switch to 1 frame gain delay.
- Adjusted delays for reliable operation.

arch/arm/configs/bcm2711_defconfig
```diff
@@ -944,6 +944,7 @@ CONFIG_VIDEO_ARDUCAM_PIVARIETY=m
 CONFIG_VIDEO_IMX219=m
 CONFIG_VIDEO_IMX258=m
 CONFIG_VIDEO_IMX290=m
+CONFIG_VIDEO_IMX296=m
```
6by9, line 947: Unless there is a reason for a module to only apply to eg CM4, they would normally be added to all 5 normal defconfigs. eg https://github.com/raspberrypi/linux/commit/c0286215da0d807ac3717bb51ec813bd0784f398

drivers/media/i2c/imx296.c
```diff
@@ -610,7 +610,7 @@ static int imx296_enum_mbus_code(struct v4l2_subdev *sd,
 		return -EINVAL;
 
 	code->code = sensor->mono ? MEDIA_BUS_FMT_Y10_1X10
-		   : MEDIA_BUS_FMT_SBGGR10_1X10;
```
6by9: Commit text would more normally start as
```
media: i2c: imx296:
```
or possibly only 
```
media: imx296:
```
Same for the other patches.

arch/arm/boot/dts/overlays/README
```diff
@@ -1978,6 +1978,19 @@ Params: 4lane                   Enable 4 CSI2 lanes. This requires a Compute
         cam0                    Adopt the default configuration for CAM0 on a
                                 Compute Module (CSI0, i2c_vc, and cam0_reg).
```
pelwell, line 1980: You need an extra blank line either side of the new entry.

drivers/media/i2c/imx296.c
```diff
@@ -171,6 +171,10 @@
 #define IMX296_GTTABLENUM				IMX296_REG_8BIT(0x4114)
 #define IMX296_CTRL418C					IMX296_REG_8BIT(0x418c)
 
+#define IMX296_STANDBY_DELAY		1500
+#define IMX296_STREAM_ON_DELAY		2000
+#define IMX296_STREAM_OFF_DELAY		2000
```
6by9, line 176: Where have these delays come from? The only numbers I can see are in "Sensor Setting Flow (Sensor Master Mode)" for a >1ms delay "Wait for internal regulator stabilization".

drivers/media/i2c/imx296.c
```diff
@@ -539,6 +543,7 @@ static int imx296_stream_off(struct imx296 *sensor)
 	int ret = 0;
 
 	imx296_write(sensor, IMX296_CTRL0A, IMX296_CTRL0A_XMSTA, &ret);
+	usleep_range(IMX296_STREAM_OFF_DELAY, 2*IMX296_STREAM_OFF_DELAY);
```
6by9, line 633: "Sensor Setting Flow (Sensor Master Mode)"  lists the correct sequence as setting STANDBY=1, wait >1ms, and then XMSTA=1.

drivers/media/i2c/imx296.c
```diff
@@ -970,6 +975,8 @@ static int imx296_identify_model(struct imx296 *sensor)
 		return ret;
 	}
 
+	usleep_range(IMX296_STANDBY_DELAY, IMX296_STANDBY_DELAY);
```
6by9: That's a very tight usleep when min and max are identical.
Why do we need to delay here anyway? Any chance of verifying Laurent's comment about needing to be out of standby?

drivers/media/i2c/imx296.c
```diff
@@ -256,6 +258,36 @@ static int imx296_write(struct imx296 *sensor, u32 addr, u32 value, int *err)
...
+	MEDIA_BUS_FMT_SRGGB10_1X10,
+	MEDIA_BUS_FMT_SGRBG10_1X10,
+	MEDIA_BUS_FMT_SGBRG10_1X10,
+	MEDIA_BUS_FMT_SBGGR10_1X10,
+};
+
+static u32 imx296_mbus_code(struct imx296 *sensor)
+{
+	unsigned int i = 0;
+
```
6by9, line 282: The imx219 version of this function includes
```
lockdep_assert_held(&imx219->mutex);
```
as this function needs to be called with the mutex held

pelwell:
Are you happy your comments have been addressed, @6by9?

drivers/media/i2c/imx296.c
```diff
@@ -591,7 +591,7 @@ static int imx296_setup(struct imx296 *sensor)
 	imx296_write(sensor, IMX296_CTRL418C, sensor->clk_params->ctrl418c,
 		     &ret);
 
-	imx296_write(sensor, IMX296_GAINDLY, IMX296_GAINDLY_NONE, &ret);
+	imx296_write(sensor, IMX296_GAINDLY, IMX296_GAINDLY_1FRAME, &ret);
```
6by9, line 604: Commit text:
>In order to behave more simlar to the rolling shutter sensors, set the gain dealy to 1 frame. This simplifies userland control of the gain value.

"In order to behave in a similar manner to rolling shutter sensors..."
s/dealy/delay

drivers/media/i2c/imx296.c
```diff
@@ -461,6 +465,10 @@ static int imx296_ctrls_init(struct imx296 *sensor)
 				     ARRAY_SIZE(imx296_test_pattern_menu) - 1,
 				     0, 0, imx296_test_pattern_menu);
 
+	v4l2_ctrl_new_int_menu(&sensor->ctrls, NULL, V4L2_CID_LINK_FREQ,
```
6by9, line 470: The ```v4l2_ctrl_handler_init(&sensor->ctrls, 9);``` line hasn't been updated to advise the of the extra control. Did it get updated with the HFLIP and VFLIP patch as well?

drivers/media/i2c/imx296.c
naushir: I recall these delays directly from Sony as things were just plain unreliable without them.  I'll dig out the emails (if I still have them) to verify.

drivers/media/i2c/imx296.c
naushir: I'll put 2x for the max like the others.
We do need to come out of standby with a delay to read the id register, otherwise the I2C does fail. From what I can tell, this is because the register is in a different chip bank that is powered down in standby.

drivers/media/i2c/imx296.c
naushir: Hmmm, do I need this to guard against changing h/v flip ctrls?  I can't see any other reason for it...

drivers/media/i2c/imx296.c
naushir: Will fix.

drivers/media/i2c/imx296.c
6by9: It looks like the mutex is acquired in the correct places, at least based on imx219.

I believe the theory is that two different clients can open the subdev simultaneously, one changing the flip and the other getting/setting the format. The mutex ensures they are consistent.
See https://github.com/raspberrypi/linux/commit/745d4612d2c853c00abadbf69799c8aee7f99c39 and https://github.com/raspberrypi/linux/commit/81015221a269e0d9bf6dbc3fda099b09a876ebb7

arch/arm/configs/bcm2709_defconfig
```diff
@@ -928,6 +928,7 @@ CONFIG_VIDEO_ARDUCAM_PIVARIETY=m
 CONFIG_VIDEO_IMX219=m
 CONFIG_VIDEO_IMX258=m
 CONFIG_VIDEO_IMX290=m
+CONFIG_VIDEO_IMX296=m
```
6by9, line 931: Of course now the commit text is wrong ;-)

drivers/media/i2c/imx296.c
6by9: My comment was more that the datasheet states ```IMX296_CTRL00, IMX296_CTRL00_STANDBY``` first, then wait >1ms, then ```IMX296_CTRL0A, IMX296_CTRL0A_XMSTA```. The register writes are currently reversed.

drivers/media/i2c/imx296.c
6by9: Commit text still reads oddly, and has a typo in ```simlar```

arch/arm/configs/bcm2709_defconfig
naushir: Arghhhhh

drivers/media/i2c/imx296.c
naushir: Yes, I did notice that as well.  Both my driver and Laurent's driver use the given sequence (don't ask me why, it's been too long), and I kept it unchanged because it does seem to work.

drivers/media/i2c/imx296.c
naushir: Arghhhhh again!

drivers/media/i2c/imx296.c
6by9: OK, we're not using Master/Slave mode at present, so it may be irrelevant. One to bear in mind should we introduce it.

6by9:
LGTM. Ship it!

drivers/media/i2c/imx296.c
naushir: In time we will definitely add support for this!

Issue #4588, opened 2021-09-17
incorrect lines doubled in 480i (only) using kms/fkms


alanbork:
480i (but not 1080i) content seems to have some kind of resampling/line skipping in both x and y directions, see test image below. I have verified that this occurs under both kms and fkms on a pi4, with releases as new as efe79c1df6df24e666c08dd68e1636ed24210fc1** and at least as old as 9dbcd6dfed16cc7de4a67b9ca445c4847f68c4c2 - I haven't tested outside this range, but I suspect it goes way further back in time I just didn't have a good test image to catch it until now.

![image](https://user-images.githubusercontent.com/56519925/133853945-414ca1bc-d650-4255-9c3b-52d57c1d181a.png)

(observed on left, desired on right).

this is not a TV deinterlacing artifact as the same image is displayed just fine when rendered on a pi0 using the dispmanx firmware. note also that if it were deinterlacing you'd expect it to be an issue in one direction only, but it's clear that vertical lines are being dropped/doubled, too. In this particular case it's a TCL S423 TV but I suspect it's not TV specific, either.  The same TV shows the test image just fine in 480p. 

test environment: bare console, no x11, using drm/kms to initialize and draw to screen.

footnote:
**actually, fkms doesn't work at all for 480i with this release, you have to go back a few releases before it can produce 480i, but when it does work at all, it has this distortion.

popcornmix:
Sounds like [this issue](https://forum.libreelec.tv/thread/24415-le-10-beta-for-i4-force-hdmi-resolution/) which had a recent fix for fkms, but hasn't been fixed on kms.

alanbork:
looks like that's not yet hit rpi-update? or is there a proper hash to grab and test this  from the Hexxeh github releases?

I know the below rpi-update shouldn't work but figured I'd try anyway ;-)

/root #rpi-update b80f36b3fb1566c028771d4a5ac9b23edf22069a
 *** Raspberry Pi firmware updater by Hexxeh, enhanced by AndrewS and Dom
[...]
 *** Invalid hash given

alanbork:
also, to be clear, both fkms and kms show this under efe79c1df6df24e666c08dd68e1636ed24210fc1, the most recent release I have access to.

pelwell:
rpi-update gets its releases from a different repo: https://github.com/Hexxeh/rpi-firmware/commits/master
The equivalent release in this case is f8357e4d1dbc6b98dcec9acae9dbba5d9fb86040

[...]

PR #5040, opened 2022-05-17
HDMI Pixel doubled modes use HDMI instead of PV

6by9:
#4588 
These patches move to using the HDMI block for pixel doubling instead of the pixel valve.

It means that the PV config is restricted to horizontal timings being a multiple of 2, but does try to fix the mode up as the standard 480i and 576i modes fail this criteria.


popcornmix:
For me the following modes produced pictures on two monitors here (a Dell 4k and a Dell 1920x1200) with this PR
```
modetest -M vc4 -s 32:1920x1080i-30
modetest -M vc4 -s 32:1920x1080i-25
modetest -M vc4 -v -s 32:720x480i-30
modetest -M vc4 -v -s 32:720x576i-25
```
I don't have a test card that shows the reported pixel order issue, but I believe this method of doubling won't have the issue.

PR #5041, opened 2022-05-18
clk-raspberrypi: Minimise firmware clocks

popcornmix:
A follow-on from https://github.com/raspberrypi/linux/pull/4940

Currently, we don't try to minimise most firmware clocks.
e.g. if hevc requests a 550MHz clock, it continues to run at 550MHz after hevc driver has dropped its request,
and the same happens for all other non-minimised clocks.

This results in a higher required chip voltage, and so more heat/power than is necessary.

When minimising clocks you do need to switch from the `clk_set_rate` API, to `clk_set_min_rate` API.
I believe all fw clocks are using that API, apart from v3d which I've updated.


popcornmix:
@mripard does this seem okay?
In testing, I do see clocks return to low frequencies where previously they get stuck high.

mripard:
It looks good to me :)

popcornmix:
@pelwell okay to merge?

PR #5043, opened 2022-05-19
bcm2835-codec encode format fixes

6by9:
@jc-kynesim Hopefully these mean you can set the buffer geometry happily via the selection API.

The encoder is given the resolution specified on the output (V4L2 CAPTURE) side, so that resolution does need to be set correctly.

### DIFF
--- a/drivers/staging/vc04_services/bcm2835-codec/bcm2835-v4l2-codec.c
+++ b/drivers/staging/vc04_services/bcm2835-codec/bcm2835-v4l2-codec.c
@@ -1899,6 +1899,8 @@ static int vidioc_s_selection(struct file *file, void *priv,
 {
 	struct bcm2835_codec_ctx *ctx = file2ctx(file);
 	struct bcm2835_codec_q_data *q_data = NULL;
+	struct vchiq_mmal_port *port = NULL;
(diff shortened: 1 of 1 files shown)


jc-kynesim:
The V4L2 spec on https://www.linuxtv.org/downloads/v4l-dvb-apis-new/userspace-api/v4l/dev-encoder.html section 4.5.2.5 has CAPTURE S_FMT width,height as ignored so requiring it be set correctly seems a bit out of spec.  Frame size (within geometry) and cropped frame size seem to be set by selection API on OUTPUT

jc-kynesim:
Though having said that, on trivial testing, it does seem to work for me :-)

6by9:
Yes you are right.that the spec does say that.
I'll update the patches to make it read only and the output size to be set from the input cropped format.

6by9:
(Sorry, messed up my branches as I thought this had been merged. Now reinstated back to where it should be).

pelwell:
Are you happy with this, @jc-kynesim ?

jc-kynesim:
Works for me - so yes I'm happy.  Sorry about the delay - took a bit before I was back in a situation where I could test.

PR #5047, opened 2022-05-25
vc-sm-cma: Restore correct cache maintainance operations

popcornmix:
We have been using the more expensive flush operations rather than
invalidate and clean since kernel rpi-5.9.y

These are exposed with:
52f1453513ba95084ab811a030032fe605b0cbe2 Re-expose some dmi APIs for use in VCSM

But I believe that commit was dropped when (non-cma) vc-sm was dropped,
and didn't get updated when the commit was restored

### DIFF
--- a/drivers/staging/vc04_services/vc-sm-cma/vc_sm.c
+++ b/drivers/staging/vc04_services/vc-sm-cma/vc_sm.c
@@ -1247,7 +1247,9 @@ static void (*cache_op_to_func(const unsigned int cache_op))
 		return NULL;
 
 	case VC_SM_CACHE_OP_INV:
+		return dmac_inv_range;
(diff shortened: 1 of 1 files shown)


popcornmix:
Ping @6by9

6by9:
Doesn't this break 64bit kernels? Those functions only appear to exist for 32bit ARM architectures.

6by9:
> Doesn't this break 64bit kernels? Those functions only appear to exist for 32bit ARM architectures.

Ah, this is all under a ```#ifndef CONFIG_ARM64``` as the UAPI is 32bit only. Forget that then.

6by9:
This will do no harm, but AFAIK it is only used by the Pi3 HEVC decode, not generic cache management of vcsm buffers.

PR #5050, opened 2022-05-30
media: i2c: imx477: Correct minimum exposure lines

davidplowman:
The minimum number of exposure lines value (IMX477_EXPOSURE_MIN) was
previously 20 but this is not correct. The datasheet is not completely
explicit, however the new value of 4 has been tested with all the
sensor modes supported by this driver, and matches the lowest exposure
value of 114us that could be achieved wtih Raspberry Pi's legacy
firmware driver.

### DIFF
--- a/drivers/media/i2c/imx477.c
+++ b/drivers/media/i2c/imx477.c
@@ -61,7 +61,7 @@ MODULE_PARM_DESC(trigger_mode, "Set vsync trigger mode: 1=source, 2=sink");
 /* Exposure control */
 #define IMX477_REG_EXPOSURE		0x0202
 #define IMX477_EXPOSURE_OFFSET		22
-#define IMX477_EXPOSURE_MIN		20
+#define IMX477_EXPOSURE_MIN		4
 #define IMX477_EXPOSURE_STEP		1
 #define IMX477_EXPOSURE_DEFAULT		0x640
 #define IMX477_EXPOSURE_MAX		(IMX477_FRAME_LENGTH_MAX - \


davidplowman:
@naushir  @6by9

naushir:
LGTM

drivers/media/i2c/imx477.c
```diff
@@ -61,7 +61,7 @@ MODULE_PARM_DESC(trigger_mode, "Set vsync trigger mode: 1=source, 2=sink");
 /* Exposure control */
 #define IMX477_REG_EXPOSURE		0x0202
 #define IMX477_EXPOSURE_OFFSET		22
-#define IMX477_EXPOSURE_MIN		20
+#define IMX477_EXPOSURE_MIN		4
```
6by9, line 64: Table 5-24 Integration time setting (In case of FRM_LENGHT_CTL = 0h) appears to be fairly explicit that a COARSE_INTEG_TIME of > FRM_LENGTH_LINES - 20 is prohibited.

Also description of register 0350h FRM_LENGTH_CTL also states that it controls whether the frame length is changed when FRM_LENGTH_LINES < COARSE_INTEG_TIME + alpha, where alpha = 20 for this sensor.

Or am I getting confused here with the value that should be set in EXPOSURE_OFFSET for max exposure time? That appears that it should be 20 instead of 22.

davidplowman: Yes, I agree things are not explained in the clearest way and there is some interpretation involved!

I took these various statements to be a restriction on FRM_LENGTH_LINES, but not actually implying anything about the min value of COARSE_INTEG_TIME. Indeed, in our old firmware driver it goes lower than 20 by some way (giving 114us min exposure rather than 571us).

I thought EXPOSURE_OFFSET being 22 rather than 20 was maybe a question of tests including the case of equality or not, but honestly I don't know!

PR #5052, opened 2022-06-01
drm/vc4_hdmi: Allow hotplug detect to be forced

popcornmix:
See: https://forum.libreelec.tv/thread/24783-tv-avr-turns-back-on-right-after-turning-them-off

While the kernel provides a :D flag for assuming device is connected,
it doesn't stop this function from being called and generating a cec_phys_addr_invalidate
message when hotplug is deasserted.

That message provokes a flurry of CEC messages which for many users results in the TV
switching back on again and it's very hard to get it to stay switched off.

It seems to only occur with an AVR and TV connected but has been observed across a
number of manufacturers.

The issue started with https://github.com/raspberrypi/linux/pull/4371
and this provides an optional way of getting back the old behaviour

### DIFF
--- a/drivers/gpu/drm/vc4/vc4_hdmi.c
+++ b/drivers/gpu/drm/vc4/vc4_hdmi.c
@@ -40,6 +40,8 @@
 #include <linux/component.h>
 #include <linux/gpio/consumer.h>
 #include <linux/i2c.h>
+#include <linux/module.h>
+#include <linux/moduleparam.h>
 #include <linux/of_address.h>
 #include <linux/of_gpio.h>
 #include <linux/of_platform.h>
@@ -112,6 +114,10 @@
 #define CEC_CLOCK_FREQ 40000
 
 #define HDMI_14_MAX_TMDS_CLK   (340 * 1000 * 1000)
+
+/* bit field to force hotplug detection. bit0 = HDMI0 */
+static int force_hotplug = 0;
+module_param(force_hotplug, int, 0644);
 
 static const char * const output_format_str[] = {
 	[VC4_HDMI_OUTPUT_RGB]		= "RGB",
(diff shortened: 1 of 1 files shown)


drivers/gpu/drm/vc4/vc4_hdmi.c
```diff
@@ -257,6 +262,9 @@ vc4_hdmi_connector_detect(struct drm_connector *connector, bool force)
 			connected = true;
 	}
 
+	if (force_hotplug)
```
6by9: Very minor optimisation (sorry about indentation)
```
if (force_hotplug) {
   connected = true
} else if (vc4_hdmi->hpd_gpio) {
		if (gpiod_get_value_cansleep(vc4_hdmi->hpd_gpio))
			connected = true;
	} else {
		if (vc4_hdmi->variant->hp_detect &&
		    vc4_hdmi->variant->hp_detect(vc4_hdmi))
			connected = true;
```
There's no point in checking the actual HPD line if forcing it, and the check of force_hotplug is unconditional either way.

popcornmix: Ok

HiassofT:
Just wondering: shouldn't that be better a per-connector bitset instead of a single global flag? eg 1 for HDMI0, 2 for HDMI1, 3 for both?

popcornmix:
Updated to address two comments.

HiassofT:
We got some positive feedback that this helps to work around the CEC issues in LibreELEC https://github.com/LibreELEC/LibreELEC.tv/issues/6432#issuecomment-1146916310

HiassofT:
Thanks for merging!

Could you also add that to rpi-5.10.y? We use that for our stable LibreELEC 10 branch.

It's not a 100% clean pick as the gpio get differs a bit (gpiod_get... in 5.15), here's the patch I used for 5.10
https://github.com/HiassofT/LibreELEC.tv/blob/le10-test/packages/linux/patches/raspberrypi/linux-010-pr5052-backport.patch

pelwell:
We're trying to restrict 5.10 changes to bug fixes. @popcornmix Do you think this qualifies?

popcornmix:
I'm happy with that. Kodi (Matrix) is still using 5.10 kernel and they were the main reporters of the issue this fixes.
I've cherry-picked it.

PR #5055, opened 2022-06-06
Add RGB332 and RGBA4444 formats to vc4

6by9:
From https://forums.raspberrypi.com/viewtopic.php?t=334760

Verified on Pi4 that they render correctly through modetest for smpte (patch needed to add RGB332/BGR233 support to modetest).
~Still building to test on a CM3, so please hold off on the merge before it is confirmed to work.~


6by9:
Tested on Pi4 and CM3 now and they all seem to be correct based on testing with modetest.

PR #5056, opened 2022-06-07
Enable kernel PPS

by:
Kernel PPS is now readily available in the kernel config to be selected


pelwell:
There are many unused options available in the kernel - why do we want this one?

by:
Increases precision of synchronisation with external clocks w/ PPS (e.g., GPS) significantly.

pelwell:
The pps Kconfig file says:
```
config NTP_PPS
	bool "PPS kernel consumer support"
	depends on !NO_HZ_COMMON
	help
	  This option adds support for direct in-kernel time
	  synchronization using an external PPS signal.

	  It doesn't work on tickless systems at the moment.
```
Note the `depends` and the last line. Raspberry Pi kernels are "tickless" - they define `NO_HZ`, and hence `NO_HZ_COMMON` - so setting `CONFIG_NTP_PPS=y` in the defconfig file will have no effect; even after running `make ARCH=arm64 bcm2711_defconfig` you will find that `CONFIG_NTP_PPS` does not appear in `.config`.

Issue #3282, opened 2019-10-11
(ds1307/ds3231) hwclock: ioctl(3, RTC_UIE_ON, 0): Invalid argument


boatbodger:
**Describe the bug**
When running hwclock on a Pi 4 with a DS3231 RTC, an "invalid argument" error occurs
`Waiting for clock tick...`
`ioctl(3, RTC_UIE_ON, 0): Invalid argument'

**Additional context**
Ultimately my problem with reading the clock was being caused by noise on the ground connection between the R-Pi and the clock chip.
The 'Invalid argment' is still occurring, but is non-fatal - many thanks to pelwell for the explanation below.

boatbodger:
Out of curiosity, I left the system alone for three hours, and then issued 'hwclock' again.
That time I got a response.
If I do it again immediately afterwards, the error recurs.
It feels like reading it "knocks it out", and after some time it is sufficiently recovered for another read:
`root@ftk-mys-svr:~# hwclock`
`2019-10-11 23:16:05.987908+01:00`
`root@ftk-mys-svr:~# hwclock`
`hwclock: ioctl(RTC_RD_TIME) to /dev/rtc0 to read the time failed: Invalid argument`
`root@ftk-mys-svr:~# top`
`top - 23:19:13 up  3:34,  4 users,  load average: 1.00, 1.04, 1.01`

boatbodger:
After more testing, using verbose mode, even on a *successful* hwclock, the ioctl Invalid argument occurs, but hwclock reports receiving a clock tick:
`root@ftk-mys-svr:~# hwclock --verbose --show`
`hwclock from util-linux 2.33.1`
`System Time: 1570885984.794421`
`Trying to open: /dev/rtc0`
`Using the rtc interface to the clock.`
etc etc then
`Waiting for clock tick...`
`ioctl(3, RTC_UIE_ON, 0): Invalid argument`
`Waiting in loop for time from /dev/rtc0 to change`
`...got clock tick`
`Time read from Hardware Clock: 2019/10/12 13:13:05`
etc etc then
`2019-10-12 14:13:04.971991+01:00`
`root@ftk-mys-svr:~#`
So despite the Invalid argument it succeeds.

By contrast, a failed attempt reads thus:
`root@ftk-mys-svr:~# hwclock --verbose --show`
`hwclock from util-linux 2.33.1`
`System Time: 1570885997.152220`
`Trying to open: /dev/rtc0`
`Using the rtc interface to the clock.`
[...]

PR #5057, opened 2022-06-07
Enable RTC UIE emulation to avoid error in hwclock output

by:
Solves https://github.com/raspberrypi/linux/issues/3282


pelwell:
The Kconfig entry says:
```
config RTC_INTF_DEV_UIE_EMUL
        bool "RTC UIE emulation on dev interface"
        depends on RTC_INTF_DEV
        help
          Provides an emulation for RTC_UIE if the underlying rtc chip
          driver does not expose RTC_UIE ioctls. Those requests generate
          once-per-second update interrupts, used for synchronization.

          The emulation code will read the time from the hardware
          clock several times per second, please enable this option
          only if you know that you really need it.
```
That doesn't sound very appealing. Are you certain this won't end penalising users who don't care about the interrupt?

by:
It will always be a trade-off, but as I've seen so many problems and errors messages with many different RTCs, it would be worthwhile, from my perspective; why not test it more widely with 5.18 first? – Here it works great without negative side-effects or performance issues since quite some months...

its0x08 (approved):

PR #5069, opened 2022-06-25
w1 bulk read per master

herby74:
A bulk read of slave devices of a `w1_bus_master` should be possible by writing `trigger` to the `therm_bulk_read` entry in the device tree of the bus master. However, the entry `therm_bulk_read` only shows up for the first w1_bus_master. (Reason: The static variable counter `u16 bulk_read_device_counter` is common for all masters in w1_therm.c.) This change fixes the behavior by using a dedicated `bulk_slave_count` variable per master.


Ka10ken1:
nice

PR #5079, opened 2022-07-04
drm/vc4: Correct interlaced timings again

6by9:
The back porch timings were correct, only the sync offset was wrong.
Correct timing is now reported for 1080i and 576i, but the h offset is
incorrect for 480i for non-obvious reasons.

Fixes: c0ba150ce468 "drm/vc4: Correct HDMI timing registers for interlaced modes"


pelwell:
> the h offset is incorrect for 480i

Is that a regression or just the result of an incomplete fix?

6by9:
Incomplete fix.

The hardware block is programmed apparently correctly, but according to the analyser it doesn't produce the timings anticipated.

PR #5080, opened 2022-07-04
IMX258 fixes for Vision Components modules

6by9:
Adds alternate compatible string to adjust register set.

### DIFF
--- a/Documentation/devicetree/bindings/media/i2c/imx258.yaml
+++ b/Documentation/devicetree/bindings/media/i2c/imx258.yaml
@@ -14,10 +14,13 @@ description: |-
   type stacked image sensor with a square pixel array of size 4208 x 3120. It
   is programmable through I2C interface.  Image data is sent through MIPI
   CSI-2.
+  There are a number of variants of the sensor which cannot be detected at
+  runtime, so multiple compatible strings are required to differentiate these.
 
 properties:
   compatible:
     const: sony,imx258
+    const: sony,imx258-pdaf
 
   assigned-clocks: true
   assigned-clock-parents: true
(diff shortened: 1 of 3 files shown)


Documentation/devicetree/bindings/media/i2c/imx258.yaml
```diff
@@ -14,10 +14,13 @@ description: |-
   type stacked image sensor with a square pixel array of size 4208 x 3120. It
   is programmable through I2C interface.  Image data is sent through MIPI
   CSI-2.
+  There are a number of variants of the sensor which can not be detected at
```
pelwell: "cannot" is preferred (ducks...)

drivers/media/i2c/imx258.c
```diff
@@ -1782,6 +1817,19 @@ static int imx258_probe(struct i2c_client *client)
 
 	imx258->csi2_flags = ep.bus.mipi_csi2.flags;
 
+#ifdef CONFIG_ACPI
```
pelwell: Is this still necessary? i2c.h already provides a null (and NULL-returning) implementation of `i2c_of_match_device` for the case that `CONFIG_OF` is not enabled.

Documentation/devicetree/bindings/media/i2c/imx258.yaml
6by9: Done

drivers/media/i2c/imx258.c
6by9: Fixed. I'd initially used of_match_device, but then swapped.

PR #5084, opened 2022-07-05
bcm2835-v4l2-codec: support H.264 5.0 and 5.1 levels

mdevaev:
For TC358743, we needed to support a capture resolution of 1920x1200 and encode it to H.264. Following [that](https://forums.raspberrypi.com/viewtopic.php?t=291447#p1762296), for resolutions greater than 1080p, we need to use H.264 levels 5.0 and 5.1. This is supported by the firmware, so I've just added the appropriate control values to the driver.

Tested on RPi4 with PiKVM V3 HAT and uStreamer. Both modes are working perfectly.

### DIFF
--- a/drivers/staging/vc04_services/bcm2835-codec/bcm2835-v4l2-codec.c
+++ b/drivers/staging/vc04_services/bcm2835-codec/bcm2835-v4l2-codec.c
@@ -2189,6 +2189,12 @@ static int bcm2835_codec_set_level_profile(struct bcm2835_codec_ctx *ctx,
 		case V4L2_MPEG_VIDEO_H264_LEVEL_4_2:
 			param.level = MMAL_VIDEO_LEVEL_H264_42;
 			break;
+		case V4L2_MPEG_VIDEO_H264_LEVEL_5_0:
+			param.level = MMAL_VIDEO_LEVEL_H264_5;
+			break;
+		case V4L2_MPEG_VIDEO_H264_LEVEL_5_1:
+			param.level = MMAL_VIDEO_LEVEL_H264_51;
+			break;
 		default:
 			/* Should never get here */
 			break;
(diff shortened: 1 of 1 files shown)


6by9:
Is 1920x1200 actually encoded correctly?

Memory says there have been issues reported with 1920x1200 video decode, and encode uses decode to create the reference frames for prediction. I can't find the threads at present, but I seem to recall that those macroblocks beyond the level 4.1 limit of 8192 ended up corrupted.

mdevaev:
Well, I didn't notice any problems. We tested it, switched modes back and forth, the video was encoded perfectly. Perhaps the problem was before the Pi4 boards. Here the screenshot for 5.1 level:

image

6by9:
Just confirmed that video decode of 1920x1200 does have issues.
```gst-launch-1.0 videotestsrc ! video/x-raw,width=1920,height=1200,framerate=30/1,format=NV12 ! x264enc ! filesink location=1920x1200.h264``` to create a file
```gst-launch-1.0 filesrc location=1920x1200.h264 ! h264parse ! v4l2h264dec ! kmssink``` to then play back, and the bottom part of the screen has corruption.

I'll fire up the firmware debugger to see whether the reference frames are corrupt or not.

mdevaev:
Could you try to do m2m encoding and software decoding? Maybe the problem is reproduced only when decoding.

mdevaev:
Tested encoding with `gst-launch-1.0 videotestsrc ! video/x-raw,width=1920,height=1200,framerate=30/1,format=UYVY ! v4l2h264enc ! 'video/x-h264,level=(string)5' ! filesink location=1920x1200.h264` and played with software decoding on my laptop. The encoding process is working fine.
image

mdevaev:
@6by9 sup?

6by9:
Sorry, been stupidly busy, and only just had a chance to hook the debugger up to look at the reconstructed images of the encoder.

It does look like the encoder is NOT hitting the same issue as the decoder, and the recon images are fine.

@pelwell I'm happy, anything from you?
I wish we could put a big caveat in that the encoder only supports parts of levels 5 and 5.1, but I don't know of a path to validate that sort of thing.

pelwell:
The implementation is uncontroversial.

mdevaev:
Many thanks!

PR #5086, opened 2022-07-06
Overlay for GalaxyCore GC9A01 displays

juliannojungle:
This includes an overlay for easy use of displays based on GalaxyCore's `GC9A01` single chip driver, attaching them on `spi0` as `fb1` framebuffer.

Originally created at https://github.com/juliannojungle/gc9a01-overlay where includes all documentation. Merging in raspberrypi kernel for easy use of said displays on future Raspberry Pi OS releases.


pelwell:
Hi @juliannojungle,

I'm not familiar with the GC9A01 display, but the overlay looks like it could work - I'm sure you've been using it enough that we can say it has been tested. I have a few problems with the Pull Request:

1. I guess you used the GitHub editor to add the overlay source - it uses DOS line endings (each line has a carriage return (0xd/Ctrl-M) as well as a line feed (0xa/Ctrl-J)). This looks ugly using the Git tools.
2. The "✨" is attention-seeking.
3. We require a README entry for each overlay - see https://github.com/raspberrypi/linux/blob/rpi-5.15.y/arch/arm/boot/dts/overlays/README. You should be able to copy most of it from a similar overlay such as piscreen. Please preserve the alphabetical ordering, strict indentation (spaces only - no TABs), and two blank lines between entries.
4. You'll also need to add an entry to the Makefile (https://github.com/raspberrypi/linux/blob/rpi-5.15.y/arch/arm/boot/dts/overlays/Makefile). Again, respect the alphabetical ordering.

juliannojungle:
Hi, @pelwell!
Yes, I've been using this overlay for quite some time. It works nicely on [the display I have](https://www.waveshare.com/wiki/1.28inch_LCD_Module).
The "2 - ✨" thing is a commit message pattern adopted by many repositories, based on [some rules](https://gitmoji.dev/). The "sparkles" means "introducing new feature". I tottaly missed that I shouldn't be using that here, sorry for that. Already removed.
I'll fix the first topic (carriage return and line feed) and adjust the files from topics 3 and 4 later today.

Thank you very much for the instructions about the merge (I searched the wiki but couldn't find any guidelines).

juliannojungle:
Hi, @pelwell. I believe I fixed said problems with the Pull Request now. Thank you for taking time to instruct me on how to proceed.

pelwell:
Nice job. I forgot to mention the compatible string - I prefer to use just `brcm,bcm2835` for overlays compatible with all Pis and `brcm,bcm2711` for those that only work with Pi 4 - so I've pushed a fix-up patch.

reece146:
Feature request:

Would it be possible to extend this overlay such that is possible to have parameters passed to attach the driver and frame buffer to a specific SPI device?

Example:

I would like to have round displays and frame buffers attached at spidev0,1 spidev1,0 and spidev1,2. These would be created as fb1, fb2, and fb3. Or something of that nature.

Is this possible?

As far as I can tell from my rudimentary understanding of how this works, this overlay sets up spidev0,1 and fb1 and that's the only functional configuration available.

Thanks

PR #5099, opened 2022-07-19
dtoverlays: Add nohdmi options to vc4-kms-v3d overlays

6by9:
There are cases where hotplug detect is incorrectly wired
or it is desired that the HDMI outputs are disabled.

Add an override to vc4-kms-v3d and vc4-kms-v3d-pi4 to do that.

https://forums.raspberrypi.com/viewtopic.php?t=337623


pelwell:
Thanks for that. While reviewing the commit, I noticed that `pixelvalve3` gets enabled by two separate fragments in the -pi4 variant - fragment 9 and fragment 21 - and the `composite` parameter disables the former and enables the latter. The second fragment appeared in this commit: https://github.com/raspberrypi/linux/commit/8b126ef52fa67217b4fea2aa76bc3fcbb9a55bf2

Is this some dark magic beyond my understanding or totally spurious?

6by9:
I got it right on first hit, including updating the README!! Time for a serious "work meeting"!

> Is this some dark magic beyond my understanding or totally spurious?

Spurious :-(
The output won't be enabled unless the "encoder" (HDMI/DPI/DSI/VEC/TXP) on the end of the chain is enabled so it doesn't cause any harm having PV3 enabled even with the VEC disabled.

JinShil:
Thank you for this.  How does one know when this is available through `apt dist-upgrade/full-upgrade`?

pelwell:
There is no rigid timetable for the next release, but it is likely to be in the next month or two. However, I think it's safe to say that there won't be another release that doesn't contain this patch. That means that if you download the overlays below (or just the one you need) and copy them to `/boot/overlays` then you will be safe to `apt upgrade` without fear of losing it (provided you are up-to-date before copying over the downloaded files).

Pi 4 (includes Pi 400 and CM4) : https://drive.google.com/file/d/1GXWqOpgQxGKiOVNobSumK-ChdgAyNLaf/view?usp=sharing
Older Pis: https://drive.google.com/file/d/1gI5pg2RinQ7LwsOo-_Qae1a01wYM047K/view?usp=sharing

PR #5101, opened 2022-07-21
overlays: Make more overlays runtime-capable

pelwell:
The kernel now rejects the creation of a DT node with the same name as
an existing node, so take care not to do that.

### DIFF
--- a/arch/arm/boot/dts/bcm270x-rpi.dtsi
+++ b/arch/arm/boot/dts/bcm270x-rpi.dtsi
@@ -132,6 +132,7 @@
 
 &cpu_thermal {
 	/delete-node/ trips;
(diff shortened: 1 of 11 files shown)


pelwell:
@6by9 Would you mind casting an eye over this to see if I've done anything stupid?

6by9:
AFAICT Those look reasonable. 

I assume the fragment ordering in apds9960-overlay is so that the interrupt bit gets merged into the apds9960 fragment before that is then merged into the base DT? That's the bit I always get wrong.

pelwell:
Thanks. You're right about the fragment ordering; in practise the overlay application logic should cope with one level of reordering (intra- first, then inter-), but I prefer to set a good example not to rely on it.

PR #5106, opened 2022-07-26
Port MultiGenerational LRU (MGLRU) from Common Android Kernel

Pzqqt:
These patches are from the `android13-5.15` branch of the [Common Android Kernel](https://android.googlesource.com/kernel/common). There are basically no conflicts when porting them.

There has been a lot of [news](https://www.phoronix.com/news/MGLRU-July-2022-Performance) that mglru is very good for Linux performance.

Although this is an old v9 patches (currently the latest [v13 patches](https://lore.kernel.org/lkml/20220706220022.968789-1-yuzhao@google.com/)), I believe it would still be better to have it.

I've tried compiling a kernel for my Raspberry Pi 4B and using it for two days. It seems to work just fine.


popcornmix:
This is interesting and I'd like to hear about any quantifiable improvements (e.g. benchmarks, or normal operations that run faster on a Pi).

It's unlikely we'll merge this at this stage - if it's not ready for upstream, then it's not ready for us.
[lwn](https://lwn.net/Articles/894859/) discusses some of the concerns with merging this upstream and that applies to us even more so with a version that is still being tweaked.

If there are significant performance advantages that normal pi users will see, then we may consider getting some testing (e.g. through an opt-in next branch of rpi-update).

Pzqqt:
> This is interesting and I'd like to hear about any quantifiable improvements (e.g. benchmarks, or normal operations that run faster on a Pi).
> 
> It's unlikely we'll merge this at this stage - if it's not ready for upstream, then it's not ready for us. [lwn](https://lwn.net/Articles/894859/) discusses some of the concerns with merging this upstream and that applies to us even more so with a version that is still being tweaked.
> 
> If there are significant performance advantages that normal pi users will see, then we may consider getting some testing (e.g. through an opt-in next branch of rpi-update).

Ok thanks for your reply. I will try to benchmark and compare, then let you know the results.

Btw I got a warning when compiling with clang, you can cherry-pick [this commit](https://github.com/Pzqqt/kernel_raspberrypi_4b/commit/7462741893ec83b33134355988cc8bf382e8000c) to fix it.

Pzqqt:
This is the result of redis-benchmark:

Command: `redis-benchmark -c 100 -n 50000 --csv`

Test items|MGLRU (v9) rps|No MGLRU rps|MGLRU (v9) avg_latency_ms|No MGLRU avg_latency_ms|
--|:--:|--:|--:|--:|
SET|17787.26|17677.21|2.843|2.862|
GET|17708.52|17667.85|2.848|2.858|
MSET|17917.94|17667.85|2.826|2.868|

There seems to be some improvement, but it's not noticeable.😂

As for the increase in normal operations speed, to be honest, this is difficult to test and not easy to perceive.

popcornmix:
> As for the increase in normal operations speed, to be honest, this is difficult to test and not easy to perceive.

Without a compelling benefit, I'm not sure it's worth bringing it in early (when it quite likely still has kinks to iron out).
We generally move to new LTS kernels (sounds likely to be 5.20), and include a testing period using next branch of rpi-update.
It sounds like it could be merged upstream by that point, where users could test.

Pzqqt:
> > As for the increase in normal operations speed, to be honest, this is difficult to test and not easy to perceive.
> 
> Without a compelling benefit, I'm not sure it's worth bringing it in early (when it quite likely still has kinks to iron out). We generally move to new LTS kernels (sounds likely to be 5.20), and include a testing period using next branch of rpi-update. It sounds like it could be merged upstream by that point, where users could test.

OK, thanks for your hard work. You keep users always having the latest version of the Linux kernel, I love that.

Let's wait for the official arrival of MGLRU in 5.20. I'll still add MGLRU to my privately built kernel, that's my interest.

popcornmix:
Comment here if you notice any clear improvements.
I think from the discussion, the plan is to make it enabled by default, but if that is left to the kernel builder (or distribution) then we may want to make that choice, so corroborating information is useful.

ghost:
Linking [the discussion](https://forums.raspberrypi.com/viewtopic.php?t=340335) from the Pi forum.

emansom:
@Pzqqt How much RAM is your Raspberry Pi equipped with? Might be interesting to see if it behaves better on e.g. 1G models.

@yuzhaogoogle Would've expected better results on the redis benchmark, would the benefit be more prominent in sustained (e.g. multiple days) real world workloads across multiple applications running (e.g. Redis+MariaDB+PHP)?

ghost:
The default `redis-benchmark` data size is 3 bytes, according to the [manual](https://redis.io/docs/reference/optimization/benchmarks/). So `redis-benchmark -c 100 -n 50000 --csv` might have only used a few megabytes of memory. `redis-cli dbsize` can show the actual database size.

MGLRU won't improve the performance for underutilized systems, i.e., systems always having free memory.

@emansom Yes, if `redis-benchmark` uses a lot of memory, especially with zram or zswap.

PR #5115, opened 2022-08-04
rpi-simple-soundcard: adds definitions for the HiFiBerry AMP3

j-schambacher:
Currently, the definition is a copy of the the Infineon Merus AMP, but it supports the devices full sample rate of 192ksps. The Infineon/Merus card allows a maximum of 96ksps and we are working with Infineon to fix their definition. (We will issue a PR later.) The Hifiberry card does use also different GPIOs in line with our other products.


pelwell:
Hi Jörg, I'm not going to be able to look at this until next week, but it is on the list.

pelwell:
Looks good to me!

j-schambacher:
Thanks, Phil! I've sent my proposed modifications for the Merus Ampal ready to the Infineon team for review. They or me will send a PR soon.

PR #5121, opened 2022-08-11
vc4-kms-v3d COB and interrupt fixes

6by9:
The COB allocation changes could do with some stress testing.

This PR has stemmed from testing triple head modes, and finding that the code enabling underrun interrupts was actually enabling other events for HVS channel 2, which then weren't handled and caused bad things to happen.

With these changes I've got 4k HDMI and dual DSI (800x480) running happily in X. I haven't got dual 4k monitors to hand to test.


drivers/gpu/drm/vc4/vc4_hvs.c
```diff
@@ -485,23 +485,25 @@ static int vc4_hvs_init_channel(struct vc4_hvs *hvs, struct drm_crtc *crtc,
...
 			    (oneshot ? SCALER_DISPCTRLX_ONESHOT : 0);
-	else
+		dispbkgndx |= SCALER_DISPBKGND_AUTOHS;
+	} else {
 		dispctrl |= VC4_SET_FIELD(mode->hdisplay,
 					  SCALER5_DISPCTRLX_WIDTH) |
 			    VC4_SET_FIELD(mode->vdisplay,
 					  SCALER5_DISPCTRLX_HEIGHT) |
 			    (oneshot ? SCALER5_DISPCTRLX_ONESHOT : 0);
+	}
```
pelwell, line 504: Are you certain that the AUTOHS can't possibly be set already in SCALER_DISPBKGNDX(chan), i.e. is it better to clear it here (or unconditionally, earlier)?

6by9: You appear to be right - the firmware sets it, therefore we ought to explicitly clear it.

6by9:
Hold fire on these - I've just seen glitches on my 4k HDMI screen.
I'm out for a week now, so it'll just have to wait.

6by9:
It looks like things may just be on the edge with 4k60.

On this CM4, a standard Raspberry Pi OS image (5.15.32 kernel) is showing occasional horizontal glitched lines (generally browny noise) if running HDMI at 4k60 and running something like iPlayer video decode. Dropping to 4k30 is clean.
The HVS doesn't appear to be signalling an underflow, so I couldn't say for certain whether it was HVS or V3D that is stuttering.

These changes don't alter that, and allow triple headed mode.
I probably ought to test with dual 4k30 though.

6by9:
It looks like my corruption was a dodgy HDMI lead! I should have thought of that one earlier.

PR #5132, opened 2022-08-18
Enable /proc/config.gz support.

arekm:
Enable IKCONFIG to enable /proc/config.gz support.
bcm2709_defconfig and bcm2711_defconfig has it but in bcm2835_defconfig
it was missing.

Noticed on:
```
# uname -a
Linux octopi 5.15.32-v8+ #1538 SMP PREEMPT Thu Mar 31 19:40:39 BST 2022 aarch64 GNU/Linux
# lsb_release -a
No LSB modules are available.
Distributor ID:	Debian
Description:	Debian GNU/Linux 11 (bullseye)
Release:	11
Codename:	bullseye
# head -n 8 /proc/cpuinfo
processor	: 0
BogoMIPS	: 38.40
Features	: fp asimd evtstrm crc32 cpuid
CPU implementer	: 0x41
CPU architecture: 8
CPU variant	: 0x0
CPU part	: 0xd03
CPU revision	: 4
# modinfo configs
modinfo: ERROR: Module alias configs not found.
# ls -ld /proc/config*
ls: cannot access '/proc/config*': No such file or directory
```

I assume it uses bcm2835_defconfig.


pelwell:
That's an upstream file - the equivalent downstream file (i.e. the one that we maintain) is bcmrpi_defconfig.

arekm:
Ok, this pull makes no sense then. Thanks.

PR #5141, opened 2022-08-23
drm/vc4: Add async update support for cursor planes

pelwell:
Now that cursors are implemented as regular planes, all cursor
movements result in atomic updates. As the firmware-kms driver
doesn't support asynchronous updates, these are synchronous, which
limits the update rate to the screen refresh rate. Xorg seems unaware
of this (or at least of the effect of this), because if the mouse is
configured with a higher update rate than the screen then continuous
mouse movement results in an increasing backlog of mouse events -
cue extreme lag.

Add minimal support for asynchronous updates - limited to cursor
planes - to eliminate the lag.

### DIFF
--- a/drivers/gpu/drm/vc4/vc4_firmware_kms.c
+++ b/drivers/gpu/drm/vc4/vc4_firmware_kms.c
@@ -672,6 +672,50 @@ static int vc4_plane_atomic_check(struct drm_plane *plane,
 	return vc4_plane_to_mb(plane, &vc4_plane->mb, new_plane_state);
 }
 
+static void vc4_plane_atomic_async_update(struct drm_plane *plane,
+					  struct drm_atomic_state *state)
+{
+	struct drm_plane_state *new_plane_state =
+		drm_atomic_get_new_plane_state(state, plane);
+
+	swap(plane->state->fb, new_plane_state->fb);
+	plane->state->crtc_x = new_plane_state->crtc_x;
+	plane->state->crtc_y = new_plane_state->crtc_y;
+	plane->state->crtc_w = new_plane_state->crtc_w;
+	plane->state->crtc_h = new_plane_state->crtc_h;
+	plane->state->src_x = new_plane_state->src_x;
+	plane->state->src_y = new_plane_state->src_y;
+	plane->state->src_w = new_plane_state->src_w;
+	plane->state->src_h = new_plane_state->src_h;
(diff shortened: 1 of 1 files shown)


pelwell:
How does this look to you, @mripard? It has the desired effect in the limited testing I have done (with `usbhid.mousepoll=5` to increase the severity of the problem).

popcornmix:
For background https://github.com/raspberrypi/linux/pull/4971 introduced the issue with fkms.
It was reported in https://github.com/raspberrypi/linux/issues/4988, so @Dark-Sky may be interested in testing.
We've seen a few reports on forum as 4971 has made it to apt.

While the issue only affects fkms users (which is deprecated), its still useful as a debugging test or a workaround for an upstream bug, so I agree this workaround is worthwhile (it fixed the issue in my testing).

drivers/gpu/drm/vc4/vc4_firmware_kms.c
```diff
@@ -672,6 +672,50 @@ static int vc4_plane_atomic_check(struct drm_plane *plane,
...
+					struct drm_atomic_state *state)
+{
+	struct drm_plane_state *new_plane_state =
+		drm_atomic_get_new_plane_state(state, plane);
+	int ret = -EINVAL;
+
+	if (plane->type == 2 &&
+	    plane->state->fb &&
+	    new_plane_state->crtc->state->active)
+		ret = vc4_plane_atomic_check(plane, state);
```
mripard: atomic_async_check has a slightly different semantic than atomic_check. atomic_check indicates whether the state is sane or not, but atomic_async_check only needs to return whether we can commit it asynchronously or not.

It means that both atomic_check and atomic_async_check will be run:
https://elixir.bootlin.com/linux/latest/source/drivers/gpu/drm/drm_atomic_helper.c#L994

So you don't need to call vc4_plane_atomic_check here, you just need to return whether it can be committed asynchronously (0) or not (< 0)

pelwell: That also works for me. Fixing commit added, will squash later. Thanks!

mripard: It looks good to me now

pelwell: Thanks again for the review.

PR #5149, opened 2022-08-29
Add support for I2C TPMs such as SLB9673

AndreasFuchsTPM:
This PR add support for TPMs with I2C interface by adding the necessary driver to the defconfig.
This driver is available upstream since kernel version 6.0

Further, it adds a device tree overlay for such TPMs.

This support is similar to #2585 but for I2C.
Given that it's configured as module, there is no impact on the zImage.


arch/arm/boot/dts/overlays/tpm-slb9673-overlay.dts
```diff
@@ -0,0 +1,50 @@
+/*
+ * Device Tree overlay for the Infineon SLB9673 Trusted Platform Module add-on
+ * boards, which can be used as a secure key storage and hwrng.
+ */
+
+/dts-v1/;
+/plugin/;
+
+/ {
+	compatible = "brcm,bcm2835", "brcm,bcm2708", "brcm,bcm2709";
```
pelwell: Just `"brcm,bcm2835";`, please.

pelwell (commented):
This looks pretty good. Apart from the compatible string change, my only issue is the omission of `arch/arm/configs/bcmrpi_defconfig` (the Pi 1 defconfig) - was that conscious decision?

AndreasFuchsTPM:
> This looks pretty good. Apart from the compatible string change, my only issue is the omission of `arch/arm/configs/bcmrpi_defconfig` (the Pi 1 defconfig) - was that conscious decision?

No, that slipped by... So I added it now.

> Just "brcm,bcm2835";, please.

Now that I added it to bcmrpi_defconfig, I guess I'll leave the compat string as is ?

pelwell:
> Now that I added it to bcmrpi_defconfig, I guess I'll leave the compat string as is ?

No - overlays that work on BCM2835, BCM2836, BCM2837 and BCM2711 should use "brcm,bcm2835". Those which are only for BCM2711 (e.g. ones that use the additional I2C and SPI interfaces) should use "brcm,bcm2711".

AndreasFuchsTPM:
> No - overlays that work on BCM2835, BCM2836, BCM2837 and BCM2711 should use "brcm,bcm2835". Those which are only for BCM2711 (e.g. ones that use the additional I2C and SPI interfaces) should use "brcm,bcm2711".

Interesting. Thanks for clarifying.

PR Updated.

PR #5154, opened 2022-08-30
Add a pullup/down parameter to PPS DTBO

rlcamp:
Square wave output from RTCs such as the DS3231 requires a pullup resistor, this commit allows the pin's internal pullup resistor to be used by specifying "pull=2" in the parameters, or "pull=1" for pulldown if necessary on other parts

### DIFF
--- a/arch/arm/boot/dts/overlays/README
+++ b/arch/arm/boot/dts/overlays/README
@@ -3102,8 +3102,8 @@ Params: gpiopin                 Input GPIO (default "18")
                                 off)
         capture_clear           Generate clear events on the trailing edge
                                 (default off)
-        pull                    Specify whether the pin is pulled up (2), down
-                                (1), or neither (0, default)
+        pull                    Desired pull-up/down state (off, down, up)
+                                Default is "off".
 
 
 Name:   pwm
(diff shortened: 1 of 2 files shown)


arch/arm/boot/dts/overlays/README
```diff
@@ -3102,6 +3102,8 @@ Params: gpiopin                 Input GPIO (default "18")
                                 off)
         capture_clear           Generate clear events on the trailing edge
                                 (default off)
+        pull                    Specify whether the pin is pulled up (2), down
```
pelwell: You don't need the magic numbers - the values `up`, `down` and `off` are also accepted (and preferred) the parameter mechanism converts them to the appropriate integers.

rlcamp: Is this a suggested documentation fix, or would it require an actual change to the .dts commit? Struggling to find documentation of this other than the existing dtbo readme sections, which have precedents for both ways.

pelwell: It's just a documentation change. The overlay handling code supports several string-to-integer conversions:
* `off`, `no`, `n`, `false` -> 0
* `on`, `yes`, `y`, `true` , `down` -> 1
* `up` -> 2

rlcamp: Thanks, updated documentation (and tested it both ways for good measure).

PR #5158, opened 2022-09-01
Add support for AT30TS74 into hwmon

rosterloh:
Atmel (now Microchip) AT30TS74 is an LM75 compatible sensor. Add it.

Signed-off-by: Richard Osterloh <richard.osterloh@gmail.com>

### DIFF
--- a/drivers/hwmon/Kconfig
+++ b/drivers/hwmon/Kconfig
@@ -1222,6 +1222,7 @@ config SENSORS_LM75
 	  temperature sensor chip, with models including:
 
 		- Analog Devices ADT75
+		- Atmel (now Microchip) AT30TS74
 		- Dallas Semiconductor DS75, DS1775 and DS7505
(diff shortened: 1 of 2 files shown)


pelwell:
My response to this PR was going to be "this is an unmodified upstream file - please submit the patch upstream", but then I looked at linux-next and found that support for the AT30TS74 has already been added. In fact, your PR is actually a verbatim (character for character identical) clone of upstream commit c851b715d38de0, which is already in rpi-5.19.y and rpi-6.0.y, except that you've changed the commit message and signed it off yourself.

We're (almost) always happy to back-port commits that have been accepted into newer kernels because it means that any maintenance overhead of the downstream patch is short-lived.

See a59a3781987e1e6331a88df8d7abb1a1a9524e66.

rosterloh:
Perfect, thank you so much @pelwell

PR #5164, opened 2022-09-05
imx290 fixes

6by9:
Minor issues noted from https://github.com/raspberrypi/libcamera-apps/issues/344

### DIFF
--- a/drivers/media/i2c/imx290.c
+++ b/drivers/media/i2c/imx290.c
@@ -927,12 +927,6 @@ static int imx290_set_fmt(struct v4l2_subdev *sd,
 						 IMX290_VMAX_MAX - mode->height,
 						 1,
 						 mode->vmax - mode->height);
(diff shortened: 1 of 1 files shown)


naushir:
Looks like a reasonable set of changes. Not tested it myself though.

6by9:
I've just been running with these, and if horizontal_blanking is increased to the max of 63615, then I can get just over 7 second exposures.  I've not checked with the exposure pattern generator, but the frame timings certainly all fit with the expected values.

pelwell:
I can't parse the second commit message, but otherwise it look reasonable.

6by9:
> I can't parse the second commit message, but otherwise it look reasonable.

I think I've lost a line. I'll update and repush with it making sense!

6by9:
Branch updated. Hopefully the commit text makes more sense now.

PR #5173, opened 2022-09-13
usb: xhci: expand mitigations for VLI_SS_BULK_OUT_BUG quirk

P33M:
After several months back-and-forth with VIA, we have a candidate root-cause for #4844 and a suggested fix. My known-bad pendrive now gets written to endlessly without suffering data corruption.

### DIFF
--- a/drivers/usb/host/xhci-mem.c
+++ b/drivers/usb/host/xhci-mem.c
@@ -1437,6 +1437,7 @@ int xhci_endpoint_init(struct xhci_hcd *xhci,
 	unsigned int ep_index;
 	struct xhci_ep_ctx *ep_ctx;
 	struct xhci_ring *ep_ring;
+	struct usb_interface_cache *intfc;
 	unsigned int max_packet;
 	enum xhci_ring_type ring_type;
 	u32 max_esit_payload;
@@ -1446,6 +1447,8 @@ int xhci_endpoint_init(struct xhci_hcd *xhci,
 	unsigned int mult;
 	unsigned int avg_trb_len;
 	unsigned int err_count = 0;
+	unsigned int is_ums_dev = 0;
+	unsigned int i;
 
 	ep_index = xhci_get_endpoint_index(&ep->desc);
 	ep_ctx = xhci_get_ep_ctx(xhci, virt_dev->in_ctx, ep_index);
@@ -1477,8 +1480,34 @@ int xhci_endpoint_init(struct xhci_hcd *xhci,
 
 	mult = xhci_get_endpoint_mult(udev, ep);
 	max_packet = usb_endpoint_maxp(&ep->desc);
-	max_burst = xhci_get_endpoint_max_burst(udev, ep);
 	avg_trb_len = max_esit_payload;
+
+	/*
+	 * VL805 errata - Bulk OUT bursts to superspeed mass-storage
+	 * devices behind hub ports can cause data corruption with
+	 * non-wMaxPacket-multiple transfers.
+	 */
+	for (i = 0; i < udev->config->desc.bNumInterfaces; i++) {
+		intfc = udev->config->intf_cache[i];
+		/*
+		 * Slight hack - look at interface altsetting 0, which
+		 * should be the UMS bulk-only interface. If the class
+		 * matches, then we disable out bursts for all OUT
+		 * endpoints because endpoint assignments may change
+		 * between alternate settings.
+		 */
+		if (intfc->altsetting[0].desc.bInterfaceClass ==
+		    USB_CLASS_MASS_STORAGE) {
+			is_ums_dev = 1;
+			break;
+		}
+	}
+	if (xhci->quirks & XHCI_VLI_SS_BULK_OUT_BUG &&
+	    usb_endpoint_is_bulk_out(&ep->desc) && is_ums_dev &&
+	    udev->route)
+		max_burst = 0;
+	else
+		max_burst = xhci_get_endpoint_max_burst(udev, ep);
 
 	/* FIXME dig Mult and streams info out of ep companion desc */
 


drivers/usb/host/xhci-mem.c
```diff
@@ -1477,9 +1480,33 @@ int xhci_endpoint_init(struct xhci_hcd *xhci,
...
+		/*
+		 * Slight hack - look at interface altsetting 0, which
+		 * should be the UMS bulk-only interface. If the class
+		 * matches, then we disable out bursts for all OUT
+		 * endpoints because endpoint assignments may change
+		 * between alternate settings.
+		 */
+		if (intfc->altsetting[0].desc.bInterfaceClass ==
+		    USB_CLASS_MASS_STORAGE)
+			is_ums_dev = 1;
```
pelwell, line 1501: Why not `break;` here? It would be to prevent someone going through the same thought process as me more than for optimisation.

P33M: Yes, the search need not be exhaustive - a single hit is all that's needed.

drivers/usb/host/xhci-mem.c
```diff
@@ -1477,9 +1480,35 @@ int xhci_endpoint_init(struct xhci_hcd *xhci,
...
+		intfc = udev->config->intf_cache[i];
+		/*
+		 * Slight hack - look at interface altsetting 0, which
+		 * should be the UMS bulk-only interface. If the class
+		 * matches, then we disable out bursts for all OUT
+		 * endpoints because endpoint assignments may change
+		 * between alternate settings.
+		 */
+		if (intfc->altsetting[0].desc.bInterfaceClass ==
+		    USB_CLASS_MASS_STORAGE) {
```
hselasky, line 1500: It doesn't make sense this only applies to mass storage!

jglathe: Sorta. Minimum tweak, though. The actual fix needs to be in the VL805 firmware, I guess. 
Another "fun" fact: I am following and testing this out of curiosity. On an RPi 4B 1.5 with 8GB RAM this works fine (as it seems, with an Argon 40 m.2 SATA enclosure and 174c:1156 ASMedia adapter). On a CM4 I/O board with WaveShare USB3.2 extension card (also VL805, but apparrently other firmware), I don't get the corruption, but uas sporadic errors with device resets. Need to disable uas for the external drives (JMicron and UGreen adapters). This also slows things down a bit.

hselasky: By chance, could you run the same tests under FreeBSD? We have a different USB stack and XHCI host controller!
We have some pre-built images here, which you can easily boot the RPI4 from:
https://download.freebsd.org/ftp/snapshots/arm64/aarch64/ISO-IMAGES/13.1/

P33M: > It doesn't make sense this only applies to mass storage!

Do you have an example of a superspeed device that is not a mass-storage class device which is a victim of this bug? Two factors are required - a driver that enqueues a vulnerable sequence of TRBs, and a device that both supports bursts and frequently signals flow control on its OUT endpoints.

jglathe: Well, I did - still searching for my micro HDMI to HDMI cable... anyway. I didn't come that far, just trying to mount the ntfs volume fails with command errors. wich is sorta consistent with VL805 firmware.

debian_uas_errors
freebsd_trymount

That the JMicron adapter throws errors with uas is well known. Couldn't mount the BTRFS UGreen device on FreeBSD. Anyway, the UGreen device is mounted and used as uas without errors on another RPi 4B 1.5 8GB.

grafik

since I have a FreeBSD SD card and a setup with keyboard/screen now, I can do other tests. It looks to me like this actually is a firmware issue with the VL805.

dobo90: Isn't this workaround related also to the wireless cards? I've noticed that a mt76 driver maintainer has suggested usage of @P33M's workaround because of Mediatek MT7612U random failures ([link](https://github.com/openwrt/mt76/issues/405#issuecomment-1066549535)). AFAIR those issues were mostly found on Raspberry Pi 4.
I've noticed similar issues on my setup (RPi4 + MT7612U). Disabling scatter-gatter (_disable_usb_sg_ parameter in _mt76_usb_ module) on mt76 seems to drop occurrence of those failures but in my case it didn't solve it ultimately (crashes occurred like once a week).

jglathe: This would be the one case you need to disable bursts for all burst-capable devices behind a hub. Bad news.

hselasky: Is there a descriptor which tells how many bursts a given USB 3.0 HUB supports? I'm trying to look in the USB specification.

P33M: @dobo90 if you think that the failure is related, then please create a new issue. Try building a custom kernel from this repo with the above interface class check disabled, and see if your wireless dongle still crashes.

https://www.raspberrypi.com/documentation/computers/linux_kernel.html#building-the-kernel

P33M: Hubs don't need to know protocol layer details about the packets they are forwarding. They are store-and-forward, and manage link state and packet credit between upstream and downstream ports. A data burst simply refers to the number of outstanding DPs that a transfer source can post before it has to receive an ACK from the destination, and a "burst" can continue indefinitely if there is no other pending endpoint activity. The rules for starting and ending a burst are only functions of the controller and the device.

Edit: one incidental way in which hubs can influence bursts is the number of header credits they advertise, which could cause bursts to be paced slower on a link that has a reduced buffer capacity. I've not seen a hub advertise fewer than 4 though, which is the maximum.

hselasky: It's called bMaxBurst, but it's not defined for HUBs (yet).

leezu: @dobo90 If I understand the current fix correctly, it forces `max_burst = 0` if a USB_CLASS_MASS_STORAGE device is present on a hub (which should include the root_hub, and thus all usb devices plugged into the RPi4?). I'm able to reproduce the firmware hangs on CF-953AX (mt7921au) on kernel 6.0.0-rc7-v8+ which includes [the commit](https://github.com/raspberrypi/linux/commit/3157603c3925a30c6b26dc8f7a1f2c23f7b7bb55), both in the case when a USB_CLASS_MASS_STORAGE device is present during boot and remains present, as well as if no USB_CLASS_MASS_STORAGE device is present during boot or at any time. Thus the fix does not appear to help with the wireless card hangs. I've opened a new issue to track it https://github.com/raspberrypi/linux/issues/5193 containing more details as well as a general issue regarding "USB disconnect causes failure on other USB devices" https://github.com/raspberrypi/linux/issues/5192

P33M: Having a discussion on a merge request line comment isn't how github is supposed to be used - if you have a bug that is related to this commit, then please create a new issue, fill out the template, and refer to this with a link.

PR #5175, opened 2022-09-19
Backport of HDMI Hotplug Rework 

mripard:
Hi

These patches will be part of 6.1. They fix an issue we currently have in the driver, even though no one complained so far.

Indeed, when we detect that a cable has been plugged into the HDMI connector, we enable the scrambler again if the mode we had programmed requires. This breaks the spec though, since it requires the scrambler to be enabled in a fairly short time frame after the video has been started.

The solution that was suggested was to do an empty mode set on the HDMI connector that will disable and enable the HDMI link.


pelwell:
Thanks!

PR #5181, opened 2022-09-25
ARM: dts: Add dtparams to disable PCIe, HDMI, and SD card/eMMC

0dvictor:
Under certain situations, it is desired to disable some unused devices.
This patch introduces three parameters allowing disabling PCIe, HDMI, and SD card (or eMMC on CM4).

### DIFF
--- a/arch/arm/boot/dts/bcm2708-rpi.dtsi
+++ b/arch/arm/boot/dts/bcm2708-rpi.dtsi
@@ -14,8 +14,10 @@
 	};
 
 	__overrides__ {
+		hdmi = <&hdmi>,"status";
 		i2c2_iknowwhatimdoing = <&i2c2>,"status";
 		i2c2_baudrate = <&i2c2>,"clock-frequency:0";
+		sd = <&sdhost>,"status";
 		sd_poll_once = <&sdhost>,"non-removable?";
 	};
 };
--- a/arch/arm/boot/dts/bcm2711-rpi-ds.dtsi
+++ b/arch/arm/boot/dts/bcm2711-rpi-ds.dtsi
@@ -4,6 +4,10 @@
 / {
 	__overrides__ {
 		arm_freq;
+		hdmi = <&hdmi0>,"status",
+		       <&hdmi1>,"status";
+		pcie = <&pcie0>,"status";
+		sd = <&emmc2>,"status";
 	};
 
 	v3dbus: v3dbus {
--- a/arch/arm/boot/dts/overlays/README
+++ b/arch/arm/boot/dts/overlays/README
@@ -190,6 +190,9 @@ Params:
                                 to negotiate. Legal values are 10, 100 and
                                 1000 (default 1000). Pi3B+ only.
 
+        hdmi                    Set to "off" to disable the HDMI interface
+                                (default "on")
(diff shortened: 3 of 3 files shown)


pelwell:
These overlays are Pi4/BCM2711-specific, so critiquing them as they are I would require the `compatible` strings to be changed to `"brcm,bcm2711"`. However, I'd rather consider an alternative implementation, having been bitten by compatibility issues before.

The required implementation may be board-specific, but the concepts are common. All Pis have some form of HDMI output capability, so why shouldn't you be able to disable HDMI on a 3B? But overlays don't know what they are being applied to, so they can't adapt (the correct label on Pi Zero-3 would be `hdmi`, not `hdmi0` and `hdmi1`).

The appropriate solution for small, board-specific changes such as these is to build them into the base dts files as parameters - they are sufficiently simple that no fragments are needed. For 2711-based devices your overlays translate to:
```
__overrides__ {
        emmc = <&emmc2>, "status";
        hdmi = <&hdmi0>, "status", <&hdmi1>, "status";
        pcie = <&pcie0>, "status";
};
```
You would use them by adding `dtparam=emmc=off` etc. to `config.txt`.

I would add them to the top of `bcm2711-rpi-ds.dtsi`, which is used . For the Pi 0-3 equivalents I would put them in `bcm2708-rpi.dtsi` it is included by all Pis except those that 2711-based.

There's a small matter of the best name for the `emmc` parameter - on most Pis that is used to drive the SD card, so `emmc` might be confusing or harder to discover. You could add `sd` as an alias, or perhaps name it `sdhci` instead.

What do you think?

0dvictor:
Good suggestions. Let me update this commit accordingly.

pelwell:
The implementation and README update look good. However, I'm not sure about the parameter name `sd_host` - is it too close to `sdhost`, which is not used on Pi 4? Any opinions, @popcornmix and @6by9?

0dvictor:
Indeed, I am also struggling to name it. I initially named it as `emmc` because I'm using a CM4 with eMMC, and now I realized that CM4 is the abnormality. `sdhci` might not be a good name either, since [it is the SDIO for wireless](https://github.com/raspberrypi/linux/blob/22f046f2b421f1753f87f0dfe8c252e41c63dadf/arch/arm/boot/dts/bcm2711-rpi-4-b.dts#L181).

For the Pi 0-3, it seems no-brainer to name it as `sd_host` since the DT node is named as `sdhost`. However, on BCM2711-based Pis (Pi 4, CM4, etc.), `sdhost` is `mmc2` and looks like [an unused SD host](https://github.com/raspberrypi/linux/blob/22f046f2b421f1753f87f0dfe8c252e41c63dadf/arch/arm/boot/dts/bcm2711-rpi-4-b.dts#L508). So I hoped that might be okay.

Maybe I'm overthinking on this, and we should simply name it as `sd`? Because the goal of this `dtparam` is simply disable whatever driver of the SD card?

popcornmix:
`sd` or `sdcard` seem okay to me.

0dvictor:
Nice, let me rename it to `sd`.

pelwell:
Thanks - I'm happy with that.

PR #5183, opened 2022-09-27
ARM: dts: Fix missing rpi CM4 serial console

juergh:
Without commandline argument '8250.nr_uarts=1' there is no 8250 UART and hence no serial console on the CM4. Add it back.

Fixes: b9e62329e096 ("ARM: dts: Adopt the upstream snd_bcm2835 handling")
Signed-off-by: Juerg Haefliger <juerg.haefliger@canonical.com>


pelwell:
Thanks!

PR #5187, opened 2022-09-28
WIP: Switch from downstream ov9281 driver to mainline ov9282

6by9:
Needed a load of fixups to the mainline driver (which I'm about to upstream).
Partly creating the PR so I can review it more easily, hence WIP tag for now.


pelwell:
These commit message changes would make them easier to read:
* media: i2c: ov9282: Correct HTS register for configured pixel rate
    * "as is" -> "as-is"
* media: i2c: ov9282: Make V4L2_CID_HBLANK r/w
    * "read only" -> "read-only"

Otherwise it all looked fine to me.

6by9:
Branch updated with the commit text changes suggested. 
Needs a smidge  more testing, and then sending upstream, then I'll remove the WIP tag.

6by9:
Updated based on v2 patches sent to linux-media.

6by9:
All patches are in mainline 6.2, and #5304 switches rpi-6.2.y to use it.

Closing as backporting to 5.15 has limited gain. It may be worth backporting to 6.1 assuming it is the next LTS (when finally announced).

PR #5196, opened 2022-10-03
Rpi 5.19/bcm2835 audio fixes

juergh:
On 5.19, HDMI audio outputs are always disabled. This PR fixes that plus some other minor issues.

### DIFF
--- a/drivers/staging/vc04_services/bcm2835-audio/bcm2835.c
+++ b/drivers/staging/vc04_services/bcm2835-audio/bcm2835.c
@@ -299,21 +299,28 @@ static void set_hdmi_enables(struct device *dev)
 		of_node_put(firmware_node);
 	}
 
-	if (!firmware)
+	if (!firmware) {
+		dev_err(dev, "Failed to get fw structure\n");
 		return;
+	}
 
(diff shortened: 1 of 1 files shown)


popcornmix:
This looks good to me. @pelwell ?

drivers/staging/vc04_services/bcm2835-audio/bcm2835.c
```diff
@@ -360,10 +360,12 @@ static int snd_bcm2835_alsa_probe(struct platform_device *pdev)
 	    !of_property_read_bool(dev->of_node, "brcm,disable-hdmi"))
 		set_hdmi_enables(dev);
 
-	of_property_read_u32(dev->of_node,
-			     "brcm,disable-headphones",
-			     &disable_headphones);
-	enable_headphones = !disable_headphones;
+	if (enable_headphones) {
```
pelwell, line 363: Is this better than the smaller change of `enable_headphones &= !disable_headphones;`?

juergh: I opted for the proposed implementation to save a DT property read if enable_headphones == 0.

PR #5197, opened 2022-10-03
bcm2835-codec: Allow encoder cmds on ISP resize and deinterlace

6by9:
@jc-kynesim Hopefully this pair will do the trick for you.


jc-kynesim:
Looks good to me.
Ta

PR #5198, opened 2022-10-05
media: i2c: imx477: Reset hblank on mode switch

naushir:
Reset the hblank control to the minimum value on every mode switch. This is to account for userland instances that do not yet control hblank, otherwise it gets set to a non-optimal value.


6by9:
I'm removing patches like this from other drivers. 
Or at least insert a 
```if (imx477->mode != mode)```
at https://github.com/raspberrypi/linux/blob/e3c6ee4b0197deeee200e3c53c3e460c64538760/drivers/media/i2c/imx477.c#L1631 so that reselecting the same mode doesn't reset it.

naushir:
Yes, makes sense!

6by9:
LGTM

Issue #5200, opened 2022-10-06
Artefact in IMX462 camera frames after rpi-update


sandyol55:
### Describe the bug

When in the 1280 : 720 : 12bit mode the sensor outputs a noisy coloured 'overlay' over the rightmost 32 pixels. Colour seems to vary depending on the content of main image.
Does not happen in any of the three other sensor modes. The artefact appears in preview, captured stills, or video,  and in both libcamera-apps and Picamera2 applications, (so presumed to be a low level change perhaps associated with recent IMX290 driver modifications???) 

### Steps to reproduce the behaviour

IMX462 board directly driven by 2G Pi4b via short (150mm) cable.
Using imx290 dtoverlay.
A simple libcamera-hello is enough to demonstrate  the issue, as it opens in the 1280:720:12bit mode, the only mode that seems to exhibit the artefact. 
Artefact appeared after recent rpi-updates to both .70 and .72 kernels.
Reverting to 5.15.61-v7l+ reverts to normal behaviour.

### Device (s)

Raspberry Pi 4 Mod. B

### System

OS 32bit Bullseye 

$ vcgencmd version
Oct  5 2022 15:51:11 
Copyright (c) 2012 Broadcom
version 6c57d97fa8066929a2fee6e4fac6d64ad62506c9 (clean) (release) (start)

$ uname -a
Linux raspberrypi 5.15.72-v7l+ #1591 SMP Wed Oct 5 12:05:33 BST 2022 armv7l GNU/Linux

6by9:
I've reproduced this with an Arducam IMX462 (UC-755). ```dtoverlay=imx290``` is all that I've applied for configuration as the oscillator is 37.125MHz on that board.

It's an odd error - it looks like the CSI side of the sensor is running too slow to keep up with the sensor array. The numbers look to match the datasheet though, so it needs some more investigation.

6by9:
Found.
In 720p mode the CSI link is running slower, and therefore the HMAX value (width + HBLANK) needs to be increased to 3300 rather than being left at 2200 (the same ratio as the 891 vs 594Mbit/s/lane speed).
The latest patches had set blanking based on a constant HMAX of 2200.

[...]

PR #5204, opened 2022-10-10
media: i2c: imx290: Correct min HBLANK.

6by9:
In the 720p mode the CSI link is run at a lower frequency, and the minimum HBLANK value has to be increased to avoid generating more data from the sensor than the link can carry.

Set the minimum based on the mode.

See #5200


pelwell:
Nice - here's to deletion as a coding technique.

PR #5209, opened 2022-10-17
drivers/gpu/drm/vc4: Add missing 32-bit RGB formats

davidplowman:
The missing 32-bit per pixel ABGR and various "RGB with an X value" formats are added. Change sent by Dave Stevenson.

### DIFF
--- a/drivers/gpu/drm/vc4/vc4_firmware_kms.c
+++ b/drivers/gpu/drm/vc4/vc4_firmware_kms.c
@@ -181,18 +181,22 @@ static const struct vc_image_format {
 		.drm = DRM_FORMAT_ARGB8888,
 		.vc_image = VC_IMAGE_ARGB8888,
 	},
-/*
- *	FIXME: Need to resolve which DRM format goes to which vc_image format
- *	for the remaining RGBA and RGBX formats.
- *	{
- *		.drm = DRM_FORMAT_ABGR8888,
- *		.vc_image = VC_IMAGE_RGBA8888,
- *	},
- *	{
- *		.drm = DRM_FORMAT_XBGR8888,
- *		.vc_image = VC_IMAGE_RGBA8888,
- *	},
- */
+	{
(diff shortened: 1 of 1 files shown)


davidplowman:
@6by9 in case you don't get notified anyway!

6by9:
Looks a familiar diff.

Are we happy that we don't need DRM_FORMAT_RGBA8888 and DRM_FORMAT_BGRA8888? Adding them is a faff.

davidplowman:
Picamera2 seems to work with this as it stands so I'm happy, and wouldn't want to cause any further work  on account of fkms!

6by9:
LGTM then.

drivers/gpu/drm/vc4/vc4_firmware_kms.c
```diff
@@ -181,18 +181,22 @@ static const struct vc_image_format {
...
- */
+	{
+		.drm = DRM_FORMAT_XBGR8888,
+		.vc_image = VC_IMAGE_RGBX32,
+	},
+	{
+		.drm = DRM_FORMAT_ABGR8888,
+		.vc_image = VC_IMAGE_RGBA32,
+	},
+	{
```
pelwell, line 192: The last two don't follow the same pattern as the first two:
* XBGR <-> RGBX
* ABGR <-> RGBA
* RGBX <-> BGRX (not XRGB)
* BGRX <-> RGBX (not XRGB)

 Where does the inconsistency come from?

6by9: > Where does the inconsistency come from?

VC_IMAGE_TYPE_T. The formats are all over the place in name.
https://github.com/raspberrypi/userland/blob/master/interface/vctypes/vc_image_types.h#L58
The comments are testament to just how messed up formats are.

I worked from the HVS format definitions in scalerlib, mapped it back through VC_IMAGE_TYPE_T, and then to DRM_FORMAT_T. Even then I got it wrong on first attempt.

PR #5212, opened 2022-10-18
ARM: dts: Disable uart1 in the CM4 dts

iluminat23:
It is a wrong assumption that a CM4 based device usese uart1. As the comment in the dts says the uart1 is enabled, because it is mapped to the pin header. This is not a valid assumption for the CM4. It might do more harm then good to eneable uart1. Even more that the BCM2711 has more uarts which better features then the mini uart.

Signed-off-by: Philipp Rosenberger <p.rosenberger@kunbus.com>

### DIFF
--- a/arch/arm/boot/dts/bcm2711-rpi-cm4.dts
+++ b/arch/arm/boot/dts/bcm2711-rpi-cm4.dts
@@ -254,13 +254,6 @@
 	};
 };
 
-/* uart1 is mapped to the pin header */
-&uart1 {
-	pinctrl-names = "default";
-	pinctrl-0 = <&uart1_gpio14>;
-	status = "okay";
-};
-
 &vchiq {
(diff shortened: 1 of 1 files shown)


pelwell:
Have you looked at the state of uart1 (a.k.a. ttyS0) when `enable_uart` is not set (or set to 0)?
```
$ ls -l /dev/ttyS0
ls: cannot access '/dev/ttyS0': No such file or directory
$ raspi-gpio get 14-15
GPIO 14: level=1 fsel=0 func=INPUT pull=NONE
GPIO 15: level=1 fsel=0 func=INPUT pull=UP
```
Although the .dtb has the node enabled, the firmware disables it if it isn't wanted.

The CM4 is no different to any other Pi in that regard (except that you need to substitute uart0 and ttyAMA0 on non-Bluetooth models).

PR #5215, opened 2022-10-19
Emc2305 updates

pelwell:
This is an attempt to forward-port 1100b06c10681. It wasn't the easiest of ports, hence the PR. In particular, the intention for the `pwm-channel` attribute isn't 100% clear - should it be replaced by a `pwm-separate` flag?


6by9:
It looks like it.
Memory says that it was a route to choose whether all fans adopted the same settings, or acted as individuals.

I'll have a look tomorrow at the changes made with the mainline driver, and update the overlay or driver appropriately.

6by9:
Just looking at the comments when it was submitted to mainline
https://lore.kernel.org/all/20220810171552.56417-3-michaelsh@nvidia.com/
v4->v5
Changes added by Michael Shych:
**- Remove OF interface;**
v2->v3
Change pointed out by Guenter Roeck:
- Change casting from const pointer;
**- Remove pwm_max per channel parameter.**
Changes added by Michael Shych:
**- Add pwm_separate field instead using of a special case of pwm_channel;**
- Change the mechanism of obtaining a number of pwm channels.

Well who would need OF?!

We've also lost the block
```
	/* Acknowledge any existing faults. Stops the device responding on the
	 * SMBus alert address.
	 */
	i2c_smbus_read_byte_data(client, EMC2305_REG_FAN_STALL_STATUS);
	i2c_smbus_read_byte_data(client, EMC2305_REG_FAN_STATUS);
```
from the end of probe.

I've got hardware to test this, so will take your patch, add back the bits we need, and create a new PR.

6by9:
Ah, Rob Herring read the riot act over the binding not being generic - https://patchwork.kernel.org/project/linux-hwmon/patch/20220623165217.59252-3-michaelsh@nvidia.com/

pelwell:
Superseded by #5218.

PR #5218, opened 2022-10-20
hwmon: emc2305: fixups for driver submitted to mailing lists

6by9:
The driver had a number of issues, checkpatch warnings/errors, and other limitations, so fix these up to make it usable.

Fixed up version of #5215.


pelwell:
Thanks!

PR #5220, opened 2022-10-21
Increase number of DRM planes supported per device

6by9:
DRM had a somewhat arbitrary limit of 32 planes per device due to internally using a u32 for a bitmask.

HVS can support more planes than that, so switch to a u64 and increase the limit to 64.

@mripard thoughts?
I don't see the bitmask being exported to uapi, therefore this shouldn't cause grief.

### DIFF
--- a/drivers/gpu/drm/drm_atomic.c
+++ b/drivers/gpu/drm/drm_atomic.c
@@ -451,7 +451,7 @@ static void drm_atomic_crtc_print_state(struct drm_printer *p,
 	drm_printf(p, "\tactive_changed=%d\n", state->active_changed);
 	drm_printf(p, "\tconnectors_changed=%d\n", state->connectors_changed);
 	drm_printf(p, "\tcolor_mgmt_changed=%d\n", state->color_mgmt_changed);
-	drm_printf(p, "\tplane_mask=%x\n", state->plane_mask);
+	drm_printf(p, "\tplane_mask=%llx\n", state->plane_mask);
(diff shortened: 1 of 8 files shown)


6by9:
Only against 6.1 as that was the branch I happened to be on here.

mripard:
Sorry for the belated answer.

I've reviewed the UAPI, and I think you're right, it doesn't look like the plane mask is exported to the userspace. Only the plane ID is, but it's going to be far lower than the maximum value for u64.

6by9:
Is it worth sending this upstream? With wlroots starting to use multiple planes when available, 32 planes does become a limitation.

I guess the other thing to look at is abstracting it out so that it is only internal to drm core. The size could then be totally arbitrary. One for another day.

mripard:
We can definitely try to send it, yeah. Worst case scenario is that it's rejected for some reason :)

pelwell:
Is this PR ready for merging? It at least needs to be rebased.

6by9:
> Is this PR ready for merging? It at least needs to be rebased.

Let's see what mainline think to it.
If they reject it then it will also direct our efforts with regard wayfire and punting planes to DRM for composition.

6by9:
(Sorry for the churn on here - I pushed to the wrong branch. This is now rebased, but I still haven't sent it upstream).

6by9:
Closing as 6.1 is definitely superceded, and being investigated with #6319

PR #5223, opened 2022-10-25
drivers: usb: dwc_otg: fix incorrect reference passing when checking bandwidth

P33M:
This was an obvious bug (since January 2012), but I'm a bit skeptical that a wild read could cause lockups. Normally you get some sort of kernel splat...

### DIFF
--- a/drivers/usb/host/dwc_otg/dwc_otg_hcd_linux.c
+++ b/drivers/usb/host/dwc_otg/dwc_otg_hcd_linux.c
@@ -807,7 +807,6 @@ static int dwc_otg_urb_enqueue(struct usb_hcd *hcd,
 	struct usb_host_endpoint *ep = urb->ep;
 #endif
 	dwc_irqflags_t irqflags;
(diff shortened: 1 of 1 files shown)


popcornmix:
@HiassofT does this resolve your LE issue?

pelwell:
How considered is the use of `ep->hcpriv` over `*ref_ep_hcpriv` (as used later on in the same function)?

P33M:
[s]Oh dear. That's also broken - `ep->hcpriv` is not a `struct urb *`.[/s] - urgh, what a mess - function calls nested inside function calls.

pelwell:
Merging so that this might get some wider testing, on the basis that it looks more consistent and rpi-6.0.y is a very much a beta branch.

HiassofT:
Thanks for the fix, KASAN error on 64bit kernels is now gone but unfortunately the FIQ / stackprotector per task issue is still present. I'll do some more tests and then post on the issue

PR #5224, opened 2022-10-25
configs: enable pressure stall information

### LINKED ISSUE #4683: PSI support
**Describe the bug**
``/cat/proc/pressure/*`` is not available using the latest raspberrypi-kernel package for Raspbian bullseye. Tools like oomd can't be used.

**Expected behaviour**
Linux kernel from 4.20 and up include PSI enabled, it's up to the maintainers of the distro to disable it. It should be built into the kernel (I tried setting ``psi=1`` in cmdline.txt, but it`s still not available).

**Actual behaviour**
PSI can't be used without building the kernel.

**System**
 Copy and paste the results of the raspinfo command in to this section. Alternatively, copy and paste a pastebin link, or add answers to the following questions:

* Which model of Raspberry Pi? Pi 4B+ 4GB
* Which OS and version (`cat /etc/rpi-issue`)?
```
Raspberry Pi reference 2018-11-13
Generated using pi-gen, https://github.com/RPi-Distro/pi-gen, 7e0c786c641ba15990b5662f092c106beed40c9f, stage2
```
* Which firmware version (`vcgencmd version`)?
```
Oct 29 2021 10:47:49
Copyright (c) 2012 Broadcom
version b8a114e5a9877e91ca8f26d1a5ce904b2ad3cf13 (clean) (release) (start_cd)
```
* Which kernel version (`uname -a`)?
```
Linux ferferga-server 5.10.63-v7l+ #1459 SMP Wed Oct 6 16:41:57 BST 2021 armv7l GNU/Linux
```

ChemicalXandco:
Fixes #4683 

This is necessary for waydroid (android container) to work.


pelwell:
On how many of these platforms do you think somebody might realistically be running Waydroid? Is BCM2835 (bcmrpi_defconfig) a suitable platform? There's no point in enabling the feature and forcing people to pay the penalty of a larger kernel if they won't be using it.

ChemicalXandco:
I think it is quite reasonable to run waydroid on those older platforms and the penalty of PSI is negligible. I also think it makes sense to maintain feature parity across all of them in that regard. Do you agree with my assessment? I don't mind either way since I only use BCM2711 for this.

ferferga:
@pelwell Any updates about this PR, whether it's approved or not?

pelwell:
If it's not been merged then it's not been approved. This is one of those cases where it takes significant effort to evaluate the effects of enabling a feature for our kernel builds. If the feature benefits all users then it's a no-brainer, but if it benefits a few but costs everybody (including us) then it is likely to be overtaken by other demands on our time.

ZanderBrown:
As mentioned in the other issue, it also allows the likes of systemd-oomd to be enabled, which could help reduce lockups due to high swapping even in regular (non-waydroid) setups

In fact, it also enables [GMemoryMonitor](https://docs.gtk.org/gio/iface.MemoryMonitor.html) (and any other low-memory-monitor consumers) so that applications can free caches or similar in low-memory scenarios, so possibly even avoid the need for OOM in the first place

pelwell:
Enabling pressure stall statistics adds 17kB to a Pi 1 kernel and 67kB to a 64-bit kernel, neither of which is excessive. The developers of the PSI feature claim that the overhead of enabling it is too small to measure except in artificial benchmarks, but despite that they include the option (CONFIG_PSI_DEFAULT_DISABLED) to leave it disabled unless explicitly enabled with `psi=1` on the command line. Without this additional option, the reverse is also possible - to explicitly disable it on the command line (`psi=0`).

To me the overhead in kernel size and execution time is sufficiently low that the feature can be added to all kernels, but I'm leaning towards leaving it disabled by default. Any thoughts, @popcornmix?

popcornmix:
I'm of the same view. Happy to add the option, but perhaps disabled by default is best (assuming enabling it only benefits a very small percentage of users).

pelwell:
Added by fcfe018feab97279f946bb7fc9049c28268a1a8f.

Issue #5189, opened 2022-09-29
6.0 locks up on RPi2/3 if FIQ is enabled in dwc_otg and kernel is built with STACKPROTECTOR_PER_TASK 


HiassofT:
### Describe the bug

Trying to boot 32bit rpi-6.0.y kernel on RPi3B+ results in a hard lockup, usually during USB enumeration.

This seems to be FIQ related, adding `dwc_otg.fiq_enable=0 dwc_otg.fiq_fsm_enable=0` to cmdline.txt allows the kernel to boot.

Very likeley it's a crash somewhere in dwc_otg, but unfortunately serial logs stop after `8<--- cut here ---` so the details what exactly went wrong are missing - not much more on HDMI screein either in a couple of tries.

### Steps to reproduce the behaviour

* build 32bit rpi-6.0.y kernel with bcm2709_defconfig
* install on RPi3B+
* boot

### Device (s)

Raspberry Pi 3 Mod. B+

### System

I first noticed the lockup on LibreELEC, running 6.0-rc1 a while ago and now tested LE and RPiOS with current 6.0-rc7 kernel as well (all with the same results)

```
root@raspberrypi:~# cat /etc/rpi-issue
Raspberry Pi reference 2022-01-28
Generated using pi-gen, https://github.com/RPi-Distro/pi-gen, f01430c9d8f67a4b9719cc00e74a2079d3834d5d, stage2
root@raspberrypi:~# vcgencmd version
Aug 26 2022 14:04:36 
Copyright (c) 2012 Broadcom
version 102f1e848393c2112206fadffaaf86db04e98326 (clean) (release) (start)
root@raspberrypi:~# uname -a
Linux raspberrypi 6.0.0-rc7-v7+ #1 SMP Thu Sep 29 16:33:35 CEST 2022 armv7l GNU/Linux
```

Kernel built from commit  3fb5ca89c844eb9fda924a5e76ffab9e8c068675

### Logs

[rpi3b+.log.txt](https://github.com/raspberrypi/linux/files/9679100/rpi3b%2B.log.txt)

### Additional context

Most likely RPi0-3 are affected as well, haven't tested those though

lategoodbye:
@HiassofT Your error description doesn't fit exactly to this, but could you give it a try?
https://lore.kernel.org/lkml/20220929-rpi-pi3-unplugged-fixes-v1-0-cd22e962296c@cerno.tech/

pelwell:
[...]

PR #5226, opened 2022-10-27
stop GCC from patching FIQ functions

P33M:
Round 2 for https://github.com/raspberrypi/linux/issues/5189

Patching the option in the makefile rather than using explicit attributes preserves the remaining compile options.


HiassofT:
Thanks a lot, the change is working fine here!

Tested both with gcc 10 and plugin support on Debian Bullseye and gcc 12 in LE buildsystem.

PR #5227, opened 2022-10-28
WIP: EMC2301 fan driver updates from mainline

6by9:
The EMC2301 driver was merged in 6.1 and has a few differences from the earlier version that we took on our branches.

Revert the earlier version, backport the final mainline version, and then fix it up as they dropped DT support

https://forums.raspberrypi.com/viewtopic.php?t=342115


6by9:
Hold fire on this - there seems to be something not joining up as I've got my CM4 up to 83 deg and the fan controller is still not requesting a PWM drive to speed up the fan. I'll have to have a further look on Monday.

drivers/hwmon/emc2305.c
```diff
@@ -0,0 +1,624 @@
...
+	ret = i2c_smbus_read_word_swapped(client, EMC2305_REG_FAN_TACH(channel));
+	if (ret <= 0)
+		return ret;
+
+	ret = ret >> EMC2305_TACH_REGS_UNUSE_BITS;
+	ret = EMC2305_RPM_FACTOR / ret;
+	if (ret <= EMC2305_TACH_RANGE_MIN)
+		return 0;
+
+	return ret * EMC2305_TACH_CNT_MULTIPLIER;
```
pelwell, line 260: While you are revisiting this, lines 249 and 242 are at least confusing, given that the multiplier is 2.

6by9: Those lines are unmodified from the mainline driver, so do we really want to make too many mods for aesthetics?

Or is your issue that the min range is EMC2305_TACH_RANGE_MIN, but the comparison is done before the multiplication by EMC2305_TACH_CNT_MULTIPLIER, so ret is actually half the value it should be, and we're really returning 0 for any speed below EMC2305_TACH_RANGE_MIN * 2.

6by9:
It looks like the thermal DT nodes have all shuffled around between 5.15 and 6.1, so backporting the driver without updating all the relevant DT doesn't work.

Now trying to work out who changed the DT thermal nodes and why.

drivers/hwmon/emc2305.c
pelwell: Yes, the minimum non-zero value returned is 961. I must admit I'm doubtful about the whole notion of a minimum value specified for the controller in RPM when different fans can have different multipliers. It seems more natural to have a minimum in pulses per second, with RPM then calculated from that.

6by9:
Closing as this is against 5.15.
Issue with why it wasn't working found and fixed on 6.1 via #5440

PR #5228, opened 2022-10-29
Add support for mcp3004

DwfSw:
Add overlay support for the mcp3004. The new overlay includes the voltage reference to avoid the warning: "supply vref not found, using dummy regulator". I left the voltage regulator reference default to 3.3V

### DIFF
--- a/arch/arm/boot/dts/overlays/Makefile
+++ b/arch/arm/boot/dts/overlays/Makefile
@@ -139,6 +139,7 @@ dtbo-$(CONFIG_ARCH_BCM2835) += \
 	mcp2515-can0.dtbo \
 	mcp2515-can1.dtbo \
 	mcp251xfd.dtbo \
+	mcp3004.dtbo \
 	mcp3008.dtbo \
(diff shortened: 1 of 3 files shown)


pelwell:
I'm happy with the idea of mcp3004 overlay supporting all the SPI busses, but this is now an outdated implementation. It is possible to eliminate the duplication using some features of more recent firmwares. Look at the [maxtherm](https://github.com/raspberrypi/linux/blob/rpi-5.15.y/arch/arm/boot/dts/overlays/maxtherm-overlay.dts) overlay for an example of how it can be done.

Please also include a README entry.

DwfSw:
@pelwell thanks for the comments! Updated as suggested. Thanks.

pelwell:
P.S. I'm aware that this makes mcp3004 different to the other mcp300x overlays, but this is better. At some point we may add the new style of parameters to the existing overlays, or even move them all under a single umbrella overlay, similar to `i2c-sensor`.

Issue #5180, opened 2022-09-24
H.264 encoder corruption on repeatable image pattern


mdevaev:
### Describe the bug

The H.264 M2M encoder breaks down when trying to encode something containing a contrasting repeating (at least partially) pattern, for example, a lot of black text on a white background. Below is an example of the original image and the result in PiKVM.

**Source:**
![image](https://user-images.githubusercontent.com/416259/192109080-3abf598a-e9ce-4c6f-b985-9e24a002dc71.png)

**Encoded:**
![image](https://user-images.githubusercontent.com/416259/192109098-ef07778c-c458-4b7e-89f4-3264468bb609.png)

Encoding parameters:
```
V4L2_CID_MPEG_VIDEO_BITRATE = 5000k
V4L2_CID_MPEG_VIDEO_H264_I_PERIOD = 30
V4L2_CID_MPEG_VIDEO_H264_PROFILE = V4L2_MPEG_VIDEO_H264_PROFILE_CONSTRAINED_BASELINE
V4L2_CID_MPEG_VIDEO_H264_LEVEL = V4L2_MPEG_VIDEO_H264_LEVEL_4_0 // Also tried 5.1, no changes
V4L2_CID_MPEG_VIDEO_REPEAT_SEQ_HEADER = 1
V4L2_CID_MPEG_VIDEO_H264_MIN_QP = 16
V4L2_CID_MPEG_VIDEO_H264_MAX_QP = 32
```

On increasing the MIN_QP parameter to 20 or 24, the bug will be reproduced only if the screen is completely filled with text. Probably something breaks inside the encoder because the I-frame does not restore the image. Also corruption does not occur immediately, but after 10-20 seconds after showing the text pattern.

### Steps to reproduce the behaviour

Use TC358743 capture and  [https://github.com/pikvm/ustreamer](ustreamer). Open the first screenshot on the video source. Run ustreamer and record the video:
 ```
ustreamer --dv-timings --persistent --format=uyvy --verbose --h264-sink=test
(in the another console)
ustreamer-dump --sink=test -o test.h264
(Ctrl+C after a minute)
```
Then play file `test.h264`.

### Device (s)

Raspberry Pi 4 Mod. B

### System

`Arch Linux ARM`

```
Aug 26 2022 14:03:16
Copyright (c) 2012 Broadcom
version 102f1e848393c2112206fadffaaf86db04e98326 (clean) (release) (start)
```

[...]

PR #5232, opened 2022-10-31
vc04_services: bcm2835-codec: Increased H.264 buffer size

mdevaev:
When encoding an image containing a large amount of small text, larger buffers are required:
- 1MB for 720p and lower;
- 1.5MB for 1080p.

Signed-off-by: Maxim Devaev <mdevaev@gmail.com>

### DIFF
--- a/drivers/staging/vc04_services/bcm2835-codec/bcm2835-v4l2-codec.c
+++ b/drivers/staging/vc04_services/bcm2835-codec/bcm2835-v4l2-codec.c
@@ -138,11 +138,14 @@ static const char * const components[] = {
  * V4L2 mandates that the encoded frame must fit in a single buffer. Sizing
  * that buffer is a compromise between wasting memory and risking not fitting.
  * The 1080P version of Big Buck Bunny has some frames that exceed 512kB.
- * Adopt a moderately arbitrary split at 720P for switching between 512 and
- * 768kB buffers.
+ * However, if the frame is an image of a large amount of small text,
+ * the size of the encoded frame increases dramatically and can reach
+ * 750kB for 720p and 1.2MB for 1080p.
+ * Adopt a moderately arbitrary split at 720P for switching between 1MB and
+ * 1.5MB buffers.
  */
-#define DEF_COMP_BUF_SIZE_GREATER_720P	(768 << 10)
-#define DEF_COMP_BUF_SIZE_720P_OR_LESS	(512 << 10)
+#define DEF_COMP_BUF_SIZE_GREATER_720P	(1536 << 10)
+#define DEF_COMP_BUF_SIZE_720P_OR_LESS	(1024 << 10)
 /* JPEG image can be very large. For paranoid reasons 4MB is used */
 #define DEF_COMP_BUF_SIZE_JPEG (4096 << 10)
 


mdevaev:
Fixes (details are also here): #5180

6by9:
Userspace can fix this anyway
https://www.kernel.org/doc/html/latest/userspace-api/media/v4l/pixfmt-v4l2.html#single-planar-format-structure
<html><body>


__u32 | sizeimage | Size in bytes of the buffer to hold a complete image, set by the driver. Usually this is bytesperline times height. When the image consists of variable length compressed data this is the number of bytes required by the codec to support the worst-case compression scenario. The driver will set the value for uncompressed images. **_Clients are allowed to set the sizeimage field for variable length compressed data flagged with V4L2_FMT_FLAG_COMPRESSED at ioctl VIDIOC_ENUM_FMT_**, but the driver may ignore it and set the value itself, or it may modify the provided value based on alignment requirements or minimum/maximum size requirements. If the client wants to leave this to the driver, then it should set sizeimage to 0.
-- | -- | --



</body>
</html>

```V4L2_PIX_FMT_H264``` has the ```V4L2_FMT_FLAG_COMPRESSED``` flag set, and therefore you should be able to set ```sizeimage``` via ```VIDIOC_S_FMT```.
I believe that is all implemented.
https://github.com/raspberrypi/linux/blob/rpi-5.15.y/drivers/staging/vc04_services/bcm2835-codec/bcm2835-v4l2-codec.c#L1506-L1513

mdevaev:
It's also working, thank you. In this case I don't know if it's worth applying a patch, since the default values were suitable for almost all users. On the other hand, the difference in memory consumption is not too big. The question is how much it affects users with old Raspberry boards.

6by9:
The current defaults are valid for most use cases - it's only where bitrate control can't keep a handle on buffer sizes (probably because of overriding QP settings on detailed scenes) that they become excessive.

On a Pi0 with IIRC FFmpeg defaulting to allocating 20 buffers, allocating another 15MB of buffers is significant, so I really don't want to increase the default.

V4L2 not permitting fragmenting of compressed buffers is one of the biggest limitations I know of. Android certainly used to suffer from the same restriction.
There have been discussions at Linux Media summits about it, but not in a way that is actually functional for realtime pipelined encoders. Basically throw a big error (not nice to userspace), reencode the frame (not possible with realtime as your reference frame is probably already updated and you're time limited), or drop the frame (not possible as the reference frame has been updated). It's also an issue to decode as there is no one answer to how big a frame might be in a pre-encoded video clip. One of the Big Buck Bunny clips hits this as some frames are far bigger than you would expect for the bitrate.
 
Double checking the firmware source, there is ```MMAL_PARAMETER_VIDEO_ENCODE_FRAME_LIMIT_BITS```.
If set and a buffer exceeds the limit, then it appears that the encoder does abort the encode and discard the frame. I'm not 100% clear at what stage it happens, or what the codec then produces (V4L2 expects a buffer out even if the encode failed), but it might be worth a quick experiment.

mdevaev:
Thank you for the detailed information. I agree that this change doesn't make sense for all users, especially since we already have a fix in userspace.

PR #5233, opened 2022-10-31
WIP: add support for Lontium lt6911uxc HDMI-RX to MIPI CSI2-TX

nicolasb565:
Tested on a rpi cm4 with io-board. I had to remove chip U11 on the io-board.

I don't think the driver support libcamera, and the device tree is hardcoded to use CAM1.

Code based on https://github.com/InES-HPMM/Lontium_lt6911uxc/blob/main/source/lt6911uxc.c
See https://forums.raspberrypi.com/viewtopic.php?t=341781 for discussion.


6by9:
I'll try and review when I get a chance.

Adding any driver is normally 4 patches:
- device tree bindings
- driver
- defconfig changes
- overlay & README.

checkpatch checks that any compatible strings defined in the driver patch are documented in the bindings. Without the binding it will complain.

Only working on CAM1 is fine as I believe it requires 4 data lanes, and that is only present on CAM1. (CAM0 is 2 lanes only)
libcamera is only relevant for raw image sensors, and this device doesn't fall into that category.

arch/arm/boot/dts/overlays/Makefile
```diff
@@ -130,6 +130,8 @@ dtbo-$(CONFIG_ARCH_BCM2835) += \
 	justboom-dac.dtbo \
 	justboom-digi.dtbo \
 	ltc294x.dtbo \
+	lt6911uxc-overlay.dtbo \
```
pelwell: Don't include the `-overlay` here - there is a make rule to create `*.dtbo` from `*-overlay.dts`.

Also, both new entries belong before `ltc294x.dtbo'.

drivers/media/i2c/Makefile
```diff
@@ -115,6 +115,7 @@ obj-$(CONFIG_VIDEO_S5C73M3)	+= s5c73m3/
 obj-$(CONFIG_VIDEO_ADP1653)	+= adp1653.o
 obj-$(CONFIG_VIDEO_LM3560)	+= lm3560.o
 obj-$(CONFIG_VIDEO_LM3646)	+= lm3646.o
+obj-$(CONFIG_VIDEO_LT6911UXC) += lt6911uxc.o
```
pelwell: The other lines have a TAB character before the `+=`.

pelwell (commented):
Aside from the inline review comments you will need to write README entries for both overlays. Remember to include two blank lines between entries, to use spaces for indents, and to keep to the 80-character line length limit.

arch/arm/boot/dts/overlays/Makefile
nicolasb565: resolved in last commit

drivers/media/i2c/Makefile
nicolasb565: resolved in last commit

arch/arm/boot/dts/overlays/README
```diff
@@ -2424,6 +2424,21 @@ Info:   This overlay has been deprecated and removed - see gpio-ir
 Load:   <Deprecated>
 
 
+Name:   lt6911uxc
+Info:   Adds support for Lontium lt6911uxc HDMI to CSI
+Load:   dtoverlay=lt6911uxc,<param>=<val>
+Params: media-controller        Configure use of Media Controller API for
+                                configuring the sensor (default off)
```
pelwell: Remove the trailing whitespace, please.

arch/arm/boot/dts/overlays/README
```diff
@@ -2424,6 +2424,21 @@ Info:   This overlay has been deprecated and removed - see gpio-ir
...
+                                configuring the sensor (default off)    
+
+
+Name:   lt6911uxc-audio
+Info:   Used in combination with the lt6911uxc overlay to route the audio
+        from the lt6911uxc over I2S to the Pi.
+        Wiring is LRCK/WFS to GPIO 19, BCK/SCK to GPIO 18, and DATA/SD to GPIO
+        20.
+Load:   dtoverlay=lt6911uxc-audio
+
```
pelwell, line 2441: Document the `card-name` parameter.

pelwell (commented):
With those fixed I approve the overlay side of the PR. It still needs @6by9's approval, and dividing into the 4 commits as discussed previously - probably in that order.

drivers/media/i2c/lt6911uxc_regs.h
```diff
@@ -0,0 +1,72 @@
+/* SPDX-License-Identifier: (GPL-2.0)
```
6by9, line 1: Using SPDX licence headers means you don't need the boilerplate of the licence.
SPDX, description, and copyright is sufficient.

Same for the C file(s)

drivers/media/i2c/lt6911uxc_regs.h
```diff
@@ -0,0 +1,72 @@
...
+#define H_BP_0P5	0x8676	/* horizontal half back porch pixel */
+#define H_SW_0P5	0x8672	/* hsync half length pixel  */
+#define V_TOTAL		0x867A	/* vertical total lines */
+#define V_ACTIVE	0x867E	/* vertical active lines */
+#define V_BP		0x8674	/* vertical back porch lines */
+#define V_FP		0x8675	/* vertical front porch lines */
+#define V_SW		0x8671	/* vsync length lines */
+
+#define SYNC_POL	0x8670	/* hsync/vsync polarity flags */
+#define MASK_VSYNC_POL	(1 << 1)
```
6by9, line 42: Generally use the BIT macro instead of (1 << N) defines.

arch/arm/boot/dts/overlays/README
nicolasb565: Resolved in last commit

arch/arm/boot/dts/overlays/README
nicolasb565: Resolved in last commit

drivers/media/i2c/lt6911uxc.c
```diff
@@ -0,0 +1,989 @@
...
+
+/* v4l2 debug level */
+static int debug;
+module_param(debug, int, 0644);
+MODULE_PARM_DESC(debug, "debug level (0-3)");
+
+/* custom v4l2 controls */
+#define V4L2_CID_USER_LT6911UXC_BASE (V4L2_CID_USER_BASE + 0x1090)
+#define LT6911UXC_CID_AUDIO_SAMPLING_RATE (V4L2_CID_USER_LT6911UXC_BASE + 1)
+#define LT6911UXC_CID_AUDIO_PRESENT       (V4L2_CID_USER_LT6911UXC_BASE + 2)
```
6by9, line 48: Ah, the bane of adding custom ioctls - you need to reserve a bank of CID values.
```V4L2_CID_USER_BASE + 0x1090``` clashes with an existing assignment [```V4L2_CID_USER_MAX217X_BASE```](https://elixir.bootlin.com/linux/latest/C/ident/V4L2_CID_USER_MAX217X_BASE).

These really do need to be registered with mainline, otherwise you get conflicts every time they move on. We ran into the same issue with the BCM2835_ISP driver as that is still waiting to be upstreamed #4440 

Regardless, the control block needs to be defined in the uapi headers for use by applications.

File: drivers/media/i2c/lt6911uxc.c [hunk 8]
```diff
@@ -0,0 +1,989 @@
...
+
+#define POLL_INTERVAL_MS 1000
+
+/* v4l2 dv timings */
+static struct v4l2_dv_timings default_timing = V4L2_DV_BT_CEA_1920X1080P60;
```
6by9, line 53: This looks odd as it isn't a variable, however the macro for a structure assignment.
This should either be ```const``` as it never changes, or used explicitly in ```lt6911uxc_initial_setup``` (although it should be ```const``` there too).

drivers/media/i2c/lt6911uxc.c
```diff
@@ -0,0 +1,989 @@
+// SPDX-License-Identifier: (GPL-2.0)
+/*
+ * lt6911uxc.c - Lontium 4k60 HDMI-CSI bridge driver
+ *
+ * Copyright (c) 2020, Lukas Neuner <neur@zhaw.ch>
+ *
+ * This program is based on the tc358840 - Toshiba HDMI to CSI-2 bridge from
+ * Armin Weiss
```
6by9, line 8: The tc358840 driver isn't in mainline, so this comment is largely irrelevant without a link. 
https://patchwork.linuxtv.org/project/linux-media/patch/44d2d84d-f18d-9c80-977f-8a2b8675e70a@xs4all.nl/ is probably the most useful as Hans Verkuil appears to have taken that driver, tidied it up a fair amount, and at least started upstreaming it.

drivers/media/i2c/lt6911uxc.c
```diff
@@ -0,0 +1,989 @@
...
+		V4L2_DV_BT_STD_CVT,
+		V4L2_DV_BT_CAP_PROGRESSIVE | V4L2_DV_BT_CAP_CUSTOM |
+		V4L2_DV_BT_CAP_REDUCED_BLANKING)
+};
+
+struct lt6911uxc_state {
+	struct i2c_client *i2c_client;
+	struct v4l2_subdev sd;
+	struct media_pad pad;
+	struct mutex lock;
```
6by9, line 74: What does the mutex protect? It seems to be initialised in ```lt6911uxc_initial_setup```, but then only taken/released in ```lt6911uxc_isr```. I could be wrong, but I don't believe ISRs are re-entrant.

drivers/media/i2c/lt6911uxc.c
```diff
@@ -0,0 +1,989 @@
...
+}
+
+static int lt6911uxc_isr(struct v4l2_subdev *sd, bool *handled)
+{
+	struct lt6911uxc_state *state = to_state(sd);
+
+	mutex_lock(&state->lock);
+	dev_dbg(sd->dev, "%s in kthread %d\n", __func__, current->pid);
+
+	lt6911uxc_ext_control(sd, true);
```
6by9, line 520: lt6911uxc_ext_control is only called from here, and released below. It therefore doesn't need to keep track of the state in ```enable_i2c```, which drops the complexity.

drivers/media/i2c/lt6911uxc.c
```diff
@@ -0,0 +1,989 @@
...
+	if (err != 1) {
+		dev_err(&client->dev, "%s: write register 0x%x from 0x%x failed\n",
+			__func__, reg, client->addr);
+		return;
+	}
+	dev_dbg(&client->dev, "i2c: write register: 0x%04X = 0x%02X\n",
+		reg, val);
+}
+
+static void lt6911uxc_i2c_rd(struct v4l2_subdev *sd, u16 reg, u8 *values, u32 n)
```
6by9, line 171: Returning void from an I2C read function which can genuinely fail seems flawed. Logging a message is not error handling, particularly as the values array will not have been touched, so the calling function will quite likely be dealing with uninitialised stack.

drivers/media/i2c/lt6911uxc.c
```diff
@@ -0,0 +1,989 @@
...
+	return 0;
+}
+
+static int lt6911uxc_subscribe_event(struct v4l2_subdev *sd, struct v4l2_fh *fh,
+		struct v4l2_event_subscription *sub)
+{
+	switch (sub->type) {
+	case V4L2_EVENT_SOURCE_CHANGE:
+		return v4l2_src_change_event_subdev_subscribe(sd, fh, sub);
+	default:
```
6by9, line 390: As per https://elixir.free-electrons.com/linux/latest/source/drivers/media/i2c/tc358743.c#L1496, you should allow for subscribing to V4L2_EVENT_CTRL too.

drivers/media/i2c/lt6911uxc.c
```diff
@@ -0,0 +1,989 @@
...
+		byte_clock = fm2 << 16 | fm1 << 8 | fm0;
+		dev_dbg(dev, "byte clock %d [kHz], MIPI clock %d [kHz]\n",
+			byte_clock, byte_clock*4);
+
+		/* MIPI */
+		lanes = lt6911uxc_i2c_rd8(&state->sd, MIPI_LANES);
+		state->csi_lanes_in_use = lanes;
+		dev_dbg(dev, "MIPI lanes %d\n", lanes);
+
+		lt6911uxc_csi_enable(&state->sd, true);
```
6by9, line 449: Enabling the CSI interface without checking the streaming status of the device seems a little unusual.

drivers/media/i2c/lt6911uxc.c
```diff
@@ -0,0 +1,989 @@
...
+
+		if (!state->detected_timings.bt.width) {
```
6by9, line 451: Why the difference between ```!state->detected_timings.bt.width``` and the normal ```else``` clause. Both update the timings because they've changed, and the different debug message seems an insufficient reason to have a separate clause.

drivers/media/i2c/lt6911uxc.c
```diff
@@ -0,0 +1,989 @@
...
+			state->signal_present = false;
+			v4l2_subdev_notify_event(&state->sd,
+						&lt6911uxc_ev_source_change);
+			dev_dbg(dev, "event: no signal\n");
+
+			memset(&state->timings, 0, sizeof(state->timings));
+			memset(&state->detected_timings, 0,
+			       sizeof(state->detected_timings));
+		}
+		if (handled)
```
6by9, line 424: This function is only called from one place, and handled is passed in. It seems redundant checking if handled is not NULL.
If the polled thread hadn't passed in a handled variable to lt6911uxc_isr then I would have understood it.

drivers/media/i2c/lt6911uxc.c
```diff
@@ -0,0 +1,989 @@
...
+	dev_dbg(sd->dev, "%s in kthread %d\n", __func__, current->pid);
+
+	lt6911uxc_ext_control(sd, true);
+
+	/* Retrieve interrupt event */
+	lt6911uxc_hdmi_int_handler(state, handled);
+
+	lt6911uxc_audio_int_handler(state, handled);
+
+	lt6911uxc_log_status(sd);
```
6by9: This is going to create a fair amount of logging information if the HDMI input is connected/disconnected frequently. The same info is available via userspace calls, so seems redundant here.

drivers/media/i2c/lt6911uxc.c
```diff
@@ -0,0 +1,989 @@
...
+		v4l2_print_dv_timings(sd->name, "query_dv_timings: ",
+				timings, true);
+	return 0;
+}
+
+static int lt6911uxc_s_stream(struct v4l2_subdev *sd, int enable)
+{
+	v4l2_dbg(3, debug, sd, "%s(): enable %d\n", __func__, enable);
+
+	/* handled by ISR */
```
6by9, line 642: It really shouldn't be. Some CSI2 receivers require an idle bus with LP-11 state when initialising, and you can't achieve that.

The ISR potentially automatically stops on a source change or disconnect, but it shouldn't automatically start.

drivers/media/i2c/lt6911uxc.c
```diff
@@ -0,0 +1,989 @@
...
+		return -EINVAL;
+
+	/* retrieve mbus pixelcode and active video frame size */
+	fmt->code = state->mbus_fmt_code;
+	fmt->width  = state->timings.bt.width;
+	fmt->height = state->timings.bt.height;
+	fmt->field  = V4L2_FIELD_NONE;
+	fmt->colorspace = V4L2_COLORSPACE_SRGB;
+	fmt->ycbcr_enc = V4L2_YCBCR_ENC_601;
+	fmt->quantization = V4L2_QUANTIZATION_LIM_RANGE;
```
6by9, line 665: Is it really always 601 and limited range? Are the conversion matrices used to get RGB HDMI data to YUYV CSI2 data actually controllable.
What about YCbCr HDMI data which can also be 601, 709, or 2020.

drivers/media/i2c/lt6911uxc.c
```diff
@@ -0,0 +1,989 @@
...
+	format->format.code = code;
+
+	if (ret)
+		return ret;
+
+	switch (code) {
+	case MEDIA_BUS_FMT_UYVY8_1X16:
+		break;
+	default:
+		return -EINVAL;
```
6by9, line 693: I'm fairly certain this is wrong, but note that TC358743 does the same.
AFAIK V4L2 requires that you return a valid format, not an error code. Same if you get passed in an invalid resolution, it should be fixed up to something that the device can produce.

Admittedly for the video device not subdevice API, but https://www.kernel.org/doc/html/latest/userspace-api/media/v4l/vidioc-g-fmt.html 
> -EINVAL The struct [v4l2_format](https://www.kernel.org/doc/html/latest/userspace-api/media/v4l/vidioc-g-fmt.html#c.V4L.v4l2_format) type field is invalid or the requested buffer type not supported.

drivers/media/i2c/lt6911uxc.c
```diff
@@ -0,0 +1,989 @@
...
+	const u32 mask = V4L2_MBUS_CSI2_LANE_MASK;
+
+	cfg->type = V4L2_MBUS_CSI2_DPHY;
+	cfg->flags = (state->csi_lanes_in_use << __ffs(mask)) & mask;
+
+	/* In DT mode, only report the number of active lanes */
+	if (sd->dev->of_node)
+		return 0;
+
+	/* Support for non-continuous CSI-2 clock is missing in pdate mode */
```
6by9, line 746: pdata, not pdate.

But we're not using pdata, and that feels very much like a copy/paste rather than necessarily applying here. Confirmation would be nice.

drivers/media/i2c/lt6911uxc.c
```diff
@@ -0,0 +1,989 @@
...
+				lt6911uxc_g_timings_cap(state), NULL, NULL);
+}
+
+/* ------ Register OPS ------------------------------------------------------ */
+
+static int lt6911uxc_open(struct v4l2_subdev *sd, struct v4l2_subdev_fh *fh)
+{
+	struct i2c_client *client = v4l2_get_subdevdata(sd);
+
+	return 0;
```
6by9, line 788: Why bother? You're doing nothing.

drivers/media/i2c/lt6911uxc.c
```diff
@@ -0,0 +1,989 @@
...
+	.get_mbus_config	= lt6911uxc_get_mbus_config,
+};
+
+static struct v4l2_subdev_ops lt6911uxc_ops = {
+	.core	= &lt6911uxc_subdev_core_ops,
+	.video	= &lt6911uxc_subdev_video_ops,
+	.pad	= &lt6911uxc_pad_ops,
+};
+
+#ifdef CONFIG_MEDIA_CONTROLLER
```
6by9, line 824: Just ```select MEDIA_CONTROLLER``` in Kconfig and remove the ```#ifdef```s. 
You're calling ```media_entity_pads_init``` unconditionally below, so I don't believe this will compile without MEDIA_CONTROLLER being enabled.

drivers/media/i2c/lt6911uxc.c
```diff
@@ -0,0 +1,989 @@
...
+	},
+	.id_table = lt6911uxc_id,
+	.probe	  = lt6911uxc_probe,
+	.remove   = lt6911uxc_remove,
+};
+module_i2c_driver(lt6911uxc_driver);
+
+MODULE_DESCRIPTION("Driver for Lontium lt6911uxc HDMI to CSI-2 Bridge");
+MODULE_AUTHOR("Lukas Neuner <neur@zhaw.ch>");
+MODULE_AUTHOR("Alexey Gromov <groo@zhaw.ch>");
```
6by9, line 1030: Do they know that you're listing them as author. Any support requests are likely to head back to them if this is left here.

drivers/media/i2c/lt6911uxc.c
```diff
@@ -0,0 +1,989 @@
+// SPDX-License-Identifier: (GPL-2.0)
+/*
+ * lt6911uxc.c - Lontium 4k60 HDMI-CSI bridge driver
```
6by9, line 3: 4k60, but you only allow 4k30. I'd define it as HDMI2.0 or similar rather than a resolution and rate.

6by9 (commented):
There are a load of ```checkpatch.pl --strict``` errors over alignment. I would like to see this sent upstream, therefore it may as well be clean to start with.

MAINTAINERS really should be updated too. I'd suggest you approach the original authors with regard whether they want to support forks of their code or not.

nicolasb565:
I will address the comments above.

I will add support for actually detecting if the lt6911uxc is there on i2c bus. If the device is not there, I suspect we don't want to create a v4l2 device and we should print some kind of warning?

What I was also thinking was to read the flash and check we got the firmware that was expected. If not, we write the firmware to the flash and restart the Lontium so that it will run properly using the expected firmware.

6by9:
> I will add support for actually detecting if the lt6911uxc is there on i2c bus. If the device is not there, I suspect we don't want to create a v4l2 device and we should print some kind of warning?

If you can check an ID register, then that is a good plan. Most CSI subdevs will, eg https://github.com/torvalds/linux/blob/master/drivers/media/i2c/ov9282.c#L752 called from https://github.com/torvalds/linux/blob/master/drivers/media/i2c/ov9282.c#L1035. It logs an error message and fails the probe if the device isn't present.

> What I was also thinking was to read the flash and check we got the firmware that was expected. If not, we write the firmware to the flash and restart the Lontium so that it will run properly using the expected firmware.

I'd forgotten you'd mentioned that when reviewing the patches. Doing so would be ideal, or initially just checking the firmware version against your supported one, and again erroring out if it differs.

nicolasb565:
> > I will add support for actually detecting if the lt6911uxc is there on i2c bus. If the device is not there, I suspect we don't want to create a v4l2 device and we should print some kind of warning?
> 
> If you can check an ID register, then that is a good plan. Most CSI subdevs will, eg https://github.com/torvalds/linux/blob/master/drivers/media/i2c/ov9282.c#L752 called from https://github.com/torvalds/linux/blob/master/drivers/media/i2c/ov9282.c#L1035. It logs an error message and fails the probe if the device isn't present.
> 
> > What I was also thinking was to read the flash and check we got the firmware that was expected. If not, we write the firmware to the flash and restart the Lontium so that it will run properly using the expected firmware.
> 
> I'd forgotten you'd mentioned that when reviewing the patches. Doing so would be ideal, or initially just checking the firmware version against your supported one, and again erroring out if it differs.

Yes I manage to see if the device is there over i2c. I probe a chip id and check it's the lt6911uxc id.

I'm finishing up the code for firmware programming.

For the firmware blob binary, how/where do I include this in the pull request?

6by9:
> Yes I manage to see if the device is there over i2c. I probe a chip id and check it's the lt6911uxc id.

That sounds fine - it's always useful to validate the connected hardware.

> I'm finishing up the code for firmware programming.
>
> For the firmware blob binary, how/where do I include this in the pull request?

Do you have permission to distribute the firmware blob, and under what licence?

For mainline they would normally be submitted via linux-firmware@kernel.org for merging into https://git.kernel.org/pub/scm/linux/kernel/git/firmware/linux-firmware.git/ alongside the driver.

Seeing as this driver isn't being sent upstream yet, open an issue on https://github.com/RPi-Distro/repo/issues linking to the firmware file and requesting it to be merged into the linux-firmware package, and the relevant person should pick it up.

nicolasb565:
> > Yes I manage to see if the device is there over i2c. I probe a chip id and check it's the lt6911uxc id.
> 
> That sounds fine - it's always useful to validate the connected hardware.
> 
> > I'm finishing up the code for firmware programming.
> > For the firmware blob binary, how/where do I include this in the pull request?
> 
> Do you have permission to distribute the firmware blob, and under what licence?
> 
> For mainline they would normally be submitted via [linux-firmware@kernel.org](mailto:linux-firmware@kernel.org) for merging into https://git.kernel.org/pub/scm/linux/kernel/git/firmware/linux-firmware.git/ alongside the driver.
> 
> Seeing as this driver isn't being sent upstream yet, open an issue on https://github.com/RPi-Distro/repo/issues linking to the firmware file and requesting it to be merged into the linux-firmware package, and the relevant person should pick it up.

We will not do firmware programming in the driver. License and NDA issues... You can buy those chips already programmed by Lontium. They also provide example code to program a new firmware under NDA.

drivers/media/i2c/lt6911uxc.c
nicolasb565: Seams like it's to make a distinction between no hdmi input -> hdmi input compared to a simple timing change?

drivers/media/i2c/lt6911uxc.c
nicolasb565: What does it do? We don't notify for those type of events?

drivers/media/i2c/lt6911uxc.c
nicolasb565: Yes I noticed it's quite noisy and I'm going to make it a lot more quiet tomorrow. Only log something when there is an actual change.

drivers/media/i2c/lt6911uxc.c
nicolasb565: What do you mean by that? I don't really know what is considered usual or not.

drivers/media/i2c/lt6911uxc.c
nicolasb565: Yes I agree.

6by9:
I'm not sure where to go with this.

If Lontium have multiple different firmware, then the driver needs to either be written to work with all of them, or verify which version is programmed. Without that there will be numerous support issues that will come back to "wrong firmware, but nothing you can do about that".
That sort of fairly fundamental question needs to be addressed by the mainline Linux folks, not us, therefore the driver will really need to be submitted upstream. I've already given review comments that should head off the initial issues they'd raise, but expect it to be a moderately involved process.

PR #5236, opened 2022-11-03
vc04_services: bcm2835_codec: Allow larger images through the ISP

6by9:
Whilst the codecs are restricted to 1920x1080 / 1080x1920, the ISP isn't, but the limits advertised via V4L2 was 1920x1920 for all roles.

Increase the limit to 16k x 16k for the ISP.


6by9:
@naushir Could I have a review?

This is the simple 1-in, 1-out ISP driver, and is just making the max size the same as the complex one. Someone wanted to pass 4k video through it, and the old 1920x1920 limit prevented them.

naushir:
LGTM!

PR #5237, opened 2022-11-04
Final Clock Range API 

mripard:
Hi,

The clock range API has finally been merged upstream, with some changes compared to what we carried downstream.

This PR undoes the work that has been rejected, adds the API that was merged, and converts the users of the old API to the new one.


popcornmix:
This appeared to be fine in testing.

Desktop worked with two monitors (one 4kp60) and I could change resolutions.
kodi could play 4kp60 video (and hevc clock was clamped to 550Mz as desired).

No complaints in dmesg.

pelwell:
Thanks!

PR #5238, opened 2022-11-04
Final Clock Range API, for 6.1

mripard:
Hi,

Here's a PR equivalent to #5237 for the 6.1 branch. It differs a bit because we carried another intermediate API, and there was some conflicts.


popcornmix:
This appeared to be fine in testing.

Desktop worked with two monitors (one 4kp60) and I could change resolutions.
kodi could play 4kp60 video (and hevc clock was clamped to 550Mz as desired).

No complaints in dmesg.

pelwell:
Thanks!

PR #5243, opened 2022-11-09
Adding support for PTP on RPI 3B+

tiagofreire-pt:
The PR https://github.com/raspberrypi/linux/pull/4856 added on the lan driver support for `SOF_TIMESTAMPING_TX_SOFTWARE`. Therefore, only kernel modules support is still lacking for full support of PTP (while on software) kernel timestamping.

The use case is to allow a stratum-1 PPS disciplined NTP server basic support for PTP and increased accuracy across the LAN, without the need of a kernel compiling effort.


tiagofreire-pt:
Delete this PR.

pelwell:
I don't see the option to delete it - it's **Closed**, and that's enough.

Issue #5153, opened 2022-08-30
Unusual Uasp file


symbios24:
### Describe the bug

Hello there

Is it possible please to add to the file: https://github.com/raspberrypi/linux/blob/rpi-5.15.y/drivers/usb/storage/unusual_uas.h

Some usb 3.0 chipsets/controllers more??

The Jmicron Jms578

The Asmedia Asm1153E

The Asmedia Asm235CM

And The Initio Inic3639

Thanks in advance 

### Steps to reproduce the behaviour

.

### Device (s)

Raspberry Pi 4 Mod. B

### System

Raspbian Buster latest version

pelwell:
Do you have the information required? A Pull Request would be ideal.

symbios24:
What information you need??

I dont know how to do pull requests

pelwell:
> What information you need??

The text to add to unusual_uas.h.

symbios24:
You need the output of the lsusb command ??

pelwell:
lsusb would give the vendor and device IDs, but that doesn't tell us which quirks are needed for each one.

symbios24:
My asm1153e quirks are: 174c:55aa

pelwell:
No  - that's a device id for an ASMedia Technology ASM1051E SATA 6Gb/s bridge. Quirks are things like `US_FL_NO_REPORT_OPCODES` and `US_FL_IGNORE_UAS`.

symbios24:
How can i see that,with what command??

pelwell:
Why do you think the controllers you listed need to be added to `unusual_uas.h`? What have you been doing to set the quirk(s) up to now?

symbios24:
The jms567 that is included in the unusual_uas file works good with quirks enabled and both mechanical and ssds have great performance 

Im using both mechanical and ssds with that controller's most of them perform somehow ok with quirks enabled and ssds but with mechanical disks their performance is very bad with quirks or not there is no difference 

If you cannot add them all at least add the Asmedia Asm1153E
And the Initio inic3639

pelwell:
What I'm hearing is "I have several devices that use a number of SATA bridges. One of them works well, the others don't. The one that works well is included in unusual_uas.h, the others aren't, so please add the others to it to fix them."

What have I got wrong, or missed out?

symbios24:
If you cannot add any device,can you at least tell me how to add it my self manual to the file in my os directly??

pelwell:
[...]

PR #5246, opened 2022-11-10
Fix productName for JMS578

asmello:
Quirk detection did not work correctly for my device, leading to UAS being used, which in turn led to a lot of troubleshooting of random failures. Since adding a quirk override to `cmdline.txt`, my device has been working flawlessly in BOT mode.

I suspect the quirk did not get applied automatically because the `productName` value is incorrectly recorded in the kernel entry. Even if I'm wrong, there is no harm in making this change - this name _is_ wrong, and was probably a copy-paste mistake.

### DIFF
--- a/drivers/usb/storage/unusual_devs.h
+++ b/drivers/usb/storage/unusual_devs.h
@@ -2126,7 +2126,7 @@ UNUSUAL_DEV(  0x152d, 0x0567, 0x0114, 0x0117,
 /* Reported by David Kozub <zub@linux.fjfi.cvut.cz> */
 UNUSUAL_DEV(0x152d, 0x0578, 0x0000, 0x9999,
 		"JMicron",
-		"JMS567",
+		"JMS578",
 		USB_SC_DEVICE, USB_PR_DEVICE, NULL,
 		US_FL_BROKEN_FUA),
 
--- a/drivers/usb/storage/unusual_uas.h
+++ b/drivers/usb/storage/unusual_uas.h
@@ -114,7 +114,7 @@ UNUSUAL_DEV(0x152d, 0x0567, 0x0000, 0x9999,
 /* Reported-by: David Kozub <zub@linux.fjfi.cvut.cz> */
 UNUSUAL_DEV(0x152d, 0x0578, 0x0000, 0x9999,
 		"JMicron",
-		"JMS567",
+		"JMS578",
 		USB_SC_DEVICE, USB_PR_DEVICE, NULL,
 		US_FL_BROKEN_FUA),
 


pelwell:
The quirk entry was provided by the following upstream commit - https://github.com/raspberrypi/linux/commit/62354454625741f0569c2cbe45b2d192f8fd258e - in which the author says:

> USB: uas and storage: Add US_FL_BROKEN_FUA for another JMicron JMS567 ID
>
> There is another JMS567-based USB3 UAS enclosure (152d:0578) that fails
> with the following error:

It is clear that "JMS567" is the intended device name, so I see nothing wrong with the change. If JMicron have confusing or reused product IDs then you need to take it up with them.

asmello:
Hey, I think you are being a bit too premature in closing this. Two things:

1. My device gets reported as JMS578. See the output of `lsusb -v`:

```
Bus 002 Device 002: ID 152d:0578 JMicron Technology Corp. / JMicron USA Technology Corp. JMS578 SATA 6Gb/s
Couldn't open device, some information will be missing
Device Descriptor:
  bLength                18
  bDescriptorType         1
  bcdUSB               3.00
  bDeviceClass            0
  bDeviceSubClass         0
  bDeviceProtocol         0
  bMaxPacketSize0         9
  idVendor           0x152d JMicron Technology Corp. / JMicron USA Technology Corp.
  idProduct          0x0578 JMS578 SATA 6Gb/s
  bcdDevice            2.09
  iManufacturer           1 JMicron
  iProduct                2 USB to ATA/ATAPI Bridge
  iSerial                 3 0123456789ABCDEF
  bNumConfigurations      1
  Configuration Descriptor:
    bLength                 9
    bDescriptorType         2
    wTotalLength       0x0079
    bNumInterfaces          1
    bConfigurationValue     1
    iConfiguration          0
    bmAttributes         0x80
      (Bus Powered)
    MaxPower              896mA
    Interface Descriptor:
      bLength                 9
      bDescriptorType         4
      bInterfaceNumber        0
      bAlternateSetting       0
      bNumEndpoints           2
      bInterfaceClass         8 Mass Storage
      bInterfaceSubClass      6 SCSI
      bInterfaceProtocol     80 Bulk-Only
      iInterface              0
      Endpoint Descriptor:
        bLength                 7
        bDescriptorType         5
        bEndpointAddress     0x81  EP 1 IN
        bmAttributes            2
          Transfer Type            Bulk
          Synch Type               None
          Usage Type               Data
        wMaxPacketSize     0x0400  1x 1024 bytes
        bInterval               0
        bMaxBurst              15
      Endpoint Descriptor:
        bLength                 7
        bDescriptorType         5
        bEndpointAddress     0x02  EP 2 OUT
        bmAttributes            2
          Transfer Type            Bulk
          Synch Type               None
          Usage Type               Data
        wMaxPacketSize     0x0400  1x 1024 bytes
        bInterval               0
        bMaxBurst              15
    Interface Descriptor:
      bLength                 9
      bDescriptorType         4
      bInterfaceNumber        0
      bAlternateSetting       1
      bNumEndpoints           4
      bInterfaceClass         8 Mass Storage
      bInterfaceSubClass      6 SCSI
      bInterfaceProtocol     98
      iInterface              0
      Endpoint Descriptor:
        bLength                 7
        bDescriptorType         5
        bEndpointAddress     0x01  EP 1 OUT
        bmAttributes            2
          Transfer Type            Bulk
          Synch Type               None
          Usage Type               Data
        wMaxPacketSize     0x0400  1x 1024 bytes
        bInterval               0
        bMaxBurst               0
        Command pipe (0x01)
      Endpoint Descriptor:
        bLength                 7
        bDescriptorType         5
        bEndpointAddress     0x82  EP 2 IN
        bmAttributes            2
          Transfer Type            Bulk
          Synch Type               None
          Usage Type               Data
        wMaxPacketSize     0x0400  1x 1024 bytes
        bInterval               0
        bMaxBurst               0
        MaxStreams             32
        Status pipe (0x02)
      Endpoint Descriptor:
        bLength                 7
        bDescriptorType         5
        bEndpointAddress     0x83  EP 3 IN
        bmAttributes            2
          Transfer Type            Bulk
          Synch Type               None
          Usage Type               Data
        wMaxPacketSize     0x0400  1x 1024 bytes
        bInterval               0
        bMaxBurst              14
        MaxStreams             32
        Data-in pipe (0x03)
      Endpoint Descriptor:
        bLength                 7
        bDescriptorType         5
        bEndpointAddress     0x04  EP 4 OUT
        bmAttributes            2
          Transfer Type            Bulk
          Synch Type               None
          Usage Type               Data
        wMaxPacketSize     0x0400  1x 1024 bytes
        bInterval               0
        bMaxBurst               7
        MaxStreams             32
        Data-out pipe (0x04)
```

If this is in error, we have a mistake somewhere else in the code.

2. My device didn't get detected as a quirk device. Was I wrong in assuming this was because the `productName` didn't match? If so, can you help me diagnose why this happened? Given both `vendorId` and `productId` match exactly.

pelwell:
This page (https://usb-ids.gowdy.us/read/UD/152d/0578) suggests that there is confusion about the identity of devices with the ID 152d:057 - perhaps JMicron have reused the same ID or given some 567's the wrong ID?

Either way, this isn't something that we have changed in our kernel tree - it comes direct from the upstream maintainers - and the name is not used in determining whether or not to apply quirks.

asmello:
Ack, thanks for confirming. Happy to keep the old name then, but do you have a hypothesis as to why the quirk was not applied automatically?

pelwell:
Can you confirm the cmdline.txt addition that makes it work?

If you read the rest of unusual_devs.h and unusual_uas.h you'll see that a small number of devices are given the `US_FL_IGNORE_UAS` flag. Is the problem actually that special flags are being applied to your JMS device but not the right ones?

pelwell:
Note that the cmdline `quirks=` flag `u` sets the flag `US_FL_IGNORE_UAS`. which seems to be the important one:
```
	if (flags & US_FL_IGNORE_UAS) {
		dev_warn(&udev->dev,
			"UAS is ignored for this device, using usb-storage instead\n");
		return 0;
	}
```

asmello:
This is what I added: `usb-storage.quirks=152d:0578:u`

The rest is the default raspbian settings:
```
usb-storage.quirks=152d:0578:u console=serial0,115200 console=tty1 root=PARTUUID=5487497e-02 rootfstype=ext4 fsck.repair=yes rootwait
```

> If you read the rest of unusual_devs.h and unusual_uas.h you'll see that a small number of devices are given the US_FL_IGNORE_UAS flag. Is the problem actually that special flags are being applied to your JMS device but not the right ones?

Ah, I think you are right on the dot. The current entry only specified `US_FL_BROKEN_FUA` which I guess does _not_ imply UAS is disabled. I just assumed everything in this list had UAS disabled by default.

Is this something we want to change? I assume David Kozub didn't have the same issues I'm having when he submitted the original entry, so maybe not every device with this id is affected.

pelwell:
I see the subject has been covered before in https://github.com/raspberrypi/linux/issues/5153, which seems to have stalled. Perhaps you should comment there and we can decide whether to do something or mark it as Won't Fix.

PR #5249, opened 2022-11-14
User selection of HDMI RGB/YCbCr444/YCbCr422 mode

6by9:
Replicates the old hdmi_pixel_encoding config.txt option.

### DIFF
--- a/drivers/gpu/drm/vc4/vc4_hdmi.c
+++ b/drivers/gpu/drm/vc4/vc4_hdmi.c
@@ -1315,12 +1315,12 @@ static void vc5_hdmi_set_csc_coeffs_swap(struct vc4_hdmi *vc4_hdmi,
 	lockdep_assert_held(&vc4_hdmi->hw_lock);
 
 	/* YUV444 needs the CSC matrices using the channels in a different order */
-	HDMI_WRITE(HDMI_CSC_12_11, (coeffs[2][1] << 16) | coeffs[2][0]);
-	HDMI_WRITE(HDMI_CSC_14_13, (coeffs[2][3] << 16) | coeffs[2][2]);
-	HDMI_WRITE(HDMI_CSC_22_21, (coeffs[0][1] << 16) | coeffs[0][0]);
-	HDMI_WRITE(HDMI_CSC_24_23, (coeffs[0][3] << 16) | coeffs[0][2]);
-	HDMI_WRITE(HDMI_CSC_32_31, (coeffs[1][1] << 16) | coeffs[1][0]);
-	HDMI_WRITE(HDMI_CSC_34_33, (coeffs[1][3] << 16) | coeffs[1][2]);
+	HDMI_WRITE(HDMI_CSC_12_11, (coeffs[1][1] << 16) | coeffs[1][0]);
+	HDMI_WRITE(HDMI_CSC_14_13, (coeffs[1][3] << 16) | coeffs[1][2]);
+	HDMI_WRITE(HDMI_CSC_22_21, (coeffs[2][1] << 16) | coeffs[2][0]);
+	HDMI_WRITE(HDMI_CSC_24_23, (coeffs[2][3] << 16) | coeffs[2][2]);
+	HDMI_WRITE(HDMI_CSC_32_31, (coeffs[0][1] << 16) | coeffs[0][0]);
+	HDMI_WRITE(HDMI_CSC_34_33, (coeffs[0][3] << 16) | coeffs[0][2]);
 }
 
 static void vc5_hdmi_csc_setup(struct vc4_hdmi *vc4_hdmi,
--- a/drivers/gpu/drm/vc4/vc4_hdmi.c
+++ b/drivers/gpu/drm/vc4/vc4_hdmi.c
@@ -538,6 +538,7 @@ static int vc4_hdmi_connector_atomic_check(struct drm_connector *connector,
 
 	if (old_state->colorspace != new_state->colorspace ||
 	    old_vc4_state->broadcast_rgb != new_vc4_state->broadcast_rgb ||
+	    old_vc4_state->requested_output_format != new_vc4_state->requested_output_format ||
 	    !drm_connector_atomic_hdr_metadata_equal(old_state, new_state)) {
 		struct drm_crtc_state *crtc_state;
 
@@ -572,6 +573,8 @@ int vc4_hdmi_connector_get_property(struct drm_connector *connector,
 
 	if (property == vc4_hdmi->broadcast_rgb_property) {
 		*val = vc4_conn_state->broadcast_rgb;
+	} else if (property == vc4_hdmi->output_format_property) {
+		*val = vc4_conn_state->requested_output_format;
 	} else {
 		DRM_DEBUG_ATOMIC("Unknown property [PROP:%d:%s]\n",
 				 property->base.id, property->name);
@@ -602,6 +605,9 @@ int vc4_hdmi_connector_set_property(struct drm_connector *connector,
 
 	if (property == vc4_hdmi->broadcast_rgb_property) {
 		vc4_conn_state->broadcast_rgb = val;
+		return 0;
+	} else if (property == vc4_hdmi->output_format_property) {
+		vc4_conn_state->requested_output_format = val;
 		return 0;
 	}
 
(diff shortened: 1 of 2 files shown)


6by9:
@mripard You may want to look at the first of these two patches - the swap in ```vc5_hdmi_set_csc_coeffs_swap``` is wrong. (I haven't checked that it is the same as mainline, but assume it is).

drivers/gpu/drm/vc4/vc4_hdmi.c
```diff
@@ -2028,6 +2064,20 @@ vc4_hdmi_encoder_compute_format(const struct vc4_hdmi *vc4_hdmi,
...
+		int ret;
+
+		ret = vc4_hdmi_encoder_compute_clock(vc4_hdmi, vc4_state,
+						     mode, bpc, format);
+		if (!ret) {
+			vc4_state->output_format = format;
+			return 0;
+		}
+	}
+
```
mripard, line 2086: This patch looks mostly ok to me, but it looks like if we fail that check, we'll fallback onto the old behaviour.

I can see two things a bit wrong with that:

- I'd expect that if the user explicitly sets a format, they'd also expect to fail if we can't match that
- Also, let's assume that the user sets RGB, it doesn't work for that resolution, we fallback to YUV422. The property will still report the output format as RGB, which seems wrong?

I guess we could use a default (ie, 0) value of "auto" that would trigger the old behaviour. If the user forces it to something else, we just try that configuration and fail if we can't. What do you think?

6by9: Sure can go for an "auto", with any other setting being explicit. That is closer to the behaviour of "Broadcast RGB" for range, so makes sense.
It may fail if hardware or the monitor is not be able to do it, but that's the user's lookout.

mripard: If it's ever a concern, we could add another property listing the available formats based on the monitor and controller capabilities.

6by9: I think the only case it can really fail is if trying to force RGB when 422 is required due to bandwidth limits.
The monitor should tell you what is available via the EDID, but it is overriding this when the EDID is dodgy that is the whole point of this property.

I'll try to respin in the next couple of days - rather a lot on at present.

6by9:
Patch updated to have a (default) auto mode that behaves as before, and overrides for the other modes.

It also drops the requirement for 4:2:2 to be driven in 12bpc mode. The HDMI spec appears to allow 24, 30, and 36 bpc modes for 4:2:2, even though the packing on the wire is always 36bit. It certainly works (presumably stuffing the data into the MSBs), but I wonder if others agree with handling this.

nicolasb565:
Is there a way to set that at runtime from userspace in a way that could avoid color format conversion in a gstreamer pipeline?

The use case I have in mind would be to send input from uvc camera YUYV/YUY2 and send that to hdmi 0 output in 4:2:2

Thanks!

6by9:
> Is there a way to set that at runtime from userspace in a way that could avoid color format conversion in a gstreamer pipeline?
>
> The use case I have in mind would be to send input from uvc camera YUYV/YUY2 and send that to hdmi 0 output in 4:2:2

This is changing the output format of the HDMI block.
All composition is done in a fully sampled RGB space (ie 420 and 422 have the chroma upsampled and YUV converted to RGB), and the HVS doesn't support YUYV pixel interleaved formats on the input side, so this is of limited use in your use case.

mripard:
I can't really comment for the last one, I don't know the spec well enough. @HiassofT maybe?

The two first patches look fine to me though, so I'm fine with merging this if there's no further comment on the third

6by9:
@mripard - thanks for the response.

As per the commit text, the HDMI 2.0 spec refers to 24, 30, or 36 bit support in 4:2:2, even though on the wire 4:2:2 is always packed as 36 bits.
Having discussed with @popcornmix we're not aware of anything telling the display how many of those 36 bits are being used, so it will presumably use all 12bits, but the HDMI block correctly packs the data in the MSBs anyway, so fundamentally it works.

nicolasb565:
> 

Would there be a way to just do some kind of "pass through" without having any composition? Basically just send the data over hdmi. My use case does not need composition.

6by9:
> Would there be a way to just do some kind of "pass through" without having any composition? Basically just send the data over hdmi. My use case does not need composition.

Please don't hikack a pull request (or unrelated issue) with random requests.

HiassofT:
IMO we should leave the YCC 4:2:2 12-bit check in place.

The max_bpc connector property controls output on the wire, which will always be 12 bits for Y and 12 bits for both C comonents in case of YCC 4:2:2 and we added the check to have an easy method of disabling 4:2:2 in auto mode - eg you wouldn't ever want your desktop to run in 4:2:2 or you may also want to disable YCC 4:2:2 because a TV/monitor can't really cope with it.

So we chose to make 4:2:2 opt-in by requring to set max_bpc to 12

Transmitting 8 or 10bit content in 12bit 4:2:2 and padding LSBs is a different thing

6by9:
Auto mode will still try RGB first, so there is no way that you'll get 4:2:2 by accident, and that behaviour hasn't changed.

This new DRM property is intended as an override for the odd cases where displays or capture devices only accept YCbCr 444 or 422, same as we originally had ```hdmi_pixel_encoding``` on the legacy stack.

There are several places in HDMI2.0 such as:
Section 5 Overview
>  HDMI 1.4b also defines a mechanism for transporting YCBCR 4:2:2 Pixels with Pixel sizes of 24, 30, or 36 bits. 

Table 7-6 for BT 2020
> Bits/component 10 or 12. Pixel encoding RGB or YCBCR 4:2:2 or YCBCR 4:4:4

So it is in the spec that bit depths other than 12 are valid in YCbCr 422, although it's fairly scant on details of how it is intended to be implemented. 

It was the annoyance that in order to force 422 you potentially have to also force max_bpc to 12 in the same atomic commit, otherwise the commit fails. You can't necessarily set max_bpc to 12 first in a separate commit as it wouldn't be valid for 4k60 RGB.

> we added the check to have an easy method of disabling 4:2:2 in auto mode

Things should be technically correct, not an easy way around a problem.

HiassofT:
The spec is quite clear how less than 12 bits per component have to be transmitted, LSB zero padding up to 12 bits
```
The YC B C R 4:2:2 pixel encoding on HDMI closely resembles standard ITU-R BT.601. The
high-order 8 bits of the Y samples are mapped onto the 8 bits of Channel 1 and the low-order 4
bits are mapped onto the low-order 4 bits of Channel 0. If fewer than 12 bits are used, the valid bits
shall be left-justified (i.e. MSb=MSb) with zeroes padding the bits below the LSb.
```
It's a 12-bit per component mode and unless you want to change the meaning of max_bpc to mean average bits per component (resembling TMDS character rate) instead of real bits per component then allowing the 12bit 4:2:2 mode with max_bpc != 12 is wrong

HiassofT:
Ah, and setting max_bpc to 12 is fine and will work with 4kp60, too - it's only max and the max TMDS rate checks take care that 12 or 10 bit 4:4:4 modes will be rejected

pelwell:
So is this PR good to merge or not?

6by9:
@Hiassoft Thanks for the reference - I'd looked at that section several times and totally missed it.

Let me see if I can sensibly drop the non-12bpc YUV422 patch.
If there's no sensible path to selecting the 422 mode from modetest and xrandr, then it's worthless.

It's getting to the mode in 4k60 that is the niggle.
output_format=auto, max_bpc=12 (will switch to 422), and then you can set output_format=422.
If output_format!=auto, then trying to set max_bpc=12 will fail in 4k60, and it's these weird forcing the order of some changes that I never like. Admittedly 422 support is patchy, but no other DRM driver appears to enforce this.

I can't test that path here at home as I have no 4k monitor, so I'll drop it for now. I may come back to it later.

I've also just noted that explicitly setting "Output format" still checks the EDID for whether that mode is reported as supported. That (correctly) stops me forcing 444 or 422 on the DVI monitor that I happen to have hooked up. We may want to revisit that but too.

6by9:
Branch updated to drop that patch.

drivers/gpu/drm/vc4/vc4_hdmi.c
pelwell: It's not clear what happened to this idea.

6by9: Github seems to be attributing the conversation to now totally different code.

I went with a new setting of VC4_HDMI_OUTPUT_AUTO that retained the old behaviour. 

The other settings of RGB, YUV444, and YUV422 will all try to force the output format, and fail if not supportable.

PR #5251, opened 2022-11-17
ASoC: da7219: Improve pole orientation detection on certain headsets

davidrau-renesas-opensource:
Enhanced the control of ground switches in AAD logic. 
The current mechanism can solve the pervious detection issue better.


pelwell:
You have submitted a patch for a file that is unmodified from the upstream kernel, and for a driver that isn't even enabled in the Raspberry Pi kernels. A change like this should be sent to the kernel maintainers. However, an `msleep` within an interrupt handler, even a threaded interrupt handler, especially one which is unconditional, seems like a bad idea.

Before you send your patch upstream you'll need to write a better commit message explaining what you have changed and why. The submission process is explained here: https://www.kernel.org/doc/html/latest/process/submitting-patches.html

Once the patch is accepted we'll happily back-port it, and enable the codec in our builds.

davidrau-renesas-opensource:
Hi Phil,

Thanks for the kind reponse and suggestions. :)

I will recheck this patch with my co-worker and provide more details to kernel maintainer.

PR #5257, opened 2022-11-25
Added Digi2 Pro audio card

reinisb:
Back in 2021 Hifiberry released a new card called `Digi2 Pro` ([ref](
https://www.hifiberry.com/blog/the-new-digi2-pro/)). It uses the same dtoverlay as their `Digi+ Pro` card. To reduce confusion about whether `Digi2 Pro` is supported, I've extended the README to explicitly mention also `Digi2 Pro`.


pelwell:
Thanks!

PR #5260, opened 2022-12-01
media: i2c: ov7251: Add module param to select ext trig mode

6by9:
As there isn't currently a defined mechanism for selecting an external trigger mode on image sensors, copy the imx477 approach of using a module parameter to enable ext trig.

### DIFF
--- a/drivers/media/i2c/ov7251.c
+++ b/drivers/media/i2c/ov7251.c
@@ -23,6 +23,10 @@
 #include <media/v4l2-fwnode.h>
 #include <media/v4l2-subdev.h>
 
+static int trigger_mode;
+module_param(trigger_mode, int, 0644);
+MODULE_PARM_DESC(trigger_mode, "Set vsync trigger mode: 0=standalone, (1=source - not implemented), 2=sink");
+
 #define OV7251_SC_MODE_SELECT		0x0100
 #define OV7251_SC_MODE_SELECT_SW_STANDBY	0x0
 #define OV7251_SC_MODE_SELECT_STREAMING		0x1
@@ -244,7 +248,6 @@ static const struct reg_value ov7251_setting_vga[] = {
 	{ 0x3662, 0x01 },
 	{ 0x3663, 0x70 },
 	{ 0x3664, 0x50 },
-	{ 0x3666, 0x0a },
 	{ 0x3669, 0x1a },
 	{ 0x366a, 0x00 },
 	{ 0x366b, 0x50 },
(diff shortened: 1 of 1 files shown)


drivers/media/i2c/ov7251.c
```diff
@@ -23,6 +23,10 @@
 #include <media/v4l2-fwnode.h>
 #include <media/v4l2-subdev.h>
 
+static int trigger_mode;
+module_param(trigger_mode, int, 0644);
+MODULE_PARM_DESC(trigger_mode, "Set vsync trigger mode: 0=standalone, non-zero=sink");
```
pelwell: This is the only place you refer to a `sink` trigger. I think `external` is more obvious, as well as matching the code.

Documenting the two settings as 0 and 1 is logical and leaves the way open to supporting other modes, without any loss of functionality (or change of implementation).

drivers/media/i2c/ov7251.c
```diff
@@ -309,9 +312,8 @@ static const struct reg_value ov7251_setting_vga[] = {
 	{ 0x3c00, 0x89 },
 	{ 0x3c01, 0x63 },
 	{ 0x3c02, 0x01 },
-	{ 0x3c03, 0x00 },
 	{ 0x3c04, 0x00 },
-	{ 0x3c05, 0x03 },
+	{ 0x3c05, 0x01 },
```
pelwell, line 316: What does this change do?

6by9: {0x3C04, 0x3C05} number of active frames

0x3C03 when 0x17: External trigger snapshot mode A rising edge on FSIN pin wakes the sensor up and streams out {0x3404, 0x3405} frames

So previously each rising edge triggered 3 frames (the default for some reason), which sort of defeats the point of it being solely externally triggered.
https://datasheetspdf.com/pdf-file/1252891/OmniVision/OV7750/1

drivers/media/i2c/ov7251.c
6by9: In theory the OV7251 supports generating a trigger signal from the same FSIN/VSYNC pin, but I haven't got that working as yet. (Same with OV9281)
We already have IMX477 with a module parameter called "trigger_mode" with 0=standalone, 1=trigger source, and 2=trigger sink.

The plan is that once I have a mainline driver able to both source and sink trigger signals, it will become standardised via a V4L2 control and upstreamed. This was discussed at the Linux Media Summit in Dublin back in Sept, but I haven't got there yet.

Perhaps retain exactly the same syntax as imx477, but ignore trigger_mode=1?

drivers/media/i2c/ov7251.c
pelwell: Understood.

drivers/media/i2c/ov7251.c
pelwell: > Perhaps retain exactly the same syntax as imx477, but ignore trigger_mode=1?

I prefer that. I'd even be happy with treating all non-zero values the same for now as long as the comment said to use 2.

6by9: Updated so that 2 sets for ext trig sink mode, and anything else disables.

1 is documented as being ext trig source, but noted as currently not being implemented.

PR #5262, opened 2022-12-02
usb: xhci: add XHCI_VLI_HUB_TT_QUIRK

P33M:
This one's pretty horrible, but in the absence of a hub firmware update it'll have to do for now.

Spinning is unavoidable but not sufficient (the hardware can still cause babble by itself). There will be a maximum of 125us delay if submitting only one URB to the endpoint at once. Generally multiple URBs are pipelined to maintain throughput.

### DIFF
--- a/drivers/usb/host/xhci-ring.c
+++ b/drivers/usb/host/xhci-ring.c
@@ -3585,6 +3585,48 @@ static int xhci_align_td(struct xhci_hcd *xhci, struct urb *urb, u32 enqd_len,
 	return 1;
 }
 
+static void xhci_vl805_hub_tt_quirk(struct xhci_hcd *xhci, struct urb *urb,
+				    struct xhci_ring *ring)
+{
+	struct list_head *tmp;
+	struct usb_device *udev = urb->dev;
+	unsigned int timeout = 0;
+	unsigned int single_td = 0;
+
+	/*
+	 * Adding a TD to an Idle ring for a FS nonperiodic endpoint
+	 * that is behind the internal hub's TT will run the risk of causing a
+	 * downstream port babble if submitted late in uFrame 7.
+	 * Wait until we've moved on into at least uFrame 0
+	 * (MFINDEX references the next SOF to be transmitted).
+	 *
+	 * Rings for IN endpoints in the Running state also risk causing
+	 * babble if the returned data is large, but there's not much we can do
+	 * about it here.
+	 */
+	if (udev->route & 0xffff0 || udev->speed != USB_SPEED_FULL)
+		return;
+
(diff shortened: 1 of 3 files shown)


drivers/usb/host/xhci-ring.c
```diff
@@ -3585,6 +3585,48 @@ static int xhci_align_td(struct xhci_hcd *xhci, struct urb *urb, u32 enqd_len,
...
+static void xhci_vl805_hub_tt_quirk(struct xhci_hcd *xhci, struct urb *urb,
+				    struct xhci_ring *ring)
+{
+	struct list_head *tmp;
+	struct usb_device *udev = urb->dev;
+	unsigned int timeout = 0;
+	unsigned int single_td = 0;
+
+	/*
+	 * Adding a TD to to an Idle ring for a FS nonperiodic endpoint
```
pelwell: `to to`

drivers/usb/host/xhci-ring.c
```diff
@@ -3585,6 +3585,48 @@ static int xhci_align_td(struct xhci_hcd *xhci, struct urb *urb, u32 enqd_len,
...
+		}
+	}
+	if (single_td) {
+		while (timeout < 20 &&
+		       (readl(&xhci->run_regs->microframe_index) & 0x7) == 0) {
+			udelay(10);
+			timeout++;
+		}
+		if (timeout >= 20)
+			xhci_warn(xhci, "MFINDEX didn't advance - %u.%u dodged\n",
```
pelwell, line 3624: Have you seen this warning in testing?

P33M: No, but falling out of the loop is a theoretical possibility if:
- The xhci controller is in the process of failing because of PCIe link disruption
- The xhci driver is incorrectly allowing submission of URBs when the integrated hub's upstream port is in the Disabled state
- There's some other bug in the VL805 implementation that causes this register to not get updated or stuck at zero.

pelwell: Sure. I was just checking that it doesn't need to be limited in any way.

drivers/usb/host/xhci-ring.c
P33M: done

eMPee584:
.. Would this quirk be applicable to other architectures too?
Asking because I've been having severe troubles with my VL805 on debian amd64 kernels > 5.19 ..

P33M:
The patch is architecture-agnostic. I have no reason to believe that the TT bug is fixed in any other version of non-Pi4 VL805 firmware.

eMPee584:
So would you consider this fit for mainline inclusion? What symptom does it fix?

PR #5263, opened 2022-12-05
Fixed SMI driver issue when writing and using DMA

madimario:
When using the SMI device and writing data larger than DMA_THRESHOLD_BYTES, the data was copied from user space to kernel space after the DMA had already completed. This means that the data written out will not be the data stored in user_ptr, but whatever was already in the kernel buffers.

I've added code that will first copy the user data from user space to kernel space before starting the DMA transfer. The code I've added may not be the best way to handle the problem, but I didn't want to change the files too much.


madimario:
Hi, just wondering if this is being worked on behind the scenes. Is there anything you would like me to change or do to help move this forward?

pelwell:
Any comments, @Wren6991 ?

Issue #4575, opened 2021-09-11
5.10.63 causes 1000s of lines of errors relating to hdmi-audio-codec hdmi-audio-codec.2.auto: ASoC: error at snd_soc_dai_startup on i2s-hifi: -19


graysky2:
**Describe the bug**
1. When kodi starts, dmesg is polluted with thousands of lines like this:
```
hdmi-audio-codec hdmi-audio-codec.2.auto: ASoC: error at snd_soc_dai_startup on i2s-hifi: -19
```
2. If I start my RPi4 with the attached TV off, the RPi4 does not detect that an external monitor is connected and the only way to enable HDMI output is to reboot the RPi4 with the TV on.

Downgrading to 5.10.59-1 fixes both issues.  There are only 5 lines in dmesg displaying that error. 
 That kernel is based on this commit: https://github.com/raspberrypi/linux/commit/22201d4868f66aa7773d5c4bad7bab1795ee6757

**To reproduce**
* Install 5.10.63-1 on Arch ARM aarch64 (https://github.com/raspberrypi/linux/commit/597cd63b48c6dc3746827504f191744c120fbb4d)
* Start kodi

**Expected behaviour**
1. dmesg should not be spammed with 1000s of lines
2. HDMI detection should work regardless of TV on/off state

**Actual behaviour**
1. dmesg is spammed with 1000s of lines.
2. HDMI detection is broken relative to 5.10.59

**System**
 Copy and paste the results of the raspinfo command in to this section. Alternatively, copy and paste a pastebin link, or add answers to the following questions:

* Which model of Raspberry Pi? RPi4B Rev 1.4
* Which OS and version?  Arch ARM aarch64
* Which firmware version (`vcgencmd version`)?
```
Aug 31 2021 14:52:47 
Copyright (c) 2012 Broadcom
version 67615e950e1e28b92dfae6303cf7a8b879a8908f (clean) (release) (start)
```
* Which kernel version (`uname -a`)?
`Linux treetop 5.10.63-1-ARCH #1 SMP PREEMPT`

**Logs**
If applicable, add the relevant output from `dmesg` or similar.

**Additional context**
Add any other relevant context for the problem.

lategoodbye:
I submitted a RFC patch last year, but it wasn't accepted because the issue must be solved differently:
https://www.spinics.net/lists/dri-devel/msg284535.html

More recent discussion:
https://www.spinics.net/lists/alsa-devel/msg127150.html

[...]

PR #5265, opened 2022-12-06
vc4_hdmi: Avoid log spam for audio start failure

popcornmix:
We regularly get dmesg error reports of:
[   18.184066] hdmi-audio-codec hdmi-audio-codec.3.auto: ASoC: error at snd_soc_dai_startup on i2s-hifi: -19
[   18.184098]  MAI: soc_pcm_open() failed (-19)

Currently I get 30 of these when booting to desktop. We always say, ignore they are harmless, but removing them would be good.

A bit of investigation shows, for me, the errors are all generated by second, unused hdmi interface.

It shows as an alsa device, and pulseaudio attempts to open it (numerous times), generating a kernel error message each time.

systemctl --user restart pulseaudio.service generates 6 additional error messages.

The error messages all come through:
https://github.com/raspberrypi/linux/blob/a009a9c0d79dfec114ee5102ec3d3325a172c952/sound/soc/soc-pcm.c#L39

which suggests returning ENOTSUPP, rather that ENODEV will be quiet. And indeed it is.

Note: earlier kernels do not have the quiet ENOTSUPP, so additional cherry-picks will be needed to backport

### DIFF
--- a/drivers/gpu/drm/vc4/vc4_hdmi.c
+++ b/drivers/gpu/drm/vc4/vc4_hdmi.c
@@ -2359,7 +2359,7 @@ static int vc4_hdmi_audio_startup(struct device *dev, void *data)
 	}
 
 	if (!vc4_hdmi_audio_can_stream(vc4_hdmi)) {
-		ret = -ENODEV;
+		ret = -ENOTSUPP;
 		goto out_dev_exit;
 	}
 


popcornmix:
See: https://github.com/raspberrypi/linux/issues/4575

6by9:
Looks reasonable to me.

popcornmix:
I've also cherry-picked:
https://github.com/raspberrypi/linux/commit/b63605b13ef77f0720ce5bba6864e76f07429c97

to rpi-5.15.y and rpi-6.0.y trees which is needed to make ENOTSUPP quiet.
(it was a clean cherry-pick on 6.0, but needed minor fixing up in 5.15).

PR #5270, opened 2022-12-08
KUnit Support and Github CI

mripard:
Hi,

This is mostly a backport of most of the KUnit work that have been contributed upstream recently and is likely to hit 6.3.

It introduces some core helpers and some unit tests for the VC4 driver.

Finally, it adds (for downstream only) a github workflow file that will run the generic KMS unit tests and the VC4 tests, on both ARM and ARM64


pelwell:
Do you have a feel for how long it would take to run the workflow using the default, free GitHub runner?

mripard:
During my test runs, the generic tests took about 3 minutes, the vc4 tests 5 minutes each. They run in parallel though.

pelwell:
I'm tempted to add the tests but make them opt-in, since you are essentially the only person working on the code and the chances of an apparently unrelated change affecting the outcomes seems small. What do you think, @popcornmix?

mripard:
If it's a concern, we can run them only if something under drivers/gpu was modified?

popcornmix:
How does opt in work on github?
Do we get an optional button to run tests on a PR? Or does it need to be kicked off from a separate interaface?

pelwell:
The completely manual way is via the Actions tab. PRs from first-time users require approval for the test to be run, but that could be extended to requiring approval for PRs from all users.

pelwell:
> that could be extended to requiring approval for PRs from all users.

No, that's not correct - the most I can do from the Settings menu is require approval for outside contributors.

mripard:
I just had a look, we can definitely run a workflow only if a given path is modified: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-including-paths

Would that be acceptable?

6by9:
The workflow needs to include ```workflow_dispatch:``` to allow it to be triggered manually from the Actions tab.

Seeing as the tests will run in parallel with the kernel builds, and take significantly less time to run than those kernel builds, is there a reason to not run them?

popcornmix:
I think my view is that while evaluating exactly what tests we want to run, we get more information by enabling tests by default.
If we determine that a test isn't providing value (too slow, false fails, unlikely to ever fail etc) we can switch it to manual or remove it completely.

@mripard This needs rebasing after the 6.1.0 bump (sorry!)

pelwell:
OK - rebase it and I'll merge.

mripard:
> I think my view is that while evaluating exactly what tests we want to run, we get more information by enabling tests by default. If we determine that a test isn't providing value (too slow, false fails, unlikely to ever fail etc) we can switch it to manual or remove it completely.

We for sure need to get rid of unreliable tests, and a good argument can indeed be made for slow tests. But I don't really get why we would need to remove tests that are fast and reliable if they always pass. Isn't it the ideal we should tend to?

If we strive to make all tests pass all the time, pushing that logic a bit would make us remove all tests eventually, which is odd :) 

> @mripard This needs rebasing after the 6.1.0 bump (sorry!)

I'll do it

pelwell:
There are limits to how much free testing is available, and limits to how long we want to wait for tests to pass. Pruning unnecessary tests helps on both counts.

6by9:
> There are limits to how much free testing is available, and limits to how long we want to wait for tests to pass. Pruning unnecessary tests helps on both counts.

Limits on testing?
https://github.blog/2019-08-08-github-actions-now-supports-ci-cd/#simple-pay-as-you-go-pricing
> Actions is free for the 40 million developers on GitHub to use with public repositories.

https://docs.github.com/en/billing/managing-billing-for-github-actions/about-billing-for-github-actions#about-billing-for-github-actions
> GitHub Actions usage is free for standard GitHub-hosted runners in public repositories

There is a limit on the number of concurrent jobs, but even the Free layer at 20 concurrent jobs (and that's test jobs, not threads or compile jobs) is fairly generous. I don't know what level the company subscribes to.
https://docs.github.com/en/actions/learn-github-actions/usage-limits-billing-and-administration#usage-limits

And all Actions are currently running in parallel (this will change as and when we add running tests on real hardware), so as long as they take less time than the longest kernel build (about 45mins), there is no delay in running it.

popcornmix:
Bah - github is misbehaving and is showing this PR as needing rebasing (with rebase and merge disabled with "This branch cannot be rebased due to too many changes").
However looking directly at your tree, the commits are based on:
`5725471b0d1eff179278191722c811a7ebd851d0 (HEAD -> rpi-6.1.y, origin/rpi-6.1.y`
which should be fine.

I can push the commits directly, but it would be nice to know why github gets unhappy (it has happened before with your PRs).

mripard:
I don't know, the PR is all green on my end: 

image

pelwell:
I see the same messages as Dom, and yet my checkout of the PR appears to be 35 commits ahead of rpi-6.1.y, i.e. a simple fast-forward.

Merged offline.

PR #5273, opened 2022-12-13
Add 8DEV USB2CAN driver support

lihgong:
Enable compiling the 8DEV USB2CAN as module, which makes my day easier without build that myself


pelwell:
Although I appreciate the purity, separate PRs for each affected defconfig is over the top. I've added the remaining defconfigs and merged it offline. See https://github.com/raspberrypi/linux/commit/d864f9360da700994c425af41a3ff10770685db3.

PR #5274, opened 2022-12-13
Add 8DEV CAN2USB support

lihgong:
Enable compiling the 8DEV USB2CAN as module, which makes my day easier without build that myself


pelwell:
Resolving via #5273.

Issue #4289, opened 2021-04-17
[Pi 4B 8GB] When EEE (Energy Efficient Ethernet) is enable and active on eth0 at gigabit speed, ethernet becomes slow to unresponsive with packets dropping


Photopuppet:
**Summary**
When Energy Efficient Ethernet is enabled and active from a new install of Raspberry Pi OS on a Pi 4B 8GB, **gigabit ethernet** transfer with the built-in ethernet adapter becomes slow and unresponsive with packets being dropped. Other computers supporting EEE plugged into the same network behave normally, even if they are tested in the same port that the Pi was connected. When the Pi is connected with EEE enabled, it also results in some odd behaviour on other devices connected to the same switch such as causing the other device to drop packets and/or reconnect constantly. Removing the Pi from the network causes the behaviour on the switch and other device to return to normal.

**To reproduce**
When Pi 4B 8GB is connected to gigabit switch with default EEE behaviours, the network performance of the Pi deteriorates as above.

Status from ethtool --show-eee eth0:

EEE Settings for eth0:                                                                                                                                                                                                                                       
         EEE status: enabled - active                                                                                                                                                                                                                           
         Tx LPI: inactive                                                                                                                                                                                                                                   
         Supported EEE link modes:  100baseT/Full                                                                                                                                                                                                               
                                    1000baseT/Full                                                                                                                                                                                                              
         Advertised EEE link modes:  100baseT/Full                                                                                                                                                                                                              
                                     1000baseT/Full                                                                                                                                                                                                             
         Link partner advertised EEE link modes:  100baseT/Full                                                                                                                                                                                                 
                                                  1000baseT/Full 

**Temporary fix**
As soon as EEE is disabled with 'ethtool --set-eee eth0 eee off', ethernet performance returns to normal at gigabit speeds.

**Logs**
I am unsure which logs would be necessary to help diagnose the problem... please let me know and I will be happy to add them!

Equipment in use: Home network with BT Business Hub in use as modem and switch + Powerline ethernet with built in switch (BT Mini Connector)

A relevant forum thread: https://www.raspberrypi.org/forums/viewtopic.php?t=305820

ganzgustav22:
I've read here: https://www.raspberrypi.org/forums/viewtopic.php?t=305820 that it cannot simply be disable via the ethtool command because it gets re-enabled later for whatever reason. One user fixed this with a script running in a constant loop that disables it again and again once a second, which I'd like to avoid.

[...]

PR #5277, opened 2022-12-14
Add the ability to disable EEE on Pi 4s

pelwell:
The EEE implementation on BCM2711 doesn't work well with all switches. Add a module parameter and dtparam that allows it to be disabled without the need for a service to run ethtool after boot.

See #4289 for more details.


popcornmix:
Looks reasonable to me.

Wenri:
Coming from Issue #4289. I built the 64f1b87b and confirm it works. EEE can be disabled by adding `genet.eee=N` to cmdline.txt and no packet loss in 1000 sent packets. Thanks a lot for the patch and hope it can be merged soon.

Issue #4446, opened 2021-07-09
HDMI output is broken when /boot/config.txt contains hdmi_enable_4kp60=1


graysky2:
**Describe the bug**
Booting with `hdmi_enable_4kp60=1` in `/boot/config.txt` results in no video output.  Removing this line restores HDMI output.

**To reproduce**
Use the following for `/boot/config.txt` and boot.
```
dtoverlay=vc4-kms-v3d,cma-512
dtoverlay=rpivid-v4l2
disable_overscan=1
disable_fw_kms_setup=1
hdmi_enable_4kp60=1
```

**Expected behaviour**
The machine should boot displaying video output (boot messages, console login etc.)

**Actual behaviour**
The machine boots but there is not HDMI output available.  The monitor does not detect any signal.

**System**
> Copy and paste the results of the raspinfo command in to this section. Alternatively, copy and paste a pastebin link, or add answers to the following questions:

This is ArchARM which does not ship with `raspinfo`

* Which model of Raspberry Pi? RPi4B
* Which firmware version (`vcgencmd version`): https://github.com/raspberrypi/userland/commit/97bc8180ad682b004ea224d1db7b8e108eda4397
* Which kernel version (`uname -a`): `Linux workbench 5.10.48-1-ARCH #1 SMP PREEMPT Fri Jul 9 14:19:24 EDT 2021 aarch64 GNU/Linux`

**Logs**
I see the following in `dmesg`:
```
[  +0.007207] vc4-drm gpu: bound fe400000.hvs (ops vc4_hvs_ops [vc4])
[  +0.001395] Registered IR keymap rc-cec
[  +0.000085] rc rc0: vc4 as /devices/platform/soc/fef00700.hdmi/rc/rc0
[  +0.000074] input: vc4 as /devices/platform/soc/fef00700.hdmi/rc/rc0/input1
[  +0.000423] vc4_hdmi fef00700.hdmi: Could not register sound card: -517
[  +0.008635] platform regulatory.0: Direct firmware load for regulatory.db failed with error -2
[  +0.000026] cfg80211: failed to load regulatory.db
[  +0.018276] brcmfmac: F1 signature read @0x18000000=0x15264345
[  +0.004858] FAT-fs (mmcblk0p1): Volume was not properly unmounted. Some data may be corrupt. Please run fsck.
[  +0.010943] brcmfmac: brcmf_fw_alloc_request: using brcm/brcmfmac43455-sdio for chip BCM4345/6
[...]

PR #5281, opened 2022-12-15
vc4_crtc: Increase setup cost in core clock calculation to handle extreme reduced blanking

popcornmix:
See: https://github.com/raspberrypi/linux/issues/4446

The formula that determines the core clock requirement based on pixel clock and blanking has been determined experimentally to minimise the clock while supporting all modes we've seen.

A new reduced blanking mode (4kp60 at 533MHz rather than the standard 594MHz) has been seen that doesn't produce a high enough clock and results in "flip_done timed out" error.

Increase the setup cost in the formula to make this work. The result is a reduced blanking mode increases by up to 7MHz while leaving the standard timing mode untouched

### DIFF
--- a/drivers/gpu/drm/vc4/vc4_crtc.c
+++ b/drivers/gpu/drm/vc4/vc4_crtc.c
@@ -720,7 +720,7 @@ static int vc4_crtc_atomic_check(struct drm_crtc *crtc,
 
 		mode = &crtc_state->adjusted_mode;
 		if (vc4_encoder->type == VC4_ENCODER_TYPE_HDMI0) {
-			vc4_state->hvs_load = max(mode->clock * mode->hdisplay / mode->htotal + 1000,
+			vc4_state->hvs_load = max(mode->clock * mode->hdisplay / mode->htotal + 8000,
 						  mode->clock * 9 / 10) * 1000;
 		} else {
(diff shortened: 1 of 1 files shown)


pelwell:
> while leaving the standard timing mode untouched

Is that taken care of by the max()? I was expecting the change to be larger.

popcornmix:
Yes - see calculations in linked thread.

Basically we do core_clock = pixel_clock * max(active_ratio, 90%)
The standard mode has an active_ratio (hactive/htotal) of 3840 / 4400 = 87%, the reduced one is 96%.

So standard is clamped at 90% (right side of max), reduced is using 87% (left side of max).

The setup cost increase doesn't affect the standard mode. It was:
`max(594.00 * 3840 / 4400 + 1, 594.00 * 9 / 10), = max(519.4, 534.6) = 534MHz`

it becomes:
`max(594.00 * 3840 / 4400 + 8, 594.00 * 9 / 10), = max(526.4, 534.6) = 534MHz`

timg236:
As discussed on the other thread. LGTM.

popcornmix:
@mripard I think this change would be useful upstream.

njhollinghurst:
^ Please ignore the above link, which was a typo.

PR #5282, opened 2022-12-15
Camera dt muxing options for CMs

6by9:
Make life easier for using cameras on CM1, 3, and 4S by providing easy options for rerouting the I2C.

Corrects the default shutdown GPIO on CM1&3

Adds overrides for CM4 and 4S to reassign the camera shutdown GPIOs.

### DIFF
--- a/arch/arm/boot/dts/overlays/cm-swap-i2c0-overlay.dts
+++ b/arch/arm/boot/dts/overlays/cm-swap-i2c0-overlay.dts
@@ -1,0 +1,27 @@
+// SPDX-License-Identifier: GPL-2.0-only
+// Definitions for IMX708 camera module on VC I2C bus
+/dts-v1/;
+/plugin/;
+
+#include <dt-bindings/gpio/gpio.h>
+
+/{
+	compatible = "brcm,bcm2835";
+
+	fragment@0 {
+		target = <&i2c0mux>;
+		i2c0mux_frag: __overlay__ {
+			pinctrl-0 = <&i2c0_gpio28>;
+			pinctrl-1 = <&i2c0_gpio0>;
+		};
+	};
+
+	__overrides__ {
+		i2c0-gpio0 = <&i2c0mux_frag>, "pinctrl-0:0=",<&i2c0_gpio0>;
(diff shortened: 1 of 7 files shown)


arch/arm/boot/dts/overlays/cm-swap-i2c0-overlay.dts
```diff
@@ -0,0 +1,27 @@
...
+// Definitions for IMX708 camera module on VC I2C bus
+/dts-v1/;
+/plugin/;
+
+#include <dt-bindings/gpio/gpio.h>
+
+/{
+	compatible = "brcm,bcm2835";
+
+	i2c0mux_frag: fragment@0 {
```
pelwell: I think this label is in the wrong place - it should be on the `__overlay__` node, otherwise the changes fall down the cracks.

6by9: I'll have to check that tomorrow then. I was copying https://github.com/raspberrypi/linux/blob/rpi-5.15.y/arch/arm/boot/dts/overlays/imx219-overlay.dts#L18

6by9: Ah, there they are changing the target which makes sense. Here we're changing the contents, so you're almost certainly right.

pelwell:
```
$ ovmerge arch/arm/boot/dts/overlays/cm-swap-i2c0-overlay.dts,i2c0-gpio44
/dts-v1/;
/plugin/;

/ {
	compatible = "brcm,bcm2835";
	fragment@0 {
		target = <&i2c0mux>;
		pinctrl-0 = <&i2c0_gpio44>;
		__overlay__ {
			pinctrl-0 = <&i2c0_gpio28>;
			pinctrl-1 = <&i2c0_gpio0>;
		};
	};
};
```

6by9: As expected, you were right. Now fixed and tested on CM3 with i2c10-gpio44 and rerouting the I2C.

PR #5283, opened 2022-12-15
[RFC | DO NOT MERGE] vchiq_core: Please fillout comments

lategoodbye:
Hi, this is again a request for comments. The VCHIQ staging todo list contains an item to  review the remote_event_*() functions. So to make this job easier explaining comments of the higher function for each them would be nice. In order to make this task easier, i prepared a branch based on 6.1 with FIXME comments which need to be filled out.

This branch is not intended for merge.

@pelwell

Thanks


drivers/staging/vc04_services/interface/vchiq_arm/vchiq_core.c
```diff
@@ -498,6 +498,7 @@ vchiq_set_conn_state(struct vchiq_state *state, enum vchiq_connstate newstate)
 	vchiq_platform_conn_state_changed(state, oldstate, newstate);
 }
 
+/* FIXME */
```
pelwell, line 501: Remote events provide a way of presenting several virtual doorbells to a peer (ARM host to VPU) using only one physical doorbell. They can be thought of as a way for the peer to signal a semaphore, in this case implemented as a workqueue.

Remote events remain signalled until acknowledged by the receiver, and they are non-counting. They are designed in such a way as to minimise the number of interrupts and avoid unnecessary waiting.

A remote_event is as small data structures that live in shared memory. It comprises two booleans - armed and fired:
* The sender sets fired when they signal the receiver.
* If fired is set, the receiver has been signalled and need not wait.
* The receiver sets the armed field before they begin to wait.
* If armed is set, the receiver is waiting and wishes to be woken by interrupt.

remote_event_create initialises a single remote_event, and the associated wait_queue.

drivers/staging/vc04_services/interface/vchiq_arm/vchiq_core.c
```diff
@@ -536,6 +537,7 @@ remote_event_wait(wait_queue_head_t *wq, struct remote_event *event)
 	return 1;
 }
 
+/* FIXME */
```
pelwell, line 540: Acknowledge that the event has been signalled, and wake any waiters. Usually called as a result of the doorbell being rung.

drivers/staging/vc04_services/interface/vchiq_arm/vchiq_core.c
```diff
@@ -544,13 +546,15 @@ remote_event_signal_local(wait_queue_head_t *wq, struct remote_event *event)
 	wake_up_all(wq);
 }
 
+/* FIXME */
```
pelwell, line 549: Check if a single event has been signalled, waking the waiters if it has.

drivers/staging/vc04_services/interface/vchiq_arm/vchiq_core.c
```diff
@@ -544,13 +546,15 @@ remote_event_signal_local(wait_queue_head_t *wq, struct remote_event *event)
...
 static inline void
 remote_event_poll(wait_queue_head_t *wq, struct remote_event *event)
 {
 	if (event->fired && event->armed)
 		remote_event_signal_local(wq, event);
 }
 
+/* FIXME */
```
pelwell, line 557: VCHIQ used a small, fixed number of remote events. It is simplest to enumerate them here for polling.

pelwell (commented):
There you go. Don't ask me to prove the correctness of the implementation.

lategoodbye:
Thanks

Patch has been send to the staging mailing list.

PR #5285, opened 2022-12-16
overlays: Add disable-emmc2

timg236:
Add a new overlay that disables the EMMC2 controller on BCM2711. This can be useful on a Compute Module 4 if the onboard EMMC2 storage is unreliable and the system can be booted by other means e.g Network / USB.

Signed-off-by: Tim Gover <tim.gover@raspberrypi.com>

### DIFF
--- a/arch/arm/boot/dts/overlays/README
+++ b/arch/arm/boot/dts/overlays/README
@@ -873,6 +873,14 @@ Load:   dtoverlay=disable-bt
 Params: <None>
 
 
+Name:   disable-emmc2
+Info:   Disable EMMC2 controller on BCM2711.
+        The allows the onboard EMMC storage on Compute Module 4 to be disabled
+        e.g. if a fault has occurred.
+Load:   dtoverlay=disable-emmc2
+Params: <None>
+
+
 Name:   disable-wifi
 Info:   Disable onboard WLAN on Pi 3B, 3B+, 3A+, 4B and Zero W.
 Load:   dtoverlay=disable-wifi
--- a/arch/arm/boot/dts/overlays/disable-emmc2-overlay.dts
+++ b/arch/arm/boot/dts/overlays/disable-emmc2-overlay.dts
@@ -1,0 +1,13 @@
+/dts-v1/;
+/plugin/;
+
+/{
+	compatible = "brcm,bcm2711";
+
+	fragment@0 {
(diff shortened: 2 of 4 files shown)


arch/arm/boot/dts/overlays/README
```diff
@@ -872,6 +872,12 @@ Info:   Disable onboard Bluetooth on Pi 3B, 3B+, 3A+, 4B and Zero W, restoring
 Load:   dtoverlay=disable-bt
 Params: <None>
```
pelwell, line 874: Two blank lines between entries.

arch/arm/boot/dts/overlays/README
```diff
@@ -872,6 +872,12 @@ Info:   Disable onboard Bluetooth on Pi 3B, 3B+, 3A+, 4B and Zero W, restoring
...
 
+Name:   disable-emmc2
+Info:   Disable EMMC2 controller on BCM2711.
+        The allows the onboard EMMC storage on Compute Module 4 to be disabled
+        e.g. if a fault has occurred.
+Load:   dtoverlay=disable-emmc2
+Params: <None>
```
pelwell, line 882: And here.

arch/arm/boot/dts/overlays/disable-emmc2-overlay.dts
```diff
@@ -0,0 +1,13 @@
+/dts-v1/;
+/plugin/;
+
+/{
+	compatible = "bcm2711";
```
pelwell: `brcm,bcm2711`.

pelwell (commented):
In addition to the inline comments, please add an entry to overlay_map.dts saying:
```
	disable-emmc2 {
		bcm2711;
	};
```

arch/arm/boot/dts/overlays/README
timg236: done

arch/arm/boot/dts/overlays/README
timg236: done

arch/arm/boot/dts/overlays/disable-emmc2-overlay.dts
timg236: done

PR #5288, opened 2022-12-19
Restore dropped support for Innolux panel

pelwell:
In the move from 5.15 to 6.0 & 6.1, a commit adding support for an Innolux LCD panel (the catchily-named AT056tN53V1) was lost. Restore that support in a separate commit.

See https://forums.raspberrypi.com/viewtopic.php?p=2063287#p2063287.

### DIFF
--- a/Documentation/devicetree/bindings/display/panel/panel-simple.yaml
+++ b/Documentation/devicetree/bindings/display/panel/panel-simple.yaml
@@ -174,6 +174,8 @@ properties:
       - ivo,m133nwf4-r0
         # Innolux AT043TN24 4.3" WQVGA TFT LCD panel
       - innolux,at043tn24
+        # Innolux AT056tN53V1 5.6" VGA (640x480) TFT LCD panel
+      - innolux,at056tn53v1
         # Innolux AT070TN92 7.0" WQVGA TFT LCD panel
       - innolux,at070tn92
         # Innolux G070Y2-L01 7" WVGA (800x480) TFT LCD panel
--- a/Documentation/userspace-api/media/v4l/subdev-formats.rst
+++ b/Documentation/userspace-api/media/v4l/subdev-formats.rst
@@ -627,7 +627,7 @@ The following tables list existing packed RGB formats.
     * .. _MEDIA_BUS_FMT_RGB565_1X24_CPADHI:
 
       - MEDIA_BUS_FMT_RGB565_1X24_CPADHI
-      - 0x1020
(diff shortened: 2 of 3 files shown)


pelwell:
@aBUGSworstnightmare-rpi Does this look OK to you?

6by9:
It looks like we've got the panel patch, adding MEDIA_BUS_FMT_BGR666_1X24_CPADHI, and dt-binding merged into one. C'est la vie. 

Just be aware that official support for MEDIA_BUS_FMT_BGR666_1X24_CPADHI is coming in 6.2 via https://github.com/torvalds/linux/commit/2468e0195c09706896712a29030132e7f1bd0bcc, so this will have a conflict.

pelwell:
Noted re 6.2. How many commits would you like this in?

6by9:
> How many commits would you like this in?

I'm not fussed over splitting it, just noting that we're potentially causing ourselves grief in forward porting.

aBUGSworstnightmare-rpi:
> It looks like we've got the panel patch, adding MEDIA_BUS_FMT_BGR666_1X24_CPADHI, and dt-binding merged into one. C'est la vie.
> 
> Just be aware that official support for MEDIA_BUS_FMT_BGR666_1X24_CPADHI is coming in 6.2 via [torvalds/linux@2468e01](https://github.com/torvalds/linux/commit/2468e0195c09706896712a29030132e7f1bd0bcc), so this will have a conflict.

Sorry, but MEDIA_BUS_FMT_BGR666_1X24_CPADHI is available in 6.1 as well https://github.com/raspberrypi/linux/blob/3fff7f771b67562955c02775e1ca472e244eab61/include/uapi/linux/media-bus-format.h#L52 . bus_format no got changed (compared to what is was when submitting the pull-request).

In case you want to add the timing to panel simple you may want to change the clock to below (but can do that with a pull-request later); gives a 60Hz output as the panel is running a little out of spec atm.
	`.pixelclock = { 25200000, 25200000, 25200000}`

pelwell:
Am I to assume, because you didn't mention my PR, that it is at least correct (if not complete)?

pelwell:
> Am I to assume, because you didn't mention my PR, that it is at least correct (if not complete)?

It appears not. The new subdev format numbers, included upstream but not documented, don't match the documentation this PR adds. I've added a follow-up patch, to be squashed on merge, that uses the new numbers (and fixes a likely cut-and-paste error in the original commit).

PR #5294, opened 2022-12-29
overlays: audremap: Fix setting of the pin function

pocek:
The `brcm,function` property must be set in the overlay fragment too - otherwise the parametrization won't work. At least that's the case for bcm2711-rpi-cm4.dts which starts with empty properties:

```
&gpio {
        audio_pins: audio_pins {
                brcm,pins = <>;
                brcm,function = <>;
        };
};
```

This was broken since a56df85d2f42fd461fdc05f33617141aca5ba465.

/cc @pelwell

### DIFF
--- a/arch/arm/boot/dts/overlays/audremap-overlay.dts
+++ b/arch/arm/boot/dts/overlays/audremap-overlay.dts
@@ -8,6 +8,7 @@
                 target = <&audio_pins>;
                 frag0: __overlay__ {
                         brcm,pins = <12 13>;
+                        brcm,function = <4>; /* alt0 alt0 */
                 };
         };
 


pelwell:
Thanks for this. I think what you have done is correct, but I'm bothered by the comment that:

> otherwise the parametrization won't work

Can you confirm that the existing overlay (i.e. before this patch) works with `dtoverlay=audremap,pins_12_13`, and that only the default case is broken? If not, the problem is more complex than I think.

pelwell:
The `,pins_12_13` workaround works for me. I think this change is correct, even if it turns out not to be sufficient. Merging in the absence of confirmation, and any remaining issues can be addressed in a separate commit.

pelwell:
The PR is now merged, and the change will be in future kernel builds.

PR #5296, opened 2023-01-01
Add ms pressure sensors overlays

PolarStarWalker:
- add overlays for ms5637, ms5803, ms5805, ms5837, ms8607 sensors
- enable module building in bcm2711_defconfig


pelwell:
Your overlays look OK, but they could be smaller and more flexible as part of the i2c-sensor container overlay. See #5301 for an alternative implementation.

The defconfig commit should target all 5 current defconfigs (and there was a typo in the commit message). Again, see #5301.

PolarStarWalker:
> Your overlays look OK, but they could be smaller and more flexible as part of the i2c-sensor container overlay. See #5301 for an alternative implementation.
> 
> The defconfig commit should target all 5 current defconfigs (and there was a typo in the commit message). Again, see #5301.

Thanks!

PR #5299, opened 2023-01-03
drm/vc4: Fix YUV plane handling when planes are in different buffers

6by9:
YUV images can either be presented as one allocation with offsets for the different planes, or multiple allocations with 0 offsets.

The driver only ever drm_fb_[dma|cma]_get_gem_obj with plane index 0, therefore any application using the second approach was incorrectly rendered.

Correctly determine the address for each plane, removing the assumption that the base address is the same for each.

Fixes: fc04023fafec ("drm/vc4: Add support for YUV planes.")


pelwell:
There's a word missing (calls?) from the second sentence of the commit message, but otherwise it LGTM.

6by9:
> There's a word missing (calls?) from the second sentence of the commit message

Fixed.

PR #5301, opened 2023-01-04
overlays: i2c-sensor: MS temp & pressure sensors

pelwell:
[ This is an alternative implementation of #5296 that supports all the I2C
interfaces and has a smaller footprint ]

Add support for the Measurement Specialities temperature and pressure sensors to the i2c-sensor overlay. The supported devices are MS5637, MS5803, MS5805, MS5837 and MS8607.


PolarStarWalker:
Very nice, thanks!

PR #5303, opened 2023-01-05
imx708: Do not reset vblank to a default value

naushir:
imx708_set_framing_limits resets the vblank control to the mode default
value unconditionally. This causes it to overwrite the user specified
vblank and exposure control values when starting the sensor, since
it is called when handling V4L2_CID_WIDE_DYNAMIC_RANGE.

Remove this call to s_ctrl as it is unnecessary.


naushir:
@6by9 @davidplowman @pelwell

davidplowman:
Thanks - have checked that this fixes the nasty "winking" when switching to/from the HDR mode.

PR #5304, opened 2023-01-05
6.2 - ov9281 deprecated for ov9282 driver

6by9:
As we're still rebasing, I've done reverts of the downstream commits.

The compatible string "ovt,ov9281" has been adopted by the new mainline driver, so no dtoverlay changes are required.


popcornmix:
Dropping downstream patches is always good.

6by9:
> Dropping downstream patches is always good.

It's why I faffed around sending the patches upstream!

pelwell:
Thanks.

PR #5306, opened 2023-01-06
drivers: media: imx708: Enable long exposure mode

naushir:
Enable long exposure modes by using the long exposure shift register setting
in the imx708 sensor.


naushir:
@njhollinghurst

njhollinghurst:
Works for me -- both non-HDR and HDR modes tested, on Pi4 and Pi3.

naushir:
I think this is now ready to be merged if there's no other feedback.

PR #5313, opened 2023-01-11
drm/vc4: bo: Fix drmm_mutex_init memory hog

### LINKED ISSUE #5302: Memory leak in video driver
### Describe the bug

When playing video files with Kodi the kernel memory (namely kmalloc-128 and kmalloc-256) is constantly increasing until the system eventually runs OOM.

Play a video with kodi and watch slabtop.

### Device (s)

Raspberry Pi 3 Mod. B+

### System

```
# cat /etc/rpi-issue
Raspberry Pi reference 2018-10-09
Generated using pi-gen, https://github.com/RPi-Distro/pi-gen, 7b40fd11bc13d6496bb6bd1bc02f946b7e0ac7cb, stage2
# vcgencmd version
Dec 12 2022 11:59:10
Copyright (c) 2012 Broadcom
version ed6f6b8fcdc6476410b9cf75d141633461d34bdd (clean) (release) (start)
# uname -a
Linux Raspberry 5.15.84-v8+ #1611 SMP PREEMPT Wed Jan 4 16:19:37 GMT 2023 aarch64 GNU/Linux
# cat /etc/issue
Debian GNU/Linux 11
```

### Logs

```
# cat kmalloc-128/alloc_traces | sort -rn | head
  39453 kstrdup_const+0x3c/0x48 age=9949/204277/303649 pid=0-712 cpus=0-3
....
# cat kmalloc-256/alloc_traces | sort -rn | head
  33110 drmm_mutex_init+0x54/0x68 [drm] age=4618/195716/305391 pid=178-587 cpus=0-3
....
```

### Additional context

I am using

```
dtoverlay=vc4-kms-v3d,cma-256
```

mripard:
Commit 374146cad469 ("drm/vc4: Switch to drmm_mutex_init") converted, among other functions, vc4_create_object() to use drmm_mutex_init().

However, that function is used to allocate a BO, and therefore the mutex needs to be freed much sooner than when the DRM device is removed from the system.

For each buffer allocation we thus end up allocating a small structure as part of the DRM-managed mechanism that is never freed, eventually leading us to no longer having any free memory anymore.

Let's switch back to mutex_init/mutex_destroy to deal with it properly.

Fixes: 374146cad469 ("drm/vc4: Switch to drmm_mutex_init")

Fixes #5302

### DIFF
--- a/drivers/gpu/drm/vc4/vc4_bo.c
+++ b/drivers/gpu/drm/vc4/vc4_bo.c
@@ -177,6 +177,7 @@ static void vc4_bo_destroy(struct vc4_bo *bo)
 		bo->validated_shader = NULL;
 	}
 
+	mutex_destroy(&bo->madv_lock);
 	drm_gem_cma_free_object(obj);
 }
 
@@ -406,9 +407,7 @@ struct drm_gem_object *vc4_create_object(struct drm_device *dev, size_t size)
 	bo->madv = VC4_MADV_WILLNEED;
 	refcount_set(&bo->usecnt, 0);
 
-	ret = drmm_mutex_init(dev, &bo->madv_lock);
-	if (ret)
-		return ERR_PTR(ret);
+	mutex_init(&bo->madv_lock);
 
 	mutex_lock(&vc4->bo_lock);
 	bo->label = VC4_BO_TYPE_KERNEL;


popcornmix:
Just played Big Buck Bunny on Pi3 with this PR and the leak of kstrdup_const and drmm_mutex_init is not present.
Pi4 still works.

So, looks good to me.

6by9:
Thanks Maxime.

Is vc4_bo_destroy always the release for vc4_create_object? vc4_bo_create feels more like the matching create function.

Chasing through, vc4_bo_create calls drm_gem_cma_create, which calls __drm_gem_cma_create, which calls the gem_create_object function, so that does match.

However __drm_gem_cma_create also appears to be called from drm_gem_cma_prime_import_sg_table. It does also return a drm_gem_dma_object, so presumably it is handled the same way as a created object, and will be released through vc4_bo_destroy. Sorry, this is only me not knowing the detail of the code paths involved.

mripard:
I'm not super familiar with the GPU/BO management side either, but my understanding is that BO are refcounted and we're supposed to give up our reference using drm_gem_object_put. If the refcount is 0, it will call drm_gem_object_free, which will call the drm_gem_object_funcs->free implementation. On vc4, it's vc4_free_object, which will then rely on vc4_bo_destroy it it indeed needs to be freed (instead of put back into a cache I reckon?).

6by9:
Fine, your knowledge seems to tally with my digging, it is just slightly counter-intuitive function naming.

LGTM.

PR #5314, opened 2023-01-11
drm_probe_helper: Cancel previous job before starting new one

popcornmix:
Currently we schedule a call to output_poll_execute from drm_kms_helper_poll_enable for 10s in future. Later we try to replace that in drm_helper_probe_single_connector_modes with a 0s schedule with delayed_event set.

But as there is already a job in the queue this fails, and the immediate job we wanted with delayed_event set doesn't occur until 10s later.

And that call acts as if connector state has changed, reprobing modes. This has a side effect of waking up a display that has been blanked.

Make sure we cancel the old job before submitting the immediate one.


popcornmix:
@mripard agreed this fix looked correct

PR #5315, opened 2023-01-12
Make Kunit work again on 6.2 

mripard:
Hi, 

Here are a few patches that make the DRM/KMS kunit compile again.

Those patches should be squashed into their original backports, hence why they are marked as fixups.

git rebase -i v6.2-rc3 should make everything fine.


pelwell:
Thanks. I'll merge them as is, then append an extra patch to (I think) work with older binutils. We can take care of the squashing in  a separate pass.

PR #5316, opened 2023-01-13
.github/workflows: Set warnings-as-errors for builds

pelwell:
To avoid code with build warnings being introduced into the tree, force
CONFIG_WERROR=y in the build workflow.

The second commit, reverting a fix that fixed a warning, will be dropped once the workflow change has been shown to be successful.


pelwell:
That seems to work - all kernel builds failed with same `unused variable 'ret'` warning/error. Now dropping the second patch to test the build succeeds. See https://github.com/raspberrypi/linux/actions/runs/3912267646.

popcornmix:
Looks to have failed as desired.
LGTM assuming it passes with test revert removed.

pelwell:
I can't seem to retrigger the workflows for the same content, but having read the errors from the previous build I'm happy it will pass.

PR #5317, opened 2023-01-13
vc4/hdmi: Always enable GCP with AVMUTE cleared

popcornmix:
See: https://forum.libreelec.tv/thread/24780-le-10-0-1-rpi4-no-picture-after-update-from-le-10-0-0

Issue is some displays go blank at the point of firmware to kms handover.
Plugging/unplugging hdmi cable, power cycling display, or switching standby off/on
typically resolve this case.

Finally managed to find a display that suffers from this, and track down the issue.

The firmware uses AVMUTE in normal operation. It will set AVMUTE before disabling hdmi
clocks and phy. It will clear AVMUTE after clocks and phy are set up for a new hdmi mode.

But with the hdmi handover from firmware to kms, AVMUTE will be set by firmware.

kms driver typically has no GCP packet (except for deep colour modes).
The spec isn't clear on whether to consider the AVMUTE as continuing indefinitely
in the absense of a GCP packet, or to consider that state to have ended.

Most displays behave as we want, but there are a number (from mutiple manufacturers)
which need to see AVMUTE cleared before displaying a picture.

Lets just always enable GCP packet with AVMUTE cleared. That resolves the issue on
problematic displays.

From HDMI 1.4 spec:
A CD field of zero (Color Depth not indicated) shall be used whenever the Sink does
not indicate support for Deep Color. This value may also be used in Deep Color mode
to transmit a GCP indicating only non-Deep Color information (e.g. AVMUTE).

So use CD=0 where we were previously not enabling a GCP.

### DIFF
--- a/drivers/gpu/drm/vc4/vc4_hdmi.c
+++ b/drivers/gpu/drm/vc4/vc4_hdmi.c
@@ -108,6 +108,10 @@
 #define VC5_HDMI_GCP_WORD_1_GCP_SUBPACKET_BYTE_1_SHIFT	8
 #define VC5_HDMI_GCP_WORD_1_GCP_SUBPACKET_BYTE_1_MASK	VC4_MASK(15, 8)
 
+#define VC5_HDMI_GCP_WORD_1_GCP_SUBPACKET_BYTE_0_MASK	VC4_MASK(7, 0)
+#define VC5_HDMI_GCP_WORD_1_GCP_SUBPACKET_BYTE_0_SET_AVMUTE	BIT(0)
+#define VC5_HDMI_GCP_WORD_1_GCP_SUBPACKET_BYTE_0_CLEAR_AVMUTE	BIT(4)
+
 # define VC4_HD_M_SW_RST			BIT(2)
 # define VC4_HD_M_ENABLE			BIT(0)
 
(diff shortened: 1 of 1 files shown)


popcornmix:
I checked with an analyser what my dell xps laptop (intel graphics) does, and it always outputs a gpc with:
Set AV Mute:off Clear AV Mute: On

which matches this PR's behaviour.

HiassofT:
As discussed on LibreELEC Slack the vc5/RPi4 change makes my HDFury Vrroom very unhappy - when kms takes over I get a flickering display (switching between black screen and signal at fast, irregular intervals). It seems to be some issue at RPi4 output/HDfury input, the OSD of the HDFury is shown fine (only the RPi4 input signal flickers).

However after switching to another mode (eg with modetest) the display is fine - also after modetest switches back to the initial mode.

vc4/RPi3B+ seem to work fine with this change, I haven't seen the flickering issue with it

popcornmix:
I've dropped the vc4 support as it's makes the PR more complicated and so far we haven't had a report of failure.
This also avoids the issue @HiassofT reported with his HD Fury.

There has been a positive report on LE forum:
https://forum.libreelec.tv/thread/24780-le-10-0-1-rpi4-no-picture-after-update-from-le-10-0-0/?postID=175991#post175991

vc4 can be added later.

HiassofT:
Testing with my HD Fury Vrroom (hooked up to a 1920x1200p60 monitor) and my LG 55C8 4k OLED TV (which didn't suffer from any problems though) looks fine so far :+1:

PR #5318, opened 2023-01-14
bcm2835-mmc: Honor return value of mmc_of_parse()

l1k:
bcm2835_mmc_probe() ignores errors returned by mmc_of_parse() and in particular ignores -EPROBE_DEFER, which may be returned if the power sequencing driver configured in the devicetree is compiled as a module.

The user-visible result is that access to the SDIO device fails because its power sequencing requirements have not been observed.  Fix it.

### DIFF
--- a/drivers/mmc/host/bcm2835-mmc.c
+++ b/drivers/mmc/host/bcm2835-mmc.c
@@ -1474,7 +1474,9 @@ static int bcm2835_mmc_probe(struct platform_device *pdev)
 	}
 
 	if (node) {
-		mmc_of_parse(mmc);
+		ret = mmc_of_parse(mmc);
+		if (ret)
+			goto err;
 
 		/* Read any custom properties */
 		of_property_read_u32(node,


pelwell:
Thanks!

l1k:
FWIW, having this also on the rpi-6.1.y branch would probably be useful as that's the next LTS release. Should I file a separate pull request or do you prefer cherry-picking yourself? Thanks a lot.

pelwell:
I was planning to back-port to 6.1, but PRs also work for me

lategoodbye:
@l1k Do you have any plans to send this mainline?

l1k:
@lategoodbye This particular driver seems to exist only in the Foundation's downstream tree. The drivers in mainline already check the return value of `mmc_of_parse()` AFAICS. Am I missing something?

lategoodbye:
Ops, i got confused by the amount of mmc drivers. Thanks for clarifying.

PR #5321, opened 2023-01-15
drm/vc4: Calculate bpc based on max_requested_bpc

HiassofT:
This aligns vc4 with Intel, AMD and Synopsis drivers and fixes max bpc connector property not working as expected as conn_state->max_bpc isn't neccessarily up to date at the point we need the property value.

I initially noticed that on LibreELEC with kernel 6.1 and was wondering why switching to 12bpc didn't work. I've tested this now on RPiOS with 64bit 5.15 kernel on a RPi4 and a 1920x1200 Monitor which supports 12-bit YCC 4:2:2 and it has the same issues.

eg enable drm debugging and set max bpc to 12 - this should switch the output to YCC 422
```
root@raspberrypi:~# echo 0x02 > /sys/module/drm/parameters/debug
root@raspberrypi:~# modetest -M vc4 -w "32:max bpc:12"
```
dmesg shows that vc4 isn't even trying to check for 12-bit support (that's because max_bpc in connector state is stil 8, only max_requested_bpc is 12)
```
[   78.346261] vc4-drm gpu: [drm:vc4_atomic_check [vc4]] crtc-3: Trying to find a channel.
[   78.346376] vc4-drm gpu: [drm:vc4_atomic_check [vc4]] crtc-3: Already enabled, reusing channel 0.
[   78.346504] vc4-drm gpu: [drm:vc4_hdmi_encoder_atomic_check [vc4]] Trying with a 8 bpc output
[   78.346620] vc4-drm gpu: [drm:vc4_hdmi_encoder_atomic_check [vc4]] Trying with an RGB output
[   78.346718] vc4-drm gpu: [drm:vc4_hdmi_sink_supports_format_bpc.isra.29 [vc4]] RGB Format, checking the constraints.
[   78.346790] vc4-drm gpu: [drm:vc4_hdmi_sink_supports_format_bpc.isra.29 [vc4]] RGB format supported in that configuration.
[   78.346862] vc4-drm gpu: [drm:vc4_hdmi_encoder_atomic_check [vc4]] Mode 1920x1200 @ 60Hz: Found configuration: bpc: 8, fmt: RGB, clock: 154000000
```

With this change, using max_requested_bpc, it works as expected:
```
[   53.821214] vc4-drm gpu: [drm:vc4_atomic_check [vc4]] crtc-3: Trying to find a channel.
[   53.821334] vc4-drm gpu: [drm:vc4_atomic_check [vc4]] crtc-3: Already enabled, reusing channel 0.
[   53.821459] vc4-drm gpu: [drm:vc4_hdmi_encoder_atomic_check [vc4]] Trying with a 12 bpc output
[   53.821537] vc4-drm gpu: [drm:vc4_hdmi_encoder_atomic_check [vc4]] Trying with an RGB output
[   53.821634] vc4-drm gpu: [drm:vc4_hdmi_sink_supports_format_bpc.isra.0 [vc4]] RGB Format, checking the constraints.
[   53.821708] vc4-drm gpu: [drm:vc4_hdmi_sink_supports_format_bpc.isra.0 [vc4]] 12 BPC but sink doesn't support Deep Color 36.
[   53.821780] vc4-drm gpu: [drm:vc4_hdmi_encoder_atomic_check [vc4]] Failed, Trying with an YUV422 output
[   53.821870] vc4-drm gpu: [drm:vc4_hdmi_sink_supports_format_bpc.isra.0 [vc4]] YUV422 format, checking the constraints.
[   53.821942] vc4-drm gpu: [drm:vc4_hdmi_sink_supports_format_bpc.isra.0 [vc4]] YUV422 format supported in that configuration.
[   53.822015] vc4-drm gpu: [drm:vc4_hdmi_encoder_atomic_check [vc4]] Mode 1920x1200 @ 60Hz: Found configuration: bpc: 12, fmt: YUV 4:2:2, clock: 154000000
```

The AMD commit 01c22997bed059 has some more info in the commit message and in the comment added to the driver

ping @mripard

### DIFF
--- a/drivers/gpu/drm/vc4/vc4_hdmi.c
+++ b/drivers/gpu/drm/vc4/vc4_hdmi.c
@@ -2124,7 +2124,7 @@ vc4_hdmi_encoder_compute_config(const struct vc4_hdmi *vc4_hdmi,
 {
 	struct drm_device *dev = vc4_hdmi->connector.dev;
 	struct drm_connector_state *conn_state = &vc4_state->base;
-	unsigned int max_bpc = clamp_t(unsigned int, conn_state->max_bpc, 8, 12);
+	unsigned int max_bpc = clamp_t(unsigned int, conn_state->max_requested_bpc, 8, 12);
 	unsigned int bpc;
 	int ret;
 


pelwell:
This PR looks very plausible and I'm happy to take it on trust, but does anyone more knowledgeable about DRM have any feedback?

mripard:
I'm not entirely sure this is the proper fix. max_bpc is already documented as being max_requested_bpc minus the limitations expressed by the EDIDs:

https://github.com/raspberrypi/linux/blob/rpi-5.15.y/include/drm/drm_connector.h#L784

And max_bpc is assigned by the framework through drm_atomic_connector_check(), so it should be there, and correct. If it isn't, then I think that's the bug we should be fixing instead of this one which is, I believe, a symptom but not the root cause.

What is the use case that triggers that bug?

HiassofT:
The use case is being able to get YCbCr 4:2:2 on a HDMI monitor with 4:2:2 support but without 10/12 bit deep color support in auto mode when the max_bpc connector property is set to 12.

The max_bpc display info in drm only accounts for deep color modes, so max_bpc will be clamped to 8 which means we will never consider the 12bit 4:2:2 mode in this case.

As we are doing our own 10/12 bit and 4:2:2 support checks and consider 4:2:2 to be a 12-bit mode we can't use the clamped max_bpc.

I fear changing the clamping so that 4:2:2 support on HDMI monitors would set max_bpc in display info might have unwanted side effects on other display drivers as then 10/12 bit 4:4.4 deep color modes might also get selected.

I''m open to rework 4:4:4/4:2:2 selection in auto mode but it's tricky, and I think using max_requested_bpc is the best solution for now.

pelwell:
I saw your concerns, @mripard - should I merge this until @HiassofT is able to rework the 4:4:4/4:2:2 selection?

mripard:
> The use case is being able to get YCbCr 4:2:2 on a HDMI monitor with 4:2:2 support but without 10/12 bit deep color support in auto mode when the max_bpc connector property is set to 12.

Sorry, it wasn't really clear but what I meant was what sequence of commits are you doing to end up bypassing the call to atomic_check?

The goal itself makes sense to me.

> The max_bpc display info in drm only accounts for deep color modes, so max_bpc will be clamped to 8 which means we will never consider the 12bit 4:2:2 mode in this case.

That's what we should be fixing imo, the doc indicates that driver should be using max_bpc, and most drivers do so it looks like what you just described is the root cause.

> As we are doing our own 10/12 bit and 4:2:2 support checks and consider 4:2:2 to be a 12-bit mode we can't use the clamped max_bpc.
> 
> I fear changing the clamping so that 4:2:2 support on HDMI monitors would set max_bpc in display info might have unwanted side effects on other display drivers as then 10/12 bit 4:4.4 deep color modes might also get selected.

Right... Yeah, you're right I don't think there's any other way that wouldn't imply fixing half the other drivers.

Sorry for the noise. 

Could you still add a comment next to your change to explain what you just explained?

HiassofT:
I've changed the commit message to explain the 4:2:2 / deep color clamping issue and dropped the "not up to date" part.

When looking at the AMD/Intel driver the commits switching to max_requested_bpc mentioned potential problems with non-atomic commits and max_bpc containing old values and initially I thought this is what happened on RPi as well.

But on a second thought and glance through the code I'm no longer sure that's actually the case - the code looks sane to me. I might still be missing something and RPi could also have the issue though.

PR #5323, opened 2023-01-16
overlays: add BNO055 i2c and enable build in config

PolarStarWalker:
Add basic overlay and enable build for BNO055 sensor


pelwell:
This looks good at first glance. Is there any reason not to add it to `i2c-sensor` and get the benefit of support for other i2c buses?

Do you happen to know which addresses are legal?

PolarStarWalker:
> This looks good at first glance. Is there any reason not to add it to `i2c-sensor` and get the benefit of support for other i2c buses?
> 
> Do you happen to know which addresses are legal?

As I understand sensors in `i2c-sensor` are simple devices that have simple overlay only with address field. But according to [bno055.yml](https://github.com/raspberrypi/linux/blob/rpi-6.1.y/Documentation/devicetree/bindings/iio/imu/bosch%2Cbno055.yaml), the driver also have fields like `reset-gpios` and `clocks`. I don't need this functional, so don't include it, but maybe in the future someone may need it's.

According to bosch datasheet the default address of bno055 is 0x29, but it cat be changed to 0x28

pelwell:
I've folded bno055 into the i2c-sensor overlay, adding a `reset_pin` parameter. I can't see the `clocks` property being used.

See https://github.com/raspberrypi/linux/commit/2510d79b0a41095ef4a8b3c59e4b6d946c65da56 (until the next rebase).

PolarStarWalker:
Thank you!

PR #5326, opened 2023-01-18
.github/workflows: Correct kernel builds artifacts

pelwell:
Modify the kernel build workflow to create artifacts with the correct names and structure, both as an example of what we expect and in case anyone wants to use the output.

### DIFF
--- a/.github/workflows/kernel-build.yml
+++ b/.github/workflows/kernel-build.yml
@@ -37,14 +37,15 @@ jobs:
     - name: Build kernel
       run: |
         mkdir ${{github.workspace}}/build
-        make ARCH=arm KERNEL=kernel CROSS_COMPILE=arm-linux-gnueabihf- O=${{github.workspace}}/build bcm2835_defconfig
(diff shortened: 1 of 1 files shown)


pelwell:
This PR currently includes the reversion of a recent commit in order to trigger the build workflow; simply changing the workflow file is not sufficient. The reversion will be dropped before merging.

pelwell:
I think this is good to go now. Can you cast your eye over it, @6by9?

6by9:
Looks reasonable.

There was a comment made on the initial workflows to use the matrix stuff, but it's not bubbled up the priorities again yet.
https://github.com/nbuchwitz/actions-examples/blob/main/.github/workflows/test1.yml was given as the example that removed almost all the duplication.

pelwell:
It does look better after deduplication (I met @nbuchwitz recently), but I'll save that for another PR.

pelwell:
Matrix build PR here: https://github.com/raspberrypi/linux/pull/5333

PR #5327, opened 2023-01-18
drm/vc4: hvs: Defer dlist slots deallocation

popcornmix:
During normal operations, the cursor position update is done through an asynchronous plane update, which on the vc4 driver basically just modifies the right dlist word to move the plane to the new coordinates.

However, when we have the overscan margins setup, we fall back to a regular commit when we are next to the edges. And since that commit happens to be on a cursor plane, it's considered a legacy cursor update by KMS.

The main difference it makes is that it won't wait for its completion (ie, next vblank) before returning. This means if we have multiple commits happening in rapid succession, we can have several of them happening before the next vblank.

In parallel, our dlist allocation is tied to a CRTC state, and each time we do a commit we end up with a new CRTC state, with the previous one being freed. This means that we free our previous dlist entry (but don't clear it though) every time a new one is being committed.

Now, if we were to have two commits happening before the next vblank, we could end up freeing reusing the same dlist entries before the next vblank.

Indeed, we would start from an initial state taking, for example, the dlist entries 10 to 20, then start a commit taking the entries 20 to 30 and setting the dlist pointer to 20, and freeing the dlist entries 10 to
20. However, since we haven't reach vblank yet, the HVS is still using the entries 10 to 20.

If we were to make a new commit now, chances are the allocator are going to give the 10 to 20 entries back, and we would change their content to match the new state. If vblank hasn't happened yet, we just corrupted the active dlist entries.

A first attempt to solve this was made by creating an intermediate dlist buffer to store the current (ie, as of the last commit) dlist content, that we would update each time the HVS is done with a frame. However, if the interrupt handler missed the vblank window, we would end up copying our intermediate dlist to the hardware one during the composition, essentially creating the same issue.

Since making sure that our interrupt handler runs within a fixed, constrained, time window would require to make Linux a real-time kernel, this seems a bit out of scope.

Instead, we can work around our original issue by keeping the dlist slots allocation longer. That way, we won't reuse a dlist slot while it's still in flight. In order to achieve this, instead of freeing the dlist slot when its associated CRTC state is destroyed, we'll queue it in a list.

A naive implementation would free the buffers in that queue when we get our end of frame interrupt. However, there's still a race since, just like in the shadow dlist case, we don't control when the handler for that interrupt is going to run. Thus, we can end up with a commit adding an old dlist allocation to our queue during the window between our actual interrupt and when our handler will run. And since that buffer is still being used for the composition of the current frame, we can't free it right away, exposing us to the original bug.

Fortunately for us, the hardware provides a frame counter that is increased each time the first line of a frame is being generated. Associating the frame counter the image is supposed to go away to the allocation, and then only deallocate buffers that have a counter below or equal to the one we see when the deallocation code should prevent the above race from occuring.

### DIFF
--- a/drivers/gpu/drm/vc4/vc4_crtc.c
+++ b/drivers/gpu/drm/vc4/vc4_crtc.c
@@ -1052,14 +1052,8 @@ void vc4_crtc_destroy_state(struct drm_crtc *crtc,
 	struct vc4_dev *vc4 = to_vc4_dev(crtc->dev);
 	struct vc4_crtc_state *vc4_state = to_vc4_crtc_state(state);
 
-	if (drm_mm_node_allocated(&vc4_state->mm)) {
-		unsigned long flags;
-
-		spin_lock_irqsave(&vc4->hvs->mm_lock, flags);
-		drm_mm_remove_node(&vc4_state->mm);
-		spin_unlock_irqrestore(&vc4->hvs->mm_lock, flags);
-
-	}
+	vc4_hvs_mark_dlist_entry_stale(vc4->hvs, vc4_state->mm);
+	vc4_state->mm = NULL;
 
 	drm_atomic_helper_crtc_destroy_state(crtc, state);
 }
--- a/drivers/gpu/drm/vc4/vc4_drv.h
+++ b/drivers/gpu/drm/vc4/vc4_drv.h
@@ -340,6 +340,9 @@ struct vc4_hvs {
 	struct drm_mm lbm_mm;
 	spinlock_t mm_lock;
 
+	struct list_head stale_dlist_entries;
+	struct work_struct free_dlist_work;
+
 	struct drm_mm_node mitchell_netravali_filter;
 
 	struct debugfs_regset32 regset;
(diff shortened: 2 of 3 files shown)


popcornmix:
A while back we had an easily reproducible graphics glitch with the cursor moving between the edges of two displays.
https://github.com/raspberrypi/linux/pull/4895 was the fix.

Upstream suggested removing the legacy cursor support, and we followed that including a revert here: https://github.com/raspberrypi/linux/pull/4971

That avoided the easily reproducible case.

But there's been a number of glitches seen since then that don't involve the cursor.
Typically they appear as multicoloured pixels that occupy the bottom part (up to 20%) of screen.

I've been seeing it in kodi when playing video. It affects both hw and sw decoded video.
It can also occur in kodi's gui (when library scan busy spinner is updating is one case).
It also occurs in RpiOS when running mutter (typically with something animating like a youtube video, windowed in browser)

Some reports:
https://forum.kodi.tv/showthread.php?tid=369090
https://forums.raspberrypi.com/viewtopic.php?p=2068441#p2068441
https://forums.raspberrypi.com/viewtopic.php?t=340078

Here's one I caught on kodi's gui.
kodi_glitch

I've resurrected Maxime's cursor fix and I've been unable to reproduce any glitches since. I think it is still wanted.
I believe we are freeing dlist allocation while it is still on screen. If it gets reused, the hvs processes a corrupt dlist
(typically towards the end of a display).

popcornmix:
@mripard do you think reintroducing this is correct?

I've had a couple of positive reports from LibreELEC slack.

pelwell:
I'm happy to merge on the basis of your testing alone, whether or not it is ultimately the best solution.

popcornmix:
Lets get it in for wider testing.

popcornmix:
I've picked it to 6.0 and 6.1 trees (it needed slightly different fixups).

pelwell:
Did you notice the KUnit test failures? https://github.com/raspberrypi/linux/actions/runs/3988799056
They seem to start with this commit.

popcornmix:
@mripard can you see why kunit test fails with this commit (on 6.1 and 6.2).
It seems to boot and work fine on my Pi4.

mripard:
Sorry for the delay

> @mripard do you think reintroducing this is correct?

For the record, yes, if it still fixes some issue then it makes sense to merge it.

I've had a look at the kunit crash and we now access a register (the frame count) as part of the CRTC state destruction. Since kunit runs with a mock device, it's a immediate failure condition. It's not clear to me what the best solution is at the moment, but I'm on it.

pelwell:
Many thanks, Maxime.

PR #5328, opened 2023-01-19
media: bcm2835-v4l2-codec: Enable selection ioctl for ISP

jc-kynesim:
The ISP cases do nothing. Remove the break that separates them from the deinterlace case so they now do the same as deinterlace. This enables simple width & height setting, but does not enable setting left and top coordinates.

### DIFF
--- a/drivers/staging/vc04_services/bcm2835-codec/bcm2835-v4l2-codec.c
+++ b/drivers/staging/vc04_services/bcm2835-codec/bcm2835-v4l2-codec.c
@@ -1855,7 +1855,6 @@ static int vidioc_g_selection(struct file *file, void *priv,
 		}
 		break;
 	case ISP:
-		break;
 	case DEINTERLACE:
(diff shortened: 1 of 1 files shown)


jc-kynesim:
@6by9 I can't think of any reason why the ISP shouldn't have at least the selection capabilities of the deinterlacer - can you? A better patch would enable left & top as well, but this fixes bottom line corruption I was seeing when testing my ffmpeg v4l2 scaler.

6by9:
LGTM.
This is only the simple ISP instance (/dev/video12), so it won't affect libcamera anyway.

jc-kynesim:
I've got a failing test - can anyone give me a clue what is bust?

6by9:
> I've got a failing test - can anyone give me a clue what is bust?

Looks like an internal Github workflow issue as it was failing to upload the resulting tarball. I've had a couple of those today.
I've kicked that one test off again.

PR #5330, opened 2023-01-20
6.2 fixes for imx290, unicam, and vcsm-cma

6by9:
popcornmix had dropped imx290 changes due to significant rework upstream, and further changes are queued for 6.3 so have been backported.
All downstream work now reapplied (and I will upstream those patches).

Minor tweak to unicam for raw16 formats.

Fix for vcsm-cma where it was detaching from the wrong dmabuf, which with the dma-buf locking changes of 6.2 was hitting a WARN for every frame when running libcamera.

### DIFF
--- a/drivers/media/i2c/imx290.c
+++ b/drivers/media/i2c/imx290.c
@@ -512,28 +512,21 @@ static int imx290_set_register_array(struct imx290 *imx290,
 
 static int imx290_set_data_lanes(struct imx290 *imx290)
 {
-	int ret = 0, laneval, frsel;
+	int ret = 0;
+	u32 frsel;
 
 	switch (imx290->nlanes) {
 	case 2:
-		laneval = 0x01;
+	default:
 		frsel = 0x02;
 		break;
 	case 4:
-		laneval = 0x03;
 		frsel = 0x01;
 		break;
-	default:
-		/*
-		 * We should never hit this since the data lane count is
-		 * validated in probe itself
-		 */
-		dev_err(imx290->dev, "Lane configuration not supported\n");
-		return -EINVAL;
 	}
 
-	imx290_write(imx290, IMX290_PHY_LANE_NUM, laneval, &ret);
-	imx290_write(imx290, IMX290_CSI_LANE_MODE, laneval, &ret);
+	imx290_write(imx290, IMX290_PHY_LANE_NUM, imx290->nlanes - 1, &ret);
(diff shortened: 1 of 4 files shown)


pelwell:
Is there a reason for including both "media: i2c: imx290: Use V4L2 subdev active" and its reversion (immediately afterwards)? While we're still rebasing this will probably disappear rapidly.

6by9:
> Is there a reason for including both "media: i2c: imx290: Use V4L2 subdev active" and its reversion (immediately afterwards)?

My error. I'd picked up the wrong patch initially which had a build error, reverted it and then applied the correct one, but forgot to remove the intermediate step.
The second version of that patch is the desired one. I'll drop the two erroneous ones.

drivers/media/i2c/imx290.c
```diff
@@ -107,6 +107,7 @@
```
pelwell, line 110: The commit message isn't the clearest.

drivers/staging/vc04_services/vc-sm-cma/vc_sm.c
```diff
@@ -255,7 +255,7 @@ static void vc_sm_clean_up_dmabuf(struct vc_sm_buffer *buffer)
 		buffer->import.sgt = NULL;
```
pelwell, line 255: Commit message: "adding checking" -> "added checking"

pelwell: That's in https://github.com/raspberrypi/linux/pull/5330/commits/cd321f8abac2547f63f8b6f86259901613bbc2e9

drivers/media/i2c/imx290.c
pelwell: That's in https://github.com/raspberrypi/linux/pull/5330/commits/44e09a8aa799689f0f690ab12ab831fd5c749f99

6by9: Reworded to hopefully be clearer.

drivers/staging/vc04_services/vc-sm-cma/vc_sm.c
6by9: Done

6by9:
Sorry, I hadn't meant to include "media: v4l2-core: Extend dev_debug logging to subdevices" in the PR. I've dropped it.

PR #5333, opened 2023-01-24
.github/workflows: Switch to a matrix build

pelwell:
Remove the per-build duplication by putting build parameters in a matrix.

### DIFF
--- a/.github/workflows/kernel-build.yml
+++ b/.github/workflows/kernel-build.yml
@@ -15,9 +15,40 @@ env:
   NUM_JOBS: 3
 
 jobs:
-  build-bcm2835:
-
+  build:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


pelwell:
This PR won't trigger a build because the workflow explicitly ignores changes to the workflow file, but you can see the results of running a build on exactly the same tree (except with the previous commit reverted to force a build) here: https://github.com/pelwell/linux/actions/runs/3994747733

pelwell:
Any objections?

popcornmix:
+59 / - 234 lines sounds good to me.

pelwell:
I've eyeballed the tar creation logs and they _look_ OK, but obviously I'll fix any issues found down the line (which will be quicker and less error-prone thanks this).

PR #5337, opened 2023-01-26
configs: add CONFIG_DM_INTEGRITY=m

ep69:
This is necessary for DM_CRYPT volumes with integrity protection.


pelwell:
The PR looks OK in itself. A Signed-off-by: is always welcomed, but config changes like this soon get squashed so it matters less.

Can you say a bit about the integrity protection option, and why it is desirable for Pi users?

ep69:
@pelwell Thanks for looking! This is my firs PR so feel free to point out any mistakes I made.
About integrity protection: In general, most of current software (such as md-raid) is not able to detect silent data corruption (good read here: https://securitypitfalls.wordpress.com/2018/05/08/raid-doesnt-work/). This is not fortunate for data storage solutions built on RPi. DM_INTEGRITY is able to either add a layer with block-device level checksums or do this together with DM_CRYPT encryption.

pelwell:
You now should be able to install an early build including the `dm-integrity` module using a new feature of `rpi-update`:
```
$ sudo rpi-update pulls/5337
```
After a reboot you will be running the latest 5.15 kernel.

PR #5340, opened 2023-01-30
media: bcm2835-v4l2-codec: Add profile & level ctrls to decode

jc-kynesim:
In order to support discovery of what profile & levels are supported by stateful decoders implement the profile and level controls where they are defined by V4L2.

### DIFF
--- a/drivers/staging/vc04_services/bcm2835-codec/bcm2835-v4l2-codec.c
+++ b/drivers/staging/vc04_services/bcm2835-codec/bcm2835-v4l2-codec.c
@@ -3193,6 +3193,89 @@ static int queue_init(void *priv, struct vb2_queue *src_vq,
 	return vb2_queue_init(dst_vq);
 }
 
+static void dec_add_profile_ctrls(struct bcm2835_codec_dev *const dev,
+				  struct v4l2_ctrl_handler *const hdl)
+{
+	unsigned int i;
+	const struct bcm2835_codec_fmt_list *const list = &dev->supported_fmts[0];
+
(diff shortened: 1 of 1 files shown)


jc-kynesim:
I know a couple of lines are a bit long but I can't see how to rework this without making it noticeably less legible. (I've already checked this is OK with 6by9.)

drivers/staging/vc04_services/bcm2835-codec/bcm2835-v4l2-codec.c
```diff
@@ -3332,11 +3415,12 @@ static int bcm2835_codec_open(struct file *file)
 	break;
 	case DECODE:
 	{
-		v4l2_ctrl_handler_init(hdl, 1);
+		v4l2_ctrl_handler_init(hdl, 1 + dev->supported_fmts[0].num_entries * 2);
```
6by9, line 3418: This is a little annoying as there are formats that won't add the associated controls, but such is life other than a loop to return the number of controls, and a second loop that then creates the controls.

6by9 (approved):
Some ugliness can't be avoided at times without obfuscating the code.

drivers/staging/vc04_services/bcm2835-codec/bcm2835-v4l2-codec.c
jc-kynesim: I thought about that but decided that there was nowhere near enough gain. The added code size would probably be bigger than the wasted memory.

PR #5341, opened 2023-01-31
drm/vc4_plane: Add support for YUV444 formats

popcornmix:
Support displaying DRM_FORMAT_YUV444 and DRM_FORMAT_YVU444 formats. Tested with kmstest and kodi. e.g.

kmstest -r 1920x1080@60 -f 400x300-YU24

Note: without the shift of width, only half the chroma is fetched, resulting in correct left half of image and corrupt colours on right half.

The increase in width shouldn't affect fetching of Y data, as the hardware will clamp at dest width.

### DIFF
--- a/drivers/gpu/drm/vc4/vc4_plane.c
+++ b/drivers/gpu/drm/vc4/vc4_plane.c
@@ -108,6 +108,18 @@ static const struct hvs_format {
 		.pixel_order_hvs5 = HVS_PIXEL_ORDER_XYCRCB,
 	},
 	{
+		.drm = DRM_FORMAT_YUV444,
+		.hvs = HVS_PIXEL_FORMAT_YCBCR_YUV422_3PLANE,
+		.pixel_order = HVS_PIXEL_ORDER_XYCBCR,
+		.pixel_order_hvs5 = HVS_PIXEL_ORDER_XYCBCR,
+	},
+	{
+		.drm = DRM_FORMAT_YVU444,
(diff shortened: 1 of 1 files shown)


popcornmix:
@6by9 as discussed. Does this look okay?

6by9:
The width change doesn't mess up the luma? I'll believe you.

Has this been tested on a Pi0-3?

popcornmix:
> The width change doesn't mess up the luma? I'll believe you.

kmstest says no. I think you can always set source width/height larger with no visible effect (it will stop when it runs out of dest pixels, consuming source pixels at the scaled rate).

> Has this been tested on a Pi0-3?

Yes - it works on Pi3.

6by9:
Thinking it through, controlling the amount of each component read is likely why there is a YUV420 and YUV422 pixel format in the HVS dlist.

That probably means it's overreading the luma and therefore wasting SDRAM bandwidth, but for the very few use cases which want to use YUV4:4:4, I think it can be overlooked. We're not expecting 4k YUV444 decode from anywhere.

My only complaint is from checkpatch
```
WARNING: line length of 105 exceeds 100 columns
#46: FILE: drivers/gpu/drm/vc4/vc4_plane.c:1130:
+	/* for YUV444 hardware wants double the width, otherwise it doesn't fetch full width of chroma */
```

popcornmix:
Updated the comment to be multi-line

6by9:
LGTM.
(I'm not going to fuss over a fullstop at the end of the comment!)

PR #5350, opened 2023-02-15
Add optional regulator support for ak7375 driver

ArduCAM:


pelwell:
On the whole this PR looks plausible, but there are a few issues:
1. It's targeting 5.15, while we're already close to switching to 6.1 for our main releases.
2. The ak7375 driver is an unmodified upstream file, for which we don't want to host downstream patches without a very good reason.

Item 2 in particular should not be a surprise to you.

ArduCAM:
Hi @pelwell 

Thank you for your reply,
For the second one, I understand your approach. At present, we mainly want to use the ak7375 driver on the Raspberry Pi with the imx519 camera. We will also submit the patch to the upstream, but the process of submitting the patch to the upstream is slow for us. Since our main use is on the Raspberry Pi and we want to get it merged as soon as possible, that's why we made this PR, is this acceptable for you guys?

6by9:
Checking the upstream repos I see someone has [done the job for you](https://patchwork.linuxtv.org/project/linux-media/patch/20221209143741.214242-4-y.oudjana@protonmail.com/), and it should be in 6.3 (it's already in the [linux-media master branch](https://git.linuxtv.org/media_tree.git/log/drivers/media/i2c/ak7375.c)).

It doesn't include using the regulator framework callback to power up the VCM with the sensor, but libcamera doesn't need that as it will open the lens subdev for itself.

If you wished to backport that series to 6.1, with the first line of each commit message being "Commit <hash> upstream" to make it quickly identifiable as a backport, then that would be acceptable.

You do appear to be missing an update to the imx519 overlay to reference the VCM - there's no point in having the driver with no users of it.

pelwell:
Closing for the reasons given above.

ArduCAM:
Hi @6by9 
Thank you for your information,
We saw a patch for ak7375 submitted earlier, but it hadn't been successfully merged at the time.
I will resubmit a PR for "Commit upstream" and imx519 overlay, thanks again.

PR #5351, opened 2023-02-17
Commit upstream ak7375

ArduCAM:
Upstream patch:
https://patchwork.linuxtv.org/project/linux-media/patch/20221209143741.214242-4-y.oudjana@protonmail.com/

PR #5355, opened 2023-02-20
dtoverlays: Add VCM option to imx519

ArduCAM:

vcm related PR: #5351

### DIFF
--- a/arch/arm/boot/dts/overlays/README
+++ b/arch/arm/boot/dts/overlays/README
@@ -2384,6 +2384,8 @@ Params: rotation                Mounting rotation of the camera sensor (0 or
                                 configuring the sensor (default on)
         cam0                    Adopt the default configuration for CAM0 on a
                                 Compute Module (CSI0, i2c_vc, and cam0_reg).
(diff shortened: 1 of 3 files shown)


ArduCAM:
#5351

pelwell:
@6by9 is the expert here, but in my opinion adding the VCM declaration into `imx519.dtsi`, even if it is disabled, breaks the layering. The imx219 support, which I'm sure you are familar with, puts it in the parent overlay `imx219-overlay.dts`.

6by9:
This takes the same approach as taken by imx708, and that's the only sensor we really care about with VCM (ov5647 and imx219 with VCM are 3rd party products). Having the VCM in the sensor dtsi makes life easier in camera-mux-2port and camera-mux-4port.

Potentially we could have generic VCM dtsi files for each driver chip that get pulled in, but I'm not convinced it is worth it as currently only ov5647 and imx219  use the same VCM.
It would keep the layering clean though, and potentially we could see variants of the same sensor with different VCMs.

pelwell:
I'm going to take that as approval.

PR #5358, opened 2023-02-24
vc04_services: bcm2835_codec: Ignore READ_ONLY ctrls in s_ctrl

6by9:
In adding the MPEG2/MPEG4/H264 level and profile controls to the decoder, they weren't declared as read-only, nor handlers added to bcm2835_codec_s_ctrl. That resulted in an error message "Invalid control" being logged every time v4l2_ctrl_handler_setup was called from bcm2835_codec_create_component.

Define those controls as read only, and exit early from s_ctrl on read only controls.

Fixes: 18d8d0236d45 ("media: bcm2835-v4l2-codec: Add profile & level ctrls to decode")

I'm aware that this has 3 lines that exceed 100 columns, but I think these are reasonable ones for ignoring the coding standard and checkpatch. The alternative is that to create a define for the valid values bitmask and use that, but it's less readable.

### DIFF
--- a/drivers/staging/vc04_services/bcm2835-codec/bcm2835-v4l2-codec.c
+++ b/drivers/staging/vc04_services/bcm2835-codec/bcm2835-v4l2-codec.c
@@ -2222,6 +2222,9 @@ static int bcm2835_codec_s_ctrl(struct v4l2_ctrl *ctrl)
 		container_of(ctrl->handler, struct bcm2835_codec_ctx, hdl);
 	int ret = 0;
 
+	if (ctrl->flags & V4L2_CTRL_FLAG_READ_ONLY)
+		return 0;
+
 	switch (ctrl->id) {
 	case V4L2_CID_MPEG_VIDEO_BITRATE:
 		ctx->bitrate = ctrl->val;
@@ -2374,7 +2377,7 @@ static int bcm2835_codec_s_ctrl(struct v4l2_ctrl *ctrl)
 		break;
 
 	default:
-		v4l2_err(&ctx->dev->v4l2_dev, "Invalid control\n");
+		v4l2_err(&ctx->dev->v4l2_dev, "Invalid control %08x\n", ctrl->id);
 		return -EINVAL;
 	}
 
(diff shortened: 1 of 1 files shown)


popcornmix:
The 5.15 kernel is unlikely to be used by any future rpi-update/apt builds.
This would be more useful as a 6.1 PR.

6by9:
> The 5.15 kernel is unlikely to be used by any future rpi-update/apt builds.
This would be more useful as a 6.1 PR.

Happy to make it so, it's just that the driver is currently spamming the kernel logs with the error message after #5340

drivers/staging/vc04_services/bcm2835-codec/bcm2835-v4l2-codec.c
```diff
@@ -2222,6 +2222,9 @@ static int bcm2835_codec_s_ctrl(struct v4l2_ctrl *ctrl)
 		container_of(ctrl->handler, struct bcm2835_codec_ctx, hdl);
 	int ret = 0;
 
+	if (ctrl->flags & V4L2_CTRL_FLAG_READ_ONLY)
+		return 0;
```
soheil, line 2226: Good job defining controls as read-only and exiting early from s_ctrl on read-only controls.

soheil (commented):
This is an interesting code change. You set the MPEG2/MPEG4/H264 level and profile controls in the decoder that were not declared as read-only, and as handlers were not added to bcm2835_codec_s_ctrl, this resulted in an error message "Invalid control" being logged each time v4l2_ctrl_handler_setup was called from bcm2835_codec_create_component. 

To fix this error message, you defined the controls as read-only and made changes to the s_ctrl function so that it exits early from s_ctrl on read-only controls.

Your code looks good to me, and I am happy with the changes made to the code. I understand that this has 3 lines that exceed 100 columns, but I think that these are reasonable ones for ignoring the coding standard and checkpatch.

Thank you for making this changes.

6by9:
We're focusing on 6.1 now, hence the equivalent there (#5359) is merged. I'm going to close this one.

Indeed I ignored the long line warning - these V4L2_CID_ defines are so long that it's almost impossible not to have long lines. The only other alternative was a ```#define``` for the valid bitmask, or making a big assumption with regard all the levels being contiguous bits in the mask, but that still doesn't help for the profiles.

JamesH65:
That last non-Pi  comment was 99.39% fake, as tested by the Chat GPT output detector!

pelwell:
Yup - https://github.com/soheil shows 13 pull requests in unrelated repos with a similar style, having started the day by forking https://github.com/openai/openai-cookbook.
@soheil You are on the way to being blocked.

PR #5361, opened 2023-02-27
dtoverlays: Add VCM option to Arducam64MP

ArduCAM:

### DIFF
--- a/arch/arm/boot/dts/overlays/README
+++ b/arch/arm/boot/dts/overlays/README
@@ -613,6 +613,8 @@ Params: rotation                Mounting rotation of the camera sensor (0 or
                                 configuring the sensor (default on)
         cam0                    Adopt the default configuration for CAM0 on a
                                 Compute Module (CSI0, i2c_vc, and cam0_reg).
+        vcm                     Select lens driver state. Default is enabled,
+                                but vcm=off will disable.
 
 
 Name:   arducam-pivariety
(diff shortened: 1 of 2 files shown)


ArduCAM:
Hi @6by9 
I want to add Arducam64MP to camera mux, but its vcm is the same as imx708, should I move vcm's dts to a separate dtsi?

6by9:
> I want to add Arducam64MP to camera mux, but its vcm is the same as imx708, should I move vcm's dts to a separate dtsi?

The dw9817 in ```dw9817@c``` is just a label. There's nothing stopping you having dw9817_imx708@c and dw9817_arducam64mp@c.

It would be possible to reduce the duplication, but as I noted in https://github.com/raspberrypi/linux/pull/5355#issuecomment-1438090886 I'm not sure it really makes any significant gain. This clash of node name doesn't make it any easier as in the camera mux case you'd either need to rename the node for each instance, or have it only defined once and then have multiple sensors linking to it.

ArduCAM:
> > I want to add Arducam64MP to camera mux, but its vcm is the same as imx708, should I move vcm's dts to a separate dtsi?
> 
> The dw9817 in `dw9817@c` is just a label. There's nothing stopping you having dw9817_imx708@c and dw9817_arducam64mp@c.
> 
> It would be possible to reduce the duplication, but as I noted in [#5355 (comment)](https://github.com/raspberrypi/linux/pull/5355#issuecomment-1438090886) I'm not sure it really makes any significant gain. This clash of node name doesn't make it any easier as in the camera mux case you'd either need to rename the node for each instance, or have it only defined once and then have multiple sensors linking to it.

Got it, just not sure if it's general behavior to do this.

ArduCAM:
I'm going to close this PR for now and reopen it when I'm done with the changes.

PR #5362, opened 2023-02-27
drivers: media: i2c: imx258: Update imx258 with macros

luigi311:
This brings in some of the changes that Megi has made to get the imx258 mainline driver to work on the Pinephone Pro. I have remade all their changes on top of your driver as it contains way more features than the current mainline driver. The hope is to get this validated and working with the raspberry pi as well and then upstream it so the imx258 will work on Pinephone Pro with minimal 3rd party patches since it seems like the PPP has trouble running in 1272mbps mode.

2104x1560 and 1048x780 seem to generate images with a few green lines going across them but it seems like that appears without only the PPP required changes and on megis kernel so shouldnt be a new issue with this PR.


luigi311:
Pinging @6by9 as they were the original developer of the RPI imx258 driver to make sure everything is still good and to make sure they can test this new driver on their system.

6by9:
Initial comments based on a quick read through:

We're about to switch to the rpi-6.1.y branch as standard, so ideally this PR wants to be against that. Unfortunately all the IMX258 patches have been squashed there which will make upstreaming more tricky. I'll try to make up a clean branch for upstream with these patches. The change for V4L2_MBUS_CSI2_NONCONTINUOUS_CLOCK is already made on 6.1.

Any additional GPIOs need to be documented in the dtbinding.

My datasheet for the sensor only lists 3 supply rails to the sensor - vana @ 2.7V, vdig @ 1.2V, and vif @ 1.8V. Where has this i2c @ 1.8V supply come from? Does it really belong to the sensor, or the I2C bus?

The debug g_register and s_register calls are really ugly. Personally I always just use something like ```i2ctransfer -f -y <bus> w3@0x1a 0x01 0x00 0x01```, or ```i2ctransfer -f -y <bus> w2@0x1a 0x01 0x00 r1@0x1a```. I doubt mainline will be very keen on adding those.

I haven't checked through the register set changes, but the swapping from hex to decimal for values would make more sense to be part of that patch rather than mixed into the split out of common registers.
All macros should really start IMX258_REG_ to identify them.
You have definitely introduced duplication in the register sets - COARSE_INTEG_TIME = IMX258_REG_EXPOSURE, FRM_LENGTH_LINES = IMX258_REG_VTS. Your last patch adds COARSE_INTEG_TIME and FRM_LENGTH_LINES into the mode register tables which is redundant when there are controls for them.

Arbitrarily dropping the link rate from 1267 to 1220Mbit/s is unlikely to be acceptable to mainline as it's effectively a regression. I'm expecting link rate to also be twice the link frequency, so you've now misdeclared that. There's nothing stopping you detecting at probe time what link frequency setting is configured via device tree, and switching registers based on that.
Is it really a CSI-2 bandwidth issue or pixel processing rate issue? If you increase IMX258_PPL_DEFAULT and HTS to increase the horizontal blanking, does that solve the problem without altering the clock setup? Implementing V4L2_CID_HBLANK is relatively simple, and then gives the control to userspace.

pelwell:
Are these additional features useful to us?

Any serious attempt at creating something upstreamable should probably start with rpi-6.2.y, handling any back-ports to rpi-6.1.y as additional commits.

6by9:
> Are these additional features useful to us?

Adding macro names to more registers is nice, but none of it adds any functionality on the Pi.

If my patches apply relatively cleanly to mainline, then I may send them upstream anyway - fewer downstream patches to maintain has benefits.

luigi311:
>  My datasheet for the sensor only lists 3 supply rails to the sensor - vana @ 2.7V, vdig @ 1.2V, and vif @ 1.8V. Where has this i2c @ 1.8V supply come from? Does it really belong to the sensor, or the I2C bus?

This doesnt seem to affect anything since my initial test with your driver didnt include it so i can remove it if it isnt in your datasheet.

> The debug g_register and s_register calls are really ugly. Personally I always just use something like i2ctransfer -f -y <bus> w3@0x1a 0x01 0x00 0x01, or i2ctransfer -f -y <bus> w2@0x1a 0x01 0x00 r1@0x1a. I doubt mainline will be very keen on adding those.

This can be omitted since its not too important atleast for what we require to be  upstreamed.


> Arbitrarily dropping the link rate from 1267 to 1220Mbit/s is unlikely to be acceptable to mainline

Testing with your original driver with only the necessary changes to get the driver to load was failing to produce an image at 4208x3120 and it didnt work until i dropped the rate down to 1220/1224.

> Your last patch adds COARSE_INTEG_TIME and FRM_LENGTH_LINES into the mode register tables which is redundant when there are controls for them.

I added those in since they existed in megi's  version of the mode controls though his was based on the mainline kernel which misses a lot of the controls. I can go ahead and remove.

luigi311:
Let me use all of your registry values for now and then i can take a look at the PPL and HTS changes you referenced to see if we can make 1267 work.

6by9:
Actually checking the datasheet, registers 0x342/3 have the comment
```
Set to 5352d. Any other value change require, please confirm with SONY
```
So implementing V4L2_CID_HBLANK may be possible, but results may be unpredictable.

luigi311:
That should be fine then, we can keep it how it is for now so we can atleast upstream these changes and bring the PPP to as few patches as possible for now and submit any more patches upstream afterwards. Im getting a bunch of merge conflicts rebasing to the 6.2 branch so let me just blow out everything and then add in what we are planning to add.

6by9:
Also be aware of the series at https://patchwork.linuxtv.org/project/linux-media/cover/20230117100603.51631-1-jacopo.mondi@ideasonboard.com/
Whilst it has Reviewed-by responses on most of the patches, it hasn't been merged and is likely to cause conflicts.

It's the minor annoyance with Linux media drivers that maintainers seem to batch merging patches up so you have limited visibility as to what has been accepted or not.

6by9:
Branch based on https://git.linuxtv.org/sailus/media_tree.git/tree/ with my patches pushed to https://github.com/6by9/linux/tree/linux-media-imx258

6by9:
Jacopo has just pinged Sakari as to why that series wasn't part of a pull for 6.3, so there may be movement on it.

luigi311:
Ok this now contains all the changes on a clean 6.2 base so we can start making changes now. I have removed the two duplicate regs and kept the 1267 mbps and kept your register values for everything else.

luigi311:
> Jacopo has just pinged Sakari as to why that series wasn't part of a pull for 6.3, so there may be movement on it.

Yes i have contacted Jacopo and let him know about this PR since you mentioned his patch request.

luigi311:
This now  has the dtbindings for the gpio.  I think this now has everything

PR #5363, opened 2023-02-28
dtoverlays: Add VCM option to Arducam64MP and Add  Arducam64MP support to camera mux overlays 

ArduCAM:


6by9:
Looks reasonable to me.

Issue #5210, opened 2022-10-17
Sitronix ST7701 changes for SPI displays need to be forward-ported to 6.1


6by9:
### Describe the bug

Upstream driver has changed significantly, so downstream changes need reworking.
Placeholder for myself so we don't forget.

### Steps to reproduce the behaviour

Patches can't be easily forward ported.

### Device (s)

Raspberry Pi 4 Mod. B

### System

rpi-6.1.y

pelwell:
See https://github.com/raspberrypi/linux/pull/5370 for a potential fix.

PR #5370, opened 2023-03-06
drm/panel/panel-sitronix-st7701: Support SPI config and RGB data

pelwell:
The ST7701 supports numerous different interface mechanisms for MIPI DSI, RGB, or SPI. The driver was only implementing DSI input, so add RGB parallel input with SPI configuration.

### DIFF
--- a/drivers/gpu/drm/panel/panel-sitronix-st7701.c
+++ b/drivers/gpu/drm/panel/panel-sitronix-st7701.c
@@ -7,15 +7,20 @@
 #include <drm/drm_mipi_dsi.h>
 #include <drm/drm_modes.h>
 #include <drm/drm_panel.h>
+#include <drm/drm_print.h>
 
 #include <linux/bitfield.h>
 #include <linux/gpio/consumer.h>
 #include <linux/delay.h>
+#include <linux/media-bus-format.h>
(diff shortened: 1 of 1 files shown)


pelwell:
@6by9 This commit got dropped in the move to 6.1, and had some merge conflicts (which may be why it got dropped).

Does this commit have a role in 6.1 (its compatible string is used by an overlay, so I think so), and do the changes look OK (it was b3782d3f0fd8 on 5.15)?

pelwell:
Ah - this fixes https://github.com/raspberrypi/linux/issues/5210.

6by9:
The mainline driver gained a load of abstractions for things like gamma tables. Those are all relevant for SPI configured DPI as well as DSI, so it would be nice to make use of them.
Making the macro magic with ```ST7701_DSI``` work with two different data types is more of a problem. It's probably easiest to make a st7701_spi_write take a ```u8``` array, and convert it internally to ```u16``` with bit 8 set on each data word.

Merge this for now, but leave #5210 open for making use of the mainline improvements.

jackoalan:
It looks like an SPI/DSI condition was missed in the 6.1 merge. Fix up for review here: #5484

PR #5372, opened 2023-03-08
imx327/290/462 backports to 6.2

6by9:
This patch set hasn't actually been merged upstream as yet, but I'm still hoping it'll hit 6.3.

Reverts all our downstream patches, applies the 2 upstream series, and then updates the overlays.


pelwell:
Thanks - looks good to me.

6by9:
Shall I backport to 6.1 as well whilst we're still rebasing and it's going to be LTS?

pelwell:
Yes, if you think it may save us from having to do so later when rebasing is not an option.

6by9:
Will do then.

PR #5373, opened 2023-03-08
Rpi workflows checkpatch

6by9:


6by9:
Sorry for the noise - meant to create this PR locally to my repo to test the workflow, but it pushed it to here instead.

PR #5375, opened 2023-03-08
Add workflow for checkpatch on PRs

6by9:
Sorry for the noise earlier.

This workflow can only be triggered on PR as it uses the Github REST API to retrieve specifically a "pull" list of commits.

We may also need to add "webispy/checkpatch-action@v9" under settings/actions/general/"allow specified actions and reusable workflows", or we fork it into our own repo for security.

The two patches to the rpi-panel-attiny are largely so that the workflow has something to test, but they will also want to be merged in due course.

### DIFF
--- a/.github/workflows/checkpatch.yml
+++ b/.github/workflows/checkpatch.yml
@@ -1,0 +1,18 @@
+name: Advisory checkpatch review
+on: [pull_request]
+
+jobs:
+  review:
+    name: checkpatch review
+    runs-on: ubuntu-latest
+    steps:
+    - name: 'Calculate PR commits + 1'
+      run: echo "PR_FETCH_DEPTH=$(( ${{ github.event.pull_request.commits }} + 1 ))" >> $GITHUB_ENV
+    - uses: actions/checkout@v3
+      with:
+        ref: ${{ github.event.pull_request.head.sha }}
+        fetch-depth: ${{ env.PR_FETCH_DEPTH }}
+    - name: Copy checkpatch.conf
+      run: cp ${{github.workspace}}/.github/workflows/ci_checkpatch.conf ${{github.workspace}}/.checkpatch.conf
+    - name: Run checkpatch review
+      uses: webispy/checkpatch-action@v9
--- a/.github/workflows/ci_checkpatch.conf
+++ b/.github/workflows/ci_checkpatch.conf
@@ -1,0 +1,4 @@
+--no-tree
+--ignore FILE_PATH_CHANGES
+--ignore GIT_COMMIT_ID
+--ignore SPDX_LICENSE_TAG


pelwell:
`workflows: Enable patch checking in CI workflows`?

6by9:
OK, too coarse at present as warnings are treated as errors, so the warnings for this workflow cause the failure:
```
2023-03-08T16:54:42.0508370Z [1;31mWARNING: A patch subject line should describe the change not the tool that found it[0m
2023-03-08T16:54:42.0733104Z #4:
2023-03-08T16:54:42.0733688Z Subject: [PATCH] workflows: We all love checkpatch, so add it to the CI
2023-03-08T16:54:42.0771746Z ##[error]WARNING: A patch subject line should describe the change not the tool that found it
2023-03-08T16:54:42.0782180Z 
2023-03-08T16:54:42.0782189Z 
2023-03-08T16:54:42.0782865Z [1;31mWARNING: added, moved or deleted file(s), does MAINTAINERS need updating?[0m
2023-03-08T16:54:42.0790362Z #18:
2023-03-08T16:54:42.0790629Z new file mode 100644
2023-03-08T16:54:42.0795511Z ##[error]WARNING: added, moved or deleted file(s), does MAINTAINERS need updating?
2023-03-08T16:54:42.0797772Z 
2023-03-08T16:54:42.0797782Z 
2023-03-08T16:54:42.0802122Z [1mtotal: 0 errors, 2 warnings, 16 lines checked[0m
2023-03-08T16:54:42.0803857Z [1;31m>> Failure[0m
```
The other two patches pass, and actually the warning over the reference to checkpatch is invalid in this case (I do describe what the patch is doing).

These tests aren't blocking merge, so perhaps we want to have warnings flagged such that we can make the call on whether we agree or not.

pelwell:
That's certainly going to be easier than teaching it what we do and don't allow. At least we know that commits should only be checked once, so that merging a PR with errors shouldn't cause further failures down the line.

Shall we?

6by9:
Added ignores for FILE_PATH_CHANGES (MAINTAINERS not updated), SPDX_LICENSE_TAG (which frequently hits dtoverlays), and UNKNOWN_COMMIT_ID (use of full commit hashes) as properties to the script.

I still expect this patch to trip up the use of "checkpatch" in the commit subject.

I forgot to rename the test name to "Advisory checkpatch review" - I will do that once I've seen the results from the current run.

6by9:
MAINTAINERS error now suppressed, along with the other 2 mentioned.
My patch still fails because I reference "checkpatch" in the subject, and that is probably how it should be.

Test renamed to include "Advisory".

If people are happy, then I can drop the two display regulator patches (or feel free to review here and we can push the lot - I have got a couple of others to add that allows the flips to be controlled, but it doesn't matter on order).

> We may also need to add "webispy/checkpatch-action@v9" under settings/actions/general/"allow specified actions and reusable workflows", or we fork it into our own repo for security.

Checked the settings, and it is currently permitted (kind of obvious as the test ran). Settings ought to be reviewed though.

6by9:
Dropped to just the one patch with the workflow.
All happy now?

PR #5377, opened 2023-03-10
Adapt downstream imx219 patches due to mainline updates

6by9:
The first two patches cherry-picked from rpi-5.15.y.
The other two have been recreated.

They should all apply cleanly to 6.1 and 6.3.

I am still compiling these, so they could yet have errors.


6by9:
A couple of false starts, but my local build has just completed OK.

popcornmix:
Also cherry-pickd to 6.1.

PR #5381, opened 2023-03-13
IMX708 link frequency changes

naushir:
~~This is to assist with https://github.com/raspberrypi/libcamera/issues/43.~~

Now ready for review.


naushir:
Slight change of plans.  Will use a overlay parameter instead of a v4l2 ctrl.

drivers/media/i2c/imx708.c
```diff
@@ -1738,6 +1784,15 @@ static int imx708_init_controls(struct imx708 *imx708)
 					       IMX708_INITIAL_PIXEL_RATE, 1,
 					       IMX708_INITIAL_PIXEL_RATE);
 
+	imx708->link_freq =
+		v4l2_ctrl_new_int_menu(ctrl_hdlr, &imx708_ctrl_ops,
+				       V4L2_CID_LINK_FREQ,
+				       ARRAY_SIZE(link_freqs) - 1,
+				       imx708->link_freq_idx,
+				       link_freqs);
```
6by9: Nit: As this is read-only, advertising the other values is just taunting the user.
Does it work if we use (watch me get the indirection wrong)
```
	v4l2_ctrl_new_int_menu(ctrl_hdlr, &imx708_ctrl_ops,
				       V4L2_CID_LINK_FREQ,
				       1,
				       0,
				       &link_freqs[imx708->link_freq_idx]);
```
So we have one entry with the correct value?

drivers/media/i2c/imx708.c
```diff
@@ -1738,6 +1784,15 @@ static int imx708_init_controls(struct imx708 *imx708)
 					       IMX708_INITIAL_PIXEL_RATE, 1,
 					       IMX708_INITIAL_PIXEL_RATE);
 
+	imx708->link_freq =
```
6by9: Less of a nit: we don't need to access ```imx708->link_freq``` outside of this function. 
Create a local ```struct v4l2_ctrl * ctrl```, ```ctrl = v4l2_ctrl_new_int_menu(...), if (ctrl) ctrl->flags |= V4L2_CTRL_FLAG_READ_ONLY```.

Or the ```imx708->link_freq->flags |= V4L2_CTRL_FLAG_READ_ONLY;``` can move down to beside the others at line 1866.

drivers/media/i2c/imx708.c
```diff
@@ -1474,6 +1510,16 @@ static int imx708_start_streaming(struct imx708 *imx708)
 		return ret;
 	}
 
+	/* Update the link frequency registers */
+	freq_regs = &link_freq_regs[imx708->link_freq->val];
```
6by9: ```freq_regs = &link_freq_regs[imx708->link_freq_idx];```

drivers/media/i2c/imx708.c
```diff
@@ -1835,11 +1890,15 @@ static void imx708_free_controls(struct imx708 *imx708)
 
 static int imx708_check_hwcfg(struct device *dev)
 {
+	struct i2c_client *client = to_i2c_client(dev);
+	struct v4l2_subdev *sd = i2c_get_clientdata(client);
+	struct imx708 *imx708 = to_imx708(sd);
```
6by9: That's a heck of a dance. Just pass ```struct imx708 *imx708``` in from ```imx708_probe```.

6by9:
Unrelated but just noticed in reviewing. 
```
	struct v4l2_ctrl *red_balance;
	struct v4l2_ctrl *blue_balance;
```
in ```struct imx708``` appear unused.

```struct v4l2_ctrl *notify_gains;``` I think is also redundant, as I believe you can use
```
		ret = imx708_write_reg(imx708, IMX708_REG_COLOUR_BALANCE_BLUE,
				       IMX708_REG_VALUE_16BIT,
				       ctrl->p_new.p_u32[0]);
		if (ret)
			break;
		ret = imx708_write_reg(imx708, IMX708_REG_COLOUR_BALANCE_RED,
				       IMX708_REG_VALUE_16BIT,
				       ctrl->p_new.p_u32[3]);
```
in ```imx708_set_ctrl```. It should come back to the same place.

drivers/media/i2c/imx708.c
naushir: I use ``imx708->link_freq->val``  as the index (as below), but I'll use ``imx708->link_freq_idx`` directly instead.

drivers/media/i2c/imx708.c
naushir: Yeah, I wasn't too sure what to do here.  I assumed advertising all available link speeds in the menu ctrl + setting it read only so you could only "ctrl select" one would be ok.  I'll change it over.

drivers/media/i2c/imx708.c
naushir: I can just pass in struct imx708 * into the function to make it cleaner.

naushir:
> Unrelated but just noticed in reviewing.
> 
> ```
> 	struct v4l2_ctrl *red_balance;
> 	struct v4l2_ctrl *blue_balance;
> ```
> 
> in `struct imx708` appear unused.
> 
> `struct v4l2_ctrl *notify_gains;` I think is also redundant, as I believe you can use
> 
> ```
> 		ret = imx708_write_reg(imx708, IMX708_REG_COLOUR_BALANCE_BLUE,
> 				       IMX708_REG_VALUE_16BIT,
> 				       ctrl->p_new.p_u32[0]);
> 		if (ret)
> 			break;
> 		ret = imx708_write_reg(imx708, IMX708_REG_COLOUR_BALANCE_RED,
> 				       IMX708_REG_VALUE_16BIT,
> 				       ctrl->p_new.p_u32[3]);
> ```
> 
> in `imx708_set_ctrl`. It should come back to the same place.

I'll fix these up in a separate commt.

naushir:
Updated PR with the changes.  I've had to remove the 444Mhz link frequency as it does not work correctly, even with a lower fps.  The next link frequency available 445.5Mhz doesn't work either.  I've added an option for 453Mhz instead, which does seem to be stable.

I've also added a 3rd patch to this PR to address the upstream review feedback comments.  I'll send upstream a new revision of the patch including all these changes once they are merged into our tree.

6by9:
Patches look good.

Didn't upstream want the regulator names to be in lower case? I seem to recall updating the bindings based on that.

Again for upstream, I thought I'd seen a comment over using a v4l2_ctrl cluster for HFLIP & VFLIP so that the s_ctrl only gets called the once instead of twice. It may have been with regard a different driver, but it does apply here too.

naushir:
> Didn't upstream want the regulator names to be in lower case? I seem to recall updating the bindings based on that.
> 
> Again for upstream, I thought I'd seen a comment over using a v4l2_ctrl cluster for HFLIP & VFLIP so that the s_ctrl only gets called the once instead of twice. It may have been with regard a different driver, but it does apply here too.

I don't recall those specific comments.  I'm following patchworks [here](https://patchwork.linuxtv.org/project/linux-media/patch/20230124150546.12876-3-naush@raspberrypi.com/).  But I can fix them up as well.

6by9:
Case of regulator names is on the dtbindings patch - https://patchwork.linuxtv.org/project/linux-media/patch/20230124150546.12876-2-naush@raspberrypi.com/#143941
The driver obviously has to match the binding.

Clusters was a comment from Laurent on imx290 - https://patchwork.linuxtv.org/project/linux-media/patch/20230131192016.3476937-12-dave.stevenson@raspberrypi.com/#144338
It is trivial to add - see https://git.linuxtv.org/sailus/media_tree.git/commit/drivers/media/i2c/imx290.c?id=924350ae03259b613a29a38c1e6f5b439eb9ce66

naushir:
Added 3 new patches on top for the dtb changes and h/v flip cluster control change.

6by9:
LGTM, and hopefully in reasonable shape for a v2 to mainline.
If only we had the multiplexed stream support to get rid of the multiple pads bit then we could actually use mainline!

pelwell:
1. There are a number of typos in the commit messages - "dtbinding"->"dt-bindings", "documenation"->"documentation", "cluser"->"cluster".
2. The first commit ("drivers: media: imx708: Increase usable link frequencies") and fifth commit ("drivers: media: imx708: Follow the documented devicetree labels") include driver _and_ overlay changes, which makes the usual DT squashing more awkward.
3. Checkpatch is complaining about the fourth commit changing both yaml and headers, but I can't see it (unless the MAINTAINERS file counts as a header).

Let me know when you've made whatever changes you are going to and I'll merge it.

naushir:
I think that should cover it all.  Let me know if I've missed anything!

Issue #5379, opened 2023-03-12
6.1.16 kernel broke wifi on RPi 3b+


iucoen:
### Describe the bug

After upgrading to the latest rpi-6.1.y branch, version 6.1.16 cannot connect to wifi for me anymore. I did a diff between 6.1.15 and 6.1.16 and was able to narrow it down to this commit: 22dfb21bf1cd876616d45cda1bc6daa89eec6747

After reverting that commit wifi connects fine. 

### Steps to reproduce the behaviour

Compile 6.1.16 kernel, install, reboot, and wifi can no longer connect.

### Device (s)

Raspberry Pi 3 Mod. B+

### System

OS is Gentoo Linux.
Kernel is 6.1.16-v8+

DaniPhii:
I experienced the same issue using Arch Linux ARM, downgraded kernel to 6.1.14 and Wi-Fi is working again.

iucoen:
Upstream confirmed the issue, patch here: https://lore.kernel.org/all/20230311141914.24444-1-marcan@marcan.st/

pelwell:
Thanks everyone - we'll cherry-pick the fix ASAP.

popcornmix:
We've bumped to 6.1.19 which includes the fix.
Latest rpi-update kernel contains this.

PR #5382, opened 2023-03-13
wifi: cfg80211: Partial revert "wifi: cfg80211: Fix use after free fo…

pelwell:
…r wext"

This reverts part of commit 015b8cc5e7c4 ("wifi: cfg80211: Fix use after free for wext")

This commit broke WPA offload by unconditionally clearing the crypto modes for non-WEP connections. Drop that part of the patch.

Fixes: 015b8cc5e7c4 ("wifi: cfg80211: Fix use after free for wext")
Cc: stable@vger.kernel.org
Link: https://lore.kernel.org/linux-wireless/ZAx0TWRBlGfv7pNl@kroah.com/T/#m11e6e0915ab8fa19ce8bc9695ab288c0fe018edf


pelwell:
See https://github.com/raspberrypi/linux/issues/5379 for background.

HiassofT:
It might be best to just update to 6.1.19 which contains the partial revert commit - 6.1.17 reverted two more commits from 6.1.16 which caused issues
https://cdn.kernel.org/pub/linux/kernel/v6.x/ChangeLog-6.1.17

seems 6.1.16 wasn't a too great version...

pelwell:
Yes, of course, if that's available.

popcornmix:
I've bumped to 6.1.19 which should include this fix. rpi-update contains this update.

PR #5384, opened 2023-03-14
Imx296

naushir:
This PR removes the downstream imx296 driver and replaces it with the upstream version.  Some additional features/fixes have also been added:

- H/V flip support.
- Sensor crop support.
- Fix gain delay to 1 frame.
- Disable binning on colour sensor variant.


6by9:
I'm assuming the two base commits are 
https://github.com/torvalds/linux/commit/d680dc5805745e68e3c8d769f40685707ca7fe4f and 
https://github.com/torvalds/linux/commit/cb33db2b6ccfe3ccc13347755ab3ef38691d59c3
which are both in 6.3, but they would seem to be the patches direct from the list as they don't have Mauro's SoB.
Normal approach would be to cherry-pick and add "Upstream commit <hash>" to the commit text.

I'm not reviewing those as they should be good from upstream. Review on the other minor mods to come....

naushir:
I actually pulled them from linux-media/master, I assume they are the same.  They seem to match the commits in rpi-6.3.y.

drivers/media/i2c/imx296.c
```diff
@@ -384,10 +384,35 @@ static const struct v4l2_ctrl_ops imx296_ctrl_ops = {
 	.s_ctrl = imx296_s_ctrl,
 };
 
+static void imx296_setup_hblank(struct imx296 *sensor, unsigned int width)
+{
+	/*
+	 * Horizontal blanking is controlled through the HMAX register, which
+	 * contains a line length in INCK clock units. The INCK frequency is
+	 * fixed to 74.25 MHz. The HMAX value is currently fixed to 1100,
```
6by9: I realise this is a move of code from the base driver, however INCK is not fixed at 74.25MHz. ```imx296_clk_params``` has entries for 37.125, 54, and 74.25MHz.

Use sensor->clk_params->freq instead of a hardcoded 74250000.
If I read "Readout drive modes" on page 44 of the datasheet, it lists the values for "Number of INCK in 1H" for each INCK. 
Seeing as I believe (from the overlay) we're using a 54MHz INCK, I'm slightly surprised that HMAX is correctly set when fixed at 1100.

drivers/media/i2c/imx296.c
```diff
@@ -597,6 +646,10 @@ static int imx296_stream_on(struct imx296 *sensor)
 	usleep_range(2000, 5000);
 	imx296_write(sensor, IMX296_CTRL0A, 0, &ret);
 
+	/* vflip and hflip cannot change during streaming */
+	__v4l2_ctrl_grab(sensor->vflip, 1);
```
6by9: Does changing flip create a corrupt frame? If not then we don't need to grab the control on a mono variant of the sensor.

drivers/media/i2c/imx296.c
```diff
@@ -427,6 +466,16 @@ static int imx296_ctrls_init(struct imx296 *sensor)
 			  V4L2_CID_ANALOGUE_GAIN, IMX296_GAIN_MIN,
 			  IMX296_GAIN_MAX, 1, IMX296_GAIN_MIN);
 
+	sensor->hflip = v4l2_ctrl_new_std(&sensor->ctrls, &imx296_ctrl_ops,
+					  V4L2_CID_HFLIP, 0, 1, 1, 0);
+	if (sensor->hflip)
```
6by9: ```if (sensor->hflip && !sensor->mono```
The mono sensor doesn't change the format.

drivers/media/i2c/imx296.c
```diff
@@ -427,6 +466,16 @@ static int imx296_ctrls_init(struct imx296 *sensor)
...
+		sensor->hflip->flags |= V4L2_CTRL_FLAG_MODIFY_LAYOUT;
+
+	sensor->vflip = v4l2_ctrl_new_std(&sensor->ctrls, &imx296_ctrl_ops,
+					  V4L2_CID_VFLIP, 0, 1, 1, 0);
+	if (sensor->vflip)
+		sensor->vflip->flags |= V4L2_CTRL_FLAG_MODIFY_LAYOUT;
```
6by9, line 478: Ditto

drivers/media/i2c/imx296.c
naushir: I believe this is correct.  The sensor internally multiplies any INCLK value to 74.25 MHz.  This is done by setting the INCLK value in one of the clk related registers. This can be verified by the frame time computations, which do give the right numbers, implying HMAX is at the right value.

drivers/media/i2c/imx296.c
naushir: Nothing is mentioned in the datasheet.  But I would imagine there may be some timing related issues doing it mid-frame.  To be safe, I would consider keeping this as-is.

drivers/media/i2c/imx296.c
naushir: Ack.

drivers/media/i2c/imx296.c
naushir: Ack.

drivers/media/i2c/imx296.c
6by9: I only raised it as I had the same question on [imx290](https://patchwork.linuxtv.org/project/linux-media/patch/20230131192016.3476937-12-dave.stevenson@raspberrypi.com/#144355) where it is explicitly stated in the datasheet that it will create a corrupt frame, however that is a rolling shutter sensor compared to global.

drivers/media/i2c/imx296.c
6by9: That would imply that the comment is misleading. Not your comment, so not a blocker for this PR though.
These patches really ought to be upstreamed though, hence me being pedantic.

naushir: I could reword the comment as:

```
	 * Horizontal blanking is controlled through the HMAX register, which
	 * contains a line length in contains a line length in units of an
	 * internal 74.25 MHz clock derived from the INCLK. The HMAX value is
	 * currently fixed to 1100, convert it to a number of pixels based on
	 * the nominal pixel rate.
```

naushir: > These patches really ought to be upstreamed though, hence me being pedantic.

Yes, no reason not to send these patches upstream.

naushir:
Added one more commit on top to limit the crop window to 96x88.  Below this and the sensor does not produce valid frames.

6by9:
Looks good to me.

I guess it's hard to tell if the sensor just doesn't stream, or if the interrupt rate is just too high for Unicam. It makes little difference in reality.

naushir:
It streams, only returns a frame with all 0 pixels :-)

pelwell:
You just need to rebase against the current rpi-6.1.y and I'll merge.

PR #5386, opened 2023-03-17
WIP: HDMI HPD debounce

6by9:
Debounce the HPD signal for those people with dodgy cables.

WIP as the second patch pushes the delay up to 2.5secs. 250ms should be sufficient in most cases and not cause a noticeable delay.

### DIFF
--- a/drivers/gpu/drm/vc4/vc4_hdmi.c
+++ b/drivers/gpu/drm/vc4/vc4_hdmi.c
@@ -2870,12 +2870,14 @@ static void vc4_hdmi_hpd_con_wq(struct work_struct *work)
 		drm_connector_helper_hpd_irq_event(connector);
 }
 
-#define HPD_DEBOUNCE_DELAY_MS 250
+#define HPD_DEBOUNCE_DELAY_MS 2500
 
 static irqreturn_t vc4_hdmi_hpd_irq_con_thread(int irq, void *priv)
 {
 	struct vc4_hdmi *vc4_hdmi = priv;
+	struct device *pdev = &vc4_hdmi->pdev->dev;
 
+	dev_err(pdev, "HPD IRQ status conn\n");
 	mod_delayed_work(system_wq, &vc4_hdmi->hpd_con_work,
 			 msecs_to_jiffies(HPD_DEBOUNCE_DELAY_MS));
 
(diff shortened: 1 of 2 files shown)


6by9:
@spl237 Once the CI workflows have done their thing, you should be able to do ```sudo rpi-update pulls/5386``` to update your system to a kernel with this patch.

drivers/gpu/drm/vc4/vc4_hdmi.c
```diff
@@ -2858,12 +2858,40 @@ static int vc4_hdmi_audio_init(struct vc4_hdmi *vc4_hdmi)
...
+						 struct vc4_hdmi,
+						 hpd_con_work);
+	struct drm_connector *connector = &vc4_hdmi->connector;
+	struct drm_device *dev = connector->dev;
+
+	if (dev && dev->registered)
+		drm_connector_helper_hpd_irq_event(connector);
+}
+
+#define HPD_DEBOUNCE_DELAY_MS 2500
```
awesomelewis2007, line 2873: I would move this to line 121 or somewhere near the top of the file

0x4248, line 2873: Please dont put defines here instead place it near the top of the file

```diff
- #define HPD_DEBOUNCE_DELAY_MS 2500
```

You can put it near line 121

```diff
#define HSM_MIN_CLOCK_FREQ	120000000
#define CEC_CLOCK_FREQ 40000

#define HDMI_14_MAX_TMDS_CLK   (340 * 1000 * 1000)
+ #define HPD_DEBOUNCE_DELAY_MS 2500
/* bit field to force hotplug detection. bit0 = HDMI0 */
static int force_hotplug = 0;
module_param(force_hotplug, int, 0644);

0x4248 (changes requested):

drivers/gpu/drm/vc4/vc4_hdmi.c
pelwell: Who are you?

0x4248: Why?

0x4248: I'm only suggesting changes.

PR #5392, opened 2023-03-21
ARM: dts: Disable headphone audio on CM, CM2 and CM3

juergh:
Similar to 79075960b723b3b0c15fad27bc555ab4385ae4ab. Not sure why this wasn't done back then.

### DIFF
--- a/arch/arm/boot/dts/bcm2708-rpi-cm.dts
+++ b/arch/arm/boot/dts/bcm2708-rpi-cm.dts
@@ -164,6 +164,7 @@ cam0_reg: &cam0_regulator {
 &audio {
 	pinctrl-names = "default";
 	pinctrl-0 = <&audio_pins>;
+	brcm,disable-headphones = <1>;
 };
 
 &hdmi {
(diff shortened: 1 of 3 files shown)


pelwell:
Hi Juerg,

Although I take your point about CMs not having headphone outputs, the baseboards they are attached to _could_. They would need to have a custom dt-blob.bin to select the PWM pins and enable the firmware's audio processing, but that could easily have been done. The DTBs have been like this since 5.4, and I'm reluctant to change them now - would an overlay or dtparam work for you?

juergh:
Hi Phil,
Ok, but that is true for the CM4 as well, no? Just wondering about the different behavior. And yes, an overlay or dtparam would work just fine. I'll see if I can come up with something when I find the time...

pelwell:
CM4 is different because, apart from specific interfaces such as I2C, only the GPIOs that normally appear on the 40-pin header are accessible to baseboards. As a result we know that there can never be a headphone output without using an overlay that remaps it GPIOs 12 & 13 or 18 & 19, i.e. `dtoverlay=audremap,pins_12_13`.

PR #5393, opened 2023-03-22
Add PDAF support for IMX519

ArduCAM:


6by9:
As this driver is supported by yourselves I've only given it a cursory review. I don't seen any glaring issues.

Changing the link frequency is a bit annoying as it means that other overlays using the driver will no longer work. In mainline that would not be acceptable, but I'm not expecting anyone to be using the driver elsewhere.

pelwell:
Is there not a standard DT binding for link speed?

pelwell:
Huh? The DT patch appears to override the link frequency, but then the driver patch changes the default. Are they both required?

6by9:
> Is there not a standard DT binding for link speed?

It's a standard property, but the driver only supports one value, and that value has been changed in this PR.
The binding defines what the hardware can support (which is a nearly unbounded range) and therefore doesn't specify anything. It's the driver implementation for this sensor that has the limitation.

An old overlay with the new driver (or vice versa) will fail to probe when the link frequency configuration is checked.

It does mean that the example in the dtbinding won't work with an actual system - https://github.com/raspberrypi/linux/blob/rpi-6.1.y/Documentation/devicetree/bindings/media/i2c/imx519.yaml#L107.
Then again the binding here is significantly different to the binding that has been reviewed in mainline - https://patchwork.linuxtv.org/project/linux-media/patch/20221209084831.000048ab@arducam.com/

6by9:
> Huh? The DT patch appears to override the link frequency, but then the driver patch changes the default. Are they both required?

As just posted, DT defines the desired value. This driver happens to be limited to one specific value, and fails if it doesn't match.

There are drivers that can configure their PLLs based on the DT requested link frequency (eg media/i2c/ccs), or can operate with one of a number of specific frequencies (eg media/i2c/imx290 which needs different values depending on whether using 2 or 4 lanes).
All drivers are therefore expected to read the DT value and validate it.

pelwell:
As it just alters a third-party driver and overlay that we merely host...

ArduCAM:
Hi @6by9 

> As this driver is supported by yourselves I've only given it a cursory review. I don't seen any glaring issues.
> 
> Changing the link frequency is a bit annoying as it means that other overlays using the driver will no longer work. In mainline that would not be acceptable, but I'm not expecting anyone to be using the driver elsewhere.

Thank you for your comment, I was thinking that the upstream branch has not yet merged the imx519 driver, I am going to submit a v2 version to the upstream and reduce the link frequency, is this also unacceptable? Maybe I should add support for two link frequencies?

6by9:
If your driver had been merged to mainline then dropping the existing link frequency value would have been considered a regression. Whilst it is still under review you can change anything you like.

Is there a specific reason for dropping the link frequency? I note it changes the max frame rate in each of the modes, and HTS is also changed, but none of that is critical to the sensor's operation.

ArduCAM:
Hi @6by9 ,

> If your driver had been merged to mainline then dropping the existing link frequency value would have been considered a regression. Whilst it is still under review you can change anything you like.
> 
> Is there a specific reason for dropping the link frequency? I note it changes the max frame rate in each of the modes, and HTS is also changed, but none of that is critical to the sensor's operation.

The main reason for modifying the link frequency is that some of our products do not support faster speeds, (such as the network cable extension version), and we choose to reduce the link frequency for maximum compatibility.
Due to the modification of the PLL, the related HTS has also been modified.
There is something special about 720 resolution, it needs bigger hts when PDAF is enabled (don't know the reason yet)

6by9:
> The main reason for modifying the link frequency is that some of our products do not support faster speeds, (such as the network cable extension version), and we choose to reduce the link frequency for maximum compatibility.

If it's for compatibility with other devices that may or may not be in use, then it would be nice to retain the "full" feature set and only degrade performance when necessary. Allowing one of a number of link-frequency values in DT would permit this.

However it's your product, so do as you wish.

IMX708 is in a similar position at present where the standard link frequency appears to cause issues with GPS receivers, so it is likely to gain DT control of chosen link freq fairly soon. There's nothing particularly magic in doing so.

PR #5394, opened 2023-03-22
drivers: media: i2c: imx708: Fix WIDE_DYNAMIC_RANGE control with long…

davidplowman:
… exposure

@naushir @njhollinghurst @6by9 

This is fixing a problem when a long frame length (requiring the special "long exposure" shift counter) is set before starting the sensor. Once the sensor is started, we will apply all the default control values, plus of course any updated frame length and exposure values. Note that we will also set the WIDE_DYNAMIC_RANGE control to its default value as well.

At this point we have no control over the order in which these controls happen - and the WIDE_DYNAMIC_RANGE control resets the "long exposure" counter. If this happens after the frame length had carefully set it... you'll end up with completely incorrect exposure times.

This doesn't seem like a nice fix to me. If anyone has a suggestion for something nicer, that would be great...!

### DIFF
--- a/drivers/media/i2c/imx708.c
+++ b/drivers/media/i2c/imx708.c
@@ -171,7 +171,7 @@ struct imx708_mode {
 	/* Not all modes have the same exposure lines step. */
 	u32 exposure_lines_step;
 
-	/* HDR flag, currently not used at runtime */
+	/* HDR flag, used for checking if the current mode is HDR */
 	bool hdr;
 };
 
@@ -1060,9 +1060,6 @@ static void imx708_set_framing_limits(struct imx708 *imx708)
 	unsigned int hblank;
 	const struct imx708_mode *mode = imx708->mode;
 
-	/* Default to no long exposure multiplier */
-	imx708->long_exp_shift = 0;
-
 	__v4l2_ctrl_modify_range(imx708->pixel_rate,
 				 mode->pixel_rate, mode->pixel_rate,
 				 1, mode->pixel_rate);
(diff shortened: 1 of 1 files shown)


6by9:
I thought it was just iterating through the list of controls which would be in the order that ```v4l2_ctrl_new_std``` added them to the driver. Is that not the case?

Actually the code in [s_ctrl](https://github.com/raspberrypi/linux/blob/rpi-6.1.y/drivers/media/i2c/imx708.c#L1160) for ```V4L2_CID_WIDE_DYNAMIC_RANGE``` is in the wrong place. The control is grabbed when streaming starts, therefore you'll never get passed the ```if (pm_runtime_get_if_in_use(&client->dev) == 0)``` at line 1105 when set.
It should be alongside ```if (ctrl->id == V4L2_CID_VBLANK)``` at line 1098 (change to a switch/case there). You can't change the mode at stream_on!

naushir:
>  It should be alongside `if (ctrl->id == V4L2_CID_VBLANK)` at line 1098 (change to a switch/case there). You can't change the mode at stream_on!

This was my initial suggestion as well, but it seems not to work for unknown reasons...

davidplowman:
Revised attempt!

drivers/media/i2c/imx708.c
```diff
@@ -1098,6 +1095,23 @@ static int imx708_set_ctrl(struct v4l2_ctrl *ctrl)
 	if (ctrl->id == V4L2_CID_VBLANK)
 		imx708_adjust_exposure_range(imx708, ctrl);
 
+	/*
```
6by9:
```
switch (ctrl->id) {
   case V4L2_CID_VBLANK:
...
   break;
   case V4L2_CID_WIDE_DYNAMIC_RANGE:
....
    break;
}

File: drivers/media/i2c/imx708.c [hunk 2]
```diff
@@ -1098,6 +1095,23 @@ static int imx708_set_ctrl(struct v4l2_ctrl *ctrl)
...
+	 * The WIDE_DYNAMIC_RANGE control can also be applied immediately as it
+	 * doesn't set any registers, though we shouldn't do anything if the
+	 * control value already matches the mode.
+	 */
+	if (ctrl->id == V4L2_CID_WIDE_DYNAMIC_RANGE &&
+	    imx708->mode && imx708->mode->hdr != ctrl->val) {
```
6by9: I thought we could get the old value from the control handler rather than having to cache it ourselves, but that may well be my memory. ctrl->val is certainly the new value, and I don't know where else it would be kept.

Isn't imx708->mode meant to be protected by the mutex? Or as we're using the same mutex for both driver and ctrl_handler, perhaps that is all safe anyway?

davidplowman: imx708->mode seemed to be fair game before I came along, so I'll defer to the camera gods on that one! I'll put in the switch statement though...

6by9:
Assuming it works, LGTM.

naushir:
LGTM

Issue #5390, opened 2023-03-21
On a CM3+, the regenerate_ssh_host_keys script in rasapberrypi-sys-mods takes a very long time to run on kernel v6.1


JinShil:
### Describe the bug

On a CM3+, the `regenerate_ssh_host_keys` script in `rasapberrypi-sys-mods` is taking a very long time to run after installing 6.1 via apt using our pi-gen-based OS creation script. The offending line appears to be `dd if=/dev/hwrng of=/dev/urandom count=1 bs=4096 status=none`

```
$ time sudo dd if=/dev/hwrng of=/dev/urandom count=1 bs=4096 status=none
real    3m28.951s
user    0m0.024s
sys     0m0.041s
```

The same command runs in under a second on a CM4 or CM4S.

### Steps to reproduce the behaviour

Using a CM3+, install a 64-bit Bullseye OS with kernel v6.1 and run the `regenerate_ssh_host_keys` script.

### Device (s)

Raspberry Pi CM3+, Raspberry Pi CM3+ Lite

### System

OS was generated using a script based on pi-gen.

Kernel version 6.1.19-v8+

pelwell:
I can confirm the regression - a 3B+ shows the same problem (I've not checked other models yet).

popcornmix:
Pi 4 seemed quick for me.

popcornmix:
On Pi4:
```
sudo cat /dev/hwrng
+K���(��u<��pP�3�<0�\R�⾬}���F����
```
(spurts lots of random stuff).

On Pi3+ nothing comes out.

pelwell:
It's interesting that there should be a difference (which I'll investigate), but on the surface the problem appears to come from this upstream commit: 96cb9d0554457086664d3bd10630b11193d863f1, the payload of which is:
```diff
diff --git a/drivers/char/hw_random/bcm2835-rng.c b/drivers/char/hw_random/bcm2835-rng.c
index e7dd457e9b22b..e98fcac578d66 100644
--- a/drivers/char/hw_random/bcm2835-rng.c
+++ b/drivers/char/hw_random/bcm2835-rng.c
@@ -71,7 +71,7 @@ static int bcm2835_rng_read(struct hwrng *rng, void *buf, size_t max,
        while ((rng_readl(priv, RNG_STATUS) >> 24) == 0) {
                if (!wait)
                        return 0;
-               cpu_relax();
+               hwrng_msleep(rng, 1000);
        }
 
[...]

PR #5396, opened 2023-03-22
hwrng: bcm2835 - sleep more intelligently

pelwell:
While waiting for random data, use sleeps that are proportional
to the amount of data expected. Prevent indefinite waits by
giving up if nothing is received for a second.

See: https://github.com/raspberrypi/linux/issues/5390

### DIFF
--- a/drivers/char/hw_random/bcm2835-rng.c
+++ b/drivers/char/hw_random/bcm2835-rng.c
@@ -71,7 +71,7 @@ static int bcm2835_rng_read(struct hwrng *rng, void *buf, size_t max,
 	while ((rng_readl(priv, RNG_STATUS) >> 24) == 0) {
 		if (!wait)
 			return 0;
-		cpu_relax();
+		hwrng_msleep(rng, 1000);
 	}
 
 	num_words = rng_readl(priv, RNG_STATUS) >> 24;
--- a/drivers/char/hw_random/bcm2835-rng.c
+++ b/drivers/char/hw_random/bcm2835-rng.c
@@ -14,6 +14,7 @@
 #include <linux/printk.h>
 #include <linux/clk.h>
 #include <linux/reset.h>
+#include <linux/delay.h>
 
 #define RNG_CTRL	0x0
 #define RNG_STATUS	0x4
@@ -27,6 +28,9 @@
 #define RNG_WARMUP_COUNT 0x40000
 
 #define RNG_INT_OFF	0x1
+
+#define RNG_FIFO_WORDS	4
+#define RNG_US_PER_WORD	34 /* Tuned for throughput */
 
 struct bcm2835_rng_priv {
 	struct hwrng rng;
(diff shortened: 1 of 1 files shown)


drivers/char/hw_random/bcm2835-rng.c
```diff
@@ -65,23 +69,33 @@ static int bcm2835_rng_read(struct hwrng *rng, void *buf, size_t max,
...
+		num_words = min(rng_readl(priv, RNG_STATUS) >> 24,
+				max_words - count);
 
-	for (count = 0; count < num_words; count++)
-		((u32 *)buf)[count] = rng_readl(priv, RNG_DATA);
+		for (; num_words; num_words--, count++)
+			((u32 *)buf)[count] = rng_readl(priv, RNG_DATA);
+
+		/* Reset the IDLE timeout */
+		idle_endtime = jiffies + HZ;
```
lategoodbye: Doesn't make this the timeout test in the outer loop useless?

drivers/char/hw_random/bcm2835-rng.c
```diff
@@ -65,23 +69,33 @@ static int bcm2835_rng_read(struct hwrng *rng, void *buf, size_t max,
...
 	u32 max_words = max / sizeof(u32);
-	u32 num_words, count;
+	u32 num_words, count = 0;
 
-	while ((rng_readl(priv, RNG_STATUS) >> 24) == 0) {
-		if (!wait)
-			return 0;
-		hwrng_msleep(rng, 1000);
-	}
+	while ((count < max_words) && time_before(jiffies, idle_endtime)) {
```
lategoodbye: It would be better to have the timeout checkout within the inner loop.

lategoodbye (commented):
As a generate note, this change is too complex to be applied as a fix in mainline.

drivers/char/hw_random/bcm2835-rng.c
pelwell: I think we can afford to miss a 1 second deadline by a few microseconds.

drivers/char/hw_random/bcm2835-rng.c
pelwell: As the commit message says, the timeout protects against no data arriving for a second, not against whether the whole read takes more than a second.

pelwell:
> As a generate note, this change is too complex to be applied as a fix in mainline.

I copied the timeout logic from the iproc-rng200 driver: https://github.com/raspberrypi/linux/blob/rpi-6.1.y/drivers/char/hw_random/iproc-rng200.c#L112

pelwell:
I'd be perfectly happy without the additional timeout mechanism.

lategoodbye:
> > As a generate note, this change is too complex to be applied as a fix in mainline.
> 
> I copied the timeout logic from the iproc-rng200 driver: https://github.com/raspberrypi/linux/blob/rpi-6.1.y/drivers/char/hw_random/iproc-rng200.c#L112

Yes, the iproc implementation is rather complex.

How about the omap implementation?
https://github.com/raspberrypi/linux/blob/rpi-6.1.y/drivers/char/hw_random/omap-rng.c#L186

Replacing the while loop with a for loop and move the register check within. The timeout is necessary to address the original RCU stall.

pelwell:
Here's a simpler version.

Issue #5398, opened 2023-03-24
Sense Hat matrix not working on 6.1 kernel.


steveiliop56:
### Describe the bug

After upgrading the raspberry pi kernel to 6.1 the sense hat matrix does no longer work. Thonny and python report that they can communicate with it and seem like they are trying to use the matrix but without dropping any error. I can confirm that the gyro, accelerometer, temperature, humidity and joystick work fine.

### Steps to reproduce the behaviour

You can check it by yourself by just upgrading the pi to the newest 6.1 kernel.

### Device (s)

Raspberry Pi 4 Mod. B, Sense Hat V1.0

### System

The `rpi-issue` file:

```
Raspberry Pi reference 2023-02-21
Generated using pi-gen, https://github.com/RPi-Distro/pi-gen, 25e2319effa91eb95edd9d9209eb9f8a584d67be, stage4
```

Firmware version:

```
Mar 17 2023 10:50:39
Copyright (c) 2012 Broadcom
version 82f3750a65fadae9a38077e3c2e217ad158c8d54 (clean) (release) (start)
```

Kernel version:

```
Linux raspberrypi4 6.1.19-v8+ #1637 SMP PREEMPT Tue Mar 14 11:11:47 GMT 2023 aarch64 GNU/Linux
```

### Logs

Here are the dmesg logs, if needed:

```
[    0.000000] Booting Linux on physical CPU 0x0000000000 [0x410fd083]
[    0.000000] Linux version 6.1.19-v8+ (dom@buildbot) (aarch64-linux-gnu-gcc-8 (Ubuntu/Linaro 8.4.0-3ubuntu1) 8.4.0, GNU ld (GNU Binutils for Ubuntu) 2.34) #1637 SMP PREEMPT Tue Mar 14 11:11:47 GMT 2023
[    0.000000] random: crng init done
[    0.000000] Machine model: Raspberry Pi 4 Model B Rev 1.2
[    0.000000] efi: UEFI not found.
[    0.000000] Reserved memory: created CMA memory pool at 0x000000000ec00000, size 512 MiB
[    0.000000] OF: reserved mem: initialized node linux,cma, compatible id shared-dma-pool
[    0.000000] Zone ranges:
[    0.000000]   DMA      [mem 0x0000000000000000-0x000000003fffffff]
[    0.000000]   DMA32    [mem 0x0000000040000000-0x00000000fbffffff]
[    0.000000]   Normal   empty
[    0.000000] Movable zone start for each node
[...]

PR #5399, opened 2023-03-24
rpisense-fb: Flush any deferred updates on release

pelwell:
Since [1], writing to the Sense HAT framebuffer and immediately closing the file descriptor is likely to have no effect. This is because the framebuffer deferred IO framework is now careful to cancel pending updates.

To retain the old behaviour, add an fb_release method that flushes the updates instead.

See: https://github.com/raspberrypi/linux/issues/5398

[1] 3efc61d95259 ("fbdev: Fix invalid page access after closing deferred I/O devices")

### DIFF
--- a/drivers/video/fbdev/rpisense-fb.c
+++ b/drivers/video/fbdev/rpisense-fb.c
@@ -183,8 +183,18 @@ static int rpisense_fb_ioctl(struct fb_info *info, unsigned int cmd,
 	return 0;
 }
 
+static int rpisense_fb_release(struct fb_info *info, int user)
+{
+	/* Flush any pending updates */
+	cancel_delayed_work(&info->deferred_work);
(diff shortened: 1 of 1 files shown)


bsimmo:
This has made the LEDs light up. But there are still something majorly wrong.
https://forums.raspberrypi.com/viewtopic.php?p=2094406#p2094218
Test before mine and my test. I can give video if really needed.
Redraw is very slow.

pelwell:
Without trying this and/or reading the Python library (both of which I will), my guess would be that library is opening and closing the frame buffer for each access, each of which will incur at least a 10ms penalty. strace would confirm either way.

pelwell:
Yup: https://github.com/astro-pi/python-sense-hat/blob/master/sense_hat/sense_hat.py#L377

PR #5404, opened 2023-03-27
rpisense-fb: Add explicit fb_deferred_io_mmap hook

pelwell:
As of commit [1], introduced in 5.18, fbdev drivers that use deferred IO and need mmap support must include an explicit fb_mmap pointer to the fb_deferred_io_mmap.

[1] 590558510327 ("fbdev: Put mmap for deferred I/O into drivers")


platu:
Memory mapping of the led matrix works fine again with this fix 👍🏻

PR #5407, opened 2023-03-28
dtoverlays: Add inverted override property to ssd1306-spi

martinezjavier:
The new ssd130x DRM driver supports both SSD1306 I2C and SPI panels and is compatible with the ssd1307fb driver bindings. So the "solomon,com-invdir" DT property (to invert the COM pin scan dir) can also be used for SPI now.

This allows to configure panels whose scan direction needs to be inverted.

### DIFF
--- a/arch/arm/boot/dts/overlays/README
+++ b/arch/arm/boot/dts/overlays/README
@@ -4074,6 +4074,8 @@ Params: speed                   SPI bus speed (default 10000000)
         dc_pin                  GPIO pin for D/C (default 24)
         reset_pin               GPIO pin for RESET (default 25)
         height                  Display height (32 or 64; default 64)
+        inverted                Set this if display is inverted and mirrored.
+                                (default=not set)
 
 
 Name:   ssd1331-spi
--- a/arch/arm/boot/dts/overlays/ssd1306-spi-overlay.dts
+++ b/arch/arm/boot/dts/overlays/ssd1306-spi-overlay.dts
@@ -80,5 +80,6 @@
 		reset_pin = <&ssd1306>,"reset-gpios:4",
 		            <&ssd1306_pins>,"brcm,pins:0";
 		height    = <&ssd1306>,"solomon,height:0";
+		inverted = <&ssd1306>,"solomon,com-invdir?";
 	};
 };


pelwell:
As the automated checks will have told you, a README entry is required for the added parameter:
```
...
ssd1306-spi undocumented parameters:
  inverted
Failed
```
Other than that I'm happy with this PR.

6by9:
It looks like this property has been present since the driver was added in [5.18](https://github.com/torvalds/linux/commit/a61732e808672cfa8c8c6028bcf9feacb953ef40). Any reason why you're specifically targeting 6.3?
6.1 is the LTS that we're using for our vendor kernel releases.

Otherwise it LGTM too.

martinezjavier:
> As the automated checks will have told you, a README entry is required for the added parameter:

It should be fixed now.

martinezjavier:
> It looks like this property has been present since the driver was added in [5.18](https://github.com/torvalds/linux/commit/a61732e808672cfa8c8c6028bcf9feacb953ef40). Any reason why you're specifically targeting 6.3?

No, other that I'm not familiar with how this downstream tree is managed.

> 6.1 is the LTS that we're using for our vendor kernel releases.

Ok, should I change the target branch to 6.1 and will get merged back to the newer branches or?

pelwell:
6.1 is the current default branch, and PRs should target it unless there is some specific reason not to.

martinezjavier:
> 6.1 is the current default branch, and PRs should target it unless there is some specific reason not to.

Ok. I changed the target branch now.

arch/arm/boot/dts/overlays/README
```diff
@@ -4074,7 +4074,8 @@ Params: speed                   SPI bus speed (default 10000000)
         dc_pin                  GPIO pin for D/C (default 24)
         reset_pin               GPIO pin for RESET (default 25)
         height                  Display height (32 or 64; default 64)
-
+        inverted                Set this if display is inverted and mirrored.
+                                (default=not set)
```
pelwell, line 4079: Unfortunately you've deleted one of the two required blank lines between entries, as the error log says:
```
* Missing blank lines after overlay? (4080)
```

martinezjavier: Ups, sorry about that. It should be fixed now.

pelwell:
Yes, that's better.

PR #5409, opened 2023-03-28
media: bcm2835-unicam: Start and stop media_pipeline with same node

6by9:
media_pipeline_start and media_pipeline_stop now validate that the pipeline is being started and stopped with the same pipe and pad handles.
When running with embedded metadata (eg imx477 and imx708), the start typically happens from the metadata pad, whilst stop is always from the image pad.

Always pass the image pad to media_pipeline_start to ensure that the calls are balanced.

### DIFF
--- a/drivers/media/platform/bcm2835/bcm2835-unicam.c
+++ b/drivers/media/platform/bcm2835/bcm2835-unicam.c
@@ -2524,7 +2524,8 @@ static int unicam_start_streaming(struct vb2_queue *vq, unsigned int count)
 		goto err_streaming;
 	}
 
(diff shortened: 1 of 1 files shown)


6by9:
@naushir This fixes the issue for imx477.
Not observed for imx219 as it appears libcamera has disabled embedded metadata.
I'd not tested imx708 prior to this change, but don't see a splat with it.

naushir:
Looks good!

6by9:
Any other comments, or are we good to merge?
Without this we're getting a kernel WARN every time we stop IMX708 or IMX477 streaming (camera modules that use embedded data).

PR #5413, opened 2023-03-29
Alternative SenseHAT fix for 6.1

pelwell:
Since [1], the fbdev deferred IO framework is careful to cancel
pending updates on close to prevent dirty pages being accessed after
they may have been reused. However, this is not necessary in the case
that the pagelist is empty, and drivers that don't make use of the
pagelist may have wanted updates cancelled for no good reason.

Avoid penalising fbdev drivers that don't make use of the pagelist by
making the cancelling of deferred IO on close conditional on there
being a non-empty pagelist.

See: https://github.com/raspberrypi/linux/issues/5398

[1] https://github.com/raspberrypi/linux/commit/3efc61d95259956db25347e2a9562c3e54546e20 ("fbdev: Fix invalid page access after closing deferred I/O devices")

### DIFF
--- a/drivers/video/fbdev/core/fb_defio.c
+++ b/drivers/video/fbdev/core/fb_defio.c
@@ -321,7 +321,8 @@ static void fb_deferred_io_lastclose(struct fb_info *info)
 	struct page *page;
 	int i;
 
-	cancel_delayed_work_sync(&info->deferred_work);
+	if (!list_empty(&info->fbdefio->pagereflist))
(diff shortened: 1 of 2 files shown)


pelwell:
This alternative patch avoids the horrendous slowdown seen with the sense_hat.py library, which opens and closes the framebuffer, incurring an 18ms delay while the data is pushed over I2C, for every `get_pixel`  and `set_pixel` call.

PR #5414, opened 2023-03-30
configs: Disable LRU_GEN_ENABLED on 32-bit kernels

popcornmix:
We seem to be getting increased OOM failures with MGLRU enabled on 32-bit kernels. 
Lets disable it by default (it can still be manually enabled at runtime).

See: https://github.com/raspberrypi/linux/issues/5395


pelwell:
Checkpatch doesn't like the long commit message, but otherwise it LGTM.

pelwell:
Apart from the missing apostrophe.

PR #5420, opened 2023-04-08
configs: Add CONFIG_CHARGER_GPIO=m

xrh0905:
Related issue: https://github.com/raspberrypi/firmware/issues/1801


arch/arm/configs/bcm2709_defconfig
```diff
@@ -725,6 +725,7 @@ CONFIG_RPI_POE_POWER=m
 CONFIG_BATTERY_DS2760=m
 CONFIG_BATTERY_MAX17040=m
 CONFIG_BATTERY_GAUGE_LTC2941=m
+CONFIG_CHARGER_GPIO=m
```
pelwell, line 727: All of these additions need to go before CONFIG_BATTERY_GAUGE_LTC2941. When creating patches like this, use `make ... xxx_defconfig`, `make ... savedefconfig` and compare the output file (`defconfig`) with the `xxx_defconfig` file in questions. They should be identical.

pelwell (commented):
When config settings are added like this they should be added for all build variants unless they are incompatlble (e.g. PCIE on a Pi 3). You've omitted `arm/bcm2711_defconfig` and `arm64/bcm2711_defconfig`.

PR #5421, opened 2023-04-08
overlays: Add gpio-charger support

xrh0905:
Related issue: https://github.com/raspberrypi/firmware/issues/1801
Related PR: #5420

### DIFF
--- a/arch/arm/boot/dts/overlays/gpio-charger-overlay.dts
+++ b/arch/arm/boot/dts/overlays/gpio-charger-overlay.dts
@@ -1,0 +1,42 @@
+// Definitions for gpio-charger module
+/dts-v1/;
+/plugin/;
+
+/ {
+	compatible = "brcm,bcm2835";
+
+	fragment@0 {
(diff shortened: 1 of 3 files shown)


arch/arm/boot/dts/overlays/gpio-charger-overlay.dts
```diff
@@ -0,0 +1,41 @@
...
+				gpios = <&gpio 4 0>;
+				charger-type = "mains";
+			};
+		};
+	};
+
+	__overrides__ {
+		gpio =       <&charger>,"reg:0",
+			     <&pin_state>,"brcm,pins:0",
+			     <&pin_state>,"reg:0";
```
pelwell: You've missed out the most important one - `<&charger>,"gpios:4"`.

pelwell:
And as with #5422, please leave a blank line between the first line of the commit message (the Subject) and the body of the message, i.e. leave the second line blank, otherwise the output of `git log --oneline` includes the second line.

PR #5427, opened 2023-04-12
configs: CONFIG_USB_F_*=m CONFIG_USB_CONFIGFS_*=m (Complete USB Gadget Support for bcmrpi3_defconfig)

ilovegabagoolsomuch:
The changes I've added are cross-referenced with bcm2711_defconfig to include the same USB gadgets.

It adds the actual USB Gadget modules (which were not added by default, despite being in the bcm2711_defconfig) for the Pi 3 and deriviative devices (BCM2837) such as the Pi Zero 2 W.


pelwell:
Your patch includes some Kconfig settings that are implied by others - they should not appear in a _defconfig file, which is meant to only include variations from the defaults. I've added a patch to remove the unnecessary settings, and it will be squashed at merge time.

I'm happy to accept this patch with the above changes, but you need to be aware that the 64-bit kernels that we release are all built from arm64/bcm2711_defconfig - bcmrpi3_defconfig is a user-contributed file that may be useful if you are trying to slim down your kernel for a BCM2710-based Pi.

ilovegabagoolsomuch:
Thanks for that tip! It seems like some distributions (such as Alpine Linux) use the `bcmrpi3_defconfig`.

```
case $flavor in
		rpi) defconfig=bcmrpi_defconfig
		[ "$CARCH" = "aarch64" ] && defconfig=bcmrpi3_defconfig ;;
		rpi2) defconfig=bcm2709_defconfig ;;
		rpi4) defconfig=bcm2711_defconfig ;;
		*) die "Unknown flavor: $flavor" ;;
	esac
```

I see how I could use Alpine's `linux-rpi4` kernel, but I think USB Gadget support on even a minimal config is necessary, especially given that this group of devices contains the Pi Zero. Perhaps I should file an issue on their end. Thanks for the help!

PR #5428, opened 2023-04-13
ARM: dts: bcm27xx: Add i2c_arm/vc and friends

pelwell:
Since there is now a dedicated dts file for the rev1 Model B (the only Pi to drive the primary camera with i2c1), move the creation of the i2c_arm, i2c_vc and i2c labels, aliases and overrides into the base dts files.

### DIFF
--- a/arch/arm/boot/dts/bcm2708-rpi-b-plus.dts
+++ b/arch/arm/boot/dts/bcm2708-rpi-b-plus.dts
@@ -186,6 +186,12 @@
 cam0_reg: &cam_dummy_reg {
 };
 
+i2c_arm: &i2c1 {
+};
+
+i2c_vc: &i2c0 {
+};
+
 / {
 	__overrides__ {
 		audio = <&chosen>,"bootargs{on='snd_bcm2835.enable_headphones=1 snd_bcm2835.enable_hdmi=1',off='snd_bcm2835.enable_headphones=0 snd_bcm2835.enable_hdmi=0'}";
--- a/arch/arm/boot/dts/bcm2708-rpi-b-rev1.dts
+++ b/arch/arm/boot/dts/bcm2708-rpi-b-rev1.dts
@@ -197,6 +197,12 @@ i2c_csi_dsi: &i2c1 {
 cam0_reg: &cam_dummy_reg {
 };
 
+i2c_arm: &i2c0 {
+};
+
+i2c_vc: &i2c1 {
+};
+
 / {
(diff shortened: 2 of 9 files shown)


timg236:
Looks like a good idea

vfazio:
@pelwell I noticed that the commit message talks about making the `i2c` label but i do not see `i2c` in the `__symbols__` list on newly generated device trees, so any overlays binding to `<&i2c>` will not load.

The fix seems pretty straightforward `i2c: &i2c1 {}` (i2c0 for rev 1 model B) however I wanted to confirm the intent about exposing `i2c` in `__symbols__` before writing an issue or submitting an MR... or if there's some expectation (or documentation) that overlays should _not_ bind to &i2c and instead should bind to &i2c_arm.

The `i2c` symbol is available on 5.15 kernels from what I can tell

pelwell:
You're correct that, according to the commit message, there should be an `i2c` label, but perhaps that's not the best outcome. There appears to be only one user of it - the `adau1977-adc` overlay - so perhaps that should be migrated to use `i2c_arm`. How have you discovered its absence - are you using `adau1977-adc`?

vfazio:
@pelwell we manufacture custom carrier boards for the RPi CM 3 and have our own custom dts overlays for those carriers. We are in the process of migrating to the rpi-6.1.y branch and noticed this

pelwell:
Fair enough. I've gone ahead and replaced `&i2c` with `&i2c_arm` in `adau1977-adc`. Are you happy to do the same with your custom overlays, or is that going to be a problem?

vfazio:
@pelwell I think we can make the change in our overlays without issue, especially since it should be backwards compatible with 5.15.

Should there some documentation update about `i2c` no longer being available? I'm not sure where that'd be other than maybe `arch/arm/boot/dts/overlays/README`.

Just thinking that other vendors may be impacted by the change.

pelwell:
I don't think there is a good place to document that change except here - the README covers the user-facing aspects of the overlays, and the overlay documentation doesn't mention `&i2c` - in https://www.raspberrypi.com/documentation/computers/configuration.html#part3.3 it says:

> For people writing overlays, the same aliasing has been applied to the labels on the I2C DT nodes. Thus, you should write:
```
    fragment@0 {
        target = <&i2c_arm>;
        __overlay__ {
            status = "okay";
        };
    };
````

vfazio:
OK, thanks for the quick responses @pelwell !

Issue #5429, opened 2023-04-13
I2C NACK being ignored by ioctl I2C_RDWR


jdf-advanced:
### Describe the bug

I'm using C to interface with some I2C devices, but I've come across a corner case that I think could cause really infrequent issues. I want to see if a community has spotted a way round it, or if I'm doing something wrong. 

I'm using ioctl with I2C_RDWR to do a read of a register using a repeated start. The ideal procedure is this:
```
[START CONDITION]
[SEND DEVICE ADDR + W]
[RECEIVE ACK]
[SEND REGISTER ADDR]
[RECEIVE ACK]
[START CONDITION]
[SEND DEVICE ADDR + R]
[RECEIVE ACK]
loop {
  [RECEIVE DATA]
  [SEND ACK]
}
[STOP CONDITION]
```

If the first ACK is instead a NACK (i.e. the slave doesn't respond to it's address) I'd expect the master (the pi) to send a STOP condition and cease the transmission. Instead it proceeds to do a repeated start and try the read anyway. 

If the initial NACK is caused by the slave device being busy, but it's not busy by the time the read is attempted, this could result in invalid data being received, and possibly unintended behaviour from the slave. 

It could cause issues if a slave device performs some sort of state change (clearing interrupts, kicking off the next ADC sample conversion etc.) when a read occurs. As far as the slave is concerned, it may have just done a perfectly valid read, but the master has ignored the data. 

### Steps to reproduce the behaviour

To reliably simulate this happening, I've been doing the write to a device that doesn't exist, and then a read from a device that does. Here's the code:

```C
file_i2c = open("/dev/i2c-1", O_RDWR); 

uint8_t inbuf[20];
uint8_t outbuf[20];
struct i2c_rdwr_ioctl_data packets;
struct i2c_msg messages[2];

outbuf[0] = 0x40;
messages[0].addr = 0x36;
messages[0].flags = 0;
messages[0].len = 1;
messages[0].buf = outbuf;

messages[1].addr = 0x49;
messages[1].flags = I2C_M_RD;
messages[1].len = 4;
messages[1].buf = inbuf;

packets.msgs = messages;
packets.nmsgs = 2;

int result = ioctl(file_i2c, I2C_RDWR, &packets);
```

This is the logic analyser's output of what's happening:

![image](https://user-images.githubusercontent.com/98110076/231705770-5ca96219-31fc-40ed-8f1f-cb9b26e1bffc.png)

[...]

PR #5434, opened 2023-04-15
i2c: bcm2835: Add NAK delay

pelwell:
The BCM2835 I2C controllers do not have an interrupt-on-error capability, and this makes it easy to proceed past a write NAK condition to the remainder of a multi-part transfer. In the event that a write is followed by other transfers, introduce a short delay to allow the NAK to be detected.

Dallying in interrupt context is not ideal, even for 100us (with a 100kHz bus) so better implementations are welcome.

See: https://github.com/raspberrypi/linux/issues/5429

### DIFF
--- a/drivers/i2c/busses/i2c-bcm2835.c
+++ b/drivers/i2c/busses/i2c-bcm2835.c
@@ -15,6 +15,7 @@
 #include <linux/of_device.h>
 #include <linux/platform_device.h>
 #include <linux/slab.h>
+#include <linux/delay.h>
 
 #define BCM2835_I2C_C		0x0
 #define BCM2835_I2C_S		0x4
@@ -64,6 +65,10 @@ static unsigned int clk_tout_ms = 35; /* SMBUs-recommended 35ms */
 module_param(clk_tout_ms, uint, 0644);
 MODULE_PARM_DESC(clk_tout_ms, "clock-stretch timeout (mS)");
 
+static unsigned int nak_delay_cycles;
+module_param(nak_delay_cycles, uint, 0644);
+MODULE_PARM_DESC(nak_delay_cycles, "NAK delay (cycles/bits)");
+
 #define BCM2835_DEBUG_MAX	512
 struct bcm2835_debug {
 	struct i2c_msg *msg;
@@ -80,6 +85,7 @@ struct bcm2835_i2c_dev {
 	struct completion completion;
 	struct i2c_msg *curr_msg;
 	struct clk *bus_clk;
+	unsigned long nak_delay_us;
 	int num_msgs;
 	u32 msg_err;
 	u8 *msg_buf;
@@ -232,6 +238,12 @@ static int clk_bcm2835_i2c_set_rate(struct clk_hw *hw, unsigned long rate,
 
 	bcm2835_i2c_writel(div->i2c_dev, BCM2835_I2C_CLKT, clk_tout);
 
+	/* Set the NAK wait time to be 10 bits in microseconds */
+	div->i2c_dev->nak_delay_us = (nak_delay_cycles * 1000000ul) / rate;
+	if (nak_delay_cycles)
+		dev_warn(div->i2c_dev->dev, "NAK delay: %d cycles -> %ld us\n",
+			 nak_delay_cycles, div->i2c_dev->nak_delay_us);
+
 	return 0;
 }
 
@@ -380,6 +392,13 @@ static irqreturn_t bcm2835_i2c_isr(int this_irq, void *data)
 {
 	struct bcm2835_i2c_dev *i2c_dev = data;
 	u32 val, err;
+
+	/* Allow time for any NAK to be detected */
+	if (i2c_dev->num_msgs &&
+	    !(i2c_dev->curr_msg->flags & I2C_M_RD) &&
+	    (i2c_dev->msg_buf_remaining == i2c_dev->curr_msg->len) &&
+	    i2c_dev->nak_delay_us)
+		udelay(i2c_dev->nak_delay_us);
 
 	val = bcm2835_i2c_readl(i2c_dev, BCM2835_I2C_S);
 	bcm2835_debug_add(i2c_dev, val);


jdf-advanced:
According to my testing, this doesn't appear to solve the issue. I can't see a difference in behaviour after doing the update. 
```
$ uname -a
Linux raspberrypi 6.1.23-v7l+ #1 SMP Sat Apr 15 10:53:47 UTC 2023 armv7l GNU/Linux
```
I'm pretty sure I have updated it correctly, but running the same C test shown in #5429 results in the same activity on the I2C bus

pelwell:
Strange - it works for me. Let me make the delay larger.

pelwell:
I've gone one better, and made the NAK delay a module parameter - `nak_delay_cycles`. It defaults to 0, which effectively disables the mechanism and makes me happier to merge this. A value of 12 ought to be sufficient (10 was working for me). You can make it larger, but note that this will increase CPU load and decrease I2C throughput.

To test (after the autobuilds have completed):
1. `sudo rpi-update pulls/5434`
2. Edit cmdline.txt and add (say) `i2c_bcm2835.nak_delay_cycles=12`.
3. After rebooting, confirm that the setting has taken effect:
    ```
    $ dmesg | grep NAK
    [    7.937021] i2c-bcm2835 fe804000.i2c: NAK delay: 12 cycles -> 120 us
    ```

pelwell:
This is with `nak_delay_cycles=12` - notice the small gap between the address 0x68 and the value 0x00:
i2cnak

jdf-advanced:
I'm afraid I'm still not seeing a difference. I wonder if there's something masking it.

```
master@raspberrypi:~ $ cat /boot/cmdline.txt
console=serial0,115200 console=tty1 root=PARTUUID=3ccf3caf-02 rootfstype=ext4 fsck.repair=yes rootwait i2c_bcm2835.nak_delay_cycles=12
master@raspberrypi:~ $ uname -a
Linux raspberrypi 6.1.24-v7l+ #1 SMP Thu Apr 20 09:31:30 UTC 2023 armv7l GNU/Linux
master@raspberrypi:~ $ dmesg | grep NAK
[    2.222006] dwc_otg: NAK holdoff enabled
[    6.375446] i2c-bcm2835 fe804000.i2c: NAK delay: 12 cycles -> 120 us
```

image

pelwell:
You'll only see it if it's not the last part of a multi-part message. As I understand it, even the old driver would correctly handle a NAK in the scenario above.

jdf-advanced:
Outcome of a more thorough test:

image

Really not sure why I'm getting different results than you at this stage

pelwell:
Have you tried making the delay larger, to see if it is having any effect?

jdf-advanced:
At `nak_delay_cycles=23` it starts to have an affect. However, at that value, I start to see bus errors, presumably caused by some sort of timing conflict. Approx 25% of the transfers fail in an interesting manner:

image
It stays in that state, with the SDA line low, until the next time it tries to do a transaction, at which point there is still no stop bit. 
image 

At `nak_delay_cycles=25` it was consistently not continuing with either transaction, as shown in the first image. 

At `nak_delay_cycles=30` it appears to be exhibiting the correct/expected behaviour, adding about 85us to each transaction.

pelwell:
Give me your self-contained test code, a single .c file I can just run gcc on, to eliminate any potential differences there,

jdf-advanced:
```C
#include "pigpio.h"
#include "stdio.h"
#include "stdlib.h"
#include "string.h"
#include <unistd.h>
#include <fcntl.h>
#include <sys/ioctl.h>
#include <linux/i2c.h>
#include <linux/i2c-dev.h>

int file_i2c;

int main(int c, char *v[])
{
        file_i2c = open("/dev/i2c-1", O_RDWR);

        uint8_t inbuf[20];
        uint8_t outbuf[20];
        struct i2c_rdwr_ioctl_data packets;
        struct i2c_msg messages[2];

        outbuf[0] = 0x40;
        messages[0].addr = 0x36;
        messages[0].flags = 0;
        messages[0].len = 1;
        messages[0].buf = outbuf;

        messages[1].addr = 0x49;
        messages[1].flags = I2C_M_RD;
        messages[1].len = 4;
        messages[1].buf = inbuf;

        packets.msgs = messages;
        packets.nmsgs = 2;

        int result = ioctl(file_i2c, I2C_RDWR, &packets);

        printf("message 0 len: %d; message 1 len: %d; result: %d\n", messages[0].len, messages[1].len, result);

        messages[0].addr = 0x54;
        messages[0].flags = 0;
        messages[0].len = 1;
        messages[0].buf = outbuf;
        outbuf[0] = 0x10;

        messages[1].addr = 0x54;
        messages[1].flags = I2C_M_RD;
        messages[1].len = 4;
        messages[1].buf = inbuf;

        ioctl(file_i2c, I2C_RDWR, &packets);

        return 0;
}
```

pelwell:
I see what the problem is now. Are you running with "dtoverlay=disable-bt", or in some other way disabling uart ttyS0? If so, try running with this while loop in the background to keep Linux active:
```
$ while true; do true; done &
```
This will have the effect of keeping the core at its higher setting of 500MHz, which in turn will give a stable 100kHz I2C clock. From the logic analyser captures I can see that your I2C clock is dropping down to 40kHz, which throws out the delays.

For your system performance you probably want to keep the core clock (and therefore I2C clock) stable. If you must disable the other UART then you can keep the core clock high with (on a Pi 4):
```
core_freq=500
core_freq_min=500
```

jdf-advanced:
```
core_freq=500
core_freq_min=500
```
With those two lines in config.txt the problem appears to be fixed. 

I don't think I'm willing to directly approve the PR for a few reasons:

- It seems really fragile to me, if the core frequency has to be set just to get it to work properly. The I2C stuff is generally independent of clock speed. Adding this one tiny aspect that's sort of fixed feels like a fragile thing to do.
- I don't know how this would interplay with clock stretching. On the ACK or NACK is definitely a point where a clock stretch is likely to happen.
- Something that fixes bus integrity feels to me like it should be part of the driver permanently, rather than an optional added feature. I appreciate that the silicon may make it impossible to add without a significant recurring penalty for a tiny edge case. But when people do come across that edge case, knowing that this argument fixes it would be a huge challenge. This is a _really_ hard problem to Google!
- I always understood that ISRs should be exited as soon as reasonably possible, and certainly shouldn't have blocking delays in them. I don't know if this is so much of an issue with a Linux ISR vs the embedded ones I'm used to dealing with, but it's another reason why I'll let someone with more experience of this codebase approve or disapprove the PR.

So, I can tell you that the code does fix the issue I've been seeing. I'll let someone who knows more than I do approve the PR!

PR #5435, opened 2023-04-18
configs: Enable LRU_GEN_ENABLED on non-pi4 32-bit kernels

popcornmix:
We believe the OOM issue with MGLRU on 32-bit kernel only affects Pi4 which uses CONFIG_VMSPLIT_3G.

Pi0-3 which use CONFIG_VMSPLIT_2G is expected not to suffer from this OOM issue.

So lets enable MGLRU by default for these users,
who many benefit from the advantages it has.

See: https://github.com/raspberrypi/linux/issues/5395


popcornmix:
FYI: @yuzhaogoogle

PR #5437, opened 2023-04-19
Let the kernel initialise the onboard Bluetooth modem by default

pelwell:
Allowing the kernel to initialise the Bluetooth modem means it can be reset and reloaded if necessary, and eventually we'll be able to get rid of the commands `hciattach` and `btuart`, and the `hciuart` service that calls them.

### DIFF
--- a/arch/arm/boot/dts/bcm2708-rpi-bt.dtsi
+++ b/arch/arm/boot/dts/bcm2708-rpi-bt.dtsi
@@ -5,22 +5,34 @@
 		compatible = "brcm,bcm43438-bt";
 		max-speed = <3000000>;
 		shutdown-gpios = <&gpio 45 GPIO_ACTIVE_HIGH>;
-		status = "disabled";
+		local-bd-address = [ 00 00 00 00 00 00 ];
+		fallback-bd-address; // Don't override a valid address
+		status = "okay";
 	};
 };
 
 &uart1 {
 	minibt: bluetooth {
 		compatible = "brcm,bcm43438-bt";
-		max-speed = <460800>;
+		max-speed = <230400>;
 		shutdown-gpios = <&gpio 45 GPIO_ACTIVE_HIGH>;
+		local-bd-address = [ 00 00 00 00 00 00 ];
+		fallback-bd-address; // Don't override a valid address
 		status = "disabled";
 	};
 };
 
 / {
+	aliases {
+		bluetooth = &bt;
+	};
+
 	__overrides__ {
+		bdaddr = <&bt>,"local-bd-address[",
+		       <&bt>,"fallback-bd-address?=0",
+		       <&minibt>,"local-bd-address[",
+		       <&minibt>,"fallback-bd-address?=0";
 		krnbt = <&bt>,"status";
-		krnbt_baudrate = <&bt>,"max-speed:0";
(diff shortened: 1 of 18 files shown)


hailfinger:
Does dtoverlay=disable-bt still disable all Bluetooth functionality even with this patch applied or are other config.txt entries necessary as well?

pelwell:
Good question. It will be made to do so.

I spent much of last week trying to get the miniuart case more reliable, so treat the current state of this PR as more of a statement of intent than an indication of how it will look finally. I hope to update it in the next day or so.

pelwell:
This is ready for review/testing now, but first a warning.

Some early Bluetooth-equipped Pis do not have the BDADDR programmed into OTP and rely on a variety of userspace applications and scripts to set them. Updating the kernel and dtbs makes some of those applications irrelevant (good) and breaks the corrective scripts (not so good).There will be an accompanying firmware update that supplies the same fallback BDADDR by patching the dtb, but as of writing that version of the firmware hasn't been released.

There is a workaround for anyone who finds that hci0 disappears - add the following to config.txt:
```
dtparam=bdaddr=aa:aa:aa:aa:aa:aa
```
This will evade the kernel's invalid address checks, but be picked up and corrected by the bthelper script.

pelwell:
The latest rpi-update release includes the BDADDR patching described above.

macmpi:
Notifying @gentoo-root  @lategoodbye who contributed to initial discussion on that topic https://github.com/raspberrypi/linux/issues/3675
Would make sense some of these changes also land into mainline device tree files then?

lategoodbye:
I cannot speak for the Bluetooth/devicetree maintainer, but at least:
Bluetooth: hci_bcm: Add more invalid BDADDRs
serial: 8250: Add NOMSI bug for bcm2835aux
looks very good. Since the fallback bd address doesn't look really device specific to me, a module parameter would be a alternative. But no strong opinion about this.

mykhani:
> Allowing the kernel to initialise the Bluetooth modem means it can be reset and reloaded if necessary, and eventually we'll be able to get rid of the commands `hciattach` and `btuart`, and the `hciuart` service that calls them.

Was there any specific problem that this kernel based initialisation of the Bluetooth modem addressed? More specifically, allowing resetting and reloading the Bluetooth modem, was that the main reason behind this change?

We are facing some problems with the Bluetooth HCI interface for the internal Bluetooth (`BCM34455`) getting timed out and stuck on a custom CM4 based device, running Yocto Linux with kernel version `5.15.34` from `meta-raspberrypi`.  I can see online similar issues happening on different RaspberryPi devices with different kernel versions.

Right now, we can recover by restarting the device, which I guess, resets and power cycles the Bluetooth modem. So apparently, this feature will allow us to recover just by reloading the Bluetooth driver instead of restarting the whole device?

And If yes, I am curious to understand how this might be a valid workaround? I mean, it could be a problem/bug in the Bluetooth firmware (`BCM4345C0.hcd`)? Or some internal handling of Bluetooth by the Linux kernel? Won't this hide solving the root cause of the problem?

pelwell:
> Was there any specific problem that this kernel based initialisation of the Bluetooth modem addressed? More specifically, allowing resetting and reloading the Bluetooth modem, was that the main reason behind this change?

No - the motivation was removing the dependence on hciattach.

Issue #5438, opened 2023-04-20
imx219 sensor 200fps not possible anymore (max. 103fps currently)


Hermann-SW:
### Describe the bug

In January "--list-cameras" reports >200fps:
```
pi@pi:~ $ libcamera-hello --list-cameras
Available cameras
-----------------
0 : imx219 [3280x2464] (/base/soc/i2c0mux/i2c@1/imx219@10)
    Modes: 'SRGGB10_CSI2P' : 640x480 [206.65 fps - (1000, 752)/1280x960 crop]
                             1640x1232 [41.85 fps - (0, 0)/3280x2464 crop]
                             1920x1080 [47.57 fps - (680, 692)/1920x1080 crop]
                             3280x2464 [21.19 fps - (0, 0)/3280x2464 crop]
           'SRGGB8' : 640x480 [206.65 fps - (1000, 752)/1280x960 crop]
                      1640x1232 [41.85 fps - (0, 0)/3280x2464 crop]
                      1920x1080 [47.57 fps - (680, 692)/1920x1080 crop]
                      3280x2464 [21.19 fps - (0, 0)/3280x2464 crop]
```

After "sudo apt update" and "sudo apt upgrade" only 103fps.
There are only 6 commits to imx219.c since January:
https://github.com/raspberrypi/linux/commits/rpi-6.1.y/drivers/media/i2c/imx219.c
```
pi@raspberrypi4B2:~ $ libcamera-hello --list-cameras
Available cameras
-----------------
0 : imx219 [3280x2464] (/base/soc/i2c0mux/i2c@1/imx219@10)
    Modes: 'SRGGB10_CSI2P' : 640x480 [103.33 fps - (1000, 752)/1280x960 crop]
                             1640x1232 [41.85 fps - (0, 0)/3280x2464 crop]
                             1920x1080 [47.57 fps - (680, 692)/1920x1080 crop]
                             3280x2464 [21.19 fps - (0, 0)/3280x2464 crop]
           'SRGGB8' : 640x480 [103.33 fps - (1000, 752)/1280x960 crop]
                      1640x1232 [41.85 fps - (0, 0)/3280x2464 crop]
                      1920x1080 [47.57 fps - (680, 692)/1920x1080 crop]
                      3280x2464 [21.19 fps - (0, 0)/3280x2464 crop]
pi@raspberrypi4B2:~ $ 
```

### Steps to reproduce the behaviour

Watch output of "libcamera-hello --list-cameras"

### Device (s)

[...]

PR #5439, opened 2023-04-20
media: i2c: imx219: Scale the pixel clock rate for the 640x480 mode

6by9:
The 640x480 mode uses a special binning mode for high framerate operation where the pixel rate is effectively doubled. Account for this when setting up the pixel clock rate, and applying the vblank and exposure controls.


6by9:
Untested, but should resolve #5438.

Hermann-SW:
Thanks, does resolve, see https://github.com/raspberrypi/linux/issues/5438#issuecomment-1516734111

PR #5440, opened 2023-04-21
hwmon: emc2305: Add calls to initialise of cooling maps

6by9:
Commit 46ef9d4ed26b ("hwmon: emc2305: fixups for driver submitted to mailing lists") missed adding the call to thermal_of_cooling_device_register required to configure any cooling maps for the device, hence stopping it from actually ever changing speed.

### DIFF
--- a/drivers/hwmon/emc2305.c
+++ b/drivers/hwmon/emc2305.c
@@ -348,9 +348,17 @@ static int emc2305_set_single_tz(struct device *dev, int idx)
 	cdev_idx = (idx) ? idx - 1 : 0;
 	pwm = data->pwm_min[cdev_idx];
 
-	data->cdev_data[cdev_idx].cdev =
-		thermal_cooling_device_register(emc2305_fan_name[idx], data,
-						&emc2305_cooling_ops);
+	if (dev->of_node)
+		data->cdev_data[cdev_idx].cdev =
+			devm_thermal_of_cooling_device_register(dev, dev->of_node,
+								emc2305_fan_name[idx],
+								data,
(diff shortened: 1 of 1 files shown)


ilg-ul:
I just updated to **6.1.34-v8+** via `rpi-update` and the `/sys/class/thermal/cooling_device0` still does not show. :-(

The full story is at <https://forums.raspberrypi.com/viewtopic.php?p=2115841#p2115730>.

Is there anything else that must be configured to make the fan work automatically?

Any suggestion on how to diagnose this will be highly appreciated.

ilg-ul:
I commented out `i2c_vc=on` and explicitly added `i2c_csi_dsi` to `i2c-fan`, and the fan now works as expected.

```console
$ grep i2c /boot/config.txt 
dtparam=i2c_arm=on
#dtparam=i2c_vc=on
dtoverlay=i2c-fan,emc2301,i2c_csi_dsi
```

However, there is a small drawback, I can no longer monitor the fan RPM, since the the 0x2f device is no longer visible.

Is there any trick to read the fan RPM without affecting the automation?

pelwell:
I don't have an EMC2305 to test with, but you may find something in /sys/class/hwmon:
```
$ find /sys/class/hwmon -name "*fan*"
```
I would expect there to be a `fan<n>_input` value which shows the speed.

ilg-ul:
Thank you, Phil, your hint was useful.

The variable that returns the actual RPM is:

```sh
cat /sys/class/hwmon/hwmon2/fan1_input
```

PR #5442, opened 2023-04-24
arm/boot/dts/overlays/mipi-dbi-spi: fix default brightness

dlech:
There is an off-by-one error in the default brightness for the PWM backlight in the MIPI DBI SPI overlay that produces the following message in the kernel logs:

    pwm-backlight backlight_pwm: invalid default brightness level: 16, using 15

The value is 0-based, so the max brightness is 15, not 16.


pelwell:
Thanks!

PR #5444, opened 2023-04-24
drm/vc4: Limit max_bpc to 8 on Pi0-3

6by9:
Pi 0-3 have no deep colour support and only 24bpp output, so max_bpc should remain as 8.


popcornmix:
LGTM

PR #5452, opened 2023-04-28
[backport rpi-5.10.y] dtoverlays: Add overlays for the IMX708 image sensor

quaresmajose:


6by9:
rpi-5.10.y doesn't have the imx708 driver, so why are you adding the overlays for it?

That branch is obsolete and hasn't been updated since Aug 2022, so there's little reason for backporting things.

pelwell:
Yes, we'll only accept major bugfixes into rpi-5.10.y.

PR #5456, opened 2023-05-03
drm/v3d: New debugfs end-points to query GPU usage stats.

popcornmix:
Two new debugfs interfaces are implemented:

- gpu_usage: exposes the total runtime since boot of each of the 5 scheduling queues available at V3D (BIN, RENDER, CSD, TFU, CACHE_CLEAN). So if the interface is queried at two different points of time the usage percentage of each of the queues can be calculated.

- gpu_pid_usage: exposes the same information but to the level of detail of each process using the V3D driver. The runtime for process using the driver is stored. So the percentages of usage by PID can be calculated with measures at different timestamps.

The storage of gpu_pid_usage stats is only done if the debugfs interface is polled during the last 70 seconds. If a process does not submit a GPU job during last 70 seconds its stats will also be purged.

### DIFF
--- a/drivers/gpu/drm/v3d/v3d_debugfs.c
+++ b/drivers/gpu/drm/v3d/v3d_debugfs.c
@@ -6,6 +6,7 @@
 #include <linux/debugfs.h>
 #include <linux/seq_file.h>
 #include <linux/string_helpers.h>
+#include <linux/sched/clock.h>
 
 #include <drm/drm_debugfs.h>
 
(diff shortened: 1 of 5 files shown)


popcornmix:
Here is an example script (run with sudo) to view gpu usage with this kernel
[gpu_usage.py.txt](https://github.com/raspberrypi/linux/files/11386595/gpu_usage.py.txt)

popcornmix:
@spl237 Once the CI workflows have done their thing, you should be able to do `sudo rpi-update pulls/5456` to update your system to a kernel with this patch.

popcornmix:
I'm not sure it's worth fixing checkpatch warning `Error: WARNING: Prefer seq_puts to seq_printf`
when seq_printfis already used here.

popcornmix:
@pelwell okay to merge?

PR #5457, opened 2023-05-05
pinctrl: bcm2835: Workaround for edge IRQ loss

pelwell:
It has been observed that edge events can be lost when GPIO edges occur close to each other. Investigation suggests this is due to a hardware bug, although no mechanism has been identified.

Work around the event loss by moving the IRQ acknowledgement into the main ISR, adding missing events by explicit level-change detection.

See: https://forums.raspberrypi.com/viewtopic.php?t=350295

### DIFF
--- a/drivers/pinctrl/bcm/pinctrl-bcm2835.c
+++ b/drivers/pinctrl/bcm/pinctrl-bcm2835.c
@@ -412,14 +412,31 @@ static void bcm2835_gpio_irq_handle_bank(struct bcm2835_pinctrl *pc,
 	unsigned long events;
 	unsigned offset;
 	unsigned gpio;
+	u32 levs, levs2;
 
 	events = bcm2835_gpio_rd(pc, GPEDS0 + bank * 4);
+	levs = bcm2835_gpio_rd(pc, GPLEV0 + bank * 4);
 	events &= mask;
 	events &= pc->enabled_irq_map[bank];
+	bcm2835_gpio_wr(pc, GPEDS0 + bank * 4, events);
+
+retry:
 	for_each_set_bit(offset, &events, 32) {
 		gpio = (32 * bank) + offset;
 		generic_handle_domain_irq(pc->gpio_chip.irq.domain,
 					  gpio);
+	}
+	events = bcm2835_gpio_rd(pc, GPEDS0 + bank * 4);
+	levs2 = bcm2835_gpio_rd(pc, GPLEV0 + bank * 4);
+
+	events |= levs2 & ~levs & bcm2835_gpio_rd(pc, GPREN0 + bank * 4);
+	events |= ~levs2 & levs & bcm2835_gpio_rd(pc, GPFEN0 + bank * 4);
+	events &= mask;
+	events &= pc->enabled_irq_map[bank];
+	if (events) {
+		bcm2835_gpio_wr(pc, GPEDS0 + bank * 4, events);
+		levs = levs2;
+		goto retry;
 	}
 }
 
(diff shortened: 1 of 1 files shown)


pelwell:
W.r.t. the checkpatch warning, "acknowledgement" is the correct UK spelling.

marcarla:
Dear Pelwell,

Many thanks for the workaround, it seems to work nice.

I wonder if another approach may be viable, mainly if one is interested both to rising and falling edges.  In the irqflow module (https://github.com/marcarla/RPi_gpio_interrupt_check), using level instead of edge interrupts, I switched between the high and low levels using the irq_set_irq_type() function. I mean, irq_set_irq_type() was called from inside the interrupt handler at each service to register the other level for the next service. All tests gave positive results: no interrupt lost and (up to now) no freeze or panic. The time overhead of irq_set_irq_type() to the handler is 3 ... 4 us. The doubt is: is the use of irq_set_irq_type() correct in interrupt context? I could find no explicit statement about this. Thanks if you can help to clear this point.

pelwell:
Why are you looking for a different workaround? Is there something undesirable about this one?

marcarla:
Absolutely no! For what I can see, your workaround works quite well.

I had been considering the other way, that I have described, before
seeing your patch. And I remained with the curiosity if it is safe
the use of irq_set_irq_type() in interrupt context. When working with
drivers its use may come handy. The use of disable_irq() is explicitly
forbidden in interrupt context. I could find no indication about
irq_set_irq_type() and I cannot trust the fact that in a test, whatever
long, my system did not crash. Can you give me this piece of information?

Thanks.

pelwell:
I can't see anything that would make irq_set_irq_type unsafe in an IRQ handler - just try it. However, apart from changing timing I can't see how it might address the problem of the disappearing events.

marcarla:
Well, the point is not to use any more edges for interrupts, but only levels.
I setup to wait, e.g., a high level. When the interrupt  is triggered by the high level, the interrupt service routine is executed and, first of all, it changes the setup, requesting a low-level interrupt. At next execution the setup is again switched to a high level and so on. Clearly, what was a single rising edge interrupt now is doubled into a high level and a low level interrupt plus the overhead of irq_set_irq_type(), that is about 3-4 us each time. Not a small burden, all considering, however in the test it worked. So, I repeat, my interest is purely cultural. If it is possible (and safe) to use irq_set_irq_type in interrupt context, one day it may come out useful. Many thanks for your attention and your reply.

pelwell:
I follow your logic now, but you are assuming that it is the "edge-ness" of the interrupts that causes the problem. We're unlikely to really understand where the events get lost, but my instinct is that it is around accesses to the GPEDS registers, rather than in the edge detectors themselves. Level high and level low events are also routed through the same GPEDS registers, so I don't think there's any guarantee that your scheme will be be immune to event loss. But if you are curious and don't mind the possibility of failure then I'll be interested to hear the outcome.

marcarla:
Indeed I am very curious, and your interest encouraged me to spend a rainy afternoon to make some more tests.

The irqflow module, already described in https://github.com/marcarla/RPi_gpio_interrupt_check, has been modified to wait in a sequence for a high and low level, instead of a rising and falling edge, changing the setup in the interrupt handler with irq_set_irq_type.
 
A first test was made to obtain the histogram of missed interrupts as a function of the delay between signals to gpio21 and gpio16, with same test conditions as described in https://forums.raspberrypi.com/viewtopic.php?t=350295.

No interrupt was lost, but this is not a surprise: with levels, if a request goes to be canceled without being processed, it is immediately repeated, as long as the line level does not change. It happened I verified this point the hard way, when my ISR did not take proper action: the CPU was flooded by continuously repeating requests.

I was certain, however, that some interrupt requests should go lost because of the GPEDS registers, for the same reason it happens in your workaround if the levels on the gpios are too short in time. So I modified the signals: gpio21 was fed with the usual 1 kHz square wave, while signal to gpio16 was modified into a 50 ns pulse, with 1000.001 Hz repetition rate. This way the pulse slowly travels the whole 1 ms length of gpio21 square wave in 1000 sec.

To my very great surprise, no interrupt was lost in a 6000 sec test.

The test has been made a first time looking for the L->H level change, then it has been repeated for the H->L level change and again using a 999.95 us long pulse. No one interrupt was ever missed.

It looks like the interrupt system in RPi works quite well with levels.

These are the results, if I am not making some big mistake. Let me know if in your opinion any other test may be useful.

PR #5458, opened 2023-05-05
Fixes for DMA40 and use for hdmi audio

popcornmix:
The desire to move from a DMA lite channel to a DMA40 channel
on 2711 arises from discussion with Broadcom about losing samples
with HBR audio when system is busy.

It turns out the DMA40 code path has never been used and there 
are a number of fatal issues with the code, so this PR fixes those first.

The AXI path from DMA controller to HDMI audio fifo
is long, and may have considerable delay.
    
When using DMA without waiting for responses it is
very easy to overfill the fifo as when the fifo
removes DREQ there may be large numbers of writes
in flight.
    
This means the DREQ fifo threshold must be set low
enough to accommodate the maximum number of in flight
writes (unknown by something like 24),
which means the 32 element fifo only requests data
when it contains fewer than 8 entries, making it
susceptible to underflow.
    
If we wait for write responses we can set the DREQ
fifo threshold much higher as there are a controlled
number of writes in flight. That makes us more robust
to underflow.

However the overall bandwidth is reduced by setting
this, so also want to enable a burstsize of 4 to improve
bandwidth.


drivers/dma/bcm2835-dma.c
```diff
@@ -818,20 +818,25 @@ static enum dma_status bcm2835_dma_tx_status(struct dma_chan *chan,
 		struct bcm2835_desc *d = c->desc;
 		dma_addr_t pos;
 
-		if (d->dir == DMA_MEM_TO_DEV && c->is_40bit_channel)
```
6by9, line 822: Commit text could be expanded, and would normally reference bcm2835-dma rather than just "dma".

drivers/dma/bcm2835-dma.c
```diff
@@ -673,20 +673,43 @@ static void bcm2835_dma_abort(struct bcm2835_chan *c)
...
+		/*
+		* A zero control block address means the channel is idle.
+		* (The ACTIVE flag in the CS register is not a reliable indicator.)
+		*/
+		if (!readl(chan_base + BCM2835_DMA_ADDR))
+			return;
+
 		/* Write 0 to the active bit - Pause the DMA */
-		writel(0, chan_base + BCM2835_DMA_CS);
+		writel(readl(chan_base + BCM2835_DMA_CS) &~ BCM2835_DMA_ACTIVE,
```
6by9: Whitespace as per checkpatch warning, 
```
readl(chan_base + BCM2835_DMA_CS) & ~BCM2835_DMA_ACTIVE,
```

drivers/gpu/drm/vc4/vc4_hdmi.c
```diff
@@ -2766,9 +2767,14 @@ static int vc4_hdmi_audio_init(struct vc4_hdmi *vc4_hdmi)
 	if (index < 0)
 		index = 1;
 
+	/*
+	 * TODO: Retrieve the address through the dma-ranges mapping
+	 */
 	addr = of_get_address(dev->of_node, index, NULL, NULL);
+	complete_addr = (u64)0x4 << 32;
+	complete_addr |= be32_to_cpup(addr);
```
pelwell: Can you explain these lines? They appear to be adding an 0x400000000 offset.

drivers/dma/bcm2835-dma.c
```diff
@@ -284,7 +290,7 @@ struct bcm2835_desc {
 /* the max dma length for different channels */
 #define MAX_DMA40_LEN SZ_1G
 
-#define BCM2711_DMA40_BURST_LEN(x)	((min(x,16) - 1) << 8)
+#define BCM2711_DMA40_BURST_LEN(x)	(((x) & 15) << 8)
```
6by9, line 297: This has changed behaviour. Was it just totally wrong to start with, or something more subtle?

drivers/gpu/drm/vc4/vc4_hdmi.c
popcornmix: The 40-bit alias for peripherals is 0x4. We need that 4 in the DMA DESTI high bits of peripheral address.

pelwell: That knowledge does not belong here. c.f. the second commit in this series

drivers/dma/bcm2835-dma.c
popcornmix: Just to clarify - the dma40 code has never been run before (in locks up kernel in any possibly use), so regressions aren't an issue.

This is semantics.  A burst describes the number of additional beats added to the original access.
Storing 0 in this field would just be the original access.
Storing 3 in this field would be original access plus 3 extra beats.

I think it makes sense to consider this fields as "extra burst beats", rather than "total burst beats".
That is how the non-dma40 code works, and how the hardware works.

drivers/gpu/drm/vc4/vc4_hdmi.c
popcornmix: Okay - I think the pair of commits can be dropped.

pelwell: And in return I'll find a kosher way to achieve the same result.

popcornmix:
A few of these commits were in @mripard 's tree, so I've switched to those.
Code is identical, but he may have a more upstream friendly commit message.

popcornmix:
Any remaining issues with this?
The checkpatch issue is just not finding a fixes commit hash (which is in this tree).

pelwell:
I'd merge if I could - GitHub says no.

popcornmix:
I've rebased.

pelwell:
This is causing problems for SPI (at least): https://forums.raspberrypi.com/viewtopic.php?p=2108782#p2108782
For regular DMA channels I get:
```
[   12.656306] bcm2835-dma fe007000.dma: failed to abort dma
```
and for 40-bit channels it's:
```
[   98.834943] bcm2835-dma fe007b00.dma: failed to halt dma
```
Reverting 9cc6a5314b8a (easy) and 2510319a3a21 (less easy, but works if you also revert 654368fe848c) fixes it with no obvious side effects.

BTW, I think 654368fe848c is pointless for us because the PROT fields can only be written in secure mode. Is there a use case where this has been problematic?

popcornmix:
Try reading DMA registers with busybox devmem after the PROT fields have been written to 0.
The whole DMA channel reads back as zero (including the fixed bits like dma channel ID in debug register).
I think kernel *can* set them, but it makes debug tools impossible to use.

pelwell:
> I think kernel can set them, but it makes debug tools impossible to use.

How does that work? Both should be operating at supervisor level (with `sudo busybox devmem`) - EL1 on aarch64.

popcornmix:
Try it - I just have.
Run dma.py (from pyregs) when audio is playing with and without the PROT commit.

pelwell:
It seems that the devmem accesses come through with a different protection level, but it would be good to understand how the kernel can be performing a secure access. I propose reverting all three patches and reworking the PROT patch accordingly.

popcornmix:
Do you have a means I can reproduce this?
I'd like to step through how the halt is failing.

PR #5459, opened 2023-05-08
Bluetooth: btusb: Improve support for Actions Semi ATS2851 based devices

EzPzDev:
Add two more quirks to resume the device initialization and basic operation as the device seems not to support "Read Transmit Power" and "Set Extended Scan Parameters".

### DIFF
--- a/drivers/bluetooth/btusb.c
+++ b/drivers/bluetooth/btusb.c
@@ -4017,6 +4017,7 @@ static int btusb_probe(struct usb_interface *intf,
 		/* Support is advertised, but not implemented */
 		set_bit(HCI_QUIRK_BROKEN_ERR_DATA_REPORTING, &hdev->quirks);
 		set_bit(HCI_QUIRK_BROKEN_READ_TRANSMIT_POWER, &hdev->quirks);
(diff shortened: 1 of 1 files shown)


pelwell:
@holtmann You added the HCI_QUIRK_BROKEN_ERR_DATA_REPORTING quirk for the ATS2851. Does also setting HCI_QUIRK_BROKEN_READ_TRANSMIT_POWER and HCI_QUIRK_BROKEN_EXT_SCAN sound reasonable to you?

EzPzDev:
Sorry, this was originally from upstream. I forgot to add Co-authored-by: Raul Cheleguini.

pelwell:
Unfortunately what you've done is to remove the useful commit message and delay the approval process. We are (almost) always happy to accept back-ports from newer kernels because it means i) that they have been approved, and ii) that the minimal support burden disappears once we reach that kernel version.

In future, cherry-pick the back-ports, adding a line at the start of the commit message saying "commit <hash> upstream" but otherwise leaving them unchanged, Alternatively, create an Issue asking us to back-port them.

pelwell:
See https://github.com/raspberrypi/linux/pull/5465 for how it should be done.

EzPzDev:
Thank you for the explanation! I was unaware, but you've provided useful information I will use in the future.

PR #5443, opened 2023-04-24
i2c-mux and i2c-pwm overlay updates for alternate i2c buses

6by9:

PR #5461, opened 2023-05-09
Set default pins for the extra I/O interfaces

pelwell:
As per the discussion in #5443, add the references to the default pinctrl groups for the BCM2711 additional I/O interfaces.


6by9:
Works for me on i2c4 (once I'd noticed the default is GPIOs 8&9, not 6&7!).
dtoverlay=i2c-mux,i2c3 also sets up the right GPIOs now.

PR #5463, opened 2023-05-10
fbtft: Flush any deferred updates on release

mfudpi:
Since [1], writing to the fbtft framebuffer and immediately closing the file descriptor is likely to have no effect. This is because the framebuffer deferred IO framework is now careful to cancel pending updates.

To retain the old behaviour, add an fb_release method that flushes the updates instead.

See: #5399 

[1] 3efc61d ("fbdev: Fix invalid page access after closing  #deferred I/O devices")

### DIFF
--- a/drivers/staging/fbtft/fbtft-core.c
+++ b/drivers/staging/fbtft/fbtft-core.c
@@ -475,6 +475,15 @@ static int fbtft_fb_blank(int blank, struct fb_info *info)
 	return ret;
 }
 
+static int fbtft_fb_release(struct fb_info *info, int user)
+{
+	/* Flush any pending updates */
+	cancel_delayed_work(&info->deferred_work);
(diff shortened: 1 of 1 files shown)


pelwell:
You're submitting a minor tweak on my PR (#5399) but without mentioning the fact in the PR? I guess the See: counts.

If you'd read around a bit more you'd find that I submitted an alternative workaround (#5413) that should work for all drivers using fb_deferred_io without causing slow-down for user-space code that rapidly opens and closes the device.

mfudpi:
Thanks for pointing me that direction, must have missed your alternative workaround.

Unfortunately #5413 does not work in my case.

Writing to the framebuffer and closing the file descriptor immediately does not have any effect.
Running `cat /dev/urandom > /dev/fb0` afterwards does not update the display.

PR #5466, opened 2023-05-11
Handle DMA slave address mapping correctly

pelwell:
Contrary to popular belief, DMA slave addresses are meant to be passed as CPU physical addresses, not DMA addresses (see https://github.com/raspberrypi/linux/blob/rpi-6.1.y/include/linux/dmaengine.h#L437 for proof). Doing so allows the DMA controller to perform the conversion in its own context using a simple call to `phys_to_dma`.

This series:
1. fixes up the DMA address mappings as defined by the Device Tree `dma-ranges` property,
2. modifies the DMA driver to do the translation,
3. ensures that the client drivers pass physical addresses,
4. as part of the bcm2835-sdhost changes, retrieves the address of the system timer (used for some logging) from DT, and
5. adds an `i2s_dma4` parameter that selects 40-bit DMA for I2S sound cards.

The fact that the resulting code is shorter supports the idea that this is the right approach.


pelwell:
This series looks long, but most of the patches are similar changes to 7 slave DMA client drivers.

popcornmix:
Nice. Looks a lot cleaner.

PR #5467, opened 2023-05-11
Rpi 6.4.y imx258 patches

6by9:
I will look at upstreaming these to save grief again later. The sensor is being used on other platforms, and others have made comment that these will be useful there too.


popcornmix:
Thanks.

PR #5476, opened 2023-05-19
WIP: input: edt-ft5x06: Only read data for number of points reported

6by9:
Rather than always reading the maximum number of points supported by the chip (which may be as high as 10), read the number of active points first, and read data for just those. In most cases this will result in less data on the I2C bus, with only the maximum touch points taking more due to a second read that has to configure the start address.


pelwell:
Apologies if you wanted to do more testing - I registered the WIP after studying the patch and merging (there is a Draft PR status that is harder to ignore).

PR #5479, opened 2023-05-23
Fix I2C issues for DSI touch and V3 camera

6by9:


6by9:
Convert back to a draft based on discussions for always cleanly flushing FIFOs through the normal code paths, and implementing the I2C_M_IGNORE_NAK flag.

PR #5482, opened 2023-05-25
bcm2835-dma: Fixes for dma_abort

popcornmix:
There is a problem with the current abort scheme
when dma is blocked on a DREQ which prevents halting.

This is triggered by SPI driver which aborts dma
in this state and so leads to a halt timeout.

Discussion with Broadcom suggests the sequence:
```
CS.ACTIVE=0
while (CS.OUTSTANDING_TRANSACTIONS == 0)
  wait()
DEBUG.RESET=1
```
should be safe on a dma40 channel.

Unfortunately the non-dma40 channels don't have
OUTSTANDING_TRANSACTIONS, only WAITING_FOR_OUTSTANDING_WRITES which doesn't consider reads.

### DIFF
--- a/drivers/dma/bcm2835-dma.c
+++ b/drivers/dma/bcm2835-dma.c
@@ -245,6 +245,7 @@ struct bcm2835_desc {
 #define BCM2711_DMA40_ERR		BIT(10)
 #define BCM2711_DMA40_QOS(x)		(((x) & 0x1f) << 16)
 #define BCM2711_DMA40_PANIC_QOS(x)	(((x) & 0x1f) << 20)
+#define BCM2711_DMA40_TRANSACTIONS	BIT(25)
 #define BCM2711_DMA40_WAIT_FOR_WRITES	BIT(28)
 #define BCM2711_DMA40_DISDEBUG		BIT(29)
 #define BCM2711_DMA40_ABORT		BIT(30)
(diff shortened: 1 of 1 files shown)


popcornmix:
This seems to behave as desired for HDMI audio and SPI display.
I'm waiting for feedback on the non-dma40 case which may be unsafe with outstanding reads.

drivers/dma/bcm2835-dma.c
```diff
@@ -245,6 +245,7 @@ struct bcm2835_desc {
 #define BCM2711_DMA40_ERR		BIT(10)
```
pelwell, line 247: Can we have the `BCM2711_DMA40_PROT` bits here?

popcornmix:
Updated to do something plausible for non-dma40 case.
We do have to rely on a timeout (I've chosen 100 loops which equates to 20us, which could be increased).

popcornmix:
@pelwell any comments?

pelwell (commented):
Sorry, forgot to submit the final review. The rest looks OK, but then it looked OK before.

drivers/dma/bcm2835-dma.c
popcornmix: Added commit to move PROT bits definition.

PR #5484, opened 2023-05-26
panel-sitronix-st7701: Fix panel prepare over SPI

jackoalan:
A DSI write is issued in st7701_prepare even when the probed panel runs on SPI. In practice, this results in a panic with the vc4-kms-dpi-hyperpixel2r overlay active.

Perform the equivalent write over SPI in this case.


pelwell:
Thanks!

PR #5486, opened 2023-05-31
drivers: media: imx296: Disable 2x2 binned mode

naushir:
Disable enumerating the 2x2 binned mode entirely as it does not seem to work for either mono or colour sensor variants.

### DIFF
--- a/drivers/media/i2c/imx296.c
+++ b/drivers/media/i2c/imx296.c
@@ -745,12 +745,14 @@ static int imx296_enum_frame_size(struct v4l2_subdev *sd,
 {
 	const struct imx296 *sensor = to_imx296(sd);
 	const struct v4l2_mbus_framefmt *format;
-	/* Binning only works on the mono sensor variant */
-	unsigned int max_index = sensor->mono ? 2 : 1;
 
 	format = v4l2_subdev_get_pad_format(sd, state, fse->pad);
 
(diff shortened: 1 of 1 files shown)


naushir:
@popcornmix or @pelwell can we merge this when you get a chance please?

drivers/media/i2c/imx296.c
```diff
@@ -745,12 +745,14 @@ static int imx296_enum_frame_size(struct v4l2_subdev *sd,
...
-	unsigned int max_index = sensor->mono ? 2 : 1;
 
 	format = v4l2_subdev_get_pad_format(sd, state, fse->pad);
 
-	if (fse->index >= max_index || fse->code != imx296_mbus_code(sensor))
+	/*
+	 * Binning does not seem to work on either mono or colour sensor
+	 * variants. Disable enumerating the binned frame size for now.
+	 */
+	if (fse->index >= 1 || fse->code != imx296_mbus_code(sensor))
```
6by9, line 755: You've disabled the binned mode being enumerated which will cover libcamera, but imx296_set_format will still round to the nearest integer divide of the sensor width/height.

naushir: Yes, good point, I'll remove any reference to binning from there.

6by9:
LGTM

PR #5488, opened 2023-06-01
Add PSI support to for Waydroid

ghost:
Raspberry Pi users cannot use Waydroid without this.


pelwell:
See also https://github.com/raspberrypi/linux/issues/4683.

PR #5499, opened 2023-06-12
Remove the Zero 2 W DT-based WLAN firmware selection

pelwell:
Once https://github.com/RPi-Distro/firmware-nonfree/pull/37 has been merged, packaged, published, and several aeons allowed to pass, it will be safe to merge this patch removing the downstream-specific DT-based firmware selection mechanism.


nbuchwitz:
Even if no aeon has passed yet, I give it a shot and ask if this maybe can be merged? This would end some annoying false positive error / warnings with custom da

pelwell:
The necessary firmware symlinks are in current images. Anyone installing just the latest kernel on an old image could be caught out, but that's always going to come with a risk.

nbuchwitz:
Thanks, Phil!

PR #5507, opened 2023-06-16
drivers: media: bcm2835_unicam: Improve frame sequence count handling

naushir:
Ensure that the frame sequence counter is incremented only if a previous frame start interrupt has occurred, or a frame start + frame end has occurred simultaneously.

This corresponds the sequence number with the actual number of frames produced by the sensor, not the number of frame buffers dequeued back to userland.


naushir:
@6by9 this is what I have come up with.  Have I missed any other conditioning in the logic?

6by9:
Looks reasonable to me.

naushir:
@pelwell feel free to merge this if you are happy with the change.

PR #5508, opened 2023-06-19
dtoverlays: Fix pitft[28|35] overlays for 6.1 driver change.

6by9:
The overlays configured both irq-gpio and an interrupts/ interrupt-parent configuration for the stmpe MFD device.

irq-gpio was reworked in 6.1 and has issues with the configuration as provided. Removing it and using the interrupts/interrupt-parent configuration works fine, so do that.

### DIFF
--- a/arch/arm/boot/dts/overlays/pitft28-resistive-overlay.dts
+++ b/arch/arm/boot/dts/overlays/pitft28-resistive-overlay.dts
@@ -68,7 +68,6 @@
 				reg = <1>;
 
 				spi-max-frequency = <500000>;
-				irq-gpio = <&gpio 24 0x2>; /* IRQF_TRIGGER_FALLING */
 				interrupts = <24 2>; /* high-to-low edge triggered */
 				interrupt-parent = <&gpio>;
 				interrupt-controller;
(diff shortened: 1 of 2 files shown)


pelwell:
Do you prefer the split (interrupts, interrupt-parent) notation to the combined irq-gpio style?

6by9:
> Do you prefer the split (interrupts, interrupt-parent) notation to the combined irq-gpio style?

I couldn't get the irq-gpio route to work correctly.
https://forums.raspberrypi.com/viewtopic.php?t=351394

The irq-gpio route appears to no longer support edge triggering after conversion from gpio to gpiod calls, but it configures the hardware in a way that seems to leave the STMPE in the IRQ active state, therefore nearly spinning.

Switching to the interrupts/interrupt-parent route still allows use of edge triggering, and works as it used.

pelwell:
Fair enough - I've squeezed a link to the Forums thread into the commit message.

6by9:
> Fair enough - I've squeezed a link to the Forums thread into the commit message.

Ta, I regularly forget to add that bit.

PR #5510, opened 2023-06-20
ASoC: cs43130: Fix numerator/denominator mixup

pelwell:
In converting to using the standard u16_fract type, commit [1] made the obvious mistake and failed to take account of the difference in numerator and denominator ordering, breaking all uses of the cs43130 codec.

Fix it.

[1] commit e14bd35ef446 ("ASoC: cs43130: Re-use generic struct u16_fract")

Fixes: e14bd35ef446 ("ASoC: cs43130: Re-use generic struct u16_fract")


popcornmix:
LGTM

PR #5518, opened 2023-06-25
Fix HDMI audio passthrough issues

HiassofT:
ALSA didn't set up proper constraints for compressed audio (eg 192kHz 8ch for DTS-HD and Dolby TrueHD) which prevented applications like kodi to pass through those stream to TVs/AVRs.

@popcornmix also reported that his Sony TV refused to decode TrueHD streams because the channel info in the audio infoframe was incorrect.

We've carried both fixes in LibreELEC for a while and they were just accepted upstream and have been queued up for linux-next

### DIFF
--- a/sound/core/pcm_drm_eld.c
+++ b/sound/core/pcm_drm_eld.c
@@ -2,10 +2,24 @@
 /*
  *  PCM DRM helpers
  */
+#include <linux/bitfield.h>
 #include <linux/export.h>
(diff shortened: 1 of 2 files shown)


popcornmix:
I've been running with these for a couple of months, and they fixed an issue we saw with a Sony TV.
+1 from me.

popcornmix:
Thanks!

PR #5526, opened 2023-07-01
smsc95xx MAC address cleanups

l1k:
Here's an assortment of cleanups for the `smsc95xx` driver's parsing of the `smsc95xx.macaddr` command line parameter.

No functional change intended, just cleanups.

I'm unsure whether to submit for rpi-6.4.y or rpi-6.1.y. I'm choosing the former in the hope that this will get the commits applied to future branches such as the forthcoming rpi-6.5.y. (Barring any objections to the commits of course.)

Thanks!


l1k:
FWIW, the issues reported by checkpatch for commit [1/3] are addressed by the subsequent commits in this pull request, so I'm inclined to consider them as false-positives. Let me know if you think otherwise.

PR #5527, opened 2023-07-03
Fix Clang 16 warnings

dereference23:


drivers/usb/host/dwc_otg/dwc_otg_hcd_intr.c
```diff
@@ -402,7 +402,7 @@ int32_t dwc_otg_hcd_handle_rx_status_q_level_intr(dwc_otg_hcd_t * dwc_otg_hcd)
 			hc->xfer_count += grxsts.b.bcnt;
 			hc->xfer_buff += grxsts.b.bcnt;
 		}
-
+		fallthrough;
```
popcornmix: As the fallthrough just results in break, it seems more natural to just break here.
@P33M any reason not to do that?

popcornmix:
Looks good to me (other than minor comment).

Issue #5535, opened 2023-07-12
Debian certificates for regulatory.db missing


nbuchwitz:
### Describe the bug

Debian distributes the wireless regulatory database in the package `wireless-regdb`. The package contains the upstream db as well as a debian version and manages it through update-alternatives. Per default the Debian database is used:

```
pi@RevPi99986:~ $ ls -l /lib/firmware/regulatory.db*
lrwxrwxrwx 1 root root   31  3. Mai 04:53 /lib/firmware/regulatory.db -> /etc/alternatives/regulatory.db
-rw-r--r-- 1 root root 4388 30. Jun 2022  /lib/firmware/regulatory.db-debian
lrwxrwxrwx 1 root root   35  3. Mai 04:53 /lib/firmware/regulatory.db.p7s -> /etc/alternatives/regulatory.db.p7s
-rw-r--r-- 1 root root 1225 30. Jun 2022  /lib/firmware/regulatory.db.p7s-debian
-rw-r--r-- 1 root root 1182 30. Jun 2022  /lib/firmware/regulatory.db.p7s-upstream
-rw-r--r-- 1 root root 4388 30. Jun 2022  /lib/firmware/regulatory.db-upstream
pi@RevPi99986:~ $ ls -l /etc/alternatives/regulatory.db*
lrwxrwxrwx 1 root root 34 12. Jul 10:08 /etc/alternatives/regulatory.db -> /lib/firmware/regulatory.db-debian
lrwxrwxrwx 1 root root 38 12. Jul 10:08 /etc/alternatives/regulatory.db.p7s -> /lib/firmware/regulatory.db.p7s-debian
```

The database is signed and the signature checked by the kernel (certs need to be present in `net/wireless/certs`). This is fine as long as the official upstream version is used (cert present, signature can be verified). The Debian database though is signed by Ben and Romain and the certificates are added in the Debian build process:

https://salsa.debian.org/kernel-team/linux/-/blob/master/debian/patches/debian/wireless-add-debian-wireless-regdb-certificates.patch

On a Raspberry Pi kernel these signatures are not present and the user will find a error in the kernel log:

```
pi@cm4io:~ $ dmesg |grep cfg80211
[    5.515935] cfg80211: Loading compiled-in X.509 certificates for regulatory database
[    5.558783] cfg80211: Loaded X.509 cert 'sforshee: 00b28ddf47aef9cea7'
[    5.564276] cfg80211: loaded regulatory.db is malformed or signature is missing/invalid
[...]

PR #5536, opened 2023-07-12
cfg80211: ship debian certificates as hex files

nbuchwitz:
Loading the regulatory database from the debian files fails with

"loaded regulatory.db is malformed or signature is missing/invalid"

due to missing certificates.  Add these debian-specific certificates from upstream to fix this error.

The certificates have been imported as following:

```
patch -p1 <<<$(
curl https://salsa.debian.org/kernel-team/linux/-/raw/\
master/debian/patches/debian/\
wireless-add-debian-wireless-regdb-certificates.patch
)
```

See #5535 for details


l1k:
LGTM.

While a case could be made that the patch should be retrieved and applied during the build process of the raspberrypi-kernel package (to ensure the latest keys are used), that approach would have the disadvantage of requiring Internet connectivity.

nbuchwitz:
Valid point, but personally I'd opt for the manual approach in order to:

a) keep control about what 3rd party keys are integrated
b) be resilient against connectivity issues or other down times of the salsa gitlab

pelwell:
That works for me. I slightly object to two more lines of log spam, and having to patch the kernel in the first place for the privilege of using the Debian regulatory database, but that's where we are.

PR #5537, opened 2023-07-12
drm: vc4: Fixup duplicated macro definition in vc4_firmware_kms

6by9:
Both vc4_drv.h and vc4_firmware_kms.c had definitions for to_vc4_crtc.

Rename the fkms one to make it unique, and drop the magic define vc4_crtc vc4_kms_crtc
define to_vc4_crtc to_vc4_kms_crtc
that renamed half the variable and function names in a slightly unexpected way.


6by9:
A cursory check of fkms on Pi4 seems to say it works.

popcornmix:
Works for me.

PR #5540, opened 2023-07-13
Xhci fixups

P33M:
Downstream patch updates for 6.5.y. 
Most of the trb/ring space calculation has been reworked so these downstream patches no longer need to be applied in future:
cc068f10 xhci: correct room_on_ring() for cases where there is a single segment
839b7fcb usb: xhci: account for num_trbs_free when invalidating TDs

Also collapsed the VLI_SS_BULK_OUT_BUG into its final version which actually worked around the hardware bug.


popcornmix:
Looks good to me. The downstream commits have become somewhat smaller.

(still boots with working USB).

Issue #5539, opened 2023-07-12
Machine unresponsive after Internal error: Oops: 0000000096000045 [#1] PREEMPT SMP


sfatula:
### Describe the bug

Happens every few weeks seemingly "randomly".  GUI becomes non responsive, I can connect via another machine via ssh can can't reboot as that hangs. So, end up having to power off via removing power. 

### Steps to reproduce the behaviour

I don't rreally know how to reproduce it. I am using autofs if that matters. It just seems to happen every few weeks. Always plenty of memory available on 8gb pi. 

### Device (s)

Raspberry Pi 4 Mod. B

### System

```
cat /etc/rpi-issue
Raspberry Pi reference 2022-04-04
Generated using pi-gen, https://github.com/RPi-Distro/pi-gen, 27a8050c3c06e567c794620394a8c2d74262a516, stage4
```

```
vcgencmd version
Mar 17 2023 10:50:39 
Copyright (c) 2012 Broadcom
version 82f3750a65fadae9a38077e3c2e217ad158c8d54 (clean) (release) (start)
```
```
uname -a
Linux stevepi 6.1.21-v8+ #1642 SMP PREEMPT Mon Apr  3 17:24:16 BST 2023 aarch64 GNU/Linux
```

### Logs

From syslog:

Jul 12 12:37:40 stevepi kernel: [391018.643998] Unable to handle kernel NULL pointer dereference at virtual address 0000000000000008
Jul 12 12:37:40 stevepi kernel: [391018.652983] Mem abort info:
Jul 12 12:37:40 stevepi kernel: [391018.655972]   ESR = 0x0000000096000045
Jul 12 12:37:40 stevepi kernel: [391018.659867]   EC = 0x25: DABT (current EL), IL = 32 bits
Jul 12 12:37:40 stevepi kernel: [391018.665311]   SET = 0, FnV = 0
Jul 12 12:37:40 stevepi kernel: [391018.669818]   EA = 0, S1PTW = 0
Jul 12 12:37:40 stevepi kernel: [391018.673179]   FSC = 0x05: level 1 translation fault
Jul 12 12:37:40 stevepi kernel: [391018.678651] Data abort info:
Jul 12 12:37:40 stevepi kernel: [391018.681671]   ISV = 0, ISS = 0x00000045
Jul 12 12:37:40 stevepi kernel: [391018.685606]   CM = 0, WnR = 1
[...]

PR #5542, opened 2023-07-14
Possible fixes for #5539

pelwell:
Issue #5539 appears to show a crash inside smb2_reconnect_server. Newer upstream kernels include two patches that look like they may help, so this PR back-ports them.


popcornmix:
Wonder why these commits don't have a "Fixes" tag? (it sounds like a fix).
If bug reporter is happy then so am I.

pelwell:
These have already been back-ported upstream.

PR #5548, opened 2023-07-24
bcm2835-unicam.c pixel formats code tidy + 16-bit support

octopus-russell:
Replaces hardcoded CSI-2 data type literals with defines from mipi-csi2.h and adds missing 16-bit Bayer formats

### DIFF
--- a/drivers/media/platform/bcm2835/bcm2835-unicam.c
+++ b/drivers/media/platform/bcm2835/bcm2835-unicam.c
@@ -405,10 +405,33 @@ static const struct unicam_fmt formats[] = {
 		.csi_dt		= MIPI_CSI2_DT_RAW14,
 		.valid_colorspaces = MASK_CS_RAW,
 	}, {
+		.fourcc		= V4L2_PIX_FMT_SBGGR16,
+		.code		= MEDIA_BUS_FMT_SBGGR16_1X16,
+		.depth		= 16,
+		.csi_dt		= MIPI_CSI2_DT_RAW16,
+		.valid_colorspaces = MASK_CS_RAW,
+	}, {
+		.fourcc		= V4L2_PIX_FMT_SGBRG16,
+		.code		= MEDIA_BUS_FMT_SGBRG16_1X16,
+		.depth		= 16,
+		.csi_dt		= MIPI_CSI2_DT_RAW16,
+		.valid_colorspaces = MASK_CS_RAW,
+	}, {
+		.fourcc		= V4L2_PIX_FMT_SGRBG16,
+		.code		= MEDIA_BUS_FMT_SGRBG16_1X16,
+		.depth		= 16,
+		.csi_dt		= MIPI_CSI2_DT_RAW16,
+		.valid_colorspaces = MASK_CS_RAW,
+	}, {
+		.fourcc		= V4L2_PIX_FMT_SRGGB16,
+		.code		= MEDIA_BUS_FMT_SRGGB16_1X16,
+		.depth		= 16,
+		.csi_dt		= MIPI_CSI2_DT_RAW16,
+		.valid_colorspaces = MASK_CS_RAW,
+	}, {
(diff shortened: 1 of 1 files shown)


6by9:
Thanks for the patches.

The content looks fine, but there are a couple of things to address.
(Apologies if I'm teaching you to suck eggs on using git here, but it's as easy to spell it out than have you come back asking for the git assistance)

1) Each patch needs to be complete and not break the build. The header typo should be part of the "Replaced magic number CSI data types with defines" patch, and the tabs and whitespace patch as part of "Added 16-bit Bayer formats to bcm2835-unicam.c".
The easiest way to resolve this is to use "git rebase -i" to allow you to go back through the patches on your local branch. "pick" at the start of the line means the patch will be taken as is. "fixup" (or just "f") will squash that patch into the previous one. (There is also "squash" to merge this patch into the previous, but put both commit text blocks into the one commit text, and "edit" to just stop and allow you to manually edit the patch)
If you reorder your patches and use "fixup" to squash what are currently your 3rd and 4th patches into the relevant main patches, then we end up with 2 patches and no fixups.

2) This is a driver that is being upstreamed to the mainline Linux kernel. As such all contributions need to follow the upstream rules, and particularly that all patches have a Signed-off-by: line and suitable commit logs - see https://www.kernel.org/doc/html/latest/process/submitting-patches.html#sign-your-work-the-developer-s-certificate-of-origin. The easiest approach is to use "git commit -s" to add a Signed-off-by tag.
If you use "git rebase -i" again, and ask to edit your 2 patches, it'll stop after each one, and let you run "git commit --amend -s" to add that tag. "git rebase --continue" continues to the next patch when you've finished editing each one.

Once you have a nice clean set of patches, "git push -f ...." will allow you to force push the updates to the same branch as before, and the PR gets automatically updated.

6by9:
Can I follow up and ask if you were planning to make the requested changes? If not, then I'll close this PR?
Thanks.

octopus-russell:
Hi, apologies we unfortunately haven't had the development resources to complete this. Probably best close the PR for now and when we have resources we'll create a new PR.

Thanks
Russell

6by9:
No problem.

If it's OK by you, due to lack of Signed-off-by I'll take your base patches and reauthor them. I'd like to add a "Suggested-by:" tag, but that should have a full name and email address, which your commits don't have.

It's a tricky one as you could then claim copyright on the changes, but both using the defines and adding 16bit support are things we want to do, and there is only really one way to do it. This is the reason for requiring the Signed-off-by tag in the first place.

octopus-russell:
Yeah sure, that would be great thanks. I guess you just need my name/email for the suggested-by tag? As for copyright I see its a bit tricky, but the code would end up identical either way and I'm giving permission to be the 'suggested-by' tag which implies credit I think?

6by9:
> Yeah sure, that would be great thanks. I guess you just need my name/email for the suggested-by tag? 

Yes please. It will be public in the final commit, but feel free to email it to me instead (dave.stevenson @ obvious domain).

> As for copyright I see its a bit tricky, but the code would end up identical either way and I'm giving permission to be the 'suggested-by' tag which implies credit I think?

The other option is that I can keep you as author and (with your permission) add your supplied name & email address as a Signed-off-by. I can then modify the patch and add my SoB, and all is good.

I know this is a pain, but it's covering backs legally against people turning against a project and requesting that code with their copyright is removed (I seem to recall that it has happened with the Linux kernel in the past).

6by9:
My memory is obviously failing - I made these changes to rpi-6.2.y back in January via #5330.
The imx290 changes got backported to 6.1 via #5411, but not the unicam changes.

We're imminently shifting to 6.6 as the new LTS kernel, so that already uses the defines and has 16bit support. I see little point in backporting it now to 6.1, but shout if you really need it and I'll sort out a cherry-pick.

Closing.

PR #5555, opened 2023-07-28
drivers: media: imx296: Add standby delay during probe

naushir:
Add a 2-5ms delay when coming out of standby and before reading the sensor info register durning probe, as instructed by the datasheet. This standby delay is already present when the sensor starts streaming.


naushir:
@njhollinghurst and @6by9

6by9:
LGTM.

PR #5560, opened 2023-08-01
drivers: media: i2c Updated imx296 driver for external trigger

bebon901:
Updated imx296 driver to support external trigger mode via XTR pin. Added module parameter to control this mode.

### DIFF
--- a/drivers/media/i2c/imx296.c
+++ b/drivers/media/i2c/imx296.c
@@ -20,6 +20,10 @@
 #include <media/v4l2-fwnode.h>
 #include <media/v4l2-subdev.h>
 
+static int trigger_mode;
+module_param(trigger_mode, int, 0644);
+MODULE_PARM_DESC(trigger_mode, "Set trigger mode: 0=default, 1=XTRIG");
+
 #define IMX296_PIXEL_ARRAY_WIDTH			1456
 #define IMX296_PIXEL_ARRAY_HEIGHT			1088
 
@@ -645,6 +649,12 @@ static int imx296_stream_on(struct imx296 *sensor)
 
 	imx296_write(sensor, IMX296_CTRL00, 0, &ret);
 	usleep_range(2000, 5000);
+
+	if (trigger_mode == 1) {
+		imx296_write(sensor, IMX296_CTRL0B, IMX296_CTRL0B_TRIGEN, &ret);
+		imx296_write(sensor, IMX296_LOWLAGTRG,  IMX296_LOWLAGTRG_FAST, &ret);
+	}
+
 	imx296_write(sensor, IMX296_CTRL0A, 0, &ret);
 
 	/* vflip and hflip cannot change during streaming */


naushir:
LGTM.

pelwell:
Checkpatch is complaining that you've not followed the kernel coding style w.r.t. brace placement. Other than that it looks good.

pelwell:
Thanks!

6by9:
I know this has been merged, but a minor comment that PM_RUNTIME is optional and is also using autosuspend. That means there is no guarantee that the sensor will be reset before each stream_on, and a module parameter can be changed dynamically. Once set into XTRIG mode there is no way to come back out of it unless PM_RUNTIME performs a shutdown on the sensor via the regulator framework.

The datasheet also implies that this isn't just a simple external trigger for changing frame rate. For fast trigger mode it states:
```Exposure time = (XTRIG low level pulse width) + 14.26us```
That totally changes the way the sensor would function, so I'm curious whether this works as intended when running with libcamera.

naushir:
> I know this has been merged, but a minor comment that PM_RUNTIME is optional and is also using autosuspend. That means there is no guarantee that the sensor will be reset before each stream_on, and a module parameter can be changed dynamically. Once set into XTRIG mode there is no way to come back out of it unless PM_RUNTIME performs a shutdown on the sensor via the regulator framework.

True.  We can fix this on top, but I'm not too concerned right now as I doubt users will be switching into and out of this mode at runtime just yet.

> 
> The datasheet also implies that this isn't just a simple external trigger for changing frame rate. For fast trigger mode it states: `Exposure time = (XTRIG low level pulse width) + 14.26us` That totally changes the way the sensor would function, so I'm curious whether this works as intended when running with libcamera.

Yes it does work as expected.  The pulse width sets the exposure time, and we fix the shutter speed in libcamera and the AGC will auto adjust gain as it needs.  All very cool!

6by9:
Ah, I hadn't noted the Signed-of-by, so had assumed this was a 3rd party contribution.

> Yes it does work as expected. The pulse width sets the exposure time, and we fix the shutter speed in libcamera and the AGC will auto adjust gain as it needs. All very cool!

That isn't how I expect it to work. You still have an active V4L2_CID_EXPOSURE control when in this mode, so I'd expect that to determine the exposure time, not some external trigger.

It is permitted to have a module parameter that is only checked at probe, so you could remove that control or make it read only at a nominal value if the module parameter is in external trigger mode.

I wonder if there a master mode where an IMX296 will generate a suitable trigger based on the configured exposure? I don't see anything obvious in the datasheet.

naushir:
Removing the control is a no-go as libcamera expects V4L2_CID_EXPOSURE to be available.  We can make it read-only with a nominal (but really meaningless) value.


> I wonder if there a master mode where an IMX296 will generate a suitable trigger based on the configured exposure? I don't see anything obvious in the datasheet.

There is a "stereo mode" operation that sort of does this, but it's not exactly what this trigger mode does, and does not seem to be fully deterministic in behavior.

Issue #5550, opened 2023-07-25
Waveshare 7.9" DSI not working android pi4


beletruz:
### Describe the bug

The latest 6.1.31 doesn't work with the latest raspberry pi os with this display https://www.waveshare.com/wiki/7.9inch_DSI_LCD#Overview by adding dtoverlay=vc4-kms-dsi-waveshare-panel,7_9_inch the display stays off and the led blinks but it does work (it lights up, it displays as it should) except for the touchscreen with android running on the pi (this android https://konstakang.com/devices/rpi4/LineageOS20 )
Using the "dtoverlay=vc4-kms-dsi-waveshare-panel,7_9_inch" and "goodix" for touch this is what it reports about the touchscreen

### Steps to reproduce the behaviour

Flash LieageOS20 -> flash 6.1 kernel -> the display shows the input but the touch doesn't work

### Device (s)

Raspberry Pi 4 Mod. B

### System

OS https://konstakang.com/devices/rpi4/LineageOS20
version d7f9c2b4ef7e4a8c0b04374a879ce89d7a948453 (clean) (release) (start _ x)
Linux localhost 6.1.31-v8-g84ed1b6094c9 #1 SMP PREEMPT Thu Jun 15 UTC 2023 aarch64 Toybox

### Logs

[ 3.486535] Goodix-TS 1-0014: supply AVDD28 not found, using dummy regulator
[ 3.493801] Goodix-TS 1-0014: supply VDDIO not found, using dummy regulator
[ 3.597518] Goodix-TS 1-0014: Error reading 1 bytes from 0x8140: -5
[ 3.629484] Goodix-TS 1-0014: Error reading 1 bytes from 0x8140: -5
[ 3.661185] Goodix-TS 1-0014: I2C communication failure: -5
[ 3.667235] Goodix-TS: probe of 1-0014 failed with error -5
[ 3.674567] i2c i2c-22: Added multiplexed i2c bus 0
[ 3.679906] Goodix-TS 10-0014: supply AVDD28 not found, using dummy regulator
[ 3.687263] Goodix-TS 10-0014: supply VDDIO not found, using dummy regulator
[ 3.695975] Goodix-TS 10-0014: ID 9271, version: 1070
[ 3.718404] input: Goodix Capacitive TouchScreen as /devices/platform/soc/fe205000.i2c/i2c-22/i2c-10/10-0014/input/input0

### Additional context

[...]

PR #5565, opened 2023-08-05
drm/panel: Fix default values for Waveshare 7.9 inch DSI touchscreen

ilarrain:
This fixes touchscreen calibration, axis swapping and inversion.

As referenced in https://github.com/raspberrypi/linux/issues/5550

### DIFF
--- a/arch/arm/boot/dts/overlays/vc4-kms-dsi-waveshare-panel-overlay.dts
+++ b/arch/arm/boot/dts/overlays/vc4-kms-dsi-waveshare-panel-overlay.dts
@@ -92,11 +92,11 @@
 		7_0_inchC = <&panel>, "compatible=waveshare,7.0inch-c-panel",
 				   <&touch>, "touchscreen-size-x:0=800",
 				   <&touch>, "touchscreen-size-y:0=480";
-		7_9_inch = <&panel>, "compatible=waveshare,7.9inch-panel",
-				   <&touch>, "touchscreen-size-x:0=400",
-				   <&touch>, "touchscreen-size-y:0=1280",
-				   <&touch>, "touchscreen-inverted-x?",
-				   <&touch>, "touchscreen-inverted-y?";
+                7_9_inch = <&panel>, "compatible=waveshare,7.9inch-panel",
+                                   <&touch>, "touchscreen-size-x:0=4096",
+                                   <&touch>, "touchscreen-size-y:0=4096",
+                                   <&touch>, "touchscreen-inverted-x?",
+                                   <&touch>, "touchscreen-swapped-x-y?";
 		8_0_inch = <&panel>, "compatible=waveshare,8.0inch-panel",
 				   <&touch>, "touchscreen-size-x:0=800",
 				   <&touch>, "touchscreen-size-y:0=1280",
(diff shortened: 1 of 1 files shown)


pelwell:
I've read the other thread, and I see that it fixes the original user's problem - thanks for that. Could you say something about changing the dimensions to 4096? Otherwise it looks ready to squash and merge.

6by9:
I could have sworn I'd commented on this PR previously, but it seems to have disappeared.

I'd like to confirm what inversions the Waveshare driver uses for the 7.9" panel.
From https://github.com/waveshare/Waveshare-DSI-LCD/blob/main/6.1.21/64/WS_xinchDSI_MAIN.sh#L10 I'm expecting it to be SCREEN_type = 5, which means at https://github.com/waveshare/Waveshare-DSI-LCD/blob/main/6.1.21/64/pi4/WS_xinchDSI.sh#L64 means I'm expecting it to be inverted-x and inverted-y as in the original.
It's possible that Waveshare have an error in their script, but https://github.com/waveshare/Waveshare-DSI-LCD/issues/4#issuecomment-1509490884 largely confirms that was the setting used for that display.

6by9:
Ah, the PR I'd commented on was https://github.com/ilarrain/linux/pull/1

ilarrain:
> I've read the other thread, and I see that it fixes the original user's problem - thanks for that. Could you say something about changing the dimensions to 4096? Otherwise it looks ready to squash and merge.

With the default config, only a small touch area (about 2x2 cm) was mapped to the whole screen, rendering it unusable.
I started experimenting with the value on each axis, and the touch area was getting bigger and bigger: 400 > 1080 >
2000 > 3000 > 4000 (it was close to perfect, but not quite) > 4200 (it was too big) > 4096 (it was perfect, ok both axis).
That's how I got to 4096. (2^12, 12 bit resolution on each axis).

I don't know if it applies to other panels, but at list on mine, @beletruz and @mbhangui, the patch fixed the issues. Maybe other panel sizes also use 4096 on both axis. I have no way to experiment, as this is the only panel I've got.

> I'd like to confirm what inversions the Waveshare driver uses for the 7.9" panel. From https://github.com/waveshare/Waveshare-DSI-LCD/blob/main/6.1.21/64/WS_xinchDSI_MAIN.sh#L10 I'm expecting it to be SCREEN_type = 5, which means at https://github.com/waveshare/Waveshare-DSI-LCD/blob/main/6.1.21/64/pi4/WS_xinchDSI.sh#L64 means I'm expecting it to be inverted-x and inverted-y as in the original. It's possible that Waveshare have an error in their script, but [waveshare/Waveshare-DSI-LCD#4 (comment)](https://github.com/waveshare/Waveshare-DSI-LCD/issues/4#issuecomment-1509490884) largely confirms that was the setting used for that display.

It is type 5 in the scripts for Wavershare installation, which corresponds to those values, you are correct. I have no way to know if it is an error by the manufacturer, or they are apllying an obscure logic in the overlays or kernel modules, as those are closed source. I just know what values work on my panel (and at least 2 other users).

> Ah, the PR I'd commented on was [ilarrain#1](https://github.com/ilarrain/linux/pull/1)

Sorry about that. I don't know much about git, and ended sending the pull request on the wrong repository. I thought about deleting it, but would have deleted the discussion too.

6by9:
I went shopping, and you appear to be right that the display manufacturer (or Waveshare as integrator) have done something very weird in their touch configuration, and plausibly emulated a Goodix touch controller rather than actually using one.

@pelwell The second patch is a fixup of the first. I think we can squash the two commits into one whilst merging, dropping the second commit in the process?
Do we care about the lack of SoB on a dtoverlay patch?

pelwell:
Yes, no, and done.

PR #5580, opened 2023-08-23
media: i2c: imx477: Add control of FSTROBE via module parameters

6by9:
Copy the firmware's handling of FSTROBE, but using module parameters instead of config.txt entries.


6by9:
Rebased onto 6.12 as #6945 6945

Issue #5582, opened 2023-08-25
kernel: bcm2835-codec bcm2835-codec: bcm2835_codec_buf_prepare data will not fit into plane (1536 < 195840)


qrp73:
### Describe the bug

There is a lot of errors in the log from bcm2835-codec

### Steps to reproduce the behaviour

Install Raspberry Pi OS aarch64 on Raspberry Pi 4B 4G. Boot it, open youtube and play some music, then open log with journalctl and scroll it down. Make sure there are a lot of errors from bcm2835-codec.

### Device (s)

Raspberry Pi 4 Mod. B

### System

$ cat /etc/rpi-issue
Raspberry Pi reference 2023-05-03
Generated using pi-gen, https://github.com/RPi-Distro/pi-gen, 7c750947a959fb626a70c09fd17c65815df192ac, stage4

$ vcgencmd version
Mar 17 2023 10:50:39 
Copyright (c) 2012 Broadcom
version 82f3750a65fadae9a38077e3c2e217ad158c8d54 (clean) (release) (start)

$ uname -a
Linux raspi 6.1.21-v8+ #1642 SMP PREEMPT Mon Apr  3 17:24:16 BST 2023 aarch64 GNU/Linux

### Logs

```
Aug 25 08:44:58 raspi kernel: bcm2835-codec bcm2835-codec: bcm2835_codec_buf_prepare data will not fit into plane (1536 < 622080)
Aug 25 08:44:58 raspi kernel: bcm2835-codec bcm2835-codec: bcm2835_codec_buf_prepare data will not fit into plane (1536 < 622080)
Aug 25 08:44:58 raspi kernel: bcm2835-codec bcm2835-codec: bcm2835_codec_buf_prepare data will not fit into plane (1536 < 622080)
Aug 25 08:44:58 raspi kernel: bcm2835-codec bcm2835-codec: bcm2835_codec_buf_prepare data will not fit into plane (1536 < 622080)
Aug 25 08:44:58 raspi kernel: bcm2835-codec bcm2835-codec: bcm2835_codec_buf_prepare data will not fit into plane (1536 < 622080)
Aug 25 08:44:58 raspi kernel: bcm2835-codec bcm2835-codec: bcm2835_codec_buf_prepare data will not fit into plane (1536 < 622080)
Aug 25 08:44:58 raspi kernel: bcm2835-codec bcm2835-codec: bcm2835_codec_buf_prepare data will not fit into plane (1536 < 622080)
[...]

PR #5588, opened 2023-08-30
TinyDRM ILI9486 overlay, and minor bcm2835-codec logging tweak

6by9:
https://forums.raspberrypi.com/viewtopic.php?t=355484 for the TinyDRM overlay.

bcm2835-codec change for #5582


arch/arm/boot/dts/overlays/piscreen-overlay.dts
```diff
@@ -98,5 +100,7 @@
 		fps =		<&piscreen>,"fps:0";
 		debug =		<&piscreen>,"debug:0";
 		xohms =		<&piscreen_ts>,"ti,x-plate-ohms;0";
+		drm =		<&piscreen>,"compatible=waveshare,rpi-lcd-35",
+				<&piscreen>,"reset-gpios:8=GPIO_ACTIVE_HIGH";
```
pelwell: Are you sure this works?:
```
$ dtmerge arch/arm/boot/dts/bcm2711-rpi-4-b.dtb base.dtb -
$ dtmerge base.dtb merged.dtb arch/arm/boot/dts/overlays/piscreen.dtbo drm
DTOVERLAY[error]: invalid override value 'GPIO_ACTIVE_HIGH' - ignored
```

6by9: You're right - my update script to push the change to the Pi obviously didn't for some reason (probably PEBKAC).

6by9:
Actually there's a bigger problem. In the non-drm case (compatible ```ilitek,ili9486```), both drivers get loaded
```
pi@pi:~ $ modinfo fb_ili9486
filename:       /lib/modules/6.1.47-v8+/kernel/drivers/staging/fbtft/fb_ili9486.ko.xz
license:        GPL
author:         Noralf Tronnes
description:    FB driver for the ILI9486 LCD Controller
alias:          platform:ili9486
alias:          spi:ili9486
alias:          platform:fb_ili9486
alias:          spi:fb_ili9486
srcversion:     A87A9FEB7E5C3553E7CDE5A
alias:          of:N*T*Cilitek,ili9486C*
alias:          of:N*T*Cilitek,ili9486
depends:        fbtft
staging:        Y
intree:         Y
name:           fb_ili9486
vermagic:       6.1.47-v8+ SMP preempt mod_unload modversions aarch64

pi@pi:~ $ modinfo ili9486
filename:       /lib/modules/6.1.47-v8+/kernel/drivers/gpu/drm/tiny/ili9486.ko.xz
license:        GPL
author:         Kamlesh Gurudasani <kamlesh.gurudasani@gmail.com>
description:    Ilitek ILI9486 DRM driver
srcversion:     6FBFCC0E15DDF43CA36470A
alias:          of:N*T*Cozzmaker,piscreenC*
alias:          of:N*T*Cozzmaker,piscreen
alias:          of:N*T*Cwaveshare,rpi-lcd-35C*
alias:          of:N*T*Cwaveshare,rpi-lcd-35
alias:          spi:ili9486
depends:        drm_kms_helper,drm,drm_mipi_dbi,backlight,drm_dma_helper
intree:         Y
name:           ili9486
vermagic:       6.1.47-v8+ SMP preempt mod_unload modversions aarch64
```
Presumably it's ```of:N*T*Cilitek,ili9486``` vs ```spi:ili9486``` causing the issue.

pelwell:
> Presumably it's of:N*T*Cilitek,ili9486 vs spi:ili9486 causing the issue.

Sorry, missed this pseudo-question - yes, that is probably the cauase.

Can we just disable the fbtft version?

6by9:
> Can we just disable the fbtft version?

Not really unless we change the overlay to only supporting the DRM one.

Both are mainline drivers, so probably worth reporting to mainline to see what they want to do. The fact that SPI has this secondary matching (and complaining if the SPI alias is missing) is really a bit ugly.

pelwell:
In the same way that the fbtft driver uses `spi:fb_ili9486`, we could patch the upstream DRM driver so it also uses a prefix for the SPI alias, e.g. `spi:drm_ili9486`.

6by9:
I'm not totally sure on how the spi_id table is used, except that the SPI core whinges if a compatible string device name isn't present in the spi_id table as well. So it complains on ili9486 as ```rpi-lcd-35``` and ```piscreen``` aren't present.
(It also complains that ```ili9486``` is missing for fb_ili9486).

Easiest fix is
```
@ -187,7 +187,8 @@ static const struct of_device_id ili9486_of_match[] = {
 MODULE_DEVICE_TABLE(of, ili9486_of_match);
 
 static const struct spi_device_id ili9486_id[] = {
-       { "ili9486", 0 },
+       { "rpi-lcd-35", 0 },
+       { "piscreen", 0 },
        { }
 };
 MODULE_DEVICE_TABLE(spi, ili9486_id);
```
thus silencing the SPI core code on the mismatch, and removing the clash.

Whilst at it I'll add the missing spi_device_id entries in ads7846 (the touchscreen driver) so at least with the DRM option the SPI core is happy.

pelwell:
This also works, if you prefer:
```
<&piscreen>,"reset-gpios:8=",<GPIO_ACTIVE_HIGH>;
```
but I'm happy to merge as is.

6by9:
> This also works, if you prefer:

Give me two mins to update to that - it's much nicer.

Issue #5214, opened 2022-10-18
spurious interrupts from a gpio interrupt line


marcarla:
### Describe the bug

At random, interrupts from a gpio line are received by the interrupt handler twice. The problem is present only with a 64 bit system, while 32 bit systems are clean

### Steps to reproduce the behaviour

I send a square wave with frequency 1 kHz and duty cycle 50% to a gpio line (gpio4), programmed to generate interrupts both at the rising and falling edge. Interrupts should occur with cadence 500 us with alternating 0/1 values. A small module handles the interrupts flow and checks its coherency.
Sometimes (about once in a thousand events) a few usec after an interrupt has been handled (typically 4 ... 10), the service routine is entered again.
To reproduce the problem, these is the module source code:
[irqtest.c.txt](https://github.com/raspberrypi/linux/files/9814996/irqtest.c.txt)
and the Makefile (remove the .txt from names):
[Makefile.txt](https://github.com/raspberrypi/linux/files/9815008/Makefile.txt)

This same test, run on the same machine with a 32 bit system, goes clean through millions of events.

### Device (s)

Raspberry Pi 3 Mod. B

### System

pi@raspberrypi:~/irqtest $ raspinfo
System Information
------------------

Raspberry Pi 3 Model B Rev 1.2
PRETTY_NAME="Debian GNU/Linux 11 (bullseye)"
NAME="Debian GNU/Linux"
VERSION_ID="11"
VERSION="11 (bullseye)"

Raspberry Pi reference 2022-09-22
Generated using pi-gen, https://github.com/RPi-Distro/pi-gen, 005a8c73b05a2cab394073150208bf4f069e861a, stage4

Linux raspberrypi 5.15.61-v8+ #1579 SMP PREEMPT Fri Aug 26 11:16:44 BST 2022 aarch64 GNU/Linux
Revision	: a02082
Serial		: 00000000fd306629
Model		: Raspberry Pi 3 Model B Rev 1.2
Throttled flag  : throttled=0x0
Camera          : supported=0 detected=0, libcamera interfaces=0

Videocore information
---------------------

Aug 26 2022 14:04:36 
Copyright (c) 2012 Broadcom
version 102f1e848393c2112206fadffaaf86db04e98326 (clean) (release) (start)

alloc failures:     0
compactions:        0
legacy block fails: 0

Filesystem information
----------------------
Filesystem     1K-blocks    Used Available Use% Mounted on
/dev/root       14978128 3586248  10726152  26% /
[...]

PR #5589, opened 2023-08-31
drivers: irqchip: irq-bcm2835: Concurrency fix

pelwell:
The commit shown in Fixes: aims to improve interrupt throughput by getting the handlers invoked on different CPU cores. It does so (*) by using an irq_ack hook to change the interrupt routing.

Unfortunately, the IRQ status bits must be cleared at source, which only happens once the interrupt handler has run - there is no easy way for one core to claim one of the IRQs before sending the remainder to the next core on the list, so waking another core immediately results in a race with a chance of both cores handling the same IRQ. It is probably for this reason that the routing change is deferred to irq_ack, but that doesn't guarantee no clashes - after irq_ack is called, control returns to bcm2836_chained_handler_irq which proceeds to check for other pending IRQs at a time when the next core is probably doing the same thing.

Since the whole point of the original commit is to distribute the IRQ handling, there is no reason to attempt to handle multiple IRQs in one interrupt callback, so the problem can be solved (or at least made much harder to reproduce) by changing a "while" into an "if", so that each invocation only handles one IRQ.

(*) I'm not convinced it's as effective as claimed since irq_ack is called _after_ the interrupt handler, but the author thought it made a difference.

See: https://github.com/raspberrypi/linux/issues/5214
     https://github.com/raspberrypi/linux/pull/1794

Fixes: fd4c9785bde86 ("ARM64: Round-Robin dispatch IRQs between CPUs.")

### DIFF
--- a/drivers/irqchip/irq-bcm2835.c
+++ b/drivers/irqchip/irq-bcm2835.c
@@ -343,7 +343,8 @@ static void bcm2836_chained_handle_irq(struct irq_desc *desc)
 {
 	u32 hwirq;
 
-	while ((hwirq = get_next_armctrl_hwirq()) != ~0)
+	hwirq = get_next_armctrl_hwirq();
+	if (hwirq != ~0)
 		generic_handle_domain_irq(intc.domain, hwirq);
 }
 


popcornmix:
Okay with me.

Reverting the round robin commit is another option (in an ideal world combined with a an arm64 fiq solution for dwc).

P33M:
BCM2836/7 never had any sort of automatic interrupt dispatching to not-busy cores, so round-robin servicing would be a useful compromise in the general sense. Fiq_fsm is just the canary in the coalmine since it needs <100us servicing latency or things start to break.

PR #5597, opened 2023-09-08
ASoC: hdmi-codec: Fix broken channel map reporting

popcornmix:
Commit 5fe680b23327 ("ASoC: hdmi-codec: fix channel info for compressed formats") accidentally changed hcp->chmap_idx from ca_id, the CEA channel allocation ID, to idx, the index to the table of channel mappings ordered by preference.

This resulted in wrong channel maps being reported to userspace, eg for 5.1 "FL,FR,LFE,FC" was reported instead of the expected "FL,FR,LFE,FC,RL,RR":

~ # speaker-test -c 6 -t sine
...
 0 - Front Left
 3 - Front Center
 1 - Front Right
 2 - LFE
 4 - Unknown
 5 - Unknown

~ # amixer cget iface=PCM,name='Playback Channel Map' | grep ': values'
  : values=3,4,8,7,0,0,0,0

Revert this incorrect change so that channel maps are properly reported again.

Fixes: 5fe680b23327 ("ASoC: hdmi-codec: fix channel info for compressed formats")
Cc: stable@vger.kernel.org

### DIFF
--- a/fs/f2fs/segment.c
+++ b/fs/f2fs/segment.c
@@ -661,9 +661,7 @@ init_thread:
 				"f2fs_flush-%u:%u", MAJOR(dev), MINOR(dev));
 	if (IS_ERR(fcc->f2fs_issue_flush)) {
 		err = PTR_ERR(fcc->f2fs_issue_flush);
-		kfree(fcc);
-		SM_I(sbi)->fcc_info = NULL;
-		return err;
+		fcc->f2fs_issue_flush = NULL;
 	}
 
 	return err;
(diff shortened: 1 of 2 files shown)


popcornmix:
From @HiassofT here: https://github.com/LibreELEC/LibreELEC.tv/pull/8122

6by9:
Commit hash of the fixes looks dubious.

On rpi-6.1.y it's 3f388718331b5ce2acd34730448db001759868aa, but that has "commit 4e0871333661d2ec0ed3dc00a945c2160eccae77 upstream" in the commit text for the backport, and that would appear to be the commit in mainline (https://github.com/torvalds/linux/commit/4e0871333661d2ec0ed3dc00a945c2160eccae77)

HiassofT:
There was a bit of a mix-up as both broonie and tiwai applied the patch to their trees - seems the other commit hash survived when it was finally merged into mainline
https://patchwork.kernel.org/project/alsa-devel/patch/20230624165232.5751-1-hias@horus.com/#25403695

HiassofT:
Ah, sorry, my fault, 4e0871333661 is indeed the right githash - the hash in the commit message is wrong (glad you noticed that before I sent it upstream)

HiassofT:
I've sent the fix upstream now, with corrected hashes in the commit message. Best if you pick that one instead
https://patchwork.kernel.org/project/alsa-devel/patch/20230909114633.3193-1-hias@horus.com/

pelwell:
The version of the patch sent upstream as been applied and pushed offline.

PR #5498, opened 2023-06-11
media: i2c: imx219: fix binning and rate_factor for 480p and 1232p

vvarma:
Fix for #5493 

Ran the following tests with the changes:

```
$ libcamera-hello --list-camera
Available cameras
-----------------
0 : imx219 [3280x2464] (/base/soc/i2c0mux/i2c@1/imx219@10)
    Modes: 'SRGGB10_CSI2P' : 640x480 [206.65 fps - (1000, 752)/1280x960 crop]
                             1640x1232 [41.85 fps - (0, 0)/3280x2464 crop]
                             1920x1080 [47.57 fps - (680, 692)/1920x1080 crop]
                             3280x2464 [21.19 fps - (0, 0)/3280x2464 crop]
           'SRGGB8' : 640x480 [206.65 fps - (1000, 752)/1280x960 crop]
                      1640x1232 [83.70 fps - (0, 0)/3280x2464 crop]
                      1920x1080 [47.57 fps - (680, 692)/1920x1080 crop]
                      3280x2464 [21.19 fps - (0, 0)/3280x2464 crop]
```
Note the higher fps for RAW8 1232p mode

Verified no frame corruption with 480p and 1232p in RAW8 and RAW10 formats using variations of:
```
$ libcamera-vid --level 4.2 --framerate 200 --buffer-count 10 --mode 640:480:10:P --save-pts t640x480x10x200.pts -o t640x480x10x200.h264 -t 3000 --denoise cdn_off -n
```

Issue #5493, opened 2023-06-08
[imx219] 10bit 640x480 video at high fps (200+) 


vvarma:
### Describe the bug

In commit [9ffaf55e](https://github.com/raspberrypi/linux/commit/9ffaf55e126cf4aad3d46198a2fe7b54cc13fbec), the change switches between analog 2x2 binning and the 2x2 binning based on whether its an 8-bit codec or 10-bit one. But in commit [7a41d0f](https://github.com/raspberrypi/linux/pull/5439/commits/7a41d0f88d60d9520f3c11e93a7875a30f097115) the rate factor is set based on mode.

So in this case of a 10bit 640x480 video capture, the normal 2x2 binning is used but with rate_factor set to 2.

I believe this causes the video at high FPS to be "corrupted"

https://github.com/raspberrypi/linux/assets/2176033/b5199fc5-0315-4e9d-a0f2-a3f404cb8146

as compared to the same at 8bit codec or 10-bit codec with lower FPS

https://github.com/raspberrypi/linux/assets/2176033/ab178929-78b7-4039-b0ed-5c2498c0ea72

https://github.com/raspberrypi/linux/assets/2176033/02098d57-d4c3-405d-ae7e-415c4fe5a7cd

### Steps to reproduce the behaviour

```
libcamera-vid --level 4.2 --framerate 200 --buffer-count 10 --mode 640:480:10:P --save-pts t640x480x10x200.pts -o t640x480x10x200.h264 -t 3000 --denoise cdn_off -n
```
vs
```
libcamera-vid --level 4.2 --framerate 200 --buffer-count 10 --mode 640:480:8:P --save-pts t640x480x8x200.pts -o t640x480x8x200.h264 -t 3000 --denoise cdn_off -n
```
or 
```
libcamera-vid --level 4.2 --framerate 100 --buffer-count 10 --mode 640:480:10:P --save-pts t640x480x10x100.pts -o t640x480x10x100.h264 -t 3000 --denoise cdn_off -n
```

### Device (s)

Raspberry Pi 4 Mod. B

### System

> cat /etc/rpi-issue
```
Raspberry Pi reference 2023-05-03
Generated using pi-gen, https://github.com/RPi-Distro/pi-gen, 7c750947a959fb626a70c09fd17c65815df192ac, stage2
```
>  vcgencmd version
```
Mar 17 2023 10:50:39
Copyright (c) 2012 Broadcom
[...]

PR #5607, opened 2023-09-19
[media][imx219] selection API

vvarma:
This change primarily adds the set_selection APIs (crop and compose). While the hardware supports other binning ratios, only the `BINNING_NONE`, `BINNING_2x2` and `BINNING_2x2_ANALOG` are supported. The rule to use the "special binning" is based on the frame size and whether it is an 8-bit format. At the same time with this change, the driver allows binning on x or y dimensions or both based on the crop and compose fields. 

The biggest change in terms of the implementation is to reduce the scope of `imx219->mode`. Instead, the `crop` and `compose` are stored in the `imx219` struct. All the usages of `imx219->mode` have been transitioned to use these fields instead except in the `imx219_init_controls` to set defaults. The `supported_modes` continues to be used for `S_FMT` operations.

One change in the API is that the v_blank value is not set to the default of the mode selected on `S_FMT` but is based on whatever the user set last. It should be possible to avoid this if not maintaining the `v_blanking` in the `S_FMT` op is a required feature.

### Note
The commits are separated out in order to help review the code. I plan to squash all of them into 1 or 2 commits if and when it's close to getting approval unless the reviewers prefer I squash it right away.

### Links
- https://github.com/raspberrypi/linux/pull/4935

### DIFF
--- a/drivers/media/i2c/imx219.c
+++ b/drivers/media/i2c/imx219.c
@@ -147,10 +147,12 @@ enum pad_types {
 	NUM_PADS
 };
 
+static const u32 binning_ratio[] = { 1, 2 };
+
 enum binning_mode {
 	BINNING_NONE,
 	BINNING_DIGITAL_2x2,
-	BINNING_ANALOG_2x2,
+	BINNING_ANALOG_2x2
 };
 
 enum binning_bit_depths {
@@ -879,6 +881,50 @@ static int imx219_get_pad_format(struct v4l2_subdev *sd,
 	return ret;
 }
 
+static int imx219_refresh_ctrls(struct imx219 *imx219,
+				struct v4l2_rect *prev_compose)
+{
+	int exposure_max, exposure_def, hblank, pixel_rate, rate_factor;
+	u32 prev_hts = prev_compose->width + imx219->hblank->val;
+	u32 prev_vts = prev_compose->height + imx219->vblank->val;
+
+	rate_factor = imx219_get_rate_factor(imx219);
+	if (rate_factor < 0)
+		return rate_factor;
+	/* Update limits and set FPS to default */
+	__v4l2_ctrl_modify_range(imx219->vblank, IMX219_VBLANK_MIN,
+				 IMX219_VTS_MAX - imx219->compose.height, 1,
+				 prev_vts - imx219->compose.height);
+	__v4l2_ctrl_s_ctrl(imx219->vblank, prev_vts - imx219->compose.height);
+	/* Update max exposure while meeting expected vblanking */
+	exposure_max = prev_vts - 4;
+	exposure_def = (exposure_max < IMX219_EXPOSURE_DEFAULT) ?
+			       exposure_max :
+			       IMX219_EXPOSURE_DEFAULT;
+	__v4l2_ctrl_modify_range(imx219->exposure, imx219->exposure->minimum,
+				 exposure_max, imx219->exposure->step,
+				 exposure_def);
+	/*
+   * Retain PPL setting from previous mode so that the
+   * line time does not change on a mode change.
+   * Limits have to be recomputed as the controls define
+   * the blanking only, so PPL values need to have the
+   * mode width subtracted.
+   */
+	hblank = prev_hts - imx219->compose.width;
+	__v4l2_ctrl_modify_range(imx219->hblank,
+				 IMX219_PPL_MIN - imx219->compose.width,
+				 IMX219_PPL_MAX - imx219->compose.width, 1,
+				 IMX219_PPL_MIN - imx219->compose.width);
+	__v4l2_ctrl_s_ctrl(imx219->hblank, hblank);
+
+	/* Scale the pixel rate based on the mode specific factor */
+	pixel_rate = IMX219_PIXEL_RATE * rate_factor;
+	__v4l2_ctrl_modify_range(imx219->pixel_rate, pixel_rate, pixel_rate, 1,
+				 pixel_rate);
+	return 0;
+}
+
 static int imx219_set_pad_format(struct v4l2_subdev *sd,
 				 struct v4l2_subdev_state *sd_state,
 				 struct v4l2_subdev_format *fmt)
@@ -886,7 +932,7 @@ static int imx219_set_pad_format(struct v4l2_subdev *sd,
 	struct imx219 *imx219 = to_imx219(sd);
 	const struct imx219_mode *mode;
 	struct v4l2_mbus_framefmt *framefmt;
-	int exposure_max, exposure_def, hblank, pixel_rate, rate_factor;
+	int ret = 0;
 	unsigned int i;
 
 	if (fmt->pad >= NUM_PADS)
@@ -919,10 +965,7 @@ static int imx219_set_pad_format(struct v4l2_subdev *sd,
 			*framefmt = fmt->format;
 		} else if (imx219->mode != mode ||
 			   imx219->fmt.code != fmt->format.code) {
-			u32 prev_hts =
-				imx219->compose.width + imx219->hblank->val;
-			u32 prev_vts =
-				imx219->compose.height + imx219->vblank->val;
+			struct v4l2_rect curr_compose = imx219->compose;
 
 			imx219->mode = mode;
 			imx219_update_crop_from_mode(mode, &imx219->crop);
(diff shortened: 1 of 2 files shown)


6by9:
Sorry I'm going to have to direct this one upstream to the linux-media mailing list.

There is already at least one significant patch set on there (https://patchwork.linuxtv.org/project/linux-media/list/?series=11202), and accepting something like this downstream is going to lead to major merge conflicts when that hits (all have a Reviewed-by tag, so should be in the next merge cycle).

The other issue is that adding support for selection API on top of a mode based sensor driver was discussed within the linux-media community last September (https://www.spinics.net/lists/linux-media/msg219442.html). A basic route forward was proposed, but no implementation has been forthcoming since.

We have defined modes. Those can not be dropped as it would be considered a regression in stopping existing applications working.
It is undefined how a set_pad_format call should interact with a controllable selection crop, so if you set the crop to say 2x2 binned 1292x972, and then set the format to 2592x1944 you can't say what the driver should do.

One approach would be that setting the format is allowed to update the selections, thereby allowing you to retain enumeration and selection of the current modes, but also amend them afterwards via the selection API. Implementing something like that will require buy in from the linux-media maintainers though. I haven't followed through the complete details of your patch set to see if that is what you've implemented.

vvarma:
Thanks for the quick response @6by9, replying inline -
> Sorry I'm going to have to direct this one upstream to the linux-media mailing list.

> There is already at least one significant patch set on there (https://patchwork.linuxtv.org/project/linux-media/list/?series=11202), and accepting something like this downstream is going to lead to major merge conflicts when that hits (all have a Reviewed-by tag, so should be in the next merge cycle).

Ah too bad! Looking at the changes in the patch set you shared, I think there shouldn't be too many major conflicts. I'll pull in those commits to my tree and try rebasing my changes on top of the same. 

> The other issue is that adding support for selection API on top of a mode based sensor driver was discussed within the linux-media community last September (https://www.spinics.net/lists/linux-media/msg219442.html). A basic route forward was proposed, but no implementation has been forthcoming since.

"Another alternative is a new API to cover freely configurable and mode-based
drivers. Backwards compatible, and must not break freely configurable drivers
that now work well. This seems to be the way forward, Jacopo will work towards
this."
Is there a document or some specs to refer to what the new APIs will look like? Personally (and with my limited experience), I found the current selection API to be quite capable when dealing with binning except when dealing with the special binning mode of the imx219 which has quite a few cases to handle. 

>We have defined modes. Those can not be dropped as it would be considered a regression in stopping existing applications working.
>It is undefined how a set_pad_format call should interact with a controllable selection crop, so if you set the crop to say 2x2 binned 1292x972, and then set the format to 2592x1944 you can't say what the driver should do.

>One approach would be that setting the format is allowed to update the selections, thereby allowing you to retain enumeration and selection of the current modes, but also amend them afterwards via the selection API. Implementing something like that will require buy in from the linux-media maintainers though. I haven't followed through the complete details of your patch set to see if that is what you've implemented.

That is indeed how I have implemented it. The enumerated formats are retained and the S_FMT allows only one of the enumerated modes to be set, maintaining backwards compatibility with existing users. This means, as you mentioned, that the set_pad_fmt is allowed to override the crop and compose targets. The expectation is that users of the selection API first set the format and then change the selection targets as required.

6by9:
> > There is already at least one significant patch set on there (https://patchwork.linuxtv.org/project/linux-media/list/?series=11202), and accepting something like this downstream is going to lead to major merge conflicts when that hits (all have a Reviewed-by tag, so should be in the next merge cycle).
> 
> Ah too bad! Looking at the changes in the patch set you shared, I think there shouldn't be too many major conflicts. I'll pull in those commits to my tree and try rebasing my changes on top of the same.

The rpi-6.5.y branch should be working if you want to be closer to linux-media. There are a couple of fixups required for imx219 on rpi-6.6.y at present which should be sorted in the next few days.
You may be best taking rpi-6.6.y whilst it hasn't got the fixups, as Laurent's patches are more likely to apply cleanly to that.

> > The other issue is that adding support for selection API on top of a mode based sensor driver was discussed within the linux-media community last September (https://www.spinics.net/lists/linux-media/msg219442.html). A basic route forward was proposed, but no implementation has been forthcoming since.
> 
> "Another alternative is a new API to cover freely configurable and mode-based drivers. Backwards compatible, and must not break freely configurable drivers that now work well. This seems to be the way forward, Jacopo will work towards this." Is there a document or some specs to refer to what the new APIs will look like?

@jmondi You were the one who raised the subject of sensor mode configuration and selection API in Dublin last year. I'm guessing it's not gone anywhere as yet.

> Personally (and with my limited experience), I found the current selection API to be quite capable when dealing with binning except when dealing with the special binning mode of the imx219 which has quite a few cases to handle.

It's these holes that mean that the current APIs don't fully work, and why the CCS becomes so much more complex with multiple subdevs (and AFAIK nothing uses it). There's also no way to select between binning and skipping.

vvarma:
While testing this patch I also came across an issue that I do not yet know how to handle. I added it in this thread since it's related to the PR here, please let me know if I should move this elsewhere.
the v_blank ctrl, when set to its minimum with the special binning mode does not always work. To be more descriptive, I wrote a small test script that captures the issue
```
#!/bin/bash

function set_selection(){
  height=$1;
  printf "set height to %d " ${height}
  v4l2-ctl --set-selection target=crop,width=1280,height="$((height*2))"
  v4l2-ctl --set-selection target=compose,width=640,height="${height}"
}

function find_vts_limit(){
  vblank=32
  while true ; do
    v4l2-ctl -c vertical_blanking=${vblank}
    if timeout 1s v4l2-ctl --stream-mmap=2 --stream-count 10; then
      printf "Vertical blanking limit: %d\n" ${vblank}
      break
    fi
    vblank=$((vblank+4))
  done
}

# set initial format
v4l2-ctl -v pixelformat=RGGB,height=480,width=640

# reduce height using the selection api and find the vblank level at which the stream doesnt get stuck
height=480;
while [ $height -gt 0 ]; do
  set_selection $height
  find_vts_limit
  height=$((height-4))
done
```
Running this (results attached), between the compose height 480 and 256, the vblank of 32 works well. From there till 128, the minimum vblank is not working as intended but increases to 156. From 128 then all the way to the height of 4 the minimum vblank reduces till it reaches 32 again. 

In the logs, I can see 
```
[ 3145.603602] unicam 3f801000.csi: ISR: ISTA: 0x1, STA: 0xD001, sequence 264, lines done 0
[ 3145.603621] unicam 3f801000.csi: ISR: [0] Dropping frame, buffer not available at FS
[ 3145.603628] unicam 3f801000.csi: Scheduling dummy buffer for node 0
```
In the unicam driver, in the ISR there is a comment concerning this case 
```			 * If cur_frm == next_frm, it means we have not had
			 * a chance to swap buffers, likely due to having
			 * multiple interrupts occurring simultaneously (like FE
			 * + FS + LS). In this case, we cannot signal the buffer
			 * as complete, as the HW will reuse that buffer.
```

However, I do not understand these results and how they are affected by vblanking and specifically how there seems to be a point of inflexion at the height of 128.

[limit_vts.txt](https://github.com/raspberrypi/linux/files/12662593/limit_vts.txt)

6by9:
Welcome to the wonderful world of image sensors. They don't always behave in the logical way one might expect, hence your issue #5493 and subsequent PR #5498.

Sensor vendors generally provide the tables of validated mode registers for specific modes, and that is what they support. There are often niggles when you try and generalise them.

vvarma:
The above issues were due to a bug in the code. Closing this PR for now as discussed.

vvarma:
> > > There is already at least one significant patch set on there (https://patchwork.linuxtv.org/project/linux-media/list/?series=11202), and accepting something like this downstream is going to lead to major merge conflicts when that hits (all have a Reviewed-by tag, so should be in the next merge cycle).
> > 
> > 
> > Ah too bad! Looking at the changes in the patch set you shared, I think there shouldn't be too many major conflicts. I'll pull in those commits to my tree and try rebasing my changes on top of the same.
> 
> The rpi-6.5.y branch should be working if you want to be closer to linux-media. There are a couple of fixups required for imx219 on rpi-6.6.y at present which should be sorted in the next few days. You may be best taking rpi-6.6.y whilst it hasn't got the fixups, as Laurent's patches are more likely to apply cleanly to that.

@6by9 I believe the patchset linked above has been merged into the rpi-6.7.7 branch, hence, I plan to redo the above PR. Based on the newer changes, I think it should be a fairly small PR. Please let me know if you have any suggestions for me going into the change set and whether I should be on the lookout for any other conflicting changes that you may be aware of.

PR #5613, opened 2023-09-25
Camera fixups for 6.6

6by9:
Brings in the imx219 and video-mux patches that didn't rebase easily, and fixes up unicam patch.

### DIFF
--- a/drivers/media/i2c/imx219.c
+++ b/drivers/media/i2c/imx219.c
@@ -100,8 +100,8 @@
 /* Binning  Mode */
 #define IMX219_REG_BINNING_MODE		0x0174
 #define IMX219_BINNING_NONE		0x0000
-#define IMX219_BINNING_2X2		0x0101
-#define IMX219_BINNING_2X2_ANALOG	0x0303
+#define IMX219_BINNING_2X2_NORMAL	0x0101
+#define IMX219_BINNING_2X2_SPECIAL	0x0303
 
 /* Test Pattern Control */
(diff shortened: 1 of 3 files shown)


drivers/media/i2c/imx219.c
```diff
@@ -756,48 +796,57 @@ static int imx219_set_pad_format(struct v4l2_subdev *sd,
...
-
-	imx219_update_pad_format(imx219, mode, &fmt->format, fmt->format.code);
-	format = v4l2_subdev_get_pad_format(sd, sd_state, 0);
-
-	if (imx219->mode == mode && format->code == fmt->format.code)
-		return 0;
-
-	if (fmt->which == V4L2_SUBDEV_FORMAT_ACTIVE) {
-		u32 prev_hts = imx219->mode->width + imx219->hblank->val;
+	if (fmt->pad >= NUM_PADS)
```
pelwell, line 888: I don't know if this is an old patch, but why handle the error case early? Why not "if IMAGE, else if METADATA. else ERROR"? At least that way it's obvious where METADATA is handled.

6by9: Old patch - https://github.com/raspberrypi/linux/commit/e3376fb94fda798d2a322e9c70789286132a1a9f

Upstream don't like two independent pads for image and metadata, and are working on adding support for multiplexed pads (6.7 if we're incredibly lucky, but otherwise 6.8). Little point in polishing old patches that are going to be ditched RSN.

PR #5617, opened 2023-09-28
Fan/thermal fix

pelwell:
Since 6.3, the thermal/cooling fan support has been broken due to an uninitialised variable. Fix it, and remove a spurious duplicate function call.


pelwell:
Fix confirmed in https://forums.raspberrypi.com/viewtopic.php?t=356881.

PR #5618, opened 2023-09-28
Add support for Pi 5

popcornmix:
Add support for Pi 5.

This supports the new hardware in BCM2712 and RP1.

There are two 64-bit defconfig options.

bcm2711_defconfig will support all 64-bit Pi devices and uses a pagesize of 4K
bcm2712_defconfig switches to a 64K pagesize, and only works on Pi 5.


6by9:
@popcornmix We want to drop the top 3 patches marked DNI (Do Not Integrate) that mess with the workflow files.

pelwell:
And the first PCIe reset revert can be replaced with upstream commit 8eb8c2735306526929141052c82bd118a7a2119b.

pelwell:
On the subject of workflow files, as long as we continue to guarantee that the only difference between bcm2712_defconfig and bcm2711_defconfig is the page size, we can omit a Pi 5-specific test build.

6by9:
> On the subject of workflow files, as long as we continue to guarantee that the only difference between bcm2712_defconfig and bcm2711_defconfig is the page size, we can omit a Pi 5-specific test build.

It probably has some gain though in order to be able to use rpi-update to grab the appropriate build artifacts though (I'm hoping rpi-update looks at that bit of the extension too - I've not checked).

pelwell:
I've bumped `overlaycheck` in the utils repo, so subsequent pushes _should_ pass.

pelwell:
`rpi-update` will need to grow a `WANT_PI5` flag.

popcornmix:
Updated (@pelwell please check pcie-bcmstb.c - it is now not the same as initial PR).

popcornmix:
rpi-firmware will need the additional kernel2712.img built (and rpi-update will need checking that it copies it).
I'll also check what rpi-update from a PR does.

But I think we'll need the additional build artifact - otherwise rpi-update will bump you from a 16K pagesize to a 4K pagesize.

pelwell:
> Updated (@pelwell please check pcie-bcmstb.c - it is now not the same as initial PR).

pcie-brcmstb.c now matches rpi-6.6.y in that respect. I'm happy this part needed changing to sit on top of that upstream change, but perhaps this is the wrong outcome. The question is whether PERST# also needs to be asserted on BCM2712 - the safe thing would be to add that case to the condition, since that matches the old behaviour, and then drop it if we decide it isn't necessary. I'll add a patch to that effect.

> But I think we'll need the additional build artifact - otherwise rpi-update will bump you from a 16K pagesize to a 4K pagesize.

OK - I'll add the build type to the workflow.

pelwell:
> I'll add a patch to that effect.

After discussion with Tim this is currently unnecessary because the bootloader does leave it reset, so I won't add such a patch at this time.

popcornmix:
> After discussion with Tim this is currently unnecessary because the bootloader does leave it reset, so I won't add such a patch at this time.

So current PR is good (wrt pcie-bcmstb.c)?

pelwell:
Yes, I think so.

I've bumped the workflows. There is still some issue with one of the overlays that I'll look at next.

pelwell:
@6by9 in vc4-kms-dsi-7inch-overlay.dts it references ts_i2c_frag, but no such label exists. I think the node you are aiming for is currently labelled frag13, but that doesn't fit with the nice user-friendly labels the rest of that file uses. However, changing edt-ft5406.dtsi could break edt-ft5406-overlay.dts

Would you like me to:
a. add ts_i2c_frag as a second label into edt-ft5406.dti,
b. add ts_i2c_frag as an alias but inside vc4-kms-v3d-pi5-overlay.dts,
c. change vc4-kms-v3d-pi5-overlay.dts to use frag13, or
d. change everything to use ts_i2c_frag instead?

6by9:
> in vc4-kms-dsi-7inch-overlay.dts it references ts_i2c_frag, but no such label exists. 

See commit d6e6b9afa8 on the private branch. It just missed the cutoff yesterday, and looks like popcornmix hasn't included it in this PR or merged it into "dts: 2712: Update for device tree"

6by9:
DRM/KMS unit tests failing appears to be a build failure due to part of raspberrypi-gpiomem.c
```
2023-09-28T14:52:58.7457288Z ../drivers/char/raspberrypi-gpiomem.c:104:37: error: implicit declaration of function ‘phys_mem_access_prot’; did you mean ‘phys_mem_access_prot_allowed’? [-Werror=implicit-function-declaration]
2023-09-28T14:52:58.7457878Z [14:49:36] Configuring KUnit Kernel ...
2023-09-28T14:52:58.7500841Z   104 |                 vma->vm_page_prot = phys_mem_access_prot(file, base, len,
2023-09-28T14:52:58.7501241Z Generating .config ...
2023-09-28T14:52:58.7501769Z       |                                     ^~~~~~~~~~~~~~~~~~~~
2023-09-28T14:52:58.7502250Z       |                                     phys_mem_access_prot_allowed
2023-09-28T14:52:58.7503042Z ../drivers/char/raspberrypi-gpiomem.c:104:37: error: incompatible types when assigning to type ‘pgprot_t’ from type ‘int’
```
I'm not totally clear why it's even building gpiomem.

KUnit Tests/VC4 are failing as the test vc4/tests/vc4_test_lbm_size.c needs updating for the lbm_size having moved.
That should be easy to fix.

pelwell:
So, option d then. Cherry-picked.

pelwell:
Upstream build fixup pushed.

6by9:
VC4 unit tests fixed with https://github.com/6by9/linux/commit/950224c8c28ea11b2642f98f2b59625fe2afc3ca (I'm not sure I can do a push to popcornmix's branch.

6by9:
drivers/char/Kconfig has ```RASPBERRYPI_GPIOMEM``` default to y. That would explain the build breakage for the generic unit tests, but is that what we want?
I guess if we default to not building it then the defconfig needs updating. As it's providing userspace access it could be a module, and only on 2711/2712 rather than all kernels.

The old BCM2835_DEVGPIOMEM option for bcm2835-gpiomem was a module.

pelwell:
Defaulting the gpiomem to a module makes sense - it's an anomaly that it isn't already.

6by9:
Commits https://github.com/6by9/linux/commit/430bb5b1f3253abab25bbf059d7b7c0b7e781ed4 and https://github.com/6by9/linux/commit/bde073f4c395ff601c41f3651fc8709c4a134b04 sort the generic unit test CI failure.

430bb5b1f wants to go before "drivers: char: delete bcm2835-gpiomem", and bde073f4c3 fixes up "drivers: char: delete bcm2835-gpiomem".

popcornmix:
I believe I've squashed in commits that have been pushed/mentioned. Anything missing?

pelwell:
Nope - that looks good to me if the checks are passing.

PR #5627, opened 2023-10-02
media/rpivid: Make iommu work on rpivid

jc-kynesim:
In order to use iommu on hevc add appropriate iommu to dts description and set dma mask_and_coherent in probe.
I am assured dma_set_mask_and_coherent is benign on Pi4 (which has no iommu) and it seems to be so in practice.
Also adds a bit of debug to make internal buffer allocation failure easier to spot in future.

### DIFF
--- a/drivers/staging/media/rpivid/rpivid_h265.c
+++ b/drivers/staging/media/rpivid/rpivid_h265.c
@@ -2495,11 +2495,19 @@ static int rpivid_h265_start(struct rpivid_ctx *ctx)
 	for (i = 0; i != ARRAY_SIZE(ctx->pu_bufs); ++i) {
 		// Don't actually need a kernel mapping here
 		if (gptr_alloc(dev, ctx->pu_bufs + i, pu_alloc,
-			       DMA_ATTR_NO_KERNEL_MAPPING))
+			       DMA_ATTR_NO_KERNEL_MAPPING)) {
+			v4l2_err(&dev->v4l2_dev,
+				 "Failed to alloc %#zx PU%d buffer\n",
+				 pu_alloc, i);
 			goto fail;
+		}
 		if (gptr_alloc(dev, ctx->coeff_bufs + i, coeff_alloc,
-			       DMA_ATTR_NO_KERNEL_MAPPING))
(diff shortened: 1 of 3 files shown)


drivers/staging/media/rpivid/rpivid_h265.c
```diff
@@ -2495,11 +2495,19 @@ static int rpivid_h265_start(struct rpivid_ctx *ctx)
 	for (i = 0; i != ARRAY_SIZE(ctx->pu_bufs); ++i) {
 		// Don't actually need a kernel mapping here
 		if (gptr_alloc(dev, ctx->pu_bufs + i, pu_alloc,
-			       DMA_ATTR_NO_KERNEL_MAPPING))
+			       DMA_ATTR_NO_KERNEL_MAPPING)) {
+			v4l2_err(&dev->v4l2_dev,
+				 "Failed to alloc %#zx PU%d buffer\n",
+				 pu_alloc, i);
```
6by9, line 2501: Do we need these error messages? Most of the kernel alloc functions log errors themselves.

6by9 (commented):
I suspect Dom will want the dt change to be independent of the driver one, otherwise we get into a mess in trying to squash things down.

drivers/staging/media/rpivid/rpivid_h265.c
jc-kynesim: Well these ones didn't!

jc-kynesim:
dt change split from rpivid. Why do we squash the dt files?

6by9:
> dt change split from rpivid. Why do we squash the dt files?

So that the list of downstream patches that have to be forward ported between kernel versions doesn't just grow and grow.

eg https://github.com/raspberrypi/linux/commits/rpi-6.6.y/arch/arm/boot/dts
1 mega-squash, and so far 2 fixups.

drivers/staging/media/rpivid/rpivid_h265.c
6by9: Fair enough.

Checkpatch will actually warn over kzalloc and similar calls having logging messages on failure.

PR #5628, opened 2023-10-02
Replace hdmi-codec channel map fix with accepted verison

HiassofT:
v2 of the patch, with a minor change (report "unknown" channel mappings in case of non-pcm audio) has been accepted for 6.7:

https://lore.kernel.org/alsa-devel/169625985045.65718.7789607105387288563.b4-ty@kernel.org/

Replace the v1 version in the RPi tree so we are up to date


popcornmix:
Thanks!

PR #5629, opened 2023-10-02
Add DT aliases to map to DRM card names

6by9:
Using this so that the card defined as DSI1 in DT will always be assigned the DRM display name DSI-1, and likewise for DSI-2. 

It can also work for HDMI, DPI, etc, but there's no need as the two HDMI's belong to the same DRM card, and there is only one DPI display.

There is a potential for confusion as alias ```SPI1``` (in capitals) will be the alias the DRM device, whilst ```spi1``` (lower case) is the SPI bus alias. At least capitalisation is preserved in DT so they don't actually collide, but I also don't see an easy way around the potential confusion.

### DIFF
--- a/drivers/gpu/drm/drm_connector.c
+++ b/drivers/gpu/drm/drm_connector.c
@@ -81,6 +81,7 @@ struct drm_conn_prop_enum_list {
 	int type;
 	const char *name;
 	struct ida ida;
+	int first_dyn_num;
 };
 
 /*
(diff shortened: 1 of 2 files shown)


pelwell:
> I also don't see an easy way around the potential confusion.

Have you considered a prefix, e.g. `drm_`?

6by9:
> Have you considered a prefix, e.g. `drm_`?

Doing so means having to do an sprintf (of some form) for drm_connector_enum_list.name alongside the prefix.
Is it worth messing around with it, or just using it as is? I didn't feel it was worth it.

pelwell:
You've expressed a reservation, and I agree that the names as they stand are not ideal. Can we even guarantee that there will never be a clash?

6by9:
> Can we even guarantee that there will never be a clash?

Only by getting it upstreamed.

I don't know what other uses of spi aliases there would be as there is already the mechanism for using that lower case.

There is now DRM_MODE_CONNECTOR_USB which would use the alias USB1, USB2, etc for USB connected displays, but those wouldn't generally be described in DT.

DT generally uses lower case. I didn't fancy having to convert the strings to lower case, but perhaps I should.

pelwell:
> DT generally uses lower case. I didn't fancy having to convert the strings to lower case, but perhaps I should.

I think you'll regret it if you don't. A simple open-coded strcat/strcpy with a `tolower()` in the middle shouldn't be more than a few lines.

6by9:
Updated to mangle the names to drm_<lower_case_type>.

drivers/gpu/drm/drm_connector.c
```diff
@@ -110,12 +111,40 @@ static struct drm_conn_prop_enum_list drm_connector_enum_list[] = {
...
+	int i = 0;
+
+	strcpy(node_name, DT_DRM_NODE_PREFIX);
+
+	do {
+		node_name[i + strlen(DT_DRM_NODE_PREFIX)] =
+				tolower(drm_connector_enum_list[type].name[i]);
+
+	} while (drm_connector_enum_list[type].name[i++] &&
+		 i < MAX_DT_NODE_NAME_LEN);
```
pelwell: `length` is the limit. And if you're accepting the possibility of string overflow you'd better ensure the result is NUL-terminated.

pelwell (commented):
So much better - just the one inline comment.

drivers/gpu/drm/drm_connector.c
6by9: Sorted.

6by9:
Fixed the length error, and dropped my erroneous debug pr_err.

pelwell:
Running an overlaycheck locally has kindly pointed out that `aliases property name must include only lowercase and '-'`. I have no idea why - it's not in the DT specs I've read. Can you change the separator to `-`?

6by9:
> Running an overlaycheck locally has kindly pointed out that `aliases property name must include only lowercase and '-'`. I have no idea why - it's not in the DT specs I've read. Can you change the separator to `-`?

Presumably numerics are allowed as well, otherwise we're totally stuffed.
Updating.....

PR #5630, opened 2023-10-04
cfe improvements/fixes

naushir:


tomba:
@naushir looks good to me

drivers/media/platform/raspberrypi/rp1_cfe/cfe.c
```diff
@@ -997,6 +997,14 @@ static int cfe_start_streaming(struct vb2_queue *vq, unsigned int count)
...
 	}
 
+	/* When using the Frontend, we must enable the FE_CONFIG node. */
+	if (is_fe_enabled(cfe) &&
+	    !check_state(cfe, NODE_ENABLED, cfe->node[FE_CONFIG].id)) {
+		cfe_err("FE enabled, but FE_CONFIG node is not\n");
+		ret = -EINVAL;
+		goto err_streaming;
+	}
+
```
tomba, line 1007: @naushir  Ah, noticed only now. The above code block is on the wrong side of the pm_runtime_resume_and_get. It will miss a runtime put in error case.

PR #5632, opened 2023-10-04
Cfe improvements

tomba:
More CFE improvements and fixes. These may need more testing, as they start to change things that affect the userspace APIs and behavior.

These require https://github.com/raspberrypi/libcamera/pull/82 to work (or more specifically, the last 3 patches require).

### DIFF
--- a/drivers/media/platform/raspberrypi/rp1_cfe/cfe.c
+++ b/drivers/media/platform/raspberrypi/rp1_cfe/cfe.c
@@ -461,7 +461,7 @@ const struct cfe_fmt *find_format_by_code(u32 code)
 	return NULL;
 }
 
-static const struct cfe_fmt *find_format_by_pix(u32 pixelformat)
+const struct cfe_fmt *find_format_by_pix(u32 pixelformat)
 {
 	unsigned int i;
 
--- a/drivers/media/platform/raspberrypi/rp1_cfe/cfe.c
+++ b/drivers/media/platform/raspberrypi/rp1_cfe/cfe.c
@@ -473,6 +473,46 @@ const struct cfe_fmt *find_format_by_pix(u32 pixelformat)
 	return NULL;
 }
 
+/*
+ * Given the mbus code, find the 16 bit remapped code. Returns 0 if no remap
+ * possible.
+ */
+u32 cfe_find_16bit_code(u32 code)
+{
+	const struct cfe_fmt *cfe_fmt;
+
+	cfe_fmt = find_format_by_code(code);
+
+	if (!cfe_fmt || !cfe_fmt->remap[CFE_REMAP_16BIT])
+		return 0;
+
+	cfe_fmt = find_format_by_pix(cfe_fmt->remap[CFE_REMAP_16BIT]);
+	if (!cfe_fmt)
+		return 0;
+
+	return cfe_fmt->code;
+}
+
+/*
+ * Given the mbus code, find the 8 bit compressed code. Returns 0 if no remap
+ * possible.
+ */
+u32 cfe_find_compressed_code(u32 code)
+{
+	const struct cfe_fmt *cfe_fmt;
+
+	cfe_fmt = find_format_by_code(code);
+
+	if (!cfe_fmt || !cfe_fmt->remap[CFE_REMAP_COMPRESSED])
+		return 0;
+
+	cfe_fmt = find_format_by_pix(cfe_fmt->remap[CFE_REMAP_COMPRESSED]);
+	if (!cfe_fmt)
+		return 0;
+
+	return cfe_fmt->code;
+}
+
 static int cfe_calc_format_size_bpl(struct cfe_device *cfe,
 				    const struct cfe_fmt *fmt,
 				    struct v4l2_format *f)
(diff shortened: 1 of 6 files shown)


tomba:
Cc: @naushir

drivers/media/platform/raspberrypi/rp1_cfe/cfe.c
```diff
@@ -473,6 +473,46 @@ const struct cfe_fmt *find_format_by_pix(u32 pixelformat)
...
+	cfe_fmt = find_format_by_pix(cfe_fmt->remap[CFE_REMAP_16BIT]);
+	if (!cfe_fmt)
+		return 0;
+
+	return cfe_fmt->code;
+}
+
+/*
+ * Given the mbus code, find the 16 bit compressed code. Returns 0 if no remap
+ * possible.
```
naushir, line 516: s/find the 16 bit compressed code/find the 8 bit compressed code/

drivers/media/platform/raspberrypi/rp1_cfe/cfe_fmts.h
```diff
@@ -197,24 +197,28 @@ static const struct cfe_fmt formats[] = {
 		.code = MEDIA_BUS_FMT_SBGGR16_1X16,
 		.depth = 16,
 		.flags = CFE_FORMAT_FLAG_FE_OUT,
+		.remap = { 0, V4L2_PIX_FMT_PISP_COMP1_BGGR },
```
naushir: For the remap, instead of 0, should we use V4L2_PIX_FMT_*?  Not sure it matters.

drivers/media/platform/raspberrypi/rp1_cfe/pisp_fe.c
```diff
@@ -433,26 +433,63 @@ static int pisp_fe_pad_set_fmt(struct v4l2_subdev *sd,
...
-		format->format.code = MEDIA_BUS_FMT_FIXED;
-		break;
-	}
+		fmt = v4l2_subdev_get_pad_format(sd, state, FE_OUTPUT0_PAD);
+		*fmt = format->format;
 
-	fmt = v4l2_subdev_get_pad_format(sd, state, format->pad);
-	*fmt = format->format;
+		fmt = v4l2_subdev_get_pad_format(sd, state, FE_OUTPUT1_PAD);
+		*fmt = format->format;
```
naushir, line 445: Does this mean that userland *has* to set the FE_STREAM_PAD format before FE_OUTPUT0_PAD and FE_OUTPUT1_PAD?

drivers/media/platform/raspberrypi/rp1_cfe/cfe.c
```diff
@@ -250,7 +258,9 @@ struct cfe_node {
 	/* Pointer pointing to next v4l2_buffer */
 	struct cfe_buffer *next_frm;
 	/* Used to store current pixel format */
-	struct v4l2_format fmt;
+	struct v4l2_format vid_fmt;
+	/* Used to store current meta format */
+	struct v4l2_format meta_fmt;
```
naushir, line 263: Why do we need both these since you can only have one of them populated?

naushir (commented):
Sorry for the late comments, but this slipped from my radar.

drivers/media/platform/raspberrypi/rp1_cfe/pisp_fe.c
tomba: Yes. You should always configure the sink side of the subdev first, and then the source side.

Of course, once you have configured the sink side, you now what the sink side format is, and you can then configure the source side multiple times (presuming all the source side configs are compatible with the sink side).

I think you've asked about this, or similar cases, earlier too. Do you see some issues with this? Afaic, this is how all well-behaved subdev drivers work. See Documentation/userspace-api/media/v4l/dev-subdev.rst, search for "propagate".

drivers/media/platform/raspberrypi/rp1_cfe/cfe.c
tomba: Good question. You _can_ have both of the populated. This is one of those v4l2 things again...

When you have a video node that can support both video and meta, you can call e.g. VIDIOC_S_FMT with type of V4L2_BUF_TYPE_VIDEO_CAPTURE, and VIDIOC_S_FMT with type of V4L2_BUF_TYPE_META_CAPTURE. Both will succeed.

It's only when you queue the buffers the type is "locked in". Until that, it's... both/neither =).

Having separate fields for vid and meta was the most sensible solution, which still makes v4l2-compliance happy, I was able to come up with.

drivers/media/platform/raspberrypi/rp1_cfe/pisp_fe.c
naushir: Ok sounds reasonable.

drivers/media/platform/raspberrypi/rp1_cfe/cfe.c
naushir: Thanks for the explanation.  Let's keep it as-is.

naushir:
All tests seem to pass fine.  Nothing else to do but merge now...

PR #5635, opened 2023-10-05
Port Pi 5 changes to rpi-6.5.y

pelwell:


popcornmix:
LGTM

Compiles and runs on pi5 (nfs boot, keyboard/mouse, kodi etc).

PR #5638, opened 2023-10-06
gpio: brcmstb: Use dynamic GPIO base numbers

pelwell:
Forcing a gpiochip to have a fixed base number now leads to a warning message. Remove the need to do so by calculating hwirq numbers based on bank numbers.

Fixes: 3b0213d56eb7 ("gpio: Add GPIO support for Broadcom STB SoCs")


6by9:
Seeing as the main use of this is having stable numbers when exporting via /sys/class/gpio, do we follow Ubuntu 21.04 (Hirsute Hippo) and drop that deprecated API?

https://waldorf.waveform.org.uk/2021/the-pins-they-are-a-changin.html

pelwell:
We've not used fixed GPIO numbers for years, but the sysfs interface still works. Arguably the switch to Bookworm is an opportunity to drop the sysfs API, but it's a bit short notice. I'd rather try get the word out that it will disappear when we move our stable kernels to 6.6/6.7 - whatever is the next LTS.

pelwell:
Although this patch was prompted by the warnings in 6.5, it should work equally well in 6.1 and may stop people assuming the GPIO numbers will be nice (they probably will be effectively fixed though).

Yes, that looks better (i.e. horrible):
```
lrwxrwxrwx 1 root gpio    0 Oct  6 17:24 gpiochip399 -> ../../devices/platform/axi/1000120000.pcie/1f000d0000.gpio/gpio/gpiochip399
lrwxrwxrwx 1 root gpio    0 Oct  6 17:24 gpiochip453 -> ../../devices/platform/soc/107d517c00.gpio/gpio/gpiochip453
lrwxrwxrwx 1 root gpio    0 Oct  6 17:24 gpiochip459 -> ../../devices/platform/soc/107d517c00.gpio/gpio/gpiochip459
lrwxrwxrwx 1 root gpio    0 Oct  6 17:24 gpiochip476 -> ../../devices/platform/soc/107d508500.gpio/gpio/gpiochip476
lrwxrwxrwx 1 root gpio    0 Oct  6 17:24 gpiochip480 -> ../../devices/platform/soc/107d508500.gpio/gpio/gpiochip480
```

6by9:
We've not relied on them, but until this patch in 6.5 at least brcmstb has been asking for a specific base GPIO.

Dropping it from 6.5 onwards would be reasonable in my book, so it would hit with the next LTS.

I guess we need to drop https://github.com/raspberrypi/linux/commit/2afa8798ada2b03fd6ae3ef5c5dfcb39a5877ad8 from 6.5 and later to avoid the same warning on pinctrl-2835?

pelwell:
Reversion added.

6by9:
Looks like the warn has been in since 6.2, but we just haven't noticed until now.
https://github.com/raspberrypi/linux/commit/502df79b860563d79143be7a1453c2b3224cd836

PR #5640, opened 2023-10-07
Add panel-dsi for device-tree side configuration of generic DSI panels

timonsku:
This adds an analog to the already existing panel-dpi to panel-simple which allows for timing configuration in device-tree for simple DSI panels.
In addition this also allows configuration of the DSI parameters in device tree.
I tried to be as consistent as possible with how panel-dpi was implemented.
Code was tested with a 840x480 panel driven by an externally configured bridge IC.

This addition was discussed with @6by9 over in https://github.com/raspberrypi/firmware/issues/1819#issuecomment-1725769614
The main motivation is with not needing to run custom kernels for simple DSI devices that require nothing but an entry in panel-simple for the timings and DSI config flags.
This becomes very cumbersome when a large amount of displays need to be maintained. Essentially the exact same reasoning that the panel-dpi patch had for the initial implementation for this type of feature in panel-simple but expanding it to also be possible for DSI panels.

Any feedback would be welcome.

### DIFF
--- a/Documentation/devicetree/bindings/display/panel/panel-dsi.yaml
+++ b/Documentation/devicetree/bindings/display/panel/panel-dsi.yaml
@@ -1,0 +1,118 @@
+# SPDX-License-Identifier: (GPL-2.0-only OR BSD-2-Clause)
+%YAML 1.2
+---
+$id: http://devicetree.org/schemas/display/panel/panel-dsi.yaml#
+$schema: http://devicetree.org/meta-schemas/core.yaml#
+
+title: Generic MIPI DSI Panel
+
+maintainers:
+  - Timon Skerutsch <kernel@diodes-delight.com>
+
+allOf:
+  - $ref: panel-common.yaml#
+
+properties:
+  compatible:
+    description:
+      Shall contain a panel specific compatible and "panel-dsi"
+      in that order.
+    items:
+      - {}
+      - const: panel-dsi
+
+  dsi-color-format:
+    description: |
+      The color format used by the panel. Only DSI supported formats are allowed.
+    enum:
+      - RGB888
+      - RGB666
+      - RGB666_PACKED
+      - RGB565
+
+  port:
+    $ref: /schemas/graph.yaml#/$defs/port-base
+    unevaluatedProperties: false
+    description:
+      Panel MIPI DSI input
+
+    properties:
+      endpoint:
+        $ref: /schemas/media/video-interfaces.yaml#
+        unevaluatedProperties: false
+
+        properties:
+          data-lanes: true
+
+        required:
+          - data-lanes
+
+  mode:
+    description: |
+      DSI mode flags. See DSI Specs for details.
+      These are driver independent features of the DSI bus.
+    items:
+      - const: MODE_VIDEO
+      - const: MODE_VIDEO_BURST
+      - const: MODE_VIDEO_SYNC_PULSE
+      - const: MODE_VIDEO_AUTO_VERT
+      - const: MODE_VIDEO_HSE
+      - const: MODE_VIDEO_NO_HFP
+      - const: MODE_VIDEO_NO_HBP
+      - const: MODE_VIDEO_NO_HSA
+      - const: MODE_VSYNC_FLUSH
+      - const: MODE_NO_EOT_PACKET
+      - const: CLOCK_NON_CONTINUOUS
+      - const: MODE_LPM
+      - const: HS_PKT_END_ALIGNED
+
+  reg: true
+  backlight: true
+  enable-gpios: true
+  width-mm: true
+  height-mm: true
+  panel-timing: true
+  power-supply: true
+  reset-gpios: true
+  ddc-i2c-bus: true
+
+required:
+  - panel-timing
+  - reg
+  - power-supply
+  - dsi-color-format
+  - port
+
+additionalProperties: false
+
+examples:
+  - |
+    panel {
+        compatible = "panel-mfgr,generic-dsi-panel","panel-dsi";
+        power-supply = <&vcc_supply>;
+        backlight = <&backlight>;
+        dsi-color-format = "RGB888";
+        reg = <0>;
+        mode = "MODE_VIDEO", "MODE_VIDEO_BURST", "MODE_NO_EOT_PACKET";
+
+        port {
+            panel_dsi_port: endpoint {
+                data-lanes = <1 2>;
+                remote-endpoint = <&dsi_out>;
+            };
+        };
+
+        panel-timing {
+            clock-frequency = <9200000>;
+            hactive = <800>;
+            vactive = <480>;
+            hfront-porch = <8>;
+            hback-porch = <4>;
+            hsync-len = <41>;
+            vback-porch = <2>;
+            vfront-porch = <4>;
+            vsync-len = <10>;
+        };
+    };
+
+...
--- a/arch/arm/boot/dts/overlays/Makefile
+++ b/arch/arm/boot/dts/overlays/Makefile
@@ -294,6 +294,7 @@ dtbo-$(CONFIG_ARCH_BCM2835) += \
 	vc4-kms-dpi-hyperpixel4sq.dtbo \
 	vc4-kms-dpi-panel.dtbo \
 	vc4-kms-dsi-7inch.dtbo \
+	vc4-kms-dsi-generic.dtbo \
 	vc4-kms-dsi-lt070me05000.dtbo \
 	vc4-kms-dsi-lt070me05000-v2.dtbo \
 	vc4-kms-dsi-waveshare-panel.dtbo \
--- a/arch/arm/boot/dts/overlays/README
+++ b/arch/arm/boot/dts/overlays/README
@@ -4789,6 +4789,37 @@ Params: sizex                   Touchscreen size x (default 800)
                                 the default DSI1 and i2c_csi_dsi).
 
 
+Name:   vc4-kms-dsi-generic
+Info:   Enable a generic DSI display under KMS.
+        Default timings are for a 840x480 RGB888 panel.
+        Requires vc4-kms-v3d to be loaded.
+Load:   dtoverlay=vc4-kms-dsi-generic,<param>=<val>
+Params: clock-frequency         Display clock frequency (Hz)
+        hactive                 Horizontal active pixels
+        hfp                     Horizontal front porch
+        hsync                   Horizontal sync pulse width
+        hbp                     Horizontal back porch
+        vactive                 Vertical active lines
+        vfp                     Vertical front porch
+        vsync                   Vertical sync pulse width
+        vbp                     Vertical back porch
+        width-mm                Define the screen width in mm
+        height-mm               Define the screen height in mm
+        rgb565                  Change to RGB565 output
+        rgb666                  Change to RGB666 output
+        rgb666p                 Change to RGB666 output with pixel packing
+        rgb888                  Change to RGB888 output, this is the default
+        one-lane                Use one DSI lane for data transmission
+                                This is the default
+        two-lane                Use two DSI lanes for data transmission
+        three-lane              Use three DSI lanes for data transmission
+                                Only supported on Pi5 and CM
+        four-lane               Use four DSI lanes for data transmission
+                                Only supported on Pi5 and CM
+        dsi0                    Switch DSI port to DSI0
+                                Only supported on Pi5 and CM
+
+
 Name:   vc4-kms-dsi-lt070me05000
 Info:   Enable a JDI LT070ME05000 DSI display on DSI1.
         Note that this is a 4 lane DSI device, so it will only work on a Compute
--- a/drivers/gpu/drm/panel/panel-simple.c
+++ b/drivers/gpu/drm/panel/panel-simple.c
@@ -40,6 +40,7 @@
 #include <drm/drm_edid.h>
 #include <drm/drm_mipi_dsi.h>
 #include <drm/drm_panel.h>
+#include <drm/drm_of.h>
 
 /**
  * struct panel_desc - Describes a simple panel.
@@ -4659,6 +4660,9 @@ static const struct panel_desc_dsi osd101t2045_53ts = {
 	.lanes = 4,
 };
 
+// for panels using generic panel-dsi binding
+static struct panel_desc_dsi panel_dsi;
+
 static const struct of_device_id dsi_of_match[] = {
 	{
 		.compatible = "auo,b080uan01",
@@ -4682,14 +4686,118 @@ static const struct of_device_id dsi_of_match[] = {
 		.compatible = "osddisplays,osd101t2045-53ts",
 		.data = &osd101t2045_53ts
 	}, {
+		/* Must be the last entry */
+		.compatible = "panel-dsi",
+		.data = &panel_dsi,
+	}, {
 		/* sentinel */
 	}
 };
 MODULE_DEVICE_TABLE(of, dsi_of_match);
 
+
+/* Checks for DSI panel definition in device-tree, analog to panel_dpi */
+static int panel_dsi_dt_probe(struct device *dev,
+			  struct panel_desc_dsi *desc_dsi)
+{
+	struct panel_desc *desc;
+	struct display_timing *timing;
+	const struct device_node *np;
+	const char *dsi_color_format;
+	const char *dsi_mode_flags;
+	struct property *prop;
+	int dsi_lanes, ret;
+
+	np = dev->of_node;
+
+	desc = devm_kzalloc(dev, sizeof(*desc), GFP_KERNEL);
+	if (!desc)
+		return -ENOMEM;
+
+	timing = devm_kzalloc(dev, sizeof(*timing), GFP_KERNEL);
+	if (!timing)
+		return -ENOMEM;
+
+	ret = of_get_display_timing(np, "panel-timing", timing);
+	if (ret < 0) {
+		dev_err(dev, "%pOF: no panel-timing node found for \"panel-dsi\" binding\n",
+			np);
+		return ret;
+	}
+
+	desc->timings = timing;
+	desc->num_timings = 1;
+
+	of_property_read_u32(np, "width-mm", &desc->size.width);
+	of_property_read_u32(np, "height-mm", &desc->size.height);
+
+	dsi_lanes = drm_of_get_data_lanes_count_ep(np, 0, 0, 1, 4);
+
+	if (dsi_lanes < 0) {
+		dev_err(dev, "%pOF: no or too many data-lanes defined", np);
+		return dsi_lanes;
+	}
+
+	desc_dsi->lanes = dsi_lanes;
+
+	of_property_read_string(np, "dsi-color-format", &dsi_color_format);
+	if (!strcmp(dsi_color_format, "RGB888")) {
+		desc_dsi->format = MIPI_DSI_FMT_RGB888;
+		desc->bpc = 8;
+	} else if (!strcmp(dsi_color_format, "RGB565")) {
+		desc_dsi->format = MIPI_DSI_FMT_RGB565;
+		desc->bpc = 6;
+	} else if (!strcmp(dsi_color_format, "RGB666")) {
+		desc_dsi->format = MIPI_DSI_FMT_RGB666;
+		desc->bpc = 6;
+	} else if (!strcmp(dsi_color_format, "RGB666_PACKED")) {
+		desc_dsi->format = MIPI_DSI_FMT_RGB666_PACKED;
+		desc->bpc = 6;
+	} else {
+		dev_err(dev, "%pOF: no valid dsi-color-format defined", np);
+		return -EINVAL;
+	}
+
+
+	of_property_for_each_string(np, "mode", prop, dsi_mode_flags) {
+		if (!strcmp(dsi_mode_flags, "MODE_VIDEO"))
+			desc_dsi->flags |= MIPI_DSI_MODE_VIDEO;
+		else if (!strcmp(dsi_mode_flags, "MODE_VIDEO_BURST"))
+			desc_dsi->flags |= MIPI_DSI_MODE_VIDEO_BURST;
+		else if (!strcmp(dsi_mode_flags, "MODE_VIDEO_SYNC_PULSE"))
+			desc_dsi->flags |= MIPI_DSI_MODE_VIDEO_SYNC_PULSE;
+		else if (!strcmp(dsi_mode_flags, "MODE_VIDEO_AUTO_VERT"))
+			desc_dsi->flags |= MIPI_DSI_MODE_VIDEO_AUTO_VERT;
+		else if (!strcmp(dsi_mode_flags, "MODE_VIDEO_HSE"))
+			desc_dsi->flags |= MIPI_DSI_MODE_VIDEO_HSE;
+		else if (!strcmp(dsi_mode_flags, "MODE_VIDEO_NO_HFP"))
+			desc_dsi->flags |= MIPI_DSI_MODE_VIDEO_NO_HFP;
+		else if (!strcmp(dsi_mode_flags, "MODE_VIDEO_NO_HBP"))
+			desc_dsi->flags |= MIPI_DSI_MODE_VIDEO_NO_HBP;
+		else if (!strcmp(dsi_mode_flags, "MODE_VIDEO_NO_HSA"))
+			desc_dsi->flags |= MIPI_DSI_MODE_VIDEO_NO_HSA;
+		else if (!strcmp(dsi_mode_flags, "MODE_VSYNC_FLUSH"))
+			desc_dsi->flags |= MIPI_DSI_MODE_VSYNC_FLUSH;
+		else if (!strcmp(dsi_mode_flags, "MODE_NO_EOT_PACKET"))
+			desc_dsi->flags |= MIPI_DSI_MODE_NO_EOT_PACKET;
+		else if (!strcmp(dsi_mode_flags, "CLOCK_NON_CONTINUOUS"))
+			desc_dsi->flags |= MIPI_DSI_CLOCK_NON_CONTINUOUS;
+		else if (!strcmp(dsi_mode_flags, "MODE_LPM"))
+			desc_dsi->flags |= MIPI_DSI_MODE_LPM;
+		else if (!strcmp(dsi_mode_flags, "HS_PKT_END_ALIGNED"))
+			desc_dsi->flags |= MIPI_DSI_HS_PKT_END_ALIGNED;
+	}
+
+	desc->connector_type = DRM_MODE_CONNECTOR_DSI;
+	desc_dsi->desc = *desc;
+
+	return 0;
+}
+
 static int panel_simple_dsi_probe(struct mipi_dsi_device *dsi)
 {
 	const struct panel_desc_dsi *desc;
+	struct panel_desc_dsi *dt_desc;
 	const struct of_device_id *id;
 	int err;
 
@@ -4697,7 +4805,20 @@ static int panel_simple_dsi_probe(struct mipi_dsi_device *dsi)
 	if (!id)
 		return -ENODEV;
 
-	desc = id->data;
+	if (id->data == &panel_dsi) {
+		/* Handle the generic panel-dsi binding */
+		dt_desc = devm_kzalloc(&dsi->dev, sizeof(*dt_desc), GFP_KERNEL);
+		if (!dt_desc)
+			return -ENOMEM;
+
+		err = panel_dsi_dt_probe(&dsi->dev, dt_desc);
+		if (err < 0)
+			return err;
+
+		desc = dt_desc;
+	} else {
+		desc = id->data;
+	}
 
 	err = panel_simple_probe(&dsi->dev, &desc->desc);
 	if (err < 0)
--- a/arch/arm/boot/dts/overlays/vc4-kms-dsi-generic-overlay.dts
+++ b/arch/arm/boot/dts/overlays/vc4-kms-dsi-generic-overlay.dts
@@ -1,0 +1,106 @@
+/dts-v1/;
+/plugin/;
+
+/ {
+	compatible = "brcm,bcm2835";
+
+	dsi_frag: fragment@0 {
+		target = <&dsi1>;
+		__overlay__{
+			status = "okay";
+			#address-cells = <1>;
+			#size-cells = <0>;
+			port {
+				dsi_out:endpoint {
+					remote-endpoint = <&panel_dsi_port>;
+				};
+			};
+			panel: panel-dsi-generic@0 {
+				// See panel-dsi.yaml binding
+				// Using dummy name for panel model
+				compatible = "Generic,panel-dsi","panel-dsi";
+				reg = <0>;
+				power-supply = <0>;
+				backlight = <0>;
+				dsi-color-format = "RGB888";
+				mode = "MODE_VIDEO";
+				width-mm = <0>;
+				height-mm = <0>;
+
+				port {
+					panel_dsi_port: endpoint {
+						data-lanes = <1>;
+						remote-endpoint = <&dsi_out>;
+					};
+				};
+
+				timing: panel-timing {
+					clock-frequency = <30000000>;
+					hactive = <840>;
+					vactive = <480>;
+					hback-porch = <44>;
+					hfront-porch = <46>;
+					hsync-len = <2>;
+					vback-porch = <18>;
+					vfront-porch = <16>;
+					vsync-len = <2>;
+				};
+			};
+		};
+	};
+
+	fragment@1 {
+		target = <&panel_dsi_port>;
+		__dormant__ {
+			data-lanes = <1>;
+		};
+	};
+
+	fragment@2 {
+		target = <&panel_dsi_port>;
+		__dormant__ {
+			data-lanes = <1 2>;
+		};
+	};
+
+	fragment@3 {
+		target = <&panel_dsi_port>;
+		__dormant__ {
+			data-lanes = <1 2 3>;
+		};
+	};
+
+	fragment@4 {
+		target = <&panel_dsi_port>;
+		__dormant__ {
+			data-lanes = <1 2 3 4>;
+		};
+	};
+
+	__overrides__ {
+		dsi0 = <&dsi_frag>, "target:0=",<&dsi0>;
+
+		clock-frequency = <&timing>, "clock-frequency:0";
+		hactive = <&timing>, "hactive:0";
+		hfp = <&timing>, "hfront-porch:0";
+		hsync = <&timing>, "hsync-len:0";
+		hbp = <&timing>, "hback-porch:0";
+		vactive = <&timing>, "vactive:0";
+		vfp = <&timing>, "vfront-porch:0";
+		vsync = <&timing>, "vsync-len:0";
+		vbp = <&timing>, "vback-porch:0";
+
+		width-mm = <&panel>, "width-mm:0";
+		height-mm = <&panel>, "height-mm:0";
+
+		rgb565 = <&panel>, "dsi-color-format=RGB565";
+		rgb666p = <&panel>, "dsi-color-format=RGB666_PACKED";
+		rgb666 = <&panel>, "dsi-color-format=RGB666";
+		rgb888 = <&panel>, "dsi-color-format=RGB888";
+		one-lane = <0>,"+1";
+		two-lane = <0>,"+2";
+		three-lane = <0>,"+3";
+		four-lane = <0>,"+4";
+	};
+
+};


drivers/gpu/drm/panel/panel-simple.c
```diff
@@ -4681,23 +4684,123 @@ static const struct of_device_id dsi_of_match[] = {
...
+	} else if (!strcmp(dsi_color_format, "RGB666")) {
+		desc_dsi->format = MIPI_DSI_FMT_RGB666;
+		desc->bpc = 6;
+	} else if (!strcmp(dsi_color_format, "RGB666_PACKED")) {
+		desc_dsi->format = MIPI_DSI_FMT_RGB666_PACKED;
+		desc->bpc = 6;
+	}
+
+	of_property_read_string(np, "mode", &dsi_mode_flags);
+	if (!strcmp(dsi_mode_flags, "MODE_VIDEO"))
```
6by9: The MIPI_DSI_MODE_xxx flags are generally used in combination, not mutually exclusive.
eg MODE_VIDEO switches between command mode and video mode displays. Almost all the others are only applicable to video mode, so have to have that set as well.

If your example binding with
```
mode = "MODE_VIDEO", "MODE_VIDEO_BURST", "MODE_NO_EOT_PACKET";
```
is valid, then that is great. However I don't see this parsing code reading all of those flags.

Documentation/devicetree/bindings/display/panel/panel-dsi.yaml
```diff
@@ -0,0 +1,111 @@
...
+
+properties:
+  compatible:
+    description:
+      Shall contain a panel specific compatible and "panel-dsi"
+      in that order.
+    items:
+      - {}
+      - const: panel-dsi
+
```
6by9, line 23: Referencing https://github.com/torvalds/linux/blob/master/Documentation/devicetree/bindings/display/panel/panel-simple-dsi.yaml#L58-L60, you ought to be supporting a reg property to set the DSI virtual channel.

drivers/gpu/drm/panel/panel-simple.c
```diff
@@ -4681,23 +4684,123 @@ static const struct of_device_id dsi_of_match[] = {
...
 	int err;
 
 	id = of_match_node(dsi_of_match, dsi->dev.of_node);
 	if (!id)
 		return -ENODEV;
 
-	desc = id->data;
+	if (id->data == &panel_dsi) {
+		/* Handle the generic panel-dsi binding */
+		err = panel_dsi_dt_probe(&dsi->dev, dt_desc);
```
6by9, line 4814: The handlng of dt_desc here looks suspicious.

Defined as an uninitialised pointer, that uninitialised value is passed to panel_dsi_dt_probe which dereferences it and writes the desc field. We then store that uninitialised value in desc for use later :-/

drivers/gpu/drm/panel/panel-simple.c
```diff
@@ -4681,23 +4684,123 @@ static const struct of_device_id dsi_of_match[] = {
...
+		dev_err(dev, "%pOF: no panel-timing node found for \"panel-dsi\" binding\n",
+			np);
+		return ret;
+	}
+
+	desc->timings = timing;
+	desc->num_timings = 1;
+
+	of_property_read_u32(np, "width-mm", &desc->size.width);
+	of_property_read_u32(np, "height-mm", &desc->size.height);
```
6by9, line 4732: Up to here in this function is all common with panel_dpi_probe.

If we change the 
```
if (desc == &panel_dpi) {
```
at https://github.com/torvalds/linux/blob/master/drivers/gpu/drm/panel/panel-simple.c#L593 to
```
if (desc == &panel_dpi || desc == &panel_dsi) {
```
don't you get that for free?

drivers/gpu/drm/panel/panel-simple.c
```diff
@@ -500,6 +500,7 @@ static int panel_dpi_probe(struct device *dev,
 	return 0;
 }
 
+
```
6by9: Random whitespace change

drivers/gpu/drm/panel/panel-simple.c
pelwell: Hint: Look at `of_property_for_each_strings`, which uses the length of the property to decide if there may be more strings.

Aside: I find the following related code in `of_prop_next_string` worrying:
```
	curv += strlen(cur) + 1;
	if (curv >= prop->value + prop->length)
		return NULL;
```
The second line acknowledges that the last string may not be NUL-terminated by comparing the apparent length of the string with the overall length of the property, but by then it may be too late - the strlen may already have stumbled into an invalid page. I should work up the enthusiasm to raise the matter upstream.

timonsku: Ooh wow yea that fell of my todo, thanks for spotting that. My panel for testing only requires MODE_VIDEO hence working fine despite this bug..

drivers/gpu/drm/panel/panel-simple.c
timonsku: I might be missing what you mean but panel_dsi_dt_probe allocates the desc struct that gets assigned?

drivers/gpu/drm/panel/panel-simple.c
timonsku: The flow of DSI panels is quite seperate to all the other types. What you propose is what I had done previously but then decided against it as it breaks with how the driver is currently constructed (note that two drivers are actually registered, one for panel-simple and one for panel-simple-dsi).
I'm unsure tbh what the best approach is but yes there is a fair chunk of duplicate code but panel_dpi_probe assumes a dpi panel and writes dpi specific properties.
You could to some degree merge these two probe functions but it would require then two seperate spaces where we parse the device tree, one for the DSI port specific aspects set in panel_desc_dsi in panel_simple_dsi_probe and in panel_dpi probe with some if else required to set the dpi or dsi specific settings of panel_desc.

Separating the two despite the overlap in function and following the currently set flow of panel-simple-dsi seemed like the better approach.
Generally speaking panel-simple might require a broader restructuring as there are two competing ideas in there.
If I were to do that I would remove all of the existing panel definitions, move them into device tree overlays and have the driver be fully device tree defined but I don't think that would be acceptable as it would be a breaking change and given how contentious this was in the past would probably mean a lengthy discussion with upstream first.

drivers/gpu/drm/panel/panel-simple.c
timonsku: Oh wait yea no I think you have a point, the outer member pointer itself does not get allocated. This might be running on undefined behaviour right now, otherwise it should blow up which it weirdly doesn't atm.

What would you think about making dt_desc static? My main gripe there was that the panel definitions are supposed to be treated as const there so I can't reuse panel_dsi which would be the better way really.

drivers/gpu/drm/panel/panel-simple.c
6by9: `panel_dpi_probe` currently takes `*dev` and a `struct panel_simple *panel`. It only uses `*panel` to fill in the `desc` field.

Modify it to take a `struct panel_desc **desc`, and pass in &panel-desc from both dpi and dsi paths. The DSI probe can then continue and read the DSI specific bits.

Yes it would leave a couple of minor and inconsequential DPI related things executed in the DSI path, but if viewed necessary those could be split out.

drivers/gpu/drm/panel/panel-simple.c
6by9: You can't reuse `panel_dsi` as you could have 2 displays configured through the same `panel-dsi` compatible.

You are right that there is a conflict that `panel_simple_probe` gets a dummy desc as panel_dpi, and then `panel_dpi_probe` handles allocating the desc. That bit almost does want to be pulled out in order to make it potentially common.

Your current implementation does all the work before `panel_simple_probe` gets a look in.

aBUGSworstnightmare-rpi:
May I ask which bridge device this is needed/intended for?
for a MIPI-to-DPI bridge (as used on the official 7in) a "panel-dpi" compatible is there already 
```
		panel_generic: __overlay__  {
			compatible = "panel-dpi";

			width-mm = <154>;
			height-mm = <83>;
			bus-format = <0x1009>;

			timing: panel-timing {
				clock-frequency = <29500000>;
				hactive = <800>;
				hfront-porch = <24>;
				hsync-len = <72>;
				hback-porch = <96>;
				hsync-active = <1>;
				vactive = <480>;
				vfront-porch = <3>;
				vsync-len = <10>;
				vback-porch = <7>;
				vsync-active = <1>;

				de-active = <1>;
				pixelclk-active = <1>;
			};
		};
```

for a MIPI-to-LVDS bridge one needs to enable 'Generic LVDS panel driver' in order to use  "panel-lvds" compatible from DT
i.e.
```
		panel: panel {
				compatible = "panel-lvds";

				backlight = <&backlight_lvds>;

				/* Physical dimensions of active area */
				width-mm = <217>;
				height-mm = <136>;

				data-mapping = "vesa-24";

				panel-timing {
					clock-frequency = <71100000>;
					hactive = <1280>;
					hfront-porch = <160>;	
					hsync-len = <48>;
					hback-porch = <48>;
					
					vactive = <800>;
					vfront-porch = <23>;
					vsync-len = <6>;					
					vback-porch = <8>;
				};
```

Bridge input timing will have to be calculated by the driver based on the required output (in most cases) and other parameters (i.e. no data-lanes can be configured via overrides).

so, what is the real target here? Having the possibility for adding 'dumb panels' (panels which just require a suitable timing on the DSI interface)? But that would not be in-line with

> https://github.com/raspberrypi/firmware/issues/1819
I'm currently working on product for Adafruit that would support different panels via a bridge IC like it is used in the official touchscreen.

Would be glad if you can share some more details so others might be able to support.

drivers/gpu/drm/panel/panel-simple.c
timonsku: Oh right, I guess that might need fixing then too for panel-dpi, that is currently re-useing the single dummy struct :)
Though I'm not sure how you would handle multiple panels like this properly that does not rely on hard coded dummy panels to refer to. I think that would require some deeper changes to remove the assumption that a panel is in some way already statically defined. I might need to test this if it might already work unintentionally, the whole driver is a bit of a rabbit hole of multiple layers of structs being passed around and torn apart.

drivers/gpu/drm/panel/panel-simple.c
timonsku: Ok! So my approach so far was to only add and not change anything. The only reason for that was to appease to an upstreaming process as much as possible. If you think it would be fine to make some changes to the existing things then I would propose to keep the probing functions seperate but pull out the panel timing and panel-common property parsing into a new function that takes on the same role as the EDID parsing does. That function adds additional timings to a panel that were found on the EDID connected to the DT defined I2C port.
That feels like a more generic way and would allow any type of "connector" to get timings from device tree and the same wouldn't need to be done for say LVDS.
I'm a bit unsure about the dynamic creation of a panel as discussed above though. Could stick with having some dummy structs for now or see to handle this a bit more properly. A bit of a question of scope I guess, there is a lot that could technically be improved overall.
What are your thoughts?

timonsku:
@aBUGSworstnightmare-rpi 
There are multiple things where this would be needed. The use case mentioned in the firmware issue is one of them but does not really relate to the IC itself, that could be any bridge really but yes its a DSI to DPI bridge IC in that instance but might also change down the line which one specifically or new products with different bridges being added where its desirable to keep software compatibility so that the process for the end user looks the same, so the bootstrapping would happen externally and the requirement to the host would only be to send a suitable DSI stream.
A fixed port configuration would not be a great solution as different applications have different needs in terms of DSI bus flags.

"Native" panels without a bridge IC are also a target. The wish is really only to be able to define what one can define in a panel_desc_dsi but doing that at runtime without needing to get a patch merged or distribute a kernel fork.
There are potentially dozens or even hundreds of different panels in the future and defining them all in the driver is just not sustainable, again exact same motivation as with the original pane-dpi patch that enabled DT timings but wanting control over the DSI bits too.

aBUGSworstnightmare-rpi:
> "Native" panels without a bridge IC are also a target.

Hmm .. still don't get that!
A 'native' DSI for me is a command mode display. Find a typical power-on-sequence example from the image linked here (https://drive.google.com/file/d/1Vi3UZlcxU89LIjbEj2nI8zCIRfpN5BTa/view?usp=share_link)

You can't fire this display up with a panel-simple-dsi compatible; you will have to use panel-dpi-cm. That's pretty much the same for all native DSI displays which autoload TCON register settings from NVM; one still has to send 0x51/0x53/0x55 before you can switch the display on (0x29) and exit sleep mode (0x11) for normal operation. 
Looking at the sequence, a 'dumb' display simply needs 'Host Display Data transfer'

6by9:
Native DSI can be either command mode or video mode.
I've never encountered a command mode panel in reality and have no idea how well (or not) they'll behave. When someone actually finds one, I'll worry about it.

Whilst I wouldn't advocate it, the Waveshare DSI panels could be considered native DSI panels because the MCU handles all the initialisation. The regulator and backlight side could be handled by a regulator driver similar to rpi_panel_attiny_regulator, which leaves the panel side as a simple DSI device. That is a valid use case in my book.

I would advocate that bridge chips are configured from Linux rather than hidden behind an MCU. At least that way there is visibility over what the hardware is actually doing, and if a compatibility issue is observed with a future chip then it can be addressed. Most bridge chip drivers should be configured automatically by the upstream panel device, so other than a bit of DT it adds no overhead.

timonsku:
@aBUGSworstnightmare-rpi see the panels in the `dsi_of_match` array for examples of DSI video mode panels that do no require initialization.
And what 6by9 described, sometimes its useful to init externally for either systems reasons or because the init sequence is inexplicably under NDA. In those cases you can turn them into simple panels.

@6by9 generally agree with the sentiment about bridges! It gets a bit more complicated in some specific dev board style applications as I found out as the needs break with some assumptions in existing drivers or guidelines of how drivers should behave. That is a very niche thing though where I feel re-using generic infrastructure might be the nicer approach.

ladyada:
(subscribing myself by commenting)

pelwell:
That works - there's also a Subscribe button in the controls on the right at the top of the page (it probably says Unsubscribe now).

pelwell:
Can you stop force-pushing your branch unless you are actively changing your code? GitHub is pretty good at rebasing when we want to merge, and every update invalidates the previous build check results.

timonsku:
sorry, there were commits that were not supposed to go into this branch that I wanted to roll back. Not going to happen again.

drivers/gpu/drm/panel/panel-simple.c
timonsku: Got some DSI cables for PI5 today so I could test multiple panels working correctly.
That works well with the latest commit.
Regarding my earlier comment in the other thread. While I think that is probably the best approach down the line I found that it becomes quite far reaching to add a general "get timing from dt" functionality in that isn't port specific like right now.
I think that would be something that should probably be tested across more devices and I don't currently have hardware here to also test this with LVDS.
Keeping the ~15 duplicate lines feels like the better compromise for now as its otherwise awkward to merge just the dpi and dsi specific probing bits.
Let me know what you think of this.

arturo182:
Just wanted to add that I tested this PR on the CM4 with a custom carrier I'm working on, and it works great! I no longer have to patch `panel-simple.c` to add my custom panel :)

ladyada:
@pelwell ive also had much success with fun displays connecting through the ICN6211 (with a microcontroller doing the i2c/spi configuration during power-up) - can you approve the workflow so it could get reviewed before the next kernel release is made? :)

image

pelwell:
Is this actually ready? I see that the Draft status has been removed (GitHub doesn't give notifications about that, so a comment is always a good idea), but some of the comments on the PR have been acknowledged but not acted on.

Documentation/devicetree/bindings/display/panel/panel-dsi.yaml
```diff
@@ -0,0 +1,114 @@
...
+  - dsi-color-format
+  - lanes
+
+additionalProperties: false
+
+examples:
+  - |
+    panel {
+        compatible = "panel-mfgr,generic-dsi-panel","panel-dsi";
+        label = "Simple DSI Panel";
```
6by9: https://github.com/torvalds/linux/blob/master/Documentation/devicetree/bindings/display/panel/panel-common.yaml#L34-L46
>       If no such name exists, and unless the property is mandatory according to
>      device tree bindings, it shall rather be omitted than constructed of
>      non-descriptive information. For instance an LCD panel in a system that
>      contains a single panel shall not be labelled "LCD" if that name is not
>      inscribed on the system or used in a descriptive fashion in system
>      documentation.

Documentation/devicetree/bindings/display/panel/panel-dsi.yaml
```diff
@@ -0,0 +1,114 @@
...
+            hback-porch = <4>;
+            hsync-len = <41>;
+            vback-porch = <2>;
+            vfront-porch = <4>;
+            vsync-len = <10>;
+
+            hsync-active = <0>;
+            vsync-active = <0>;
+            de-active = <1>;
+            pixelclk-active = <1>;
```
6by9: None of these 4 are applicable to DSI so should be omitted.

drivers/gpu/drm/panel/panel-simple.c
6by9: Still present

Documentation/devicetree/bindings/display/panel/panel-dsi.yaml
```diff
@@ -0,0 +1,114 @@
...
+      - RGB888
+      - RGB666
+      - RGB666_PACKED
+      - RGB565
+
+  lanes:
+    description: |
+      The number of lanes used by the panel.
+    $ref: /schemas/types.yaml#/definitions/uint32
+    enum: [1, 2, 3, 4]
```
6by9: The more consistent approach would be to use data-lanes as an array of data lane indexes. This is used across a number of bridges, and is consistent with V4L2.
eg https://github.com/torvalds/linux/blob/master/Documentation/devicetree/bindings/display/bridge/chipone,icn6211.yaml#L64-L71

drivers/gpu/drm/panel/panel-simple.c
6by9: panel_dpi is fine as `desc` gets swapped for `panel->desc`, which is allocated in `panel_dpi_probe`.

Now that you devm_kzalloc the desc, I think you're OK too.

drivers/gpu/drm/panel/panel-simple.c
```diff
@@ -4681,31 +4685,143 @@ static const struct of_device_id dsi_of_match[] = {
...
-	desc = id->data;
+	if (id->data == &panel_dsi) {
+		/* Handle the generic panel-dsi binding */
+		dt_desc = devm_kzalloc(&dsi->dev, sizeof(*dt_desc), GFP_KERNEL);
+		if (!dt_desc)
+			return -ENOMEM;
+		err = panel_dsi_dt_probe(&dsi->dev, dt_desc);
+		if (err < 0)
+			return err;
+		err = panel_simple_probe(&dsi->dev, &dt_desc->desc);
```
6by9: Massive duplication here.
```
if (id->data == &panel_dsi) {
  ...
   desc = &dt_desc->desc;
} else {
   desc = id->data;
}

err = panel_simple_probe(...);
...
```

6by9 (changes requested):
Sorry, still unaddressed issues.

There needs to be 2 commits - one device tree binding, and one driver change. The third "fixup" patch touches both binding and driver.

Checkpatch complaint as there is no commit description.

How about a 3rd commit that adds an overlay using this, along similar lines to https://github.com/raspberrypi/linux/blob/rpi-6.1.y/arch/arm/boot/dts/overlays/vc4-kms-dpi-generic-overlay.dts or vc4-kms-dpi-panel-overlay.dts

For those adding panels using this driver change, I would like to see it done in a similar way to https://github.com/raspberrypi/linux/blob/rpi-6.1.y/arch/arm/boot/dts/overlays/vc4-kms-dpi-panel-overlay.dts, where there is one overlay that uses an override to select the relevant panel. The timing and config can be put in that overlay.

Documentation/devicetree/bindings/display/panel/panel-dsi.yaml
timonsku: right what I was trying to convey is that this should be replaced with actual panel information by the user as its done with the dpi binding. Though for the `Label` property thats probably not obvious enough.
I think it is useful to have them so that they show up as the actual panel name but of course should be filled out properly. So f.e.
```
        compatible = "TDO,TL040HD","panel-dsi";
        label = "TL040HD";
```

Documentation/devicetree/bindings/display/panel/panel-dsi.yaml
timonsku: I don't quite understand the use of that format. The intended use seems to then be `drm_of_get_data_lanes_count_ep` to get the number of elements in the array to get the number that you currently just set directly. That seems very convoluted?
Happy to change it if its the canonical way to set DSI lane count, just feels quite odd but maybe I'm missing an aspect here. If there are PHYs that support picking specific lanes the dsi_panel struct does not support that at least as its just an unsigned int. Not sure if its good to convey a support for that if the information is just dropped.

drivers/gpu/drm/panel/panel-simple.c
timonsku: Added in last commit.

timonsku:
Thanks for the input @6by9 !
Regarding splitting binding and driver patch. I thought the initial commit of a new binding should be split into its own commit (which I did).
So what you are saying I should always keep changes to bindings completely separated too? Should I then revert the latest (third) commit and split it into two (making a total of 4 excluding upcoming fixes for the other things you addressed)

For the requested overlay, you mean this as a practical example for how to use this?

6by9:
> Thanks for the input @6by9 ! Regarding splitting binding and driver patch. I thought the initial commit of a new binding should be split into its own commit (which I did). So what you are saying I should always keep changes to bindings completely separated too? Should I then revert the latest (third) commit and split it into two (making a total of 4 excluding upcoming fixes for the other things you addressed)

With a very few exceptions, individual commits should never cross subsystem boundaries. For mainline it causes grief over which tree it should be committed through. 

We also don't want your development history here - you've been asked to make changes as the original driver has issues, therefore please amend the existing commits as they've gone nowhere except your local branch.

> For the requested overlay, you mean this as a practical example for how to use this?

Yes please.

drivers/gpu/drm/panel/panel-simple.c
timonsku: Yes, having a second a think on this I realize the reason I did this actually has a different solution that wouldn't require this. Will try to remove it.

timonsku:
> Is this actually ready? I see that the Draft status has been removed (GitHub doesn't give notifications about that, so a comment is always a good idea), but some of the comments on the PR have been acknowledged but not acted on.

Sorry I just realized that my added comments where part of making a review and did not show up publicly as I hadn't submitted the review until now.

Documentation/devicetree/bindings/display/panel/panel-dsi.yaml
6by9: AFAIK it's never passed on to the user, and it's more for those analysing the DT to be able to understand which device is being derived.
Looking at the other bindings, very few define a label (amusingly panel/lgphilips,lb035q02.yaml includes `label = "lcd";` almost to spite the writer of the docs!), and a very quick grep in arch/arm/boot/dts doesn't appear to include any displays with labels.

Please drop it.

Documentation/devicetree/bindings/display/panel/panel-dsi.yaml
6by9: I'll admit that DRM is less consistent than V4L2, but some devices support reordering of the data lanes. Using the array allows for a consistent way of configuring the interface.

Any non-standard property is meant to be prefixed with a vendor code. `data-lanes` is standardised. `lanes` is not. Friendly advice if upstreaming - use the standard properties.

drivers/gpu/drm/panel/panel-simple.c
6by9: I would like to see the duplication removed, but only if you're committing to upstream the differences.
Handling major changes to files such as this that get a fair amount of churn is not easy.

timonsku: As in for an upstream submission you would like to see the larger change that abstracts the device tree side timing definition?

Documentation/devicetree/bindings/display/panel/panel-dsi.yaml
timonsku: Thats fair yea. I will change it to `data-lanes`.

Documentation/devicetree/bindings/display/panel/panel-dsi.yaml
timonsku: thats fair, was mostly following the other panel-*.yaml bindings which make use of it but you are right its actually not passed to user space. I somehow remembered that it would be used for the display name in the window manager but that does not seem to be the case, at least not with Wayland. Does not have any value in that case.

timonsku:
Sorry for the force push but not aware of any other way to make a clean history rewrite/amend.
I think I covered all the outstanding change requests. A third commit added for an example overlay.

Documentation/devicetree/bindings/display/panel/panel-dsi.yaml
```diff
@@ -0,0 +1,111 @@
...
+
+examples:
+  - |
+    panel {
+        compatible = "panel-mfgr,generic-dsi-panel","panel-dsi";
+        power-supply = <&vcc_supply>;
+        backlight = <&backlight>;
+        dsi-color-format = "RGB888";
+        reg = <0>;
+        data-lanes = <1 2>;
```
6by9: This should be under the endpoint, not the top level of the panel.

Definition should also match.

arch/arm/boot/dts/overlays/panel-simple-dsi.dts
```diff
@@ -0,0 +1,55 @@
...
+            status = "okay";
+            #address-cells = <1>;
+            #size-cells = <0>;
+            port {
+                    dsi_out:endpoint {
+                            remote-endpoint = <&panel_dsi_port>;
+                    };
+            };
+            panel@0 {
+                // See panel-dsi.yml binding for more information
```
6by9: .yaml

drivers/gpu/drm/panel/panel-simple.c
```diff
@@ -4681,23 +4685,141 @@ static const struct of_device_id dsi_of_match[] = {
...
+		return ret;
+	}
+
+	desc->timings = timing;
+	desc->num_timings = 1;
+
+	of_property_read_u32(np, "width-mm", &desc->size.width);
+	of_property_read_u32(np, "height-mm", &desc->size.height);
+
+	dsi_lanes = drm_of_get_data_lanes_count(np, 1, 4);
```
6by9: Prototype of drm_of_get_data_lanes_count is 
```
int drm_of_get_data_lanes_count(const struct device_node *endpoint,
				const unsigned int min, const unsigned int max)
```
You've passed the device_node of the base device, not the endpoint.

Use ```drm_of_get_data_lanes_count_ep(np, 0, -1, 1, 4);```

drivers/gpu/drm/panel/panel-simple.c
```diff
@@ -4681,23 +4685,141 @@ static const struct of_device_id dsi_of_match[] = {
...
+
+	if (dsi_lanes < 0){
+		dev_err(dev, "%pOF: no or too many data-lanes defined",np);
+		return dsi_lanes;
+	}else{
+		desc_dsi->lanes = dsi_lanes;
+	}
+
+	of_property_read_u32(np, "lanes", &desc_dsi->lanes);
```
6by9: Why are we still trying to read this?

6by9 (commented):
Force pushing is the correct way to update the branch backing a PR.

Thank you - the general implementation is far cleaner.

A number of checkpatch errors.

Please give a full description in the commit text. Titles should always start with the subsystem (do `git log` on the file if in doubt, and copy all the other commits).

I was hoping that the overlay would actually be compiled and useful, with similar overrides to https://github.com/raspberrypi/linux/blob/rpi-6.1.y/arch/arm/boot/dts/overlays/vc4-kms-dpi-generic-overlay.dts

drivers/gpu/drm/panel/panel-simple.c
timonsku: oh yea thats not supposed to still be there, may have clicked the wrong thing during rebase

drivers/gpu/drm/panel/panel-simple.c
timonsku: the property does not sit in the endpoint though but is part of the panel-dsi binding to be able to pass that into the dsi_mode for setup.
`drm_of_get_data_lanes_count_ep` would not be able to find the property then. If you check the implementation of the functions the use of `endpoint` of the non `_ep` prototype for it is a bit misleading because it does not actually expect an endpoint it just looks for the data-lanes property within the node it was given and does a boundary check, nothing specific to endpoints in there.
`drm_of_get_data_lanes_count_ep` does look for an endpoint though and expects to find the property there.
So I think in this case `drm_of_get_data_lanes_count` is more apropriate. Otherwise I would not know how to specify the binding properly without modifying the inherited bindings.

Documentation/devicetree/bindings/display/panel/panel-dsi.yaml
timonsku: I guess relates to the other comment. Would you then put all the dsi properties under the endpoint?
Feels inconsistent otherwise to me if the dsi specific bits are set in different places.
Though then not really sure how that would be defined in a binding when it essentially touches the binding of something it only inherits?

timonsku:
Ah yes sorry about the missing description you mentioned that before, will run checkpatch on the actual patch. I only ran it on the source files.

@ Overlay, my bad I understood that way different. More as a template for new overlays, not to be used as is.
But yes that makes sense, so the intention is full configuration with overrides so that panels don't even necessarily have to have an overlay installed? (e.g. same function as the dpi overlay)

timonsku:
@6by9 do you have a suggestion for how to represent the mode stringarray with overrides?
As I understand there is currently no direct support for array types in the overrides syntax and I'm not aware of any devicetree "native" way of appending values to an array with overlays (in which case I could use fragments).
The only way I currently see is having a fragment for every possible combination but that is not practical.

pelwell:
We can do arrays of cells, but not arrays of strings. Exactly what change do you want the overlay to make?

6by9:
> @6by9 do you have a suggestion for how to represent the mode stringarray with overrides? As I understand there is currently no direct support for array types in the overrides syntax and I'm not aware of any devicetree "native" way of appending values to an array with overlays (in which case I could use fragments). The only way I currently see is having a fragment for every possible combination but that is not practical.

- Seeing as we don't really support command mode (I've never seen a command mode panel in the flesh), you can assume MODE_VIDEO is set. Command mode panels would be supported through [panel-dsi-cm](https://github.com/raspberrypi/linux/blob/rpi-6.1.y/drivers/gpu/drm/panel/panel-dsi-cm.c) anyway, and not panel-simple.
- MODE_LPM is redundant as this is the simple panel-dsi that sends no commands.
- HS_PKT_END_ALIGNED is a Mediatek special.
- And actually the vc4 driver currently doesn't look at the other flags, so all of them can be ignored.

So just set MODE_VIDEO and have done with it.

If we add support for the extra flags, then a vc4-kms-dsi-panel-overlay.dts overlay would allow those who wish to define their panel by name can do support that level of configuration, but the generic overlay won't support it. Overlays don't go upstream, so it's not an issue for that.

timonsku:
That's fair enough, good with me. I don't see a big need for more than video_mode either. Most other options are usually just optimizations anyway.

timonsku:
@6by9 Please see updated commits. checkpatch should be happy except for license header which does not seem to be wanted for overlays.
I enabled the I2C interface in the overlay as well, while nothing technically needs it it seems sensible for a generic overlay to allow access in case there is anything attached that you might want to talk to from userspace.

The only thing unaddressed is the location of the data-lanes property. See my comments on the last review.
If you feel like things should move then I'll revise that.
If not I think things should hopefully have no more form issues. Thanks for bearing with me there.

6by9:
> I enabled the I2C interface in the overlay as well, while nothing technically needs it it seems sensible for a generic overlay to allow access in case there is anything attached that you might want to talk to from userspace.

If you don't need it in the overlay, don't enable it.

i2c0mux will claim GPIOs 0&1 which may be used for other purposes. Your overlay would be incompatible with that for no good reason.
There's already `dtparam=i2c_vc=on` to enable it via a 1-liner, or you can use `dtoverlay=i2c0,pins44_45` if you want just the single I2C port with no mux.

> The only thing unaddressed is the location of the data-lanes property. See my comments on the last review. If you feel like things should move then I'll revise that. If not I think things should hopefully have no more form issues. Thanks for bearing with me there.

data-lanes is a standard property, and it goes in the endpoint as covered by https://www.kernel.org/doc/Documentation/devicetree/bindings/media/video-interfaces.yaml

If you have a bridge or similar DSI device, then you have an input port and an output port, each with their embedded endpoint. The number of lanes is property of the port/endpoint, not the overall device. 
Whilst you only have 1 port, the property remains under the endpoint for consistency.

timonsku:
> i2c0mux will claim GPIOs 0&1 which may be used for other purposes. Your overlay would be incompatible with that for no good reason. There's already `dtparam=i2c_vc=on` to enable it via a 1-liner, or you can use `dtoverlay=i2c0,pins44_45` if you want just the single I2C port with no mux.

That is a fair point, will revert.

> If you have a bridge or similar DSI device, then you have an input port and an output port, each with their embedded endpoint. The number of lanes is property of the port/endpoint, not the overall device. Whilst you only have 1 port, the property remains under the endpoint for consistency.

Okay yea, the disconnect to the other bus properties feels a bit odd but fair enough if there is precedence for having lanes defined in the endpoint. Will revise.

drivers/gpu/drm/panel/panel-simple.c
timonsku: switched data-lanes to be in endpoint

timonsku:
Pushed the change. I'm a bit uncertain if the bindings are defined correctly this way but given that there should only ever be a single endpoint this feels like the correct approach to me.
I'm relying on the address for port and endpoint to default to 0, afaik they don't need to/shouldn't be defined specifically if the expectation is that there can't be multiple instances?

arch/arm/boot/dts/overlays/Makefile
```diff
@@ -309,7 +309,8 @@ dtbo-$(CONFIG_ARCH_BCM2835) += \
 	waveshare-can-fd-hat-mode-a.dtbo \
 	waveshare-can-fd-hat-mode-b.dtbo \
 	wittypi.dtbo \
-	wm8960-soundcard.dtbo
+	wm8960-soundcard.dtbo \
+	panel-dsi-generic.dtbo
```
arturo182: These seem to be sorted alphabetically and putting this at the end of the list breaks the sort.

timonsku: oh true

arch/arm/boot/dts/overlays/panel-dsi-generic-overlay.dts
```diff
@@ -0,0 +1,113 @@
...
+			#address-cells = <1>;
+			#size-cells = <0>;
+			port {
+				dsi_out:endpoint {
+					remote-endpoint = <&panel_dsi_port>;
+				};
+			};
+			panel: panel-dsi-generic@0 {
+				// See panel-dsi.yaml binding
+				compatible = "panel-dsi";
```
6by9: Your bindings state
> Shall contain a panel specific compatible and "panel-dsi" in that order

except you haven't included a panel specific compatible, dummy or otherwise.

arch/arm/boot/dts/overlays/panel-dsi-generic-overlay.dts
```diff
@@ -0,0 +1,113 @@
...
+		vsync = <&timing>, "vsync-len:0";
+		vbp = <&timing>, "vback-porch:0";
+
+		width-mm = <&panel>, "width-mm:0";
+		height-mm = <&panel>, "height-mm:0";
+
+		rgb565 = <&panel>, "dsi-color-format=RGB565";
+		rgb666p = <&panel>, "dsi-color-format=RGB666_PACKED";
+		rgb666 = <&panel>, "dsi-color-format=RGB666";
+		rgb888 = <&panel>, "dsi-color-format=RGB888";
```
6by9: Not needed as this is the default already present in the panel definition.

arch/arm/boot/dts/overlays/panel-dsi-generic-overlay.dts
```diff
@@ -0,0 +1,113 @@
...
+	fragment@4 {
+		target = <&panel_dsi_port>;
+		__dormant__ {
+			data-lanes = <1 2 3 4>;
+		};
+	};
+
+	fragment@5 {
+		target = <&panel>;
+		__dormant__ {
```
6by9: Why is this dormant? 

As per earlier review comments:
> Seeing as we don't really support command mode (I've never seen a command mode panel in the flesh), you can assume MODE_VIDEO is set. Command mode panels would be supported through [panel-dsi-cm](https://github.com/raspberrypi/linux/blob/rpi-6.1.y/drivers/gpu/drm/panel/panel-dsi-cm.c) anyway, and not panel-simple.
>
> So just set MODE_VIDEO and have done with it.

arch/arm/boot/dts/overlays/Makefile
```diff
@@ -177,6 +177,7 @@ dtbo-$(CONFIG_ARCH_BCM2835) += \
 	ov5647.dtbo \
 	ov7251.dtbo \
 	ov9281.dtbo \
+	panel-dsi-generic.dtbo \
```
6by9: Please follow the pattern of the other dsi and dpi panels and name the overlay vc4-kms-dsi-generic-overlay.dts.

6by9 (commented):
All overlays must have an entry in the README file documenting the overlay and the associated overrides. In this case most of it can be copied from vc4-kms-dpi-generic.

Icing on the cake would have been to include configuration of a backlight node (as vc4-kms-dpi-generic does), but I'm not going to worry about that one now.

arch/arm/boot/dts/overlays/panel-dsi-generic-overlay.dts
pelwell: I'm OK with having parameters that select values that are also the defaults. It can take some of the guesswork out of reading a config file.

arch/arm/boot/dts/overlays/panel-dsi-generic-overlay.dts
timonsku: heh right, yea I'm was sure if it would be sensible to have dummy string there but yea for the sake of consistency it would probably be better to have a dummy string to the generic overlay

arch/arm/boot/dts/overlays/panel-dsi-generic-overlay.dts
timonsku: That was sort of my thinking, a lot of people who would use the generic overlay might never see the dts file and just follow a guide. Being explicit in the parameters seems like it would avoid potential confusion.

arch/arm/boot/dts/overlays/panel-dsi-generic-overlay.dts
timonsku: right, I wanted to possibly keep LP video mode as an option and see if it is equivalent to video mode and LP flag, then decided against it and forgot to just remove all of that entirely.
Will prune.

timonsku:
I'm with you for the backlight but I'm not sure what the best options for that are. Unlike for dpi you are not guaranteed to have a connection to the 40pin header and you might not wanna touch that with a default if you don't know if there isn't likely some form of HAT on there.
With Pi5 we now also have these two AUX IO but I'm not sure what your plans are for that and if they support PWM?
Happy to do a second PR later on when got hardware that could test that.

Have pushed the requested changes except for leaving RGB888 in as an option. If there is disagreement let me know.

arch/arm/boot/dts/overlays/Makefile
6by9: Your overlay is now named vc4-kms-dsi-generic, so the Makefile needs to match or it won't get built.

arch/arm/boot/dts/overlays/README
```diff
@@ -173,6 +173,30 @@ Params:
...
+        drm_fb1_vc4             Assign /dev/fb1 to the vc4 outputs
+
+        drm_fb2_rp1_dpi         Assign /dev/fb2 to the RP1 DPI output
+
+        drm_fb2_rp1_dsi0        Assign /dev/fb2 to the RP1 DSI0 output
+
+        drm_fb2_rp1_dsi1        Assign /dev/fb2 to the RP1 DSI1 output
+
+        drm_fb2_vc4             Assign /dev/fb2 to the vc4 outputs
+
```
6by9: Something funny has gone on with your push. These are part of 61b138adaeaddefe749f421a3b69c67d4a49a8e3, and the mcp23017 changes are part of 27f0e0e195568c06f21ce380f0736bdd219baf3c.

Your branch is based on feb4ea51e77af09c1a4c7c3e5dd33237750f95cd from 5th Oct.

Please do `git fetch` and `git rebase` to pull in all updates that have been made upstream, and that should resolve the mismatch with these erroneous fragments.

6by9 (commented):
Makefile and weird rebase fixes required.

For vc4-kms-dpi-generic the backlight is disabled by default, so you could have done the same here. If the panel implementation has suitable backlight control then the wiring is sorted via overrides.
(For those designing hardware, an I2C controlled backlight makes most sense - TI look to have a number of options in the LP855x range).
Leave it for now - this has rumbled on long enough.

timonsku:
oh yea sorry, divergent feature branch, looks clean to me now. Makefile fixed.

And yea agree, I2C backlight probably the best for now and what we've been using.

arch/arm/boot/dts/overlays/README
```diff
@@ -4712,6 +4712,35 @@ Params: clock-frequency         Display clock frequency (Hz)
...
+        height-mm               Define the screen height in mm
+        rgb565                  Change to RGB565 output
+        rgb666                  Change to RGB666 output
+        rgb666p                 Change to RGB666 output with pixel packing
+        rgb888                  Change to RGB888 output, this is the default
+        one-lane                Use one DSI lane for data transmission
+                                This is the default
+        two-lane                Use two DSI lanes for data transmission
+        three-lane              Use three DSI lanes for data transmission
+                                Only supported on Pi5+ and CM
```
6by9: Pi5+ ?

6by9 (approved):

6by9:
Nothing obviously wrong other than the Pi5+ comment. 

CI pipeline running, so will just check whether that picks anything up, but otherwise I think I'm happy. Thanks for following through on this.

arch/arm/boot/dts/overlays/README
timonsku: As in for Pi5 and "going forward" assuming that a Pi6 would keep with the 4 channels.

timonsku:
Glad to hear and thanks for the guidance!
Should I remove the comment for 3 and 4 channel option before you pull it in?

E: Hm why is the dtoverlaycheck unhappy about the Readme? Don't quite understand the error output.

6by9:
dtoverlaycheck ensures:
- README is in alphabetical order. vc4-kms-dsi-generic doesn't come between vc4-kms-dpi-generic and vc4-kms-dpi-hyperpixel2r
- All parameters are documented, and all documented parameters exist. The overlay uses XXX_lane whilst the documentation says XXX-lane. Hyphen vs underscore. Hyphens are preferred.

If you could update the overlay to use hyphens, correct the alphabetical order, and drop the erroneous `+` from `5B+`, then I think we're there.

timonsku:
oh yea, thats what I get for typing it out..
dsi0 was indeed also not in the readme. Pushed fix.

pelwell:
Yay - thanks for persevering and getting this over the line.

PR #5641, opened 2023-10-10
drivers: media: rp1_cfe: Remove PISP specific MBUS formats

naushir:
Remove the MEDIA_BUS_FMT_PISP* format codcs entirely. For the image pad formats, use the 16-bit Bayer format mbus codes instead. For the config and stats pad formats, use MEDIA_BUS_FMT_FIXED.

Remove the format validation check in cfe_video_link_validate(), as there is no 1:1 mapping from mbus codes to V4L2 format fourcc any more.


naushir:
@tomba this change gets rid of the custom pisp mbus codes.  Note we need an accompanying change in libcamera to keep userland in sync.  I'll merge that once this goes in.

EDIT: The userland tree can be found at https://github.com/raspberrypi/libcamera/tree/next_mbus

tomba:
> Remove the format validation check in cfe_video_link_validate(), as there is no 1:1 mapping from mbus codes to V4L2 format fourcc any more.

If I recall right, in my not-merged CFE series I have code that does checks like these. We can (and should) check that the v4l2 pix format is compatible with the mbus format.

naushir:
> If I recall right, in my not-merged CFE series I have code that does checks like these. We can (and should) check that the v4l2 pix format is compatible with the mbus format.

Agree - i'm in the process of adding it back.

naushir:
> Agree - i'm in the process of adding it back.

Done!

njhollinghurst (approved):
LGTM

FWIW the data on the CSI2->ISP-FE link are indeed 16-bit Bayer, albeit packed in pairs; and so are internal data in CSI2DMA channels somewhere between unpack and (optional) compress stages.

naushir:
I'm going to merge this one now.  Note that once this goes in, you will need to rebuild libcamera from https://github.com/raspberrypi/libcamera/tree/next to get the enum values to match again.

PR #5642, opened 2023-10-10
Revert "Revert "xhci: add quirk for host controllers that don't updat…

pelwell:
…e endpoint DCS""

This downstream reversion of an upstream reversion got lost, probably do to confusion caused by weird behaviour from git log --oneline when used on a named file.

This reverts commit 96a0b80eb1b02e1330d525d4c866ccdfa8c67434.

### DIFF
--- a/drivers/usb/host/xhci-ring.c
+++ b/drivers/usb/host/xhci-ring.c
@@ -601,8 +601,11 @@ static int xhci_move_dequeue_past_td(struct xhci_hcd *xhci,
 	struct xhci_ring *ep_ring;
 	struct xhci_command *cmd;
 	struct xhci_segment *new_seg;
+	struct xhci_segment *halted_seg = NULL;
 	union xhci_trb *new_deq;
 	int new_cycle;
+	union xhci_trb *halted_trb;
+	int index = 0;
 	dma_addr_t addr;
 	u64 hw_dequeue;
 	bool cycle_found = false;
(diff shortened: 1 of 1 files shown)


drivers/usb/host/xhci-pci.c
```diff
@@ -307,7 +307,6 @@ static void xhci_pci_quirks(struct device *dev, struct xhci_hcd *xhci)
 
 	if (pdev->vendor == PCI_VENDOR_ID_VIA && pdev->device == 0x3483) {
 		xhci->quirks |= XHCI_LPM_SUPPORT;
-		xhci->quirks |= XHCI_EP_CTX_BROKEN_DCS;
```
popcornmix: I'm confused. If this is a revert revert of `xhci: add quirk for host controllers that don't update endpoint DCS` then shouldn't it look similar to the original  `xhci: add quirk for host controllers that don't update endpoint DCS` commit?

Yet,  `xhci: add quirk for host controllers that don't update endpoint DCS` adds the line `xhci->quirks |= XHCI_EP_CTX_BROKEN_DCS` and this is removing it? (which sounds like a single revert?)

drivers/usb/host/xhci-ring.c
```diff
@@ -640,7 +643,27 @@ static int xhci_move_dequeue_past_td(struct xhci_hcd *xhci,
...
+	if (halted_seg) {
+		index = ((dma_addr_t)(hw_dequeue & ~0xf) - halted_seg->dma) /
+			 sizeof(*halted_trb);
+		halted_trb = &halted_seg->trbs[index];
+		new_cycle = halted_trb->generic.field[3] & 0x1;
+		xhci_dbg(xhci, "Endpoint DCS = %d TRB index = %d cycle = %d\n",
+			 (u8)(hw_dequeue & 0x1), index, new_cycle);
+	} else {
+		new_cycle = hw_dequeue & 0x1;
+	}
```
popcornmix, line 667: The changes in this file look how I'd expect.

drivers/usb/host/xhci-pci.c
pelwell: No, I'm the confused one. It appears that that line wasn't removed by the initial revert, even though it was added by the original commit - either a problem fixing up the revert or something strange with merge commits. It should be OK now.

popcornmix:
LGTM

PR #5644, opened 2023-10-11
media: i2c: ov9282: Read chip ID via 2 reads

6by9:
Vision Components have made an OV9281 module which blocks reading back the majority of registers to comply with NDAs, and in doing so doesn't allow auto-increment register reading as used when reading the chip ID.

Use two reads and manually combine the results.

We used to have an equivalent patch for this for the downstream ov9281 driver. Mainline view Vision Components' icky I2C behaviour as needing to be supported in either regmap or the v4l2_cci helpers, but doing so looks fairly non-trivial to me, and needs this driver converting to use either of those.

### DIFF
--- a/drivers/media/i2c/ov9282.c
+++ b/drivers/media/i2c/ov9282.c
@@ -1078,12 +1078,16 @@ error_unlock:
 static int ov9282_detect(struct ov9282 *ov9282)
 {
 	int ret;
-	u32 val;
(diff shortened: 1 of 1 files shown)


pelwell:
ov9282_read_reg feels like the right place for the workaround, but the end result is likely to be more complicated, so this is OK.

PR #5645, opened 2023-10-11
input: touchscreen: edt-ft5x06: Suppress bogus data on startup

njhollinghurst:
When polled without the use of IRQ, FT5x06 registers may hold undefined data. Filter out the initial static state (allowing a few transient zeroes which may occur around 700ms after reset).

### DIFF
--- a/drivers/input/touchscreen/edt-ft5x06.c
+++ b/drivers/input/touchscreen/edt-ft5x06.c
@@ -75,6 +75,8 @@
 #define EDT_DEFAULT_NUM_X		1024
 #define EDT_DEFAULT_NUM_Y		1024
 
+#define RESET_DELAY_MS			300	/* reset deassert to I2C */
+#define FIRST_POLL_DELAY_MS		300	/* in addition to the above */
 #define POLL_INTERVAL_MS		17	/* 17ms = 60fps */
 
 enum edt_pmode {
@@ -134,6 +136,7 @@ struct edt_ft5x06_ts_data {
 
 	char name[EDT_NAME_LEN];
 	char fw_version[EDT_NAME_LEN];
+	int init_td_status;
 
 	struct edt_reg_addr reg_addr;
 	enum edt_ver version;
@@ -268,12 +271,33 @@ static irqreturn_t edt_ft5x06_ts_isr(int irq, void *dev_id)
 		 * points.
 		 */
 		num_points = min(rdbuf[2] & 0xf, tsdata->max_support_points);
+
+		/* When polling FT5x06 without IRQ: initial register contents
+		 * could be stale or undefined; discard all readings until
+		 * TD_STATUS changes for the first time (or num_points is 0).
+		 */
+		if (tsdata->init_td_status) {
+			if (tsdata->init_td_status < 0)
+				tsdata->init_td_status = rdbuf[2];
+
+			if (num_points && rdbuf[2] == tsdata->init_td_status)
+				goto out;
+
+			tsdata->init_td_status = 0;
+		}
+
 		if (num_points) {
 			datalen = tplen * num_points + crclen;
 			cmd = offset;
 			error = edt_ft5x06_ts_readwrite(tsdata->client,
 							sizeof(cmd), &cmd,
 							datalen, &rdbuf[offset]);
+			if (error) {
+				dev_err_ratelimited(dev,
+						    "Unable to fetch data, error: %d\n",
+						    error);
+				goto out;
+			}
 		}
 	}
 
@@ -1294,7 +1318,7 @@ static int edt_ft5x06_ts_probe(struct i2c_client *client,
 	if (tsdata->reset_gpio) {
 		usleep_range(5000, 6000);
 		gpiod_set_value_cansleep(tsdata->reset_gpio, 0);
-		msleep(300);
+		msleep(RESET_DELAY_MS);
 	}
 
 	input = devm_input_allocate_device(&client->dev);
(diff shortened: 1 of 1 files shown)


njhollinghurst:
Not pretty, but it seems to do the trick. (We could avoid special handling for "transient zeroes" either by slowing down the reset by another 400ms or delaying the initial polling timer by 400ms.)

What I'm not doing right now is checking for duplicated IDs, because the bad state is not "guaranteed" to have any. But duplicated IDs are bad news generally as they can lead to input event spamming.

drivers/input/touchscreen/edt-ft5x06.c
```diff
@@ -1292,6 +1324,7 @@ static int edt_ft5x06_ts_probe(struct i2c_client *client,
 	}
 
 	if (tsdata->reset_gpio) {
+		gpiod_set_value_cansleep(tsdata->reset_gpio, 1);
```
6by9: Not necessary as you've already had
```
 	tsdata->reset_gpio = devm_gpiod_get_optional(&client->dev,
						     "reset", GPIOD_OUT_HIGH);
```
above

6by9 (commented):
Other than the one comment, that looks reasonable.

drivers/input/touchscreen/edt-ft5x06.c
njhollinghurst: Ah. Thanks. Will remove it.

njhollinghurst:
Now I think I still need to do some more testing for the "transient zeros" effect, to see whether or not we need to delay the first polling interval by 400ms or whatever (it's a bit shocking that reset seems to take so long).

If we do that, the code change could be a little simpler.

drivers/input/touchscreen/edt-ft5x06.c
```diff
@@ -274,7 +283,31 @@ static irqreturn_t edt_ft5x06_ts_isr(int irq, void *dev_id)
...
+				dev_err_ratelimited(dev,
+						    "Unable to fetch data, error: %d\n",
+						    error);
+				goto out;
+			}
+
+			/* When polling FT5x06 without IRQ: initial register
+			 * contents could be stale or undefined; ignore them
+			 * until something changes or num_points becomes 0.
+			 */
```
6by9: As per the comment, now that you're delaying and don't see the awkward drop to 0 points, can't we now just filter until num_points to hits 0?

njhollinghurst: EDITED.  I previously had a largely bogus objection about pre-existing touches.

However there is worse case, which is that it could change from the undefined data to the first real touch (which could occur after an arbitrarily long delay) without num_points _ever_ dropping to zero.

So I think the workaround really does need to do a `memcmp` to detect the first change.

njhollinghurst: Can probably simplify by detecting change to `num_points` only. The risk is that if the bogus data had exactly one point, we would miss the first touch.

drivers/input/touchscreen/edt-ft5x06.c
```diff
@@ -1384,11 +1409,12 @@ static int edt_ft5x06_ts_probe(struct i2c_client *client,
 			return error;
 		}
 	} else {
+		tsdata->init_td_status = UINT_MAX; /* filter bogus initial data */
 		INIT_WORK(&tsdata->work_i2c_poll,
 			  edt_ft5x06_ts_work_i2c_poll);
 		timer_setup(&tsdata->timer, edt_ft5x06_ts_irq_poll_timer, 0);
 		tsdata->timer.expires = jiffies +
-					msecs_to_jiffies(POLL_INTERVAL_MS);
+			msecs_to_jiffies(FIRST_POLL_DELAY_MS);
```
pelwell: The previous indentation looks better.

pelwell (commented):
I don't love the fact that this relies on both 0 and UINT_MAX being illegal TD_STATUS values, especially when there are no comments or named constants to make that clear, but if that is indeed the case then I can live with this.

drivers/input/touchscreen/edt-ft5x06.c
njhollinghurst: That was emacs' default for linux style - but I've just noticed that checkpatch allows both (surprisingly). I can increase it again.

pelwell: Either that or push jiffies onto the next line

njhollinghurst: Would a signed integer and -1 be a nicer way to represent "uninitialized"?  Actual values are u8, and u8->int casting is presumably uncontroversial.

To anticipate another potential question: we deliberately do include the "reserved" upper nybble in the test, because it does help to discriminate between bogus and real touch data.

pelwell: Since `init_td_status` is a signed `int`, and the `rdbuf` values are `u8`s, making the magic values negative is clearer and guaranteed to work.

pelwell:
With the latest change you are still relying on the first TD_STATUS value not being 0 - is that guaranteed to be the case?

njhollinghurst:
> With the latest change you are still relying on the first TD_STATUS value not being 0 - is that guaranteed to be the case?

Functionally, we don't rely on this: even if we chose some other bit pattern to represent the no-filtering state, a first value of `TD_STATUS==0` would imply `num_points==0` and would put it into that state.

0 is a benign value that we don't need to filter, and indeed is what we'd like the first TD_STATUS to have been, so it felt natural to use it for that. It's also the simplest value to test for (silly micro-optimization!)

pelwell:
> a first value of TD_STATUS==0 would imply num_points==0

Is a reviewer supposed to know that? Anyway, I don't object.

njhollinghurst:
Sorry... I am having second thoughts about the fix above It's possible to get into a state where the first touch is ignored. This can happen either if the initial data happened to have TD_STATUS == 0x01, which I have never observed; or quite reproducibly by the following steps:
- Hold one finger on the touchscreen
- Reboot
- Release finger as soon as the Plymouth splash appears (TS is held in reset at this point)
- Now the _next_ touch will be ignored.

To fix it I may need to revert the simplification which looked at TD_STATUS only, and go back to comparing the entire buffer.

The safer-but-more-complicated version is sitting on branch [ft5x06-filter2](/raspberrypi/linux/tree/ft5x06-filter2). I won't PR it for now.

PR #5650, opened 2023-10-15
MCP23017: allow specification of the i2c bus

janisstreib:
Analogous to `i2c-rtc-overlay.dts`

### DIFF
--- a/arch/arm/boot/dts/overlays/mcp23017-overlay.dts
+++ b/arch/arm/boot/dts/overlays/mcp23017-overlay.dts
@@ -7,7 +7,7 @@
 	compatible = "brcm,bcm2835";
 
 	fragment@0 {
-		target = <&i2c1>;
+		target = <&i2cbus>;
 		__overlay__ {
 			status = "okay";
 		};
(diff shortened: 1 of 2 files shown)


arch/arm/boot/dts/overlays/mcp23017-overlay.dts
```diff
@@ -57,13 +57,22 @@
 			microchip,irq-mirror;
 		};
 	};
+        frag100: fragment@100 {
+		target = <&i2c_arm>;
```
pelwell: i2c_arm isn't wrong, but let's keep it as `i2c1` anyway for regularity (and because `i2c_arm` only exists after the firmware has added it as an alias for `i2c1`).

pelwell (commented):
That looks good so far. There's one minor in-line comment, but otherwise I'd just like you to add the other i2c buses as well, so that it properly matches i2c-rtc.

janisstreib:
Thank you very much! I've used the wrong branch as template for the i2c block. I've amended the suggested changes analogous to this branch's I2C DTS.

janisstreib:
Sorry for the multiple amends afterwards; I needed to fix my formatting. I hope it's all good now :see_no_evil:

PR #5651, opened 2023-10-15
vc_mem: Add the DMA memcpy support from bcm2708_fb

pelwell:
bcm2708_fb is disabled by the vc4-kms-v3d overlay, which means that the DMA memcpy support it provides is not available to allow vclog to read the VC logs from the top 16MB on Pi 3. Add the code to the vc_mem driver, which will still be available.

It ought to be possible to do a proper DMA_MEM_TO_MEM copy via the generic DMA customer API, but that can be a later step.

### DIFF
--- a/drivers/char/broadcom/vc_mem.c
+++ b/drivers/char/broadcom/vc_mem.c
@@ -23,8 +23,20 @@
 #include <linux/uaccess.h>
 #include <linux/dma-mapping.h>
 #include <linux/broadcom/vc_mem.h>
+#include <linux/compat.h>
+#include <linux/platform_data/dma-bcm2708.h>
+#include <soc/bcm2835/raspberrypi-firmware.h>
 
 #define DRIVER_NAME  "vc-mem"
+
+/* N.B. These use a different magic value for compatibility with bmc7208_fb */
+#define VC_MEM_IOC_DMACOPY   _IOW('z', 0x22, struct vc_mem_dmacopy)
+#define VC_MEM_IOC_DMACOPY32 _IOW('z', 0x22, struct vc_mem_dmacopy32)
+
+/* address with no aliases */
+#define INTALIAS_NORMAL(x) ((x) & ~0xc0000000)
+/* cache coherent but non-allocating in L1 and L2 */
+#define INTALIAS_L1L2_NONALLOCATING(x) (((x) & ~0xc0000000) | 0x80000000)
 
 /* Device (/dev) related variables */
 static dev_t vc_mem_devnum;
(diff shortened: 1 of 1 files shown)


pelwell:
https://github.com/raspberrypi/utils/pull/36 is the corresponding vclog change.

pelwell:
@timg236 Updated to bail out on everything but a BCM2710.

drivers/char/broadcom/vc_mem.c
```diff
@@ -99,6 +139,186 @@ vc_mem_get_current_size(void)
...
+	u32 revision;
+	int rc;
+
+	if (vcdma->dev)
+		return 0;
+
+	fwnode = of_find_node_by_path("/system");
+	rc = of_property_read_u32(fwnode, "linux,revision", &revision);
+	if (((revision >> 12) & 0xf) != 2) {
+		/* Only BCM2710 has a memory hole to work around */
```
popcornmix: By hole, do you mean peripherals mapping low (below 1GB) .
Doesn't that include 2709 (pi2) and 2710 (pi3)?
Wasn't pi4 still mapping low by default?

pelwell: I think you're right about 2709 - I was thinking Pi 2 was still 512MB, but that's not the case. Pi 4 is OK because the firmware puts the logs in the first 1GB while the peripherals are mapped just below 4GB, so they don't obstruct the host's view of the VPU's RAM usage.

I'll confirm that Pi 2 is affected, then add it to the conditional.

drivers/char/broadcom/vc_mem.c
```diff
@@ -99,6 +139,189 @@ vc_mem_get_current_size(void)
...
+	u32 revision;
+	int rc;
+
+	if (vcdma->dev)
+		return 0;
+
+	fwnode = of_find_node_by_path("/system");
+	rc = of_property_read_u32(fwnode, "linux,revision", &revision);
+	revision = (revision >> 12) & 0xf;
+	if (revision != 1 && revision != 12) {
```
popcornmix: Did you mean 12 or 2?

pelwell: Indeed - more haste, less speed

timg236:
Looks good

popcornmix:
Tested on Pi2 (32-bit), Pi3 (64-bit) and Pi4 (64-bit).
Checked with strace that it behaved as desired.

Would be nice to get this using a higher level dma api, but this is fine as fix for now.

PR #5654, opened 2023-10-16
drm/vc4: Correct address offset for planes with src_[xy] offsets

6by9:
11cf37e741b4 switched to using drm_fb_dma_get_gem_addr instead of drm_fb_dma_get_gem_obj and adding fb->offset[].

However the tiled formats need to compute the offset in a more involved manner than drm_fb_dma_get_gem_addr applies, and we were ending up with the offset for src_[xy] being applied twice.

Switch back to using drm_fb_dma_get_gem_obj and fully computing the offsets ourselves.

Fixes: 11cf37e741b4 ("drm/vc4: Move the buffer offset out of the vc4_plane_state")

### DIFF
--- a/drivers/gpu/drm/vc4/vc4_plane.c
+++ b/drivers/gpu/drm/vc4/vc4_plane.c
@@ -1447,9 +1447,9 @@ static int vc4_plane_mode_set(struct drm_plane *plane,
 	vc4_state->ptr0_offset[0] = vc4_state->dlist_count;
 
 	for (i = 0; i < num_planes; i++) {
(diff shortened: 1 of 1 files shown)


popcornmix:
Tested with dual monitor and X. Without the PR I see same desktop on both screens. With the PR I see separate desktops.
Kodi still seems functional a number of formats.

LGTM

6by9:
AFAIK Kodi doesn't use source rectangles (only destination rects for letterboxing), so wouldn't be affected anyway.

PR #5657, opened 2023-10-17
drivers: media: rp1_cfe: Fix link validate test for pixel format

naushir:
Now that we have removed unique PISP media bus codes, the cfe format table has multiple entries with the same media bus code for 16-bit formats. The test in cfe_video_link_validate() did not account for this. Fix it by testing the media bus code and the V4L2 pixelformat 4cc together.

As a drive-by, ensure we have a valid CSI2 datatype id when programming the hardware block.


naushir:
@davidplowman this ought to fix our issue.

davidplowman:
LGTM!

kbingham:
Thanks - this fixes the issue I had!

Issue #5470, opened 2023-05-16
USB 2.0 driver fails with kernel update from 6.1.27 to  6.1.28-3-rpi-ARCH 


baslking:
### Describe the bug

USB 2.0 generates kernel trace at boot and then does not correctly detect or use USB connected devices.

expected behavior of 
~lsusb
Bus 001 Device 003: ID 0463:ffff MGE UPS Systems UPS
Bus 001 Device 002: ID 0424:2514 Microchip Technology, Inc. (formerly SMSC) USB 2.0 Hub
Bus 001 Device 001: ID 1d6b:0002 Linux Foundation 2.0 root hub
Kernel 28 behavior:
~lsusb
Bus 001 Device 001: ID 1d6b:0002 Linux Foundation 2.0 root hub

### Steps to reproduce the behaviour

Updated kernel no longer detects USB devices boot trace shown below.
Reverting to 6.1.27 kernel resolves the issue 

### Device (s)

Raspberry Pi CM4 8GB

### System

NAME="Arch Linux ARM"
PRETTY_NAME="Arch Linux ARM"
ID=archarm
ID_LIKE=arch
BUILD_ID=rolling
ANSI_COLOR="38;2;23;147;209"
HOME_URL="https://archlinuxarm.org/"
DOCUMENTATION_URL="https://archlinuxarm.org/wiki"
SUPPORT_URL="https://archlinuxarm.org/forum"
BUG_REPORT_URL="https://github.com/archlinuxarm/PKGBUILDs/issues"

pacman -Qs raspberry
local/firmware-raspberrypi 20230125-1
    Additional firmware for Raspberry Pi
local/raspberrypi-bootloader 20230512-1
    Bootloader files for Raspberry Pi

This kernel works:
Linux pi-nas 6.1.27-1-rpi-ARCH #1 SMP PREEMPT Tue May  2 18:32:20 MDT 2023 aarch64 GNU/Linux
This kernel does not:
Linux pi-nas 6.1.28-3-rpi-ARCH #1 SMP PREEMPT Mon May 15 12:00:19 MDT 2023 aarch64 GNU/Linux

### Logs

journalctl -b snippet:
------------[ cut here ]------------
May 16 11:09:01 pi-nas kernel: bcm2835-isp bcm2835-isp: Register capture node 1 with media controller
[...]

PR #5658, opened 2023-10-17
ARM: dts: bcm2712-rpi: Set a 1GB ZONE_DMA limit

pelwell:
The arm64 initialisation uses the physical address reachable by all DMA controllers to set the size of ZONE_DMA. This fails on BCM2712 with the current dts files because the declaration of the I/O space fools it into thinking the legacy 30-bit DMA channels can see most of the 32-bit space.

Take advantage of the simple nature of the implementation by adding a node with a restricted dma-ranges property solely so as to act as the limiting factor in the calculation.

See: https://github.com/raspberrypi/linux/issues/5470

### DIFF
--- a/arch/arm/boot/dts/bcm2712-rpi.dtsi
+++ b/arch/arm/boot/dts/bcm2712-rpi.dtsi
@@ -64,6 +64,16 @@
 };
 
 / {
+	/*
+	 * Add a node with a dma-ranges value that exists only to be found
+	 * by of_dma_get_max_cpu_address, and hence limit the DMA zone.
+	 */
+	zone_dma {
+		#address-cells = <1>;
+		#size-cells = <1>;
+		dma-ranges = <0x0  0x0 0x0  0x40000000>;
+	};
+
 	__overrides__ {
 		arm_freq;
 	};


pelwell:
I'm slightly concerned that this may be unnecessarily restrictive for Pi 5 most of the time - is the old DWC_OTG block the only block still (sometimes) in use that can only address the first 1GB? - but let's give this some wider testing.

popcornmix:
I can't think of anything else limited to 1GB (apart from VC).
In fact pretty much everything can access the full (40-bits-ish) address bus on Pi5.

pelwell:
Are you OK with this being merged, or would you rather tie it to dwc_otg use in some way?

popcornmix:
What exactly does this change affect?
I assume some buffers were more freely allocated before this, and are now limited to bottom 1GB?
Does this affect cma? dmabufs? kmalloc?

pelwell:
DMA-coherent buffers will have to be in the bottom 1GB.
CMA is going to contend for the same space, exactly as we've seen on Pi 4.
In 64-bit mode the kernel has a large reach so although the kmalloc heap
may move it should make no difference.

pelwell:
This is likely to be superseded by #5661.

PR #5660, opened 2023-10-18
Forcing /dev/fbN assignments for DRM cards

6by9:
Following on from #5629 for DRM device naming, add the option of forcing the mapping of /dev/fbN to particular DRM cards.

In a similar way to I2C etc, it looks for the highest numbered alias, and unassigned devices will be put beyond that number on the normal first-come, first-served basis.

If you assign something to fb1 but nothing to fb0, then you won't get a console unless you also use ```fbcon=map=<1>``` in the kernel command line.

Aliases only created for Pi5 as that's the one we hit this with more with DSI (x2), DPI, and VEC all being separate cards, but we could add it for the SPI display overlays and Pi0-4 as well. The SPI overlays are currently largely using fbdev drivers rather than DRM ones. I am working on a new tinydrm overlay to use mimic the fbtft overlay but using the TinyDRM drivers, but it's only got 2 drivers working at present.
(There is also panel-mipi-dbi which takes a firmware file to configure an SPI display, but AFAIK there is no repository yet of those firmware files to make it useful).


drivers/video/fbdev/core/fbmem.c
```diff
@@ -52,6 +52,7 @@ static DEFINE_MUTEX(registration_lock);
 
 struct fb_info *registered_fb[FB_MAX] __read_mostly;
 int num_registered_fb __read_mostly;
+int min_dynamic_fb;
```
pelwell: This code is paying heavy lip service to optimisation - does `min_dynamic_fb` deserve a `__read_mostly` as well?

drivers/gpu/drm/drm_fb_helper.c
```diff
@@ -1967,6 +1967,15 @@ __drm_fb_helper_initial_config_and_unlock(struct drm_fb_helper *fb_helper,
 	 * register the fbdev emulation instance in kernel_fb_helper_list. */
 	mutex_unlock(&fb_helper->lock);
 
+	id = of_alias_get_highest_id("drm_fb");
```
pelwell, line 1970: It's a nit pick, but the comment says `drm-fb`.

drivers/video/fbdev/core/fbmem.c
6by9: Done

drivers/gpu/drm/drm_fb_helper.c
6by9: Commit text updated to also say ```drm_fb```

pelwell:
Looks good to me - nice use of the new "assign a literal path string" feature.

Feel free to merge when the auto-builds have completed.

hailfinger:
Is there a variant of this for Pi 4B which I could test? Sometimes my tinydrm SPI display shows up as /dev/fb0, sometimes as /dev/fb1. HDMI is attached as well, so I get the bootup log randomly on HDMI or SPI display.

6by9:
> Is there a variant of this for Pi 4B which I could test? Sometimes my tinydrm SPI display shows up as /dev/fb0, sometimes as /dev/fb1. HDMI is attached as well, so I get the bootup log randomly on HDMI or SPI display.

That's exactly the situation we have on Pi5 with DSI, DPI, or the VEC being separate cards.
NB This is only for DRM drivers, not fbdev.

I hadn't added it for Pi0-4 as it's only going to be SPI type displays that have the potential to cause grief, and the main niggle with this override is that if the card assigned fb0 doesn't load correctly, you end up with no console.

Can I ask which SPI display you're using?
I've started creating a tinydrm overlay to replace fbtft (https://github.com/6by9/linux/blob/rpi-6.1.y-tinydrm/arch/arm/boot/dts/overlays/tinydrm-overlay.dts).
piscreen, pitft28-resistive, pitft35-resistive, watterott-display all have a drm override to select the TinyDRM driver instead of fbdev, but add in touch controllers too.
adafruit-st7735r is TinyDRM only and should probably be moved into tinydrm only.

Once the SPI display overlays are tidied up some more, I'll add an override to them, and for Pi0-4. The main issue I have is with regard testing as I only have about 5 SPI displays (piscreen, pitft35-resistive, the 2 in tinydrm-overlay, and I think I have one other in my parts drawer at home).

hailfinger:
> > Is there a variant of this for Pi 4B which I could test? Sometimes my tinydrm SPI display shows up as /dev/fb0, sometimes as /dev/fb1. HDMI is attached as well, so I get the bootup log randomly on HDMI or SPI display.
> 
> That's exactly the situation we have on Pi5 with DSI, DPI, or the VEC being separate cards. NB This is only for DRM drivers, not fbdev.

Understood. I'm using tinydrm.


> I hadn't added it for Pi0-4 as it's only going to be SPI type displays that have the potential to cause grief, and the main niggle with this override is that if the card assigned fb0 doesn't load correctly, you end up with no console.
> 
> Can I ask which SPI display you're using? I've started creating a tinydrm overlay to replace fbtft (https://github.com/6by9/linux/blob/rpi-6.1.y-tinydrm/arch/arm/boot/dts/overlays/tinydrm-overlay.dts). piscreen, pitft28-resistive, pitft35-resistive, watterott-display all have a drm override to select the TinyDRM driver instead of fbdev, but add in touch controllers too. adafruit-st7735r is TinyDRM only and should probably be moved into tinydrm only.

Hardware: Waveshare 2.0 inch 240x320 display with ST7789V controller.
I'm using the mipi-dbi-spi overlay and it works reasonably well (needs to run kmsprint if the display wasn't activated otherwise by Xorg or something). See https://forums.raspberrypi.com/viewtopic.php?t=337019 for details including an init sequence for generating the firmware file.

> Once the SPI display overlays are tidied up some more, I'll add an override to them, and for Pi0-4. The main issue I have is with regard testing as I only have about 5 SPI displays (piscreen, pitft35-resistive, the 2 in tinydrm-overlay, and I think I have one other in my parts drawer at home).

I have a bunch of other SPI display models here, but the ST7789V display is the one attached right now, so any immediate tests will target that one. Willing to test other displays later this week.

6by9:
> Hardware: Waveshare 2.0 inch 240x320 display with ST7789V controller. I'm using the mipi-dbi-spi overlay and it works reasonably well (needs to run kmsprint if the display wasn't activated otherwise by Xorg or something). See https://forums.raspberrypi.com/viewtopic.php?t=337019 for details including an init sequence for generating the firmware file.

This is where I'm a little lost over TinyDRM display specific drivers vs mipi-dbi-spi firmware files.
AFAIK there is no repository of firmware files for those not covered. For a general user this makes mipi-dbi-spi too big a step to be accessible. At least the dedicated drivers can be instantiated from overlays with no further involvement.

Also note that currently mipi-dbi-spi is not registered in Mesa as a kmsro driver, so Wayfire currently barfs on it due to falling back to software rendering. It needs the equivalent of https://gitlab.freedesktop.org/mesa/mesa/-/commit/8cfc17bdda31a381bfbaadc75f0d34dada0e8c91 (as does ili9486) - unless someone beats me to it, I'll be making a PR to Mesa with the relevant change soon.

> I have a bunch of other SPI display models here, but the ST7789V display is the one attached right now, so any immediate tests will target that one. Willing to test other displays later this week.

I may take you up on that. The main niggle with the panel specific drivers is noting that many of them have the opposite sense for the reset line, so it's not necessarily a trivial conversion of the overlay.

Just for you - #5675.
dtparam=drm_fb0_vc4 should lock /dev/fb0 to the vc4 HDMI/DSI/DPI display on Pi0-4.

PR #5661, opened 2023-10-19
overlays: Add Pi 5 version of dwc2

pelwell:
Most bus masters on Pi 5 have access to at least the lower 4GB, but the old DWC OTG controller is still restricted to the first 1GB. Create a Pi 5-specific version of the overlay including a fragment that ensures the swiotlb bounce buffers are allocated in that area.

This PR replaces https://github.com/raspberrypi/linux/pull/5658.

### DIFF
--- a/arch/arm/boot/dts/overlays/Makefile
+++ b/arch/arm/boot/dts/overlays/Makefile
@@ -59,6 +59,7 @@ dtbo-$(CONFIG_ARCH_BCM2835) += \
 	draws.dtbo \
 	dwc-otg.dtbo \
 	dwc2.dtbo \
+	dwc2-pi5.dtbo \
 	edt-ft5406.dtbo \
 	enc28j60.dtbo \
 	enc28j60-spi2.dtbo \
--- a/arch/arm/boot/dts/overlays/README
+++ b/arch/arm/boot/dts/overlays/README
@@ -1107,6 +1107,10 @@ Params: dr_mode                 Dual role mode: "host", "peripheral" or "otg"
                                 mode
 
 
+Name:   dwc2-pi5
+Info:   See dwc2 (this is the Pi 5 version)
+
+
 [ The ds1307-rtc overlay has been deleted. See i2c-rtc. ]
 
 
--- a/arch/arm/boot/dts/overlays/dwc2-pi5-overlay.dts
+++ b/arch/arm/boot/dts/overlays/dwc2-pi5-overlay.dts
@@ -1,0 +1,18 @@
+#include "dwc2-overlay.dts"
+
+/{
+	fragment@1 {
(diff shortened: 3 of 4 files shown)


pelwell:
@6by9 Can you confirm that this is a suitable alternative to #5658 (if we're nervous about placing unnecessary restrictions on RAM layout for all Pi 5s)?

timg236:
This seems like a good idea

6by9:
Should work - I'll double check in a mo.

Is it cleaner to use an include instead of being 50% duplication of dwc2-overlay.dts?

popcornmix:
Yes, I prefer this solution.

6by9:
It works to the same level that I had the old "always apply the range" change did.

pelwell:
> Is it cleaner to use an include instead of being 50% duplication of dwc2-overlay.dts?

Much cleaner - done.

6by9:
Having said this worked, it looks like my system hadn't updated properly :-(

#5658 put a dma-ranges in /, whereas this puts it in /soc.
/soc already has a dma-ranges, and it doesn't get overwritten by the overlay.

pelwell:
>  it looks like my system hadn't updated properly :-(

Oh well.

> /soc already has a dma-ranges, and it doesn't get overwritten by the overlay.

I would have expected that to be okay, but the dma-ranges above the zone_dma's dma-ranges must be interfering. I'll fix it.

pelwell:
Fixup pushed. Success looks like this:
```
$ pi@raspberrypi:~$ dmesg | grep -A3 "Zone ranges"
[    0.000000] Zone ranges:
[    0.000000]   DMA      [mem 0x0000000000000000-0x000000003fffffff]
[    0.000000]   DMA32    [mem 0x0000000040000000-0x00000000ffffffff]
[    0.000000]   Normal   [mem 0x0000000100000000-0x00000001ffffffff]
```

6by9:
Previous patch
```
pi@bookworm64-rel:~ $ dmesg | grep -A3 "Zone ranges"
[    0.000000] Zone ranges:
[    0.000000]   DMA      [mem 0x0000000000000000-0x00000000ffffffff]
[    0.000000]   DMA32    empty
[    0.000000]   Normal   [mem 0x0000000100000000-0x00000001ffffffff]
```
New
```
pi@bookworm64-rel:~ $ dmesg | grep -A3 "Zone ranges"
[    0.000000] Zone ranges:
[    0.000000]   DMA      [mem 0x0000000000000000-0x000000003fffffff]
[    0.000000]   DMA32    [mem 0x0000000040000000-0x00000000ffffffff]
[    0.000000]   Normal   [mem 0x0000000100000000-0x00000001ffffffff]
```
:-)

PR #5662, opened 2023-10-19
Swap DT aliases from underscore to hyphen.

6by9:
Fixes the warning reported by dtoverlaycheck.

(That warning probably wants to be promoted to an error).


pelwell:
Thanks.

PR #5663, opened 2023-10-19
Typo in overlays README

lurch:
touchscreen-size-y for rpi-ft5406 defaults to 480, not 600

I _assume_ this is a typo, as https://github.com/raspberrypi/linux/blob/rpi-6.1.y/arch/arm/boot/dts/overlays/edt-ft5406.dtsi and https://github.com/raspberrypi/linux/blob/rpi-6.1.y/arch/arm/boot/dts/overlays/rpi-ft5406-overlay.dts both say the touchscreen is 800x480, and that's also the pixel-resolution of the [DSI screen](https://www.raspberrypi.com/products/raspberry-pi-touch-display/).


6by9:
Correct that the default is 480, not 600.

PR #5666, opened 2023-10-20
configs: stop defaulting to the powersave governor

clopez:
* The default configs for the RPi shouldn't be defaulting to the powersave governor because it reduces the performance by setting the frequency to the minimum.

* This is an issue even if user-space changes the governor later because it is slowing down the boot time (the kernel will boot with the minimum CPU frequency until user-space fixes it)

* It is also a major issue for another projects than RaspberryPiOS that rely on this kernel and use this configs as base or default. For example: I have been bitten by this issue when using Yocto with meta-raspberrypi because their kernel config is based on this ones, so it will boot with powersave by default and there user space is not configured to change the governor. So it will remain with powersave always :(

* Just don't set any default governor and let the Linux's defaults apply: which means selecting governor `schedutil` in this kernel version, but it was governor `ondemand` on older versions and on the future it can be a newer scheduler. See the logic to select the default at the file `drivers/cpufreq/Kconfig`.

* Enable also the config option to ensure that the `powersave` governor is built and available if user space wants to use it.

* For config bcmrpi_defconfig enable also the `schedutil` governor, which is the only config missing to enable this one.


timg236:
That's a pretty major change to push out to everyone and would break the press SHIFT to disable the performance governor behaviour (initd raspi-config)

pelwell:
We default to powersave to allow boards to be booted successfully when higher clocks speeds are causing problems, e.g. with an inadequate power supply. Trying to reduce the clock with the Shift key having started up running faster may already be too late.
 
You can always override the default with a module parameter (which sets the governor as early as the kernel's compiled-in default) - add the following to cmdline.txt:
```
cpufreq.default_governor=ondemand
```

clopez:
This is a feature I wasn't aware of so I checked how it works by flashing RaspberryOS and doing a few tests:

This is what I discovered how it works:

 * The kernel boots with the `powersave` governor and switchs to governor `ondemand` when user-space starts.
 * If you hold down shift during boot then the `powersave` governor will remain.


Now I have some comments about this feature:

 1) Starting the system with the `powersave` governor slows-down the boot time.
    This feature should work the other way around: Start with the `ondemand` governor
    by default and switch to `powersave` if requested by the user

  2) It is not clear at all for how much time or exactly when you have to press-down
     the shift key as there is no visual indication at all that the governor switched
     So you end pressing the shift key until boot ends and you hope that it worked,
     because you have no way of knowing if it worked other than checking the current
     governor manually (or maybe you can infer it because everything is much slower).

  3) You can enable this by mistake by pressing shift during boot without the intention
     of slowing down your RPi and you won't notice because there is no visual indication.
     I bet this happened more than once and people were asking themselves why the RPi was so slow.

  4) The [documentation about this is really miss-leading](https://www.raspberrypi.com/documentation/computers/configuration.html#overclock). It literally says this:

> Be aware that overclocking may reduce the lifetime of your Raspberry Pi.
> If overclocking at a certain level causes system instability, try a more modest overclock.
> Hold down the Shift key during boot to temporarily disable overclocking.

Switching to `powersave` governor is *not* disabling overcloking, but enabling *underclocking*



I think this feature should be reworked to boot with the default governor (which nowadays is `schedutil` and not `ondemand`)
and then switch to `powersave` if requested by the user by pressing the shift key.

I also think the documentation should be fixed (the user is really underclocking the RPi, not disabling overclocking)

And it would be nice to have some visual indication about the status of the governor, maybe some widget around the network/volume at the top.

Also, please not ignore that this kernel and their configs is used by other distros that may not be aware of this  settting of powersave as the default governor.  See for example:  https://github.com/agherzan/meta-raspberrypi/pull/1231

IMHO the default governor should be policed in the RaspberryOS repository and not in this kernel repo.

pelwell:
Noted.

I'd be surprised if there were distros that weren't aware of our policy.

We're not changing it.

clopez:
For your surprise I can tell you that not only the default Yocto builds are affected by this but also the Buildroot ones.
I just submitted a patch for them to workaround the issue here: https://lists.buildroot.org/pipermail/buildroot/2023-October/677641.html

I understand that RaspberryPiOS wants to boot the boards with the `powersave` CPU governor as default. But I think the defconfigs on the kernel repository is not the place to set that because other distros depend on this defconfigs.

This should be a specific configuration on the RaspberryPiOS repository for the kernel cmdline (`/boot/cmdline.txt`). You can change the default by setting: `cpufreq.default_governor=powersave` there. That seems a more appropriate place to set this than on the kernel defconfig.

PR #5672, opened 2023-10-23
drivers: media: imx477: Disable the scaler

naushir:
The horizontal scaler was enabled for the 2028x1520 and 2028x1080 modes, with a scale factor of 1. It caused a single column of bad pixels on the right edge of the image. Since scaling is not needed for these modes, disable it entirely.


naushir:
@davidplowman

davidplowman:
I'm convinced. Bit weird how we never noticed...!

naushir:
I think it's partially to do with these being video modes and we don't pixel peek at them often.

PR #5679, opened 2023-10-24
drm/vc4: crtc: Support odd horizontal timings on BCM2712

popcornmix:
BCM2711 runs pixelvalve at two pixels per clock cycle which results in an unfortunate limitation that odd horizontal timings are not possible. This is apparent on the standard DMT mode of 1366x768@60 which cannot be driven with correct timing.

BCM2712 defaults to the same behaviour, but has a mode to support odd timings. While internally it still runs at two pixels per clock, setting the PV_VCONTROL_ODD_TIMING bit makes it appear externally to behave as it is one pixel per clock.

Switching to this mode fixes 1366x768@60 mode, and other custom resultions with odd horizontal timings.


pelwell:
It looks like the patch sets the bit whenever it is available. Are there no downsides?

popcornmix:
> It looks like the patch sets the bit whenever it is available. Are there no downsides?

I don't believe so. It was just added as opt-in in case it breaks something.
From a long ago email:
> Odd values for (at least some of) the horizontal timings are now supported by adding an upper-half-valid signal to the 2 pixels/clock interface into HDMI -- there's even been some testing of it, as it was tweaked a couple of weeks ago.

It shouldn't harm performance (we still run two pixels per clock). There is some chance of a bug, but 1366x768@60 with odd timings works:
```
pi@pi5:~ $ kmsprint
Connector 0 (32) HDMI-A-1 (connected)
  Encoder 0 (31) TMDS
    Crtc 2 (88) 1366x768@60.02 85.885 1366/73/144/217/- 768/1/3/23/+ 60 (60.02) U 
      Plane 2 (78) fb-id: 300 (crtcs: 2) 0,0 1366x768 -> 0,0 1366x768 (XR24 AR24 AB24 XB24 RG16 BG16 AR15 XR15 RG24 BG24 YU16 YV16 YU24 YV24 YU12 YV12 NV12 NV21 NV16 NV61 P030 XR30 AR30 AB30 XB30 RGB8 BGR8 XR12 AR12 XB12 AB12 BX12 BA12 RX12 RA12)
        FB 300 1366x768 XR24
      Plane 22 (280) fb-id: 312 (crtcs: 2) 0,0 64x64 -> 1363,716 64x64 (XR24 AR24 AB24 XB24 RG16 BG16 AR15 XR15 RG24 BG24 YU16 YV16 YU24 YV24 YU12 YV12 NV12 NV21 NV16 NV61 P030 XR30 AR30 AB30 XB30 RGB8 BGR8 XR12 AR12 XB12 AB12 BX12 BA12 RX12 RA12)
        FB 312 64x64 AR24
```
and dual 4kp60 still works. I guess we won't know about any subtle failures without getting more people testing it.

Maybe leave it a couple of days for the stable kernel to get out before merging. @6by9 may want to try it.

6by9:
Works on my 1366x768 monitor which has a slightly different timing from popcornmix's
```
Connector 1 (42) HDMI-A-2 (connected)
  Encoder 1 (41) TMDS
    Crtc 3 (99) 1366x768@59.79 85.500 1366/70/143/213/+ 768/3/3/24/+ 60 (59.79) P|D 
      Plane 3 (89) fb-id: 302 (crtcs: 3) 0,0 1366x768 -> 0,0 1366x768 (XR24 AR24 AB24 XB24 RG16 BG16 AR15 XR15 RG24 BG24 YU16 YV16 YU24 YV24 YU12 YV12 NV12 NV21 NV16 NV61 P030 XR30 AR30 AB30 XB30 RGB8 BGR8 XR12 AR12 XB12 AB12 BX12 BA12 RX12 RA12)
        FB 302 1920x1080 RG16
```

We could drop the comment at https://github.com/raspberrypi/linux/blob/rpi-6.1.y/drivers/gpu/drm/vc4/vc4_crtc.c#L401-L406 as it has no real benefit now that the bit is set. Not a requirement though.

popcornmix:
Dropped outdated comment.

PR #5684, opened 2023-10-26
KMS stale dlist alloc changes

6by9:
#5674 "ERROR Failed to allocate DLIST entry: -28"

I haven't reproduced any failures, but have noted that we were getting stale allocs stuck for some time due to issues in the logic that freed them.

I have some reservations about actually disabling the HVS channel. It could be worked around by keeping an `enabled` state variable, but it feels nicer to actually clear the hardware. It needs a bit more testing to ensure we don't get the stuck pixel issue (not observed so far).

Currently not tested on a Pi5, hence draft PR.


popcornmix:
I've been running this on kodi system for a couple of days without issue (but I wasn't getting the dlist error before).
Code changes look okay to me.

6by9:
Rebased and updated.
I've added one extra patch that logs the size of the allocation that failed, just to give more debug should there be further problems.

PR #5685, opened 2023-10-27
drivers: media: imx477: Set horizontal binning when disabling the scaler

davidplowman:
The horizontal scaler has been disabled but actually the sensor is not binning horizontally, resulting in images that are stretched 2x horizontally (missing the right half of the field of view completely).

Therefore we must additionally set the horizontal binning mode. There is only marginal change in output quality and noise levels.

Fixes: f075893e9b0e ("drivers: media: imx477: Disable the scaler")


davidplowman:
@naushir You probably want to look and see if you agree with me on this. As far as I can tell, these modes were scaling horizontally and not binning at all! But this does seem to fix the headline problem for me.

I think it might be very slightly more alias-y, it's hard to be sure. But not offensively so. I didn't find a way to go back to the scaler and not mess up the final pixel column.

naushir:
Hmmm, makes sense, but I was sure I looked at the images with the scaling disabled and they seemed ok.  How did I miss this?

davidplowman:
Yes, I really don't understand. Maybe an independent check that (a) the problem exists and (b) this fixes it would be advisable...

naushir:
I'll run this first thing on Monday.

naushir:
Yup this is definitely an issue! I went back to the old captures I took to show the original fix worked, and they are all stretched as well.  It was hardly noticeable as i was pointing it to the flat board without many objects in the frame!  Also confirmed that this fixes it, so going to merge it now.

PR #5690, opened 2023-10-31
fixup! arch/arm64: Add Revision, Serial, Model to cpuinfo

pelwell:
Delete the Hardware string, which is pointless and misleading.

See: https://github.com/raspberrypi/bookworm-feedback/issues/129


popcornmix:
LGTM

PR #5691, opened 2023-11-01
drivers: media: cfe: Set the CSI-2 link frequency correctly

naushir:
Use the sensor provided link frequency to set the DPHY timing parameters on stream_on. This replaces the hard-coded 999 MHz value currently being used. As a fallback, revert to the original 999 Mhz link frequency.

As a drive-by, fix a 80-character line formatting error.

### DIFF
--- a/drivers/media/platform/raspberrypi/rp1_cfe/cfe.c
+++ b/drivers/media/platform/raspberrypi/rp1_cfe/cfe.c
@@ -780,7 +780,8 @@ static void cfe_start_channel(struct cfe_node *node)
 			__func__, node_desc[FE_OUT0].name,
 			cfe->fe_csi2_channel);
 
-		source_fmt = v4l2_subdev_get_pad_format(&cfe->csi2.sd, state, cfe->fe_csi2_channel);
+		source_fmt = v4l2_subdev_get_pad_format(&cfe->csi2.sd, state,
+							cfe->fe_csi2_channel);
 		fmt = find_format_by_code(source_fmt->code);
 
 		width = source_fmt->width;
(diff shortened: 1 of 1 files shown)


drivers/media/platform/raspberrypi/rp1_cfe/cfe.c
```diff
@@ -982,6 +983,35 @@ static void cfe_buffer_queue(struct vb2_buffer *vb)
...
+	s64 link_freq;
+
+	state = v4l2_subdev_lock_and_get_active_state(&cfe->csi2.sd);
+	source_fmt = v4l2_subdev_get_pad_format(&cfe->csi2.sd, state,
+						cfe->fe_csi2_channel);
+	fmt = find_format_by_code(source_fmt->code);
+	link_freq = v4l2_get_link_freq(cfe->sensor->ctrl_handler,
+				       fmt->depth,
+				       cfe->csi2.active_data_lanes * 2);
+	if (link_freq < 0) {
```
njhollinghurst: Do we want any bounds checks here? ISTR the legal range is 80Mbps to 1.5Gbps. But perhaps that's too many magic numbers.

njhollinghurst: (er, halved for the *2 or whatever)

naushir: There are further bounds checking when we go to program the DPHY frequency from ``csi2_open_rx()``.

njhollinghurst:
LGTM

6by9:
This is only checking the adjacent node for a link frequency control. If the adjacent node is a mux or bridge, then it may not have a link frequency. 

I'm just comparing to qcom camss which follows the full chain until it finds a MEDIA_ENT_F_CAM_SENSOR, but actually even that isn't perfect as a parallel to CSI2 bridge, or FPD-LinkIII bridge would mangle the link frequency required.
CIO2, IMX8, and IMX7 all appear to do the simple thing of just looking at the adjacent node.

I think the ideal is to follow the chain until you find a node which has a LINK_FREQUENCY control and use that. Both adjacent and end have issues.

naushir:
> I think the ideal is to follow the chain until you find a node which has a LINK_FREQUENCY control and use that. Both adjacent and end have issues.

Hmmm, more annoyances.  Hopefully the new rev does this right...?

drivers/media/platform/raspberrypi/rp1_cfe/cfe.c
```diff
@@ -982,6 +983,63 @@ static void cfe_buffer_queue(struct vb2_buffer *vb)
...
+	struct v4l2_subdev *subdev;
+	const struct cfe_fmt *fmt;
+	struct media_entity *entity;
+	struct media_pad *pad;
+	s64 link_freq;
+
+	state = v4l2_subdev_lock_and_get_active_state(&cfe->csi2.sd);
+	source_fmt = v4l2_subdev_get_pad_format(&cfe->csi2.sd, state,
+						cfe->fe_csi2_channel);
+	fmt = find_format_by_code(source_fmt->code);
```
naushir, line 998: This is probably wrong?  I think I should do this after finding the entity/subdev through the loop below.

njhollinghurst: Maybe. I suppose you want the nearest thing that truthfully declares either a link frequency, or a (pixel format, pixel rate) pair, which does imply putting it inside the loop.

naushir: Ok, I've done this now, but can revert if we think it's the wrong thing to do in the future.

naushir: Actually, this is wrong.  We *want* to be looking at the csi entity pad for the format, that's what's connected to the HW after all.  I will revert my update back to the above code.

naushir:
I think this now looks to be the correct behavior for those pesky bridge/link entities!

naushir:
If no objections, I'll merge this in later today.

PR #5701, opened 2023-11-08
drivers: media: imx708: Tweak broken line correction parameter

njhollinghurst:
In QBC full-resolution mode, reduce QBC_LPF_INTENSITY parameter to reduce the occurrence of re-mosaic artefacts in colourful scenes, at the slight expense of black-and-white ones.

See IMX708_IQ_Tuning_Operation_Manual_v1.0.0 section 4.

Higher values reduce broken-line artifacts in black-and-white gratings with perfect white balance gains, but can introduce "ladder" artefacts in coloured gratings and textures. It is a trade-off, but existing test images and user feedback suggest the latter are more important. Reducing the register value reduces the frequency (but not the severity) of these artifacts in many cases.

### DIFF
--- a/drivers/media/i2c/imx708.c
+++ b/drivers/media/i2c/imx708.c
@@ -20,6 +20,14 @@
 #include <media/v4l2-fwnode.h>
 #include <media/v4l2-mediabus.h>
 
+/*
+ * Parameter to adjust Quad Bayer re-mosaic broken line correction
+ * strength, used in full-resolution mode only. Set zero to disable.
+ */
+static int qbc_adjust = 2;
+module_param(qbc_adjust, int, 0644);
+MODULE_PARM_DESC(qbc_adjust, "Quad Bayer broken line correction strength [0,2-5]");
+
 #define IMX708_REG_VALUE_08BIT		1
 #define IMX708_REG_VALUE_16BIT		2
 
@@ -99,10 +107,16 @@
 
 /* HDR exposure ratio (long:med == med:short) */
 #define IMX708_HDR_EXPOSURE_RATIO       4
-#define IMX708_REG_MID_EXPOSURE	0x3116
-#define IMX708_REG_SHT_EXPOSURE	0x0224
+#define IMX708_REG_MID_EXPOSURE		0x3116
+#define IMX708_REG_SHT_EXPOSURE		0x0224
 #define IMX708_REG_MID_ANALOG_GAIN	0x3118
 #define IMX708_REG_SHT_ANALOG_GAIN	0x0216
+
+/* QBC Re-mosaic broken line correction registers */
+#define IMX708_LPF_INTENSITY_EN		0xC428
+#define IMX708_LPF_INTENSITY_ENABLED	0x00
+#define IMX708_LPF_INTENSITY_DISABLED	0x01
+#define IMX708_LPF_INTENSITY		0xC429
 
 /*
  * Metadata buffer holds a variety of data, all sent with the same VC/DT (0x12).
@@ -171,6 +185,9 @@ struct imx708_mode {
 
 	/* HDR flag, used for checking if the current mode is HDR */
 	bool hdr;
+
+	/* Quad Bayer Re-mosaic flag */
+	bool remosaic;
 };
 
 /* Default PDAF pixel correction gains */
@@ -363,8 +380,6 @@ static const struct imx708_reg mode_4608x2592_regs[] = {
 	{0x341f, 0x20},
 	{0x3420, 0x00},
 	{0x3421, 0xd8},
-	{0xC428, 0x00},
-	{0xC429, 0x04},
 	{0x3366, 0x00},
 	{0x3367, 0x00},
 	{0x3368, 0x00},
@@ -677,7 +692,8 @@ static const struct imx708_mode supported_modes_10bit_no_hdr[] = {
 		.pixel_rate = 595200000,
 		.exposure_lines_min = 8,
 		.exposure_lines_step = 1,
-		.hdr = false
+		.hdr = false,
+		.remosaic = true
 	},
 	{
 		/* regular 2x2 binned. */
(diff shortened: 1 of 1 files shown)


njhollinghurst:
Sony don't specify definitive values for these QBC  parameters. I also experimented with QBC Re-mosaic noise-reduction and false-colour curves but those had a less obvious effect, so have left them at the defaults (for minimal filtering).

njhollinghurst:
Another thought: Since the change can't be proven to be monotonically beneficial, maybe it should be a module parameter?
 (I see we have a couple of those already in imx477, e.g. `dpc_enable` which is very much in the same ballpark.)
@6by9 @naushir opinions?

6by9:
No particularly strong views from me.
If you can come up with an appropriate module name then you could do that.

njhollinghurst:
Hmm. The register name `LPF_INTENSITY` makes it easily confused with a different sharpness parameter, also documented as "LPF intensity", and doesn't seem to capture its purpose. Might just call it `qbc_adjust`.

6by9:
> Hmm. The register name `LPF_INTENSITY` makes it easily confused with a different sharpness parameter, also documented as "LPF intensity", and doesn't seem to capture its purpose. Might just call it `qbc_adjust`.

If you assign specific values rather than allowing the user to write any arbitrary rubbish, then you can describe what those values are in the parameter text, eg `modinfo drm`
```
pi@bookworm64:~ $ modinfo drm
filename:       /lib/modules/6.6.0-v8+/kernel/drivers/gpu/drm/drm.ko.xz
...
parm:           debug:Enable debug output, where each bit enables a debug category.
		Bit 0 (0x01)  will enable CORE messages (drm core code)
		Bit 1 (0x02)  will enable DRIVER messages (drm controller code)
		Bit 2 (0x04)  will enable KMS messages (modesetting code)
		Bit 3 (0x08)  will enable PRIME messages (prime code)
		Bit 4 (0x10)  will enable ATOMIC messages (atomic code)
		Bit 5 (0x20)  will enable VBL messages (vblank code)
		Bit 7 (0x80)  will enable LEASE messages (leasing code)
		Bit 8 (0x100) will enable DP messages (displayport code) (ulong)
parm:           edid_fixup:Minimum number of valid EDID header bytes (0-8, default 6) (int)
```
https://elixir.bootlin.com/linux/latest/source/drivers/gpu/drm/drm_print.c#L45

naushir:
``qbc_adjust`` seems a reasonable name to me.

drivers/media/i2c/imx708.c
```diff
@@ -1515,6 +1521,10 @@ static int imx708_start_streaming(struct imx708 *imx708)
 		return ret;
 	}
 
+	/* Quad Bayer re-mosaic adjustments (for full-resolution mode only) */
```
6by9, line 1537: So does the sensor handle this only affecting full res mode? The registers were in imx708_reg mode_4608x2592_regs so inherently only affected full res mode, but now this is common for all modes.

drivers/media/i2c/imx708.c
```diff
@@ -1515,6 +1521,10 @@ static int imx708_start_streaming(struct imx708 *imx708)
...
+	imx708_write_reg(imx708, 0xC428, IMX708_REG_VALUE_08BIT, !qbc_adjust);
+	imx708_write_reg(imx708, 0xC429, IMX708_REG_VALUE_08BIT, qbc_adjust);
```
6by9: 0xc428 is an enable and 0xc429 is a strength?

Checking datasheet it is, but with 0xc428 = 0 being enabled, and = 1 being disabled. Quirky!

I'd be tempted to use `qbc_adjust ? 0x01 : 0x00` instead of `!qbc_adjust` to make it more obvious the values being programmed, otherwise you end up checking the C specification to confirm that !0 = 1. Even better would be macros defining the bits in the register.

drivers/media/i2c/imx708.c
njhollinghurst: I presumed there were no side-effects from redundantly writing them, so it seemed unnecessary to spend a line of code checking the mode.

Previously they would have been left unchanged when switching from full-res to another mode (not that that happened often). More commonly they would keep their reset values (0x01,0x01). Yes I suppose it would make sense to explicitly set the "disable" flag in other modes.

drivers/media/i2c/imx708.c
njhollinghurst: I've seen many examples of ! and !! for this elsewhere but I agree ?: is clearer.

drivers/media/i2c/imx708.c
6by9: > Previously they would have been left unchanged when switching from full-res to another mode (not that that happened often).

Did they actually persist? imx708_set_stream calls pm_runtime_get_sync on enable and pm_runtime_put on disable, which in theory enables and disables regulators (may be delayed, I've never checked). This is why Laurent is pushing for using `pm_runtime_set_autosuspend_delay` / `pm_runtime_use_autosuspend` / `pm_runtime_put_autosuspend` to definitely avoid dropping regulators on a mode switch.

If tested then I'm happy for it to remain unconditional.

njhollinghurst: A running switch can happen in libcamera-still when taking a timelapse, or some Python cases -- not common ones. Anyway, it now sets the enable conditionally.

Yes, tested. I'd forgotten about HDR mode but have now tested that too.

(I swapped the order of the writes in the hope it might do a tail merge and save a few bytes of code...)

drivers/media/i2c/imx708.c
pelwell: Just make sure you get them the right way round....

njhollinghurst: Now reworked and hopefully clearer, if a bit less terse.

6by9:
LGTM (I'll ignore the unrelated whitespace change).

naushir:
LGTM

PR #5702, opened 2023-11-08
defconfigs: Drop FB_UDL from all Pi defconfigs

6by9:
CONFIG_FB_UBL is the fbdev driver for DisplayLink USB displays. CONFIG_UDL is the DRM driver for DisplayLink USB displays.

Both attach themselves to the same USB device alias, so both get loaded for any attached display. You now have 2 drivers for the same hardware, and a race as to which one will load.

DRM emulates the framebuffer interface, so disable the fbdev driver.

https://github.com/raspberrypi/bookworm-feedback/issues/133


popcornmix:
Make sense.

PR #5703, opened 2023-11-08
Include I2C bus details in ft5x06 and goodix touch driver names

6by9:
https://github.com/raspberrypi/bookworm-feedback/issues/138

The name is only taken as a `const char *`, so there's no easy way for the framework to do this munge.

Adds the I2C bus number and address to the name, so DSI-2 wants to be associated with `4-0038 generic ft5x06 (79)`, and DSI-1 with `6-0038 generic ft5x06 (79)` for Pi5, and `10-0038 generic ft5x06 (79)` with DSI-1 on Pi0-4.
Goodix is used by the HyperPixel and a few other displays, hence converted that one too.

Needs the relevant wayfire.ini updates before merging / releasing, hence draft for now.

### DIFF
--- a/drivers/input/touchscreen/edt-ft5x06.c
+++ b/drivers/input/touchscreen/edt-ft5x06.c
@@ -67,6 +67,7 @@
 #define TOUCH_EVENT_RESERVED		0x03
 
 #define EDT_NAME_LEN			23
+#define EDT_NAME_PREFIX_LEN		8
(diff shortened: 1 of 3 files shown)


lurch:
> with DSI-1 on Pi0-4.

Pi1-4, as Pi0 doesn't have a DSI connector :wink:

6by9:
Simon has made the relevant ini file changes, so this is ready for review and merge.

PR #5711, opened 2023-11-14
media/i2c: ov7251: Switch from V4L2_CID_GAIN to V4L2_CID_ANALOGUE_GAIN

6by9:
The mainline driver has implemented analogue gain using the control V4L2_CID_GAIN instead of V4L2_CID_ANALOGUE_GAIN.

libcamera requires V4L2_CID_ANALOGUE_GAIN, and therefore fails.

Update the driver to use V4L2_CID_ANALOGUE_GAIN.

https://forums.raspberrypi.com/viewtopic.php?p=2157035
Needs to be confirmed to work before merging, but want CI.


6by9:
Rebased and tested. I guess this ought to be sent upstream as well (such joys).

6by9:
Any objections on this?

pelwell:
There's not much room for subtle errors in that change.

Issue #5295, opened 2022-12-30
Compiling the kernel with LLVM=1 fails for ARCH=arm CROSS_COMPILE=armv6l-unknown-linux-gnueabihf- kernels


eliasnaur:
### Describe the bug

The issue seems to be that `https://github.com/raspberrypi/linux/blob/rpi-5.15.y/arch/arm/lib/memcmp_rpi.S` (and perhaps others) are not compatible with the clang assembler.

### Steps to reproduce the behaviour

On a nix system with flakes enabled,

```sh
$ nix build github:eliasnaur/rpikernel-clang-issue
error: builder for '/nix/store/3zc9nww3crck85wir5fs78zfyfv1nm5q-Raspberry-Pi-Linux-kernel.drv' failed with exit code 2;
last 10 log lines:
>         ldrmib DAT5, [S_2]
>         ^
> <instantiation>:11:9: note: while in macro instantiation
>         memcmp_trailing_15bytes unaligned
>         ^
> arch/arm/lib/memcmp_rpi.S:265:6: note: while in macro instantiation
> 140: memcmp_short_inner_loop 1
>      ^
> make[1]: *** [scripts/Makefile.build:391: arch/arm/lib/memcmp_rpi.o] Error 1
> make: *** [Makefile:1900: arch/arm/lib] Error 2
For full logs, run 'nix log /nix/store/3zc9nww3crck85wir5fs78zfyfv1nm5q-Raspberry-Pi-Linux-kernel.drv'.
```

The relevant configuration and building scripts are in the [repository](https://github.com/eliasnaur/rpikernel-clang-issue/blob/d8ae540103c27c52023c306b61c3c5af4a87d851/flake.nix#L24). I'm using [clang 14](https://github.com/eliasnaur/rpikernel-clang-issue/blob/d8ae540103c27c52023c306b61c3c5af4a87d851/flake.nix#L28).

### Device (s)

Raspberry Pi Zero, Raspberry Pi Zero W / WH

### System

The issue is during building of custom kernel, not at runtime.

### Additional context

I'm building a custom kernel with LLVM because I have problems producing deterministic output from a gcc build.

The issue can be worked around by disabling `BCM2835_FAST_MEMCPY` in the kernel config.

pelwell:
Are there no further error messages? The ones you include are showing where the problem is but not what.

eliasnaur:
Here's the full `nix log` output:

[...]

PR #5718, opened 2023-11-18
arch/arm/lib: Fix the assembly sources for the kernel build by LLVM

altimeter-130ft:
The LLVM assembler is stricter on the source syntax than GCC.

- Prepend a backslash (\) to each macro argument.

- The condition field is always the last portion of an instruction mnemonic.

- The terms of the relational operators in a conditional assemble directive are integers, not strings.

The assembly source files complying above assemble by both LLVM and GCC / Binutils.

Issue: [5295](https://github.com/raspberrypi/linux/issues/5295)
Build tested on: clang 14.0.6 and gcc Raspbian 12.2.0-14+rpi1, both armhf native on qemu-static.
Kernel exec tested on: Raspberry Pi 3B+ and Zero W.

### DIFF
--- a/arch/arm/lib/memcpymove.h
+++ b/arch/arm/lib/memcpymove.h
@@ -27,111 +27,111 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 */
 
 .macro unaligned_words  backwards, align, use_pld, words, r0, r1, r2, r3, r4, r5, r6, r7, r8
- .if words == 1
-  .if backwards
-        mov     r1, r0, lsl #32-align*8
+ .if \words == 1
+  .if \backwards
+        mov     \r1, \r0, lsl #32-\align*8
         ldr     r0, [S, #-4]!
-        orr     r1, r1, r0, lsr #align*8
-        str     r1, [D, #-4]!
+        orr     \r1, \r1, \r0, lsr #\align*8
+        str     \r1, [D, #-4]!
   .else
-        mov     r0, r1, lsr #align*8
-        ldr     r1, [S, #4]!
-        orr     r0, r0, r1, lsl #32-align*8
-        str     r0, [D], #4
+        mov     \r0, \r1, lsr #\align*8
+        ldr     \r1, [S, #4]!
+        orr     \r0, \r0, \r1, lsl #32-\align*8
+        str     \r0, [D], #4
   .endif
- .elseif words == 2
-  .if backwards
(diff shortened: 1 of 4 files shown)


arch/arm/lib/memcpymove.h
```diff
@@ -419,43 +419,43 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
+ .if \backwards
         ands    DAT2, D, #15
         beq     164f
  .else
         ands    DAT2, D, #15
         beq     164f
         rsb     DAT2, DAT2, #16
  .endif
-        memcpy_leading_15bytes backwards, align
+        memcpy_leading_15bytes \backwards, 1
```
iucoen, line 443: ~~Should this be \align ?~~
Align is not an argument to this macro. How did the old code even compile?

altimeter-130ft: I want to know why, too 😂

**The test assembly source for a quick reproduction**
```
    .text
    .arch armv6
    .object_arch armv4
    .arm
    .p2align 2

.macro should_not_assemble_macro arg
    .if \arg == 0
        add       a1, a2, #4
    .else
        sub       a1, a3, #2
    .endif
.endm

should_not_assemble:
    should_not_assemble_macro arg

    .type should_not_assemble, %function

    .end
```

**GCC**
```
(raspbian-bookworm-armhf)a130ft@berget:~/work/should-not-assemble$ gcc -o should-not-assemble.o -c should-not-assemble.S 
(raspbian-bookworm-armhf)a130ft@berget:~/work/should-not-assemble$ objdump --disassemble should-not-assemble.o

should-not-assemble.o:     file format elf32-littlearm


Disassembly of section .text:

00000000 <should_not_assemble>:
   0:	e2420002 	sub	r0, r2, #2
(raspbian-bookworm-armhf)a130ft@berget:~/work/should-not-assemble$ 
```

**LLVM (clang)**
```
(raspbian-bookworm-armhf)a130ft@berget:~/work/should-not-assemble$ clang -o should-not-assemble.o -c should-not-assemble.S 
<instantiation>:1:5: error: expected absolute expression
.if arg == 0
    ^
should-not-assemble.S:16:5: note: while in macro instantiation
    should_not_assemble_macro arg
    ^
(raspbian-bookworm-armhf)a130ft@berget:~/work/should-not-assemble$ 
```

So gcc insists that `arg == 0` expanded out of `\arg == 0` has the acceptable syntax and evaluates to false.

PR #5720, opened 2023-11-19
drm/vc4: Set TV margins on the composite connector state

sly547:
The `overscan_*` settings in the config.txt do not work on the composite video output with the VC4 KMS driver. This patch fixes it (tested on a Raspberry Pi 4). See the commit message for a more detailed description.

Here are some forum entries that I collected where people are reporting this issue:
https://forums.raspberrypi.com/viewtopic.php?t=330579
https://forums.raspberrypi.com/viewtopic.php?t=330924
https://forums.raspberrypi.com/viewtopic.php?t=342062

I did not test this with an upstream kernel, but it already does the necessary `drm_atomic_helper_connector_tv_reset()` in the conncetor reset function. On the other hand it does not call `drm_connector_attach_tv_margin_properties()` in the connector init function, which I think is necessary?!

What is the correct approach here: patch the raspberrypi kernel, then try to send a similar patch upstream?

### DIFF
--- a/drivers/gpu/drm/vc4/vc4_vec.c
+++ b/drivers/gpu/drm/vc4/vc4_vec.c
@@ -436,6 +436,9 @@ static void vc4_vec_connector_reset(struct drm_connector *connector)
 	/* preserve TV standard */
 	if (connector->state)
 		connector->state->tv.mode = vc4_vec_get_default_mode(connector);
+
+	/* apply TV margins from the cmdline_mode */
+	drm_atomic_helper_connector_tv_reset(connector);
 }
 
 static int vc4_vec_connector_atomic_check(struct drm_connector *conn,
@@ -482,6 +485,8 @@ static int vc4_vec_connector_init(struct drm_device *dev, struct vc4_vec *vec)
 		return ret;
 
 	drm_connector_helper_add(connector, &vc4_vec_connector_helper_funcs);
+
+	drm_connector_attach_tv_margin_properties(connector);
 
 	drm_object_attach_property(&connector->base,
 				   dev->mode_config.tv_mode_property,


6by9:
This partly looks to have been fixed in mainline since 6.3 with https://github.com/torvalds/linux/commit/c104b23147c725dcdf27558c5f37258aa66e38ed, however I think that has missed the call to `drm_connector_attach_tv_margin_properties`.

Downstream has gained https://github.com/raspberrypi/linux/commit/745e0786d92efd77c2cbd59be2f94a0997419789 in the meantime which allows the TV standard to be set via a module parameter.

I'll try to test it tomorrow to see what the state of play is in both cases. I'm happy to have a patch for it on this 6.1 branch, and we'll look at fixing mainline another day (I have a BIG pile of patches to send for Pi5 support)

sly547:
That sounds like a good approach to me. Thanks for looking into this. Let me know if I can help out with testing.

The margins seem to be different from the TV standard in that the comma-separated options in `video=` more or less directly translate into the numbers in the connector state. So my guess is that a new module parameter would not make sense here?

I'll push the commit again with my sign-off message.

sly547:
I couldn't stop myself from setting up a Raspberry Pi 1 with an upstream linux-6.6.2, and getting PAL output to work.

The trick was to manually set PAL mode with one of these:
```
modetest -M vc4 -w 50:"mode":2
modetest -M vc4 -w 50:"TV mode":3
```

The margins from the kernel cmdline are getting applied without further changes, thanks to `drm_atomic_helper_connector_tv_reset()` being called.

When I modify `vc4_vec_connector_init()` to do `drm_connector_attach_tv_margin_properties()`, the margin properties are also available to libdrm's `drmModeObjectGetProperties()` and can be changed with e.g.
```
modetest -M vc4 -w 50:"left margin":100
```

6by9:
Sorry for the delay in testing this.
All looks good for this patching 6.1.

I'll sort out the patch to add `drm_connector_attach_tv_margin_properties` to the 6.6 kernel, and will upstream it in due course.

6by9:
#5745 for the 6.6 fix.

6.2 to 6.5 are unlikely to get updated as they've been superceded by 6.6.

6by9:
@popcornmix Could you cast an extra set of eyes over this one.

Upstream 6.3 gained some changes in this area, hence 6.6 needing fewer changes.

sly547:
Perfect, thank you!

PR #5733, opened 2023-11-22
drm: vc4: Free the dlist alloc immediately if it never hit the hw

6by9:
atomic_check creates a state, and allocates the dlist memory for it such that atomic_flush can not fail.

On destroy that dlist allocation was being put in the stale list, even though it had never been programmed into the hardware, therefore doing lots of atomic_checks could consume all the dlist memory and fail.

If the dlist has never been programmed into the hardware, then free it immediately.


popcornmix:
LGTM.
A quick test on kodi didn't show any issues on Pi4 or Pi5.

Issue #5727, opened 2023-11-21
RPI4: Hardware plane offloading is broken for NV12 (zero-copy video playback)


rmader:
#### Note: Wayfire, as currently shipped, does not seem to support NV12 offloading. So this issue was only reproduced so far with other compositors, even though I assume that Wayfire will gain support for it eventually.

Context: The RPI4 advertises hardware plane support for NV12 (linear modifier) which is a useful feature to display content from the hardware video decoder or attached cameras without extra copy. Unfortunately this appears to be broken and the only thing displayed is solid green - the equivalent to black/all zero input for NV12.

This has been reproduced on multiple compositors, on this kernel and upstream (see below).

A visual example, showing the `vivid` kernel module test pattern:

![2023-11-21-01-42-47-575](https://github.com/raspberrypi/linux/assets/6313774/315c8cfb-72c6-4174-88cf-815f3c224386)

For comparison the pattern composited (also works on Wayfire):
![2023-11-21-01-42-54-888](https://github.com/raspberrypi/linux/assets/6313774/fa3bb22d-59b8-4022-8b92-33263bc3ff83)

This happens with different source and destination resolutions, also in fullscreen. I get the exact same behavior on different compositors (Weston and Mutter, both of which work on other hardware), kernels (this downstream and upstream as shipped in Fedora), OSs (see below) and Mesa versions (and double-checked already that the values passed to KMS are sane (minus https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/26283).

So I can say with high confidence that this is most likely a kernel driver issue (or firmware - in any case not on layers above).

### Steps to reproduce the behaviour

I have reproduced the problem in various ways, but the following should make it comparatively easy. If you want to reproduce and struggle at any point, please feel free to ask for more elaborate steps:
1. build `weston` (and the included `weston-simple-dmabuf-v4l`) from `main` - or backport https://gitlab.freedesktop.org/wayland/weston/-/merge_requests/1177 so `weston-simple-dmabuf-v4l` can correctly upload `NV12`
2. enable the `vivid` test pattern as explained by `weston-simple-dmabuf-v4l -h`, i.e.
    - `options vivid node_types=0x1 num_inputs=1 input_types=0x00` > `/etc/modprobe.d/vivid.conf`
    - `modprobe vivid`
    - `v4l2-ctl -d /dev/video0 --set-fmt-video=width=1920,height=1080,pixelformat=NV12`
3. run `weston` as compositor (just on any TTY)
[...]

PR #5734, opened 2023-11-22
drm: vc4: Block swiotlb bounce buffers being imported as dmabuf

6by9:
The dmabuf import already checks that the backing buffer is contiguous and rejects it if it isn't. vc4 also requires that the buffer is in the bottom 1GB of RAM, and this is all correctly defined via dma-ranges.

However the kernel silently uses swiotlb to bounce dma buffers around if they are in the wrong region. This relies on dma sync functions to be called in order to copy the data to/from the bounce buffer.

DRM is based on all memory allocations being coherent with the GPU so that any updates to a framebuffer will be acted on without the need for any additional update. This is fairly fundamentally incompatible with needing to call dma_sync_ to handle the bounce buffer copies, and therefore we have to detect and reject mappings that use bounce buffers.

### DIFF
--- a/drivers/gpu/drm/vc4/vc4_drv.c
+++ b/drivers/gpu/drm/vc4/vc4_drv.c
@@ -29,6 +29,7 @@
 #include <linux/of_platform.h>
 #include <linux/platform_device.h>
 #include <linux/pm_runtime.h>
+#include <linux/dma-direct.h>
 
 #include <drm/drm_aperture.h>
 #include <drm/drm_atomic_helper.h>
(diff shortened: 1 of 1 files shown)


6by9:
Fixes https://github.com/raspberrypi/linux/issues/5727

pelwell:
This is going to reject buffers that end up in swiotlb pages. How does this rejection manifest itself? Does it not cause another problem?

A very brief skim of some documentation on dmabufs (https://docs.kernel.org/driver-api/dma-buf.html) made me think that the DMA synchronisation is done at the request of the application, using the DMA_BUF_IOCTL_SYNC ioctl.

6by9:
This rejection will cause the userspace `ìoctl(DRM_IOCTL_PRIME_FD_TO_HANDLE)` to fail, which is permitted.
If the buffer isn't contiguous then you'll get the same error - https://github.com/raspberrypi/linux/blob/rpi-6.1.y/drivers/gpu/drm/drm_gem_dma_helper.c#L469-L470

Yes the DMABUF API covers syncing to the main dmabuf allocation, but that doesn't cover syncing to the bounce buffer which only happens when the DMA consumer (ie HVS) says it wants to sync. DRM totally ignores doing that, therefore the bounce buffer never gets copied into.
V3D having an iommu and able to address all of memory will import this buffer happily, and the DMABUF API will ensure that the data that it sees is coherent whenever the CPU has modified it.

This is effectively the same issue that the firmware used to have to handle with the transposer and `display_rotate` for 90 and 270 deg. The system sees the framebuffer and thinks it is on the screen, but something has to be kicked to generate the rotated (or bounced) buffer every frame. That existed in the firmware, but doesn't in DRM.

pelwell:
I'd like to discuss this more - in person is easier - but I'll merge the PR as at least an improvement.

Issue #5726, opened 2023-11-20
No working hysteresis support in gov_step_wise.c => fan switching on/off very frequently.


jkreileder:
### Describe the bug

gov_step_wise.c has hysteresis code but this actually does nothing as ops->get_trip_hyst is an optional operation and there's no implementation of it.
On the pi5 hyst is set to 5000 for all trip points but because of the missing get_trip_hyst, it's basically always 0 in gov_step_wise.c.

The result are very frequent switches around trip temperatures.  E.g. on my pi5 I see more than 1000 on/off switches around 50 degrees in 1 hour. Sometimes the fan only stops for a fraction of a second before spinning up again.  Not sure if that is good for the fan in the long term.

Note that Linux 6.7 removes the get_trip_hyst operation altogether: https://github.com/torvalds/linux/commit/35d8dbbb25add265a880ab0dc48a229f06b08325
Apparently the hyst values should be read directly from the tables. 

### Steps to reproduce the behaviour

On a PI 5:

echo 'file gov_step_wise.c +p' | sudo tee /sys/kernel/debug/dynamic_debug/control
dmesg

=> temp and hyst (which is/should be "temp - configured hyst" value here) are always the same.

### Device (s)

Other

### System

Ubuntu 23.10 
Linux pi5 6.5.0-1006-raspi #9~fanfix1-Ubuntu SMP PREEMPT_DYNAMIC Wed Nov 15 15:03:29 UTC 20 aarch64 aarch64 aarch64 GNU/Linux

pelwell:
This is clearly a problem that will need to be resolved before we move to 6.6 - the next LTS kernel.

PR #5736, opened 2023-11-22
Fix hysteresis support in gov_step_wise.c

jkreileder:
Directly get hyst value instead of going through an optional and, now, unimplemented function.

Fixes https://github.com/raspberrypi/linux/issues/5726
(Note that there are more drivers with this problem. But this PR fixes it for stand Pi fans.)

### DIFF
--- a/drivers/thermal/gov_step_wise.c
+++ b/drivers/thermal/gov_step_wise.c
@@ -98,15 +98,9 @@ static void thermal_zone_trip_update(struct thermal_zone_device *tz, int trip_id
 		trace_thermal_zone_trip(tz, trip_id, trip.type);
 	}
 
-	hyst_temp = trip_temp = trip.temperature;
+	trip_temp = trip.temperature;
+	hyst_temp = trip_temp - trip.hysteresis;
 	trip_type = trip.type;
-
-	if (trip.hysteresis)
-		hyst_temp = trip_temp - trip.hysteresis;
-	else
-		dev_info_once(&tz->device,
-			"Zero hysteresis value for Trip%d[type=%d]\n",
-			trip_id, trip_type);
 
 	dev_dbg(&tz->device,
 		"Trip%d[type=%d,temp=%d,hyst=%d]:trend=%d,throttle=%d\n",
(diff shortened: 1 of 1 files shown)


drivers/thermal/gov_step_wise.c
```diff
@@ -99,12 +99,15 @@ static void thermal_zone_trip_update(struct thermal_zone_device *tz, int trip_id
 	}
 
 	hyst_temp = trip_temp = trip.temperature;
-	if (tz->ops->get_trip_hyst) {
-		tz->ops->get_trip_hyst(tz, trip_id, &hyst_temp);
-		hyst_temp = trip_temp - hyst_temp;
-	}
 	trip_type = trip.type;
 
+	if (trip.hysteresis)
```
pelwell: I don't see why you've put this code here - the location of the old version is more logical.

drivers/thermal/gov_step_wise.c
```diff
@@ -99,12 +99,15 @@ static void thermal_zone_trip_update(struct thermal_zone_device *tz, int trip_id
...
+		hyst_temp = trip_temp - trip.hysteresis;
+	else
```
pelwell: I don't think we need to warn about a zero hysteresis - what's wrong with this?:
```
    trip_temp = trip.temperature;
    hyst_temp = trip_temp - trip.hysteresis;
```

pelwell (commented):
Thanks for the patch - the new mechanism is much simpler.

drivers/thermal/gov_step_wise.c
jkreileder: Nothing really, the only reason for this was to make it more like the existing code in gov_bang_bang.c: https://github.com/torvalds/linux/blob/9b6de136b5f0158c60844f85286a593cb70fb364/drivers/thermal/gov_bang_bang.c#L22-L24

I'll make a simplified PR later.

jkreileder: Updated now.

pelwell:
Thanks!

PR #5738, opened 2023-11-23
Add support for adt7410 (module and overlay)

aSmig:
The Analog Devices ADT7410 is a high precision temperature sensor with growing popularity. This PR seeks to add the module and overlay.

### DIFF
--- a/arch/arm/boot/dts/overlays/README
+++ b/arch/arm/boot/dts/overlays/README
@@ -2148,10 +2148,14 @@ Name:   i2c-sensor
 Info:   Adds support for a number of I2C barometric pressure, temperature,
         light level and chemical sensors on i2c_arm
 Load:   dtoverlay=i2c-sensor,<param>=<val>
-Params: addr                    Set the address for the BH1750, BME280, BME680,
-                                BMP280, BMP380, CCS811, DS1621, HDC100X, JC42,
-                                LM75, MCP980x, MPU6050, MPU9250, MS5637, MS5803,
-                                MS5805, MS5837, MS8607, SHT3x or TMP102
+Params: addr                    Set the address for the ADT7410, BH1750, BME280,
+                                BME680, BMP280, BMP380, CCS811, DS1621, HDC100X,
+                                JC42, LM75, MCP980x, MPU6050, MPU9250, MS5637,
+                                MS5803, MS5805, MS5837, MS8607, SHT3x or TMP102
+
+        adt7410                 Select the Analog Devices ADT7410 and ADT7420
+                                temperature sensors
+                                Valid address 0x48-0x4b, default 0x48
 
         aht10                   Select the Aosong AHT10 temperature and humidity
                                 sensor
--- a/arch/arm/boot/dts/overlays/i2c-sensor-common.dtsi
+++ b/arch/arm/boot/dts/overlays/i2c-sensor-common.dtsi
@@ -508,6 +508,21 @@
 		};
 	};
 
+	fragment@34 {
+		target = <&i2cbus>;
+		__dormant__ {
+			#address-cells = <1>;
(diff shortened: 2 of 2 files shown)


aSmig:
File sizes are fairly small:
2.0K    /boot/firmware/overlays/adt7410.dtbo
4.0K    /lib/modules/6.1.61-v8+/kernel/drivers/hwmon/adt7410.ko.xz
4.0K    /lib/modules/6.1.61-v8+/kernel/drivers/hwmon/adt7x10.ko.xz
10.0K   total

pelwell:
This needs a few small changes:
1. Without an entry in overlays/Makefile it won't be compiled and installed.
2. Don't include `CONFIG_SENSORS_ADT7X10=m` to the defconfigs - it is implied by `CONFIG_SENSORS_ADT7410=m`.

Otherwise it looks good.

aSmig:
Thanks for the quick review! How is this? I appreciate the extra help.

pelwell:
Looks good. I'll let the build checks run, squash them down to two commits, and merge offline.

pelwell:
On second thoughts (first thought, if I'm honest), as this is such a simple overlay for an I2C-attached sensor, I'd rather add it to the `i2c-sensor` overlay. There's a new commit that adds `adt7410` as a parameter - does that work for you?

aSmig:
Putting this into i2c-sensor sounds very reasonable. Unfortunately I haven't been successful at testing this approach on my system.

```
$ sudo dtoverlay i2c-sensor adt7410 addr=0x4a
* Failed to apply overlay '0_i2c-sensor' (kernel)
```

The simpler adt7410.dtbo did load successfully. This is new territory for me so I'm not sure how to debug. I also tried loading i2c-sensors,adt7410,addr=0x4a via config.txt with no luck. I must be doing something wrong when compiling the i2c-sensor overlay, so here are my steps:

 ```
wget https://raw.githubusercontent.com/raspberrypi/linux/5f9c19e21de5ed2a57d5560698a4ce2440518f62/arch/arm/boot/dts/overlays/i2c-sensor-common.dtsi
 cpp -nostdinc -I ~/linux/include -undef -x assembler-with-cpp i2c-sensor-common.dtsi i2c-sensor-common.dtsi.preprocessed
 mv i2c-sensor-common.dtsi.preprocessed i2c-sensor-common.dtsi
 cp ~/linux/arch/arm/boot/dts/overlays/i2c-sensor-overlay.dts .
 dtc -O dtb i2c-sensor-common.dtsi -o i2c-sensor.dtbo
 sudo cp i2c-sensor.dtbo /boot/overlays/
```

If you are confident that everything is fine and will just work with a complete kennel build, no need to help me troubleshoot my broken testing process. I'll trust you. Thanks again for your work!

**Comment by pelwell** [comment]
There was an unwanted underscore in one of the labels, which then ended up being undefined. I've tested and pushed a modified version. Until recently the dtoverlay command (not the config.txt directive) would fail because it lacked support for the newer overlay features - that should all be fixed in the latest version - but I think the i2c-sensor overlay is sufficiently simple that it would work even with older installs.

**Comment by pelwell** [comment]
By the way, I would have written your compilation step like this:
```
dtc -@ -Hepapr -I dts -O dtb -o i2c-sensor.dtbo i2c-sensor-overlay.dts
```
Some of the difference are cosmetic or nice-to-haves, but some of them (the `-@`, using the correct source file) are not.

aSmig:
This is great, thanks! Your updated i2c-sensor-common.dtsi is working on my system when compiled correctly, thanks for the tips. Don't we still need CONFIG_SENSORS_ADT7410=m added to the configs?

pelwell:
You're right - commit 057e09bbfd357 added.

aSmig:
Perfect. Thank you so much!

Issue #3593, opened 2020-05-03
Missing PPS modules for 64-bit builds of 5.4


trejan:
64-bit builds of the 5.4 kernel are missing the PPS modules (pps_core, pps-gpio and pps-ldisc). The other 32-bit builds of 5.4 and all builds of 4.19 do have the PPS modules.

```
./4.19.97+/kernel/drivers/pps/clients/pps-gpio.ko
./4.19.97+/kernel/drivers/pps/clients/pps-ldisc.ko
./4.19.97+/kernel/drivers/pps/pps_core.ko
./4.19.97-v7+/kernel/drivers/pps/clients/pps-gpio.ko
./4.19.97-v7+/kernel/drivers/pps/clients/pps-ldisc.ko
./4.19.97-v7+/kernel/drivers/pps/pps_core.ko
./4.19.97-v7l+/kernel/drivers/pps/clients/pps-gpio.ko
./4.19.97-v7l+/kernel/drivers/pps/clients/pps-ldisc.ko
./4.19.97-v7l+/kernel/drivers/pps/pps_core.ko
./4.19.97-v8+/kernel/drivers/pps/clients/pps-gpio.ko
./4.19.97-v8+/kernel/drivers/pps/clients/pps-ldisc.ko

./5.4.35+/kernel/drivers/pps/clients/pps-gpio.ko
./5.4.35+/kernel/drivers/pps/clients/pps-ldisc.ko
./5.4.35+/kernel/drivers/pps/pps_core.ko
./5.4.35-v7+/kernel/drivers/pps/clients/pps-gpio.ko
./5.4.35-v7+/kernel/drivers/pps/clients/pps-ldisc.ko
./5.4.35-v7+/kernel/drivers/pps/pps_core.ko
./5.4.35-v7l+/kernel/drivers/pps/clients/pps-gpio.ko
./5.4.35-v7l+/kernel/drivers/pps/clients/pps-ldisc.ko
./5.4.35-v7l+/kernel/drivers/pps/pps_core.ko

```
Diff between 4.19 and 5.4 version of arch/arm64/configs/bcm2711_defconfig shows these two config statements are missing:

```
CONFIG_PPS_CLIENT_LDISC=m
CONFIG_PPS_CLIENT_GPIO=m
```

pelwell:
I can't explain how those options got lost - the history has been squashed - but they have been restored in rpi-5.4.y and rpi-5.6.y, and hopefully will remain from now on.

[...]

PR #5739, opened 2023-11-23
configs: enable pps_core module

lightswitch05:
It seems the pps_core module has disappeared. I've been able to enable the [pps_gpio](https://github.com/raspberrypi/linux/issues/3593#issuecomment-1823795799) module, but not `pps_core`. Perhaps this configuration needs to be added? There is also a chance I'm way off base, thank you in advance for having a look at the PR.


pelwell:
What you've done is changed PPS from being built-in to being an external module, which is a step backwards. ntp_core (enabled with CONFIG_NTP_PPS) is not supported because we set `CONFIG_NO_HZ_COMMON=y`:
```
config NTP_PPS
	bool "PPS kernel consumer support"
	depends on !NO_HZ_COMMON
	help
	  This option adds support for direct in-kernel time
	  synchronization using an external PPS signal.

	  It doesn't work on tickless systems at the moment.
```
But that just means you need an external program to perform the synchronisation - isn't that what chronyd does?

FYI none of this has changed since 5.10, except for PPS being built-in.

lightswitch05:
Thank you for your response @pelwell. Your example is showing `NTP_PPS`, but this PR is for `CONFIG_PPS`. Is that not a different configuration?

The issue I'm having isn't setting up time sync with either chrony or NTP, but just getting the system to recognize the PPS device and create `/dev/pps0`.

I was torn between the `m` setting and the `y`. Or maybe this `CONFIG_PPS` config is being defaulted to `y` somewhere else?

pelwell:
CONFIG_PPS=y is being selected by something else. For each of the defconfigs, run:
```
$ make ARCH=arm <thing>_defconfig
$ grep PPS .config
$ make ARCH=arm savedefconfig
$ diff <thing>_defconfig defconfig
```
(or ARCH=arm64)
You'll see that CONFIG_PPS=y is set in all cases, and that when the defconfig is regenerated your setting is dropped.

lightswitch05:
Ah ok, thank you for clearing that up for me!

PR #5744, opened 2023-11-24
drivers: media: cfe: Find the source pads on the sensor entity

6by9:
The driver was assuming that pad 0 on the sensor entity was the appropriate source pad, but this isn't necessarily the case. With video-mux, it has the sink pads first, and then the source pad as the last one.

Iterate through the sensor pads to find the relevant source pads.

### DIFF
--- a/drivers/media/platform/raspberrypi/rp1_cfe/cfe.c
+++ b/drivers/media/platform/raspberrypi/rp1_cfe/cfe.c
@@ -1826,7 +1826,7 @@ static void cfe_unregister_nodes(struct cfe_device *cfe)
 
 static int cfe_link_node_pads(struct cfe_device *cfe)
 {
-	unsigned int i;
+	unsigned int i, source_pad = 0;
 	int ret;
 
 	for (i = 0; i < CSI2_NUM_CHANNELS; i++) {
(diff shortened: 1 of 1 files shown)


6by9:
@naushir - fixes up use of video-mux on Pi5. I've had 2 OV9281's running on camera-mux-4port with this patch.

naushir:
LGTM.  Logic seems correct and the embedded pad should be linked correctly as well.

drivers/media/platform/raspberrypi/rp1_cfe/cfe.c
```diff
@@ -1826,7 +1826,7 @@ static void cfe_unregister_nodes(struct cfe_device *cfe)
 
 static int cfe_link_node_pads(struct cfe_device *cfe)
 {
-	unsigned int i;
+	unsigned int i, source_pad = 0;
```
pelwell, line 1829: Putting the initialisation in the obvious slot in the `for` statement requires less mental effort from the reader.

pelwell: But the logic depends on it not being reinitialised between iterations of the outer loop.

pelwell: Would you consider this instead?
```
		while (source_pad < cfe->sensor->entity.num_pads &&
			!(cfe->sensor->entity.pads[source_pad].flags & MEDIA_PAD_FL_SOURCE))
			source_pad++
```
@naushir Do you have a preference? I'll bow to the majority.

6by9: It would be in the outer for loop, ie
```
for (i = 0, source_pad = 0; i < CSI2_NUM_CHANNELS; i++) {
```
which also looks a little strange.

Quite happy to swap to a while loop.

6by9: Done

pelwell:
LGTM. Merge when ready, @naushir.

naushir:
LGTM!

PR #5745, opened 2023-11-24
drm/vc4: vec: Add the margin properties to the connector

6by9:
All the handling for the properties was present, but they were never attached to the connector to allow userspace to change them.

Add them to the connector.


popcornmix:
LGTM

Issue #5748, opened 2023-11-26
Rpi5 sampling rate dac hat


antonellocaroli:
### Describe the bug

Set in config.txt to load
`dtoverlay=i-sabre-q2m`

but I don't understand why it shows me a sample rate of 192000
I remember that on Rpi4 it was up to 384000

Where does he get 192000 saple rate from?

```
*** Scanning for playback devices ***
Card 0, ID `DAC', name `I-Sabre Q2M DAC'
  Device 0, ID `I-Sabre Q2M DAC i-sabre-codec-dai-0', name `I-Sabre Q2M DAC i-sabre-codec-dai-0', 1 subdevices (1 available)
    2 channels, sampling rate 8000..192000 Hz
    Sample formats: S16_LE, S32_LE
    Buffer size range from 32 to 131072
    Period size range from 16 to 1024

      Subdevice 0, name `subdevice #0'

```
    
It seems that on anything the maximum samplerate is always 192000
      

### Steps to reproduce the behaviour

Set in config.txt to load
`dtoverlay=i-sabre-q2m`

### Device (s)

Other

### System

6.1.0-rpi4-rpi-2712

### Logs

[    0.000000] Booting Linux on physical CPU 0x0000000000 [0x414fd0b1]
[    0.000000] Linux version 6.1.0-rpi4-rpi-2712 (debian-kernel@lists.debian.org) (gcc-12 (Debian 12.2.0-14) 12.2.0, GNU ld (GNU Binutils for Debian) 2.40) #1 SMP PREEMPT Debian 1:6.1.54-1+rpt2 (2023-10-05)
[    0.000000] random: crng init done
[    0.000000] Machine model: Raspberry Pi 5 Model B Rev 1.0
[    0.000000] efi: UEFI not found.
[    0.000000] Reserved memory: created CMA memory pool at 0x000000003a000000, size 64 MiB
[    0.000000] OF: reserved mem: initialized node linux,cma, compatible id shared-dma-pool
[    0.000000] Zone ranges:
[    0.000000]   DMA      [mem 0x0000000000000000-0x00000000ffffffff]
[    0.000000]   DMA32    empty
[    0.000000]   Normal   [mem 0x0000000100000000-0x00000001ffffffff]
[    0.000000] Movable zone start for each node
[    0.000000] Early memory node ranges
[    0.000000]   node   0: [mem 0x0000000000000000-0x000000000007ffff]
[...]

PR #5751, opened 2023-11-27
ASoC: dwc: Permit sample rates up to 384kHz

pelwell:
The BCM2835 I2S block advertises clock rates up to 384kHz, and there's no reason why RP1's DWC I2S block shouldn't do the same.

See: https://github.com/raspberrypi/linux/issues/5748


popcornmix:
LGTM

PR #5755, opened 2023-11-28
ADS1115: allow specification of the i2c bus

JinShil:
Analogous to https://github.com/raspberrypi/linux/pull/5650

### DIFF
--- a/arch/arm/boot/dts/overlays/README
+++ b/arch/arm/boot/dts/overlays/README
@@ -489,6 +489,16 @@ Params: addr                    I2C bus address of device. Set based on how the
         cha_gain                Set the gain of the Programmable Gain
                                 Amplifier for this channel. (Default 1 sets the
                                 full scale of the channel to 4.096 Volts)
+        i2c0                    Choose the I2C0 bus on GPIOs 0&1
+        i2c_csi_dsi             Choose the I2C0 bus on GPIOs 44&45
+        i2c3                    Choose the I2C3 bus (configure with the i2c3
+                                overlay - BCM2711 only)
+        i2c4                    Choose the I2C4 bus (configure with the i2c4
+                                overlay - BCM2711 only)
+        i2c5                    Choose the I2C5 bus (configure with the i2c5
+                                overlay - BCM2711 only)
+        i2c6                    Choose the I2C6 bus (configure with the i2c6
+                                overlay - BCM2711 only)
 
         Channel parameters can be set for each enabled channel.
         A maximum of 4 channels can be enabled (letters a thru d).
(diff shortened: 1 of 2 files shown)


pelwell:
Even allowing for the checkpatch warnings (*), this is not going to work. The problem is the merge order of the fragments.

Fragments are merged in two linear passes:
1. Fragments that target other fragments (I call them _intra-_ fragments because they operate within the overlay) are applied.
2. Fragments that target the base dtb (_inter-_ fragments) are applied.

This scheme is more forgiving than the obvious one-pass, top-to-bottom scheme, but perhaps it lulls people into a false sense of security. What it lacks is a pass 0 that applies fragments that target intra fragments (_intra-intra_?), and a pass -1 that applies fragments that target intra fragments that target intra fragments (_intra-intra-intra_). As a result, all intra fragments (or at least intra fragments with dependency relationships) have to be written in the order in which they are applied.

Your modifications have fallen foul of this restriction by introducing an extra level to the hierarchy, and for no good reason. You could just as easily have put the frag100 label (ideally with a different name) on fragment 0 and deleted fragment 100 itself.

(*) I'm annoyed because I thought I had added enough infrastructure to the automated build checks to spot this kind of problem, but it clearly doesn't work. You can prove the problem exists yourself like this:
```
$ make ARCH=arm bcm2711_defconfig
$ make ARCH=arm dtbs
$ cd arch/arm/boot/dts
$ dtmerge bcm2711-rpi-4-b.dtb merged1.dtb overlays/ads1115.dtbo
$ dtmerge bcm2711-rpi-4-b.dtb merged2.dtb overlays/ads1115.dtbo chb_enable
$ dtdiff merged1.dtb merged2.dtb
```
(`dtmerge` is from the `raspi-utils` package which should be installed by default, or you can build it from our `raspberrypi/utils` repo)

JinShil:
How is this any different from https://github.com/raspberrypi/linux/blob/rpi-6.1.y/arch/arm/boot/dts/overlays/mcp23017-overlay.dts and https://github.com/raspberrypi/linux/blob/rpi-6.1.y/arch/arm/boot/dts/overlays/i2c-rtc-overlay.dts?  Don't those also suffer from the same flaw?

pelwell:
In your overlay, fragment 0 will be added to fragment 100 before e.g. fragment 1 is added to fragment 0.

Show me a problematic fragment in either of those overlays - I can't see one which doesn't obey the rules.

pelwell:
I've fixed the problem with the overlay checker, which now correctly reports:
```
* fragment@1 should precede fragment@0
*   Error ^ in overlay ads1115
```
Note that this is after checking _all_ of the overlays, not just `ads1115`.

JinShil:
I think I understand, but I'm not able to get `overlaycheck` to reproduce the error locally, and the time I've allotted to this has expired for now.

pelwell:
Are you happy for me to fix this version of the overlay and add it for you?

pelwell:
See: https://github.com/raspberrypi/linux/pull/5762

Issue #5741, opened 2023-11-24
Pisound on Pi 5


gtrainavicius:
### Describe the bug

Pisound is the I²S master, the default eeprom dt overlay we shipped it with is not working on Pi 5.

I've noticed issue #5724, and based on it I found that doing this procedure manages to load the driver and register the card successfully, the card then is usable:

1. Add `dtoverlay=pisound,master` in `/boot/config.txt`.
2. Power up Pi 5 without Pisound attached, to skip loading its hat eeprom.
3. Run `sudo modprobe -r snd_soc_pisound` to unload the Pisound's driver (as it obviously failed to load without Pisound attached).
4. Carefully attach Pisound on powered on Pi 5.
5. Run `sudo modprobe snd_soc_pisound` to load the Pisound's driver, it loads ok this time.
6. Check `amidi -l`, `aplay -l`, `arecord -l` - it's listed as expected.

Quick tests indicate that Pisound is fully operational once the driver manages to register the card.

Powering on the Pi 5 without Pisound attached, but with the dtoverlay line in config.txt, is necessary, otherwise the system does not boot at all when the hat eeprom is available, and the dtoverlay line for the same hat is there.

What should be done to make Pisound working on Pi 5 right out of the box?

### Steps to reproduce the behaviour

Have Pisound attached as normal, the driver fails to register the card:

```
[    3.200062] pisound: Detected Pisound card:
[    3.200072] pisound:         Serial:           PS-3...
[    3.200075] pisound:         Firmware Version: 1.03
[    3.200078] pisound:         Hardware Version: 1.1
[    3.200081] pisound:         Id:               ...
[    3.200622] designware-i2s 1f000a0000.i2s: ASoC: error at snd_soc_dai_set_fmt on 1f000a0000.i2s: -22
[    3.203481] pisound: snd_soc_register_card() failed: -22
```

### Device (s)

Other

### System

```
Raspberry Pi reference 2023-10-10
Generated using pi-gen, https://github.com/RPi-Distro/pi-gen, 962bf483c8f326405794827cce8c0313fd5880a8, stage4

2023/10/30 16:45:10
Copyright (c) 2012 Broadcom
version 30de0ba5 (release) (embedded)

Linux raspberrypi 6.1.0-rpi4-rpi-2712 #1 SMP PREEMPT Debian 1:6.1.54-1+rpt2 (2023-10-05) aarch64 GNU/Linux
```

### Additional context

[...]

PR #5757, opened 2023-11-28
ASoC: dwc: Fix full-duplex mode

pelwell:
Configuration of the DMA register was carelessly zeroing bits that may used by a stream in the other direction. Preserve them instead.

See: https://github.com/raspberrypi/linux/issues/5741


sound/soc/dwc/dwc-i2s.c
```diff
@@ -210,10 +212,6 @@ static void dw_i2s_config(struct dw_i2s_dev *dev, int stream)
 			dmacr |= (DMACR_DMAEN_RXCH0 << ch_reg);
```
popcornmix, line 217: It looks to me with the old code, unused channels would be cleared (by the `dmacr=0` initialisation and then not setting the bit). Does this still happen with new code? (e.g. if this is called once with `config->chan_nr=4` and then with `config->chan_nr=2`.

(Note: I'm not familiar with how this code is called, so ignore if you are happy with new behaviour).

pelwell: That's annoyingly correct. I've added clearing of the relevant group of channel enable flags before setting the appropriate number.

Issue #5766, opened 2023-12-02
[QCA7000] SPI missed interrupt


samsamfire:
### Describe the bug

Dear all,

I have a compute module 3 board connected to 2 MCP2515 chips and 1 QCA7000 chip (EV charging applications).
2-3 years ago we noticed a problem in the interrupt handling of the QCA7000. After a variable amount of time (couple of minutes to couple of hours) we would miss an interrupt on the QCA7000 chip, which would basically "freeze" the communication. 
We found out that basically, if the interrupt between QCA7000 and MCP2515 arrived at exactly the same time (us level), the interrupt would be missed on the QCA7000 interface. We spent some time digging trying to understand why this happened, but gave up in the end thinking that it was a hardware issue. We implemented a "dirty" fix in the qca_spi.c driver additionnaly to poll the QCA7000 every 50ms.
Several years have passed, and I am digging up this old problem that we were having at the time. We have seen on the field that sometimes the driver seems to "stop" receiving messages. I have also seen that sometimes, some packets can get dropped, and after reading extensively the QCA7000 datasheet and knowing that the RX buffer is pretty small, it turns out that this 50ms polling has just reduced the occurence of the problem but I don't think it has resolved the root cause. I have done some tests on STM32MP1 with only QCA as SPI device (CAN is integrated) and never get dropped packets / frozen interface.

Anyway, after reading this very interesting thread #1490 about 2 MCP2515 missing interrupts, I think that we are in the same case. The QCA7000 is indeed a **level-triggered** (high) interrupt, just like the MCP2515. We never saw this problem with the MCP2515 because we were running a 5.4.y kernel with this patch.

In #2267 @pelwell said "**If this is effective we should reconsider all usage of edge-triggered interrupts**". I don't know what the status is on this ? Also "ONESHOT" flag mentionned is not present in qcaspi driver. I don't exactly understand what it's supposed to do, but if we were to use a level-triggered interrupt shouldn't that also be added ?

What are your thoughts on this ? Has anybody experienced similar issues with other SPI devices ?

Thanks !

### Steps to reproduce the behaviour

Initiate repeated SLAC  with QCA, after a certain time interface will freeze (no more read messages).
CAN0 & CAN1 are also running.

### Device (s)

Raspberry Pi CM3+

### System

We are using raspbian buster with kernel v5.4.77.
I'll get back to you for the "firmware version".

lategoodbye:
[...]

PR #5771, opened 2023-12-05
overlays: qca7000: Use a LEVEL_HIGH interrupt

pelwell:
The QCA7000 interrupt should be level-triggered.

See: https://github.com/raspberrypi/linux/issues/5766


lategoodbye:
Please don't merge as this change cause serious trouble.

PR #5772, opened 2023-12-05
dts: bcm2712: put usb under /axi not /soc

P33M:
With dtoverlay=dwc2,dr_mode=host in /boot/firmware/config.txt, and the associated EEPROM firmware bump to turn the PHY on, I have a functional USB2 root port on the USB-C connector.

Still some rough edges if the port goes into suspend, which you get with a hub that has no occupied downstream ports.
- Frequent "mode mismatch interrupt" spam
- Disconnect interrupts aren't signalled.

### DIFF
--- a/arch/arm/boot/dts/bcm2712.dtsi
+++ b/arch/arm/boot/dts/bcm2712.dtsi
@@ -133,19 +133,6 @@
 			status = "disabled";
 		};
 
-		usb: usb@7c480000 {
-			compatible = "brcm,bcm2835-usb";
-			reg = <0x7c480000 0x10000>;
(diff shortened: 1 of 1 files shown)


6by9:
Do we need to revert https://github.com/raspberrypi/linux/commit/ff2db5847beb33038d658657e5720aef499a4669 and https://github.com/raspberrypi/linux/commit/74d906b6ca1b4b4bc63c531e38c314fc83d9223b now so that we don't have a restricted DMA region with dwc2?

pelwell:
We certainly can revert them, since the resulting dma zone will then be the first 4GB, matching the dwc2 range. To not do so, apart from retaining a spurious extra overlay, would unnecessarily limit the locations for the swiotlb bufffers, but that probably wouldn't be harmful (or even noticeable).

PR #5774, opened 2023-12-06
Allow IMX477 and IMX296 to be always on, to simplify multi-camera sync

njhollinghurst:
The issue is that when the cameras are powered down, they clamp sync signals (XVS, XHS, XTRIG) to ground. As they are active-low, and not all the cameras start streaming at exactly the same time, it could result in spurious triggers.

A previous workaround involved low-impedance pull-ups so that all cameras would see a high quiescent level. This is potentially unsafe and does not scale to large multi-camera setups.

This change allows cameras to be brought up with XVS, XHS, XTRIG pins in a high-impedance state.

### DIFF
--- a/drivers/media/i2c/imx477.c
+++ b/drivers/media/i2c/imx477.c
@@ -1742,25 +1742,20 @@ static int imx477_start_streaming(struct imx477 *imx477)
 	imx477_write_reg(imx477, 0x0b05, IMX477_REG_VALUE_08BIT, !!dpc_enable);
 	imx477_write_reg(imx477, 0x0b06, IMX477_REG_VALUE_08BIT, !!dpc_enable);
 
-	/* Set vsync trigger mode */
-	if (trigger_mode != 0) {
-		/* trigger_mode == 1 for source, 2 for sink */
-		const u32 val = (trigger_mode == 1) ? 1 : 0;
-
-		imx477_write_reg(imx477, IMX477_REG_MC_MODE,
-				 IMX477_REG_VALUE_08BIT, 1);
-		imx477_write_reg(imx477, IMX477_REG_MS_SEL,
-				 IMX477_REG_VALUE_08BIT, val);
-		imx477_write_reg(imx477, IMX477_REG_XVS_IO_CTRL,
-				 IMX477_REG_VALUE_08BIT, val);
-		imx477_write_reg(imx477, IMX477_REG_EXTOUT_EN,
-				 IMX477_REG_VALUE_08BIT, val);
-	}
-
 	/* Apply customized values from user */
 	ret =  __v4l2_ctrl_handler_setup(imx477->sd.ctrl_handler);
 	if (ret)
 		return ret;
+
+	/* Set vsync trigger mode: 0=standalone, 1=source, 2=sink */
+	imx477_write_reg(imx477, IMX477_REG_MC_MODE,
+			 IMX477_REG_VALUE_08BIT, (trigger_mode > 0) ? 1 : 0);
+	imx477_write_reg(imx477, IMX477_REG_MS_SEL,
+			 IMX477_REG_VALUE_08BIT, (trigger_mode <= 1) ? 1 : 0);
+	imx477_write_reg(imx477, IMX477_REG_XVS_IO_CTRL,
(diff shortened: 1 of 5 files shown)


njhollinghurst:
Draft pending more testing.

drivers/media/i2c/imx477.c
```diff
@@ -1742,20 +1742,15 @@ static int imx477_start_streaming(struct imx477 *imx477)
...
-		imx477_write_reg(imx477, IMX477_REG_XVS_IO_CTRL,
-				 IMX477_REG_VALUE_08BIT, val);
-		imx477_write_reg(imx477, IMX477_REG_EXTOUT_EN,
-				 IMX477_REG_VALUE_08BIT, val);
-	}
+	/* Set vsync trigger mode: 0=standalone, 1=source, 2=sink */
+	imx477_write_reg(imx477, IMX477_REG_MC_MODE,
+			 IMX477_REG_VALUE_08BIT, (trigger_mode > 0) ? 1 : 0);
+	imx477_write_reg(imx477, IMX477_REG_MS_SEL,
+			 IMX477_REG_VALUE_08BIT, (trigger_mode <= 1) ? 1 : 0);
```
pelwell, line 1754: Is the '<=' a typo? It doesn't fit the pattern of `IMX477_REG_XVS_IO_CTRL` and `IMX477_REG_EXTOUT_EN`, but if trigger_mode is guaranteed to only be 1 or 2 then the behaviour is the same.

6by9:
Looks reasonable to me.

drivers/media/i2c/imx477.c
```diff
@@ -1742,20 +1742,15 @@ static int imx477_start_streaming(struct imx477 *imx477)
...
-		imx477_write_reg(imx477, IMX477_REG_MS_SEL,
-				 IMX477_REG_VALUE_08BIT, val);
-		imx477_write_reg(imx477, IMX477_REG_XVS_IO_CTRL,
-				 IMX477_REG_VALUE_08BIT, val);
-		imx477_write_reg(imx477, IMX477_REG_EXTOUT_EN,
-				 IMX477_REG_VALUE_08BIT, val);
-	}
+	/* Set vsync trigger mode: 0=standalone, 1=source, 2=sink */
+	imx477_write_reg(imx477, IMX477_REG_MC_MODE,
+			 IMX477_REG_VALUE_08BIT, (trigger_mode > 0) ? 1 : 0);
```
pelwell, line 1752: All the `? 1 : 0`s look unnecessary, but they may be acceptable

njhollinghurst: It (almost) matches the previous style, but can be removed.

drivers/media/i2c/imx477.c
njhollinghurst: It is to get the correct state for trigger mode 0. We want it to match the reset state (the Software Reference Manual p137 mentions only "default" which I guess is the same thing: values 0,1,0,0 respectively. The doc doesn't explain the difference between MS_SEL and EXTOUT_EN but from the names it sounds reasonable. It might be that they are ignored when MC_MODE==0).

naushir (commented):
LGTM

njhollinghurst:
I have re-drafted this while investigating an interaction with XVS sync out -- there may be some (pre-existing?) glitch on start streaming which perhaps needs those register writes to be re-ordered; and when we _stop_ streaming it now does something weird (which we didn't previously see because the camera would be powered down).

njhollinghurst:
Update: Nearly got this right, for IMX477 at least.  There's still an awkward case for multi-camera sync. The problem is that after stopping the master camera streaming (now without POR) and configuring it to start again, XVS can start to emit spurious pulses every 2ms or so (which can cause the sink camera to get out of step).

By moving the I2C writes for trigger mode closer to the "start streaming" write, the problem can be prevented 90% of the time, but it remains sensitive to the real time interval between I2C writes. Using a multi-write transfer may improve it further but would clutter the driver (and is probably not watertight). cc @P33M 

One hopes that a different set of register incantations might cure this, but I haven't discovered one yet.

njhollinghurst:
I've pushed the mitigations against IMX477 jabbering (not needed for IMX296, as we only support that with external input).
I think this is as good as we can manage for now, and a monotonic improvement.

PR #5775, opened 2023-12-07
dwc_otg: Avoid the use of align_buf for short packets

popcornmix:
Recent kernels (from 6.5) fail to boot on Pi0-3.

This has been tracked down to the call to:
ret = usb_get_std_status(hdev, USB_RECIP_DEVICE, 0, &hubstatus);

returning garbage in hubstatus (it gets the uninitialised contents of a kmalloc buffer that is not overwritten as expected).

As we don't have strong evidence that this code path has ever worked, and it is causing a clear problem currently, lets disable it to allow wider use of newer kernels.

### DIFF
--- a/drivers/usb/host/dwc_otg/dwc_otg_hcd.c
+++ b/drivers/usb/host/dwc_otg/dwc_otg_hcd.c
@@ -1339,8 +1339,10 @@ static void assign_and_init_hc(dwc_otg_hcd_t * hcd, dwc_otg_qh_t * qh)
 			 * buffer.
 			 */
 			wLength = ((uint16_t *)urb->setup_packet)[3];
+			#if 0
 			if (hc->ep_is_in && wLength < 4)
 				ptr = hc->xfer_buff;
+			#endif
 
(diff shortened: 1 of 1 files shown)


popcornmix:
From investigation, it wasn't clear why the problem only became apparent with 6.5 kernel.

6.4 kernel has working usb.
6.5 kernel has non-working usb.
6.5 kernel with whole of `drivers/usb` reverted to 6.4 has non-working usb.

so it seems a non-usb related change causes the issue.
I suspect it may be some unrelated change provokes it by chance (e.g. timing, contents of uninitialised memory, alignment of an allocated buffer), rather that a specific change in 6.5.

But I'd suggest applying this to 6.6 and 6.7 kernels (6.5 makes sense, but it EOL, so less critical).
And leave older kernels like 6.1 as they are as problem is not visible, and it's not clear if this will cause any new issue.

popcornmix:
@P33M can you confirm this seems like a reasonable step?

P33M:
LGTM. The fact that this previously worked for #3148 supports the theory that it's some other change that broke the fix.

PR #5776, opened 2023-12-07
configs: rpi: Compile TSC2007 as module

timonsku:
Compile the TSC2007 touchscreen driver as a module and enable the optional IIO module.
Used by most newer Adafruit products.


timonsku:
Don't understand why the checkpatch is failing with missing commit description. The commit has a description?

pelwell:
I don't understand either, but we can ignore it if everything else is ok

timonsku:
I think it was missing extra line break, added now. Local checkpatch is happy now

pelwell:
I had to fix the order because it wasn't as it would be generated by make savedefconfig, but otherwise it's fine.

timonsku:
oh will keep that in mind, had it manually inserted for the commit after changing it with menuconfig on the test system

pelwell:
We should add it to the build checks.

PR #5777, opened 2023-12-08
input: goodix: Support output-only reset lines

pelwell:
Although the motivation behind initialising the reset GPIO as an input is sound, it prevents the driver from being used with an output-only GPIO controller. Fall back to trying to create it as an output.

See: https://forums.raspberrypi.com/viewtopic.php?t=361103

### DIFF
--- a/drivers/input/touchscreen/goodix.c
+++ b/drivers/input/touchscreen/goodix.c
@@ -1060,6 +1060,8 @@ retry_get_irq_gpio:
 
 	/* Get the reset line GPIO pin number */
 	gpiod = devm_gpiod_get_optional(dev, GOODIX_GPIO_RST_NAME, ts->gpiod_rst_flags);
+	if (IS_ERR(gpiod))
+		gpiod = devm_gpiod_get_optional(dev, GOODIX_GPIO_RST_NAME, GPIOD_OUT_LOW);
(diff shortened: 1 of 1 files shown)


pelwell:
@Gadgetoid Does this help? Wait about 45 minutes for the builds to complete then run `sudo rpi-update pulls/5777` to install a trial build.

Gadgetoid:
Build failed with:

```
../drivers/input/touchscreen/goodix.c: In function ‘goodix_get_gpio_config’:
../drivers/input/touchscreen/goodix.c:1064:76: error: ‘GPIOD_OUT’ undeclared (first use in this function); did you mean ‘GPIOD_IN’?
 1064 |                 gpiod = devm_gpiod_get_optional(dev, GOODIX_GPIO_RST_NAME, GPIOD_OUT);
      |                                                                            ^~~~~~~~~
      |                                                                            GPIOD_IN
../drivers/input/touchscreen/goodix.c:1064:76: note: each undeclared identifier is reported only once for each function it appears in
```

pelwell:
It should be good now - wait about half an hour for the build.

pelwell:
Once you've confirmed this works I'll update the branch, remove the draft status and merge. There's no rush, though.

PR #5778, opened 2023-12-08
usb: dwc3: disable bulk IN bursts for non-stream endpoints

P33M:
This fixes the broken case with disk-to-disk copies for me. It will slow USB mass-storage devices down "a bit", but UAS will be unaffected.

See #5753


P33M:
Closing as at least 1 user reports this isn't a robust fix.

PR #5781, opened 2023-12-11
drivers/gpu/drm/panel:fix waveshare panel software restart/shutdown display is abnormal

waveshare:


6by9:
Thanks for the patch - it looks to be the correct solution.

If being a stickler then it should call `drm_panel_disable` so that the framework gets notified of everything too, but as this is in an unload path it doesn't really matter.

6by9:
I have no intention of upstreaming this driver, so I'm really not worried over the checkpatch warning for no commit description.

PR #5785, opened 2023-12-11
firmware/psci: Pass given partition number through

pelwell:
Pi 5 uses BL31 as its armstub file, so the reset goes via PSCI. Parse any "reboot" parameter as a partition number to reboot into.

This code path is only used if reboot mode has been set to warm or soft, so the second patch enables warm reboot mode.


pelwell:
@timg236 This should fix PINN on Pi 5, at least in our downstream tree.

timg236:
Thanks will give it a try tomorrow. Need to check tryboot works as expected but IIRC BL31 didn’t modify that bit in the register

PR #5791, opened 2023-12-14
Update TSC2007 to common touch axis handling

timonsku:
The driver currently lacks most of the essential configuration properties like invert, swap, min max values etc.
This adds support for the common handling of these values via touchscreen.yaml.
I also removed/updated some of the old vendor specific properties that can be covered via the generic touchscreen.yaml


ladyada:
cc @makermelissa

drivers/input/touchscreen/tsc2007_iio.c
```diff
@@ -6,6 +6,7 @@
 
 #include <linux/i2c.h>
 #include <linux/iio/iio.h>
+#include <linux/input/touchscreen.h>
```
6by9, line 9: Is this needed?

drivers/input/touchscreen/tsc2007_core.c
```diff
@@ -141,9 +142,8 @@ static irqreturn_t tsc2007_soft_irq(int irq, void *handle)
 
 			rt = ts->max_rt - rt;
 
+			touchscreen_report_pos(ts->input, &ts->prop, tc.x, tc.y, false);
```
6by9, line 145: You've changed the order of this (which triggers ABS_X and ABS_Y events), wrt to BTN_TOUCH. Does that matter?

I suspect not as I believe `input_sync` is the trigger to do anything with all the events, but I couldn't say so definitively.

drivers/input/touchscreen/tsc2007_core.c
```diff
@@ -234,29 +234,20 @@ static int tsc2007_probe_properties(struct device *dev, struct tsc2007 *ts)
 	u32 val32;
 	u64 val64;
 
-	if (!device_property_read_u32(dev, "ti,max-rt", &val32))
```
6by9, line 237: Dropping these causes a regression for any existing users. AIUI you need to parse these, and then overwrite with values from the common parsing if present.

drivers/input/touchscreen/tsc2007_core.c
```diff
@@ -234,29 +234,20 @@ static int tsc2007_probe_properties(struct device *dev, struct tsc2007 *ts)
...
+	if (!device_property_read_u32(dev, "touchscreen-max-pressure", &val32))
```
6by9, line 237: `touchscreen-max-pressure` is read by [touchscreen_parse_properties](https://elixir.bootlin.com/linux/latest/C/ident/touchscreen_parse_properties). Is there a need to read it in two locations?

Set the default to MAX_12BIT before calling `touchscreen_parse_properties`, and it will overwrite that default if found.

drivers/input/touchscreen/tsc2007_core.c
```diff
@@ -353,10 +344,13 @@ static int tsc2007_probe(struct i2c_client *client,
 
 	input_set_capability(input_dev, EV_KEY, BTN_TOUCH);
 
+	/* Apply */
```
6by9, line 347: This and line 350 appear to only be irrelevant changes.

6by9 (commented):
Sorry for having missed this one.

DT binding change needs to be at least 2 patches - conversion of TXT to YAML as is, and then changing the functionality.

DT bindings are ABI, so you can't break existing users. Renaming parameters will cause regressions.

You've put yourself as maintainer in the binding - update MAINTAINERS too?

This is an unmodified driver from mainline, so I'm afraid we really do have to point you to upstream for a proper review and merge. We'll happily backport an accepted patch, but can't really take on the maintenance burden.

timonsku:
Sorry, stuff happened on my end as well. Thank your for the feedback and review!

> DT bindings are ABI, so you can't break existing users. Renaming parameters will cause regressions.

Right that is true. How is this typically handled, do old driver just never get updated or should it support both old and new bindings?

> You've put yourself as maintainer in the binding - update MAINTAINERS too?

I was really unsure how this is handled and haven't found guidance on how things would look if something did not have a yaml binding before. Should I put myself as a maintainer there? It felt off to put the current driver maintainer in there if they haven't authored the yaml binding.

> This is an unmodified driver from mainline, so I'm afraid we really do have to point you to upstream for a proper review and merge. We'll happily backport an accepted patch, but can't really take on the maintenance burden.

That I don't fully understood, could you expand what you mean with unmodified driver from mainline? Last time I asked the policy was that existing drivers could get a modifications accepted directly without going through mainline first and completely new drivers would have to go through mainline first. Did I misunderstand that?

drivers/input/touchscreen/tsc2007_iio.c
timonsku: it was because of the added property in the tsc2007.h that the iio driver also uses but in that case its I think better to have the header in the common include instead of the .c files respectively

drivers/input/touchscreen/tsc2007_core.c
timonsku: yes they will only be interpreted after the sync call so order of setting the report should be irrelevant.
https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/Documentation/input/input-programming.rst?id=HEAD#n121

6by9:
I totally missed your comment until you commented earlier - sorry.

> > DT bindings are ABI, so you can't break existing users. Renaming parameters will cause regressions.
> 
> Right that is true. How is this typically handled, do old driver just never get updated or should it support both old and new bindings?

The driver should support both. It doesn't really matter whether the generic or proprietary option is read first, just as long as the other is read if it fails.

> > You've put yourself as maintainer in the binding - update MAINTAINERS too?
> 
> I was really unsure how this is handled and haven't found guidance on how things would look if something did not have a yaml binding before. Should I put myself as a maintainer there? It felt off to put the current driver maintainer in there if they haven't authored the yaml binding.

If they were listed as maintainer of the text file, then it's reasonable to list them as the maintainer of the yaml conversion. The patch would be sent to them as well as the generic mailing list, so they are likely the one who would Ack or review it, hence have a chance to object.

> > This is an unmodified driver from mainline, so I'm afraid we really do have to point you to upstream for a proper review and merge. We'll happily backport an accepted patch, but can't really take on the maintenance burden.
> 
> That I don't fully understood, could you expand what you mean with unmodified driver from mainline? Last time I asked the policy was that existing drivers could get a modifications accepted directly without going through mainline first and completely new drivers would have to go through mainline first. Did I misunderstand that?

We would like to see the patches submitted to the relevant mailing list for review. Otherwise we can end up with rubbish left kicking around, and then we start getting conflicts if mainline changes. At least if the change looks likely to be accepted upstream, then we'll generally accept it downstream.
There are some changes that won't be acceptable upstream, and those we will tend to carry. Each extra patch becomes a support burden for each kernel release, which is why we're also upstreaming our own stuff.

timonsku:
> I totally missed your comment until you commented earlier - sorry.

no worries! Did not realize they notify, was preparing the changes you had requested and was gonna ping again after pushing them. Things were on hold on my end too.

> The driver should support both. It doesn't really matter whether the generic or proprietary option is read first, just as long as the other is read if it fails.

ok good thats mostly how I implemented it now, though I read both and the newer ones take precedence. Does that sound ok? It seemed like good behavior in case both equivalent properties are present.

> If they were listed as maintainer of the text file, then it's reasonable to list them as the maintainer of the yaml conversion. The patch would be sent to them as well as the generic mailing list, so they are likely the one who would Ack or review it, hence have a chance to object.

makes sense, that sounds reasonable

> We would like to see the patches submitted to the relevant mailing list for review. Otherwise we can end up with rubbish left kicking around, and then we start getting conflicts if mainline changes. At least if the change looks likely to be accepted upstream, then we'll generally accept it downstream. There are some changes that won't be acceptable upstream, and those we will tend to carry. Each extra patch becomes a support burden for each kernel release, which is why we're also upstreaming our own stuff.

Ok gotcha, will proceed like that then.

PR #5794, opened 2023-12-15
Allow disabling FIFOs on PL011 UART

ITachiLab:
The BCM2835 PL011 implementation assumes that Rx/Tx FIFOs should be always enabled. Although this is desirable for most of the time, there are rare occasions when FIFOs introduce unwanted delay on the line. The delay is unnoticeable for the common UART use cases but it can lead to communication failures in UART-based protocols expecting fast responses, such as eBUS.

The solution is to add a new `disable-fifos` property to Pi DTBs and use the presence of that property to forcefully set FIFOs size to 1, and prevent FIFOs from being enabled in the PL011 driver.

Example issues that might be originating from the delay introduced by FIFO buffering:

- https://forums.raspberrypi.com/viewtopic.php?t=170945&p=1095466
- john30/ebusd#91

Tested with **ebusd** on Raspberry Pi Zero W v1.1. I was able to communicate with eBUS's participants after adding `dtoverlay=uart0,disable_fifos` to `config.txt`. If it's not appropriate to keep this flag in the "uart0" overlay, a dedicated overlay for the flag can be created instead.

## Results on logic analyzer

For a simple test I tied TX and RX pins together, and I ran a simple program written in C in which I'm writing a single character, reading it back, and then writing and reading it again. This way I can test if there are any delays in both write and read queues.

FIFO is enabled

The first screenshot shows that when **FIFOs are enabled**, the gap between consecutive write/read/write operations is roughly **13 milliseconds**.

FIFO is disabled

The same test repeated when **FIFOs are disabled** shows that the gap is now only **30 microseconds**. This is 430 times smaller than when FIFOs were enabled.

The code I used for testing:

```c
#include <stdio.h>
#include <fcntl.h>
#include <termios.h>
#include <unistd.h>

int main() {
    int port = open("/dev/ttyAMA0", O_RDWR);

    if (port < 0) {
        printf("Cannot open port\n");
        return -1;
    }

    struct termios tty;

    tty.c_cflag &= ~(PARENB | CSTOPB | CRTSCTS);
    tty.c_cflag |= CREAD | CLOCAL | CS8;

    tty.c_lflag &= ~(ICANON | ECHO | ECHOE | ECHONL | ISIG);

    tty.c_iflag &= ~(IGNBRK | BRKINT | PARMRK | ISTRIP | INLCR | IGNCR 
            | ICRNL | IXON | IXOFF | IXANY);

    tty.c_oflag &= ~(OPOST | ONLCR);

    tty.c_cc[VTIME] = 10;
    tty.c_cc[VMIN] = 0;

    cfsetispeed(&tty, B2400);

    if (tcsetattr(port, TCSANOW, &tty) != 0) {
        printf("Cannot write attributes\n");
        close(port);

        return -1;
    }

    char read_buffer[1];
    unsigned char msg[] = { 'a' };
    int read_chars;

    write(port, msg, 1);
    read_chars = read(port, read_buffer, 1);

    write(port, msg, 1);
    read(port, read_buffer, 1);

    close(port);

    printf("Read: %d\n", read_chars);

    return 0;
}
```

### DIFF
--- a/Documentation/devicetree/bindings/serial/pl011.yaml
+++ b/Documentation/devicetree/bindings/serial/pl011.yaml
@@ -106,6 +106,12 @@ properties:
       CTS-induced TX lockup.
     type: boolean
 
+  disable-fifos:
+    description:
+      Disable FIFOs (character mode) that is, the FIFOs become 1-byte-deep
+      holding registers.
+    type: boolean
+
 required:
   - compatible
   - reg
--- a/arch/arm/boot/dts/overlays/README
+++ b/arch/arm/boot/dts/overlays/README
@@ -4542,7 +4542,7 @@ Params: <None>
 
 
 Name:   uart0
-Info:   Change the pin usage of uart0
+Info:   Change the pin usage and parameters of uart0
 Load:   dtoverlay=uart0,<param>=<val>
 Params: txd0_pin                GPIO pin for TXD0 (14, 32 or 36 - default 14)
 
@@ -4550,6 +4550,7 @@ Params: txd0_pin                GPIO pin for TXD0 (14, 32 or 36 - default 14)
 
         pin_func                Alternative pin function - 4(Alt0) for 14&15,
                                 7(Alt3) for 32&33, 6(Alt2) for 36&37
+        disable_fifos           Disable FIFOs (character mode)
 
 
 Name:   uart0-pi5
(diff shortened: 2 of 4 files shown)


ITachiLab:
CC: @pelwell (Based on recent changes done to the same source files)

pelwell:
1. I'm not particularly interested in taking on a niche patch such as this, given that we will have to maintain it.
2. "The BCM2835 PL011 implementation assumes that Rx/Tx FIFOs should be always enabled." Do other drivers make FIFO usage conditional in some way?
3. You don't need to patch `sbsa_uart_probe` - that code path is not used on Raspberry Pis.

PR #5798, opened 2023-12-18
input: edt-ft5x06: Correct prefix length in snprintf

6by9:
snprintf takes the length of the array that we can print into, and has to fit the NULL terminator in there too.
Printing the prefix is generally "12-3456 " which is 8 desired characters (the length of EDT_NAME_PREFIX_LEN) and the NULL. The space is therefore being truncated to fit the NULL in.

Increase the length snprintf is allowed to use.

### DIFF
--- a/drivers/input/touchscreen/edt-ft5x06.c
+++ b/drivers/input/touchscreen/edt-ft5x06.c
@@ -966,7 +966,7 @@ static int edt_ft5x06_ts_identify(struct i2c_client *client,
 	char *model_name = tsdata->name;
 	char *fw_version = tsdata->fw_version;
 
-	snprintf(model_name, EDT_NAME_PREFIX_LEN, "%s ", dev_name(&client->dev));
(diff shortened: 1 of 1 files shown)


JamesH65:
The string being written into doesn't seem to have space for the terminator if both the prefix and name are the maximum length they can be. If I read it right.

6by9:
> The string being written into doesn't seem to have space for the terminator if both the prefix and name are the maximum length they can be. If I read it right.

`snprintf(model_name, EDT_NAME_PREFIX_LEN + 1, "%s ", dev_name(&client->dev));` will print max EDT_NAME_PREFIX_LEN chars and the NULL. 
`model_name += strlen(model_name);` will move forward by max EDT_NAME_PREFIX_LEN chars.

Therefore `snprintf(model_name, EDT_NAME_LEN, "EP0%i%i0M09", rdbuf[0] >> 4, rdbuf[0] & 0x0F);` will move forward by max EDT_NAME_LEN chars (including the NULL), so we don't run off the end of `char name[EDT_NAME_PREFIX_LEN + EDT_NAME_LEN];`

Issue #5799, opened 2023-12-18
Disable `CONFIG_IRQSOFF_TRACER` for better WireGuard performance


cyyself:
### Describe the bug

Recently, I found the WireGuard Performance on Respbian is much worse than OpenWRT using [my wireguard benchmark script](https://github.com/cyyself/wg-bench). After some investigations, I found the performance was hindered by `CONFIG_PREEMPT` and `CONFIG_IRQSOFF_TRACER`, which hindered from ~1.1Gbps to only ~400Mbps. As Raspbian may be used as Desktop OS, we can only turn off `CONFIG_IRQSOFF_TRACER`. I also found that Debian has not enabled this configuration by default. IMHO such configuration which hindered too much performance should be removed. If we only remove `CONFIG_IRQSOFF_TRACER=y` but keep `CONFIG_PREEMPT=y`, we can also see the performance improvements from ~390Mbps to ~680Mbps. You can also see my investigation procedures [here](https://github.com/cyyself/wg-bench/issues/5).

### Steps to reproduce the behaviour

Using my WireGuard benchmark:

```bash
sudo apt install wireguard-tools iperf3 git
git clone https://github.com/cyyself/wg-bench.git
cd wg-bench
sudo ./setup-netns.sh
sudo ./benchmark.sh
sudo ./clean-up.sh
```

Pi 4 should reach ~680Mbps instead of ~390Mbps.

### Device (s)

Raspberry Pi 4 Mod. B

### System

`/etc/rpi-issue`:

```console
Raspberry Pi reference 2023-12-11
Generated using pi-gen, https://github.com/RPi-Distro/pi-gen, 2acf7afcba7d11500313a7b93bb55a2aae20b2d6, stage2
```

`vcgencmd version`:

```console
Oct 17 2023 15:39:16 
Copyright (c) 2012 Broadcom
version 30f0c5e4d076da3ab4f341d88e7d505760b93ad7 (clean) (release) (start)
```

`uname -a` (self-compiled):

```console
Linux raspberrypi 6.1.68-v8+ #29 SMP PREEMPT Mon Dec 18 22:58:01 CST 2023 aarch64 GNU/Linux
```

PR #5800, opened 2023-12-18
configs: rpi: Disable CONFIG_IRQSOFF_TRACER by default

cyyself:
This configuration hindered performance by ~74% measured from RPI 4B from ~680Mbps to ~390Mbps when benchmarking wireguard locally using netns and iperf3. Remove it by default for better performance.

Related issue: #5799


popcornmix:
@pelwell I'm inclined to agree with this.

My Ubuntu install has this disabled:
```
/boot/config-6.2.0-39-generic:# CONFIG_IRQSOFF_TRACER is not set
```

pelwell:
No objections here - it's been enabled since rpi-3.17.y back in 2013!

popcornmix:
Thanks for investigating this - hopefully it will benefit some other use cases.

PR #5803, opened 2023-12-19
ARM: dts: bcm2712-rpi-5-b: Allow RTC to be disabled

pelwell:
Add a dtparam "rtc", so that "dtparam=rtc=off" can be used to disable the Pi 5's onboard RTC.

See: https://forums.raspberrypi.com/viewtopic.php?t=361813


popcornmix:
LGTM

timg236 (approved):
LGTM

PR #5805, opened 2023-12-21
Update the 3B LED support

pelwell:
As reported in https://forums.raspberrypi.com/viewtopic.php?p=2171315#p2171315, the Pi 3B LEDs were broken in rpi-6.6.y. This is because the downstream brcmvirt-gpio driver (used to optimise the activity LED handling) needed some work to make it 6.6-compatible. It's also been brought slightly more up to date in general.


popcornmix:
LGTM

PR #5818, opened 2024-01-03
drivers/gpu/drm/panel：Fixed display issues with several screens

waveshare:
1.  Modify the DSI mode to fix the problem that 7.9inch cannot be displayed
2.  Modified the timing of 11.9inch to fix the issue that 11.9inch was displayed abnormally


drivers/gpu/drm/panel/panel-waveshare-dsi.c
```diff
@@ -340,9 +340,8 @@ static int ws_panel_probe(struct i2c_client *i2c,
 	 */
 	drm_panel_add(&ts->base);
 
-	ts->dsi->mode_flags = (MIPI_DSI_MODE_VIDEO |
-			   MIPI_DSI_MODE_VIDEO_SYNC_PULSE |
-			   MIPI_DSI_MODE_LPM);
+	ts->dsi->mode_flags =  MIPI_DSI_MODE_VIDEO_HSE | MIPI_DSI_MODE_VIDEO |
+			   MIPI_DSI_CLOCK_NON_CONTINUOUS;
```
6by9, line 344: The need for this surprises as I have one of your 7.9" panels and it works fine on a Pi4. Is it another variant of Pi that has issues?

The ICN6211 datasheet says:
> ICN6211 supports Non-Burst Mode with Sync Pulses, Non-Burst Mode with Sync Events and Burst mode

so changing from Sync Pulses to Sync Events (not setting MIPI_DSI_MODE_VIDEO_SYNC_PULSE) should be irrelevant.
However also note that the vc4_dsi driver doesn't look at the VIDEO_SYNC_PULSE or VIDEO_HSE flags. Pi5 and RP1 does support SYNC_PULSE mode

MIPI_DSI_CLOCK_NON_CONTINUOUS will result in the clock lane dropping to LP during blanking (where time allows). That shouldn't make any difference as long as you are providing the ICN6211 a reference clock rather than trying to clock it off the DSI signal.

pelwell:
I presume you have no problem with this being merged, @6by9?

6by9 (commented):
Thanks for the patch.

I have no significant issue with the patches, but the change of mode_flags looks odd and would like to understand it.

drivers/gpu/drm/panel/panel-waveshare-dsi.c
waveshare: Thank you for your further additions.
Yes, the 7.9 inch panel works fine on the pi4 but not on the pi5.
After modifying the mode, it ran fine when tested on both pi4 and pi5.

pelwell: That's not a very encouraging answer - it reads like "we changed some stuff until it worked on Pi 5, but we don't understand why". If that's not the case, please explain further.

6by9: I've just tried the 7.9" panel out on my Pi5 (actually on a 6.6.9 kernel, but that's irrelevant).

You're right that the existing driver results in no output on Pi5.
Swapping from `MIPI_DSI_MODE_VIDEO_SYNC_PULSE` to `MIPI_DSI_MODE_VIDEO_HSE`, or just removing `MIPI_DSI_MODE_VIDEO_SYNC_PULSE` (to get sync events) both seem to work. That probably means that the video timings don't quite allow both HSYNC_START and HSYNC_END packets to be put out the bus.

Switching to `MIPI_DSI_CLOCK_NON_CONTINUOUS` is not needed.

In passing I note that the refresh rate for the 7.9" panel appears to come out at 73.71Hz. I did my best to copy the timings from those selected by your binary blob drivers, but that seems an odd refresh rate.

TBH I'm grateful for your input in actually making this driver do what is required. It would be nice to understand the full reasoning, but also happy enough if it works. I'm never intending to upstream this driver (others could as it has the relevant Signed-off-by tags), so it can be a little hacky if necessary.

samdowrickstr (approved):

drivers/gpu/drm/panel/panel-waveshare-dsi.c
pelwell: I received no response to my request for explanation from @waveshare, but do you now approve this PR, @6by9?

6by9: It's only for their panels, so adding the random MIPI_DSI_MODE_VIDEO_HSE and MIPI_DSI_CLOCK_NON_CONTINUOUS flags makes little difference to the world.
Likewise the odd refresh rate of 73.71Hz is quirky, but causes no harm anywhere else.

Merge away.

waveshare: I apologize for my delayed response, and I appreciate your understanding and tolerance.

The addition of MIPI_DSI_CLOCK_NON_CONTINUOUS was referenced from the panel-simple.c driver.

MIPI_DSI_MODE_VIDEO_SYNC_PULSE works well on the Raspberry Pi 4, but we are uncertain why it is not functioning on the Pi 5.

As for the refresh rate of 73.71Hz for the 7.9-inch panel, this specific refresh rate was determined during the development phase for debugging purposes. The stability of the display on the special panel is correlated with the main control clock frequency. It operates smoothly near the clock frequency of the Raspberry Pi's main control, ensuring that operations like sleep and wake do not encounter abnormal bugs. Therefore, at this frequency, the corresponding refresh rate is calculated as 73.71Hz.

We will continue to improve our documentation, and any additional discoveries will be updated and explained accordingly. Thank you once again.

6by9: Thanks for the update.

Quirks like your 73.1Hz are almost inevitable, but it's nice to know that they are intentional rather than an oversight.

`MIPI_DSI_CLOCK_NON_CONTINUOUS` isn't necessarily wrong, it was just a change that seemed unnecessary.

I will try to find time to validate `MIPI_DSI_MODE_VIDEO_SYNC_PULSE` on Pi5 and ensure it is compliant with the DSI spec, but I'm expecting it to be. (I've been looking at it on other panels recently).
The flag is ignored on Pi4, so it won't have mattered whether the flag was incorrect previously.

Darmur:
after back-porting the waveshare patch on kernel 6.1.69, the display started to work fine.

however, the touch-panel doesn't work as it should, it detects the touch (i can see the mouse cursor popping-up or moving), but it's not aligned at all with the position where I touch. I've tried all possible combinations of swapxy, invy and invx, declaring explicitly the value 0 or 1. none of the combinations made the touch working properly, most of the time the mouse cursor is all the way towards one of the edges (left or right).

the touch-panel works fine with the image provided by waveshare, that includes their own drivers.

6by9:
#6017 ?

I only have the 2.8", 7.9", and 10.1" panels. Transcription errors may well exist with the other panels, but I'm very grateful that Waveshare are actually contributing fixes, and are now referencing these drivers in their wiki (eg https://www.waveshare.com/wiki/2.8inch_DSI_LCD#For_Bookworm_and_Bullseye_System)

Darmur:
I tried the fix from the PR #6017 

I can confirm it solves the problem

I had to add some parameters to the overlay declaration, for having the correct orientation of the axes with the display rotated horizontally

`dtoverlay=vc4-kms-dsi-waveshare-panel,11_9_inch,rotation=270,swapxy=0,invx=0`

Issue #5817, opened 2024-01-02
Wolfson wm8731 audio codec in Pi 5 does not work (works with Pi 4!)


rafael2k:
### Describe the bug

The sBitx radio uses a Pi 4 and a wm8731 audio codec. With the Pi 5, the kernel shows strange messages (same config.txt options used), and the sound card does not work.

### Steps to reproduce the behaviour

Just boot the radio with a Pi 5. Error messages appear every 10s.
```
[   21.818092] designware-i2s 1f000a4000.i2s: ASoC: error at snd_soc_dai_set_bclk_ratio on 1f000a4000.i2s: -22
[   21.818113]  AudioInjector audio: ASoC: error at snd_soc_link_hw_params on AudioInjector audio: -22
[   21.818458]  AudioInjector audio: ASoC: error at __soc_pcm_hw_params on AudioInjector audio: -22
[   21.821546] designware-i2s 1f000a4000.i2s: ASoC: error at snd_soc_dai_set_bclk_ratio on 1f000a4000.i2s: -22
[   21.821553]  AudioInjector audio: ASoC: error at snd_soc_link_hw_params on AudioInjector audio: -22
[   21.821567]  AudioInjector audio: ASoC: error at __soc_pcm_hw_params on AudioInjector audio: -22
[   31.650617] designware-i2s 1f000a4000.i2s: ASoC: error at snd_soc_dai_set_bclk_ratio on 1f000a4000.i2s: -22
[   31.650625]  AudioInjector audio: ASoC: error at snd_soc_link_hw_params on AudioInjector audio: -22
[   31.650638]  AudioInjector audio: ASoC: error at __soc_pcm_hw_params on AudioInjector audio: -22
[   31.651472] designware-i2s 1f000a4000.i2s: ASoC: error at snd_soc_dai_set_bclk_ratio on 1f000a4000.i2s: -22
[   31.651478]  AudioInjector audio: ASoC: error at snd_soc_link_hw_params on AudioInjector audio: -22
[   31.651815]  AudioInjector audio: ASoC: error at __soc_pcm_hw_params on AudioInjector audio: -22
[   41.684473] designware-i2s 1f000a4000.i2s: ASoC: error at snd_soc_dai_set_bclk_ratio on 1f000a4000.i2s: -22
[   41.684483]  AudioInjector audio: ASoC: error at snd_soc_link_hw_params on AudioInjector audio: -22
[...]

PR #5819, opened 2024-01-03
ASoC: dwc: Defer bclk_ratio handling to hw_params

pelwell:
bclk_ratio is only a factor in clock producer mode, and needs to override the default value of num_channels * sample_size. Move the bclk_ratio handling into the hw_params method, only latching the value in set_bclk_ratio, to address both of those matters.

See: https://github.com/raspberrypi/linux/issues/5817


pelwell:
This seems to be working as expected, and fixes #5817.

PR #5827, opened 2024-01-04
drm/vc4: Add 2712 support to vc4_plane_async_set_fb

6by9:
vc4_plane_async_set_fb directly overwrites the plane address in the dlist entry, but hadn't been updated for the GEN6 / 2712 dlist format, corrupting the address in the process.

Add support for the 2712 dlist format to the function.

Fixes: 1ab1fbbb7e76 ("drm/vc4: hvs: Support BCM2712 HVS")

### DIFF
--- a/drivers/gpu/drm/vc4/vc4_plane.c
+++ b/drivers/gpu/drm/vc4/vc4_plane.c
@@ -1877,7 +1877,7 @@ static int vc6_plane_mode_set(struct drm_plane *plane,
 				 * The UPM buffer will be allocated in
 				 * vc6_plane_allocate_upm().
 				 */
-				VC4_SET_FIELD(upper_32_bits(paddr) & 0xf,
+				VC4_SET_FIELD(upper_32_bits(paddr) & 0xff,
 					      SCALER6_PTR0_UPPER_ADDR));
 
 		/* Pointer Word 1 */
@@ -2079,7 +2079,8 @@ void vc4_plane_async_set_fb(struct drm_plane *plane, struct drm_framebuffer *fb)
 {
 	struct vc4_plane_state *vc4_state = to_vc4_plane_state(plane->state);
 	struct drm_gem_dma_object *bo = drm_fb_dma_get_gem_obj(fb, 0);
-	uint32_t addr;
+	struct vc4_dev *vc4 = to_vc4_dev(plane->dev);
+	dma_addr_t dma_addr = bo->dma_addr + fb->offsets[0];
 	int idx;
 
 	if (!drm_dev_enter(plane->dev, &idx))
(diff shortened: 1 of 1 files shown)


drivers/gpu/drm/vc4/vc4_plane.c
```diff
@@ -2077,19 +2078,38 @@ void vc4_plane_async_set_fb(struct drm_plane *plane, struct drm_framebuffer *fb)
...
+		writel(value, &vc4_state->hw_dlist[vc4_state->ptr0_offset[0]]);
+		vc4_state->dlist[vc4_state->ptr0_offset[0]] = value;
+
+		value = lower_32_bits(dma_addr);
+		writel(value, &vc4_state->hw_dlist[vc4_state->ptr0_offset[0] + 1]);
+		vc4_state->dlist[vc4_state->ptr0_offset[0] + 1] = value;
+	} else {
+		u32 addr;
+
+		addr = (u32)dma_addr;
```
pelwell, line 2111: `lower_32_bits(dma_addr)`?

drivers/gpu/drm/vc4/vc4_plane.c
```diff
@@ -2077,19 +2078,38 @@ void vc4_plane_async_set_fb(struct drm_plane *plane, struct drm_framebuffer *fb)
...
+		 * scanout will start from this address as soon as the FIFO
+		 * needs to refill with pixels.
+		 */
+		writel(addr, &vc4_state->hw_dlist[vc4_state->ptr0_offset[0]]);
+
+		/* Also update the CPU-side dlist copy, so that any later
+		 * atomic updates that don't do a new modeset on our plane
+		 * also use our updated address.
+		 */
+		vc4_state->dlist[vc4_state->ptr0_offset[0]] = addr;
```
pelwell, line 2123: It's surprising there isn't a helper function to update `hw_dlist` and the shadow `dlist`.

drivers/gpu/drm/vc4/vc4_plane.c
6by9: OK. I was just leaving the code largely as was and ignoring the potential for overflow.

drivers/gpu/drm/vc4/vc4_plane.c
```diff
@@ -2077,19 +2078,38 @@ void vc4_plane_async_set_fb(struct drm_plane *plane, struct drm_framebuffer *fb)
...
+		u32 value;
 
-	/* Also update the CPU-side dlist copy, so that any later
-	 * atomic updates that don't do a new modeset on our plane
-	 * also use our updated address.
-	 */
-	vc4_state->dlist[vc4_state->ptr0_offset[0]] = addr;
+		value = vc4_state->dlist[vc4_state->ptr0_offset[0]] &
+					~SCALER6_PTR0_UPPER_ADDR_MASK;
+		value |= VC4_SET_FIELD(upper_32_bits(dma_addr) & 0xf,
```
popcornmix: shouldn't this be ` & 0xff`? There are 8 bits of upper address.

drivers/gpu/drm/vc4/vc4_plane.c
pelwell: The two are functionally equivalent, but my version looks nicer and avoids the visible cast.

drivers/gpu/drm/vc4/vc4_plane.c
pelwell: It's good to be ready for the 32GB variant.

drivers/gpu/drm/vc4/vc4_plane.c
6by9: This it the only place that the dlist gets mangled in this way.

The fragments for each plane get created in vc[46]_plane_set_mode.
vc4_hvs_atomic_flush then assembles the fragments into a complete list in the dlist memory.

This async update is the only place that does both in one hit.

Reviewing the code it feels a little fragile based on a number of major assumptions.
AIUI it only works on the primary plane, which would normally be the bottom-most but that isn't guaranteed. vc4_state->ptr0_offset[0] is the offset within the fragment, so it only works as long as it is the bottom-most plane and hence first in the dlist.

drivers/gpu/drm/vc4/vc4_plane.c
6by9: Copied from https://github.com/raspberrypi/linux/blob/rpi-6.1.y/drivers/gpu/drm/vc4/vc4_plane.c#L1862-L1869

I'll check to see what the docs say...

6by9: > It's good to be ready for the 32GB variant.

This is actually the iommu VA, so it depends on where Nick has mapped it in the virtual address space as to how big the value is.

drivers/gpu/drm/vc4/vc4_plane.c
pelwell: There's also this: https://github.com/raspberrypi/linux/blob/e2020fe9bd678ba229b9448bb72a111f5632bc9f/drivers/gpu/drm/vc4/vc4_plane.c#L2170-L2186

6by9: As amending that one needs to be a separate patch that is ideally applied before any 2712 stuff starts to be merged, can I take it as a task to be dealt with in the mainline patches?

pelwell: Of course.

bluestang2006:
@6by9 I've tested your branch recently and it does fix the issue at hand. I've added some more info to the issue report...the corruption happens when vsync is disabled only.

6by9:
Had missed that I hadn't updated this after the review comments.
Done now to address popcornmix's comment about upper address bit masking. pelwell's comment on helper function will be left to deal with when upstreaming.

popcornmix (approved):

6by9:
Are we planning on another 6.1 release, and hence does this warrant a backport to 6.1?

pelwell:
Yes we are, and yes it does. We need to bump the 6.1 stable so that it gets feature parity with the upcoming 6.6 stable.

6by9:
> Yes we are, and yes it does. We need to bump the 6.1 stable so that it gets feature parity with the upcoming 6.6 stable.

OK, I'll create a backport PR.

Checking the email thread with Igalia, there's a 2nd cleanup required in the async path too, but less critical.

PR #5833, opened 2024-01-05
Rpi 6.7.y iommu support

6by9:
Based the changes on the matching upstream commits, but not tested on hardware yet (it does build now).

### DIFF
--- a/drivers/iommu/Kconfig
+++ b/drivers/iommu/Kconfig
@@ -487,4 +487,11 @@ config SPRD_IOMMU
 
 	  Say Y here if you want to use the multimedia devices listed above.
 
+config BCM2712_IOMMU
+       tristate "BCM2712 IOMMU driver"
+       depends on ARM64 && ARCH_BCM
+       select IOMMU_API
+       help
+	 IOMMU driver for BCM2712
+
 endif # IOMMU_SUPPORT
--- a/drivers/iommu/Makefile
+++ b/drivers/iommu/Makefile
@@ -29,3 +29,4 @@ obj-$(CONFIG_VIRTIO_IOMMU) += virtio-iommu.o
 obj-$(CONFIG_IOMMU_SVA) += iommu-sva.o io-pgfault.o
 obj-$(CONFIG_SPRD_IOMMU) += sprd-iommu.o
 obj-$(CONFIG_APPLE_DART) += apple-dart.o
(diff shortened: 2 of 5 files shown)


6by9:
@njhollinghurst FYI

njhollinghurst:
I am surprised that the `count` argument is not used (in the exynos example). It looks like it ought to be important!

6by9:
> I am surprised that the `count` argument is not used (in the exynos example). It looks like it ought to be important!

The commit text does include
> Trivially update map/unmap to the new interface, which is quite happy for drivers to still process just one page per call.

so I'm assuming it is processing one page per call and hence can ignore `count`.
I was going to try running it today to see if it works.

njhollinghurst:
I see: so that return value suffices to say "I only did one of the pages you asked for"?

Anyway, I think the quick hack / fix-up might be as simple as inserting `bytes *= count;` at the top of the function (since we already have a loop to break the given page size into 4K basic units). That might be more efficient.

I suppose we should handle the (very unexpected) case where only some of the pages fit below `APERTURE_TOP`.

6by9:
> Anyway, I think the quick hack / fix-up might be as simple as inserting `bytes *= count;` at the top of the function (since we already have a loop to break the given page size into 4K basic units). That might be more efficient.

Probably, I just haven't looked into the API closely enough to determine the detail of it.

njhollinghurst:
Not yet tested... but I might try adding this before `/* large page and write enable flags */`. It doesn't really matter if it goes before or after computing `align`.
```
		/* Handle multiple contiguous pages up to APERTURE_TOP */
		if (bytes && iova + count * bytes > APERTURE_TOP)
			count = (APERTURE_TOP - iova) / bytes;
		bytes *= count;
```
By modifying `bytes` we modify both the loop over `p` and also the return value; hopefully correctly!

njhollinghurst:
I expect the fix-up is fine as it is!  I might try to optimize the count > 1 case at a later date in a separate PR.

6by9:
Rebased. I'll do a test, but @njhollinghurst has said it works for him.

6by9:
Checked this tree, and the dlist pointer words include an address up at 0xAxxxxxxxx, which is correct for the iommu virtual address.
libcamera also works, so that checks out RP1 for the CFE, and ISPBE.

I think that makes me happy with this - dropping the Draft tag.

njhollinghurst:
LGTM  (I'll have a go at my subsequent "optimization" later).

PR #5838, opened 2024-01-08
watchdog: bcm2835_wdt: Fix WDIOC_SETTIMEOUT handling

popcornmix:
Users report about the unexpected behavior for setting timeouts above 15 sec on Raspberry Pi. According to watchdog-api.rst the ioctl WDIOC_SETTIMEOUT shouldn't fail because of hardware limitations. But looking at the code shows that max_timeout based on the register value PM_WDOG_TIME_SET, which is the maximum.

Since 664a39236e71 ("watchdog: Introduce hardware maximum heartbeat in watchdog core") the watchdog core is able to handle this problem.

This fix has been tested with watchdog-test from selftests.

Link: https://bugzilla.kernel.org/show_bug.cgi?id=217374
Fixes: 664a39236e71 ("watchdog: Introduce hardware maximum heartbeat in watchdog core")

### DIFF
--- a/drivers/watchdog/bcm2835_wdt.c
+++ b/drivers/watchdog/bcm2835_wdt.c
@@ -36,6 +36,7 @@
 
 #define SECS_TO_WDOG_TICKS(x) ((x) << 16)
 #define WDOG_TICKS_TO_SECS(x) ((x) >> 16)
+#define WDOG_TICKS_TO_MSECS(x) ((x) * 1000 >> 16)
 
 struct bcm2835_wdt {
 	void __iomem		*base;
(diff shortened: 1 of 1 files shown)


popcornmix:
See: https://lore.kernel.org/linux-watchdog/979cc901-6583-4e7c-b181-1b506c29193c@broadcom.com/T/

Ping @lategoodbye

This is a simpler alternative to https://github.com/raspberrypi/linux/pull/5664
to get longer watchdog timeouts.

In testing this works fine, with, say 120s timeout (longer than the single-shot 15s duration of watchdog hardware).

lategoodbye:
Not sure what's the intension of the ping, but my mainline patch has been applied for watchdog-next. So the patch will be in Linux 6.8 and then hopefully backported. Some subsystems need a lot of time.

lategoodbye:
FWIW the original patch is now in Mainline stable: 6.1.75 / 6.6.14

popcornmix:
Great - I'll drop cherry-picked commit on next update.

PR #5840, opened 2024-01-09
Patches for the DSI touchscreen being shifted by 1 pixel

6by9:
https://forums.raspberrypi.com/viewtopic.php?t=361866

I've backported the patches from mainline, but one actually looks to break things and hence has a revert.
It then also passes the DPI timings to the TC358762, and corrects the panel timings.


6by9:
Closing as obsolete on 6.1

PR #5841, opened 2024-01-10
Add link freq Control for IMX519 and Arducam64MP

ArduCAM:


drivers/media/i2c/arducam_64mp.c
```diff
@@ -2291,6 +2297,19 @@ static int arducam_64mp_init_controls(struct arducam_64mp *arducam_64mp)
 						     ARDUCAM_64MP_PIXEL_RATE,
 						     ARDUCAM_64MP_PIXEL_RATE, 1,
 						     ARDUCAM_64MP_PIXEL_RATE);
+	if (arducam_64mp->pixel_rate)
+		arducam_64mp->pixel_rate->flags |= V4L2_CTRL_FLAG_READ_ONLY;
```
6by9: This happens automatically due to the core - https://elixir.bootlin.com/linux/latest/source/drivers/media/v4l2-core/v4l2-ctrls-defs.c#L1512

drivers/media/i2c/arducam_64mp.c
```diff
@@ -2291,6 +2297,19 @@ static int arducam_64mp_init_controls(struct arducam_64mp *arducam_64mp)
...
+	/* LINK_FREQ is also read only */
+	link_freq_max = ARRAY_SIZE(arducam_64mp_link_freq_menu) - 1;
+	arducam_64mp->link_freq =
+		v4l2_ctrl_new_int_menu(ctrl_hdlr, &arducam_64mp_ctrl_ops,
+				       V4L2_CID_LINK_FREQ,
+				       link_freq_max,
+				       0,
+				       arducam_64mp_link_freq_menu);
+	if (arducam_64mp->link_freq)
+		arducam_64mp->link_freq->flags |= V4L2_CTRL_FLAG_READ_ONLY;
```
6by9: If you're only needing the `struct v4l2_ctrl` for setting the READ_ONLY flag, then store it as a local to this function rather than in the sensor's state.

drivers/media/i2c/imx519.c
```diff
@@ -1836,6 +1841,17 @@ static int imx519_init_controls(struct imx519 *imx519)
...
+		imx519->pixel_rate->flags |= V4L2_CTRL_FLAG_READ_ONLY;
+
+	/* LINK_FREQ is also read only */
+	imx519->link_freq =
+		v4l2_ctrl_new_int_menu(ctrl_hdlr, &imx519_ctrl_ops,
+				       V4L2_CID_LINK_FREQ,
+				       ARRAY_SIZE(imx519_link_freq_menu) - 1, 0,
+				       imx519_link_freq_menu);
+	if (imx519->link_freq)
+		imx519->link_freq->flags |= V4L2_CTRL_FLAG_READ_ONLY;
```
6by9: Ditto to the above.

drivers/media/i2c/arducam_64mp.c
ArduCAM: Got it, I was wondering why the read-only flag wasn't added before, but recently imx477 had the same commit, so I guess it's redundant too.

drivers/media/i2c/arducam_64mp.c
ArduCAM: Done.

drivers/media/i2c/imx519.c
ArduCAM: Done.

6by9:
LGTM - thanks for the updates.

You're right it did get added to imx477 back in Nov 2023. I can't find the PR for it, as I would have made the same comment had I seen the PR.

pelwell:
> I can't find the PR for it,

It looks like it was a direct push.

PR #5842, opened 2024-01-10
Remove 2712 SDHCI accessors

P33M:
Prompted by an email discusson from a freeBSD user, there doesn't seem to be a need to guard register accesses on C1 parts.

A full apt update and fio benchmark went fine with these removed, whereas if these were required, the controller usually falls over when attempting to mount the filesystem.


pelwell:
Should the second commit be in its own PR?

PR #5845, opened 2024-01-11
Underscan, overscan, wombling free

njhollinghurst:
Previously the RP1 VEC driver supported a fixed set of 8 "canned" video modes.
There appears to be some demand for different active widths and heights, to fine-tune overscan compensation.

This change supports arbitrary widths and heights within the maximum available space. E.g. for BBC Micro fans:
```
xrandr --newmode 640x256 15.429 640 752 824 987 256 274 277 313
```
It still requires the pixel rate to be one of {13.5MHz, 15.429MHz} and the line and frame rates to match one of the known TV standards (with a little leeway for rounding errors). ~At present it ignores [hv]sync_start and [hv]sync_end (the middle two numbers of each group of four), and always centres the image on the screen.~

### DIFF
--- a/drivers/gpu/drm/rp1/rp1-vec/rp1_vec.c
+++ b/drivers/gpu/drm/rp1/rp1-vec/rp1_vec.c
@@ -19,7 +19,6 @@
 #include <linux/list.h>
 #include <linux/platform_device.h>
 #include <linux/clk.h>
-#include <linux/printk.h>
 #include <linux/console.h>
 #include <linux/debugfs.h>
 #include <linux/uaccess.h>
@@ -208,26 +207,26 @@ static void rp1vec_connector_destroy(struct drm_connector *connector)
 static const struct drm_display_mode rp1vec_modes[4] = {
 	{ /* Full size 525/60i with Rec.601 pixel rate */
 		DRM_MODE("720x480i", DRM_MODE_TYPE_DRIVER, 13500,
-			 720, 720 + 14, 720 + 14 + 64, 858, 0,
-			 480, 480 + 7, 480 + 7 + 6, 525, 0,
+			 720, 720 + 16, 720 + 16 + 64, 858, 0,
+			 480, 480 + 6, 480 + 6 + 6, 525, 0,
 			 DRM_MODE_FLAG_INTERLACE)
 	},
 	{ /* Cropped and horizontally squashed to be TV-safe */
 		DRM_MODE("704x432i", DRM_MODE_TYPE_DRIVER, 15429,
-			 704, 704 + 72, 704 + 72 + 72, 980, 0,
-			 432, 432 + 31, 432 + 31 + 6, 525, 0,
+			 704, 704 + 76, 704 + 76 + 72, 980, 0,
+			 432, 432 + 30, 432 + 30 + 6, 525, 0,
 			 DRM_MODE_FLAG_INTERLACE)
 	},
 	{ /* Full size 625/50i with Rec.601 pixel rate */
 		DRM_MODE("720x576i", DRM_MODE_TYPE_DRIVER, 13500,
 			 720, 720 + 20, 720 + 20 + 64, 864, 0,
-			 576, 576 + 4, 576 + 4 + 6, 625, 0,
+			 576, 576 + 5, 576 + 5 + 5, 625, 0,
 			 DRM_MODE_FLAG_INTERLACE)
 	},
 	{ /* Cropped and squashed, for square(ish) pixels */
 		DRM_MODE("704x512i", DRM_MODE_TYPE_DRIVER, 15429,
 			 704, 704 + 80, 704 + 80 + 72, 987, 0,
-			 512, 512 + 36, 512 + 36 + 6, 625, 0,
+			 512, 512 + 37, 512 + 37 + 5, 625, 0,
 			 DRM_MODE_FLAG_INTERLACE)
 	}
 };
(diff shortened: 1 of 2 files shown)


njhollinghurst:
BTW allowing arbitrary pixel, frame or line rates would be more difficult! Many parameters depend on the frequency of the video clock (currently 108MHz), and the pixel rate must be a fraction of that. Since DRM can't specify timings for things like colour bursts, the driver would have to guess what was desired if the given timings didn't match any TV standard. So I don't propose to support arbitrary modes.

We might however choose to pay more attention to [hv]sync_start, to allow the active area to be shifted around the screen?

drivers/gpu/drm/rp1/rp1-vec/rp1_vec.c
```diff
@@ -298,27 +298,43 @@ static enum drm_mode_status rp1vec_mode_valid(struct drm_device *dev,
...
+	} else {
+		return MODE_BAD;
 	}
+
+	/* Check pixel rate (kHz) and horizontal size limit */
+	if (mode->clock == 13500 && mode->hdisplay <= 720)
+		return MODE_OK;
+	if (mode->clock >= 15428 && mode->clock <= 15429 &&
+	    mode->hdisplay <= 800)
+		return MODE_OK;
```
6by9, line 335: Minor nit that it would be nicer if `mode_valid` didn't switch from `if X return MODE_BAD` to `if X return MODE_OK` , and just had one `return MODE_OK` at the very end.

drivers/gpu/drm/rp1/rp1-vec/rp1_vec.c
```diff
@@ -298,27 +298,43 @@ static enum drm_mode_status rp1vec_mode_valid(struct drm_device *dev,
...
+		if (vdisplay_full > 576)
+			return MODE_BAD;
+	} else {
+		return MODE_BAD;
 	}
+
+	/* Check pixel rate (kHz) and horizontal size limit */
+	if (mode->clock == 13500 && mode->hdisplay <= 720)
+		return MODE_OK;
+	if (mode->clock >= 15428 && mode->clock <= 15429 &&
```
6by9, line 333: Very small range.
`if ((mode->clock == 15428 || mode->clock == 15429) ...`
?

6by9 (commented):
Looks reasonable, just a couple of minor nits.

drivers/gpu/drm/rp1/rp1-vec/rp1_vec.c
njhollinghurst: Our standard VEC clock rate is 108 MHz. Dividing by 8 gives the Rec.601 rate of 13.5MHz, but for emulators I thought it would be nice to offer a faster rate (so that 640 columns will comfortably fit on any CRT). 108/7 is a suitable rate but alas not an integer!

Since the back-end ignores most of the timings anyway, we could make the range broader. So somebody asking for 16MHz (the BBC Micro pixel rate) would get something that worked more or less as expected.

drivers/gpu/drm/rp1/rp1-vec/rp1_vec.c
njhollinghurst: Yes it bothered me as well but seemed nicer than the nested MODE_BADs in the vertical test. I'll see if I can tidy it.

njhollinghurst:
Hmm, now I wonder if I should also be implementing `mode_fixup()`?  I'm a little resistant as some of the definitive timings are not integers.

And I can't remember how it interacts with `mode_valid()`: does the latter mean that a mode is fixable, or that it requires no fixup?

6by9:
As just discussed, I suspect `mode_valid` is the wrong place to check this as it only filters the modes that are generally presented via an EDID, so not applicable here.

`atomic_check` probably of the connector is likely more appropriate. For vc4 HDMI it is done in the encoder `atomic_check`, but they should all be equal.

njhollinghurst:
`mode_valid()` does seem to be getting called usefully, perhaps due to some legacy/compatibility code.

Now I'm doing a bit of an overhaul that will also pay attention to `[hv]sync_start`, allowing the active region to be shifted around the screen. That should restore the old (pre-Pi5) overscan margin functionality, by other means.

drivers/gpu/drm/rp1/rp1-vec/rp1_vec.c
njhollinghurst: Now re-worked so the vertical test looks more like the horizontal one (both look for good cases).

njhollinghurst:
I think this is good now. I'm not sure we desperately need it... It does allow KMS/DRM applications to fine-tune the image size and position to control overscan, and relaxes some restrictions on mode; but there's still no way to set a custom mode at boot time for the console/FB.

While investigating I also got the impression that `tv_mode` (previously `tv_norm`) is not being passed correctly from the command line, but that should be a separate issue.

Renalds-B:
How to set overscan? Traditional margin_left/right/top/bottom in cmdline.txt doesn't change a thing.

njhollinghurst:
> How to set overscan? Traditional margin_left/right/top/bottom in cmdline.txt doesn't change a thing.

That's an outstanding issue. The current PR allows 'custom' DRM/KMS modes to be used, but it doesn't define any: it's necessary but maybe not sufficient. A future change is needed which may involve parsing command line options.

Renalds-B:
That means no way to impact overscan at the moment? Or is there a way for user to use custom KMS modes, that can help with this?

njhollinghurst:
> That means no way to impact overscan at the moment?

Try using one of these existing modes: 704x512i (for PAL) or 704x432i (for NTSC), or their progressive variants. They should be a fairly snug fit on CRT TVs though they may suffer from black borders on some LCDs.

> Or is there a way for user to use custom KMS modes, that can help with this?

Yes, right now you would have to create a custom mode (with the same pixel clock and totals as one of the existing modes, but different image size and porches). In X it can be done with `xrandr --newmode`. For Wayfire I think you have to edit a configuration file. Some future change may make this easier.

Renalds-B:
> > That means no way to impact overscan at the moment?
> 
> Try using one of these existing modes: 704x512i (for PAL) or 704x432i (for NTSC), or their progressive variants. They should be a fairly snug fit on CRT TVs though they may suffer from black borders on some LCDs.
> 
> > Or is there a way for user to use custom KMS modes, that can help with this?
> 
> Yes, right now you would have to create a custom mode (with the same pixel clock and totals as one of the existing modes, but different image size and porches). In X it can be done with `xrandr --newmode`. For Wayfire I think you have to edit a configuration file. Some future change may make this easier.

704x resolutions also suffer from black borders on my CRT.
And I have 480x theme applied in Retropie that doesn't quite look good with 407x resolution. 
On the other hand, 480x resolution is going outside of the CRT borders, and image is being cut off.

Although the most annoying thing is that interlaced resolutions also drop system frame rate to 30-25fps.

With older Pi models and old drivers this wasn't a problem (everything worked out of the box, really), but that's understandeable. It would be truly amazing, if Pi5 could support some sort of margin/overscan settings with standart resolutions like 407/480 and if interlaced video wouldn't impact the framerate of system.

Thank you for your hard work!

njhollinghurst:
Yes, we're still investigating ways to specify custom image sizes and offset, within the frameworks provided by Linux. (There's no convenient opportunity for extra scaling though: the size of the framebuffer must match the image sent to the TV.)

Due to limitations of the hardware (and maybe also of Linux DRM), we get one vertical blank interrupt / page-flip only for every frame (30fps), not for every field. I don't think that can be worked around.

PR #5846, opened 2024-01-11
drivers: w1: HACK do dummy reads while delaying

P33M:
This is the first, probably of many cases, where variable latency to RP1 breaks timing-sensitive GPIO bitbash.

See https://forums.raspberrypi.com/viewtopic.php?t=362339

Some other alternative approaches would be to
- Ask users to set "performance" in /sys/module/pcie_aspm/parameters/policy to force all links to L0 while bitbashing (also affects NVMe, if in use)
- Clobber ASPM on PCIE2 via overlay fragments, which means that specifying dtoverlay=w1-gpio in config.txt costs you ~400mW
- Add a pile of heuristic warts to the RP1 firmware to inhibit endpoint L1 entry if it detects RIO activity

So pick your poison...


JamesH65:
Just out of interest, why do you need to do the dummy reads during the delay? (Should that be commented in the code?)

pelwell:
To keep the link out of a powersave mode.

pelwell:
FYI, this gives me a reliable temperature from a DS18B20.

P33M:
Closing - rebasing offline doesn't seem to want to work, and I have a better version of the fix

PR #5847, opened 2024-01-12
Add support for 2712D0 variant

popcornmix:
Get ready for forthcoming 2712D0 variant of processor, which purges some unused peripherals and has some register layout tweaks.


drivers/pinctrl/bcm/pinctrl-bcm2712.c
```diff
@@ -60,22 +60,24 @@
...
 	}
 
-#define REG_BIT_INVALID 0xffff
+#define MUX_BIT_VALID	0x8000
+#define REG_BIT_INVALID	0xffff
 
 #define BIT_TO_REG(b) (((b) >> 5) << 2)
 #define BIT_TO_SHIFT(b) ((b) & 0x1f)
 
+#define MUX_BIT(mr,mb) (MUX_BIT_VALID + ((mr)*4)*8 + (mb)*4)
```
pelwell: Can you add the missing space that checkpatch is complaining about?

arch/arm/boot/dts/broadcom/bcm2712d0-rpi-5-b.dts
```diff
@@ -0,0 +1,109 @@
+// SPDX-License-Identifier: GPL-2.0
```
pelwell, line 1: This whole commit could be dropped now that we have the `bcm2712d0` overlay, unless we're catering for upstream preferences.

arch/arm/boot/dts/broadcom/bcm2712-rpi-5-b.dts
```diff
@@ -24,8 +24,6 @@
 #define spi6 _spi6
 #define uart0 _uart0
```
pelwell, line 25: This commit could go before the addition of the bcm2712d0 variant, to avoid modifying it.

arch/arm/boot/dts/broadcom/bcm2712d0-rpi-5-b.dts
```diff
@@ -86,6 +86,10 @@
 	reg = <0x7d510700 0x1c>;
 };
 
+&vc4 {
```
pelwell, line 89: Fold this change into the bcm2712d0-rpi-5-b.dts commit (putting this commit first), if we're keeping it.

arch/arm/boot/dts/broadcom/bcm2712d0-rpi-5-b.dts
popcornmix: Yeah, I was thinking it will be wanted in future for upstream.

popcornmix:
Updated to address comments.

pelwell (approved):
Nice - thanks.

6by9:
As it's only the HVS that differs with D0, the flag could be in vc4_hvs instead of vc4_dev. It would also then be the compatible for the HVS that gets updated instead of the overall driver.
I'm not worrying about it for now, but will probably do that when upstreaming.

PR #5848, opened 2024-01-13
overlays: Add overlay for the Solomon SSD1327 OLED

martinezjavier:
The ssd130x driver now supports for the Solomon SSD132x controller family. Add a Device Tree Overlay for the SSD1327 OLED panel.


pelwell:
Thanks!

PR #5859, opened 2024-01-17
imx290 HCG and analogue gain updates

6by9:


6by9:
Hit the wrong button and created the PR before I'd described it :-(

For testing on https://forums.raspberrypi.com/viewtopic.php?t=363696

geoffrey-vl (approved):

geoffrey-vl (approved):

geoffrey-vl:
Hi 6by9, I'm wondering if you could also add IMX462 to the driver:

`{
		.compatible = "sony,imx462lqr",
		.data = &imx290_models[IMX290_MODEL_IMX327LQR],
	},`

6by9:
IIRC then there's little point in doing so as it is identical to one of the others (I'd need to check which one) 

When there was the belief that imx462 could do 12bit 120fps readout then it would have been worth it, but that has been confirmed by Sony not to be the case and was an error on the flyer.

geoffrey-vl:
OK, well, I was suggesting it just to have the binding explicitly defined. Anyway, now that there is a small difference in max analog gain wouldn't we also need to modify the device three overlays for IMX462 so that it binds more correctly?

6by9:
IMX462 has a max gain of 29.4dB, same as IMX327.

Now 2d41947ec2c0140c65783982692c2e3d89853c47 did add differences in the register initialisation between 327 and 290, and I haven't checked through which option is correct for 462. If we had that, then it would be worth adding in 462 as well.
I'm afraid it's not a priority for at the moment though, and I can't share the required datasheets as they are under NDA.

geoffrey-vl:
Would it be OK enough to check if the IMX327 binding works for IMX462? Because that's something I could do.

6by9:
I know the 462 produces frames with the 290 or 327 configuration. It's very minor changes to some image quality that are different between the sensor models.

6by9:
I got distracted, so did the comparison of the registers. There's only one register different from imx290, however max analog gain is only 29.4dB vs 30.0dB.

PR updated to include adding a new compatible for imx462 (both mono and colour variants)

geoffrey-vl:
Looks to be working. Will you further upstream this one to mainline?

6by9:
Sent upstream - https://lore.kernel.org/linux-media/20241114-media-imx290-imx462-v1-0-c538a2e24786@raspberrypi.com/

I noted one minor change needed whilst reviewing it, so I'll update this again and then it should be good to merge.

naushir:
All seems reasonable to me.  I do wonder if this HCG enable should be automatic after a certain gain value maybe?  Are there any downsides of using it unconditionally at high gains?

naushir (commented):
LGTM

geoffrey-vl:
> All seems reasonable to me. I do wonder if this HCG enable should be automatic after a certain gain value maybe? Are there any downsides of using it unconditionally at high gains?

This was discussed in following topic: https://forums.raspberrypi.com/viewtopic.php?p=2182756&hilit=exposure#p2182756

6by9:
https://forums.raspberrypi.com/viewtopic.php?p=2186273#p2186273 confirmed that enabling HCG adds x5.8 of gain, so trying to enable it programmatically would cause instability in AE.

I'll hold fire on merging this just now, as there are a couple of comments from upstreaming the imx462 support patches.

6by9:
I updated this branch last week with the accepted patches from upstream, but forgot to comment on it.

I'm happy with this PR now.
@naushir another quick review please?

naushir:
LGTM

pelwell:
Would it not be cruel to merge before this PR's birthday?

6by9:
> Would it not be cruel to merge before this PR's birthday?

Do I have to buy the cake?!

PR #5863, opened 2024-01-18
ARM: pl011: Add rs485 to the RP1 support

pelwell:
pl011_axi_probe, added for RP1 support, lacks the rs485 additions that appeared during its development.

Suggested in https://forums.raspberrypi.com/viewtopic.php?p=2182033#p2182033.


pelwell:
I think this is low risk and correct by analogy with the original pl011_probe.

PR #5868, opened 2024-01-18
mmc: bcm2835-sdhost: use Host Software Queueing mechanism

P33M:
See commit 511ce378e16f ("mmc: Add MMC host software queue support")

Introduced in 5.8, this feature lets the block layer issue up to 2 pending requests to the MMC layer which in certain cases can improve throughput, but in the case of this driver can significantly reduce context switching when performing random reads.

On bcm2837 with a performant class A1 card, context switches under FIO random 4k reads go from ~8800 per second to ~5800, with a reduction in hardIRQs per second from ~5800 to ~4000. There is no appreciable difference in throughput.

For bcm2835, and for workloads other than random read, HSQ is a wash in terms of throughput and CPU load.

So, use it by default.

### DIFF
--- a/drivers/mmc/host/Kconfig
+++ b/drivers/mmc/host/Kconfig
@@ -37,6 +37,7 @@ config MMC_BCM2835_PIO_DMA_BARRIER
 config MMC_BCM2835_SDHOST
 	tristate "Support for the SDHost controller on BCM2708/9"
 	depends on ARCH_BCM2835
+	select MMC_HSQ
 	help
 	  This selects the SDHost controller on BCM2835/6.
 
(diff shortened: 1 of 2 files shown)


P33M:
I don't know why checkpatch claims the commit doesn't exist - github found it. Is CI doing a shallow clone?

nbuchwitz:
> I don't know why checkpatch claims the commit doesn't exist - github found it. Is CI doing a shallow clone?

It's configured via env var `PR_FETCH_DEPTH`  in this repo

https://github.com/raspberrypi/linux/tree/rpi/.github

 which seems to be set to 2:

https://github.com/raspberrypi/linux/actions/runs/7572151730/job/20621303831?pr=5868#step:4:4

PR #5872, opened 2024-01-19
Pisound updates for Pi 5

gtrainavicius:
* Don't export the button's GPIO via sysfs GPIO class from the kernel module.
* Use 100kHz SPI speed for Pi 5.

Btw, I'm unsure if these lines are needed in pisound-pi5-overlay.dts:

```
/ {
	compatible = "brcm,bcm2712";
};
```

Let me know if the lines should be removed. :)

### DIFF
--- a/arch/arm/boot/dts/overlays/Makefile
+++ b/arch/arm/boot/dts/overlays/Makefile
@@ -192,6 +192,7 @@ dtbo-$(CONFIG_ARCH_BCM2835) += \
 	piscreen.dtbo \
 	piscreen2r.dtbo \
 	pisound.dtbo \
+	pisound-pi5.dtbo \
 	pitft22.dtbo \
 	pitft28-capacitive.dtbo \
 	pitft28-resistive.dtbo \
--- a/arch/arm/boot/dts/overlays/README
+++ b/arch/arm/boot/dts/overlays/README
@@ -3486,8 +3486,14 @@ Params: speed                   Display SPI bus speed
 
 
 Name:   pisound
-Info:   Configures the Blokas Labs pisound card
+Info:   Configures the Blokas Labs Pisound card
 Load:   dtoverlay=pisound
+Params: <None>
+
+
+Name:   pisound-pi5
+Info:   Pi 5 specific overlay override for Blokas Labs Pisound card, see pisound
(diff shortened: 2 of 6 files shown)


pelwell:
That looks OK to me. The bcm2712 compatible string helps the checker tools know which devices it can be used on.

If you can just add a placeholder README entry I'll merge it - something in the style of:
```
Name:   disable-wifi-pi5
Info:   See disable-wifi
```

pelwell:
Sorry - one thing I should have said before is that we appreciate it if the DT patches are kept separate from the driver patches, because all the DT patches get squashed together when the tree is rebased.

pelwell:
"one think" - ha ha, something I obviously didn't do.

gtrainavicius:
Do you mean you'd like the pisound.c changes to be placed in a separate commit, or in another pull request entirely?

gtrainavicius:
I've split off the Device Tree changes into its own commit, and there's 2 commits for `pisound.c`

pelwell:
That's great.

gtrainavicius:
@pelwell, quick question - when is the next kernel release via `rpi-update` planned?

pelwell:
You just missed one yesterday. Ordinarily it would be a few days to a week until the next one, but this time it's more complicated because our master branch is about to switch over to rpi-6.6.y. Fortunately your changes have been ported over to 6.6 so they will be in the next release, one way or the other.

gtrainavicius:
We have some eager Pi 5 and Pisound users waiting for this fix to be released, do you have any estimate on when a new build will be out? Could a build on rpi-6.1y be triggered?

pelwell:
They can `sudo rpi-update rpi-6.1.y` to get the latest build at any time.

gtrainavicius:
This pull request didn't make it to [6.1.74](https://github.com/raspberrypi/firmware/commit/5f795d147fb21ecb0cf2e6d42929130b6b049de0), or did it?

pelwell:
Passing kernel branch names to rpi-update gets you the latest auto-build of the top-of-tree for that branch.

Try it.

gtrainavicius:
Ah ok, I didn't know this, thank you!

PR #5885, opened 2024-01-23
RP1 VEC: Use tv_mode from command line, align enum with drm_connector; tidy

### LINKED ISSUE #5864: RP1 VEC: "tv_mode" property is not correctly inherited from command line "video"
### Describe the bug

RP1 VEC has a `tv_norm` string parameter which affects
- The set of named DRM modes advertised
- The colour system and pedestal used in some of those modes

(by default it supports all TV modes, prefers NTSC, and guesses NTSC or PAL depending on the active mode).

It can successfully be set as a module parameter from the command line (so there is no urgent problem). However, the more standard `video/tv_mode` command line setting appears to be ignored.

Furthermore, the private enum of tv_norms used by RP1 VEC driver does not quite match DRM's `enum drm_connector_tv_mode`.  It would cause confusion if an enumerated value were to get passed from the connector to our driver (which we do at least _attempt_ to do, via `drm_object_attach_property()`). The same thing is true for VC4 VEC.

When using RP1 VEC, compare different behaviour between
`video=Composite-1,tv_mode=NTSC-J`
and
`video=Composite-1 drm-rp1-vec.tv_norm=NTSC-J`

### Device (s)

Other

### System

Raspberry Pi 5
rpi-6.1.y as at Thu 18 Jan 11:32:27 GMT 2024

### Logs

_No response_

### Additional context

May also be relevant to VC4.

njhollinghurst:
This attempts to fix #5864, targeting Linux 6.6

I narrowly kept the "tv_norm" module parameter  in this draft; but it's somewhat redundant now that we can use video/tv_mode.

### DIFF
--- a/drivers/gpu/drm/rp1/rp1-vec/rp1_vec.c
+++ b/drivers/gpu/drm/rp1/rp1-vec/rp1_vec.c
@@ -48,52 +48,6 @@
 
 #include "rp1_vec.h"
 
-/*
- * Default TV standard parameter; it may be overridden by the OF
- * property "tv_norm" (which should be one of the strings below).
- *
- * The default (empty string) supports various 60Hz and 50Hz modes,
- * and will automatically select NTSC[-M] or PAL[-BDGHIKL]; the two
- * "fake" 60Hz standards NTSC-443 and PAL60 also support 50Hz PAL.
- * Other values will restrict the set of video modes offered.
- *
- * Finally, the DRM connector property "mode" (which is an integer)
- * can be used to override this value, but it does not prevent the
- * selection of an inapplicable video mode.
- */
-
-static char *rp1vec_tv_norm_str;
-module_param_named(tv_norm, rp1vec_tv_norm_str, charp, 0600);
-MODULE_PARM_DESC(tv_norm, "Default TV norm.\n"
-		 "\t\tSupported: NTSC, NTSC-J, NTSC-443, PAL, PAL-M, PAL-N,\n"
-		 "\t\t\tPAL60.\n"
-		 "\t\tDefault: empty string: infer PAL for a 50 Hz mode,\n"
-		 "\t\t\tNTSC otherwise");
-
-const char * const rp1vec_tvstd_names[] = {
-	[RP1VEC_TVSTD_NTSC]     = "NTSC",
-	[RP1VEC_TVSTD_NTSC_J]   = "NTSC-J",
-	[RP1VEC_TVSTD_NTSC_443] = "NTSC-443",
-	[RP1VEC_TVSTD_PAL]      = "PAL",
-	[RP1VEC_TVSTD_PAL_M]    = "PAL-M",
-	[RP1VEC_TVSTD_PAL_N]    = "PAL-N",
-	[RP1VEC_TVSTD_PAL60]    = "PAL60",
-	[RP1VEC_TVSTD_DEFAULT]  = "",
-};
-
-static int rp1vec_parse_tv_norm(const char *str)
-{
-	int i;
-
-	if (str && *str) {
-		for (i = 0; i < ARRAY_SIZE(rp1vec_tvstd_names); ++i) {
-			if (strcasecmp(str, rp1vec_tvstd_names[i]) == 0)
-				return i;
-		}
-	}
-	return RP1VEC_TVSTD_DEFAULT;
-}
-
 static void rp1vec_pipe_update(struct drm_simple_display_pipe *pipe,
 			       struct drm_plane_state *old_state)
 {
@@ -143,7 +97,7 @@ static void rp1vec_pipe_update(struct drm_simple_display_pipe *pipe,
 
 static void rp1vec_pipe_enable(struct drm_simple_display_pipe *pipe,
 			       struct drm_crtc_state *crtc_state,
-			      struct drm_plane_state *plane_state)
+			       struct drm_plane_state *plane_state)
 {
 	struct rp1_vec *vec = pipe->crtc.dev->dev_private;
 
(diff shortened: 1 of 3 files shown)


njhollinghurst:
@6by9  Shall I go ahead and remove the "tv_norm" parameter altogether? I am in two minds about it. 

(I think the VEC driver will need at least one parameter, either for enabling 'vintage' modes and/or defining a custom mode, prior to evaluating the "video=" mode; but it need not be the existing "tv_norm" parameter.)

drivers/gpu/drm/rp1/rp1-vec/rp1_vec.c
```diff
@@ -231,36 +198,47 @@ static const struct drm_display_mode rp1vec_modes[4] = {
 	}
 };
 
+/*
+ * Advertise standard and preferred video modes.
+ *
+ * From each interlaced mode in the table above, derive a progressive one.
+ *
+ * This driver always support all 50Hz and 60Hz video modes, regardless
```
6by9: nit: s/support/supports

drivers/gpu/drm/rp1/rp1-vec/rp1_vec.c
```diff
@@ -231,36 +198,47 @@ static const struct drm_display_mode rp1vec_modes[4] = {
...
+ * (except that "PAL" with 525/60 will be implemented as "PAL60").
+ */
+
 static int rp1vec_connector_get_modes(struct drm_connector *connector)
 {
-	struct rp1_vec *vec = container_of(connector, struct rp1_vec, connector);
-	bool ok525 = RP1VEC_TVSTD_SUPPORT_525(vec->tv_norm);
-	bool ok625 = RP1VEC_TVSTD_SUPPORT_625(vec->tv_norm);
 	int i, prog, n = 0;
+	int tvm = rp1vec_initial_tv_mode(connector);
```
6by9: Do we really want the command line tv_norm here, or use the currently selected standard?

drivers/gpu/drm/rp1/rp1-vec/rp1_vec.c
```diff
@@ -269,11 +247,9 @@ static int rp1vec_connector_get_modes(struct drm_connector *connector)
 
 static void rp1vec_connector_reset(struct drm_connector *connector)
 {
-	struct rp1_vec *vec = container_of(connector, struct rp1_vec, connector);
-
 	drm_atomic_helper_connector_reset(connector);
 	if (connector->state)
-		connector->state->tv.mode = vec->tv_norm;
+		connector->state->tv.mode = rp1vec_initial_tv_mode(connector);
```
6by9: I'd need to check when the connector state reset is used. It feels like we really want to take the value from the DRM property, not the original default.

If it comes from the DRM property, then it feels better to do it in the framework reset, not the driver specific one. I vaguely recall having to fix up the planes reset for colourspace or rotation previously for the same reasons.

Actually `drm_atomic_helper_connector_tv_reset` uses exactly that, although a downstream patch then messes that up.

drivers/gpu/drm/rp1/rp1-vec/rp1_vec.c
```diff
@@ -231,36 +198,47 @@ static const struct drm_display_mode rp1vec_modes[4] = {
...
-				}
-
-				if (mode->hdisplay == 704 &&
-				    mode->vtotal == ((ok525) ? 525 : 625))
-					mode->type |= DRM_MODE_TYPE_PREFERRED;
-
-				drm_mode_set_name(mode);
-				drm_mode_probed_add(connector, mode);
-				n++;
+		for (prog = 0; prog < 2; prog++) {
```
6by9, line 214: The default timings for the defined TV standards can added with
```
	int count = drm_connector_helper_tv_get_modes(connector);
```
vc4 does this in the 6.6 tree.

Adding the cropped modes is a little trickier and may still require modes to be stored, but it's nicer to remove as much duplication as possible.

drivers/gpu/drm/rp1/rp1-vec/rp1_vec.c
njhollinghurst: I wasn't sure... the latter is in "state" which I thought might have locking issues?
The only thing it affects is which mode is marked as "preferred".

6by9: Use `drm_connector_helper_tv_get_modes` and let it worry about which one gets the preferred flag set.

drivers/gpu/drm/rp1/rp1-vec/rp1_vec.c
njhollinghurst: Gosh there's a lot of code under `drm_connector_helper_tv_modes()`. Hope it's correct.
Will it produce both 525/60 and 625/50 modes at the same time?

drivers/gpu/drm/rp1/rp1-vec/rp1_vec.c
njhollinghurst: OK. I'm a bit vague as to the difference between the "default value", the "property value" and the "state", and what is the intention of the reset, but it looks like we should follow (or maybe just call) `drm_atomic_helper_connector_tv_reset()`

drivers/gpu/drm/rp1/rp1-vec/rp1_vec.c
njhollinghurst: ...or should it do as discussed below and use `drm_object_property_get_default_value()`?

drivers/gpu/drm/rp1/rp1-vec/rp1_vec.c
6by9: Maxime wrote it

It'll add 480i and/or 576i based on the standards listed `drm_mode_create_tv_properties` was told were supported, and in the order (+ preferred flag) as based on the default value registered when you `drm_object_attach_property` the tv_mode property.

drivers/gpu/drm/rp1/rp1-vec/rp1_vec.c
6by9: Discussed below. Resolving.

drivers/gpu/drm/rp1/rp1-vec/rp1_vec.c
njhollinghurst: Ah, so that's what "default property value" means.

njhollinghurst: Currently marking one of the cropped modes as "preferred": this is to stop the FB console spilling off the screen (in the absence of any better way of handling overscan).

As I've just checked, `drm_connector_helper_tv_modes()` does return both Rec.601 modes, though not the progressive ones (maybe that needs an extra command line option). Is there a way to stop it marking one of those as preferred?

6by9: The cropped modes being preferred probably does mean you need to roll your own.

I seem to recall there is a function to nuke the preferred flag from all modes so you can add it as you see fit, but it may or may not help.

njhollinghurst: I just tried an experiment to see if that helper function would pay attention to "margin_left" etc. Unsurprisingly it seems not to. I think I may need to roll my own. I'll ensure the standard modes match its output (one of my porches differs by all of 4 pixels right now).

I'll tidy up the preferred flag, to use the property's own default rather than go back to the command line options.

njhollinghurst: I think it's as good as it's going to get, for now.

Will look at it again when we add the "vintage" TV modes (for which a "progressive" variant is probably not useful).

njhollinghurst:
Cherry-picked this back onto the rebased rpi-6.6.y. Will do a quick re-test and then merge.

PR #5889, opened 2024-01-24
PCI: brcmstb: Enable CRS software visibility after linkup

P33M:
This fell out of the discussions around "why does the bootloader nvme probe behave differently to Linux" - as part of aligning the behaviours of both, we should make maximal use of the features offered by the RC.

The RC supports neither Device Readiness nor Function Readiness notifications, therefore we should use a combination of the 100ms grace period after link-up with CRS.

EPs are permitted to issue Retry statuses for up to 1 second, which if CRSSVE=0 will
a) completely stall the CPU until the Ubus/completion timeout is reached (<1s, but some tens of milliseconds)
b) cause the RC to return 0xffffffff for reads of the vendor ID register instead of 0xffff0001 which will cause probe failure instead of retrying up to the 1 second limit.

For some reason the associated root complex register bit is reset by perst_n, so enable it after link-up.

### DIFF
--- a/drivers/pci/controller/pcie-brcmstb.c
+++ b/drivers/pci/controller/pcie-brcmstb.c
@@ -1385,7 +1385,7 @@ static int brcm_pcie_start_link(struct brcm_pcie *pcie)
 {
 	struct device *dev = pcie->dev;
 	void __iomem *base = pcie->base;
-	u16 nlw, cls, lnksta;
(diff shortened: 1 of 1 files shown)


timg236:
Is this something that we'd backport to 6.1?

P33M:
As we're about to do one final 6.1 release, no - this needs more soak testing with various PCIe not-HAT setups.

PR #5891, opened 2024-01-24
HACK: drm/vc4: Disable overrun interrupts

6by9:
For testing hypothesis that the lockup of not clearing stale dlist allocs is down to a read-modify-write race between DSPEISLUR and DSPEIEOF.


6by9:
Close as superceded by #5935 on 6.6.

Issue #5814, opened 2023-12-30
[6.6] regression: no composite video output on RPi4


HiassofT:
### Describe the bug

With kernel 6.6 (currently 6.6.8 from rpi-update next) composite output doesn't work on RPi4. kmsprint shows the connector and encoder, but no crtc or planes and the video output stays black.
```
root@raspberrypi:~# kmsprint 
Connector 0 (45) Composite-1 (connected)
  Encoder 0 (44) TVDAC
```
On kernel 6.1 from RPiOS apt-update composite output is fine.

### Steps to reproduce the behaviour

Enable composite via config.txt and cmdline.txt.

config.txt add the composite parameter to the vc4kms-v3d overlay
```
dtoverlay=vc4-kms-v3d,composite=1
```
cmdline.txt add `video=Composite-1:720x576@50ie` to the end
```
console=serial0,115200 console=tty1 root=PARTUUID=5372fcc0-02 rootfstype=ext4 fsck.repair=yes rootwait cfg80211.ieee80211_regdom=AT video=Composite-1:720x576@50ie
```

### Device (s)

Raspberry Pi 4 Mod. B

### System

```
System Information
------------------

Raspberry Pi 4 Model B Rev 1.1
PRETTY_NAME="Debian GNU/Linux 12 (bookworm)"
NAME="Debian GNU/Linux"
VERSION_ID="12"
VERSION="12 (bookworm)"

Raspberry Pi reference 2023-10-10
Generated using pi-gen, https://github.com/RPi-Distro/pi-gen, 962bf483c8f326405794827cce8c0313fd5880a8, stage4

Linux raspberrypi 6.6.8-v8+ #1709 SMP PREEMPT Thu Dec 21 12:54:25 GMT 2023 aarch64 GNU/Linux
Revision	: c03111
Serial		: 100000008eb25f41
Model		: Raspberry Pi 4 Model B Rev 1.1
Throttled flag  : throttled=0x0
Camera          : supported=0 detected=0, libcamera interfaces=0

Videocore information
---------------------

Dec  8 2023 12:36:15 
Copyright (c) 2012 Broadcom
version e02d33b3122450accf9dea471a177d3b5623f5ad (clean) (release) (start)

alloc failures:     0
compactions:        0
legacy block fails: 0


Filesystem information
----------------------
Filesystem     1K-blocks    Used Available Use% Mounted on
udev             1668532       0   1668532   0% /dev
tmpfs             388300    1096    387204   1% /run
[...]

Issue #5894, opened 2024-01-25
vc4_vec broken on 6.7


6by9:
### Describe the bug

Observing that the VEC fails to enable modes on 6.7 (probably 6.6 as well).
vc4_vec_encoder_atomic_check is failing at https://github.com/raspberrypi/linux/blob/rpi-6.6.y/drivers/gpu/drm/vc4/vc4_vec.c#L760
```
[    6.654936] vc4_vec_encoder_atomic_check: mode "720x480i": 60 13500 720 736 800 858 480 486 492 525 0x48 0x1a
[    6.655043] vc4_vec_encoder_atomic_check: mode->crtc_(timings) 720 736 800 858 240 243 246 262
```

This check was added via #4406 and is only a downstream patch. 

The drm_mode for the analog interlaced modes are now coming from the core framework (`fill_analog_mode` via `drm_analog_tv_mode`) vs being baked into vc4, and the timing has changed slightly
Was `13.500 720/14/64/60/-     480/7/6/32/-`
Now `13.500 720/16/64/58/-     480/6/6/33/-`
Horizontal timing change to the porches will be a slight horizontal shift.
Vertical timing changes may imply the PV is wanting a slight munge so that it give the VEC the right signals at the desired time.

### Steps to reproduce the behaviour

Enable `dtoverlay=vc4-kms-v3d,composite`, reboot, and observe the framebuffer not come up, and `kmstest` can't enable the default modes (encoder atomic_check fails).

### Device (s)

Raspberry Pi CM4

### System

6.7.1

6by9:
Simplest solution is to revert https://github.com/raspberrypi/linux/commit/2ef74e9687c92e26d4b047c2499c96b215ec7f00 (on 6.6).

6by9:
Fixed via #5898

PR #5898, opened 2024-01-25
6.6 VEC fixups

6by9:
Fixes up the downstream VEC patches so that the VEC works on 6.6. Also needed on 6.7.
See #5894

### DIFF
--- a/drivers/gpu/drm/vc4/vc4_vec.c
+++ b/drivers/gpu/drm/vc4/vc4_vec.c
@@ -404,10 +404,6 @@ static void vc4_vec_connector_reset(struct drm_connector *connector)
 {
 	drm_atomic_helper_connector_reset(connector);
 	drm_atomic_helper_connector_tv_reset(connector);
-
-	/* preserve TV standard */
-	if (connector->state)
-		connector->state->tv.mode = vc4_vec_get_default_mode(connector);
 }
 
 static int
--- a/drivers/gpu/drm/vc4/vc4_vec.c
+++ b/drivers/gpu/drm/vc4/vc4_vec.c
@@ -757,18 +757,6 @@ static int vc4_vec_encoder_atomic_check(struct drm_encoder *encoder,
 		if ((mode->crtc_vtotal - mode->crtc_vsync_end) < 4)
 			return -EINVAL;
 
-		if ((mode->flags & DRM_MODE_FLAG_INTERLACE) &&
-		    (mode->vdisplay % 2 != 0 ||
-		     mode->vsync_start % 2 != 1 ||
-		     mode->vsync_end % 2 != 1 ||
-		     mode->vtotal % 2 != 1))
-			return -EINVAL;
-
-		/* progressive mode is hard-wired to 262 total lines */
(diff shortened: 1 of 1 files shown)


6by9:
As I believe we're still rebasing 6.6, I'm assuming the revert will become a dropped patch, hence not needing a SoB.
The fixes tag will also become erroneous as the git hash changes, but that's life. I'll be picking this up when upstreaming.

HiassofT:
Thanks, this fixes #5814 

I did a quick test on LibreELEC on RPi4 with `video=Composite-1:720x576@50ie` and `video=Composite-1:720x480@60ie` and both resulted in proper PAL/NTSC output

6by9:
> Thanks, this fixes #5814
> 
> I did a quick test on LibreELEC on RPi4 with `video=Composite-1:720x576@50ie` and `video=Composite-1:720x480@60ie` and both resulted in proper PAL/NTSC output

Thanks for testing.
I'd forgotten you'd reported the issue already (sorry), and was looking to make vc4 and rp1_vec align in behaviour when I found vc4 was broken.

Ideally you want to use `video=Composite-1:720x576@50ie,tv_mode=PAL` and `video=Composite-1:720x480@60ie,tv_mode=NTSC` to be explicit over which standard you want. vc4 previously had code that looks at the resolution and makes a guess, but we haven't added that to RP1 at present (see #5885).
Setting the tv_mode is the "offficial" way of doing it, so is the one that should be recommended.

drivers/gpu/drm/vc4/vc4_vec.c
```diff
@@ -404,10 +404,6 @@ static void vc4_vec_connector_reset(struct drm_connector *connector)
 {
 	drm_atomic_helper_connector_reset(connector);
 	drm_atomic_helper_connector_tv_reset(connector);
```
njhollinghurst, line 408: (I'll copy this to the parallel RP1 change)

drivers/gpu/drm/vc4/vc4_vec.c
```diff
@@ -370,7 +370,9 @@ static const struct drm_prop_enum_list legacy_tv_mode_names[] = {
 enum drm_connector_tv_mode
 vc4_vec_get_default_mode(struct drm_connector *connector)
 {
-	if (vc4_vec_tv_norm) {
+	if (connector->cmdline_mode.tv_mode_specified) {
+		return connector->cmdline_mode.tv_mode;
+	} else if (vc4_vec_tv_norm) {
```
njhollinghurst, line 375: Are the enums still mismatched (e.g. NTSC-443, NTSC-J order)?

6by9: The vc4 driver internally only uses the `DRM_MODE_TV_MODE_` enums.

The legacy `VC4_VEC_TV_MODE_` enums are only used if userspace uses the legacy (and non-standardised) "mode" property, and it converts to/from `DRM_MODE_TV_MODE_` on the fly.

The tv_norm module parameter is matching on strings, and is now using `drm_get_tv_mode_from_name` to match against the standardised enum strings and provide a `DRM_MODE_TV_MODE_`

Issue #31, opened 2012-06-02
kernel oops. 


rewolff:
Hi, 

While my 'pi was doing real work for me it crashed 15 minutes after I had left. :-( 

It produced the attached "ooops" (provided I can find how to attach files) . 

It's reading from an SD card (microSD to be exact), and writing the data over a TPC connection to a server. This is progressing at 300k/sec. (with 32G total, that is expected tot take 29 hours). 

It has now been running since this morning: about 14 hours.

rewolff:
I can't find how to attach files here, so I'll have to host the file myself: http://prive.bitwizard.nl/DSC04643.JPG

rewolff:
I checked my current copy of system.map. I did recompile the kernel adding a few modules, so I don't think the symbols moved, but I can't be sure. 

Crash occurred in memcpy (sounds reasonable). 
linkregister contains DWC_TASK_SCHEDULE (0xc before the end). (which as far as I can tell doesn't call memcpy).

popcornmix:
Can you reproduce this on latest 3.2.27 kernel?

ghollingworth:
Closing since there's been no further investigation

rewolff:
Sounds like a plan. It was hard for me to reproduce. I was hoping the report might fit into a pattern or something like that.

PR #5899, opened 2024-01-26
Pispbe/rpi 6.7.y/be unsquashed

jmondi:
NOT FOR MERGE only for review.

Un-squashed list of changes to the BE driver to prepare for upstreaming for earlier review/comments.
Mostly cosmetics, apart from a few commits that rework slightly the job scheduling routines.

Tested with this branch based on Dave's mainline rpi1 support
https://github.com/jmondi/raspberrypi-linux/tree/pispbe/6by9_mainline_2712_rp1/libcamera-testing

with single and dual camera capture.

The actual commits for the mainline patches, based on media-staging are at
https://github.com/jmondi/raspberrypi-linux/tree/pispbe/media-staging/be-upstream


jmondi:
This last push adds documentation for the image formats and the isp and breaks the uapi header out from the driver directory.

The only missing part is the documentation of the opaque format as I've found none in the PiSP specs.

All of these commits are now collected in:
https://github.com/jmondi/raspberrypi-linux/commits/pispbe/media-staging/be-upstream/

which represent what I intend to send upstream once this PR has been approved

drivers/media/platform/raspberrypi/pisp_be/pisp_be.c
```diff
@@ -47,9 +47,9 @@ MODULE_LICENSE("GPL v2");
 #define PISP_BE_INTERRUPT_STATUS_OFFSET		0x1c
 #define PISP_BE_AXI_OFFSET			0x20
 #define PISP_BE_CONFIG_BASE_OFFSET		0x40
-#define PISP_BE_IO_INPUT_ADDR0_LO_OFFSET 	PISP_BE_CONFIG_BASE_OFFSET
-#define PISP_BE_GLOBAL_BAYER_ENABLE_OFFSET (PISP_BE_CONFIG_BASE_OFFSET + 0x70)
-#define PISP_BE_GLOBAL_RGB_ENABLE_OFFSET   (PISP_BE_CONFIG_BASE_OFFSET + 0x74)
+#define PISP_BE_IO_INPUT_ADDR0(n)		(0x40 + 8 * (n))
+#define PISP_BE_GLOBAL_BAYER_ENABLE		0xb0
+#define PISP_BE_GLOBAL_RGB_ENABLE		0xb4
```
naushir: Should we not be using ``PISP_BE_CONFIG_BASE_OFFSET`` in these macros, e.g.:
```
#define PISP_BE_IO_INPUT_ADDR0(n) (PISP_BE_CONFIG_BASE_OFFSET+ 8 * (n))
```

include/uapi/linux/pisp_be_config.h
```diff
@@ -5,12 +5,12 @@
  * Copyright (C) 2021 - Raspberry Pi Ltd
  *
  */
-#ifndef _PISP_BE_CONFIG_H_
-#define _PISP_BE_CONFIG_H_
+#ifndef _UAPI_PISP_BE_CONFIG_H_
```
naushir: Just to note that we should copy this modified file to libpisp.

drivers/media/platform/raspberrypi/pisp_be/pisp_be.c
njhollinghurst: It's a style question -- the original was derived from our hardware specification and test code, which uses offsets wrt the block base address, but wasn't particularly Linux-style.

Since we ignore the user-specified configuration here and write directly to the registers, either style of basing seems acceptable to me.

I agree that a parameterized macro would be nice for HW addresses, though FWIW I might remove the `INPUT` (will we use the macro for things that are _not_ inputs?) and the `0` (the parameter _n_ replaces the 0) and add a `LO`/`HI` distinction for the two 32-bit halves.

include/uapi/linux/pisp_common.h
```diff
@@ -1,12 +1,18 @@
...
 /*
- * Raspberry Pi PiSP common types.
+ * Raspberry Pi PiSP common configuration definitions.
  *
  * Copyright (C) 2021 - Raspberry Pi (Trading) Ltd.
  *
  */
-#ifndef _PISP_TYPES_H_
-#define _PISP_TYPES_H_
+#ifndef _UAPI_PISP_COMMON_H_
```
naushir: Just to note that we should copy this modified file to libpisp.

drivers/media/platform/raspberrypi/pisp_be/pisp_be.c
```diff
@@ -303,51 +270,50 @@ static void hw_queue_job(struct pispbe_dev *pispbe,
...
 	 */
-	for (u = 0; u < N_HW_ADDRESSES; ++u) {
-		write_reg(pispbe, PISP_BE_IO_INPUT_ADDR0_LO_OFFSET + 8 * u,
-			  (u32)(hw_dma_addrs[u]));
-		write_reg(pispbe, PISP_BE_IO_INPUT_ADDR0_LO_OFFSET + 8 * u + 4,
-			  (u32)(hw_dma_addrs[u] >> 32));
+	for (unsigned int u = 0; u < N_HW_ADDRESSES; ++u) {
+		pispbe_wr(pispbe, PISP_BE_IO_INPUT_ADDR0(u),
+			  lower_32_bits(job->hw_dma_addrs[u]));
+		pispbe_wr(pispbe, PISP_BE_IO_INPUT_ADDR0(u) + 4,
```
njhollinghurst: See earlier comment at L63 - the macro shouldn't be called `INPUT`...`0` if it doesn't refer to input component 0.
And the + 4 isn't nice.

drivers/media/platform/raspberrypi/pisp_be/pisp_be.c
```diff
@@ -303,51 +270,50 @@ static void hw_queue_job(struct pispbe_dev *pispbe,
...
 			return;
 		}
 	}
 
 	/*
-	 * Write tile pointer to hardware. XXX Tile offsets and sizes not
-	 * checked (and even if checked, the user could subsequently modify
-	 * them)!
+	 * Write tile pointer to hardware. Tile offsets and sizes not checked
+	 * (and even if checked, the user could subsequently modify them)!
```
njhollinghurst: Maybe we can excuse ourselves for not checking, by adding something like this?
```
	 * The IOMMU should prevent out-of-bounds offsets reaching non-ISP buffers.
```
though this implicitly admits that one ISP-BE user could mess with buffers (including config buffers) belonging to a _different_ ISP-BE user.

drivers/media/platform/raspberrypi/pisp_be/pisp_be.c
```diff
@@ -578,119 +538,103 @@ static int pispbe_schedule_internal(struct pispbe_node_group *node_group,
...
+		if (buf[i])
+			list_del(&buf[i]->ready_list);
+		spin_unlock_irqrestore(&node->ready_lock, flags);
 
-	/* Pull a buffer from each V4L2 queue to form the queued job */
-	for (i = 0; i < PISPBE_NUM_NODES; i++) {
-		if (buf[i]) {
-			node = &node_group->node[i];
+		if (!buf[i] && !ignore_buffers)
+			return -ENODEV;
```
njhollinghurst: Have we changed the logic here, by amalgamating these two loops, and is this safe to do?
Previously in the "nothing to do" case, no state would be changed. It is also not an error but a common case.

drivers/media/platform/raspberrypi/pisp_be/pisp_be.c
```diff
@@ -40,32 +34,36 @@ MODULE_LICENSE("GPL v2");
...
+#define PISP_BE_INTERRUPT_STATUS_REG		0x1c
+#define PISP_BE_AXI_REG				0x20
+#define PISP_BE_CONFIG_BASE_REG			0x40
+#define PISP_BE_IO_INPUT_ADDR0(n)		(0x40 + 8 * (n))
+#define PISP_BE_GLOBAL_BAYER_ENABLE		0xb0
+#define PISP_BE_GLOBAL_RGB_ENABLE		0xb4
+#define N_HW_ADDRESSES				14
+#define N_HW_ENABLES				2
+
+#define PISP_BE_VERSION_2712C1			0x02252700
```
njhollinghurst: The reference to C1 is strictly correct but confusing -- D0 differs only in the Minor bits, which we don't check.
I wonder if we should omit the `C1` suffix?

naushir: Sounds reasonable.

drivers/media/platform/raspberrypi/pisp_be/pisp_be.c
jmondi: I kept `INPUT` as it's actually the register name used in the specs, I could drop it but matching specs seems a good idea. I agree we can create a `LO/HI` version though.

To be honest I would remove `PISP_BE_CONFIG_BASE_OFFSET` completely but replacing 

```
	begin =	offsetof(struct pisp_be_config, global.bayer_order)
	      / sizeof(u32);
	end = offsetof(struct pisp_be_config, axi) / sizeof(u32);
	for (unsigned int u = begin; u < end; u++)
		pispbe_wr(pispbe, PISP_BE_CONFIG_BASE_REG + 4 * u,
			  ((u32 *)job->config)[u]);
```

seems not a good idea, but I would only use `PISP_BE_CONFIG_BASE_OFFSET` here and not for defining other registers where we write directly to.

drivers/media/platform/raspberrypi/pisp_be/pisp_be.c
jmondi: I agree with all but `INPUT` as that's the name used in specs

drivers/media/platform/raspberrypi/pisp_be/pisp_be.c
jmondi: ok, do you mean

```
		if (!buf[i] && !ignore_buffers)
			return -ENODEV;

		pispbe->queued_job.buf[i] = buf[i];
```

we shouldn't `pispbe->queued_job.buf[i] = buf[i];` if `!buf[i]` ?

I thought setting it to (a possible) `NULL` was safe as that was its default state anyway. I can protect this with

```
 if (buf[i])
		pispbe->queued_job.buf[i] = buf[i];
```

or drop this patch completely, as even if there's a tiny window between the two loops, it should be safe ?

drivers/media/platform/raspberrypi/pisp_be/pisp_be.c
jmondi: oh, now I see, it's not only used for `INPUT` but for all addresses. I'll remove it!

drivers/media/platform/raspberrypi/pisp_be/pisp_be.c
njhollinghurst: I was being a bit pedantic: `INPUT0_LO ` is the name of the _first_ of a series of address registers, it's not the name of _all_ of the address registers. Maybe suitable macro names might be something like `PISP_BE_DMA_ADDRESS_LO(n)` and `PISP_BE_DMA_ADDRESS_HI(n)`?

BTW I think we made a mistake when we included DMA addresses in `struct pisp_be_config` -- they are not for the user to specify -- but now that it's in UAPI, we can't change the structure! We just have to skip over them. I'm not sure how to make it cleaner.

jmondi: > I was being a bit pedantic: `INPUT0_LO ` is the name of the _first_ of a series of address registers, it's not the name of _all_ of the address registers. Maybe suitable macro names might be something like `PISP_BE_DMA_ADDRESS_LO(n)` and `PISP_BE_DMA_ADDRESS_HI(n)`?

Yeah, I realized it later, so I dropped `INPUT` and `0` from name. I can rename it it `PISP_BE_DMA_ADDRESS...()` if preferred

> 
> BTW I think we made a mistake when we included DMA addresses in `struct pisp_be_config` -- they are not for the user to specify -- but now that it's in UAPI, we can't change the structure! We just have to skip over them. I'm not sure how to make it cleaner.

Well, the uAPI is not upstream yet, so we're still in time to change it :)

drivers/media/platform/raspberrypi/pisp_be/pisp_be_formats.h
```diff
@@ -504,11 +504,6 @@ static const struct pisp_be_format supported_formats[] = {
 		.colorspace_mask    = V4L2_COLORSPACE_MASK_RAW,
 		.colorspace_default = V4L2_COLORSPACE_RAW,
 	},
-	/* Opaque BE format for HW verification. */
-	{
-		.fourcc		    = V4L2_PIX_FMT_RPI_BE,
-		.align		    = 32,
-	},
```
naushir: This format is also used in the non-HoG cases for our pisp HW verification tests.  @njhollinghurst perhaps we can reintroduce this patch in our downstream tree to keep the tests working?

naushir:
Just trying these changes now on top of a rpi-6.6.y tree and unfortunately when I run rpicam-apps or cam, I get the following dmesg output:
```
[   34.710292] pispbe 1000880000.pisp_be: ISP BE config error: check if ISP RAMs enabled?
```
and then the system locks up.  I'm sure I've missed something obvious, @jmondi any thoughts?

naushir:
Ah, ~~perhaps~~ this is the fix:
```c
diff --git a/drivers/media/platform/raspberrypi/pisp_be/pisp_be.c b/drivers/media/platform/raspberrypi/pisp_be/pisp_be.c
index 07f0fb8aa85c..6e88491b20ba 100644
--- a/drivers/media/platform/raspberrypi/pisp_be/pisp_be.c
+++ b/drivers/media/platform/raspberrypi/pisp_be/pisp_be.c
@@ -265,7 +265,7 @@ static void pispbe_queue_job(struct pispbe_dev *pispbe,
        for (unsigned int u = 0; u < N_HW_ADDRESSES; ++u) {
                pispbe_wr(pispbe, PISP_BE_IO_ADDR_LOW(u),
                          lower_32_bits(job->hw_dma_addrs[u]));
-               pispbe_wr(pispbe, PISP_BE_IO_ADDR_HIGH(u) + 4,
+               pispbe_wr(pispbe, PISP_BE_IO_ADDR_HIGH(u),
                          upper_32_bits(job->hw_dma_addrs[u]));
        }
        pispbe_wr(pispbe, PISP_BE_GLOBAL_BAYER_ENABLE, job->hw_enables[0]);
```

jmondi:
> Ah, ~perhaps~ this is the fix:
> 

Why I don't see it ?

` Linux version 6.7.0+ (jmondi@fedora) (aarch64-linux-gnu-gcc (GCC) 13.2.1 20230728 (Red Hat Cross 13.2.1-1), GNU ld version 2.39-4.fc38) #31 SMP PREEMPT Thu Feb  8 15:20:13 CET 2024`

```
pi@raspberrypi:~$ cam -c1 -C
[0:01:22.094832560] [982]  INFO Camera camera_manager.cpp:284 libcamera v0.1.0+118-563cd78e
[0:01:22.182109208] [987]  INFO RPI pisp.cpp:653 libpisp version v1.0.3 7d49ac3beab3 24-01-2024 (08:36:04)
[0:01:22.404264727] [987]  INFO RPI pisp.cpp:1112 Registered camera /base/axi/pcie@120000/rp1/i2c@80000/imx219@10 to CFE device /dev/media2 and ISP device /dev/media0 using PiSP variant BCM2712_C0
[0:01:22.405509467] [982]  WARN V4L2 v4l2_pixelformat.cpp:338 Unsupported V4L2 pixel format Y16 
[0:01:22.405751875] [982]  WARN V4L2 v4l2_pixelformat.cpp:338 Unsupported V4L2 pixel format RGB6
[0:01:22.405778079] [982]  WARN V4L2 v4l2_pixelformat.cpp:338 Unsupported V4L2 pixel format BGR6
[0:01:22.405818116] [982]  WARN V4L2 v4l2_pixelformat.cpp:338 Unsupported V4L2 pixel format PC1M
Using camera /base/axi/pcie@120000/rp1/i2c@80000/imx219@10 as cam0
[0:01:22.407122079] [982]  INFO Camera camera.cpp:1183 configuring streams: (0) 800x600-XRGB8888
[0:01:22.408457505] [987]  INFO RPI pisp.cpp:1396 Sensor: /base/axi/pcie@120000/rp1/i2c@80000/imx219@10 - Selected sensor format: 1640x1232-SBGGR10_1X10 - Selected CFE format: 1640x1232-PC1B
cam0: Capture until user interrupts by SIGINT
82.789795 (0.00 fps) cam0-stream0 seq: 000007 bytesused: 1920000
82.848400 (17.06 fps) cam0-stream0 seq: 000008 bytesused: 1920000
82.908455 (16.65 fps) cam0-stream0 seq: 000009 bytesused: 1920000
82.968481 (16.66 fps) cam0-stream0 seq: 000010 bytesused: 1920000
83.028505 (16.66 fps) cam0-stream0 seq: 000011 bytesused: 1920000
83.088546 (16.66 fps) cam0-stream0 seq: 000012 bytesused: 1920000
83.148504 (16.68 fps) cam0-stream0 seq: 000013 bytesused: 1920000
83.208769 (16.59 fps) cam0-stream0 seq: 000014 bytesused: 1920000
83.268785 (16.66 fps) cam0-stream0 seq: 000015 bytesused: 1920000
83.328856 (16.65 fps) cam0-stream0 seq: 000016 bytesused: 1920000
83.388942 (16.64 fps) cam0-stream0 seq: 000017 bytesused: 1920000
```

jmondi:
> Ah, ~perhaps~ this is the fix:
> 
> ```c
> diff --git a/drivers/media/platform/raspberrypi/pisp_be/pisp_be.c b/drivers/media/platform/raspberrypi/pisp_be/pisp_be.c
> index 07f0fb8aa85c..6e88491b20ba 100644
> --- a/drivers/media/platform/raspberrypi/pisp_be/pisp_be.c
> +++ b/drivers/media/platform/raspberrypi/pisp_be/pisp_be.c
> @@ -265,7 +265,7 @@ static void pispbe_queue_job(struct pispbe_dev *pispbe,
>         for (unsigned int u = 0; u < N_HW_ADDRESSES; ++u) {
>                 pispbe_wr(pispbe, PISP_BE_IO_ADDR_LOW(u),
>                           lower_32_bits(job->hw_dma_addrs[u]));
> -               pispbe_wr(pispbe, PISP_BE_IO_ADDR_HIGH(u) + 4,
> +               pispbe_wr(pispbe, PISP_BE_IO_ADDR_HIGH(u),
>                           upper_32_bits(job->hw_dma_addrs[u]));
>         }
>         pispbe_wr(pispbe, PISP_BE_GLOBAL_BAYER_ENABLE, job->hw_enables[0]);
> ```

ah yes, this is correct indeed, the + 4 is included in the macro definition.

Weird I can't trigger it :/

njhollinghurst:
> Weird I can't trigger it :/

Probably some kernel difference means you are running without IOMMU, and your addresses were all in the lower 32 bits.

naushir:
Another fix that is needed is 
```c
diff --git a/drivers/media/platform/raspberrypi/pisp_be/pisp_be_formats.h b/drivers/media/platform/raspberrypi/pisp_be/pisp_be_formats.h
index 0b9dcc7d52c5..51c8859723c2 100644
--- a/drivers/media/platform/raspberrypi/pisp_be/pisp_be_formats.h
+++ b/drivers/media/platform/raspberrypi/pisp_be/pisp_be_formats.h
@@ -27,6 +27,8 @@ struct pisp_be_format {

 #define V4L2_COLORSPACE_MASK(colorspace) BIT(colorspace)

+#define V4L2_COLORSPACE_MASK_JPEG \
+       V4L2_COLORSPACE_MASK(V4L2_COLORSPACE_JPEG)
 #define V4L2_COLORSPACE_MASK_SMPTE170M \
        V4L2_COLORSPACE_MASK(V4L2_COLORSPACE_SMPTE170M)
 #define V4L2_COLORSPACE_MASK_REC709    \
@@ -46,7 +48,8 @@ struct pisp_be_format {
  * producing an RGB format. In turn this requires us to allow all these colour
  * spaces for every YUV/RGB output format.
  */
-#define V4L2_COLORSPACE_MASK_ALL_SRGB (V4L2_COLORSPACE_MASK_SRGB |     \
+#define V4L2_COLORSPACE_MASK_ALL_SRGB (V4L2_COLORSPACE_MASK_JPEG |      \
+                                      V4L2_COLORSPACE_MASK_SRGB |      \
                                       V4L2_COLORSPACE_MASK_SMPTE170M | \
                                       V4L2_COLORSPACE_MASK_REC709)
```
Without this some of our colour space tests fail. Things should be fine for the compliance errors, as our default colour spaces are not JPEG and still SMTPE170M.

jmondi:
> Another fix that is needed is
> 
> ```c
> diff --git a/drivers/media/platform/raspberrypi/pisp_be/pisp_be_formats.h b/drivers/media/platform/raspberrypi/pisp_be/pisp_be_formats.h
> index 0b9dcc7d52c5..51c8859723c2 100644
> --- a/drivers/media/platform/raspberrypi/pisp_be/pisp_be_formats.h
> +++ b/drivers/media/platform/raspberrypi/pisp_be/pisp_be_formats.h
> @@ -27,6 +27,8 @@ struct pisp_be_format {
> 
>  #define V4L2_COLORSPACE_MASK(colorspace) BIT(colorspace)
> 
> +#define V4L2_COLORSPACE_MASK_JPEG \
> +       V4L2_COLORSPACE_MASK(V4L2_COLORSPACE_JPEG)
>  #define V4L2_COLORSPACE_MASK_SMPTE170M \
>         V4L2_COLORSPACE_MASK(V4L2_COLORSPACE_SMPTE170M)
>  #define V4L2_COLORSPACE_MASK_REC709    \
> @@ -46,7 +48,8 @@ struct pisp_be_format {
>   * producing an RGB format. In turn this requires us to allow all these colour
>   * spaces for every YUV/RGB output format.
>   */
> -#define V4L2_COLORSPACE_MASK_ALL_SRGB (V4L2_COLORSPACE_MASK_SRGB |     \
> +#define V4L2_COLORSPACE_MASK_ALL_SRGB (V4L2_COLORSPACE_MASK_JPEG |      \
> +                                      V4L2_COLORSPACE_MASK_SRGB |      \
>                                        V4L2_COLORSPACE_MASK_SMPTE170M | \
>                                        V4L2_COLORSPACE_MASK_REC709)
> ```
> 
> Without this some of our colour space tests fail. Things should be fine for the compliance errors, as our default colour spaces are not JPEG and still SMTPE170M.

I'll take it in and re-run v4l2-compliance

naushir:
With these 2 patches applied on top, all our tests are passing!

jmondi:
> With these 2 patches applied on top, all our tests are passing!

nice, thanks for testing and sorry for the `+ 4` which went unnoticed in my testing.

On top of which of your kernel branches would you like me to send a squashed version of these patches ?

One other thing, I'm working with @tomba to make sure the uAPI headers that I'll upstream for BE will be compatible with the ones used for his forthcoming version of the FE driver. The uAPI headers might change location (`include/uapi/media/` vs the current `include/uapi/linux/`). Just an heads-up!

naushir:
> On top of which of your kernel branches would you like me to send a squashed version of these patches ?

I think make a PR against rpi-6.6.y.  We will merge it to that and then backport to rpi-6.1.y manually.

popcornmix:
> I think make a PR against rpi-6.6.y. We will merge it to that and then backport to rpi-6.1.y manually.

We're not expecting to release any further 6.1 kernel updates to apt. so I don't think it's worth backporting.
We should be moving apt to 6.6 kernel soon enough.

naushir:
> We're not expecting to release any further 6.1 kernel updates to apt. so I don't think it's worth backporting. We should be moving apt to 6.6 kernel soon enough.

That also works for me!

drivers/media/platform/raspberrypi/pisp_be/pisp_be.c
```diff
@@ -0,0 +1,1972 @@
...
+			(BIT(CONFIG_NODE) | BIT(MAIN_INPUT_NODE)))
+		return -ENODEV;
+
+	node = &node_group->node[CONFIG_NODE];
+	spin_lock_irqsave(&node->ready_lock, flags);
+	buf[CONFIG_NODE] = list_first_entry_or_null(&node->ready_queue,
+						    struct pispbe_buffer,
+						    ready_list);
+	if (buf[CONFIG_NODE])
+		list_del(&buf[CONFIG_NODE]->ready_list);
```
njhollinghurst, line 481: What happens to this buffer now if `prepare_job()` subsequently fails because other nodes/buffers were not queued by the user? Is it lost forever?

drivers/media/platform/raspberrypi/pisp_be/pisp_be.c
njhollinghurst: I was concerned with the `list_del()` which looks like it could leak buffers in case of failure?

Normally the user will enqueue all the image buffers _before_ the config meta buffer, so this won't be reached. However, we don't want to allow a user error to cause a kernel error.

drivers/media/platform/raspberrypi/pisp_be/pisp_be.c
jmondi: Hi @njhollinghurst. Can we please move the comments to the ML, otherwise keeping this and the posted version in sync is difficult. 

Anyway, I guess the same question applies to all buffers which are extracted from a queue,  for each node. 
In the for loop that iterates over the video device we have

```
		/* Pull a buffer from each V4L2 queue to form the queued job */
		spin_lock_irqsave(&node->ready_lock, flags);
		buf[i] = list_first_entry_or_null(&node->ready_queue,
						  struct pispbe_buffer,
						  ready_list);
		if (buf[i]) {
			list_del(&buf[i]->ready_list);
			pispbe->queued_job.buf[i] = buf[i];
		}
		spin_unlock_irqrestore(&node->ready_lock, flags);

		if (!buf[i] && !ignore_buffers)
			return -ENODEV;
```

if we error out earlier with -ENODEV all buf[i] which has been extracted from a queue should probably be queued back ?

jmondi: Happy to revert `205735176035 ("be: Delete buffers from queue in a single loop")` if we have to push more fixes on this version. I'm not too bothered about the two loops, as said it should mostly be harmless. The worst that can happen is that a queue that resulted empty at inspection time (inside the first for loop) receives a new buffer in between the two loops and gets ignored. The other only conflicting path is a `stop_streaming`  while we're preparing the job, which returns all buffers from the queue in between these two for loops.

njhollinghurst: I'd find it helpful to discuss here just a little more, so I don't say something misguided on the ML.

Can you explain the concern about receiving a new buffer in between the two loops? What is the difference, in terms of locking and atomicity, between the two code versions?

Is there a better design in which we can push the removed buffers back onto the _front_ of the queues?

jmondi: > I'd find it helpful to discuss here just a little more, so I don't say something misguided on the ML.
> 

no worries

> Can you explain the concern about receiving a new buffer in between the two loops? What is the difference, in terms of locking and atomicity, between the two code versions?
>

The `ready_queue` queue is accessed in three locations
1) When preparing a job, to extract buffers from its front
2) When queuing a buffer to its tail
3) When stopping streaming to clear the queue

in 1) the queue is accessed in two points, one to inspect it if there's any available buffer, and at a second time to actually remove buffers from the queue. There's then a tiny window where the queue might be accessed by 2) and 3) before actually being modified.

As said, it should be harmless, but if in example the one of the nodes has no buffers queued at the time of inspection, then a buffers gets queued, it will get ignored and could potentially cause a mis-alignement.

In general, inspecting and modifying an object which is accessed by other concurrent code paths should be done while holding a lock, to reduce potentially hard to debug issues. If the solution turns out to be more complex than the original issue was then we can evaluate if it's worth it.

> Is there a better design in which we can push the removed buffers back onto the _front_ of the queues?

Isn't pushing in front a simple `list_add()` ?

njhollinghurst: I think I understand. So removing the buffers and (conditionally) adding them again on failure, seems like a clear improvement over the old two-loops design, but is still theoretically vulnerable to an intervening `STREAMOFF`. It might be good enough.

If we acquire all the queue locks (in a safe order!?), could it be possible to do the whole operation atomically?

jmondi: > I think I understand. So removing the buffers and (conditionally) adding them again on failure, seems like a clear improvement over the old two-loops design, but is still theoretically vulnerable to an intervening `STREAMOFF`. It might be good enough.
> 

`stop_streaming` is actually protected by `hw_lock`, and `pispbe_prepare_job` is run with the same lock held, so should be safe in that regard now that I look more closely.

So it's only about making sure `pispbe_prepare_job()` and `queue_buffer` do not step on each other. I think returning all buffers to the queue front in case of error is probably the safest option.

> If we acquire all the queue locks (in a safe order!?), could it be possible to do the whole operation atomically?

Do you mean  `pispbe_prepare_job()` with "the whole operation" ?

njhollinghurst: > Do you mean `pispbe_prepare_job()` with "the whole operation" ?

I did (at least, all of the queue inspection and state changes within that function). But it sounds like that is not necessary.

jmondi: probably not.. are you ok if I send an update with a cleanup routine that re-queues the buffers to the front of the queues in case of error ?

jmondi: @njhollinghurst I've re-pushed the pull request adding on top https://github.com/raspberrypi/linux/pull/5899/commits/9fcfb8ee070ecc2ca9b4fcf4629c4545f9cf178e

I triggered the error and tested with cam, which hangs as no job gets scheduled. I wonder if in the case of errors in creating the job description wouldn't be better to stop the streaming operations completely..

njhollinghurst: Thanks - I'll try to do some testing this afternoon.

From a first look at the code, are all of `buf[]` initialized when we get to `err_return_buffers`?

jmondi: Should they ? Only the initialized ones are relevant, as they are initialized with buffers extracted from the ready_list queue. If  `buf[i]` has not been initialized it means no buffer have been extracted from the queue. Did I get your question correctly ?

njhollinghurst: If we terminate the loop with `goto`, it looks like some value of `buf[i]` will contain an undefined value.

`buf[]` might need to be initialized before the loop, or upon termination, or maybe the cleanup could be done inside the loop (without a goto) so it can see the current value of the loop counter?

jmondi: ah right, `struct pispbe_buffer *buf[PISPBE_NUM_NODES];` is not initialized.

From memory
`pispbe_buffer *buf[PISPBE_NUM_NODES] = {}; `

should be valid C99, otherwise we can easily memset() ?

njhollinghurst: Yes, that seems to work!

One way to test this stuff is to modify `src/libcamera/pipeline/rpi/pisp/pisp.cpp:prepareBe()`, to move the line
`        isp_[Isp::Input].queueBuffer(buffer)`
to the bottom of the function, after the `Config` buffer has been queued. (This is not normal user behaviour.)

Without initializing `buf[]`, that would result in a hard crash, dereferencing a bad virtual address in `pispbe_prepare_job()`. With the empty initializer, it appears to be fixed.

jmondi: Thanks for testing. Let me pass ` = {}` through the compiler matrix to make sure it doesn't cause issues with exotic arch compilers, in which case I'll memset().

I'll include this fix in v2 of the patches on the ML, thanks

jmondi:
I guess this one can safely be closed now!

PR #5900, opened 2024-01-26
drm/rp1: depends on, instead of select, MFD_RP1

peat-psuwit:
According to kconfig-language.txt [1], select should be used only for "non-visible symbols ... and for symbols with no dependencies". Since MFD_RP1 both is visible and has a dependency, "select" should not be used and "depends on" should be used instead.

In particular, this fixes the build of this kernel tree on NixOS, where its kernel config system will try to answer 'M' to as many config as possible.

[1] https://www.kernel.org/doc/html/latest/kbuild/kconfig-language.html


6by9:
Agreed - DPI and DSI drivers depend on MFD_RP1, and shouldn't select it.

peat-psuwit:
Ping the developer.

PR #5902, opened 2024-01-26
ARM: dts: Standardise downstream Pi GPIO pin names

pelwell:
Standardise the names of the pins on the 40-pin GPIO header. This makes it easier to use libgpiod (and the gpiod tools) for cross-platform GPIO manipulation.


pelwell:
@Gadgetoid This is prompted by your comment about having to use different names on different platforms.

There's a chance that some third-party app is using one of the old alt-function-based names ("SPI_MOSI", "TXD0", etc.), but I think it's a slim chance...

Gadgetoid:
Day late and a dollar short here (we're all battling covid 😭) but thank you - this should simplify my middleware and make things a little less confusing.

PR #5908, opened 2024-01-29
drivers: media: pisp_be: Update seqeuence numbers of the buffers

naushir:
Add a framebuffer sequence counter and increment on every completed job. This counter is then used to update the VB2 buffer sequence count before calling vb2_buffer_done().


naushir:
@njhollinghurst and @davidplowman.

@jmondi this will be of interest to you as well for adding it into your upstreaming branch.

jmondi:
Thanks, I can confirm the issue and that this patch fixes it. I'll collect it!

PR #5910, opened 2024-01-30
iommu/bcm2712: don't allow building as module

peat-psuwit:
Since bcm2712-iommu{,-cache}.c doesn't have usual module descriptors such as `MODULE_LICENSE`, configuring this as 'M' fails the build with `ERROR: modpost: missing MODULE_LICENSE() in <...>/bcm2712-iommu.o`. Since it seems like the code is not intended to be built as a module anyway (it registers the driver with `builtin_platform_driver()`), don't allow building this code as a module.

### DIFF
--- a/drivers/iommu/Kconfig
+++ b/drivers/iommu/Kconfig
@@ -507,7 +507,7 @@ config SPRD_IOMMU
 	  Say Y here if you want to use the multimedia devices listed above.
 
 config BCM2712_IOMMU
-       tristate "BCM2712 IOMMU driver"
+       bool "BCM2712 IOMMU driver"
        depends on ARM64 && ARCH_BCM
        select IOMMU_API
        help


njhollinghurst:
I don't know much about Linux configuration -- I probably copied the original from one of the other entries (and without spotting that the one above was virtual). Yes, it seems entirely reasonable for it _not_ to be a module.

pelwell:
Are you saying it shouldn't be a module? Is is not possible to get to point where the root filesystem are mounted without it?

6by9:
> Are you saying it shouldn't be a module? Is is not possible to get to point where the root filesystem are mounted without it?

It's not going to be very useful as a module on Raspberry Pi OS as that precludes NVM (PCIe), USB or NFS (PCIe and RP1) booting which will both go through iommus.
I don't recall for SD card boot, but think that is plausible without iommus. IMHO Being able to build as a module is therefore valid, but needs the relevant entries.

Looking through drivers/iommu/Kconfig, it looks like very few allow tristate, so likewise making it a bool isn't an invalid option.

pelwell:
OK - thanks, all.

Issue #5914, opened 2024-01-31
Pi 5 I2C timing out of spec


pelwell:
### Describe the bug

As detailed in this thread (https://forums.raspberrypi.com/viewtopic.php?p=2187613#p2187613), the new Pi 5 I2C timing doesn't adhere to the specification. In particular, the SDA line should not start to change until the SCL line has dropped below the 30% level (i.e. become a clear digital low).

### Steps to reproduce the behaviour

Use I2C, look at scope traces.

### Device (s)

Raspberry Pi 5

### System

Seen all all current Linux branches.

pelwell:
There's now a Pull Request that should help with the I2C timing. See https://github.com/raspberrypi/linux/pull/5916.

There should be a test build available in the next 45 minutes (it should definitely be done by 1900 GMT). After that, run "sudo rpi-update pulls/5916" to install it. However:
1. It's beta software, so should be used with caution.
2. It's a 6.6 kernel, which might be too big a change for you.
3. It's only a Device Tree change, so a full kernel install is overkill.

Alternatively, download just the updated Pi 5 dtb file from here: https://drive.google.com/file/d/1NGJsJoeFf_K8mVi0lNjO6FIPEOYj9lvf/view?usp=sharing

Copy it to /boot/firmware/ (or just /boot on an older image), reboot, and see how you get on.

dgedgedge:
This patch correctes our issue.
I did put a scope on the I2C bus, there is now a 300ns TsuDAT/ThDAT guard on the signal.
See discussion on forum [https://forums.raspberrypi.com/viewtopic.php?p=2187613#p2187613](https://forums.raspberrypi.com/viewtopic.php?p=2187613#p2187613 )
Sorry for the answer delay.

PR #5916, opened 2024-01-31
Improve I2C timing (again)

pelwell:
Improve I2C timing by:
1. setting the SDA hold time to 300ns, and
2. increasing the drive strength to improve the signal quality at 1MHz.

Prompted by #5914.

### DIFF
--- a/arch/arm/boot/dts/broadcom/rp1.dtsi
+++ b/arch/arm/boot/dts/broadcom/rp1.dtsi
@@ -307,6 +307,7 @@
 			clocks = <&rp1_clocks RP1_CLK_SYS>;
 			i2c-scl-rising-time-ns = <65>;
 			i2c-scl-falling-time-ns = <100>;
+			i2c-sda-hold-time-ns = <300>;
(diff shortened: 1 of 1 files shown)


pelwell:
@JinShil You may want to give this PR a try, as the fault it corrects might have been the cause of your problems with the 9306.

JinShil:
I've tested this on my existing hardware with the 9306.  It results in the same symptoms described in #5853, when the hardware is running out of spec at 1MHz.  But up to and including 400kHz it works fine.  I'm still waiting on new hardware to test a 1MHz clock running within spec.

pelwell:
Thanks for trying.

JinShil:
I tested a new I2C device with this PR today. The device is a GP8403 DAC.

With #5802, there were quite a few errors at both 100kHz and 400kHz.  However, after updating to this PR, the errors vanished.  So, this PR appears to be a significant improvement.

I'm still waiting on 1MHz hardware, and will test that when I get my hands on it.

PR #5917, opened 2024-02-01
ASoC: DACplus - fix 16bit sample support in clock consumer mode

j-schambacher:
The former code did not adjust the physical sample width when in clock consumer mode and has taken the fixed 32 bit default. This has caused the audio to be played at half its frequency due to the fixed bclk_ratio of 64.

Problem appears only on PI5 as on the former PIs the I2S module did simply run at fixed 64x rate.


j-schambacher:
Thanks, Phil! Will do the same for DAC+ADC Pro.

PR #5918, opened 2024-02-01
ASoC: adds support for Hifiberry AMP4Pro to the dacplus driver

j-schambacher:
The AMP4 Pro is a I2S master mode capable amplifier with
providing all I2S clocks.
The card driver can be shared between TAS575x amplifiers
and the PCM512x DACs as they are SW compatible.
From a HW perspective though we need to limit the sample
rates to the standard audio rates to avoid running the
onboard clocks through the PLL. Using the PLL would require
even a different HW.
DAI/stream name are also set accordingly to allow the user
a convenient identification of the soundcard

Needs the pcm512x driver with TAS575x support (already in
upstream & raspberry kernels).


j-schambacher:
...updated with the missing README inputs

PR #5922, opened 2024-02-05
staging: bcm2835-codec: Fix up for 6.8 - use ignore_cap_streaming

6by9:
Drops downstream patch to v4l2_mem2mem, and uses the new mainline flag to achieve the same functionality


popcornmix:
Merged manually.

PR #5924, opened 2024-02-05
Fixes for pcie-brcmstb on 6.8

P33M:
Not entirely sure how we want this to look in the history - I assume once the downstream reversions are all squashed together, we go back to having one mega-commit for pcie-brcmstb.


popcornmix:
Thanks - a quick test worked.

I suspect pcie-brcmstb will be squashed until such a time when useful bits are pulled out, cleaned up and submitted upstream.

Issue #5927, opened 2024-02-06
Wolfson wm8731 audio codec broken between 6.1.73 and 6.1.74


rafael2k:
### Describe the bug

6.1.74 broke the wm8731 "snd_soc_audioinjector_pi_soundcard".  
6.1.73 was working fine. I'm running in a Pi 4, latest Raspberry Pi OS, 64 bit, bookworm.
I'm getting same kind of errors:
```
[   84.950454] bcm2835-i2s fe203000.i2s: I2S SYNC error!
[  138.433994] bcm2835-i2s fe203000.i2s: I2S SYNC error!
[  211.219306] bcm2835-i2s fe203000.i2s: ASoC: error at snd_soc_dai_hw_params on fe203000.i2s: -22
[  211.220038]  AudioInjector audio: ASoC: error at __soc_pcm_hw_params on AudioInjector audio: -22
[  215.901118] bcm2835-i2s fe203000.i2s: ASoC: error at snd_soc_dai_hw_params on fe203000.i2s: -22
[  215.902089]  AudioInjector audio: ASoC: error at __soc_pcm_hw_params on AudioInjector audio: -22
```
This is similar to bug https://github.com/raspberrypi/linux/issues/5817 , but got fixed. Now it is happening also in the Pi 4.

### Steps to reproduce the behaviour

Just boot the system.

### Device (s)

Raspberry Pi 4 Mod. B

### System

```
Oct 17 2023 15:39:16 
Copyright (c) 2012 Broadcom
version 30f0c5e4d076da3ab4f341d88e7d505760b93ad7 (clean) (release) (start)
```
```
Linux estacao6.hermes.radio 6.1.74-v8+ #1725 SMP PREEMPT Mon Jan 22 13:35:32 GMT 2024 aarch64 GNU/Linux
```

### Logs

Write or read from alsa device does not work.
Sound card is stuck, just like in previous bug.

rafael2k:
I have another SD card with this kernel:
Linux estacao9.hermes.radio 6.1.0-rpi8-rpi-v8 #1 SMP PREEMPT Debian 1:6.1.73-1+rpt1 (2024-01-25) aarch64 GNU/Linux

Which is working.

pelwell:
[...]

PR #5928, opened 2024-02-06
Add I2S audio debug output for #5927

pelwell:
See: https://github.com/raspberrypi/linux/issues/5927


pelwell:
No longer needed.

PR #5930, opened 2024-02-06
vc4/hvs: Support fixed alpha correctly on 2712D0

popcornmix:
2712D0 removed alpha_mode from control word 2 for choosing fixed alpha and replaced it with the previously reserved value of 3 in alpha_mask.

Handle this to fix corrupt desktop when using X on 2712D0


popcornmix:
Ping @6by9 @lurch

6by9:
Looks good.
Really minor nit would be for a blank line between the conditional and unconditional `return` lines.

lurch:
Yup, after doing `sudo rpi-update pulls/5930; sudo reboot` X11 is now working perfectly fine on my 2712D0, both with and without glamor enabled :tada:

popcornmix:
Fixed nit.

PR #5937, opened 2024-02-09
SDHCI vs EMMC on Pi 5

P33M:
This fixes writes on the Raspikey for me and corrects a few more benign bugs.

I still can't boot from it, though - something goes wrong when attempting to switch MMC partitions - unrelated to IO errors.


pelwell:
Is this worth merging as is, or would you rather wait until booting works?

P33M:
This is an improvement (and gets CM5 HS400 working) - the raspikey is a separate issue.

geerlingguy:
Mmm... if I'm reading correctly, this could greatly speed up flashing compute module eMMC? Probably my number one pain point with CM4/eMMC modules.

PR #5938, opened 2024-02-09
imx477: make trigger-mode more configurable

erikboto:
Currently there's no good way to have two imx477 cameras and configure them so that one is vsync source and the other vsync sink, since the configuration is done as a module parameter affecting both instances.

In our case we have a Raspberry Pi5 with two HQ cameras connected (imx477 based), and we want to use the XVS pins to allow them to capture in synchronously. With these changes we can now use the dtoverlays to configure the function of the XVS pin individually.

### DIFF
--- a/arch/arm/boot/dts/overlays/README
+++ b/arch/arm/boot/dts/overlays/README
@@ -862,6 +862,10 @@ Params: cam0-arducam-64mp       Select Arducam64MP for camera on port 0
         cam1-ov7251             Select OV7251 for camera on port 1
         cam1-ov9281             Select OV9281 for camera on port 1
         cam1-imx290-clk-freq    Set clock frequency for an IMX290 on port 1
+        cam0-sync-source        Set camera on port 0 as vsync source
+        cam0-sync-sink          Set camera on port 0 as vsync sink
+        cam1-sync-source        Set camera on port 1 as vsync source
+        cam1-sync-sink          Set camera on port 1 as vsync sink
 
         cam0                    Connect the mux to CAM0 port (default is CAM1)
 
@@ -923,6 +927,14 @@ Params: cam0-arducam-64mp       Select Arducam64MP for camera on port 0
         cam3-ov7251             Select OV7251 for camera on port 3
         cam3-ov9281             Select OV9281 for camera on port 3
         cam3-imx290-clk-freq    Set clock frequency for an IMX290 on port 3
+        cam0-sync-source        Set camera on port 0 as vsync source
+        cam0-sync-sink          Set camera on port 0 as vsync sink
+        cam1-sync-source        Set camera on port 1 as vsync source
+        cam1-sync-sink          Set camera on port 1 as vsync sink
+        cam2-sync-source        Set camera on port 2 as vsync source
+        cam2-sync-sink          Set camera on port 2 as vsync sink
+        cam3-sync-source        Set camera on port 3 as vsync source
+        cam3-sync-sink          Set camera on port 3 as vsync sink
 
         cam0                    Connect the mux to CAM0 port (default is CAM1)
 
@@ -2630,6 +2642,8 @@ Params: rotation                Mounting rotation of the camera sensor (0 or
                                 Compute Module (CSI0, i2c_vc, and cam0_reg).
         always-on               Leave the regulator powered up, to stop the
                                 camera clamping I/Os such as XVS to 0V.
+        sync-source             Configure as vsync source
+        sync-sink               Configure as vsync sink
 
 
 Name:   imx462
@@ -2670,6 +2684,8 @@ Params: rotation                Mounting rotation of the camera sensor (0 or
                                 Compute Module (CSI0, i2c_vc, and cam0_reg).
         always-on               Leave the regulator powered up, to stop the
                                 camera clamping I/Os such as XVS to 0V.
+        sync-source             Configure as vsync source
+        sync-sink               Configure as vsync sink
 
 
 Name:   imx519
--- a/arch/arm/boot/dts/overlays/camera-mux-2port-overlay.dts
+++ b/arch/arm/boot/dts/overlays/camera-mux-2port-overlay.dts
@@ -536,5 +536,10 @@
 
 		cam0 = <&i2c_frag>, "target:0=",<&i2c_csi_dsi0>,
 		       <&csi_frag>, "target:0=",<&csi0>;
+
+		cam0-sync-source = <&imx477_0>, "trigger-mode:0=1";
+		cam0-sync-sink = <&imx477_0>, "trigger-mode:0=2";
+		cam1-sync-source = <&imx477_1>, "trigger-mode:0=1";
+		cam1-sync-sink = <&imx477_1>, "trigger-mode:0=2";
 	};
 };
--- a/arch/arm/boot/dts/overlays/camera-mux-4port-overlay.dts
+++ b/arch/arm/boot/dts/overlays/camera-mux-4port-overlay.dts
@@ -939,5 +939,14 @@
 
 		cam0 = <&i2c_frag>, "target:0=",<&i2c_csi_dsi0>,
 		       <&csi_frag>, "target:0=",<&csi0>;
+
+		cam0-sync-source = <&imx477_0>, "trigger-mode:0=1";
+		cam0-sync-sink = <&imx477_0>, "trigger-mode:0=2";
+		cam1-sync-source = <&imx477_1>, "trigger-mode:0=1";
+		cam1-sync-sink = <&imx477_1>, "trigger-mode:0=2";
+		cam2-sync-source = <&imx477_2>, "trigger-mode:0=1";
+		cam2-sync-sink = <&imx477_2>, "trigger-mode:0=2";
+		cam3-sync-source = <&imx477_3>, "trigger-mode:0=1";
+		cam3-sync-sink = <&imx477_3>, "trigger-mode:0=2";
 	};
 };
--- a/arch/arm/boot/dts/overlays/imx378-overlay.dts
+++ b/arch/arm/boot/dts/overlays/imx378-overlay.dts
@@ -8,3 +8,10 @@
 &cam_node {
 	compatible = "sony,imx378";
 };
+
+/{
+	__overrides__ {
+		sync-sink = <&cam_node>,"trigger-mode:0=2";
+		sync-source = <&cam_node>,"trigger-mode:0=1";
+	};
+};
--- a/arch/arm/boot/dts/overlays/imx477-overlay.dts
+++ b/arch/arm/boot/dts/overlays/imx477-overlay.dts
@@ -8,3 +8,10 @@
 &cam_node {
 	compatible = "sony,imx477";
 };
+
+/{
+	__overrides__ {
+		sync-sink = <&cam_node>,"trigger-mode:0=2";
+		sync-source = <&cam_node>,"trigger-mode:0=1";
+	};
+};
--- a/drivers/media/i2c/imx477.c
+++ b/drivers/media/i2c/imx477.c
@@ -1124,6 +1124,9 @@ struct imx477 {
 	/* Current mode */
 	const struct imx477_mode *mode;
 
+	/* Trigger mode */
+	int trigger_mode_of;
+
 	/*
 	 * Mutex for serialized access:
 	 * Protect sensor module set pad format and start/stop streaming safely.
@@ -1711,7 +1714,7 @@ static int imx477_start_streaming(struct imx477 *imx477)
 	struct i2c_client *client = v4l2_get_subdevdata(&imx477->sd);
 	const struct imx477_reg_list *reg_list;
 	const struct imx477_reg_list *extra_regs;
-	int ret;
+	int ret, tm;
 
 	if (!imx477->common_regs_written) {
 		ret = imx477_write_regs(imx477, mode_common_regs,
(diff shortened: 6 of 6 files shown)


6by9:
Thanks for the patch.

Removing the module parameter causes a regression for anyone who is already using it.
Initialise `imx477->trigger_mode = trigger_mode`, and then try reading from device tree to override that.

I'm not convinced updating the camera mux overlays is worth it as you can only run one of the sensors at a time. I guess both as sinks off a master on another interface would be applicable.
I'd suggest that you create generic `cam[01]-sync-source` and `cam[01]-sync-sink overrides, and have those set the appropriate property on all drivers that support configuring external trigger (currently only imx477, but we have other sensors that could support it). You'd therefore load with something like `dtoverlay=camera-mux-2port,cam0-imx477,cam0-sync-source,cam1-imx477,cam1-sync-sink`

Why only update camera-mux-2port and not camera-mux-4port?

imx477-overlay.dts change is incorrect - it should be an override (and documented), not forced to 0. It probably makes more sense to put it in imx477_378-overlay.dtsi instead of imx477-overlay.dts. I'm expecting imx378 to behave in the same manner as imx477 as it is very closely related. README for imx378 would then need updating too.

Finally driver changes need to be separate from overlay changes. Overlays in particular frequently get squashed down, and that's impossible if driver and overlay changes are intermingled.

erikboto:
First off, thanks for the fast feedback!

> Thanks for the patch.
> 
> Removing the module parameter causes a regression for anyone who is already using it. Initialise `imx477->trigger_mode = trigger_mode`, and then try reading from device tree to override that.

Right, I've fixed that in a follow up commit now (to be squashed later).

> 
> I'm not convinced updating the camera mux overlays is worth it as you can only run one of the sensors at a time. I guess both as sinks off a master on another interface would be applicable. I'd suggest that you create generic `cam[01]-sync-source` and `cam[01]-sync-sink overrides, and have those set the appropriate property on all drivers that support configuring external trigger (currently only imx477, but we have other sensors that could support it). You'd therefore load with something like `dtoverlay=camera-mux-2port,cam0-imx477,cam0-sync-source,cam1-imx477,cam1-sync-sink`

That sounds like a good idea. I'll try to fix this in a nicer way as you suggest.
> 
> Why only update camera-mux-2port and not camera-mux-4port?

Oversight on my side, I'll update the camera-mux-4port as well although I don't have any HW to test it with.
> 
> imx477-overlay.dts change is incorrect - it should be an override (and documented), not forced to 0. It probably makes more sense to put it in imx477_378-overlay.dtsi instead of imx477-overlay.dts. I'm expecting imx378 to behave in the same manner as imx477 as it is very closely related. README for imx378 would then need updating too.

Perhaps I shouldn't change imx477-overlay.dts (or imx477_378-overlay.dtsi)  at all, so that there's no value by in device tree by default to allow the regular module parameter to work as the default? Or is it possible to add it as an override there without any default value, just as a way to document that it's possible to use it? I'm quite new to device tree, so I'm not really sure what's possible and not yet.

> 
> Finally driver changes need to be separate from overlay changes. Overlays in particular frequently get squashed down, and that's impossible if driver and overlay changes are intermingled.

Ok, I'll split it up into different commits.

arch/arm/boot/dts/overlays/README
```diff
@@ -854,6 +856,8 @@ Params: cam0-arducam-64mp       Select Arducam64MP for camera on port 0
         cam1-imx258             Select IMX258 for camera on port 1
         cam1-imx290             Select IMX290 for camera on port 1
         cam1-imx477             Select IMX477 for camera on port 1
+        cam0-imx477-source      Select IMX477 for camera on port 1 as vsync source
```
pelwell: `cam1-imx477-source`

arch/arm/boot/dts/overlays/README
```diff
@@ -854,6 +856,8 @@ Params: cam0-arducam-64mp       Select Arducam64MP for camera on port 0
...
+        cam0-imx477-sink        Select IMX477 for camera on port 1 as vsync sink
```
pelwell: `cam1-imx477-sink`

pelwell (commented):
Fix the README, separate driver changes from overlay/README changes, and we may be in a position to merge.

6by9:
> Perhaps I shouldn't change imx477-overlay.dts (or imx477_378-overlay.dtsi) at all, so that there's no value by in device tree by default to allow the regular module parameter to work as the default? Or is it possible to add it as an override there without any default value, just as a way to document that it's possible to use it? I'm quite new to device tree, so I'm not really sure what's possible and not yet.

Having it supported through the mux overlays but not directly would be a very odd decision.
Pi5 in particular has 2 MIPI ports, so could have
```
dtoverlay=imx477,sync-source
dtoverlay=imx477,cam0,sync-sink
```

of_property_read_u8 is implemented via of_property_read_u8_array with an array size of 1.
[As documented](https://elixir.bootlin.com/linux/latest/source/include/linux/of.h#L1260)
```
* Return: 0 on success, -EINVAL if the property does not exist,
 * -ENODATA if property does not have a value, and -EOVERFLOW if the
 * property data isn't large enough.
 *
 * The out_values is modified only if a valid u8 value can be decoded.
```
So having the DT parameter absent is valid and easily supported (don't throw an error if of_property_read_u8 returns one), and an override is perfectly permitted to add additional entries to overlay.

erikboto:
> > Perhaps I shouldn't change imx477-overlay.dts (or imx477_378-overlay.dtsi) at all, so that there's no value by in device tree by default to allow the regular module parameter to work as the default? Or is it possible to add it as an override there without any default value, just as a way to document that it's possible to use it? I'm quite new to device tree, so I'm not really sure what's possible and not yet.
> 
> Having it supported through the mux overlays but not directly would be a very odd decision. Pi5 in particular has 2 MIPI ports, so could have
> 
> ```
> dtoverlay=imx477,sync-source
> dtoverlay=imx477,cam0,sync-sink
> ```
> 
> of_property_read_u8 is implemented via of_property_read_u8_array with an array size of 1. [As documented](https://elixir.bootlin.com/linux/latest/source/include/linux/of.h#L1260)
> 
> ```
> * Return: 0 on success, -EINVAL if the property does not exist,
>  * -ENODATA if property does not have a value, and -EOVERFLOW if the
>  * property data isn't large enough.
>  *
>  * The out_values is modified only if a valid u8 value can be decoded.
> ```
> 
> So having the DT parameter absent is valid and easily supported (don't throw an error if of_property_read_u8 returns one), and an override is perfectly permitted to add additional entries to overlay.

The updated imx477.c now uses the parameter as default, but will use any value from DT to override it if present. My worry was how I can make it apparent in the DT that you can provide a value if you want, but not provide a default value in DT since that would make the module parameter to be always overridden which breaks the backwards compatibility with the current functionality.

erikboto:
I've split the DT changes and the driver changes now. But I'll also try to address the issues from @6by9 to make it a bit nicer.

6by9:
You've already done it in camera-mux-2port-overlay.dts
```
__overrides__ {
   	cam0-imx477-sink = <&mux_in0>, "remote-endpoint:0=",<&imx477_0_ep>,
			      <&imx477_0_ep>, "remote-endpoint:0=",<&mux_in0>,
			      <&mux_in0>, "clock-noncontinuous?",
			      <&imx477_0>, "status=okay",
				  <&imx477_0>, "trigger-mode:0=2";
```
You've added `trigger-mode:0=2` to imx477_0 when the property wasn't present at all before.

Add to imx477_378-overlay.dtsi, in the overrides section:
```
		sync-sink= <&cam_node>, "trigger-mode:0=1";
		sync-source = <&cam_node>, "trigger-mode:0=2";
```

njhollinghurst:
Currently `trigger_mode` is sampled when the camera starts streaming, and can be changed. It is possible (though admittedly awkward) to have both a source and a sink on the same Pi 5 (or CM4). It is even possible to change the cameras' roles at runtime (after probe), which would become more difficult with the proposed change.

Edit: In fact it looks like this change would break the [currently documented method](https://www.raspberrypi.com/documentation/accessories/camera.html#boot-up-both-raspberry-pis), in which the user writes to the parameter after booting.

If we make a change for imx477, should we do something analogous for imx296? Noting that `trigger_mode` has a rather different interpretation on imx296, where we might want to call the override "ext" instead of "sink" (or some such).

erikboto:
I think I've finally understood the devicetree overlays a bit better and have something that works well and is extendable for the camera-mux-N-port cases.

I couldn't get the __overrides__ section work in imx477_378.dtsi since it is included from camera-mux- and then would cause syntax issues, so it's added to both imx477-overlay.dts and imx378-overlay.dts instead.

As @njhollinghurst writes the module parameter is read during stream start, which is not the case with my current changes. If we really want to keep the old way of working where the parameter could be changed at runtime we could keep track in the module if the trigger-mode was set it device tree and then always use that value, but if there was no device tree value we evaluate the module parameter at stream start. It adds a bit of complexity, and I wonder if it's worth it or not?

I don't think the current version of my changes will break the currently documented version though, since that is about using it on different Pi's and not about changing the setting after starting a stream. But I could add the changes to keep track of if the setting comes from device tree or not if you feel it's the correct approach.

njhollinghurst:
Well the documentation needs review, but currently it explicitly tells the user to change `trigger_mode` at runtime ([link](https://www.raspberrypi.com/documentation/accessories/camera.html#boot-up-both-raspberry-pis)).

We would have to synchronize a kernel change with a documentation change...

Edit: I am not averse to doing this, we need to change the doc anyway to edit config.txt for `always-on`. And as a hardware configuration it make sense to have this in the DT.

Should we do something similar for imx296? Except that trigger-mode means something different there.

arch/arm/boot/dts/overlays/README
```diff
@@ -862,6 +862,10 @@ Params: cam0-arducam-64mp       Select Arducam64MP for camera on port 0
         cam1-ov7251             Select OV7251 for camera on port 1
         cam1-ov9281             Select OV9281 for camera on port 1
         cam1-imx290-clk-freq    Set clock frequency for an IMX290 on port 1
+        cam0-sync-source        Set camera on port 0 as vsync source (for supported cameras)
```
pelwell: These lines are too long - I think you could drop the `(for supported cameras)` proviso.

erikboto:
> Well the documentation needs review, but currently it explicitly tells the user to change `trigger_mode` at runtime ([link](https://www.raspberrypi.com/documentation/accessories/camera.html#boot-up-both-raspberry-pis)).
> 

Ah, now I see what you mean! 

> We would have to synchronize a kernel change with a documentation change...
> 
> Edit: I am not averse to doing this, we need to change the doc anyway to edit config.txt for `always-on`. And as a hardware configuration it make sense to have this in the DT.
> 
> Should we do something similar for imx296? Except that trigger-mode means something different there.

I had a quick look into the docs + kernel code for the imx296, and it looks to me like it could make sense to also allow setting trigger mode via DT. I didn't find it in the camera-mux-N-port files, but if it's also added there I guess it could also fit into the camN-sync-sink overrides.

arch/arm/boot/dts/overlays/README
erikboto: Right, I removed that now and also fixed the missing sign-off lines

njhollinghurst:
imx296 is complicated by having 2 different mechanisms for synchronizing cameras -- one selected by a link and the other by a register setting.

So on second thoughts I now think that should be a separate change, and need not impede your PR.

njhollinghurst:
Should we in fact remove the old module parameter? Its presence will "fool" users into thinking they can follow the old instructions and change it at run-time. Better to break it noisily than silently.

I suppose the answer will depend on whether anybody is setting it from the command line or manually probing the module....

It might also be possible to combine a DT and a module parameter in a more dynamic way, e.g. by defaulting the module parameter to -1 and substituting the DT property in that case.

erikboto:
Perhaps the suggestion with a default module param value of -1 is best, at least it will keep todays functionality untouched. Just let me know if that is the solution you feel is most appropriate and I can implement it.

njhollinghurst:
How about this suggestion:

The DT/OF property has narrower scope than the module parameter, so it should take priority when it exists. But when it's not present, default to the _original_ behaviour of sampling the module parameter at run-time.

- In `probe()`: Use either a reserved value (such as -1) for `imx477->trigger_mode`, or a new `bool` flag, to indicate whether or not the OF property was found.
- In `start_streaming()`, use `int tm = (imx47->trigger_mode >= 0) ? imx477->trigger_mode : trigger_mode;`

Edit: thought I guess that might be confusing. Maybe a comment wouldn't go amiss, or maybe rename one of those variables (but not the property or parameter names) to clarify?

pelwell:
The downside with that scheme is that it is easier to add/remove a module parameter than it is to add/remove a DT property.

njhollinghurst:
> The downside with that scheme is that it is easier to add/remove a module parameter than it is to add/remove a DT property.

Yes. But it preserves the existing documented behaviour, without either having to change the default module parameter value or expose the new "magic" value to the user, or else changing the semantics of value zero...

pelwell:
>  it is easier to add/remove a module parameter than it is to add/remove a DT property.

This is true, but not relevant if the property only exists when the parameter is used (i.e. the parameter creates the property, rather than changing an existing value). Provided the overlay parameter trumps the module parameter then I'm happy. If somebody want to control trigger mode from the module parameter then they just need to not use the overlay parameter.

njhollinghurst:
OK then -- I'm happy to approve a version that adds these DT/OF properties, while still allowing the user to change trigger mode after boot using sysfs (as [previously documented](https://www.raspberrypi.com/documentation/accessories/camera.html#boot-up-both-raspberry-pis)).

I don't mind which mechanism (OF or module parameter) has priority, as long as it's clear.  If any new parameter values are exposed to the user, they should be documented (e.g. in `MODULE_PARM_DESC`).

erikboto:
I've updated the driver now to keep the original behavior if trigger-mode is not set via DT/OF. 

I also found an issue where of_property_read_u8() didn't read the value correctly since it requires the values in OF to be prefixed with "/bits/ 8" for it to work properly, so it always read 0 instead of the correct value. This is fixed by instead reading the value as u32.

pelwell (approved):

njhollinghurst:
> I've updated the driver now to keep the original behavior if trigger-mode is not set via DT/OF.
> 
> I also found an issue where of_property_read_u8() didn't read the value correctly since it requires the values in OF to be prefixed with "/bits/ 8" for it to work properly, so it always read 0 instead of the correct value. This is fixed by instead reading the value as u32.

Looks good. I will test it with a Pi 5 and dual IMX477s.
Unless you have a particular need for this on Linux 6.1, I may try re-targeting it to rpi-6.6.y

njhollinghurst (approved):
Works as expected for Pi 5 with IMX477. I have not tested with a camera MUX.

erikboto:
> > I've updated the driver now to keep the original behavior if trigger-mode is not set via DT/OF.
> > I also found an issue where of_property_read_u8() didn't read the value correctly since it requires the values in OF to be prefixed with "/bits/ 8" for it to work properly, so it always read 0 instead of the correct value. This is fixed by instead reading the value as u32.
> 
> Looks good. I will test it with a Pi 5 and dual IMX477s. Unless you have a particular need for this on Linux 6.1, I may try re-targeting it to rpi-6.6.y

We currently solved it using a more hackish way since we wanted a quick fix, but I wanted to solve it properly upstream since I feel it's the right thing to do. So if rpi-6.6 .y is more convenient I'm fine with that.

njhollinghurst:
Thanks!

It didn't auto-rebase with 6.6 which is not surprising. I'll merge it into 6.1 and then port to 6.6.

PR #5945, opened 2024-02-12
brcmfmac: Add sae_ext WPA3 support

pelwell:
This is a minimal set of patches to add WPA3 support to WLAN firmwares with the `sae_ext` (*) feature, that works with `wpa_supplicant`. A compatible firmware can be download from this Cypress release package: https://community.infineon.com/t5/Wi-Fi-Bluetooth-for-Linux/Cypress-Linux-WiFi-Driver-Release-FMAC-2023-09-01/td-p/492862

For convenience you can download it directly from here: https://drive.google.com/file/d/13u_Ipf6yUATl38HyVNiHG3H2BVxCzfkB/view?usp=drive_link

It is suitable for use on Pi 5, Pi 4 and CM4. Install it by coping to /etc/firmware/cypress/cyfmac43455-sdio-standard.bin.

(*) This appears to be different from `extsae`, as used by Synaptics, which isn't supported here yet, although both appear as `extsae` in the firmware version. It's necessary to run `strings` on the firmware and grep for `sae_ext` to be sure. And obviously, both are different to the in-firmware WPA3 support provided by the `sae` feature, but which only works with `iwd`.

### DIFF
--- a/drivers/net/wireless/broadcom/brcm80211/brcmfmac/bus.h
+++ b/drivers/net/wireless/broadcom/brcm80211/brcmfmac/bus.h
@@ -298,7 +298,7 @@ void brcmf_rx_event(struct device *dev, struct sk_buff *rxp);
 
 int brcmf_alloc(struct device *dev, struct brcmf_mp_device *settings);
 /* Indication from bus module regarding presence/insertion of dongle. */
-int brcmf_attach(struct device *dev);
+int brcmf_attach(struct device *dev, bool start_bus);
 /* Indication from bus module regarding removal/absence of dongle */
 void brcmf_detach(struct device *dev);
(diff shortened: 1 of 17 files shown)


pelwell:
You can install a beta kernel (6.6) containing this patch by running `sudo rpi-update pulls/5945`. Back up any important data, or clone your SD card, first - just in case...

AdvancedDogPlopThrower:
I needed to copy the firmware to /lib/firmware/cypress/cyfmac43455-sdio-standard.bin (not  /etc/firmware/cypress/)
Then it works using Network Manager

NB Security must be set to "WPA3 Personal" and NOT "WPA/WPA2/WPA3 Personal"
image

PR #5954, opened 2024-02-13
arm: dt: bcm2712: Reduce DDC frequency to 97.5kHz from 200kHz.

6by9:
The I2C spec says the DDC link should run at 100kHz or less, however Pi5/BCM2712 had been configured for 200kHz.
Reduce it to comply with the spec, and match Pi4.


6by9:
https://forums.raspberrypi.com/viewtopic.php?p=2192979

PR #5955, opened 2024-02-13
Set PCIe QOS chicken bits for 2712D0

njhollinghurst:
Fairly conservative small change which just sets the three top bits, which should make D0 behave the same as C1.

Still using high QOS values with almost no spread, and ignoring QOS forwarding which seems still to be broken (while it no longer always gives a QOS of zero, it doesn't prevent glitches when real-time and non-real-time traffic share the link and any traffic has QOS below the ARM cores'; regardless of AXI burst clobbering).

Not currently setting the "emergency QOS OR" field of either chip revision: I'm hoping it won't be necessary.

This is a draft PR pending extended testing; it also needs testing on 2712C1 to verify my assertion that these bits are harmlessly ignored  (if not, we'll have to do different things for the different revisions).

### DIFF
--- a/drivers/pci/controller/pcie-brcmstb.c
+++ b/drivers/pci/controller/pcie-brcmstb.c
@@ -199,6 +199,9 @@
 #define  VDM_PRIORITY_TO_QOS_MAP_MASK			0xf
 
 #define PCIE_MISC_AXI_INTF_CTRL 0x416C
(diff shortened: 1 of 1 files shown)


njhollinghurst:
OK it's had an overnight soak test on 2712D0 and a briefer test on 2712C1 and seems OK.
It's also been confirmed that these chicken bits have no effect on C1. There are no QOS-ORing bits on C1 that would affect PCIe2.

PR #5960, opened 2024-02-15
drivers: staging: bcm2835-isp: Respect caller's stride value

davidplowman:
The stride value reported for output image buffers should be at least as large as any value that was passed in by the caller (subject to correct alignment for the pixel format). If the value is zero (meaning no value was passed), or is too small, the minimum acceptable value will be substituted.


davidplowman:
@naushir @6by9

naushir:
LGTM!

6by9:
I thought we had this anyway, but I guess not.

Looks fine, although it does give userspace an easy way to allocate masses of memory by artificially inflating the stride. Then again VIDIOC_CREATE_BUFS allows you to do that anyway.

vanlayag (approved):

popcornmix:
@davidplowman this needs a rebase before merging.

davidplowman:
Sorry, failed to spot that this wanted a rebase. Hopefully all good now...!

6by9:
Still LGTM

PR #5961, opened 2024-02-15
drm: Add DRM_MODE_TV_MODE_MONOCHROME

njhollinghurst:
Add this as a value for enum_drm_connector_tv_mode, represented by the string "MONOCHROME", to represent video with standard timings but no colour encoding or bursts. Define it to have no pedestal (since only NTSC-M calls for a pedestal).

Change default mode creation to acommodate the new tv_mode value which comprises both 525-line and 625-line formats.


njhollinghurst:
It's a draft for now as I'm not sure if it's something we want to commit ourselves to, or eventually upstream?

Hopefully I have mentioned the new tv_mode in all the places where it's needed?

Carefully worded to imply that it supports both 525-line and 625-line formats but is not necessarily limited to those two.

- It might be vaguely useful  in its own right;
- It also solves my problem of how to hide/advertise "vintage" TV modes in RP1 VEC;
- It does not solve the problem of more generic custom modes in the console FB.

drivers/gpu/drm/drm_modes.c
```diff
@@ -530,7 +530,8 @@ static int fill_analog_mode(struct drm_device *dev,
  * @interlace: whether to compute an interlaced mode
  *
  * This function creates a struct drm_display_mode instance suited for
- * an analog TV output, for one of the usual analog TV mode.
+ * an analog TV output, for one of the usual analog TV modes (where this
```
6by9: Nit: make the additional bit a new sentence.

drivers/gpu/drm/drm_connector.c
```diff
@@ -1049,6 +1049,7 @@ static const struct drm_prop_enum_list drm_tv_mode_enum_list[] = {
 	{ DRM_MODE_TV_MODE_PAL_M, "PAL-M" },
 	{ DRM_MODE_TV_MODE_PAL_N, "PAL-N" },
 	{ DRM_MODE_TV_MODE_SECAM, "SECAM" },
+	{ DRM_MODE_TV_MODE_MONOCHROME, "MONOCHROME" },
```
6by9: Nit: `MONOCHROME` is pretty long. Abbreviate to `MONO`?

I don't recall if the kernel command line has a max length, but it can get pretty unwieldy anyway just in normal usage.

6by9 (commented):
Looks fine to me, with just two very minor nits if we were respinning for any reason.

drivers/gpu/drm/drm_connector.c
njhollinghurst: I guess the enum value should also be shortened to match the string?

njhollinghurst: Actually not sure about changing the enum, as it could get confused with mono audio. MONOCHROME already appears in full in a spattering of symbol names (mostly private to particular VL2 and DRM drivers).

6by9: Enum value within DRM ("MONOCHROME") or name (DRM_MODE_TV_MODE_MONOCHROME) ?

I have no issue with the enum name being that long as it's only ever in the source code. It's only that handling the long string becomes a headache at runtime.

6by9:
LGTM

njhollinghurst:
I'm still a bit wary of merging it. Once the VEC driver (either of them) starts referring to this, they will be impossible to upstream without this change (or without reverting support for it).

njhollinghurst:
A late thought:  Now I think `"Mono"` should be not all upper-case, since it's not an acronym.
That would be more in keeping with the other property names in `drm_connector.c`.
I hope the abbreviation from "Monochrome" will be acceptable.

6by9:
I'll create a patch to implement mono on vc4_vec and try upstreaming it.
I'll query the ambiguity over having resolutions quoted in the descriptions but then omitting PAL60.

6by9:
Sent upstream - https://patchwork.freedesktop.org/series/130023/

njhollinghurst:
This branch and PR are somewhat orphaned now, but for completeness I've pushed an implementation for RP1 VEC.

njhollinghurst:
Superseded by #5962

PR #5962, opened 2024-02-16
VEC mono support

6by9:
Extends #5961 to add support to vc4

@njhollinghurst

### DIFF
--- a/drivers/gpu/drm/drm_connector.c
+++ b/drivers/gpu/drm/drm_connector.c
@@ -1049,6 +1049,7 @@ static const struct drm_prop_enum_list drm_tv_mode_enum_list[] = {
 	{ DRM_MODE_TV_MODE_PAL_M, "PAL-M" },
 	{ DRM_MODE_TV_MODE_PAL_N, "PAL-N" },
 	{ DRM_MODE_TV_MODE_SECAM, "SECAM" },
+	{ DRM_MODE_TV_MODE_MONOCHROME, "Mono" },
 };
 DRM_ENUM_NAME_FN(drm_get_tv_mode_name, drm_tv_mode_enum_list)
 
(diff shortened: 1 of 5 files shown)


6by9:
Sent upstream for comment on monochrome support - https://patchwork.freedesktop.org/series/130023/

njhollinghurst:
If the DRM change turns out not to be upstreamable, perhaps we should abandon this or do it by other means. There doesn't seem to be enough user demand for it to justify maintaining separately (especially when we come to implement it for the RP1 VEC which is new and not yet upstreamed at all).

6by9:
> If the DRM change turns out not to be upstreamable, perhaps we should abandon this or do it by other means. There doesn't seem to be enough user demand for it to justify maintaining separately (especially when we come to implement it for the RP1 VEC which is new and not yet upstreamed at all).

Disabling colour does have its uses, so it would be nice to support.

Selecting it via the kernel command line is the ideal option, with it appearing through DRM UAPI is coincidental.
The core DRM patch is minimal, and is in code that is unlikely to change upstream. It also means we can implement RP1 mono support in the same way, whereas a module parameter will be either vc4.mono_vec or drm-rp1-vec.mono depending on platform.

Unless @popcornmix or @pelwell have strong objections, I'm tempted to say we add the enum value downstream and let mainline do as they wish.

pelwell:
No objection from me.

popcornmix:
I'm okay with this.

6by9:
Ta. I'll rebase this then, and we just ignore mainline.

6by9:
Branch is rebased, but Github PRs are having issues (https://www.githubstatus.com/incidents/wcl1sw4mzg60) and this hasn't updated yet.

PR #5965, opened 2024-02-17
Add support for VEML6075 and HTS221 in i2c-sensor overlay

javiercarrascocruz:
The HTS221 humidity and temperature sensor is only supported inside the
rpi-sense v1 and v2 overlays, but there is still no support as a
standalone sensor. Moreover, its interrupt (data ready) is still not
supported. When at it, irq.h has been added to use interrupt trigger
definitions instead of hard coded numbers.

Support for the VEML6075 UVA and UVB light sensor has been added to the
kernel with v6.8. This device is supported by IIO and does not support
additional parameters like interrupts or different I2C addresses.

### DIFF
--- a/arch/arm/boot/dts/overlays/README
+++ b/arch/arm/boot/dts/overlays/README
@@ -2281,10 +2281,13 @@ Params: addr                    Set the address for the ADT7410, BH1750, BME280,
         hdc100x                 Select the Texas Instruments HDC100x temp sensor
                                 Valid addresses 0x40-0x43, default 0x40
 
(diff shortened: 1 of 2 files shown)


javiercarrascocruz:
It seems that the errors have nothing to do with this PR. Are they known issues in `rpi-6.8.y`?

pelwell:
> Are they known issues in rpi-6.8.y?

It seems to have crept in with 6.8-rc3, and will be fixed before it's released.

pelwell:
I've added a patch enabling the VEML6075 driver in the defconfig files, and a temporary build fix.

PR #5966, opened 2024-02-18
media/v4l2_m2m: In buffered mode run jobs if either port is streaming

pelwell:
In order to get the intended behaviour of the stateful video decoder API where only the OUTPUT queue needs to be enabled and fed buffers in order to get the SOURCE_CHANGED event that configures the CAPTURE queue, we want the device to run should either queue be streaming.

@6by9 This patch didn't merge cleanly in the port from 6.7 to 6.8. I've tried to fix it up, and it does compile, but before merging I wanted to check:
1. if it looks OK to you, and
2. if it is still needed - it seems odd that this old patch hasn't gone upstream when so much else has.


6by9:
Already fixed via https://github.com/raspberrypi/linux/commit/3f9278d33d073d373ebc0b311c1f377d24bc093c - the new mechanism from upstream does the same thing as ours, but just uses a new flag to choose that behaviour.

pelwell:
OK - thanks.

Issue #5964, opened 2024-02-17
Wifi with 802.1x stopped working on rpi-6.6.y branch


iucoen:
### Describe the bug

- After updating to 8c590ad86d69fe72b7051f2311da133ed73e4a3b, wifi stops working. With a kernel warning message in brcmf_cfg80211_set_pmk
- Update: the problematic commit seems to be: 15f000a. After reverting EAP-TLS works again. 

### Steps to reproduce the behaviour

Boot up, error happens as soon as wpa_supplicant starts.

### Device (s)

Raspberry Pi 5

### System

I use gentoo linux. Kernel version 6.6.16-v8-16k+. For WIFI I'm using wpa_supplicant in EAP-TLS mode.

### Logs

dmesg
```
[   44.160554] ------------[ cut here ]------------
[   44.165757] WARNING: CPU: 0 PID: 310 at drivers/net/wireless/broadcom/brcm80211/brcmfmac/cfg80211.c:5987 brcmf_cfg80211_set_pmk+0x12c/0x150 [brcmfmac]
[   44.179912] Modules linked in: joydev vc4 snd_soc_hdmi_codec cec brcmfmac_wcc drm_display_helper hci_uart drm_dma_helper pisp_be drm_kms_helper rpivid_hevc(C) btbcm v4l2_mem2mem videobuf2_dma_contig spidev videobuf2_memops aes_ce_blk snd_soc_core videobuf2_v4l2 snd_compress bluetooth aes_ce_cipher snd_pcm_dmaengine brcmfmac snd_pcm videodev ghash_ce ecdh_generic ecc videobuf2_common brcmutil snd_timer gf128mul libaes snd mc spi_bcm2835 raspberrypi_hwmon sha2_ce cfg80211 raspberrypi_gpiomem rp1_adc sha256_arm64 rfkill sha1_ce nvmem_rmem uio_pdrv_genirq gpio_keys uio pwm_fan sch_fq_codel drm fuse drm_panel_orientation_quirks backlight dm_mod
[   44.239470] CPU: 0 PID: 310 Comm: wpa_supplicant Tainted: G        WC         6.6.16-v8-16k+ #48
[   44.249020] Hardware name: Raspberry Pi 5 Model B Rev 1.0 (DT)
[   44.255617] pstate: 80400009 (Nzcv daif +PAN -UAO -TCO -DIT -SSBS BTYPE=--)
[...]

PR #5969, opened 2024-02-20
brcmfmac: Fix 802.1x

iucoen:
Commit 7d239fbf9d4 broke 802.1X authentication by setting profile->use_fwsup = NONE whenever PSK is not used. However 802.1X does not use PSK and requires profile->use_fwsup set to 1X, or brcmf_cfg80211_set_pmk() fails. Fix this by also checking for if 802.1X is wanted.

Fixes: 7d239fbf9d4 (brcmfmac: Fix interoperating DPP and other encryption network access
Fixes: https://github.com/raspberrypi/linux/issues/5964


pelwell:
This makes sense - good job - but I'd like to change the test from `if (!sme->want_1x)` to `if (profile->use_fwsup != BRCMF_PROFILE_FWSUP_1X)` - it's more like the block of code above, and I think makes it clearer what the value of use_fwsup will be in each case.

iucoen:
I'll squash the commits and create a new PR.

pelwell:
That's fine, but FYI we get the option to squash into a single commit on merge. The alternative would have been to squash locally and "git push -f" to the same github branch.

Issue #5946, opened 2024-02-12
raspberrypi_axi_monitor shows activity on incorrect bus device


cibomahto:
### Describe the bug

When attempting to profile a system using raspberrypi_axi_monitor, bus traffic appears to be reported from the wrong bus.

Specifically, traffic that should have originated from the CPU is being shown as originating from the H264 block.

### Steps to reproduce the behaviour

1. Use Rasberry Pi Imager to prepare an SD card with the 'Rasbperry Pi OS Lite (64-bit)' image. Use the custom settings to enable SSH, and set a user/password
2. Place a CM4 (RPI-CM4Lite-R1-3' in the IO board. Connect the ethernet switch as 12V power.
3. Boot the device
4. Once it is on the network, ssh to it and add 'dtparam=axiperf=on' to /boot/config.txt
5. Reboot the device
6. SSH to the device and run 'sudo apt update; sudo apt install stress-ng' to install stress-ng -y
7. Run 'cat /sys/kernel/debug/raspberrypi_axi_monitor/System/data' to read the performance stats: 

```
     Bus   |    Atrans    Atwait      AMax    Wtrans    Wtwait      WMax    Rtrans    Rtwait      RMax
======================================================================================================
    DMA_L2 |        0K        0K        0K        0K        0K        0K        0K        0K        0K
     TRANS |        0K        0K        0K        0K        0K        0K        0K        0K        0K
      JPEG |        0K        0K        0K        0K        0K        0K        0K        0K        0K
 SYSTEM_UC |        0K        0K        0K        0K        0K        0K        0K        0K        0K
    DMA_UC |        0K        0K        0K        0K        0K        0K        0K        0K        0K
 SYSTEM_L2 |        0K        0K        0K        0K        0K        0K        0K        0K        0K
    CCP2TX |        0K        0K        0K        0K        0K        0K        0K        0K        0K
   MPHI_RX |        0K        0K        0K        0K        0K        0K        0K        0K        0K
   MPHI_TX |        0K        0K        0K        0K        0K        0K        0K        0K        0K
       HVS |       15K        0K        0K        0K        0K        0K       14K        0K        0K
      H264 |        4K        1K        0K        6K        1K        0K        9K        1K        0K
       ISP |        0K        0K        0K        0K        0K        0K        0K        0K        0K
       V3D |        0K        0K        0K        0K        0K        0K        0K        0K        0K
PERIPHERAL |        0K        0K        0K        0K        0K        0K        0K        0K        0K
[...]

PR #5971, opened 2024-02-20
perf/raspberry: Add support for 2712 axi performance monitors

popcornmix:
Also handle 2711 correctly which has a different configuration from 2835.

### DIFF
--- a/drivers/perf/raspberrypi_axi_monitor.c
+++ b/drivers/perf/raspberrypi_axi_monitor.c
@@ -33,7 +33,7 @@
 #define MAX_BUSES 16
 #define DEFAULT_SAMPLE_TIME 100
 
-#define NUM_BUS_WATCHER_RESULTS 9
+#define NUM_BUS_WATCHER_RESULTS 11
 
 struct bus_watcher_data {
 	union	{
@@ -48,6 +48,8 @@ struct bus_watcher_data {
 			u32 rtrans;
 			u32 rtwait;
 			u32 rmax;
+			u32 rpend;
+			u32 ratrans;
 		};
 	};
 };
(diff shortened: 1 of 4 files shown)


popcornmix:
See: https://github.com/raspberrypi/linux/issues/5946
(and it also supports 2712)

drivers/perf/raspberrypi_axi_monitor.c
```diff
@@ -515,17 +699,35 @@ static const struct file_operations fops_debug = {
...
 	struct device *dev = &pdev->dev;
 	struct device_node *np = dev->of_node;
 	struct device_node *fw_node;
+	const struct of_device_id *match;
 
 	state = kzalloc(sizeof(struct rpi_axiperf), GFP_KERNEL);
 	if (!state)
 		return -ENOMEM;
 
+	match = of_match_device(rpi_axiperf_match, dev);
```
6by9: You can use `match = of_device_get_match_data(dev);` to save having to move `rpi_axiperf_match`.

pelwell:
Somewhere (bcm2712-rpi.dtsi, perhaps) needs an axiperf override, just as bcm270x-rpi.dtsi has.
And it needs to be two commits...

drivers/perf/raspberrypi_axi_monitor.c
```diff
@@ -521,11 +706,18 @@ static int rpi_axiperf_probe(struct platform_device *pdev)
...
 
 	state = kzalloc(sizeof(struct rpi_axiperf), GFP_KERNEL);
 	if (!state)
 		return -ENOMEM;
 
+	match = of_device_get_match_data(dev);
+	if (!match)
+		return -ENODEV;
+
+	state->config = match->data;
```
6by9: `of_device_get_match_data` already gives you the data pointer, so you want to do
```
state->config = of_device_get_match_data(dev);
if (!state->config)
   return -ENODEV;
```
(possibly -EINVAL rather than -ENODEV?)

popcornmix: yeah - I noticed that locally when change stopped working

pelwell:
Other than being a combined code & DT commit, LGTM.

pelwell (approved):

PR #5977, opened 2024-02-21
iommu: bcm2712-iommu: Map and unmap multiple pages in a single call

njhollinghurst:
For efficiency, the map_pages() and unmap_pages() calls now pay attention to their "count" argument, at least in successful cases.

Fix a bugette where we omitted to set mapped to 0 in case of error. Minor style fix.

### DIFF
--- a/drivers/iommu/bcm2712-iommu.c
+++ b/drivers/iommu/bcm2712-iommu.c
@@ -241,42 +241,48 @@ static int bcm2712_iommu_map(struct iommu_domain *domain, unsigned long iova,
 			     int prot, gfp_t gfp, size_t *mapped)
 {
 	struct bcm2712_iommu *mmu = domain_to_mmu(domain);
+	u32 entry = MMMU_PTE_VALID | (pa >> MMU_PAGE_SHIFT);
+	u32 align = (u32)(iova | pa | bytes);
+	unsigned int p;
 
-	(void)gfp;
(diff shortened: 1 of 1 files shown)


njhollinghurst:
This is a leftover from #5833. The efficiency improvement is admittedly quite modest; the average `count` (when exercised by camera) seems to be about 2.

Note that `count` does not feature in the large-page/alignment flags: because pages might get freed in a different order to the order in which they were allocated. We aren't bothering to check e.g. that pages are aligned to their own size, because we trust the framework to get it right.

There is no change in the "pass-thru" default case (which perhaps should be an error case, as it's never used), which still returns one page at a time. The bug fix in the "off-the-top" error case is needed, I think.

6by9:
Looks reasonable based on my knowledge of it (ie very limited).
In theory a patch should do one thing, but this will all get squashed before upstreaming, so I don't care.

6by9 (approved):

njhollinghurst:
Tested with camera/ISP; I'll do a bit more testing with other Pi 5 blocks then de-draft it.

njhollinghurst:
I'm going to push a tidied-up version that ignores special treatment of the "pass-through" region.
IOMMU-aware DMA drivers aren't told it exists and should never use it -- they should respect the `force_aperture` flag.
Meanwhile non-IOMMU-aware ones won't use this API.

PR #5982, opened 2024-02-23
dts: bcm2712, rpi1: add definitions for multi channel I2S module

j-schambacher:
Defines the pin functions for use in clk producer mode for I2S0 module using GPIOs 18..27 to enable 4 data lanes input and 4 lanes output.
The pin definitions enable all 4 stereo channels for input and output. In the DT overlay just use <&i2s_clk_producer_mc8> instead of <&i2s_clk_producer>.
The multi channel capability is necessary for our new DAC8X which I'll submit in a separate request.


j-schambacher:
mmmh... forgotten to switch branches .... wanted to split the I2S multi channel stuff into a separate PR

PR #5983, opened 2024-02-23
ASoC: Add multichannel HiFiBerry DAC8X

j-schambacher:
Combines all necessary changes:
Defines the settings for the 8 channel version of the standard DAC by overwriting the number of channels in the DAI defs.
It will run in 8ch mode only on a PI5 as it requires the designware I2S0 module.

Necessary changes are the 'mc8' definitions in the RP1 and bcm2712 DT setting.

Seems, one can't pass the checks with separate PRs.
@pelwell Please let me know in case I need to split it though. Or any other thought on the changes. Thanks, Joerg

### DIFF
--- a/arch/arm/boot/dts/bcm2712-rpi.dtsi
+++ b/arch/arm/boot/dts/bcm2712-rpi.dtsi
@@ -168,6 +168,7 @@ i2c5: &rp1_i2c5 { };
 i2c6: &rp1_i2c6 { };
 i2s:  &rp1_i2s0 { };
 i2s_clk_producer: &rp1_i2s0 { };
+i2s_clk_producer_mc8: &rp1_i2s0 { };
 i2s_clk_consumer: &rp1_i2s1 { };
 pwm0: &rp1_pwm0 { };
 pwm1: &rp1_pwm1 { };
@@ -227,6 +228,12 @@ i2c_arm: &i2c1 {
 	pinctrl-names = "default";
 	pinctrl-0 = <&rp1_i2s0_18_21>;
 };
+
+&i2s_clk_producer_mc8 {
+	pinctrl-names = "default";
+	pinctrl-0 = <&rp1_i2s0_18_27>;
+};
+
 
 &i2s_clk_consumer {
 	pinctrl-names = "default";
--- a/arch/arm/boot/dts/overlays/Makefile
+++ b/arch/arm/boot/dts/overlays/Makefile
@@ -88,6 +88,7 @@ dtbo-$(CONFIG_ARCH_BCM2835) += \
 	hifiberry-amp3.dtbo \
 	hifiberry-amp4pro.dtbo \
 	hifiberry-dac.dtbo \
+	hifiberry-dac8x.dtbo \
 	hifiberry-dacplus.dtbo \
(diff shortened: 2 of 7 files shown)


j-schambacher:
ok, seems, w/o the main DT change in the tree, the other checks won't pass...

pelwell:
What do you think we gain by adding `i2s_clk_producer_mc8`?

pelwell:
Note that by making `i2s_clk_producer_mc8` an alias for `i2s_clk_producer`, and then getting `i2s_clk_producer_mc8` to claim 10 GPIOs, you are forcing all users of `i2s_clk_producer` to assign those 10 GPIOs to I2S.

j-schambacher:
The intention for the clk_producer_mc8 definition is to claim indeed 10 GPIOs - all are dedicated then for I2S input and output as per the RP1 manual, table 3. *_mc8 takes in addition to the 'standard' clk_producer definition I2S0_DI[1 .. 3] and I2S0_DO[1 .. 3]:
I2S0_SCLK, I2S_WS, 4 x I2S0_DI[], 4x I2S0_D[] -> GPIOs 18...27
In fact for a DAC only one could maybe reduce to just the I2S0 _DO[]s, but my thought was to keep things simple. Otherwise, we may end up with all kinds of different channel combos.
I don't see the conflict in case another overlay uses the normal clk_producer def. It loads only the standard 4: I2S0_SCLK, I2S_WS, I2S0_DI[0], I2S0_D[0] -> GPIOs 18...21. Just x-checked it again running 'blink' on one of the unused I2S0_Dx pins while having the DAC board runnning using the standard dac overlay ('hfifberry-dac') with 'clk_producer'. Maybe, I've missed something?

In case you want to try yourself, you do not even need a board. You can easily see the different channels' outputs at pins by running something like

AUDIODEV=hw:0,0 play -r 48000 -n -b 32 -c 8 synth sine 1000 sine 2000 sine 3000 sine 4000 sine 5000 sine 6000 sine 7000 sine 8000 vol -10dB

Using _mc8 will give you data on all 4 I2SD_DO[]s, standard will output only on I2S0_DO[0].

pelwell:
This is the conflict: https://github.com/raspberrypi/linux/pull/5983/files#diff-474cc6a46df20f7fbaec6f0d0f68c9c0eb42ccb8d2563822083b82112de1647fR170-R171

You are making i2s_clk_producer_mc8 an alias for i2s_clk_producer. This means that anything you do to i2s_clk_producer_mc8, you are by definition doing to i2s_clk_producer. You could rewrite this:
```
&i2s_clk_producer_mc8 {
	pinctrl-names = "default";
	pinctrl-0 = <&rp1_i2s0_18_27>;
};
```
as this:
```
&i2s_clk_producer {
	pinctrl-names = "default";
	pinctrl-0 = <&rp1_i2s0_18_27>;
};
```
and it would make absolutely zero difference to the output.

j-schambacher:
Ok, I see. The (bad) idea was to define just once the complete settings for all sound cards potentially using 8 channels.

Meanwhile, I've moved all settings/definitions into the DAC8X overlay. This means I leave RP1 and bcm2712 DT files untouched.
I guess it is easier to close this PR and open a new one with only the two commits.

Issue #5986, opened 2024-02-25
Seems like waveshare 2.8 inch Panel has wrong touchinput scale in bookworm


TheMarrt:
### Describe the bug

This panel:
https://www.waveshare.com/wiki/2.8inch_DSI_LCD#FAQ

can be activated with this dtoverlay:
dtoverlay=vc4-kms-dsi-waveshare-panel,2_8_inch,rotation=90

But the Scale seems to be flipped. Horizontal(480) scale is overexaggerated like it scales to 640px, and vertical(640) to 480).
I suspect this is because the Display is 480x640 rather than 640x480

I suspect this must be fixed, but i am new to github issue reporting:
https://github.com/raspberrypi/linux/blob/afd5f659b0453e4c710ce5cf74c577563ff16239/arch/arm/boot/dts/overlays/vc4-kms-dsi-waveshare-panel-overlay.dts#L80
"touchscreen-size-x:0=640" --> 480
"touchscreen-size-y:0=480" --> 640

### Steps to reproduce the behaviour

try https://www.waveshare.com/wiki/2.8inch_DSI_LCD on raspberry pi 4 b with bookworm installed

### Device (s)

Raspberry Pi 4 Mod. B

### System

cat /etc/rpi-issue
Raspberry Pi reference 2023-12-05
Generated using pi-gen, https://github.com/RPi-Distro/pi-gen, 70cd6f2a1e34d07f5cba7047aea5b92457372e05, stage4

vcgencmd version
Oct 17 2023 15:39:16
Copyright (c) 2012 Broadcom
version 30f0c5e4d076da3ab4f341d88e7d505760b93ad7 (clean) (release) (start)

uname -a
Linux pi8 6.1.0-rpi8-rpi-v8 #1 SMP PREEMPT Debian 1:6.1.73-1+rpt1 (2024-01-25) aarch64 GNU/Linux

6by9:
The wiki you linked to doesn't reference the rotation override at all.

Why do you need to rotate the video from device tree? If using any window manager (ie X11 or Wayfire) then both have easy mechanisms for rotating the displays as part of the composition (`xrandr` and `wlr-randr`), and similar options for mapping touch to match the display (`xinput --map-to-output` and adding an `input-device:` section to wayfire.ini).

I'm not overly bothered about fixing the touchscreen for the basic framebuffer device with touch. My preference would be dropping the rotation override.

TheMarrt:
[...]

PR #5990, opened 2024-02-26
dtoverlays: Correct width/height on Waveshare 2.8" panel

6by9:
Width and height were swapped, so touch didn't work over the whole panel.


popcornmix (approved):

popcornmix:
See: https://github.com/raspberrypi/linux/issues/5986

6by9:
> See: #5986

I was just creating the link the other way :-)

PR #5991, opened 2024-02-27
drivers: media: cfe: Add basic async virtual channel support

naushir:
Add support for asynchronous virtual channel handling in the cfe driver. This functionality is not yet enabled, it needs the streams API support to set a virtual channel for each device node.

The existing job handling and ISR logic works exactly the same as now, but grouped with the same virtual channel on other device nodes.


naushir:
Draft for now as it still needs much more testing.

PR #5993, opened 2024-02-28
ASoC: Add multichannel HiFiBerry DAC8X [V2]

j-schambacher:
Additions and changes for the 8 channel version of the standard PCM5102 DAC by setting the number of channels
in the DAI definitions.
It will run in 8ch mode only on a PI5 as it requires the designware I2S0 module.

The necessary output lanes I2S0_DO[0..3] are claimed from within the DT overlay.


pelwell:
That looks good. Can you just add this to overlay_map.dts?:
```diff

         disable-wifi-pi5 {
                 bcm2712;
         };
 
+        hifiberry-dac8x {
+                bcm2712;
+        };
+
         highperi {
                 bcm2711;
         };
```

j-schambacher:
Do I need to change the dac8x overlay to be only compatible with bcm2712 to avoid the failing check?

pelwell:
Yes - use "brcm,bcm2712".

PR #6008, opened 2024-03-04
vc4: Add jack detection to HDMI audio driver

cillian64:
Add ALSA jack detection to the vc4-hdmi audio driver so userspace knows when to add/remove HDMI audio devices.

### DIFF
--- a/drivers/gpu/drm/vc4/vc4_hdmi.c
+++ b/drivers/gpu/drm/vc4/vc4_hdmi.c
@@ -50,6 +50,7 @@
 #include <linux/reset.h>
 #include <sound/dmaengine_pcm.h>
 #include <sound/hdmi-codec.h>
+#include <sound/jack.h>
 #include <sound/pcm_drm_eld.h>
 #include <sound/pcm_params.h>
 #include <sound/soc.h>
@@ -421,7 +422,7 @@ static void vc4_hdmi_handle_hotplug(struct vc4_hdmi *vc4_hdmi,
 				    enum drm_connector_status status)
 {
 	struct drm_connector *connector = &vc4_hdmi->connector;
-	struct edid *edid;
+	struct edid *edid = NULL;
 	int ret;
 
 	/*
@@ -439,12 +440,25 @@ static void vc4_hdmi_handle_hotplug(struct vc4_hdmi *vc4_hdmi,
 	 * the lock for now.
 	 */
 
+	if (status != connector_status_disconnected)
+		edid = drm_get_edid(connector, vc4_hdmi->ddc);
+
+	/*
+	 * Report plugged/unplugged events to ALSA jack detection.  Do this
+	 * *after* EDID probing, otherwise userspace might try to bring up
+	 * audio before it's ready.
+	 */
+	mutex_lock(&vc4_hdmi->update_plugged_status_lock);
+	if (vc4_hdmi->plugged_cb && vc4_hdmi->codec_dev)
+		vc4_hdmi->plugged_cb(vc4_hdmi->codec_dev,
+				     status != connector_status_disconnected);
+	mutex_unlock(&vc4_hdmi->update_plugged_status_lock);
+
 	if (status == connector_status_disconnected) {
 		cec_phys_addr_invalidate(vc4_hdmi->cec_adap);
 		return;
 	}
 
-	edid = drm_get_edid(connector, vc4_hdmi->ddc);
 	if (!edid)
 		return;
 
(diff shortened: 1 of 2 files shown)


drivers/gpu/drm/vc4/vc4_hdmi.c
```diff
@@ -3694,6 +3748,8 @@ static int vc4_hdmi_bind(struct device *dev, struct device *master, void *data)
 	if (ret)
 		return ret;
 
+	mutex_init(&vc4_hdmi->update_plugged_status_lock);
```
6by9, line 3744: We're missing a matching `mutex_destroy()` in unbind. Admittedly it only does anything if `CONFIG_DEBUG_MUTEXES` is set, but it's nicer to keep it clean.

drivers/gpu/drm/vc4/vc4_hdmi.c
```diff
@@ -448,6 +458,17 @@ static void vc4_hdmi_handle_hotplug(struct vc4_hdmi *vc4_hdmi,
 	if (!edid)
 		return;
 
+	/*
```
6by9: Do we need to separate the plugged and unplugged events either side of the EDID parsing, or can we just always report them after (ie from here)?

6by9 (commented):
Two minor comments, otherwise looks good.

drivers/gpu/drm/vc4/vc4_hdmi.c
cillian64: If we only report here then we'd only report the plugged events (not the unplugged events), because we early-return if the state is disconnected.

The duplication really bothers me but I couldn't come up with a better way to structure it.

drivers/gpu/drm/vc4/vc4_hdmi.c
cillian64: Good point, I'll add that.

drivers/gpu/drm/vc4/vc4_hdmi.c
6by9: I'd missed the return.

Possibly just split the edid fragment?
```
	edid = drm_get_edid(connector, vc4_hdmi->ddc);

	/* We need to process the EDID before calling the plugged callback */
	if (status == connector_status_connected) {
		mutex_lock(&vc4_hdmi->update_plugged_status_lock);
		if (vc4_hdmi->plugged_cb && vc4_hdmi->codec_dev)
			vc4_hdmi->plugged_cb(vc4_hdmi->codec_dev, CONDITIONAL);
		mutex_unlock(&vc4_hdmi->update_plugged_status_lock);
	}

	if (!edid)
		return;
```

cillian64: I've done essentially this but also put an `if(connected)` before fetching the EDID.  I've convinced myself that we can't uninitialised-read `edid` (though it's less obvious than I would like)

drivers/gpu/drm/vc4/vc4_hdmi.c
```diff
@@ -3806,8 +3855,19 @@ static int vc4_hdmi_bind(struct device *dev, struct device *master, void *data)
 	return ret;
 }
 
+static void vc4_hdmi_unbind(struct device *dev, struct device *master, void *data)
+{
+	struct vc4_hdmi *vc4_hdmi = dev_get_drvdata(dev);
+
+	(void)master;
+	(void)data;
```
6by9: Does any compiler actually complain over the unused parameters? 
Most of the unbind functions in vc4 don't use master or data, and yet don't worry about having to use all parameters.

drivers/gpu/drm/vc4/vc4_hdmi.c
6by9: You are correct that reading the EDID when disconnected would be an odd thing to do, so I'm happy with that.

pelwell: Is that going to be enough to satisfy all compilers? Wouldn't initialising edid to NULL be better?

cillian64: I've changed to initialise to NULL just in case.

drivers/gpu/drm/vc4/vc4_hdmi.c
cillian64: Good point, looks like linux disables unused warnings: https://www.spinics.net/lists/newbies/msg63816.html

popcornmix:
I've given this a quick test. It creates a new control `HDMI Jack`.
```
pi@pi4:~ $ amixer -d  -D sysdefault:CARD=vc4hdmi0 controls
numid=1,iface=CARD,name='HDMI Jack'
numid=6,iface=MIXER,name='PCM Playback Volume'
numid=5,iface=PCM,name='ELD'
numid=4,iface=PCM,name='IEC958 Playback Default'
numid=3,iface=PCM,name='IEC958 Playback Mask'
numid=2,iface=PCM,name='Playback Channel Map'
```
and the value of this changes with hotplug. e.g. when connected:
```
i@pi4:~ $ amixer -d  -D sysdefault:CARD=vc4hdmi0 cget numid=1
VERIFY ID: numid=1,iface=MIXER,name=''
numid=1,iface=CARD,name='HDMI Jack'
  ; type=BOOLEAN,access=r-------,values=1
  : values=on
```
and disconnected
```
pi@pi4:~ $ amixer -d  -D sysdefault:CARD=vc4hdmi0 cget numid=1
VERIFY ID: numid=1,iface=MIXER,name=''
numid=1,iface=CARD,name='HDMI Jack'
  ; type=BOOLEAN,access=r-------,values=1
  : values=off
```
(works for both HDMI ports). Assuming this is what it's meant to do then it works for me.

cillian64:
> Assuming this is what it's meant to do then it works for me.

Pretty much, the other things needed are a patch to pipewire to recognise "HDMI Jack" instead of "HDMI/DP Jack" (which upstream have merged but I need to backport into our package), and a patch to wf-panel-pi to stop restarting pipewire on display hotplugs.

PR #6010, opened 2024-03-04
media: i2c: imx296: Updated register setting to fix Fast Trigger

njhollinghurst:
In Fast Trigger mode (external shutter control), FE packet was not sent at end of frame. Sony recommend this specific change, which fixes the issue.

The Sony registers also differ from the open source driver in a few other places (not reflected in this initial commit); pending further testing, I may add those later...


6by9 (approved):
That matches the guidance received from Sony, so LGTM.

njhollinghurst:
OK. I'll evaluate the other register differences separately so as not to complicate this fix.

PR #6013, opened 2024-03-05
Add GPCLK support on RP1

P33M:
Prompted by https://forums.raspberrypi.com/viewtopic.php?t=365761&hilit=gpclk this lets users drive gpclks via overlays (and if suitably motivated, drive off-chip sources into peripherals).

I've noticed that clocks without the enable bit set (either by firmware or Linux) report a nonzero frequency - probably a bug, but I don't know where...

### DIFF
--- a/drivers/clk/clk-rp1.c
+++ b/drivers/clk/clk-rp1.c
@@ -361,6 +361,7 @@ struct rp1_clock_data {
 	u32 div_frac_reg;
 	u32 sel_reg;
 	u32 div_int_max;
+	unsigned long max_freq;
 	u32 fc0_src;
 };
 
@@ -1211,7 +1212,15 @@ static void rp1_clock_choose_div_and_prate(struct clk_hw *hw,
 	/* Recalculate to account for rounding errors */
 	tmp = (u64)*prate << CLK_DIV_FRAC_BITS;
 	tmp = div_u64(tmp, div);
-	*calc_rate = tmp;
+	/*
+	 * Prevent overclocks - if all parent choices result in
+	 * a downstream clock in excess of the maximum, then the
+	 * call to set the clock will fail.
+	 */
+	if (tmp > clock->data->max_freq)
+		*calc_rate = 0;
+	else
+		*calc_rate = tmp;
 }
 
 static int rp1_clock_determine_rate(struct clk_hw *hw,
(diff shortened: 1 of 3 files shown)


drivers/clk/clk-rp1.c
```diff
@@ -1745,9 +1818,15 @@ static const struct rp1_clk_desc clk_desc_array[] = {
 				.name = "clk_i2s",
 				.parents = {"xosc",
 					    "pll_audio",
-					    "pll_audio_sec"},
+					    "pll_audio_sec",
+					     "clksrc_gp0",
```
pelwell: The indentation is off here.

drivers/clk/clk-rp1.c
```diff
@@ -1830,9 +1923,25 @@ static const struct rp1_clk_desc clk_desc_array[] = {
...
-				.parents = {"xosc"},
+				.parents = {"xosc",
+					    "clksrc_gp1",
+					    "clksrc_gp2",
+					    "clksrc_gp3",
+					    "clksrc_gp4",
+					    "clksrc_gp5",
+					    "pll_sys",
+					    "pll_audio",
+					    "",
```
pelwell, line 1958: I think the  "-", used elsewhere, looks nicer - could it work instead?

include/dt-bindings/clock/rp1.h
```diff
@@ -50,3 +50,7 @@
 #define RP1_CLK_DPI			40
 #define RP1_CLK_MIPI0_DPI		41
 #define RP1_CLK_MIPI1_DPI		42
+
+/* Extra PLL output channels - RP1B0 only*/
```
pelwell: Missing space.

drivers/clk/clk-rp1.c
P33M: The "-" is used as a break/delimiter between the glitchless mux sources and the aux mux sources - the choice of an empty string was deliberate to make the source selection invalid. I'm not sure why the delimiter is necessary - the driver has the information from the num_std_parents and num_aux_parents to work with.

P33M: There's a gotcha that means a delimiter is needed - the mux assignments are not sequential - the second index into the standard sources is always the aux selector (if a glitchless mux is implemented).

oOTnTOo:
So... How can i use GPCLK0? I use "pinctrl set 4 a0"，but its just "lo" or "hi"。i am try DTS but not work. my system is "Linux raspberrypi 6.6.51+rpt-rpi-2712 #1 SMP PREEMPT Debian 1:6.6.51-1+rpt3 (2024-10-08) aarch64 GNU/Linux", model is Raspberry Pi 5 , bcm 2712。

P33M:
Commenting on a closed pull request isn't recommended.

Ask your question in the support forum at: https://forums.raspberrypi.com/viewforum.php?f=29

PR #6025, opened 2024-03-11
defconfigs: Add ILI9881 to rpi and bcm2709 defconfigs

6by9:
Needs validating that we've got everything, hence draft.


6by9:
Forgot I'd set this as draft. All good as far as I can tell.

PR #6027, opened 2024-03-12
Imx296 updates

naushir:
These commits went missing during a recent kernel rebase.


naushir:
@popcornmix or @pelwell can we merge this if you are ok with it?

PR #6029, opened 2024-03-12
dtoverlays: Fixup pendown gpio polarity for ads7846 users

6by9:
The driver has been converted to use gpiod, which will normalise polarity based on DT.

The piscreen overlay (and others) incorrectly defines the pendown GPIO as being ACTIVE_HIGH (0), althought triggering on the high-low edge for pen down. It therefore tries reading the pen position when not being touched, and stops when it is touched.

Tested with piscreen and ads7846 overlays. Also fixed on others where the interrupt says high->low but the polarity was ACTIVE_HIGH.

https://github.com/raspberrypi/bookworm-feedback/issues/88


6by9:
(Pushed to the wrong branch. Now corrected again and rebased in the process).

PR #6038, opened 2024-03-14
    Add support for generic i2c-mux base-nr property

pelwell:
It can be desirable to have a fixed, known base address for the channels of a mux. However, there is no easy way to do that from a DT overlay (aliases ought to work, but the overlay application logic can't perform the required string pasting to create "i2c<n>", and creating one alias per channel is tedious). To that end, create a `base-nr` DT property and a `base` parameter that allows the bus number for channel 0 to be specified, with the remaining channels allocated sequentially after it.

### DIFF
--- a/arch/arm/boot/dts/overlays/README
+++ b/arch/arm/boot/dts/overlays/README
@@ -2026,6 +2026,9 @@ Params: pca9542                 Select the NXP PCA9542 device
 
         addr                    Change I2C address of the device (default 0x70)
 
+        base                    Set an explicit base value for the channel bus
+                                numbers
+
         i2c0                    Choose the I2C0 bus on GPIOs 0&1
 
         i2c_csi_dsi             Choose the I2C0 bus on GPIOs 44&45
--- a/arch/arm/boot/dts/overlays/i2c-mux-overlay.dts
+++ b/arch/arm/boot/dts/overlays/i2c-mux-overlay.dts
@@ -159,6 +159,10 @@
 			<&pca9545>,"reg:0",
 			<&pca9548>,"reg:0";
 
+		base =  <&pca9542>,"base-nr:0",
+			<&pca9545>,"base-nr:0",
+			<&pca9548>,"base-nr:0";
+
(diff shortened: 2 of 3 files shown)


pelwell:
This has been seen to work: https://forums.raspberrypi.com/viewtopic.php?p=2203123#p2203123

popcornmix (approved):
You have a merge commit, but otherwise I'm okay with it.

6by9:
Note comments on https://github.com/raspberrypi/linux/commit/a7c1456eb3e8f99ed908b2ccce9a4045c2d3c85e over unintended consequences of this for i2c0mux failing.
(I think it's actually due to 841bb8e98b3383fdb98b43251923b94e6c86f1cf at fault)

Changing force_nr should be conditional on `of_property_read_u32` returning a value, otherwise with no force_nr we get the channels initialised with 0, 1, 2, etc, not all with 0.

pelwell:
I've been staring at that line, and not noticing that I had an unsaved change in my editor:
```diff
 		of_property_read_u32(dev_node, "base-nr", &base_nr);
-		if (!force_nr)
+		if (!force_nr && base_nr)
 			force_nr = base_nr + chan_id;
```
Now it makes sense.

6by9:
:-)

I was about to go for 
```
@@ -358,8 +358,7 @@ int i2c_mux_add_adapter(struct i2c_mux_core *muxc,
                u32 base_nr = 0;
                u32 reg;
 
-               of_property_read_u32(dev_node, "base-nr", &base_nr);
-               if (!force_nr)
+               if (!force_nr && !of_property_read_u32(dev_node, "base-nr", &base_nr))
                        force_nr = base_nr + chan_id;
 
                if (muxc->arbitrator)
```

Technically you could have base-nr defined as 0 if the mux was the only thing in your system, and the parent had been assigned somewhere that it wasn't going to clash. The mux child buses would then be 0, 1, 2, 3, etc, with the parent as eg 10.

pelwell:
However, other parts of the bus numbering system reject 0 as a forced value, so it's academic.

PR #6046, opened 2024-03-15
drivers: mmc: add SD support for Command Queueing

P33M:
Application class A2 cards require CQ to be enabled to realise their stated performance figures. Add support to enable/disable card CQ via the Performance Enhancement extension register, and cater for the slight differences in command set versus eMMC.

--

Tested using a 64GB Sandisk Extreme U3 class A2 card. Typical performance uplift is ~2.5x for random 4k read, and 4x for random 4k write. There are sporadic instances of errors (usually command timeouts) that result in the queue being flushed, but I've not been able to identify a pattern/set of commands that cause it.

Please don't merge immediately. I want a testing thread on the forums to flush out anything that's going to be fatal to your root filesystem first...

### DIFF
--- a/arch/arm/boot/dts/broadcom/bcm2712-rpi-5-b.dts
+++ b/arch/arm/boot/dts/broadcom/bcm2712-rpi-5-b.dts
@@ -819,6 +819,7 @@ spi10_cs_pins: &spi10_cs_gpio1 {};
 		random = <&random>, "status";
 		rtc = <&rpi_rtc>, "status";
 		rtc_bbat_vchg = <&rpi_rtc>, "trickle-charge-microvolt:0";
+		sd_cqe = <&sdio1>, "supports-cqe?";
 		spi = <&spi0>, "status";
 		suspend = <&pwr_key>, "linux,code:0=205";
 		uart0 = <&uart0>, "status";
--- a/arch/arm/boot/dts/broadcom/bcm2712-rpi-cm5.dtsi
+++ b/arch/arm/boot/dts/broadcom/bcm2712-rpi-cm5.dtsi
@@ -355,6 +355,7 @@ dpi_16bit_gpio2:        &rp1_dpi_16bit_gpio2        { };
 	mmc-hs200-1_8v;
 	mmc-hs400-1_8v;
 	broken-cd;
+	supports-cqe;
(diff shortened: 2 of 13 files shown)


popcornmix:
You didn't mention it, but I assume this is Pi5 only?
Will the command queueing be disabled gracefully on non-supporting pi models?

P33M:
Yes, for us this is Pi 5 only. There should be no difference in behaviour for a) non-A2 cards and/or b) host controllers that choose to use HSQ.

P33M:
One thing I have noticed it the activity LED on Pi 5 no longer blinks in a sensible fashion - only when IO starts/stops overall.

P33M:
Fixed the random timeouts - they were actually queue flushes that ended up doing an eMMC ext_csd write instead of an SD expansion register write that resulted in an "illegal command" response.

As CQ now disabled by default on Pi 5 it's safe to go in rpi-update.

popcornmix:
To confirm this is disabled by default and adding:
```
dtparam=sd_cqe
```
to config.txt on a pi5 will enable the feature (if you have a supported card).

arch/arm/boot/dts/broadcom/bcm2712-rpi-5-b.dts
```diff
@@ -819,6 +819,7 @@ spi10_cs_pins: &spi10_cs_gpio1 {};
 		random = <&random>, "status";
 		rtc = <&rpi_rtc>, "status";
 		rtc_bbat_vchg = <&rpi_rtc>, "trickle-charge-microvolt:0";
+		sd_cqe = <&sdio1>, "supports-cqe";
```
pelwell: If the aim is to create a boolean (empty property when true, absent when empty) then `"supports-cqe?"` is the correct syntax.

P33M:
Yes. If the card supports the feature and it's enabled you get the line `mmc0: Command Queue Engine enabled` in dmesg.

arch/arm/boot/dts/broadcom/bcm2712-rpi-5-b.dts
P33M: Yes, the property only has to exist for it to have an effect

PR #6064, opened 2024-03-24
(allo-boss-dac) Mute output when changing parameters

alessandromrc:
Since I noticed that sometimes changing sample rates causes some digital quirks and noises, I've changed the function to mute the output before performing the changes and then unmute it when an error occurs or the parameters got set.

Already deeply tested on my Raspberry Pi 4b and the DAC.


sound/soc/bcm/allo-boss-dac.c
```diff
@@ -275,23 +275,40 @@ static int snd_allo_boss_hw_params(
...
-			params_rate(params));
+	if (snd_soc_allo_boss_master) {
+		snd_allo_boss_set_sclk(component, params_rate(params));
 
-		ret = snd_allo_boss_update_rate_den(
-			substream, params);
-		if (ret)
+		ret = snd_allo_boss_update_rate_den(substream, params);
+		if (ret) {
+			// Unmute if error occurs
```
pelwell: It's perfectly acceptable to use `goto`s for error handling. It would work well in this function to put a label (`error:`, perhaps) before the unmute at the end. Then, all of the:
```
	if (ret) {
		// Unmute if error occurs
		snd_allo_boss_gpio_unmute(card);
		return ret;
	}
```
can become:
```
	if (ret)
		goto error;
```

alessandromrc: Nice to know @pelwell, I thought gotos weren't liked over here : D

popcornmix: See: https://www.kernel.org/doc/html/v6.6/process/coding-style.html#centralized-exiting-of-functions

alessandromrc: I've improved commenting (following what @popcornmix sent) and added gotos like @pelwell mentioned.

PS. Also installed the kernel with this version on the raspberry pi 4 and it is fully functional.

Issue #6069, opened 2024-03-26
PCF2131 RTC support + low battery detection


giampiero7:
### Describe the bug

**Support**

Support of PCF2131 has been added to the rtc-pcf2127 driver:
https://github.com/raspberrypi/linux/blob/rpi-6.6.y/drivers/rtc/rtc-pcf2127.c

But the overlay does not include it:
https://github.com/raspberrypi/linux/blob/rpi-6.6.y/arch/arm/boot/dts/overlays/i2c-rtc-common.dtsi

**Low battery detection**

Reading RTC_VL_BACKUP_LOW via ioctl() seems not to return the correct value when battery not installed.
Tested with a custom driver 

I'll be submitting a pull request as soon as possible, but if anyone else is willing to get started I won't be offended :)

### Steps to reproduce the behaviour

**Support**

Add in config.txt:
```
dtoverlay=i2c-rtc,pcf2131
```
Not working

**Low battery detection**

Run this test code:

https://github.com/sfera-labs/rtc-pcf2131/blob/master/pcf2131_example.c

```
gcc -s -Wall pcf2131_example.c -o pcf2131_example
sudo ./pcf2131_example
```

When the battery is not installed the output should be:

```
Date/time: 2023-1-3 14:04:42
Voltage low flags: 0x2
> Backup voltage is low
```

### Device (s)

Raspberry Pi CM4

### System

```
Raspberry Pi reference 2024-03-15
Generated using pi-gen, https://github.com/RPi-Distro/pi-gen, f19ee211ddafcae300827f953d143de92a5c6624, stage2
```

```
Feb 29 2024 12:24:53 
Copyright (c) 2012 Broadcom
version f4e2138c2adc8f3a92a3a65939e458f11d7298ba (clean) (release) (start)
```

```
Linux pi 6.6.20+rpt-rpi-v8 #1 SMP PREEMPT Debian 1:6.6.20-1+rpt1 (2024-03-07) aarch64 GNU/Linux
```

pelwell:
Resolved through #6070.

PR #6070, opened 2024-03-26
RTC PCF2131 support

giampiero7:
Addressing #6069


pelwell:
The overlay change is easy to accept, but the driver patch is more tricky. There's nothing wrong with the content (apart from the checkpatch complaints), but less so the driver patch.

rtc-pcf2127.c is unmodified from the upstream kernel, and the changes are in no way Raspberry Pi-specific. Accepting the patch would mean adding to the hundreds of patches we already host and have to maintain ourselves when, for example, an upstream API changes.

I'm sure the upstream maintainers would accept it (and #6072), at which point we'll happily back-port it into rpi-6.6.y. Until then, if you isolate the overlay patch we can merge it here (or I can do so myself offline).

giampiero7:
Agreed, I reverted the driver changes and I'll try to go through the upstream for the rest.
Thanks

pelwell:
Thanks for understanding.

Issue #6057, opened 2024-03-21
i2c failing to initialize and detect devices


EdgarSouSousa:
### Describe the bug

The i2c bus fails to detect and communicate with connected i2c devices. Dmesg output:
[    1.878285] i2c_dev: i2c /dev entries driver
[    2.368416] brcmstb-i2c 107d508200.i2c:  @97500hz registered in interrupt mode
[    2.380348] brcmstb-i2c 107d508280.i2c:  @97500hz registered in interrupt mode
[   63.352260] i2c_designware 1f00074000.i2c: i2c_dw_handle_tx_abort: lost arbitration
[   63.352340] i2c_designware 1f00074000.i2c: i2c_dw_handle_tx_abort: lost arbitration
[   63.352408] i2c_designware 1f00074000.i2c: i2c_dw_handle_tx_abort: lost arbitration
[   63.352476] i2c_designware 1f00074000.i2c: i2c_dw_handle_tx_abort: lost arbitration
[   63.352564] i2c_designware 1f00074000.i2c: i2c_dw_handle_tx_abort: lost arbitration
[   63.352627] i2c_designware 1f00074000.i2c: i2c_dw_handle_tx_abort: lost arbitration
[   63.352703] i2c_designware 1f00074000.i2c: i2c_dw_handle_tx_abort: lost arbitration

### Steps to reproduce the behaviour

Enable the i2c interface using raspi-config
Connect any i2c device
run `sudo  i2cdetect -y 1`
sudo dmesg | grep i2c

### Device (s)

Raspberry Pi 5

### System

Raspberry Pi reference 2024-03-15

2024/02/16 15:28:41 
Copyright (c) 2012 Broadcom
version 4c845bd3 (release) (embedded)

Linux raspberrypi-722 6.6.21-v8-16k+ #1 SMP PREEMPT Mon Mar 18 17:00:44 UTC 2024 aarch64 GNU/Linux

### Additional context

I have already flashed a new image , installed most recent firmware and tried with multiple i2c devices

pelwell:
Likely duplicate of #6056.

1. Which devices are attached?
2. Have you set the I2C clock rate?
3. When it fails, what does `pinctrl 2-3` show?

EdgarSouSousa:
1. Currently attached a VEML7700 sensor
2. I have not , I simply enabled the i2c using raspi-config
3.  `2: a3    pu | hi // GPIO2 = SDA1
[...]

PR #6071, opened 2024-03-26
i2c: designware: Add support for bus clear feature

pelwell:
Newer versions of the DesignWare I2C block support the detection of stuck signals, and a mechanism to recover from them. Add the required software support to the driver.

This change was prompted by the observation that reading a single byte from register 0 of a VEML7700 seems to cause it to issue an ACK too early, and the controller to complain about losing arbitration. There is a suspicion that this may be a more widespread problem, but at least this patch prevents the bus from locking up.

See: https://github.com/raspberrypi/linux/issues/6057


pelwell:
I can provoke the lockup using a VEML7700 by running:
```
$ i2cget -y 1 0x10 0; i2cget -y 1 0x10 0
```
Reading pairs of bytes, as expected by the device, does not exhibit the problem:
```
$ i2cget -y 1 0x10 0 w; i2cget -y 1 0x10 0 w
```

pelwell:
It's also worth noting that the same sequence also causes an error on Pi 4 (and presumably the other members of the BCM2835 family). For anyone interested, the early ACK makes the read portion of the second access appear as another write, i.e:
```
SW10A00ASR10AvvNP
SW10A00ASW10NP

S=Start, W=Write, A=ACK, R=Read, N=NAK, P=Stop, vv=Variable value
```

pelwell:
The autobuilds have completed - running `sudo rpi-update pulls/6071` will install a trial, beta kernel including this patch. Remember to back-up important data before installing, just in case.

JinShil:
I tested this on a number of different slave devices and did not observe any regressions.

pelwell:
Thanks - that's good to hear.

PR #6072, opened 2024-03-26
rtc: pcf2127: added battery_low sysfs file

giampiero7:


pelwell:
As with #6070 this is patch which would be better sent to the upstream kernel maintainers. It appears uncontroversial and will probably be accepted quickly.

giampiero7:
Agreed

PR #6074, opened 2024-03-26
allo-boss-dac mute output when changing parameters

alessandromrc:
Since I noticed that sometimes changing sample rates causes some digital quirks and noises, I've changed the function to mute the output before performing the changes and then unmute it when an error occurs or the parameters got set.

Re-opened to add signing and fix all the things mentioned.

### DIFF
--- a/sound/soc/bcm/allo-boss-dac.c
+++ b/sound/soc/bcm/allo-boss-dac.c
@@ -285,7 +285,7 @@ static int snd_allo_boss_hw_params(
 		snd_allo_boss_set_sclk(component, params_rate(params));
 
 		ret = snd_allo_boss_update_rate_den(substream, params);
-		if (ret) 
+		if (ret)
 			goto error;
 	}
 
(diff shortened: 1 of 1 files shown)


pelwell:
Please remove the trailing whitespace and unnecessary braces, then it will be ready to merge. You can either squash your commits yourself and use `git push --force ...` to update GitHub and this PR, or push it normally as a second commit and we can squash them together in the merge.

alessandromrc:
I guess the action didn't like the sign since it picked up the github email address, will wait for a response, so if there's anything to change I will directly make a new commit with such address and name.

alessandromrc:
I guess the action is still detecting braces wrongly... I think I need to put a space under the goto to make it not see the closing brace

pelwell:
No, it's fine - it's just that the two commits are being checked separately, and the first still contains the braces.

alessandromrc:
I hope this is the run :D

PR #6075, opened 2024-03-27
dts: bcm2712: Add blpubkey nvram node

timg236:
The RPi5 firmware also supports exporting the customer public key stored in flash via device-tree. Define the node so that the firmware can populate it.


pelwell:
This PR was giving me deja-vu, and sure enough it evaporated when rebased against rpi-6.6.y top-of-tree.

pelwell:
c.f. #6073

timg236:
For some reason the original one vanished from my view!

PR #6076, opened 2024-03-27
SD command queueing take 2

P33M:
This series adds more robustness when deciding whether to use CQE, adds some user-accessible debug and fixes up some omissions in error handling.

### DIFF
--- a/drivers/mmc/core/sd_ops.c
+++ b/drivers/mmc/core/sd_ops.c
@@ -8,6 +8,7 @@
 #include <linux/slab.h>
 #include <linux/types.h>
 #include <linux/export.h>
+#include <linux/ktime.h>
 #include <linux/scatterlist.h>
 
 #include <linux/mmc/host.h>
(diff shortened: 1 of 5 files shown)


drivers/mmc/core/sd_ops.c
```diff
@@ -460,6 +464,26 @@ static int mmc_sd_cmdq_switch(struct mmc_card *card, bool enable)
 	/* Performance enhancement register byte 262 controls command queueing */
 	err = mmc_sd_write_ext_reg(card, card->ext_perf.fno, card->ext_perf.page,
 				   card->ext_perf.offset + 262, reg);
+	if (err)
+		goto out;
+
+	/* Poll the register - cards may have a lazy init/deinit sequence. */
+	timeout = ktime_add_ms(ktime_get(), 10);
+	while (1) {
+		timedout = ktime_after(ktime_get(), timeout);
```
pelwell: Do you have to check for timeout here? It's counterintuitive, and will result in one extra read after the timeout has elapsed. If intentional, it warrants a comment.

P33M: The timeout length is arbitrary, but the boolean can be removed - no real need to store it separately.

pelwell: Good - it will remove the feeling that something clever might be happening.

PR #6083, opened 2024-04-01
Add support for Chipcap 2 sensor

javiercarrascocruz:
Support for this device was added with v6.9 under hwmon.

Given its relative complexity compared to the i2c devices supported by i2c-sensors, a dedicated overlay has been added.

The config files have also been updated to include its driver as a kernel module.


pelwell:
This is good, as far as it goes, but you should update the arm64 defconfig files as well.

javiercarrascocruz:
> This is good, as far as it goes, but you should update the arm64 defconfig files as well.

I updated the following arm64 defconfig files:

- arch/arm64/configs/bcm2711_defconfig
- arch/arm64/configs/bcm2712_defconfig
- arch/arm64/configs/bcmrpi3_defconfig

PR #6087, opened 2024-04-03
Add SunFounder Pironman 5 overlay

cavonlee:


arch/arm/boot/dts/overlays/sunfounder-pironman5-overlay.dts
```diff
@@ -0,0 +1,52 @@
...
+    fragment@1 {
+        target = <&spi0>;
+        __overlay__ {
+            status = "okay";
+        };
+    };
+    fragment@2 {
+        target-path = "/";
+        __overlay__ {
+            gpio_ir: ir-receiver@12 {
```
pelwell: The convention is to use hexadecimal "addresses" after the "@". The "12" here is 0x12, i.e. 18, which matches the gpio-ir overlay you've based this on. Because you've changed the GPIO to (decimal/denary) 12, you should also change the address to "@c". In that way it can coexist with a regular gpio-ir overlay on GPIO 18.

arch/arm/boot/dts/overlays/sunfounder-pironman5-overlay.dts
```diff
@@ -0,0 +1,52 @@
...
+                linux,rc-map-name = "rc-rc6-mce";
+
+                status = "okay";
+            };
+        };
+    };
+    fragment@3 {
+        target = <&gpio>;
+        __overlay__ {
+            gpio_ir_pins: gpio_ir_pins@12 {
```
pelwell: "@c" again.

arch/arm/boot/dts/overlays/sunfounder-pironman5-overlay.dts
```diff
@@ -0,0 +1,52 @@
+/dts-v1/;
+/plugin/;
+
+/ {
+    compatible = "brcm,bcm2835";
```
pelwell, line 5: We have a lot of overlays that use spaces for indents, including gpio-ir. I'm not bothered about these warnings.

pelwell (commented):
There are a few minor issues I've discussed in inline comments, but the main problem with this overlay is that it doesn't have either a Makefile entry (so it won't get built) or a README entry (so people won't know what it does or how to use it).

Makefile entries should be in alphabetical/ASCII order, as should the README, and README entries need 2 blank lines between them, strict indentation (match the others) and no TAB characters. If you get it wrong, the dtoverlay checker will complain. For this first attempt it said:
```
Overlays without documentation:
  sunfounder-pironman5
Overlays missing from the Makefile:
  sunfounder-pironman5
Failed
```

arch/arm/boot/dts/overlays/sunfounder-pironman5-overlay.dts
```diff
@@ -0,0 +1,52 @@
...
+                brcm,function = <0>;
+                brcm,pull = <2>;
+            };
+        };
+    };
+    __overrides__ {
+        ir = <&gpio_ir>,"status";
+        ir_pins = <&gpio_ir>,"gpios:4", <&gpio_ir>,"reg:0", <&gpio_ir_pins>,"brcm,pins:0", <&gpio_ir_pins>,"reg:0";
+    };
+
```
pelwell: Delete this blank line, and put a newline at the end of the last line to keep git happy.

cavonlee: Yes! Great, Thanks for all the suggestions!

popcornmix:
Note: this was merged to 6.1 kernel which is no longer used by RPiOS.
It will need cherry-picking to 6.6 (and later) kernels before it's in standard builds.

cavonlee:
> Note: this was merged to 6.1 kernel which is no longer used by RPiOS. It will need cherry-picking to 6.6 (and later) kernels before it's in standard builds.

Thanks for the note, So do i need to cherry pick it in 6.6, than make another pull request?

and, when it will be on the raspberry pi os? just wait the next raspberry pi os update? or maybe a kernal upgrade will take affect?

pelwell:
> So do i need to cherry pick it in 6.6,

No, it's OK - it's there now (b429eb7632828f64e5399952da5db10c14f598f8).

> when it will be on the raspberry pi os?

It shouldn't be too long until the next update to the stable branch.

cavonlee:
Great! Thank you pelwell!

PR #6088, opened 2024-04-03
Unicam FS/FE timing GPIO

6by9:
Replicates the firmware option to enable a GPIO to follow CSI frame start/end timings.

https://forums.raspberrypi.com/viewtopic.php?t=368437

### DIFF
--- a/arch/arm/boot/dts/broadcom/bcm270x-rpi.dtsi
+++ b/arch/arm/boot/dts/broadcom/bcm270x-rpi.dtsi
@@ -97,6 +97,19 @@
 		drm_fb0_vc4 = <&aliases>, "drm-fb0=",&vc4;
 		drm_fb1_vc4 = <&aliases>, "drm-fb1=",&vc4;
 		drm_fb2_vc4 = <&aliases>, "drm-fb2=",&vc4;
+
+		cam1_sync = <&csi1>, "sync-gpios:0=", <&gpio>,
+			    <&csi1>, "sync-gpios:4",
+			    <&csi1>, "sync-gpios:8=0", <GPIO_ACTIVE_HIGH>;
+		cam1_sync_inverted = <&csi1>, "sync-gpios:0=", <&gpio>,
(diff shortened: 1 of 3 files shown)


arch/arm/boot/dts/broadcom/bcm270x-rpi.dtsi
```diff
@@ -97,6 +97,19 @@
 		drm_fb0_vc4 = <&aliases>, "drm-fb0=",&vc4;
 		drm_fb1_vc4 = <&aliases>, "drm-fb1=",&vc4;
 		drm_fb2_vc4 = <&aliases>, "drm-fb2=",&vc4;
+
+		cam1_sync = <&csi1>, "sync-gpios:0=", <&gpio>,
+			    <&csi1>, "sync-gpios:4",
+			    <&csi1>, "sync-gpios:8=0"; /* GPIO_ACTIVE_HIGH */
```
pelwell: This is OK, but you can use the named constant:
```
		cam1_sync = <&csi1>, "sync-gpios:0=", <&gpio>,
			    <&csi1>, "sync-gpios:4",
			    <&csi1>, "sync-gpios:8=", <GPIO_ACTIVE_HIGH>;
```

6by9:
@naushir What's your view on this? It could be done from userspace, but with a large round trip delay.

naushir:
Looks reasonable to me.  Probably want to replicate this in the cfe driver?

6by9:
> Looks reasonable to me. Probably want to replicate this in the cfe driver?

Probably yes (caveat around multiple virtual channels), but I'll leave that up to you :-)

The OP was really after the simple flash driver function from the legacy stack, but I don't believe flash has really been considered in libcamera as yet.

As this can be merged without impacting the normal flow of things, I'll drop the draft status even though the OP hasn't come back on it.

6by9:
Any objections to merging this then?

naushir:
LGTM

PR #6089, opened 2024-04-03
dts: bcm2712: cm5: Disable HS400

timg236:
Disable HS400 support on CM5 until this is proved to be stable.


timg236:
@P33M

P33M (approved):

pelwell:
Please try to avoid merge commits - I'd block them on the repo but we use them for pulling in upstream commits.

Issue #6056, opened 2024-03-21
A pulse in ACK signal stops I2C communcation on Pi 5 with kernel 6.6.20


neuralassembly:
### Describe the bug

When I used AQM0802-based LCD on Pi 5 with kernel 6.6.20, it often stops responding to the command, and then it shows nothing on LCD.To make the LCD work again, I have to turn off the LCD.

When I using Pi 4 with kernel 6.1.0/6.6.20 or Pi 5 with kernel 6.1.0, such problem does not take place.

This problem might relate to the issues #5784 and #5988 .
And the pull request #6050 did not affect this issue.

To see what is happening, I observed the SCL and SDA signals with oscilloscope when sending the setting comands to LCD. The results are shown below.

![i2c_graph_all](https://github.com/raspberrypi/linux/assets/3326784/c2d6b6b4-8a3e-4561-93f6-86fc3cefd576)

Figure (A) and (B) show successful results with Pi 5 (6.1.0) and Pi 4 (6.6.20).
Figure (A) on Pi 5 (6.1.0) shows two noisy pulses at the edges of ACK, but they do not affect the I2C communication.

On the other hand, Figure (C) shows the failure result with Pi 5 (6.6.20).
Around t=0.0007, a pluse during ACK signal is observed, and it stops I2C communication.

Figure (C) around t=0.0007 is enlarged below.

![i2c_graph_enlarged](https://github.com/raspberrypi/linux/assets/3326784/af298e87-26bb-48a9-ae12-d68f7bd9ff4f)

When this failure takes place, dmesg shows the follwing error.

`
[  161.891681] i2c_designware 1f00074000.i2c: i2c_dw_handle_tx_abort: lost arbitration
`

After that, I2C communications show the following errors.

`
[ 189.544399] i2c_designware 1f00074000.i2c: controller timed out
`

In such a situation, "i2cdetect -y 1" show the following result.

```
     0  1  2  3  4  5  6  7  8  9  a  b  c  d  e  f
00:                         08 09 0a 0b 0c 0d 0e 0f
10: 10 11 12 13 14 15 16 17 18 19 1a 1b 1c 1d 1e 1f
20: 20 21 22 23 24 25 26 27 28 29 2a 2b 2c 2d 2e 2f
30: -- -- -- -- -- -- -- -- 38 39 3a 3b 3c 3d 3e 3f
40: 40 41 42 43 44 45 46 47 48 49 4a 4b 4c 4d 4e 4f
[...]

PR #6091, opened 2024-04-04
i2c: designware: Make the SDA hold time half LCNT

pelwell:
Here's another attempt at improving the Pi 5 I2C compatibility and reliability. The observation from small LCD displays discussed in #6056 is that something causes the ACK to be released early. Comparing the traces to Pi 4 showed that the ACK signal starts much earlier on Pi 5 - it's roughly halfway through the SCL low on Pi 4. Increasing the SDA hold time to half the SCL low cycle count gives the same appearance on Pi 5 and seems to fix the issue.

The new logic is dependent on there not being an explicit override from Device Tree, so the second part of this PR is a reversion of the commit that added the sda-hold-time-ns values.

### DIFF
--- a/arch/arm/boot/dts/broadcom/rp1.dtsi
+++ b/arch/arm/boot/dts/broadcom/rp1.dtsi
@@ -307,7 +307,6 @@
 			clocks = <&rp1_clocks RP1_CLK_SYS>;
 			i2c-scl-rising-time-ns = <65>;
 			i2c-scl-falling-time-ns = <100>;
-			i2c-sda-hold-time-ns = <300>;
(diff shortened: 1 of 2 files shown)


JinShil:
I tested this on a number of different i2c slave devices and did not observe any regressions.

pelwell:
Thanks again for the testing. I'm hoping now I can draw a line under the RP1 I2C issues.

microfx:
how can I install these patches?

pelwell:
They were merged and will be in all recent kernels, including the ones installed by apt.

PR #6093, opened 2024-04-05
Add overlays and configs for several i2c sensors

javiercarrascocruz:
New nodes for i2c-sensor:
 - hdc3020
 - as73211
 - as7331

New configs to build as modules:
 - HDC3020
 - HTS221, overlay already exists.
 - AS73211
 - AS7331


arch/arm/configs/bcm2709_defconfig
```diff
@@ -1382,8 +1382,10 @@ CONFIG_BCM2835_MBOX=y
 # CONFIG_IOMMU_SUPPORT is not set
 CONFIG_RASPBERRYPI_POWER=y
 CONFIG_IIO=m
+CONFIG_IIO_BUFFER=m
```
pelwell: IIO_BUFFER is a boolean, and already enabled.

arch/arm/configs/bcm2709_defconfig
```diff
@@ -1382,8 +1382,10 @@ CONFIG_BCM2835_MBOX=y
...
 CONFIG_IIO_BUFFER_CB=m
 CONFIG_IIO_SW_TRIGGER=m
+CONFIG_AS73211=m
```
pelwell: This is in the wrong place (well, a different place to where savedefconfig puts it).

arch/arm/configs/bcm2709_defconfig
```diff
@@ -1394,6 +1396,10 @@ CONFIG_SPS30_I2C=m
 CONFIG_MAX30102=m
 CONFIG_DHT11=m
 CONFIG_HDC100X=m
+CONFIG_HDC3020=m
+CONFIG_HTS221=m
+CONFIG_HTS221_I2C=m
+CONFIG_HTS221_SPI=m
```
pelwell: These two (HTS221_I2C and HTS221_SPI) are the defaults if HTS221 is enabled - drop them.

pelwell (commented):
The overlay changes look fine, but the config changes need to be run through "make ... savedefconfig" first. defconfig files are meant to contain only the changes from the defaults, and they should appear in the order in which Kconfig encounters them. The test that you've got it right is that the output of "make ... savedefconfig" matches the _defconfig file you started with.

I don't mind having so many commits, but they will ultimately be squashed down to 2 - dts and defconfigs.

arch/arm/configs/bcm2709_defconfig
javiercarrascocruz: could you please tell me where it is enabled? I did not find it in these defconfigs, but I assume I am missing another one.

pelwell: It's probably selected by one or more of the drivers that depend on it. It's best to apply your source defconfig file then look at the content of .config to see what changes are required.

PR #6095, opened 2024-04-05
Fix UVC gadget support on 32-bit systems

w23:
Note that this is clearly a common upstream problem, not specific to Raspberry Pi. I will work with mainline kernel to get this upstreamed (when I figure out how to do so :)), which will take quite some time.

Given that currently UVC gadget support is completely broken on 6.6 32-bit kernel, I believe that this is valuable for the community to be included into Raspberry Pi kernels sooner than doing it in the roundabout way through upstream.

The problem description follows:

Commit 0df28607c5cb4fe60bba591e9858a8f7ba39aa4a introduced a helper function to parse commonly occurring strings containing lists of numbers. Original functions used `buf` buffer on stack, and queried its size using `sizeof(buf)` to avoid overrun. The newly intruduced function, however, used a dynamically sized heap-allocated buffer, but wasn't updated to use this new buffer size instead of just sizeof.

Therefore, parsing now can only read items of size 3 max on 32-bit systems, and of size 7 on 64-bit ones. Some UVC fields do need numbers with more digits, e.g. `dwFrameInteval` needs at least 5-6, so it fails to be written into, thus completely breaking UVC gadget support on 32 bit systems.

### DIFF
--- a/drivers/usb/gadget/function/uvc_configfs.c
+++ b/drivers/usb/gadget/function/uvc_configfs.c
@@ -92,10 +92,10 @@ static int __uvcg_iter_item_entries(const char *page, size_t len,
 
 	while (pg - page < len) {
 		i = 0;
-		while (i < sizeof(buf) && (pg - page < len) &&
+		while (i < bufsize && (pg - page < len) &&
 		       *pg != '\0' && *pg != '\n')
 			buf[i++] = *pg++;
-		if (i == sizeof(buf)) {
+		if (i == bufsize) {
 			ret = -EINVAL;
 			goto out_free_buf;
 		}


pelwell:
GregKH is a very busy man, and moving the code obscured the changes, but that's still really bad. I expect your patch (https://lore.kernel.org/lkml/20240411164616.4130163-1-me@provod.works/raw) to be welcomed upstream with open arms - isn't that right, @pinchartl?

Can I append your S-o-b (`Signed-off-by: Ivan Avdeev <me@provod.works>`) when I merge?

w23:
> Can I append your S-o-b (`Signed-off-by: Ivan Avdeev <me@provod.works>`) when I merge?

Absolutely!
It would probably make sense to just copy the whole patch description that I sent to the mainline instead of this description when merging. I made this PR a few days before I had a chance to read on how to make kernel patches probably, so it's not really compliant :).

pelwell:
Superseded by https://github.com/raspberrypi/linux/commit/1dc282dffafb493fed2cf4ff9d4706f810a26c8c.

PR #6099, opened 2024-04-10
module: Avoid ABI changes when debug info is disabled

popcornmix:
CI builds are done with debug info disabled, but this removes some members from struct module.  This causes builds to fail if there is an ABI reference for the current ABI.

Define these members unconditionally, so that there is no ABI change.


popcornmix:
This adds debian patch from [here](https://sources.debian.org/src/linux/6.6.15-2/debian/patches/debian/module-avoid-abi-changes-when-debug-info-is-disabled.patch/) to make manual/rpi-update kernel builds closer to apt builds.

PR #5867, opened 2024-01-18
drm: rp1: rp1-vec: Experimental support for the 405-line TV standard

njhollinghurst:
Add an additional hardware mode which displays a 720x378 frame in the 405-line TV standard, for testing.

PR #6105, opened 2024-04-16
drm: rp1: vec: Support more video modes in the RP1 VEC driver

njhollinghurst:
This is the replacement for #5867 which adds monochrome TV mode and the "vintage" TV systems, plus some other bells and whistles:

Existing progressive modes have unpleasant aspect ratios: this commit supports both low-resolution modes (e.g. 320x240) and high-resolution "fake" ones (e.g. 640x480). As part of this, it generalizes how pixel clock rates are handled.

So as not to clutter the list of advertised modes, additional progressive modes can only be selected as a "custom mode". Added a module parameter to define a custom mode from its timings. This also lets users tweak image dimensions and offsets.

Not added right now:
- Not added support for picture aspect ratio, because I think most software ignores it (and it would get conflated with WSS, which RP1 can't do).
- No attempt to implement "margins": we could hack something in the horizontal axis, but the lack of scaler precludes margins in the vertical axis.
- The "tv_norm" parameter is ~tentatively back in~ removed again, as there's no point duplicating tv_mode.

### DIFF
--- a/drivers/gpu/drm/rp1/rp1-vec/rp1_vec.c
+++ b/drivers/gpu/drm/rp1/rp1-vec/rp1_vec.c
@@ -48,6 +48,63 @@
 
 #include "rp1_vec.h"
 
+/*
+ * Linux doesn't make it easy to create custom video modes for the console
+ * with non-CVT timings; so add a module parameter for it. The format is:
+ * "<pclk>,<hact>,<hfp>,<hsync>,<hbp>,<vact>,<vfp>,<vsync>,<vbp>[,i]"
+ * (where each comma may be replaced by any sequence of punctuation).
+ * pclk should be 108000/n for 5 <= n <= 16 (twice this for "fake" modes).
+ */
+
+static char *rp1vec_cmode_str;
+module_param_named(cmode, rp1vec_cmode_str, charp, 0600);
+MODULE_PARM_DESC(cmode, "Custom video mode:\n"
+		 "\t\t<pclk>,<hact>,<hfp>,<hsync>,<hbp>,<vact>,<vfp>,<vsync>,<vbp>[,i]\n");
+
+static struct drm_display_mode *rp1vec_parse_custom_mode(struct drm_device *dev)
+{
+	char const *p = rp1vec_cmode_str;
+	struct drm_display_mode *mode;
+	unsigned int n, vals[9];
+
+	if (!p)
+		return NULL;
+
+	for (n = 0; n < 9; n++) {
+		unsigned int v = 0;
+
+		if (!isdigit(*p))
+			return NULL;
+		do {
+			v = 10u * v + (*p - '0');
+		} while (isdigit(*++p));
+
+		vals[n] = v;
+		while (ispunct(*p))
+			p++;
+	}
+
+	mode = drm_mode_create(dev);
+	if (!mode)
+		return NULL;
+
+	mode->clock	  = vals[0];
+	mode->hdisplay	  = vals[1];
+	mode->hsync_start = mode->hdisplay + vals[2];
+	mode->hsync_end	  = mode->hsync_start + vals[3];
+	mode->htotal	  = mode->hsync_end + vals[4];
+	mode->vdisplay	  = vals[5];
+	mode->vsync_start = mode->vdisplay + vals[6];
+	mode->vsync_end	  = mode->vsync_start + vals[7];
+	mode->vtotal	  = mode->vsync_end + vals[8];
+	mode->type  = DRM_MODE_TYPE_DRIVER | DRM_MODE_TYPE_PREFERRED;
+	mode->flags = DRM_MODE_FLAG_NHSYNC | DRM_MODE_FLAG_NVSYNC;
+	if (strchr(p, 'i'))
+		mode->flags |= DRM_MODE_FLAG_INTERLACE;
+
+	return mode;
+}
+
 static void rp1vec_pipe_update(struct drm_simple_display_pipe *pipe,
 			       struct drm_plane_state *old_state)
 {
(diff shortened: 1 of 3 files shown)


njhollinghurst:
I think this is ready now.

Compared to the timings in #5867 I have fixed a small boundary glitch in 405-line, and a horizontal offset in 819-line. Other modes' horizontal offsets may have shifted by a pixel or two due to the different calculation but I think are accurate enough.

drivers/gpu/drm/rp1/rp1-vec/rp1_vec.c
```diff
@@ -48,6 +48,62 @@
...
+		} while (isdigit(*++p));
+
+		if (((n >= 2 && n <= 4) || (n >= 6 && n <= 8)) && v < vals[n - 1])
+			v += vals[n - 1];
+		vals[n] = (int)v;
+		while (ispunct(*p))
+			p++;
+	}
+
+	mode = drm_mode_create(dev);
```
6by9, line 87: Can fail.
```
if (!mode)
  return NULL;
```

drivers/gpu/drm/rp1/rp1-vec/rp1_vec.c
```diff
@@ -48,6 +48,62 @@
...
+		int v = 0;
+
+		if (!isdigit(*p))
+			return NULL;
+		do {
+			v = 10 * v + (*p - '0');
+		} while (isdigit(*++p));
+
+		if (((n >= 2 && n <= 4) || (n >= 6 && n <= 8)) && v < vals[n - 1])
+			v += vals[n - 1];
```
6by9: I had to look twice at this to work out what and why.

You're assigning to the mode parameters in order lower down, so switch those to
```
	mode->hdisplay	  = vals[1];
	mode->hsync_start = mode->hdisplay + vals[2];
	mode->hsync_end	  = mode->hsync_start + vals[3];
	mode->htotal	  = mode->hsync_end + vals[4];
....
```
instead?

drivers/gpu/drm/rp1/rp1-vec/rp1_vec.c
```diff
@@ -48,6 +48,62 @@
...
+static char *rp1vec_cmode_str;
+module_param_named(cmode, rp1vec_cmode_str, charp, 0600);
+MODULE_PARM_DESC(cmode, "Custom video mode:\n"
+		 "\t\t<pclk>,<hact>,<hss>,<hse>,<htotal>,<vact>,<vss>,<vse>,<vtotal>[,i]\n");
+
+static struct drm_display_mode *rp1vec_parse_custom_mode(struct drm_device *dev)
+{
+	char const *p = rp1vec_cmode_str;
+	struct drm_display_mode *mode;
+	int n, vals[9];
```
6by9: unsigned int for vals as it'll never be negative?

drivers/gpu/drm/rp1/rp1-vec/rp1_vec.c
```diff
@@ -48,6 +48,62 @@
...
+	mode->hsync_start = vals[2];
+	mode->hsync_end	  = vals[3];
+	mode->htotal	  = vals[4];
+	mode->vdisplay	  = vals[5];
+	mode->vsync_start = vals[6];
+	mode->vsync_end	  = vals[7];
+	mode->vtotal	  = vals[8];
+	mode->type = DRM_MODE_TYPE_DRIVER | DRM_MODE_TYPE_PREFERRED;
+	mode->flags =
+		DRM_MODE_FLAG_NHSYNC | DRM_MODE_FLAG_NVSYNC | DRM_MODE_FLAG_CSYNC;
```
6by9: Whilst DRM_MODE_FLAG_CSYNC is technically correct, it is redundant as there is no other option.
I don't believe the NHSYNC and NVSYNC will do anything useful either, but it does match with [fill_analog_mode](https://elixir.bootlin.com/linux/latest/C/ident/fill_analog_mode)

drivers/gpu/drm/rp1/rp1-vec/rp1_vec.h
```diff
@@ -20,13 +20,18 @@
 #define RP1VEC_HW_BLOCK_CFG   1
 #define RP1VEC_NUM_HW_BLOCKS  2
 
+#define RP1VEC_TV_MODE_MONOCHROME (DRM_MODE_TV_MODE_SECAM + 1)
```
6by9: Why is RP1VEC_TV_MODE_MONOCHROME needed still? Our tree has `DRM_MODE_TV_MODE_MONOCHROME` merged, and we'll fight the battle over mono when we come to upstream RP1 VEC (likely just dropping it).

6by9 (commented):
I'll trust you on the register calcs, and I'm largely only commenting on the DRM API stuff.

drivers/gpu/drm/rp1/rp1-vec/rp1_vec.h
njhollinghurst: It was in case somebody wanted to pull it into another fork and have it at least compile.
Can remove.

drivers/gpu/drm/rp1/rp1-vec/rp1_vec.c
njhollinghurst: Ah. I should check the `drm_mode_duplicate` at line 351 also.

drivers/gpu/drm/rp1/rp1-vec/rp1_vec.c
njhollinghurst: There was also a `v < vals[n - 1]` in the condition. Can remove that and change the parameter description to say HFP, HSYNC_WIDTH, HBP.
It's quite annoying there are these two different conventions; as well as clock rate in MHz (fp), kHz or Hz.

drivers/gpu/drm/rp1/rp1-vec/rp1_vec.c
njhollinghurst: OK I'd better make them match.

drivers/gpu/drm/rp1/rp1-vec/rp1_vec.c
6by9: Oh sorry, I had misread your text and was thinking you were passing in `hact,hfp,hsync,hbp`, but it's `<hact>,<hss>,<hse>,<htotal>`.

In which case I don't understand that `v += vals[n - 1]`. Are you trying to handle the case of someone passing in hact,hfp etc? I'd say don't, and just through out the whole mode line as it is invalid.

(You may be able to allow it to be thrown out later as `drm_mode_validate_basic` should get called on the mode, or you could just call it yourself).

njhollinghurst: Yes it was convenient for it to accept both while I was testing. What do you think re pixel clock? `kmstest` accepts Hz (though it prints MHz) which I'm not keen on as the last 3 digits have nowhere to go in `struct drm_display_mode`.

drivers/gpu/drm/rp1/rp1-vec/rp1_vec_hw.c
```diff
@@ -46,334 +53,371 @@ static const struct rp1vec_ipixfmt my_formats[] = {
...
+			.max_rows_per_field = 240,
+			.ref_vfp = 2,
+			.interlaced = false,
+			.first_field_odd = false,
+			.scale_v = 0x0cf,
+			.scale_u = 0x074,
+			.scale_y = 0x107,
+			.de_end	 = 0x1a4f,
+			.de_bgn	 = 0x038f,
+			.half_lines_per_field = 526,
```
njhollinghurst: I might need to change this number to 524, to match the PAL-M version. typical user expectations and indeed the advertised DRM mode. Though sadly that will cause more colour artifacts.

drivers/gpu/drm/rp1/rp1-vec/rp1_vec_hw.c
```diff
@@ -46,334 +53,371 @@ static const struct rp1vec_ipixfmt my_formats[] = {
...
 	 */
-	hpad = max(0, mode->hsync_start - hwm->ref_hfp - w);
-	hpad = min(hpad, hwm->total_cols - w);
-	vpad = max(0, ((mode->vsync_start - hwm->ref_vfp) >> mode_ilaced) - h);
-	vpad = min(vpad, hwm->rows_per_field - h);
+	i = (vec->fake_31khz) ? (mode->clock >> 1) : mode->clock;
+	rate = (i < (RP1VEC_VDAC_KHZ / 16)) ? 16 : max(4, (RP1VEC_VDAC_KHZ + 256) / i);
+	wmax = min((hwm->de_end - hwm->de_bgn) / rate, 1020);
+	w = min(mode->hdisplay, wmax);
+	ref_2mid = (hwm->de_bgn + hwm->de_end) / rate + 6; /* + 6 for FIR delay */
```
njhollinghurst: Could stand to reduce the magic offset to 4, shifting everything right by 1 pixel(!), to more precisely match Pi 4's video out (in Rec.601 modes at least).

drivers/gpu/drm/rp1/rp1-vec/rp1_vec.c
6by9: In which case go with what is documented, and drop the testing code.

kHz probably makes most sense as that is what `struct drm_display_mode` uses.

drivers/gpu/drm/rp1/rp1-vec/rp1_vec.c
```diff
@@ -158,79 +214,156 @@ static void rp1vec_connector_destroy(struct drm_connector *connector)
...
+	int real_clock	  = mode->clock >> fake_31khz;
+
+	/* Check pixel clock is in a sane range */
+	if (real_clock < 6016)
+		return MODE_CLOCK_LOW;
+	else if (real_clock > 27000)
+		return MODE_CLOCK_HIGH;
+
+	/* Try to match against the 525-line 60Hz mode (System M) */
+	if (vtotal_2fld >= 500 && vtotal_2fld < 544 && vdisplay_2fld <= 486 &&
```
njhollinghurst: The line-count test here is awfully permissive, the reason was to allow command-line modes like `640x480M` (which is a CVT mode incompatible with TV systems). Now we have `cmode` that's not a good reason, so I think I shall make it stricter again.

drivers/gpu/drm/rp1/rp1-vec/rp1_vec.c
njhollinghurst: I'll go for khz and porches (rather than totals), as that keeps the command line short and feels nicest.

drivers/gpu/drm/rp1/rp1-vec/rp1_vec_hw.c
njhollinghurst: It's so simple to adapt that I've decided to offer both, depending on `cmode`. The default will be 262 lines.

njhollinghurst:
Latest push
- addresses all review comments
- tightens up mode validation a bit (to encourage correct modes) and limits real pixel clock to 21.6MHz (was 27MHz)
- makes progressive NTSC modes 262 lines by default (was 263; but we allow a total of 263 lines via a custom mode)
- tweaks an offset in horizontal timing calculations

drivers/gpu/drm/rp1/rp1-vec/rp1_vec_hw.c
njhollinghurst: Hmm, should we likewise support both 312 and 313 lines for Progressive PAL? I thought so but have changed my mind.
Unlike 263 lines, 313 seems to provide no benefit, and one video converter dislikes it (it displays in black and white).

bagelbyheart:
How would one take advantage of the new options from `config.txt` or `cmdline.txt`?

njhollinghurst:
> How would one take advantage of the new options from `config.txt` or `cmdline.txt`?

I'm waiting for this to reach a stable Raspberry Pi OS release before documenting fully.

It's done via command line. For the "vintage TV" modes see the `video=` options [here](https://forums.raspberrypi.com/viewtopic.php?t=363432&start=50#p2217025). Other modes are created using the `drm-rp1-vec.cmode` parameter of the form `pixrate,hactive/hfp/hsw/hbp,vactive/vfp/vsw/vbp[,i]`; examples [here](https://forums.raspberrypi.com/viewtopic.php?t=366652#p2214271).

There are some restrictions on custom modes:
- Pixel rate (in kHz) should be 108000/_n_ (rounded to an integer) for integer 5 <= _n_ <= 16.
- Pixel rate divided by horizontal total must match a real TV mode's line rate, i.e. 15.625kHz or 15.734kHz.
- Line count (vertical total) must match a real TV mode if interlaced, or half this +/-0.5 if progressive.
- [There are also "fake" progressive modes with twice the standard pixel-rate, line-rate and vertical total. This is a hack for aspect ratio; only every other scanline is displayed.]

bagelbyheart:
I just realized, this doesn't apply to raspberrypi0-4 does it? Is there any way to set custom composite out resolutions there that you're aware of?

njhollinghurst:
It doesn't apply there. I know rather less about the VC4 VEC but it looks more restrictive in terms of pixel clock and frame sizes.

On the other hand, the VC4 VEC is connected to the Hardware Video Scaler (and uses HVS to implement overscan/margins), which means it has less need for custom output modes.

bagelbyheart:
Yeah that's what I was thinking, I am not quite at the point of being able to untangle that code myself. cest'la. Thanks for the quick reply =)

6by9:
vc4 VEC is more restricted.
It takes the timing from the PixelValve which can do almost anything in the way of timings, but the encoder is then fairly hard coded around the various video standards.

`kmstest -r ...` does allow you to supply arbitrary timings for the connector, so you could try mangling the mode there. I suspect changing the pixel clock will perform badly, but there should be no harm in trying. If it does work then I may copy across the custom mode code from RP1.
More generally for vc4 you're better off using the HVS to scale and position your plane in the appropriate place on your display output.

Issue #6104, opened 2024-04-16
Regression with Hifiberry DAC+ Pro since switch to params_width


HiassofT:
### Describe the bug

We got a report on our forums that the Hifiberry DAC+ Pro produced choppy audio in Kodi with our latest releases, running kernel 6.6.25. The symptoms looked a lot like the DAC isn't running at the correct rate and Kodi (which outputs 24bit audio) was constantly trying to adjust for it.

https://forum.libreelec.tv/thread/28386-rpi-4-choppy-audio-with-hifiberry-dac2-pro/?postID=190889#post190889

Glancing through the commit history I stumbled upon f0a501b81926 which switched the hifiberry_dacplus driver from `snd_pcm_format_physical_width` to `snd_pcm_format_width`.

With this switch reverted the user reported that kodi worked fine again.

It looks like this driver (and possible others, too), deliberately uses physical_width to ensure the bclk_ratio is either 16 or 32 so the clock rate can be cleanly derived from the oscillators by an integer divider.

### Steps to reproduce the behaviour

Play 24bit audio on a Hifiberry DAC+ Pro (`speaker-test -F S24_LE ...` should do the trick and likely produce slightly off-pitch audio)

### Device (s)

Raspberry Pi 4 Mod. B

### System

LibreELEC 12 beta 2, kernel 6.6.25

### Logs

https://paste.libreelec.tv/logical-cowbird.log

### Additional context

Using physical_width to get either 16 or 32 isn't quite correct as well, as it would fail in case of S24_3LE audio (which, fortunately, isn't being supported on RPis yet), and without additional comments is hard to distinguish from a typo. So it's be better to make the intention clear, eg by using a switch or sth like `width = width <=16 ? 16 : 32`

I suspect other cards may have regressed as well by this change, from a quick glance eg the Allo Boss DAC driver seems to use two fixed frequency oscillators as well and the driver looks like a copycat of the hifiberry one.

ping @pelwell @j-schambacher

j-schambacher:
@HiassofT I'll take a look within the next days. Thanks for your proposal. I'll test that, but will take some time.

pelwell:
I'm sure @HiassofT is correct. I'll create a PR with the proposed fix, i.e. using snd_pcm_format_width, then forcing either 16 or 32 depending.

j-schambacher:
Agreed! Thanks, Phil!

pelwell:
#6110 is the suggested fix.

PR #6110, opened 2024-04-17
ASoC: bcm: Use power-of-2 bclk_ratios

pelwell:
The soundcard drivers originally used snd_pcm_format_physical_width, but a later commit changed that to snd_pcm_format_width because the in-memory sample storage width should not be a factor in determining the bclk_ratio. However, the physical width rounds the sample bits up to the nearest power of 2, which makes it easier to find integer clock divisors.

Restore the old behaviour, but with an implementation that makes it clear what is going on.

See: https://github.com/raspberrypi/linux/issues/6104

### DIFF
--- a/sound/soc/bcm/allo-boss-dac.c
+++ b/sound/soc/bcm/allo-boss-dac.c
@@ -278,6 +278,9 @@ static int snd_allo_boss_hw_params(
 	struct snd_soc_component *component = asoc_rtd_to_codec(rtd, 0)->component;
 	struct snd_soc_card *card = rtd->card;
 
+	/* Using powers of 2 allows for an integer clock divisor */
+	width = width <= 16 ? 16 : 32;
+
 	/* Mute before changing sample rate */
 	snd_allo_boss_gpio_mute(card);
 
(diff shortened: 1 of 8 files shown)


HiassofT:
Thanks a lot, this looks fine to me.

I don't have any of the soundcards that caused noticable issues here but I could verify that this PR works fine with my Cirrus Logic audio card (rpi-cirrus driver):

Without this PR the driver requested 2 slots of 24 bit, but the arizona driver choose to use the next higher bclk rate that was a multiple of the sample rate so it worked fine. As the interface is running in I2S mode with the RPi as a clock consumer the slight misconfiguration on the RPi side didn't matter.

dmesg with dynamic debug of snd_soc_arizona enabled showed:
```
wm5102-codec wm5102-codec: AIF1: Configuring for 2 24 bit TDM slots
wm5102-codec wm5102-codec: AIF1: BCLK 2822400Hz LRCLK 44100Hz
```

With this PR the driver now correctly requests 2 slots of 32bit which matches the actually supported bclk rate and the configuration is fine both on RPi and codec side
```
wm5102-codec wm5102-codec: AIF1: Configuring for 2 32 bit TDM slots
wm5102-codec wm5102-codec: AIF1: BCLK 2822400Hz LRCLK 44100Hz
```

pelwell:
Thanks for reporting this.

PR #6113, opened 2024-04-18
Fix up rotation support for panels

6by9:
We had a bad port from 6.1 to 6.6 which resulted in losing the generic parsing of "rotation" from DT to populate the "panel orientation" property.

Also reimplement dropping of custom handling patches to remove get_orientation as well as remove reading the orientation.

https://forums.raspberrypi.com/viewtopic.php?t=369197


6by9:
Any comments on this PR?

popcornmix:
Looks fine to me.

PR #6116, opened 2024-04-19
overlays/README: Document that vc4-(f)kms requires >=512MB

6by9:
The firmware stops vc4-kms-v3d and vc4-fkms-v3d loading if the system has less than 512MB of RAM. It can work if gpu_mem and CMA heap size are set appropriately, but can't be guaranteed. Document this restriction.

Also drops Eric's name from the overlay description as it isn't relevant or accurate anymore.

Raised due to https://github.com/raspberrypi/rpicam-apps/issues/668

### DIFF
--- a/arch/arm/boot/dts/overlays/README
+++ b/arch/arm/boot/dts/overlays/README
@@ -4864,8 +4864,10 @@ Params: <None>
 
 
 Name:   vc4-fkms-v3d
-Info:   Enable Eric Anholt's DRM VC4 V3D driver on top of the dispmanx
+Info:   Enable the kernel DRM VC4 V3D driver on top of the dispmanx
         display stack.
+        NB The firmware will not allow this overlay to load on a Pi with less
+        than 512MB as memory is too tight.
 Load:   dtoverlay=vc4-fkms-v3d,<param>
 Params: cma-512                 CMA is 512MB (needs 1GB)
         cma-448                 CMA is 448MB (needs 1GB)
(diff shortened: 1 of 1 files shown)


arch/arm/boot/dts/overlays/README
```diff
@@ -4881,8 +4883,10 @@ Params: cma-512                 CMA is 512MB (needs 1GB)
 
 
 Name:   vc4-fkms-v3d-pi4
-Info:   Enable Eric Anholt's DRM VC4 V3D driver on top of the dispmanx
+Info:   Enable the kernel DRM VC4 V3D driver on top of the dispmanx
         display stack.
+        NB The firmware will not allow this overlay to load on a Pi with less
```
pelwell: Moot (are there any Pi 4s with less than 512MB)?

arch/arm/boot/dts/overlays/README
```diff
@@ -5171,7 +5177,9 @@ Params: cma-512                 CMA is 512MB (needs 1GB)
 
 
 Name:   vc4-kms-v3d-pi4
-Info:   Enable Eric Anholt's DRM VC4 HDMI/HVS/V3D driver for Pi4.
+Info:   Enable the kernel DRM VC4 HDMI/HVS/V3D driver for Pi4.
+        NB The firmware will not allow this overlay to load on a Pi with less
```
pelwell: As above.

arch/arm/boot/dts/overlays/README
6by9: True. Removed.

PR #6117, opened 2024-04-19
Add a strict_gpiod option

pelwell:
This patch set addresses some of the comments on https://github.com/raspberrypi/linux/commit/022689f0973d87956b2e5e8aaa0c29803cdb2a71. The concern is that GPIOs should always be returned to being inputs when they are released so that, for example, a heater is switched off or a motor stops when the controlling application crashes. That is how the upstream pinctrl drivers work, but in the downstream RPi kernels this behaviour has been modified so that GPIO outputs remain as outputs with whatever drive they had, in order that changes from gpioset will persist without the need to keep gpioset as a background process.

Two of the patches add support for the `strict_gpiod` module parameter to the two pinctrl drivers. Another patch adds dtparams so that the change can be made in config.txt instead, and will apply to all models of Pi. The final patch adds a `gpio-ranges` property to Pi 5 - without it, the notification of a pin being released never reaches the pinctrl driver.

### DIFF
--- a/drivers/pinctrl/pinctrl-rp1.c
+++ b/drivers/pinctrl/pinctrl-rp1.c
@@ -573,6 +573,10 @@ static const char * const irq_type_names[] = {
 	[IRQ_TYPE_LEVEL_LOW] = "level-low",
 };
 
+static bool strict_gpiod;
+module_param(strict_gpiod, bool, 0644);
+MODULE_PARM_DESC(strict_gpiod, "unless true, outputs remain outputs when freed");
+
 static int rp1_pinconf_set(struct pinctrl_dev *pctldev,
 			   unsigned int offset, unsigned long *configs,
 			   unsigned int num_configs);
@@ -1201,8 +1205,8 @@ static int rp1_pmx_free(struct pinctrl_dev *pctldev, unsigned offset)
 	struct rp1_pin_info *pin = rp1_get_pin_pctl(pctldev, offset);
 	u32 fsel = rp1_get_fsel(pin);
 
-	/* Return non-GPIOs to GPIO_IN */
-	if (fsel != RP1_FSEL_GPIO) {
+	/* Return non-GPIOs to GPIO_IN, unless strict_gpiod is set */
+	if (strict_gpiod || fsel != RP1_FSEL_GPIO) {
 		rp1_set_dir(pin, RP1_DIR_INPUT);
 		rp1_set_fsel(pin, RP1_FSEL_GPIO);
 	}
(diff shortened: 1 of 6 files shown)


drivers/pinctrl/pinctrl-rp1.c
```diff
@@ -1201,8 +1205,9 @@ static int rp1_pmx_free(struct pinctrl_dev *pctldev, unsigned offset)
 	struct rp1_pin_info *pin = rp1_get_pin_pctl(pctldev, offset);
 	u32 fsel = rp1_get_fsel(pin);
 
-	/* Return non-GPIOs to GPIO_IN */
-	if (fsel != RP1_FSEL_GPIO) {
+	/* Return non-GPIOs to GPIO_IN, unless strict_gpiod is set */
+	pr_err("%s(%d) - strict_gpiod %d\n", __func__, offset, strict_gpiod);
```
6by9: Is the `pr_err` left over debug? If really wanted, it should be `dev_err`.

6by9 (commented):
I'm glad it's not just me that gets caught by the dtoverlaycheck failure :-)

drivers/pinctrl/pinctrl-rp1.c
pelwell: Yes - it's been deleted.

lategoodbye:
Patch for Mainline Kernel:
https://lore.kernel.org/linux-gpio/20240419204057.86078-1-wahrenst@gmx.net/T/#u

lategoodbye:
The second version with parameter name `persist_gpio_outputs` has been accepted mainline. Since i like to align the behavior between rpi and mainline, the question is how we should proceed here.
https://lore.kernel.org/linux-gpio/CACRpkdY5pbtCU=hFaQ0MmYqMy5xWCxV6HPcEoRMMNEDL657crQ@mail.gmail.com/T/#t

Can we adapt to the mainline parameter name or is it too late and we need to handle both parameter names?

pelwell:
The name change is easy enough - the abstraction through a dtparam helps. It's the inversion of the sense combined with the use of module parameters that is awkward/ugly, given that kernel updates are independent of changes to /etc/modprobe.d, etc.

pelwell:
Corresponding downstream changes can be found here: https://github.com/raspberrypi/linux/pull/6150

lategoodbye:
The upstream patch has been applied to 6.10-rc1, but there were comments after it has been applied:
https://lore.kernel.org/linux-gpio/Zjk-C0nLmlynqLAE@surfacebook.localdomain/

Does it makes sense to adjust the parameter persist_gpio_outputs via sysfs afterwards or should we make it readonly via sysfs?

I need to know your opinion before i send follow-up patches.

pelwell:
I think there are people who will want one value, and people who will want the other value, but likely nobody who wants a changeable global value (aside from the convenience of avoiding a reboot after it is set to the preferred value). Having said that, I have no problem with it being changeable at runtime.

lategoodbye:
I don't have a strong preference here. But i value the benefits of a readonly sysfs entry:
- the module parameter would be equivalent to the DT parameter
- possible applications can rely on the behavior by reading the sysfs entry once
- comment easier to fix (no need to argue about locking)

So you actually don't have a problem with making this readonly?

pelwell:
> So you actually don't have a problem with making this readonly?

That's correct. In the past it was never strict in the downstream world, and now users have the option of making it strict - that's enough.

PR #6118, opened 2024-04-19
dw-axi-dmac-platform: Avoid trampling with zero length buffer

popcornmix:
This code:
```
for_each_sg(sgl, sg, sg_len, i)
  num_sgs += DIV_ROUND_UP(sg_dma_len(sg), axi_block_len);
```
determines how many hw_desc are allocated.
If `sg_dma_len(sg)=0` we don't allocate for this sgl.

However in the next loop, we will increment loop
for this case, and loop gets higher than num_sgs
and we trample memory.

### DIFF
--- a/drivers/dma/dw-axi-dmac/dw-axi-dmac-platform.c
+++ b/drivers/dma/dw-axi-dmac/dw-axi-dmac-platform.c
@@ -916,7 +916,7 @@ dw_axi_dma_chan_prep_slave_sg(struct dma_chan *dchan, struct scatterlist *sgl,
 		mem = sg_dma_address(sg);
 		len = sg_dma_len(sg);
 		num_segments = DIV_ROUND_UP(sg_dma_len(sg), axi_block_len);
-		if (num_segments == 0)
+		if (!num_segments)
 			continue;
 
 		segment_len = DIV_ROUND_UP(sg_dma_len(sg), num_segments);
--- a/drivers/dma/dw-axi-dmac/dw-axi-dmac-platform.c
+++ b/drivers/dma/dw-axi-dmac/dw-axi-dmac-platform.c
@@ -916,6 +916,9 @@ dw_axi_dma_chan_prep_slave_sg(struct dma_chan *dchan, struct scatterlist *sgl,
 		mem = sg_dma_address(sg);
 		len = sg_dma_len(sg);
 		num_segments = DIV_ROUND_UP(sg_dma_len(sg), axi_block_len);
+		if (num_segments == 0)
+			continue;
+
 		segment_len = DIV_ROUND_UP(sg_dma_len(sg), num_segments);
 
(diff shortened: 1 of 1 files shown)


popcornmix:
[This](https://forums.raspberrypi.com/viewtopic.php?t=359876) is the forum bug report:

Let me test this at home.

In the office the TV signal is very low, so I wasn't able to run for long,
but I did have some seconds of playback, using 16k pagesize kernel with this PR,
compared to usual kernel panic and need to reboot.

popcornmix:
Testing at home is positive.
Running on Pi5 with 16k pagesize kernel and its recording the snooker just fine.

@pelwell do you agree with my logic?

drivers/dma/dw-axi-dmac/dw-axi-dmac-platform.c
```diff
@@ -918,6 +918,9 @@ dw_axi_dma_chan_prep_slave_sg(struct dma_chan *dchan, struct scatterlist *sgl,
 		num_segments = DIV_ROUND_UP(sg_dma_len(sg), axi_block_len);
 		segment_len = DIV_ROUND_UP(sg_dma_len(sg), num_segments);
```
pelwell: If num_segments can be zero on the next line, why didn't we get a division-by-zero error here?

pelwell: Looking at a disassembly of the output, the compiler seems to have moved the `(num_segments == 0)` test (which should by `(!num_segments)` BTW) between the two DIV_ROUND_UP calls, hence the lack of an exception.

popcornmix: There was no divide by zero exception before this PR where the `num_segments == 0` protection wasn't present, nor with logging added which printed the result of the divide by zero (which was 0).
For interest, @pelwell did find:
```
  ARM Architecture Reference Manual, ARMv8 [ARM DDI 0487B.b]
      http://infocenter.arm.com/help/topic/com.arm.doc.ddi0487b.b/DDI0487B_b_armv8_arm.pdf
      1) aarch64 instruction set: section C3.4.7 and C6.2.279 (UDIV)
         "A division by zero results in a zero being written to
          the destination register, without any indication that
          the division by zero occurred."
      2) aarch32 instruction set: section F1.4.8 and F5.1.263 (UDIV)
         "For the SDIV and UDIV instructions, division by zero
          always returns a zero result."
```

pelwell: The !num_segments test should go before the segment_len calculation, if only for appearances, but otherwise I agree with this commit and the reasoning behind it.

popcornmix:
Moved the test up a line to avoid the potential divide by zero.

PR #6121, opened 2024-04-22
dts: overlay: ov5647: Specify clock-noncontinuous on CSI endpoint

kleinweby:
I've encountered frontend timeouts with the ov5647 on a Raspberry PI 3+ when using a longer cable (~30cm-ish). None of the odd suggestions on the internet worked to fix that, but it suggests that the problem gets worse with longer cables. I did not test a short cable, however an imx477 works just fine with my setup.

After some trail-and-error I've stumbled across the fact, that the clock-noncontinuous parameter specified in the device tree was not honoured. Fixing this, and therefore enabling the non-continuous clock mode fixed the frontend timeouts in my setup.

### DIFF
--- a/arch/arm/boot/dts/overlays/camera-mux-2port-overlay.dts
+++ b/arch/arm/boot/dts/overlays/camera-mux-2port-overlay.dts
@@ -453,6 +453,7 @@
 			      <&imx708_0>,"lens-focus:0=", <&imx708_0_vcm>;
 		cam0-ov5647 = <&mux_in0>, "remote-endpoint:0=",<&ov5647_0_ep>,
 			      <&ov5647_0_ep>, "remote-endpoint:0=",<&mux_in0>,
+			      <&mux_in0>, "clock-noncontinuous?",
 			      <&ov5647_0>, "status=okay";
 		cam0-ov7251 = <&mux_in0>, "remote-endpoint:0=",<&ov7251_0_ep>,
 			      <&ov7251_0_ep>, "remote-endpoint:0=",<&mux_in0>,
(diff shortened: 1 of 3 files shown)


6by9:
Thanks for the patch.
I've never liked that the transmitter and receiver have to match in DT for these flags, as errors are bound to creep in. However that is the way that mainline media devs have decided it should work. (DRM devs have decided the opposite for DSI displays)

As you say, there is the get_mbus_config API call to allow the receiver to ask the transmitter what mode it is in, but very few sensor drivers implement it. I recall having the discussion over this when getting the TC358743 to work and setting clocking mode being dismissed - ah https://patchwork.kernel.org/project/linux-media/patch/20190316154801.20460-2-jacopo+renesas@jmondi.org/#22535719 appears to have gone nowhere.

get_mbus_config is implemented for video-mux as the different inputs can have different numbers of data lanes and different clocking modes, but you appear to be correct that the unicam driver isn't looking at the flags field for clock mode - that ought to get fixed. The camera-mux-2port and camera-mux-4port overlays also ought to get fixed for this missing `clock-noncontinuous` flag on ov5647 too.

I'm happy to take this as is, but if you fancied fixing up the camera-mux overlays too, then you'd have my gratitude.
I'll have to look again at get_mbus_config in unicam.

kleinweby:
Sorry for the late reply. Yeah the whole thing is somewhat confusing and redundant looking. 

> The camera-mux-2port and camera-mux-4port overlays also ought to get fixed for this missing clock-noncontinuous flag on ov5647 too.

Sure. I hope I deduced the correct solution, as I cannot test it.

6by9:
Thanks for the updates - it just saves me a little time.
I totally agree with the changes as they all match imx219, imx477, etc.

Now I just need to fix up unicam to abide by the updated clock mode....

pelwell:
Is this good to merge now?

6by9:
> Is this good to merge now?

As far as I'm concerned, yes.
Fixing up Unicam is my task.

PR #6122, opened 2024-04-23
drivers: media: cfe: Add remap entries for mono formats

naushir:
The 8-bit and 16-bit mono formats were missing the appropriate remap entries in the format table.


6by9:
I've tested with the 8bit line, and it fixes the ov9281 8 bit modes.

16bit now follows the Bayer 16 bit, so that looks reasonable to me.

naushir:
@tomba this is something to put into the next version of the upstream driver.

tomba:
> @tomba this is something to put into the next version of the upstream driver.

Alright, I have picked this up.

PR #6123, opened 2024-04-23
vc4/hdmi: Ignore hotplug interrupt with force_hotplug

popcornmix:
The intention of the vc4.force_hotplug setting is to ignore hotplug completely.

It can be used when a display switching AV inputs, or going into stanbdby or changing a KVM switch
toggles hotplug and some side effect of that is unwanted.

It turns out while vc4.force_hotplug currently makes hotplug always read as asserted, that isn't enough to stop drm doing lots of stuff, including re-reading the edid.

An example of what drm does with a hotplug deasert/assert and vc4.force_hotplug=1 currently is:

https://paste.debian.net/hidden/dc07434b/

That is unwanted. Lets ignore the hotplug interrupt completely so drm is blissfully unaware of the hotplug change.

### DIFF
--- a/drivers/gpu/drm/vc4/vc4_hdmi.c
+++ b/drivers/gpu/drm/vc4/vc4_hdmi.c
@@ -2910,7 +2910,7 @@ static irqreturn_t vc4_hdmi_hpd_irq_thread(int irq, void *priv)
 	struct drm_connector *connector = &vc4_hdmi->connector;
 	struct drm_device *dev = connector->dev;
 
-	if (dev && dev->registered)
+	if (dev && dev->registered && !force_hotplug)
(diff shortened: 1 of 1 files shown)


popcornmix:
Ping @6by9 @HiassofT

6by9:
Looks reasonable.

I was wondering whether it is better to not register the interrupt handler if force_hotplug was set, but I guess this way you can dynamically enable and disable force_hotplug via /sys/module/vc4/parameters/force_hotplug.

HiassofT:
Thanks, a few quick tests on RPi5 and RPi4 look fine, no more log messages with drm.debug=0x07 when hotplug is de- and re-asserted.

PR #6094, opened 2024-04-05
DSI0 colour swap fixes

6by9:
Patches so far trying to make DSI0 behave.

PR #6124, opened 2024-04-23
drm/bridge: tc358762: revert move ops to enable

6by9:
Reverts 8a4b2fc9c91a ("drm/bridge: tc358762: Split register programming from pre-enable to enable") as we want the config commands sent before video starts.

@lurch Please can you test this on your panel?


lurch:
This makes the panel functional on Pi 4, but unfortunately makes it non-functional on Pi 5.

6by9:
@lurch This is superseded by #6094.

That is tested with all Pi4, CM4 (both DSI0 & DSI1), CM3 (DSI0 & DSI1), and Pi5.
Pi5 is shifted one pixel horizontally, but that's another issue to fix later.

PR #6128, opened 2024-04-26
The ELECOM EDC-QUA3C is an RTL8156B based 2.5G Ethernet controller.

ose-support-ros:
Add the vendor and product ID values to the driver. This makes Ethernet work with the adapter.


popcornmix:
This file is from the upstream linux kernel and is currently untouched by this repo.
As such it's best to submit this change upstream. Once it is accepted we can pick it up here.

See [submitting patches](https://www.kernel.org/doc/html/v6.6/process/submitting-patches.html).

PR #6132, opened 2024-04-29
drm: rp1: dpi: Add support for MEDIA_BUS_FMT_RGB565_1X24_CPADHI

njhollinghurst:
This new format corresponds to the Raspberry Pi legacy DPI mode 3. This should enable it to work on Raspberry Pi 5.

I suppose I should test it before merging!

XXX we still don't have a way to get DPI mode 4. RGB666_1X24_CPADHI will put the bits in the right places, a named pinctrl is defined for Raspberry Pi 5 / RP1, but no pinctrl is defined for earlier Raspberry Pi models.

### DIFF
--- a/drivers/gpu/drm/rp1/rp1-dpi/rp1_dpi_hw.c
+++ b/drivers/gpu/drm/rp1/rp1-dpi/rp1_dpi_hw.c
@@ -319,8 +319,14 @@ static u32 set_output_format(u32 bus_format, u32 *shift, u32 *imask, u32 *rgbsz)
 		*shift |= OSHIFT_RGB(29, 19, 9);
 		return OMASK_RGB(0x3fc, 0x3fc, 0x3fc);
 
+	case MEDIA_BUS_FMT_RGB565_1X24_CPADHI:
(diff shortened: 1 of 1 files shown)


6by9:
LGTM

> XXX we still don't have a way to get DPI mode 4. RGB666_1X24_CPADHI will put the bits in the right places, a named pinctrl is defined for Raspberry Pi 5 / RP1, but no pinctrl is defined for earlier Raspberry Pi models.

I haven't heard anyone complain that mode 4 is missing from vc4 and rp1 kernel drivers. Selecting RGB666_1X24_CPADHI and amending the pinctrl is the way to do it, but it's not high on my priority list.

njhollinghurst:
A quick test using VGA666 board, `kmstest` and poking around with an oscilloscope suggests it's doing the expected things.

njhollinghurst:
Parting comment: Mode 4 is part of the MIPI DPI spec and leaves a suitable hole for I2S, so we should consider supporting it if we can, either by adding a media bus format (as was done for mode 3) or working around Linux in some other way.

PR #6133, opened 2024-04-29
Reinstate "output format" DRM property

6by9:
The "output format" property had been lost in switching to 6.6, and it appears was never implemented for Pi0-3 (only full/limited range RGB).

I want to double check the CCM for BT601, and ideally ought to add the CCM for BT709, but want the CI build for the reporter on https://forums.raspberrypi.com/viewtopic.php?t=369875


6by9:
Dropped the Pi0-3 patch and rebased. This may as well get merged. I'll create a new PR for the Pi0-3 bit.

Issue #6125, opened 2024-04-24
RP1 (Pi 5) DSI issue with 4 lanes


njhollinghurst:
### Describe the bug

Currently we make the DPI clock by dividing down the DSI byte clock. That is not going to work in cases where bpp < 8*lanes.

With the current driver it can result in DPI running too slowly, causing dropped scanlines and other misbehaviour of MIPI DSI.

If they must be clocked via different PLL and divider architectures, we may need to investigate what pixel-rate precision can be achieved while keeping DPI and DSI in effective lock-step; or to understand what happens when they're not quite synchronized (hopefully the worst that will happen is that HFP duration becomes slightly irregular)? In either case, some driver changes will be needed.

(leftover from PR 6120)

### Steps to reproduce the behaviour

Using `dtoverlay=vc4-kms-dsi-generic`, configure DSI output with four lanes and 24bpp.
`kmstest --flip` will show that the display is running at a lower frame rate than expected.
A DSI analyser may show other woes.

### Device (s)

Raspberry Pi 5

### System

rpi-6.6.y

### Additional context

cc @6by9

njhollinghurst:
Been doing some experiments with the 7" display (even though it uses only 1 lane)...

When DPI rate is imprecisely derived from the 200MHz core clock: the screen works but flickers, with a horizontal jitter.

DSI Host has the option to inhibit returning to LP mode during HBP, which may guarantee that the offset between HSYNC and data will be a fixed number of MIPI clocks.  Unfortunately that doesn't cure the jitter!*

There is a limited palette of pixel rates for which we can derive DPI rate from the core clock and DSI from the MIPI PLL, in an exact ratio. With one lane at 24bpp, 28.571Mpix/s is one such rate, and it displays perfectly.** With 4 lanes there would be about 25 such rates in the range 13.333Mpix/s to 160Mpix/s.

--
*I guess the bridge chip expects the line period to be uniform, or perhaps an exact multiple of its own re-generated (divided) DPI clock. Let's hope other displays are less fussy.

**it doesn't bear too much thinking about _how_ this displays without jitter, since the phase relationship between the two clocks is undefined.

njhollinghurst:
#6157 should fix this by allowing DPI clock to be derived from PLL_SYS where necessary.

Still no attempt has been made to force the pixel-rate to a more auspicious one to reduce jitter, since we have not tested with a 4-lane panel that would seem to need it.

njhollinghurst:
[...]

PR #6136, opened 2024-04-30
drm: rp1: rp1-dsi: Switch to PLL_SYS source for DPI when 8*lanes > bpp

njhollinghurst:
Draft which partially addresses #6125.

This sets up the DPI clock as accurately as it can to match the DSI rate; without altering HS/LP transitions or messing with the mode timings.

If 4-lane panels are as fussy as the bridge chip in our (1-lane) 7" display, there is a risk of horizontal jitter, due to irregular line intervals at most pixel-clock rates.

Not yet successfully tested with any real 4-lane panel.

### DIFF
--- a/drivers/gpu/drm/rp1/rp1-dsi/rp1_dsi.c
+++ b/drivers/gpu/drm/rp1/rp1-dsi/rp1_dsi.c
@@ -54,6 +54,7 @@ static void rp1_dsi_bridge_pre_enable(struct drm_bridge *bridge,
 	struct rp1_dsi *dsi = bridge_to_rp1_dsi(bridge);
 
 	rp1dsi_dsi_setup(dsi, &dsi->pipe.crtc.state->adjusted_mode);
+	dsi->dsi_running = true;
 }
 
 static void rp1_dsi_bridge_enable(struct drm_bridge *bridge,
@@ -443,7 +444,7 @@ static int rp1dsi_platform_probe(struct platform_device *pdev)
 	/* Hardware resources */
 	for (i = 0; i < RP1DSI_NUM_CLOCKS; i++) {
 		static const char * const myclocknames[RP1DSI_NUM_CLOCKS] = {
-			"cfgclk", "dpiclk", "byteclk", "refclk"
+			"cfgclk", "dpiclk", "byteclk", "refclk", "pllsys"
 		};
 		dsi->clocks[i] = devm_clk_get(dev, myclocknames[i]);
 		if (IS_ERR(dsi->clocks[i])) {
--- a/drivers/gpu/drm/rp1/rp1-dsi/rp1_dsi_dsi.c
+++ b/drivers/gpu/drm/rp1/rp1-dsi/rp1_dsi_dsi.c
@@ -1359,8 +1359,8 @@ static u32 get_colorcode(enum mipi_dsi_pixel_format fmt)
 #define RP1DSI_BYTE_CLK_MIN  10000000
 #define RP1DSI_BYTE_CLK_MAX 187500000
 #define RP1DSI_ESC_CLK_MAX   20000000
-#define RP1DSI_TO_CLK_DIV           5
-#define RP1DSI_LPRX_TO_VAL      0x400
+#define RP1DSI_TO_CLK_DIV        0x50
+#define RP1DSI_LPRX_TO_VAL       0x40
 #define RP1DSI_BTA_TO_VAL       0xd00
 
 void rp1dsi_dsi_setup(struct rp1_dsi *dsi, struct drm_display_mode const *mode)
@@ -1381,6 +1381,8 @@ void rp1dsi_dsi_setup(struct rp1_dsi *dsi, struct drm_display_mode const *mode)
 	vid_mode_cfg = 0xbf00;
 	if (!(dsi->display_flags & MIPI_DSI_MODE_VIDEO_SYNC_PULSE))
 		vid_mode_cfg |= 0x01;
+	else if (8 * dsi->lanes > bpp)
+		vid_mode_cfg &= ~0x400; /* PULSE && inexact DPICLK => fix HBP time */
 	if (dsi->display_flags & MIPI_DSI_MODE_VIDEO_BURST)
 		vid_mode_cfg |= 0x02;
 	DSI_WRITE(DSI_VID_MODE_CFG, vid_mode_cfg);
@@ -1390,9 +1392,10 @@ void rp1dsi_dsi_setup(struct rp1_dsi *dsi, struct drm_display_mode const *mode)
 	DSI_WRITE(DSI_MODE_CFG, 1);
 
 	/* Set timeouts and clock dividers */
-	DSI_WRITE(DSI_TO_CNT_CFG,
-		  (((bpp * mode->htotal) / (7 * RP1DSI_TO_CLK_DIV * dsi->lanes)) << 16) |
-		  RP1DSI_LPRX_TO_VAL);
+	timeout = (bpp * mode->htotal * mode->vdisplay) / (7 * RP1DSI_TO_CLK_DIV * dsi->lanes);
+	if (timeout > 0xFFFFu)
+		timeout = 0;
+	DSI_WRITE(DSI_TO_CNT_CFG, (timeout << 16) | RP1DSI_LPRX_TO_VAL);
 	DSI_WRITE(DSI_BTA_TO_CNT, RP1DSI_BTA_TO_VAL);
 	DSI_WRITE(DSI_CLKMGR_CFG,
 		  (RP1DSI_TO_CLK_DIV << 8) |
--- a/drivers/gpu/drm/rp1/rp1-dsi/rp1_dsi_dsi.c
+++ b/drivers/gpu/drm/rp1/rp1-dsi/rp1_dsi_dsi.c
@@ -1304,22 +1304,27 @@ static unsigned long rp1dsi_refclk_freq(struct rp1_dsi *dsi)
 static void rp1dsi_dpiclk_start(struct rp1_dsi *dsi, u32 byte_clock,
 				unsigned int bpp, unsigned int lanes)
 {
-	if (!dsi->clocks[RP1DSI_CLOCK_DPI])
-		return;
+	/* Dummy clk_set_rate() to declare the actual DSI byte-clock rate */
+	clk_set_rate(dsi->clocks[RP1DSI_CLOCK_BYTE], byte_clock);
 
-	if (bpp >= 8 * lanes && dsi->clocks[RP1DSI_CLOCK_BYTE] &&
-	    !clk_set_parent(dsi->clocks[RP1DSI_CLOCK_DPI], dsi->clocks[RP1DSI_CLOCK_BYTE]))
-		byte_clock = clk_get_rate(dsi->clocks[RP1DSI_CLOCK_BYTE]); /* dummy/nominal rate */
+	/*
+	 * Prefer the DSI byte-clock source where possible, so that DSI and DPI
+	 * clocks will be in an exact ratio and downstream devices can recover
+	 * perfect timings. But when DPI clock is faster, fall back on PLL_SYS.
+	 * To defeat rounding errors, specify explicitly which source to use.
+	 */
+	if (bpp >= 8 * lanes)
+		clk_set_parent(dsi->clocks[RP1DSI_CLOCK_DPI], dsi->clocks[RP1DSI_CLOCK_BYTE]);
 	else if (dsi->clocks[RP1DSI_CLOCK_PLLSYS])
 		clk_set_parent(dsi->clocks[RP1DSI_CLOCK_DPI], dsi->clocks[RP1DSI_CLOCK_PLLSYS]);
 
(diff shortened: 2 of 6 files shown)


drivers/gpu/drm/rp1/rp1-dsi/rp1_dsi_dsi.c
```diff
@@ -1124,19 +1124,23 @@ static uint8_t dphy_get_div(u32 refclk_khz, u32 vco_freq_khz, u32 *ptr_m, u32 *p
...
-	u32 best_n, best_m, best_err = 0x7fffffff;
-	unsigned int n;
+	static const u64 REF_DIVN_MAX = 40000000;
+	static const u64 REF_DIVN_MIN =  5000000;
+	unsigned int n, best_n, best_m;
+	u64 best_err = vco_freq;
 
-	for (n = 1 + refclk_khz / REF_DIVN_MAX; n * REF_DIVN_MIN <= refclk_khz && n < 100; ++n) {
-		u32 half_m = (n * vco_freq_khz + refclk_khz) / (2 * refclk_khz);
+	for (n = 1 + refclk / REF_DIVN_MAX; n * REF_DIVN_MIN <= refclk && n < 100; ++n) {
```
6by9, line 1139: Need to be careful to ensure this can't be built on a 32bit kernel as it'll error out on 64bit division. I'd hope a Kconfig change can protect against that.

drivers/gpu/drm/rp1/rp1-dsi/rp1_dsi.c
```diff
@@ -54,11 +54,18 @@ static void rp1_dsi_bridge_pre_enable(struct drm_bridge *bridge,
...
 
 static void rp1_dsi_bridge_enable(struct drm_bridge *bridge,
 				  struct drm_bridge_state *old_state)
 {
+	struct rp1_dsi *dsi = bridge_to_rp1_dsi(bridge);
+
+	if (!dsi->dsi_running) {
+		rp1dsi_dsi_setup(dsi, &dsi->pipe.crtc.state->adjusted_mode);
+		dsi->dsi_running = true;
+	}
```
6by9: How are you thinking you can get to `enable` without `pre_enable` being called?

drivers/gpu/drm/rp1/rp1-dsi/rp1_dsi_dsi.c
```diff
@@ -1290,23 +1300,25 @@ static unsigned long rp1dsi_refclk_freq(struct rp1_dsi *dsi)
...
-
-		clk_set_parent(dsi->clocks[RP1DSI_CLOCK_DPI], dsi->clocks[RP1DSI_CLOCK_BYTE]);
-		clk_set_rate(dsi->clocks[RP1DSI_CLOCK_DPI], (4 * lanes * u) / (bpp >> 1));
-		clk_prepare_enable(dsi->clocks[RP1DSI_CLOCK_DPI]);
-	}
+	if (!dsi->clocks[RP1DSI_CLOCK_DPI])
+		return;
+
+	if (bpp >= 8 * lanes && dsi->clocks[RP1DSI_CLOCK_BYTE] &&
+	    clk_set_parent(dsi->clocks[RP1DSI_CLOCK_DPI], dsi->clocks[RP1DSI_CLOCK_BYTE]) == 0)
```
6by9: `!clk_set_parent(...)`

drivers/gpu/drm/rp1/rp1-dsi/rp1_dsi_dsi.c
```diff
@@ -1290,23 +1300,25 @@ static unsigned long rp1dsi_refclk_freq(struct rp1_dsi *dsi)
...
+		byte_clock = clk_get_rate(dsi->clocks[RP1DSI_CLOCK_BYTE]); /* dummy/nominal rate */
+	else if (dsi->clocks[RP1DSI_CLOCK_PLLSYS])
+		clk_set_parent(dsi->clocks[RP1DSI_CLOCK_DPI], dsi->clocks[RP1DSI_CLOCK_PLLSYS]);
```
6by9, line 1319: You checked the return code above, but not here.

drivers/gpu/drm/rp1/rp1-dsi/rp1_dsi_dsi.c
```diff
@@ -1337,17 +1349,17 @@ static u32 get_colorcode(enum mipi_dsi_pixel_format fmt)
...
 #define RP1DSI_LPRX_TO_VAL      0x400
 #define RP1DSI_BTA_TO_VAL       0xd00
 
 void rp1dsi_dsi_setup(struct rp1_dsi *dsi, struct drm_display_mode const *mode)
 {
 	u32 timeout, mask, vid_mode_cfg;
-	int lane_kbps;
 	unsigned int bpp = mipi_dsi_pixel_format_to_bpp(dsi->display_format);
+	u32 byte_clock = clamp((bpp * 125u * (u32)min(mode->clock, 200000)) / dsi->lanes,
+			       10000000u, 187500000u);
```
6by9: Magic numbers,

6by9 (commented):
I'll trust you on the PLL setup. Largely coding style nits.

drivers/gpu/drm/rp1/rp1-dsi/rp1_dsi_dsi.c
njhollinghurst: Can a 32-bit kernel not do u64 division? I can probably re-work the arithmetic to avoid it but it would look messier.

drivers/gpu/drm/rp1/rp1-dsi/rp1_dsi_dsi.c
njhollinghurst: "not" sounds like failure. Is that the preferred style?

drivers/gpu/drm/rp1/rp1-dsi/rp1_dsi_dsi.c
njhollinghurst: Yes, not really sure what to do if they both fail, as it leaves us with an unknown clock parent (which might still work provided it has never been the dummy one).

drivers/gpu/drm/rp1/rp1-dsi/rp1_dsi_dsi.c
njhollinghurst: Max DPI pixel rate in kHz; min and max DSI rates in bytes/second. I can add them to the #defines above.

drivers/gpu/drm/rp1/rp1-dsi/rp1_dsi_dsi.c
6by9: It can, but not as a raw divide.

`do_div`, or you've got `DIV_ROUND_CLOSEST`, `DIV_ROUND_DOWN`, and `DIV_ROUND_UP` type macros (see include/linux/math.h)

pelwell: I'm sure it can - the compiler will generate calls to helper functions - provided at least one of the types is 64-bit.

drivers/gpu/drm/rp1/rp1-dsi/rp1_dsi_dsi.c
pelwell: Yes, sadly

drivers/gpu/drm/rp1/rp1-dsi/rp1_dsi_dsi.c
pelwell: The kernel build checks have all been successful.

drivers/gpu/drm/rp1/rp1-dsi/rp1_dsi_dsi.c
6by9: You're in functions that return void, so you cant actually return an error even if we could handle failing to set up the clock tree.

The normal preference is to use `!foo` rather than `foo == 0`, but it doesn't appear to be in https://www.kernel.org/doc/html/latest/process/coding-style.html

drivers/gpu/drm/rp1/rp1-dsi/rp1_dsi_dsi.c
6by9: Yes, we're in void returning functions, so you are pretty stuffed. I was more commenting on the lack of consistency.

Doing a `git grep clk_set_parent` I see a number of drivers not checking the return value at all, so it's probably OK not to.

drivers/gpu/drm/rp1/rp1-dsi/rp1_dsi_dsi.c
6by9: I would, just to make it clear.

drivers/gpu/drm/rp1/rp1-dsi/rp1_dsi_dsi.c
njhollinghurst: `div_u64()` looks like the one to use.

drivers/gpu/drm/rp1/rp1-dsi/rp1_dsi.c
njhollinghurst: Ah, I wasn't sure if it was mandatory. BTW it would be nice if pre-enable was called before any DSI commands got sent.

drivers/gpu/drm/rp1/rp1-dsi/rp1_dsi_dsi.c
6by9: CONFIG_DRM_RP1_DSI is only set in arch/arm64/configs/bcm2711_defconfig and arch/arm64/configs/bcm2712_defconfig, so no 32bit build is trying to compile it.

If it were to try, then it typically fails with errors such as 
```
ERROR: "__aeabi_ldivmod" [/home/pi/media_build/v4l/mn88436.ko] undefined!
ERROR: "__aeabi_uldivmod" [/home/pi/media_build/v4l/mn88436.ko] undefined!
```
Certainly for upstreaming you'll get the random config builds managing to try building it on 32bit and failing. Kconfig should block that.

My 32bit build + MFD_RP1 and DRM_RP1_DSI has just failed with 
```
ERROR: modpost: "msi_domain_set_affinity" [drivers/mfd/rp1.ko] undefined!
ERROR: modpost: "irq_domain_deactivate_irq" [drivers/mfd/rp1.ko] undefined!
ERROR: modpost: "irq_domain_activate_irq" [drivers/mfd/rp1.ko] undefined!
```
instead, so bigger fish to fry.

More an observation than requiring fixing immediately.

njhollinghurst:
> I'll trust you on the PLL setup. Largely coding style nits.

The "bogus fixed clock" is also something I really ought to fix -- probably by making a near clone of `clk_fixed`  whose rate can notionally be changed, since it's really not a RP1-specific thing -- but I'll try to get everything functional first.

drivers/gpu/drm/rp1/rp1-dsi/rp1_dsi_dsi.c
pelwell: (Largely to myself)
Apparently this is because the kernel doesn't use libgcc and doesn't include its own versions of these functions. The functions declared in math64.h should be used instead.

njhollinghurst:
Hmm, doing the clocks "correctly" is not that simple. The current dummy frequency of 72kHz was chosen to be highly divisible; this is not true of an an actual MIPI VCO rate rounded to integer Hz. And once rounding errors creep in we might not be able to guarantee that the divider will be the precise integer or simple fraction that we wanted it to be, or that `clk-rp1` will consistently favour an exact ratio over an inexact one.

njhollinghurst:
I've left the clocking changes unsquashed, for clarity. There are changes in several places which kinda need to be atomic.
@naushir do you have any comments on the clk-rp1 changes?

naushir:
> @naushir do you have any comments on the clk-rp1 changes?

clk-rp1 changes look reasonable to me.  Shame there is not an existing clock type like this in the framework.

drivers/gpu/drm/rp1/rp1-dsi/rp1_dsi_dsi.c
njhollinghurst: Now checking neither of them.

drivers/gpu/drm/rp1/rp1-dsi/rp1_dsi_dsi.c
njhollinghurst: I think the u64 divides are all fixed up now. refclk is back to being a u32.

drivers/gpu/drm/rp1/rp1-dsi/rp1_dsi.c
njhollinghurst: Gone.

drivers/gpu/drm/rp1/rp1-dsi/rp1_dsi_dsi.c
njhollinghurst: Done.

njhollinghurst:
Replaced by #6157

PR #6137, opened 2024-05-01
overlays: qca7000: Adjust URL & README info

lategoodbye:
in-tech smart charging is now called chargebyte. Our product name seems to be more stable than our company name, so drop them where it's possible.

https://chargebyte.com/articles/in-tech-smart-charging-wird-zu-chargebyte


pelwell:
> Our product name seems to be more stable than our company name

That got a smile...

PR #6140, opened 2024-05-01
drivers: media: pisp_be: Remove extras field from pisp_be_config

naushir:
Remove the various extras fields from struct pisp_be_config. These fields are not used by the driver nor the hardware, so do not belong in this structure.

Also put the tiles array at the top of struct pisp_be_tiles_config to ensure any size change to struct pisp_be_config does not move the tiles struct out of alignment.

This is a userland breaking change, and libpisp must be updated accordingly.

### DIFF
--- a/drivers/media/platform/raspberrypi/pisp_be/pisp_be.c
+++ b/drivers/media/platform/raspberrypi/pisp_be/pisp_be.c
@@ -312,13 +312,10 @@ static void hw_queue_job(struct pispbe_dev *pispbe,
 	write_reg(pispbe, PISP_BE_GLOBAL_BAYER_ENABLE_OFFSET, hw_enables[0]);
 	write_reg(pispbe, PISP_BE_GLOBAL_RGB_ENABLE_OFFSET, hw_enables[1]);
 
-	/*
-	 * Everything else is as supplied by the user. XXX Buffer sizes not
-	 * checked!
-	 */
+	/* Everything else is as supplied by the user. */
 	begin =	offsetof(struct pisp_be_config, global.bayer_order) /
 								sizeof(u32);
-	end = offsetof(struct pisp_be_config, axi) / sizeof(u32);
+	end = sizeof(struct pisp_be_config) / sizeof(u32);
 	for (u = begin; u < end; u++) {
 		unsigned int val = ((u32 *)config)[u];
 
--- a/drivers/media/platform/raspberrypi/pisp_be/pisp_be_config.h
+++ b/drivers/media/platform/raspberrypi/pisp_be/pisp_be_config.h
@@ -398,15 +398,6 @@ struct pisp_be_hog_buffer_config {
 };
 
 struct pisp_be_config {
-	/* I/O configuration: */
-	struct pisp_be_input_buffer_config input_buffer;
-	struct pisp_be_tdn_input_buffer_config tdn_input_buffer;
-	struct pisp_be_stitch_input_buffer_config stitch_input_buffer;
-	struct pisp_be_tdn_output_buffer_config tdn_output_buffer;
-	struct pisp_be_stitch_output_buffer_config stitch_output_buffer;
-	struct pisp_be_output_buffer_config
-				output_buffer[PISP_BACK_END_NUM_OUTPUTS];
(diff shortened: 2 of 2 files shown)


naushir:
This will break userland, so best to merge this in together with the (soon to be) upstream changes.

naushir:
@davidplowman and @njhollinghurst we won't merge this yet, but just a heads-up.

naushir:
@jmondi this is something we should put into the next revision of your upstream driver.

njhollinghurst:
I'm definitely a bit nervous about the driver/UAPI change....

But while we're at it why don't we remove the I/O addresses (buffer configs) as well, and start from GLOBAL_BAYER_ENABLE? The driver will never touch those address fields (instead it uses `dma_addr_t addrs[]`). I think the "test HAL" does use them for index numbers for TDN/Stitch buffer cycling, but I'm sure we could come up with a different and no worse hack.

naushir:
> But while we're at it why don't we remove the I/O addresses (buffer configs) as well, and start from GLOBAL_BAYER_ENABLE? The driver will never touch those address fields (instead it uses `dma_addr_t addrs[]`). I think the "test HAL" does use them for index numbers for TDN/Stitch buffer cycling, but I'm sure we could come up with a different and no worse hack.

Done!

jmondi:
> @jmondi this is something we should put into the next revision of your upstream driver.

I'll take this in and take the occasion to send a new version.

If mainline does not provide any additional feedback/comments on the driver, I'll send a pull request and that's it, the series has been there for quite some time. We're late already for the 6.10 merge window, but as soon as v6.10-rc1 is out, I'll make sure the driver is collected

drivers/media/platform/raspberrypi/pisp_be/pisp_be.c
```diff
@@ -312,14 +312,9 @@ static void hw_queue_job(struct pispbe_dev *pispbe,
 	write_reg(pispbe, PISP_BE_GLOBAL_BAYER_ENABLE_OFFSET, hw_enables[0]);
 	write_reg(pispbe, PISP_BE_GLOBAL_RGB_ENABLE_OFFSET, hw_enables[1]);
 
-	/*
-	 * Everything else is as supplied by the user. XXX Buffer sizes not
-	 * checked!
-	 */
-	begin =	offsetof(struct pisp_be_config, global.bayer_order) /
-								sizeof(u32);
-	end = offsetof(struct pisp_be_config, axi) / sizeof(u32);
```
jmondi, line 321: The previous version used as `begin` address the offset of `global.bayer_order`.

The `pisp_be_global_config` structure is defined as

```
struct pisp_be_global_config {
	u32 bayer_enables;
	u32 rgb_enables;
	u8 bayer_order;
	u8 pad[3];
} __attribute__((packed));
```

So the previous version was skipping the first two fields.

The new version writes from `struct pisp_be_config[0]` which instead includes the two fields. Am I reading this wrong ?

naushir: Good catch.  It's actually a bit inconsequential.  The start of the config register range is ``global.bayer_enables``, but in the old code, we skip writing to ``global.bayer_enables`` and ``global.rgb_enables`` in the bulk programming and do it with explicit single register writes.  The reason for this is so that the driver can selectively disable blocks if there was a problem with the block config or buffers were not correctly provided.  However, this really should not happen.

I'll revert it back to the old way just to be completely safe.

jmondi: > Good catch. It's actually a bit inconsequential. The start of the config register range is `global.bayer_enables`, but in the old code, we skip writing to `global.bayer_enables` and `global.rgb_enables` in the bulk programming and do it with explicit single register writes. The reason for this is so that the driver can selectively disable blocks if there was a problem with the block config or buffers were not correctly provided. However, this really should not happen.
> 
> I'll revert it back to the old way just to be completely safe.

As long as it was intentional, it's fine.

Up to you, let me know if I should collect this last version

naushir: I've reverted back to skipping these regs in the latest version.  Let's use that for the upstream driver as well.

PR #6143, opened 2024-05-02
LT070ME050000 DSI panel fixes

6by9:
@njhollinghurst required to just clean up the handling of that panel.


njhollinghurst:
Thanks. I've still not got the backlight to come on from Pi 5... although it is supposedly sending DSI LP escape commands and they do seem to be getting onto the wire (which no longer has the funny voltage droop).

Issue #6155, opened 2024-05-10
Missing Color Order in KMS DPI Overlay


othermod:
### Describe the bug

The vc4-kms-dpi-generic-overlay.dts overlay is missing some capability that previously existed for changing the color order of DPI output. It gives only bgr888 and rgb888 as options. I developed a product using red instead of green in the middle group, and I'm currently unable to use the KMS DPI overlay.

The previous DPI method gave 4 color orders for RGB
rgb_order:
   1: DPI_RGB_ORDER_RGB
   2: DPI_RGB_ORDER_BGR
   3: DPI_RGB_ORDER_GRB
   4: DPI_RGB_ORDER_BRG

It is possible to bring back this capability that previously existed?

### Steps to reproduce the behaviour

Use the KMS DPI overlay on a product (such as my PSPi 6) that makes use of the previously available DPI color orders that use red in the center group instead of green.

### Device (s)

Raspberry Pi CM4

### System

Raspberry Pi OS Bookworm

6by9:
Possible but painful.

All the formats are represented by MEDIA_BUS_FMT_xxx defines in the kernel from https://github.com/torvalds/linux/blob/master/include/uapi/linux/media-bus-format.h#L37.
To add GRB and BRG means new defines for each bit depth of interest, so potentially 10 new formats when you consider all the supported modes.

As it's a set of defines shared with mainline Linux, we either have to upstream the defines (a moderate pain), or somehow handle the potential for downstream to have extra defines that aren't in mainline without creating merge conflicts(*) and account for what happens should someone actually upstream the formats properly.

I'm tempted to do it as a bit of a hack via a DT override on the dpi driver node that tells it to ignore the component order from the MEDIA_BUS_FMT and adopt the specified value. Doing so won't be acceptable to mainline though.

@pelwell what's your feeling? https://github.com/raspberrypi/linux/pull/6156 is a quick patch for the override option.

(*) Probably take some values like 0x1f00 - 0x1f0a which won't be used for many years.

othermod:
Thank you. I appreciate anything you are willing to do. I tried working this out on my own without luck.
The KMS overlay also appears to lack the ability to disable HSYNC and VSYNC (using DE mode on an LCD, and using gpio2 and gpio3 for I2C). I got around this by just setting the pins manually after bootup. I can create a separate issue later on if it becomes a problem. Just wanted to bring it up while you're considering options.

6by9:
[...]

PR #6156, opened 2024-05-10
RGB order override for DPI

6by9:
Quick (untested) mod for https://github.com/raspberrypi/linux/issues/6155

It's either this or adding a load of extra MEDIA_BUS_FMTs.

### DIFF
--- a/arch/arm/boot/dts/overlays/README
+++ b/arch/arm/boot/dts/overlays/README
@@ -4988,6 +4988,10 @@ Params: clock-frequency         Display clock frequency (Hz)
                                 Set the default brightness. Normal range 1-16.
                                 (default 16).
         rotate                  Display rotation {0,90,180,270} (default 0)
+        rgb-order               Allow override of RGB order from DPI.
+                                Options for vc4 are "rgb", "bgr", "grb", and
+                                "brg". Other values will be ignored.
+
 
 
 Name:   vc4-kms-dpi-hyperpixel2r
--- a/drivers/gpu/drm/vc4/vc4_dpi.c
+++ b/drivers/gpu/drm/vc4/vc4_dpi.c
@@ -95,6 +95,8 @@ struct vc4_dpi {
 	struct clk *core_clock;
 
 	struct debugfs_regset32 regset;
+
+	int rgb_order_override;
 };
 
 #define to_vc4_dpi(_encoder)						\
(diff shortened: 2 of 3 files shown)


njhollinghurst:
Will we be wanting this on RP1 as well?  (BTW the latter can also generate several other pinouts that would leave a gap for I2S, but I expect VC4 can't match those. I mention it in case it helps forward planning...)

drivers/gpu/drm/vc4/vc4_dpi.c
```diff
@@ -205,6 +207,11 @@ static void vc4_dpi_encoder_enable(struct drm_encoder *encoder)
 			}
 		}
 
+		if (dpi->rgb_order_override) {
```
pelwell: Are we happy that it's never necessary to force DPI_ORDER_RGB?

6by9: All the RGB order formats are available via MEDIA_BUS_FMT defines, hence I saw no reason to override it.
I did think about using -1 as "not overridden", but it seemed overkill.

6by9:
> Will we be wanting this on RP1 as well? (BTW the latter can also generate several other pinouts that would leave a gap for I2S, but I expect VC4 can't match those. I mention it in case it helps forward planning...)

Yes we'll probably want something similar on RP1.
The `dpi` DT node should be remapped to the RP1 peripheral on Pi5 so the property would exist with the overlay change, but the driver would obviously need to look for it. Handling the values in a generic manner might want a little more thought.

arch/arm/boot/dts/overlays/README
```diff
@@ -4988,6 +4988,12 @@ Params: clock-frequency         Display clock frequency (Hz)
                                 Set the default brightness. Normal range 1-16.
                                 (default 16).
         rotate                  Display rotation {0,90,180,270} (default 0)
+        rgb-order               Allow override of RGB order from DPI.
+                                0 = obey MEDIA_BUS_FMT (default)
+                                1 = BGR
+                                2 = GRB
+                                3 = BRG
+
```
njhollinghurst, line 4994: There are 6 permutations including GBR, RBG; I guess those are never wanted?

6by9: Those permutations are not available on vc4.

https://github.com/raspberrypi/linux/blob/rpi-6.6.y/drivers/gpu/drm/vc4/vc4_dpi.c#L33-L41

njhollinghurst: Ah, fair enough!

Meanwhile I just remembered RP1 would have a problem with all of these for "mode 2" (which really only works when the in-memory fb format is exactly RGB565) but that can't be helped.

6by9:
Updated and now takes a string for the RGB order so that all orders can be handled in the RP1 driver.

I think this handles the use case fairly cleanly, so dropping the draft status.
(I know I should have updated the DT bindings too, but I doubt this will ever be acceptable to upstream anyway)

drivers/gpu/drm/vc4/vc4_dpi.c
```diff
@@ -361,6 +369,20 @@ static int vc4_dpi_bind(struct device *dev, struct device *master, void *data)
 	if (ret)
 		return ret;
 
+	dpi->rgb_order_override  = -1;
```
pelwell: There's a stray extra space on this line.

6by9: Fixed.

PR #6157, opened 2024-05-10
rp1-dsi: Switch to PLL_SYS source for DPI when 8*lanes > bpp

njhollinghurst:
Replaces #6136 after rebase to rpi-6.6.y and with the changes grouped into what I hope is a sensible order.

This ought in principle to fix the clock rate bug with 4-lane DSI on RP1. It does not break the (single-lane) Raspberry Pi 7" Display.

However, I still cannot get a picture (or even a backlight) out of LT070ME05000.

### DIFF
--- a/arch/arm/boot/dts/broadcom/rp1.dtsi
+++ b/arch/arm/boot/dts/broadcom/rp1.dtsi
@@ -1109,16 +1109,15 @@
 
 			interrupts = <RP1_INT_MIPI0 IRQ_TYPE_LEVEL_HIGH>;
 
-			clocks = <&rp1_clocks RP1_CLK_MIPI0_CFG>,  // required, config bus clock
-				 <&rp1_clocks RP1_CLK_MIPI0_DPI>,  // required, pixel clock
-				 <&clksrc_mipi0_dsi_byteclk>,    // internal, parent for divide
-				 <&clk_xosc>;                    // hardwired to DSI "refclk"
-			clock-names = "cfgclk", "dpiclk", "byteclk", "refclk";
+			clocks = <&rp1_clocks RP1_CLK_MIPI0_CFG>,
+				 <&rp1_clocks RP1_CLK_MIPI0_DPI>,
+				 <&rp1_clocks RP1_CLK_MIPI0_DSI_BYTECLOCK>,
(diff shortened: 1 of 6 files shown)


6by9:
These patches fix 4 lane operation for me.
I'll trust njh over the behaviour, but it looks sane.

Branch rebased.

Any objections to merging?

njhollinghurst:
No objection from me. I'd been delaying because I'd hoped to test it with an actual display (rather than only with an analyser or scope), but LT070ME05000 is not playing ball. I think this change is shown to be an improvement anyway, or at least makes nothing worse.

arch/arm/boot/dts/broadcom/rp1.dtsi
```diff
@@ -1109,16 +1109,17 @@
...
+				 <&rp1_clocks RP1_CLK_MIPI0_DSI_BYTECLOCK>,
+				 <&clk_xosc>,                // hardwired to DSI "refclk"
+				 <&rp1_clocks RP1_PLL_SYS>;  // alternate parent for divide
+			clock-names = "cfgclk", "dpiclk", "byteclk", "refclk", "pllsys";
 
 			assigned-clocks = <&rp1_clocks RP1_CLK_MIPI0_CFG>,
-					  <&rp1_clocks RP1_CLK_MIPI0_DPI>;
+					  <&rp1_clocks RP1_CLK_MIPI0_DPI>,
+					  <&rp1_clocks RP1_CLK_MIPI0_DSI_BYTECLOCK>;
 			assigned-clock-rates = <25000000>;
```
pelwell, line 1120: What are the semantics of having fewer assigned-clock-rates than assigned-clocks. Or, to put it another way, what is the effect of assigning a clock without giving it a rate?

njhollinghurst: Going by the documentation, it probably does nothing. (I imagined it might give the driver some kind of exclusive right to set the rate and parent at run-time, but can find no evidence for this). I'd better try it with this removed.

njhollinghurst: Yes it seems to work just as well without the last two assigned clocks. I will remove them and force the change into the penultimate commit.

PR #6163, opened 2024-05-14
configs: Enable EROFS support as a module

ragazenta:
EROFS filesystem stands for Enhanced Read-Only File System. It aims to form a generic read-only filesystem solution for various read-only use cases instead of just focusing on storage space saving without considering any side effects of runtime performance.

> ... with guaranteed end-to-end performance by using compact layout, transparent file compression and direct access, especially for those embedded devices with limited memory and high-density hosts with numerous containers. [(1)](https://docs.kernel.org/filesystems/erofs.html)

Wouldn't it be a good idea to enable this? Thanks.


kaysievers:
That would be nice.

greyltc:
+1. Is there any reason EROFS shouldn't be turned on here?

pelwell:
Superseded by https://github.com/raspberrypi/linux/pull/6806.

PR #6166, opened 2024-05-14
DRAFT: drm: rp1: rp1-vec: Add a parameter to set DRM picture aspect ratio

njhollinghurst:
Experimentally, allow setting picture_aspect_ratio field in all advertised DRM modes. This has no other effect on VEC's behaviour (and perhaps has no useful effects at all...)

It is set via a module parameter `drm-rp1-vec.par` which can be set from the command line, or later using sysfs. A value of 1 means 4:3, 2 means 16:9.

### DIFF
--- a/drivers/gpu/drm/rp1/rp1-vec/rp1_vec.c
+++ b/drivers/gpu/drm/rp1/rp1-vec/rp1_vec.c
@@ -61,6 +61,11 @@ module_param_named(cmode, rp1vec_cmode_str, charp, 0600);
 MODULE_PARM_DESC(cmode, "Custom video mode:\n"
 		 "\t\t<pclk>,<hact>,<hfp>,<hsync>,<hbp>,<vact>,<vfp>,<vsync>,<vbp>[,i]\n");
 
+int par;
+module_param_named(par, par, int, 0600);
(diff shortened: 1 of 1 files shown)


drivers/gpu/drm/rp1/rp1-vec/rp1_vec.c
```diff
@@ -362,11 +368,13 @@ static int rp1vec_connector_get_modes(struct drm_connector *connector)
...
 				mode->type |= DRM_MODE_TYPE_PREFERRED;
 			}
+			mode->picture_aspect_ratio = par;
 			drm_mode_set_name(mode);
+
 			drm_mode_probed_add(connector, mode);
 			n++;
 
-			if (mode->vtotal == 405 || mode->vtotal == 819)
+			if (i >= 4)
```
6by9, line 377: This 4 is creeping in as a magic number both here and line 355. It'd be nice if it came from a define in and around the rp1vec_modes array.

6by9 (commented):
Can we add validation that the module parameter is a valid `enum hdmi_picture_aspect`? https://elixir.bootlin.com/linux/latest/source/include/linux/hdmi.h#L96

njhollinghurst:
I think I will close this due to lack of interest (and the risk that it would get confused with WSS, which RP1 doesn't support).

PR #6168, opened 2024-05-15
configs: Remove obsolete ReiserFS support

ragazenta:
ReiserFS is marked as obsolete in 6.6 kernel, I propose it should be disabled. Thanks.

See https://www.phoronix.com/news/ReiserFS-Obsolete.


pelwell:
I'm reluctant to remove an option like this within a kernel version that we are already shipping - we have no way of knowing how many people may be using it. Apart from taking up space in an image, it is doing no harm.

As I understand it, ReiserFS is scheduled for removal in 2025. If that is after the next LTS, i.e. our next production kernel, we'll have to make a decision whether to drop it early, but at the moment I'm feeling no incentive to do so.

PR #6169, opened 2024-05-15
media: rp1: csi2: Fix set_fmt for the source

tomba:
The driver allows changing the source pad's format to the 16-bit or the compressed version, but doesn't allow changing it back to the sink's format. A work-around for the user is to set the format to the sink side, which then gets propagated correctly.

Fix the driver to allow changing the source format back to the sink's format.


tomba:
@naushir jfyi

tomba:
Hmm, I think we have the same issue in the FE. I'll refresh the series.

PR #6174, opened 2024-05-16
dtoverlays: Fix noints mode of mcp23017

6by9:
noints mode disables 2 fragments that configure a GPIO to be used for the interrupt line from the MCP23017, but fails to remove the pinctrl-0 reference or pinctrl-names. It therefore fails to load due to an invalid phandle.

Move the pinctrl-0 and pinctrl-names properties so they also get disabled by the noints override.

https://forums.raspberrypi.com/viewtopic.php?t=370907

Draft awaiting confirmation from OP that it works.


6by9:
Found my MCP23017 and verified that it loads with the `noints` override, although mine is on i2c-1 instead of i2c-10.

PR #6176, opened 2024-05-17
Minor panel fixups

6by9:
Working around a quirk of the panel on Pi1-3.

Shutting down cleanly.


6by9:
Shutdown hook is stopping shutdown - investigating.

6by9:
Doh, helps to call `i2c_set_clientdata` if you want to use `i2c_get_clientdata`.
Fixed.

PR #6177, opened 2024-05-17
Add association between DRM panels and backlight where defined

6by9:
For comment from SimonL for UI backlight control.

For ease of testing this also includes the patches from 6176 as it's useful to have the panel fully working.


6by9:
@pelwell @popcornmix
Simon has written a widget to control the backlight based on this. Are we happy with the core mods as downstream patches? I don't think we'll get much traction in upstreaming this sort of thing.

PR #6181, opened 2024-05-21
mfd: Add a driver for the RPI RP2040 GPIO bridge

roliver-rpi:
The Raspberry Pi RP2040 GPIO bridge is an I2C-attached MFD exposing both a Tx-only SPI controller, and a GPIO controller.

Due to the relative difference in transfer rates between standard-mode I2C and SPI, the GPIO bridge makes use of 12 MiB of non-volatile storage to cache repeated transfers. This cache is arranged in ~8 KiB blocks and is addressed by the MD5 digest of the data contained therein.

Optionally, this driver is able to take advantage of Raspberry Pi RP1 GPIOs to achieve faster than I2C data transfer rates.

### DIFF
--- a/drivers/spi/spi-rp2040-gpio-bridge.c
+++ b/drivers/spi/spi-rp2040-gpio-bridge.c
@@ -1,0 +1,1219 @@
+// SPDX-License-Identifier: GPL-2.0+
+/*
+ * RP2040 GPIO Bridge
+ *
+ * Copyright (C) 2023, 2024, Raspberry Pi Ltd
+ */
+
+#include <crypto/hash.h>
+#include <linux/crypto.h>
+#include <linux/delay.h>
+#include <linux/firmware.h>
+#include <linux/gpio/driver.h>
+#include <linux/i2c.h>
+#include <linux/kernel.h>
+#include <linux/minmax.h>
+#include <linux/of_address.h>
+#include <linux/pm_runtime.h>
+#include <linux/spi/spi.h>
+#include <linux/stddef.h>
+#include <linux/types.h>
+
+#define MODULE_NAME "rp2040-gpio-bridge"
+
+#define I2C_RETRIES 4U
+
+#define ONE_KIB 1024U
+#define MD5_SUFFIX_SIZE 9U
+
+#define RP2040_GBDG_FLASH_BLOCK_SIZE (8U * ONE_KIB)
+#define RP2040_GBDG_BLOCK_SIZE (RP2040_GBDG_FLASH_BLOCK_SIZE - MD5_SUFFIX_SIZE)
+
+/*
+ * 1MiB transfer size is an arbitrary limit
+ * Max value is 4173330 (using a single manifest)
+ */
+#define MAX_TRANSFER_SIZE (1024U * ONE_KIB)
+
+#define HALF_BUFFER (4U * ONE_KIB)
+
+#define STATUS_SIZE 4
+#define MD5_DIGEST_SIZE 16
+#define VERSION_SIZE 4
+#define ID_SIZE 8
+#define TOTAL_RD_HDR_SIZE \
+	(STATUS_SIZE + MD5_DIGEST_SIZE + VERSION_SIZE + ID_SIZE)
+
+struct rp2040_gbdg_device_info {
+	u8 md5[MD5_DIGEST_SIZE];
+	u64 id;
+	u32 version;
+	u32 status;
+};
+
+static_assert(sizeof(struct rp2040_gbdg_device_info) == TOTAL_RD_HDR_SIZE);
+
+#define MANIFEST_UNIT_SIZE 16
+static_assert(MD5_DIGEST_SIZE == MANIFEST_UNIT_SIZE);
+#define MANIFEST_HEADER_UNITS 1
+#define MANIFEST_DATA_UNITS \
+	DIV_ROUND_UP(MAX_TRANSFER_SIZE, RP2040_GBDG_BLOCK_SIZE)
+
+#define STATUS_BUSY 0x01
+
+#define DIRECT_PREFIX 0x00
+#define DIRECT_CMD_CS 0x07
+#define DIRECT_CMD_EMIT 0x08
+
+#define WRITE_DATA_PREFIX 0x80
+#define WRITE_DATA_PREFIX_SIZE 1
+
+#define FIXED_SIZE_CMD_PREFIX 0x81
+
+#define WRITE_DATA_UPPER_PREFIX 0x82
+#define WRITE_DATA_UPPER_PREFIX_SIZE 1
+
+#define NUM_GPIO 24
+
+enum rp2040_gbdg_fixed_size_commands {
+	/* 10-byte commands */
+	CMD_SAVE_CACHE = 0x07,
+	CMD_SEND_RB = 0x08,
+	CMD_GPIO_ST_CL = 0x0b,
+	CMD_GPIO_OE = 0x0c,
+	CMD_DAT_RECV = 0x0d,
+	CMD_DAT_EMIT = 0x0e,
+	/* 18-byte commands */
+	CMD_READ_CSUM = 0x11,
+	CMD_SEND_MANI = 0x13,
+};
+
+struct rp2040_gbdg {
+	struct spi_controller *controller;
+
+	struct i2c_client *client;
+	struct crypto_shash *shash;
+	struct shash_desc *shash_desc;
+
+	struct regulator *regulator;
+
+	struct gpio_chip gc;
+	u32 gpio_requested;
+	u32 gpio_direction;
+
+	bool fast_xfer_requires_i2c_lock;
+	struct gpio_descs *fast_xfer_gpios;
+	u32 fast_xfer_recv_gpio_base;
+	u8 fast_xfer_data_index;
+	u8 fast_xfer_clock_index;
+	void __iomem *gpio_base;
+	void __iomem *rio_base;
+
+	bool bypass_cache;
+
+	u8 buffer[2 + HALF_BUFFER];
+	u8 manifest_prep[(MANIFEST_HEADER_UNITS + MANIFEST_DATA_UNITS) *
+			 MANIFEST_UNIT_SIZE];
+};
+
+static int rp2040_gbdg_gpio_dir_in(struct gpio_chip *gc, unsigned int offset);
+static void rp2040_gbdg_gpio_set(struct gpio_chip *gc, unsigned int offset,
+				 int value);
+static int rp2040_gbdg_fast_xfer(struct rp2040_gbdg *priv_data, const u8 *data,
+				 size_t len);
+
+static int rp2040_gbdg_rp1_calc_offsets(u8 gpio, size_t *bank_offset,
+					u8 *shift_offset)
+{
+	if (!bank_offset || !shift_offset || gpio >= 54)
+		return -EINVAL;
+	if (gpio < 28) {
+		*bank_offset = 0x0000;
+		*shift_offset = gpio;
+	} else if (gpio < 34) {
+		*bank_offset = 0x4000;
+		*shift_offset = gpio - 28;
+	} else {
+		*bank_offset = 0x8000;
+		*shift_offset = gpio - 34;
+	}
+
+	return 0;
+}
+
+static int rp2040_gbdg_calc_mux_offset(u8 gpio, size_t *offset)
+{
+	size_t bank_offset;
+	u8 shift_offset;
+	int ret;
+
+	ret = rp2040_gbdg_rp1_calc_offsets(gpio, &bank_offset, &shift_offset);
+	if (ret)
+		return ret;
+	*offset = bank_offset + shift_offset * 8 + 0x4;
+
+	return 0;
+}
+
+static int rp2040_gbdg_rp1_read_mux(struct rp2040_gbdg *priv_data, u8 gpio,
+				    u32 *data)
+{
+	size_t offset;
+	int ret;
+
+	ret = rp2040_gbdg_calc_mux_offset(gpio, &offset);
+	if (ret)
+		return ret;
+
+	*data = readl(priv_data->gpio_base + offset);
+
+	return 0;
+}
+
+static int rp2040_gbdg_rp1_write_mux(struct rp2040_gbdg *priv_data, u8 gpio,
+				     u32 val)
+{
+	size_t offset;
+	int ret;
+
+	ret = rp2040_gbdg_calc_mux_offset(gpio, &offset);
+	if (ret)
+		return ret;
+
+	writel(val, priv_data->gpio_base + offset);
+
+	return 0;
+}
+
+static size_t rp2040_gbdg_max_transfer_size(struct spi_device *spi)
+{
+	return MAX_TRANSFER_SIZE;
+}
+
+static int rp2040_gbdg_get_device_info(struct i2c_client *client,
+				       struct rp2040_gbdg_device_info *info)
+{
+	u8 buf[TOTAL_RD_HDR_SIZE];
+	u8 retries = I2C_RETRIES;
+	u8 *read_pos = buf;
+	size_t field_size;
+	int ret;
+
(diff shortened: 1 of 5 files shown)


drivers/mfd/rpi-rp2040-gpio-bridge.c
```diff
@@ -0,0 +1,1155 @@
...
+	priv_data->shash_desc->tfm = priv_data->shash;
+
+	controller->bus_num = np ? -1 : client->adapter->nr;
+	controller->num_chipselect = 1;
+	controller->mode_bits = SPI_CPOL | SPI_CPHA;
+	controller->bits_per_word_mask = SPI_BPW_MASK(8);
+	controller->min_speed_hz = 35000000;
+	controller->max_speed_hz = 35000000;
+	controller->max_transfer_size = rp2040_gbdg_max_transfer_size;
+	controller->max_message_size = rp2040_gbdg_max_transfer_size;
```
6by9: Seeing as `rp2040_gbdg_max_transfer_size` just returns `MAX_TRANSFER_SIZE`, do we need a function for it?

drivers/mfd/rpi-rp2040-gpio-bridge.c
```diff
@@ -0,0 +1,1155 @@
...
+
+	if (!priv_data->shash_desc) {
+		crypto_free_shash(priv_data->shash);
+		return dev_err_probe(
+			dev, ENOMEM,
+			"Could not allocate memory for shash_desc\n");
+	}
+	priv_data->shash_desc->tfm = priv_data->shash;
+
+	controller->bus_num = np ? -1 : client->adapter->nr;
```
6by9: Why is this conditional on `np` (aka `dev->of_node`)? It's only accessing `client->adapter`

drivers/mfd/rpi-rp2040-gpio-bridge.c
```diff
@@ -0,0 +1,1155 @@
...
+		priv_data->fast_xfer_gpios = NULL;
+		goto cleanup;
+	}
+
+cleanup:
+	if (of_args[0].np)
+		of_node_put(of_args[0].np);
+	if (of_args[1].np)
+		of_node_put(of_args[1].np);
+	return 0;
```
6by9: So this is a success path? `cleanup` implied to me at least that it was an error path. `node_put` or similar would be more obvious.

drivers/mfd/rpi-rp2040-gpio-bridge.c
```diff
@@ -0,0 +1,1155 @@
...
+
+	i2c_set_clientdata(client, priv_data);
+
+	priv_data->shash = crypto_alloc_shash("md5", 0, 0);
+	if (IS_ERR(priv_data->shash))
+		return dev_err_probe(dev, PTR_ERR(priv_data->shash),
+				     "Could not allocate shash\n");
+
+	if (crypto_shash_digestsize(priv_data->shash) != MD5_DIGEST_SIZE) {
+		crypto_free_shash(priv_data->shash);
```
6by9: Repeated multiple times. It might be nice to make it a common exit path at the end of the function.

drivers/mfd/rpi-rp2040-gpio-bridge.c
```diff
@@ -0,0 +1,1155 @@
...
+	ret = regulator_disable(reg);
+	if (ret)
+		return dev_err_probe(dev, ret, "could not disable regulator\n");
+
+	devm_pm_runtime_enable(dev);
+	pm_runtime_use_autosuspend(dev);
+
+	controller = devm_spi_alloc_master(dev, sizeof(struct rp2040_gbdg));
+	if (!controller)
+		return dev_err_probe(dev, ENOMEM,
```
6by9: This (and subsequent) error path(s) don't clean up pm_runtime.

If `devm_pm_runtime_enable` does pass enable across to devm, then disabling in _remove feels unbalanced.

drivers/mfd/rpi-rp2040-gpio-bridge.c
```diff
@@ -0,0 +1,1155 @@
...
+static void rp2040_gbdg_gpio_set(struct gpio_chip *gc, unsigned int offset,
+				 int value)
+{
+	struct rp2040_gbdg *priv_data = gpiochip_get_data(gc);
+	struct i2c_client *client = priv_data->client;
+	u32 pattern;
+	int ret;
+
+	if (offset >= NUM_GPIO) {
+		ret = -EINVAL;
```
6by9: You do nothing with `ret`.
```
if (offset >= NUM_GPIO)
   return;
```
and no need for `else`.

drivers/mfd/rpi-rp2040-gpio-bridge.c
```diff
@@ -0,0 +1,1155 @@
...
+}
+
+static int rp2040_gbdg_gpio_dir_in(struct gpio_chip *gc, unsigned int offset)
+{
+	struct rp2040_gbdg *priv_data = gpiochip_get_data(gc);
+	struct i2c_client *client = priv_data->client;
+	int ret;
+
+	if (offset >= NUM_GPIO) {
+		ret = -EINVAL;
```
6by9:
```
   return -EINVAL;
```
and no else.

drivers/mfd/rpi-rp2040-gpio-bridge.c
```diff
@@ -0,0 +1,1155 @@
...
+}
+
+static int rp2040_gbdg_gpio_get_direction(struct gpio_chip *gc,
+					  unsigned int offset)
+{
+	struct rp2040_gbdg *priv_data = gpiochip_get_data(gc);
+	int ret;
+
+	if (offset >= NUM_GPIO)
+		ret = -EINVAL;
```
6by9: `return -EINVAL`

drivers/mfd/rpi-rp2040-gpio-bridge.c
```diff
@@ -0,0 +1,1155 @@
...
+	}
+	dev_info(dev, "%s() found dev ID: %llx, fw ver. %u\n", __func__,
+		 info.id, info.version);
+
+	ret = regulator_disable(reg);
+	if (ret)
+		return dev_err_probe(dev, ret, "could not disable regulator\n");
+
+	devm_pm_runtime_enable(dev);
+	pm_runtime_use_autosuspend(dev);
```
6by9: Do you want just autosuspend, or `pm_runtime_set_autosuspend_delay` to defer powerdown slightly in case the device is about to be used again? (See drivers/media/i2c/imx290.c for a full user).

drivers/mfd/rpi-rp2040-gpio-bridge.c
```diff
@@ -0,0 +1,1155 @@
...
+	int ret;
+
+	if (offset >= NUM_GPIO)
+		return -EINVAL;
+
+	pattern = (1 << (offset + 8));
+	if (pattern & priv_data->gpio_requested)
+		return -EBUSY;
+
+	// Resume if previously no gpio requested
```
6by9: Normal `/*   */` instead of `//`
Other places too, but this was the first I noticed.

`checkpatch.pl --strict` isn't a bad policy to adopt.

drivers/mfd/rpi-rp2040-gpio-bridge.c
```diff
@@ -0,0 +1,1155 @@
...
+{
+	int ret;
+
+	if (priv_data->fast_xfer_gpios) {
+		ret = rp2040_gbdg_10byte_cmd(
+			priv_data->client, CMD_DAT_EMIT,
+			priv_data->fast_xfer_recv_gpio_base, length);
+		return ret ? ret :
+			     rp2040_gbdg_fast_xfer(priv_data, data, length);
+	} else {
```
6by9: You've returned in the first clause, so no point in an `else`

drivers/mfd/rpi-rp2040-gpio-bridge.c
```diff
@@ -0,0 +1,1155 @@
...
+#define HALF_BUFFER (4U * ONE_KIB)
+
+#define STATUS_SIZE 4
+#define MD5_DIGEST_SIZE 16
+#define VERSION_SIZE 4
+#define ID_SIZE 8
+#define TOTAL_RD_HDR_SIZE \
+	(STATUS_SIZE + MD5_DIGEST_SIZE + VERSION_SIZE + ID_SIZE)
+
+typedef u8 md5_digest[MD5_DIGEST_SIZE];
```
6by9: typedefs are frowned upon in the kernel coding standards.

I don't immediately see it being used other than in `struct rp2040_gbdg_device_info` either.

drivers/mfd/rpi-rp2040-gpio-bridge.c
```diff
@@ -0,0 +1,1155 @@
...
+#define WRITE_DATA_PREFIX_SIZE 1
+
+#define FIXED_SIZE_CMD_PREFIX 0x81
+
+#define WRITE_DATA_UPPER_PREFIX 0x82
+#define WRITE_DATA_UPPER_PREFIX_SIZE 1
+
+#define NUM_GPIO 24
+
+enum FIXED_SIZE_COMMANDS {
```
6by9: Naming of enums shouldn't have capitals, and should generally identify the module.

drivers/mfd/rpi-rp2040-gpio-bridge.c
```diff
@@ -0,0 +1,1155 @@
...
+				       struct rp2040_gbdg_device_info *info)
+{
+	int ret;
+	u8 buf[TOTAL_RD_HDR_SIZE];
+	u8 *read_pos = buf;
+	size_t field_size;
+
+	do {
+		ret = i2c_master_recv(client, buf, sizeof(buf));
+	} while (ret == -ETIMEDOUT);
```
6by9: If I tie SCL low, do we ever get out of this loop?

drivers/mfd/rpi-rp2040-gpio-bridge.c
```diff
@@ -0,0 +1,1155 @@
...
+	field_size = sizeof_field(struct rp2040_gbdg_device_info, id);
+	memcpy(&info->id, read_pos, field_size);
+
+	return 0;
+}
+
+static int rp2040_gbdg_poll_device_info(struct i2c_client *client,
+					struct rp2040_gbdg_device_info *info)
+{
+	int ret;
```
6by9: Many kernel subsystems will request "inverted Christmas tree", ie longest lines first when defining variables.

drivers/mfd/rpi-rp2040-gpio-bridge.c
```diff
@@ -0,0 +1,1155 @@
...
+{
+	struct i2c_client *client = priv_data->client;
+	int ret;
+
+	if (len > RP2040_GBDG_BLOCK_SIZE)
+		return -EMSGSIZE;
+
+	u8 local_digest[MD5_DIGEST_SIZE];
+	u8 remote_digest[MD5_DIGEST_SIZE];
+	u8 ascii_digest[MD5_DIGEST_SIZE * 2 + 1] = { 0 };
```
6by9: Variable declarations before all code.

drivers/mfd/rpi-rp2040-gpio-bridge.c
```diff
@@ -0,0 +1,1155 @@
...
+
+	gpiod_direction_output(priv_data->fast_xfer_gpios->desc[0], 1);
+	gpiod_direction_output(priv_data->fast_xfer_gpios->desc[1], 0);
+
+	rp2040_gbdg_rp1_calc_offsets(priv_data->fast_xfer_data_index,
+				     &data_bank, &data_offset);
+	rp2040_gbdg_rp1_calc_offsets(priv_data->fast_xfer_clock_index,
+				     &clock_bank, &clock_offset);
+
+	void __iomem *data_set =
```
6by9: Variable declaration in the middle of a function.

arch/arm/boot/dts/overlays/rpi-rp2040-gpio-bridge.dtsi
```diff
@@ -0,0 +1,20 @@
+spi_bridge: spi@40 {
```
6by9: dtoverlay should be a separate patch to driver or bindings.

Documentation/devicetree/bindings/mfd/raspberrypi,rp2040-gpio-bridge.yaml
```diff
@@ -0,0 +1,77 @@
+# SPDX-License-Identifier: GPL-2.0-only
```
6by9: Bindings should be a separate patch to anything else.

Does it validate? `make dt_binding_check`

drivers/mfd/Kconfig
```diff
@@ -2362,5 +2362,14 @@ config MFD_RSMU_SPI
 	  Additional drivers must be enabled in order to use the functionality
 	  of the device.
 
+config MFD_RASPBERRYPI_RP2040_GPIO_BRIDGE
```
6by9: Is it an MFD in the kernel sense? Most MFD drivers allow for a parent that can acquire shared resources, and then you have separate drivers off it for the specific functions.

bcm2835-pm parents watchdog and power drivers
POE+ HAT uses simple-mfd-i2c as the core, with a PWM and power supply driver on top.

6by9 (commented):
Largely coding style niggles.

General minor clean up of whether there is a space before the final return of a function would be nice (it's generally expected).

roliver-rpi:
Any further comments on this?

I believe all code review comments have been addressed.

Perhaps the single remaining outstanding question is whether or not we're happy with this in mfd?

Thoughts on potential alternate locations @pelwell @6by9 @naushir ?

If we're happy with mfd then I believe this is ready to merge.

pelwell:
> whether or not we're happy with this in mfd?

What are the alternatives:
* An I2C attached SPI controller would live in `drivers/spi`. 
* Many devices also provide GPIOs. For example, sc16is75x is an I2C- or SPI-attached UART that also acts as a GPIO controller. It lives in `drivers/tty/serial`, again suggesting that `drivers/spi` is appropriate.
* Shove it in `drivers/misc`?

roliver-rpi:
Driver has been moved to drivers/spi

drivers/spi/spi-rp2040-gpio-bridge.c
```diff
@@ -0,0 +1,1219 @@
...
+
+	if (IS_ERR(reg))
+		return PTR_ERR(reg);
+
+	rp2040_gbdg->regulator = reg;
+
+	return 0;
+}
+
+static void rp2040_gbdg_parse_dt(struct rp2040_gbdg *rp2040_gbdg)
```
pelwell, line 908: Is it reasonable that this has no return value? It has many points where it returns early after an error.

pelwell (commented):
Realistically I'm not going to fully grok this new driver, but apart from the lack of error return from the DT parsing it looks OK to me.

drivers/spi/spi-rp2040-gpio-bridge.c
roliver-rpi: Only the parsing of the 'bypass-cache' property in rp2040_gbdg_parse_dt() is required for all platforms. This cannot fail due to the API of of_property_read_bool() returning the bool.

Any failure in the rest of rp2040_gbdg_parse_dt() is not an issue as the driver will use I2C to transfer blocks of data as a fallback (rather than fast-xfer using RP1 bit-bashing). In fact, we expect failure on platforms without RP1.

If we were to return a value from rp2040_gbdg_parse_dt, in it's current form, it would only be informational and not to be acted upon (other than perhaps a dev_info).

I'm happy to refactor rp2040_gbdg_parse_dt() if we can think of a cleaner/easier to understand API. Alternatively, I could comment the function as it's perhaps not sufficiently clear in its current form.

pelwell: It's OK - if you're happy justifying it being as it is now, I'm happy to merge it

PR #6184, opened 2024-05-22
ARM: dts: Fix camera sync parameters

pelwell:
The camera sync parameter declarations include are invalid in that they have both embedded string values and following cell values. The intention here was to use the cell values, but the embedded "0" was not removed as it should have been.

The implication of this mistake is that inverted sync flags would be ignored because the 0 value corresponds to GPIO_ACTIVE_HIGH. The extra cell value would be treated as the start of another override, but then ignored because the end of the parameter is reached.

These errors should have been picked up by the automated checks, but the base dts files weren't subjected to the same validation. This has now been corrected.


pelwell:
@6by9 FYI the inverted sync parameters introduced by #6088 were almost certainly broken.

Issue #6134, opened 2024-04-29
igc driver not working on Pi5


lasselj:
### Describe the bug

It appears that with an Intel i225 and i226 NIC in the Pi5 both of which use the igc driver it is not possible to read any address in the ioremapped address space. I should mention that the Intel i210 which uses the igb driver works without a hitch (the architecture of the i210 and the i225/226 is remarkably similar).

In igc_probe the call to pci_resource_start and pci_resource_len return the correct addresses of the memory also observed in lspci. The call to ioremap returns a valid address, however all calls to igc_rd32 always without exception return 0x0. Similarly a direct call to readl in the remapped space returns 0 no matter the register.

For fun I have also tried to map the pci resource using the pci_iomap call used in the igb driver but it doesn't make a difference.

Loading the driver fails with error -13 (IGC_ERR_SWFW_SYNC) which is essentially the first time the driver needs to read a set bit in a read call to acquire the software/firmware semaphore in igc_get_hw_semaphore_i225.

The same device (225/226 with igc) works fine on a CM4.

I've set the dtparam pciex1 and I've tried using the dtoverlay pcie_32bit_dma but it does not make a difference and the latter is not required on the CM4.

#help :-)

### Steps to reproduce the behaviour

(with a 225/226 nic in the PCIe slot)
modprobe igc

### Device (s)

Raspberry Pi 5

### System

Generated using pi-gen, https://github.com/RPi-Distro/pi-gen, f19ee211ddafcae300827f953d143de92a5c6624, stage4

2024/04/20 11:53:30 
Copyright (c) 2012 Broadcom
version d1744d21 (release) (embedded)

Linux pi5-pcie 6.6.28-v8-16k+ #2 SMP PREEMPT Sun Apr 28 08:30:38 BST 2024 aarch64 GNU/Linux

### Logs

Probably not worth a lot, but for the purposes of giving insight into the call stack:

[12587.426063] Intel(R) 2.5G Ethernet Linux Driver
[12587.426069] Copyright(c) 2018 Intel Corporation.
[12587.426171] igc 0000:01:00.0: PCIe PTM not supported by PCIe bus/controller
[12587.426244] igc_probe (1): 0x1b00000000-0x100000
[12587.426246] igc_probe (2): 0xffffc00083000000
[12587.426247] igc_probe (3): 0x0-0x0
[12587.426250] igc_get_invariants_base(1): 0
[12587.426251] igc_get_invariants_base(2): 0
[...]

PR #6186, opened 2024-05-24
DTS: overlays: add mmio-hi parameter to pciex1-compat-pi5

P33M:
I225-V network adapters have a buggy ROM that won't complete internal initialisation if (at least) BAR0 has an assigned address of 0x0.

Add a parameter to the pciex1-compat-pi5 overlay to make outbound addresses start at 2GB.

see #6134


geerlingguy:
Also related: https://github.com/raspberrypi/linux/issues/6278 - behavior of some devices behind multi-port PCIe switches.

PR #6194, opened 2024-05-29
media: dt-bindings: i2c: Add Sony IMX500

roliver-rpi:
Add YAML device tree binding for the Sony IMX500 CMOS image sensor / CNN inference engine.  Also, add a MAINTAINERS entry.

### DIFF
--- a/Documentation/devicetree/bindings/media/i2c/sony,imx500.yaml
+++ b/Documentation/devicetree/bindings/media/i2c/sony,imx500.yaml
@@ -1,0 +1,161 @@
+# SPDX-License-Identifier: (GPL-2.0-only OR BSD-2-Clause)
+%YAML 1.2
+---
+$id: http://devicetree.org/schemas/media/i2c/sony,imx500.yaml#
+$schema: http://devicetree.org/meta-schemas/core.yaml#
+
+title: Sony CMOS Digital Image Sensor and CNN
+
+maintainers:
+  - Raspberry Pi <kernel-list@raspberrypi.com>
+
+description: |-
+  The Sony IMX500 is a stacked 1/2.3-inch CMOS digital image sensor and inbuilt
+  AI processor with an active array CNN (Convolutional Neural Network) inference
+  engine.  The native sensor size is 4056H x 3040V, and the module also contains
+  an in-built ISP for the CNN. The module is programmable through an I2C
+  interface with firmware and neural network uploads being made over SPI. The
+  default I2C address is 0x1A, with an address of 0x10 being selectable via
+  SLASEL. The module also has a second I2C interface available with a fixed
+  address of 0x36.  Image data is sent through MIPI CSI-2, which is configured
+  as either 2 or 4 data lanes.
+
+properties:
+  compatible:
+    const: sony,imx500
+
+  reg:
+    description: I2C device address
+    maxItems: 1
+
+  clocks:
+    maxItems: 1
+
+  clock-names:
+    description: |-
+      Input clock (12 to 27 MHz)
+    items:
+      - const: inck
+
+  interrupts:
+    maxItems: 1
+
+  vana-supply:
+    description: Supply voltage (analog) - 2.7 V
+
+  vdig-supply:
+    description: Supply voltage (digital) - 0.84 V
+
+  vif-supply:
+    description: Supply voltage (interface) - 1.8 V
+
+  reset-gpios:
+    description: |-
+      Sensor reset (XCLR) GPIO
+
+      Chip clear in lieu of built-in power on reset. To be set 'High' after
+      power supplies are brought up and INCK supplied.
+
+  port:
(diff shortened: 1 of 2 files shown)


Documentation/devicetree/bindings/media/i2c/sony,imx500.yaml
```diff
@@ -0,0 +1,161 @@
...
+      Video output port
+
+    properties:
+      endpoint:
+        $ref: /schemas/media/video-interfaces.yaml#
+        type: object
+        unevaluatedProperties: false
+        properties:
+          data-lanes:
+            items:
```
6by9, line 72: Your description states it supports 2 or 4 lanes, but this only allows 2.
```
          data-lanes:
            oneOf:
              - items:
                  - const: 1
                  - const: 2
                  - const: 3
                  - const: 4
              - items:
                  - const: 1
                  - const: 2
```

Documentation/devicetree/bindings/media/i2c/sony,imx500.yaml
```diff
@@ -0,0 +1,161 @@
...
+              - const: 1
+              - const: 2
+          clock-noncontinuous: true
+          link-frequencies: true
+          clock-lanes:
```
6by9, line 77: clock-lanes is normally skipped when there is no support for lane reordering.

Documentation/devicetree/bindings/media/i2c/sony,imx500.yaml
```diff
@@ -0,0 +1,161 @@
...
+      endpoint:
+        $ref: /schemas/media/video-interfaces.yaml#
+        type: object
+        unevaluatedProperties: false
+        properties:
+          data-lanes:
+            items:
+              - const: 1
+              - const: 2
+          clock-noncontinuous: true
```
6by9, line 75: Whilst not incorrect to be in the binding, just check whether you are actually selecting non-continuous clock, or and whether it is mandatory (I'm never clear in bindings whether this "true" means supported or must be set).

I've just been looking at imx477 and found that the sensor is configured for continuous clocks.

Documentation/devicetree/bindings/media/i2c/sony,imx500.yaml
```diff
@@ -0,0 +1,161 @@
...
+        };
+      };
+    };
+
+    spi {
+      #address-cells = <1>;
+      #size-cells = <0>;
+
+      sensor@0 {
+        compatible = "sony,imx500";
```
6by9, line 154: You've not pushed the driver yet, but doesn't this instantiate a second instance of the driver? How do the 2 get associated, or do they deliberately not?

I'm thinking of what happens if you have 2 attached to the system. How do you control which SPI and I2C device are linked, or does it not matter.

6by9 (commented):
You have to love bindings :-/

Documentation/devicetree/bindings/media/i2c/sony,imx500.yaml
6by9: For reference, #6208 for adding continuous clock mode selection to imx477.

Documentation/devicetree/bindings/media/i2c/sony,imx500.yaml
roliver-rpi: I'm currently refactoring the driver but I will push it soon.

I'm doing something very similar to the s5c73m3 driver:
https://github.com/raspberrypi/linux/blob/5882bc5db17dd4a730faaedf547d073e487ec163/drivers/media/i2c/s5c73m3/s5c73m3-core.c#L1664
https://github.com/raspberrypi/linux/blob/5882bc5db17dd4a730faaedf547d073e487ec163/drivers/media/i2c/s5c73m3/s5c73m3-spi.c#L133

I'll have to examine the consequences of having 2 devices attached to the the system more closely.

6by9: So AIUI that will gate registering the SPI driver part until the I2C bit has loaded, which is necessary as the i2c part allocates the common state structure.

However I don't see how that will work with two (it may not be applicable for that driver).
`s5c73m3_register_spi_driver` will get called when the first I2C device gets probed, but that will allow both SPI instances to probe, and the `container_of(spi->dev.driver, ...)` will point back to the one I2C instance.
When the second I2C device probes, `spi_register_driver` will get called for a second time, but I sort of hope it falls down the "Error: Driver '%s' is already registered, aborting" error path at https://elixir.bootlin.com/linux/latest/source/drivers/base/driver.c#L241

I don't know exactly what it would look like, but I would have expected a form not dissimilar to the `lens-focus` or `flash` DT links that you get from an image sensor to lens or flash devices. (Those get handled by the common v4l2-fwnode code, whereas this would be driver specific).

Issue #6196, opened 2024-05-29
/soc/nvmem_* nodes in RPI5 device tree violate DTS specification


lorc:
### Describe the bug

`bcm2712-rpi.dtsi` file provide `/soc/nvmem_*` nodes with `reg` values like `<0 192>` which violates DTS specification as those nodes reside inside `/soc/` node which has `ranges` defined as follows:

```
		ranges     = <0x7c000000  0x10 0x7c000000  0x04000000>;
```

So it is impossible to calculate, say, `/soc/nvmem_otp` address. Moreover, `reg` property inside those nodes does not provide memory address at all, which also violates dts, because `/soc/` node is compatible with "simple-bus".

This causes issues for non-Linux kernel. For example Xen is not happy about this.

### Steps to reproduce the behaviour

Just look at the source code (`arch/arm/boot/dts/broadcom/bcm2712-rpi.dtsi`). And compare this with Device Tree Specification (https://readthedocs.org/projects/devicetree-specification/downloads/pdf/latest/) sections `2.3.8` and `4.5`.

### Device (s)

Raspberry Pi 5

### System

Linux version rpi-6.6.y 

### Logs

This is how Xen complains about those entries:
```
I/TC: Added guest 1
(XEN) *** LOADING DOMAIN 0 ***
(XEN) Loading d0 kernel from boot module @ 0000000000e00000
(XEN) Allocating 1:1 mappings totalling 128MB for dom0:
(XEN) BANK[0] 0x00000010000000-0x00000018000000 (128MB)
(XEN) Grant table range: 0x00000002000000-0x00000002040000
(XEN) Unable to retrieve address 0 for /soc/nvmem_otp
(XEN) Device tree generation failed (-22).
(XEN) 
(XEN) ****************************************
(XEN) Panic on CPU 0:
(XEN) Could not set up DOM0 guest OS (rc = -22)
(XEN) ****************************************
(XEN) 
(XEN) Reboot in five seconds...
```

### Additional context

Proper solution is to provide own device with own addressing mode, like this:

```
	nvmem {
		compatible = "raspberrypi,rpi-otp";
		#address-cells = <1>;
		#size-cells = <1>;
		firmware = <&firmware>;
		nvmem_otp: nvmem_otp {
			reg = <0 192>;
		};

		nvmem_cust: nvmem_cust {
			reg = <1 8>;
		};

		nvmem_mac: nvmem_mac {
			reg = <2 6>;
		};

[...]

PR #6197, opened 2024-05-30
dts: bcm2712-rpi: Move rpi-otp nodes onto a bus

pelwell:
The rpi-otp driver uses a virtualised, OTP-relative addressing scheme. However, when instance nodes are children of "/soc" they appear to be addressable directly by the host, which is wrong (but not in a way which causes an error unless one goes looking for one).

Add a wrapper (bus) node without a "ranges" property to make the separation clear.

See: https://github.com/raspberrypi/linux/issues/6196

### DIFF
--- a/arch/arm/boot/dts/broadcom/bcm2708-rpi.dtsi
+++ b/arch/arm/boot/dts/broadcom/bcm2708-rpi.dtsi
@@ -26,18 +26,24 @@
 };
 
 &soc {
-	nvmem_otp: nvmem_otp {
-		compatible = "raspberrypi,rpi-otp";
-		firmware = <&firmware>;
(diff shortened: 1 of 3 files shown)


popcornmix:
Are the rpi-otp nodes in `bcm2708-rpi.dtsi` and `bcm2711-rpi-ds.dtsi` okay as they are?

pelwell:
No, not really - I can give them the same treatment.

pelwell:
Like so.

popcornmix (approved):

lorc:
@pelwell, thank you. I checked your PR on my setup, it is working as expected, i.e. Xen is happy.

PR #6203, opened 2024-06-04
Fix HS400 on CM5

P33M:
CM5 HS400 is now reliable for me.
Also found two more commits for SD down the back of the sofa...


P33M:
@timg236 this works for me on both r0.1 and r0.2 boards.

PR #6209, opened 2024-06-06
rpi-6.10.y/vchiq dma mask updates

kbingham:
These update the (remaining) vchiq bus devices to explicitly set the DMA_MASK.

In particular, the VC-SM-CMA triggers a warning at boot without this.

This implementation updates the SM-CMA, ISP, and Codec to match the audio and camera device. Though I do wonder if we should instead make the vchiq_bus probe() just register this on all vchiq devices directly instead!

(I assume all vchiq based devices would have an identical DMA mask).

### DIFF
--- a/drivers/staging/vc04_services/bcm2835-codec/bcm2835-v4l2-codec.c
+++ b/drivers/staging/vc04_services/bcm2835-codec/bcm2835-v4l2-codec.c
@@ -3898,6 +3898,13 @@ static int bcm2835_codec_probe(struct vchiq_device *device)
 	struct media_device *mdev;
 	int ret = 0;
 
+	ret = dma_set_mask_and_coherent(&device->dev, DMA_BIT_MASK(32));
+	if (ret) {
+		dev_err(&device->dev, "dma_set_mask_and_coherent failed: %d\n",
+			ret);
+		return ret;
+	}
+
 	drv = devm_kzalloc(&device->dev, sizeof(*drv), GFP_KERNEL);
 	if (!drv)
 		return -ENOMEM;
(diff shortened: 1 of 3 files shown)


popcornmix:
Yes, VideoCore has a 32-bit (or perhaps more accurately 30-bit address bus), so any services that use vchiq will be the same. 

@pelwell ?

pelwell:
The limitation here is the bus to the VPU and its peripherals. Yes, it's a 30-bit address bus but offset at the 3GB mark, so it's effectively 32-bit. The other clients, `bcm2835-audio` and `bcm2835-camera`, also explicitly set 32-bit masks, so it could be moved into the common core instead.

pelwell:
This only builds for arm64 - ARCH=arm builds fail complaining about the absence of dmac_inv_range and dmac_clean_range.

6by9:
> This only builds for arm64 - ARCH=arm builds fail complaining about the absence of dmac_inv_range and dmac_clean_range.

Missing forward port of https://github.com/raspberrypi/linux/commit/bd7e4336b9f12cf217deeb5778bcd4cf75a5e75e ?

pelwell:
> Missing forward port of https://github.com/raspberrypi/linux/commit/bd7e4336b9f12cf217deeb5778bcd4cf75a5e75e ?

Yes - it was literally the first missing commit suggested by my forward-porting script, and now applied.

6by9:
Note that the use cases for dmac_inv_range and dmac_clean_range as cache flush functions via vcsm-cma are likely dead.

I think it was only JC's HEVC decode on Pi3 that was using them, and with the switch to V3D being driven from the kernel that is no longer functional. (It used VPU, QPUs, ARM NEON, and almost every other hardware block it could get hold of).

popcornmix:
@6by9 I've pushed to 6.10 a revert of the commit that exposes `dmac_inv_range`/`dmac_clean_range` and removed the code path from vc-sm-cma driver, as I agree that code path is no longer used.

kbingham:
Refreshed this branch, but I still think we should make vchiq bus do this at the core level (and send that patch upstream).

That will take some more minutes than I have right now though.

popcornmix:
I agree this would be best handled in common code. 
I'll merge this for now, but, happy to receive an update when you are ready with the upstream version.

PR #6215, opened 2024-06-10
Fix the IQaudIO handling of 16-bit samples

pelwell:
Between rpi-6.1.y and rpi-6.6.y the DA7213 codec driver picked up a patch selecting a BCLK ratio of 32 for 16-bit audio. This breaks the CPU DAI if it is expecting a BCLK ratio of 64. Instead of baking the knowledge of the codec's foibles into the soundcard driver, give the codec a set_bclk_ratio method and make the soundcard driver use it.


pelwell:
This is said patch: https://github.com/raspberrypi/linux/commit/c89e652e84f636354213aae42490c7f77a0eadde

timg236:
LGTM

PR #6219, opened 2024-06-11
lan78xx: Read initial EEE status from DT

pelwell:
Add two new DT properties:
* microchip,eee-enabled  - a boolean to enable EEE
* microchip,tx-lpi-timer - time in microseconds to wait before entering low power state


pelwell:
I believe this should be the 6.10 equivalent of applying 61d24cbde05c9dab66c1e9e9a1e421a9b7c324b6, d9d641b0b1061748789aaa1fff7eee034b4c1038 and 75122994b1746ba360f44c6e3a14a39f0f0f8688.

popcornmix (approved):

pelwell:
I have a feeling this patch can be dropped.

pelwell:
Nope, it's still necessary. The follow-up patch is an improvement on the original in that it preserves the value of tx_lpi_timer across the disablement of eee and/or tx_lpi.

PR #6234, opened 2024-06-20
arm64: dts: remove broken DTSI links

Noltari:
Both [bcm283x-rpi-csi1-2lane.dtsi](https://github.com/raspberrypi/linux/blob/rpi-6.6.y/arch/arm64/boot/dts/broadcom/bcm283x-rpi-csi1-2lane.dtsi) and [bcm283x-rpi-lan7515.dtsi](https://github.com/raspberrypi/linux/blob/rpi-6.6.y/arch/arm64/boot/dts/broadcom/bcm283x-rpi-lan7515.dtsi) are linked to ["../arm/boot/dts/"](https://github.com/raspberrypi/linux/blob/rpi-6.6.y/arch/arm64/boot/dts) instead of ["../arm/boot/dts/broadcom"](https://github.com/raspberrypi/linux/blob/rpi-6.6.y/arch/arm64/boot/dts/broadcom), so the links are broken.

Fixes: d060fc0b4568 ("BCM2708: Add core Device Tree support")

OpenWrt reference: https://github.com/openwrt/openwrt/pull/15762

### DIFF
--- a/arch/arm64/boot/dts/broadcom/bcm283x-rpi-csi1-2lane.dtsi
+++ b/arch/arm64/boot/dts/broadcom/bcm283x-rpi-csi1-2lane.dtsi
@@ -1,1 +1,0 @@
-../../../../arm/boot/dts/bcm283x-rpi-csi1-2lane.dtsi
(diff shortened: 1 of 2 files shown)


pelwell:
They're also unnecessary, which explains we nobody noticed.

6by9:
LGTM. All the arm64 files use includes of the arm versions, so they're redundant.
...
And Phil's faster at typing!

PR #6250, opened 2024-07-04
JPEG Control Reference in bcm2835-codec driver. Add control ID V4L2_CID_JPEG_RESTART_INTERVAL

bllem:
I added Image Process Control ID to be able to add RSTn markers to the generated JPEG file.

Tested on a live example.
Example of use:

```C++
v4l2_control controls;
controls.id = V4L2_CID_JPEG_RESTART_INTERVAL;
controls.value = 100;
        
if(0 < ioctl(fd,  VIDIOC_S_CTRL, &controls)) {
    perror("Error  VIDIOC_S_EXT_CTRLS"); 
}
```

### DIFF
--- a/drivers/staging/vc04_services/bcm2835-codec/bcm2835-v4l2-codec.c
+++ b/drivers/staging/vc04_services/bcm2835-codec/bcm2835-v4l2-codec.c
@@ -2423,6 +2423,17 @@ static int bcm2835_codec_s_ctrl(struct v4l2_ctrl *ctrl)
 		ret = 0;
 		break;
 
+	case V4L2_CID_JPEG_RESTART_INTERVAL:
+		if (!ctx->component)
+			break;
+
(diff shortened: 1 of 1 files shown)


6by9:
Fundamentally the patches look OK, but a couple of things.

- Please squash the two patches together as we don't need the intermediate step with the default being 1. (We can do that during the merge, but it's as easy for you to do it as there are other changes being requested).
- Linux contributions need to be identifiable as documented in https://www.kernel.org/doc/html/latest/process/submitting-patches.html#sign-your-work-the-developer-s-certificate-of-origin. Your commits as "root@DESKTOP-EB9E36V" with no name aren't, so can't be accepted in that form. Please use a valid email address and your name.
- If I've read it right, the top end limit of 1,000,000 would be 256MPix between restart markers. That seems a little excessive. The other drivers in mainline that support this control either use a max of 100 (coda and mediatek) or 0xFFFF (Samsung).

PR #6253, opened 2024-07-05
staging: bcm2835-codec: Disable HEADER_ON_OPEN for video encode

6by9:
Video encode can defer generating the header until the first frame is presented, which allows it to take the colourspace information from the frame rather than just the format.

Enable that for the V4L2 driver now that the firmware populates all the parameters.

https://github.com/raspberrypi/firmware/issues/1885


popcornmix (approved):

6by9:
With the corresponding firmware patch, the OP has confirmed that this works.

PR #6254, opened 2024-07-05
dts: Make camN_reg and camN_reg_gpio overrides generic

6by9:
The camera regulator GPIO can be used for other purposes, so the camN_reg override to allow disabling is potentially useful on any platform.
camN_gpio is less useful, but isn't invalid.

Move these overrides from the CM dt files to bcm270x-rpi.dtsi / bcm2712-rpi.dtsi

### DIFF
--- a/arch/arm/boot/dts/broadcom/bcm2711-rpi-cm4.dts
+++ b/arch/arm/boot/dts/broadcom/bcm2711-rpi-cm4.dts
@@ -498,13 +498,6 @@ i2c_csi_dsi0: &i2c0 {
 			<&ant2>, "output-high?=off",
 			<&ant2>, "output-low?=on";
 
-		cam0_reg = <&cam0_reg>,"status";
-		cam0_reg_gpio = <&cam0_reg>,"gpio:4",
-				  <&cam0_reg>,"gpio:0=", <&gpio>;
-		cam1_reg = <&cam1_reg>,"status";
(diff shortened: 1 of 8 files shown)


arch/arm/boot/dts/broadcom/bcm2711-rpi-cm4.dts
```diff
@@ -498,13 +498,6 @@ i2c_csi_dsi0: &i2c0 {
 			<&ant2>, "output-high?=off",
 			<&ant2>, "output-low?=on";
 
-		cam0_reg = <&cam0_reg>,"status";
-		cam0_reg_gpio = <&cam0_reg>,"gpio:4",
-				  <&cam0_reg>,"gpio:0=", <&gpio>;
```
pelwell, line 503: This overriding of the GPIO controller is lost in the replacement, common definitions. This means that users won't be able to redirect from `&expgpio` to `&gpio`. I don't think there's any harm in setting this across all platforms even though it is unnecessary, but CM4 could get a dedicated implementation.

pelwell (commented):
As per inline comment, camN_reg_gpio won't work as intended on CM4.

Also, the commit message refers to it as camN_gpio.

arch/arm/boot/dts/broadcom/bcm2711-rpi-cm4.dts
6by9: Thank you. Reinstated.

PR #6257, opened 2024-07-07
overlays: Add Pineboards HatDrive! POE+

mikegapinski:
This overlay sets sets the current supply to 5A on the upcoming Pineboards HatDrive! POE+ board. 

cc @pelwell


pelwell:
You've got a few things to fix: https://github.com/raspberrypi/linux/actions/runs/9828536077/job/27132619672?pr=6257#step:4:515

mikegapinski:
> You've got a few things to fix: https://github.com/raspberrypi/linux/actions/runs/9828536077/job/27132619672?pr=6257#step:4:515

Tabs vs spaces... Thanks for letting me know. Pushed again.

One thing to confirm - is hat_current_supply sufficient for the 5A input or should I add something else to remove the USB power limit that is active by default when the official USB-C PD is detected?

pelwell:
> is hat_current_supply sufficient for the 5A input

I think it's sufficient. You can confirm by looking in /chosen/power/*:
```
$ (cd /proc/device-tree/chosen/power/; for a in *current*; do echo -n $a; od -An -td4 --endian=big $a; done)
```

mikegapinski:
image
it looks like it is not enough sadly, should I set 'usb_max_current_enable' to 1?

pelwell:
How old is your EEPROM image? 2024-04-05 (https://github.com/raspberrypi/rpi-eeprom/commit/d8abe8c67d2af9168b7d72628d4bdc00b027bfbe) is the first version with full HAT+ support.

mikegapinski:
```
BOOTLOADER: update available
   CURRENT: Sat 20 Apr 10:53:30 UTC 2024 (1713610410)
    LATEST: Wed  5 Jun 15:41:49 UTC 2024 (1717602109)
   RELEASE: default (/lib/firmware/raspberrypi/bootloader-2712/default)
            Use raspi-config to change the release.
```

pelwell:
Does your HAT declare itself to be a stackable HAT+ by using an EEPROM address of 0x51, 0x52 or 0x53? It looks as though, using  the current sequence of operations, a `hat_current_supply` property in a non-stackable HAT/HAT+ is applied too late to be noticed by the code that sets the current limit.

mikegapinski:
I think it does, but I'll check the schematic tomorrow to be sure.  This part hasn't changed from the OG HatDrive! Top and it has the same config as your official M.2 HAT+. I can also modify the addresses or disconnect the EEPROM completely. 

BTW, if this is an issue in the firmware it also won't work fine on the Pironman overlay that was merged recently since it also sets the current in a similar way.

With the current overlay I can draw around 22W through PoE+ with a high power USB device connected and the Pi 5 overlocked (100% CPU load). I could push it a little further with the official PSU, that's why I started looking into USB current limits.

mikegapinski:
@pelwell confirmed, we are using 0x51. In my test setup I've set the dtoverlay in config.txt (the HAT is not flashed)

image

pelwell:
> I've set the dtoverlay in config.txt (the HAT is not flashed)

That definitely won't work at present. Please flash the EEPROM so we can confirm that everything else works.

pelwell:
It looks like there is some confusion here over `hat_current_supply`, and not all of it yours.

In an ideal world, one might want the firmware to:
1. look at the hardware and any loaded overlays,
2. based on both of them, decide whether sufficient power is available to support USB booting, and
3. write the results to Device Tree.

However, the fact that the overlays may be loaded from USB causes a chicken-and-egg problem - it's not possible to know that an overlay enables USB booting before USB booting has loaded the overlay. Instead, the presence of a Mode 1 Power HAT+ EEPROM address is used to indicate the presence of a high current supply. The value of `hat_current_supply`/`max_current` in DT is only there to provide more accurate information to the kernel and/or userspace.

If your POE+ HAT+ uses address 0x53, `usb_max_current_enable` will be set automatically and USB booting will be enabled.

pelwell:
> the Pironman overlay that was merged recently since it also sets the current in a similar way.

@sunfounder FYI.

mikegapinski:
We’ll change the address tomorrow and test again. Thanks for the info!

Wysyłane z aplikacji Outlook dla systemu iOS<https://aka.ms/o0ukef>
________________________________
Od: Phil Elwell ***@***.***>
Wysłane: Tuesday, July 9, 2024 5:48:32 PM
Do: raspberrypi/linux ***@***.***>
DW: Michał Gapiński ***@***.***>; Author ***@***.***>
Temat: Re: [raspberrypi/linux] overlays: Add Pineboards HatDrive! POE+ (PR #6257)


the Pironman overlay that was merged recently since it also sets the current in a similar way.

@sunfounder<https://github.com/sunfounder> FYI.

pelwell:
We've had a major documentation logjam, but I hope that when the final HAT+ doc comes out it is much clearer about the Device Tree semantics.

mikegapinski:
It makes our job a little more interesting when we have to dig deeper or guess what the schematics are 😅 that's how it all started for us last year.

With the HAT+ address changed do we still need the overlay? It's not a problem but if we can save some time on programming the boards in production we'll do it. Our POE+ HAT doesn't really need anything more than a 5A power limit

Wysyłane z aplikacji Outlook dla systemu iOS<https://aka.ms/o0ukef>
________________________________
Od: Phil Elwell ***@***.***>
Wysłane: Tuesday, July 9, 2024 5:58:54 PM
Do: raspberrypi/linux ***@***.***>
DW: Michał Gapiński ***@***.***>; Author ***@***.***>
Temat: Re: [raspberrypi/linux] overlays: Add Pineboards HatDrive! POE+ (PR #6257)


We've had a major documentation logjam, but I hope that when the final HAT+ doc comes out it is much clearer about the Device Tree semantics.

pelwell:
I can't see how not including the name of the overlay in the EEPROM image could save more than milliseconds of programming time. The only way to make a real saving is to not have (or not program) the EEPROM, but then it wouldn't be a HAT+.

pelwell:
There's a trial EEPROM image that makes the handling of overlays from the different flavours of HAT+ consistent, but as discussed previously that hat_current_supply is only a description of the power capabilities for userspace. Only MODE1 power HAT+s actually enable USB booting.

The recovery-style .zip is available here: https://drive.google.com/file/d/1uVhO08f09L2JzATppjlinGTpCeaTneup/view?usp=drive_link
Extract the files into a FAT partition on a spare SD card, boot it, and wait for the activity LED to flash.

mikegapinski:
> There's a trial EEPROM image that makes the handling of overlays from the different flavours of HAT+ consistent, but as discussed previously that hat_current_supply is only a description of the power capabilities for userspace. Only MODE1 power HAT+s actually enable USB booting.
> 
> The recovery-style .zip is available here: https://drive.google.com/file/d/1uVhO08f09L2JzATppjlinGTpCeaTneup/view?usp=drive_link Extract the files into a FAT partition on a spare SD card, boot it, and wait for the activity LED to flash.

Thank you! We changed the address on our board but I am yet to test it. We're entering production very soon, so I'll test everything and get back to you after the weekend.

mikegapinski:
@pelwell Thanks for the guidance, I just tested a final board with an 0x53 address, written the EEPROM with the overlay name from this PR and it worked flawlessly:

```
mgapinski@poepi:~/utils/eeptools $ sudo ./eepflash.sh -w -t=24c32 -a=53 -f=hatdrive_poe_plus.eep 
This will attempt to talk to an eeprom at i2c address 0x53. Make sure there is an eeprom at this address.
This script comes with ABSOLUTELY no warranty. Continue only if you know what you are doing.
Do you wish to continue? (yes/no): yes
Writing...
Closing EEPROM Device.
Done.
mgapinski@poepi:~/utils/eeptools $ sudo reboot

Broadcast message from root@poepi on pts/1 (Sat 2024-08-03 14:53:03 BST):

The system will reboot now!
mgapinski@MacBook-Pro ~ % ssh poepi
Linux poepi 6.6.36-v8-16k+ #1 SMP PREEMPT Sun Jul  7 15:17:41 BST 2024 aarch64

The programs included with the Debian GNU/Linux system are free software;
the exact distribution terms for each program are described in the
individual files in /usr/share/doc/*/copyright.

Debian GNU/Linux comes with ABSOLUTELY NO WARRANTY, to the extent
permitted by applicable law.
Last login: Sat Aug  3 14:53:27 2024
mgapinski@poepi:~ $ (cd /proc/device-tree/chosen/power/; for a in *current*; do echo -n $a; od -An -td4 --endian=big $a; done)
hat_current_supply        5000
max_current        5000
usb_max_current_enable           1
usb_over_current_detected           0
```

It looks like the overlay can be merged as is, hopefully it'll reach stable soon enough. We're hoping to launch the board in ~1 month

pelwell:
I'll merge it once you've added the extra blank lines between entries, as flagged up by the build checks: https://github.com/raspberrypi/linux/actions/runs/9840009688/job/27170998817?pr=6257#step:4:515

mikegapinski:
@pelwell I've added the blank lines + synced with the base branch

arch/arm/boot/dts/overlays/README
```diff
@@ -3703,6 +3703,12 @@ Load:   dtoverlay=pineboards-hat-ai
 Params: <None>
 
 
+Name:   pineboards-hatdrive-poe-plus
+Info:   Configures the Pineboards HatDrive! PoE+
+Load:   dtoverlay=pineboards-hat-ai
```
pelwell: Should be "dtoverlay=pineboards-hatdrive-poe-plus"

mikegapinski: Missed that, updated

Issue #6256, opened 2024-07-06
aplay maximum period_size issue for 'HIFIBERRY DAC+ ADC PRO' and RPi5 running Raspberry Pi OS bookworm


anchung-chen:
### Describe the bug
My setup is the RPi5/RPi4  with HIFIBERRY DAC+ ADC PRO, running the lately updated Raspberry Pi OS bookworm with default config.txt settings.

On the Raspberry Pi 5: aplay reports a 'period_size' of 512, which is very small and could not change to a large value.
On the Raspberry Pi 4: aplay reports a 'period_size' of 16384, which is normal and could change to a large value.

p.s. Pi5 and Pi4 boot the same uSD Card with lately updated Raspberry Pi OS bookworm with kernel 6.6.31+rpt-rpi-2712.

### Steps to reproduce the behaviour
$ sudo apt install alsa-utils sox
$ sox -V -r 48000 -n -b 32 -c 2 sin_1K_32bit.wav synth 30 sin 1000 vol -10dB
$ aplay -D hw:CARD=sndrpihifiberry,DEV=0  -v sin_1K_32bit.wav 
$ aplay -D hw:CARD=sndrpihifiberry,DEV=0 --period-size=32768  -v sin_1K_32bit.wav

### Device (s)

Raspberry Pi 5 or Raspberry Pi 4 with HIFIBERRY DAC+ ADC PRO

### System

$ cat /etc/rpi-issue
Raspberry Pi reference 2023-10-10
Generated using pi-gen, https://github.com/RPi-Distro/pi-gen, 962bf483c8f326405794827cce8c0313fd5880a8, stage4
$ vcgencmd version
2024/06/05 16:41:49 
Copyright (c) 2012 Broadcom
version 6fe0b091 (release) (embedded)
$ uname -a
Linux raspberrypi 6.6.31+rpt-rpi-2712 #1 SMP PREEMPT Debian 1:6.6.31-1+rpt1 (2024-05-29) aarch64 GNU/Linux

### Logs

--------------------------------RPi5-------------------------------------------
$ aplay -l
**** List of PLAYBACK Hardware Devices ****
card 0: vc4hdmi0 [vc4-hdmi-0], device 0: MAI PCM i2s-hifi-0 [MAI PCM i2s-hifi-0]
  Subdevices: 1/1
  Subdevice #0: subdevice #0
[...]

PR #6263, opened 2024-07-10
dmaengine: dw-axi-dmac: Honour snps,block-size

pelwell:
The snps,block-size DT property declares the maximum block size for each channel of the dw-axi-dmac. However, the driver ignores these when setting max_seg_size and uses MAX_BLOCK_SIZE (4096) instead.

To take advantage of the efficiencies of larger blocks, calculate the minimum block size across all channels and use that instead.

See: https://github.com/raspberrypi/linux/issues/6256

### DIFF
--- a/drivers/dma/dw-axi-dmac/dw-axi-dmac-platform.c
+++ b/drivers/dma/dw-axi-dmac/dw-axi-dmac-platform.c
@@ -1470,6 +1470,7 @@ static int dw_probe(struct platform_device *pdev)
 	struct dw_axi_dma *dw;
 	struct dw_axi_dma_hcfg *hdata;
 	struct reset_control *resets;
+	unsigned int max_seg_size;
 	unsigned int flags;
 	u32 i;
 	int ret;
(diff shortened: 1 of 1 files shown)


drivers/dma/dw-axi-dmac/dw-axi-dmac-platform.c
```diff
@@ -1585,9 +1586,21 @@ static int dw_probe(struct platform_device *pdev)
...
+
 	dw->dma.dev->dma_parms = &dw->dma_parms;
-	dma_set_max_seg_size(&pdev->dev, MAX_BLOCK_SIZE);
+	max_seg_size = UINT_MAX;
+	for (i = 0; i < dw->hdata->nr_channels; i++) {
+		unsigned int block_size = chip->dw->hdata->block_size[i];
+
+		if (!block_size)
+			block_size = MAX_BLOCK_SIZE;
+		max_seg_size = min(chip->dw->hdata->block_size[i], max_seg_size);
```
popcornmix: should this be `min(block_size, max_seg_size)`?

pelwell: Yes, really - it's the result of a last minute change

pelwell: Fixed.

popcornmix (approved):

Issue #6266, opened 2024-07-13
RPi 3B+ cannot boot after commit e6c1e86


iucoen:
### Describe the bug

After updating to commit e6c1e86 my system no longer boots. I use Gentoo and OpenRC gets stuck at trying to mount the root file system. The commit in question seems to be relevant.

### Steps to reproduce the behaviour

- Install kernel at commit e6c1e86
- Boot process gets stuck at trying to mount the root filesystem

### Device (s)

Raspberry Pi 3 Mod. B+

### System

Gentoo aarch64 (musl + OpenRC)

P33M:
That commit should be a no-op on a 3B+. It doesn't support CQHCI?

Please post your full kernel .config and the console output from boot.

iucoen:
I'm using bcm2711_defconfig, which has CONFIG_MMC_CQHCI=y. Full MMC related configs:

```
CONFIG_MMC=y
CONFIG_PWRSEQ_EMMC=y
CONFIG_MMC_BLOCK=y
CONFIG_MMC_BLOCK_MINORS=32
# CONFIG_MMC_TEST is not set
# MMC/SD/SDIO Host Controller Drivers
CONFIG_MMC_BCM2835_MMC=y
CONFIG_MMC_BCM2835_DMA=y
CONFIG_MMC_BCM2835_PIO_DMA_BARRIER=2
CONFIG_MMC_BCM2835_SDHOST=y
# CONFIG_MMC_DEBUG is not set
# CONFIG_MMC_ARMMMCI is not set
CONFIG_MMC_SDHCI=y
CONFIG_MMC_SDHCI_IO_ACCESSORS=y
# CONFIG_MMC_SDHCI_PCI is not set
CONFIG_MMC_SDHCI_PLTFM=y
# CONFIG_MMC_SDHCI_OF_ARASAN is not set
# CONFIG_MMC_SDHCI_OF_AT91 is not set
CONFIG_MMC_SDHCI_OF_DWCMSHC=m
# CONFIG_MMC_SDHCI_CADENCE is not set
# CONFIG_MMC_SDHCI_F_SDH30 is not set
# CONFIG_MMC_SDHCI_MILBEAUT is not set
CONFIG_MMC_SDHCI_IPROC=y
# CONFIG_MMC_TIFM_SD is not set
CONFIG_MMC_SPI=m
# CONFIG_MMC_CB710 is not set
# CONFIG_MMC_VIA_SDMMC is not set
[...]

PR #6268, opened 2024-07-15
Fixup SD posted write limit and an upstream regression

P33M:
See #6266 - Pi 0-3 got broken, and probably non-A2 cards on Pi 5 as well.

Also the error `mmc0: cannot verify signal voltage switch` seen in https://forums.raspberrypi.com/viewtopic.php?t=367459&start=50#p2236478 was the result of an upstream change.

### DIFF
--- a/drivers/mmc/core/mmc.c
+++ b/drivers/mmc/core/mmc.c
@@ -1663,6 +1663,7 @@ static int mmc_init_card(struct mmc_host *host, u32 ocr,
 		card->ocr = ocr;
 		card->type = MMC_TYPE_MMC;
 		card->rca = 1;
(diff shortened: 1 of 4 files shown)


nullr0ute:
Do you happen to have a upstream commit ID that caused the regression for reference? I don't see it referenced here or in the forum link.

P33M:
Fun with github links - it referenced my fork.
The upstream change included in rpi-6.6.y is https://github.com/raspberrypi/linux/commit/485d9421719baae063d6622c9455e8d17e920af2
That references upstream commit d77dc388cd61dfdafe30b98025fa827498378199.

PR #6269, opened 2024-07-15
Rpi 6.10.y Unicam fixups

6by9:
@naushir libcamera still won't run against this as the upstream Unicam driver has changed structure, and libcamera's pipeline handler hasn't been updated to match.

With these changes I can run the couple of devices I've tried in the older video-node mode.


6by9:
Rebased, and a simple final patch added to swap the compatible strings so that we're still using the old downstream driver whilst we wait for libcamera to catch up and support multiplexed streams.

Our standard libcamera is working on 6.10 with these changes.

naushir:
LGTM, we can remove the last patch when we are finally ready in userland.

6by9:
> LGTM, we can remove the last patch when we are finally ready in userland.

That was the plan - put it together in the way we eventually want it, and then fix it up so it works now!

PR #6271, opened 2024-07-16
Revert "Update DAC8x to support 384khz"

sirfragles:
Reverts raspberrypi/linux#6187
Detailed Commit Information:

- Reason for Reverting: The update to support 384kHz for DAC8x resulted in instability during audio playback at such a high bitrate. To ensure consistent and stable audio performance, we are reverting this change.

- Affected File: sound/soc/bcm/rpi-simple-soundcard.c

- Changes Made:
   - Reverted the sample rate limit from 384kHz back to 192kHz.
   - Modified the playback rate definition accordingly.

### DIFF
--- a/sound/soc/bcm/rpi-simple-soundcard.c
+++ b/sound/soc/bcm/rpi-simple-soundcard.c
@@ -324,10 +324,10 @@ static int hifiberry_dac8x_init(struct snd_soc_pcm_runtime *rtd)
 	struct snd_soc_dai *codec_dai = asoc_rtd_to_codec(rtd, 0);
 
 	/* override the defaults to reflect 4 x PCM5102A on the card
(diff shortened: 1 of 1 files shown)


pelwell:
Why is it necessary to remove the ability to support 384kHz? Is not using it not sufficient?

sirfragles:
The reason for removing the ability to support 384kHz is due to stability issues. Playback at such a high bitrate has proven to be unstable. Additionally, the designer from HiFiBerry specifically requested that I revert the changes due to these instability concerns. In the future, I plan to introduce a dtparam option to allow users to enable 384kHz support if they choose to experiment with it.

pelwell:
> the designer from HiFiBerry specifically requested that I revert the changes

That's the kind of information to put in a PR.

I'll merge it.

PR #6272, opened 2024-07-17
arm64: dts: Add cm5l files

pelwell:
CM5 Lite DTBs require minor changes compared to the "heavy" variants.


pelwell:
Do these look OK to you, @6by9 (mainly the former commit) and @timg236 (mainly the latter)?

6by9:
Moving 2712 to arm64 is fine by me.
The upstream DT (should be in 6.11) is in arm64.

timg236:
> Do these look OK to you, @6by9 (mainly the former commit) and @timg236 (mainly the latter)?

Sorry missed this yesterday. It looks ok to me

PR #6274, opened 2024-07-19
added support for attpm20p

David-P-SP:
Was testing with LetsTrust TPM, needed to add this so it can be used.


pelwell:
That's an unmodified upstream file that isn't specific to Raspberry Pi. In this situation we ask users to send contributions to the kernel maintainers. However, fortunately for you, this oversight has already been spotted and fixed: 0543f29408a151c1c4a12e5da07ec45c2779b9b8

Since that commit is marked with a Fixes: tag, I expect that to be backported fairly soon to both 6.10 and 6.9.

Issue #6275, opened 2024-07-20
configs: enable CONFIG_BRIDGE_VLAN_FILTERING


t0xicCode:
### Describe the bug

This has been asked before on the forums (https://forums.raspberrypi.com/viewtopic.php?t=341675) but not here yet.

Debian/Ubuntu have `CONFIG_BRIDGE_VLAN_FILTERING` enabled with `CONFIG_VLAN_8021q` configured as a loadable module, so we could do the same (and `8021q` is already setup to be a loadable module).

Enabling this would only slightly increase the uncompressed kernel size, but would allow bridges to be automatically network aware, which isn't the case at the moment.

### Steps to reproduce the behaviour

Install Raspberry Pi OS, do `grep VLAN_FILTERING /boot/config*`.
Result is `#CONFIG_BRIDGE_VLAN_FILTERING is not set` for every kernel config file.

### Device (s)

Raspberry Pi 4 Mod. B

### System

`/etc/rpi-issue`
```
Raspberry Pi reference 2023-12-11
Generated using pi-gen, https://github.com/RPi-Distro/pi-gen, 2acf7afcba7d11500313a7b93bb55a2aae20b2d6, stage2
```

`vcgencmd version`
```
Apr 17 2024 17:27:09 
Copyright (c) 2012 Broadcom
version 86ccc427f35fdc604edc511881cdf579df945fb4 (clean) (release) (start)
```

`uname -a`
```
Linux husk 6.6.28+rpt-rpi-v8 #1 SMP PREEMPT Debian 1:6.6.28-1+rpt1 (2024-04-22) aarch64 GNU/Linux
```

PR #6277, opened 2024-07-20
configs: rpi: enable CONFIG_BRIDGE_VLAN_FILTERING

t0xicCode:
Enabling CONFIG_BRIDGE_VLAN_FILTERING allows bridges to automatically receive and forward packets based on the 802.1q VLAN information and the device/port configuration.

See: #6275


pelwell:
Interestingly, this makes no clear difference in memory usage, at least when testing with no VLANs configured. The uncompressed image size is unchanged, and the compressed image is a few bytes smaller.

PR #6279, opened 2024-07-23
pinctrl: rp1: jump through hoops to avoid PCIe latency issues

P33M:
Automatic link power saving plus the ability of a root complex to buffer pending posted write transfers (and consider them complete before being transmitted on the wire) causes compression of updates to GPIO state.

The large bandwidth of a Gen 2 x4 link means the writes toggle state inside RP1 as fast as it can go (~20MHz), which is bad for applications wanting bitbash with at least a few microseconds of delay between updates.

By tailoring IO access patterns to a special Root Complex register, writes to GPIOs can be stalled until the link wakes - meaning all writes end up with a reasonably consistent minimum pacing (~200ns).

Additionally, write barriers have no effect other than to arbitrarily delay some writes by a small, variable amount - so remove the vast majority of these in areas that could be hot-paths.

Although the IO memory is mapped with Device strongly-ordered semantics, this doesn't prevent the splitter inside BCM2712 from letting an MMIO read request to a GPIO register get ahead of the pacing writes to the Root Complex register. So each pin state read must flush writes out to the Outer-Shareable domain.

### DIFF
--- a/drivers/pinctrl/pinctrl-rp1.c
+++ b/drivers/pinctrl/pinctrl-rp1.c
@@ -197,6 +197,7 @@ struct rp1_pin_info {
 	void __iomem *inte;
 	void __iomem *ints;
 	void __iomem *pad;
+	void __iomem *dummy;
 };
 
 enum funcs {
@@ -276,6 +277,7 @@ struct rp1_pinctrl {
 	void __iomem *gpio_base;
 	void __iomem *rio_base;
 	void __iomem *pads_base;
+	void __iomem *dummy_base;
 	int irq[RP1_NUM_BANKS];
 	struct rp1_pin_info pins[RP1_NUM_GPIOS];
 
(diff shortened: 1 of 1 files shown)


drivers/pinctrl/pinctrl-rp1.c
```diff
@@ -1592,10 +1667,23 @@ static int rp1_pinctrl_probe(struct platform_device *pdev)
 	pinctrl_add_gpio_range(pc->pctl_dev, &pc->gpio_range);
 
 	return 0;
+out_iounmap:
```
pelwell, line 1672: I think this would deserve a preceding blank line, but see devm_ comment.

drivers/pinctrl/pinctrl-rp1.c
```diff
@@ -652,32 +690,33 @@ static void rp1_set_fsel(struct rp1_pin_info *pin, u32 fsel)
 		FLD_SET(ctrl, RP1_GPIO_CTRL_OEOVER, RP1_OEOVER_PERI);
 	}
 	FLD_SET(ctrl, RP1_GPIO_CTRL_FUNCSEL, fsel);
-	writel(ctrl, pin->gpio + RP1_GPIO_CTRL);
+
```
pelwell: I don't see a reason to insert this blank line.

drivers/pinctrl/pinctrl-rp1.c
```diff
@@ -1528,6 +1568,38 @@ static int rp1_pinctrl_probe(struct platform_device *pdev)
...
+	 *
+	 * of_iomap() is used directly here as the address overlaps with the RC driver's
+	 * usage.
+	 */
+	rp1_node = of_find_node_by_name(NULL, "rp1");
+	if (!rp1_node)
+		dev_err(&pdev->dev, "failed to find RP1 DT node\n");
+	else if (pace_pin_updates &&
+		 of_device_is_compatible(rp1_node->parent, "brcm,bcm2712-pcie")) {
+		pc->dummy_base = of_iomap(rp1_node->parent, 0);
```
pelwell, line 1597: Use devm_of_iomap(), and the error handling becomes simpler.

P33M: I don't know why I didn't think there'd be a devm variant...

pelwell: It's probably a recent addition.

P33M: It goes through the "check if the resource range is free" step, which causes a failure. This has bitten at least one other attempted conversion - 
https://lkml.iu.edu/hypermail/linux/kernel/2006.1/02364.html

But this block stil needs to warn on failure, as otherwise you're getting the broken write pacing without knowing why.

pelwell: I'm disappointed in the kernel devs - why not give it an "exclusive" flag, etc? So much for the principal of least surprise.

P33M: Ho hum. Warning added.

pelwell (approved):

PR #6280, opened 2024-07-23
Align AXI DMAC configuration and peripheral client use

P33M:
This is a bit of an omni-PR, but it solves two important issues:
- TV hat unreliability on Pi 5
- Hardware limitations of the DMAC causing mismatches between client expectations of bytes transferred vs descriptor lengths

In the first case, the limitations of the DMAC severely restrict SPI throughput so operating a) in a unidirectional mode b) with corrected burst lengths are both required. Doing b) means that other peripheral drivers using the DMAC need to set burst thresholds (particularly RX) correctly.

### DIFF
--- a/drivers/dma/dw-axi-dmac/dw-axi-dmac-platform.c
+++ b/drivers/dma/dw-axi-dmac/dw-axi-dmac-platform.c
@@ -261,6 +261,15 @@ static u32 axi_chan_get_xfer_width(struct axi_dma_chan *chan, dma_addr_t src,
 	return __ffs(src | dst | len | BIT(max_width));
 }
 
+static u32 axi_dma_encode_msize(u32 max_burst)
+{
+	if (max_burst <= 1)
+		return DWAXIDMAC_BURST_TRANS_LEN_1;
+	if (max_burst > 1024)
+		return DWAXIDMAC_BURST_TRANS_LEN_1024;
+	return fls(max_burst) - 2;
+}
+
 static inline const char *axi_chan_name(struct axi_dma_chan *chan)
 {
 	return dma_chan_name(&chan->vc.chan);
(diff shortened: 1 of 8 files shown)


drivers/dma/dw-axi-dmac/dw-axi-dmac-platform.c
```diff
@@ -261,6 +261,15 @@ static u32 axi_chan_get_xfer_width(struct axi_dma_chan *chan, dma_addr_t src,
 	return __ffs(src | dst | len | BIT(max_width));
 }
 
+static u32 axi_dma_encode_msize(u32 max_burst)
+{
+	if (max_burst == 0)
+		return DWAXIDMAC_BURST_TRANS_LEN_1;
+	if (max_burst > 1024)
+		return DWAXIDMAC_BURST_TRANS_LEN_1024;
+	return fls(max_burst) - 2;
```
P33M, line 270: This will underflow for a max_burst of 1 - not typically set by drivers (using 0 to denote "no bursts please" instead) but will fix.

drivers/spi/spi-dw-core.c
```diff
@@ -200,7 +200,17 @@ int dw_spi_check_status(struct dw_spi *dws, bool raw)
...
-		dw_spi_reset_chip(dws);
+		/*
+		 * Forcibly halting the controller can cause DMA to hang.
+		 * Defer to dw_spi_handle_err outside of interrupt context
+		 * and mask further interrupts for the current transfer.
+		 */
+		if (dws->dma_mapped) {
+			dw_spi_mask_intr(dws, 0xff);
+			dw_readl(dws, DW_SPI_ICR);
+		} else
```
pelwell: Sorry, but the kernel coding style is to use braces in both paths or neither.

Issue #6281, opened 2024-07-24
Raspberry Pi 5 HDMI interlaced does not work properly


zoli0726:
### Describe the bug

Since Pi 5 DSI does not support interlaced anymore I went towards the route of using an HDMI->VGA dongle to produce 15khz signal for my CRT tv.

240p works very well, however 480i and 576i looks to be missing one of the fields, means the signal looks like  progressive 240p with slight flickering and scanlines.
I checked whether the dongle might be an issue, but on my PC with the same modeline it produces proper 480i or 576i.

I messed around in 

`static void vc5_hdmi_set_timings(struct vc4_hdmi *vc4_hdmi,
				 struct drm_connector_state *state,
				 const struct drm_display_mode *mode)`

especiall with 

`	HDMI_WRITE(HDMI_VERTA0, verta);
	HDMI_WRITE(HDMI_VERTA1, verta);

	HDMI_WRITE(HDMI_VERTB0, vertb_even);
	HDMI_WRITE(HDMI_VERTB1, vertb); `

and looks like some of these registers do not have any effect, writing them to 0 changes absolutely nothing about the image. (if I remember correctly, HDMI_VERTB1) , I'm wondering whether it might be, that these are not backwards compatible with Pi4, because the code itself looks like was inherited from Pi4. 
	

### Steps to reproduce the behaviour

Connected to HDMI-A-2
I loaded a custom edid using cmdline.txt.
Set the resolution to the interlace mode the edid supports. Tested it with Raspberry OS and Batocera 40 beta build.

extended cmdline.txt: 

drm.edid_firmware=HDMI-A-2:edid-hdmi-2.bin

extended config.txt: 

hdmi_edid_file=1
hdmi_edid_filename:0=edid-hdmi-1.bin
hdmi_edid_filename:1=edid-hdmi-2.bin

### Device (s)

Raspberry Pi 5

### System

Raspberry Pi OS with desktop
Release date: July 4th 2024
System: 64-bit
Kernel version: 6.6
Debian version: 12 (bookworm)

### Additional context

I can upload my edid file somewhere if needed.

popcornmix:
> hdmi_edid_file=1
hdmi_edid_filename:0=edid-hdmi-1.bin
hdmi_edid_filename:1=edid-hdmi-2.bin

remove these - config.txt settings apply to the deprecated firmware driver.

Can you confirm if you are booting to console or desktop? Note that wayland doesn't support interlaced modes.
[...]

PR #6282, opened 2024-07-25
drm/vc4: Disable the 2pixel/clock odd timings workaround for interlaced

6by9:
Whilst BCM2712 does fix using odd horizontal timings, it doesn't work with interlaced modes.

Drop the workaround for interlaced modes and revert to the same behaviour as BCM2711.

https://github.com/raspberrypi/linux/issues/6281


6by9:
It may be possible to tweak other settings to retain the odd timings support with interlacing, but this seemed the more expedient solution for now.

PR #6284, opened 2024-07-25
drm/vc4: Add support for per plane scaling filter selection

6by9:
Seeing as the HVS can be configured with regard the scaling filter, and DRM now supports selecting scaling filters at a per CRTC or per plane level, we can implement it.

Default remains as the Mitchell/Netravali filter, but nearest neighbour is now also implemented.

https://forums.raspberrypi.com/viewtopic.php?p=2239565

Draft whilst I evaluate whether it is really that useful. The original discussion where the standard property was added was suggesting that the emulation folks wanted it to give the really blocky graphics.

### DIFF
--- a/drivers/gpu/drm/vc4/vc4_drv.h
+++ b/drivers/gpu/drm/vc4/vc4_drv.h
@@ -358,6 +358,7 @@ struct vc4_hvs {
 	struct work_struct free_dlist_work;
 
 	struct drm_mm_node mitchell_netravali_filter;
+	struct drm_mm_node nearest_neighbour_filter;
 
 	struct debugfs_regset32 regset;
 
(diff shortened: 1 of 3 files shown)


popcornmix:
I think you also want to set the "no interpolate" bit (bit 31 of horiz/vert polyphase filter word 0) in dlist
```
Disable Coefficient Interpolation
1 =	Do not interpolate coefficients to 64 phases; just use the 8 given phases.  This is required to implement a true nearest neighbour kernel.
0 =	Interpolate coefficients to 64 phases.
```

6by9:
> I think you also want to set the "no interpolate" bit (bit 31 of horiz/vert polyphase filter word 0) in dlist
> 
> ```
> Disable Coefficient Interpolation
> 1 =	Do not interpolate coefficients to 64 phases; just use the 8 given phases.  This is required to implement a true nearest neighbour kernel.
> 0 =	Interpolate coefficients to 64 phases.
> ```

Agreed. PR updated.

pelwell:
This looks OK to me.

popcornmix (approved):

6by9:
Dropped draft tag. It's a standard DRM property, so we may as well implement it.

Issue #6285, opened 2024-07-26
PR 6280 breaks SPI communication and wm5102 detection on RPi5


HiassofT:
### Describe the bug

With current rpi-update kernel the WM5102 soundchip on the Cirrus Logic Audio Card is no longer detected, dmesg shows `arizona spi0.0: Unknown device ID: 0` which is typical for a SPI comminication issue. I suspect other SPI devices might be affected, too.

Bisecting lead me to the two "Save bandwidth with the TMOD_TO/RO" commits 6014649 and cd9084c, with those two reverted the chip is detected fine.

Just reverting the second TMOD_RO commit does not help and reverting only the first TMOD_TO one isn't easy because of shared context.

Checking with my scope it looks like the initial address (RPi output) phase is terminated too early, instead of the expected 48 clock cycles (1 bit r/w, 31bit address, 16 bit padding) I only see 21 cycles. In the following 16 (RPi input) clock cycles the WM5102 doesn't yet drive MISO as it's still waiting for the missing address cycles and so the RPi reads a device id of 0 (instead of 0x5102).
![broken](https://github.com/user-attachments/assets/6a45afa1-b0d9-4ee2-83ee-f06b1c0069bb)

With the two TMOD commits reverted the SPI communication works fine:
![ok](https://github.com/user-attachments/assets/feeb0e72-b3f7-4e9f-a63d-c12bb9cb704c)

### Steps to reproduce the behaviour

Use a Cirrus Logic Audio Card on a RPi5, just enabling the `cirrus-wm5102` should also the SPI issue (detecting the chip ID by reading register 0 is the first thing the arizona-spi driver does).

### Device (s)

Raspberry Pi 5

### System

RPi OS bookworm, 64bit, kernel 
```
Linux raspberrypi 6.6.42-v8-16k+ #1787 SMP PREEMPT Thu Jul 25 17:45:51 BST 2024 aarch64 GNU/Linux
```

### Logs

https://paste.debian.net/hidden/4840050a/

### Additional context

ping @pelwell

pelwell:
#6286 should get it working again, but I don't have one of those devices (or many other SPI devices at all).

pelwell:
Thanks for the early report. #6286 is now merged - please close this issue if you are satisfied.

HiassofT:
Thanks a lot for the quick fix, testing looks good here so I'm happy to close the issue

PR #6286, opened 2024-07-29
spi: dw: Fix non-DMA transmit-only transfers

pelwell:
Ensure the transmit FIFO has emptied before ending the transfer by dropping the TX threshold to 0 when the last byte has been pushed into the FIFO. Include a similar fix for the non-IRQ paths.

See: https://github.com/raspberrypi/linux/issues/6285
Fixes: 6014649de765 ("spi: dw: Save bandwidth with the TMOD_TO feature")

### DIFF
--- a/drivers/spi/spi-dw-core.c
+++ b/drivers/spi/spi-dw-core.c
@@ -220,6 +220,32 @@ int dw_spi_check_status(struct dw_spi *dws, bool raw)
 }
 EXPORT_SYMBOL_NS_GPL(dw_spi_check_status, SPI_DW_CORE);
 
+static inline bool dw_spi_ctlr_busy(struct dw_spi *dws)
+{
+	return dw_readl(dws, DW_SPI_SR) & DW_SPI_SR_BUSY;
+}
+
+static enum hrtimer_restart dw_spi_hrtimer_handler(struct hrtimer *hr)
+{
+	struct dw_spi *dws = container_of(hr, struct dw_spi, hrtimer);
+
+	if (!dw_spi_ctlr_busy(dws)) {
+		spi_finalize_current_transfer(dws->host);
+		return HRTIMER_NORESTART;
+	}
+
+	if (!dws->idle_wait_retries) {
+		dev_err(&dws->host->dev, "controller stuck at busy\n");
+		spi_finalize_current_transfer(dws->host);
+		return HRTIMER_NORESTART;
+	}
+
+	dws->idle_wait_retries--;
+	hrtimer_forward_now(hr, dws->idle_wait_interval);
+
+	return HRTIMER_RESTART;
+}
+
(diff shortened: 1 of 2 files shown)


HiassofT:
It is a bit better now, but still not quite right, the TX transfer is often missing one or two clock cycles, see eg this screenshot where TX phase was one cycle short and the ID read back as 2881 instead of 5102:
pr-2881

I've also seen one case where the kernel reported 1440 (so 2 cycles missing) and another one where the ID was read fine but a later SPI transfer failed

pelwell:
Although the RP1 SPI interface supports transmit-only, it lacks an interrupt-on-idle. When the TX-FIFO-empty interrupt fires, there may be some data left in the TX shift register. Spinning in interrupt context is not recommended, and by the next data is ready to be sent it may not even be necessary to wait, so we  can replace any post-TX-wait with a pre-transfer idle-wait.

If this works for you I'll squash it with the previous patch.

HiassofT:
Still no joy, in case of a single write-only transfer chip-select is de-asserted too early:
pr-write-failed

pelwell:
I was hoping not to have to resort to a timer, but I can't think of an alternative with decent performance.

Third time lucky?

drivers/spi/spi-dw-core.c
```diff
@@ -246,7 +272,16 @@ static irqreturn_t dw_spi_transfer_handler(struct dw_spi *dws)
 		}
 	} else if (!dws->tx_len) {
 		dw_spi_mask_intr(dws, DW_SPI_INT_TXEI);
-		spi_finalize_current_transfer(dws->host);
+		if (dw_spi_ctlr_busy(dws)) {
+			dws->idle_wait_retries = 8;
```
P33M: If this is an analogue for bits per data element aka "word width", SPI can be configured up to 32. This value should be in one of the TX configuration registers...

pelwell: There should be somewhere more accessible.

pelwell: And there was - `dws->n_bytes`.

HiassofT:
Thanks a lot for the update, testing with the cirrus logic audio card looks fine so far (also on the scope)

pelwell:
Thanks for the feedback. I'm trying to tune the wait intervals - exponential backoff feels better than up to 32 timer interrupts - but otherwise this feels like the right fix.

pelwell:
Apart from squashing the commits, this is the release candidate. It should reduce the number of interrupts taken to detect the end-of-transfer.

pelwell:
Now tested over a wide range of speeds, during the course of which I discovered that a speed of 3kHz or lower can lead to effectively random speeds due to clock divisor overflow - now fixed in the second commit.

PR #6292, opened 2024-07-31
drivers: drm: rp1-dsi: Implement more DSI options and flags

njhollinghurst:
Now implementing:
- Per-command selection of LP or HS for commands (previously LP)
- EoTp transmission option (previously EoTp was always disabled)
- Non-continuous clock option (previously always continuous)
- Per-command enabling of ACK request (in command mode only)

Make a plausible (and possibly correct) attempt to measure the longest LP command that will fit into vertical blanking lines.

DON'T set hardware's "Burst Mode" flag: since this driver locks the DPI and DSI rates together, it shouldn't matter; but it seems Burst Mode may break if there isn't time to return to LP11 in HFP.

### DIFF
--- a/drivers/gpu/drm/rp1/rp1-dsi/rp1_dsi_dsi.c
+++ b/drivers/gpu/drm/rp1/rp1-dsi/rp1_dsi_dsi.c
@@ -103,6 +103,24 @@
 
 /* And some bitfield definitions */
 
+#define DSI_PCKHDL_EOTP_TX_EN  BIT(0)
+#define DSI_PCKHDL_BTA_EN      BIT(2)
+
+#define DSI_VID_MODE_LP_CMD_EN        BIT(15)
+#define DSI_VID_MODE_FRAME_BTA_ACK_EN BIT(14)
+#define DSI_VID_MODE_LP_HFP_EN        BIT(13)
+#define DSI_VID_MODE_LP_HBP_EN        BIT(12)
+#define DSI_VID_MODE_LP_VACT_EN       BIT(11)
+#define DSI_VID_MODE_LP_VFP_EN        BIT(10)
+#define DSI_VID_MODE_LP_VBP_EN        BIT(9)
+#define DSI_VID_MODE_LP_VSA_EN        BIT(8)
+#define DSI_VID_MODE_SYNC_PULSES      0
+#define DSI_VID_MODE_SYNC_EVENTS      1
+#define DSI_VID_MODE_BURST            2
+
+#define DSI_CMD_MODE_ALL_LP           0x10f7f00
+#define DSI_CMD_MODE_ACK_RQST_EN      BIT(1)
+
 #define DPHY_PWR_UP_SHUTDOWNZ_LSB 0
 #define DPHY_PWR_UP_SHUTDOWNZ_BITS BIT(DPHY_PWR_UP_SHUTDOWNZ_LSB)
 
@@ -1252,8 +1270,8 @@ static u32 dphy_configure_pll(struct rp1_dsi *dsi, u32 refclk, u32 vco_freq)
 			       vco_freq, actual_vco_freq, m, refclk, n,
 			       hsfreq_table[dsi->hsfreq_index].hsfreqrange);
 	} else {
-		drm_warn(dsi->drm,
-			 "rp1dsi: Error configuring DPHY PLL %uHz\n", vco_freq);
+		drm_err(dsi->drm,
+			"rp1dsi: Error configuring DPHY PLL %uHz\n", vco_freq);
 	}
 
 	return actual_vco_freq;
@@ -1321,7 +1339,7 @@ static void rp1dsi_dpiclk_start(struct rp1_dsi *dsi, u32 byte_clock,
 	clk_set_rate(dsi->clocks[RP1DSI_CLOCK_DPI], (4 * lanes * byte_clock) / (bpp >> 1));
 	clk_prepare_enable(dsi->clocks[RP1DSI_CLOCK_DPI]);
 	drm_info(dsi->drm,
-		 "rp1dsi: Nominal Byte clock %u DPI clock %lu (parent rate %lu)",
+		 "rp1dsi: Nominal Byte clock %u DPI clock %lu (parent rate %lu)\n",
 		 byte_clock,
 		 clk_get_rate(dsi->clocks[RP1DSI_CLOCK_DPI]),
 		 clk_get_rate(clk_get_parent(dsi->clocks[RP1DSI_CLOCK_DPI])));
@@ -1365,7 +1383,8 @@ static u32 get_colorcode(enum mipi_dsi_pixel_format fmt)
 
 void rp1dsi_dsi_setup(struct rp1_dsi *dsi, struct drm_display_mode const *mode)
 {
-	u32 timeout, mask, vid_mode_cfg;
+	int cmdtim;
+	u32 timeout, mask, clkdiv;
 	unsigned int bpp = mipi_dsi_pixel_format_to_bpp(dsi->display_format);
 	u32 byte_clock = clamp((bpp * 125 * min(mode->clock, RP1DSI_DPI_MAX_KHZ)) / dsi->lanes,
 			       RP1DSI_BYTE_CLK_MIN, RP1DSI_BYTE_CLK_MAX);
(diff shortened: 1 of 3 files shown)


njhollinghurst:
Some combination of these changes seems to make the LT070ME05000 display work with Pi 5. The [old] 7" panel still works too.

May need tidying and has a little too much debug printing, hence Draft PR.

I still need to investigate if/when it's appropriate to use Burst Mode and non-matching DPI and DSI pixel rates and video modes.

njhollinghurst:
LT070ME05000 is still not totally reliable... on rare occasions it doesn't seem to work after boot (possibly because commands are not transmitted correctly?)

Other times it comes up with the wrong brightness; this seems to be a side-effect of failing to read back the brightness in the previous session, causing it to be clipped at 5% ever after.

drivers/gpu/drm/rp1/rp1-dsi/rp1_dsi_dsi.c
```diff
@@ -1471,6 +1516,28 @@ void rp1dsi_dsi_send(struct rp1_dsi *dsi, u32 hdr, int len, const u8 *buf)
...
+		if (use_lpm)
+			val |= DSI_VID_MODE_LP_CMD_EN;
+		else
+			val &= ~DSI_VID_MODE_LP_CMD_EN;
+		DSI_WRITE(DSI_VID_MODE_CFG, val);
+		val = (use_lpm) ? DSI_CMD_MODE_ALL_LP : 0;
+		if (req_ack)
+			val |= DSI_CMD_MODE_ACK_RQST_EN;
+		DSI_WRITE(DSI_CMD_MODE_CFG, val);
+		(void)(0 + DSI_READ(DSI_CMD_MODE_CFG));
```
6by9: Huh? Reading, adding 0, casting it to void, and doing nothing with it.

drivers/gpu/drm/rp1/rp1-dsi/rp1_dsi_dsi.c
```diff
@@ -1471,6 +1516,28 @@ void rp1dsi_dsi_send(struct rp1_dsi *dsi, u32 hdr, int len, const u8 *buf)
 		usleep_range(100, 150);
 	}
 
+	/*
+	 * Update global configuration flags for LP/HS and ACK options.
+	 * XXX It's not clear if having empty FIFOs (checked above and below) guarantees that
+	 * the last command has completed and been ACKed, or how closely these control registers
+	 * align with command/payload FIFO writes (as each is an independent clock-crossing)?
+	 */
+	if (use_lpm != dsi->using_lpm || req_ack != dsi->using_ack) {
```
6by9: Do we have to retain the state and avoid rewriting the registers if unchanged? What's the downside of making it unconditional?

drivers/gpu/drm/rp1/rp1-dsi/rp1_dsi_dsi.c
```diff
@@ -1504,8 +1571,10 @@ int rp1dsi_dsi_recv(struct rp1_dsi *dsi, int len, u8 *buf)
 			break;
 		usleep_range(100, 150);
 	}
-	if (i == 0)
+	if (i == 0) {
```
6by9: `if (!i)`

drivers/gpu/drm/rp1/rp1-dsi/rp1_dsi_dsi.c
```diff
@@ -1531,5 +1600,6 @@ void rp1dsi_dsi_stop(struct rp1_dsi *dsi)
 
 void rp1dsi_dsi_set_cmdmode(struct rp1_dsi *dsi, int mode)
 {
+	drm_info(dsi->drm, "Change to %s Mode\n", mode ? "Command" : "Video");
```
6by9: Left over debug log?

drivers/gpu/drm/rp1/rp1-dsi/rp1_dsi_dsi.c
```diff
@@ -103,6 +103,24 @@
 
 /* And some bitfield definitions */
 
+#define DSI_PCKHDL_EOTP_TX_EN  BIT(0)
```
6by9, line 106: Toss up as to whether it is better to put the bitfields alongside the register defines but indented, eg https://elixir.bootlin.com/linux/v6.10/source/drivers/gpu/drm/vc4/vc4_regs.h#L131

6by9 (commented):
Just a quick read through. Otherwise I'll trust that you're doing the right thing.

drivers/gpu/drm/rp1/rp1-dsi/rp1_dsi_dsi.c
njhollinghurst: It forces a round-trip delay, for superstitious reasons (because Synopsys docs show multiple FIFOs straddling a clock domain boundary); the `0 +` is yet more superstition as I was fuzzy about the distinction between mentioning and using an lvalue in 'C'. Probably unnecessary!

Maybe @P33M has an opinion about that kind of thing?

drivers/gpu/drm/rp1/rp1-dsi/rp1_dsi_dsi.c
njhollinghurst: Only a slight speed penalty.

drivers/gpu/drm/rp1/rp1-dsi/rp1_dsi_dsi.c
njhollinghurst: Yes... I thought would be harmless to leave in, but since there are hardly any other messages to order it with it's probably not useful to see it.

drivers/gpu/drm/rp1/rp1-dsi/rp1_dsi_dsi.c
P33M: readl() forces the compiler to emit a load + load barrier instruction pair which means a round-trip across PCIe has to complete. Minimum latency 1.2us. PCIe ordering prevents the read request overtaking a previous write.

P33M: Generally, I'd expect a to find a comment as to why a register read is there where the return value is not used.

njhollinghurst: I actually do not know if it's needed or not.

I'm trying to look at the dsi_host RTL but have not yet tracked down how these (LP, ENA_ACK) signals are synchronized from the APB to the byteclock domain. I do hope they're not assumed to be constant. So far I suspect there's more risk the flags may be applied to a previous transmission rather than fail to be applied to the new one....

drivers/gpu/drm/rp1/rp1-dsi/rp1_dsi_dsi.c
njhollinghurst: It's actually far from clear that it's safe to change these registers while running, although the Linux API seems to require it. That said, writing them with the _same_ values again is safe enough so that is no reason not to do so unconditionally.

drivers/gpu/drm/rp1/rp1-dsi/rp1_dsi_dsi.c
njhollinghurst: I think I'll leave them where they are for now, otherwise I'd have to move the other (DPHY_* etc) definitions as well.

njhollinghurst:
This has now been verified as beneficial on two formerly-not-working panels (`lt07me05000-v2` and `microtips-st7701`).

There are two outstanding questions but I don't think they should hold up this PR:
1. When and how to use Burst Mode. It could be helpful in cases when (8 * lanes > bpp), by allowing us to match DSI and DPI clocks; currently we don't exploit it but we do set the hardware flag (which changes a FIFO threshold) when it's requested.
2. Safety of per-command LP/HS (and ACK) selection. I think it's _formally_ not safe (unsynchronized clock domain crossings into complex logic), but Linux seems to require it: e.g. lt070me05000 needs LP commands for setup and HS commands for backlight control. Assuming drivers don't change frequently between the two, it seems a net benefit to implement it.

drivers/gpu/drm/rp1/rp1-dsi/rp1_dsi_dsi.c
njhollinghurst: I've removed the `0 +`.

pelwell:
Is there any reason not to merge this?

6by9:
All good by me.

PR #6293, opened 2024-08-01
dtoverlays: Add overlay for HD44780 via I2C PCF8574 backpack

6by9:
Many HD44780 LCD displays are connected via very common I2C GPIO expander.
We have an overlay for connecting the displays directly to GPIOs, but not one for it connected via a backpack. Add such an overlay.

Requested https://forums.raspberrypi.com/viewtopic.php?t=374408

### DIFF
--- a/arch/arm/boot/dts/overlays/README
+++ b/arch/arm/boot/dts/overlays/README
@@ -1752,9 +1752,9 @@ Params: pin_d4                  GPIO pin for data pin D4 (default 6)
         pin_bl                  Optional pin for enabling/disabling the
                                 display backlight. (default disabled)
 
-        display_height          Height of the display in characters
+        display_height          Height of the display in characters (default 2)
 
-        display_width           Width of the display in characters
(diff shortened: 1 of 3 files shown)


arch/arm/boot/dts/overlays/README
```diff
@@ -1705,6 +1705,33 @@ Params: gpio_pin                GPIO pin to trigger on (default 3)
...
+        pin_en                  GPIO pin for "Enable" (default 2)
+
+        pin_rs                  GPIO pin for "Register Select" (default 0)
+
+        pin_rw                  GPIO pin for R/W select (default 1)
+
+        pin_bl                  GPIO pin for enabling/disabling the display
+                                backlight. (default 3)
+
+        display_height          Height of the display in characters
```
pelwell: The dimensions have default values but they aren't documented?

6by9: They weren't documented for hd44780-lcd on which this was based.

Now documented in both.

PR #6296, opened 2024-08-02
drivers: media: pci: Update Hailo accelerator device driver to v4.18.0

naushir:
Sourced from https://github.com/hailo-ai/hailort-drivers/


pelwell:
checkpatch.pl throws up a lot of whitespace warnings, but I doubt we're going to get them fixed.

Feel free to merge when you feel ready.

naushir:
Need to wait for a bit until we get userland in shape. Kernel driver v4.18 is incompatible with the v4.17 userland libraries that we currently ship.

PR #6299, opened 2024-08-04
Updated OTG audio drivers to make all string descriptors configurable through configfs

rithwikmakesthings:
When using the Raspberry Pi as an OTG Audio device, it currently presents itself to the host device as "Playback Inactive" and "Capture Inactive" (Particularly true when plugged into a MacOS host device). There is no easy way for a user to change this name without making minor edits to the driver and recompiling the kernel- an all in all tedious and time consuming process for configuring something so simple as a device name.

The updates in this PR allows users to set custom names of terminals/controls/alt modes at the time of setup, in the same place as other options already available to them, while still leaving the previous options as defaults if not configured. The device name can now be specified in place of the previously default 'Capture Inactive' and 'Playback Inactive',

### DIFF
--- a/drivers/usb/gadget/function/f_uac2.c
+++ b/drivers/usb/gadget/function/f_uac2.c
@@ -104,25 +104,10 @@ enum {
 	STR_AS_OUT_ALT1,
 	STR_AS_IN_ALT0,
 	STR_AS_IN_ALT1,
+	NUM_STR_DESCRIPTORS,
 };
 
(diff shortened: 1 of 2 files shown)


popcornmix:
It looks your PR changes code that is untouched in the downstream pi kernel tree.

As such, you'll need to [submit this upstream](https://docs.kernel.org/process/submitting-patches.html).
Once it's merged upstream, we can pick it.

PR #6301, opened 2024-08-05
DTS: bcm2712: enable SD slot CQE by default on Pi 5

P33M:
I think this is stable enough across cards in the retail channel that we can flip the switch, i.e. all known-bad cards are ignored.

There's two unresolved questions
- What happens to the sd_cqe parameter in overlays/README, now that it does nothing?
- Does the new disable wart end up in overlays, or is it documented (somewhere) as an instruction to put `sdhci.debug_quirks=` in cmdline.txt? How have we historically handled this?

### DIFF
--- a/arch/arm/boot/dts/overlays/README
+++ b/arch/arm/boot/dts/overlays/README
@@ -378,9 +378,9 @@ Params:
                                 non-lite SKU of CM4).
                                 (default "on")
 
-        sd_cqe                  Use to enable Command Queueing on the SD
-                                interface for faster Class A2 card performance
-                                (Pi 5 only, default "off")
+        sd_cqe                  Set to "off" to disable Command Queueing if you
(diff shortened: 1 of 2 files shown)


pelwell:
Although it isn't possible for an overlay to delete a property, it is possible for a parameter to delete a property in its scope. Since the property `supports-cqe;` and the parameter `sd_cqe` are in the same scope, through the magic of inverted boolean overrides  we can avoid resorting to command line parameters:
```
		sd_cqe = <&sdio1>, "supports-cqe!";
```

pelwell:
...where for `sd_cqe` you should, of course, read `no_sd_cqe`, i.e.:
```
		no_sd_cqe = <&sdio1>, "supports-cqe!";
```

pelwell (approved):
And changing the default and documentation is even easier.

PR #6309, opened 2024-08-12
drm/vc4: Add a delay after disabling hdmi phy output

popcornmix:
There appears to be a requirement for some devices (I'm testing with a 8K VRROOM 40Gbps HDMI switch)
for a measable delay between removing the hdmi phy output from the old mode, to enabling the hdmi phy output for the new mode.

Without the delay, a mode switch has a small change of getting a permanent 'no signal', which requires a subsequent mode switch or a unplug/replug to redetect.

Switching between 4kp24/25/30 modes fails about 5% of time in my testing.

Add a delay to make it impossible to switch faster than this.


popcornmix:
`msleep 15` is not sufficient.
Currently testing `msleep 20`

popcornmix:
A bit more background.

I can reproduce "no signal" on Pi4 and Pi5.
It doesn't occur with fkms driver on Pi4.

Originally reported by @HiassofT (who uses the vrroom switch) and gets black screen when kodi stops playing a video (typically switching between 1920x1200@59.91 and 1920x1080@60).

6by9:
It feels a bit of a kludge to work around certain hardware, but without a significant delve into what's going on I'll agree that this is probably the most pragmatic route forward.

I can't immediately see anything in the HDMI spec that puts any timing requirements on stopping and restarting.

drivers/gpu/drm/vc4/vc4_crtc.c
```diff
@@ -668,6 +668,7 @@ static void vc4_crtc_atomic_disable(struct drm_crtc *crtc,
 	 * someone was waiting it.
 	 */
```
6by9, line 669: Commit text:
s/measable/measurable

popcornmix:
The original fix works for pi4. On pi5 I was still getting failures, and some digging revealed the code in `phy_disable` is ineffective on pi5 (for two reasons).

I've also lower the original delay from 30ms to 20ms as that passed an overnight switching test.

6by9 (commented):
Looks plausible.

Seeing as this wants to be folded into the original patch when upstreaming, don't worry about the typo unless you need to respin for other reasons.

popcornmix:
We'll wait for confirmation from @HiassofT that it fixes his issue (he can test at end of week).

Also be interesting to hear if it fixes @chewitt's issue with switching from 4kp50 to 1080p60 on a (JVC?) TV
(although his issue may be this or something else).

chewitt:
JVC? .. it's not the '80s :)

This patch resolves the issue I've had switching from 4K@50 back to 1080@60 on my LG B7V 👍🏻

popcornmix:
Good to hear.

HiassofT:
Thanks a lot, this changes seem to fix the "no signal" issue on my VRROOM here.

RPi3B+, RPi4 and RPi5 all managed to do over 1000 1200p60->1080p60->1200p60 modeswitch cycles without any "no signal" issue (testing with LibreELEC/kodi, playing Big Buck Bunny and stopping it). Without this PR I'd usually get a "no signal" condition after 20-50 cycles

popcornmix:
Okay to merge?

PR #6311, opened 2024-08-14
Add overlay for a generic I2S clock-master DAC

elagil:
This adds support for any generic stereo I2S DAC that acts as a clock master on the bus.
The feature is very useful when the audio sink cannot synchronize to the clock of the RaspberryPi and does not have a sample rate converter. In that case, the only solution is to use the clock from the domain of the DAC.

Tested with an STM32 as I2S clock master. 

This borrows from `audioinjector-bare-i2s-overlay.dts`, but makes the DAC the clock master, instead of the CPU.


elagil:
Addressed pipeline issues

elagil:
Made a mistake with the sign-off address :face_with_diagonal_mouth:

elagil:
Thanks for the quick merge!

PR #6312, opened 2024-08-14
drm/vc4: Allow option to transpose the output on the writeback connector

6by9:
The hardware behind the writeback connector can do a transpose as it writes the image out.
DRM offers no mechanism to select that, but as a simple test we can note when the width/height of the buffer are swapped wrt the mode, and enable the transpose. The atomic_check obviously has to be updated too to permit that combination.

Transposing a square output buffer is left as future problem to solve.

### DIFF
--- a/drivers/gpu/drm/drm_connector.c
+++ b/drivers/gpu/drm/drm_connector.c
@@ -361,7 +361,8 @@ static int __drm_connector_init(struct drm_device *dev,
 
 	drm_object_attach_property(&connector->base,
 				   config->non_desktop_property,
-				   0);
+				   (connector_type != DRM_MODE_CONNECTOR_VIRTUAL &&
+				   connector_type != DRM_MODE_CONNECTOR_WRITEBACK) ? 0 : 1);
 	drm_object_attach_property(&connector->base,
 				   config->tile_property,
 				   0);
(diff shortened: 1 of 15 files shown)


6by9:
@jc-kynesim PR updated. It also includes all the other pending DRM tweaks that are kicking around for you/DT.

6by9:
JC has said he's happy with this, so drop the draft tag.

The checkpatch complaint on spaces before tabs is deliberate as it matches all the other comments in that block.

Branch rebased.

drivers/gpu/drm/drm_connector.c
```diff
@@ -362,7 +362,7 @@ static int __drm_connector_init(struct drm_device *dev,
 	drm_object_attach_property(&connector->base,
 				   config->non_desktop_property,
 				   (connector_type != DRM_MODE_CONNECTOR_VIRTUAL &&
-				   connector_type != DRM_MODE_CONNECTOR_WRITEBACK) ? 0 : 1;
+				   connector_type != DRM_MODE_CONNECTOR_WRITEBACK) ? 0 : 1);
```
popcornmix, line 365: This should be squashed in previous commit.

6by9: Done.

PR #6314, opened 2024-08-16
drm: rp1: rp1-dsi: Add DRM_FORMAT_ABGR8888

JanKehren:
Add support for DRM_FORMAT_ABGR8888.
This is required to use the driver in Android.

### DIFF
--- a/drivers/gpu/drm/rp1/rp1-dpi/rp1_dpi.c
+++ b/drivers/gpu/drm/rp1/rp1-dpi/rp1_dpi.c
@@ -260,6 +260,8 @@ static struct drm_driver rp1dpi_driver = {
 static const u32 rp1dpi_formats[] = {
 	DRM_FORMAT_XRGB8888,
 	DRM_FORMAT_XBGR8888,
+	DRM_FORMAT_ARGB8888,
+	DRM_FORMAT_ABGR8888,
 	DRM_FORMAT_RGB888,
 	DRM_FORMAT_BGR888,
 	DRM_FORMAT_RGB565
--- a/drivers/gpu/drm/rp1/rp1-dpi/rp1_dpi_hw.c
+++ b/drivers/gpu/drm/rp1/rp1-dpi/rp1_dpi_hw.c
@@ -258,6 +258,18 @@ static const struct rp1dpi_ipixfmt my_formats[] = {
 	  .rgbsz  = BITS(DPI_DMA_RGBSZ_BPP, 3),
 	},
 	{
+	  .format = DRM_FORMAT_ARGB8888,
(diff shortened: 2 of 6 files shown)


6by9:
Thanks for the patch, however this needs some discussion.

Stating that an alpha channel is supported on a primary plane makes little sense as there will be nothing underneath(*), so it seems a slightly daft thing for Android to be requiring.
The RP1 hardware certainly doesn't support any form of alpha, and indeed only has one plane.

There are a couple of formats that are defined as mandatory for DRM, but I can't find the reference for those at present. If Android is requiring this format, it'd be worth getting that list updated upstream.

If we do accept this then you should add DRM_FORMAT_ARGB8888 as well for symmetry.

(*) Patches for setting background colour have been mooted, but never merged

JanKehren:
Thanks for a quick review.

I added DRM_FORMAT_ARGB8888 for the symmetry. 

I would also agree that Android behavior makes little sense but the default behavior of Android giving out stuff in HAL_PIXEL_FORMAT_RGBA_8888 which translates in minigbm to DRM_FORMAT_ABGR8888.

6by9:
Having worked on Android systems in previous jobs, I have no expectations of it doing anything sensible!

I'm of an opinion to accept this as it solves an issue, even if logically it makes little sense. I'll let others respond though

@njhollinghurst any thoughts?

njhollinghurst:
It seems reasonable if it solves a compatibility problem. (We have to assume of course that Alpha==0xFF or is safe to ignore.)

njhollinghurst:
DPI and VEC as well?

6by9:
> We have to assume of course that Alpha==0xFF or is safe to ignore.

We only have the one plane on the RP1 display devices, so there can never be a plane underneath.
Memory says that DRM does document that there is assumed to be an opaque black layer at the bottom of the stack should the lowest plane not cover the whole screen.
If the alpha isn't 0xff then we will do the wrong thing (the black should be reflected through), but the hardware can't do that.

> VEC as well?

This change could be made for DPI and VEC as well as they'll have the same requirements under Android. I wasn't going to chase for it.

I was looking at other drivers to compare. Things like the tiny DRM drivers largely do not to support ABGR, with arcpgu being the exception. That programs the register with the same value for XRGB8888 and ARGB8888 (https://github.com/raspberrypi/linux/blob/rpi-6.6.y/drivers/gpu/drm/tiny/arcpgu.c#L133-L136). I guess that can be taken as prior art.

JanKehren:
I have added this for DPI and VEC driver.

njhollinghurst:
Thank you

PR #6316, opened 2024-08-19
configs: Add I2C_HID_OF to Pi defconfigs

6by9:
https://forums.raspberrypi.com/viewtopic.php?t=375242

In the 5.12 kernel cycle I2C_HID was split into I2C_HID_OF and I2C_HID_ACPI, under a parent of the old I2C_HID config option that previously supported OF.

The Pi defconfigs didn't get updated, so I2C_HID stopped working.

Enable I2C_HID_OF to bring back that functionality.


pelwell:
Thanks - there is now a mechanism in place to catch the loss of config settings, but this one predates it.

PR #6318, opened 2024-08-20
ASoC: Add support for the HiFiBerry ADC sound card

j-schambacher:
Adds the driver, the DT overlays, UUID hat definition and the defconfigs for the new HiFiBerry ADC sound card.
A separate driver keeps the possibility for further updates with enhanced clock management etc.


j-schambacher:
@pelwell Shall I fix the 'style errors'? One seems to be the GPL vs GPL v2 in the driver, the other is another missing info beyond
"Say Y or M ... for Hifiberry ADC." in the Kconfig. checkpatch did indeed complain about the last one - besides 40 others with the same issue. Just let me know, what ever is the easiest/best for you. Also, please let me know any other issue/concern you have with the PR.

pelwell:
I'm not bothered about the cursory Kconfig help info or the GPL issues. Provided you stick to the kernel coding guidelines and the other checks pass, I'm happy.

j-schambacher:
Tried to fix it all, but I'm lost now. It didn't change anything. Maybe, I'm blindly focused on the wrong stuff ;-). Any hints?

pelwell (commented):
It's complaining that the README entry declares parameters but the Load: line doesn't mention them - just append `,<param>=<val>`.

j-schambacher:
@pelwell Thanks, Phil! I didn't find that.
To avoid the Kconfig warning(s) the info/help paragraph needs to have 4 lines or more. Nice idea, but probably not really useful for these simple cards.

pelwell:
Why do you need the hat_map entry? It's intended to be a get-out-of-jail mechanism for existing cards with bad built-in overlays - new cards should be programmed with the name of the overlay.

j-schambacher:
I was under the wrong impression that all hats need to be in that list. I've removed it.

PR #6319, opened 2024-08-20
DRM patches for @cillian64

6by9:
@cillian64 

I'm still looking for my patch that assigns the extra planes for the exclusive use of the writeback connector.


6by9:
Extra patch found.

`sudo rpi-update pulls/6319` once the CI pipeline has run.

PR #6325, opened 2024-08-26
media: i2c: imx290: Add support for High Conversion Gain

velkoborsky:
This update adds support for High Conversion Gain (HCG) mode, which significantly improves low-light performance.

According to the IMX290 and IMX462 datasheets, the address 3009h is configured as follows:

- **Bits 0-1: "FRSEL [1:0]"**: Frame rate (Data rate) setting
- **Bit 4: "FDG_SEL"**: Conversion gain switching; 0: LCG Mode, 1: HCG Mode
- **Other bits**: Fixed to 0

The frame rate setting has been implemented in the driver; however, conversion gain switching has not yet been implemented.

### DIFF
--- a/drivers/media/i2c/imx290.c
+++ b/drivers/media/i2c/imx290.c
@@ -41,6 +41,7 @@
 #define IMX290_WINMODE_720P				(1 << 4)
 #define IMX290_WINMODE_CROP				(4 << 4)
 #define IMX290_FR_FDG_SEL				CCI_REG8(0x3009)
+#define IMX290_60FPS_LCG				0x01
 #define IMX290_BLKLEVEL					CCI_REG16_LE(0x300a)
 #define IMX290_GAIN					CCI_REG8(0x3014)
 #define IMX290_VMAX					CCI_REG24_LE(0x3018)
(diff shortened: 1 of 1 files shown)


6by9:
We've been here already - #5859 / https://forums.raspberrypi.com/viewtopic.php?t=363696, and #4593 before it.

Adding new V4L2 controls in a downstream repo is a maintenance nightmare, which generally means it's a fairly hard "no".
Controls also need useful names, and I've still not actually seen a decent explanation of what this conversion gain property on IMX327/290/462 does.
As in #5859, a module parameter is generally the least painful route, although it does limit it to being a global option for all users of the module on a system (ie shared by 2 imx290's connected to the same system).


I'd not noticed the comment on the forum thread that says that the CG isn't being applied on first startup. It looks like I've got a clash between `imx290_set_data_lanes` and `imx290_set_ctrl(V4L2_CID_ANALOGUE_GAIN)` both trying to set the same `IMX290_FR_FDG_SEL` register. Odd as the control handler should come second.
I'm short on time at present, but have just rebased #5859 so that CI reruns. I may see if I can find 5 mins to try it out later in the week.

velkoborsky:
Thank you for reviewing the patch. A module parameter would indeed be a useful compromise. I will test #5859.

For the explanation of the HCG - even in the datasheet there is no description of what HCG is doing. Practically, it allows you to capture low light signals with much less noise. We are using it for scientific purposes, greatly improving signal-to-noise ratio. 

There are some good measurements on the following website: https://player-one-astronomy.com/product/mars-color-camera/ The effect of HCG is visible as a step change in the readout noise.

PR #6328, opened 2024-08-28
rtc: pcf8523: Fix oscillator stop bit handling reading from Control_1 register

Paladinking:
The check if the oscillator stop bit is set was reading from Control_1 register instead of the Seconds register. This caused the Seconds register to be incorrectly changed if bit 7 of Control_1 happens to be set. 

Without this, the pcf8523 driver is completely broken, as reading from the clock causes the value to change by up to 1 minute.

### DIFF
--- a/drivers/rtc/rtc-pcf8523.c
+++ b/drivers/rtc/rtc-pcf8523.c
@@ -108,10 +108,10 @@ static int pcf8523_rtc_read_time(struct device *dev, struct rtc_time *tm)
 	if (err < 0)
 		return err;
 
-	if ((regs[0] & PCF8523_CONTROL1_STOP) || (regs[3] & PCF8523_SECONDS_OS))
+	if (regs[PCF8523_REG_CONTROL1] & PCF8523_CONTROL1_STOP)
 		return -EINVAL;
 
-	if (regs[0] & PCF8523_SECONDS_OS) {
+	if (regs[PCF8523_REG_SECONDS] & PCF8523_SECONDS_OS) {
 		/*
 		 * If the oscillator was stopped, try to clear the flag. Upon
 		 * power-up the flag is always set, but if we cannot clear it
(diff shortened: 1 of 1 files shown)


drivers/rtc/rtc-pcf8523.c
```diff
@@ -108,10 +108,10 @@ static int pcf8523_rtc_read_time(struct device *dev, struct rtc_time *tm)
 	if (err < 0)
 		return err;
 
-	if ((regs[0] & PCF8523_CONTROL1_STOP) || (regs[3] & PCF8523_SECONDS_OS))
+	if ((regs[0] & PCF8523_CONTROL1_STOP))
```
pelwell: Drop the extra level of parentheses.

drivers/rtc/rtc-pcf8523.c
```diff
@@ -108,10 +108,10 @@ static int pcf8523_rtc_read_time(struct device *dev, struct rtc_time *tm)
...
 		return -EINVAL;
 
-	if (regs[0] & PCF8523_SECONDS_OS) {
+	if (regs[3] & PCF8523_SECONDS_OS) {
```
pelwell: You could replace `3` with `PCF8523_REG_SECONDS` here and in the rest of this `if` block.

pelwell (commented):
You're right - it's just using the wrong register.  In general I approve of your patch, but see the inline comments, and the warnings from the automatic builds (https://github.com/raspberrypi/linux/actions/runs/10592534573/job/29356327864#step:6:76).

Paladinking:
Okay, should be better now.

> You're right - it's just using the wrong register. In general I approve of your patch, but see the inline comments, and the warnings from the automatic builds (https://github.com/raspberrypi/linux/actions/runs/10592534573/job/29356327864#step:6:76).

pelwell:
That looks good to me. I'll just wait for the auto-builds to complete before merging.

pelwell:
You can ignore the remaining checkpatch warning:
```
ERROR: Missing Signed-off-by: line by nominal patch author 'Axel <48924884+Paladinking@users.noreply.github.com>'
```

pelwell:
In future this can be squashed into `rtc: pcf8523: Fix oscillator stop bit handling` (the irony of the name is not lost on me...)

PR #6334, opened 2024-09-04
Added waveshare 5.0inch, 6.25inch, and 8.8inch dsi screen support

waveshare:


6by9:
Thanks for the PR.

[README](https://github.com/raspberrypi/linux/blob/rpi-6.6.y/arch/arm/boot/dts/overlays/README#L5214-L5242) needs to be updated for the new overrides, which is what the dtoverlay checks are complaining about.

I'm happy to ignore the checkpatch complaint as it's only for a missing commit message.

The URLs in the driver patch don't work - I assume they just aren't live yet.

waveshare:
1. The following content has been added to the README:
```
5_0_inch              5.0" 720x1280
6_25_inch            6.25" 720x1560
8_8_inch              8.8" 480x1920
```
2. Yes, they are not online yet

6by9:
Ideally the dtoverlay and README changes should be squashed into the same commit. (The driver change remains as a separate commit).
I'll leave it up to @pelwell to say if he wants that fixed or not.

I'm happy otherwise.

pelwell:
> Ideally the dtoverlay and README changes should be squashed into the same commit.

You would get extra "credit" for that -  it reduces the commit count by one on rpi-6.6.y, but all the DTS patches get squashed together on later branches so it doesn't matter that much.

pelwell:
Merged, but no gold star.

Issue #3263, opened 2019-10-02
fbtft broken on kernels newer than 5.0


satmandu:
I am loading the [waveshare35a.dts](https://github.com/swkim01/waveshare-dtoverlays) overlay in config.txt thus to get a lcd screen connected:
```
[all]
device_tree_address=0x03000000
enable_uart=1
dtparam=spi=on
dtoverlay=pi3-disable-bt
dtoverlay=waveshare35a:rotate=270,speed=21000000,debug=1
dtoverlay=vc4-kms-v3d
dtparam=i2c_arm=on
initramfs initrd.img followkernel
```
cmdline.txt:
```
net.ifnames=0 dwc_otg.lpm_enable=0 console=tty1 root=/dev/mmcblk0p2 rootfstype=ext4 elevator=deadline rootwait fsck.mode=force fsck.repair=yes fbcon=map:10 fbcon=font:ProFont6x11 logo.nologo nosplash cma=128M@256M
```

This works with the testing kernel8.img kernel at https://github.com/Hexxeh/rpi-firmware/tree/master
This also works with the ubuntu 19.10 5.0.0-1017-raspi2 arm64 kernel.
This does not work with a rpi-5.3.y kernel from today. (Kernel is modified with BCM2708_VCMEM so that I get vcdbg messages.)

Working ubuntu 5.0.0-1017-raspi2 kernel:
5.0.0-1017-raspi2 dmesg: https://paste.ubuntu.com/p/GKPnqw86MC/
5.0.0-1017-raspi2 config: https://paste.ubuntu.com/p/mT7RDnYQS3/
Broken:
5.3.0-v8-gb03118b64 dmesg: https://paste.ubuntu.com/p/f8tjHTPk2Z/
5.3.0-v8-gb03118b64 kernel config: https://paste.ubuntu.com/p/gyYyRM8jJH/

```
sudo vcgencmd version 
Sep  3 2019 23:52:29 
Copyright (c) 2012 Broadcom
version 6dc34ddf57a9cb5e234d00a4490d0b6a31688526 (clean) (release) (start)

sudo vcdbg log msg
Failed to allocate -1716518942 bytes for message buffer

sudo dtoverlay -d /boot/firmware/overlays -l
No overlays loaded

uname -a
Linux bridge 5.0.0-1017-raspi2 #17-Ubuntu SMP Thu Sep 12 14:25:40 UTC 2019 aarch64 aarch64 aarch64 GNU/Linux
```

[...]

PR #6337, opened 2024-09-04
piscreen-overlay: Add invx, invy and swapxy

satmandu:
This PR adds touchscreen x,y swap and inversion options, using the `invx`, `invy`, and `swapxy` as per the discussion at https://github.com/raspberrypi/linux/issues/3263?notification_referrer_id=MDE4Ok5vdGlmaWNhdGlvblRocmVhZDEyMjAwNTIwNzExOjEwOTY3MDE%3D#issuecomment-2327607601 , which will ideally work for the Waveshare 3.5a touchscreen.

I have changed the parameter names to match the common parameter name usage in other device tree overlay files in the same dir.

### DIFF
--- a/arch/arm/boot/dts/overlays/README
+++ b/arch/arm/boot/dts/overlays/README
@@ -3694,6 +3694,12 @@ Params: speed                   Display SPI bus speed
         drm                     Select the DRM/KMS driver instead of the FBTFT
                                 one
 
+        invx                    Touchscreen inverted x axis
+
+        invy                    Touchscreen inverted y axis
+
+        swapxy                  Touchscreen swapped x y axis
+
 
 Name:   piscreen2r
 Info:   PiScreen 2 with resistive TP display by OzzMaker.com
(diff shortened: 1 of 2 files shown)


satmandu:
The dtbo appears to work for me. For whatever reason with the way the screen is oriented on the wavshare35a, I had to use the `invx` parameter to get the `invy` behavior I wanted (the touchscreen responding as expected to roughly where I poke at the screen.)

Is this because of the default `swapxy`?

Should I flip things so the settings are thus @6by9 ?
```
		invx =		<&piscreen_ts>,"touchscreen-inverted-y?";
		invy =		<&piscreen_ts>,"touchscreen-inverted-x?";
```

6by9:
> Is this because of the default `swapxy`?

Almost certainly.
With swapxy and invx/invy it matters which order the two operations are done in. It should be documented (somewhere) which order the two operations are done, and that is the expected behaviour.

> Should I flip things so the settings are thus @6by9 ?
> 
> ```
> 		invx =		<&piscreen_ts>,"touchscreen-inverted-y?";
> 		invy =		<&piscreen_ts>,"touchscreen-inverted-x?";
> ```

No, because then using invx without swapxy would give the wrong result.

satmandu:
Happy to make documentation changes here as you might suggest.

Otherwise I'm happy to leave this as is, as this completes the functionality I desire.

6by9:
It is already documented - https://github.com/raspberrypi/linux/blob/rpi-6.6.y/Documentation/devicetree/bindings/input/touchscreen/touchscreen.yaml#L73-L76
```
  touchscreen-swapped-x-y:
    description: X and Y axis are swapped
                 Swapping is done after inverting the axis
    type: boolean
```

I'm happy.
I'll leave it with @pelwell to make any further comments and merge.

PR #6339, opened 2024-09-05
7inch panel ts tweaks.

6by9:
For @lurch in his testing.
@spl237 for reference.

(Hyperpixel4 updated as it happens to use the same touch controller and I had one I could test with).

### DIFF
--- a/drivers/input/touchscreen/goodix.h
+++ b/drivers/input/touchscreen/goodix.h
@@ -78,6 +78,7 @@ struct goodix_chip_data {
 struct goodix_ts_data {
 	struct i2c_client *client;
 	struct input_dev *input_dev;
+	u32 num_keys;
 	struct input_dev *input_pen;
 	const struct goodix_chip_data *chip;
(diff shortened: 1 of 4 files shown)


drivers/input/touchscreen/goodix.h
```diff
@@ -78,6 +78,7 @@ struct goodix_chip_data {
 struct goodix_ts_data {
 	struct i2c_client *client;
 	struct input_dev *input_dev;
+	u32 num_keys;
```
pelwell, line 81: Do you envision this value being used outside the configure function? As it stands, it could just be a local.

pelwell:
As discussed elsewhere today, there's no reason for two overlay patches. Aside from that and the inline questions, LGTM.

lurch:
This doesn't actually fix the problems I was having, so this probably shouldn't be merged yet. I've emailed Dave with the details.

6by9:
Closing - Simon's fixing it in arandr.

PR #6340, opened 2024-09-05
drivers: drm: rp1-vec: Increase width limit, for PAL 16:9 @ 18MHz

njhollinghurst:
There was no technical reason for the DRM mode's width limit of 848; increase it to 960 (720*18MHz/13.5MHz) to support ~square pixels on 16:9 screens.

Tweak the PAL active window to start slightly earlier. (The maximum number of visible columns at 18MHz is about 942.)


njhollinghurst:
This begs the question: Should this (or any other) limit be increased further? I don't think it would be useful. The hardware limit is something like 1023 columns. However, 960 already exceeds the horizontal active window at 18MHz.

The next convenient pixel rate is 21.6MHz which would fall foul of the hardware limit for the 625/25 and 525/50 systems (not to mention bandwidth), and is really supported only for System E.

PR #6344, opened 2024-09-09
media: i2c: ov5647: Add control of V4L2_CID_HBLANK

6by9:
The driver did expose V4L2_CID_HBLANK, but as a READ_ONLY control.

The sensor only uses the HTS register to control the line length, so convert this control to read/write, with the appropriate ranges. Adopt the old fixed values as the minimum values permitted in each mode to avoid issues of it not streaming.

This should allow exposure times up to ~3 seconds (up from ~1sec).

### DIFF
--- a/drivers/media/i2c/ov5647.c
+++ b/drivers/media/i2c/ov5647.c
@@ -53,6 +53,8 @@
 #define OV5647_REG_AEC_AGC		0x3503
 #define OV5647_REG_GAIN_HI		0x350a
 #define OV5647_REG_GAIN_LO		0x350b
+#define OV5647_REG_HTS_HI		0x380c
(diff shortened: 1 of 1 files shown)


6by9:
https://forums.raspberrypi.com/viewtopic.php?p=2251691

6by9:
@naushir to review.

Verified good images at 1/3fps.
I didn't check that the exposure control would also go all that way up, but I don't believe that it won't -  it's controlled in lines, and we've only extended the duration of each line rather than adding any extras.

naushir:
Looks good to me.

6by9:
Reported as working in https://forums.raspberrypi.com/viewtopic.php?p=2252037#p2252037

PR #6345, opened 2024-09-09
DRM Writeback connector priority changes

6by9:
@jc-kynesim

These should get the 10bittest working on Pi3.

### DIFF
--- a/drivers/gpu/drm/vc4/vc4_kms.c
+++ b/drivers/gpu/drm/vc4/vc4_kms.c
@@ -674,17 +674,26 @@ static int vc4_load_tracker_atomic_check(struct drm_atomic_state *state)
 	for_each_oldnew_plane_in_state(state, plane, old_plane_state,
 				       new_plane_state, i) {
 		struct vc4_plane_state *vc4_plane_state;
+		struct vc4_crtc *vc4_crtc;
 
 		if (old_plane_state->fb && old_plane_state->crtc) {
 			vc4_plane_state = to_vc4_plane_state(old_plane_state);
-			load_state->membus_load -= vc4_plane_state->membus_load;
-			load_state->hvs_load -= vc4_plane_state->hvs_load;
+			vc4_crtc = to_vc4_crtc(old_plane_state->crtc);
+
+			if (!vc4_crtc->feeds_txp) {
+				load_state->membus_load -= vc4_plane_state->membus_load;
+				load_state->hvs_load -= vc4_plane_state->hvs_load;
+			}
 		}
 
 		if (new_plane_state->fb && new_plane_state->crtc) {
 			vc4_plane_state = to_vc4_plane_state(new_plane_state);
-			load_state->membus_load += vc4_plane_state->membus_load;
-			load_state->hvs_load += vc4_plane_state->hvs_load;
+			vc4_crtc = to_vc4_crtc(new_plane_state->crtc);
+
+			if (!vc4_crtc->feeds_txp) {
+				load_state->membus_load += vc4_plane_state->membus_load;
+				load_state->hvs_load += vc4_plane_state->hvs_load;
+			}
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


jc-kynesim:
Good oh! Thanks. Do you want me to test (I'm only going to do the obvious) or are you happy that you've already done that?

drivers/gpu/drm/vc4/vc4_kms.c
```diff
@@ -241,15 +241,22 @@ static void vc4_hvs_pv_muxing_commit(struct vc4_dev *vc4,
...
+		 * TXP can also run with a lower panic level than a live display,
+		 * as it doesn't have the same real-time constraint.
 		 */
-		if (vc4_crtc->feeds_txp)
-			dsp3_mux = VC4_SET_FIELD(3, SCALER_DISPCTRL_DSP3_MUX);
-		else
-			dsp3_mux = VC4_SET_FIELD(2, SCALER_DISPCTRL_DSP3_MUX);
+		if (vc4_crtc->feeds_txp) {
+			dsp3_mux_pri = VC4_SET_FIELD(3, SCALER_DISPCTRL_DSP3_MUX);
+			dsp3_mux_pri |= VC4_SET_FIELD(1, SCALER_DISPCTRL_PANIC2);
```
popcornmix: Any reason for this being 1 (Use panic mode for first two of the frame or if empty) rather than 0 (Fixed priority of 2’b00 (never panic).) ?

6by9: Copied from the firmware (middleware/hvs/hvs.c)
```
   static const uint32_t priority_0 = 0x2;  // panic when <2 lines in FIFO
   static const uint32_t priority_1 = 0x2;  // panic when <2 lines in FIFO
   static const uint32_t priority_2 = 0x1;  // raised but no panic
```

If you think we can work with never panic, then I'm happy to switch it to that.

6by9:
> Good oh! Thanks. Do you want me to test (I'm only going to do the obvious) or are you happy that you've already done that?

Give it the quick tests that you were looking at - I've only done `10bittest -P NV12 -w`. I don't see it being affected by anything else, but you never know.

As per my email, there seems to be another issue lurking:
I have just hit a funny one in that I had a DSI display connected. That took HVS channel 2, which is the only one that can feed the transposer. 10bittest then couldn't start as it couldn't get the appropriate channel :-(
The channel allocation code in the kernel was always a little funky and made some assumptions, so it may want to be revisited.

I wasn't going to try fixing that under this PR, but may limit our ability to use writeback generically. That's possibly of more interest to DT than you, but thought it worth flagging.

drivers/gpu/drm/vc4/vc4_kms.c
popcornmix: The full options are:
```
0 = Fixed priority of 2’b00 (never panic).
1 = Use panic mode for first two of the frame or if empty.
2 = Use panic mode whenever less than two complete lines in FIFO.
3 = Fixed priority of 2’b11 (always panic).
```

I don't see this as real time, so I don't think it needs to panic ever.

But there could be a concern that a non-panicking memory access from a plane being fetched for writeback will hold up a following panicking access from a plane needed by HDMI, effectively killing it's priority (I don't believe there's any way to overtake).

6by9: I know very little on how the AXI priorities are going to interact, hence just copying the firmware settings.

Happy to take guidance on which to use.

popcornmix: I think logically 0 (never panic) makes sense - we are not real time when writing back, and don't want to interfere with actual real time masters like camera.

If we see the theoretical issue of low priority writeback accesses holding off subsequent high priority accesses needed by hdmi, then we can revisit (it will likely have to be the same as hdmi priority) - but I have no evidence of that actually occurring.

So, I'd recommend 0 for this PR.

6by9:
I'd forgotten about updating this PR based on popcornmix's request for priority 0. Now done.

popcornmix (approved):

PR #6350, opened 2024-09-11
DRM source rectangle fixes

6by9:
@cillian64  As discussed:
- fix for your sub 1 pixel source rectangle issue (even if we don't fully understand how it became <1pixel)
- potential patch to save LBM for your upscaled 1x1 solid colour fills. (I need to patch modetest or kmstest to allow me to test source rectangles easily)


6by9:
Converted to draft as using the TPZ filter really doesn't work at present, and needs to either be fixed (if possible) or dropped.

PR #6353, opened 2024-09-12
ASoC: add the HiFiBerry 8-channel ADC soundcard

j-schambacher:
Additions and changes for the new 8 channel ADC card for the Pi 5.
It's a HW controlled device, so we can use the snd-soc-dummy device
for the DAI declarations.

### DIFF
--- a/arch/arm/boot/dts/overlays/hifiberry-adc8x-overlay.dts
+++ b/arch/arm/boot/dts/overlays/hifiberry-adc8x-overlay.dts
@@ -1,0 +1,50 @@
+// Definitions for HiFiBerry ADC8x
+/dts-v1/;
+/plugin/;
+
+/ {
+	compatible = "brcm,bcm2712";
+
+	fragment@0 {
+		target = <&gpio>;
+		__overlay__ {
+			rp1_i2s0_adc8x: rp1_i2s0_adc8x {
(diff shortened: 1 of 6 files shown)


arch/arm/boot/dts/overlays/hifiberry-adc8x-overlay.dts
```diff
@@ -0,0 +1,51 @@
...
+		};
+	};
+
+	fragment@2 {
+		target-path = "/";
+		__overlay__ {
+			dummy-codec {
+				#sound-dai-cells = <0>;
+				compatible = "snd-soc-dummy";
+//				num-channels = <8>;
```
pelwell: Did you mean to leave this line commented out?

j-schambacher: Ooops! Forgotten to delete and checkpatch does not complain ;-) But yes, it is not necessary as it is ultimately set in the driver init. But I can fix it together with the 75 char issues.

pelwell: If you can do that soon I'll get it merged this afternoon before the next rpi-update build

j-schambacher:
Thanks!

PR #6358, opened 2024-09-13
mm/vmscan: Maintain TLB coherency in LRU code

pelwell:
As a workaround (and possibly a fix) for CPU spins observed on BCM2837, use ptep_clear_flush_young instead of ptep_test_and_clear_young inside lru_gen_look_around in order to expose PTE changes to the MMU. Note that on architectures that don't require an explicit flush, ptep_clear_flush_young just calls ptep_test_and_clear_young.

### DIFF
--- a/mm/vmscan.c
+++ b/mm/vmscan.c
@@ -4736,7 +4736,7 @@ void lru_gen_look_around(struct page_vma_mapped_walk *pvmw)
 		if (!folio)
 			continue;
 
-		if (!ptep_test_and_clear_young(vma, addr, pte + i))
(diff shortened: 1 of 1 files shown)


pelwell:
There is a mailing-list discussion about this issue: https://marc.info/?l=linux-mm&m=172614610701899&w=2

pelwell:
A timelapse camera test which is known to trigger the CPU spins has now reached 3000 frames in 5 hours without any significant slowdown (a monitoring application reported an unexpected 3 second delay, but there is no clear gap in the timestamps).

PR #6361, opened 2024-09-13
dts: 2712: Reduce default cma usage on Pi5

popcornmix:
Significant cma shouldn't really be needed on Pi5 as the hardware blocks support iommu and can access system memory.

We've migrated codec and camera support to system memory, and 3d has always (even on Pi4) used system memory.

A large cma block causes issues with enabling NUMA on a low memory (2G) Pi5, as cma cannot span numa regions.

### DIFF
--- a/arch/arm/boot/dts/overlays/vc4-kms-v3d-pi5-overlay.dts
+++ b/arch/arm/boot/dts/overlays/vc4-kms-v3d-pi5-overlay.dts
@@ -3,7 +3,7 @@
 #include "cma-overlay.dts"
 
 &frag0 {
-	size = <((320-4)*1024*1024)>;
+	size = <(64*1024*1024)>;
 };
 
 / {


arch/arm/boot/dts/overlays/vc4-kms-v3d-pi5-overlay.dts
```diff
@@ -3,7 +3,7 @@
 #include "cma-overlay.dts"
 
 &frag0 {
-	size = <((320-4)*1024*1024)>;
+	size = <(64*1024*1024)>;
```
pelwell, line 6: The fact that this is a multiple of 64MB means it will not be automatically capped to the platform-limit, but this is way below all the limits so not an issue.

popcornmix:
I'll leave this here for a while to see if anyone has objections.

I've tested dual 4k desktop use (with 3d active).
Kodi playing 4K hevc videos (desktop and standalone).
libcamera-vid records okay at 4608x2592.

arch/arm/boot/dts/overlays/vc4-kms-v3d-pi5-overlay.dts
popcornmix: Yes - I spent a while reading the firmware code and came to same conclusion.

popcornmix:
Removed draft tag

Issue #6365, opened 2024-09-17
Pi 5 UARTs broken in 6.6.44 build of rpi-6.6.y.


pelwell:
### Describe the bug

As detailed by @lgeitner in the thread starting at https://github.com/raspberrypi/linux/pull/6144#issuecomment-2349489287, the MeteoPi application which requires UART access seems to be broken in the 6.6.44 build of rpi-6.6.y. It has also been shown to have worked in the 6.6.42 build. The corresponding source commits for those two builds are:

"kernel: Bump to 6.6.42" (1969fd7): bfbd468e81defa70d4356a1a3e467cc5d67ca89b
"kernel: Bump to 6.6.44" (44a287f): 2806f39528464ddcc56f77dd8ea4d81f71b7e619

### Steps to reproduce the behaviour

Run MeteoPi. Other test cases have yet to be determined.

### Device (s)

Raspberry Pi 5

### System

`Linux 6.6.44-v8-16k+ #1789 SMP PREEMPT Mon Aug 5 15:24:18 BST 2024 aarch64 GNU/Linux`

### Logs

N/A

pelwell:
The current list of suspect commits is:
```
2806f39528464 Revert "mfd: Partial revert of Add rp1 driver"
d0ea54e1c941a mfd: Partial revert of Add rp1 driver
216df57950849 DTS: bcm2712: enable SD slot CQE by default on Pi 5
3c319a466a1c7 dtoverlays: Document display_[width|height] on hd44780-lcd overlay
e94e761305fa2 dtoverlays: Add overlay for HD44780 via I2C PCF8574 backpack
a4bf61fad9fe1 fixup! pinctrl: bcm2712 pinctrl/pinconf driver
16d0ee22d2c0b hwmon: (adt7410) Add DT compatible strings
05e3687c6c973 overlays: i2c-rtc: Correct bq32000 property name
e9294823cf020 spi: dw: Clamp the minimum clock speed
199e611183de0 spi: dw: Fix non-DMA transmit-only transfers
70636ad110715 ARM: dts: bcm2712: Fix invalid polling-delay-passive setting
485d11cfa7df2 drm/vc4: Disable the 2pixel/clock odd timings workaround for interlaced
062434ab3be76 dts: rp1: restrict i2s burst lengths to 4
b6b4260fa546d sound/soc: dwc-i2s: choose FIFO thresholds based on DMA burst constraints
[...]

PR #6366, opened 2024-09-17
Trial builds for Issue #6365

pelwell:
This PR will be used to create test builds while debugging Issue #6365 with @lgeitner.

This first build just reverts all the downstream patches since the known good build, with the exception of the final two which just move an item within a Makefile.


pelwell:
Following the discussion in #6365, this PR is now just a single revert on top of the current rpi-6.6.y tree. It will be merged later today unless there is negative feedback.

pelwell:
Superseded by #6369.

PR #6367, opened 2024-09-17
Fix use-after-free in rpisense-js driver

paulober:
Currently the driver calls `input_free_device` after `input_unregister_device` if it fails in the probe function or in the remove function. Because the `input_unregister_device` function frees the device at any time:

> Once device is unregistered the caller should not try to access it as it may get freed at any moment.

the `input_free_device` will cause a 
```console
refcount_t: underflow; use-after-free.
```

Linux kernel docs I use for reference:
- https://www.kernel.org/doc/html/v4.14/driver-api/input.html#c.input_allocate_device
- https://www.kernel.org/doc/html/v4.14/driver-api/input.html#c.input_unregister_device
- https://www.kernel.org/doc/html/v4.14/driver-api/input.html#c.input_free_device

### DIFF
--- a/drivers/input/joystick/rpisense-js.c
+++ b/drivers/input/joystick/rpisense-js.c
@@ -107,8 +107,11 @@ static int rpisense_js_probe(struct platform_device *pdev)
 		goto err_keys_reg;
 	}
 	return 0;
+
 err_keys_reg:
 	input_unregister_device(rpisense_js->keys_dev);
+	return ret;
+
 err_keys_alloc:
 	input_free_device(rpisense_js->keys_dev);
 	return ret;
@@ -119,7 +122,6 @@ static int rpisense_js_remove(struct platform_device *pdev)
 	struct rpisense_js *rpisense_js = &rpisense->joystick;
 
 	input_unregister_device(rpisense_js->keys_dev);
-	input_free_device(rpisense_js->keys_dev);
 	return 0;
 }
 


6by9:
Agreed that this looks dubious, however my gut feel is that is probably better fixed by using the devm_ versions of the functions.

Better still is that the Sensehat drivers have actually been upstreamed to https://github.com/raspberrypi/linux/blob/rpi-6.6.y/drivers/input/joystick/sensehat-joystick.c, and that already uses the devm_ versions.

The overlay will need to be updated as the compatible string is different (`rpi,rpi-sense-js` vs `raspberrypi,sensehat-joystick`), and I haven't looked at which other bits of the SenseHAT have been upstreamed so there may be tweaks required for the rest. Slightly dubious is that it appears to have been upstreamed without a DT binding.

paulober:
@6by9 Ok, so should i update it to also use the devm versions or will it just be replaced by the other driver?

6by9:
Ideally we'd use the upstream driver as long as it all works. All downstream patches are an extra maintenance burden with each new kernel release.

I'm just trying to find the thread that describes it upstream. https://lore.kernel.org/all/20220419205158.28088-3-cmirabil@redhat.com/ was where the joystick driver was merged.
It looks like the other patches haven't been landed for whatever reason.

I've just used `b4 shazam https://lore.kernel.org/all/20220419205158.28088-3-cmirabil@redhat.com/` to apply those patches to an rpi-6.6.y tree. Draft PR at #6368 so we can see what might be needed.

6by9:
As just discussed, I think we can just switch to using the newer joystick driver and simple-mfd, whilst leaving the older framebuffer driver (the new one has switched to an auxdisplay driver).

The current rpisense-core driver seems to have a number of other bugs, like not deleting the device properly on remove, hence preventing a reload. Switching to simple-mfd will avoid those too.

XECDesign:
...this is why nobody should let me go near the kernel source.

6by9:
Closing as #6368 removes the core and downstream joystick drivers, and fixes up the framebuffer driver to work with simple-i2c-mfd.

PR #6368, opened 2024-09-17
Upstream SenseHAT patches

6by9:
From #6367, noting that there are upstream patches proposed for the SenseHAT.

The joystick driver was merged, but not these ones. PR for reference and test.


6by9:
@paulober PR updated to use simple-i2c-mfd, upstream joystick driver, and a modified downstream display driver.

Hopefully that will solve your use-after-free in the downstream joystick driver, and gets rid of the other nasties observed in the core driver.

This does still need the UUID of the v1 sense HAT adding to hat_map.dts.

paulober:
@6by9 Thanks, this should solve them.

I was actually doing exactly the same changes (about 80% I would say) but didn't finished getting the FB driver running again after removing the MFD core driver. Sorry for not communicating.
May I ask, why you did not add `depends on FB && I2C` and `select MFD_SIMPLE_MFD_I2C` instead of the custom MFD driver?

Also do you think it can be a good idea to add
```diff
+config MFD_RASPBERRYPI_SENSE_HAT
+       tristate "Raspberry Pi Sense HAT MFD"
+       depends on I2C
+       select MFD_SIMPLE_MFD_I2C
+       help
+         This module supports the LED Matrix controller found on the Raspberry Pi
+         Sense HAT v1 and v2 boards, and the joystick on the Sense HAT.
+         (Functionally it relies on MFD_SIMPLE_MFD_I2C to provide the framework
+         that loads the child drivers).
```
to `drivers/mfd/Kconfig` like the POE HAT MFD record in there?

pelwell:
> This does still need the UUID of the v1 sense HAT adding to hat_map.dts.

What, like this?:
```diff
diff --git a/arch/arm/boot/dts/overlays/hat_map.dts b/arch/arm/boot/dts/overlays/hat_map.dts
index be38b817022f6..ae5d9925c5f4e 100644
--- a/arch/arm/boot/dts/overlays/hat_map.dts
+++ b/arch/arm/boot/dts/overlays/hat_map.dts
@@ -96,6 +96,11 @@ recalbox-rgbdual {
                overlay = "recalboxrgbdual";
        };
 
+       sensehat-v1 {
+               uuid = [ 5d960035 8e87 428f 95d8 59852d697754 ];
+               overlay = "rpi-sense";
+       };
+
        sensehat-v2 {
                uuid = [ 1aa9c428 72eb 48da 9306 8c3706ed3653 ];
                overlay = "rpi-sense-v2";
```
That does seem to do the right thing:
```
  6.43 Read /overlays/hat_map.dtb bytes     1921 hnd 0xc441
  6.66 PCIEx1: PWR 0 DET_WAKE 0
  6.81 Read /config.txt bytes     1487 hnd 0xc
  6.91 Read /overlays/rpi-sense.dtbo bytes     1082 hnd 0xc4f4
  6.05 MESS:00:00:06.405349:0: Loaded overlay 'rpi-sense'
```

6by9:
> May I ask, why you did not add `depends on FB && I2C` and `select MFD_SIMPLE_MFD_I2C` instead of the custom MFD driver?

An oversight.
 
> Also do you think it can be a good idea to add
> 
> ```diff
> +config MFD_RASPBERRYPI_SENSE_HAT
> +       tristate "Raspberry Pi Sense HAT MFD"
> +       depends on I2C
> +       select MFD_SIMPLE_MFD_I2C
> +       help
> +         This module supports the LED Matrix controller found on the Raspberry Pi
> +         Sense HAT v1 and v2 boards, and the joystick on the Sense HAT.
> +         (Functionally it relies on MFD_SIMPLE_MFD_I2C to provide the framework
> +         that loads the child drivers).
> ```
> 
> to `drivers/mfd/Kconfig` like the POE HAT MFD record in there?

No, it just pollutes the config space as it'll only enable the parent, not the subdevices.
Enabling either the joystick or FB driver will select MFD_SIMPLE_MFD_I2C, so that is the correct route.

6by9:
> > This does still need the UUID of the v1 sense HAT adding to hat_map.dts.
> 
> What, like this?:

Thank you. I only found a V2 to test with.

6by9:
Pi5 not getting any joystick events is due to no GPIO23 not having been configured as an input.
Pi3 is an input by default, and hence works.

I'm slightly surprised that the pinmuxing doesn't automatically become an input when you request interrupts from a GPIO, but I guess you could want to know when something else has changed the state of an output.

@pelwell What's the correct route for configuring pinmuxing on Pi5? Still `brcm,pins` and `brcm,function`?

pelwell:
I'll have to look at why it doesn't just work (and if it should), but the RP1 pinctrl driver accepts the old-style `brcm,` declarations (with a compatibility mapping), so it's best to use that unless the overlay has to be Pi 5-specific.

6by9:
Thanks. I've tried adding the old brcm, declarations, but don't appear to be doing anything.

It also feels odd that asking for the interrupt doesn't appear to claim the GPIO or pinctrl. It's quite possible I've just messed up though.

pelwell:
Push what you've got so far.

6by9:
Sorry, I'd left the office.

Changes pushed now.

arch/arm/boot/dts/overlays/rpi-sense-overlay.dts
```diff
@@ -12,11 +12,23 @@
...
 				status = "okay";
+
+				display {
+					compatible = "raspberrypi,rpi-sense-fb";
+					status = "okay";
+				};
+				joystick {
+					compatible = "raspberrypi,sensehat-joystick";
+					interrupts = <23 1>;
+					pinctrl-name = "default";
```
6by9: Found my error - a missing "s" on "pinctrl-name" here.
I'd copied mcp23017-overlay, which is also wrong.

6by9 (commented):
I now agree with upstream that using the BTN_DPAD_* defines is more sensible. The current buttons are treated as straight keyboard inputs. I wondered why my quick evtest wiggles resulted in the Pi rebooting, but I kept on managing to do "up" "enter", and the previous command in my bash history tended to be "sudo reboot".

arch/arm/boot/dts/overlays/rpi-sense-overlay.dts
pelwell: I missed it too

6by9:
One slighty odd thing is that whilst the overlay defines all the nodes for the relevant kernel modules for the sensors, however none of those modules are built by the default defconfigs. I'm guessing that the devices are being accessed from userspace via some random I2C library.

pelwell:
Yes, sadly,

6by9:
Draft status removed.

The display side still works (it's the same driver), and we get button presses with evtest from the joystick.
As above, all the other devices appear to be driven from userspace, so there's nothing further needed to test those.

I'm happy with it. Any comments from others?

PR #6369, opened 2024-09-17
Another trial build for 6365

pelwell:
This is another trial build for #6365. It is a rewind to 2806f39528464, followed by a reversion of the downstream commits since bfbd468e81def (when the UARTs worked).


pelwell:
Superseded by #6371.

PR #6370, opened 2024-09-18
ILI9881 and DSI fixes

6by9:
@lurch


6by9:
Discussed with GSH.
As long as I haven't messed up the rebase, this is a significant improvement to the reliability and is worth merging for the new image. Anything further can be released via apt at a later date.

Issue #6379, opened 2024-09-20
amba-pl011 platform_driver bug


TopgunZh:
### Describe the bug

You missed platform_driver_unregister in the pl011_exit function

### Steps to reproduce the behaviour

1.sudo rmmod amba-pl011.ko.xz
2.sudo insmod amba-pl011.ko.xz

miss  platform_driver_unregister(&pl011_axi_platform_driver);

### Device (s)

Raspberry Pi 5

### System

![image](https://github.com/user-attachments/assets/508ac456-837d-4c9a-9646-e77bf58d0be8)

### Logs

![image](https://github.com/user-attachments/assets/568f5237-0c2a-48eb-b077-fa6a190afa7e)

popcornmix:
Does https://github.com/raspberrypi/linux/pull/6380 work for you?

TopgunZh:
> Does #6380 work for you?
It had already been fixed when I opened the issue. The bug is relatively simple.

PR #6380, opened 2024-09-20
tty/serial: pl011: Also unregister pl011_axi_platform_driver

popcornmix:
See: https://github.com/raspberrypi/linux/issues/6379


pelwell:
Hmm, I don't know - it looks a bit risky.

PR #6383, opened 2024-09-24
Fix speaker disabling issue after closing microphone app

alitekin-saha:
This pull request addresses an issue in the Raspberry Pi Voice HAT driver where the speaker is disabled after the microphone app is closed. The root cause of the issue was incorrect handling of the SDMODE pin, which led to unintended behavior when both the microphone and speaker were used together.

### DIFF
--- a/sound/soc/bcm/googlevoicehat-codec.c
+++ b/sound/soc/bcm/googlevoicehat-codec.c
@@ -95,8 +95,7 @@ static int voicehat_daiops_trigger(struct snd_pcm_substream *substream, int cmd,
 				   struct snd_soc_dai *dai)
 {
 	struct snd_soc_component *component = dai->component;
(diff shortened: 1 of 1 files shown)


pelwell:
From the looks of it, the problem would occur if both playback and record streams are active - stopping the record stream would mute the playback stream. An explanation along those lines needs to go in the commit message, along with your "Signed-off-by:" line.

pelwell:
Hint: We don't (always?) receive notifications for pushes, so it's always wise to leave a comment to say if you've pushed a change that requires attention.

alitekin-saha:
> Hint: We don't (always?) receive notifications for pushes, so it's always wise to leave a comment to say if you've pushed a change that requires attention.

Thanks for the reminder, I've added more lines to the comment.

PR #6385, opened 2024-09-25
drm/vc4: Rework UPM allocation to avoid double buffering

6by9:
The previous UPM allocation was done per plane state, when there is no overlap of usage between frames.

Allocate per plane, and only change the allocation if the required size changes. We have to reference count as plane states can be duplicated.

### DIFF
--- a/drivers/gpu/drm/vc4/tests/vc4_test_lbm_size.c
+++ b/drivers/gpu/drm/vc4/tests/vc4_test_lbm_size.c
@@ -188,6 +188,7 @@ static void drm_vc4_test_vc4_lbm_size(struct kunit *test)
 	struct drm_framebuffer *fb;
 	struct drm_plane *plane;
 	struct drm_crtc *crtc;
+	struct vc4_dev *vc4;
 	unsigned int i;
 	int ret;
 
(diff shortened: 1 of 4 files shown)


6by9:
These patches are now cleanly rejecting commits that exceed LBM or UPM limits. We reuse the allocations if they are of the same size as the previous user of the plane, but will allocate a new buffer when it changes (that will double buffer, but it is a relatively rare event to have significant changes).

Probably hold off merging until the stable image has been bumped, but review comments welcome.

dividuum:
FYI: I've been running this version for a while now and everything seems to work nicely even when giving it quite a beating by shuffling a lot of DRM planes around.

pelwell:
Any objections to merging now?

6by9:
> FYI: I've been running this version for a while now and everything seems to work nicely even when giving it quite a beating by shuffling a lot of DRM planes around.

Great. Thanks for testing.

> Any objections to merging now?

Assuming all the apt updating is done and we're happy, then go for it.

pelwell:
Given that the last release included a cherry-pick, I think we're essentially past the freeze point.

PR #6387, opened 2024-09-25
overlays: wm8960-soundcard: Fix clock declaration

pelwell:
How did this ever work? The static-clock declaration is fine, but the reference to it is attached to part of the simple-audio-card node, rather than the instantiation of the codec driver (a subnode of the I2C controller).

Move the clock reference where it should be, and fix a few minor cosmetic issues at the same time.

### DIFF
--- a/arch/arm/boot/dts/overlays/wm8960-soundcard-overlay.dts
+++ b/arch/arm/boot/dts/overlays/wm8960-soundcard-overlay.dts
@@ -13,7 +13,7 @@
 	};
 
 	fragment@1 {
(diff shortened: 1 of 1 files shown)


Demon000:
I can confirm that this pull request fixes audio playback on wm8960, previously `aplay` would error out with the following.

```
[   69.158923] wm8960 1-001a: No MCLK configured
[   69.158950] wm8960 1-001a: ASoC: error at snd_soc_dai_hw_params on wm8960-hifi: -22
```

PR #6388, opened 2024-09-26
drivers: media: imx500: Simplify the vblank control init

naushir:
Set the VBLANK control minimum and default values to IMX500_VBLANK_MIN unconditionally everywhere.

Remove the mode specific framerate_default parameter, it is now unused.

### DIFF
--- a/drivers/media/i2c/imx500.c
+++ b/drivers/media/i2c/imx500.c
@@ -274,9 +274,6 @@ struct imx500_mode {
 	/* Analog crop rectangle. */
 	struct v4l2_rect crop;
 
-	/* Default framerate. */
-	unsigned int framerate_default;
-
 	/* Default register values */
 	struct imx500_reg_list reg_list;
 };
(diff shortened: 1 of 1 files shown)


roliver-rpi (approved):

drivers/media/i2c/imx500.c
```diff
@@ -1744,40 +1739,22 @@ static int imx500_get_pad_format(struct v4l2_subdev *sd,
...
 	__v4l2_ctrl_modify_range(
 		imx500->vblank, IMX500_VBLANK_MIN,
 		((1 << IMX500_LONG_EXP_SHIFT_MAX) * IMX500_FRAME_LENGTH_MAX) -
-			mode->height,
-		1, frm_length_default - mode->height);
+			mode->height, 1, IMX500_VBLANK_MIN);
 
 	/* Setting this will adjust the exposure limits as well. */
-	__v4l2_ctrl_s_ctrl(imx500->vblank, frm_length_default - mode->height);
+	__v4l2_ctrl_s_ctrl(imx500->vblank, IMX500_VBLANK_MIN);
```
6by9, line 1757: Technically not needed as __v4l2_ctrl_modify_range will update the value if the current value is out of range and call s_ctrl.

There was a debate previously over whether setting the mode should reset the frame rate controls or not, but I don't think there was ever a final decision.

6by9 (approved):

drivers/media/i2c/imx500.c
naushir: Agree, but I'd prefer to leave it in to preserve existing code that I've been testing with 😄

PR #6393, opened 2024-10-01
Revert switch to upstream HAT patches

popcornmix:
Temporary revert until the firmware side non-UUID HAT mapping feature is ready.
See: https://forums.raspberrypi.com/viewtopic.php?t=377333


popcornmix:
@6by9 does this look correct?
I didn't revert "[dtoverlays: Correct pinctrl assignment on mcp23017](https://github.com/raspberrypi/linux/pull/6368/commits/cf37a0fc6363ce9ce8ae978160523f811afbdd31)" as that appeared to be a genuine bug fix that is not dependent on the other commits.

6by9:
Looks fine.

"drivers/mfd: sensehat: Add Raspberry Pi Sense HAT to simple_mfd_i2c" isn't reverted either, but that will have no effect.
We could keep "Input: sensehat-joystick : Revert to downstream keymap" as again it would have no effect, but not an issue to reapply later on.

Issue #6398, opened 2024-10-03
Add support for SquashFS ZStandard compressed


mortegai:
Please add support for kernel module "squashfs" ZStandard compressed file systems.
CONFIG_SQUASHFS_ZSTD

pelwell:
PR #6400 adds SQUASHFS_ZSTD. After about 40 minutes you should be able to install a trial kernel build using `sudo rpi-update pulls/6400`.

mortegai:
Tested and working properly.

PR #6400, opened 2024-10-03
configs: Enable SQUASHFS_ZSTD

pelwell:
ZSTD is the current favourite compression standard. Enable SquashFS to use it.

Link: https://github.com/raspberrypi/linux/issues/6398


pelwell:
The ZSTD module was already enabled, and this option only seems to add about 3KB to the squashfs module size.

popcornmix (approved):

PR #6403, opened 2024-10-06
Add uConsole driver support 

crossplatformdev:


pelwell:
We don't normally take whole drivers for devices which aren't Pi-exclusive, and certainly not without a lot more information.

crossplatformdev:
These PR is a rpi-6.6.y base with @oesmith patch:
https://raw.githubusercontent.com/oesmith/uconsole/refs/heads/main/kernel/clockworkpi-kernel-6.6.patch

Extracted from @ak-rex's kernel fork:
https://github.com/ak-rex/ClockworkPi-linux/commits/rpi-6.6.y/

That allows CM4 to be used with the uConsole landing board, screen, battery, and other drivers.

@clockworkpi's uConsole description is below:
https://www.clockworkpi.com/product-page/uconsole-kit-rpi-cm4-lite

6by9:
Panel drivers want to go upstream, and would need devicetree bindings.
CWD686 panel driver has been sent before at https://patchwork.freedesktop.org/patch/503489/, but no v4 ever seemed to appear.

The change to vc4_dsi.c I'd want some more description on - I'm not aware of any issue in coming out of ULPS mode from power on, and ULPS isn't used at any other time now.

ACP20 driver changes would want to go upstream too with justification.

Backlight driver wants to go upstream. Based on the [datasheet I found](https://en.visvie.com/static/upload/file/20220902/1662097252339999.pdf) it's a 1-wire interface, so should ideally be using the 1-wire support rather than bit-bashing.

bcm2835-audio change is wrong as it should be handled by updating the kernel command line in device tree.

Any patch should be a single self-contained entity, one per driver or per feature being added.

Issue #6349, opened 2024-09-11
After "apt upgrade" every other boot fails


ViRb3:
### Describe the bug

I ran `apt update && apt upgrade` today. Since then, approximately every other reboot fails (50% chance) with the following error:
```
Started systemd-udevd.service - Rule-based Manager for Device Events and Files.
[ TIME ] Timed out waiting for device dev-disk-by\2xdpartuuid-4e639091\x2d01.device - /dev/disk/by-partuuid/4e639091-01.
[DEPEND] Dependency failed for failed for boot-firmware.mount - /boot/firmware.
[DEPEND] Dependency failed for local-fs.target - Local File Systems.
`````

I was unfortunate to update 3 of my Raspberry Pi 5s, and **every single one experiences this issue**.

### Steps to reproduce the behaviour

1. Run `apt update && apt upgrade`
2. Repeat reboot until reproduced

### Device (s)

Raspberry Pi 5

### System

```bash
$ cat /etc/rpi-issue
Raspberry Pi reference 2023-12-11
Generated using pi-gen, https://github.com/RPi-Distro/pi-gen, 2acf7afcba7d11500313a7b93bb55a2aae20b2d6, stage2

$ vcgencmd version
2024/07/30 15:25:46
Copyright (c) 2012 Broadcom
version 790da7ef (release) (embedded)

$ uname -a
Linux server 6.6.47+rpt-rpi-v8 #1 SMP PREEMPT Debian 1:6.6.47-1+rpt1 (2024-09-02) aarch64 GNU/Linux
```

pelwell:
Which medium are these Pis booting from?

ViRb3:
@pelwell Micro SD card.

P33M:
What sort of micro sd card?

ViRb3:
It's a SanDisk Extreme Pro 128GB on all three of the devices. For what is worth, I also have a bunch of external HDDs attached to each Pi, however those have their own power supply and are connected via a powered hub.

I also use an initramfs with dropbear for rootfs decryption according to https://github.com/ViRb3/pi-encrypted-boot-ssh.

This setup hasn't ever caused issues until yesterday.

pelwell:
Is your debug output above from a serial port? Or rather, do you have a Pi 5-compatible serial cable you can use to capture debug output? If so, try with `ignore_loglevel` in cmdline.txt (keeping it to a singe line) - there should be some messages about the SD card, but whatever you get, post it all.

[...]

PR #6405, opened 2024-10-07
mmc: sd: disable sd cqhci by default

P33M:
The fixup method isn't ideal as it doesn't allow for date range specifiers, so this array will become unwieldy.

For now, constrain to whole years which span production dates of cards that misbehave.

Link: https://forums.raspberrypi.com/viewtopic.php?t=367459
Also see #6349

### DIFF
--- a/arch/arm/boot/dts/overlays/README
+++ b/arch/arm/boot/dts/overlays/README
@@ -378,9 +378,9 @@ Params:
                                 non-lite SKU of CM4).
                                 (default "on")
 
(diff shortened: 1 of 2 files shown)


pelwell:
The auto-builds are ready, if you want people to try them.

I assume are new official cards require no random quirks to be specified?

P33M:
The new cards have had weeks of soak time running continuous stress tests, so no. However I still haven't ruled out a regression that's crept in between testing and the most recent apt release which could also affect them.

PR #6408, opened 2024-10-09
dts: bcm2712d0: Add non-d0 vc6 compatible string

pelwell:
Although the VC4/VC6 driver requires a special compatible string for the "d0" stepping, the removal of the old compatible string upsets Mesa.

Satisfy both requirements by adding the old "brcm,bcm2712-vc6" string as a fallback.

### DIFF
--- a/arch/arm/boot/dts/overlays/bcm2712d0-overlay.dts
+++ b/arch/arm/boot/dts/overlays/bcm2712d0-overlay.dts
@@ -41,7 +41,7 @@
 	fragment@4 {
 		target = <&vc4>;
 		__overlay__ {
-			compatible = "brcm,bcm2712d0-vc6";
+			compatible = "brcm,bcm2712d0-vc6", "brcm,bcm2712-vc6";
 		};
 	};
 
--- a/arch/arm64/boot/dts/broadcom/bcm2712d0-rpi-5-b.dts
+++ b/arch/arm64/boot/dts/broadcom/bcm2712d0-rpi-5-b.dts
@@ -87,7 +87,7 @@
 };
 
 &vc4 {
-	compatible = "brcm,bcm2712d0-vc6";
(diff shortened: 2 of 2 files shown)


6by9:
Is the `brcm,bcm2712-vc6` string submitted to upstream Mesa, or just our downstream patches? 

The patches I'm working on for upstream were going to use `brcm,bcm2712c-vc6` and `brcm,bcm2712d-vc6`, which will have similar issues.

popcornmix:
This is the block that cares (in vulkan part of mesa)
https://gitlab.freedesktop.org/mesa/mesa/-/blob/main/src/broadcom/vulkan/v3dv_device.c?ref_type=heads#L1544

(and yes, changing to brcm,bcm2712c-vc6 will break it if you don't include the old name as an alternate).

6by9:
> This is the block that cares (in vulkan part of mesa) https://gitlab.freedesktop.org/mesa/mesa/-/blob/main/src/broadcom/vulkan/v3dv_device.c?ref_type=heads#L1544

Interesting that they are doing an strncmp, so `brcm,bcm2712-vc6-d0` would have matched happily. 
 
> (and yes, changing to brcm,bcm2712c-vc6 will break it if you don't include the old name as an alternate).

Hmm, that means the bindings have to list the generic name as well. Perhaps we do need to add it as a suffix, or update Mesa.

pelwell:
Let's go with the suffix for now.

popcornmix:
I did question Igalia why the vulkan part of mesa checked vc4 compatibles and the gl part did not, and they agreed it wasn't necessary and proposed a patch:
https://gitlab.freedesktop.org/itoral/mesa/-/tree/v3dv_device_name?ref_type=heads

So hopefully going forward we won't have to be so careful with changes here.
But for now (with existing mesa in bookworm), this dts change seems good.

PR #6409, opened 2024-10-10
configs: Enable HOTPLUG_CPU

popcornmix:
This works and can be useful for
debugging/benchmarking/low power use

$ nproc
4
$ echo 0 | sudo tee  /sys/devices/system/cpu/cpu3/online
0
$ echo 0 | sudo tee  /sys/devices/system/cpu/cpu2/online
0
$ echo 0 | sudo tee  /sys/devices/system/cpu/cpu1/online
0
$ nproc
1


popcornmix:
Tested on Pi4 and while nothing bad happens, we are unable to offline a cpu (as you can on pi5).
The `online` and `offline` nodes are not present in  `/sys/devices/system/cpu/cpuN/`.

I think this is due to PSCI support in pi5 arm stub (using arm-trusted-firmware), which is not present in the more minimal stubs of Pi0-4.

I suspect this is the code that checks:
https://github.com/raspberrypi/linux/blob/0985b32279d47e7ea01152233d26eedcdf721d09/arch/arm64/kernel/psci.c#L50

I've restricted the config changes to `arch/arm64/configs/bcm2712_defconfig` and `arch/arm64/configs/bcm2712_defconfig`.

timg236 (approved):

PR #6415, opened 2024-10-14
dtoverlays: Fix up imx500 overlays to have unique clock nodes

6by9:
The overlay was creating DT nodes /clocks/clk-aicam and /clocks/clk-aicam-gated for both cam0 and cam1, which resulted in one failing.

The clock infrastructure creates the clock name from the node name without any @N reg extension, so we can't just use that. The nodes therefore have to be renamed.

@naushir
@roliver-rpi 
https://forums.raspberrypi.com/viewtopic.php?t=377822 (alpha team sub-forum, so not generally accessible).

### DIFF
--- a/arch/arm/boot/dts/overlays/imx500-overlay.dts
+++ b/arch/arm/boot/dts/overlays/imx500-overlay.dts
@@ -72,16 +72,16 @@
 		};
 	};
 
-	clocks_frag: fragment@104 {
+	fragment@104 {
 		target-path = "/clocks";
 		__overlay__ {
-			clk_aicam: clk-aicam {
+			clk_aicam: clk-aicam1 {
 				compatible = "fixed-clock";
 				#clock-cells = <0>;
 				clock-frequency = <24000000>;
 			};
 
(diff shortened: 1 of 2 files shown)


pelwell:
> The nodes therefore have to be renamed.

For the curious, there's no technical reason why the existing nodes in the overlay have to be renamed, as long as the cam0 parameter causes them to have different names, but this is more aesthetically pleasing.

naushir:
LGTM

6by9:
> > The nodes therefore have to be renamed.
> 
> For the curious, there's no technical reason why the existing nodes in the overlay have to be renamed, as long as the cam0 parameter causes them to have different names, but this is more aesthetically pleasing.

I'd tried having a separate fragment for the two cases, but then you end up having to change the clock node that &cam_node uses (labels in dormant fragments still count as labels and have to be unique).

More curious is why the clock framework drops the register number extension when it creates the clock name - that's just annoying! It's possible that `clock-output-names` would have worked, but this is as easy.

PR #6419, opened 2024-10-15
CQE incremental fixes part 2

P33M:
If #6418 fixes the generic case of a hang on boot with an A2 card, then this PR should be merged.

The 2nd commit should be a no-op, as I placed an assert in cqe_off() for cq_host->qcnt != 0 and it never fired. As it turns out, mmcblk /does/ serialise async requests if a synchronous one arrives, and the existing claim mechanism should prevent other contexts from running in parallel. Retuning is a bit special as it is invoked prior to starting either a cqe or regular request, but it appears to have been thought about.

The third commit should disable CQ on older Sandisk cards such as https://forums.raspberrypi.com/viewtopic.php?t=367459&start=25#p2208495 (and I have one of these).

### DIFF
--- a/arch/arm/boot/dts/overlays/README
+++ b/arch/arm/boot/dts/overlays/README
@@ -378,9 +378,9 @@ Params:
                                 non-lite SKU of CM4).
                                 (default "on")
 
-        sd_cqe                  Use to enable Command Queueing on the SD
-                                interface for faster Class A2 card performance
-                                (Pi 5 only, default "off")
+        sd_cqe                  Set to "off" to disable Command Queueing if you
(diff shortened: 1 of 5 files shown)


P33M:
From my testing, the loop that previously failed in 20 minutes survived an overnight run. There's one positive response in the issue, and one pending.

P33M:
This PR now fixes two definite bugs in the CQ implementation so the default can switch back to enabled for supported cards. During testing there was a mild panic as the other "known good" A2 card from Integral would reliably break the filesystem if trimmed. However, that's unrelated to CQ - it just doesn't erase properly.

PR #6423, opened 2024-10-17
OV5647 link frequency control

6by9:
I found I had the OV spreadsheet for PLL config, so could determine the link frequency.

I don't know why we had a different clock setup for 1080p cropped and binned modes vs the full res. Increasing it that fraction gives a slightly higher max frame rate in those modes (32.81 fps at 1080p and 46.34fps at 1296x972), and removes one extra link frequency.

There will hopefully be a follow-up to these which will add selection of the slower rate for 6sec exposures. 
Simplest approach is to set {0x3037, 0x06} to halve all the clocks (default is 0x03). Arranging it will be the awkward one - selection via link frequency in DT is the cleanest approach, but the current overlays actually have a single incorrect value defined, so we'll need a fallback.

(I will find a few minutes to send these patches upstream in due course).


drivers/media/i2c/ov5647.c
```diff
@@ -715,6 +565,13 @@ static int ov5647_set_mode(struct v4l2_subdev *sd)
 	if (ret < 0)
 		return ret;
 
+	ret = ov5647_write_array(sd, ov5647_common_regs,
```
naushir, line 581: Do we want a guard like ``if (!ov5647->common_regs_written)`` like we do with other drivers?

naushir (commented):
Only the one minor thing, otherwise LGTM

drivers/media/i2c/ov5647.c
6by9: The `{0x0103, 0x01},` as the second register in the common list is a reset.
The mode tables do not all cover the superset of those set by any of the modes, hence requiring the reset.

Another followup is to remove all the registers that are set to the default value - there are a fair number of those which are all redundant.

naushir: In which case, let's merge as-is.

Bajec:
why are link frequencyes not added also to the [device tree](https://github.com/raspberrypi/linux/blob/84ab77459e61c648299d32464127b89ca65de40a/arch/arm/boot/dts/overlays/ov5647.dtsi#L21)?

6by9:
> why are link frequencyes not added also to the [device tree](https://github.com/raspberrypi/linux/blob/84ab77459e61c648299d32464127b89ca65de40a/arch/arm/boot/dts/overlays/ov5647.dtsi#L21)?

The driver doesn't check the link frequencies defined in devicetree, so there was little point in updating DT at this stage. The 297MHz value currently in there is totally wrong for this sensor.

As I've said, there will be a follow up patch that will read the link frequencies and configure the clock setup accordingly.

PR #6427, opened 2024-10-18
Input: matrix-keypad - don't map irqs when atomic

pelwell:
Calling gpio_to_irq can result in irqs being mapped for the first time. which is not safe in atomic context. Enabling CONFIG_PROVE_LOCKING will trigger a kernel splat from _lock_acquire in a call chain from matrix_keypad_scan.

Fix the bug (and improve efficiency) by mapping all GPIO irqs from within the probe function.

Link: https://forums.raspberrypi.com/viewtopic.php?t=378044
Fixes: bab7614d6d1b ("Input: add support for generic GPIO-based matrix keypad")

### DIFF
--- a/drivers/input/keyboard/matrix_keypad.c
+++ b/drivers/input/keyboard/matrix_keypad.c
@@ -27,6 +27,7 @@ struct matrix_keypad {
 	const struct matrix_keypad_platform_data *pdata;
 	struct input_dev *input_dev;
 	unsigned int row_shift;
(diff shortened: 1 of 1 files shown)


pelwell:
Perhaps not so surprisingly, an almost identical patch is already in kernels from 6.9 onwards. PR updated with a back-port of that commit instead.

pelwell:
@dtor FYI your patch is a fix, not just an optimisation.

popcornmix (approved):

PR #6429, opened 2024-10-22
drivers: media: bcm2835_isp: Cache LS table dmabuf

naushir:
Clients such as libcamera do not change the LS table dmabuf on every frame. In such cases instead of mapping/remapping the same dmabuf on every frame to send to the firmware, cache the dmabuf once and only update and remap if the dmabuf has been changed by the userland client.

### DIFF
--- a/drivers/staging/vc04_services/bcm2835-isp/bcm2835-v4l2-isp.c
+++ b/drivers/staging/vc04_services/bcm2835-isp/bcm2835-v4l2-isp.c
@@ -139,6 +139,8 @@ struct bcm2835_isp_dev {
 	/* Image pipeline controls. */
 	int r_gain;
 	int b_gain;
+	struct dma_buf *last_ls_dmabuf;
+	struct mmal_parameter_lens_shading_v2 ls;
 };
 
 struct bcm2835_isp_buffer {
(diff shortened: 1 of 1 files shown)


drivers/staging/vc04_services/bcm2835-isp/bcm2835-v4l2-isp.c
```diff
@@ -1740,6 +1756,7 @@ static int bcm2835_isp_probe_instance(struct platform_device *pdev,
 	}
 
 	atomic_set(&dev->num_streaming, 0);
+	dev->last_ls_dmabuf = NULL;
```
6by9: Not necessary as we've used `devm_kzalloc` for dev.

drivers/staging/vc04_services/bcm2835-isp/bcm2835-v4l2-isp.c
```diff
@@ -677,6 +679,7 @@ static void bcm2835_isp_node_stop_streaming(struct vb2_queue *q)
 				 "%s: Failed disabling component, ret %d\n",
 				 __func__, ret);
 		}
+		dev->last_ls_dmabuf = NULL;
```
6by9: Nit pick: Putting this at line 674 so straight after the set_isp_param call associates the two parts more cleanly.

6by9: You could also update `dev->ls` instead of having a local here.

Is there a path in stopping and restarting where it reuses the existing `dev->ls` struct or control value, and is then using a stale memhandle? In theory it should be possible to disable a LS table after having been configured, but I'm not sure that is supported.

naushir: I'll reuse ``dev->ls`` and null it.  That's probably the safest thing.  The libcamera client will always re-populate the fd on startup.

naushir:
@6by9 I think this is ready to merge if you are happy with the changes.

6by9:
Not the easiest to read on a phone, but it looks OK.

PR #6434, opened 2024-10-23
dts: 2712: Drop some numa options from bootargs

popcornmix:
iommu_dma_numa_policy=interleave is not valid in the current tree It generates an unknown setting will be passed to usespace warning

system_heap.max_order=0 is wanted when numa is enabled, but may not be when it is disabled.

Add it on firmware side when we know if `numa=fake=<n>` is used.

### DIFF
--- a/arch/arm64/boot/dts/broadcom/bcm2712-rpi.dtsi
+++ b/arch/arm64/boot/dts/broadcom/bcm2712-rpi.dtsi
@@ -99,7 +99,7 @@
 
 / {
 	chosen: chosen {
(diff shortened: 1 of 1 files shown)


popcornmix:
Draft until bootloader side is merged and available.

popcornmix:
Bootloader should [now do this](https://github.com/raspberrypi/rpi-eeprom/commit/3eecae99756865406bf30693ee2e7c20dad8f947), so marking this as ready.

pelwell:
You'll need to resolve the merge conflict.

popcornmix:
Rebased

PR #6421, opened 2024-10-16
dts: bcm2712-rpi: Add aliases for the CSI/DSI I2Cs

pelwell:
Older Pis arrange that the camera I2C ports appear as /dev/i2c-10. Add aliases so that on the Pi 5 family, i2c_csi_dsi0 becomes i2c-10 and i2c_csi_dsi1 becomes i2c-11. Only the I2C buses that appear on the 40-pin header, i.e. I2C0 to I2C3, get a low bus number.

Also add hints for our udev rules about which symlinks to create for backwards-compatibility with the previous bus numbers. Note that lower numbers have priority, so i2c-0 on CM5 masks i2c-11, forcing i2c-11 to be a symlink to i2c-0, not vice versa.

PR #6441, opened 2024-10-24
arm64: dts: Sort out CM5 and I/O board I2C ports

pelwell:
There is a difference in I2C usage between CM4IO and CM5IO. Present a simple, consistent view of the world by moving the assignment of the bus IDs into carrier-specific files.

Fixes: 36faab69e8ee ("dts: bcm2712-rpi: Add aliases for the CSI/DSI I2Cs")
See: https://github.com/raspberrypi/linux/pull/6421

### DIFF
--- a/arch/arm64/boot/dts/broadcom/bcm2712-rpi-cm4io.dtsi
+++ b/arch/arm64/boot/dts/broadcom/bcm2712-rpi-cm4io.dtsi
@@ -1,0 +1,28 @@
+// SPDX-License-Identifier: GPL-2.0
+
+i2c_csi_dsi0: &i2c0 { // Note: For CAM0 and DISP0 connectors
+};
+
+i2c_csi_dsi1: &i2c6 { // Note: For CAM1, DISP1, on-board RTC, and fan controller
+	pinctrl-0 = <&rp1_i2c6_38_39>;
+	pinctrl-names = "default";
+	clock-frequency = <100000>;
+	symlink = "i2c-6";
+};
+
+i2c_csi_dsi: &i2c_csi_dsi1 { }; // The connector that needs no jumper to enable
+
+&aliases {
+    /delete-property/ i2c11;
+    i2c10 = &i2c_csi_dsi;
+};
+
+// The RP1 USB3 interfaces are not usable on CM4IO
+
+&rp1_usb0 {
+	status = "disabled";
+};
+
+&rp1_usb1 {
+	status = "disabled";
+};
--- a/arch/arm64/boot/dts/broadcom/bcm2712-rpi-cm5-cm4io.dts
+++ b/arch/arm64/boot/dts/broadcom/bcm2712-rpi-cm5-cm4io.dts
@@ -2,19 +2,4 @@
 /dts-v1/;
 
 #include "bcm2712-rpi-cm5.dtsi"
-
-// The RP1 USB3 interfaces are not usable on CM4IO
-
-&rp1_usb0 {
-	status = "disabled";
-};
-
-&rp1_usb1 {
-	status = "disabled";
-};
-
-/ {
-	__overrides__ {
-		i2c_csi_dsi = <&i2c_csi_dsi>, "status";
-	};
-};
+#include "bcm2712-rpi-cm4io.dtsi"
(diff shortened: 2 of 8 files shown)


arch/arm64/boot/dts/broadcom/bcm2712-rpi-cm4io.dtsi
```diff
@@ -0,0 +1,24 @@
+// SPDX-License-Identifier: GPL-2.0
+
+i2c_csi_dsi0: &i2c0 { // Note: For CAM0 and DISP0 connectors
```
6by9, line 3: I haven't looked at the pinout, but CM5 only has 2 MIPI PHY ports, so do CAM0 and DISP0 actually connect to anything other than the I2C?

I'm sure Dominic made a comment in passing yesterday that CAM0 and DISP0 pins on the CM were repurposed for USB3.

arch/arm64/boot/dts/broadcom/bcm2712-rpi-cm4io.dtsi
```diff
@@ -0,0 +1,24 @@
...
+	symlink = "i2c-10";
+};
+
+i2c_csi_dsi1: &i2c6 { // Note: For CAM1, DISP1, on-board RTC, and fan controller
```
6by9, line 6: And seeing as CAM1 and DISP1 share an I2C port, if the above is true about the MIPI wiring, that means we can't run two identical cameras or displays on CM5 with CM4IO (fine on CM5IO).

arch/arm64/boot/dts/broadcom/bcm2712-rpi-cm4io.dtsi
pelwell: This is the connectivity table I ended up with:
```
CM5        CM4IO        CM5IO
---        -----        -----
DPHY0      CAM1         CAMDISP0
DPHY1      DISP1        CAMDISP1
GPIO0/1    CAM0, DISP0  CAMDISP1
GPIO38/39  CAM1, DISP1  CAMDISP0
```

arch/arm64/boot/dts/broadcom/bcm2712-rpi-cm4io.dtsi
pelwell: No, you've got it backwards: CAM0 and DISP0 shared DPHY0, CAM1 and DISP1 share DPHY1.

pelwell: On CM4IO, of course. CM5IO only has the two sockets.

6by9: CAM0 and DISP0 only have 2 data lanes each on the CM4 pinout, so how have we got the 4 data lanes of second PHY on CM5 routed out?

I'll have to find a schematic....

pelwell: We don't - the 3rd and 4th data lanes of CAM0 and DISP0 are unconnected.

pelwell: Yes, I believe that's correct.

arch/arm64/boot/dts/broadcom/bcm2712-rpi-cm4io.dtsi
pelwell: This means that CAM0 and DISP0 are not usable for cameras or displays, which is the better of the two options because that I2C port is effectively shared with ID_SC and ID_SD. The `i2c_csi_dsi0` parameter is there for compatibility with any existing I2C setup. `i2c_csi_dsi1` (and its alias `i2c_csi_dsi`) is assigned to CAM1 and DISP1.

pelwell:
I believe this is the best we can do for CM4IO - people should use CM5IO to get the most out of CM5.

pelwell:
Unless somebody discovers a problem, this is the final iteration. From the updated commit message:
```
CM5 has reduced connectivity on CM4IO - the DPHYs are connected to
CAM1 and DISP1. Keep i2c-10 for use with that pair, as is the case for
CM4 on CM4IO.
```

Issue #6446, opened 2024-10-29
Feature Request:  Add support for TCS3725 ahd VEML6040 RGB Color Sensors


Jibun-no-Kage:
Feature Request:  Add support for TCS34725 ahd VEML6040 RGB Color Sensors as device tree overlays.

Example of TCS34725 RGB Color Sensor...
https://www.adafruit.com/product/1334

VEML6040 RGB Color Sensor...
https://www.taydaelectronics.com/veml6040-color-sensor-rgbw-i2c-module.html

pelwell:
TCS34725 will need it's driver enabling - `CONFIG_TCS3472=m`.  It makes no use of Device Tree configuration, so it should just be a matter of adding the name to the `i2c-sensor` overlay.

The VEML6040 (not VMEML6040) driver doesn't exist in 6.6, but it is there in 6.12. It also doesn't seem to require any configuration.

If I have a moment I might throw together the necessary config and overlay updates against 6.12. What's your application?

pelwell:
And which I2C addresses can the devices appear on - don't make me read the datasheets.

Jibun-no-Kage:
Oh, right, TCS34725 is 0x29...  VEML6040 is 0x10...

The use case is that I am using these sensors to read a 'tally' light on a remote floating head camera, at the end of 25 or so foot  jib or boom crane... which is beyond visibility to anyone at ground level.  A micro controlller, i.e. in this case Pi Zero 2W forwards this color state to a production broadcast monitor at ground level, that has a custom GPIO interface,  I have not decided if I will use solid-state relays, opto-couplers, or NPN transistors, or even mosfet to do the actual GPIO pin state changes between the Pi and monitor GPIO interface yet.  but since i have the solid state relays, likely use those for the initial testing.,  And of course, the relays provide isolation between Pi GPIO and Production Monitor GPIO.

Sensor->i2c->Pi->Evaluate RGB levels->Pi GPIO->Trigger applicable relay.  On production broadcast monitor GPIO interface, Pin to GND, i.e. Low 'On', Pin High 'Off', for the respective color of tally status on-screen of monitor.  So I just via relay, short pin to GND for the given pin, i.e. color, red or green.

Typical broadcast switchers and cameras... Red is on-air, Green is stand-by or preview.  Or none... if camera is not pending preview or on-air.

[...]

PR #6448, opened 2024-10-29
Add TCS3472 and VEML6040 sensor support (and a few others)

pelwell:
Add support for the I2C sensors requested in #6446, and a few other related devices.


gnustomp:
Hi, might it be possible to update this for 6.12.y? There should be no need to backport patches, just add the sensors to the devicetree overlay and defconfigs. I can test for VEML6040.

pelwell:
Are you in a position to test it? The reason the PR was never merged is that the original requester stopped responding.

gnustomp:
I only have a VEML6040, but yes I am happy to test. (In fact, I had manually added it to the devicetree locally, only to discover the kernel driver wasn't built.)

arch/arm64/configs/bcm2712_defconfig
```diff
@@ -1536,6 +1536,7 @@ CONFIG_INV_MPU6050_I2C=m
 CONFIG_APDS9960=m
 CONFIG_AS73211=m
 CONFIG_BH1750=m
+CONFIG_TCS3472=m
 CONFIG_TSL4531=m
 CONFIG_VEML6070=m
```
gnustomp, line 1542: I think `CONFIG_VEML6040=m` needs to be added to the defconfigs.

gnustomp (commented):
With `sudo rpi-update pulls/6448`, the devicetree node of VEML6040 is present but there is no driver loaded.

```
$ uname -a
Linux dispi 6.12.22-v8-16k+ #1 SMP PREEMPT Thu Apr 24 17:25:53 UTC 2025 aarch64 GNU/Linux
$ cat /boot/firmware/config.txt
<snipped>
dtoverlay=i2c-sensor,bme280,veml6040
$ cat /sys/firmware/devicetree/base/axi/pcie@1000120000/rp1/i2c@74000/veml6040@10/status; echo
okay
$ lsmod | grep veml6040
$ ls /lib/modules/6.12.22-v8-16k+/kernel/drivers/iio/light
apds9960.ko.xz  as73211.ko.xz  bh1750.ko.xz  tcs3472.ko.xz  tsl4531.ko.xz  veml6070.ko.xz  veml6075.ko.xz
```

arch/arm64/configs/bcm2712_defconfig
pelwell: Indeed - it's been missing from the start, but not tested. It's there now, and there will be another test build in about 40 minutes.

gnustomp:
Confirmed that veml6040 is loaded and functional:

```
$ uname -a
Linux dispi 6.12.22-v8-16k+ #1 SMP PREEMPT Fri Apr 25 14:43:02 UTC 2025 aarch64 GNU/Linux
$ cat /sys/bus/iio/devices/iio\:device0/name
veml6040
$ cat /sys/bus/iio/devices/iio\:device0/in_intensity_clear_raw 
120
```

Thanks!

PR #6450, opened 2024-11-01
Backport the upstream PWM GPIO driver from 6.11 to rpi-6.6.y

timg236:
Backport this driver to provide bit bashed PWM support on an arbitrary GPIO pin. 

On Pi5 this is considerably faster than doing this from user space but this is still constrained somewhat by PCIe latency and jitter if the PCIe link is under a high load (e.g. USB3). This should still be suitable for simple applications like servos.


pelwell:
The usual set of defconfigs is:
```
arm bcmrpi_defconfig
arm bcm2709_defconfig
arm bcm2711_defconfig
arm64 bcm2711_defconfig
arm64 bcm2712_defconfig
arm64 bcmrpi3_defconfig
```
We don't use `bcmrpi3_defconfig` - it was originally user-constributed before we embraced 64-bit builds, but I've been keeping it up-to-date with the others.

timg236:
@Gadgetoid FYI, the backport of the pwm-gpio should be in rpi-update relatively soon.

It worked ok on a Pi5 with a MG946 servo and seemed solid under normal loads. Maxing out the PCIe link can cause a very small but observable wobble on the servo.

Gadgetoid:
Thank you!

Can confirm this works from source (rpi-update hasn't got the memo just yet).

From my earlier testing jitter and accuracy were lightyears ahead of the common userspace software PWM alternatives.

It's just a shame the `/sys/class/pwm` interface is a disaster. I don't know how far out the switch to the character device ABI is, but it'll be welcome.

Issue #6220, opened 2024-06-11
DHT11.ko (32-bit Bookworm Raspberry Pi OS with a 64-bit kernel) (raspberry pi 4b)


Abdulrahman-Yasser:
### Describe the bug

I think DHT11 module doesn't work with my kernel image

### Steps to reproduce the behaviour

**My aim is to read dht11 from the file system (kernel module)** 
Steps : 

1. Changed my device tree bcm2711-rpi-4-b.dts and added the following lines 
``` 
humidity-sensor {
             compatible = "dht11";
             gpios = <&gpio 17 GPIO_ACTIVE_HIGH>;
         }; 
```
2. connected my dht data to gpio17 - vcc to rpi.5v - gnd to rpi.gnd
3. i can find the file in ls -l /sys/bus/iio/devices/iio:device0
4. output of `ls -l /sys/bus/iio/devices/iio:device0` is : 
```
-rw-r--r-- 1 root root 4096 Jun  7 01:30 in_humidityrelative_input
-rw-r--r-- 1 root root 4096 Jun  7 01:30 in_temp_input
-r--r--r-- 1 root root 4096 Jun  7 01:29 name
lrwxrwxrwx 1 root root    0 Jun  7 01:30 of_node -> ../../../../firmware/devicetree/base/humidity-sensor
drwxr-xr-x 2 root root    0 Jun  7 01:30 power
lrwxrwxrwx 1 root root    0 Jun  7 01:30 subsystem -> ../../../../bus/iio
-rw-r--r-- 1 root root 4096 Jun  7 01:29 uevent
-r--r--r-- 1 root root 4096 Jun  7 01:30 waiting_for_supplier
```
5.  I try to read the temperature using the following code :
```
sudo cat in_temp_input 
```
It shows 
```
cat: in_temp_input: Input/output error
```
6. output of `dmesg | tail` is 
```
[   59.967095] dht11 humidity-sensor: Don't know how to decode data: 42 0 31 4
[   62.719064] dht11 humidity-sensor: Don't know how to decode data: 42 0 31 3
[   64.866734] dht11 humidity-sensor: Don't know how to decode data: 42 0 31 3
[  193.956733] dht11 humidity-sensor: Don't know how to decode data: 42 0 31 3
```
7. The reading in dmesg is correct. I tried to test it with a lighter and the temperature increased correctly. 
[...]

PR #6454, opened 2024-11-04
iio: humidity: dht11: Allow non-zero decimals

pelwell:
The DHT11 datasheet is pretty cryptic, but it does suggest that after each integer value (humidity and temperature) there are "decimal" values. Validate these as integers in the range 0-9 and treat them as tenths of a unit.

Link: https://github.com/raspberrypi/linux/issues/6220


pelwell:
Merged for lack of interest.

PR #6457, opened 2024-11-06
More dwc3 quirks

P33M:
More fallout from #6141 -  I'm not sure why it was just the BT dongle that tripped over the TRB race - but the suspicious conditions are all met - only a single TRB is ever posted, and there's very little other activity on the bus at the same time as interrupt/control transfers are posted in lockstep.


pelwell (approved):

P33M:
fyi this also applies cleanly to 6.12, will do so...

PR #6458, opened 2024-11-08
configs: Switch default cpufreq governor to ondemand

popcornmix:
Historically we have booted in powersave mode, and switched to ondemand with the /etc/init.d/raspi-config scipt if the shift key is not pressed.

This was intended to protect users from an unsafe overclock, or inadequate power supply.

But there's little evedence that this option is benefitting anyone, and we can boot faster without it.


popcornmix:
I have noticed that `CONFIG_CPU_FREQ_GOV_PERFORMANCE=y` disappears from defconfigs with this commit.
However it is still present in .config, so I think that is just a defconfig curiosity.

pelwell:
It is, because the cpufreq Kconfig says:
```
config CPU_FREQ_DEFAULT_GOV_PERFORMANCE
	bool "performance"
	select CPU_FREQ_GOV_PERFORMANCE
	help
	  Use the CPUFreq governor 'performance' as default. This sets
	  the frequency statically to the highest frequency supported by
	  the CPU.
```

popcornmix:
But I didn't enable `CPU_FREQ_DEFAULT_GOV_PERFORMANCE` - I enabled `CPU_FREQ_DEFAULT_GOV_ONDEMAND`

But for whatever reason, `CPU_FREQ_DEFAULT_GOV_ONDEMAND` does select `CPU_FREQ_GOV_PERFORMANCE`:
```
config CPU_FREQ_DEFAULT_GOV_ONDEMAND
        bool "ondemand"
        depends on !(X86_INTEL_PSTATE && SMP)
        select CPU_FREQ_GOV_ONDEMAND
        select CPU_FREQ_GOV_PERFORMANCE
        help
          Use the CPUFreq governor 'ondemand' as default. This allows
          you to get a full dynamic frequency capable system by simply
          loading your cpufreq low-level hardware driver.
          Be aware that not all cpufreq drivers support the ondemand
          governor. If unsure have a look at the help section of the
          driver. Fallback governor will be the performance governor.
```

pelwell:
Start again...

Setting the various governors as being the default implies a requirement for that governor, therefore no longer setting POWERSAVE as the default means that POWERSAVE is not automatically enabled. ONDEMAND is slightly different because it isn't available on all platforms, so asking for ONDEMAND to the default also enables PERFORMANCE as a fallback.

PR #6461, opened 2024-11-11
drivers: usb: xhci: set HID bit in streaming endpoint contexts

P33M:
This fixes @naushir's reboot IO errors and also stops shallow IO depths from getting spurious NAKed packets.

The rationale in the commit message applies globally, which is why this isn't an RP1-specific quirk. I believe the VL805 implements this field as ignored-1.


naushir:
Tested and this fixes my issue, LGTM!

PR #6463, opened 2024-11-12
drm/vc4: hvs: Defer updating the enable_bg_fill until vblank

6by9:
The register to enable/disable background fill was being set from atomic flush, however that will be applied immediately and can be a while before the vblank. If it was required for the current frame but not for the next one, that can result in corruption for part of the current frame.

Store the state in vc4_hvs, and update it on vblank.

@cillian64 for test.

### DIFF
--- a/drivers/gpu/drm/vc4/vc4_drv.h
+++ b/drivers/gpu/drm/vc4/vc4_drv.h
@@ -365,6 +365,8 @@ struct vc4_hvs {
 		unsigned int enabled: 1;
 	} eof_irq[HVS_NUM_CHANNELS];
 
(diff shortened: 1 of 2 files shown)


cillian64:
Thanks! Confirmed that this fixes the issue on my Pi5, and I don't see any other issues with background fill being enabled/disabled.  Do you reckon this wants separately testing on D0?

6by9:
D0 should make no difference here, although I have a 2GB Pi5 with D0 here if you want to do the quick test.

cillian64:
Happy to leave it if there's no particular D0 changes around this, it'll be a while before any software in the wild is actually using this.

popcornmix (approved):

PR #6464, opened 2024-11-13
dmaengine: dw-axi-dmac: Only start idle channels

pelwell:
Attempting to start a non-idle channel causes an error message to be logged, and is inefficient. Test for emptiness of the desc_issued list before doing so.


pelwell:
Any comments, @P33M? Given that the decision is taken under the same lock as all the queue handling, I believe it is safe. I2S and UART DMA have been tested.

popcornmix:
This has been merged through another route.

Issue #6004, opened 2024-03-03
HQ camera link frequency can cause GPS distruption


6by9:
### Describe the bug

From https://github.com/raspberrypi/libcamera/issues/43, the same issue of GPS interference has been noted on the HQ camera as well as the v3 that has already been fixed.
Look at tweaking the PLL settings to move the peak away from GPS on 1575MHz.

### Steps to reproduce the behaviour

See https://github.com/raspberrypi/libcamera/issues/43

### Device (s)

Raspberry Pi 4 Mod. B

### System

All kernels as the link frequency has never varied.

ranulphus:
Hi, any update on this? I see that it hasn't been taken on yet. Any tips on how to resolve it as I have a whole load of these cameras inside units with GPS and when then camera is on you can kiss goodbye to GPS signal!

6by9:
I had 30mins spare, so have done the quick patch to add it (based on imx708).
https://github.com/raspberrypi/linux/pull/6468

Give it 40mins for CI to run, and you should be able to get the test kernel using `sudo rpi-update pulls/6468`. Please ensure anything critical is backed up first.

Same as imx708, you can load `dtoverlay=imx477,link-frequency=447000000` or `dtoverlay=imx477,link-frequency=453000000` to shift the link frequency by those few MHz. 447MHz needs a little more testing before merging, but 453MHz should have no issues.

ranulphus:
This is great, I will give it a try with a new kernel.

Question...

I also have a legacy version that I need to support, with kernel 5.4.154. I build from a snapshot so I can apply code changes. Is it sufficient to just edit the line:

#define IMX477_DEFAULT_LINK_FREQ        450000000

Or could that cause problems and I need to apply the other changes to make it configurable etc?

ranulphus:
Ok that didn't work on the old kernel. Do I need to change some registers as well?

6by9:
Just changing `IMX477_DEFAULT_LINK_FREQ` will do nothing other than change the value advertised to userspace. Actually it may stop the driver probing as then the configured link frequency isn't the supposedly supported one.

If you want to do the quick hack, then changing register 0x030f from 0x96 to 0x97 or 0x98 in the table of registers would make the change without telling the rest of the system about it (it's not a critical thing for anything else to know about). Do note that it occurs multiple times.

ranulphus:
[...]

PR #6468, opened 2024-11-14
Add GPS avoidance option to IMX477

6by9:
https://github.com/raspberrypi/linux/issues/6004

Copied from imx708.
Not tested max data rates at lowest link frequency yet.


6by9:
1332x990 10bit mode fails at 447MHz.

6by9:
Updated to support 450MHz (default), but also 453 and 456MHz, dropping 447MHz.

naushir:
LGTM

PR #6470, opened 2024-11-15
RP1 PIO support

pelwell:
This patch set provides support for accessing the PIO hardware in RP1. Aside from the data FIFOs, the registers are inaccessible over PCIe, so all interactions must be proxied via the RP1 firmware.

The included pwm-pio driver and overlay provides up to 4 PWM channels on arbitrary header pins, and acts as an example of in-kernel usage.

Requires the latest Pi 5 EEPROM image. User-space support and some more examples can be found in the utils repo: https://github.com/raspberrypi/utils/pull/102

(Incorporates #6464.)


pelwell:
Note: I think there's a problem running on the 32-bit OS - probably the missing CONFIG_COMPAT support.

igorpecovnik:
and when compiling:

https://paste.armbian.de/yufeboteja
`error: include/uapi/misc/rp1_pio_if.h: missing "WITH Linux-syscall-note" for SPDX-License-Identifier`

pelwell:
That doesn't look like a regular build error - are you installing headers?

igorpecovnik:
> are you installing headers?

[Armbian build automation](https://github.com/armbian/os/actions/workflows/complete-artifact-matrix-all.yml) throws those errors out after this commit. I haven't look closer, but yes, it looks like headers are being installed in the process, line 20030 of paste. And I think they have to work.

pelwell:
Does the patch in https://github.com/raspberrypi/linux/pull/6472 solve the problem?

pelwell:
32-bit OS support now resolved by https://github.com/raspberrypi/linux/pull/6476.

PR #6472, opened 2024-11-16
fixup! misc: Add RP1 PIO driver

pelwell:
Fix for some Linux header licence rule.

Link: #6470


igorpecovnik (approved):
Fixes the problem with headers compilation.

pelwell:
Thanks - merged.

PR #6477, opened 2024-11-18
dtoverlays: enable SPI CS active-high

ikerexxe:
The documentation isn't very clear explaining how to enable SPI CS active-high and it takes a long time to understand it. Adding a specific overlay as a simple example on how to invert this signal can help understand the solution.

Link: https://forums.raspberrypi.com/viewtopic.php?t=378222

### DIFF
--- a/arch/arm/boot/dts/overlays/Makefile
+++ b/arch/arm/boot/dts/overlays/Makefile
@@ -259,6 +259,7 @@ dtbo-$(CONFIG_ARCH_BCM2835) += \
 	spi-rtc.dtbo \
 	spi0-0cs.dtbo \
 	spi0-1cs.dtbo \
+	spi0-1cs-inverted.dtbo \
 	spi0-2cs.dtbo \
 	spi1-1cs.dtbo \
 	spi1-2cs.dtbo \
--- a/arch/arm/boot/dts/overlays/README
+++ b/arch/arm/boot/dts/overlays/README
@@ -4438,6 +4438,14 @@ Params: cs0_pin                 GPIO pin for CS0 (default 8)
                                 it for other uses.
 
 
+Name:   spi0-1cs-inverted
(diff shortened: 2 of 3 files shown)


ikerexxe:
CC'ing @martinezjavier

pelwell:
You would be better off submitting your PR to rpi-6.6.y, from where we can forward port it. Otherwise you are in danger of having rpi-6.12.y rebased under your feet - which is what has just happened.

ikerexxe:
> You would be better off submitting your PR to rpi-6.6.y, from where we can forward port it. Otherwise you are in danger of having rpi-6.12.y rebased under your feet - which is what has just happened.

Done!

pelwell:
You'll need to fix these problems: https://github.com/raspberrypi/linux/actions/runs/11895664817/job/33146805156?pr=6477#step:4:559

The syntax error occurs because you aren't including `<dt-bindings/gpio/gpio.h>`, which you need for GPIO_ACTIVE_HIGH.

ikerexxe:
> You'll need to fix these problems: https://github.com/raspberrypi/linux/actions/runs/11895664817/job/33146805156?pr=6477#step:4:559
> 
> The syntax error occurs because you aren't including `<dt-bindings/gpio/gpio.h>`, which you need for GPIO_ACTIVE_HIGH.

Ok, thank you for the information.

pelwell:
The syntax error wasn't the only problem: https://github.com/raspberrypi/linux/actions/runs/11895664817/job/33146805156#step:4:561

The overlay itself is now fine, but you need entries in the README and the Makefile.

You may find it helpful to install `overlaycheck` from our utils repo: https://github.com/raspberrypi/utils/tree/master/overlaycheck

ikerexxe:
Yeah, that helped a lot

PR #6480, opened 2024-11-20
Enable Raspberry Touch 2 rotation with overlay

foonerd:
The panel-ilitek-ili9881c.c which Touch Display 2 is based on already contains orientation.
By minor update to the vc4-kms-dsi-ili9881-7inch-overlay.dts panel can be rotated with minimal effort.

Issue opened here: 
https://github.com/raspberrypi/documentation/issues/3923

Complete discussion and working dbto in the community thread:
https://community.volumio.com/t/guide-official-raspberry-touch-display-2/69791

### DIFF
--- a/arch/arm/boot/dts/overlays/README
+++ b/arch/arm/boot/dts/overlays/README
@@ -5249,6 +5249,7 @@ Params: sizex                   Touchscreen size x (default 720)
         invy                    Touchscreen inverted y axis
         swapxy                  Touchscreen swapped x y axis
         disable_touch           Disables the touch screen overlay driver
+        rotatation              Display rotation {0,90,180,270} (default 0)
         dsi0                    Use DSI0 and i2c_csi_dsi0 (rather than
                                 the default DSI1 and i2c_csi_dsi).
 
--- a/arch/arm/boot/dts/overlays/README
+++ b/arch/arm/boot/dts/overlays/README
@@ -5249,7 +5249,7 @@ Params: sizex                   Touchscreen size x (default 720)
         invy                    Touchscreen inverted y axis
         swapxy                  Touchscreen swapped x y axis
         disable_touch           Disables the touch screen overlay driver
-        rotation              Display rotation {0,90,180,270} (default 0)
+        rotation                Display rotation {0,90,180,270} (default 0)
         dsi0                    Use DSI0 and i2c_csi_dsi0 (rather than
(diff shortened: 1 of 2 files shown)


6by9:
Commits should include a commit text and `Signed-off-by:` line., but otherwise looks reasonable.

foonerd:
@6by9 
For my benefit, the commit line should be:
dtoverlays: Enable Raspberry Touch 2 rotation

Did I get this right?

Who should be listed as "Signed-off-by: "?

arch/arm/boot/dts/overlays/README
```diff
@@ -5249,6 +5249,7 @@ Params: sizex                   Touchscreen size x (default 720)
         invy                    Touchscreen inverted y axis
         swapxy                  Touchscreen swapped x y axis
         disable_touch           Disables the touch screen overlay driver
+        rotatation              Display rotation {0,90,180,270} (default 0)
```
6by9: As the dtoverlaycheck docs noticed, typo in `rotatation`

6by9 (commented):
`Signed-off-by:` is documented in https://www.kernel.org/doc/html/latest/process/submitting-patches.html#sign-your-work-the-developer-s-certificate-of-origin

It should match the commit author in this case. For upstream both would need use your real name rather than any nickname, but for dtoverlays that isn't as critical.

arch/arm/boot/dts/overlays/README
foonerd: Corrected.

6by9:
The patches get applied to the tree exactly as they are presented in the PR. The 3 changes therefore want to get squashed into one, and force pushed (`git push -f ... `) to your Github branch.

foonerd:
I see. Lets close this PR and open another, correct one. - Do you agree?

pelwell:
No - if the end result is one patch then we can squash at the point of merging.

foonerd:
Thanks. In such case, is there anything else needed from me at this point?

pelwell:
I think we're good - I'll just give it the once over then merge.

harmjanr:
Any clue when we can expect this PR / 6.6.63 via `sudo apt full-upgrade`?

6by9:
> Any clue when we can expect this PR / 6.6.63 via `sudo apt full-upgrade`?

Unlikely to be before Christmas as we've just done a kernel release.

PR #6481, opened 2024-11-20
drm: vc4: dsi: Handle the different command FIFO widths

6by9:
DSI0 and DSI1 have different widths for the command FIFO (24bit vs 32bit), but the driver was assuming the 32bit width of DSI1 in all cases.

Handle the difference via the variant structure.


6by9:
Confirmed as working.

PR #6482, opened 2024-11-20
Adding Pimidi kernel module.

gtrainavicius:
Submitting a new kernel module and its DT overlay for an upcoming Blokas product.


pelwell:
Let the auto-builds complete to look for any obvious problems before submitting any changes, but we would appreciate this PR being split into 3 commits: driver (including Kconfig and Makefile), defconfig files, and the overlay files.

pelwell:
Note: We're not bothered about long lines in overlays, required changes to the list of MAINTAINERS, or new compatible strings requiring dt-bindings changes, but the other checkpatch errors and warnings should be fixed.

sound/drivers/pimidi.c
```diff
@@ -0,0 +1,1050 @@
...
+	.trigger = pimidi_midi_input_trigger,
+};
+
+static int pimidi_register(struct pimidi_instance *instance)
+{
+	int err = 0;
+
+	mutex_lock(&pimidi_global.lock);
+	printd(instance, "Registering...");
+	if (!pimidi_global.reset_gpio) {
```
6by9, line 438: This looks very weird. You have unique reset gpios defined in the overlay for each of the address options, and yet you have this global state that after the first device has registered and acquired the GPIO, all subsequent ones don't bother. Why?

In fact why do you need this global state at all? I can't see any actually shared resource between multiple devices.

6by9: Take it back - I see it's the `data-ready` GPIOs that are unique, and reset isn't. That's rather a shame as it makes the driver far more complex.

gtrainavicius: Yes, the reset line is shared, each device in the stack gets a unique data-ready pin. Nothing to be ashamed of for trading some manageable software complexity for far more valuable commodity - GPIOs available to use for other purposes by the end users. :)

gtrainavicius:
I have split out the changes into the 3 commits as requested, as well as fixed the patch checks. Some warnings about pr_cont and a logging macro remain, but they are harmless for the module code.

sound/drivers/pimidi.c
```diff
@@ -0,0 +1,1119 @@
...
+		printe(instance, "snd_card_register failed: %d", err);
+		goto finalize;
+	}
+
+finalize:
+	if (err) {
+		instance->stopping = true;
+		cancel_work_sync(&instance->drdy_handler);
+		if (irq_initialized)
+			devm_free_irq(&client->dev, client->irq, instance);
```
pelwell: It should not be necessary to call `devm_free_irq` unless you are intended to release the IRQ before the device is released. Returning an error from the probe method is not one of those situations, so you can delete this line (along with all mentions of `irq_initialized`).

sound/drivers/pimidi.c
```diff
@@ -0,0 +1,1119 @@
...
+}
+
+static void pimidi_drdy_continue(struct pimidi_instance *instance)
+{
+	if (instance->stopping) {
+		printd(instance, "Refusing to queue work / enable IRQ due to stopping.");
+		return;
+	}
+
+	if (gpiod_get_value(instance->data_ready_gpio)) {
```
pelwell, line 688: Deciding whether to queue work or enable an interrupt based on the current state of the GPIO feels racy. Are you satisfied that it is safe? I guess it depends on whether the data ready line can spontaneously go high.

 I also notice that you are taking care to release the comm_lock mutex before calling this function - why?

sound/drivers/pimidi.c
```diff
@@ -0,0 +1,1119 @@
...
+		goto finalize;
+	}
+
+	err = pimidi_register(instance);
+	if (err < 0) {
+		printe(instance, "pimidi_register failed: %d", err);
+		goto finalize;
+	}
+
+	pimidi_perform_reset();
```
pelwell, line 907: Can you explain to me why you go to the trouble of resetting the other devices when each instance is probed? I understand that it is a shared reset line, but why can't you just reset when the first instance is probed? Without this, the driver would be much simpler.

sound/drivers/pimidi.c
gtrainavicius: The DRDY line remains high as set by the Pimidi device until all of its output is exhausted, this is a safe check.

The comm mutex is released to let the kernel module communicate with other Pimidi instances.

sound/drivers/pimidi.c
gtrainavicius: This supports the `sudo dtoverlay pimidi sel=x` usage when the devices must be loaded dynamically. As the reset is shared, each device must be re-initialized.

pelwell: One of us isn't understanding. If the reset is shared, isn't all the hardware reset by the first probe?

sound/drivers/pimidi.c
gtrainavicius: You are right, thank you, I've removed the lines in the latest push.

sound/drivers/pimidi.c
gtrainavicius: Each device must be initialized by a special 'handshake' via the DRDY GPIO. When a new device is added dynamically to the device tree, the kernel module can't know what is the current state of the new device, therefore, it must reset and do the 'handshake' procedure to all of them all over again, to ensure the state is consistent.

gtrainavicius: Keep in mind that when initial device is added via dtoverlay, the kernel module only knows about 1 of the DRDY pins, and a probe gets executed immediately. When a 2nd device gets added, it now knows about 2 DRDY pins, so it must repeat the procedure, as the new device must be reset, and as the reset is a shared line, it must re-initialize all the devices it knew of up till now.

gtrainavicius:
Thank you! :)

Issue #6484, opened 2024-11-21
[RPi5] [rpi-6.12.y]: nvme no worky


ElDavoo:
### Describe the bug

On 6.12 kernels, my nvme SSD does not get detected (while the PCIe bridge does), and the status LED of the shield does not turn on.
It does work fine with 6.6.y and 6.11.y.

### Steps to reproduce the behaviour

Compiling the 6.12.y (currently at ab9bb48 ) with bcm2712_defconfig.

### Device (s)

Raspberry Pi 5

### System

```
Raspberry Pi reference 2024-11-13
Generated using pi-gen, https://github.com/RPi-Distro/pi-gen, 919f1e86b959321edaa8266ee271e5d0870f5298, stage2
```

```
2024/11/12 16:10:44 
Copyright (c) 2012 Broadcom
version 4b019946 (release) (embedded)
```

`Linux pi5 6.11.9-v8-16k+ #3 SMP PREEMPT Thu Nov 21 13:14:34 GMT 2024 aarch64 GNU/Linux`
`Linux pi5 6.12.0-v8-16k+ #4 SMP PREEMPT Thu Nov 21 17:37:01 GMT 2024 aarch64 GNU/Linux`

## 6.11
[kmsg11.txt](https://github.com/user-attachments/files/17850079/kmsg11.txt)

## 6.12
[kmsg12.txt](https://github.com/user-attachments/files/17850074/kmsg12.txt)

### Additional context

# Context
1.  Shield model: Geekworm X1001
2.  SSD model: Ediloca EN600 Pro

config.txt contains `dtparam=pciex1_gen=3`

# 6.11
## lspci
```
0000:00:00.0 PCI bridge: Broadcom Inc. and subsidiaries BCM2712 PCIe Bridge (rev 21)
0000:01:00.0 Non-Volatile memory controller: MAXIO Technology (Hangzhou) Ltd. NVMe SSD Controller MAP1202 (rev 01)
0001:00:00.0 PCI bridge: Broadcom Inc. and subsidiaries BCM2712 PCIe Bridge (rev 21)
0001:01:00.0 Ethernet controller: Raspberry Pi Ltd RP1 PCIe 2.0 South Bridge
```
## lspci -v
```
0000:01:00.0 Non-Volatile memory controller: MAXIO Technology (Hangzhou) Ltd. NVMe SSD Controller MAP1202 (rev 01) (prog-if 02 [NVM Express])
[...]

PR #6487, opened 2024-11-22
Set pcie link speed before unreset, not after

P33M:
Trial fix for https://github.com/raspberrypi/linux/issues/6484


ElDavoo:
Works perfectly, many thanks.

PR #6494, opened 2024-11-26
spi: atmel-quadspi: minor bug in register name in debug message

magdalena-trivina:


ragazenta:
This patch has been already applied at upstream.

See: https://lore.kernel.org/linux-spi/20241122141302.2599636-1-csokas.bence@prolan.hu/

magdalena-trivina:
> This patch has been already applied at upstream.
> 
> See: https://lore.kernel.org/linux-spi/20241122141302.2599636-1-csokas.bence@prolan.hu/

Thanks for the info.

PR #6496, opened 2024-11-26
TESTING: V6.12 dwc hcd fixes

popcornmix:
Do not merge - just to create a test build.


lategoodbye:
I think, this can be closed because the relevant patches have been applied mainline.

Thanks

popcornmix:
Thanks. Look like the patches are available from 6.6.67 (and are also on current 6.12 and 6.13 trees).

PR #6499, opened 2024-11-29
waveshare panel driver update

waveshare:
1. add waveshare 13.3 inch panel support 
2. add 4lane support

### DIFF
--- a/arch/arm/boot/dts/overlays/README
+++ b/arch/arm/boot/dts/overlays/README
@@ -5324,6 +5324,8 @@ Params: 2_8_inch                2.8" 480x640
         8_0_inch                8.0" 1280x800
         10_1_inch               10.1" 1280x800
         11_9_inch               11.9" 320x1480
+        13_3_inch_4lane         13.3" 1920x1080 4lane
+        13_3_inch_2lane         13.3" 1920x1080 2lane
         i2c1                    Use i2c-1 with jumper wires from GPIOs 2&3
         disable_touch           Disable the touch controller
         rotation                Set the panel orientation property
--- a/arch/arm/boot/dts/overlays/vc4-kms-dsi-waveshare-panel-overlay.dts
+++ b/arch/arm/boot/dts/overlays/vc4-kms-dsi-waveshare-panel-overlay.dts
@@ -51,6 +51,11 @@
 				reg = <0x14>;
 				compatible = "goodix,gt911";
 			};
+
+			touch2: ilitek@41 {
+				compatible = "ilitek,ili251x";
+				reg = <0x41>;
+			};
 		};
 	};
 
@@ -120,6 +125,8 @@
 				   <&touch>, "touchscreen-inverted-x?",
 				   <&touch>, "touchscreen-inverted-y?";
 		8_8_inch = <&panel>, "compatible=waveshare,8.8inch-panel";
+		13_3_inch_4lane = <&panel>, "compatible=waveshare,13.3inch-4lane-panel";
+		13_3_inch_2lane = <&panel>, "compatible=waveshare,13.3inch-2lane-panel";
 		i2c1 = <&i2c_frag>, "target:0=",<&i2c1>,
 		       <0>, "-3-4+5";
 		disable_touch = <&touch>, "status=disabled";
--- a/drivers/input/touchscreen/ili210x.c
+++ b/drivers/input/touchscreen/ili210x.c
@@ -67,6 +67,8 @@ struct ili210x {
 	u8 version_proto[2];
 	u8 ic_mode[2];
 	bool stop;
+	struct timer_list poll_timer;
+	struct work_struct poll_work;
 };
 
 static int ili210x_read_reg(struct i2c_client *client,
@@ -358,6 +360,34 @@ static irqreturn_t ili210x_irq(int irq, void *irq_data)
 	} while (!priv->stop && keep_polling);
 
 	return IRQ_HANDLED;
+}
+
+static void ili210x_poll_work(struct work_struct *work)
+{
+	struct ili210x *priv = container_of(work, struct ili210x, poll_work);
+	struct i2c_client *client = priv->client;
+	const struct ili2xxx_chip *chip = priv->chip;
+	u8 touchdata[ILI210X_DATA_SIZE] = { 0 };
+	bool touch;
+	int error;
+
+	error = chip->get_touch_data(client, touchdata);
+	if (error) {
+		dev_err(&client->dev, "Unable to get touch data: %d\n", error);
+		return;
+	}
+
+	touch = ili210x_report_events(priv, touchdata);
+}
+
+static void ili210x_poll_timer_callback(struct timer_list *t)
+{
+	struct ili210x *priv = from_timer(priv, t, poll_timer);
+
+	schedule_work(&priv->poll_work);
+
+	if (!priv->stop)
+		mod_timer(&priv->poll_timer, jiffies + msecs_to_jiffies(ILI2XXX_POLL_PERIOD));
 }
 
 static int ili251x_firmware_update_resolution(struct device *dev)
@@ -945,11 +975,6 @@ static int ili210x_i2c_probe(struct i2c_client *client)
 		return -ENODEV;
 	}
 
-	if (client->irq <= 0) {
-		dev_err(dev, "No IRQ!\n");
-		return -EINVAL;
-	}
-
 	reset_gpio = devm_gpiod_get_optional(dev, "reset", GPIOD_OUT_HIGH);
 	if (IS_ERR(reset_gpio))
 		return PTR_ERR(reset_gpio);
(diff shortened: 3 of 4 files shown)


arch/arm/boot/dts/overlays/README
```diff
@@ -5324,6 +5324,8 @@ Params: 2_8_inch                2.8" 480x640
         8_0_inch                8.0" 1280x800
         10_1_inch               10.1" 1280x800
         11_9_inch               11.9" 320x1480
+	13.3inch-30fps        13.3″ 1920x1080 2lane
+	13.3inch-60fps        13.3″ 1920x1080 4lane
```
pelwell: Replace the TABs with spaces, fix the indentation so it all lines up, and get the parameter names right!

arch/arm/boot/dts/overlays/vc4-kms-dsi-waveshare-panel-overlay.dts
```diff
@@ -120,6 +126,8 @@
 				   <&touch>, "touchscreen-inverted-x?",
 				   <&touch>, "touchscreen-inverted-y?";
 		8_8_inch = <&panel>, "compatible=waveshare,8.8inch-panel";
+		13_3_60fps_inch = <&panel>, "compatible=waveshare,13.3inch-60fps-panel";
+		13_3_30fps_inch = <&panel>, "compatible=waveshare,13.3inch-30fps-panel";
```
pelwell: These names look wrong - `13_3_inch_60fps` and `13_3_inch_30fps` make more sense.

waveshare: Yeah, I found the problem. I'm working on it.
I think it would be more appropriate to read `13_3_inch_4lane`.

waveshare:
Wait a minute. It looks like some of the files haven't been updated yet.

waveshare:
Changes have been completed. PTAL.

pelwell:
Fix the tabs again, then all the checkpatch errors and warnings (all of them look reasonable) and squash back down to 3 commits - driver, config and DT.

drivers/input/touchscreen/ilitek_2511/ilitek_platform_init.c
```diff
@@ -0,0 +1,419 @@
...
+}
+
+#endif
+
+
+#endif
+module_i2c_driver(ilitek_touch_device_driver);
+
+MODULE_AUTHOR("waveshare");
+MODULE_DESCRIPTION("WS_xinchDSI_Touch Driver");
```
6by9: Really? The copyright headers all list ILI Technology Corporation and Luca Hsu.

drivers/input/touchscreen/ilitek_2511/ilitek_platform_init.c
```diff
@@ -0,0 +1,419 @@
...
+/* The I2C device list is used for matching I2C device and I2C device driver. */
+static const struct i2c_device_id tpd_device_id[] = {
+	{ILITEK_TS_NAME, 0},
+	{},			/* should not omitted */
+};
+
+MODULE_DEVICE_TABLE(i2c, tpd_device_id);
+
+const struct of_device_id touch_dt_match_table[] = {
+	{.compatible = "tchip,ilitek2" },
```
6by9: Who is tchip as a manufacturer? It's not defined in https://github.com/torvalds/linux/blob/master/Documentation/devicetree/bindings/vendor-prefixes.yaml

6by9 (commented):
Is there a need for this new touch driver?

There is already https://github.com/torvalds/linux/blob/master/drivers/input/touchscreen/ilitek_ts_i2c.c which claims to be for the Ilitek 23XX, 25XX, and Lego chips. Which chip are you actually using on this display, and is it compatible?

drivers/input/touchscreen/ilitek_2511/ilitek_platform_init.c
waveshare: I'll check with my coworker that this TP driver came directly to us from our supplier

6by9:
I've just noticed that we already have an [ilitek251x overlay](https://github.com/raspberrypi/linux/blob/rpi-6.6.y/arch/arm/boot/dts/overlays/ilitek251x-overlay.dts) in the tree using the ili210x driver that lists an [`ilitek,ili251x` compatible](https://github.com/raspberrypi/linux/blob/rpi-6.6.y/drivers/input/touchscreen/ili210x.c#L1045).

I'd hope either ili210x or ilitek_ts_i2c would support your touch controller.

drivers/input/touchscreen/ili210x.c
```diff
@@ -1001,12 +1027,18 @@ static int ili210x_i2c_probe(struct i2c_client *client)
...
-		return error;
+	if (client->irq) {
+		error = devm_request_threaded_irq(dev, client->irq, NULL, ili210x_irq,
+					IRQF_ONESHOT, client->name, priv);
+		if (error) {
+			dev_err(dev, "Unable to request touchscreen IRQ, err: %d\n",error);
+			return error;
+		}
+	}
+	else{
```
6by9: Space before the {

drivers/input/touchscreen/ili210x.c
```diff
@@ -1001,12 +1027,18 @@ static int ili210x_i2c_probe(struct i2c_client *client)
...
+		timer_setup(&priv->poll_timer, ili210x_poll_timer_callback, 0);
+		mod_timer(&priv->poll_timer, jiffies + msecs_to_jiffies(ILI2XXX_POLL_PERIOD));
+		INIT_WORK(&priv->poll_work, ili210x_poll_work);
```
6by9, line 1039: You don't cancel the workqueue or delete the timer on remove.
See https://github.com/raspberrypi/linux/commit/818bb6d49db23300ab816fe6e1e71b6387f79d96 for doing the same on edt-ft5x06(*).

In this case it probably wants to go into ili210x_stop though as they've avoided adding a remove function.

(*) Looking at it, I think the cancel_work_sync should come first.
Comparing to [tc358743](https://elixir.bootlin.com/linux/v6.12.1/source/drivers/media/i2c/tc358743.c#L2186) which also has a polling option, that uses slightly different semantics again - there never is one clear option, and it needs careful thought over what triggers what.

6by9 (commented):
Thank you for the update - using the mainline drivers is always preferable to adding yet another downstream one.

I've made comment on one coding style issue that checkpatch will complain over. This is the sort of patch that does want to be sent upstream, so we will be quite strict over it. CI has now run checkpatch for this PR, so please fix the errors.

drivers/input/touchscreen/ili210x.c
waveshare: Thank you very much for your valuable guidance on maintaining coding standards, especially for patches intended for upstream submission.I will promptly address the issues identified by checkpatch and correct the coding style problem you pointed out. Once the updates are complete, I will submit the revised version for further review.

waveshare:
Thank you for your feedback and guidance, the changes are as above, please feel free to let me know if further improvements are needed or additional testing is recommended.
 I'm happy to make adjustments to ensure the patch meets upstream standards.

pelwell:
You keep repeating the same mistakes. Download `overlaycheck` from our utils repo (https://github.com/raspberrypi/utils/tree/master/overlaycheck) to get it right before submitting.

waveshare:
> You keep repeating the same mistakes. Download `overlaycheck` from our utils repo (https://github.com/raspberrypi/utils/tree/master/overlaycheck) to get it right before submitting.

Ok, we will pay attention to this problem in the future

waveshare:
Already passed `overlaycheck`.image

pelwell:
Yes, this one is looking OK.

6by9:
Looks reasonable to me.

I'm not sure which order the ili210x_stop devm_action and the remove hook will be called in, and that could cause issues. Testing with `sudo rmmod ili210x` to confirm it doesn't crash, or better still adding a debug line in `remove` confirms that `stop` has already been set would be nice to know it's OK. However unloading modules is a rare enough thing that I'm not going to fuss.

waveshare:
> Looks reasonable to me.
> 
> I'm not sure which order the ili210x_stop devm_action and the remove hook will be called in, and that could cause issues. Testing with `sudo rmmod ili210x` to confirm it doesn't crash, or better still adding a debug line in `remove` confirms that `stop` has already been set would be nice to know it's OK. However unloading modules is a rare enough thing that I'm not going to fuss.

Thank you for your suggestions! I tested unloading the module using sudo rmmod ili210x, and there were no crashes during the process. I also added a debug line in the remove function to confirm that stop had already been set. All tests ran smoothly without any issues.
image
I agree that unloading modules is not a common operation, so it’s unlikely to be a major concern in practice. Still, it’s good to have verified this just in case.

Thank you all again for taking the time to share your suggestions and insights!

PR #6506, opened 2024-12-02
dtoverlays: Add Arducam override for ov9281

6by9:
The Arducam module is slow starting up, so add an override to slow the regulator down.
https://forums.raspberrypi.com/viewtopic.php?t=380236


pelwell (approved):

6by9:
Confirmed as working at https://forums.raspberrypi.com/viewtopic.php?t=380236, so merging.

Issue #6504, opened 2024-11-30
Pi 5 kernel8.img for 6.6.62 causes audio players to freeze with a looping sound


theflyingape:
### Describe the bug

**config.txt**
```
[pi5]
kernel=kernel8.img
```

### Steps to reproduce the behaviour

`ffplay sound.wav` race condition with studdering sound until killed.
If I comment out kernel8.img, ffplay runs fine, but some other software will segfault without kernel8.

### Device (s)

Raspberry Pi 5

### System

`uname -r`
6.6.62+rpt-rpi-v8

If I rollback to 6.6.51, it kernel8.img and sound events run fine.

popcornmix:
What device are you outputting audio to? (e.g. hdmi?)
Is this RPiOS bookworm 64-bit? Lite or desktop image?

theflyingape:
Yes, latest Bookworm release and output to HDMI-0 -- both from LightDM with Wayland+labwc and from console mode.

theflyingape:
I can confirm rolling back the linux-image-6.6.62 --> linux-image-6.6.51 with the kernel8.img enabled works as expected.

pelwell:
I'm not a  regular ffpmeg user, but it's working for me to HDMI0. There is no audible difference between kernel8.img and kernel_2712.img.

1. What sample format and rate is `sound.wav`? If you've used several formats, does the behaviour differ between them?
2. Does `speaker-test -c 2` stutter?
3. How about `aplay sound.wav`?

theflyingape:
Just tried `sudo apt install linux-image-6.6.62+rpt-rpi-2712 linux-image-6.6.62+rpt-rpi-v8` on Pi 5, rebooted with and without kernel8.img, and sound stutters regardless of format (wav, mp3) which includes playing video (flv, mp4) and the studder stalls video playback. I also tested with and without wireplumber and pipewire services.

`speaker-test` does stutter and I cannot Ctrl-C to stop -- must wait 30-seconds.
`aplay` the same -- but interesting that ffplay, mplayer, mpv all work on Pi 4 (not Pi 5 as reported) but aplay fails.

Only on Pi4, I must stop wireplumber (not pipewire) else mplayer immediately reports "Audio device got stuck!"
On Pi 5, stopping all and sending it to ALSA device reports "Audio device got stuck!" repeatedly.

pelwell:
[...]

PR #6510, opened 2024-12-03
nvme-pci: Disable Host Memory Buffer usage

pelwell:
Some NVME drives seem to request significant amounts of DMA coherent memory - enough to exhaust our standard 64MB CMA allocation.

Try disabling the feature to see what effect it has - drives should continue to function without it.

Link: https://github.com/raspberrypi/linux/issues/6504

### DIFF
--- a/drivers/nvme/host/pci.c
+++ b/drivers/nvme/host/pci.c
@@ -1936,6 +1936,7 @@ static void nvme_free_host_mem(struct nvme_dev *dev)
 	dev->nr_host_mem_descs = 0;
 }
 
+#if 0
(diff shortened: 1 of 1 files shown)


P33M:
NVM Express v1.4a s8.9:

> 8.9 Host Memory Buffer (Optional)

>The Host Memory Buffer (HMB) feature allows the controller to utilize an assigned portion of host memory exclusively. The use of the host memory resources is vendor specific. Host software may not be able to provide any or a limited amount of the host memory resources requested by the controller. The controller shall function properly without host memory resources.

So drives shouldn't fail if the allocation fails, but throughput may suffer.

PR #6514, opened 2024-12-04
drm/v3d: Enable Performance Counters before clearing them

mairacanal:
On the Raspberry Pi 5, performance counters are not being cleared when `v3d_perfmon_start()` is called, even though we write to the CLR register. As a result, their values accumulate until they overflow.

The expected behavior is for performance counters to reset to zero at the start of a job. When the job finishes and the perfmon is stopped, the counters should accurately reflect the values for that specific job.

To ensure this behavior, the performance counters are now enabled before being cleared. This allows the CLR register to function as intended, zeroing the counter values when the job begins.

---

This patch was already submitted to the dri-devel mailing list [1] and will eventually reach stable-6.6. The idea is just to speed up the process of reaching the downstream repository.

[1] https://lore.kernel.org/dri-devel/20241204122831.17015-1-mcanal@igalia.com/T/

### DIFF
--- a/drivers/gpu/drm/v3d/v3d_perfmon.c
+++ b/drivers/gpu/drm/v3d/v3d_perfmon.c
@@ -51,9 +51,9 @@ void v3d_perfmon_start(struct v3d_dev *v3d, struct v3d_perfmon *perfmon)
 		V3D_CORE_WRITE(0, V3D_V4_PCTR_0_SRC_X(source), channel);
 	}
 
+	V3D_CORE_WRITE(0, V3D_V4_PCTR_0_EN, mask);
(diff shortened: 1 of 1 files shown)


popcornmix:
Thanks - looks sensible to me.

mairacanal:
It was already reviewed and applied to drm-misc-fixes [1]

[1] https://gitlab.freedesktop.org/drm/misc/kernel/-/commit/c98b10496b2f3c4f576af3482c71aadcfcbf765e

popcornmix:
@pelwell any objections?

pelwell:
Nope.

PR #6515, opened 2024-12-04
input: edt-ft5x06: Extend the power on delay

6by9:
There's a report of a touch panel not being identified correctly, but unloading and reloading the module brings it to life.

About the only thing this can be is that it hasn't come out of reset properly, so extend that delay.

https://github.com/raspberrypi/bookworm-feedback/issues/330


pelwell:
This patch is TL;DR - it's probably fine, though.

6by9:
It doesn't help in the OP's case :-( 

The touch controller fails to probe at boot, but probes fine if you `rmmod` and then `modprobe` the module again.

6by9:
Closing as it didn't help the issue.

PR #6522, opened 2024-12-09
media: i2c: ov9282: Correct the exposure offset

6by9:
The datasheet lists that "Maximum exposure time is frame length -25 row periods, where frame length is set by registers {0x380E, 0x380F}".
However this driver had OV9282_EXPOSURE_OFFSET set to 12 which allowed that restriction to be violated, and would result in very under-exposed images.

Correct the offset.

Fixes: 14ea315bbeb7 ("media: i2c: Add ov9282 camera sensor driver")


6by9:
Reported https://forums.raspberrypi.com/viewtopic.php?p=2274970#p2274970
I'll be sending the patch upstream when I get a chance.

6by9:
Sent upstream and has a R-b
https://lore.kernel.org/linux-media/20241209-media-ov9282-fix-v1-1-d06bb7546f18@raspberrypi.com/

PR #6524, opened 2024-12-09
cgroup: Add cgroup_enable option

pelwell:
The upstream addition of the kernel parameter cgroup_disable makes it possible to configure cgroups at boot time. In theory, re-enabling a disabled cgroup is simply a case of removing the relevant cgroup_disable setting, but this is difficult if the setting comes from Device Tree.

Re-introduce cgroup_enable as a way around the problem.


popcornmix (approved):

pelwell:
Thanks - I believe this is correct, but it may not be sufficient if legacy V1 memory cgroup support is needed.

PR #6525, opened 2024-12-09
drm/v3d: Correct clock settng calls to new APIs

popcornmix:
There was a report that 6.12 kernel has lower benchmark scores than 6.6.

I can confirm, and found it started with 6.8 kernel which moved some code into a new file (v3d_submit.c) and in two places the change to the clock api were missed.

The effect of the bug is the v3d clock sometimes
unwantedly drops to a lower rate.

With this patch the benchmark scores are good again.

Fixes: 86963038cb0c0ba99dad14665ca92936ca723f89

### DIFF
--- a/drivers/gpu/drm/v3d/v3d_submit.c
+++ b/drivers/gpu/drm/v3d/v3d_submit.c
@@ -18,7 +18,7 @@ v3d_clock_down_work(struct work_struct *work)
 		container_of(work, struct v3d_dev, clk_down_work.work);
 	int ret;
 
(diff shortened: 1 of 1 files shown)


popcornmix:
This was original [bug report](https://forums.raspberrypi.com/viewtopic.php?p=2273685#p2273685).

6by9:
Is this only a downstream patch? I'm not sure why this hasn't been upstreamed as all the clock APIs have been.

popcornmix:
Yes, downstream only. Upstream doesn't reduce clock when idle.

PR #6529, opened 2024-12-10
RP1 DPI: Add interlaced modes

njhollinghurst:
Add support for interlaced DPI modes on Raspberry Pi 5. (Draft for testing, based on rpi-6.6.y)

In order for this to work correctly, DPI's DE output must be enabled on GPIO1 (whether used or not). PIO can then snoop on DE and HSYNC signals to generate a fixed-up VSYNC on GPIO2.

This is the cut-down version of the patch which omits CSYNC support, but it should still be possible to implement CSYNC (and equalizing pulses) using an external PIO program; I will need to update the example in [utils/piolib/examples](https://github.com/raspberrypi/utils/blob/master/piolib/examples/dpi_interlace.c).

### DIFF
--- a/drivers/gpu/drm/bridge/panel.c
+++ b/drivers/gpu/drm/bridge/panel.c
@@ -82,6 +82,8 @@ static int panel_bridge_attach(struct drm_bridge *bridge,
 		return ret;
 	}
 
+	connector->interlace_allowed = true;
+
 	drm_panel_bridge_set_orientation(connector, bridge);
 
 	drm_connector_attach_encoder(&panel_bridge->connector,
--- a/drivers/gpu/drm/rp1/rp1-dpi/rp1_dpi_hw.c
+++ b/drivers/gpu/drm/rp1/rp1-dpi/rp1_dpi_hw.c
@@ -202,7 +202,7 @@
 // Status
 #define DPI_DMA_STATUS  0x3c
 
-#define BITS(field, val) (((val) << (field ## _SHIFT)) & (field ## _MASK))
+#define BITS(field, val) FIELD_PREP((field ## _MASK), val)
 
 static unsigned int rp1dpi_hw_read(struct rp1_dpi *dpi, unsigned int reg)
 {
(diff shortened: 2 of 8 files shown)


drivers/gpu/drm/bridge/panel.c
```diff
@@ -82,6 +82,8 @@ static int panel_bridge_attach(struct drm_bridge *bridge,
 		return ret;
 	}
 
+	connector->interlace_allowed = true;
```
6by9, line 85: Commit text should be `drm: bridge: panel: Connector to allow interlaced modes`

drivers/gpu/drm/rp1/rp1-dpi/rp1_dpi_hw.c
```diff
@@ -417,6 +401,80 @@ void rp1dpi_hw_setup(struct rp1_dpi *dpi,
 			BITS(DPI_DMA_QOS_LLEV, 0x8) |
 			BITS(DPI_DMA_QOS_LQOS, 0x7));
 
+	if (!(mode->flags & DRM_MODE_FLAG_INTERLACE)) {
+		rp1dpi_hw_write(dpi, DPI_DMA_VISIBLE_AREA,
+				BITS(DPI_DMA_VISIBLE_AREA_ROWSM1, mode->vdisplay - 1) |
+				BITS(DPI_DMA_VISIBLE_AREA_COLSM1, mode->hdisplay - 1));
+
+		rp1dpi_hw_write(dpi, DPI_DMA_SYNC_WIDTH,
+				BITS(DPI_DMA_SYNC_WIDTH_ROWSM1, mode->vsync_end - mode->vsync_start - 1) |
```
6by9: As per the checkpatch error, break the line after `DPI_DMA_SYNC_WIDTH_ROWSM1,`

Same for the other errors.

drivers/gpu/drm/rp1/rp1-dpi/rp1_dpi_hw.c
```diff
@@ -417,6 +401,80 @@ void rp1dpi_hw_setup(struct rp1_dpi *dpi,
...
+		rp1dpi_hw_write(dpi, DPI_DMA_BACK_PORCH,
+				BITS(DPI_DMA_BACK_PORCH_ROWSM1, mode->vtotal - mode->vsync_start - 1) |
+				BITS(DPI_DMA_BACK_PORCH_COLSM1, mode->htotal - mode->hsync_start - 1));
+
+		rp1dpi_hw_write(dpi, DPI_DMA_FRONT_PORCH,
+				BITS(DPI_DMA_FRONT_PORCH_ROWSM1, mode->vsync_start - mode->vdisplay - 1) |
+				BITS(DPI_DMA_FRONT_PORCH_COLSM1, mode->hsync_start - mode->hdisplay - 1));
+
+		vctrl = BITS(DPI_DMA_CONTROL_VSYNC_POL, !!(mode->flags & DRM_MODE_FLAG_NVSYNC)) |
+			BITS(DPI_DMA_CONTROL_VBP_EN,    (mode->vtotal != mode->vsync_start))    |
```
6by9: Remove the excess white space. Normal coding style wouldn't align this type, only in tables.

drivers/gpu/drm/rp1/rp1-dpi/rp1_dpi_hw.c
```diff
@@ -417,6 +401,80 @@ void rp1dpi_hw_setup(struct rp1_dpi *dpi,
...
+		 *
+		 * RP1 DPI hardware wasn't designed to support interlace, but lets us change
+		 * both the VFP line count and the next DMA address while running. That allows
+		 * pixel data to be correctly timed for interlace, but VSYNC remains wrong.
+		 *
+		 * It is necessary to use external hardware (such as PIO) to regenerate VSYNC
+		 * based on HSYNC, DE (which *must* both be mapped to GPIOs 1, 3 respectively).
+		 * This driver includes a PIO program to do that, when DE is enabled.
+		 *
+		 * An alternative fixup is to synthesize CSYNC from HSYNC and modified-VSYNC.
```
6by9, line 450: Have you still got CSYNC support? I thought you'd said it had been dropped.

drivers/gpu/drm/rp1/rp1-dpi/rp1_dpi_pio.c
```diff
@@ -0,0 +1,225 @@
...
+ * (one half-line delay is needed to sample DE; we always wait for the next
+ * half-line boundary to improve VSync start accuracy).
+ */
+
+static int rp1dpi_pio_vsync_ilace(struct rp1_dpi *dpi,
+				  struct drm_display_mode const *mode)
+{
+	static const int wrap_target = 14;
+	static const int wrap = 26;
+	u16 instructions[] = {  /* This is mutable */
```
6by9, line 84: Using statics means we can't instantiate the driver twice. Unlikely ever to happen, but I'd be tempted to put it in struct rpi1_dpi instead.

6by9:
BTW I've just noticed that the BITS macro is a roll-it-yourself shift.
There are macros `FIELD_FIT` `FIELD_PREP`, and `FIELD_SET` in include/linux/bitfield.h which also do size checking to ensure you don't end up writing too many bits into a field. See vc4_regs.h for example usage.

drivers/gpu/drm/rp1/rp1-dpi/rp1_dpi_hw.c
njhollinghurst: DPI VSYNC waveform is somewhat arbitrary (since it can't be correct for interlace). Here it's designed to assist CSYNC generation. But I removed the PIO code that actually does generate CSYNC (for code bloat reduction).

drivers/gpu/drm/rp1/rp1-dpi/rp1_dpi_pio.c
njhollinghurst: These are just named constants. The PIO compiler spits out lower-case `#define`s but I avoided those because they seemed unsightly, and there were formerly 3 different functions that defined similar-named constants differently.

6by9: I'd misread `instructions` as being static.

I assume the program gets copied into RP1 / PIO in some shape or form, so this doesn't need to remain in context after the `pio_add_program` call.

njhollinghurst:
> There are macros `FIELD_FIT` `FIELD_PREP`, and `FIELD_SET` in include/linux/bitfield.h which also do size checking to ensure you don't end up writing too many bits into a field. See vc4_regs.h for example usage.

I'll see if I can use those... ~I'd like to keep the field definitions unmodified though, as they're taken straight from VRBuild output.~ Oh, I see that I've already edited them since the VRB.

drivers/gpu/drm/rp1/rp1-dpi/rp1_dpi_pio.c
njhollinghurst: Ah, it's the other way around: instructions[] is mutable because it gets subjected to some runtime fixups (rather than have 4 or 8 different versions for different sync polarities). But I can change the style if it's unclear.

I assumed (perhaps wrongly) that using a run-time initializer is no less efficient than explicitly making a copy. And yes, it's not needed once it's been loaded into PIO instruction memory.

njhollinghurst:
Ugh. FIELD_PREP() can't be used outside a function (e.g. as a static initializer for a table) because it contains bracketed code. So I would need separate `IMASK_RGB` and `IMASK_RGB_CONST` macros.

pelwell (approved):

njhollinghurst:
OK I just need to check it still works with the external/example program, and may adjust the offset of the "VSync helper" signal (which should make no difference to the built-in H/VSync case, that doesn't use it).

Would be nice to keep the door open for equalizing pulses, but as the generally-accepted "PAL" modeline has 576 _complete_ lines, and I don't want to make a special case for it, it seems slightly doomed: obviously we can't put an equalizing pulse in the middle of an active scanline.

igorpecovnik:
This PR breaks our auto-compilation:
https://github.com/armbian/os/actions/runs/12292566594/job/34305984530
https://paste.armbian.de/qiqusamowu

```
   ERROR: modpost: "pio_close" [drivers/gpu/drm/rp1/rp1-dpi/drm-rp1-dpi.ko] undefined!
   ERROR: modpost: "pio_set_error" [drivers/gpu/drm/rp1/rp1-dpi/drm-rp1-dpi.ko] undefined!
   ERROR: modpost: "rp1_pio_sm_unclaim" [drivers/gpu/drm/rp1/rp1-dpi/drm-rp1-dpi.ko] undefined!
   ERROR: modpost: "pio_open" [drivers/gpu/drm/rp1/rp1-dpi/drm-rp1-dpi.ko] undefined!
   ERROR: modpost: "rp1_pio_sm_claim" [drivers/gpu/drm/rp1/rp1-dpi/drm-rp1-dpi.ko] undefined!
   ERROR: modpost: "rp1_pio_sm_init" [drivers/gpu/drm/rp1/rp1-dpi/drm-rp1-dpi.ko] undefined!
   ERROR: modpost: "rp1_pio_sm_set_enabled" [drivers/gpu/drm/rp1/rp1-dpi/drm-rp1-dpi.ko] undefined!
   ERROR: modpost: "rp1_pio_add_program" [drivers/gpu/drm/rp1/rp1-dpi/drm-rp1-dpi.ko] undefined!
   ERROR: modpost: "rp1_pio_sm_exec" [drivers/gpu/drm/rp1/rp1-dpi/drm-rp1-dpi.ko] undefined!
   ERROR: modpost: "rp1_pio_gpio_set_function" [drivers/gpu/drm/rp1/rp1-dpi/drm-rp1-dpi.ko] undefined!
   WARNING: modpost: suppressed 3 unresolved symbol warnings because there were too many)
   make[3]: *** [scripts/Makefile.modpost:145: Module.symvers] Error 1
   make[2]: *** [Makefile:1873: modpost] Error 2
   make[1]: *** [/armbian/cache/sources/linux-kernel-worktree/6.6__bcm2711__arm64/Makefile:359: __build_one_by_one] Error 2
   make: *** [Makefile:234: __sub-make] Error 2
```

6by9:
drivers/gpu/drm/rp1/rp1-dpi/Kconfig needs a `depends on RP1_PIO` or `select RP1_PIO`.

CONFIG_RP1_PIO is in our standard bcm2712_defconfig, so does that mean you're using your own?

pelwell:
Odd - our autobuilds are happy, as are my own builds.

Are you using a different defconfig? I notice that CONFIG_DRM_RP1_DPI is lacking a dependency on CONFIG_RP1_PIO, which could explain the build failure. However, CONFIG_PWM_PIO_RP1 is also lacking that dependency, and your build of pwm-pio-rp1 isn't failing.

njhollinghurst:
I'll fix it. I'll make it an ifdef rather than a dependency as DPI should still be usable without it (apart from the interlaced sync fixup).

pelwell:
I would add a `select RP1_PIO` in the Kconfig.

6by9:
> I would add a `select RP1_PIO` in the Kconfig.

I'd agree.
It's possible to make things work with ifdef's, but it's non-trivial to ensure that all combinations of modules and built-ins are valid.

igorpecovnik:
> Are you using a different defconfig?

Yes. Still this should not break this way. I will update our configs. Thanks for the tip.

pelwell:
Be careful not to introduce a recursive dependency, as I seem to have just done.

igorpecovnik:
> Be careful not to introduce a recursive dependency,

Yeah, I just did ;)

```
[🔨]   drivers/misc/Kconfig:20:error: recursive dependency detected!
[🔨]   drivers/misc/Kconfig:20:	symbol RP1_PIO is selected by PWM_PIO_RP1
[🔨]   drivers/pwm/Kconfig:468:	symbol PWM_PIO_RP1 depends on FIRMWARE_RP1
[🔨]   drivers/firmware/Kconfig:156:	symbol FIRMWARE_RP1 is selected by RP1_PIO
```

pelwell:
That's a harsh test - it's objecting to "A selects B, B selects C, and A depends on C".

pelwell:
I've unbroken the builds while we come up with a better scheme.

PR #6531, opened 2024-12-10
Revert "PCI: Warn if no host bridge NUMA node info"

popcornmix:
This warning doesn't mean anyting on our platform and the warning causes confusion.

See: https://forums.raspberrypi.com/viewtopic.php?p=2276125#p2276125

This reverts commit ad5086108b9f0361929aa9a79cf959ab5681d249.


popcornmix:
The alternative would be making `pcibus_to_node(bus)` return something other than `NUMA_NO_NODE` but I haven't investigated that (and don't know if that would come with any side effects).

pelwell (approved):

PR #6533, opened 2024-12-11
drm: vc4: txp: Do not allow 24bpp formats when transposing

6by9:
The hardware doesn't support transposing to 24bpp (RGB888/BGR888) formats. There's no way to advertise this through DRM, so block it from atomic_check instead.

https://forums.raspberrypi.com/viewtopic.php?t=380129

### DIFF
--- a/drivers/gpu/drm/vc4/vc4_txp.c
+++ b/drivers/gpu/drm/vc4/vc4_txp.c
@@ -272,6 +272,13 @@ static int vc4_txp_connector_atomic_check(struct drm_connector *conn,
 		return -EINVAL;
 	}
 
+	if (conn_state->rotation & DRM_MODE_TRANSPOSE &&
+	    (fb->format->format == DRM_FORMAT_RGB888 ||
+	     fb->format->format == DRM_FORMAT_BGR888)) {
+		DRM_DEBUG_KMS("24bpp formats not supported when transposing\n");
+		return -EINVAL;
+	}
+
 	for (i = 0; i < ARRAY_SIZE(drm_fmts); i++) {
 		if (fb->format->format == drm_fmts[i])
 			break;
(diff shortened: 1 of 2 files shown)


drivers/gpu/drm/vc4/vc4_txp.c
```diff
@@ -272,6 +272,13 @@ static int vc4_txp_connector_atomic_check(struct drm_connector *conn,
 		return -EINVAL;
 	}
 
+	if (conn_state->rotation == (DRM_MODE_ROTATE_0 | DRM_MODE_TRANSPOSE) &&
```
dividuum: Not sure if I'm violating the API, but I've been using the transpose feature by just setting `DRM_MODE_TRANSPOSE` without explicitly or'ing `DRM_MODE_ROTATE_0` as not rotating seems to be the default when setting the rotation property to `0`. As such my old code still slipped by this condition. With the explicit  `DRM_MODE_ROTATE_0` also or'ed, the commit fails as expected.

drivers/gpu/drm/vc4/vc4_txp.c
```diff
@@ -272,6 +272,13 @@ static int vc4_txp_connector_atomic_check(struct drm_connector *conn,
...
+	    (fb->format->format == DRM_FORMAT_RGB888 ||
+	     fb->format->format == DRM_FORMAT_BGR888)) {
+		DRM_DEBUG_KMS("24bpp formats not supported when tranposing\n");
```
dividuum: Typo "tranposing"

drivers/gpu/drm/vc4/vc4_txp.c
6by9: Yes it's violating the API. Added an extra patch that should validate that.

drivers/gpu/drm/vc4/vc4_txp.c
6by9: Fixed.

dividuum:
Triggered both error cases and confirmed that using the API correctly works as expected. LGTM.

6by9:
Thank you for testing.

dividuum:
Neat. Thanks everyone!

PR #6535, opened 2024-12-12
drm: rp1: rp1-dpi: Fix optional dependency on RP1_PIO

njhollinghurst:
Add optional dependency to Kconfig, and conditionally compile PIO-dependent code. Add a mode validation function to reject interlaced modes when RP1_PIO is not present.

### DIFF
--- a/drivers/gpu/drm/rp1/rp1-dpi/Kconfig
+++ b/drivers/gpu/drm/rp1/rp1-dpi/Kconfig
@@ -7,5 +7,10 @@ config DRM_RP1_DPI
 	select DRM_VRAM_HELPER
 	select DRM_TTM
 	select DRM_TTM_HELPER
+	depends on RP1_PIO || !RP1_PIO
 	help
-	  Choose this option to enable Video Out on RP1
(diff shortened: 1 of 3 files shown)


njhollinghurst:
Purported fix for #6529, in testing with various defconfigs...

6by9:
I think you want [`IS_REACHABLE`](https://elixir.bootlin.com/linux/v6.12.4/source/include/linux/kconfig.h#L65).
Simple use of ifdef has a tendency to do bad things if one half gets built-in and the other half is a module.

njhollinghurst:
Ah yes, I think I do want that.

At the Kconfig level, this [documentation](https://github.com/raspberrypi/linux/blob/rpi-6.6.y/Documentation/kbuild/kconfig-language.rst#optional-dependencies) suggests there's already a trap for built-in depending on module (though it's far from obvious from the syntax how it achieves that...)

pelwell (approved):

6by9:
LGTM

njhollinghurst:
Tested and works as expected with `CONFIG_RP1_PIO=m`, `y` and undefined.

PR #6542, opened 2024-12-16
dtoverlays: Use continuous clock mode for ov9281

6by9:
This increases the maximum frame rate from 247 to 260fps in 10-bit mode.

### DIFF
--- a/arch/arm/boot/dts/overlays/ov9281-overlay.dts
+++ b/arch/arm/boot/dts/overlays/ov9281-overlay.dts
@@ -29,7 +29,6 @@
 				csi_ep: endpoint {
 					remote-endpoint = <&cam_endpoint>;
 					data-lanes = <1 2>;
-					clock-noncontinuous;
 				};
 			};
 		};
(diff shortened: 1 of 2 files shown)


pelwell:
Can you explain briefly why this change can't be applied to all cameras, and why it mostly (but not always) must be applied to both ends?

6by9:
Support of continuous vs non-continuous mode is down to the sensor driver and whether it checks DT for the setting and acts accordingly. Many drivers don't check and assume it is correct.
The CSI2 receiver driver generally needs to know in order to avoid looking for the LP->HS transition on the clock lane and therefore not syncing.

Why V4L2 doesn't pass the setting from source to sink is their policy decision, and you're not meant to go rummaging in other DT entities to extract it yourself. There is actually an API call that does allow passing the setting, but it's not widely implemented, and whenever anyone has implemented it just to relay this clock mode information, it has been rejected. (DRM does do the sensible thing and passes it from panel/bridge driver to the DSI host controller).

Not dropping to LP mode on the clock lane saves a little time per line, hence this sensor driver allows the blanking period to be reduced slightly to match. AFAIK that is unique to OV928[12] as I noticed it and fixed it up (https://github.com/torvalds/linux/commit/a387834c16ec0264cc0503f71f3d5a5d7de02478)

pelwell:
Thanks - that makes sense.

PR #6545, opened 2024-12-17
spi: dw: Add SPI_NO_CS to the controller mode_list

pelwell:
If using gpiod CS lines it must be possible to run in no-chip-select mode, so declare SPI_NO_CS in the mode_list field for the controller.


pelwell:
See also https://forums.raspberrypi.com/viewtopic.php?t=381111.

PR #6546, opened 2024-12-18
pcm3168a: driver workaround to support ezsound soundcard

gordoste:
ASoC has a limitation preventing my new soundcard from working. Basically, it can't configure the codec driver correctly to use bidirectional I2S with an external crystal as the clock and a single set of clock lines for Rx & Tx (like the RPi has).
This workaround forces the driver to configure itself correctly and enables my new soundcard to work perfectly for 6.12 kernels. As the soundcard is only for RPi5, it doesn't make sense to do this upstream.

See https://forums.raspberrypi.com/viewtopic.php?t=380915 for links to the linux-sound mail threads discussing the SoC limitation. 

Merry Christmas!

### DIFF
--- a/Documentation/devicetree/bindings/sound/ti,pcm3168a.yaml
+++ b/Documentation/devicetree/bindings/sound/ti,pcm3168a.yaml
@@ -58,6 +58,14 @@ properties:
   VCCDA2-supply:
     description: DAC power supply regulator 2 (+5V)
 
+  adc-force-cons:
+    description: Force ADC to operate in consumer mode. Useful if ADC and DAC
+      clock pins are tied together with DAC as producer.
+
+  dac-force-cons:
+    description: Force DAC to operate in consumer mode. Useful if ADC and DAC
+      clock pins are tied together with ADC as producer.
+
   ports:
     $ref: audio-graph-port.yaml#/definitions/port-base
     unevaluatedProperties: false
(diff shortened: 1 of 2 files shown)


pelwell:
We are constantly trying to reduce the number of downstream patches as they carry a support burden, and accepting this PR would be moving in the opposite direction. However, we are more likely to accept Pi-specific patches. I think your change falls in a grey area, being a change to common code for a Pi-specific product.

I took a look at some of the kernel mail threads, and this message (https://lore.kernel.org/linux-sound/8734is88xn.wl-kuninori.morimoto.gx@renesas.com/) suggests that there may be a more generic upstream change. Have I understood correctly?

gordoste:
Yes, that’s correct. The fix is under review, and if it is adopted then this workaround won’t be needed any more. But the fix won’t be in the 6.12 tree.

pelwell:
We are happy to back-port upstream patches because the support lifetime is limited.

gordoste:
OK, I thought this would have a limited lifetime as well, because you can drop it when you move to the next LTS kernel (assuming the upstream fix is available). Incidentally, this driver is not currently built in bcm2712_defconfig.
I think the backporting of the fix will be time-consuming as there have been plenty of other changes in ASoC in the 6.13 tree.

pelwell:
OK - I'll accept your patches, but first take a look at the errors that checkpatch flagged up (the Advisory checkpatch review). It's also good practice the follow the conventions for commit naming used for files and subsystems that you modify - `git log --oneline <filename>` should give you enough info. You'll also need to provide the defconfig changes to enable the codec and the overlay that makes use of it (and soundcard driver if there is one). I see little point in merging some before the others.

gordoste:
Thanks, I will address those issues and raise a new PR. Agree that it makes sense to do it all at once.

PR #6547, opened 2024-12-18
overlays: goodix: Allow override i2c address

ragazenta:
Some Goodix devices e.g. gt911 use address 0x5d instead of 0x14. So, make the address overridable.


6by9:
LGTM.

The GT911 goes and uses the state of the interrupt line as it comes out of reset to set the address (0x14 if high, 0x5b if low), and the driver does check the address and control the state of the interrupt line during reset accordingly. I might dig out the HyperPixel4 to ensure that it does indeed do the right thing there, as that overlay hacks it by having both addresses defined.

PR #6552, opened 2024-12-19
add ams-osram image sensors.

fiepfiep:
Add support for the following image sensors:
mira050
mira130
mira220 
mira016


pelwell:
Why have you structured your files like this, putting the code in `.inl` files and including it into `.c` files? The inclusions are:
```
drivers/media/i2c/mira016.c:10:#include "mira016.inl"
drivers/media/i2c/mira050.c:10:#include "mira050.inl"
drivers/media/i2c/mira050color.c:10:#include "mira050.inl"
drivers/media/i2c/mira130.c:10:#include "mira130.inl"
drivers/media/i2c/mira220.c:10:#include "mira220.inl"
drivers/media/i2c/mira220color.c:10:#include "mira220.inl"
```
2 of the .inls are only included in one place - just embed them directly into the .c code.

For the `color` variants, where 2 .c files share a .inl, the only difference is the driver name, the compatible string, and a way of determining which is being used. This could all be achieved with a single .c file with the .inl embedded, saving the space occupied by a separate module build.

arch/arm64/configs/bcm2711_defconfig
```diff
@@ -1,4 +1,4 @@
-CONFIG_LOCALVERSION="-v8"
+CONFIG_LOCALVERSION="-v8+"
```
6by9, line 1: You shouldn't be changing  CONFIG_LOCALVERSION here.

arch/arm64/configs/bcmrpi3_defconfig
```diff
@@ -1,4 +1,4 @@
-CONFIG_LOCALVERSION="-v8"
+CONFIG_LOCALVERSION="-v8+"
```
6by9, line 1: You shouldn't be changing  CONFIG_LOCALVERSION here.

drivers/media/i2c/mira016.inl
```diff
@@ -0,0 +1,6501 @@
...
+#include <media/v4l2-event.h>
+#include <media/v4l2-fwnode.h>
+#include <media/v4l2-mediabus.h>
+#include <asm/unaligned.h>
+
+/*
+ * Introduce new v4l2 control
+ */
+#include <linux/v4l2-controls.h>
+#define AMS_CAMERA_CID_BASE (V4L2_CTRL_CLASS_CAMERA | 0x2000)
```
6by9, line 31: It's more normal to claim a range above V4L2_CID_USER_BASE, and then use those.

drivers/media/i2c/mira016.inl
```diff
@@ -0,0 +1,6501 @@
...
+// For test pattern with 2D gradiant
+#define MIRA016_DELTA_TEST_IMG_REG 0x0056
+// For setting test pattern type
+#define MIRA016_TEST_PATTERN_REG 0x0062
+#define MIRA016_TEST_PATTERN_DISABLE 0x00
+#define MIRA016_TEST_PATTERN_FIXED_DATA 0x01
+#define MIRA016_TEST_PATTERN_2D_GRADIENT 0x02
+
+/* Embedded metadata stream structure */
+#define MIRA016_EMBEDDED_LINE_WIDTH 0
```
6by9, line 185: If you have no embedded data, then you've chosen a bad base driver to base yours on. There is no point in having the metadata pad at all, so it should be dropped.

Using a second pad for metadata is also not the way that mainline are doing it, and we're about to switch to that approach.

drivers/media/i2c/mira016.inl
```diff
@@ -0,0 +1,6501 @@
...
+	u32 code;
+
+	/* bit_depth needed for analog gain selection */
+	u8 bit_depth;
+};
+
+// converted_Draco_i2c_configuration_sequence_hex_10bit_1x_360fps_Version3
+static const struct mira016_reg full_400_400_100fps_10b_1lane_reg_pre_soft_reset[] = {
+	// Sensor Operating Mode
+	{57344, 0},
```
6by9, line 274: The use of decimal here looks odd, particularly seeing as you converted something that included hex in the name. 57344 = 0xe000.

Can any of this be documented with register names?

drivers/media/i2c/mira016.inl
```diff
@@ -0,0 +1,6501 @@
...
+
+	{0x1dd, 1},	 // # vsspc sel
+	{0x1ed, 14}, // # vsspc val
+
+	{0x1df, 1}, // #cp sel
+	{0x0ee, 4}, // #cp trim,
+	};
+
+// converted_Draco_i2c_configuration_sequence_hex_12bit_1x_200fps_Version3
+static const struct mira016_reg full_400_400_100fps_12b_1lane_reg_pre_soft_reset[] = {
```
6by9, line 937: I see only a handful of registers different between this and  `full_400_400_100fps_10b_1lane_reg_pre_soft_reset`. Can the bit depth be controlled through a separate table to the resolution setting?

drivers/media/i2c/mira016.inl
```diff
@@ -0,0 +1,6501 @@
...
+	{435, 21},
+
+
+
+	{0xE000, 0},
+	{0xE004, 0},
+	{0x33D, 1},
+};
+
+static const struct mira016_reg full_400_400_100fps_12b_1lane_reg_post_soft_reset[] = {
```
6by9, line 1587: Identical to `full_400_400_100fps_10b_1lane_reg_post_soft_reset`

drivers/media/i2c/mira016.inl
```diff
@@ -0,0 +1,6501 @@
...
+	{435, 28},
+
+
+
+	{0xE000, 0},
+	{0xE004, 0},
+	{0x33D, 1},
+};
+
+static const struct mira016_reg full_400_400_100fps_8b_1lane_reg_post_soft_reset[] = {
```
6by9, line 3600: Identical to `full_400_400_100fps_12b_1lane_reg_post_soft_reset`

drivers/media/i2c/mira016.inl
```diff
@@ -0,0 +1,6501 @@
...
+	MIRA016_TEST_PATTERN_DISABLE,
+	MIRA016_TEST_PATTERN_FIXED_DATA,
+	MIRA016_TEST_PATTERN_2D_GRADIENT,
+};
+
+/* regulator supplies */
+static const char *const mira016_supply_name[] = {
+	// TODO(jalv): Check supply names
+	/* Supplies can be enabled in any order */
+	"VANA", /* Analog (2.8V) supply */
```
6by9, line 3633: IIRC this is an error with imx219, and regulator names should use lower case.

drivers/media/i2c/mira016.inl
```diff
@@ -0,0 +1,6501 @@
...
+	else
+	{
+		dev_dbg(&client->dev, "%s: i2c write error, reg: %x\n",
+				__func__, reg);
+		if (ret >= 0)
+			ret = -EINVAL;
+	}
+
+	/*
+	 * The code below is for debug purpose.
```
6by9, line 3861: Please remove debug code when submitting drivers.

drivers/media/i2c/mira016.inl
```diff
@@ -0,0 +1,6501 @@
...
+		dev_dbg(&client->dev, "%s: OTP memory busy, skip raeding addr: 0x%X\n",
+				__func__, addr);
+		ret = -EINVAL;
+	}
+
+	return ret;
+}
+
+/* Write PMIC registers, and can be reused to write microcontroller reg. */
+static int mira016pmic_write(struct i2c_client *client, u8 reg, u8 val)
```
6by9, line 4108: Drivers are generally moving to use the V4L2_CCI library to handle the data packing and endianness such as this. Please see https://github.com/torvalds/linux/commit/af73323b97702e53b0a336972aaf23e7dd92c850 where it was done for imx290, and many others.

drivers/media/i2c/mira016.inl
```diff
@@ -0,0 +1,6501 @@
...
+					__func__);
+			goto reg_off;
+		}
+		usleep_range(MIRA016_XCLR_MIN_DELAY_US,
+					 MIRA016_XCLR_MIN_DELAY_US + MIRA016_XCLR_DELAY_RANGE_US);
+		mira016->powered = 1;
+	}
+	else
+	{
+		printk(KERN_INFO "[MIRA016]: Skip regulator and clk enable, because mira016->powered == %d.\n", mira016->powered);
```
6by9, line 4194: This would imply a serious error in the driver, and if true then needs to be reference counted.

drivers/media/i2c/mira016.inl
```diff
@@ -0,0 +1,6501 @@
...
+	// Enable or disable illumination trigger
+	printk(KERN_INFO "[MIRA016]: Writing EN_TRIG_ILLUM to %d.\n", mira016->illum_enable);
+	ret = mira016_write(mira016, MIRA016_EN_TRIG_ILLUM, mira016->illum_enable);
+	if (ret)
+	{
+		dev_err(&client->dev, "Error setting EN_TRIG_ILLUM to %d.", mira016->illum_enable);
+		return ret;
+	}
+
+	if (MIRA016_LPS_DISABLED)
```
6by9, line 4265: Dead code

drivers/media/i2c/mira016.inl
```diff
@@ -0,0 +1,6501 @@
...
+	// Set conetxt bank 0 or 1
+	ret = mira016_write(mira016, MIRA016_BANK_SEL_REG, 1);
+	if (ret)
+	{
+		dev_err(&client->dev, "Error setting BANK_SEL_REG.");
+		return ret;
+	}
+
+	// Enable or disable illumination trigger
+	printk(KERN_INFO "[MIRA016]: Writing EN_TRIG_ILLUM to %d.\n", mira016->illum_enable);
```
6by9, line 4257: Debug info.

All desired logging should be using dev_err, dev_info, and dev_dbg, not raw printk.

drivers/media/i2c/mira016.inl
```diff
@@ -0,0 +1,6501 @@
...
+	u16 reg_addr = (value >> 8) & 0xFFFF;
+	u8 reg_val = value & 0xFF;
+	u8 reg_flag = (value >> 24) & 0xFF;
+
+	// printk(KERN_INFO "[MIRA016]: %s reg_flag: 0x%02X; reg_addr: 0x%04X; reg_val: 0x%02X.\n",
+	// 		__func__, reg_flag, reg_addr, reg_val);
+
+	if (reg_flag & AMS_CAMERA_CID_MIRA016_REG_FLAG_CMD_SEL)
+	{
+		if (reg_flag == AMS_CAMERA_CID_MIRA016_REG_FLAG_SLEEP_US)
```
6by9, line 4356: All these flags need a fair amount of explanation, or a datasheet.

drivers/media/i2c/mira016.inl
```diff
@@ -0,0 +1,6501 @@
...
+		}
+		else
+		{
+			printk(KERN_INFO "[MIRA016]: %s unknown command from flag %u, ignored.\n", __func__, reg_flag);
+		}
+	}
+	else if (reg_flag & AMS_CAMERA_CID_MIRA016_REG_FLAG_FOR_READ)
+	{
+		// If it is for read, skip reagister write, cache addr and flag for read.
+		mira016->mira016_reg_w_cached_addr = reg_addr;
```
6by9, line 4457: So you write the register address that you want to read? Not nice.

drivers/media/i2c/mira016.inl
```diff
@@ -0,0 +1,6501 @@
...
+	{
+		dev_err(&client->dev, "Error setting CMD_REQ_1 to 0 for REQ_EXP.");
+		return ret;
+	}
+	usleep_range(10, 20);
+
+
+
+
+	// u8 busy_status = 1;
```
6by9, line 4779: Remove dead code.

drivers/media/i2c/mira016.inl
```diff
@@ -0,0 +1,6501 @@
...
+};
+
+#define MIRA016_NUM_SUPPLIES ARRAY_SIZE(mira016_supply_name)
+
+/*
+ * The supported formats. All flip/mirror combinations have the same byte order because the sensor
+ * is monochrome
+ */
+static const u32 codes[] = {
+	MEDIA_BUS_FMT_SGRBG10_1X10,
```
6by9, line 3645: Why are you using Bayer colour filter formats if the sensor is mono? MEDIA_BUS_FMT_Y10_1X10.

drivers/media/i2c/mira016.inl
```diff
@@ -0,0 +1,6501 @@
...
+	.s_ctrl = mira016_set_ctrl,
+};
+
+static const struct v4l2_ctrl_ops mira016_custom_ctrl_ops = {
+	.g_volatile_ctrl = mira016_g_ctrl,
+	.s_ctrl = mira016_s_ctrl,
+};
+
+/* list of custom v4l2 ctls */
+static struct v4l2_ctrl_config custom_ctrl_config_list[] = {
```
6by9, line 5239: If this is being used to allow userspace to poke random registers, then please look at implementing `s_register` and `g_register` in `v4l2_subdev_core_ops`. It also gets around your need to set and cache the register address through Write in order to be able to implement read.

drivers/media/i2c/mira016.inl
```diff
@@ -0,0 +1,6501 @@
...
+
+	return codes[i];
+}
+
+static void mira016_set_default_format(struct mira016 *mira016)
+{
+	struct v4l2_mbus_framefmt *fmt;
+
+	fmt = &mira016->fmt;
+	fmt->code = MEDIA_BUS_FMT_SGRBG10_1X10; // MEDIA_BUS_FMT_Y10_1X10;
```
6by9, line 5013: Set it based on the mode tables element [0], not hard coding here.

drivers/media/i2c/mira016.inl
```diff
@@ -0,0 +1,6501 @@
...
+	struct v4l2_ctrl *gain;
+	// custom v4l2 control
+	struct v4l2_ctrl *mira016_reg_w;
+	struct v4l2_ctrl *mira016_reg_r;
+	u16 mira016_reg_w_cached_addr;
+	u8 mira016_reg_w_cached_flag;
+
+	/* Current mode */
+	const struct mira016_mode *mode;
+	/* current bit depth, may defer from mode->bit_depth */
```
6by9, line 3751: Typo for differ.

And I see no route in the code where it can differ. Can you point it out to me?

drivers/media/i2c/mira016.inl
```diff
@@ -0,0 +1,6501 @@
...
+		case V4L2_CID_TEST_PATTERN:
+			ret = mira016_write(mira016, MIRA016_BANK_SEL_REG, 0);
+			// Fixed data is hard coded to 0xAB.
+			ret = mira016_write(mira016, MIRA016_TRAINING_WORD_REG, 0xAB);
+			// Gradient is hard coded to 45 degree.
+			ret = mira016_write(mira016, MIRA016_DELTA_TEST_IMG_REG, 0x01);
+			ret = mira016_write(mira016, MIRA016_TEST_PATTERN_REG,
+								mira016_test_pattern_val[ctrl->val]);
+			break;
+		case V4L2_CID_HFLIP:
```
6by9, line 5120: If not implemented, then don't add the control.

drivers/media/i2c/mira016.inl
```diff
@@ -0,0 +1,6501 @@
...
+
+// Power on function timing
+#define MIRA016_XCLR_MIN_DELAY_US 150000
+#define MIRA016_XCLR_DELAY_RANGE_US 3000
+
+// pixel_rate = link_freq * 2 * nr_of_lanes / bits_per_sample
+// 0.9Gb/s * 2 * 1 / 12 = 157286400
+// 1.5 Gbit/s * 2 * 1 / 12 = 250 000 000
+#define MIRA016_PIXEL_RATE (400000000)
+/* Should match device tree link freq */
```
6by9, line 154: It should also be correct for the sensor. Below you reference `MIRA016_DATA_RATE 1500 //Mbit/s`, which would imply the link frequency is 750MHz 

NB technically this exceeds the spec for the Pi0-4 CSI2 receiver, however you may get away with it.

drivers/media/i2c/mira016.inl
```diff
@@ -0,0 +1,6501 @@
...
+	ret = mira016_get_regulators(mira016);
+	if (ret)
+	{
+		dev_err(dev, "failed to get regulators\n");
+		return ret;
+	}
+
+	{
+		printk(KERN_INFO "[MIRA016]: Init PMIC and uC and led driver.\n");
+		mira016->pmic_client = i2c_new_dummy_device(client->adapter,
```
6by9, line 6364: If you have a separate PMIC device, please write a regulator driver for it, and configure device tree accordingly. Particularly if it is common to all these sensors.

drivers/media/i2c/mira016.inl
```diff
@@ -0,0 +1,6501 @@
...
+													MIRA016PMIC_I2C_ADDR);
+		if (IS_ERR(mira016->pmic_client))
+			return PTR_ERR(mira016->pmic_client);
+		mira016->uc_client = i2c_new_dummy_device(client->adapter,
+												  MIRA016UC_I2C_ADDR);
+		if (IS_ERR(mira016->uc_client))
+			return PTR_ERR(mira016->uc_client);
+		mira016->led_client = i2c_new_dummy_device(client->adapter,
```
6by9, line 6372: Other than a couple of init commands, what does the kernel do with this LED driver?

6by9 (commented):
I've only gone through mira016 as I'm assuming they are all similar.
I'll agree with Phil's comment that the minimal .c file and a huge .inl file is very strange.

I'm reviewing as if you're going to submit this to the mainline linux-media mailing list. I hope you will do in due course as otherwise libcamera won't merge the support for your sensor.

You've chosen an awkward moment to try and add new drivers, as with the switch to 6.12 being imminent, the structure of drivers changes fairly significantly, and the use of the subdev state is mandated. Using 6.12 and basing on something like imx290.c would be a better start point.

General comment of lots of dead or commented out code.

Allowing userspace to generally poke random sensor registers is generally not acceptable as it means requiring a random (probably closed-source) binary to get the sensor working.

Mainline will want a device tree binding. They're a bit of a faff, but not too bad. Various parameters (eg link frequency, and PMIC / LED components) will need to be nailed down before doing that.

6by9:
> For the `color` variants, where 2 .c files share a .inl, the only difference is the driver name, the compatible string, and a way of determining which is being used. This could all be achieved with a single .c file with the .inl embedded, saving the space occupied by a separate module build.

There are already image sensor drivers that are a single driver with multiple compatibles to select between colour and mono variants, and also slight variations in the sensor. The best example is [imx290](https://github.com/torvalds/linux/blob/master/drivers/media/i2c/imx290.c#L1642-L1656) which currently supports mono and colour imx290, colour imx327, and [support will be in 6.14](https://git.linuxtv.org/sailus/media_tree.git/commit/?h=devel&id=5d9f48c93f99a4bc41c8a9d90933bc8ff137668b) for mono and colour imx462 as well.

Thinking about the overlays, are we better off having one overlay per sensor, and then have an override for mono (or colour) which just changes the compatible? Again we have that already [with imx290](https://github.com/raspberrypi/linux/blob/rpi-6.6.y/arch/arm/boot/dts/overlays/imx290-overlay.dts) (imx327 also shares the same overlay include to reduce the duplication).

fiepfiep:
still needs some work from my side, will close for now.

PR #6553, opened 2024-12-19
overlays: Add and document i2c_csi_dsi0 parameters

pelwell:
Add "i2c_csi_dsi0" parameters to overlays that already have an "i2c_csi_dsi" parameter.

The I2C bus and GPIO mapping of i2c_csi_dsi and i2c_csi_dsi0 varies between platforms. Document the associations against the dtparams "i2c_csi_dsi" and "i2c_csi_dsi0" - run "dtparam -h i2c_csi_dsi" and "dtparam -h i2c_csi_dsi0" to read it.


6by9:
I haven't checked you have all the overlays that currently include an i2c_csi_dsi override, but those you have changed all look good.

And I believe you're correct with all the GPIO assignments too.

6by9 (approved):

PR #6560, opened 2024-12-21
Update config.yml - Raspbian -> Raspberry Pi OS

andrum993:


pelwell:
Can you add https:// to the URL, or whatever it takes for it to look like a link?

andrum993:
I've added https:// as suggested, but I've been unable to test whether this has any effect, as the .github/ISSUE_TEMPLATE/config.yml file does nothing when I create a test repo with it in, for some reason. A quick google suggests linkify-ing URLs in the about field may not be supported.

The resultant page (https://github.com/raspberrypi/linux/issues/new/choose) does have 'open' buttons on the righthand side which are clickable links to the specified URL.

pelwell:
Thanks!

PR #6564, opened 2024-12-22
Add support for ezsound 6x8 soundcard

gordoste:
Apologies for multiple PRs. I just realised I should be force-pushing to update the PR. Hopefully this one will pass all of the checks.

Summarising previous discussion for the record:
This is a new multichannel soundcard for the Pi5.
Some pcm3168a driver customisation is necessary to override a limitiation in ASoC. Hopefully, this limitation will be addressed in the next LTS kernel, at which point the customisations will no longer be needed.
The relevant modules need to be enabled in bcm2712_defconfig and DT overlays provided.

Technical details on ASoC limitation:
[scenario](https://lore.kernel.org/linux-sound/5011ceef-5100-441d-b169-dabba135d27f@iinet.net.au/)
[limitation](https://lore.kernel.org/linux-sound/87o71kjm66.wl-kuninori.morimoto.gx@renesas.com/)


pelwell:
It all looks ok to me apart from the HAT map entry. It's intended to be a get-out-of-jail mechanism for old devices where the overlay was built into the EEPROM rather than the name of the overlay - if it's a new HAT+ then you should not need to trouble the HAT map.

gordoste:
OK, I've removed the HAT map entry and tidied up the commit log a little bit. Incidentally, it could be worth adding a line in the overlay check Github action:
`echo "To run overlaycheck yourself, build and install github.com/raspberrypi/utils and run 'make scripts' in your kernel dir. You may need to install the gcc-arm-linux-gnueabihf package."`

pelwell:
Noted about the README, but the overlay check tool was updated last week to be more flexible about which version is cpp to use, such that there should no longer be a dependency on gcc-arm-linux-gnueabihf.

gordoste:
That would be nice, but I think you updated kdtc not overlaycheck. [commit](https://github.com/raspberrypi/utils/commit/56a219639a2016c37b42f096e908b2098df610e1)

PR #6566, opened 2024-12-28
Add driver for the Waveshare DSI-TOUCH series panels.

waveshare:
Hello, everyone:
This is the driver for the DSI-TOUCH series of panels to be released by Waveshare.
Regarding the missing help warning, it should be caused by the length of the help description being less than 4, so i think it allowed to ignore things.

If you have any other suggestions we'd be happy to make changes!


drivers/regulator/waveshare-panel-regulator.c
```diff
@@ -0,0 +1,306 @@
+// SPDX-License-Identifier: GPL-2.0
+/*
+ * Copyright (C) 2022 Raspberry Pi Ltd.
```
pelwell: This is incorrect.

drivers/regulator/waveshare-panel-regulator.c
```diff
@@ -0,0 +1,306 @@
...
+ *
+ * Based on rpi-panel-v2-regulator.c by Broadcom
```
pelwell: This is also incorrect.

drivers/gpu/drm/panel/panel-waveshare-dsi-v2.c
```diff
@@ -0,0 +1,1353 @@
+// SPDX-License-Identifier: GPL-2.0
+/*
+ * Copyright © 2023 Raspberry Pi Ltd
```
pelwell: More copy pasta.

pelwell (changes requested):

drivers/gpu/drm/panel/panel-waveshare-dsi-v2.c
waveshare: Okay, this is a copy. I'll fix this.
Any other suggestions? I'll standardize the changes.

waveshare: Previously mentioned issues have been fixed. PTAL.

pelwell:
Check the auto-build errors. And as a wider question, does this warrant a new driver? Are the devices so different from the v1s?

waveshare:
Yes, I think a new driver needs to be developed. v2 versions have significant architectural changes compared to the v1 series.
* v1 series uses bridge chips to convert DSI to other display protocols
* v2 series utilizes a 2-lane/4-lane mipi DSI direct-drive architecture that completely eliminates the need for a bridge chip.
The two have significant differences in hardware architecture, communication protocols, initialization sequences, and scalability
We believe that the v2 version will be more cost-effective, maintainable and scalable, and will bring a better experience and continued development to the Raspberry Pi DSI display series.

btw，The previous check auto-build errors has been resolved

drivers/gpu/drm/panel/panel-waveshare-dsi-v2.c
```diff
@@ -0,0 +1,1353 @@
...
+				     "Couldn't get our reset GPIO\n");
+
+	ret = of_drm_get_panel_orientation(dsi->dev.of_node, &ctx->orientation);
+	if (ret) {
+		dev_err(&dsi->dev, "%pOF: failed to get orientation: %d\n",
+			dsi->dev.of_node, ret);
+		return ret;
+	}
+
+	ctx->panel.prepare_prev_first = true;
```
6by9: Duplicate of line 1205.

drivers/gpu/drm/panel/panel-waveshare-dsi-v2.c
```diff
@@ -0,0 +1,1353 @@
...
+	.htotal = 800 + 40 + 20 + 20,
+	.vdisplay = 1280,
+	.vsync_start = 1280 + 30,
+	.vsync_end = 1280 + 30 + 12,
+	.vtotal = 1280 + 30 + 12 + 4,
+	.width_mm = 107,
+	.height_mm = 172,
+};
+
+static const struct drm_display_mode ws_panel_8_a_2lane_mode = {
```
6by9: This mode is identical to `ws_panel_8_a_mode`, so you could remove the duplication.

arch/arm/configs/bcm2711_defconfig
```diff
@@ -794,6 +794,7 @@ CONFIG_REGULATOR_ARIZONA_MICSUPP=m
 CONFIG_REGULATOR_GPIO=y
 CONFIG_REGULATOR_RASPBERRYPI_TOUCHSCREEN_ATTINY=m
 CONFIG_REGULATOR_RASPBERRYPI_TOUCHSCREEN_V2=m
+CONFIG_REGULATOR_WAVESHARE_TOUCHSCREEN=m
```
6by9, line 801: You've only added these to the Pi4 and Pi5 defconfigs. Is there a reason not to support them on earlier Pis? All the 2 lane configs should work.

If these genuinely can not work on earlier Pis, then overlay_map.dts can be used to block the use of the overlays on other platforms.

drivers/gpu/drm/panel/panel-waveshare-dsi-v2.c
```diff
@@ -0,0 +1,1353 @@
...
+
+static const struct drm_display_mode ws_panel_10_1_a_2lane_mode = {
+	.clock = 70000,
+	.hdisplay = 800,
+	.hsync_start = 800 + 40,
+	.hsync_end = 800 + 40 + 20,
+	.htotal = 800 + 40 + 20 + 20,
+	.vdisplay = 1280,
+	.vsync_start = 1280 + 30,
+	.vsync_end = 1280 + 30 + 10,
```
6by9: Is there a need to vary the vertical front porch and sync width between the 2 and 4 lane modes? Both appear to work as well as each other in 2 lane mode (haven't tested 4) when using `kmstest --flip -r 70000000,800/40/20/20/+,1280/30/10/4/+` and `kmstest --flip -r 70000000,800/40/20/20/+,1280/20/20/4/+`.

drivers/gpu/drm/panel/panel-waveshare-dsi-v2.c
```diff
@@ -0,0 +1,1353 @@
...
+	.lanes = 2,
+	.format = MIPI_DSI_FMT_RGB888,
+};
+
+static const struct of_device_id ws_panel_of_match[] = {
+	{ .compatible = "waveshare,10.1-dsi-touch-a,4lane",
+	  &ws_panel_10_1_inch_a_4lane_desc },
+	{ .compatible = "waveshare,10.1-dsi-touch-a",
+	  &ws_panel_10_1_inch_a_desc },
+	{ .compatible = "waveshare,8.0-dsi-touch-a,4lane",
```
6by9: I don't believe commas are permitted within compatibles other than separating the vendor and product name. @pelwell Could you confirm?

Ditto for `waveshare,10.1-dsi-touch-a,4lane`

6by9 (commented):
Sorry, I had a couple of these comments pending when I thought I'd submitted them.

Other than possibly EMC, seeing as both the 2 and 4 lane modes achieve the same refresh rate, is there a benefit in providing options for both?

drivers/gpu/drm/panel/panel-waveshare-dsi-v2.c
pelwell: Yes. The current version of the spec says:

> The recommended format is "manufacturer,model", where manufacturer is a string describing the name
of the manufacturer (such as a stock ticker symbol), and model specifies the model number.
> The compatible string should consist only of lowercase letters, digits and dashes, and should start with a letter.
A single comma is typically only used following a vendor prefix. Underscores should not be used.

drivers/gpu/drm/panel/panel-waveshare-dsi-v2.c
waveshare: Do I need to push it again to solve this problem? Avoid repeated reviews, or do I directly push the modified parts?

drivers/gpu/drm/panel/panel-waveshare-dsi-v2.c
waveshare: And the same as above, we will modify it

arch/arm/configs/bcm2711_defconfig
waveshare: Alright, when we first PR, we only tested 4B and 5B.
Now we have found 3B+, 2B, and 1B+ for testing, this driver can run normally on them, and we will add it to the default configuration of earlier Pi

drivers/gpu/drm/panel/panel-waveshare-dsi-v2.c
waveshare: Both seem to work normally, we will unify them
```suggestion
	.vsync_end = 1280 + 20+ 10,
```

drivers/gpu/drm/panel/panel-waveshare-dsi-v2.c
waveshare: Thank you for the reminder, we will change it
```suggestion
	{ .compatible = "waveshare,8.0-dsi-touch-a-4lane",
```

drivers/gpu/drm/panel/panel-waveshare-dsi-v2.c
6by9: Update the patches on your local repo (`git commit --amend` and `git rebase -i` ) to make a consistent set of what you want to merge, and then `git push -f ...` to push to the same waveshare DSI-LCD branch on Github. That automatically updates the PR, and we'll review the changes.

waveshare: The issues raised have been fixed , PTAL.

drivers/gpu/drm/panel/panel-waveshare-dsi-v2.c
```diff
@@ -0,0 +1,1324 @@
...
+	return 0;
+}
+
+static int ws_panel_prepare(struct drm_panel *panel)
+{
+	struct ws_panel *ctx = panel_to_ws(panel);
+	int ret;
+
+	/* And reset it */
+	if (ctx->reset != NULL) {
```
6by9: Can be simplified to `if (ctx->reset) {`
This occurs in number of places.

drivers/gpu/drm/panel/panel-waveshare-dsi-v2.c
```diff
@@ -0,0 +1,1324 @@
...
+	}
+
+	ret = ws_panel_init_dcs_cmd(ctx);
+	if (ret < 0)
+		dev_err(panel->dev, "failed to init panel: %d\n", ret);
+
+	return 0;
+}
+
+static int ws_panel_enable(struct drm_panel *panel)
```
6by9: All the panel functions are optional, so you can just omit these.

drivers/gpu/drm/panel/panel-waveshare-dsi-v2.c
```diff
@@ -0,0 +1,1324 @@
...
+{
+	return 0;
+}
+
+static int ws_panel_unprepare(struct drm_panel *panel)
+{
+	struct ws_panel *ctx = panel_to_ws(panel);
+
+	mipi_dsi_dcs_set_display_off(ctx->dsi);
+	mipi_dsi_dcs_enter_sleep_mode(ctx->dsi);
```
6by9, line 1030: Slight mismatch with these two as `mipi_dsi_dcs_set_display_on` and `mipi_dsi_dcs_exit_sleep_mode` aren't called in `prepare`

drivers/regulator/waveshare-panel-regulator.c
```diff
@@ -0,0 +1,306 @@
...
+
+static int waveshare_panel_gpio_direction_out(struct gpio_chip *gc,
+					      unsigned int offset, int val)
+{
+	struct waveshare_panel_lcd *state = gpiochip_get_data(gc);
+	u16 last_val;
+
+	state->direction_state &= ~(1 << offset);
+
+	last_val = state->poweron_state;
```
6by9, line 67: `state->poweron_state` is being accessed here without the mutex being claimed.

Is the mutex needed to protect `state->direction_state`? You are doing a read/modify/write with the `&=`, and `|=` in `waveshare_panel_gpio_direction_in`

drivers/regulator/waveshare-panel-regulator.c
```diff
@@ -0,0 +1,306 @@
...
+
+	if (bl->props.power != FB_BLANK_UNBLANK ||
+	    bl->props.state & (BL_CORE_SUSPENDED | BL_CORE_FBBLANK))
+		brightness = 0;
+
+	mutex_lock(&state->lock);
+
+	last_val = state->poweron_state;
+	if (brightness)
+		last_val |= (1 << 2); // Enable BL_EN
```
6by9: Bit 2 is going to be advertised as a GPIO that another driver could request, however you are modifying it internally from the driver. You have the potential for issues there if others make use of this driver from their own overlays.

arch/arm/boot/dts/overlays/vc4-kms-dsi-waveshare-panel-v2-overlay.dts
```diff
@@ -0,0 +1,93 @@
...
+			status = "okay";
+
+			display_mcu: display_mcu@45 {
+				compatible = "waveshare,touchscreen-panel-regulator";
+				reg = <0x45>;
+				gpio-controller;
+				#gpio-cells = <2>;
+			};
+
+			touch: goodix@5d {
```
6by9, line 28: Seeing as we have no control over the interrupt line, I assume it is pulled down so that the touch controller will reliably come out of reset on address 0x5d. (Using the INT line to determine I2C address is a right pain with these Goodix controllers).

drivers/regulator/waveshare-panel-regulator.c
```diff
@@ -0,0 +1,306 @@
...
+	if (ret == 0)
+		dev_info(&i2c->dev, "waveshare panel size = %d\n", data);
+
+	ret = waveshare_panel_i2c_read(i2c, REG_VERSION, &data);
+	if (ret == 0)
+		dev_info(&i2c->dev, "waveshare panel mcu version = 0x%x\n",
+			 data);
+
+	state->direction_state = 0;
+	state->poweron_state = (1 << 9) | (1 << 8) | (1 << 4) |
```
6by9: There's a general preference to use the `BIT(n)` macro to denote bit-usage such as these. It could probably replace all your `1 << n` uses in the driver.

drivers/regulator/waveshare-panel-regulator.c
```diff
@@ -0,0 +1,306 @@
...
+{
+	struct waveshare_panel_lcd *state = gpiochip_get_data(gc);
+
+	state->direction_state |= (1 << offset);
+
+	return 0;
+}
+
+static int waveshare_panel_gpio_direction_out(struct gpio_chip *gc,
+					      unsigned int offset, int val)
```
6by9, line 57: Do you need to be able to control the direction with the current panels? If not, then keep the driver simple for now, and add direction control later when it is needed.

6by9 (approved):
A few comments, however as this driver isn't going to be upstreamed in the current form, I don't view any of them as critical to be resolved.

If you want to implement any of the changes then please respond in the next 48 hours to say so, otherwise I'm happy for this to be merged.

drivers/gpu/drm/panel/panel-waveshare-dsi-v2.c
luckfox-eng33: That's a good suggestion, we'll deal with it.

waveshare: That's a good suggestion, we'll deal with it.

drivers/gpu/drm/panel/panel-waveshare-dsi-v2.c
waveshare: OK, we'll delete this part.

drivers/gpu/drm/panel/panel-waveshare-dsi-v2.c
waveshare: Initialized to add display and postpone sleep instructions. Because some screens need to add other instructions to the mipi_dsi_dcs_set_display_on (0x29) and mipi_dsi_dcs_exit_sleep_mode (0x11) , there is no direct function call.

drivers/regulator/waveshare-panel-regulator.c
waveshare: OK, we'll add a mutex

drivers/regulator/waveshare-panel-regulator.c
waveshare: OK, we will use gpiod_set_value_cansleep to modify the BL_EN hardware instead of writing the registers directly

arch/arm/boot/dts/overlays/vc4-kms-dsi-waveshare-panel-v2-overlay.dts
waveshare: The muc is connected to the INT pin. If necessary, we can set the INT pin level when exiting the reset through the MCU. By default, the I2C address of Goodix controllers is 0x5d

drivers/regulator/waveshare-panel-regulator.c
waveshare: Ok, we'll use `BIT(n)` replace, maybe we're a bit old-school，lol

arch/arm/boot/dts/overlays/vc4-kms-dsi-waveshare-panel-v2-overlay.dts
6by9: As long as you have it in hand.

Avoiding a repeat of https://github.com/raspberrypi/linux/issues/6538 would be good, and I've been in exactly the same position with an early version of the Raspberry Pi Touch Display 2.
The vc4-dpi-hyperpixel4 overlay hacks around the issue by defining the touch controller on both addresses, and accepts that one will fail to probe.

drivers/regulator/waveshare-panel-regulator.c
6by9: I would often go for the `1<<n` style too, but the Linux coding style can be a bit fussy and I'm now used to reviewing based on that.

drivers/regulator/waveshare-panel-regulator.c
waveshare: It can control the direction of some GPIOs, such as the INT pin mentioned above, which we hope to keep. If necessary, we can detect the level of some pins.

drivers/regulator/waveshare-panel-regulator.c
6by9: So you now read state->poweron_state without the mutex held, and only take the mutex to write the value?! You need to take the mutex to make the entire read/modify/write operation atomic.

6by9 (commented):

Other than this one issue, I'm happy.

pelwell:
Since this PR was opened we've gained another defconfig - arch/arm64/configs/bcm2711_rt_defconfig. Please include it in the set of defconfigs updated by these patches.

waveshare:
> Since this PR was opened we've gained another defconfig - arch/arm64/configs/bcm2711_rt_defconfig. Please include it in the set of defconfigs updated by these patches.

We have just returned from the Qingming Festival holiday and will supplement the defconfig and resubmit shortly. Additionally, I inadvertently clicked the reviewers button earlier - please disregard that action.

waveshare:
@pelwell As requested:
We confirm `arch/arm64/configs/bcm2711_rt_defconfig` only exists in the `rpi-6.12.y` branch. We've therefore:
* Redirected this PR's patches to target `rpi-6.12.y`
* Force-pushed the updated series with commit [916bd958fc76988071ffd7c3fae9d317074bfb89]

waveshare:
To be reviewed.  
Please take a look. @pelwell
If there are any questions, we'd be happy to continue adjusting.

pelwell:
No, I think we're good here.

PR #6572, opened 2024-12-31
New PWM cooling fan overlay for Pi 3/4

herbingk:
This is my 3rd attempt, to make a pull request on the topic. I got stuck with your workflow scripts, that always checked older commits and failed with errors, so I started over again. Please bear with me, as I am entirely new to github and still on a learning curve. I you have an advice how to fix workflow errors with newer commits, without the need to start all over again, I greatly appreciate to learn.

I recently discovered that there is a new kernel based software PWM solution available since November 2024. A software-based PWM kernel module is available since then, back-ported from the Linux kernel 6.11 to the Raspberry Pi OS Bookworm kernel 6.6.62. This made me write a new pwm-gpio-fan overlay for my own use and after extensive testing on a Pi 3, I have published it for the community.

Moderator 6by9 on the Raspberry Pi forum suggested that I make a pull request, so that it gets merged for all Pi users.

Using the new kernel-based software PWM imposes several advantages over other implementations I have found on the internet and this forum:

- Compared with the existing gpio-fan overlay, the fan runs much quieter, often not even noticeable, due to the use of real PWM duty cycles.
- No hardware PWM (PWM0 or PWM1) of the Pi is occupied, so there is no conflict with the Pi's analogue audio output anymore.
- Reliable on even high CPU loads, as it's part of the kernel.
- Doesn't consume noticeable CPU power even on a Pi 3.

### DIFF
--- a/arch/arm/boot/dts/overlays/Makefile
+++ b/arch/arm/boot/dts/overlays/Makefile
@@ -219,6 +219,7 @@ dtbo-$(CONFIG_ARCH_BCM2835) += \
 	pwm.dtbo \
 	pwm-2chan.dtbo \
 	pwm-gpio.dtbo \
+	pwm-gpio-fan.dtbo \
 	pwm-ir-tx.dtbo \
 	pwm-pio.dtbo \
 	pwm1.dtbo \
--- a/arch/arm/boot/dts/overlays/README
+++ b/arch/arm/boot/dts/overlays/README
@@ -4036,6 +4036,46 @@ Load:   dtoverlay=pwm-gpio,<param>=<val>
 Params: gpio                    Output pin (default 4)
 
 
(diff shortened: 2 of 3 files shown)


peterharperuk:
I would suggest waiting for feedback rather than trying with new merge requests. The RPI engineers will advise if changes are required

pelwell:
I've not tested this, but it looks reasonable and the auto-build checks are happy.

When you are making changes to a PR, I suggest the following: 
1. Add new commits that fix the problems. 
2. Verify that the problem is solved, ignoring checkpatch warnings about the original commits.
3. Use `git rebase -i HEAD~n` to squash the fixups into the original commits. 
4. Use `git push -f ...` to force push the new commits to your branch, which will update the PR.

PR #6578, opened 2025-01-02
Backport upstream CFE driver to 6.12

6by9:
Our downstream CFE driver blows up on 6.12 in link validation :-(

This PR removes the downstream driver and backports the mainline one that has been merged in 6.13.
I haven't updated all necessary sensor drivers as yet (I was testing with imx415), hence draft status for the time being.

It needs the updated libcamera branch in https://github.com/raspberrypi/libcamera/pull/200. That doesn't rebase cleanly on the next branch at present due to the CameraSensorFactory stuff having been merged through official libcamera, but not quite in the same form as in the PR.


popcornmix:
Do we want this in before 6.12 reaches stable?

6by9:
> Do we want this in before 6.12 reaches stable?

No. We ended up managing to find the cause of the downstream driver blowing up, and that is working fine on 6.12.

I'll close this for now, and we'll work on the upstream driver in our own time.

PR #6584, opened 2025-01-07
IMX500: Add support for rotation and image flip

katsuykk:
This PR adds support for rotation and image flip functionality for the IMX500.

### Rotation Support
- 0: 0 degrees
- 1: 90 degrees
- 2: 180 degrees
- 3: 270 degrees

### Image Flip Support
- 0: No flip
- 1: Flip horizontal
- 2: Flip vertical
- 3: Flip horizontal and vertical

### DIFF
--- a/drivers/media/i2c/imx500.c
+++ b/drivers/media/i2c/imx500.c
@@ -217,6 +217,9 @@
 /* Rotate angle */
 #define IMX500_REG_ADDR_ROTATION CCI_REG8(0xD680)
 
+/* Flip vertical/horizontal */
+#define IMX500_REG_ADDR_IMG_ORIENTATION CCI_REG8(0x0101)
+
 /* Embedded sizes */
 #define IMX500_MAX_EMBEDDED_SIZE \
 	(2 * ((((IMX500_PIXEL_ARRAY_WIDTH * 10) >> 3) + 15) & ~15))
@@ -239,6 +242,7 @@ enum pad_types { IMAGE_PAD, METADATA_PAD, NUM_PADS };
 #define V4L2_CID_USER_IMX500_INFERENCE_WINDOW (V4L2_CID_USER_IMX500_BASE + 0)
 #define V4L2_CID_USER_IMX500_NETWORK_FW_FD (V4L2_CID_USER_IMX500_BASE + 1)
 #define V4L2_CID_USER_IMX500_ROTATION (V4L2_CID_USER_IMX500_BASE + 2)
+#define V4L2_CID_USER_IMX500_IMG_FLIP (V4L2_CID_USER_IMX500_BASE + 3)
 
 #define ONE_MIB (1024 * 1024)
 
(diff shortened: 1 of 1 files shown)


drivers/media/i2c/imx500.c
```diff
@@ -235,6 +238,7 @@ enum pad_types { IMAGE_PAD, METADATA_PAD, NUM_PADS };
 
 #define V4L2_CID_USER_IMX500_INFERENCE_WINDOW (V4L2_CID_USER_IMX500_BASE + 0)
 #define V4L2_CID_USER_IMX500_NETWORK_FW_FD (V4L2_CID_USER_IMX500_BASE + 1)
+#define V4L2_CID_USER_IMX500_ROTATION (V4L2_CID_USER_IMX500_BASE + 2)
```
6by9, line 244: Why does this need to be a custom control rather than V4L2_CID_ROTATE? (Documented in https://www.kernel.org/doc/html/latest/userspace-api/media/v4l/control.html)

How does this affect the resolution of the output image? Normally you'd rotate the 4056x3040 image to give a 3040x4056 output, so the output format needs to change.

drivers/media/i2c/imx500.c
```diff
@@ -2828,6 +2835,19 @@ static const struct v4l2_ctrl_config network_fw_fd = {
 	.def		= -1,
 };
 
+/* Custom control for rotation angle */
+static const struct v4l2_ctrl_config rotation_angle = {
+	.name		= "IMX500 Rotation Angle",
+	.id		= V4L2_CID_USER_IMX500_ROTATION,
+	.ops		= &imx500_ctrl_ops,
+	.type		= V4L2_CTRL_TYPE_INTEGER,
+	.flags		= V4L2_CTRL_FLAG_EXECUTE_ON_WRITE,
```
6by9, line 2851: Do you really need to rewrite the register even if setting it to the same value?

drivers/media/i2c/imx500.c
```diff
@@ -217,6 +217,9 @@
 /* Rotate angle */
 #define IMX500_REG_ADDR_ROTATION CCI_REG8(0xD680)
 
+/* Flip vertical/horizontal */
+#define IMX500_REG_ADDR_IMG_ORIENTATION CCI_REG8(0x0101)
```
6by9, line 221: Already defined and used as `IMX500_REG_ORIENTATION`

drivers/media/i2c/imx500.c
```diff
@@ -239,6 +242,7 @@ enum pad_types { IMAGE_PAD, METADATA_PAD, NUM_PADS };
 #define V4L2_CID_USER_IMX500_INFERENCE_WINDOW (V4L2_CID_USER_IMX500_BASE + 0)
 #define V4L2_CID_USER_IMX500_NETWORK_FW_FD (V4L2_CID_USER_IMX500_BASE + 1)
 #define V4L2_CID_USER_IMX500_ROTATION (V4L2_CID_USER_IMX500_BASE + 2)
+#define V4L2_CID_USER_IMX500_IMG_FLIP (V4L2_CID_USER_IMX500_BASE + 3)
```
6by9, line 245: Again why a custom control? V4L2_CID_HFLIP and V4L2_CIF_VFLIP exist and are already implemented.

6by9 (commented):
The flips patch looks like a straight NAK as flips are already implemented via V4L2_CID_HFLIP and V4L2_CID_VFLIP. There is no need for a custom control that would conflict with that.

Questions over implementing rotation.

PR #6586, opened 2025-01-07
media: rp1-cfe: Fix up link validation for CFE CFG input

6by9:
After commit 5fd3e2412ade ("media: v4l2-subdev: Support hybrid links in v4l2_subdev_link_validate()") link_validate is called on V4L2 OUTPUT devices such as the CFE cfg buffers input.
The CFE link_validate function was assuming it was always the sink of a link, which goes wrong on that port and does an invalid dereference.


6by9:
Only tested so far with imx415 (ie without embedded data).

6by9:
Tested with IMX477 and IMX708 now, and all good.

pelwell:
Any objection to merging, @naushir?

naushir:
Nope.  Merging now...

PR #6588, opened 2025-01-07
PCI: brcmstb: don't use ASPM state defines for register bits

P33M:
In commit b478e162f227 ("PCI/ASPM: Consolidate link state defines") PCIE_LINK_STATE_L1 and PCIE_LINK_STATE_L0s grew some bits for more granular control of ASPM.

This broke the aspm-no-l0s override, instead disabling link ASPM completely if this DT property was specified.

Just open-code the link capability register field bits.

Fixes: caab002d5069 ("PCI: brcmstb: Disable L0s component of ASPM if requested")
Fixes: 0693b4207fd7 ("PCI: brcmstb: Split post-link up initialization to brcm_pcie_start_link()")


pelwell:
Can we at least have comments as to what bits 0 and 1 do?

Issue #6527, opened 2024-12-10
Kernel 6.6.y and DAC Allo Piano 2.1+Kali fails init


foonerd:
### Describe the bug

It seems that sound/soc/bcm/allo-piano-dac-plus.c is not parsing parameters for alsa use. What I can observe is that the dtoverlay partly fails register the card:

```
[   19.836848] snd-allo-piano-dac-plus soc:sound: snd_soc_register_card() failed: -517
[   19.882239] snd-allo-piano-dac-plus soc:sound: snd_soc_register_card() failed: -517

```
However module somewhat loads:
ls -la /sys/module/snd_soc_allo_piano_dac_plus/

```
drwxr-xr-x   6 root root    0 Dec 10 08:08 .
drwxr-xr-x 150 root root    0 Jan  1  1970 ..
-r--r--r--   1 root root 4096 Dec 10 08:08 coresize
drwxr-xr-x   2 root root    0 Dec 10 08:08 drivers
drwxr-xr-x   2 root root    0 Dec 10 08:08 holders
-r--r--r--   1 root root 4096 Dec 10 08:08 initsize
-r--r--r--   1 root root 4096 Dec 10 08:08 initstate
drwxr-xr-x   2 root root    0 Dec 10 08:08 notes
-r--r--r--   1 root root 4096 Dec 10 08:08 refcnt
drwxr-xr-x   2 root root    0 Dec 10 08:08 sections
-r--r--r--   1 root root 4096 Dec 10 08:08 srcversion
-r--r--r--   1 root root 4096 Dec 10 08:08 taint
--w-------   1 root root 4096 Dec 10 08:08 uevent
```

Simple test with alsa-info shows that the expected DAC's control parameters are missing:

```
ALSA Information Script v 0.5.1
--------------------------------

This script visits the following commands/files to collect diagnostic
information about your ALSA installation and sound related hardware.

  dmesg
  lspci
  aplay
  amixer
  alsactl
  rpm, dpkg
  /proc/asound/
  /sys/class/sound/
  ~/.asoundrc (etc.)

See '/usr/sbin/alsa-info --help' for command line options.

cat: /tmp/alsa-info.2lPuin51C4/acpidevicestatus.tmp: No such file or directory
pgrep: pattern that searches for process name longer than 15 characters will result in zero matches
Try `pgrep -f' option to match against the complete command line.
cat: '/sys/module/snd_soc_allo_piano_dac_plus/parameters/*': No such file or directory
```

[...]

PR #6590, opened 2025-01-08
Trial patches for the second part of issue 6527

pelwell:
These patches should clear up the remaining problems outlined in issue #6527. If I'm right, the third patch is purely cosmetic. If not, it can be dropped.


pelwell:
For your consideration, @foonerd.

PR #6591, opened 2025-01-08
CQE improvements for 6.6

P33M:
The decision has been taken to soft-disable CQ on SD cards. Don't enable CQ unless there's a matching CID with a proven implementation.

To aid testing, and additionally let CM5 users control CQ, override the upstream boolean `supports-cqe` to a one-cell property which tristates probe behaviour, and add/modify the sd_cqe dtparam to suit.

The slight wart is the wonky way in which sdhci-brcmstb parts out the platform support from MMC/SD layer probe, which means it needs to comprehend the downstream DT modification.


pelwell (approved):

P33M:
I will need to test on CM5 hardware, that means finding lite and eMMC samples...

P33M:
Testing did reveal that the 32GB cards have a lower CID hwrev/fwrev than the 64GB cards. Which is quite annoying, as this CID is shared by the Amazon Basics cards that don't implement CQ - however, allowing these is benign as they get filtered by the mandatory features list.

PR #6592, opened 2025-01-08
add missing kmsg_panic attribute struct used in drm_panic.c

adminy:


pelwell:
You're going to have to explain how _only_ adding a structure to another structure does anything other than increase memory usage.

pelwell:
No, I see - adding CONFIG_DRM_PANIC=y results in a build failure.

pelwell:
This was caused by a bad merge of a downstream change. I've chosen to fix the original patch to append the new fields - see da53e41b64aa3 (before it gets squashed or rebased).

pelwell:
Closing, as an equivalent patch has been added. Thanks!

adminy:
the changes you pointed me to also seem to remove it

pelwell:
The patch was right, the hash was wrong (now corrected) - try da53e41b64aa3.

adminy:
@pelwell Do you have a list of `CONFIG_*=n` kernel configs that could aid in compiling the kernel faster? Because with what's in the default branch, it seems to be compiling amd and intel drivers which I doubt will be of much use on a Pi.

pelwell:
I think you must be doing it wrong. Have you read our guide to building your own kernel? https://www.raspberrypi.com/documentation/computers/linux_kernel.html#building

adminy:
yep that's the one I'm following @pelwell

popcornmix:
Can you point to some specific drivers we are building and shouldn't be?

Issue #3648, opened 2020-05-31
Please enable CRYPTO_ADIANTUM for 5.x kernels


iustin:
Hi, not sure if this is the proper place, let me know where to report it otherwise.

Using rpi-update I finally got a 5.x kernel (`5.4.42-v8+`, more precisely), and I was hoping to be able to use the Adiantum cipher for higher dm-crypt speed. However, it seems it's still not available, as:
```
# zgrep -i adiantum /proc/config.gz
# CONFIG_CRYPTO_ADIANTUM is not set
```

Furthermore, `find /lib/modules/|grep -i adiantum` finds nothing at all.

Could you please enable this? The difference in speed from the plain aes (non-accelerated on RPI boards) is quite high, about 3× according to various reports, so this would be useful (it also means 3× lower CPU usage for a given throughput).

Various links about it:

* https://forum.manjaro.org/t/enable-aes-adiantum-for-linux-rpi4-mainline/116038
* https://www.raspberrypi.org/forums/viewtopic.php?t=252980

I also tried to find the actual configs being used, but I'm not familiar with the raspbian build processes (I don't think they're in this repo per se, or at least not as a monolithic config file).

Thanks!

JamesH65:
Easy enough to build you own kernel. See https://www.raspberrypi.org/documentation/linux/kernel/building.md

Run menuconfig or similar to set the required options.

JamesH65:
Sorry, suggested that just so you could test it, going to be a while before anything change like this makes it in, and we likely want details on the improvements  and any memory/CPU hit.

iustin:
Thanks for the suggestion, that I can do (although I'm not familiar with rpi kernel builds - why is there no debian package built, would make it so much easier… anyway).

I was however hoping that simply including a new module can be done easily, as it wouldn't get used unless someone actually chooses to use this particular encryption algorithm. So it should be a no-op/safe change to anyone else.

jfikar:
I'm running custom arm64 kernel 5.4.45 with Adiantum enabled.

```
CONFIG_CRYPTO_ADIANTUM=m
CRYPTO_NHPOLY1305_NEON=m
```
The speed on RPi4 is

```
cryptsetup benchmark -c xchacha20,aes-adiantum
# Tests are approximate using memory only (no storage IO).
#            Algorithm |       Key |      Encryption |      Decryption
xchacha20,aes-adiantum        256b       211.1 MiB/s       214.1 MiB/s


[...]

PR #6594, opened 2025-01-09
configs: add CONFIG_CRYPTO_NHPOLY1305_NEON=m

plumbeo:
Enable nhpoly1305_neon module to speed up Adiantum disk encryption.


plumbeo:
When Adiantum was enabled (issue #3648) the CONFIG_CRYPTO_NHPOLY1305_NEON option, which significantly speeds up the encryption/decryption speed, was overlooked.

On a Raspberry Pi 3B tests with `cryptsetup benchmark -c xchacha12,aes-adiantum -s 256` with the current 6.6.62 64bit kernel (rebuilt from the original linux_6.6.62-1+rpt1 source package) show a 20% improvement:

```
encryption/decryption speed (average of 5 runs)

governor        w/o nhpoly1305_neon     w/ nhpoly1305_neon      improvement
powersave       61.7/61.7               74.8/74.7               1.21x
performance     118.3/118.4             142.2/142.1             1.20x
```

The option doesn't select anything else (the only diff between the config files is `CONFIG_CRYPTO_NHPOLY1305_NEON=m`) and the new `nhpoly1305-neon.ko.xz` module's size is only 2.2kb so the impact on users who don't set up Adiantum encryption is minimal. 

Can the option be enabled on all kernels?

pelwell:
BCM2835 has no NEON support, and if you use your bcmrpi_defconfig you'll find that `CONFIG_CRYPTO_NHPOLY1305_NEON=m` does not stick.

plumbeo:
> BCM2835 has no NEON support, and if you use your bcmrpi_defconfig you'll find that `CONFIG_CRYPTO_NHPOLY1305_NEON=m` does not stick.

I didn't notice that, sorry. I force pushed a new version of the patch that only touches `arch/arm/configs/{bcm2709_defconfig,bcm2711_defconfig}` and `arch/arm64/configs/{bcm2711_defconfig,bcm2712_defconfig,bcmrpi3_defconfig}`, is that ok?

pelwell:
That looks better. I'll just let the auto-builds complete before merging. Don't worry about the checkpatch Signed-off-by error - all the config patches get squashed together.

plumbeo:
Thank you! Do you need a separate PR for 6.12.y/6.13.y or will the commit be picked up from this branch later?

pelwell:
It will be automatically forward-ported.

PR #6595, opened 2025-01-09
Enable multicodec sound cards

gordoste:
This patch prepares to remove the custom ezsound patches that were introduced in the 6.12 tree.
 
As a reminder, devices with:
- A single DAI link with multiple codecs
- Clocks produced by one of the codecs
- Other codec(s) and CPU are clock consumers

could not be configured properly under ASoC, and the patchset fixed the problem by having the codec driver override the configuration provided by SoC.

A patch series ("ext_fmt") has been developed and tested ([email thread](https://lore.kernel.org/linux-sound/02b6d112-e4bf-4bc7-8cee-6d6fcd4df5e4@iinet.net.au/T/#t)) to address the ASoC limitation, and should be introduced to the 6.13 mainline before too long. However, one of the changes made to the rpi tree as part of [the usual soundcard support patch](https://github.com/raspberrypi/linux/commit/dedad3a267a8715bff11a161450eaae0523badd5) explicitly disables support for multiple codecs. This patch reverts that change.

I compiled kernels from your 6.13.y tree with and without this patch.

6.13.0-rc6 with ext_fmt patchset hangs on arecord (expected, since it has no clock)
6.13.0-rc6 with ext_fmt patchset plus this patch works perfectly (without any of the prior pcm3168a customisations).

I inserted debugging messages which show what is happening:
```
[   28.518387]  ezsound I2S PCM: DAI #0 - NOT tweaking dai_fmt
[   28.518390]  ezsound I2S PCM: DAI #1 - tweaking dai_fmt
```

You might wonder why the current ezsound patches that you accepted previously don't have this problem. It's because it sets up two separate DAI links, one for playback and one for capture, which both use the same CPU. This one is a single DAI link with 1 CPU and 2 codecs. As a result, the ALSA card has a single device used for both playback and capture, which is better IMHO.

Obviously, this is not urgent, but I thought I'd float it in case you wanted to make this change before the upcoming move to 6.12.

EDIT: Just realized I forgot Signed-Off-By. I can fix that of course, I was focussed on trying to explain this clearly.


pelwell:
This will have to be rebased before merging, but hold fire for now until I work out why the change was originally added.

pelwell:
Well, here's the where: https://github.com/raspberrypi/linux/pull/3337

pelwell:
It might speed things up if you could share any thoughts you have about the patch you are reverting, and how the use case outlined can be preserved.

gordoste:
Thanks for digging up why this was originally changed. It looks the Justboom folks had the exact same challenge that I had due to the lack of ASoC support for multicodec setups with a codec as master. The "ext_fmt" patchset I mentioned previously will remove this constraint and allow creation of a DT overlay to support the JustBoom Digi + JustBoom DAC scenario using audio_graph_card2 driver. But that patchset won't be in until 6.13.
I can see 3 options:
1. Undo #3337 in 6.12.y. Users of the JustBoom Digi + JustBoom DAC will need to stay on 6.6 forever. PR #6564 will stay in 6.12 but be removed after that.
2. Undo #3337 in 6.13.y. Users of the JustBoom Digi + JustBoom DAC will need to stay on 6.12 unless the JustBoom guys add support to 6.13. PR #6564 will stay in 6.12 but be removed after that.
3. Keep PR #6564 and PR #3337 permanently.
Note that a lot of the things on JustBoom's site are out of stock (including the JustBoom DAC HAT). It may be unlikely any further development will occur, particularly for old hats.

I suspect the best option is the 2nd one.... if you agree, then you can close this and once the "ext_fmt" patchset actually gets accepted into upstream, I will resubmit it.

gordoste:
FYI, the "ext_fmt" patch set was accepted and should get into the next merge window.

pelwell:
I'm happy to accept backports of upstream commits once merged to linux-next (at least). The cleanest solution appears to be to revert both patches and replace them with suitable use of the ext_fmt feature. I don't know if @shawaj still has an interest in the maintenance of the JustBoom drivers, but either way it would be appreciated if you created a best-effort update of any JustBoom software support that needs to change.

gordoste:
I am not really keen to backport the ext_fmt function from 6.13 into 6.12 (if that's what your saying). It does not apply cleanly and so it seems a much riskier proposition than keeping the pcm3168a patches for 6.12 kernels... that module was not even enabled in the default kernel config previously so I think it's safe to say that no other hardware uses it.

@pelwell let me know if you agree with "keep pcm3168a patches until ext_fmt is available" strategy - if so this PR can be closed and I'll submit a new one when ext_fmt makes it into linux-next.

@johannesk as you added the initial justboom-both support, your comments would definitely be appreciated.

pelwell:
I'm happy to defer the upheaval to rpi-6.13.y (currently on -rc7), meaning that it's unlikely to see much action until 2026.

Issue #6505, opened 2024-11-30
Missing Color Order in KMS DPI Overlay on Compute Module 5 and Pi 5


othermod:
### Describe the bug

The vc4-kms-dpi-generic-overlay.dts overlay is missing some capability that previously existed for changing the color order of DPI output. It gives only bgr888 and rgb888 as options. I developed a product using red instead of green in the middle group, and I'm currently unable to use the KMS DPI overlay.

The previous DPI method gave 4 color orders for RGB
rgb_order:
1: DPI_RGB_ORDER_RGB
2: DPI_RGB_ORDER_BGR
3: DPI_RGB_ORDER_GRB
4: DPI_RGB_ORDER_BRG

This is an identical issue to [https://github.com/raspberrypi/linux/issues/6155](https://github.com/raspberrypi/linux/issues/6155), but is now needed for the Raspberry Pi 5 and CM5. 

The previous fix was effective for the CM4, and I was able to compile an overlay that modifed the color order and enabled i2c, but this does not work on the CM5. Some discussion occurred about doing something similar for the RP1 [https://github.com/raspberrypi/linux/pull/6156](https://github.com/raspberrypi/linux/pull/6156). 

### Steps to reproduce the behaviour

Use the KMS DPI overlay on the CM5 with a product (such as my PSPi 6) that makes use of the previously available DPI color orders that use red in the center group instead of green.

### Device (s)

Raspberry Pi 5, Other

### System

Raspberry Pi OS Bookworm

6by9:
Yes it would be possible to add something similar for RP1, but it hasn't happened yet.

It's probably a more complex version of [set_output_format](https://github.com/raspberrypi/linux/blob/rpi-6.6.y/drivers/gpu/drm/rp1/rp1-dpi/rp1_dpi_hw.c#L300-L346).
Currently RGB vs BGR handling is done by [changing the input format configuration though](https://github.com/raspberrypi/linux/blob/rpi-6.6.y/drivers/gpu/drm/rp1/rp1-dpi/rp1_dpi_hw.c#L401-L402), so fiddling the shifts there may be the better solution.

othermod:
@6by9 I've got something functioning. Is this along the lines of what you are thinking?
[https://github.com/raspberrypi/linux/compare/rpi-6.6.y...othermod:linux:rpi-6.6.y](https://github.com/raspberrypi/linux/compare/rpi-6.6.y...othermod:linux:rpi-6.6.y)

[...]

PR #6597, opened 2025-01-09
drm: rp1: rp1-dpi: Add "rgb_order" property (to match VC4 DPI)

njhollinghurst:
As on VC4, the OF property overrides the order implied by media bus format. Only 4 of the 6 possible orders are supported. New add-on hardware designs should not rely on this "legacy" feature.

This is is an attempt to deal with #6505.

Not all combinations of DRM pixel format, bus format and rgb_order have been tested yet.

### DIFF
--- a/drivers/gpu/drm/rp1/rp1-dpi/rp1_dpi.c
+++ b/drivers/gpu/drm/rp1/rp1-dpi/rp1_dpi.c
@@ -292,6 +292,7 @@ static int rp1dpi_platform_probe(struct platform_device *pdev)
 	struct device *dev = &pdev->dev;
 	struct rp1_dpi *dpi;
 	struct drm_bridge *bridge = NULL;
+	const char *rgb_order = NULL;
 	struct drm_panel *panel;
 	int i, j, ret;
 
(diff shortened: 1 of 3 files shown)


njhollinghurst:
I think this is correct, my only reticence (and the reason it's still a draft) is code size.
There might well be a smaller or cleaner way to achieve the permutation of input (shift,mask,scaling).
Alternatively, if we drop GRB, BRG support for `RGB565_1X16`, `RGB666_1X18` there might be a simpler version that permutes output shifts (though BGR666 will still need asymmetrical settings).

drivers/gpu/drm/rp1/rp1-dpi/rp1_dpi.c
```diff
@@ -353,6 +354,20 @@ static int rp1dpi_platform_probe(struct platform_device *pdev)
 	if (ret)
 		goto done_err;
 
+	dpi->rgb_order_override = -1;
```
6by9: My only comment is that this could potentially be defined as RP1DSPI_ORDER_UNCHANGED, and check for that by name instead of `< 0` at line 423.

njhollinghurst: It's copied from vc4_dpi.c (though in that instance the values were from a register field).

BTW I also toyed with encoding the lookup order more directly in the variable (values like 0x020100) to avoid the switch statement.

njhollinghurst:
Tested on RGB666_CPADHI (Mode 6) and RGB666 (Mode 5) HATs; kmstest with all supported 4CCs over all values of rgb_order. The colours seemed to be going to the right places as far as I could see.

Also tested RGB565 (Mode 2) and RGB888 (Mode 7) albeit without appropriate hardware, and the output looked plausible.

othermod:
Tested on my hardware as well. Verified rgb888 with rgb-order of brg and grb work as expected.

drivers/gpu/drm/rp1/rp1-dpi/rp1_dpi.c
6by9: vc4_dpi.c isn't always the best example, and there is no requirement to copy how it is implemented in the code, only the DT side.

njhollinghurst: Added `RP1DSI_ORDER_UNCHANGED` as an explicit value, and re-worked the lookup to use bit-shifts (which makes both object and source code slightly smaller. That last `& 3` is superstitious since the values are internal).

PR #6598, opened 2025-01-09
iomap: avoid avoid truncating 64-bit offset to 32 bits

marcone:
on 32-bit kernels, iomap_write_delalloc_scan() was inadvertently using a 32-bit position due to folio_next_index() returning an unsigned long. This could lead to an infinite loop when writing to an xfs filesystem.


marcone:
Apologies for the double "avoid" in the title. That's how I accidentally sent it to linux mailing lists, and so that's how it ended up in [vfs.fixes](https://git.kernel.org/pub/scm/linux/kernel/git/vfs/vfs.git/log/?h=vfs.fixes) and will soon in linux-next.

The problem this fixes also exists in 6.6, but this patch doesn't apply cleanly there. Let me know if you want me to upload another PR for 6.6, since without it, xfs isn't reliably usable on 32-bit kernels (i.e. Raspberry Pi Zero)

pelwell:
I'm less happy about this patch, and it's nothing to do with the commit message. I don't understand why you are switching to a completely different way of obtaining the new position, one that isn't obviously equivalent (except for the lack of truncation). The simpler, more obviously correct patch would be to cast the result of `folio_next_index` to an `loff_t` before it is left-shifted.

marcone:
I mainly did it this way because a [previous patch by one of the kernel developers](https://github.com/raspberrypi/linux/commit/eee2d2e6ea5550118170dbd5bb1316ceb38455fb) did it the same way.
Also note that there are already several other places in this file that use `folio_pos(folio) + folio_size(folio)`, so it's a common pattern in this file.
Happy to change it if you want, but it looks like this patch is what's going to go into the upstream kernel, so changing it for Raspberry Pi will just lead to inconvenience down the road.

pelwell:
It would have been nice to word your commit message similarly, including a reference to this prior art.

pelwell:
I'm happy to wait for it to be merged upstream.

marcone:
(and for what it's worth, one of the maintainers did muse that perhaps there should be a folio_next_pos() function that returns a loff_t)

marcone:
> I'm happy to wait for it to be merged upstream.

That's too bad, but it's understandable that Raspberry Pi Zero isn't high on the priority list anymore.

pelwell:
I'm surprised they merged it as is - there isn't even a "Fixes:" tag.

pelwell:
Patch now cherry-picked from linux-next.

PR #6611, opened 2025-01-13
drm/v3d: CPU job submissions shouldn't affect V3D GPU clock

txenoo:
We can avoid calling the v3d_clock_up_put and v3d_clock_up_get when a job is submitted to a CPU queue. This way we avoid delaying timestamps CPU jobs by 4.5ms that is the time that it takes to increase the V3D frequency when the V3D subsystem changes its clock frequency.

Fixes: fe6a858096c1 ("drm/v3d: Correct clock settng calls to new APIs")

### DIFF
--- a/drivers/gpu/drm/v3d/v3d_submit.c
+++ b/drivers/gpu/drm/v3d/v3d_submit.c
@@ -125,9 +125,9 @@ v3d_lookup_bos(struct drm_device *dev,
 }
 
 static void
-v3d_job_free(struct kref *ref)
+v3d_job_free_common(struct v3d_job *job,
+		    bool is_gpu_job)
 {
-	struct v3d_job *job = container_of(ref, struct v3d_job, refcount);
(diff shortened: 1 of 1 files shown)


txenoo:
@popcornmix @mairacanal can you review this, it is a downstream patch as the V3D clock frequency management as this is only enabled downstream. We started seeing this issue with 6.12.

It solves some issues we have to make timestamp queries work using GPU jobs in V3D, so we can land Mesa [v3d: Add support for timestamp, time elapsed, and timestamp disjoint queries](https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/32547)

mairacanal:
LGTM, feel free to add my:

```
Reviewed-by: Maíra Canal <mcanal@igalia.com>

popcornmix:
A quick test looks fine. v3d clock still goes up and down, 3d renders and framerate from aquarium looks reasonable.

pelwell:
Thanks!

PR #6613, opened 2025-01-15
drivers: media: pisp_be: Add support for YUV422 planar format

naushir:
List V4L2_PIX_FMT_YUV422P as supported by the PiSP backend hardware.


6by9:
Is there a gain in having V4L2_PIX_FMT_YUV422P vs V4L2_PIX_FMT_YUV422M (which is already supported)?

I guess it adds symmetry with all the duplicated NV12 / NV12M and YVU420 / YUV420M defines, and the hardware can cope with either.

naushir:
MPlanar formats are not fully supported by libcamera yet, so this is the only way to get 422 planar out.

PR #6620, opened 2025-01-21
v3d backports from 6.13

6by9:
@roliver-rpi for testing.
I have compile tested it, but not booted it.


popcornmix:
Need to confirm this does not introduce the NULL job issue in 6.13 reported here: https://github.com/raspberrypi/linux/issues/6624

pelwell:
Either way, the result would be a useful datapoint.

roliver-rpi:
This PR works to gather v3d gpu counters using perfetto. Also, it doesn't appear to introduce the issue noted in #6624 (no freezing up playing video in both firefox / chromium).

pelwell (approved):
Approval in principal (after rebasing...)

6by9:
Branch rebased.
I haven't compile tested it as the rebase was clean, so please let CI complete. 

This PR doesn't include e4b5ccd392b9 drm/v3d: Ensure job pointer is set to NULL after job completion, so doesn't need the fix discussed in #6624.
e4b5ccd392b9 was merged to drm-misc-fixes as a fix of 14d1d1908696 (from Jun 2018) . It hasn't made it to drm-misc-next which is where I cherry-picked patches from. We can pick up both here, but they should come back to stable releases automatically anyway.

6by9:
Oops, forgot this had the draft tag. Now removed.

PR #6623, opened 2025-01-22
configs: add CONFIG_SND_SOC_WM8904=m

JinShil:


pelwell:
You missed out the original arm/bcmrpi_defconfig, i.e. the early Pis and Pluses, but everything else was fine. It was easier for me to recreate it than patch this PR, so see fed8a0934de84.

JinShil:
Thanks.

PR #6632, opened 2025-01-26
Fix incorrect propagation of frozen state during unfreeze

umarsync:
The existing freezer subsystem incorrectly propagates the frozen state when unfreezing a cgroup with multiple descendants. The critical bug was in cgroup_propagate_frozen() where ancestor cgroups' nr_frozen_descendants were being decremented by an increasing counter ('desc') instead of 1, leading to:
1. Premature unfreezing of parent cgroups
2. Negative nr_frozen_descendants counts
3. Incorrect frozen state reporting

### DIFF
--- a/kernel/cgroup/freezer.c
+++ b/kernel/cgroup/freezer.c
@@ -34,12 +34,17 @@ static void cgroup_propagate_frozen(struct cgroup *cgrp, bool frozen)
 				desc++;
 			}
 		} else {
-			cgrp->freezer.nr_frozen_descendants -= desc;
+			/* Critical fix: Decrement by 1 instead of 'desc' */
+			cgrp->freezer.nr_frozen_descendants -= 1;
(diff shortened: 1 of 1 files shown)


pelwell:
That seems like a good catch. However, that file is not one that we've changed in our version of the kernel, so the patch should be sent to the Linux kernel maintainers. Given the clarity of your explanation, I'm sure @htejun would be he happy to guide you through to its acceptance.

umarsync:
> That seems like a good catch. However, that file is not one that we've changed in our version of the kernel, so the patch should be sent to the Linux kernel maintainers. Given the clarity of your explanation, I'm sure @htejun would be he happy to guide you through to its acceptance.

I appreciate the guidance, I’ll submit this patch upstream to the Linux kernel maintainers as suggested. I’ll also reach out to @htejun for feedback and to ensure alignment with the cgroup subsystem’s workflow.

PR #6635, opened 2025-01-27
6 chip selects on SPI0

jsarrett:
I need 6 chip selects on a custom board of mine, and I'm sure someone else out there might also need more than 2, but I'm not at all sure if this is the right way to implement this. It's kinda ugly that we have to have a separate dts for every possible number of CS# lines. Looking for feedback/ideas.


pelwell:
I'm not keen on the way you've changed the GPIO for CS1 to suit your application - that could cause some confusion. 

Whether there's a cleaner way to implement this is an interesting question. It would be pretty easy to write an overlay that adds a single  spidev at position n, but patching up the cs-gpios is harder to do piecewise like this because I can't see how to patch at the property at varying offsets derived from the assigned value, and it's not possible to just append to an existing property.

Perhaps the cleanest way to do this would be to have one underlying implementation .dtsi that can conditionally add any of the spidevs controlled by macros, then make each spi0-ncs overlay include the implementation with different macro flags.

jsarrett:
is there some way to guarantee fragment application order?  maybe the <&spi0_cs_pins> of different lengths could be enabled when the spidev@n of that length is chosen?  it only works if the longest pin declaration happens last though.  thoughts?

PR #6636, opened 2025-01-27
overlays: Factor out the common i2c bus selection

pelwell:
This PR reduces some of the duplication in I2C overlays, factoring out the bus selection parameters into a .dtsi file. To allow for easier comparison and to improve the codebase, the first commit fixes a few minor problems with the old versions of the overlays. The second commit rewrites them to use the dtsi file, a step which is essentially cosmetic.

The README is also updated, moving the common parameters into a pseudo-overlay called `i2c-bus`. Overlays that use these common parameters are shown as supporting the parameter `i2c-bus`, whose documentation refers the user to the common documentation.

### DIFF
--- a/arch/arm/boot/dts/overlays/vc4-kms-dsi-7inch-overlay.dts
+++ b/arch/arm/boot/dts/overlays/vc4-kms-dsi-7inch-overlay.dts
@@ -1,4 +1,4 @@
-/*
+ /*
  * Device Tree overlay for RaspberryPi 7" Touchscreen panel
  *
  */
@@ -7,6 +7,11 @@
 /plugin/;
 
 #include "edt-ft5406.dtsi"
+
+&ft5406 {
+	vcc-supply = <&reg_display>;
+	reset-gpio = <&reg_display 1 1>;
+};
 
 / {
 	/* No compatible as it will have come from edt-ft5406.dtsi */
@@ -77,7 +82,7 @@
 
 	i2c_frag: fragment@2 {
 		target = <&i2c_csi_dsi>;
-		__overlay__ {
+		i2cbus: __overlay__ {
 			#address-cells = <1>;
 			#size-cells = <0>;
 			status = "okay";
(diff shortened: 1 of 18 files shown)


arch/arm/boot/dts/overlays/vc4-kms-dsi-7inch-overlay.dts
```diff
@@ -1,4 +1,4 @@
-/*
+ /*
```
6by9, line 1: Extra space added.

pelwell:
FYI I'm most interested in feedback for the first commit, since that makes real changes to the resulting configuration. I'm happy to bear the burden for verifying the changes made by the second, but confirmation that the overall scheme looks okay would be nice.

6by9 (commented):
One rogue space, but otherwise looks fine.

(The edt-ft5x06 include had my head hurting for a bit)

6by9:
I think the patch order will mean that some of the overlays will be functionally broken for bisection purposes if using i2c0 or 10 - they appear to drop enabling i2c0mux and i2c0if, which gets added back in with the second patch.
C'est la vie.

Otherwise those cleanups all look reasonable.

arch/arm/boot/dts/overlays/vc4-kms-dsi-7inch-overlay.dts
pelwell: Will fix.

pelwell:
When using i2c0, some overlays enabled i2c0mux and i2c0if while others didn't. It looks I've chosen the wrong option (kind of obvious now looking at some of the dts files). I'll rework it the other way - not a big change.

pelwell:
Updated to address the feedback. With a fixed (not nobbled) ovmerge, a test script that compares the results of applying all the old parameters to each updated overlay before and after the second commit finds zero differences. Therefore I think that if you're happy with the new first commit then its safe to merge.

6by9:
Looks good to me.

I wonder if the original i2c0 support predated the mux, hence some i2c0 options didn't worry about it.
In theory you could use `dtoverlay=i2c0` to enable BSC0 without the mux, and then this is going to cause grief. I wouldn't worry about that case though.

pelwell:
> In theory you could use dtoverlay=i2c0 to enable BSC0 without the mux

For some reason, that's what I was expecting it to do.

PR #6638, opened 2025-01-28
drm/edid: When reset, assume HDMI displays support RGB444

6by9:
If an HDMI connector has no EDID and the mode is set via the kernel command line, then drm_reset_display_info() is the only thing that will have set up any of connector->display_info.

With commit 26ff1c38fc29 ("drm/connector: hdmi: Compute bpc and format automatically"), it is now checked that DRM_COLOR_FORMAT_RGB444 is supported. Whilst it doesn't fail the request, it does log dev_warn for every commit, spamming the log.

For HDMI connectors initialise the color_format field to say it supports RGB444.

https://forums.raspberrypi.com/viewtopic.php?p=2290252


pelwell (approved):

popcornmix:
Merged offline.

PR #6639, opened 2025-01-28
README: show rpi-6.12.y build status

marcone:
Remove rpi-5.15.y build status since it doesn't appear to be built anymore, and add rpi-6.12.y build status.


pelwell:
Thanks!

Issue #6642, opened 2025-02-01
Pi5 failed to contact RP1 firmware


llllllllqq:
### Describe the bug

dmesg says:

...
`rp1-pio 1f00178000.pio: error -ENOENT: failed to contact RP1 firmware
[    3.957885] rp1-pio: probe of 1f00178000.pio failed with error -2`
...

Although nothing strange occurred. 
The OS still works fine without reboot.

### Steps to reproduce the behaviour

It randomly occurs after `sudo reboot --halt` and then manually press the on-board power button to power on (with default eeprom config).

~~It also randomly occurs when use raspi-config to change power off behavior to "full power off" and then reboot/poweroff.~~ (Sorry I cannot remember whether it is reboot or poweroff)

Update: It also occurs after a suddenly plug-out the usb-c power cable and then plug-in back.
~~It seems only poweroff-related behaviors will trigger this bug. I tested 30 times reboot, no bugs occurred.~~
Update2: reboot can indeed also trigger this bug, only less frequent.

### Device (s)

Raspberry Pi 5

### System

Raspberry Pi reference 2024-11-19                     Generated using pi-gen, https://github.com/RPi-Distro/pi-gen, 891df1e21ed2b6099a2e6a13e26c91dea44b34d4, stage4

2025/01/22 00:16:51
Copyright (c) 2012 Broadcom                           version a7753063 (release) (embedded)

Linux rpi5 6.6.74+rpt-rpi-2712 #1 SMP PREEMPT Debian 1:6.6.74-1+rpt1 (2025-01-27) aarch64 GNU/Linux

### Logs

[1.txt](https://github.com/user-attachments/files/18627781/1.txt)

### Additional context

Can it be safely ignored?

pelwell:
That's interesting - it looks like a dependency/ordering issue, and not one I've seen before:
```
[    3.922665] rp1-pio 1f00178000.pio: error -ENOENT: failed to contact RP1 firmware
[    3.922671] rp1-pio: probe of 1f00178000.pio failed with error -2
...
[    3.932284] rp1-firmware rp1_firmware: RP1 Firmware version eb39cfd516f8c90628aa9d91f52370aade5d0a55
```
I get:
```
[   13.445656] rp1-firmware rp1_firmware: RP1 Firmware version eb39cfd516f8c90628aa9d91f52370aade5d0a55
[   13.550428] rp1-pio 1f00178000.pio: Created instance as pio0
```
[...]

PR #6645, opened 2025-02-03
Handle probe dependencies and hard errors better

pelwell:
#6642 is a report of intermittent failures for the PIO driver to contact the firmware. These failures can be traced to the failure of PIO driver to distinguish hard errors and transient failures. The commits in this PR fix the PIO driver in this regard, but also improve the firmware driver to avoid pointless retries in the event that the firmware is not compatible.

### DIFF
--- a/drivers/firmware/rp1.c
+++ b/drivers/firmware/rp1.c
@@ -114,7 +114,8 @@ static void rp1_firmware_delete(struct kref *kref)
 
 void rp1_firmware_put(struct rp1_firmware *fw)
 {
-	kref_put(&fw->consumers, rp1_firmware_delete);
+	if (!IS_ERR_OR_NULL(fw))
+		kref_put(&fw->consumers, rp1_firmware_delete);
 }
 EXPORT_SYMBOL_GPL(rp1_firmware_put);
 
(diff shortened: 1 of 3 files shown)


drivers/firmware/rp1.c
```diff
@@ -270,19 +270,18 @@ static int rp1_firmware_probe(struct platform_device *pdev)
...
 	if (ret == sizeof(version)) {
 		dev_info(dev, "RP1 Firmware version %08x%08x%08x%08x%08x\n",
 			 version[0], version[1], version[2], version[3], version[4]);
-		ret = 0;
-	} else if (ret >= 0) {
-		ret = -EIO;
+		platform_set_drvdata(pdev, fw);
+	} else {
+		kfree(fw);
+		platform_set_drvdata(pdev, ERR_PTR(-ENOENT));
```
popcornmix, line 282: We do dereference this in `rp1_firmware_remove` (without a check). 
I'm not sure if that can happen, but looks like a potential pitfall.

pelwell: Testing the obvious fix has uncovered some deficiencies in the end-of-life code. Thank goodness for runtime overlays.

pelwell:
With the latest updates, the rp1 firmware driver and pio driver can be repeatedly instantiated and removed with no crashes or obvious leaks.

popcornmix (approved):

PR #6646, opened 2025-02-03
spi: dw: Wait for idle after TX

pelwell:
This is probably part 1 of a fix for TX only transfers. Part 2 is likely to change dw_spi_dma_transfer to not call dw_spi_wait_tx_done in the event that there is an RX buffer.

Link: https://forums.raspberrypi.com/viewtopic.php?t=383027

### DIFF
--- a/drivers/spi/spi-dw-dma.c
+++ b/drivers/spi/spi-dw-dma.c
@@ -304,6 +304,11 @@ static int dw_spi_dma_wait_tx_done(struct dw_spi *dws,
 		return -EIO;
 	}
 
+	if (!xfer->rx_buf) {
+		while (dw_readl(dws, DW_SPI_SR) & DW_SPI_SR_BUSY)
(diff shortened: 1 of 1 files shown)


pelwell:
Updated with a more complete solution, which needs retesting in its current form. Restructuring `dw_spi_dma_transfer` to potentially only call `dw_spi_wait_rx_done` is not an option given that `dw_spi_wait_rx_done` assumes that `dw_spi_wait_tx_done ` has already completed.

P33M:
The maximum time between the tx_busy check passing (==txfifo empty) and the SR_BUSY check passing is one complete wordsize. I think the cpu_relax() should be another invocation of spi_delay_exec() but with delay.value set to just n_bytes * BITS_PER_BYTE - in the case where you have a slow SPI SCLK and a large wordsize, the driver could be spinning for a nontrivial amount of time.

PR #6649, opened 2025-02-04
fixup! spi: dw: Wait for idle after TX

pelwell:
Relax a bit harder - transmission of the last bits may take a while.

### DIFF
--- a/drivers/spi/spi-dw-dma.c
+++ b/drivers/spi/spi-dw-dma.c
@@ -305,8 +305,9 @@ static int dw_spi_dma_wait_tx_done(struct dw_spi *dws,
 	}
 
 	if (!xfer->rx_buf) {
+		delay.value = dws->n_bytes * BITS_PER_BYTE;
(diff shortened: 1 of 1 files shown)


pelwell:
@P33M Surprisingly, this seems to reduce the delay between the final SCLK and CS going high to around 2.6us, whereas before it was at least 4us.

P33M:
It'll be using a more accurate guess as to when the shift register will go idle, but I'm surprised that spamming reads of the status register isn't faster. `cpu_relax()` compiles to `yield` on aarch64 which is basically a nop on non-hyperthreaded ARM cores.

That said, round-trip time is ~1.6us for RP1 register reads.

pelwell:
The test program I'm using sets SCLK to 2MHz. At that speed, a single byte is transmitted in 4us. Perhaps the more heavy-weight `spi_delay_exec` actually has the effect of delaying the first read of `DW_SPI_SR` sufficiently that it has already gone idle, meaning that a second read is not required.

PR #6650, opened 2025-02-04
arm64: dts: Prepare for #size-cells = 2

pelwell:
With a future release of the firmware, it will be possible to use dts files with a top-level #size-cells of 2. This patch adds the remaining necessary changes to make that work, gated by the macro symbol FIRMWARE_UDPATED.


pelwell:
Given that all changes are conditional on an undefined symbol, merging now has no effect on the output.

PR #6653, opened 2025-02-05
PCI: Release unused bridge resources during resize

P33M:
From https://patchwork.kernel.org/project/linux-pci/patch/20240507213125.804474-1-alex.williamson@redhat.com/

Take the suggested patch which frees upstream bridge resources if a BAR is being resized.


geerlingguy:
See also: https://github.com/raspberrypi/linux/pull/7072

PR #6662, opened 2025-02-11
ASoC: enable simultaneous usage of ADC8x and DAC8x

j-schambacher:
This modification enables the DAC8x driver to auto-detect a stacked ADC8x board. The combo allows 8 channel capture and play.
It uses all I2S data pins, and GPIO5 for detection. Needs to use the dummy-DAI instead of the 'play'-only formerly used pcm5102-DAI. Symmetric operation only for obvious reasons.

For potential standalone operation of the ADC8x we keep the ADC8x-dedicated code as is.

### DIFF
--- a/arch/arm/boot/dts/overlays/hifiberry-dac8x-overlay.dts
+++ b/arch/arm/boot/dts/overlays/hifiberry-dac8x-overlay.dts
@@ -1,6 +1,9 @@
+// SPDX-License-Identifier: GPL-2.0
 // Definitions for HiFiBerry DAC8x
 /dts-v1/;
 /plugin/;
+
+#include <dt-bindings/gpio/gpio.h>
 
 / {
 	compatible = "brcm,bcm2712";
@@ -10,8 +13,10 @@
 		__overlay__ {
 			rp1_i2s0_dac8x: rp1_i2s0_dac8x {
 				function = "i2s0";
-				pins = "gpio18", "gpio19", "gpio21",
-				       "gpio23", "gpio25", "gpio27";
+				pins = "gpio18", "gpio19", "gpio20",
+				       "gpio21", "gpio22", "gpio23",
+				       "gpio24", "gpio25", "gpio26",
+				       "gpio27";
 				bias-disable;
 				status = "okay";
 			};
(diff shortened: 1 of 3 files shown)


arch/arm/boot/dts/overlays/hifiberry-dac8x-overlay.dts
```diff
@@ -43,6 +48,7 @@
 		__overlay__ {
 			compatible = "hifiberry,hifiberry-dac8x";
 			i2s-controller = <&i2s_clk_producer>;
+			hasadc-gpio = <&gpio 5 GPIO_ACTIVE_HIGH>;
```
pelwell: I don't know the details of the hardware, but it looks strange to have a detection mechanism where the "detected" value is the same as the default with nothing connected - GPIO 5 pulls high by default, and this overlay doesn't change that.

sound/soc/bcm/rpi-simple-soundcard.c
```diff
@@ -354,16 +354,46 @@ static struct snd_rpi_simple_drvdata drvdata_hifiberry_dac = {
...
 	codec_dai->driver->playback.rates = SNDRV_PCM_RATE_8000_192000;
 
+	/* Activate capture based on ADC8x detection */
+	gpio_desc = devm_gpiod_get(card->dev, "hasadc", GPIOD_IN);
+	if (IS_ERR(gpio_desc)) {
+		dev_err(card->dev, "Failed to get GPIO: %ld\n", PTR_ERR(gpio_desc));
+		return PTR_ERR(gpio_desc);
+	}
+
+	has_adc = !gpiod_get_value(gpio_desc);
```
pelwell: Oh - then you invert it. Wouldn't it be more accurate, more obvious and generally cleaner to declare the signal as active low in the overlay, removing the need for this negation?

j-schambacher: Yes, you are absolutely right. It's just that I don't like 'negative' logic ... I'll wait for the checks to complete and change it.

pelwell: The checks are going to pass - I've run a cut-down version on a faster machine already.

pelwell:
Much better.

j-schambacher:
As usual, thanks for your suggestions and support!

PR #6666, opened 2025-02-12
Backport HEVC decoder as submitted to linux-media as v2

6by9:
This PR:
- reverts the downstream HEVC driver and supporting patches
- backports the driver as submitted as v2 (https://lore.kernel.org/r/20250206-media-rpi-hevc-dec-v2-0-69353c8805b2@raspberrypi.com) with DT binding fixup requested by Rob H, and dependencies.
- adds in the new SAND handling for DRM
- reinstates the downstream single planar pixel formats for decode
- reinstates being able to set the /dev/videoN node

This doesn't work against the existing FFmpeg as it is looking for extra flags to identify a device that it thinks can decode HEVC. @jc-kynesim was going to release his updated FFmpeg that handles this, otherwise I can look at whether we can expose a second /dev/videoN node that adds in those flags as an interim.

This wants more thorough testing, hence looking to push it to rpi-6.12.y.


popcornmix:
I think we want ffmpeg update public (or the second /dev/videoN node) before merging this.

6by9:
> I think we want ffmpeg update public (or the second /dev/videoN node) before merging this.

I'd asked @jc-kynesim whether the FFmpeg side had been released on Friday, and he'd replied
> Not yet - but that is only the work of an hour or two. I'll do that today

I don't know if that happened.

6by9:
Rebased and reworked so that the client is required to pass the column height in the pitch field if the modifier is DRM_FORMAT_MOD_BROADCOM_SAND128.

It does require removing one validation clause in the core DRM code if a modifier is set, and that may be an argument with upstream (I'll look forward to it!)

We can now have a single new path that works with both DRM and EGL (as of Mesa 24.2.5) rendering.

This still has the revert of the new pixel format so that it shouldn't change anything on systems without Mesa 24.

6by9:
Rebased.

AFAIK this is ready to go.
The FFmpeg changes should already be in Raspberry Pi OS.
(The latest changes for the newer image formats are not needed at present, nor is Mesa 24)

The only question is whether Libreelec folks are aware of the change, and have picked up the base FFmpeg changes before taking this kernel.

popcornmix:
> The only question is whether Libreelec folks are aware of the change, and have picked up the base FFmpeg changes before taking this kernel.

I've let them know (but they'd already been following this PR).

popcornmix:
I can't get hevc decode to work with kodi with this PR. Without it, it is working.
I do have latest ffmpeg from internal repo "5.1.6-0+deb12u1+rpt3"

popcornmix:
Ah - this breaks hevc decode with kodi21 in RPiOS, as it uses embedded ffmpeg (kodi21 wants a newer ffmpeg).
I'll need to rebuild the package with updated ffmpeg.

It looks like the older kodi package still works as that uses system ffmpeg.

I'm happy with this being merged for now (we can ensure kodi21 is updated before this kernel reaches apt).

popcornmix:
While kodi 20 worked on a Pi5. I can't play hevc on a Pi4.
ffmpeg is updated. hevc works without this PR but stops with it.
```
2025-03-05 19:52:14.984 T:2002    error <general>: ffmpeg[0x0]: [hevc] mediabufs_poll_cb: Timeout
2025-03-05 19:52:14.984 T:2002     info <general>: ffmpeg[0x0]: Unable to reinit media request: Device or resource busy
2025-03-05 19:52:14.984 T:1996    error <general>: ffmpeg[0x0]: [hevc] Unable to set controls: Device or resource busy
2025-03-05 19:52:14.984 T:1996    error <general>: ffmpeg[0x0]: [hevc] send_slice: Failed to set req ctls
2025-03-05 19:52:14.984 T:1996     info <general>: ffmpeg[0x0]: Unable to reinit media request: Device or resource busy
2025-03-05 19:52:14.984 T:1996    error <general>: ffmpeg[0x0]: [hevc] hardware accelerator failed to decode picture
2025-03-05 19:52:14.984 T:1996    debug <general>: ffmpeg[0x0]: v4l2_req_frame_free: avctx=0x7ef400f5f0 data=0x7ee8081640
2025-03-05 19:52:14.999 T:2002     info <general>: ffmpeg[0x0]: Unable to reinit media request: Device or resource busy
```
@HiassofT also saw similar behaviour is libreelec.

Suprisingly Pi 5 is okay with the same hevc file (sony football one).

arch/arm/boot/dts/broadcom/bcm2711-rpi-ds.dtsi
```diff
@@ -147,18 +147,11 @@
...
-		reg-names = "intc",
-			    "hevc";
-		interrupts = <GIC_SPI 98 IRQ_TYPE_LEVEL_HIGH>;
-
-		clocks = <&firmware_clocks 11>;
-		clock-names = "hevc";
-	};
+&hevc_dec {
+	reg = <0x0 0x7eb10000  0x0 0x1000>,  /* INTC */
+	      <0x0 0x7eb00000  0x0 0x10000>; /* HEVC */
```
HiassofT: INTC and HEVC are swapped compared to bcm2711.dtsi - this breaks HEVC on RPi4.

~likely the hevc_dec override should just be dropped as reg is set in bcm2711.dtsi anyways~ EDIT: scratch that

6by9: Thanks.
Yes it is needed as `size-cells` differs between upstream and downstream, but the swapping is wrong.

6by9:
Updated to correct bcm2711-rpi-ds.dtsi

pelwell:
I'm happy for this to be merged if somebody can confirm that Pi 4 is working now.

popcornmix:
Yes - hevc is working on Pi4 for me with this update.

PR #6667, opened 2025-02-13
drm/vc4: PV1 can be driven via any HVS channel, so adapt to allow it

6by9:
It was noted that if PV1 was in use to drive DSI1, then the writeback connector could not be used as HVS channel 2 was already in use.
The HVS allows PV1 (HVS output 2) to be driven by any HVS channel via the DSP3_MUX setting, but that was hardcoded to be either 2 (for PV1) or disabled for TXP.

Expand the available channels field for PV1, and configure DSP3_MUX accordingly.

https://forums.raspberrypi.com/viewtopic.php?p=2294596


6by9:
Confirmed on the forum thread as doing the job for them too.
I'll be sending this upstream as an improvement.

popcornmix (approved):

PR #6668, opened 2025-02-13
RP1 Audio Out

njhollinghurst:
Candidate version of an RP1 Audio Out driver

48kHz stereo only. GPIO 12/13 only.

Now rewritten as an ASOC DAI driver; a generic sound card and "dummy" codec must also be enabled via OF.


njhollinghurst:
@pelwell could you cast an eye over the first two commits please (since renamed for rebase), to check they don't break I2S or anything else? Thanks.

pelwell:
Do you need a card to test with?

njhollinghurst:
Successfully tested JustBoom DAC Zero and IQaudIO Pi-Codec+ hats at various sample rates.

I could even load and use the rp1aout driver at the same time, but (predictably) I2S stomped on rp1aout's clock when using the (crystal-less) JustBoom hat; they did seem able to coexist with the IQaudIO.

Getting 48000Hz and 44100Hz-based rates at the same time is impractical with a single audio PLL. Running two outputs at (48000\*32, 48000\*3200) _ought_ to be achievable, but the RP1 clk driver doesn't currently oblige. It's not a vital use case.

arch/arm64/boot/dts/broadcom/rp1.dtsi
```diff
@@ -789,6 +805,16 @@
 				       "gpio24", "gpio25";
 				bias-disable;
 			};
+			rp1_dpi_18bit_cpadhi_gpio1: rp1_dpi_18bit_cpadhi_gpio1 { /* Mode 6 with GPIO1(DEN) but not GPIO0 */
```
P33M: dpi? in my audio driver? (it's more likely than you think.)

sound/soc/raspberrypi/rp1_aout.c
```diff
@@ -0,0 +1,360 @@
...
+					      &rp1_aout_component,
+					      &rp1_aout_dai, 1);
+	if (ret)
+		return dev_err_probe(&pdev->dev, ret, "failed to register dai\n");
+
+	ret = devm_snd_dmaengine_pcm_register(&pdev->dev, NULL, 0);
+	if (ret)
+		return dev_err_probe(&pdev->dev, ret, "failed to register pcm\n");
+
+	/* TODO: More PM stuff? */
```
P33M: I don't think this will have an effect unless you define dev_pm_ops with suspend/resume/idle callbacks.

sound/soc/raspberrypi/rp1_aout.c
```diff
@@ -0,0 +1,360 @@
...
+static int rp1_aout_trigger(struct snd_pcm_substream *substream, int cmd,
+			    struct snd_soc_dai *dai)
+{
+	struct rp1_aout *aout = snd_soc_dai_get_drvdata(dai);
+
+	if (cmd == SNDRV_PCM_TRIGGER_STOP ||
+	    cmd == SNDRV_PCM_TRIGGER_SUSPEND ||
+	    cmd == SNDRV_PCM_TRIGGER_PAUSE_PUSH) {
+		/* Push a zero sample (assuming DMA has stopped already?) */
+		aout_reg_wr(aout, AUDIO_OUT_SAMPLE_FIFO, 0);
```
P33M, line 243: Underflow should be handled gracefully by the hardware (the last sample plays forever) - ASoC decouples DMA ops from driver visibility so you may introduce a glitch here if DMA is still running. I don't know if it synchronously drains the DMA buffers or flushes them.

arch/arm64/boot/dts/broadcom/rp1.dtsi
njhollinghurst: That one was a DO NOT MERGE. That particular DPI mode is incompatible with audio out. It really should have gone on a different branch...

sound/soc/raspberrypi/rp1_aout.c
njhollinghurst: Yes, that is what I was wondering. I suppose it ought to be done.

sound/soc/raspberrypi/rp1_aout.c
njhollinghurst: In tests it does seem  to use the given value quite reliably -- I'll remove the question mark

It wasn't clear to me whether a step to zero would be more or less objectionable than a nonstandard DC value (perhaps followed by a step later). Maybe 'stop' and 'pause' cases should be treated differently?

njhollinghurst: After a bit more debug: I haven't actually observed a PAUSE_PUSH or SUSPEND -- when I Ctrl-Z `aplay`, it does a STOP. I think I will keep the zero as it seems to work, and may be better than sticking at an arbitrary value.

arch/arm64/boot/dts/broadcom/rp1.dtsi
njhollinghurst: Gone now

njhollinghurst:
Some unwanted stuff has now been pruned, including the DPI weak.

This PR will likely be abandoned in favour of a more squashed version -- mainly so I can keep the "grungy" version commits, for reference.

We'll need a way to encapsulate enabling Audio Out with cross-Pi compatibility, which may require either more DTS trickery or perhaps firmware support.

P33M:
Rebooting with the carrier active doesn't result in a clean shutdown - it goes thump. snd_soc_dai_ops needs a shutdown callback...

njhollinghurst:
> Rebooting with the carrier active doesn't result in a clean shutdown - it goes thump. snd_soc_dai_ops needs a shutdown callback...

I think platform shutdown is the one I want: DAI shutdown happens after every stream, and we decided we wanted to leave the "bias" on in those cases. But I am currently battling a suspected race condition here...

njhollinghurst:
Created a tag https://github.com/njhollinghurst/linux/tree/audio_out_wip with current history.
Merciless squashing will now ensue.

njhollinghurst:
Sigh... Something seems to go wrong on about 1 boot in 20.  I get this error:
```
pi@raspberrypi:~ $ aplay -v arpanet_20201206.raw -D hw:CARD=RP1AudioOut -f dat
aplay: main:831: audio open error: Invalid argument
```
The DAI driver sees successful probe, startup and shutdown calls only. In the bad case there is never a call to HW params.

pelwell:
For extra credit, create an audremap-pi5 entry in the map that only says bcm2712, so that attempts to load it on other platforms will be blocked.

sound/soc/raspberrypi/rp1_aout.c
njhollinghurst: Or maybe there's nothing to be done, as there's very little in this component to consume or save power.

njhollinghurst:
Sorry for the churn -- I think it's all good now.

- audremap-pi5 is marked as 2712 only
- References to runtime pm removed (as I didn't intend to implement them here/now)
- Taking more care about spiv async register synchronization (at startup and when polling the mute FSM)
- eschewing SET/CLR aliases for superstitious reasons (and because we hardly used and don't need them)
- Reordered PCM and ASOC component registration, to fix a probe-time race condition.
- After moving them around to try to influence probe order, the generic audio nodes are back at top level again.

ak-rex:
Will this be backported to 6.6.y?

njhollinghurst:
> Will this be backported to 6.6.y?

Wasn't planning to -- we're about to migrate to 6.12.

It needed a tweak for 6.6. A version is sitting at: https://github.com/njhollinghurst/linux/tree/audio_out_6.6 but probably won't be merged to avoid potential conflicts.

popcornmix:
@ak-rex the next apt kernel released will be 6.12. There will be no more official 6.6 kernel releases.
Are you wanting this patch in 6.6 for some other reason?

ak-rex:
nope, i'll rebase to the newer kernel.

Issue #6673, opened 2025-02-18
media: i2c: arducam-pivariety - Uninitialized mutex causes NULL pointer dereference


Fenix-125:
### Describe the bug

The arducam-pivariety driver in drivers/media/i2c/ does not initialize a mutex before use, leading to a NULL pointer dereference under certain conditions.

### Steps to reproduce the behaviour

1. Load the arducam-pivariety module.
2. Add it to the device tree or load the module in runtime.
3. Start a video stream from the camera.
4. Observe a kernel crash in dmesg.

### Device (s)

Raspberry Pi 4 Mod. B

### System

```
$ uname -a
Linux raspberrypi 5.10.168-v8+ #XXX SMP PREEMPT Wed Jan 4 15:30:05 GMT 2023 aarch64 GNU/Linux
$ vcgencmd version
Feb  6 2023 14:32:01 
Copyright (c) 2012 Broadcom
version 7d9f0c801f040204dfed308a51d77ef33fbd980d (clean) (release) (start)
$ cat /etc/rpi-issue
Raspberry Pi OS 10 (buster)
```

### Logs

```
[    8.564976] Unable to handle kernel NULL pointer dereference at virtual address 0000000000000000
[    8.573813] Mem abort info:
[    8.576630]   ESR = 0x96000044
[    8.579679]   EC = 0x25: DABT (current EL), IL = 32 bits
[    8.584986]   SET = 0, FnV = 0
[    8.588093]   EA = 0, S1PTW = 0
[    8.591228] Data abort info:
[    8.594095]   ISV = 0, ISS = 0x00000044
[    8.597919]   CM = 0, WnR = 1
[    8.600882] user pgtable: 4k pages, 48-bit VAs, pgdp=00000008b0120000
[    8.607307] [0000000000000000] pgd=0000000000000000, p4d=0000000000000000
[    8.614107] Internal error: Oops: 96000044 [#1] PREEMPT SMP
[    8.685323] pstate: 20000005 (nzCv daif -PAN -UAO -TCO BTYPE=--)
[    8.691324] pc : __mutex_lock.constprop.0+0x1ec/0x5c0
[    8.696364] lr : __mutex_lock.constprop.0+0x1b8/0x5c0
[    8.701400] sp : ffff800012c0bad0
[    8.704701] x29: ffff800012c0bb00 x28: ffff000832d3b900 
[    8.709997] x27: 0000000000000000 x26: 00000000c040563d 
[...]

PR #6674, opened 2025-02-18
media: i2c: arducam-pivariety: Fix mutex init and NULL pointer

Fenix-125:
The mutex used in arducam-pivariety was not properly initialized,
which could lead to undefined behavior. This also caused a NULL
pointer dereference under certain conditions. 

This patch ensures the mutex is correctly initialized during probe
and prevents NULL pointer dereferences.

### DIFF
--- a/drivers/media/i2c/arducam-pivariety.c
+++ b/drivers/media/i2c/arducam-pivariety.c
@@ -1208,6 +1208,8 @@ static int pivariety_enum_controls(struct pivariety *pivariety)
 	if (ret)
 		return ret;
 
+	mutex_init(&pivariety->mutex);
+
 	index = 0;
 	while (1) {
 		ret = pivariety_write(pivariety, CTRL_INDEX_REG, index);
@@ -1295,6 +1297,7 @@ static int pivariety_enum_controls(struct pivariety *pivariety)
 	v4l2_ctrl_handler_setup(ctrl_hdlr);
 	return 0;
 err:
+	mutex_destroy(&pivariety->mutex);
 	return -ENODEV;
 }
 


Fenix-125:
Fixes: #6673

pelwell:
Is that the latest point you can safely initialise the mutex? Moving it later, if safe, would reduce the changes needed.

Any comments, @ArduCAM?

Fenix-125:
In the probe function, the mutex is not utilized before the assignment of asynchronous callbacks. The latest appropriate point for its initialization would be just before calling [pivariety_enum_controls](https://github.com/raspberrypi/linux/blob/rpi-6.6.y/drivers/media/i2c/arducam-pivariety.c#L1395). However, I recommend initializing the mutex immediately after memory allocation to mitigate potential future issues related to uninitialized mutex usage.

This approach is inspired by the [imx319.c](https://github.com/raspberrypi/linux/blob/rpi-6.6.y/drivers/media/i2c/imx319.c#L2446) driver, which follows a similar pattern to ensure robustness. Please let me know your thoughts on this.

pelwell:
I think you could probably move it much later, but I'm not going to get hung up on it - what's there looks OK, even if it isn't optimal.

pelwell:
I'll give Lee a chance to respond, but otherwise I'm happy to merge.

6by9:
You've added the error handling to the error paths in pivariety_probe to destroy the mutex, but not in pivariety_remove.

Otherwise looks fine to me.
(ack to Phil's comments on placement of the init, but if you have to include the error path anywhere, then it makes little difference)

Fenix-125:
@6by9 please note that the mutex is [destroyed](https://github.com/raspberrypi/linux/blob/rpi-6.6.y/drivers/media/i2c/arducam-pivariety.c#L991) in [pivariety_remove](https://github.com/raspberrypi/linux/blob/rpi-6.6.y/drivers/media/i2c/arducam-pivariety.c#L1434) by [pivariety_free_controls](https://github.com/raspberrypi/linux/blob/rpi-6.6.y/drivers/media/i2c/arducam-pivariety.c#L988)

pelwell:
That means you've got a double free in some error paths. Wouldn't it make more sense to move the mutex initialisation into `pivariety_enum_controls`, which looks like the counterpart to `pivariety_free_controls`?

pelwell:
See the imx500 driver for a similar pattern: https://github.com/raspberrypi/linux/blob/rpi-6.6.y/drivers/media/i2c/imx500.c#L2832

Fenix-125:
> That means you've got a double free in some error paths. Wouldn't it make more sense to move the mutex initialisation into `pivariety_enum_controls`, which looks like the counterpart to `pivariety_free_controls`?

You are right there is a double initialisation.
I moved the destroy to the outer scope. 
My motivation is initializing the mutex immediately after memory allocation to mitigate potential future issues related to uninitialized mutex usage.

pelwell:
I'm certain there won't be - this is simple linear initialisation code, and the mutex should only be needed after the probe functions returns.

Fenix-125:
Your suggestion follows the approach used in imx500, and I see many examples of this pattern. I will update the implementation accordingly.

However, regarding [this line](https://github.com/raspberrypi/linux/blob/rpi-6.6.y/drivers/media/i2c/imx500.c#L2845), which copies the mutex to the control handler—do we need to include this as well? I believe it is unnecessary in our case. What are your thoughts?

ArduCAM:
@Fenix-125  Thank you very much for submitting the bug fix, @pelwell  Thank you for your reminder.
I reviewed the above conversation and code, and I agree with pelwell that mutex is only needed after probe, and the current fix has a double release problem in some specific cases.

Fenix-125:
Please review my new changes.

pelwell:
I'm on a phone so review is not easy, but aren't you missing the mutex destroy from the free controls function? With that in place you should be done.

popcornmix:
`pivariety_free_controls` did contain a `mutex_destroy` call (prior to this PR).

Fenix-125:
@pelwell please note that the mutex is [destroyed](https://github.com/raspberrypi/linux/blob/rpi-6.6.y/drivers/media/i2c/arducam-pivariety.c#L991) in [pivariety_remove](https://github.com/raspberrypi/linux/blob/rpi-6.6.y/drivers/media/i2c/arducam-pivariety.c#L1434) by [pivariety_free_controls](https://github.com/raspberrypi/linux/blob/rpi-6.6.y/drivers/media/i2c/arducam-pivariety.c#L988)

Fenix-125:
mutex_destroy -> [arducam-pivariety.c Line 991](https://github.com/raspberrypi/linux/blob/rpi-6.6.y/drivers/media/i2c/arducam-pivariety.c#L991)

pelwell:
So we're good to go. Feel free to merge when the builds complete.

pelwell (approved):

Fenix-125:
I updated the commit message. Please rerun the test.

Fenix-125:
Please merge. I assume that I do not have access to merge this PR.

PR #6675, opened 2025-02-18
Forward-port downstream pcie-brcmstb patches to 6.14

P33M:
This applies on top of the upstream bcm2712-aware pcie-brcmstb driver and restores downstream bugfixes and compatibility features.

The intent is to keep all the conflated 2712/platform-specific adjustments relatively self-contained, hopefully avoiding merge conflicts until they land upstream.


popcornmix (approved):

popcornmix:
This booted for me, with working ssd/m2hat and ethernet.

Documentation/devicetree/bindings/pci/brcm,stb-pcie.yaml
```diff
@@ -101,6 +101,12 @@ properties:
     minItems: 1
     maxItems: 3
 
+  brcm,tperst-clk-ms:
+    description: u32 giving the number of milliseconds to extend
+      the time between internal release of fundamental reset and
+      the deassertion of the external PERST# pin. This has the
+      effect of increasing the Tperst_clk phase of link init.
+
```
6by9, line 112: Describe the units and range, eg
```
    $ref: /schemas/types.yaml#/definitions/uint32
    maximum: 5000
```

Documentation/devicetree/bindings/pci/brcm,stb-pcie.yaml
```diff
@@ -84,6 +84,11 @@ properties:
     $ref: /schemas/types.yaml#/definitions/string
     enum: [ safe, no-l1ss, default ]
 
+  brcm,fifo-qos-map:
+    description: 2712 only. Each nibble assigns all per-TC FIFOs an AXI
+      priority based on fullness (backpressure signalling).
+      Mutually exclusive with vdm-qos-map.
```
6by9, line 90: Units and range.

Do we have to encode it within DT, or could it be an array of N values?

Is it mandatory to have one or other of these, or is neither acceptable?

The bit you'll really love is that commenting "2712 only" doesn't wash in bindings. YAML has lots of lovely syntax to denote which compatibles support which properties.
eg https://github.com/torvalds/linux/blob/master/Documentation/devicetree/bindings/i2c/i2c-mux-pca954x.yaml#L95-L126

`make dt_binding_check` checks whether your bindings are valid.
`make CHECK_DTBS=y broadcom/bcm2712-rpi-5-b.dts` checks whether your dtb file complies with the requirements of the binding.
(Add the relevant environment vars if cross-compiling).

drivers/pci/controller/pcie-brcmstb.c
```diff
@@ -1762,7 +1803,7 @@ static const struct pcie_cfg_data bcm2712_cfg = {
 	.perst_set	= brcm_pcie_perst_set_7278,
 	.bridge_sw_init_set = brcm_pcie_bridge_sw_init_set_generic,
 	.post_setup	= brcm_pcie_post_setup_bcm2712,
-	.quirks		= CFG_QUIRK_AVOID_BRIDGE_SHUTDOWN,
+	.quirks		= CFG_QUIRK_AVOID_BRIDGE_SHUTDOWN | CFG_QUIRK_NO_SSC,
```
6by9, line 1961: Handling of this SSC quirk should be a separate patch to the UBUS-AXI changes. The commit text doesn't reference this at all.

Documentation/devicetree/bindings/pci/brcm,stb-pcie.yaml
```diff
@@ -96,6 +101,12 @@ properties:
     minItems: 1
     maxItems: 3
 
+  brcm,vdm-qos-map:
+    description: 2712 only. Each nibble assigns each per-TC FIFO a base AXI
+      priority with elevation depending on Vendor Messages from the EP -
+      specifically, RP1.
+      Mutually exclusive with fifo-qos-map.
```
6by9, line 117: Ditto

drivers/pci/controller/pcie-brcmstb.c
```diff
@@ -1320,10 +1320,11 @@ static int brcm_pcie_setup(struct brcm_pcie *pcie)
 		pcie->msi_target_addr = BRCM_MSI_TARGET_ADDR_GT_4GB;
 
 
-	/* Don't advertise L0s capability if 'aspm-no-l0s' */
-	aspm_support = PCIE_LINK_STATE_L1;
+	/* Always advertise L1 capability */
+	aspm_support = BIT(1);
```
6by9: Nicer to define the bits up around the define of register `PCIE_RC_CFG_PRIV1_LINK_CAPABILITY`, and then use them here.

drivers/pci/controller/pcie-brcmstb.c
```diff
@@ -1473,13 +1475,32 @@ static int brcm_pcie_start_link(struct brcm_pcie *pcie)
...
+		 * Increase Tperst_clk time by forcing PERST# output low while
+		 * the internal reset is released, so the PLL generates stable
+		 * refclk output further in advance of PERST# deassertion.
+		 */
+		tmp = readl(pcie->base + HARD_DEBUG(pcie));
+		u32p_replace_bits(&tmp, 1, PCIE_MISC_HARD_PCIE_HARD_DEBUG_PERST_ASSERT_MASK);
+		writel(tmp, pcie->base + HARD_DEBUG(pcie));
+
+		ret = pcie->cfg->perst_set(pcie, 0);
+		msleep(pcie->tperst_clk_ms);
```
6by9: What is the typical value for your sleep?
There were issues for short sleeps with msleep (<30ms), but I believe those are now resolved.

`fsleep(usecs)` is the recommended call - https://elixir.bootlin.com/linux/v6.13.3/source/Documentation/timers/delay_sleep_functions.rst (was Documentation/timers/timers-howto.rst)

6by9 (commented):
Seeing as much of this should be destined for upstream, I'm being picky.

And welcome to the lovely world of dt-bindings.

drivers/pci/controller/pcie-brcmstb.c
P33M: The worst-case EP delay (also the reason for implementing it) is 100ms and that is a minimum. I was unaware of fsleep.

drivers/pci/controller/pcie-brcmstb.c
P33M: Oops - chunk got intermingled during rebase.

Documentation/devicetree/bindings/pci/brcm,stb-pcie.yaml
P33M: This is where it gets fun: downstream uses nibbles. Would the upstream preference be a u32 array? Bit wasteful to encode 4 bits in 32.

Also, the useful combinations of these properties are either nothing, or "brcm,fifo-qos-map" or "brcm,vdm-qos-map". Can optional properties be specified as mutually exclusive?

6by9: I would have expected upstream to want an array of u8 or u32, but the only way of finding out would be to ask them.
Wasting <1kB isn't really worth worrying about.

You can convert the DT config directly back into the relevant nibbles for the hardware during probe, so it only throws a small amount extra on the stack during probe.

Yes I believe you can encode mutually exclusion on optional properties. Don't ask me for the exact runes at the moment though.

Documentation/devicetree/bindings/pci/brcm,stb-pcie.yaml
P33M: Strangely, the bindings checker complains if I define this. But not for uint8 (?)

6by9:
Minor heads up that there's another pcie-brcmstb patch set just queued up for merging
https://lore.kernel.org/linux-arm-kernel/20250214173944.47506-1-james.quinlan@broadcom.com/

P33M:
https://lore.kernel.org/linux-arm-kernel/20250214173944.47506-2-james.quinlan@broadcom.com/ obsoletes commit 5ae7f04df6cb5a5223a0d670f3ce7f20b3e43b26 but that's a problem for future us.

6by9:
> https://lore.kernel.org/linux-arm-kernel/20250214173944.47506-2-james.quinlan@broadcom.com/ obsoletes commit [5ae7f04](https://github.com/raspberrypi/linux/commit/5ae7f04df6cb5a5223a0d670f3ce7f20b3e43b26) but that's a problem for future us.

Ish.
These appear to have been accepted, so will be in 6.15/6.16. Whilst we're working on this already, it'd be better to backport those to 6.14 first and fix the problems once, rather than having to revisit it in 3 months time.

It also means that the your patches should apply upstream for when we try submitting them.

(Quickest way to apply the series to the current branch is `b4 shazam 20250214173944.47506-1-james.quinlan@broadcom.com/` - see [b4 docs](https://b4.docs.kernel.org/en/latest/contributor/prep.html#working-with-series-dependencies))

P33M:
This now 
- reverts v1/v5 of the upstream patch set
- applies v2/v6 of the same
- puts downstream fixes on top of these

The bindings pass dt_binding_check (at least, I think so - lots of others fail).

Documentation/devicetree/bindings/pci/brcm,stb-pcie.yaml
P33M: I went with an array of u8 - breaks DT compatibility across versions, but them's the breaks.

popcornmix:
This booted for me, with working ssd/m2hat and ethernet.

popcornmix (approved):

pelwell:
@6by9 Is there any reason I shouldn't hit the Merge button?

6by9:
> @6by9 Is there any reason I shouldn't hit the Merge button?

I don't think there's a good reason not to.

Picking up the newer versions of the upstream patches is fine, and P33M is the expert on what the downstream patches do.
DT bindings will always be a pain when it comes to upstreaming them, but they look reasonable to me.

6by9 (approved):

pelwell:
Thanks!

PR #6678, opened 2025-02-20
Overlays: Adding generic overlay for Cytron standard HAT+ boards that also supply power to the Raspberry Pi.

waiweng83:
We are creating HAT+ boards which will be supplying the power (5A capable) to the Raspberry Pi 5.
These boards are using some of the GPIOs and are not stackable, thus they don't belong to the mode 1 power class.

Instead, we configured them as standard class and we need to use the overlay to enable USB full current support.
All the boards will be using this common overlay and thus the generic name.


pelwell:
Setting the `usb_max_current_enable` property in your overlay has no effect except to add the property to the DTB - the firmware doesn't read it because it sees it too late. The only way to indicate to the firmware that a full 5A is available is to use an EEPROM with an address of 0x53 - see the specification: https://datasheets.raspberrypi.com/hat/hat_plus_specification.pdf

You may also benefit from reading this thread: https://github.com/raspberrypi/linux/pull/6257
If you are supplying 

I'm also not happy about the idea of multiple instances of an overlay that simply sets hat_power_supply to 5000 (which, to be clear, also doesn't alter the firmware behaviour in a significant way). I think there needs to be a truly generic HAT+ power overlay that anyone can use, if that's all their device needs.

waiweng83:
> Setting the `usb_max_current_enable` property in your overlay has no effect except to add the property to the DTB - the firmware doesn't read it because it sees it too late. The only way to indicate to the firmware that a full 5A is available is to use an EEPROM with an address of 0x53 - see the specification: https://datasheets.raspberrypi.com/hat/hat_plus_specification.pdf
> 
> You may also benefit from reading this thread: #6257 If you are supplying
> 
> I'm also not happy about the idea of multiple instances of an overlay that simply sets hat_power_supply to 5000 (which, to be clear, also doesn't alter the firmware behaviour in a significant way). I think there needs to be a truly generic HAT+ power overlay that anyone can use, if that's all their device needs.

The HAT+ specification mentioned that the Power Hat+ with EEPROM address 0x53 should not use the GPIO 2-27.

image

In my case, the HAT+ will be using the GPIOs and at the same time supplying power to the Raspberry Pi. If I'm still configuring the EEPROM address to be 0x53, will it cause my board to be not compliance with the HAT+ specification and I can't market it as a HAT+?

pelwell:
There does seem to be at least a lack of clarity in the spec. I've raised the question internally.

waiweng83:
> There does seem to be at least a lack of clarity in the spec. I've raised the question internally.

Thanks. So using EEPROM address 0x53 will be the correct way to do it? I will go this way then.

pelwell:
Yes, do that - what you are doing is reasonable, and I can't think of a better way to support that other than the 0x53 address. Just don't commit a lot of time or money to it until you have confirmation here, which I hope won't be long in coming.

waiweng83:
> Yes, do that - what you are doing is reasonable, and I can't think of a better way to support that other than the 0x53 address. Just don't commit a lot of time or money to it until you have confirmation here, which I hope won't be long in coming.

Alright. Will do.
Thanks for the help and clarification.

pelwell:
After some discussion here we've decided the following:
1. There is a hole in the specification - it doesn't address the concept of non-stackable HATs (those that use the GPIOs other than 0 & 1) that also supply power.
2. We will add an EEPROM property of some description (TBD) to allow Standard (0x50) EEPROMs to declare the power that they supply. This will be the preferred way to implement non-stackable HATs that supply power.
3. We will relax the requirements on power HATs that they do not interact with the other GPIOs, with the proviso that any product description must clearly state that the result is not stackable.

In other words, go ahead with using the 0x53 address for this product. For future products, there will be another route to the same outcome, the details of which are to be confirmed.

waiweng83:
> After some discussion here we've decided the following:
> 
> 1. There is a hole in the specification - it doesn't address the concept of non-stackable HATs (those that use the GPIOs other than 0 & 1) that also supply power.
> 2. We will add an EEPROM property of some description (TBD) to allow Standard (0x50) EEPROMs to declare the power that they supply. This will be the preferred way to implement non-stackable HATs that supply power.
> 3. We will relax the requirements on power HATs that they do not interact with the other GPIOs, with the proviso that any product description must clearly state that the result is not stackable.
> 
> In other words, go ahead with using the 0x53 address for this product. For future products, there will be another route to the same outcome, the details of which are to be confirmed.

Alright. Noted on this.
Thanks a lot.

waiweng83:
> After some discussion here we've decided the following:
> 
> 1. There is a hole in the specification - it doesn't address the concept of non-stackable HATs (those that use the GPIOs other than 0 & 1) that also supply power.
> 2. We will add an EEPROM property of some description (TBD) to allow Standard (0x50) EEPROMs to declare the power that they supply. This will be the preferred way to implement non-stackable HATs that supply power.
> 3. We will relax the requirements on power HATs that they do not interact with the other GPIOs, with the proviso that any product description must clearly state that the result is not stackable.
> 
> In other words, go ahead with using the 0x53 address for this product. For future products, there will be another route to the same outcome, the details of which are to be confirmed.

I redesigned the board with EEPROM Address 0x53 and the Pi 5 still showing up the message "This power supply is not capable of supplying 5A..."

Seems like the maximum current is still limited to 3A only.
image

May I know how should I increase it to 5A? Should I load any dtoverlay?

waiweng83:
This is the content of my eeprom_settings.txt

```########################################################################
# EEPROM settings text file
#
# Edit this file for your particular board and run through eepmake tool,
# then use eepflash tool to write to attached HAT ID EEPROM
#
# Tools available:
#  eepmake   Parses EEPROM text file and creates binary .eep file
#  eepdump   Dumps a binary .eep file as human readable text (for debug)
#  eepflash  Write or read .eep binary image to/from HAT EEPROM
#
########################################################################

# 128 bit UUID. If left at zero eepmake tool will auto-generate
# RFC 4122 compliant UUID
product_uuid 00000000-0000-0000-0000-000000000000

# 16 bit product id
product_id 0x0000

# 16 bit product version
product_ver 0x0000

# ASCII vendor string  (max 255 characters)
vendor "Cytron Technologies Sdn Bhd"

# ASCII product string (max 255 characters)
product "Industry HAT+"

# How much current the HAT+ can supply, in milliamps
current_supply 5000

# Which Device Tree overlay to load
#dt_blob "acme-sensor"
```

pelwell:
I agree that the value of the `max_current` property is concerning/confusing, but `usb_max_current_enable` is 1, which is the only thing that the firmware actually cares about.

Set `BOOT_UART=1` in the EEPROM configuration and connect a serial cable to the Debug UART port. What messages are you getting from the firmware? I would expect to see:
```
MODE1 power HAT+ detected - selecting USB high current limit
```
In shipping firmware, the `current_supply` value is only being used in the case of a normal (non-stacked) HAT+, but I think that is unnecessarily restrictive. Also, the presence of MODE0 and MODE1 power HATs is not reflected in the value of `max_current`.

I propose to close the loop and fill the gaps as follows:
1. For each HAT, calculate a current supply value which is:
    * for a MODE0 power HAT, its `current_supply` property if set, otherwise 3000,
    * for a MODE1 power HAT, its `current_supply` property if set, otherwise 5000, and
    * for a non-power HAT, the value of its `current_supply` property,
    where an absent `current_supply` property is equivalent to 0.
2. `max_current` is calculated as the maximum of all the sources in the system, including any USB-C supply. In practice the available current may be higher than this, but probably not a simple summation, and multiple power supplies is not really recommended.
3. `usb_max_current_enable` is set whenever `max_current` is at least 5000 (or if running on a CM5)..
4. Any value of the `hat_current_supply` property set by an overlay is too late to influence `usb_max_current_enable`, but it will be included in `max_current`.

In other words, what you have in your EEPROM looks correct, and it should have the desired effect in all ways except for the value of the `max_current` property in DT.

pelwell:
A patch implementing the updated logic described above has been merged to our internal software repo. You can download a trial build from here: https://drive.google.com/file/d/1j56M1orl7WinaUVYEpOI_GB5AN-7d3Ot/view?usp=sharing

pelwell:
The patch mentioned above has now made its way into a master EEPROM release: https://github.com/raspberrypi/rpi-eeprom/pull/721

waiweng83:
@pelwell Thanks for your quick response. I'm still unable to test it out as I'm currently tied up with other projects. Will revert once there is any finding. Thanks.

waiweng83:
> The patch mentioned above has now made its way into a master EEPROM release: [raspberrypi/rpi-eeprom#721](https://github.com/raspberrypi/rpi-eeprom/pull/721)

I just tested the pieeprom-2025-07-03.bin and it works as expected. Thanks for the help @pelwell

PR #6679, opened 2025-02-20
configs: Enable more ZRAM backends

popcornmix:
6.12 kernel reworked how ZRAM backends were configured. Let's enable the ones we've lost.

I've chosen zstd as the default.

6.6 kernel supported:
$ cat /sys/block/zram0/comp_algorithm
[lzo-rle] lzo lz4 zstd

6.12 currently supports:
$ cat /sys/block/zram0/comp_algorithm
[lzo-rle] lzo

With this PR 6.12 supports:
$ cat /sys/block/zram0/comp_algorithm
lzo-rle lzo lz4 [zstd]

See: https://forums.raspberrypi.com/viewtopic.php?p=2296678#p2296678


popcornmix:
I'm happy to change the default and/or what's enabled if anyone feels strongly.

pelwell (approved):

PR #6684, opened 2025-02-24
Remove unnecessary call to gpiochip_remove

AccraNET:
No gpio_chip has been added (registered ) with the core at this point. Calling gpiochip_remove is therefore not necessary


pelwell:
I agree with your patch, but this is a problem that exists in the upstream kernel: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/drivers/pinctrl/bcm/pinctrl-bcm2835.c?h=v6.14-rc4#n1351

You should send it to the upstream maintainers:
```
PIN CONTROL SUBSYSTEM
M:	Linus Walleij <linus.walleij@linaro.org>
L:	linux-gpio@vger.kernel.org
S:	Maintained
T:	git git://git.kernel.org/pub/scm/linux/kernel/git/linusw/linux-pinctrl.git
F:	Documentation/devicetree/bindings/pinctrl/
F:	Documentation/driver-api/pin-control.rst
F:	drivers/pinctrl/
F:	include/dt-bindings/pinctrl/
F:	include/linux/pinctrl/
```

6by9:
It'll also want a `Signed-off-by: author <author_email@domain>` and 
```
Fixes: 266423e60ea1 ("pinctrl: bcm2835: Change init order for gpio hogs")
```
as it's been wrong since 5.16 so needs backporting to stable.
https://github.com/torvalds/linux/commit/266423e60ea1b953fcc0cd97f3dad85857e434d1

There is also a `devm_gpiochip_add_data` function. There may be interactions because the driver is also an interrupt controller, but I wonder if the device lifetime hooks can be used to make cleanup easier.

For curiosity, have you observed this error path being taken in reality, or just noticed it in code review?

PR #6687, opened 2025-02-26
iommu/dma: Add ability to configure NUMA allocation policy for remapped allocations

popcornmix:
Add iommu_dma_numa_policy= kernel parameter which can be used to modify the NUMA allocation policy of remapped buffer allocations.

Policy is only used for devices which are not associated with a NUMA node.

Syntax identical to what tmpfs accepts as it's mpol argument is accepted.

Some examples:

 iommu_dma_numa_policy=interleave
 iommu_dma_numa_policy=interleave=skip-interleave
 iommu_dma_numa_policy=bind:0-3,5,7,9-15
 iommu_dma_numa_policy=bind=static:1-2

### DIFF
--- a/drivers/iommu/dma-iommu.c
+++ b/drivers/iommu/dma-iommu.c
@@ -21,6 +21,7 @@
 #include <linux/iova.h>
 #include <linux/irq.h>
 #include <linux/list_sort.h>
+#include <linux/mempolicy.h>
 #include <linux/memremap.h>
 #include <linux/mm.h>
 #include <linux/mutex.h>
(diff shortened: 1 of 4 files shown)


popcornmix:
In reference to [this issue](https://forums.raspberrypi.com/viewtopic.php?p=2298671#p2298671).

To test update to this kernel (`sudo rpi-update pulls/6687`) once the github CI has completed and add to cmdline.txt (on existing line) `iommu_dma_numa_policy=interleave`.

popcornmix:
This has been confirmed to fix the issue reported so removing draft.

I've also added a commit that allows random numa allocation which can be useful for testing.
I dropped it for 6.6 where as it needed a number of additional commits to work, but it's fairly clean on 6.12.

I believe both commits have no effect without changes to cmdline.txt.

pelwell (approved):

PR #6688, opened 2025-02-26
configs: Delete bcmrpi3_defconfig

pelwell:
Acknowledge the fact that bcmrpi3_defconfig is neither used nor supported by us, and avoid a bunch of future merge conflicts since it is already gone from rpi-6.14.y.


popcornmix (approved):

floion:
Hi @pelwell @popcornmix so what is the defconfig to be used for RPi3 in 64 bits mode?

floion:
Is it bcm2711_defconfig as per https://github.com/raspberrypi/linux/pull/5618#issue-1917603822?

popcornmix:
Yes, bcm2711_defconfig supports all 64-bit Pi's.

floion:
Thank you for the quick reply.

PR #6691, opened 2025-02-27
drivers: staging: fbtft: Fix fb_ssd1306 driver for 128x32 Display

pawel-soja:
The driver incorrectly sets the column and page addresses for a 128x32 display.
Display used is from the [Environmental Sensor Board](https://coral.ai/products/environmental/) kit.
The final values ​​of the addresses should be:
- 0x7F for cols because it is `128 - 1`
- 0x03 for rows because it is `32 / 8 (bits, vertical addressing) - 1`

Before:
image

After:
image

### DIFF
--- a/drivers/staging/fbtft/fb_ssd1306.c
+++ b/drivers/staging/fbtft/fb_ssd1306.c
@@ -133,7 +133,6 @@ static void set_addr_win_64x48(struct fbtft_par *par)
 	write_reg(par, 0x5);
 }
 
-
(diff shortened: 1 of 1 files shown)


6by9:
Slight dilemma here in that fbtft is orphaned and considered deprecated by upstream. DRM drivers (mainly mipi-dbi-spi) should be used instead.

However we do have 2 downstream patches to fbtft already, so adding this isn't the end of the world.

PR #6692, opened 2025-02-28
drm/v3d: Fix GPU reset issues on the Raspberry Pi 5

### LINKED ISSUE #6660: Dispatching a long-running compute shader causes system hang or abnormal behavior
### Describe the bug

When we dispatch a shader program using ioctl(SUBMIT_CSD) on a Raspberry Pi 5, if the shader program’s execution time exceeds 500 ms, ioctl(WAIT_BO) returns "Timer expired" or the system hangs.

Once "Timer expired" occurs, even subsequent shader programs that should complete within 500 ms also result in "Timer expired."

When the system hangs, I can’t do anything. Pressing the power button has no effects, and the LED stays green (on).

I suspect [this line](https://github.com/raspberrypi/linux/blob/fab0093d88452972f05f4b13e91e31d00b55421a/drivers/gpu/drm/v3d/v3d_sched.c#L510). Is there any difficulty in relaxing this limit? I think it is too tight for GPGPU.

[This is an example program](https://gist.github.com/notogawa/4dcebe6db14f5898dee85babb85f7d37#file-main-c) to reproduce. In this example, a shader is a busy nop loop.

```console
$ git clone https://gist.github.com/notogawa/4dcebe6db14f5898dee85babb85f7d37
$ cd 4dcebe6db14f5898dee85babb85f7d37
$ gcc -o main main.c
$ ./main N (N is nop-loop count)
```

Case 1: Normal
```console
$ ./main 1000000
[loop:1000000]
0.008614 sec
$ ./main 1000000
[loop:1000000]
0.008624 sec
$ ./main 64000000
[loop:64000000]
0.271148 sec
```

Case 2: Timer expired
```
$ ./main 128000000
[loop:128000000]
wait_bo: Timer expired <- display after 10sec
$ ./main 1000000
[loop:1000000]
wait_bo: Timer expired
```

Case 3: System hang
```
$ ./main 128000000
[loop:128000000]
(hang.)
```

This example is a minimal reproducible program, so it’s just a no-op loop. In reality, however, we’re submitting programs like massive matrix–matrix multiplications.

### Device (s)

Raspberry Pi 5

### System

```console
$ cat /etc/rpi-issue
Raspberry Pi reference 2024-11-19
Generated using pi-gen, https://github.com/RPi-Distro/pi-gen, 891df1e21ed2b6099a2e6a13e26c91dea44b34d4, stage2
$ vcgencmd version
2024/09/23 14:02:56
Copyright (c) 2012 Broadcom
version 26826259 (release) (embedded)
$ uname -a
Linux pi5 6.6.62+rpt-rpi-2712 #1 SMP PREEMPT Debian 1:6.6.62-1+rpt1 (2024-11-25) aarch64 GNU/Linux
```

### Logs

_No response_

### Additional context

_No response_

mairacanal:
This series addresses GPU reset issues reported in #6660, where running a long compute job would trigger repeated GPU resets, leading to a UI freeze.

Patch 1 is the backport of an upstream commit that adds proper clock handling to V3D.

Patches 2 to 6 are the fixes, which are currently under review in the mailing list [1].

Patch 7 adds the new register to the BCM2712 DT.

[1] https://lore.kernel.org/dri-devel/20250226-v3d-gpu-reset-fixes-v1-0-83a969fdd9c1@igalia.com/T/

Closes: #6660


drivers/gpu/drm/v3d/v3d_drv.c
```diff
@@ -198,6 +198,36 @@ static const struct of_device_id v3d_of_match[] = {
 };
 MODULE_DEVICE_TABLE(of, v3d_of_match);
 
+static void
+v3d_idle_sms(struct v3d_dev *v3d)
+{
+	if (v3d->ver < 71)
```
pelwell: Shouldn't this (and others) be `V3D_GEN_71`?

popcornmix:
I can confirm this PR resolves the issue of GPU resets being unrecoverable (i.e. the test code is linked issue recovers after a timeout).

popcornmix (approved):

pelwell:
Replace the numbers in the version checks with the symbols you've defined and I'll merge this.

drivers/gpu/drm/v3d/v3d_drv.c
mairacanal: It is not needed. In the end, `V3D_GEN_71 = 71` by the enum definition. But I can change it if you believe it will ease the code comprehension.

pelwell: Use them everywhere or nowhere - just be consistent.

mairacanal: Done, thanks for the feedback, I'll apply it in the upstream series as well.

drivers/gpu/drm/v3d/v3d_debugfs.c
```diff
@@ -197,11 +197,11 @@ static int v3d_v3d_debugfs_ident(struct seq_file *m, void *unused)
 		seq_printf(m, "  QPUs:         %d\n", nslc * qups);
 		seq_printf(m, "  Semaphores:   %d\n",
 			   V3D_GET_FIELD(ident1, V3D_IDENT1_NSEM));
-		if (v3d->ver <= 42) {
+		if (v3d->ver <= V3D_GEN_42) {
 			seq_printf(m, "  BCG int:      %d\n",
 				   (ident2 & V3D_IDENT2_BCG_INT) != 0);
 		}
-		if (v3d->ver < 40) {
+		if (v3d->ver < V3D_GEN_41) {
```
pelwell, line 204: Is this correct? We are still setting v3d->ver based on register values, and if V3D_GEN_41 == 41 and there is hardware with a version of 40 then this test has changed.

drivers/gpu/drm/v3d/v3d_drv.c
```diff
@@ -232,17 +267,40 @@ static int v3d_platform_drm_probe(struct platform_device *pdev)
...
 	mask = DMA_BIT_MASK(30 + V3D_GET_FIELD(mmu_debug, V3D_MMU_PA_WIDTH));
 	ret = dma_set_mask_and_coherent(dev, mask);
 	if (ret)
-		return ret;
+		goto clk_disable;
 
 	v3d->va_width = 30 + V3D_GET_FIELD(mmu_debug, V3D_MMU_VA_WIDTH);
 
 	ident1 = V3D_READ(V3D_HUB_IDENT1);
 	v3d->ver = (V3D_GET_FIELD(ident1, V3D_HUB_IDENT1_TVER) * 10 +
```
pelwell, line 297: It looks strange to change this critical value, even with the WARN. Would you consider using `gen` before here, and `v3d->ver` after?

drivers/gpu/drm/v3d/v3d_debugfs.c
mairacanal: It is correct. V3D 4.0 doesn't exist and hypothetically, if it did, the kernel doesn't support it. The kernel only supports 33, 41, 42, and 71.

drivers/gpu/drm/v3d/v3d_drv.c
mairacanal: How do you feel about this approach?

```c
v3d->ver = (enum v3d_gen)of_device_get_match_data(dev);

[...]

gen = (V3D_GET_FIELD(ident1, V3D_HUB_IDENT1_TVER) * 10 + V3D_GET_FIELD(ident1, V3D_HUB_IDENT1_REV));
WARN_ON(v3d->ver != gen);
```

This way we wouldn't change the value, but still check that the hardware returns the same value.

mairacanal: But thinking more about it, I think that we should trust the value returned by the hardware and not the value that a developer added to the Device ID table. The warning has the intention to avoid incorrect values in the Device ID table by letting the developer know that something is odd.

pelwell: I think the hardware is definitive, and the DT setting is an early hint, which is why I suggested doing it the other way round:
```c
	gen = (enum v3d_gen)of_device_get_match_data(dev);

[...]

	if (gen >= V3D_GEN_71) {
		ret = map_regs(v3d, &v3d->sms_regs, "sms");
		if (ret)
			return ret;
	}

[...]

	v3d->ver = (V3D_GET_FIELD(ident1, V3D_HUB_IDENT1_TVER) * 10 + V3D_GET_FIELD(ident1, V3D_HUB_IDENT1_REV));
	WARN_ON(v3d->ver != gen);
```

drivers/gpu/drm/v3d/v3d_debugfs.c
pelwell: I'll take your word for it.

drivers/gpu/drm/v3d/v3d_drv.c
mairacanal: The thing is: between `map_regs` and `V3D_GET_FIELD`, we have `v3d_sms_idle()`, which uses `v3d->ver`

 I'm sorry, but I don't understand what the issue is. First, we assign `v3d->ver = gen` and after we power up the GPU, we overwrite the variable with the hardware value.

In the end, `v3d->ver` will be the one from the hardware, even if `v3d->ver != gen`.

pelwell: Nothing is wrong - just trying to improve readability

naheedsa:
Hi @popcornmix , @mairacanal 
We are now using kernel version 6.18, and facing the same issue. Can you guys double-check if these changes are included in the Kernel version 6.18 or later?

`Apr 30 10:11:02 observatory kernel: v3d fec00000.v3d: [drm:v3d_reset [v3d]] *ERROR* Resetting GPU for hang.
Apr 30 10:11:02 observatory kernel: v3d fec00000.v3d: [drm:v3d_reset [v3d]] *ERROR* V3D_ERR_STAT: 0x00001000
Apr 30 10:14:00 observatory kernel: Workqueue: events drm_sched_job_timedout [gpu_sched]
Apr 30 10:14:00 observatory kernel:  drm_sched_stop+0x13c/0x218 [gpu_sched]
Apr 30 10:14:00 observatory kernel:  drm_sched_job_timedout+0x78/0x178 [gpu_sched]
Apr 30 10:16:01 observatory kernel:  drm_ioctl_kernel+0xc8/0x138 [drm]
Apr 30 10:16:01 observatory kernel:  drm_ioctl+0x35c/0x4d0 [drm]`

popcornmix:
Yes, the commits here are present:
```
dom@buildbot:~/projects/linux/rpi-6.18.y$ git log --oneline drivers/gpu/drm/v3d/ | egrep "Add V3D_SMS|tech|fence has an|flagged in it|Add clock"
76dbd0973c555 drm/v3d: Associate a V3D tech revision to all supported devices
c3e4a25602f8b drm/v3d: Set job pointer to NULL when the job's fence has an error
80cbee810e4e1 drm/v3d: Don't run jobs that have errors flagged in its fence
4dd40b5f9c3d8 drm/v3d: Add clock handling
7c78fdbace0ff drm/v3d: Add V3D tech revision to the device information
```
Probably best to create a new issue, and describe steps to reproduce your error.

mairacanal:
@naheedsa, yes,  please, create a new issue describing how to (consistently) reproduce your error and I will debug it as soon as possible.

PR #6694, opened 2025-02-28
RP1 DPI: interlaced HSync/VSync alignment (ignore)

njhollinghurst:
Instead of trying to minimize the HSync -> VSync lag, count cycles to predict the _next_ HSync.

It reduces the HSync->VSync offset from about (+30ns .. +35ns), to about (-5ns .. +10ns). There is some mode-dependent rounding error, but for a given video mode the jitter is ~5ns as before.

This is mainly for the benefit of SCART Hats, particularly those that use an XOR gate to generate Composite Sync, which otherwise could produce glitchy output when used in an interlaced mode with Raspberry Pi 5.

Net shortening of the DPI PIO code, leaves all of 5 instructions now free.


njhollinghurst:
Whoops wrong target branch

PR #6697, opened 2025-03-03
Revert "firmware: raspberrypi: Add backward compatible get_throttled"

popcornmix:
This reverts commit 0273165d1a72be7ed5ae90c3ab02c6ad267a25b2.

### DIFF
--- a/drivers/firmware/raspberrypi.c
+++ b/drivers/firmware/raspberrypi.c
@@ -33,7 +33,6 @@ struct rpi_firmware {
 	u32 enabled;
 
 	struct kref consumers;
-	u32 get_throttled;
 };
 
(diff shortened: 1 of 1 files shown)


popcornmix:
See: https://github.com/raspberrypi/firmware/issues/1950#issuecomment-2694537745

6by9:
The `WARN_ONCE` does seem to be a little excessive as you get the full backtrace splat, but a `dev_err_once` would seem reasonable.

popcornmix:
> The WARN_ONCE does seem to be a little excessive as you get the full backtrace splat, but a dev_err_once would seem reasonable.

We have been WARNing its deprecated since the 4.19 kernel. I think users have had long enough to switch to the correct interface.

Issue #6698, opened 2025-03-03
Can't compile for rpi5 using clang


KubaTaba1uga:
### Describe the bug

I tried to compile 6.6.x kernel using clang but compilation failed at some point:
```
drivers/dma-buf/heaps/system_heap.c:57:40: error: initializer element is not a compile-time constant
static unsigned int module_max_order = orders[0];
                                       ^~~~~~~~~
1 error generated.
make[5]: *** [scripts/Makefile.build:243: drivers/dma-buf/heaps/system_heap.o] Error 1
make[5]: *** Waiting for unfinished jobs....
  CC [M]  net/ipv6/ah6.o
make[4]: *** [scripts/Makefile.build:480: drivers/dma-buf/heaps] Error 2
make[3]: *** [scripts/Makefile.build:480: drivers/dma-buf] Error 2
make[2]: *** [scripts/Makefile.build:480: drivers] Error 2
```

### Steps to reproduce the behaviour

Get fresh 6.6.x kernel and then do:
```
export CC=clang LLVM=1 CROSS_COMPILE=aarch64-linux-gnu- KERNEL=kernel2712 ARCH=arm64

make bcm2712_defconfig

make -j$(nproc)
```

Watch how compilation fails at some point.

### Device (s)

Raspberry Pi 5

### System

I tried to cross compile the kernel on Debian 12 bookworm on amd64 for Raspberry pi5.

### Logs

[kernel_build.log](https://github.com/user-attachments/files/19055895/kernel_build.log)

KubaTaba1uga:
I had idea for the fix so i created this PR: https://github.com/raspberrypi/linux/pull/6699.
If it's not good i would like to work on improving the fix.

pelwell:
This should be resolved now #6699 has been merged.

PR #6699, opened 2025-03-03
Fix clang compilation error for rpi5

KubaTaba1uga:
I created an issue describing the bug: https://github.com/raspberrypi/linux/issues/6698.

To quickly summarize, i tried to compile 6.6.y kernel with clang and compilation failed on ` drivers/dma-buf/heaps/system_heap.c` with error:
```
drivers/dma-buf/heaps/system_heap.c:57:40: error: initializer element is not a compile-time constant
static unsigned int module_max_order = orders[0];
                                       ^~~~~~~~~
```

This PR is fixing the compilation error by using macro to specify compile-time constant properly.

### DIFF
--- a/drivers/dma-buf/heaps/system_heap.c
+++ b/drivers/dma-buf/heaps/system_heap.c
@@ -50,11 +50,16 @@ static gfp_t order_flags[] = {HIGH_ORDER_GFP, HIGH_ORDER_GFP, LOW_ORDER_GFP};
  * to match with the sizes often found in IOMMUs. Using order 4 pages instead
  * of order 0 pages can significantly improve the performance of many IOMMUs
  * by reducing TLB pressure and time spent updating page tables.
+ *
+ * Note: `module_max_order` must be set explicitly instead of using
+ * `orders[0]` to avoid Clang's "initializer element is not a
+ * compile-time constant" error.
  */
-static const unsigned int orders[] = {8, 4, 0};
+#define MAX_ORDERS_VALUE 8
+static const unsigned int orders[] = {MAX_ORDERS_VALUE, 4, 0};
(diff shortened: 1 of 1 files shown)


6by9:
Introduced by downstream commit https://github.com/raspberrypi/linux/commit/56fda5adce0c9b4dc8ff2c7be9a704599e685001

```
static const unsigned int orders[] = {8, 4, 0};
```
so `orders[0]` is a compile-time constant. The fact that other compilers aren't throwing an error on it implies it is an LLVM/Clang error.
Which versions of LLVM and Clang are you using?

KubaTaba1uga:
I'm using `Debian clang version 14.0.6`, but additionaly i tried `Debian clang version 15.0.6` with the same results.

popcornmix:
I think this is the relevant part of C standard:

> All the expressions in an initializer for an object that has static or thread storage duration shall be constant expressions or string literals.

which I think the original code technically violates. Although with a modest amount of effort a compiler could reduce the initialiser to a constant (but it doesn't have to).

I can confirm that compiling kernel with clang does fail without this PR and succeeds with it.

For interest I tried the 6.14 branch and that fails for additional reasons (in upstream file `arch/arm64/kernel/signal.c`), so clang support may need to be considered somewhat experimental. [edit it does actually build from clean if clang options are also passed at bcm2711_defconfig stage].

But I'm okay with this PR (personally I'd drop the comment).

pelwell:
I was wondering if you could make the whole array a macro:
```
#define ORDERS { 8, 4, 0 }
```
But that fails spectacularly.

It's a shame this patch is necessary, but it does appear to help old versions of GCC as well.

Do lose the comment, though.

KubaTaba1uga:
The comment is gone.

Issue #6701, opened 2025-03-05
Null pointer dereference on concurrent VC_SM_CMA_IOCTL_MEM_IMPORT_DMABUF ioctl


dividuum:
### Describe the bug

I've observed kernel null pointer dereferences while using the VC_SM_CMA_IOCTL_MEM_IMPORT_DMABUF ioctl. A traceback might look like this:

```
[  634.185377] ---[ end trace 0000000000000000 ]---
[  634.185400] hwmon hwmon1: Failed to get throttled (-110)
[  645.781772] Unable to handle kernel NULL pointer dereference at virtual address 0000000000000008
[  645.781801] Mem abort info:
[  645.781806]   ESR = 0x0000000096000046
[  645.781812]   EC = 0x25: DABT (current EL), IL = 32 bits
[  645.781819]   SET = 0, FnV = 0
[  645.781826]   EA = 0, S1PTW = 0
[  645.781831]   FSC = 0x06: level 2 translation fault
[  645.781840] Data abort info:
[  645.781844]   ISV = 0, ISS = 0x00000046, ISS2 = 0x00000000
[  645.781850]   CM = 0, WnR = 1, TnD = 0, TagAccess = 0
[  645.781856]   GCS = 0, Overlay = 0, DirtyBit = 0, Xs = 0
[  645.781866] user pgtable: 4k pages, 39-bit VAs, pgdp=00000000484d5000
[  645.781873] [0000000000000008] pgd=080000004c6eb003, p4d=080000004c6eb003, pud=080000004c6eb003, pmd=0000000000000000
[  645.781896] Internal error: Oops: 0000000096000046 [#1] PREEMPT SMP
[...]

PR #6703, opened 2025-03-05
VCSM locking fix

6by9:
Potential fix for #6701


6by9:
Nope. Ran far longer (just over 100k iterations), but still failed.

PR #6705, opened 2025-03-06
drivers: media: cfe: Add non-continuous CSI-2 clock mode

naushir:
Set the DPHY block to manual clock lane termination enable and HS mode i V4L2_MBUS_CSI2_NONCONTINUOUS_CLOCK is set in the bus flags.

### DIFF
--- a/drivers/media/i2c/imx500.c
+++ b/drivers/media/i2c/imx500.c
@@ -37,6 +37,8 @@
 #define IMX500_IMAGE_ONLY_FALSE 0x00
 #define IMX500_IMAGE_ONLY_TRUE 0x01
 
+#define IMX500_REG_SENSOR_TEMP_CTRL CCI_REG8(0x0138)
+
 #define IMX500_REG_ORIENTATION CCI_REG8(0x101)
 
 #define IMX500_XCLK_FREQ 24000000
@@ -2505,6 +2507,14 @@ static int imx500_start_streaming(struct imx500 *imx500)
 	if (ret < 0)
 		return ret;
 
+	/*
+	 * Disable the temperature sensor here - must be done else loading any
+	 * firmware fails...
+	 *
+	 * Re-enable before stream-on below.
+	 */
+	cci_write(imx500->regmap, IMX500_REG_SENSOR_TEMP_CTRL, 0, &ret);
+
 	ret = cci_write(imx500->regmap, IMX500_REG_IMAGE_ONLY_MODE,
 			imx500->fw_network ? IMX500_IMAGE_ONLY_FALSE :
 					     IMX500_IMAGE_ONLY_TRUE,
@@ -2598,6 +2608,9 @@ static int imx500_start_streaming(struct imx500 *imx500)
 
 	/* Disable any sensor startup frame drops. This must be written here! */
 	cci_write(imx500->regmap, CCI_REG8(0xD405), 0, &ret);
+
+	/* Re-enable the temperature sensor. */
+	cci_write(imx500->regmap, IMX500_REG_SENSOR_TEMP_CTRL, 1, &ret);
 
 	/* set stream on register */
 	cci_write(imx500->regmap, IMX500_REG_MODE_SELECT, IMX500_MODE_STREAMING,


naushir:
@6by9 very speculative change that sets up the DPHY for non-continuous clock.

drivers/media/platform/raspberrypi/rp1_cfe/cfe.c
```diff
@@ -1166,6 +1166,9 @@ static int cfe_start_streaming(struct vb2_queue *vq, unsigned int count)
 		goto err_disable_cfe;
 	}
 
+	cfe->csi2.dphy.noncontinuous_clock =
+		cfe->csi2.bus_flags & V4L2_MBUS_CSI2_NONCONTINUOUS_CLOCK;
+
```
6by9: Note that clocking mode can also be signaled via the `get_mbus_config` call that is handled in `cfe_start_streaming`.
That's used if you have something like a CSI2 mux with different inputs using different clocking modes (or number of lanes, which is already handled).

6by9 (commented):
Thanks - I'll give it a test with OV9281.

drivers/media/platform/raspberrypi/rp1_cfe/cfe.c
naushir: Fixed!

drivers/media/platform/raspberrypi/rp1_cfe/dphy.c
```diff
@@ -128,6 +129,19 @@ static void dphy_set_hsfreqrange(struct dphy_data *dphy, uint32_t mbps)
...
 
+static void dphy_set_manual_clock_term(struct dphy_data *dphy)
+{
+	/* See page 162 of the dphy databook */
+	const u8 tst = (1 << 0) + /* HS RX power on */
+		       (1 << 1) + /* Bypass HS RX power on */
+		       (1 << 2) + /* HS RX cal enable */
+		       (0 << 3) + /* HS RX cal control */
+		       (1 << 4) + /* Term EN */
+		       (1 << 5) + /* Term EN control */
```
6by9: Typo here. You want the line ending with `;`, not `+`
It's why the kernel build has failed.

naushir: Hmmm, why did this pass my compile test, or did it??

6by9: Warnings are errors is only enabled for CI builds, not by default.

I saw the CI error and looked more closely. My build had the warning, and it took a couple of seconds to twig as to why. The compiler was absolutely correct that tst was used uninitialised.

pelwell:
I would enable warnings-as-errors by default, but some don't like it. It's enabled in the auto-builds after the choice of defconfig by this step:
```
scripts/config --file $ROOT/.config --set-val CONFIG_WERROR y
```
where `$ROOT` may be `.` in your case.

naushir:
This change will need adjusting so that we can switch between auto/manual termination since it's possible this is mode specific.

drivers/media/platform/raspberrypi/rp1_cfe/cfe.c
6by9: Is it fixed?

naushir: Should be:

```c
	cfe->csi2.dphy.noncontinuous_clock = mbus_config.bus.mipi_csi2.flags &
					     V4L2_MBUS_CSI2_NONCONTINUOUS_CLOCK;
```

6by9: Ah, the lines looked so similar I hadn't noticed the change.

It needs to handle both DT and get_mbus_config.
Use `cfe->csi2.bus_flags & V4L2_MBUS_CSI2_NONCONTINUOUS_CLOCK;` if get_mbus_config returns non-zero

Use `mbus_config.bus.mipi_csi2.flags & V4L2_MBUS_CSI2_NONCONTINUOUS_CLOCK;` if get_mbus_config returns 0.

Just noticing the code at 1159-1161 which is effectively doing the same.
`mbus_config.bus.mipi_csi2.num_data_lanes;` is only valid if get_mbus_config returned 0, otherwise it should take `cfe->csi2.dphy.max_lanes`. The `if` is relying on the fact that mbus_config was initialised to 0.

naushir: ack

drivers/media/platform/raspberrypi/rp1_cfe/cfe.c
```diff
@@ -1156,6 +1156,10 @@ static int cfe_start_streaming(struct vb2_queue *vq, unsigned int count)
 		goto err_pm_put;
 	}
 
+	cfe->csi2.dphy.noncontinuous_clock =
+		(!ret ? mbus_config.bus.mipi_csi2.flags : cfe->csi2.bus_flags) &
+					V4L2_MBUS_CSI2_NONCONTINUOUS_CLOCK;
+
```
6by9: That'll do.

I'd have fixed up the `active_lanes` value too with
```
if (!ret) {
   cfe->csi2.dphy.noncontinuous_clock = mbus_config.bus.mipi_csi2.flags & V4L2_MBUS_CSI2_NONCONTINUOUS_CLOCK;
   cfe->csi2.dphy.active_lanes = mbus_config.bus.mipi_csi2.num_data_lanes;
	if (cfe->csi2.dphy.active_lanes > cfe->csi2.dphy.max_lanes) {
		cfe_err("Device has requested %u data lanes, which is >%u configured in DT\n",
			cfe->csi2.dphy.active_lanes, cfe->csi2.dphy.max_lanes);
		ret = -EINVAL;
		goto err_disable_cfe;
}} else {
  cfe->csi2.dphy.noncontinuous_clock = cfe->csi2.bus_flags & V4L2_MBUS_CSI2_NONCONTINUOUS_CLOCK;
   cfe->csi2.dphy.active_lanes = cfe->csi2.dphy.max_lanes;
}
```
but that can wait for another day.

I still haven't managed to get the failure case to happen that we were discussing this for though....

naushir: > I still haven't managed to get the failure case to happen that we were discussing this for though....

Even with this commit reverted?

6by9: > > I still haven't managed to get the failure case to happen that we were discussing this for though....
> 
> Even with this commit reverted?

Reproduced now without this commit.

6.12 doesn't have #6700 to add the option for selecting clock mode. Adding that makes testing far easier.
Continuous clock mode fails. Non-continuous is OK. That's similar to Pi3 which won't stream in continuous clock mode.

With the latest version of this patch (guess who tried testing the version that only looked at the get_mbus_config result version) I get timeouts in continuous mode, and black images non-continuous.

So we look to be in the right sort of area, but not quite right. I'll give it a few more minutes poking.

naushir: It's quite likely the register configuration I've used is also incorrect.  The datasheet is *very* convoluted.

naushir:
After staring a bit more at the databook, I suspect that while the register configuration may be right, the way we (now) write multiple config register through the test interface is likely wrong.

PR #6715, opened 2025-03-12
drm/v3d: Fix GPU reset issues on the Raspberry Pi 5 for 6.14

popcornmix:
Should be comparable to https://github.com/raspberrypi/linux/pull/6692/ but I used [upstream submission](https://lore.kernel.org/dri-devel/20250311-v3d-gpu-reset-fixes-v3-0-64f7a4247ec0@igalia.com/) as the source as it's closer to our source tree.


popcornmix:
Tested with https://github.com/raspberrypi/linux/issues/6660 (which it resolves).
Will also submit for 6.12 - but went for 6.14 first as it's a little closer to upstream.

PR #6717, opened 2025-03-12
add mira220 image sensor

fiepfiep:
add driver / device tree support for ams [mira220](https://ams-osram.com/products/sensor-solutions/cmos-image-sensors/ams-mira220)

needed changes also done for [libcamera](https://github.com/raspberrypi/libcamera/pull/248).


arch/arm/boot/dts/overlays/mira220-overlay.dts
```diff
@@ -0,0 +1,102 @@
...
+	};
+
+	fragment@201 {
+		target = <&csi_ep>;
+		__dormant__ {
+			data-lanes = <1 2 >;
+		};
+	};
+
+	fragment@202 {
```
6by9: Fragment is dormant and not referenced by an override.

arch/arm/boot/dts/overlays/mira220-overlay.dts
```diff
@@ -0,0 +1,102 @@
...
+	};
+
+	fragment@102 {
+		target = <&csi1>;
+		__dormant__ {
+			compatible = "brcm,bcm2835-unicam-legacy";
+		};
+	};
+
+	fragment@201 {
```
6by9: Fragment is dormant and not referenced by an override.

arch/arm/boot/dts/overlays/mira220.dtsi
```diff
@@ -0,0 +1,40 @@
...
+	rotation = <0>;
+	orientation = <2>;
+
+	port {
+		cam_endpoint: endpoint {
+			clock-lanes = <0>;
+			data-lanes = <1 2>;
+			// clock-noncontinuous;
+			link-frequencies =
+				/bits/ 64 <456000000>;
```
6by9: Is this the correct link frequency for this sensor?

drivers/media/i2c/mira220.c
```diff
@@ -0,0 +1,4366 @@
...
+ #define MIRA220_HBLANK_1600x1400_304		1440
+ 
+ 
+ #define MIRA220_REG_TEST_PATTERN	0x2091
+ #define	MIRA220_TEST_PATTERN_DISABLE	0x00
+ #define	MIRA220_TEST_PATTERN_VERTICAL_GRADIENT	0x01
+ 
+ /* Embedded metadata stream structure */
+ #define MIRA220_EMBEDDED_LINE_WIDTH 16384
+ #define MIRA220_NUM_EMBEDDED_LINES 1
```
6by9: Does this sensor actually have useful embedded metadata? It's disabled in libcamera.

I'd made the same comment in https://github.com/raspberrypi/linux/pull/6552#discussion_r1893800072 that this isn't how mainline are implementing metadata, so it's simpler to not implement it in the driver if not needed.

drivers/media/i2c/mira220.c
```diff
@@ -0,0 +1,4366 @@
...
+	 u32 min_vblank;
+	 u32 max_vblank;
+	 u32 hblank;
+	 u32 code;
+ };
+ 
+ // 1600_1400_30fps_12b_2lanes
+
+ static const struct mira220_reg full_1600_1400_1500_12b_2lanes_reg[] = {
+	{0x1003,0x2}, //,Initial Upload
```
6by9: Comment isn't useful.

drivers/media/i2c/mira220.c
```diff
@@ -0,0 +1,4366 @@
...
+	 u32 pixel_rate;
+	 u32 min_vblank;
+	 u32 max_vblank;
+	 u32 hblank;
+	 u32 code;
+ };
+ 
+ // 1600_1400_30fps_12b_2lanes
+
+ static const struct mira220_reg full_1600_1400_1500_12b_2lanes_reg[] = {
```
6by9: You have about 18 registers that differ between these 3 huge tables. Break out all the common stuff into a common table, and only send the bits that change per mode

drivers/media/i2c/mira220.c
```diff
@@ -0,0 +1,4366 @@
...
+ 
+ #define MIRA220_NATIVE_WIDTH			1600U
+ #define MIRA220_NATIVE_HEIGHT			1400U
+ 
+ #define MIRA220_PIXEL_ARRAY_LEFT		0U
+ #define MIRA220_PIXEL_ARRAY_TOP			0U
+ #define MIRA220_PIXEL_ARRAY_WIDTH		1600U
+ #define MIRA220_PIXEL_ARRAY_HEIGHT		1400U
+ 
+ #define MIRA220_ANALOG_GAIN_REG			0x400A
```
6by9: Comment made at https://github.com/raspberrypi/linux/pull/6552#discussion_r1893820023 that the V4L2_CCI is preferred rather than bare register defines as it encodes the size of each register.

drivers/media/i2c/mira220.c
```diff
@@ -0,0 +1,4366 @@
...
+	 MIRA220_TEST_PATTERN_DISABLE,
+	 MIRA220_TEST_PATTERN_VERTICAL_GRADIENT,
+ };
+ 
+ 
+ /* regulator supplies */
+ static const char * const mira220_supply_name[] = {
+	 // TODO(jalv): Check supply names
+	 /* Supplies can be enabled in any order */
+	 "VANA",  /* Analog (2.8V) supply */
```
6by9: Regulators should be named in lower case.

6by9 (changes requested):
I've started to review this, but I'm noticing a lot of my comments from #6552 haven't been addressed.

If embedded data isn't required, then do not include it.

arch/arm/boot/dts/overlays/mira220.dtsi
fiepfiep: currently operating at 1.5 Gbit/s per lane

drivers/media/i2c/mira220.c
fiepfiep: thanks, will do.

arch/arm/boot/dts/overlays/mira220-overlay.dts
fiepfiep: done, has been removed.

arch/arm/boot/dts/overlays/mira220-overlay.dts
fiepfiep: resolved.

drivers/media/i2c/mira220.c
fiepfiep: simplified, reduced driver to 1 mode only.

drivers/media/i2c/mira220.c
fiepfiep: I converted everything to the new style cci writes

pelwell:
I'll warn you now that, once the content of these commits has been approved, the end result should ideally be 3 commits - driver changes, config changes, and overlay/README changes.

arch/arm/boot/dts/overlays/mira220-overlay.dts
```diff
@@ -58,28 +58,6 @@
 		};
 	};
 
-	fragment@102 {
```
pelwell: Did you delete this fragment by mistake? The `media-controller` parameter refers to it.

fiepfiep: Yes that is a mistake.

fiepfiep:
> I'll warn you now that, once the content of these commits has been approved, the end result should ideally be 3 commits - driver changes, config changes, and overlay/README changes.

I will keep it in mind. Once all is approved by you I will refactor it to 3 commits.

arch/arm/boot/dts/overlays/mira220.dtsi
fiepfiep: solved, set to 750, as datarate is 1500Mbit/s

drivers/media/i2c/mira220.c
fiepfiep: I have removed the embedded data

drivers/media/i2c/mira220.c
fiepfiep: done.

drivers/media/i2c/mira220.c
fiepfiep: solved

arch/arm/boot/dts/overlays/mira220-overlay.dts
fiepfiep: I put it back. thanks for finding out.

fiepfiep:
I cleaned up the driver a bit, implemented the new CCI_REG style writes, and took care of 6by9's comments.
Please let me know if you find other issues before we can do the merge.

arch/arm/boot/dts/overlays/README
```diff
@@ -3632,6 +3649,7 @@ Params: rotation                Mounting rotation of the camera sensor (0 or
                                 Appears not to work on Pi3.
 
 
+
```
pelwell: This looks like an unnecessary change.

arch/arm/boot/dts/overlays/README
```diff
@@ -3511,6 +3511,23 @@ Params:
                                 (default 16).
 
 
+
```
pelwell: 2 blank lines is enough.

arch/arm/boot/dts/overlays/README
```diff
@@ -3511,6 +3511,23 @@ Params:
...
+Load:   dtoverlay=mira220,<param>=<val>
+Params: rotation                Mounting rotation of the camera sensor (0 or
+                                180, default 0)
+        orientation             Sensor orientation (0 = front, 1 = rear,
+                                2 = external, default external)
+        media-controller        Configure use of Media Controller API for
+                                configuring the sensor (default on)
+        cam0                    Adopt the default configuration for CAM0 on a
+                                Compute Module (CSI0, i2c_vc, and cam0_reg).
+
```
pelwell, line 3527: And here.

pelwell (commented):
Apart from the extra blank lines in the README, the overlay looks good.

Checkpatch has lots of complaints about the driver which you should really fix. Nobody using `printk` directly anymore, for example.

The only additional change I would like (apart from refactoring down to 3 commits as discussed) is the addition of CONFIG_VIDEO_MIRA220 to the new PREEMPT_RT config file arm64/bcm2711_rt_defconfig.

arch/arm/boot/dts/overlays/mira220-overlay.dts
```diff
@@ -0,0 +1,87 @@
...
+
+	__overrides__ {
+		rotation = <&cam_node>,"rotation:0";
+		orientation = <&cam_node>,"orientation:0";
+		media-controller = <0>,"!102";
+		cam0 = <&i2c_frag>, "target:0=",<&i2c_csi_dsi0>,
+		       <&csi_frag>, "target:0=",<&csi0>,
+		       <&clk_frag>, "target:0=",<&cam0_clk>,
+		       <&cam_node>, "clocks:0=",<&cam0_clk>,
+		       <&cam_node>, "VANA-supply:0=",<&cam0_reg>;
```
6by9: Your supply is now `vana-supply`, so this won't match.

drivers/media/i2c/Kconfig
```diff
@@ -149,6 +149,20 @@ config VIDEO_HI847
           To compile this driver as a module, choose M here: the
           module will be called hi847.
 
+config VIDEO_MIRA220
```
6by9: These are normally in alphabetical order. (I see the Arducam drivers have got out of position during forward porting).

Annoyingly this will also move the position in the defconfigs, so perhaps I don't care enough.

drivers/media/i2c/Makefile
```diff
@@ -49,6 +49,7 @@ obj-$(CONFIG_VIDEO_HI556) += hi556.o
 obj-$(CONFIG_VIDEO_HI846) += hi846.o
 obj-$(CONFIG_VIDEO_HI847) += hi847.o
 obj-$(CONFIG_VIDEO_I2C) += video-i2c.o
+obj-$(CONFIG_VIDEO_MIRA220)	+= mira220.o
```
6by9: Ditto - alphabetical order please.

drivers/media/i2c/mira220.c
```diff
@@ -0,0 +1,2481 @@
...
+#define MIRA220_NATIVE_HEIGHT 1400U
+
+#define MIRA220_PIXEL_ARRAY_LEFT 0U
+#define MIRA220_PIXEL_ARRAY_TOP 0U
+#define MIRA220_PIXEL_ARRAY_WIDTH 1600U
+#define MIRA220_PIXEL_ARRAY_HEIGHT 1400U
+
+#define MIRA220_ANALOG_GAIN_MIN 1
+#define MIRA220_ANALOG_GAIN_MAX \
+	1 /* Fixed analog gain to 1, to avoid unexpected behavior. */
```
6by9: Weird line breaking. That does seem quite a limitation though if you have no analogue gain.

drivers/media/i2c/mira220.c
```diff
@@ -0,0 +1,2481 @@
...
+#define MIRA220_EXPOSURE_MIN 1
+
+// Power on function timing
+#define MIRA220_XCLR_MIN_DELAY_US 100000
+#define MIRA220_XCLR_DELAY_RANGE_US 30
+
+// Outdated. See below.
+// pixel_rate = link_freq * 2 * nr_of_lanes / bits_per_sample
+// 1.0Gb/s * 2 * 2 / 12 = 357913941
+// #define MIRA220_PIXEL_RATE		(357913941)
```
6by9: If it's outdated, then please remove it.

drivers/media/i2c/mira220.c
```diff
@@ -0,0 +1,2481 @@
...
+  */
+#include <linux/v4l2-controls.h>
+
+#define MIRA220_NATIVE_WIDTH 1600U
+#define MIRA220_NATIVE_HEIGHT 1400U
+
+#define MIRA220_PIXEL_ARRAY_LEFT 0U
+#define MIRA220_PIXEL_ARRAY_TOP 0U
+#define MIRA220_PIXEL_ARRAY_WIDTH 1600U
+#define MIRA220_PIXEL_ARRAY_HEIGHT 1400U
```
6by9, line 33: Looking at https://look.ams-osram.com/m/2feacfd5e2a2d20e/original/Mira220-1-2-7-2-2MP-NIR-Enhanced-Global-Shutter-Image.pdf, Figure 1 says the sensor has 32 extra rows top and bottom, and 21 extra columns left and right.

drivers/media/i2c/mira220.c
```diff
@@ -0,0 +1,2481 @@
...
+#define MIRA220_HBLANK_640x480_304 2400
+#define MIRA220_HBLANK_1600x1400_304 1440
+
+#define MIRA220_REG_TEST_PATTERN CCI_REG8(0x2091)
+#define MIRA220_TEST_PATTERN_DISABLE 0x00
+#define MIRA220_TEST_PATTERN_VERTICAL_GRADIENT 0x01
+
+/* Embedded metadata stream structure */
+#define MIRA220_EMBEDDED_LINE_WIDTH 16384
+#define MIRA220_NUM_EMBEDDED_LINES 1
```
6by9: You'd said in https://github.com/raspberrypi/linux/pull/6717#discussion_r2025643185 "I have removed the embedded data", so why are these still here. Likewise METADATA_PAD at line 158.

drivers/media/i2c/mira220.c
```diff
@@ -0,0 +1,2481 @@
...
+
+	u32 pixel_rate;
+	u32 min_vblank;
+	u32 max_vblank;
+	u32 hblank;
+	u32 code;
+};
+
+static const struct cci_reg_sequence full_1600_1400_1500_12b_2lanes_reg_new[] = {
+	{ CCI_REG8(0x1003), 0x2 }, //,Initial Upload
```
6by9: This comment of `//,Initial Upload` is still present and unnecessary.

drivers/media/i2c/mira220.c
```diff
@@ -0,0 +1,2481 @@
...
+  * 
+  * Inverse the above formula to derive HBLANK from TARGET_FPS:
+  * HBLANK=1/((1/MIRA220_PIXEL_RATE)*TARGET_FPS*(HEIGHT+MIRA220_MIN_VBLANK))-WIDTH
+  * Example with TARGET_FPS of 120 fps for 640x480
+  * HBLANK=1/((1/54616828)*120*(480+13))-640=283
+  * Example with TARGET_FPS of 30 fps for 1600x1400
+  * HBLANK=1/((1/204813108)*30*(1400+13))-1600=3232
+  */
+
+#define MIRA220_HBLANK_640x480_120FPS 3860
```
6by9: You've reduced the driver to one mode, so most of these `MIRA220_HBLANK_` defines are redundant.

drivers/media/i2c/mira220.c
```diff
@@ -0,0 +1,2481 @@
...
+static const int mira220_test_pattern_val[] = {
+	MIRA220_TEST_PATTERN_DISABLE,
+	MIRA220_TEST_PATTERN_VERTICAL_GRADIENT,
+};
+
+/* regulator supplies */
+static const char *const mira220_supply_name[] = {
+	// TODO(jalv): Check supply names
+	/* Supplies can be enabled in any order */
+	"VANA", /* Analog (2.8V) supply */
```
6by9: Still in capitals, and that doesn't match the dtoverlay.

drivers/media/i2c/mira220.c
```diff
@@ -0,0 +1,2481 @@
...
+{
+	struct i2c_client *const client = v4l2_get_subdevdata(&mira220->sd);
+	int ret = 0;
+	u32 frame_time;
+	int try_cnt;
+
+	for (try_cnt = 0; try_cnt < 5; try_cnt++) {
+		ret = cci_write(mira220->regmap, MIRA220_IMAGER_STATE_REG,
+				MIRA220_IMAGER_STATE_STOP_AT_ROW, NULL);
+		if (ret) {
```
6by9: You really have such dodgy I2C comms such that you have to retry?

drivers/media/i2c/mira220.c
```diff
@@ -0,0 +1,2481 @@
...
+		     MIRA220_DEFAULT_LINE_LENGTH / MIRA220_DEFAULT_PIXEL_CLOCK;
+
+	usleep_range(frame_time, frame_time + 1000); //TODO, set to 1 frametime
+
+	return ret;
+}
+
+// Returns the maximum exposure time in row_length (reg value).
+// Calculation is baded on Mira220 datasheet Section 9.2.
+static u32 mira220_calculate_max_exposure_time(u32 vsize, u32 vblank,
```
6by9: vsize, more commonly known as height.

drivers/media/i2c/mira220.c
```diff
@@ -0,0 +1,2481 @@
...
+	unsigned int i;
+
+	lockdep_assert_held(&mira220->mutex);
+
+	for (i = 0; i < ARRAY_SIZE(codes); i++)
+		if (codes[i] == code)
+			break;
+
+	if (i >= ARRAY_SIZE(codes)) {
+		dev_err_ratelimited(&client->dev,
```
6by9: Error message not needed, and gives a route for userspace to be able to spam the kernel log (even if rate limited).

drivers/media/i2c/mira220.c
```diff
@@ -0,0 +1,2481 @@
...
+	}
+
+	printk(KERN_INFO
+	       "[MIRA220]: exposure fun width %d, hblank %d, vblank %d, row len %d, ctrl->val %d capped to %d.\n",
+	       mira220->mode->width, mira220->hblank->val, mira220->vblank->val,
+	       mira220->mode->row_length, exposure, capped_exposure);
+
+	/*test cci write*/
+	ret = cci_read(mira220->regmap, MIRA220_EXP_TIME_REG, &readval, &ret);
+	otherval = readval & 0xFFFF;
```
6by9: Why read the value before writing it? You do nothing with the result.

drivers/media/i2c/mira220.c
```diff
@@ -0,0 +1,2481 @@
...
+		return -ENOMEM;
+
+	v4l2_i2c_subdev_init(&mira220->sd, client, &mira220_subdev_ops);
+	mira220->regmap = devm_cci_regmap_init_i2c(client, 16);
+
+	/* Check the hardware configuration in device tree */
+	if (mira220_check_hwcfg(dev))
+		return -EINVAL;
+
+	/* Parse device tree to check if dtoverlay has param skip-reg-upload=1 */
```
6by9: Huh?

drivers/media/i2c/mira220.c
```diff
@@ -0,0 +1,2481 @@
...
+	mira220_free_controls(mira220);
+
+	pm_runtime_disable(&client->dev);
+	if (!pm_runtime_status_suspended(&client->dev))
+		mira220_power_off(&client->dev);
+	pm_runtime_set_suspended(&client->dev);
+}
+
+static const struct dev_pm_ops mira220_pm_ops = {
+	SET_SYSTEM_SLEEP_PM_OPS(mira220_suspend, mira220_resume)
```
6by9: You shouldn't be implementing SLEEP_OPS.

drivers/media/i2c/mira220.c
```diff
@@ -0,0 +1,2481 @@
...
+		SET_RUNTIME_PM_OPS(mira220_power_off, mira220_power_on, NULL)
+};
+
+static const struct of_device_id mira220_dt_ids[] = { { .compatible =
```
6by9: Please use the normal formatting for dt_ids.
```
static const struct of_device_id mira220_dt_ids[] = {
	{ .compatible = "ams,mira220" },
	{ /* sentinel */ }
};
```

drivers/media/i2c/mira220.c
```diff
@@ -0,0 +1,2481 @@
...
+								"ams,mira220" },
+						      { /* sentinel */ } };
+MODULE_DEVICE_TABLE(of, mira220_dt_ids);
+
+static const struct i2c_device_id mira220_ids[] = { { "mira220", 0 }, {} };
+MODULE_DEVICE_TABLE(i2c, mira220_ids);
```
6by9: You shouldn't need I2C ids. This driver needs config data, so you can't load via sysfs.

6by9 (commented):
Still a number of my previous review comments haven't been addressed.

- Please remove the references to the embedded data pad.
- Please remove all commented out code.
- Please remove all redundant comments.
- Please rationalise the logging messages. Very few should be needed at all.
- Regulator names should be consistently in lower case.

pelwell:
FYI the build checks explicitly set CONFIG_WERROR=y, overriding the defconfigs:
```
scripts/config --file ${{github.workspace}}/build/.config --set-val CONFIG_WERROR y
```

fiepfiep:
@pelwell could you run your checks again? I split my changes into 3 commits as requested.

pelwell:
It looks like somebody has started them. Please mark the PR as "Ready for review" when you think it is. However, there is no excuse for some of these (https://github.com/raspberrypi/linux/actions/runs/14470568709/job/40583365956?pr=6717#step:6:80) checkpatch errors such as trailing whitespace.

fiepfiep:
I still get this error in checkpatch, what does it mean?
@pelwell 
```
 WARNING: From:/Signed-off-by: email name mismatch: 'From: philippe baetens <philippe.baetens@ams-osram.com>' != 'Signed-off-by: Philippe Baetens <philippe.baetens@ams-osram.com>'
Error: WARNING: From:/Signed-off-by: email name mismatch: 'From: philippe baetens <philippe.baetens@ams-osram.com>' != 'Signed-off-by: Philippe Baetens <philippe.baetens@ams-osram.com>'
```

pelwell:
Is it possibly just that your name appears with and without uppercase letters? I'm not bothered about that, but it makes sense to fix it to avoid such errors in the future.

fiepfiep:
> Is it possibly just that your name appears with and without uppercase letters? I'm not bothered about that, but it makes sense to fix it to avoid such errors in the future.

that seems to be the case indeed. i modified the commit signoff message + some of the checkpatch remarks. Hope it is ok now.

pelwell:
That's fixed it. I'll give these builds time to complete, in which time 6by9 may want to chip in, but I've been taking his silence for consent.

6by9:
> That's fixed it. I'll give these builds time to complete, in which time 6by9 may want to chip in, but I've been taking his silence for consent.

No, just hadn't had a chance to look through.
It looks like several of my previous comments haven't been addressed, eg the overlay override `cam0` still setting `VANA-supply` and changing `vana-supply`.

arch/arm/configs/bcm2709_defconfig
```diff
@@ -934,6 +934,7 @@ CONFIG_VIDEO_BCM2835_UNICAM_LEGACY=m
 CONFIG_VIDEO_BCM2835_UNICAM=m
 CONFIG_VIDEO_ARDUCAM_64MP=m
 CONFIG_VIDEO_ARDUCAM_PIVARIETY=m
+CONFIG_VIDEO_MIRA220=m
```
6by9: With the move within Kconfig to keep everything in alphabetical order, I'm expecting this to be the wrong location in the defconfig.

drivers/media/i2c/mira220.c
```diff
@@ -0,0 +1,2160 @@
...
+
+
+// The supported formats. All flip/mirror combinations have the same byte order because the sensor
+// is monochrome
+
+static const u32 codes[] = {
+	//MEDIA_BUS_FMT_Y8_1X8,
+	//MEDIA_BUS_FMT_Y10_1X10,
+	//MEDIA_BUS_FMT_Y12_1X12,
+	MEDIA_BUS_FMT_SGRBG8_1X8,
```
6by9: The comment just before this says that the sensor is monochrome, and yet you've used the GRBG Bayer pattern code. So which is it?

drivers/media/i2c/mira220.c
```diff
@@ -0,0 +1,2160 @@
...
+	ret = mira220_write_stop_streaming_regs(mira220);
+	if (ret) {
+		dev_err(&client->dev, "Could not write stream-on sequence");
+		goto err_rpm_put;
+	}
+
+	reg_list = &mira220->mode->reg_list;
+	ret = cci_multi_reg_write(mira220->regmap, reg_list->regs,
+				  reg_list->num_of_regs, NULL);
+	// ret = mira220_write_regs(mira220, reg_list->regs, reg_list->num_of_regs);
```
6by9: Remove the now redundant line.

drivers/media/i2c/mira220.c
```diff
@@ -0,0 +1,2160 @@
...
+	pm_runtime_put(&client->dev);
+}
+
+static int mira220_set_stream(struct v4l2_subdev *sd, int enable)
+{
+	struct mira220 *mira220 = to_mira220(sd);
+	int ret = 0;
+
+	mutex_lock(&mira220->mutex);
+	if (mira220->streaming == enable) {
```
6by9: This should never be possible as the framework blocks it.

drivers/media/i2c/mira220.c
```diff
@@ -0,0 +1,2160 @@
...
+}
+
+static int __maybe_unused mira220_resume(struct device *dev)
+{
+	struct i2c_client *client = to_i2c_client(dev);
+	struct v4l2_subdev *sd = i2c_get_clientdata(client);
+	struct mira220 *mira220 = to_mira220(sd);
+	int ret;
+
+	if (mira220->streaming) {
```
6by9: resume shouldn't start streaming directly as the whole pipeline needs to be started correctly.

drivers/media/i2c/mira220.c
```diff
@@ -0,0 +1,2160 @@
...
+					    mira220->mode->min_vblank);
+
+	mira220->hblank =
+		v4l2_ctrl_new_std(ctrl_hdlr, &mira220_ctrl_ops, V4L2_CID_HBLANK,
+				  mira220->mode->hblank, mira220->mode->hblank,
+				  1, mira220->mode->hblank);
+
+	// Make the vblank control read only. This could be changed to allow changing framerate in
+	// runtime, but would require adapting other settings
+	// mira220->vblank->flags |= V4L2_CTRL_FLAG_READ_ONLY;
```
6by9: This has changed in that you have VBLANK control, so remove the comment and the commented out code.

fiepfiep:
So - what are the next steps to get this driver into the raspberry linux kernel?
At the moment I'm only merging into the 6.12 kernel. What about older or newer kernel versions? 
How does this driver propagate to official bookworm builds?

thanks.

drivers/media/i2c/mira220.c
```diff
@@ -0,0 +1,2160 @@
...
+
+static int mira220_power_off(struct device *dev)
+{
+	struct i2c_client *client = to_i2c_client(dev);
+	struct v4l2_subdev *sd = i2c_get_clientdata(client);
+	struct mira220 *mira220 = to_mira220(sd);
+	(void)mira220;
+
+	regulator_bulk_disable(MIRA220_NUM_SUPPLIES, mira220->supplies);
+	clk_disable_unprepare(mira220->xclk);
```
6by9, line 1140: Clock and regulators are disabled in the opposite order to being prepared.

drivers/media/i2c/mira220.c
```diff
@@ -0,0 +1,2160 @@
...
+
+	return ret;
+
+err_unlock:
+	mutex_unlock(&mira220->mutex);
+
+	return ret;
+}
+
+static int __maybe_unused mira220_suspend(struct device *dev)
```
6by9: Both mira220_suspend and mira220_resume are totally unused. Remove them.

6by9 (changes requested):
In reading through to see if my previous comments had been addressed, add a number more.

There are a fair number of commented out blocks of code. They should be reviewed and removed where redundant.

6by9:
> So - what are the next steps to get this driver into the raspberry linux kernel? At the moment I'm only merging into the 6.12 kernel. What about older or newer kernel versions? How does this driver propagate to official bookworm builds?

Older kernels won't get updated.

Later kernels will get a forward port, as long as there aren't any significant build changes required by the every-changing kernel APIs. Simple changes are likely to get fixed up. Significant changes without obvious fixups will be reported back to you to fix - we don't undertake the full support burden of 3rd party drivers, and they may get dropped from the defconfigs if not fixed and/or get the `BROKEN` tag added to the Kconfig entry if the situation persists for a significant period.

rpi-update gets kernel updates normally several time a week.
Raspberry Pi OS gets updates every few months based on when significant things occur, providing no significant issues are known about at that point.

drivers/media/i2c/mira220.c
fiepfiep: do you mean all mutex statements are not needed?

6by9: `mira220->streaming` is not required. The framework should never call `set_stream(..., 1)` if the stream is already enabled, or `set_stream(... , 0)` if there wasn't also a previous call to `set_stream(..., 1)` that returned 0.

If that means you don't need the mutex, then that's a bonus.

arch/arm/configs/bcm2709_defconfig
fiepfiep: right. will move to the line below the IMX sensors.

drivers/media/i2c/mira220.c
fiepfiep: sensor comes both as RGB and mono. For now, I'm only doing the RGB.

drivers/media/i2c/mira220.c
fiepfiep: ah, clear. Will fix that too.

drivers/media/i2c/mira220.c
fiepfiep: resume got deleted (unused)

drivers/media/i2c/mira220.c
fiepfiep: sure.

drivers/media/i2c/mira220.c
fiepfiep: ok, will swap order.

drivers/media/i2c/mira220.c
fiepfiep: will remove them.

pelwell:
Some of @6by9's comments have yet to be addressed, but the overlay and general structure look okay to me.

fiepfiep:
> Some of @6by9's comments have yet to be addressed, but the overlay and general structure look okay to me.

Thanks for reviewing. I believe I addressed all of @6by9's comments, but I might have missed something.. Can you be a bit more specific? thanks!

Also: each time I just 'force push' 3 new commits (overlay, conf, i2c driver) - not sure if this is your preferred way, or if you like to see my full commit history?

arch/arm/boot/dts/overlays/mira220.dtsi
```diff
@@ -0,0 +1,40 @@
+// SPDX-License-Identifier: GPL-2.0-only
+// Definitions for MIRA220 camera module on VC I2C bus
+
+
+//#include <dt-bindings/gpio/gpio.h>
```
pelwell: Is this commented out section left here for a reason? If so, explain why in a comment, otherwise delete.

pelwell:
On this page (https://github.com/raspberrypi/linux/pull/6717/files), scroll down and expand mira.c, then look at the comments that remain, e.g.: https://github.com/raspberrypi/linux/pull/6717/files#r2027459784 and https://github.com/raspberrypi/linux/pull/6717/files#r2027467385

drivers/media/i2c/mira220.c
fiepfiep: probably no longer needed indeed. There used to be a problem but this was resolved.

drivers/media/i2c/mira220.c
```diff
@@ -0,0 +1,2087 @@
...
+#include <linux/regulator/consumer.h>
+#include <media/v4l2-ctrls.h>
+#include <media/v4l2-device.h>
+#include <media/v4l2-event.h>
+#include <media/v4l2-fwnode.h>
+#include <media/v4l2-mediabus.h>
+#include <media/v4l2-cci.h>
+#include <linux/unaligned.h>
+
+/* Active pixel array is 1600 (H) x 1400 (V) pixels.
```
pelwell: Remove the space at the end of this line, squash the commits back to 3, then I think we're good to merge.

drivers/media/i2c/mira220.c
```diff
@@ -0,0 +1,2087 @@
...
+	u16 address;
+	u8 val;
+};
+
+struct mira220_reg_list {
+	unsigned int num_of_regs;
+	const struct cci_reg_sequence *regs;
+};
+
+struct mira220_v4l2_reg {
```
6by9: Unused.

drivers/media/i2c/mira220.c
```diff
@@ -0,0 +1,2087 @@
...
+	u32 row_length;
+	u32 pixel_rate;
+	u32 min_vblank;
+	u32 max_vblank;
+	u32 hblank;
+	u32 code;
+};
+
+static const struct cci_reg_sequence full_1600_1400_1500_12b_2lanes_reg_new[] = {
+	{ CCI_REG8(0x1003), 0x2 }, //  Initial Upload
```
6by9: All of the `// Initial upload` comments are meaningless

drivers/media/i2c/mira220.c
```diff
@@ -0,0 +1,2087 @@
...
+	struct v4l2_ctrl *pixel_rate;
+	struct v4l2_ctrl *vflip;
+	struct v4l2_ctrl *hflip;
+	struct v4l2_ctrl *vblank;
+	struct v4l2_ctrl *hblank;
+	struct v4l2_ctrl *exposure;
+	struct v4l2_ctrl *gain;
+
+	u16 mira220_reg_w_cached_addr;
+	u8 mira220_reg_w_cached_flag;
```
6by9: Neither are used.

drivers/media/i2c/mira220.c
```diff
@@ -0,0 +1,2087 @@
...
+		return -EINVAL;
+
+	if (code->pad == IMAGE_PAD) {
+		if (code->index >= ARRAY_SIZE(codes))
+			return -EINVAL;
+
+		code->code = mira220_validate_format_code_or_default(
+			mira220, codes[code->index]);
+	} else {
+		if (code->index > 0)
```
6by9: This else clause is redundant with no metadata pad.

drivers/media/i2c/mira220.c
```diff
@@ -0,0 +1,2087 @@
...
+static int mira220_enum_frame_size(struct v4l2_subdev *sd,
+				   struct v4l2_subdev_state *sd_state,
+				   struct v4l2_subdev_frame_size_enum *fse)
+{
+	struct mira220 *mira220 = to_mira220(sd);
+
+	if (fse->pad >= NUM_PADS)
+		return -EINVAL;
+
+	if (fse->pad == IMAGE_PAD) {
```
6by9: You've already guaranteed this condition with the `if (fse->pad >= NUM_PADS)` check

drivers/media/i2c/mira220.c
```diff
@@ -0,0 +1,2087 @@
...
+				    struct v4l2_subdev_format *fmt)
+{
+	if (fmt->pad >= NUM_PADS)
+		return -EINVAL;
+
+	if (fmt->which == V4L2_SUBDEV_FORMAT_TRY) {
+		struct v4l2_mbus_framefmt *try_fmt =
+			v4l2_subdev_state_get_format(sd_state, fmt->pad);
+
+		try_fmt->code = fmt->pad == IMAGE_PAD ?
```
6by9: `fmt->pad` will be `IMAGE_PAD`, so no need for the ? : conditional

drivers/media/i2c/mira220.c
```diff
@@ -0,0 +1,2087 @@
...
+			__v4l2_ctrl_modify_range(mira220->vblank,
+						 mira220->mode->min_vblank,
+						 mira220->mode->max_vblank, 1,
+						 mira220->mode->min_vblank);
+
+
+			__v4l2_ctrl_s_ctrl(mira220->vblank,
+					   mira220->mode->min_vblank);
+		}
+	} else {
```
6by9: There is no else.

drivers/media/i2c/mira220.c
```diff
@@ -0,0 +1,2087 @@
...
+	if (ret)
+		goto err_rpm_put;
+
+	ret = mira220_write_start_streaming_regs(mira220);
+	if (ret) {
+		dev_err(&client->dev, "Could not write stream-on sequence");
+		goto err_rpm_put;
+	}
+
+	__v4l2_ctrl_grab(mira220->vflip, true);
```
6by9: Grabbing the flip controls is normally only needed if changing the flip changes the Bayer order. This driver appears not to do that.

drivers/media/i2c/mira220.c
```diff
@@ -0,0 +1,2087 @@
...
+	ret = mira220_otp_read(mira220, 0x0d, 0, &val);
+	dev_err(&client->dev, "Read OTP add 0x0d with val %x\n", val);
+	ret = mira220_otp_read(mira220, 0x19, 0, &val);
+	dev_err(&client->dev, "Read OTP add 0x19 with val %x\n", val);
+	ret = mira220_otp_read(mira220, 0x19, 1, &val);
+	dev_err(&client->dev, "Read OTP add 0x19+1 with val %x\n", val);
+
+	mira220_otp_power_off(mira220);
+
+	return 0;
```
6by9: If we never fail the identity check, then make it a void function.

You never check any of the returned values from `mira220_otp_read` which would be a basic identity check.

drivers/media/i2c/mira220.c
```diff
@@ -0,0 +1,2087 @@
...
+
+	mira220->gain = v4l2_ctrl_new_std(
+		ctrl_hdlr, &mira220_ctrl_ops, V4L2_CID_ANALOGUE_GAIN,
+		MIRA220_ANALOG_GAIN_MIN, MIRA220_ANALOG_GAIN_MAX,
+		MIRA220_ANALOG_GAIN_STEP, MIRA220_ANALOG_GAIN_DEFAULT);
+
+	mira220->hflip = v4l2_ctrl_new_std(ctrl_hdlr, &mira220_ctrl_ops,
+					   V4L2_CID_HFLIP, 0, 1, 1, 0);
+	if (mira220->hflip)
+		mira220->hflip->flags |= V4L2_CTRL_FLAG_MODIFY_LAYOUT;
```
6by9, line 1804: In what way does it modify the layout? Normally this is when it changes the Bayer order, but your sensor appears not to do that.

drivers/media/i2c/mira220.c
```diff
@@ -0,0 +1,2087 @@
...
+	struct i2c_client *client = v4l2_get_subdevdata(&mira220->sd);
+	struct v4l2_ctrl_handler *ctrl_hdlr;
+	struct v4l2_fwnode_device_properties props;
+	int ret;
+
+	u32 max_exposure = 0;
+
+	ctrl_hdlr = &mira220->ctrl_handler;
+	/* v4l2_ctrl_handler_init gives a hint/guess of the number of v4l2_ctrl_new */
+	ret = v4l2_ctrl_handler_init(ctrl_hdlr, 16);
```
6by9: How many controls do you create? I make it 10, not 16.

drivers/media/i2c/mira220.c
```diff
@@ -0,0 +1,2087 @@
...
+		return -EINVAL;
+	}
+
+	ret = mira220_get_regulators(mira220);
+	if (ret) {
+		dev_err(dev, "failed to get regulators\n");
+		return ret;
+	}
+
+	usleep_range(1000000, 1000000 + 100);
```
6by9: Slightly crazy range of 100us tolerance on 1second.

`fsleep` is the normally recommended call now - https://www.kernel.org/doc/html/latest/timers/delay_sleep_functions.html

drivers/media/i2c/mira220.c
```diff
@@ -0,0 +1,2087 @@
...
+		.min_vblank = 20,
+		.max_vblank = 50000,
+		.hblank = MIRA220_HBLANK_1600x1400_304,
+		.code = MEDIA_BUS_FMT_SGRBG12_1X12,
+	},
+};
+
+struct mira220 {
+	struct v4l2_subdev sd;
+	struct media_pad pad[NUM_PADS];
```
6by9: Hardly worth having an array with one element.

6by9 (commented):
There are still a number of remnants from the metadata pad kicking around. The things I've listed below would be nice to be cleaned up, but seeing as you'll be retaining responsibility for maintaining it, I'm not going to fuss. None of them should break the system.
You would be picked up on all these things if you try upstreaming the driver.

fiepfiep:
i think the driver is now in good shape. requesting another review please..

pelwell:
This still applies (both aspects):

> The only additional change I would like (apart from refactoring down to 3 commits as discussed) is the addition of CONFIG_VIDEO_MIRA220 to the new PREEMPT_RT config file arm64/bcm2711_rt_defconfig.

fiepfiep:
> This still applies (both aspects):
> 
> > The only additional change I would like (apart from refactoring down to 3 commits as discussed) is the addition of CONFIG_VIDEO_MIRA220 to the new PREEMPT_RT config file arm64/bcm2711_rt_defconfig.

okay, if all is good - I'll start refactoring down to 3 commits. I've added the rt_defconfig too.

fiepfiep:
commits have been squashed back to 3 commits as requested. @pelwell @6by9

Issue #6721, opened 2025-03-16
Hyperpixel 4 fails to boot on 6.12 kernel


LaaZa:
---
name: vc4-kms-dpi-hyperpixel4 breaks boot on kernel 6.12

---

**Describe the bug**
Boot fails with vc4-kms-dpi-hyperpixel4 enabled on 6.12 kernel, all signal is lost and boot does not proceed.

**To reproduce**
1. Run kernel 6.12
2. Enable Hyperpixel 4 overlay with vc4-kms-dpi-hyperpixel4 in config.txt
3. Boot fails and HDMI nor the Hyperpixel has signal

**Expected behaviour**
System boots normally and the Hyperpixel is working as it does on 6.6 kernel

**Actual behaviour**
Fail to boot and loss of signal

**System**
 Copy and paste the results of the raspinfo command in to this section. Alternatively, copy and paste a pastebin link, or add answers to the following questions:

* Which model of Raspberry Pi? RPi 400
* Which OS and version (`cat /etc/rpi-issue`)? EndeavourOS(latest)/Alpine(3.21)
* Which firmware version (`vcgencmd version`)? Feb 17 2025 da7125
* Which kernel version (`uname -a`)? Any 6.12.y (official Raspberry Pi kernel)

**Logs**
I don't have logs because signal is lost and it fails during boot.

**Additional context**
Fixed on EndeavourOS by compiling 6.6 kernel. Didn't replace overlays, only kernel.img.

6by9:
Thanks for the report. Reproduced. Now investigating why it fails.

6by9:
Ah, downstream 6.6 had commit https://github.com/raspberrypi/linux/commit/7a5df2564ca1e83585326c183dedd9db662aa389 that added the driver for ili9806 as used on the HyperPixel4.

6.12 has gained a driver for the same display using the same filename via https://github.com/raspberrypi/linux/commit/baf272bac637d3275bb83c17ac849b44a4590655
The tree therefore doesn't have the downstream driver, so no configuration for the HyperPixel4.

Time to merge the two (hopefully they are close enough to compatible).

6by9:
And the bad news is that the new driver only supports DSI drive, not DPI configured over SPI.
Reworking to support both is going to be slightly more involved.

LaaZa:
I don't have one to test, but could this affect the square one too?

6by9:
> I don't have one to test, but could this affect the square one too?

[...]

PR #6722, opened 2025-03-17
Fixup HyperPixel4 display on 6.12

6by9:
See #6721


6by9:
Confirmed as working for OP on #6721

PR #6728, opened 2025-03-19
Adding Pisound Micro kernel module

gtrainavicius:
Submitting a new kernel module and its DT overlay for an upcoming Blokas product.


pelwell:
We are not going to build any more 6.6 images - could you submit this against 6.12 instead (perhaps after waiting for the first batch of feedback from the build checks)?

gtrainavicius:
Will do.

pelwell:
For the compatible string, if it's compatible with all Pis use "brcm,bcm2835".

pelwell:
And be aware that we've dropped arm64 bcmrpi3_defconfig in 6.12 and later.

gtrainavicius:
Opened #6731 for rpi-6.12.y

PR #6731, opened 2025-03-19
Adding Pisound Micro kernel module based on rpi-6.12.y

gtrainavicius:
Continuing from #6728


arch/arm/boot/dts/overlays/README
```diff
@@ -3804,6 +3804,21 @@ Params: sel                     The position used for the sel rotary switch.
                                 position. If param is omitted, sel=0 is assumed.
 
 
+Name:   pisound-micro
+Info:   Configures the Blokas Labs Pisound Micro card
+Load:   dtoverlay=pisound-micro
```
pelwell: You'll have to add `,<param>=<val>` here.

arch/arm/boot/dts/overlays/README
```diff
@@ -3804,6 +3804,21 @@ Params: sel                     The position used for the sel rotary switch.
...
+Params: input-mode             'differential' (default) or 'single-ended'.
```
pelwell: Indents are at fixed character positions. Each of these lines needs an extra space for the body text. Fortunately, there is room.

arch/arm/boot/dts/overlays/pisound-micro-overlay.dts
```diff
@@ -0,0 +1,133 @@
...
+
+	fragment@4 {
+		target = <&i2s_clk_consumer>;
+		__overlay__ {
+			status = "okay";
+		};
+	};
+
+	__overrides__ {
+		input-mode=<&pisound_micro_codec>,"input-mode{single-ended,differential}";
```
pelwell, line 121: That's a nice use of lookup tables for input validation.

arch/arm/boot/dts/overlays/pisound-micro-overlay.dts
```diff
@@ -0,0 +1,133 @@
...
+		hp-out-mode=<&pisound_micro_codec>,
+			"hp-out-mode{capless-headphone,headphone,line-out}";
+		line-out-mode=<&pisound_micro_codec>,"line-out-mode{line-out,headphone}";
+		mono-out-mode=<&pisound_micro_codec>,"mono-out-mode{line-out,headphone}";
+	};
+
+	__exports__ {
+		pisound_micro_ctrl;
```
pelwell, line 129: How do you plan to use the label outside the overlay?

pelwell (commented):
This is a huge PR. Are the difference to the standard adau1961 driver so great that a complete fork was the best way forward?

gtrainavicius:
> This is a huge PR. Are the difference to the standard adau1961 driver so great that a complete fork was the best way forward?

Yes, it does some things specific to Pisound Micro, like VGND short protection logic, as well as slightly different power management for better audio. These customizations make sense for Pisound Micro, it might not make sense for general users, so that's the main reason to have Pisound Micro specific fork of the code.

arch/arm/boot/dts/overlays/pisound-micro-overlay.dts
gtrainavicius: Here's an example how rotary-encoder driver could be used with that:

```
/dts-v1/;
/plugin/;

/ {
	compatible = "brcm,bcm2835";

	fragment@0 {
		target-path = "/";
		__overlay__ {
			rotary_encoder@0 {
				compatible = "rotary-encoder";
				gpios = <&pisound_micro_ctrl 0 GPIO_ACTIVE_LOW>, <&pisound_micro_ctrl 1 GPIO_ACTIVE_LOW>;
				linux,axis = <0>;
				rotary-encoder,encoding = "binary";
				rotary-encoder,rollover;
			};
		};
	};
};
```

arch/arm/boot/dts/overlays/README
gtrainavicius: Fixed both README issues.

pelwell:
It's with a slightly heavy heart that I'm going to merge this, on the understanding that, as with similar sound cards, if forward porting becomes non-trivial in the future then it may get dropped.

gtrainavicius:
Thank you! Please contact us in case there's any non trivial support issues in future kernel code, as you'll likely notice it sooner than we do.

Issue #6733, opened 2025-03-20
sh1106-spi invalid buffer alignment on 6.12 kernel


gtrainavicius:
### Describe the bug

Trying to load the DT overlay and module for sh1106 display (possibly other SPI displays are affected too), fails with `invalid buffer alignment`.

It is not necessary to have the actual display to be hooked up to reproduce the problem.

Doesn't happen on 6.6 kernel versions.

### Steps to reproduce the behaviour

Execute:

```
sudo dtoverlay sh1106-spi,reset_pin=17,dc_pin=27
```

### Device (s)

Raspberry Pi 5

### System

```
patch@patchbox:~ $ cat /etc/rpi-issue
Raspberry Pi reference 2024-04-04
Generated using pi-gen, https://github.com/RPi-Distro/pi-gen, 64e3efc023558a4d4634a702c467f513d5b2bca1, stage-patchbox
patch@patchbox:~ $ vcgencmd version
2025/03/19 13:41:26
Copyright (c) 2012 Broadcom
version cec1d3ae (release) (embedded)
patch@patchbox:~ $ uname -a
Linux patchbox 6.12.19-v8-16k+ #1865 SMP PREEMPT Wed Mar 19 13:48:20 GMT 2025 aarch64 GNU/Linux
```

### Logs

[6.12_error.log](https://github.com/user-attachments/files/19369888/6.12_error.log)

6.6 success log for comparison:

[6.6_success.log](https://github.com/user-attachments/files/19369893/6.6_success.log)

6by9:
Upstream have managed to configure two different kernel modules to have the same compatible string
https://elixir.bootlin.com/linux/v6.13.7/source/drivers/gpu/drm/solomon/ssd130x-spi.c#L113
https://elixir.bootlin.com/linux/v6.13.7/source/drivers/staging/fbtft/fb_sh1106.c#L173

Both are getting loaded, and it's the gpu/drm/solomon driver that is crashing.

If you want the deprecated fbtf version (which is what you appeared to have in 6.6), then create a file `/etc/modprobe.d/blacklist-ssd130x-spi.conf` with the contents `blacklist ssd130x-spi`.

P33M:
The splat after that is from the AXI DMAC driver - presumably this error path hasn't been called before...

Also, I'm not entirely sure that the DMAC requires a word-aligned start address. It should only need naturally-aligned buffers with a length of integer multiples of SRC_WIDTH.

6by9:
[...]

PR #6734, opened 2025-03-20
Fix two more RP1 DMA bugs

pelwell:
Two bugs discovered while investigating https://github.com/raspberrypi/linux/issues/6733:
1. axi_desc_put should not assume that all the LLIs owned by the descriptor have been initialised.
2. It is not illegal for a memory address or transfer length to be word-odd (not a multiple of 4). However, it is illegal for the transfer length to not be a multiple of the device/register width.

### DIFF
--- a/drivers/dma/dw-axi-dmac/dw-axi-dmac-platform.c
+++ b/drivers/dma/dw-axi-dmac/dw-axi-dmac-platform.c
@@ -367,7 +367,8 @@ static void axi_desc_put(struct axi_dma_desc *desc)
 
 	for (descs_put = 0; descs_put < count; descs_put++) {
 		hw_desc = &desc->hw_desc[descs_put];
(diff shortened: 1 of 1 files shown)


popcornmix (approved):

drivers/dma/dw-axi-dmac/dw-axi-dmac-platform.c
```diff
@@ -798,6 +794,11 @@ static int dw_axi_dma_set_hw_desc(struct axi_dma_chan *chan,
 		return -EINVAL;
 	}
 
+	if (len % reg_width) {
```
P33M, line 797: (Thinking aloud, not an issue at the moment) I wonder if use-cases that have memory buffers spanning a page boundary with odd alignment are broken when using this driver - and this error message makes this case known. 

But fixing this means rewriting dw_axi_dma_chan_prep_[cyclic|slave]_sg to dynamically allocate hardware descriptors inside for_each_sg().

PR #6735, opened 2025-03-20
arm64: defconfig: Experimental PREEMPT_RT defconfig for BCM2711 + BCM2712)

timg236:
Github CI workflow test for rpi-update


6by9:
Add it as to the list of combinations in https://github.com/raspberrypi/linux/blob/rpi-6.6.y/.github/workflows/kernel-build.yml#L23-L56 and CI would build it for you.

pelwell:
This looks OK, but we'll need to be careful that the new file is created to match the latest arm64/bcm2711_defconfig - this version is already missing three options:
```
< CONFIG_GPIO_FXL6408=m
< CONFIG_DRM_PANEL_ILITEK_ILI9806E_SPI=m
< CONFIG_SND_PISOUND_MICRO=m
```

timg236:
> This looks OK, but we'll need to be careful that the new file is created to match the latest arm64/bcm2711_defconfig - this version is already missing three options:
> 
> ```
> < CONFIG_GPIO_FXL6408=m
> < CONFIG_DRM_PANEL_ILITEK_ILI9806E_SPI=m
> < CONFIG_SND_PISOUND_MICRO=m
> ```

Yes, before this is merged we need to make sure it's in sync and stays in sync. Just using it to run a few normal / RT benchmarks and check for anything weird right now.

arch/arm64/configs/bcm2711_rt_defconfig
```diff
@@ -0,0 +1,1735 @@
+CONFIG_LOCALVERSION="v8-rt"
```
pelwell: I think this needs a leading hyphen, otherwise you get `6.12.19v8-rt+`.

timg236: Good point. I lost the hyphen when I added a script to generate this for testing. Will fix, I need to test on Pi4 before merging this and offering it as an experimental build.

.github/workflows/kernel-build.yml
```diff
@@ -50,6 +50,11 @@ jobs:
             defconfig: bcm2711_defconfig
             kernel: kernel8
 
+          - name: bcm2711_rt
+            arch: arm64
+            defconfig: bcm2711_defconfig
```
pelwell: bcm2711_rt_defconfig, surely?

.github/workflows/kernel-build.yml
```diff
@@ -50,6 +50,11 @@ jobs:
...
+            kernel: kernel8
```
pelwell: Is it not better to use a different name?

pelwell (approved):

timg236:
thanks @pelwell ... I'm clearly not in the zone today !

PR #6736, opened 2025-03-22
Backport ezsound-6x8 support to 6.6.y

gordoste:
Initially, I thought that the card (which has ADC and DAC clock pins connected to the Pi's I2S clock pin) would require support for DAI links with multiple codecs, and therefore would only work in 6.12.y. However, due to complex reasons (see #6564), the card is configured with 2 separate 1:1 DAI links and can work fine in 6.6.y. This PR brings the PCM3168A driver customisations and the card's DT overlay into the 6.6.y tree.


popcornmix:
The next apt kernel will be 6.12 and we will make no more official 6.6 kernel builds, so the benefit of merging this to 6.6 is limited. Do you have a specific reason why this is useful on the 6.6 branch?

gordoste:
No, that's ok, I didn't know you'd already released the last one. If I can just tell people to do an apt-get upgrade and they'll get a 6.12 kernel, this isn't needed.

PR #6740, opened 2025-03-25
mm/mempolicy: Ignore runtime policy changes when set through cmdline

popcornmix:
Some apps like linpack call numa_setpolicy to disable numa, but that tends to have a significant performance hit for us.

Add an option to force the default numa mode (and so ignore policy changes) for now.

Consider making this the default behaviour in the future.

Use mempolicy.force_numa=1 in cmdline.txt or
echo 1 | sudo tee  /sys/module/mempolicy/parameters/force_numa

to force the default numa policy.

### DIFF
--- a/mm/mempolicy.c
+++ b/mm/mempolicy.c
@@ -85,6 +85,7 @@
 #include <linux/highmem.h>
 #include <linux/hugetlb.h>
 #include <linux/kernel.h>
+#include <linux/ktime.h>
 #include <linux/sched.h>
 #include <linux/sched/mm.h>
 #include <linux/sched/numa_balancing.h>
@@ -138,6 +139,7 @@ static struct mempolicy default_policy = {
 	.refcnt = ATOMIC_INIT(1), /* never free it */
 	.mode = MPOL_LOCAL,
 };
+static bool mempolicy_cmdline_set;
 
 static struct mempolicy preferred_node_policy[MAX_NUMNODES];
 
@@ -1629,6 +1631,14 @@ static long kernel_set_mempolicy(int mode, const unsigned long __user *nmask,
 	int err;
 
 	err = sanitize_mpol_flags(&lmode, &mode_flags);
+
+	if (mempolicy_cmdline_set) {
+		// ignore messages during boot which are expected
+		if (ktime_get_boottime_seconds() > 40)
+			pr_info("Request to set policy to %d ignored\n", mode);
+		return 0;
+	}
+
 	if (err)
 		return err;
 
@@ -3402,6 +3412,7 @@ static int __init setup_numapolicy(char *str)
 		default_policy = pol;
 		mpol_to_str(buf, sizeof(buf), &pol);
 		pr_info("NUMA default policy overridden to '%s'\n", buf);
+		mempolicy_cmdline_set = pol.mode != MPOL_DEFAULT;
 	} else {
 		pr_warn("Unable to parse numa_policy=\n");
 	}


mm/mempolicy.c
```diff
@@ -119,6 +119,10 @@
 
 #include "internal.h"
 
+/* bit field to force hotplug detection. bit0 = HDMI0 */
```
pelwell: Really?

pelwell (commented):
Either I've misunderstood or the commit message seems backwards. Where it says "force the default numa mode (and so ignore policy changes)", it looks more like "ignore policy changes (and so force the default numa mode)", although the "default" is MPOL_INTERLEAVE, is not MPOL_DEFAULT.

mm/mempolicy.c
popcornmix: Oops - updated comment.

popcornmix:
It prevents run-time switching of the mempolicy, so the default policy chosen through cmdline.txt (which we set to `numa_policy=interleave`) persists forever.

I want to prevent changes using [numactl](https://linux.die.net/man/8/numactl) (and similar) userside apis.

I'm happy to switch to a preferred wording.

pelwell:
How about:

    Add an option to preserve the default numa mode by ignoring
    policy changes for now.

popcornmix:
An alternative would be to treat any `numa_policy=` in cmdline.txt as a reason to ignore calls to `kernel_set_mempolicy` so effectively `numa_force=` is not required (assumed true when `numa_policy=` exists.

I believe that is the behaviour that makes sense for most users (if bootloader expects numa to be enabled and so sets BANKLOW=1, and we disable numa, performance is bad).

pelwell:
I think I prefer that logic.

popcornmix:
Updated. Slightly annoying we do get some messages from a normal boot:
```
[    5.475404] mempolicy: Request to set policy ignored (mode:5 lmode:5 flags:0 err:0)
[    5.475603] mempolicy: Request to set policy ignored (mode:3 lmode:3 flags:0 err:0)
[    5.476811] mempolicy: Request to set policy ignored (mode:5 lmode:5 flags:0 err:0)
[    5.476981] mempolicy: Request to set policy ignored (mode:3 lmode:3 flags:0 err:0)
[   18.969939] mempolicy: Request to set policy ignored (mode:5 lmode:5 flags:0 err:0)
[   18.970114] mempolicy: Request to set policy ignored (mode:3 lmode:3 flags:0 err:0)
```
5=MPOL_PREFERRED_MANY 3=MPOL_INTERLEAVE.
I'm not sure who is requesting MPOL_PREFERRED_MANY (a backtrace just shows it comes from `invoke_syscall`, so userspace). I'm guessing systemd.

Ignoring these is reasonable, but the spam is unfortunate.
I'm happy to remove the debug, but there is some value in the messages if it occurs later.

pelwell:
Short of adding a time check, or perhaps some check on the calling process, I don't think there's much we can do about that.

popcornmix:
Options are:
disable messages
gate the messages on a cmdline parameter, defaulted to off.
live with the spam

Any preference?

pelwell:
How about muting them for the first 30-40 seconds? That would get us past the boot.

popcornmix:
Okay, gone for the "ignore first 40 seconds" of boot option.

pelwell (approved):

popcornmix:
I've reduced the debug message a little (I think only the mode set is actually interesting).

PR #6741, opened 2025-03-25
DRM fixes for cropping P030/SAND128 on Pi5

6by9:
Addresses 2 of the glitches seen with zooming 10bit HEVC under Kodi.

Diagonal corruption when overlaying GUI with 10bit HEVC, or flickering for 8bit HEVC is still being investigated.

### DIFF
--- a/drivers/gpu/drm/vc4/vc4_plane.c
+++ b/drivers/gpu/drm/vc4/vc4_plane.c
@@ -1970,18 +1970,18 @@ static int vc6_plane_mode_set(struct drm_plane *plane,
 
 			if (fb->format->format == DRM_FORMAT_P030) {
 				/*
-				 * Spec says: bits [31:4] of the given address
-				 * should point to the 128-bit word containing
-				 * the desired starting pixel, and bits[3:0]
-				 * should be between 0 and 11, indicating which
-				 * of the 12-pixels in that 128-bit word is the
+				 * Spec says: bits [31:5] of the given address
+				 * should point to the 256-bit word containing
+				 * the desired starting pixel, and bits[4:0]
+				 * should be between 0 and 23, indicating which
+				 * of the 24-pixels in that 256-bit word is the
 				 * first pixel to be used
 				 */
 				u32 remaining_pixels = src_x % 96;
-				u32 aligned = remaining_pixels / 12;
-				u32 last_bits = remaining_pixels % 12;
+				u32 aligned = remaining_pixels / 24;
+				u32 last_bits = remaining_pixels % 24;
 
-				x_off = aligned * 16 + last_bits;
+				x_off = aligned * 32 + last_bits;
 				pix_per_tile = 96;
 			} else {
 				pix_per_tile = tile_width / fb->format->cpp[0];
--- a/drivers/gpu/drm/vc4/vc4_plane.c
+++ b/drivers/gpu/drm/vc4/vc4_plane.c
@@ -1892,6 +1892,24 @@ static int vc6_plane_mode_set(struct drm_plane *plane,
 
 	src_x = vc4_state->src_x >> 16;
 
+	/* fetch an extra pixel if we don't actually line up with the left edge. */
+	if ((vc4_state->src_x & 0xffff) && vc4_state->src_x < (state->fb->width << 16))
+		width++;
+
+	/* same for the right side */
+	if (((vc4_state->src_x + vc4_state->src_w[0]) & 0xffff) &&
+	    vc4_state->src_x + vc4_state->src_w[0] < (state->fb->width << 16))
+		width++;
+
+	/* now for the top */
+	if ((vc4_state->src_y & 0xffff) && vc4_state->src_y < (state->fb->height << 16))
+		height++;
+
+	/* and the bottom */
+	if (((vc4_state->src_y + vc4_state->src_h[0]) & 0xffff) &&
+	    vc4_state->src_y + vc4_state->src_h[0] < (state->fb->height << 16))
+		height++;
+
 	switch (base_format_mod) {
 	case DRM_FORMAT_MOD_LINEAR:
 		tiling = SCALER6_CTL0_ADDR_MODE_LINEAR;
@@ -2012,24 +2030,6 @@ static int vc6_plane_mode_set(struct drm_plane *plane,
 			      (long long)fb->modifier);
 		return -EINVAL;
 	}
-
-	/* fetch an extra pixel if we don't actually line up with the left edge. */
-	if ((vc4_state->src_x & 0xffff) && vc4_state->src_x < (state->fb->width << 16))
-		width++;
-
-	/* same for the right side */
-	if (((vc4_state->src_x + vc4_state->src_w[0]) & 0xffff) &&
-	    vc4_state->src_x + vc4_state->src_w[0] < (state->fb->width << 16))
-		width++;
-
-	/* now for the top */
-	if ((vc4_state->src_y & 0xffff) && vc4_state->src_y < (state->fb->height << 16))
-		height++;
-
-	/* and the bottom */
-	if (((vc4_state->src_y + vc4_state->src_h[0]) & 0xffff) &&
-	    vc4_state->src_y + vc4_state->src_h[0] < (state->fb->height << 16))
-		height++;
 
 	/* for YUV444 hardware wants double the width, otherwise it doesn't
 	 * fetch full width of chroma


drivers/gpu/drm/vc4/vc4_plane.c
```diff
@@ -1943,7 +1943,7 @@ static int vc6_plane_mode_set(struct drm_plane *plane,
 
 		components_per_word = fb->format->format == DRM_FORMAT_P030 ? 24 : 32;
 		starting_offset = src_x % components_per_word;
-		fetch_count = (width + starting_offset + components_per_word - 1) /
+		fetch_count = (width + starting_offset + components_per_word) /
```
popcornmix: This feels like for the common case (e.g width=1920, starting_offset=0, components_per_word=32) we will now fetch 61 "words" rather than 60. And 60 is all that is needed.

That is both increasing sdram bandwidth and the potential for reading something you shouldn't.

I think we want to make this only fetch the extra word in the cases we know it goes wrong (perhaps when `starting_offset != 0`, or even `>= components_per_word-  2`).

Note: the fetcher spec does contain details on calculating this for different formats.

6by9: I had gone with an `if(!i && last_bits == 22) fetch_count++` in the DRM_FORMAT_P030 clause below this, and that worked. I then determined that this was equivalent, but it is only in some situations, and we're also unnecessarily changing the behaviour for 8bit.

I have a suspicion that the full error condition is where the last wanted pixel ends on the last pixel of a column, at least that is the main situation that we were hitting (first one was source rectangle offset (142,79), sized 3554x1998. 3554+142 = 3696, which is exactly 154 columns). I'll give it some more poking.

I'm not clear whether this is actually configuring the fetcher, or something else in the pipeline. I was reading the description of "Pointer Words 1-2 – Stripe and MAP8 Formats" in the dlist docs. The fetcher docs appear to be referencing reading into the UPM store.
I believe the current driver implements the allocations as documented, but the lockup demonstrates that something in there is wrong.

6by9: Updated to revert back to `if(!i && last_bits == 22) fetch_count++`
That still works with zooming in and out using Kodi, but avoids touching most unnecessary cases. It will need some more analysis before being able to upstream the fix.

6by9:
Diagonal corruption is something else, and not apparently related to background fill (tested by permanently enabling it). Suspect it is system load, but it is weird that it is only when updating the GUI layer that it causes problems.
Solving that one will have to wait.

HiassofT:
In the initial version of this PR the Interstellar trailer worked fine on all zoom levels, on a 1920x1080p60 display, with the recent however the "blank screen" issue at zoom level 1.11 is back (as before the change).

Here's kmsprint output with zoom level 1.11 and display blanking out:
```
rpi5:~ # kmsprint 
Connector 0 (33) HDMI-A-1 (connected)
  Encoder 0 (32) TMDS
    Crtc 2 (92) 1920x1080@60.00 148.500 1920/88/44/148/+ 1080/4/5/36/+ 60 (60.00) D 
      Plane 2 (81) fb-id: 684 (crtcs: 2) 0,0 3840x1608 -> -106,94 2132x894 (XR24 AR24 AB24 XB24 RG16 BG16 AR15 XR15 RG24 BG24 YU16 YV16 YU24 YV24 YU12 YV12 NV12 NV21 NV16 NV61 P030 XR30 AR30 AB30 XB30 RGB8 BGR8 XR12 AR12 XB12 AB12 BX12 BA12 RX12 RA12)
        FB 684 0x0 (�j�
      Plane 4 (105) fb-id: 681 (crtcs: 1 2 3) 0,0 1920x1080 -> 0,0 1920x1080 (XR24 AR24 AB24 XB24 RG16 BG16 AR15 XR15 RG24 BG24 YU16 YV16 YU24 YV24 YU12 YV12 NV12 NV21 NV16 NV61 P030 XR30 AR30 AB30 XB30 RGB8 BGR8 XR12 AR12 XB12 AB12 BX12 BA12 RX12 RA12)
        FB 681 1920x1080 AR24
Connector 1 (42) HDMI-A-2 (disconnected)
  Encoder 1 (41) TMDS
```

6by9:
Which version of the Interstellar trailer is it? 4k 10bit HEVC being downscaled to 1080p60?

HiassofT:
The trailer is 4k HEVC 8bit. I'm using a 1920x1200 monitor here on my test setup, kodi GUI resolution set to 1920x1080p60 which is also used while playing this sample (the monitor doesn't support 24p)
```
$ ls -l Interstellar-divx-hevc-uhd-51ch-4K.mkv 
-rwxrw-r-- 2 hias hias 403766800 Jun 18  2019 Interstellar-divx-hevc-uhd-51ch-4K.mkv
$ md5sum Interstellar-divx-hevc-uhd-51ch-4K.mkv 
ad611d6067ff79b5b45bd3c571859c23  Interstellar-divx-hevc-uhd-51ch-4K.mkv
$ mediainfo Interstellar-divx-hevc-uhd-51ch-4K.mkv 
General
Unique ID                                : 86748321257335288719775507338167402396 (0x41431E73BC456F9EC4B2746B21093F9C)
Complete name                            : Interstellar-divx-hevc-uhd-51ch-4K.mkv
Format                                   : Matroska
Format version                           : Version 1
File size                                : 385 MiB
Duration                                 : 2 min 24 s
Overall bit rate                         : 22.3 Mb/s
Frame rate                               : 24.000 FPS
Writing application                      : DivXMKVMux 9.9.12.5441
Writing library                          : libDivXMediaFormat 4.0.0.0578

Video
ID                                       : 1
Format                                   : HEVC
Format/Info                              : High Efficiency Video Coding
Format profile                           : Main@L5.1@Main
Codec ID                                 : V_MPEGH/ISO/HEVC
Duration                                 : 2 min 24 s
Bit rate                                 : 21.2 Mb/s
Width                                    : 3 840 pixels
Height                                   : 1 608 pixels
Display aspect ratio                     : 2.40:1
Frame rate mode                          : Constant
Frame rate                               : 24.000 FPS
Standard                                 : Component
Color space                              : YUV
Chroma subsampling                       : 4:2:0
Bit depth                                : 8 bits
Bits/(Pixel*Frame)                       : 0.143
Stream size                              : 366 MiB (95%)
Language                                 : English
Default                                  : Yes
Forced                                   : No
Color range                              : Full

Audio
ID                                       : 2
Format                                   : AC-3
Format/Info                              : Audio Coding 3
Commercial name                          : Dolby Digital
Codec ID                                 : A_AC3
Duration                                 : 2 min 24 s
Bit rate mode                            : Constant
Bit rate                                 : 640 kb/s
Channel(s)                               : 6 channels
Channel layout                           : L R C LFE Ls Rs
Sampling rate                            : 48.0 kHz
Frame rate                               : 31.250 FPS (1536 SPF)
Bit depth                                : 16 bits
Compression mode                         : Lossy
Stream size                              : 11.0 MiB (3%)
Language                                 : English
Service kind                             : Complete Main
Default                                  : Yes
Forced                                   : No
```

6by9:
Thanks. I hadn't seen a failure at all with 8bit content, but that does seem to provoke one.

Increasing fetch_count by one for 8bit `(x_offset & 31) == 30` covers some situations, but there's another case that is still failing at x1.58, which because it is 4k being scaled to 1080p is a greater crop/scale than all my 4k on 4k tests.

6by9:
I've found the actual cause.

https://github.com/raspberrypi/linux/blob/rpi-6.12.y/drivers/gpu/drm/vc4/vc4_plane.c#L2016-L2032 increases the source size if fractionally cropping a pixel at the edge. There are conditions for SAND where that will require an extra word to be fetched, however `fetch_count` had already been computed. Fetching insufficient words results in a total stall.

drivers/gpu/drm/vc4/vc4_plane.c
```diff
@@ -1983,6 +2001,7 @@ static int vc6_plane_mode_set(struct drm_plane *plane,
 
 				x_off = aligned * 32 + last_bits;
 				pix_per_tile = 96;
+
```
popcornmix: spurious newline

popcornmix:
Nice this feels more satisfactory that previous version.

drivers/gpu/drm/vc4/vc4_plane.c
6by9: Fixed

HiassofT:
Thanks a lot, testing with your latest update looks good here :+1:

Issue #6647, opened 2025-02-04
CM5 without wifi hangs on reboot


nbuchwitz:
### Describe the bug

We stumbled over an issue where all CM5 without wifi seem to hang when rebooted. After some waiting the reboot is completed whereas all CM5 with wifi show no such error (same base boards, same software).  As is some care cases the reboot even worked on CM5 without wifi I started to debug it further.

When reboot hangs:
```
Dec 01 13:28:51 RevPi systemd[1]: Shutting down.
Dec 01 13:28:51 RevPi systemd[1]: Using hardware watchdog 'Broadcom BCM2835 Watchdog timer', version 0, device /dev/watchdog0
Dec 01 13:28:51 RevPi systemd[1]: Watchdog running with a hardware timeout of 10min.
Dec 01 13:28:51 RevPi kernel: watchdog: watchdog0: watchdog did not stop!
Dec 01 13:28:51 RevPi systemd-shutdown[1]: Using hardware watchdog 'Broadcom BCM2835 Watchdog timer', version 0, device /dev/watchdog0
Dec 01 13:28:52 RevPi systemd-shutdown[1]: Watchdog running with a hardware timeout of 10min.
Dec 01 13:28:52 RevPi systemd-shutdown[1]: Syncing filesystems and block devices.
Dec 01 13:28:52 RevPi systemd-shutdown[1]: Sending SIGTERM to remaining processes...
Dec 01 13:28:52 RevPi systemd-journald[167]: Received SIGTERM from PID 1 (systemd-shutdow).
Dec 01 13:28:52 RevPi systemd-journald[167]: Journal stopped
```

When reboot works immediately:
```
Dec 01 13:29:57 RevPi136828 systemd[1]: Shutting down.
Dec 01 13:29:58 RevPi136828 systemd[1]: Using hardware watchdog 'Broadcom BCM2835 Watchdog timer', version 0, device /dev/watchdog0
Dec 01 13:29:58 RevPi136828 systemd[1]: Watchdog running with a hardware timeout of 10min.
Dec 01 13:29:58 RevPi136828 kernel: mmc1: Failed to initialize a non-removable card
Dec 01 13:29:58 RevPi136828 kernel: watchdog: watchdog0: watchdog did not stop!
Dec 01 13:29:58 RevPi136828 systemd-shutdown[1]: Using hardware watchdog 'Broadcom BCM2835 Watchdog timer', version 0, device /dev/watchdog0
[...]

PR #6747, opened 2025-03-28
mmc: Speed up reboot with an absent card

pelwell:
On devices with no card detect mechanism, the host ends up polling for the insertion of a card. This polling happens at multiple frequencies and with many steps (SDIO, SD, MMC), some of which involve timeouts. If a reboot is requested during one of those scans, it will stall, pointlessly, for up to a minute while it completes.

Attempt to short circuit the rescan if the MMC interface is being removed.

See: https://github.com/raspberrypi/linux/issues/6647

### DIFF
--- a/drivers/mmc/core/core.c
+++ b/drivers/mmc/core/core.c
@@ -399,9 +399,15 @@ void mmc_wait_for_req_done(struct mmc_host *host, struct mmc_request *mrq)
 	struct mmc_command *cmd;
 
 	while (1) {
-		wait_for_completion(&mrq->completion);
+		if (!host->rescan_disable)
+			wait_for_completion(&mrq->completion);
 
 		cmd = mrq->cmd;
+
+		if (host->rescan_disable) {
+			cmd->error = -ENOENT;
+			break;
+		}
 
 		if (!cmd->error || !cmd->retries ||
 		    mmc_card_removed(host->card))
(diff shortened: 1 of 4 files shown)


pelwell:
N.B.:
1. This is Friday afternoon patch, posted so that people can take a look at it.
2. Even with this patch there can be stalls of up to 10 seconds - however long it takes for the current outstanding command to timeout.

hasan-akbulak:
Testing with cm5 during startup still getting stuck on systemd-journald.socket for 30-60 seconds and dtparam=sd=off is not detected. seems to be unsuccessfully halting/rebooting without sd card present. still gets stuck for amount a minute. And twice encountering a error like this:

Afbeelding van WhatsApp op 2025-03-31 om 10 07 15_bb5dab3d

Problem could have occured from the system being unstable and hanging often even when using simple commands like dmesg.

pelwell:
That looks like a new crash - I've added a fix for that and another one to this PR, as well as hopefully reducing the maximum stall to 5 seconds.

Wait another 40 minutes before trying to update.

hasan-akbulak:
still stalling on systemd-journal.socket when booting

ssh takes about 30 seconds to connect. 

however when booted stalls are less than 5 seconds for commands and halting and when actually halting/rebooting no stalls
after disabling wifi bt and sd card ofc.

still sometimes small hangs with commands for 2 seconds during halt/shutdown for example trigger happy when pressing power button

Issue #6752, opened 2025-03-31
Cannot trace syscalls because CONFIG_FTRACE_SYSCALLS is not set


bukatlib:
### Describe the bug

Please consider to enable CONFIG_FTRACE_SYSCALLS kernel option to enable syscall tracing via 'perf trace', currently only raw syscalls can be observed.

  $ perf list | grep syscall
  raw_syscalls:sys_enter                             [Tracepoint event]
  raw_syscalls:sys_exit                              [Tracepoint event]

Performance impact should be close to zero if tracing is disabled (nop instructions).

### Steps to reproduce the behaviour

perf stat -e 'syscalls:sys_enter_\*'
event syntax error: 'syscalls:sys_enter_\*'
                     \___ unknown tracepoint

### Device (s)

Raspberry Pi 5

### System

Raspberry Pi reference 2024-07-04
Generated using pi-gen, https://github.com/RPi-Distro/pi-gen, 48efb5fc5485fafdc9de8ad481eb5c09e1182656, stage5

popcornmix:
This seems reasonable to me.

pelwell:
See https://github.com/raspberrypi/linux/pull/6759. About 40 minutes from now you should be able to install a trial build using `sudo rpi-update pulls/6759`.

bukatlib:
Thank you, I tested pulls/6759 and it works:

```
$ perf trace -e 'syscalls:sys_enter_open*' -a -- sleep 3
     0.000 sleep/2393 syscalls:sys_enter_openat(dfd: CWD, filename: "", flags: RDONLY|CLOEXEC)
     0.053 sleep/2393 syscalls:sys_enter_openat(dfd: CWD, filename: "", flags: RDONLY|CLOEXEC)
     0.340 sleep/2393 syscalls:sys_enter_openat(dfd: CWD, filename: "", flags: RDONLY|CLOEXEC)

$ perf stat -e 'syscalls:sys_enter_open*' -a -- sleep 3

 Performance counter stats for 'system wide':

                 0      syscalls:sys_enter_openat2                                            
                 3      syscalls:sys_enter_openat                                             
                 0      syscalls:sys_enter_open_tree                                          
                 0      syscalls:sys_enter_open_by_handle_at                                      

       3,001874110 seconds time elapsed

$ perf list -v  | grep syscalls | wc -l
620
```

[...]

PR #6759, opened 2025-04-01
configs: Add FTRACE_SYSCALLS=y

pelwell:
FTRACE_SYSCALLS enables the syscall entry and exit events to be traced.

See: https://github.com/raspberrypi/linux/issues/6752


pelwell:
This option adds 170kB to the uncompressed ARMv7 kernel - the one which is most size sensitive, since the memories are smallest:

| Build | File | size |
|--------|--------|--------|
| Before | Image | 16229536 |
|  | kernel.img | 7316892 | 
| After | Image | 16408704 |
| | kernel.img | 7371224 | 

Are you happy with this, @popcornmix?

popcornmix:
The size increase is more than I expected. Perhaps this needs more thought to see exactly how many users it would benefit.

I did ask a colleague who has been using perfetto for tracing system activity if this would be useful and he says perfetto already supports tracing syscalls using these events:
```
$ sudo cat /sys/kernel/debug/tracing/available_events | grep syscalls
raw_syscalls:sys_exit
raw_syscalls:sys_enter
```

popcornmix (approved):

Issue #5694, opened 2023-11-03
 Raspberry Pi 5 Bookworm ADS7846 can not rotate touch


DRAgon734465502:
### Describe the bug

I use a Raspberry Pi 5 or 4B with the lastest Bookworm OS
and with a Resistance touch using XPT2046
I add touch calibration at 99-calibration.conf
I add dtoverlay=ads7846 at config.txt

Then touch work but “99-calibration.conf” not work ， it not keep Touch Reverse
I can not rotate Touch direction by 99-calibration.conf
this same configuration work at 3B （same SD card with same LCD and same SD card OS）But not work at 4B and 5

please tell me how can I solve this problem？

Thanks !!!

### Steps to reproduce the behaviour

At config.txt
dtoverlay=ads7846,cs=1,penirq=25,penirq_pull=2,speed=50000,keep_vref_on=0,swapxy=0,pmax=255,xohms=150,xmin=200,xmax=3900,ymin=200,ymax=3900

sudo apt-get install xserver-xorg-input-evdev xinput-calibrator
sudo cp -rf /usr/share/X11/xorg.conf.d/10-evdev.conf /usr/share/X11/xorg.conf.d/45-evdev.conf
sudo nano /usr/share/X11/xorg.conf.d/99-calibration.conf
Section "InputClass"
        Identifier      "calibration"
        MatchProduct    "ADS7846 Touchscreen"
        Option  "Calibration"   "117 3992 3743 156"
        Option  "SwapAxes"      "1"
        Option "EmulateThirdButton" "1"
        Option "EmulateThirdButtonTimeout" "1000"
        Option "EmulateThirdButtonMoveThreshold" "300"
EndSection

### Device (s)

Raspberry Pi 4 Mod. B

### System

At config.txt
dtoverlay=ads7846,cs=1,penirq=25,penirq_pull=2,speed=50000,keep_vref_on=0,swapxy=0,pmax=255,xohms=150,xmin=200,xmax=3900,ymin=200,ymax=3900

sudo apt-get install xserver-xorg-input-evdev xinput-calibrator
sudo cp -rf /usr/share/X11/xorg.conf.d/10-evdev.conf /usr/share/X11/xorg.conf.d/45-evdev.conf
sudo nano /usr/share/X11/xorg.conf.d/99-calibration.conf
Section "InputClass"
        Identifier      "calibration"
        MatchProduct    "ADS7846 Touchscreen"
        Option  "Calibration"   "117 3992 3743 156"
[...]

PR #6760, opened 2025-04-02
ads7846 touch overlay fixups

6by9:
From comments on #5694 add in all the overrides to ads7846, and tidy them up on piscreen.

### DIFF
--- a/arch/arm/boot/dts/overlays/README
+++ b/arch/arm/boot/dts/overlays/README
@@ -602,6 +602,8 @@ Params: cs                      SPI bus Chip Select (default 1)
         pmax                    Maximum reported pressure value (default 65535)
         xohms                   Touchpanel sensitivity (X-plate resistance)
                                 (default 400)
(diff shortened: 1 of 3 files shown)


pelwell:
Given that the two commits are in the same PR and will be merged simultaneously by the same person, I don't see the point in setting up a problem in the first commit and solving it in the second. I happened to read the second commit message first and was left thinking that you were fixing an existing problem.

Otherwise, approved.

6by9:
> Given that the two commits are in the same PR and will be merged simultaneously by the same person, I don't see the point in setting up a problem in the first commit and solving it in the second. I happened to read the second commit message first and was left thinking that you were fixing an existing problem.

I don't think I have.
One patch is on ads7846-overlay, and the other is on piscreen-overlay. They both use the same touch controller, but each had their own deficiencies.

pelwell:
Sorry, you're right - I read it as being all about the ads7846 touch overlay.

PR #6762, opened 2025-04-02
Show process name in set_mempolicy() ignored message

trejan:


popcornmix:
Looks good to me. Can you add a signed off by message?
(`git commit -s --amend` and `git push -f` should do it)

pelwell (approved):

trejan:
Looks like they're getting rid of get_task_comm() and using current->comm directly. I'll amend it.

trejan:
Ignore my previous comment. The get_task_comm() change is in 6.14.

pelwell:
Good to know - we'll cross that bridge when we come to it.

PR #6764, opened 2025-04-03
Added waveshare 7.0inch h dsi screen support

waveshare:
Product link：https://www.waveshare.com/7inch-dsi-lcd-h.htm


6by9 (approved):

6by9:
Technically a Signed-off-by needs to be by an identifiable individual, but as we'll never upstream this driver I don't care.
https://www.kernel.org/doc/html/latest/process/submitting-patches.html#developer-s-certificate-of-origin-1-1

PR #6773, opened 2025-04-08
[RFC] Staging: Adding Sfera Labs drivers

giampiero7:
Hi,

This pull request represents an initial step toward integrating drivers for our boards and modules directly into the kernel, with the goal of improving distribution and long-term maintainability.

I'm fully aware this is not the ideal or final structure, but I wanted to get the conversation started with something concrete to gather feedback and suggestions on how best to proceed.

Our devices combine multiple functionalities, which makes it difficult to fit them under existing driver categories. To bootstrap the discussion, I've added a new `drivers/staging/sferalabs` directory.

In this first commit:
- I've included the driver for an upcoming product, *Iono Pi v3*.
- I've added a `commons` directory with shared components that will be reused across future drivers.

All drivers expose a `sysfs` interface to access the board/module functionalities. The Iono Pi v3 driver also registers RTC and Watchdog device nodes.

For reference, the driver (distributed as an out-of-tree kernel module) of the current Iono Pi version can be found here:  
https://github.com/sfera-labs/iono-pi-kernel-module

Before diving into formatting and implementation specifics, I'd really appreciate your thoughts on whether this structure and general approach would be acceptable.

Looking forward to your feedback!

Thank you,    
Giampiero

### DIFF
--- a/drivers/staging/Kconfig
+++ b/drivers/staging/Kconfig
@@ -64,4 +64,6 @@ source "drivers/staging/fieldbus/Kconfig"
 
 source "drivers/staging/vme_user/Kconfig"
 
+source "drivers/staging/sferalabs/Kconfig"
+
 endif # STAGING
--- a/drivers/staging/Makefile
+++ b/drivers/staging/Makefile
@@ -21,3 +21,4 @@ obj-$(CONFIG_GREYBUS)		+= greybus/
 obj-$(CONFIG_BCM2835_VCHIQ)	+= vc04_services/
 obj-$(CONFIG_XIL_AXIS_FIFO)	+= axis-fifo/
 obj-$(CONFIG_FIELDBUS_DEV)     += fieldbus/
+obj-y                           += sferalabs/
--- a/drivers/staging/sferalabs/Kconfig
+++ b/drivers/staging/sferalabs/Kconfig
@@ -1,0 +1,3 @@
+# SPDX-License-Identifier: GPL-2.0
+
+source "drivers/staging/sferalabs/ionopi-v3/Kconfig"
--- a/drivers/staging/sferalabs/Makefile
+++ b/drivers/staging/sferalabs/Makefile
@@ -1,0 +1,2 @@
+# SPDX-License-Identifier: GPL-2.0
+obj-$(CONFIG_IONO_PI_V3)     += ionopi-v3/
--- a/drivers/staging/sferalabs/commons/atecc/atecc.c
+++ b/drivers/staging/sferalabs/commons/atecc/atecc.c
@@ -1,0 +1,124 @@
+#include "atecc.h"
+
+#include <linux/delay.h>
+#include <linux/i2c.h>
+#include <linux/module.h>
+#include <linux/version.h>
+
+struct AteccBean {
+  uint8_t serialNumber[9];
+  bool probed;
+};
+
+static struct AteccBean _atecc = {
+    .probed = false,
+};
+
+static void _getCRC16LittleEndian(size_t length, const uint8_t *data,
+                                  uint8_t *crc_le) {
+  size_t counter;
+  uint16_t crc = 0;
+  uint16_t polynom = 0x8005;
+  uint8_t shift;
+  uint8_t data_bit, crc_bit;
+
+  for (counter = 0; counter < length; counter++) {
+    for (shift = 0x01; shift > 0x00; shift <<= 1) {
+      data_bit = (data[counter] & shift) ? 1 : 0;
+      crc_bit = crc >> 15;
+      crc <<= 1;
+      if (data_bit != crc_bit) {
+        crc ^= polynom;
+      }
+    }
+  }
+  crc_le[0] = (uint8_t)(crc & 0x00FF);
+  crc_le[1] = (uint8_t)(crc >> 8);
+}
+
+#if LINUX_VERSION_CODE >= KERNEL_VERSION(6, 5, 0)
+static int _atecc_i2c_probe(struct i2c_client *client) {
+#else
+static int _atecc_i2c_probe(struct i2c_client *client,
+                            const struct i2c_device_id *id) {
+#endif
+  uint8_t i;
+  int ret = -1;
+  uint8_t i2c_response[35];
+  uint8_t crc_le[2];
+
+  /*
+   * 0x03 = normal command
+   * 0x07 = total bytes for CRC generation (2 CRC bytes included)
+   * 0x02 = read operation
+   * 0x80 = read 32 bytes from configuration memory area
+   * 0x00 = configuration memory address part 1
+   * 0x00 = configuration memory address part 2
+   * 0x09 = CRC byte 1 in little endian format
+   * 0xAD = CRC byte 2 in little endian format
+   */
+  uint8_t cmd_read_sn[8] = {0x03, 0x07, 0x02, 0x80, 0x00, 0x00, 0x09, 0xAD};
+  uint8_t cmd_wake = 0x00;
+
+  for (i = 0; i < 10; i++) {
+    i2c_master_send(client, &cmd_wake, 1);
+    msleep(1);
+    if (i2c_master_send(client, cmd_read_sn, 8) == 8) {
+      msleep(1);
+      if (i2c_master_recv(client, i2c_response, 35) == 35) {
+        _getCRC16LittleEndian(33, i2c_response, crc_le);
+        if (crc_le[0] == i2c_response[33] && crc_le[1] == i2c_response[34]) {
+          memcpy(&_atecc.serialNumber[0], &i2c_response[1], 4);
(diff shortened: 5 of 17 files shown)


nbuchwitz:
Hi Giampiero,

I might be a bit biased, but wanted to share my thoughts on this too.

From the skeleton it looks like lots of it would be better suited in a board support package with proper userspace interfaces / tools for the end user (like the atecc security ship or the configuration of the pfc2131 watchdog). This will massively reduce the maintenance efforts and the end user would benefit from the upstream maintained device drivers. Or are there any use cases which don't work with the traditional approach of a device tree and the usual kernel driver? 

Imho a custom kernel driver only makes sense if there is a specific protocol  or (hardware) interface to be implemented which is not already there. Wrapping existing userspace interfaces into sysfs is the wrong approach. 

If you're interested I'm more than happy to discuss with you and Ulderico how me manage our product related modifications and make it easy for our customers to integrate it with native Raspberry Pi OS :-)

Nicolai

giampiero7:
Hello Nicolai,  
Thank you very much for your feedback! Much appreciated.

I probably should have started with a simpler and more fitting case, and should have emphasized more clearly not to focus on the implementation at this stage.

Let me outline the general goals and cornerstones:

- Many of our boards include MCUs with custom firmware and protocols, so a kernel driver is the natural path in these cases.  
- The sysfs interface is a very handy one, easy to access from userspace under any context, and much appreciated by our users.  
- Having a unified interface for functionalities that would otherwise be accessible through different subsystems is a nice-to-have. I understand this might not be the orthodox approach, and I'm open to reconsidering it.  
- Having a consistent interface across our different products, with a clear and straightforward structure, is also a nice-to-have, and something our users value.

When higher-level application logic is needed, we do provide userspace utilities, but only when it truly falls outside the scope of a kernel module. We prefer not to, when not strictly necessary, in order to maintain a unified interface.

Finally, we want our users to be able to use the system they’re familiar with (native RPi OS or any of its "mods") with as little intrusion as possible.

We’re aware this approach requires more maintenance effort on our side, but we're okay with that if it ultimately benefits our users.

So, as a starting point, perhaps we can focus on defining where to place the core driver for a specific product and how to structure it — e.g., core driver in `drivers/mfd` and subsystems in their respective directories such as `drivers/iio`, `drivers/watchdog`, etc. — and where, if possible, to place a common library.

In this specific case, the RTC and watchdog functionalities are exposed via the standard `/dev/rtcN` and `/dev/watchdogN` interfaces. Both are implemented using the PCF2131 and its interrupt pins, which made using the upstream driver unfeasible — though I believe with a small patch, it could become compatible. That, however, might be better left for a later discussion.

I chose to start with this driver because it includes many of the edge cases, which I thought would help address all the relevant points throughout this discussion.

Finally, I’d be happy to have a chat about all this — feel free to reach out at [giampiero@sferalabs.cc](mailto:giampiero@sferalabs.cc).

Thank you!

giampiero7:
Hello Nicolai,
Thank you very much for your feedback! Much appreciated.

I probably should have started with a simpler and more fitting case, and should have emphasized more clearly not to focus on the implementation at this stage.

Let me outline the general goals and cornerstones:

- Many of our boards include MCUs with custom firmware and protocols, so a kernel driver is the natural path in these cases.
- The sysfs interface is a very handy one, easy to access from userspace under any context, and much appreciated by our users.
- Having a unified interface for functionalities that would otherwise be accessible through different subsystems is a nice-to-have. I understand this might not be the orthodox approach, and I'm open to reconsidering it.
- Having a consistent interface across our different products, with a clear and straightforward structure, is also a nice-to-have, and something our users value.

When higher-level application logic is needed, we do provide userspace utilities, but only when it truly falls outside the scope of a kernel module. We prefer not to, when not strictly necessary, in order to maintain a unified interface.

Finally, we want our users to be able to use the system they’re familiar with (native RPi OS or any of its "mods") with as little intrusion as possible.

We’re aware this approach requires more maintenance effort on our side, but we're okay with that if it ultimately benefits our users.

So, as a starting point, perhaps we can focus on defining where to place the core driver for a specific product and how to structure it — e.g., core driver in `drivers/mfd` and subsystems in their respective directories such as `drivers/iio`, `drivers/watchdog`, etc. — and where, if possible, to place a common library.

In this specific case, the RTC and watchdog functionalities are exposed via the standard `/dev/rtcN` and `/dev/watchdogN` interfaces. Both are implemented using the PCF2131 and its interrupt pins, which made using the upstream driver unfeasible — though I believe with a small patch, it could become compatible. That, however, might be better left for a later discussion.

I chose to start with this driver because it includes many of the edge cases, which I thought would help address all the relevant points throughout this discussion.

Finally, I’d be happy to have a chat about all this — feel free to reach out at [giampiero@sferalabs.cc](mailto:giampiero@sferalabs.cc).

Thank you!

6by9:
All kernel drivers should follow the kernel coding style - https://www.kernel.org/doc/html/latest/process/coding-style.html
Indentation not using tabs and the use of camel case for function names are the 2 main issues here.
Please use scripts/checkpatch.pl to check for the obvious issues.
Whilst you've said you weren't worried about formatting issues, it makes life harder reviewing code when it's not the format expected.

In tree drivers should never have 
```
#if LINUX_VERSION_CODE >= KERNEL_VERSION(6, 5, 0)
```
or similar, as they are inherently in a tree of some version and are expected to work there.

Nothing I see in this PR actually defines what hardware you have attached, and you haven't included a dtoverlay that would give that information. The overlay in the repo you link to also appears to be missing any reference to half the drivers here.

Where existing drivers exist then you at least need to provide a very good justification for not using them.

- Your PCF2131 driver appears to be some mods to the existing PCF2127 driver that supports PCF2131. What needed to change and why?
- What is the atecc? Microchip's Crypto chip https://www.microchip.com/en-us/product/ATECC608A ? There's already a driver in the kernel for that https://github.com/torvalds/linux/blob/master/drivers/crypto/atmel-ecc.c
- All of commons/gpio/gpio.c appears to be accessor functions that look like the client should just be using libgpiod.
- There's a driver for MCP320x ADCs in drivers/iio/adc/mcp320x. Use libiio.
- The functions in misc.c almost certainly already have standard versions in the kernel. I haven't looked into why you're trying to mangle time values, but something feels quite wrong there.
- Drivers for specific ICs should be independent and in the relevant subsystem directory.
- All files should have an SPDX licence line, and Copyright header.
- If you need a module driver (currently I don't see a need for one), then it shouldn't be implementing an MCP3462 ADC and MAX4896 relay driver. Again MCP3462 already has a driver in drivers/iio/adc/mcp3564. MAX4896 as an 8 output relay driver should be a GPIO driver (I don't see an existing driver for it).

> The sysfs interface is a very handy one, easy to access from userspace under any context, and much appreciated by our users.

If there are defined kernel APIs to access the functionality, then you should be using them and not abusing sysfs.

Requesting merging into the Raspberry Pi kernel should really be a forerunner to getting drivers merged upstream. Overlays, DT files, and possibly minor fixes to drivers are one thing, but not wholesale dumps of large drivers. Otherwise you're just trying to pass the support burden onto Raspberry Pi.

giampiero7:
Thank you for your feedback, @6by9!
I realize now that my top-down approach didn't quite work as intended... I really wasn't looking for a code review at this stage, the code was just meant to serve as a reference point for later discussions. Apologies for the confusion and any inconvenience caused.

Let me start over: since these devices are Raspberry Pi-centric, would it be appropriate to upstream their drivers here, once they're properly structured, split into subsystems, and reusing existing drivers where possible?

Thank you very much!
Giampiero

6by9:
Minimal additional drivers which wouldn't be acceptable upstream would probably be accepted here. eg we have a number of [audio drivers](https://github.com/raspberrypi/linux/tree/rpi-6.12.y/sound/soc/bcm) which are basically just configuration for the I2S hardware. 

From what I see, most of what you've presented should condense down to 
- a couple of drivers that would be acceptable upstream (GPIO and ADC drivers are pretty trivial, and it's normally the dt-bindings that are the hardest work)
- some patches to existing upstream drivers
- a dtoverlay defining all the bits
- if wiegand.c is implementing the [Wiegand card reader interface](https://en.wikipedia.org/wiki/Wiegand_interface) then probably a module for that as I suspect it would not be acceptable upstream without a lot of effort in defining APIs etc (I do wonder if it can be implemented in userspace instead, but the timing accuracy looks quite tight. Depending on exactly what data it produces, the input subsystem may be better to act as a keyboard).

If you want to bounce a block diagram with basic details of ICs implementing functionality to us, then we can perhaps advise better. Either post it here, or use "applications at raspberrypi.com" and include a reference to here / mention 6by9 so that they can direct it my way.

giampiero7:
Thank you @6by9, this is the kind of input I was hoping to get :)
Maybe it's better to start from a more straightforward driver, for which schematics and documentation are already public.
In the upcoming weeks I'll craft a new pull request and we'll start from there.
Thank you again!

PR #6789, opened 2025-04-15
drm/vc4: tests: Patch up the vc4 kunit tests

pelwell:
Commit [1] removed the dynamic allocation of drm_modeset_acquire_ctx structures, with clients expected to allocate them by other means, such as on the stack. This usage model works less well when tests are split into separate initialisation and run phases, because a ctx allocated on the stack in the init function will go out of scope and likely be overwritten by the run function(s).

As kunit tests seem to be single threaded, make ctx static to avoid the issue.

[1] commit 30188df0c387 ("drm/tests: Drop drm_kunit_helper_acquire_ctx_alloc()")

### DIFF
--- a/drivers/gpu/drm/vc4/tests/vc4_test_lbm_size.c
+++ b/drivers/gpu/drm/vc4/tests/vc4_test_lbm_size.c
@@ -279,7 +279,7 @@ static struct kunit_case vc4_lbm_size_tests[] = {
 
 static int vc4_lbm_size_test_init(struct kunit *test)
 {
-	struct drm_modeset_acquire_ctx *ctx;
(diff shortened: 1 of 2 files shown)


pelwell:
FYI, @mripard.

mripard:
That doesn't work either. kunit isn't single-threaded, and lockdep will complain about it. You really need to initialize the state and context in the test function itself, see https://github.com/torvalds/linux/commit/7e0351ae91ed2b6178abbfae96c3c6aaa1652567

pelwell:
It's nice of you to finally respond. This was sufficient to pass the tests, but we can adopt the more complex fix in the future.

PR #6791, opened 2025-04-15
drm/vc4: plane: Increase UPM allocation size for YUV444

6by9:
YUV444 support isn't officially supported by the hardware, but worked if you told it the image was YUV422 with double the width and altered chroma scaling.

Adding BCM2712 support gained a fetcher memory (UPM). The code handling UPM allocations didn't have a case for YUV444, so only allocated based on the base width, and therefore underflowed.

Increase the UPM allocation size for the luma plane of YUV444 to match.

Fixes: 076eedaf762f ("drm/vc4: hvs: Add support for BCM2712 HVS")


6by9:
@cillian64

pelwell (approved):

cillian64:
Confirmed the latest patches fixes the issue.  Thanks!

PR #6793, opened 2025-04-15
media: i2c: imx219: Restore the 1920x1080 to using a 1:1 PAR

6by9:
Commit 0af46fbc333d ("media: i2c: imx219: Calculate crop rectangle dynamically") meant that the 1920x1080 switched from using no binning to using vertical binning but no horizontal binning.

Restore the original behaviour by ensuring the two binning settings are the same.

Fixes: 0af46fbc333d ("media: i2c: imx219: Calculate crop rectangle dynamically")


6by9:
https://forums.raspberrypi.com/viewtopic.php?p=2309972

PR #6795, opened 2025-04-15
dtoverlays: Create Pi5 variant of tc358743 overlay

6by9:
The Pi0-4 variant wants to change the compatible string for the CSI2 node to "brcm,bcm2835-unicam-legacy" to disable Media Controller by default, which would switch to Unicam when Pi5 needs the CFE driver.

Split the overlay into an include and base pi0-4 and pi5 overlay files


arch/arm/boot/dts/overlays/tc358743-pi5-overlay.dts
```diff
@@ -0,0 +1,10 @@
+// SPDX-License-Identifier: GPL-2.0-only
+// Definitions for Toshiba TC358743 HDMI to CSI2 bridge on VC I2C bus. Pi5 variant
+/dts-v1/;
+/plugin/;
+
+#include "tc358743.dtsi"
+
+/ {
+	compatible = "brcm,bcm2835";
```
pelwell: Can you make this `brcm,bcm2712`?

pelwell:
FYI I'd have been happy with you including the existing overlay directly into the new pi5 variant and patching it up, but this way is also okay.

arch/arm/boot/dts/overlays/tc358743-pi5-overlay.dts
6by9: Done.

pelwell (approved):

acardoso-plana (approved):

pelwell:
Thanks for the nudge - GitHub doesn't tend to give notifications about pushes to a PR.

PR #6796, opened 2025-04-16
drm/vc4: plane: Use nearest neighbour filter with YUV444 workaround

6by9:
As a follow-up to commit ef79eea9e4b8 ("drm/vc4: plane: Enable scaler for YUV444 on GEN6"), the image looks a little soft when rendering at 1:1 due to the scaling filter being enabled.

Switch to using the nearest neighbour filter automatically when not scaling in YUV444 to compensate.


6by9:
@cillian64

cillian64:
Confirmed this fixes the blurriness issue, thanks again!

Issue #2730, opened 2018-10-26
vchiq: Top-down description of driver


lategoodbye:
I wasn't sure where to raise this issue (here or in the documentation repo), but i think this fits better. There is [one point](https://github.com/raspberrypi/linux/blob/rpi-4.19.y/drivers/staging/vc04_services/interface/vchi/TODO#L48) on the TODO list for the vchiq driver before it can leave staging: documentation

This driver is too complex to get the idea behind it by looking at the code. So it would be nice if someone at RPF can provide so kind of overview / top-down description of the driver and its interfaces.

lategoodbye:
Would be an issue in documentation better?

ghost:
The linked document says: 

> "VCHI is just a tool to let drivers talk to the firmware."

A slight expansion on that would be:

"The Raspberry Pi runs two separate operating systems: Linux on the ARM, and a threedx RTOS called vcos running on the Videocore 4 VPU: vcos is commonly called "the firmware". Vcos is tasked with performing certain system services, for example running the display and audio outputs. The VCHI driver provides an interface to allow Linux drivers to talk to vcos. Raspberry Pi develop vcos and the VCHI interface based on demand for certain features and bug fixes, in concert with Broadcom".

The documentation issue is part of the process of getting VCHI included in the upstream Linux kernel. As such, the issue the linked document raises belongs in this repository. If you feel the need to explain the function of VCHI beyond the Linux kernel community, then you should file a separate issue in the documentation repository. A good starting point for any addition to the documentation repository might be my short explanation above.

lategoodbye:
@andrum99 Thanks for your feedback. We need a better explanation of the software architecture (e.g. what is the function of those kernel threads: keepalive, recycle, slot handler, sync) and the userspace interfaces of VCHI ( /dev/vchiq , ioctl calls, debugfs ). So it should be focused on the kernel driver.

JamesH65:
@andrum99 vcos isn't an OS, its an abstraction layer over the underlying ThreadX RTOS that runs on the Videocore. We actually have vcos versions for ThreadX, Linux and MS Windows OS calls, so a lot of the stuff can be recompiled for other OS architectures, as was done when Broadcom originally did the dev work. Useful for testing, and is also used in the userland repo. Nowadays, Broadcom do no firmware development/bug fixing for the Raspberry Pi firmware, it's all done by us. VCHI is the comms stack between the VC4 and the host OS.

ghost:
[...]

PR #6801, opened 2025-04-19
[RFC | DO NOT MERGE] vchiq_arm: Please fillout comments

lategoodbye:
There is still an open [TODO](https://github.com/raspberrypi/linux/blob/rpi-6.15.y/drivers/staging/vc04_services/interface/TODO#L19) about some high-level descriptions mostly related to the VideoCore interface (ABI, IOCTL) and the thread architecture. The idea is to document most of them inline, which should be more helpful than a separate document describing VCHIQ. I tried to mark the relevant points.

This branch is not intended for merge.

This is an alternative approach to #2730


lategoodbye:
@pelwell Gentle ping

pelwell:
Plenty of people have worked on this driver without such documentation. What practical difference will it make now? Is this the one thing blocking it from coming out of staging?

kbingham:
I think this was indeed one of the blocking points for us - as we never got a response from RPi on the Pi side documentation requirements.

If we were to document the vchiq ioctls we'd just be guessing what's happening on the other side and wouldn't be any better than an LLM respnose ;-)

(Clarification - it was blocking points to us completing the documentation - as for the requirement to document in the first place - that came from upstream previously)

kbingham:
Also - I'm still really looking forward to vchiq being destaged - as this was the key blocker to getting the RPi4 ISP support merged from what I recall.

lategoodbye:
Yes, as a developer i hate documentation, too. But userspace interfaces like IOCTL needs to be documented. [Greg Kroah-Hartman](https://lore.kernel.org/lkml/Z-2XfqNIBsMRR1Jp@kroah.com/) was recently a bit uneasy about destaging of VCHIQ. In most cases a short sentences should be enough.

Thanks

pelwell:
Added comments and rebased.

PR #6803, opened 2025-04-20
hwmon: aht10: Fix AHT20 initialization

joshermar:
The existing driver claims AHT20 support in i2c_device_id, but fails to:
1. Use the correct init command (0xBE for AHT20 vs 0xE1 for AHT10)
2. Omit AHT10_MODE_CYC which AHT20 doesn't support/require

This patch adds the proper initialization sequence and includes "aosong,aht20" in the device tree match table to fully support the AHT20.

AHT20 data sheet: http://www.aosong.com/uploadfiles/2025/04/20250417112732054.pdf

Tested on Raspberry Pi 3B+ and Zero 2 W.

### DIFF
--- a/drivers/hwmon/aht10.c
+++ b/drivers/hwmon/aht10.c
@@ -37,6 +37,8 @@
 #define AHT10_CMD_MEAS	0b10101100
 #define AHT10_CMD_RST	0b10111010
 
+#define AHT20_CMD_INIT	0b10111110
+
 /*
  * Flags in the answer byte/command
  */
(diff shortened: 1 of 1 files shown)


6by9:
I was going to say that this should go upstream and needed a dtbinding update, but it looks like we added the DT support in 812b4aa3dee6bea5ec6e557a2b4e9168b7bafae9 and it hasn't been upstreamed or had a binding written.

(The updated init commands could go upstream).

The changes look good to me.

pelwell:
I'm OK with accepting such a small downstream patch, since it is only extending an existing one.

pelwell:
There's a corresponding addition to the i2c-sensor overlay here: https://github.com/raspberrypi/linux/pull/6805

PR #6804, opened 2025-04-22
6.12/iob/topic/imx335

kbingham:
Add fixes for the IMX335 camera sensor.

The kernel driver patches are all posted and submitted (or already included) in the linux-media trees.

The dtoverlay adds support for connecting an IMX335 to the RPi5.

The driver already supports two lane operation, so it should be possible to update this to use 2 lanes too .... I suspect that's easy with the overlays?

Anyhow - posting for first review - and consideration.


kbingham:
Aha, my first review feedback:

"Overlays without documentation:
  imx335"

arch/arm/boot/dts/overlays/imx335-overlay.dts
```diff
@@ -0,0 +1,91 @@
...
+	__overrides__ {
+		rotation = <&cam_node>,"rotation:0";
+		orientation = <&cam_node>,"orientation:0";
+		media-controller = <&csi>,"brcm,media-controller?";
+		cam0 = <&i2c_frag>, "target:0=",<&i2c_csi_dsi0>,
+		       <&csi_frag>, "target:0=",<&csi0>,
+		       <&clk_frag>, "target:0=",<&cam0_clk>,
+		       <&reg_frag>, "target:0=",<&cam0_reg>,
+		       <&cam_node>, "clocks:0=",<&cam0_clk>,
+		       <&cam_node>, "avdd1-supply:0=",<&cam0_reg>;
```
6by9: `avdd1-supply`? The regulator appears to be `avdd-supply` in the driver and imx335.dtsi.

arch/arm/boot/dts/overlays/imx335-overlay.dts
```diff
@@ -0,0 +1,91 @@
...
+		target = <&csi1>;
+		csi: __overlay__ {
+			status = "okay";
+			brcm,media-controller;
+
+			port {
+				csi_ep: endpoint {
+					remote-endpoint = <&cam_endpoint>;
+					clock-lanes = <0>;
+					data-lanes = <1 2 3 4>;
```
6by9: Seeing as the driver can work over 2 lanes, I just wondered if you'd tried it.
The imx290/327 overlay supports both if you want a copy/paste for implementing the override.

arch/arm/boot/dts/overlays/imx335-overlay.dts
```diff
@@ -0,0 +1,91 @@
...
+					clock-noncontinuous;
```
6by9: Is it really non-continuous? It's not configurable in the driver, so depends on how it was written in the first place

arch/arm/boot/dts/overlays/imx335-overlay.dts
```diff
@@ -0,0 +1,91 @@
...
+				};
+			};
+		};
+	};
+
+	__overrides__ {
+		rotation = <&cam_node>,"rotation:0";
+		orientation = <&cam_node>,"orientation:0";
+		media-controller = <&csi>,"brcm,media-controller?";
```
6by9: 6.12 needs to switch the compatible string to `brcm,bcm2835-unicam-legacy`. 

See https://github.com/raspberrypi/linux/blob/rpi-6.12.y/arch/arm/boot/dts/overlays/imx219-overlay.dts#L67-L72 and https://github.com/raspberrypi/linux/blob/rpi-6.12.y/arch/arm/boot/dts/overlays/imx219-overlay.dts#L93

6by9 (commented):
We tend to use eg "Commit c0aa40f45fef80b4182704d1bc089cbf8ae8bed0 upstream" in the commit message when backporting, just to make life easier when forward porting to a kernel release which already includes it.
(That hash would apply for "media: imx335: Set vblank immediately").

You say you have some modules. Are they still commercially available? I try and have a sample of most modules we nominally support.

kbingham:
> We tend to use eg "Commit [c0aa40f](https://github.com/raspberrypi/linux/commit/c0aa40f45fef80b4182704d1bc089cbf8ae8bed0) upstream" in the commit message when backporting, just to make life easier when forward porting to a kernel release which already includes it. (That hash would apply for "media: imx335: Set vblank immediately").

Ack - will handle that in the next update.

> You say you have some modules. Are they still commercially available? I try and have a sample of most modules we nominally support.

The IMX335 support we did was for a different platform - and Arducam made the modules for us. I've asked them to make more modules (they use the RPi-22 pin connector) and they work well on Raspberry Pi so I hope I can ask Arducam to send you a sample.

arch/arm/boot/dts/overlays/imx335-overlay.dts
kbingham: I haven't tried it on RPi yet. I've used / tested 2 lane on i.MX8MP where we developed support for this sensor ... but yes I would like to test on RPi too - as then this module could hook directly to an RPi4 for instance too.

I'll give that a go next to test it.

arch/arm/boot/dts/overlays/imx335-overlay.dts
kbingham: argh - I suspect I made a copy paste error... sorry. Cleaning up...

arch/arm/boot/dts/overlays/imx335-overlay.dts
6by9: Fine. It was only a query rather than an obligation. As you say, it does make it functional with Pi0-4 if it'll work with 2 lanes which is beneficial to anyone making modules.

arch/arm/boot/dts/overlays/imx335-overlay.dts
6by9: That's why we do code reviews!

arch/arm/boot/dts/overlays/imx335-overlay.dts
kbingham: Ack, no idea how this part is being handled, but I've copied the same as imx219. That also gave me a clearer insight into the 2lane/4lane option too.

arch/arm/boot/dts/overlays/imx335-overlay.dts
kbingham: Err - no - I don't think this should be in here unless its supported by the driver.

arch/arm/boot/dts/overlays/imx335-overlay.dts
kbingham: Fixed, and also now tested by swapping on to cam0 ;-)

kbingham:
Updated, and now with updated link-frequency support, fixed cam0 support, and configurable 2-lane (default) 4-lane options so this will be able to work on other Pi platforms too. And the upstream reference has been added to the commit which is already merged upstream.

Further added documentation to the Overlay README, and provided a defconfigs update to enable the imx335 module in kernel builds.

And as a last treat, while adding the 4lane options and documenting I saw that other sensors hadn't been updated to reflect CM5 and Pi5 support - so a cleanup commit on top for that.

arch/arm/boot/dts/overlays/imx335-overlay.dts
6by9: This is one of those things I hate about some drivers. It's routinely not checked or specified when the driver is first merged. Initial DT files will often be wrong, and so correcting it later may mean breaking existing DT or at least changing the behaviour slightly.

The Starvis datasheets seem to lack this information, so I may dig out an oscilloscope to see what the others in the family are doing.

arch/arm/boot/dts/overlays/README
```diff
@@ -2822,6 +2822,23 @@ Params: 4lane                   Enable 4 CSI2 lanes. This requires a Compute
         cam0                    Adopt the default configuration for CAM0 on a
                                 Compute Module (CSI0, i2c_vc, and cam0_reg).
 
+Name:   imx335
+Info:   Sony IMX335 camera module.
+        Uses Unicam 1, which is the standard camera connector on most Pi
+        variants.
+Load:   dtoverlay=imx415,<param>
```
6by9: s/imx415/imx335

arch/arm/boot/dts/overlays/README
```diff
@@ -2822,6 +2822,23 @@ Params: 4lane                   Enable 4 CSI2 lanes. This requires a Compute
         cam0                    Adopt the default configuration for CAM0 on a
                                 Compute Module (CSI0, i2c_vc, and cam0_reg).
```
6by9, line 2824: 2 blank lines required.

arch/arm/boot/dts/overlays/README
```diff
@@ -2822,6 +2822,23 @@ Params: 4lane                   Enable 4 CSI2 lanes. This requires a Compute
...
+        rotation                Mounting rotation of the camera sensor (0 or
+                                180, default 0)
+        media-controller        Configure use of Media Controller API for
+                                configuring the sensor (default on)
+        link-frequency          Allowable link frequency values to use in Hz:
+                                594000000 (default), 445500000.
+        cam0                    Adopt the default configuration for CAM0 on a
+                                Compute Module (CSI0, i2c_vc, and cam0_reg).
+        4lane                   Enable 4 CSI2 lanes. This requires a Compute
+                                Module (1, 3, 4, or 5) or Pi 5.
```
6by9, line 2843: 2 blank lines required here too.

6by9 (commented):
I'm happy to ignore the checkpatch errors - the unknown commits are because it doesn't check out a full tree.

Phil's overlaycheck script has the 3 complaints flagged.

arch/arm/boot/dts/overlays/README
kbingham: bah humbug ;-) Fixing. ...

arch/arm/boot/dts/overlays/README
kbingham: Ohh, sorry - I hadn't noticed there were two line gaps ... that's an easy fix too ;-)

pelwell (approved):

pelwell:
Feel free to merge if you're happy now, @6by9.

6by9:
I'm happy. Just letting the CI builds complete (I'm not expecting them to fail though).

6by9:
Merged. Thanks Kieran.

6by9:
@kbingham  My module arrived this morning - thanks for arranging.

Works OK, but `rpicam-hello --list-cameras` gives me multiple identical modes for the sensor
```
pi@pi:~/rpicam-apps $ ~/rpicam-apps/build/apps/rpicam-hello --list-cameras
Available cameras
-----------------
0 : imx335 [2592x1944 12-bit RGGB] (/base/soc/i2c0mux/i2c@1/imx335@1a)
    Modes: 'SRGGB10_CSI2P' : 2592x1944 [29.99 fps - (0, 0)/2592x1944 crop]
                             2592x1944 [29.99 fps - (0, 0)/2592x1944 crop]
                             2592x1944 [29.99 fps - (0, 0)/2592x1944 crop]
           'SRGGB12_CSI2P' : 2592x1944 [29.99 fps - (0, 0)/2592x1944 crop]
                             2592x1944 [29.99 fps - (0, 0)/2592x1944 crop]
                             2592x1944 [29.99 fps - (0, 0)/2592x1944 crop]
```

imx335_enum_frame_size - https://github.com/torvalds/linux/blob/master/drivers/media/i2c/imx335.c#L663-L664
```
	if (fsize->index > ARRAY_SIZE(imx335_mbus_codes))
		return -EINVAL;
```
should be `if (fsize->index)`, as index is the index of the frame size, not of the format code.

Just letting you know as you're the maintainer for this driver.

kbingham:
Aha - oops! Yes, I'll get this fixed.

kbingham:
Fix tested. Will post upstream and a new pr here! Thanks!

PR #6807, opened 2025-04-22
IMX415 higher link rate support

6by9:
I need to upstream this one, but leaving it here for the time being.


6by9:
12bit readout mode needs checking that it behaves with all the link frequency options as the datasheet says it is only valid for certain values.

6by9:
Closing and moving to a 6.18 PR - #7433

Issue #6808, opened 2025-04-18
kernel unicode support missing - ext4 casefold feature not avaailabe


bjacke:
when trying to mount a ext4 fs with casefold enabled this error comes up:

EXT4-fs (dm-4): Filesystem with casefold feature cannot be mounted without CONFIG_UNICODE

I want the casefold feature enabled because this allows a huge speed-up for samba fileservers.

pelwell:
This feature looks like it can be enabled as a module, with loading only when actually needed, so the only cost should be a bit of storage space.

pelwell:
See https://github.com/raspberrypi/linux/pull/6809. In about 40 minutes you should be able to install a trial kernel using `sudo rpi-update pulls/6809`. Back up any important data first.

pelwell:
#6809 has now been merged.

bjacke:
thank you very much for the super quick change! :-) It works well for me now!

PR #6809, opened 2025-04-23
configs: Enable UNICODE support (for filesystems)

pelwell:
CONFIG_UNICODE enables "UTF-8 normalization and casefolding support". Enable it for the standard defconfigs as a module, so that it is only loaded when needed, e.g. to support a case-insensitive filing system.

See: https://github.com/raspberrypi/linux/issues/6808


pelwell:
This results in a small module being added - 21kB when compressed, and people who need it won't care what size it is when uncompressed and loaded.

Issue #6810, opened 2025-04-24
EGL rendering via SDL2 broken under 6.12 on 2712D0


6by9:
### Describe the bug

Reported in https://forums.raspberrypi.com/viewtopic.php?p=2311618

Running `/usr/libexec/installed-tests/SDL2/testgl2` from `libsdl2-tests` with a 6.12 kernel on a 2712D0 gives random noise on the screen. 6.6 kernel or C1 gives correct output.

### Steps to reproduce the behaviour

```
sudo apt install libegl1 libsdl2-tests
/usr/libexec/installed-tests/SDL2/testgl2
```

### Device (s)

Raspberry Pi CM5, Raspberry Pi 5

### System

Raspberry Pi OS Bookworm
Tested OK on 6.6.78 kernel.
Fails on 6.12.22

6by9:
Tested on rpi-6.15.y (6.15.0-rc3) as well and also fails.
However I've just noticed that I get a flicker of the correct image as I quit the application. I think this is the DRM async updates in vc4 again - I'll have a look.

6by9:
https://github.com/raspberrypi/linux/blob/rpi-6.12.y/drivers/gpu/drm/vc4/vc4_plane.c#L2376
`== VC4_GEN_6_C` should be `>= VC4_GEN_6_C`.

#6811

PR #6811, opened 2025-04-24
drm/vc4: plane: Fix incorrect handling of GEN_6_D in vc4_plane_async_…

6by9:
…set_fb

A conditional had been left as == GEN_6_C, when it also applied to GEN_6_D, resulting in an invalid change to the dlist on async updates.

Fixes: b7b14b31c886 ("drm/vc4: plane: Add support for 2712 D-step.")


6by9:
Fixes #6810 6810

6by9:
(`Fixes:` hash is deliberately the upstream one as I need to upstream this patch).

popcornmix (approved):

PR #6820, opened 2025-04-28
Add RS485 mode support for UART2, UART3, UART4, and UART5.

TheMightyMadman:
This patch adds RS485 mode support for UART2/3/4/5.
This allows the RTS line to be used to drive the OE pin of an RS485 transceiver (i.e. MAX3078E).
I needed this for a project & it would be useful to have in-tree.

The "ctsrts" override is automatically activated, as this is required to enable the RTS line; there are also overrides to invert the RTS signal, and to set the RTS on and off delays to change the RTS timing around data transmission.

I haven't made changes to uart0-overlay.dts or uart1-overlay.dts as the syntax is quite different, I'm guessing this is legacy.
I also haven't made changes to uartX-pi5-overlay.dts as I'm unable to test these due to them being Pi5-only.

Tested and working OK on a CM4 with a pair of MAX3078E transceivers, overrides in /boot/firmware/config.txt as follows:
`dtoverlay=uart3,rs485`
`dtoverlay=uart4,rs485`

Credit to 6by9 for pretty much all of the device tree modifications here.
The associated forum thread is available here:
https://forums.raspberrypi.com/viewtopic.php?p=2311800

### DIFF
--- a/arch/arm/boot/dts/overlays/README
+++ b/arch/arm/boot/dts/overlays/README
@@ -5181,6 +5181,19 @@ Name:   uart2
 Info:   Enable uart 2 on GPIOs 0-3. BCM2711 only.
 Load:   dtoverlay=uart2,<param>
 Params: ctsrts                  Enable CTS/RTS on GPIOs 2-3 (default off)
+        rs485                   Enable RS485 mode for using the RTS line to
+                                drive the OE pin of an RS485 transceiver (i.e.
+                                MAX3078E); also enables the UARTx ctsrts
+                                parameter, as RTS is required (default off).
+        rs485_invert_rts        When RS485 mode is enabled, inverts the RTS
+                                line from active-high (default) to active-low
+                                (default off).
+        rs485_rts_on_delay      When RS485 mode is enabled, sets the delay (in
+                                milliseconds) between data transmission starting
+                                and the RTS line being asserted (default 0).
+        rs485_rts_off_delay     When RS485 mode is enabled, sets the delay (in
+                                milliseconds) between data transmission ending
+                                and the RTS line being deasserted (default 0).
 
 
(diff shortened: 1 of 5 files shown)


arch/arm/boot/dts/overlays/README
```diff
@@ -5181,6 +5181,19 @@ Name:   uart2
 Info:   Enable uart 2 on GPIOs 0-3. BCM2711 only.
 Load:   dtoverlay=uart2,<param>
 Params: ctsrts                  Enable CTS/RTS on GPIOs 2-3 (default off)
+        rs485                   Enable RS485 mode for using the RTS line to
+                                drive the OE pin of an RS485 transceiver (i.e.
+                                MAX3078E); also enables the UARTx ctsrts
+                                parameter, as RTS is required (default off).
+        rs485_invert_rts        When RS485 mode is enabled, inverts the RTS
+                                line from active-high (default) to active-low
+                                (default off).
```
6by9: I'm not quite sure what you mean by "default off" here. You've already said the default is active-high. Applies to the other instances too.

6by9 (commented):
Just the one minor comment from me.

arch/arm/boot/dts/overlays/README
TheMightyMadman: Thank you :) I meant that "rs485_invert_rts" is default off, but I appreciate that could cause confusion - would it be better if I just removed the last line of rs485_invert_rts on each instance?

TheMightyMadman:
> Just the one minor comment from me.

Thank you - I've amended the potentially confusing documentation in a new commit.

pelwell:
FYI Several of the parameters are incorrectly declared and have no effect. See #7177 for fix.

Issue #6817, opened 2025-04-27
Huge memory leak in the Linux kernel (400-500 MB, likely in the video driver) triggered on every opening KiCad with a moderately complex schematic.


zxmak:
### Describe the bug

Every time I open a moderately complex schematic file in KiCad, and then immediately close KiCad without doing anything, the amount of free memory decreases by approximately 400 - 500 MB.

Each new launch of KiCad causes an additional 400 - 500 MB of memory loss. After 8 - 10 launches, the system runs out of memory and becomes unresponsive. After KiCad closes, the system shows reduced available memory, but no running processes account for the missing memory.

The leak appears to occur at the kernel level, most likely within the video driver.
I am attaching a test schematic file (test.kicad_sch) that reproduces the issue reliably.

[test.kicad_sch.zip](https://github.com/user-attachments/files/19929393/test.kicad_sch.zip)

Please investigate and fix this issue.

### Steps to reproduce the behaviour

1) Install a clean Raspberry Pi OS Bookworm 64-bit with the Wayland Labwc compositor.

2) Add the Debian backports repository by running:
```
sudo nano /etc/apt/sources.list.d/bookworm-backports.list
```
and adding the following line:
```
deb http://deb.debian.org/debian bookworm-backports main contrib non-free
```

3) Install KiCad by running:
```
sudo apt update && sudo apt install -t bookworm-backports kicad
```

4) Open **[test.kicad_sch](https://github.com/user-attachments/files/19929393/test.kicad_sch.zip)** with the KiCad schematic editor

5) Close the KiCad schematic editor.

6) Observe that approximately **400–500 MB** of memory is leaked.

7) Repeat steps 4 and 5 until the system freezes due to running out of memory.

### Device (s)

Raspberry Pi 4 Mod. B

### System

```
$ cat /etc/rpi-issue
Raspberry Pi reference 2023-09-22
Generated using pi-gen, https://github.com/RPi-Distro/pi-gen, 40f37458ae7cadea1aec913ae10b5e7008ebce0a, stage4

$ vcgencmd version
Mar 19 2025 18:24:21 
Copyright (c) 2012 Broadcom
version ca6e8171a80ea46924ffaa629250bfb482f3a02c (clean) (release) (start)

$ uname -a
Linux rpi4 6.12.20+rpt-rpi-v8 #1 SMP PREEMPT Debian 1:6.12.20-1+rpt1~bpo12+1 (2025-03-19) aarch64 GNU/Linux
```

[...]

PR #6822, opened 2025-04-29
drm/v3d: Add job to pending list if the reset was skipped

pelwell:
[ Potential fix for #6817 ]

When a CL/CSD job times out, we check if the GPU has made any progress since the last timeout. If so, instead of resetting the hardware, we skip the reset and let the timer get rearmed. This gives long-running jobs a chance to complete.

However, when `timedout_job()` is called, the job in question is removed from the pending list, which means it won't be automatically freed through `free_job()`. Consequently, when we skip the reset and keep the job running, the job won't be freed when it finally completes.

This situation leads to a memory leak, as exposed in [1].

Similarly to commit 704d3d60fec4 ("drm/etnaviv: don't block scheduler when GPU is still active"), this patch ensures the job is put back on the pending list when extending the timeout.

Cc: stable@vger.kernel.org # 6.0
Link: https://gitlab.freedesktop.org/mesa/mesa/-/issues/12227 [1]
Reported-by: Daivik Bhatia <dtgs1208@gmail.com>

### DIFF
--- a/drivers/gpu/drm/v3d/v3d_sched.c
+++ b/drivers/gpu/drm/v3d/v3d_sched.c
@@ -744,11 +744,6 @@ v3d_gpu_reset_for_timeout(struct v3d_dev *v3d, struct drm_sched_job *sched_job)
 	return DRM_GPU_SCHED_STAT_NOMINAL;
 }
 
-/* If the current address or return address have changed, then the GPU
- * has probably made progress and we should delay the reset.  This
- * could fail if the GPU got in an infinite loop in the CL, but that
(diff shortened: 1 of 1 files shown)


mairacanal:
I just sent a v2 of this patch [1] as v1 doesn't use the DRM sched internal lock, which can lead to a possible race condition. I'd appreciate it if you could take v2 in the tree.

[1] [[PATCH v2] drm/v3d: Add job to pending list if the reset was skipped](https://lore.kernel.org/dri-devel/20250430210643.57924-1-mcanal@igalia.com/T/)

PR #6823, opened 2025-04-29
Add adxl355 to i2c-sensor

6by9:
https://forums.raspberrypi.com/viewtopic.php?t=387270

I don't have hardware to test this, so will ask OP to test.


6by9:
OP reported this as working.

arch/arm/boot/dts/overlays/i2c-sensor-common.dtsi
```diff
@@ -815,7 +829,8 @@
 			<&bmp380>,"reg:0", <&adt7410>,"reg:0", <&ina238>,"reg:0",
 			<&hdc3020>,"reg:0", <&as73211>,"reg:0",
 			<&as7331>,"reg:0", <&adxl345>,"reg:0",
-			<&ads7828>,"reg:0", <&ads7830>,"reg:0";
+			<&ads7828>,"reg:0", <&ads7830>,"reg:0",
+			<&adxl345>,"reg:0";
```
pelwell: adxl355, surely?

6by9: Indeed.
I'd accidentally pushed to this branch as I thought the change had been merged, so was just reinstating it.

Then again changing the address is rather redundant when the datasheet says
>However, when sharing an SDA bus, the ADXL355 may prevent communica-
tion with other devices on that bus. If at any point, even when
the ADXL355 is not being addressed, the 0x3A and 0x3B bytes
(when the ADXL355 device address is set to 0x1D), or the 0xA6
and 0xA7 bytes (when the ADXL355 device address is set to
0x53) are transmitted on the SDA bus, the ADXL355 responds with
an acknowledge bit and pulls the SDA line down. For example,
this response can occur when reading or writing the data bytes
(0x3A/0x3B or 0xA6/0xA7) to another sensor on the bus. When
the ADXL355 pulls the SDA line down, communication with other
devices on the bus may be interrupted. To resolve this interruption,
the ADXL355 must be connected to a separate SDA bus, or the
CS/SCL pin must be switched high when communication with the
ADXL355 is not desired (it is normally grounded).

If you can't share the bus with any other device, then you may as well leave the address at the default.

6by9:
Fixed s/adxl345/adxl355, and rebased.

6by9:
As this was still open, I've tagged the requested changes for BMP580 on to it as well
https://github.com/raspberrypi/firmware/issues/1994

arch/arm/boot/dts/overlays/README
```diff
@@ -2475,6 +2478,10 @@ Params: addr                    Set the address for the ADS7828, ADS7830,
         bmp380                  Select the Bosch Sensortronic BMP380
                                 Valid addresses 0x76-0x77, default 0x76
 
+        bmp580                  Select the Bosch Sensortronic BMP580, BMP581,
```
pelwell: Wouldn't `bcm58x` be a more obvious (and more consistent) name, in that it hints that multiple devices are supported?

6by9: Sure. Done.

PR #6825, opened 2025-04-29
Mulligan for #6392

P33M:
Upstream's base dts exposes a full-range 32bit MMIO window starting at a PCI bus address of 0. This causes sufficient interop annoyance to be worth permanently applying the mmio-hi property of the pciex1-compat-pi5 overlay to the downstream dts.

### DIFF
--- a/arch/arm64/boot/dts/broadcom/bcm2712-ds.dtsi
+++ b/arch/arm64/boot/dts/broadcom/bcm2712-ds.dtsi
@@ -488,6 +488,11 @@
 &pcie1 {
 	brcm,fifo-qos-map = /bits/ 8 <3 3 3 3>;
 	status = "disabled";
+	ranges =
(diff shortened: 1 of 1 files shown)


arch/arm64/boot/dts/broadcom/bcm2712-ds.dtsi
```diff
@@ -488,6 +488,11 @@
 &pcie1 {
 	brcm,fifo-qos-map = /bits/ 8 <3 3 3 3>;
 	status = "disabled";
+	ranges =
+		/* 2GiB, 32-bit, non-prefetchable at PCIe 00_80000000
```
pelwell: This comment is unterminated.

P33M: Oops. Of course it booted with an nvme drive (64-bit BAR).

PR #6826, opened 2025-04-29
drm/rp1/rp1_dpi: Move Composite Sync generation into the kernel

njhollinghurst:
Move RP1 DPI's PIO-assisted Composite Sync generation code, previously released as a separate utility, into the kernel driver. There are 3 variants for progressive, generic interlaced and TV- style interlaced CSync, alongside the existing VSync fixup.

Check that all of GPIOs 1-3 are mapped to DPI, so PIO won't try to snoop on a missing output, or override another device's pins.

Add "force-csync" as both an OF property and a module parameter, for convenience of testing, as few tools set DRM_MODE_FLAG_CSYNC.

### DIFF
--- a/drivers/gpu/drm/rp1/rp1-dpi/rp1_dpi.c
+++ b/drivers/gpu/drm/rp1/rp1-dpi/rp1_dpi.c
@@ -61,6 +61,16 @@
 static unsigned int default_bus_fmt = MEDIA_BUS_FMT_RGB666_1X18;
 module_param(default_bus_fmt, uint, 0644);
 
+/*
+ * Override DRM mode flags to force the use of Composite Sync on GPIO1.
+ * This is mostly for testing, as neither panel-timing nor command-line
+ * arguments nor utilities such as "kmstest" can set DRM_MODE_FLAG_CSYNC.
+ * Sampled on each enable/mode-switch. Default polarity will be -ve.
+ * (Setting this may break Vertical Sync on GPIO2 for interlaced modes.)
+ */
+static bool force_csync;
+module_param(force_csync, bool, 0644);
+
 /* -------------------------------------------------------------- */
 
 static void rp1dpi_pipe_update(struct drm_simple_display_pipe *pipe,
@@ -89,7 +99,8 @@ static void rp1dpi_pipe_update(struct drm_simple_display_pipe *pipe,
 						dpi->bus_fmt,
 						dpi->de_inv,
 						&pipe->crtc.state->mode);
-				rp1dpi_pio_start(dpi, &pipe->crtc.state->mode);
+				rp1dpi_pio_start(dpi, &pipe->crtc.state->mode,
+						 force_csync);
 				dpi->dpi_running = true;
 			}
 			dpi->cur_fmt = fb->format->format;
@@ -294,6 +305,7 @@ static int rp1dpi_platform_probe(struct platform_device *pdev)
 	struct drm_bridge *bridge = NULL;
 	const char *rgb_order = NULL;
 	struct drm_panel *panel;
+	u32 missing_gpios;
 	int i, j, ret;
 
 	dev_info(dev, __func__);
@@ -354,6 +366,7 @@ static int rp1dpi_platform_probe(struct platform_device *pdev)
 	if (ret)
 		goto done_err;
 
+	/* RGB order property - to match VC4 */
 	dpi->rgb_order_override = RP1DPI_ORDER_UNCHANGED;
 	if (!of_property_read_string(dev->of_node, "rgb_order", &rgb_order)) {
 		if (!strcmp(rgb_order, "rgb"))
@@ -368,9 +381,9 @@ static int rp1dpi_platform_probe(struct platform_device *pdev)
 			DRM_ERROR("Invalid dpi order %s - ignored\n", rgb_order);
 	}
 
-	/* Check if PIO can snoop on or override DPI's GPIO1 */
-	dpi->gpio1_used = false;
-	for (i = 0; !dpi->gpio1_used; i++) {
+	/* Check if all of GPIOs 1, 2 and 3 are assigned to DPI */
+	missing_gpios = BIT(1) | BIT(2) | BIT(3);
+	for (i = 0; missing_gpios; i++) {
 		u32 p = 0;
 		const char *str = NULL;
 		struct device_node *np1 = of_parse_phandle(dev->of_node, "pinctrl-0", i);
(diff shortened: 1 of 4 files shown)


drivers/gpu/drm/rp1/rp1-dpi/rp1_dpi.c
```diff
@@ -379,21 +392,26 @@ static int rp1dpi_platform_probe(struct platform_device *pdev)
...
 				if (!strcmp(str, "gpio1"))
-					dpi->gpio1_used = true;
+					ret |= BIT(1);
+				else if (!strcmp(str, "gpio2"))
+					ret |= BIT(2);
+				else if (!strcmp(str, "gpio3"))
+					ret |= BIT(3);
 			}
-			for (j = 0; !dpi->gpio1_used; j++) {
+			for (j = 0; ret < 3; j++) {
```
pelwell: This condition looks wrong - shouldn't it also be 0xE (although the capital bugs me)?

drivers/gpu/drm/rp1/rp1-dpi/rp1_dpi.c
```diff
@@ -379,21 +392,26 @@ static int rp1dpi_platform_probe(struct platform_device *pdev)
...
 					break;
-				if (p == 1)
-					dpi->gpio1_used = true;
+				if (p >= 1 && p <= 3)
+					ret |= (1 << p);
 			}
 		}
 		of_node_put(np1);
 	}
+	dpi->sync_gpios_mapped = (ret >= 0xE);
```
pelwell: It will never by > 0xE.

drivers/gpu/drm/rp1/rp1-dpi/rp1_dpi.c
njhollinghurst: Ah yes, thanks. You can see how the code evolved from checking only GPIO1 to 2 to 3 (all of which are needed)...
0xe doesn't look much nicer though. Is there a more Linuxy way to do this?

pelwell (commented):
I may go through the body of the patch later, but I think the initialisation stuff would be nicer if you invert the logic and have a bitmask of outstanding GPIOs, initialised to 0x7 (or 0xe if you prefer), clearing the bits as they are found.

drivers/gpu/drm/rp1/rp1-dpi/rp1_dpi.c
popcornmix: `BIT(3)|BIT(2)|BIT(1)`?

njhollinghurst: Yes - I think I'll do Phil's suggestion below of set them first and then clear them. I should also stop overloading `ret` since comparing it to zero now gives a misleading impression of intent.

njhollinghurst:
Any views about the OF property and module parameter `"force-csync"`?

`DRM_MODE_FLAG_CSYNC` is difficult to use. DRM TV mode helper code and command-line mode parser _don't_ set it even for "PAL" and "NTSC" modes. `panel-timings` can't set it. Neither can utilities like `kmstest`. About the only reliable way it could be set would be a custom mode-list or dummy EDID. Although the EDID flag (and, I presume, the DRM one) defines it to mean CSync-on-HSync or CSync-on-Green (which doesn't map perfectly to RP1 hardware, or to VC4).

So an alternative mechanism to enable Composite Sync over DEN arguably needed. I'm not sure we need both OF and module properties? The former could be set by a dtparam. The latter could be changed at runtime.

pelwell:
I have no problem with the idea of the DT property (we're missing an implementation of the overlay parameter to set it), but if you are happy with csync being a system-wide setting then you could just make the override add the module parameter - any string assigned to the `bootargs` of `/chosen` is appended, e.g. from the `audremap` overlays:
```
	fragment@1 {
		target = <&chosen>;
		__overlay__  {
			bootargs = "snd_bcm2835.enable_headphones=1";
		};
	};
```

drivers/gpu/drm/rp1/rp1-dpi/rp1_dpi_pio.c
```diff
@@ -364,30 +364,30 @@ static int rp1dpi_pio_csync_ilace(struct rp1_dpi *dpi,
...
-		//     .wrap
+		//     .wrap_target                           ; while (true) {
+		0x3083, //  2: wait   1 gpio, 3    side 1     ;   do { @HSync
+		0xa422, //  3: mov    x, y         side 0 [4] ;   CSYNC: x = VSW - 1
+		0x2003, //  4: wait   0 gpio, 3    side 0     ;   CSYNC: HSync->CSync
+		0x12c2, //  5: jmp    pin, 2       side 1 [2] ;   } while (VSync)
+		0x3083, //  6: wait   1 gpio, 3    side 1     ;   @HSync
+		0xc442, //  7: irq    clear 2      side 0 [4] ;   CSYNC: flush IRQ
+		0x2003, //  8: wait   0 gpio, 3    side 0     ;   CSYNC: Hsync->CSync
+		0x30c2, //  9: wait   1 irq, 2     side 1     ;   @midline
```
njhollinghurst, line 375: This instruction is probably redundant but I don't feel like messing with it right now.

njhollinghurst:
Ah yes, I had forgotten that trick.
I think I'll leave them both in for now (as with "default_bus_fmt") but might yet remove the OF one.
I think this is just about ready to squash and de-Draft.

6by9 (approved):
I see nothing too untoward, although upstreaming those changes is going to be almost impossible. Happy to kick that one into the long grass though.

njhollinghurst:
I think I'll remove the OF property anyway, as in theory it would still need documenting.

And I don't propose to check in any overlay changes, so as not to raise issues of VC4 compatibility -- not now anyway!

igorpecovnik:
K6.14.y compilation breaks with:
```
   drivers/gpu/drm/rp1/rp1-dpi/rp1_dpi_pio.c:599:5: error: conflicting types for 'rp1dpi_pio_start'; have 'int(struct rp1_dpi *, const struct drm_display_mode *)'
     599 | int rp1dpi_pio_start(struct rp1_dpi *dpi, const struct drm_display_mode *mode)
         |     ^~~~~~~~~~~~~~~~
   In file included from drivers/gpu/drm/rp1/rp1-dpi/rp1_dpi_pio.c:28:
   drivers/gpu/drm/rp1/rp1-dpi/rp1_dpi.h:93:5: note: previous declaration of 'rp1dpi_pio_start' with type 'int(struct rp1_dpi *, const struct drm_display_mode *, bool)' {aka 'int(struct rp1_dpi *, const struct drm_display_mode *, _Bool)'}
      93 | int rp1dpi_pio_start(struct rp1_dpi *dpi, const struct drm_display_mode *mode,
         |     ^~~~~~~~~~~~~~~~
     CC      net/ipv4/igmp.o
   make[8]: *** [scripts/Makefile.build:222: drivers/gpu/drm/rp1/rp1-dpi/rp1_dpi_pio.o] Error 1
   make[7]: *** [scripts/Makefile.build:472: drivers/gpu/drm/rp1/rp1-dpi] Error 2
   make[6]: *** [scripts/Makefile.build:472: drivers/gpu/drm/rp1] Error 2
   make[5]: *** [scripts/Makefile.build:472: drivers/gpu/drm] Error 2
   make[4]: *** [scripts/Makefile.build:472: drivers/gpu] Error 2
```
Full logs: https://paste.armbian.com/afuqupofuy

pelwell:
You must have a config where CONFIG_RP1_DPI is enabled but not CONFIG_RP1_PIO. We'll patch it up.

igorpecovnik:
> You must have a config where CONFIG_RP1_DPI is enabled but not CONFIG_RP1_PIO. We'll patch it up.

Quite possible, will enable. Still this needs to be done different.

Thanks for prompt response!

pelwell:
See #6851. It will be forward-ported to the newer kernels once it is merged to 6.12.

PR #6827, opened 2025-04-29
Add DT controlled XTRIG control to imx296

6by9:
https://forums.raspberrypi.com/viewtopic.php?t=386482


pelwell:
The code is arranged so that the DT overrides the module parameter, whether the trigger is being enabled or disabled, whereas the overlay only allows it to be enabled. Removing the `=1` from the parameter declaration would allow it to be disabled as well.

6by9:
> The code is arranged so that the DT overrides the module parameter, whether the trigger is being enabled or disabled, whereas the overlay only allows it to be enabled. Removing the `=1` from the parameter declaration would allow it to be disabled as well.

Updated to drop the `=1`. I'd just copied it across from imx477 which has separate overrides for setting sink or source, with absence being disabled.

PR #6829, opened 2025-04-30
drivers: media: i2c: imx335: Fix frame size enumeration

kbingham:
In commit cfa49ff0558a ("media: i2c: imx335: Support 2592x1940 10-bit mode") the IMX335 driver was extended to support multiple output modes.

This incorrectly extended the frame size enumeration to check against the supported mbus_codes array instead of the supported mode/frame array. This has the unwanted side effect of reporting the currently supported frame size 2592x1944 three times.

Fix the check accordingly to report a frame size for each supported size, which is presently only a single entry.

Fixes: cfa49ff0558a ("media: i2c: imx335: Support 2592x1940 10-bit mode")
Upstream: https://lore.kernel.org/linux-media/20250430073649.1986018-1-kieran.bingham@ideasonboard.com/T/#u

### DIFF
--- a/drivers/media/i2c/imx335.c
+++ b/drivers/media/i2c/imx335.c
@@ -718,7 +718,8 @@ static int imx335_enum_frame_size(struct v4l2_subdev *sd,
 	struct imx335 *imx335 = to_imx335(sd);
 	u32 code;
 
-	if (fsize->index > ARRAY_SIZE(imx335_mbus_codes))
+	/* Only a single supported_mode available. */
+	if (fsize->index > 0)
 		return -EINVAL;
 
 	code = imx335_get_format_code(imx335, fsize->code);


6by9 (approved):
Thanks Kieran.

With your comment on the upstream commit that you're expecting to add in binning support which will increase the number of modes, there's no point in debating the various logical operations that could be used.
It looks like Sakari has accepted it as-is presumably on that basis, so I'm good with it too.

Do you want me to see if David can do a basic tune on the sensor?

6by9:
(Ignore the checkpatch CI failure - it only pulls the diffs, so doesn't have a full tree to find the git hashes of all commits).

kbingham:
> Thanks Kieran.
> 
> With your comment on the upstream commit that you're expecting to add in binning support which will increase the number of modes, there's no point in debating the various logical operations that could be used. It looks like Sakari has accepted it as-is presumably on that basis, so I'm good with it too.

Yes, I stated about the 2/2 binning mode precisely because I put "if (fsize->index > 0)" instead of "if (fsize->index)" like you suggested because I want to make sure (for me, later) it's clear that this is going to later compare a size.

> Do you want me to see if David can do a basic tune on the sensor?

Yes please! That would be great.

I'm probably going to try to do this myself - but it's not easy for me yet, and I don't know when I'll get to it - but it could give me a way to compare and see how I do compared to David's result (I suspect if David does the tune we'd merge his even if I attempt it though!)

6by9:
> > Do you want me to see if David can do a basic tune on the sensor?
> 
> Yes please! That would be great.
> 
> I'm probably going to try to do this myself - but it's not easy for me yet, and I don't know when I'll get to it - but it could give me a way to compare and see how I do compared to David's result (I suspect if David does the tune we'd merge his even if I attempt it though!)

Did you have a cam_helper file for imx335?
I think I copied imx290 for my quick testing on the basis they were both Starvis sensors, but I didn't check the content against the datasheet.

kbingham:
Oh - yes - let me send a pr to rpi's libcamera or something if it's not already covered. It's already in the libipa handlers - but I probably didn't handle the RPi somewhere.

PR #6830, opened 2025-04-30
drm/v3d: Don't retrieve the clock twice

mairacanal:
Upstream commit d3976b4e3830 ("drm/v3d: Add clock handling") introduced clock handling. Therefore, delete the clock handling previously introduced in downstream.


popcornmix (approved):

popcornmix:
This looks correct and a quick test shows neverball and webgl aquarium are working.

PR #6838, opened 2025-05-06
arm64: dts: bcm2712-rpi: Add uart0_dma parameter

pelwell:
It seems that some applications don't work with UART DMA enabled, while others don't work without it. No DMA seems to be the safer default choice, but add a dtparam - uart0_dma - to re-enable it.

See: https://github.com/raspberrypi/linux/issues/6365


pelwell:
Merging following positive feedback.

PR #6844, opened 2025-05-08
nvme-pci: Re-enable NVME Host Memory Block (HMB) support 

altserg:
This pull request restores the NVME Host Memory Block (HMB) support that was disabled in RPi kernel in December 2024.
Instead, the default kernel command line is changed for RPi 4/5 to set maximum allowed HBM size to 0.

This preserves current kernel functionality, but gives the user possibility to manually enable HMB in `cmdline.txt` file by adding `nvme.max_host_mem_size_mb=<size>` parameter. 

Optionally the user can increase the default CMA size in `config.txt` by adding/changing the 
`dtoverlay=vc4-kms-v3d,cma-<size>` line

### DIFF
--- a/arch/arm/boot/dts/broadcom/bcm2711-rpi-ds.dtsi
+++ b/arch/arm/boot/dts/broadcom/bcm2711-rpi-ds.dtsi
@@ -3,7 +3,7 @@
 
 / {
 	chosen {
-		bootargs = "coherent_pool=1M 8250.nr_uarts=1 snd_bcm2835.enable_headphones=0 cgroup_disable=memory numa_policy=interleave";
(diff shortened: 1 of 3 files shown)


pelwell:
This seems reasonable to me. Any concerns, @P33M?

P33M:
Not really - it's a module parameter so can be defaulted in the driver itself, but doing it this way drops a downstream code patch.

P33M (approved):

pelwell:
Lightly squashed, and merged offline. Thanks!

altserg:
Is the original HBM disable patch reverted as well ? Changing the .dtsi files only will not have any effect
This patch has to be applied as well for HMB to be usable with custom setting in cmdline.txt:
(https://github.com/raspberrypi/linux/pull/6844/commits/b0e1f022e52005d83f64eedff4f628cf19339f2d)

pelwell:
Oh ye of little faith: https://github.com/raspberrypi/linux/commits/rpi-6.12.y/

PR #6848, opened 2025-05-12
pisound-micro: Added pin_pull and pin_b_pull sysfs attributes for Pisound Micro.

gtrainavicius:
These attributes are available only for GPIO input and Encoder elements.


pelwell:
That was easy.

PR #6851, opened 2025-05-13
Fix up rp1_dpi_pio for CONFIG_RP1_PIO=n builds

pelwell:
If CONFIG_RP1_DPI_PIO is enabled but not CONFIG_RP1_PIO, the fallback version of rp1dpi_pio_start is used. This was not updated in #6826 when the force_csync parameter was added to the prototype.

Add the extra parameter, and take the opportunity to correct the commit name.


njhollinghurst (approved):
Whoops. The flag was a later addition and I forgot to test this configuration.

pelwell:
Are you happy with my addition of the force_csync flag to the conditional?

pelwell:
I'll take that as a yes.

njhollinghurst:
Yes. The return value is ignored (and DPI still kinda words) but the warning may be useful for debug.

PR #6858, opened 2025-05-16
overlays: sc16is75x: Disable spidev0 first

nbuchwitz:
Ensure spidev is disabled before the cs pin is referenced. Otherwise the overlay will fail when loaded during runtime with
"spi spi0.0: chipselect 0 already in use"


pelwell:
Thanks, Nicolai!

Issue #6854, opened 2025-05-14
SPI Bus performance loss of ~9% under Python moving from Kernel 6.6.31 -> 6.12.25


Roland-Smith-Imperial:
### Describe the bug

I am using python (3.11.2) running on a Pi5 to drive an AD7606 ADC with a loop that (1) triggers a hardware sample event using a pulse from a GPIO line (libgpio 1.6.3-1) (2) waits 3-uSec (the hardware acquisition time) followed by (3) read back of 12 bit ADC data via the SPI bus formatted as a pair of int16 values using the spidev (3.6-1) xfer2 function followed by (4) recording a time-stamp using time.perf_counter_ns.

Updating from RPi OS kernel version 6.6.31 -> 6.12.25 the average data rate drops from 35.5 -> 32.5 kHz, a ~9% performance hit.  The culprit seems to be the SPI read using spidev's xfer2 function, that step increases from an average of 21.6 -> 24.3 uSec, along with a significant in increase in timing jitter.

Reverting back to 6.6.31 recovers the performance.

Stripping the python code back to a minimalist loop that only runs the spidev xfer2 call (with no hardware in place) shows the same performance drop.  Plotting a histogram of the time taken for each sample loop also highlights much more loop-to-loop timing jitter with the 6.12.25 kernel.

Demonstration Python code below.

### Steps to reproduce the behaviour

Steps to reproduce the behaviour*

1) Run the python code below to benchmark the performance of the SPI bus using spidev xfer2 on an RPi5 with a 6.6.31 OS Kernel.  No additional hardware required.

2) Upgrade the Kernel to 6.12.25 (no other packages updates).

3) Re-run the python code to benchmark the performance of spidev xfer2 and compare before / after average loop time and loop-to-loop timing jitter.

 ########### Test code ############
```
import spidev                    # SPI bus IO tools
import struct                     # Conversion from 8-bit pairs to int16
import time                       # Access clock and sleep delays
import numpy as np         # Maths tools
import matplotlib.pyplot as plt   # Plotting functions, data display
import copy

from time import perf_counter_ns  # Access ns resolution timer
from struct import unpack

# SPI Bus set up and read back tools.
# Will run with no hardware connected.

class ADC_Pi(object):
    spiADC = spidev.SpiDev()          # Link to SPI bus and init
    spiADC.open(0, 0)                 # Expects SPI bus at dev/SPI(0,0)
[...]

PR #6862, opened 2025-05-19
gpiolib: Avoid the hotplug performance reduction

pelwell:
The 6.9 kernel introduced a large patchset [1] designed to make gpiochip usage safe in the presence of potential hotplugging events. The changes included protecting every gpiochip access with a claim of an interlock.

Running on a Pi 5 these changes reduce GPIO performance from userspace by around 10%. The penalty would be proportionally higher from kernel, as seen by SPI speed reductions.

Patch the gpiolib implementation to remove the protection of gpiochip accesses. By providing alternative implementations of the relevant macros, the changes are localised and therefore easier to verify.

See: https://github.com/raspberrypi/linux/issues/6854

[1] https://lwn.net/Articles/960024/

### DIFF
--- a/drivers/gpio/gpiolib.c
+++ b/drivers/gpio/gpiolib.c
@@ -53,6 +53,42 @@
 
 #define dont_test_bit(b,d) (0)
 
+#define REJECT_HOTPLUG_TAX 1
+
+#if REJECT_HOTPLUG_TAX
+
+#undef srcu_dereference
+#define srcu_dereference(obj, srcu) obj
+
+#undef srcu_dereference_check
+#define srcu_dereference_check(obj, srcu, held) obj
+
+#undef guard
+#define guard(srcu) (void)
+
+#undef scoped_guard
+#define scoped_guard(x, y)
+#undef list_for_each_entry_srcu
+#define list_for_each_entry_srcu(pos, head, member, srcu) \
+	list_for_each_entry(pos, head, member)
+
+#undef CLASS
+#define CLASS(gcg, gd) \
+	struct gcg gd = gcg ## _init
+
+static struct gpio_chip_guard gpio_chip_guard_init(struct gpio_desc *desc)
+{
+	struct gpio_chip_guard _guard;
+
+	_guard.gdev = desc->gdev;
+	_guard.idx = 0;
+	_guard.gc = _guard.gdev->chip;
+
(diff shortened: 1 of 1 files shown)


pelwell:
Here are the results of running a performance test on a Pi 5. The test drives a GPIO high and low 1 million times in a tight loop, repeating the process 10 times and calculating the mean and standard deviation of the per-pass execution time (in microseconds):
```
# The last 6.6 release
Linux raspberrypi 6.6.78-v8+ #1854 SMP PREEMPT Mon Feb 17 16:36:07 GMT 2025 aarch64 GNU/Linux
mean: 1125847, std dev: 4469.304004

# Top-of-tree 6.12
Linux raspberrypi 6.12.28-v8+ #46 SMP PREEMPT Mon May 19 14:35:01 BST 2025 aarch64 GNU/Linux
mean: 1251365, std dev: 19056.459418

# And with this PR applied
Linux raspberrypi 6.12.28-v8+ #50 SMP PREEMPT Mon May 19 15:25:05 BST 2025 aarch64 GNU/Linux
mean: 1133894, std dev: 16224.792260
```
You can see that there is still a small performance regression with the patched kernel, but it has been reduced from ~11% to less than 1%.

Roland-Smith-Imperial:
Many thanks for the patch, downloaded and running with no obvious issues.

Testing with 1E6 dsequential samples running a time-stamp, GPIO -> out, 3uSec delay, SPI Read Back on an AD7606 using Python code, priority set as chrt 99 gives the following sample rates :-

6.6.31 = 37.0 kHz

6.12.25 (unpatched) = 34.9k Hz (~6% performance drop from 6.6.31)

6.12.28 (GPIO Patch applied) = 35.5 kHz (~4% performance drop from 6.6.31)

Patched performance is clearly an improvement over unpached 6.12.25 thanks, not quite the 1% hit for pure GPIO high-low toggles in a loop, though we are alternating GPIO / SPI read so a slightly more complex situation.  Histograms to follow.
SPI_AD7606_ADC_Sample_Time_Performance_6 6 31
SPI_AD7606_ADC_Sample_Time_Performance_6 12 25
SPI_AD7606_ADC_Sample_Time_Performance_6 12 28_GPIO_Patch

pelwell:
This seems like a step in the right direction for performance.

Issue #5821, opened 2024-01-04
Pi 5 CSI-2: Camera can drop the first frame after stop and restart


njhollinghurst:
### Describe the bug

When using a CSI-2 camera with Raspberry Pi 5 (RP1 CFE), after stopping streaming and starting anew (e.g. using `rpicam-vid`), the first frame seems intermittently to be lost.

It's most likely because the CFE driver puts the SNPS CSI-2 Host into reset during `cfe_stop_streaming()` -> `csi2_close_rx()` -> `dphy_stop()`. Simulations suggest that this can result in a truncated data packet over the IDI bus, which confuses CSI2AXI and causes one or more later packets to be discarded before it gets back into sync.

Possibly the fix (if there is one) is either to reset CSI-2 Host _less_ aggressively or reset CSI2AXI _more_ aggressively, so that they remain in sync; or failing that, to defer stopping CFE until the camera has finished streaming.

cc @naushir @6by9 

### Steps to reproduce the behaviour

Not easily - need e.g. IM477 with synchronized cameras or external trigger to show the lost frame.

### Device (s)

Other

### System

rpi-6.1.y

njhollinghurst:
Moving the call to `dw_csi_host_write(dphy, RESETN, 0);` out of `dphy_stop()` and into the start of of `dphy_start()` [sic], and removing `dw_csi_host_write(dphy, NLANES, 0);` altogether, seems to help.

But now we are never shutting down the CSI-2 receiver. There may be power consumption implications. Not clear if leaving it active unused might open the door to some other bad state?

Maybe it would be better to use a timer to delay the shutdown (assuming the camera will stop). Will have a think about it and might PR it.

naushir:
Perhaps we call ``dphy_stop()`` when we close all the nodes, i.e. ``v4l2_file_operations.release``?  We use the default handler for this right now, but could easily override that.

njhollinghurst:
It may depend on the camera and whether it has a "fast stop" setting (in which case it might stop after a scanline?) or not (in which case it might be an entire frame time?)

6by9:
> It may depend on the camera and whether it has a "fast stop" setting (in which case it might stop after a scanline?) or not (in which case it might be an entire frame time?)

I was about to say that the power to the sensor is normally killed fairly soon after stop, however that can't now be guaranteed with the DT updates to make external trigger modes happy.

njhollinghurst:
[...]

PR #6863, opened 2025-05-19
Candidate fixes for camera sync issues

njhollinghurst:
Two issues currently, one is #5821 and is specific to Raspberry Pi 5.

There is also an IMX477 issue, where there are very rarely extra XVS pulses, causing receiving cameras to record a blank frame.

### DIFF
--- a/drivers/media/i2c/imx477.c
+++ b/drivers/media/i2c/imx477.c
@@ -112,6 +112,9 @@ MODULE_PARM_DESC(trigger_mode, "Set vsync trigger mode: 1=source, 2=sink");
 #define IMX477_REG_XVS_IO_CTRL		0x3040
 #define IMX477_REG_EXTOUT_EN		0x4b81
 
+/* Temperature sensor */
+#define IMX477_REG_TEMP_SEN_CTL		0x0138
+
 /* Embedded metadata stream structure */
 #define IMX477_EMBEDDED_LINE_WIDTH 16384
 #define IMX477_NUM_EMBEDDED_LINES 1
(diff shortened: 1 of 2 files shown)


drivers/media/i2c/imx477.c
```diff
@@ -112,6 +112,9 @@ MODULE_PARM_DESC(trigger_mode, "Set vsync trigger mode: 1=source, 2=sink");
 #define IMX477_REG_XVS_IO_CTRL		0x3040
 #define IMX477_REG_EXTOUT_EN		0x4b81
 
+/* Temperature sensor */
+#define IMX477_REG_TEMP_SEN_CTL		0x0138
```
6by9, line 116: As you've defined this, it'd be nice to use that define in `mode_common_regs` so all instances of 0x0138 are by name.

drivers/media/i2c/imx477.c
```diff
@@ -1833,6 +1836,8 @@ static int imx477_start_streaming(struct imx477 *imx477)
 
 	/* Set vsync trigger mode: 0=standalone, 1=source, 2=sink */
 	tm = (imx477->trigger_mode_of >= 0) ? imx477->trigger_mode_of : trigger_mode;
+	imx477_write_reg(imx477, IMX477_REG_TEMP_SEN_CTL,
```
6by9: Seeing as the enable is conditional at line 1855, make this line conditional too?
We've already enabled it in `mode_common_regs` otherwise.

6by9 (commented):
Largely looks fine. Two very minor comments.

6by9:
My comments are largely in regard to the imx477 change. I know very little of the CFE DPHY side.

drivers/media/i2c/imx477.c
njhollinghurst: I was trying to cover the case where we don't re-send the common registers (e.g. on a mode switch). Not that anyone should be trying to synchronize cameras across a mode switch!

OTOH keeping that write in the common setup is an attempt to ensure we never get a frame with a spurious temperature of zero.

njhollinghurst: Ah, but I see it is safe to make them both conditional in that case

njhollinghurst:
I think we need these changes (or something similar) for reliable cam->cam sync of IMX477 on Pi5.

But I guess we should hold off merging, until we get a comment from Sony about the temperature sensor issue. Perhaps they will recommend some simpler fix -- a chicken bit perhaps?

njhollinghurst:
Swapped two register writes in the second (IMX477) commit to make it closer to manufacturer's recommendation.
I think this is good to go now.

drivers/media/i2c/imx477.c
pelwell: There are currently no names used in mode_common_regs, and counterexamples where suitable names exist (IMX477_REG_LINE_LENGTH, IMX477_REG_LINE_LENGTH, ...).

PR #6865, opened 2025-05-21
overlays: sc16is75x: Add generic SPI overlay

nbuchwitz:
Currently four nearly identical overlays (sc16is750-spi0 / -spi1 and sc16is752-spi0 / -spi1) are provided. Besides the redundant configuration all of them lack support for other SPI interfaces than spi0 and spi1.

Thus refactor the common definitions into a generic sc16is75x-spi overlay which provides support for all known spi / cs combinations. Also choose the chip type via dtparam rather than different overlay.

In order to keep existing setups keep the old overlays, but add a deprecation message via overlay map.


nbuchwitz:
As talked about yesterday, I will think about some generic solution in order to avoid duplication of the boilerplate spi cs tuples. But this is stuff for another PR (once I found something which will work for most of the overlays)

pelwell:
I'm looking at this now. The deprecation breaking the overlay check is unfortunate, but I think it would be fixed by actually deleting the original overlays. A more serious problem is the reference to labels that don't exist in the older Pis' dtbs, eg. `spi3`. This can be avoided by replacing the `target` property with `target-path=spi3` etc., making use of the fact that paths that don't start with a `/` are assumed to begin with the name of an alias, which is then resolved.

If you don't mind sticking to the old naming scheme - `int_pin` and `xtal` - we should be able to replace the `deprecated` error messages with e.g.:
```
	sc16is750-spi0 {
		rename = "sc16is75x,sc16is750,spi0-0";
	};
```
I'm currently getting an error about overlay 'sc16is750-spi0' not being supported on the bcm2835 platform, but I should be able to fix (or at least understand) that.

pelwell:
Doh, it's obvious - the correct syntax is:
```
	sc16is750-spi0 {
		renamed = "sc16is75x,sc16is750,spi0-0";
	};
```
Note the `d` on `renamed`.

nbuchwitz:
Thanks, the changes sound resonable. I will update  the branch

pelwell:
I'd hang on until I've got it all working, especially the automated checks. When I've done that I'll push a patch with my changes.

pelwell:
Assuming the other tests pass, we can fix up the over-long commit message when we "Squash and merge".

nbuchwitz:
Looks good so far. Will do some tests with real hardware and then this should be ready

nbuchwitz:
Tests look good so far. From my side this is ready to be squashed and merged

PR #6868, opened 2025-05-23
drivers: pci: hailo: Better lock handling when calling find_vdma()

naushir:
Due to possible instabilities, reduce the mmap read lock time to only cover the call to find_vdma().


naushir:
Still a WIP - don't want to mark as draft yet until the CI produces a kernel image.

pelwell:
That shouldn't affect the builds - it has no effect on the tree.

naushir:
> That shouldn't affect the builds - it has no effect on the tree.

Hmm, I was sure creating draft PRs would hold off running the actions...

naushir:
This can be merged now.

PR #6877, opened 2025-06-02
drivers: media: imx500: Enable sensor temperature monitoring

naushir:
The register needs to be disbled before loading any firmware, otherwise the upload fails for unknown reasons. Re-enable before starting the sensor streaming.


roliver-rpi:
Looks good to me.

roliver-rpi (approved):

Issue #4358, opened 2021-05-23
Add CM4 supporting for kernel 4.19.y (back porting)


micronrobot:
Hello,
Due to personal reason, I must use kernel 4.19.y for my system with CM4 module. I have tried to add CM4 supporting to kernel 4.19.y by creating the bcm2711-rpi-cm4.dts file from bcm2711-rpi-4-b.dts file. I've patched my bcm2711-rpi-cm4.dts file based on the differences between bcm2711-rpi-cm4.dts and  bcm2711-rpi-4-b.dts in kernel 5.4.y. My CM4 system can boot with new bcm2711-rpi-cm4.dtb file but I pcie and usb doesn't work. 
Does anyone have tip what I should do more to get CM4 fully work in kernel 4.19.y? 
Thank you.

wirelessOJ:
I am trying to do the same. I need to use kernel 4.19y, but need PCIe to work. Did you ever manage to get this to work?

JamesH65:
It's likely this will never be possible without a lot of effort. You should find out why you need such an ancient kernel (2018 release, although EOL wasn't until Dec 2024) . Fixing that would be a better use of time.

wirelessOJ:
I am trying to use a patch from the WiFi PTP project repo. It is an ath9k kernel patch designed specifically for kernel 4.19. It seemed like it might have been simpler at first to modify the kernel to enable PCIe rather than trying to modify the patch. But perhaps I will look into modifying the patch for a newer kernel. Thank you.

popcornmix:
> But perhaps I will look into modifying the patch for a newer kernel

That sounds like a much more useful solution.

I'll close this as we'll obviously not be backporting anything to 4.19 kernel.

pelwell:
Although this forum isn't meant for general support, if you do get stuck with the port we may be able to help.

wirelessOJ:
It seems that I might not be backporting as some people think it is not possible to enable PCIe this way. Do you have other thoughts on this?

pelwell:
If you can do it on 6.12 then you can do it on 4.19, but you may have to back-port so many commits that you end up with a  kernel looking suspiciously like 6.12.

wirelessOJ:
hmmmm. I am no expert so I am not sure if it is worth to try that or to modify the kernel patch designed for 4.19 to be adapted for a later kernel

pelwell:
You should definitely try the forward-port to 6.12.

[...]

PR #6882, opened 2025-06-04
wifi: ath9k: Add PTP patch from wifi-ptp project

6by9:
Applies and updates the patch from
https://github.com/zlab-pub/wifi-ptp/blob/main/ath9k.diff

### DIFF
--- a/drivers/net/wireless/ath/ath9k/ptp.c
+++ b/drivers/net/wireless/ath/ath9k/ptp.c
@@ -1,0 +1,124 @@
+#include "ath9k.h"
+
+#include <linux/ktime.h>
+#include <linux/clocksource.h>
+#include <linux/ptp_clock_kernel.h>
+
+static int ath9k_phc_adjfine(struct ptp_clock_info *ptp, long scaled_ppm) {
+#if 0
+    struct ath_softc *sc = container_of(ptp, struct ath_softc, ptp_clock_info);
+    unsigned long flags;
+    int neg_adj = 0;
+    u32 mult, diff;
+    u64 adj;
+
+    if (scaled_ppm < 0) {
+        neg_adj = -1;
+        scaled_ppm = -scaled_ppm;
+    }
+    mult = sc->cc_mult;
+    adj = mult;
+    adj *= scaled_ppm;
+    diff = div_u64(adj, 1000000000ULL);
+
+    spin_lock_irqsave(&sc->systim_lock, flags);
+    timecounter_read(&sc->tc);
+    sc->cc.mult = neg_adj ? mult - diff : mult + diff;
+    spin_unlock_irqrestore(&sc->systim_lock, flags);
+
+    ath_warn(ath9k_hw_common(sc->sc_ah), "phc adjust adj=%llu freq=%u\n", adj, diff);
+#endif
+
+    return 0;
+}
+
+static int ath9k_phc_adjtime(struct ptp_clock_info *ptp, s64 delta) {
+    struct ath_softc *sc = container_of(ptp, struct ath_softc, ptp_clock_info);
+    unsigned long flags;
+
+    spin_lock_irqsave(&sc->systim_lock, flags);
+    timecounter_adjtime(&sc->tc, delta);
+    spin_unlock_irqrestore(&sc->systim_lock, flags);
+
+    ath_warn(ath9k_hw_common(sc->sc_ah), "phc adjust abs: %lld\n", delta);
+
+    return 0;
+}
+
+static int ath9k_phc_gettime(struct ptp_clock_info *ptp, struct timespec64 *ts) {
+    struct ath_softc *sc = container_of(ptp, struct ath_softc, ptp_clock_info);
+    unsigned long flags;
+    u64 ns;
+
+    spin_lock_irqsave(&sc->systim_lock, flags);
+    ns = timecounter_read(&sc->tc);
+    spin_unlock_irqrestore(&sc->systim_lock, flags);
+
+    *ts = ns_to_timespec64(ns);
+
+    return 0;
+}
+
+static int ath9k_phc_settime(struct ptp_clock_info *ptp,
+                const struct timespec64 *ts) {
+    struct ath_softc *sc = container_of(ptp, struct ath_softc, ptp_clock_info);
+    unsigned long flags;
+    u64 ns;
+
+    ns = timespec64_to_ns(ts);
+    spin_lock_irqsave(&sc->systim_lock, flags);
+    timecounter_init(&sc->tc, &sc->cc, ns);
+    spin_unlock_irqrestore(&sc->systim_lock, flags);
+
+    return 0;
+}
+
+static int ath9k_phc_enable(struct ptp_clock_info __always_unused *ptp,
+                struct ptp_clock_request __always_unused *request,
+                int __always_unused on) {
+    return -EOPNOTSUPP;
+}
+
+static const struct ptp_clock_info ath9k_ptp_clock_info = {
+    .owner      = THIS_MODULE,
+    .n_alarm    = 0,
+    .n_ext_ts   = 0,
+    .n_per_out  = 0,
+    .n_pins     = 0,
+    .pps        = 0,
+    .adjfine    = ath9k_phc_adjfine,
+    .adjtime    = ath9k_phc_adjtime,
+    .gettime64  = ath9k_phc_gettime,
+    .settime64  = ath9k_phc_settime,
+    .enable     = ath9k_phc_enable,
+};
+
+void ath9k_ptp_init(struct ath_softc *sc) {
+    sc->ptp_clock = NULL;
+
+    sc->ptp_clock_info = ath9k_ptp_clock_info;
+
+    snprintf(sc->ptp_clock_info.name,
+        sizeof(sc->ptp_clock_info.name), "%pm",
+        sc->hw->wiphy->perm_addr);
+
+    sc->ptp_clock_info.max_adj = 1e6;
+
+    sc->ptp_clock = ptp_clock_register(&sc->ptp_clock_info, sc->dev);
+
+    if (IS_ERR(sc->ptp_clock)) {
+        sc->ptp_clock = NULL;
+        ath_err(ath9k_hw_common(sc->sc_ah), "ptp_clock_register failed\n");
+    } else if (sc->ptp_clock) {
+        ath_info(ath9k_hw_common(sc->sc_ah), "registered PHC clock\n");
+    }
+}
+
+void ath9k_ptp_remove(struct ath_softc *sc) {
+    if (sc->ptp_clock) {
+        ptp_clock_unregister(sc->ptp_clock);
+        sc->ptp_clock = NULL;
+        ath_info(ath9k_hw_common(sc->sc_ah), "removed PHC clock\n");
+    }
+}
+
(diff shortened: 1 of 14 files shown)


6by9:
I am NOT proposing that this gets merged here, but created it for #4358

drivers/net/wireless/ath/ath9k/ptp.c
```diff
@@ -0,0 +1,124 @@
...
+    u64 adj;
+
+    if (scaled_ppm < 0) {
+        neg_adj = -1;
+        scaled_ppm = -scaled_ppm;
+    }
+    mult = sc->cc_mult;
+    adj = mult;
+    adj *= scaled_ppm;
+    diff = div_u64(adj, 1000000000ULL);
```
jacob-keller: The mistake here is still using 1,000,000,000 (1 billion). You need to use 1million * 2^16, but you can actually just replace this all with adjust_by_scaled_ppm.

drivers/net/wireless/ath/ath9k/ptp.c
```diff
@@ -0,0 +1,124 @@
...
+
+    spin_lock_irqsave(&sc->systim_lock, flags);
+    timecounter_read(&sc->tc);
+    sc->cc.mult = neg_adj ? mult - diff : mult + diff;
+    spin_unlock_irqrestore(&sc->systim_lock, flags);
+
+    ath_warn(ath9k_hw_common(sc->sc_ah), "phc adjust adj=%llu freq=%u\n", adj, diff);
+#endif
+
+    return 0;
```
jacob-keller:
```suggestion
    struct ath_softc *sc = container_of(ptp, struct ath_softc, ptp_clock_info);
    unsigned long flags;

    spin_lock_irqsave(&sc->systim_lock, flags);
    timecounter_read(&sc->tc);
	sc->cc.mult = adjust_by_scaled_ppm(sc->cc_mult, scaled_ppm);
    spin_unlock_irqrestore(&sc->systim_lock, flags);

    ath_dbg(ath9k_hw_common(sc->sc_ah), ATH_DBG_CONFIG, "phc adjust adj=%llu freq=%u\n", adj, diff);

    return 0;
```

jacob-keller (commented):
Here's what I think you want to implement, assuming sc->cc_mult is the base multiplier. I also suggested converting the ath_warn to ath_dbg so that its not spammy unless enabled as frequency can be adjusted multiple times per second depending on the synchronization rate. Hope this helps!

drivers/net/wireless/ath/ath9k/ptp.c
jacob-keller: The whitespace is probably not correct here for kernel style, but you should get the idea. You could also pull adjust_by_scaled_ppm to outside the lock, but I don't think its necessary, since its really just a few multiplications and division.

6by9: Whitespace and coding style are a bit of a mess generally with these changes, but that can be a task for others to fix up if they were ever to think of upstreaming them (I have no idea if it would even be considered upstream).
I did try a `checkpatch --fix-inplace`, but it only addressed a small fraction of the issues.

drivers/net/wireless/ath/ath9k/ptp.c
wirelessOJ: thank you!

wirelessOJ:
@jacob-keller @6by9  @CALMorACT When I ran` sudo./ptp4l -I wlan0 -p /dev/pptp0 -m `there was an error that suggested a driver issue. I wondered if this was due to how the ath9k.diff was forward ported. Looking at the logs using `sudo dmesg | grep ath9k` 

[    4.181657] ath9k 0000:01:00.0: enabling device (0000 -> 0002)
[    4.360628] ath9k 0000:01:00.0: Failed to initialize device
[    4.360662] ath9k 0000:01:00.0: probe with driver ath9k failed with error -12

It looks like the device (PCI 0000:01:00.0) tried to load with ath9k but failed.
 
I am wondering if I could have assistance troubleshooting this. 

Thank you!

pelwell:
Error 12 is ENOMEM, a pretty fundamental failure while probing. I don't believe it's actually run out of memory, but perhaps it's because of the PCIe space exposed by the card. What does `lspci -v` report?

jacob-keller:
Ya. That looks like a pretty fundamental issue with loading the wireless driver.

wirelessOJ:
`lspci -v` reports:

```
00:00.0 PCI bridge: Broadcom Inc. and subsidiaries BCM2711 PCIe Bridge (rev 20) (prog-if 00 [Normal decode])
      Device tree node: /sys/firmware/devicetree/base/scb/pcie@7d500000/pci@0,0
      Flags: bus master, fast devsel, latency 0, IRQ 27
      Bus: primary=00, secondary=01, subordinate=01, sec-latency=0
      Memory behind bridge: 00000000-000fffff [size=1M] [32-bit]
      Prefetchable memory behind bridge: [disabled] [64-bit]
      Capabilities: <access denied>
      Kernel driver in use: pcieport

01:00.0 Network controller: Qualcomm Atheros AR93xx Wireless Network Adapter (rev 01)
      Subsystem: Qualcomm Atheros AR93xx Wireless Network Adapter
      Flags: fast devsel, IRQ 27
      Memory at 600000000 (64-bit, non-prefetchable) [size=128K]
      Expansion ROM at 600020000 [virtual] [disabled] [size=64K]
      Capabilities: <access denied>
      Kernel modules: ath9k
```

wirelessOJ:
so isn't this a problem with how the ath9k driver was forward ported?

wirelessOJ:
It looks like it might be an issue with the dma_mask settings in the ath9k driver kernel module code.  Can someone point me to where this in the in driver code?

```
grep -i cma /proc/meminfo

CmaTotal:              0 kB

CmaFree:               0 kB
```

`CmaTotal: 0 kB` → I have no Contiguous Memory Allocator (CMA) region reserved

wirelessOJ:
@6by9 @pelwell @jacob-keller @CALMorACT If anyone has any advice, I believe something is wrong with the kernel and I am uncertain how to move forward

jacob-keller:
Unfortunately I don't have much experience with wireless drivers :(

wirelessOJ:
I ended up adding `cma=256M `and `dtoverlay=pcie-32bit-dma` to the` config.txt` file and that has resolved the ath9k initialization errors I was seeing. When I run `lscpi -v` I can now see too that the kernel driver ath9k is being used.

jacob-keller:
Try increasing the tx_timestamp_timeout config value for ptp4l. The timestamps may not be returned within the default time, depending on driver/hardware implementation. If a value of 100 or 200 milliseconds still fails then it's likely a bug with the driver never reporting a timestamp for one of the packets.

wirelessOJ:
100 and 200 result in the same error. I am wondering... 

when I run `ls /sys/class/ptp` there is a `ptp0` and when I then run `readlink /sys/class/ptp/ptp0/device`

it points to:

`../../../unimac-mdio--19:00`

but then when I run  `ls -l /sys/class/net/*/device`

lrwxrwxrwx 1 root root 0 Jul 14 10:39 /sys/class/net/eth0/device -> ../../../fd580000.ethernet

lrwxrwxrwx 1 root root 0 Jul 14 10:39 /sys/class/net/wlan0/device -> ../../../0000:01:00.0

lrwxrwxrwx 1 root root 0 Jul 14 10:39 /sys/class/net/wlan1/device -> ../../../mmc1:0001:1

It shows that the ptp0 devices is the onboard ethernet MDIO interface. 

**IF** the ath9k driver patch was applied correctly, shouldn't there be a `wlan0 ptp0` device present? Likewise, I am wondering if the driver is imported correctly, if the error I first see after running` ./ptp4l `that says` interface 'wlan0' does not support requested timestamping mode` would still show up as an error no matter what? 

I am looking at this issue: https://github.com/zlab-pub/wifi-ptp/issues/8

"Please ensure the "ls /sys/class/net/${WLAN_DEV}/device/ptp" has the right output, it means there is a available PHC/PTP clock for you and the driver patch successfully. You can read the 'dmesg' output for more info if this step is error for you"

I currently do not have a device under wlan0 that is a ptp device. I am thinking that this might indicate a problem with the way the driver patch was applied to this kernel.

jacob-keller:
Does your ptp4l command hardcode the ptp device? If so, you should stop doing that and rely on the ethtool -T autodetection. What does ethtool -T show for the wlan0 device?

jacob-keller:
If you use the wrong ptp device for a given network interface things will not work. Maybe the ath9k driver didn't implement the get_ts_info appropriately?

wirelessOJ:
Yes, I don't believe it is implementing as it should. There is no phc_device associated with wlan0.

This is what it shows for the` wlan0` interface

```
ethtool -T wlan0
Time stamping parameters for wlan0:
Capabilities:
      software-receive
      software-system-clock
PTP Hardware Clock: none
Hardware Transmit Timestamp Modes: none
Hardware Receive Filter Modes: none
```

jacob-keller:
You need to make sure the .get_ts_info callback for ethtool is implemented
and reports the clock number. When i am back at a real PC I will take a
look and see if I can propose something

wirelessOJ:
@jacob-keller thank you! I am also just worried because I do not see the correct output for `ls /sys/class/net/${WLAN_DEV}/device/ptp`

when I run `ethtool -T wlan0` the output is:

```
Time stamping parameters for wlan0:
Capabilities:
              software-receive
              software-system-clock
PTP Hardware Clock: none
Hardware Transmit Timestamp Modes: none
Hardware Receive Filter Modes: none 

```

I would think if the linuxptp patch was implemented correctly with the driver patch it should display `PTP Hardware Clock `option for the `wlan0`. 

@CALMorACT would you have any advice on this?

jacob-keller:
This has nothing to do with linuxptp project, it's the kernel driver which needs updating :)

wirelessOJ:
the unfortunate truth

wirelessOJ:
I now have the correct kernel patches compiled and applied, but when I run:

`sudo ./ptp4l -i wlan1 -p /dev/ptp0 -m -H `

the output stalls at` assuming the grade master role `

any thoughts?

jacob-keller:
That typically means it's not seeing any other announce messages on the network

wirelessOJ:
@6by9 @pelwell @jacob-keller @CALMorACT I now can correctly see a ptp device under wlan0. However, when I run ptp, software or hardware, it aborts after assuming grandmaster role. 

This is part of the output from the verbose log:
```
ptp4l[85387.147]: port 1: received link status notification
ptp4l[85387.147]: interface index 3 is up
ptp4l[85393.320]: port 1: announce timeout
ptp4l[85393.320]: port 1: LISTENING to MASTER on ANNOUNCE_RECEIPT_TIMEOUT_EXPIRES
ptp4l[85393.320]: selected local clock 000e8e.fffe.5a7295 as best master
ptp4l[85393.320]: assuming the grand master role
ptp4l[85393.321]: port 1: master tx announce timeout
ptp4l[85393.321]: port 1: setting asCapable
ptp4l[85394.320]: port 1: master sync timeout
ptp4l[85394.321]: port 1: tx sync is calling interpolate
Aborted
```
@CALMorACT do you have any advice for this?

jacob-keller:
That's very weird...

wirelessOJ:
Do you have any concept of where I can look at the patched kernel driver to resolve this?

PR #6884, opened 2025-06-05
GPIO Acquire/Release Semantics for 'AI Camera'

roliver-rpi:


drivers/clk/clk-gpio.c
```diff
@@ -44,9 +45,37 @@
...
+		return PTR_ERR(clk->gpiod);
+
+	return 0;
+}
+
+static int clk_gpio_gate_is_acquired(struct clk_hw *hw)
+{
+	struct clk_gpio *clk = to_clk_gpio(hw);
+
+	return clk->gpiod;
```
6by9: `gpiod` is a pointer, not an `int`. Returning it as an int is therefore a weird cast (potentially truncating).

I'd be tempted to return a `bool` rather than `int`.

6by9: The CI builds have failed due to this one, as they turn on warnings by default.

`../drivers/clk/clk-gpio.c:68:19: error: returning ‘struct gpio_desc *’ from a function with return type ‘int’ makes integer from pointer without a cast [-Werror=int-conversion]`

drivers/clk/clk-gpio.c
```diff
@@ -246,4 +363,5 @@ static struct platform_driver gpio_clk_driver = {
 		.of_match_table = gpio_clk_match_table,
 	},
 };
+
```
6by9: (Unneeded extra blank line)

6by9 (commented):
The rest all looks sane, even if the whole thing is rather icky in the first place.

roliver-rpi:
I believe all the review comments have been addressed. Are we happy for this to go in @6by9?

6by9 (approved):

6by9:
Looks good to me.

drivers/clk/clk-gpio.c
```diff
@@ -44,9 +45,37 @@
...
+		return PTR_ERR(clk->gpiod);
+
+	return 0;
+}
+
+static bool clk_gpio_gate_is_acquired(struct clk_hw *hw)
+{
+	struct clk_gpio *clk = to_clk_gpio(hw);
+
+	return clk->gpiod;
```
pelwell: Here you are declaring non-zero to indicate success (although I'd have preferred `!!clk->gpiod` to make the type conversion more obvious), but `clk_gpio_gate_acquire` leaves it containing an error code on failure.

pelwell (commented):
Given that there is no reference counting and no intention of re-entrancy or simultaneous use (at least by this driver), doesn't it point to a programming error for clk_gpio_gate_is_acquired to find that the GPIO is already acquired? In which case, given that clk_gpio_gate_acquire has an error return, why not drop clk_gpio_gate_is_acquired altogether and call ...acquire unconditionally?

roliver-rpi:
> Given that there is no reference counting and no intention of re-entrancy or simultaneous use (at least by this driver), doesn't it point to a programming error for clk_gpio_gate_is_acquired to find that the GPIO is already acquired? In which case, given that clk_gpio_gate_acquire has an error return, why not drop clk_gpio_gate_is_acquired altogether and call ...acquire unconditionally?

I've done some re-work; clk_gpio_gate_is_acquired cannot be removed completely though as it's required to prevent a potential nullptr dereference in .is_enabled/.is_prepared functions.

drivers/clk/clk-gpio.c
```diff
@@ -44,9 +45,37 @@
...
 
 #define to_clk_gpio(_hw) container_of(_hw, struct clk_gpio, hw)
+static int clk_gpio_gate_acquire(struct clk_hw *hw)
+{
+	struct clk_gpio *clk = to_clk_gpio(hw);
+	struct device *dev = clk->dev;
+
+	clk->gpiod = devm_gpiod_get(dev, "enable", GPIOD_OUT_LOW);
+	if (IS_ERR(clk->gpiod))
+		return PTR_ERR(clk->gpiod);
```
pelwell: You still need to NULL out clk->gpiod on error (or assign the gpiod to a temporary and only set clk->gpiod on success).

drivers/clk/clk-gpio.c
```diff
@@ -44,9 +45,40 @@
...
 
 #define to_clk_gpio(_hw) container_of(_hw, struct clk_gpio, hw)
+static int clk_gpio_gate_acquire(struct clk_hw *hw)
+{
+	struct clk_gpio *clk = to_clk_gpio(hw);
+	struct device *dev = clk->dev;
+
+	clk->gpiod = devm_gpiod_get(dev, "enable", GPIOD_OUT_LOW);
+	if (IS_ERR(clk->gpiod)) {
+		int ret = PTR_ERR(clk->gpiod);
```
pelwell, line 59: This line triggers a checkpatch warning about a missing blank line after a declaration, which is a shame since `ret` is otherwise unused. I see nothing wrong with:
```
		clk->gpiod = NULL;
		return PTR_ERR(clk->gpiod);
```

pelwell: > I see nothing wrong with

I'll get my coat.

PR #6892, opened 2025-06-11
Side-port dwc_otg fixes to dwc2

P33M:
This is a grab-bag of commits accumulated as a result of fixing things in dwc_otg and not dwc2. As dwc2 was derived from some variant of dwc_otg far in the past, the codebases have diverged.

With the recent emphasis on moving more platforms to dwc2, it's time to remove these inherited bugs. Or at least the ones that are obvious from the dwc_otg commit history.

### DIFF
--- a/drivers/usb/dwc2/hcd.c
+++ b/drivers/usb/dwc2/hcd.c
@@ -2469,10 +2469,23 @@ static int dwc2_alloc_dma_aligned_buffer(struct urb *urb, gfp_t mem_flags)
 {
 	void *kmalloc_ptr;
 	size_t kmalloc_size;
+	bool small_ctrl;
 
-	if (urb->num_sgs || urb->sg ||
-	    urb->transfer_buffer_length == 0 ||
-	    !((uintptr_t)urb->transfer_buffer & (DWC2_USB_DMA_ALIGN - 1)))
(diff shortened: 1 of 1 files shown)


P33M:
I wasn't as disciplined with linking changes to github issues as I thought - I may have to trawl back through all the PRs mentioning dwc_otg to find the failure case / hardware involved to replicate the test.

pelwell:
It will be make upstreaming easier, should somebody decide to do it.

6by9:
> It will be make upstreaming easier, should somebody decide to do it.

They seem fairly self-contained, and dwc2 is relatively low churn in mainline.
It's probably worth sending them upstream ourselves to see if anyone else cares.

6by9:
And upstream likely won't care about links to our development or issue reports if the commit texts are good (which they seem to be)

lategoodbye:
Thanks for porting this to dwc2. Since all of them are bugs, they will need a Fixes tag upstream. Otherwise these won't be backported.

Additionally there might be discussions about [699b9da](https://github.com/raspberrypi/linux/pull/6892/commits/699b9daea762d05bb79b212d9ebc191dfe8db458), because it's not clear if all dwc2 platforms are affected or just those which based on BCM2835 IP.

PR #6893, opened 2025-06-11
FIXUP: "fbdev: Allow client to request a particular /dev/fbdevN node"

6by9:
Fixes up forward port to 6.16
Squash in (currently) ee76d8549ca02d29a5add4f2596659a61b89a2bc.

### DIFF
--- a/drivers/video/fbdev/core/fbmem.c
+++ b/drivers/video/fbdev/core/fbmem.c
@@ -446,12 +446,12 @@ static int do_register_framebuffer(struct fb_info *fb_info)
 	if (num_registered_fb == FB_MAX)
 		return -ENXIO;
 
-	i = fb_info->node;
 	if (!fb_info->custom_fb_num || fb_info->node >= FB_MAX || registered_fb[fb_info->node]) {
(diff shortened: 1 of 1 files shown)


drivers/video/fbdev/core/fbmem.c
```diff
@@ -460,7 +460,6 @@ static int do_register_framebuffer(struct fb_info *fb_info)
 	if (err < 0)
 		return err;
 
-	fb_info->node = i;
 	refcount_set(&fb_info->count, 1);
```
popcornmix, line 463: Doesn't this deletion mean no one is ever setting fb_info->node?

6by9: If the caller has set `custom_fb_num` then it is also expected to have set `fb_info->node` to the desired node number.

We search if:
```
if (!fb_info->custom_fb_num || fb_info->node >= FB_MAX || registered_fb[fb_info->node]) {
```
ie caller hasn't requested to use the desired number, it's too high, or already registered.

dfc8f67c is the one user of this, and sets both `info->node` and `info->custom_fb_num`.

PR #6897, opened 2025-06-12
drivers: media: imx500: Add device id readback control

naushir:
Add a new custom control V4L2_CID_USER_GET_IMX500_DEVICE_ID to allow userland to query the device id from the IMX500 sensor eeprom.

Note that this device id can only be accessed when a network firmware has been upoloaded to the device, so cannot be cached on probe.

### DIFF
--- a/drivers/media/i2c/imx500.c
+++ b/drivers/media/i2c/imx500.c
@@ -239,6 +239,7 @@ enum pad_types { IMAGE_PAD, METADATA_PAD, NUM_PADS };
 
 #define V4L2_CID_USER_IMX500_INFERENCE_WINDOW (V4L2_CID_USER_IMX500_BASE + 0)
 #define V4L2_CID_USER_IMX500_NETWORK_FW_FD (V4L2_CID_USER_IMX500_BASE + 1)
+#define V4L2_CID_USER_GET_IMX500_DEVICE_ID (V4L2_CID_USER_IMX500_BASE + 2)
 
 #define ONE_MIB (1024 * 1024)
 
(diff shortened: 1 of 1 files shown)


naushir:
@roliver-rpi FYI

drivers/media/i2c/imx500.c
```diff
@@ -1997,7 +2021,35 @@ static int imx500_set_ctrl(struct v4l2_ctrl *ctrl)
...
+					     ctrl_handler);
+	struct i2c_client *client = v4l2_get_subdevdata(&imx500->sd);
+	u32 device_id[4] = {0};
+	int ret;
+
+	switch (ctrl->id) {
+	case V4L2_CID_USER_GET_IMX500_DEVICE_ID:
+		if (!imx500->network_written) {
+			dev_err(&client->dev, "Unable to get device id without network fw loaded\n");
+			return -EINVAL;
```
roliver-rpi: If we initialised the control as inactive and then __v4l2_ctrl_modify_flags(..., V4l2_CTRL_FLAG_INACTIVE,) when then network had been written, would this prevent userspace attempting to read the flag when it's not permissible?

naushir: I couldn't find any reference to ``__v4l2_ctrl_modify_flags`` but I can set the flags manually if I store a handle to the ctrl.  Now done!

roliver-rpi: Looks like v4l2_ctrl_activate() exists to test_and_set FLAG_INACTIVE (or test_and_clear as appropriate). It seems to send_event() to any subscribers as well.

naushir: Done and tested.

roliver-rpi:
Looks good to me!

PR #6901, opened 2025-06-16
media: i2c: Tweak default PDAF gain table in imx708 driver

njhollinghurst:
After analyzing more Raspberry Pi V3 cameras, adjust the default PDAF shield-pixel gain tables (they can still be overridden by camera OTP where programmed).


njhollinghurst:
Fortunately it's not critical to have super-accurate calibration (which is how we get away with a single line rather than a 2D table), and there is variation between camera models and individual cameras. But recent tests suggest the old values were not central.

PR #6902, opened 2025-06-16
drm/probe-helper: Ensure cmdline mode matches interlace mode

6by9:
drm_helper_probe_add_cmdline_mode was looking for a match for the width, height, and refresh rate within the EDID modes, but didn't check the interlacing flag. That meant that with video=1920x1080@50i would match any 1920x1080@50 mode that was found.
The converse would be possible too if an interlaced mode with matching resolution & refresh rate was found first.

Check the interlacing flag as well.

https://github.com/raspberrypi/bookworm-feedback/issues/422

### DIFF
--- a/drivers/gpu/drm/drm_probe_helper.c
+++ b/drivers/gpu/drm/drm_probe_helper.c
@@ -163,6 +163,10 @@ static int drm_helper_probe_add_cmdline_mode(struct drm_connector *connector)
 				continue;
 		}
 
+		if (cmdline_mode->interlace !=
+		    !!(mode->flags & DRM_MODE_FLAG_INTERLACE))
+			continue;
+
 		/* Mark the matching mode as being preferred by the user */
 		mode->type |= DRM_MODE_TYPE_USERDEF;
 		return 0;


drivers/gpu/drm/drm_probe_helper.c
```diff
@@ -163,6 +163,10 @@ static int drm_helper_probe_add_cmdline_mode(struct drm_connector *connector)
 				continue;
 		}
 
+		if (cmdline_mode->interlace ==
```
pelwell: Would you consider `!= !!`, since it's a difference we're checking for?

6by9: I had originally written it as `!= !!`, and changed it to this.
I don't mind as the compiler should be smart enough to reduce it to the same final logic.

This one ought to get upstreamed, so it would be interesting to see which way they want things at that stage.

6by9: Updated

popcornmix (approved):

PR #6903, opened 2025-06-17
configs: arm64: Enable the UDMABUF driver

pelwell:


pelwell:
For the record, the uncompressed kernel increases by 130kB (it can't be built as a module), i.e. it's smaller than the spare legacy USB driver we're currently carrying:
```
-rw-rw-r-- 1 phil phil 27433472 Jun 17 10:01 arch/arm64/boot/Image
-rw-rw-r-- 1 phil phil  9861373 Jun 17 10:09 kernel8.img
```
which becomes:
```
-rw-rw-r-- 1 phil phil 27564544 Jun 17 10:20 arch/arm64/boot/Image
-rw-rw-r-- 1 phil phil  9864471 Jun 17 10:20 kernel8.img
```

popcornmix:
Am I reading that right.
Uncompressed: +130kB
Compressed: +3kB

So the extra code/data compressed by a factor of 43x?
What is it - lots of zeros?

pelwell:
Yes, you're reading that correctly. That result is not uncommon; sometimes adding features reduces the compressed size, presumably because the changes make the resulting binary more self-similar in some way.

popcornmix (approved):

PR #6904, opened 2025-06-17
Replace imx500 network weight loading mechanism

djrscally:
Hello

Working with the imx500 lately, the requirement for the network weights to be passed as a file descriptor gave me some pause, as I was instead storing them as database objects which meant I'd have had to write them out before passing them to the driver. I thought that switching the interface to expect a buffer / size would be a bit more flexible and possibly also a bit more upstream friendly.

I also opened a parallel pull request for rpicam-apps that switches it to the new ioctl: https://github.com/raspberrypi/rpicam-apps/pull/825

Thanks
Dan

### DIFF
--- a/drivers/media/i2c/imx500.c
+++ b/drivers/media/i2c/imx500.c
@@ -24,6 +24,7 @@
 #include <media/v4l2-event.h>
 #include <media/v4l2-fwnode.h>
 #include <media/v4l2-mediabus.h>
+#include <uapi/linux/imx500.h>
 
 /* Chip ID */
 #define IMX500_REG_CHIP_ID CCI_REG16(0x0016)
@@ -238,7 +239,6 @@
 enum pad_types { IMAGE_PAD, METADATA_PAD, NUM_PADS };
 
 #define V4L2_CID_USER_IMX500_INFERENCE_WINDOW (V4L2_CID_USER_IMX500_BASE + 0)
-#define V4L2_CID_USER_IMX500_NETWORK_FW_FD (V4L2_CID_USER_IMX500_BASE + 1)
 #define V4L2_CID_USER_GET_IMX500_DEVICE_ID (V4L2_CID_USER_IMX500_BASE + 2)
 
 #define ONE_MIB (1024 * 1024)
@@ -1365,7 +1365,6 @@ struct imx500 {
 	struct v4l2_ctrl *hflip;
 	struct v4l2_ctrl *vblank;
 	struct v4l2_ctrl *hblank;
-	struct v4l2_ctrl *network_fw_ctrl;
 	struct v4l2_ctrl *device_id;
 
 	struct v4l2_rect inference_window;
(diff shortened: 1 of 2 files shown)


naushir:
@roliver-rpi for comments.

I'm not opposed to this change, although it would be nice to have the ctrl still active for a bit to maintain backward compatibility. 

From the commit message:
> It additionally requires that the network weights be stored as a file, which is not necessarily the case and would currently mean userspace would need to write them out as a file before passing them to the driver

How is this expected or intended to work?  The conversion tools all work on offline files, including the fpk generation and fpk to rpk cpio generation.  I don't see how a rpk network can be generated through the toolchain without going to a file.

You probably also want to update Picamera2 to use the new ioctl: (https://github.com/raspberrypi/picamera2/blob/c50033891450f5082740da8acba2f5f3fa94301d/picamera2/devices/imx500/imx500.py#L591).

pelwell:
Excuse the intrusion, but can't the current file descriptor mechanism be used with a pipe? That would eliminate the need for an intemediate file.

djrscally:
@naushir

> I'm not opposed to this change, although it would be nice to have the ctrl still active for a bit to maintain backward compatibility.

So both methods at the same time for a while?

> How is this expected or intended to work? The conversion tools all work on offline files, including the fpk generation and fpk to rpk cpio generation. I don't see how a rpk network can be generated through the toolchain without going to a file.

What we're envisioning is an ecosystem in which they'll be generated remotely and deployed to a board by some service, which might choose to store them as a file or as blobs in a database or whatever.

> You probably also want to update Picamera2 to use the new ioctl: (https://github.com/raspberrypi/picamera2/blob/c50033891450f5082740da8acba2f5f3fa94301d/picamera2/devices/imx500/imx500.py#L591).

Ah, thanks, I had missed that indeed.

@pelwell 

> Excuse the intrusion

No intrusion :)

> but can't the current file descriptor mechanism be used with a pipe? That would eliminate the need for an intemediate file.

I did briefly try a pipe which gave me an -EBADF error from the kernel. I didn't really spend any time trying to diagnose that problem though as I was already leaning towards this approach being better. I did also try memfd which seems to work ok though.

naushir:
> What we're envisioning is an ecosystem in which they'll be generated remotely and deployed to a board by some service, which might choose to store them as a file or as blobs in a database or whatever.

This type of workflow might not be suitable for the picamere2 library as-is.  Picamera2 is a bit more advanced compared to rpicam-apps, and we extract the neural network info from the cpio archive in the rpk file.  Without this, some of the library functionality will not be available.  Input tensor extraction comes to mind, but there may be other things.

roliver-rpi:
There's no simple way of applying atomic updates to our kernel / libcamera / picamera2 / etc packages so this would potentially be a breaking change if the control was removed immediately (especially for end-users who update their kernel via rpi-update).

Does this ioctl approach give you any additional flexibility that's not available via memfd?

PR #6905, opened 2025-06-17
Pispbe/rpi 6.6.y/split jobs handling v8

jmondi:
Version of https://gitlab.freedesktop.org/linux-media/users/jmondi/-/commits/b4/pispbe-mainline-split-jobs-handling-v6 rebased on rpi-6.6.y

This series includes a few cleanups and splits jobs creation and scheduling to reduce the time spent in irq context.

This set of changes will be required for proper supporting multi-context operations without registering multiple media graphs to userspace

UPDATE: Now changed to target rpi-6.12.y (@pelwell)

### DIFF
--- a/drivers/media/platform/raspberrypi/pisp_be/Kconfig
+++ b/drivers/media/platform/raspberrypi/pisp_be/Kconfig
@@ -2,6 +2,7 @@ config VIDEO_RASPBERRYPI_PISP_BE
 	tristate "Raspberry Pi PiSP Backend (BE) ISP driver"
 	depends on V4L_PLATFORM_DRIVERS
 	depends on VIDEO_DEV
+	depends on PM
 	depends on ARCH_BCM2835 || COMPILE_TEST
 	select VIDEO_V4L2_SUBDEV_API
 	select MEDIA_CONTROLLER
--- a/drivers/media/platform/raspberrypi/pisp_be/pisp_be.c
+++ b/drivers/media/platform/raspberrypi/pisp_be/pisp_be.c
@@ -388,10 +388,7 @@ static void pispbe_xlate_addrs(struct pispbe_job_descriptor *job,
 	ret = pispbe_get_planes_addr(addrs, buf[MAIN_INPUT_NODE],
 				     &node_group->node[MAIN_INPUT_NODE]);
 	if (ret <= 0) {
-		/*
-		 * This shouldn't happen; pispbe_schedule_internal should insist
-		 * on an input.
-		 */
+		/* Shouldn't happen, we have validated an input is available. */
 		dev_warn(node_group->pispbe->dev, "ISP-BE missing input\n");
 		hw_en->bayer_enables = 0;
 		hw_en->rgb_enables = 0;
--- a/drivers/media/platform/raspberrypi/pisp_be/pisp_be.c
+++ b/drivers/media/platform/raspberrypi/pisp_be/pisp_be.c
@@ -12,6 +12,7 @@
 #include <linux/module.h>
 #include <linux/platform_device.h>
 #include <linux/pm_runtime.h>
+#include <linux/slab.h>
 #include <media/v4l2-device.h>
 #include <media/v4l2-ioctl.h>
 #include <media/videobuf2-dma-contig.h>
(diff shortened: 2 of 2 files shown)


naushir:
@njhollinghurst FYI

njhollinghurst:
Should this be targeting 6.12 instead?

naushir:
Yes, it should be 6.12 not 6.6.  I've tagged the patches on linux-media, so if Nick is happy with these changes, we can merge this series.

njhollinghurst:
I think it's OK but I'll need to take a little time getting back up to speed on this.

- The PM changes look good. Although "hard" suspend and resume probably won't work; `pispbe_hw_init()` is never called again so the hardware won't be reinitialized (amongst other things).
- With respect to "impossible" checks: One has to be a bit careful because the user can overwrite the config buffer _at any time_ (whether it's cache clean or not). We don't have to support it, but it would be nice to survive those cases, preferably without locking up the ISP hardware.

(Edit) Ah, but now there is a `memcpy()` of course, which makes things a bit simpler!

njhollinghurst:
LGTM

jmondi:
Hi @njhollinghurst 

> I think it's OK but I'll need to take a little time getting back up to speed on this.
> 
> * The PM changes look good. Although "hard" suspend and resume probably won't work; `pispbe_hw_init()` is never called again so the hardware won't be reinitialized (amongst other things).

Commit https://github.com/jmondi/raspberrypi-linux/commit/1f2322fcc469be7dac53e05aa110190a641ef7a9 shouldn't introduce any functional difference. It simply make use of the runtime_pm helpers to properly balance its usage count, but in the end the resume/suspend function that gets called are the same ones that were called explicitly by the driver ?

> * With respect to "impossible" checks: One has to be a bit careful because the user can overwrite the config buffer _at any time_ (whether it's cache clean or not). We don't have to support it, but it would be nice to survive those cases, preferably without locking up the ISP hardware.
> 
> (Edit) Ah, but now there is a `memcpy()` of course, which makes things a bit simpler!

Yes, we copy in `.buf_prepare` into the `pispbe->config[]` array of config buffers, so later modification by users should not be a problem.

@naushir  should I push another branch for v6.12 ?

naushir:
> @naushir should I push another branch for v6.12 ?

Yes please, or simply update this PR to point to 6.12.  We don't want to update 6.6 any more.

naushir:
Actually I seem to have the magic powers to update the base, so let me do that and see if there are any conflicts.

naushir:
@jmondi annoyingly I think you are going to have to do this manually with a new PR.

pelwell:
I've tried to patch it up (updating @jmondi's 6.6 branch to point to 6.12 in the process). Take a look to check I've not broken anything - there was a minor merge conflict with the Kconfig change.

njhollinghurst:
@jmondi Yes, I think the change is good; it just reminded me of some pre-existing issues: that we are not implementing runtime PM completely, and the extra copy of `hw_enables` made at [L384](https://github.com/raspberrypi/linux/blob/d4c4662b43e94cfb67da3ecf8c0fa2c525a77169/drivers/media/platform/raspberrypi/pisp_be/pisp_be.c#L384) might be redundant now.  But those are separate matters.

jmondi:
> I've tried to patch it up (updating @jmondi's 6.6 branch to point to 6.12 in the process). Take a look to check I've not broken anything - there was a minor merge conflict with the Kconfig change.



> I've tried to patch it up (updating @jmondi's 6.6 branch to point to 6.12 in the process). Take a look to check I've not broken anything - there was a minor merge conflict with the Kconfig change.

I've locally rebased on v6.12 and the only conflict I had was in Kconfig indeed.

jmondi:
> > I've tried to patch it up (updating @jmondi's 6.6 branch to point to 6.12 in the process). Take a look to check I've not broken anything - there was a minor merge conflict with the Kconfig change.
> 
> > I've tried to patch it up (updating @jmondi's 6.6 branch to point to 6.12 in the process). Take a look to check I've not broken anything - there was a minor merge conflict with the Kconfig change.
> 
> I've locally rebased on v6.12 and the only conflict I had was in Kconfig indeed.

And I've now compared my local changes with the updated PR content and there's nothing there, so it's all good here

PR #6915, opened 2025-06-20
Add support for the Pi Vision 10.1" HMI

kronoidz:
This pull request introduces support for the upcoming SECO [Pi Vision](https://edge.seco.com/en_eu/pi-vision-10-1-cm5.html) product, which consists of an LVDS adapter and a display.
 
Specifically, it:
* Adds a driver for the Diodes pi4ioe5v6408 GPIO expander
* Enables the following four necessary modules in the bcm2712 defconfig:
  + Atmel MaxTouch driver
  + panel-lvds
  + pi4ioe5v6408 GPIO driver
  + sn65dsi83 bridge driver
* Adds the pivision dtoverlay, which gets automatically loaded by the Raspberry Pi bootloader if the EEPROM on the adapter is correctly programmed


pelwell:
How does the GPIO driver differ materially from https://github.com/raspberrypi/linux/blob/rpi-6.12.y/drivers/gpio/gpio-fxl6408.c? What happens if you put `compatible = "fcs,fxl6408";` in your overlay instead?

kronoidz:
You're right, they are equivalent and the board works with the fxl6408 driver as well.
I've removed the other driver.

pelwell:
1. Does the product really only work with Pi 5 hardware?
2. You've only enabled the required drivers for the Pi 5-specific defconfig, which uses 16kB pages. Is there a reason that your product can't be used with the more generic `kernel8.img` (built from arch/arm64/configs/bcm2711_defconfig) or the new "realtime" `kernel8_rt.img`?

arch/arm/boot/dts/overlays/pivision-overlay.dts
```diff
@@ -0,0 +1,155 @@
...
+				};
+			};
+		};
+	};
+
+	fragment@2 {
+		target-path = "/";
+		__overlay__ {
+			panel: panel {
+				compatible = "panel-lvds";
```
6by9: The guidance for use of panel-lvds is that you should have a panel specific compatible first, and fall back to panel-lvds.
eg https://github.com/raspberrypi/linux/blob/rpi-6.12.y/arch/arm/boot/dts/allwinner/sun8i-a83t-tbs-a711.dts#L74

Having any form of display timings in DT was accepted to mainline very begrudgingly, and this was their compromise to allow fixing up the timings without requiring a DT change (ABI and all that).

6by9:
> 1. Does the product really only work with Pi 5 hardware?

Use of GPIO 34 (normally that camera GPIO) for interrupts precludes it working on any board that uses a GPIO expander for the CAM_GPIO lines.
The spec also lists it as exposing PCIe and USB3 ports too, so that does make it CM5 only.

arch/arm/boot/dts/overlays/pivision-overlay.dts
kronoidz: Added panel-specific compatible: `auo,g101ean02-2`

kronoidz:
> 2. You've only enabled the required drivers for the Pi 5-specific defconfig, which uses 16kB pages. Is there a reason that your product can't be used with the more generic `kernel8.img` (built from arch/arm64/configs/bcm2711_defconfig) or the new "realtime" `kernel8_rt.img`?

No, there's no specific reason. I've updated and tested both `bcm2711_defconfig` and `bcm2711_rt_defconfig` to include the necessary drivers.

PR #6926, opened 2025-06-26
drivers: media: rpi: cfe: Avoid unpack operation for 16-bit formats

naushir:
The unpack operation is redundant for 16-bit sensor formats, don't set the hardware to do it in these cases.

### DIFF
--- a/drivers/media/platform/raspberrypi/rp1_cfe/cfe.c
+++ b/drivers/media/platform/raspberrypi/rp1_cfe/cfe.c
@@ -884,7 +884,9 @@ static void cfe_start_channel(struct cfe_node *node)
 			width = source_fmt->width;
 			height = source_fmt->height;
 
-			if (node->vid_fmt.fmt.pix.pixelformat ==
+			/* We don't need to repack in the case of 16-bit output. */
+			if (fmt->depth != 16 &&
+			    node->vid_fmt.fmt.pix.pixelformat ==
 					fmt->remap[CFE_REMAP_16BIT])
 				mode = CSI2_MODE_REMAP;
 			else if (node->vid_fmt.fmt.pix.pixelformat ==


drivers/media/platform/raspberrypi/rp1_cfe/cfe.c
```diff
@@ -884,7 +884,9 @@ static void cfe_start_channel(struct cfe_node *node)
 			width = source_fmt->width;
 			height = source_fmt->height;
 
-			if (node->vid_fmt.fmt.pix.pixelformat ==
+			/* We don't need to repack in the case of 16-bit output. */
+			if (fmt.depth != 16 &&
```
pelwell: `fmt->depth`

naushir: Groan, I even had that locally!

pelwell (approved):

naushir:
@njhollinghurst sadly I think this fix is only partially complete.  We only remove the REMAP mode for direct CSI-2 output, not the FE output.

From reading the spec, if ``mode == 2’b11`` (route to FE), then REMAP happens unconditionally.  This means we cannot apply a similar bit of logic to remove the REMAP 😢

njhollinghurst:
Yes, the only workaround when using FE is to use the 'wildcard' DT (which is equivalent to reverting https://github.com/raspberrypi/linux/commit/98b1f0dbb23faf3e96005a25f6acb95b64afce43) with the side-effect that it will then capture metadata and other DTs.

FE could _in theory_ crop out the different components, but better to avoid FE altogether if we can.

naushir:
Merging this now as it's correct, but fixing the core issue still needs quite a bit of additional work.

PR #6928, opened 2025-06-27
drm/vc4: tests: Use the correct return type in vc4_mock_atomic_add_output()

mairacanal:
The function vc4_mock_atomic_add_output() should return a pointer, even during error treatment. Use the proper macros to create pointers from the error code.


popcornmix (approved):

mairacanal:
@pelwell, does this PR need anything else to be merged?

pelwell:
Just a nudge, which you've now provided.

Issue #6815, opened 2025-04-25
vc_sm_cma: BUG: sleeping function called from invalid context


pobrn:
### Describe the bug

`vc_sm.c:get_kernel_id()` calls `idr_alloc()` with `GFP_KERNEL` under a spinlock.

(Also, does `lookup_kernel_id()` need to take the lock as well?)

### Steps to reproduce the behaviour

Start capturing from a camera with e.g. `cam` (from libcamera).

### Device (s)

Raspberry Pi 4 Mod. B

### System

```
Linux <hostname> 6.15.0-rc3-v8+ #2 SMP PREEMPT Thu Apr 24 18:47:39 CEST 2025 aarch64 GNU/Linux
```

### Logs

```
# cam -c1 -C32
[0:00:34.572735259] [311]  INFO Camera camera_manager.cpp:326 libcamera v0.5.0+155-94e7d330-dirty (2025-04-25)
[0:00:34.780727131] [312]  WARN CameraSensorProperties camera_sensor_properties.cpp:463 No static properties available for 'imx708_wide'
[0:00:34.780877790] [312]  WARN CameraSensorProperties camera_sensor_properties.cpp:465 Please consider updating the camera sensor properties database
[0:00:35.064659868] [312]  WARN RPiSdn sdn.cpp:40 Using legacy SDN tuning - please consider moving SDN inside rpi.denoise
[0:00:35.067119191] [312]  WARN CameraSensor camera_sensor_legacy.cpp:501 'imx708_wide': No sensor delays found in static properties. Assuming unverified defaults.
[0:00:35.070015603] [312]  INFO RPI vc4.cpp:407 Registered camera /base/soc/i2c0mux/i2c@1/imx708@1a to Unicam device /dev/media4 and ISP device /dev/media0
Using camera /base/soc/i2c0mux/i2c@1/imx708@1a as cam0
[0:00:35.071686595] [311]  INFO Camera camera.cpp:1258 configuring streams: (0) 800x600-XRGB8888
[0:00:35.074298429] [312]  INFO RPI vc4.cpp:576 Sensor: /base/soc/i2c0mux/i2c@1/imx708@1a - Selected sensor format: 1536x864-SBGGR10_1X10 - Selected unicam format: 1536x864-pBAA
[   35.406296] BUG: sleeping function called from invalid context at ./include/linux/sched/mm.h:321
[   35.406348] in_atomic(): 1, irqs_disabled(): 0, non_block: 0, pid: 312, name: cam
[...]

Issue #6814, opened 2025-04-24
vc_sm_cma: lockdep warnings while capturing from camera


pobrn:
### Describe the bug

dmabuf related lockdep warnings are emitted while capturing using `libcamera`

### Steps to reproduce the behaviour

1. enable lockdep in the kernel
2. start `cam -c1 -C` (or similar)
3. observe the lockdep warnings in `dma_buf_{un,}map_attachment()`

### Device (s)

Raspberry Pi 4 Mod. B

### System

* ` Linux <hostname> 6.15.0-rc3-v8+ #2 SMP PREEMPT Thu Apr 24 18:47:39 CEST 2025 aarch64 GNU/Linux`
* rpi camera module 3 (imx708)

### Logs

```
[   93.311225] ------------[ cut here ]------------
[   93.311240] WARNING: CPU: 1 PID: 325 at drivers/dma-buf/dma-buf.c:1125 dma_buf_map_attachment+0x94/0x160
[   93.325350] Modules linked in: vc4 snd_soc_hdmi_codec drm_exec drm_display_helper cec dw9807_vcm imx708 drm_client_lib drm_dma_helper v3d snd_soc_core bcm2835_codec(C) gpu_sched snd_compress bcm2835_v4l2(C) b
cm2835_isp(C) bcm2835_unicam_legacy drm_shmem_helper brcmfmac rpi_hevc_dec bcm2835_mmal_vchiq(C) drm_kms_helper v4l2_dv_timings snd_pcm_dmaengine snd_bcm2835(C) v4l2_fwnode drm v4l2_mem2mem vc_sm_cma(C) videobuf
2_vmalloc snd_pcm brcmutil v4l2_async videobuf2_dma_contig videobuf2_memops videobuf2_v4l2 i2c_mux_pinctrl drm_panel_orientation_quirks raspberrypi_hwmon videodev snd_timer i2c_mux cfg80211 i2c_brcmstb snd i2c_b
cm2835 videobuf2_common backlight rfkill raspberrypi_gpiomem mc nvmem_rmem uio_pdrv_genirq uio sch_fq_codel fuse nfnetlink ipv6
[   93.391721] CPU: 1 UID: 0 PID: 325 Comm: cam Tainted: G        WC          6.15.0-rc3-v8+ #2 PREEMPT 
[   93.400939] Tainted: [W]=WARN, [C]=CRAP
[   93.404767] Hardware name: Raspberry Pi 4 Model B Rev 1.5 (DT)
[...]

PR #6929, opened 2025-06-30
vcsm-cma fixes for lockdep warnings

6by9:
Raised as #6814 and #6815. The analysis appears correct, hence update the driver.


popcornmix (approved):

pobrn:
Thanks!

PR #6931, opened 2025-07-02
drm/tiny: Add support for Mayqueen Pixpaper e-ink panel

lc-wang:
This series adds support for the Mayqueen Pixpaper e-ink display panel, controlled via SPI.

The series includes:
- A vendor-prefix entry for "mayqueen"
- Device tree binding documentation for the Pixpaper panel
- A new DRM tiny driver using the simple KMS helper framework

Panel specs:
- Resolution: 122x250
- Format: XRGB8888
- Interface: SPI

The driver has been tested on:
- Raspberry Pi 2 Model B
- Linux kernel 6.16

All patches are signed off and follow the existing DRM/tiny structure. The goal is to provide a lightweight, open-source reference driver for integrating small e-paper displays in embedded systems using the DRM subsystem.

---

**Tested-by:** LiangCheng Wang <zaq14760@gmail.com>  
**Signed-off-by:** LiangCheng Wang <zaq14760@gmail.com>  
**Co-developed-by:** Wig Cheng <onlywig@gmail.com>


6by9:
This series really wants to go upstream to the dri-devel mailing list for review, particularly as it's missing a dtoverlay to make it useful on a Raspberry Pi.

I can make a couple of comments based on a very quick read through:
- Whilst other tinydrm drivers are using `drm_simple_display_pipeline`, recent guidance has been not to use it as it is viewed as an unneeded layer of indirection. https://lore.kernel.org/dri-devel/61d317d3-d6dd-4b9c-bf2b-4d5aa51f60bd@suse.de/
- Handling around `init_completed_successfully` looks suspicious. It shouldn't be possible to call update before init is complete.
- SPI drivers need a `struct spi_device_id` table as well as the compatible string match table, otherwise you'll get a warning logged from https://elixir.bootlin.com/linux/v6.15.4/source/drivers/spi/spi.c#L487
- Having a load of `vmap` type operations in there makes me think something may be off, but it may be necessary for this display. UDL is a different display that needs to do similar packing of image buffers, and that doesn't have to do similar mappings.

Hope that helps.

lc-wang:
Understood, thank you for the clarification.

I'll prioritize getting this driver upstream first, following the suggestions you gave. Once it lands upstream or gets reviewed, I can revisit integration with Raspberry Pi if needed.

Thanks again for the guidance!

6by9:
Closing as this should be heading upstream.

PR #6933, opened 2025-07-02
dtoverlay: Add autorepeat option to gpio-key overlay

6by9:
It's a standard option of the gpio_keys driver, so add an override for it in the overlay.


6by9:
https://forums.raspberrypi.com/viewtopic.php?t=389526

pelwell:
It looks correct - have you verified that it works?

6by9:
> It looks correct - have you verified that it works?

Yes I've tested it. 
`dtoverlay=gpio-key,gpio=4,keycode=16,autorepeat`, and I get the letter q autorepeating on my console.
Keydown event when the GPIO goes low, repeats whilst held low, and stops when taken high. (RTFM to note that `active_low` defaults to 1)

PR #6935, opened 2025-07-04
drivers: media: cfe: Workaround for 16-bit mismatch in the hardware

naushir:
Set the data type for 16-bit modes to 0 (wildcard) to workarond the 16-bit mismatch in hardware.

We also need to supress the warning about DT == 0 on start stream in these cases.


will127534:
Hi @naushir , sorry for the late testing on my side but I have issues even with this latest kernel update.
16bit mode doesn't seems to work unless I remove the following line for all 16bit modes.
`.csi_dt = 0, /* Avoid RP1 HW mismatch for 16-bit modes. */`

Any ideas?

naushir:
> 16bit mode doesn't seems to work unless I remove the following line for all 16bit modes

What errors/problems are you seeing with the latest kernel?

PR #6936, opened 2025-07-04
usb: dwc2: masquerade split-interrupt transfers

P33M:
Second attempt at porting the dwc_otg hack to avoid dropped split transactions.

Transactions now "look OK" on the bus analyser but this needs more soak testing with different hub implementations.


pelwell:
GitHub notifications about status changes and pushes are hit and miss.

jose1711:
Related: #7282

PR #6937, opened 2025-07-04
Update bcm2712_defconfig

storm-345016:
Add CLS_U32_PERF, NET_CLS_FLOWER and NET_CLS_MATCHALL to match standard Debian and allow use of these for tc


pelwell:
Matching Debian is not enough of a justification. Explain why these settings are useful, and what "tc" is. If we agree that the change is acceptable, we'd want it applied to more than just the 16k page variant.

storm-345016:
> Matching Debian is not enough of a justification. Explain why these settings are useful, and what "tc" is. If we agree that the change is acceptable, we'd want it applied to more than just the 16k page variant.

In this case we're talking about tc (traffic control) in the kernel - the matchall filter is needed (amongst other reasons) if you want to apply inbound queue discipline. The flower filter is for more fine-grained packet matching.

CLS_U32_PERF gives performance information very useful for finetuning qdisc/filter settings.

pelwell:
That sounds a long way from the interests of most users. What's the impact on kernel size, particularly of  CONFIG_CLS_U32_PERF=y?

storm-345016:
The kernel_2712.img file I have produced with the standard build instructions and just these changes applied shows a size of 9968014 as opposed to the one provided via the normal packages at 9962173. I think that's 6.12.34 vs 6.12.35 though - would have to rebuild to get the exact number (and I will if you want!)

If you use it as a router (the Pi 5 performs shockingly well as an NGFW) these are very useful things to have as you can then use things like the CAKE qdisc to knock bufferbloat entirely on the head both ways rather than just upload with the present kernel options.

pelwell:
Our rule of thumb for weighing up the cost of adding a new feature is to divide the unit cost by the fraction of the user population who are likely to benefit from it (not how many _could_ benefit from it if they were made to try it), so the smaller the fraction the higher the perceived cost.

In this case the increase in kernel size is minimal, and the added modules are also not that large (a total of 26 kB compressed), so I'm inclined to accept the additions for all configuration.

Any objections, @popcornmix?

pelwell:
Now implemented as https://github.com/raspberrypi/linux/pull/6938.

storm-345016:
Closing as implemented - that was fast, thanks!!!

PR #6939, opened 2025-07-04
iommu: bcm2712-iommu: Add locking; fix address offset; tidy

njhollinghurst:
- Now using spin_lock_irqsave in map, unmap, sync and iova_to_phys.
- Simplify bounds checks as all allocations should be in aperture.
- Use iommu_iotlb_gather_add_range(); NB gather range is inclusive.
- Fix missing address offset in bcm2712_iommu_sync_all.


njhollinghurst:
Briefly tested with `rpicam-hello` and `vlc file.h265` which between them test RP1, ISP-BE/Codec and HVS IOMMUs.

This does not fix any observed bug, but it feels like an improvement. The spin_lock was previously initialized but never used.

PR #6940, opened 2025-07-04
overlays: ili9881-7inch: Fix overlay name in preamble

nbuchwitz:
Nit pick found while investigating something else.


pelwell:
Thanks, Nicolai!

PR #6942, opened 2025-07-05
Add support for phy interrupts on BCM54213PE

nbuchwitz:
While working on the eee support for rp1's macb driver, I noticed that CM5 defines a phy interrupt, but it isn't used at all.  It's caused by missing pointers in the ops struct of BCM54213PE. Also the mdio address differes from pi5 (probably different strapping on the board?), which is fixed by this series.

Before:
```
[    3.501498] macb 1f00100000.ethernet eth0: PHY [1f00100000.ethernet-ffffffff:00] driver [Broadcom BCM54213PE] (irq=POLL)
```

After:
```
[    3.597582] macb 1f00100000.ethernet eth0: PHY [1f00100000.ethernet-ffffffff:00] driver [Broadcom BCM54213PE] (irq=168)
```

```
$ cat /proc/interrupts | grep 1f00100000.ethernet-ffffffff:00
168:     701264          0          0          0  pinctrl-rp1  37 Level     1f00100000.ethernet-ffffffff:00
```

### DIFF
--- a/arch/arm64/boot/dts/broadcom/bcm2712-rpi-cm5.dtsi
+++ b/arch/arm64/boot/dts/broadcom/bcm2712-rpi-cm5.dtsi
@@ -159,10 +159,10 @@ rp1_target: &pcie2 {
 	phy-reset-gpios = <&rp1_gpio 32 GPIO_ACTIVE_LOW>;
 	phy-reset-duration = <5>;
 
-	phy1: ethernet-phy@1 {
-		reg = <0x1>;
+	phy1: ethernet-phy@0 {
+		reg = <0x0>;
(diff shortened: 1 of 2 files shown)


pelwell:
The code change looks great - simply opting into the interrupt facility. The DT changes are unnecessary - the firmware already patches the PHY address (in case the address changes in the future), and the `gpio` and `rp1_gpio` are equivalent on the Pi 5 family.

nbuchwitz:
Yes, the firmware patches the address, but why not specify the correct one in the node? Agree that this is maybe a nit pit, but I prefer when the dt doesn't rely on some firmware automatisms when possible.

PR #6946, opened 2025-07-08
hevc_d: Put command FIFO into coherent memory

jc-kynesim:
Change command FIFO alloction to be from dma coherent memory rather than from generic kernel memory that we map and unmap. This removes a potential issue with unmap in an IRQ tail. In general it should also be faster as it avoids the overhead inherent in map and unmap and most streams only create short command streams.

### DIFF
--- a/drivers/media/platform/raspberrypi/hevc_dec/hevc_d_h265.c
+++ b/drivers/media/platform/raspberrypi/hevc_dec/hevc_d_h265.c
@@ -21,6 +21,15 @@
 #include "hevc_d_hw.h"
 #include "hevc_d_video.h"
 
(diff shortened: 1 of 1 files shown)


jc-kynesim:
Fixes a long standing (years) issue where very occasionally you get a bad frame out of the decoder that looks like it used the wrong reference frames.

6by9:
@jc-kynesim Is this 6.12 PR now obsolete with the changes for the mainline patchset? Can we close it?

jc-kynesim:
This patch is incorporated with the mainline stuff

6by9:
Closing then.

PR #6949, opened 2025-07-09
Sergey/ti sn65/initial

dev-riv:


pelwell:
1. We are no longer accepting patches to 6.6.
2. This is not the place to develop your code - take out all the debug messages before submitting.
3. You will have no success here without commit messages that explain what you are doing and why.

Issue #5349, opened 2023-02-13
OV9281 External Trigger Support


fxweidinger:
Hi! 

In one of our Projects, we use the Arducam OV9281. We want to transition to libcamera, since they have deprecated their MIPI Driver.

Now for the actual problem: Our product relies on the external trigger functionality. After skimming through this repo and a bit of trial-and-error, I'm not really sure this is supported in libcamera with this module.
I've seen that #5260 e6a7eed8db3c41fce11d8bf7005819f1f302a9ee added support for the OV7251 external trigger, so maybe the process would be similar?

6by9:
I thought I'd already added it, but don't see such a patch.

OV9281 was a downstream patch. Full support has now been merged into the mainline ov9282 driver, so currently from 6.2 we will be using that. I suspect I'll backport it to 6.1 as well. Changes aren't going to be made to the ov9281 driver on 5.15 now.

The process will be near identical to ov7251, but the registers involved may be different. I'll see what I can dig out.

Seeing as this repo is a request for a Linux kernel driver change, https://github.com/raspberrypi/linux is the appropriate repo. I'll see if I can find someone with appropriate permissions to move this issue.

fxweidinger:
Ah, I'm sorry, I had both repos open and got confused when opening the issue... I was wondering why the Commit Hash wouldn't link automatically, that explains it.
Sounds like good news though, thanks!

fxweidinger:
@6by9 So I looked into this a little: The way I see it, the only changes to be made are: 
| Register  | Desc.|
| ------------- | ------------- |
| 0x3030 | 0x04 Bit[2]: External trigger snapshot mode |
| 0x3666  |0x00  Bit[3:0]: Internal frame sync input select 0x0:From FSIN pin, used for both frame sync and frame trigger function 0xA: Fixed value 0|
1. Add module param for external trigger
2. Add statement in ov9282_set_steam(...) to enable/disable the trigger  https://github.com/raspberrypi/linux/blob/ddaf4dc206056ee0da4435c45db3d6a7c606b4ce/drivers/media/i2c/ov9282.c#L1035

Anything else I have perhaps overlooked? 
I might have some time on my hands to do this when I get my hands on a ov9281 again.

6by9:
Can I ask which driver you have been using that supports external trigger?

I've made the register changes that I'd expected to need for external trigger based on the datasheet, but the sensor keeps free-running.

fxweidinger:
[...]

PR #6954, opened 2025-07-11
media: ov9282: Add external trigger mode support

omeredemen:
Adds DT property `trigger-mode` to enable FSIN-triggered frame capture. Includes overlay and README update for ov9281_trig.

https://github.com/raspberrypi/linux/issues/5349#issuecomment-3058162605

### DIFF
--- a/arch/arm/boot/dts/overlays/ov9281-overlay.dts
+++ b/arch/arm/boot/dts/overlays/ov9281-overlay.dts
@@ -96,6 +96,7 @@
 		       <&reg_frag>, "target:0=",<&cam0_reg>;
 		arducam = <0>, "+6";
 		clk-continuous = <0>, "-7-8";
+		trigger-mode = <&cam_node>, "trigger-mode:0";
 	};
 };
 
--- a/drivers/media/i2c/ov9282.c
+++ b/drivers/media/i2c/ov9282.c
@@ -65,6 +65,14 @@
 #define OV9282_REG_MIPI_CTRL00	0x4800
 #define OV9282_GATED_CLOCK	BIT(5)
 
+/* Trigger mode registers */
+#define OV9282_REG_POWER_CTRL 0x4F00
+#define OV9282_REG_LOW_POWER_MODE_CTRL 0x3030
+#define OV9282_REG_NUM_FRAME_ON_TRIG 0x303F
+#define OV9282_REG_SLEEP_PERIOD_CTRL0 0x302C
+#define OV9282_REG_SLEEP_PERIOD_CTRL3 0x302F
+#define OV9282_REG_TIMING_23 0x3823
+
 /* Input clock rate */
 #define OV9282_INCLK_RATE	24000000
 
@@ -187,6 +195,7 @@ struct ov9282 {
 	const struct ov9282_mode *cur_mode;
 	u32 code;
 	struct mutex mutex;
+	int trigger_mode;
 };
 
 static const s64 link_freq[] = {
@@ -948,6 +957,52 @@ static int ov9282_get_selection(struct v4l2_subdev *sd,
 }
 
 /**
+ * ov9282_apply_trigger_config() - Configure sensor for FSIN external trigger mode
+ * @ov9282: pointer to ov9282 device
+ *
+ * Return: 0 on success, error code otherwise.
+ */
+static int ov9282_apply_trigger_config(struct ov9282 *ov9282)
+{
+	int ret;
+
+	ret = ov9282_write_reg(ov9282, OV9282_REG_MODE_SELECT,
+			1, OV9282_MODE_STANDBY);
+	if (ret)
+		return ret;
+
+	/* Low power mode */
+	ret = ov9282_write_reg(ov9282, OV9282_REG_POWER_CTRL, 1, 0x01);
+	if (ret)
+		return ret;
+
+	/* External trigger snapshot */
+	ret = ov9282_write_reg(ov9282, OV9282_REG_LOW_POWER_MODE_CTRL, 1, 0x04);
+	if (ret)
+		return ret;
+
+	/* 1 frame per trigger */
+	ret = ov9282_write_reg(ov9282, OV9282_REG_NUM_FRAME_ON_TRIG, 1, 0x01);
+	if (ret)
+		return ret;
+
+	ret = ov9282_write_reg(ov9282, OV9282_REG_SLEEP_PERIOD_CTRL0, 1, 0x00);
+	if (ret)
+		return ret;
+
+	ret = ov9282_write_reg(ov9282, OV9282_REG_SLEEP_PERIOD_CTRL3, 1, 0x7F);
+	if (ret)
+		return ret;
+
+	/* No auto wake */
+	ret = ov9282_write_reg(ov9282, OV9282_REG_TIMING_23, 1, 0x00);
+	if (ret)
+		return ret;
+
+	return 0;
+}
+
+/**
  * ov9282_start_streaming() - Start sensor stream
  * @ov9282: pointer to ov9282 device
  *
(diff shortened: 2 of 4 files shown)


pelwell:
Please squash your ov9282 commits into one, then split that in two - one for the driver changes and one for the overlay changes. You can then run `scripts/checkpatch.pl -g HEAD^` and `scripts/checkpatch.pl -g HEAD` to check for remaining indentation complaints. It would also be nice if the Signed-off-by lines weren't "Your Name". After all that, use `git push -f ...` to update your branch (and hence the PR).

With those changes, I presume you are happy with this @6by9?

pelwell:
Any comments or progress on this?

drivers/media/i2c/ov9282.c
```diff
@@ -992,15 +1060,25 @@ static int ov9282_start_streaming(struct ov9282 *ov9282)
...
+			dev_err(ov9282->dev, "failed to config external trigger mode");
+			return ret;
+		}
+		return 0;
+	}
+
 	/* Start streaming */
 	ret = ov9282_write_reg(ov9282, OV9282_REG_MODE_SELECT,
-			       1, OV9282_MODE_STREAMING);
+			1, OV9282_MODE_STREAMING);
```
6by9: Odd whitespace change for no reason

drivers/media/i2c/ov9282.c
```diff
@@ -992,15 +1060,25 @@ static int ov9282_start_streaming(struct ov9282 *ov9282)
 	}
 
 	/* Setup handler will write actual exposure and gain */
-	ret =  __v4l2_ctrl_handler_setup(ov9282->sd.ctrl_handler);
+	ret = __v4l2_ctrl_handler_setup(ov9282->sd.ctrl_handler);
```
6by9: Whitespace change

drivers/media/i2c/ov9282.c
```diff
@@ -947,6 +956,65 @@ static int ov9282_get_selection(struct v4l2_subdev *sd,
...
+
+	/* External trigger snapshot */
+        ret = ov9282_write_reg(ov9282, OV9282_REG_LOW_POWER_MODE_CTRL, 1, 0x04);
+        if (ret) {
+                return ret;
+        }
+
+	/* 1 frame per trigger */
+        ov9282_write_reg(ov9282, OV9282_REG_NUM_FRAME_ON_TRIG, 1, 0x01);
+        if (ret) {
```
6by9: You've not saved the result of `ov9282_write_reg` to `ret`, so this will never fail.

Ditto subsequent calls.

6by9 (changes requested):

stalinbeltran:
I need external trigger in order to capture laser light reflected from a moving mirror, so I tried this code with:

`sudo rpi-update pulls/6954`

It works. Tested with CAM-MIPIOV9281V2 (Innomaker) on a Raspberry 5.
If I can help somehow, please let me know. This should be made available in the official kernel (not everybody can find this pull request)

Thank you.

omeredemen:
> I need external trigger in order to capture laser light reflected from a moving mirror, so I tried this code with:
> 
> `sudo rpi-update pulls/6954`
> 
> It works. Tested with CAM-MIPIOV9281V2 (Innomaker) on a Raspberry 5. If I can help somehow, please let me know. This should be made available in the official kernel (not everybody can find this pull request)
> 
> Thank you.

Sorry for the latency. @stalinbeltran i am glad it worked for you.

drivers/media/i2c/ov9282.c
6by9: Looking again I'll acknowledge that it's wrong in the original, but changing it should be a separate patch.

drivers/media/i2c/ov9282.c
```diff
@@ -992,15 +1050,24 @@ static int ov9282_start_streaming(struct ov9282 *ov9282)
...
+			dev_err(ov9282->dev, "failed to config external trigger mode");
+			return ret;
+		}
+		return 0;
+	}
+
 	/* Start streaming */
-	ret = ov9282_write_reg(ov9282, OV9282_REG_MODE_SELECT,
-			       1, OV9282_MODE_STREAMING);
+	ret = ov9282_write_reg(ov9282, OV9282_REG_MODE_SELECT, 1, OV9282_MODE_STREAMING);
```
6by9: Again a whitespace change, and this one breaks the linux-media coding standard of max 80 chars per line. https://www.kernel.org/doc/html/latest/driver-api/media/maintainer-entry-profile.html#coding-style-addendum
```
	ret = ov9282_write_reg(ov9282, OV9282_REG_MODE_SELECT, 1,
			       OV9282_MODE_STREAMING);
```
would be permitted and cleaner, but again it shouldn't be mixed in with a functional change.

drivers/media/i2c/ov9282.c
```diff
@@ -964,13 +1023,12 @@ static int ov9282_start_streaming(struct ov9282 *ov9282)
 			{OV9282_REG_ANA_CORE_2, OV9282_ANA_CORE2_RAW8},
 		}
 	};
-	const struct ov9282_reg_list *reg_list;
+	struct ov9282_reg_list *reg_list;
```
6by9: Why have we lost the `const`?
And then we regain it in your second patch.

arch/arm/boot/dts/overlays/ov9281-overlay.dts
```diff
@@ -1,8 +1,9 @@
 // SPDX-License-Identifier: GPL-2.0-only
-// Definitions for OV9281_trig camera module on VC I2C bus
+// Definitions for OV9281 camera module on VC I2C bus
 /dts-v1/;
 /plugin/;
 
+#include <dt-bindings/gpio/gpio.h>
```
6by9: Why is this required? I don't see any references in the overlay to defines that it provides.

6by9 (changes requested):
Sorry but this seems to be worse than last time.

You haven't address pelwell's comment
> Please squash your ov9282 commits into one, then split that in two - one for the driver changes and one for the overlay changes. 

We have a number of whitespace changes. One is correct, but whitespace cleanups should never be merged into the same commit as functional changes.

omeredemen:
So i think i had made some changes in the original. and didn't squash the commits with the correct changes. i just changed the necessary part for the trigger mode and didn't make any changes on the other parts.

drivers/media/i2c/ov9282.c
```diff
@@ -998,6 +1059,16 @@ static int ov9282_start_streaming(struct ov9282 *ov9282)
...
 	}
 
+	/* Configure FSIN external trigger mode */
+	if (ov9282->trigger_mode > 0) {
+		ret = ov9282_apply_trigger_config(ov9282);
+		if (ret) {
+			dev_err(ov9282->dev, "failed to config external trigger mode");
+			return ret;
+		}
+		return 0;
```
6by9: There's a bit of a gotcha here with this early bail out that I hadn't noticed it in the previous review pass.

My preference would be to keep setting OV9282_REG_MODE_SELECT here in both cases, ie.
```
	/* Configure FSIN external trigger mode */
	if (ov9282->trigger_mode > 0) {
		ret = ov9282_apply_trigger_config(ov9282);
		if (ret) {
			dev_err(ov9282->dev, "failed to config external trigger mode");
			return ret;
		}
		/* stay in standby mode and wait for trigger signal */
		ret = ov9282_write_reg(ov9282, OV9282_REG_MODE_SELECT,
				       1, OV9282_MODE_STANDBY);
	} else {
		/* Start streaming */
		ret = ov9282_write_reg(ov9282, OV9282_REG_MODE_SELECT,
				       1, OV9282_MODE_STREAMING);
	}

	if (ret)
		dev_err(ov9282->dev, "fail to start streaming");

	return ret;
}
```

6by9 (commented):
One slight nasty here.

Seeing as this won't get upstreamed in this form I am prepared to accept it as is, but it would be nicer if you'd update.

Thanks.

drivers/media/i2c/ov9282.c
```diff
@@ -947,6 +956,58 @@ static int ov9282_get_selection(struct v4l2_subdev *sd,
...
+	if (ret)
+		return ret;
+
+	/* No auto wake */
+	ret = ov9282_write_reg(ov9282, OV9282_REG_TIMING_23, 1, 0x00);
+	if (ret)
+		return ret;
+
+	/* stay standby mode and wait for trigger signal */
+	ret = ov9282_write_reg(ov9282, OV9282_REG_MODE_SELECT,
```
6by9: Sorry I didn't say it explicitly, but we now have a duplication here. Just drop this write.

pelwell (approved):

pelwell:
FYI we don't (always?) get notifications of branch updates, so adding a comment once you've done so helps to avoid long delays.

I think this is good now. Feel free to merge if you agree, @6by9.

PR #6957, opened 2025-07-14
Update panel-waveshare-dsi-v2 driver

waveshare:
1) Adjust power enable sequence
2) Added support for:
	A) Add support for 3.4-DSI-TOUCH-C
	B) Add support for 4-DSI-TOUCH-C
	C) Add support for 8-DSI-TOUCH-A
	D) Add support for 9-DSI-TOUCH-B
	E) Add support for 10.1-DSI-TOUCH-B
	F) Add support for 12.3-DSI-TOUCH-A


6by9:
Thanks for the patches.

As this is a downstream driver that I never intend to upstream, I'll ignore various things like having `avdd-gpio` and `iovcc-gpio` rather than using the regulator framework to control the power rails - it just isn't worth the effort to implement it.
Similarly each commit should do a single thing, and "Update display driver" adds power control and a load of modes.

Nothing jumps out as being an issue, so I'm happy as long as all the CI checks pass.

PR #6959, opened 2025-07-15
Additional YUV420 10bit fmts

6by9:
Backport of Robert Mader's patch that is on drm-misc-next adding the format definitions, and then add that support to vc4.

@cillian64 

libdrm updates at https://gitlab.freedesktop.org/6by9/libdrm/-/tree/add_formats?ref_type=heads (I'll create a PR for that once I've added support for the other test patterns).


6by9:
Actually we can add DRM_FORMAT_Q01[026] as well.
I'll update this in a bit.

6by9:
> Actually we can add DRM_FORMAT_Q01[026] as well. I'll update this in a bit.

No I won't as they don't exist! Only Q410 and Q401 which are YUV444 formats.

Issue #6953, opened 2025-07-10
Ethernet Not Getting IP After Upgrade from 6.12.34 to 6.12.36


ak-rex:
### Describe the bug

After upgrading the Linux kernel from 6.12.34 to 6.12.36, eth0 is present but no longer connects.

### Steps to reproduce the behaviour

System fully updated
Boot kernel 6.12.34 with ethernet connected
run `ifconfig` in the terminal, received IP address from DHCP
Boot kernel 6.12.36 with ethernet connected
run `ifconfig` in the terminal, no IP address from DHCP
Switching back to 6.12.34 resolves the issue.

### Device (s)

Raspberry Pi CM5 Lite

### System

Model: Raspberry Pi CM5 / DevKit
OS: Debian 12 Bookworm

Kernel versions:
Working: 6.12.34
Broken: 6.12.36

nbuchwitz:
Is there any link detected? What does `dmesg` report?

Just updated a CM5 on the io board with `rpi-update` to `Linux rpicm5 6.12.36-v8+ #1892 SMP PREEMPT Mon Jul  7 17:12:23 BST 2025 aarch64 GNU/Linux` and got a IP via DHCP.

pelwell:
You say this is "OS: Debian 12 Bookworm". Is it Raspberry Pi OS or some other variant?

ak-rex:
Raspberry Pi OS

pelwell:
On 6.12.36, what does `ethtool eth0` report?

ak-rex:
running 6.12.36
<details>

<summary>dmesg</summary>

```
[    0.000000] Booting Linux on physical CPU 0x0000000000 [0x414fd0b1]
[    0.000000] Linux version 6.12.36-v8+ (rex@CM5-IO-BOARD) (aarch64-linux-gnu-gcc (Debian 12.2.0-14+deb12u1) 12.2.0, GNU ld (GNU Binutils for Debian) 2.40) #2 SMP PREEMPT Tue Jul  8 22:44:36 EDT 2025
[    0.000000] KASLR enabled
[    0.000000] random: crng init done
[    0.000000] Machine model: Raspberry Pi Compute Module 5 Rev 1.0
[    0.000000] efi: UEFI not found.
[    0.000000] Reserved memory: created CMA memory pool at 0x000000003b800000, size 64 MiB
[    0.000000] OF: reserved mem: initialized node linux,cma, compatible id shared-dma-pool
[...]

PR #6963, opened 2025-07-18
Enable interrupts on CM5 ethernet phy (again)

nbuchwitz:
Second attempt to enable interrupts on CM5's ethernet phy (see https://github.com/raspberrypi/linux/issues/6953). Took me some time, but I think I finally sorted it out:

- Interrupt seems to be attached to LED4/INT_N, but was previously overridden by the led shadow patch (magically worked on my machine as I had a led_modes config which had link status on LED4 ...)
- Interrupts were disable in ECR, but never re-enabled
- Active low GPIO is missing a pull up, so just use the bias pull up in rp1 

Happy for feedback and also some tests, so we won't missing anything out.

### DIFF
--- a/arch/arm64/boot/dts/broadcom/bcm2712-rpi-cm5.dtsi
+++ b/arch/arm64/boot/dts/broadcom/bcm2712-rpi-cm5.dtsi
@@ -158,6 +158,8 @@ rp1_target: &pcie2 {
 	phy-handle = <&phy1>;
 	phy-reset-gpios = <&rp1_gpio 32 GPIO_ACTIVE_LOW>;
 	phy-reset-duration = <5>;
+	pinctrl-names = "default";
+	pinctrl-0 = <&eth0_irq_pins>;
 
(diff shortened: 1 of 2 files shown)


nbuchwitz:
Tested on cm5io (booted with cable attached, unplug, re-plug cable):

```
pi@rpicm5:~ $ grep ethernet /proc/interrupts
168:          2          0          0          0  pinctrl-rp1  37 Level     1f00100000.ethernet-ffffffff:00
pi@rpicm5:~ $ grep ethernet /proc/interrupts
168:          3          0          0          0  pinctrl-rp1  37 Level     1f00100000.ethernet-ffffffff:00
pi@rpicm5:~ $ grep ethernet /proc/interrupts
168:          5          0          0          0  pinctrl-rp1  37 Level     1f00100000.ethernet-ffffffff:00
```

2 interrupts when cable was attached make sense (probably one for link status + one for auto neg and other stuff)

PR #6964, opened 2025-07-18
overlays: sc16is75x: Fix spi1 cs-gpios

### LINKED ISSUE #6962: 6.12.34 breaks Waveshare RS232 HAT
### Describe the bug

After upgrading from 6.12.20 to 6.12.34 the Waveshare overlay still gets loaded, however, communication via ttySC1 is no longer possible, the user app just freezes.

dmesg output for 6.12.34
```
[    4.693022] sc16is7xx spi1.0: Native CS is not supported - please configure cs-gpio in device-tree
[    4.697683] spi1.0: ttySC0 at I/O 0x0 (irq = 59, base_baud = 921600) is a SC16IS752
[    4.701597] spi1.0: ttySC1 at I/O 0x1 (irq = 59, base_baud = 921600) is a SC16IS752
```

dmesg output for 6.12.20
```
[    5.635965] spi1.0: ttySC0 at I/O 0x0 (irq = 44, base_baud = 921600) is a SC16IS752
[    5.636656] spi1.0: ttySC1 at I/O 0x1 (irq = 44, base_baud = 921600) is a SC16IS752

```

The config.txt snippet to load the overlay:
```
dtoverlay=sc16is752-spi1,int_pin=24
```

Upgrade kernel to 6.12.30, try to communicate with /dev/ttySC1, notice the app freezes and the activity LEDs on the Waveshare don't blink

Downgrade to kernel 6.12.20, communication is fine and activity LEDs blink as expected

### Device (s)

Raspberry Pi 4 Mod. B

### System

PRETTY_NAME="Debian GNU/Linux 12 (bookworm)"
NAME="Debian GNU/Linux"
VERSION_ID="12"
VERSION="12 (bookworm)"
VERSION_CODENAME=bookworm
ID=debian
HOME_URL="https://www.debian.org/"
SUPPORT_URL="https://www.debian.org/support"
BUG_REPORT_URL="https://bugs.debian.org/"

Firmware:
Apr 30 2025 13:33:39
Copyright (c) 2012 Broadcom
version 5560078dcc8591a00f57b9068d13e5544aeef3aa (clean) (release) (start)

Kernel (working):
6.12.20+rpt-rpi-v8

Kernel (failing):
6.12.34+rpt-rpi-v8

### Logs

_No response_

### Additional context

_No response_

nbuchwitz:
Native cs is somewhat broken on spi1 (see [1]), so re-add the explicit cs-gpios definition, as was done before in sc16is752-spi1-overlay.

Fixes https://github.com/raspberrypi/linux/issues/6962

[1] https://github.com/torvalds/linux/commit/519f2c22a6c71a9fefed1166c36d48246e010514

### DIFF
--- a/arch/arm/boot/dts/overlays/sc16is75x-spi-overlay.dts
+++ b/arch/arm/boot/dts/overlays/sc16is75x-spi-overlay.dts
@@ -61,6 +61,13 @@
 		};
 	};
 
+	spi_cs_fixup: fragment@100 {
+		target = <&spi1>;
+		__dormant__ {
+			cs-gpios = <&gpio 18 1>;
+		};
+	};
+
 	__overrides__ {
 		sc16is750 = <&sc16is75x>,"compatible=nxp,sc16is750",
 			<&sc16is75x>, "name=sc16is750@0";
@@ -79,7 +86,8 @@
 			<&sc16is75x_pins>, "name=sc16is75x_spi0_1_pins";
 		spi1-0 = <0>, "-0",
 			<&sc16is75x_frag>, "target:0=", <&spi1>,
-			<&sc16is75x_pins>, "name=sc16is75x_spi1_0_pins";
+			<&sc16is75x_pins>, "name=sc16is75x_spi1_0_pins",
+			<0>,"+100";
 		spi1-1 = <0>, "-0",
 			<&sc16is75x_frag>, "target:0=", <&spi1>,
 			<&sc16is75x>, "reg:0=1",


nbuchwitz:
This is a potential fix for https://github.com/raspberrypi/linux/issues/6962

pelwell:
That's not going to work - you can either build up the cs-gpios property one cell at a time, or add a dormant fragment that is enabled by this override (which feels a bit cleaner).

nbuchwitz:
I've updated the PR with a different approach and tested it successfully with the bespoke waveshare board

nbuchwitz:
@pelwell can you please re-trigger the builds, so the reporter can verify this too? Thanks.

nbuchwitz:
Another thought: Do we want to have this fixup in each and every overlay targeting spi1 or would it make more sense to leave it to the `spi1-[123]cs` overlays? `mcp251xfd-overlay.dts` for example assumes that spi is correctly configured by the spi overlays. 

In case of the sc1675x on spi1:
```
dtoverlay=spi1-1cs,cs0_spidev=off
dtoverlay=sc16is75x-spi,sc16is752,spi1-0,int_pin=24
```
(the `cs0_spidev=off` is probably something which can be done in the downstream overlays though)

pelwell:
What about 1.1, 1.2,  2.0, etc.?

nbuchwitz:
> What about 1.1, 1.2, 2.0, etc.?

Guess they are broken as before ("backwards compatibility")? I mean they could be added to the overlay, but imho the split between the spi overlays and device overlay would be a cleaner and more maintenance friendly approach?  I was also thinking about a dtsi shared between different spi device overlays, but still lots of boilerplate code in each ...

pelwell:
Having had a proper chance to look at this, I'm not convinced this patch is enough to solve the problem.

As well as enabling CS GPIOs, the old `sc16is752-spi1` overlay, like the `spi1-<n>cs` overlays, enables the SPI1 controller node and selects the correct ALT function on GPIOs 19-21. It also enables the AUX resets block (needed for SPI1 and SPI2).

In an ideal world we would have always required the use of `spi1-1cs` as well as `sc16is752-spi1`, but it's too late for that now. I considered generating a new `sc16is752-spi1` by combining `spi1-1cs` and `sc16is75x-spi` (using ovmerge via the `// redo` a comment), but that would break the `int_pin` and `xtal` parameters.

https://github.com/raspberrypi/linux/pull/6996 is an alternate approach that adds just enough to allow `sc16is75x-spi` to emulate `sc16is752-spi1`.

nbuchwitz:
#6996 looks sane. I will close this one in favor

Issue #6971, opened 2025-07-23
pwm-pio OOPS when enabling on 6.12.34


rpdom:
### Describe the bug

When testing pwm-pio on a Pi 5 2GB running an up-to-date Raspberry Pi OS 64-bit Bookworm Lite, I encountered errors which killed the current process.

The dmesg report showed WARNINGS and an OOPS.

### Steps to reproduce the behaviour

Install Bookworm Lite using imager.

Run `sudo apt update && sudo apt full-upgrade`.

Enable pwm-pio in config.txt

`dtoverlay=pwm-pio`

Reboot.

Then the following command line sequence caused the problem
```
rpdom@raspberrypi:~ $ cd /sys/class/pwm/pwmchip1/
rpdom@raspberrypi:/sys/class/pwm/pwmchip1 $ cat npwm 
1
rpdom@raspberrypi:/sys/class/pwm/pwmchip1 $ echo "0" >export 
rpdom@raspberrypi:/sys/class/pwm/pwmchip1 $ cd pwm0
rpdom@raspberrypi:/sys/class/pwm/pwmchip1/pwm0 $ echo 50 >duty_cycle 
rpdom@raspberrypi:/sys/class/pwm/pwmchip1/pwm0 $ echo 1000 > period 
rpdom@raspberrypi:/sys/class/pwm/pwmchip1/pwm0 $ echo 1 > enable 
```
At this point the shell exited and dmesg showed the trace.

### Device (s)

Raspberry Pi 5

### System

https://pastebin.com/c3WcrYjV

### Logs

rpdom@raspberrypi:~ $ dmesg
[    0.000000] Booting Linux on physical CPU 0x0000000000 [0x414fd0b1]
[    0.000000] Linux version 6.12.34+rpt-rpi-2712 (serge@raspberrypi.com) (aarch64-linux-gnu-gcc-12 (Debian 12.2.0-14+deb12u1) 12.2.0, GNU ld (GNU Binutils for Debian) 2.40) #1 SMP PREEMPT Debian 1:6.12.34-1+rpt1~bookworm (2025-06-26)
[    0.000000] KASLR enabled
[    0.000000] random: crng init done
[    0.000000] Machine model: Raspberry Pi 5 Model B Rev 1.0
[    0.000000] efi: UEFI not found.
[    0.000000] Reserved memory: created CMA memory pool at 0x000000003a000000, size 64 MiB
[...]

PR #6972, opened 2025-07-23
pwm: rp1: use pwmchip_get_drvdata() instead of container_of()

nbuchwitz:
The PWM framework may not embed struct pwm_chip within the driver’s private data. Using container_of() can result in accessing invalid memory or NULL pointers, especially after recent kernel changes.

Switch to pwmchip_get_drvdata() to reliably access the driver data. This resolves kernel warnings and probe failures seen after updating from kernel 6.12.28 to 6.12.34 [1]

[1] https://github.com/raspberrypi/linux/issues/6971

### DIFF
--- a/drivers/pwm/pwm-pio-rp1.c
+++ b/drivers/pwm/pwm-pio-rp1.c
@@ -22,7 +22,6 @@
 #include <linux/pwm.h>
 
 struct pwm_pio_rp1 {
-	struct pwm_chip chip;
 	struct device *dev;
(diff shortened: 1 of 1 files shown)


nbuchwitz:
Probably broken since the very beginning of 6.12.y

6by9:
Tested and LGTM.

You could lose the `struct pwm_chip chip;` from `struct pwm_pio_rp1` as it's not used now.

nbuchwitz:
Thanks. I've updated the PR

PR #6985, opened 2025-08-04
pisp_be: Stride alignment.

naushir:


naushir:
PS, I think some of our "optimal" alignment fields are wrong?  e.g. RGB24 should be 64 and not 32...

drivers/media/platform/raspberrypi/pisp_be/pisp_be_formats.h
```diff
@@ -494,15 +542,17 @@ static const struct pisp_be_format supported_formats[] = {
 	{
 		.fourcc		= V4L2_PIX_FMT_GREY,
 		.bit_depth	= 8,
-		.align		= 32,
+		.opt_align	= 32,
+		.min_align	= 32,
```
njhollinghurst: Should this (and the next two) min_aligns be 16 bytes?

naushir: Yes they should!  Fixed.

njhollinghurst:
Yes the 32's are a bit strange. I can only guess it was done for compatibility with Unicam or Codecs (perhaps some earlier versions of software that didn't pay enough attention to strides), or because we felt 64 bytes (for the smallest plane) was too onerous...

I think it is harmless to keep this, at least for non-wallpaper image formats and when the width is much greater than 64.

njhollinghurst (approved):

Issue #6856, opened 2025-05-15
lanes parameter missing for a few ili9881c_desc like nwe080


berkutta:
### Describe the bug

On rpi-6.6.y the nwe080 description have the lanes parameter set to 4:
https://github.com/raspberrypi/linux/blob/rpi-6.6.y/drivers/gpu/drm/panel/panel-ilitek-ili9881c.c#L2044-L2050

On rpi-6.12.y, this parameters is missing:
https://github.com/raspberrypi/linux/blob/rpi-6.12.y/drivers/gpu/drm/panel/panel-ilitek-ili9881c.c#L2506-L2511

Without lanes being specified, the display doesn't work.

### Steps to reproduce the behaviour

- rpi-6.6.y working
- rpi-6.12.y not working
- rpi-6.6.y with added lanes working

### Device (s)

Raspberry Pi CM4

### System

Raspbian

jack-obrien:
I'm running into this issue too, looks like a merge issue as the RPi devs created their kernel 6.12. The panel nwe080 is not present in the upstream kernel:
https://github.com/torvalds/linux/blob/master/drivers/gpu/drm/panel/panel-ilitek-ili9881c.c

So, the RPi devs will need to keep merging their panel configs each time they pull from upstream. It makes sense why the parameter might be missing in the 6.12 release here.

If I get the time later, I will put a pull request on this repo to just re-add the missing lanes parameter from all the panels in this driver.

6by9:
Sorry, I missed this one when it was first opened.

Whilst Raspberry Pi accept PRs for third party hardware (nwe080 was added for CutiePi https://github.com/raspberrypi/linux/pull/4742) and will do the basic forward porting of the drivers, we do NOT take on full responsibility for maintaining or necessarily testing it (in many cases we don't have hardware to test with).
If there are major issues in forward porting to a new kernel version, then drivers will get dropped and kicked back to the original contributor to fix.

PRs to fix up forward porting errors are welcome, as is upstreaming of patches (there's no reason nwe080 couldn't go upstream).

jack-obrien:
Yes, I have tested on my panel that the lanes parameter is still needed. Will get a PR ready soon!

PR #6987, opened 2025-08-05
drm/panel: ilitek-ili9881c: Restore missing lanes configuration for nwe080 panel

jack-obrien:
drm/panel: ilitek-ili9881c: Restore missing lanes configuration for nwe080

This config was missing with the forward porting of the rasp pi kernel to 6.12. Refer to https://github.com/raspberrypi/linux/issues/6856

I have tested this fix on my panel hardware. Previously it worked well with the nwe080 panel desc on rpi kernel 6.6. After updating to rpi 6.12, the panel stopped working. After I made this fix and recompiled the driver, the panel worked again.

I couldn't see any other panels with missing "lanes" parameter which were missed in the forward port from 6.6 to 6.12.

### DIFF
--- a/drivers/gpu/drm/panel/panel-ilitek-ili9881c.c
+++ b/drivers/gpu/drm/panel/panel-ilitek-ili9881c.c
@@ -2512,6 +2512,7 @@ static const struct ili9881c_desc nwe080_desc = {
 	.init_length = ARRAY_SIZE(nwe080_init),
 	.mode = &nwe080_default_mode,
 	.mode_flags = MIPI_DSI_MODE_VIDEO_SYNC_PULSE | MIPI_DSI_MODE_VIDEO,
+	.lanes = 4,
 };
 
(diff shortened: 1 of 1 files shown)


jack-obrien:
Ah, needs a whitespace fix, I'll sort that out soon

6by9:
> Ah, needs a whitespace fix, I'll sort that out soon

Thank you.
This file originates from upstream, so we do try to follow all their requirements so that our changes could be upstreamed as time allows.
Likewise wrapping the commit text appropriately at max 75chars and a Signed-off-by tag would be appreciated.

jack-obrien:
Thanks for the feedback, will sort it out now

pelwell:
I'm happy the build errors are nothing to do with this change - my local builds are fine.

6by9:
> I'm happy the build errors are nothing to do with this change - my local builds are fine.

I also had a build fail earlier with the toolchain install timing out. Rerunning worked, but it would imply that Github CI is having some issues.

PR #6990, opened 2025-08-06
HEVC decoder updates

6by9:
This brings rpi-6.12.y into line with the next version of the patchset for upstream, and allows for testing the GStreamer MR at https://gitlab.freedesktop.org/gstreamer/gstreamer/-/merge_requests/9247

We may be able to get away with just "[Revert "media: hevc_dec: Drop the new image formats until we have Mes…](https://github.com/raspberrypi/linux/commit/6998d18820cd3029a4791246daeae85c9c91543f)" to get GStreamer working, and I might be tempted to do that rather than this slightly more invasive change. I'll talk to John next week.

### DIFF
--- a/arch/arm/boot/dts/broadcom/bcm2711.dtsi
+++ b/arch/arm/boot/dts/broadcom/bcm2711.dtsi
@@ -614,7 +614,7 @@
 		};
 
 		hevc_dec: codec@7eb10000 {
-			compatible = "brcm,bcm2711-hevc-dec", "raspberrypi,hevc-dec";
+			compatible = "brcm,bcm2711-hevc-dec";
 			reg = <0x0 0x7eb00000  0x10000>, /* HEVC */
 			      <0x0 0x7eb10000  0x1000>;  /* INTC */
 			reg-names = "hevc",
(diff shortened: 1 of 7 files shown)


hcfman:
> We may be able to get away with just "[Revert "media: hevc_dec: Drop the new image formats until we have Mes…](https://github.com/raspberrypi/linux/commit/6998d18820cd3029a4791246daeae85c9c91543f)" to get GStreamer working, and I might be tempted to do that rather than this slightly more invasive change. I'll talk to John next week.

I tested  a checkout that was just this one commit reverted. The result was an images that were solid blocks of green at the end of a gstreamer pipeline.

And I tested this full pull request and at 1920/15 it pulled 175 frames then crashed. It was also able pull a couple of correct 4K frames.

At one stage I saw an error about calling a certain function, with I think a null, was an indication of a possible bug.

but it certainly translated several frames correctly through to the end of a gstreamer pipeline. So it might just be a some resources handling and edge cases.

6by9:
6.12 is obsolete now, so closing.
6.18 already has something fairly close to the mainline patchset.

PR #6994, opened 2025-08-12
Improve PIO DMA performance

pelwell:
The RP1 DMA controller configuration makes channels 1 and 2 more capable than the others. Add a mechanism to reserve those channels for the PIO block. There may be other ways to improve the DMA bandwidth when talking to the PIO block, but this alone more than doubles the throughput.


P33M:
Channels 0 and 1 were configured with MSIZE=8, not 16. Those channels will only issue 8-beat bursts in response to a request, so there will be a mismatch in FIFO fill levels vs thresholds.

pelwell:
I knew that, on some level. Updated with the 16s changed to 8s, and with the rp1-pio code changes broken out into a separate commit (as it should always have been).

PR #6995, opened 2025-08-12
arm64/dma-mapping: Fix arch_sync_dma_for_device to respect dir parameter

jc-kynesim:
All other architectures do different cache operations depending on the dir parameter. Fix arm64 to do the same.

This fixes udmabuf operations when syncing for read e.g. when the CPU reads back a V4L2 decoded frame buffer.

I suspect it also fixes less common dma_heap/system issues too. It can be argued that these problems should be fixed in udmabuf & dma_heap but fixing here also works around any other hidden bugs in other drivers. This fix does not preclude fixing the other drivers as well.

### DIFF
--- a/arch/arm64/mm/dma-mapping.c
+++ b/arch/arm64/mm/dma-mapping.c
@@ -16,8 +16,22 @@ void arch_sync_dma_for_device(phys_addr_t paddr, size_t size,
 			      enum dma_data_direction dir)
 {
 	unsigned long start = (unsigned long)phys_to_virt(paddr);
+	unsigned long end = start + size;
 
-	dcache_clean_poc(start, start + size);
+	switch (dir) {
+	case DMA_BIDIRECTIONAL:
+		dcache_clean_inval_poc(start, end);
+		break;
+	case DMA_TO_DEVICE:
+		dcache_clean_poc(start, end);
+		break;
+	case DMA_FROM_DEVICE:
+		dcache_inval_poc(start, end);
+		break;
+	case DMA_NONE:
+	default:
+		break;
+	}
 }
 
 void arch_sync_dma_for_cpu(phys_addr_t paddr, size_t size,


pelwell (approved):
This is not a comfortable patch, because it make me wonder how we got this far without it, but I think it is correct.

6by9:
Have you pinged Maxime with context as I'd suggested? Otherwise it definitely deserves a query to mainline.

jc-kynesim:
No, so far I've come back from holiday & discussed it with you. Where/how do you suggest that I ping Maxime?

pelwell:
We could try just giving Maxime (@mripard) a mention here...

6by9:
Pinging here is fine, but need to provide the context of the use case that goes wrong.

(Otherwise Libreelec slack if no reply).

jc-kynesim:
Longer form context of what I was doing to find this issue:

[Userland]
Alloc memory with memfd_create + ftruncate
Derive dmabuf from memfd with udmabuf
Close memfd
Queue dmabuf into V4L2 CAPTURE with QBUF
\<decode a video frame\>
Extract dmabuf from V4L2 CAPTURE with DQBUF
Map dmabuf for read with mmap
Sync for read with DMA_BUF_IOCTL_SYNC with (DMA_BUF_SYNC_START |
DMA_BUF_SYNC_READ)
Read buffer
Sync end
Unmap

I get old (zero) data out of the "Read buffer" stage in some cache lines.
It doesn't matter which way round the mmap & sync are

What happens in the Sync is:

udmabuf finds that none of the pages are currently mapped and maps them with dma_map_sgtable passing the direction flags to the mapping process. The mapping process picks some pages and after a few more calls in the stack get to dma_direct_map_page which flushes them with arch_sync_dma_for_device with the direction flags. This ignores the direction and does a clear, which in arm-speak appears to be writeback without invalidate - other drivers (including arm32) respect the dir flags and do one of invalidate, writeback, writeback + invalidate depending on them.

udmabuf then assumes that this is enough and returns. It appears that the pages picked can still have cached contents which are not invalidated.

In subsequent calls when the memory is mapped udmabuf does a dma_sync_sg_for_cpu which does work. (Doing the sync twice from userspace goes down this path the second time and works.)

Now it is arguable that udmabuf shouldn't rely on dma_map_sgtable to give it clear pages or that dma_direct_map_page is picking the wrong flush. However it is almost certainly a good idea for the arm64 arch_sync_dma_for_device to do the same as all the other drivers as there may be other similar fails elsewhere.

jc-kynesim:
Looks like this fixes my, noticeably rarer, dma_heap allocated dmabuf issues too so this isn't just an issue with udmabuf. If I can find 2 things that this fixes I'm not going to bet that there aren't others.

pelwell:
I'm going to merge this on the basis that it seems reasonable to all of us, and because it should get some in-the-field testing.

jc-kynesim:
Did we want to leave this open for further comment, possibly by Maxime? or is merged, but still open not an available option? Not sure what the correct thing to do is?

pelwell:
Until we lock the conversation everyone is free to comment here.

jc-kynesim:
I note that there is a set of udmabuf patches going though at the moment that might have fixed my udmabuf issues without this: "udmabuf: Sync to attached devices"

6by9:
Link to that series for those interested - https://lkml.org/lkml/2025/8/14/1574

jc-kynesim:
Well, I've had a go at upstreaming that now. On the plus side they responded very quickly and agreed that this is no less functional than what was there before and probably makes such bugs rarer but on the minus side the upstream consensus is that this should all be fixed in the other drivers that are doing stuff wrong. This isn't going to upstream. Having said that I think it is still worth having in our kernel even if all it does is make issues rarer.

For those that care the thread is "[PATCH] arm64/dma-mapping: Fix arch_sync_dma_for_device to respect dir" in inux-arm-kernel

6by9:
Link to the thread - https://lore.kernel.org/linux-arm-kernel/9bb085e7-9833-480b-8804-e9a0d2167f25@arm.com/T/#m89f5e38fad8bd74f09bdd6a70cd7be8dcac99eb9

6by9:
Interesting that other architectures do handle the different directions there. The implication would therefore be that those are all doing the wrong thing, and therefore should be simplified. Hmm.

If the udmabuf patches also fix it, then perhaps they are right. It's not an easy one to answer though.

jc-kynesim:
Their argument is that writeback should be enough and invalidate isn't a perfect solution to the issue due to readahead (a valid point that I hadn't really thought about) so there is no point changing anything. Also the existing code will make things go wrong faster so bugs will be found quicker.
My viewpoint is that my code is negligibly more complicated, at least as fast and suppresses some bugs. Whilst udambuf may be getting a redo, dma_heap/system is also bugged in a manner that this fixes to the point I cannot detect the issue anymore so the patch does useful stuff.

PR #6996, opened 2025-08-12
overlays: Fix sc16is752-spi1 emulation

pelwell:
[1] Removed the sc16is752-spi1 overlay, replacing it with an entry in overlay_map.dts that invokes sc16is75x-spi with specific parameters. This does not work because it fails to configure the SPI1 interface. Most such overlays would require the respective SPI interface to have already been configured using one of the spi<n>-<m> overlays, but it is not possible to do that using overlay_map, and it is unreasonable to suddenly impose that requirement on users.

Work around that specific problem by adding an extra parameter to sc16is75x to configure SPI1. It's not ideal, but better than a complete dedicated overlay.

Link: https://github.com/raspberrypi/linux/issues/6962
Fixes: ce20a8fdbf58 ("overlays: sc16is75x: Add generic SPI overlay")

[1] commit ce20a8fdbf58 ("overlays: sc16is75x: Add generic SPI overlay")

### DIFF
--- a/arch/arm/boot/dts/overlays/README
+++ b/arch/arm/boot/dts/overlays/README
@@ -4543,7 +4543,7 @@ Load:   <Deprecated>
 
 Name:   sc16is752-spi1
 Info:   This overlay is now deprecated. Use
-        "dtoverlay=sc16is75x-spi,sc16is752,spi1-0,..." instead.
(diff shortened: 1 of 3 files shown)


nbuchwitz:
A better approach, agreed.

Only little nit: name the new arg something like spi1-1cs-compat to even more state that this is to deal with legacy stuff

PR #6997, opened 2025-08-12
BCM2712 IOMMU: Allocate tables on demand; make IOVA base address and size configurable

njhollinghurst:
~The first change is a draft of something we might actually want to do for Raspberry Pi 5.
The second change is just for testing, to see if increasing IOVA can help resolve an issue with two 64Mpix cameras.~

This has now been rebased and includes allocate-on-demand.

### DIFF
--- a/drivers/iommu/bcm2712-iommu.c
+++ b/drivers/iommu/bcm2712-iommu.c
@@ -2,14 +2,16 @@
 /*
  * IOMMU driver for BCM2712
  *
- * Copyright (c) 2023 Raspberry Pi Ltd.
+ * Copyright (c) 2023-2025 Raspberry Pi Ltd.
  */
 
 #include "bcm2712-iommu.h"
+#include "iommu-pages.h"
 
 #include <linux/dma-mapping.h>
 #include <linux/err.h>
 #include <linux/iommu.h>
+#include <linux/minmax.h>
 #include <linux/of_platform.h>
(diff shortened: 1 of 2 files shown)


njhollinghurst:
As it seems not to have been the main cause of the reported multi-camera crash, I propose to sit on this PR until we can make IOMMU tables allocate-on-demand, rather than all allocated at boot time.

njhollinghurst:
Briefly tested with `labwc` (4K HDMI plus DPI), `chromium-browser`, `rpicam-still`, `ffplay`; to exercise all three IOMMUs for RP1 (CFE, DPI), Codecs (ISP-BE, HEVC) and HVS, while also testing bypass for RP1.

In this scenario, IOMMU allocated (16K) pages reached 93, so the change seems to be saving about 4.5 MBytes.

More soak testing would be advisable.

njhollinghurst:
After speaking to @jc-kynesim  I've reverted the freeing of unused tables (seen in some of the intermediate pushes but no longer present in the commit); tables pages are retained for re-use.

This reduces the risk that `dma_map*()` calls will fail on -ENOMEM, and reduces "thrashing" of the kernel page allocator.

In practice the IOVA allocator seems to prefer to re-use the highest available addresses rather than walking through the entire space, so it seems benign.

njhollinghurst:
This has become a mess. I'm going to start again

PR #6999, opened 2025-08-13
drivers: media: pisp_be: Fix for job queue removal in stop_streaming()

naushir:
The existing code unconditionally removes jobs from the job_queue list when all the nodes in a node group have stopped streaming. This will also remove jobs for any other node groups as the job_queue is a common list. Fix this by only conditionally deleting jobs that belong to the current node group.

Additionally, delete jobs as soon as the first node in the node group stops streaming. Running a job with an incomplete set of active nodes is invalid.

Fixes: 880153e0d0a9 ("media: pisp_be: Split jobs creation and scheduling")

### DIFF
--- a/drivers/media/platform/raspberrypi/pisp_be/pisp_be.c
+++ b/drivers/media/platform/raspberrypi/pisp_be/pisp_be.c
@@ -941,7 +941,6 @@ static void pispbe_node_stop_streaming(struct vb2_queue *q)
 	struct pispbe_dev *pispbe = node_group->pispbe;
 	struct pispbe_job_descriptor *job, *temp;
 	struct pispbe_buffer *buf;
-	LIST_HEAD(tmp_list);
 
 	/*
 	 * Now this is a bit awkward. In a simple M2M device we could just wait
@@ -968,19 +967,17 @@ static void pispbe_node_stop_streaming(struct vb2_queue *q)
 	spin_lock_irq(&pispbe->hw_lock);
 	node_group->streaming_map &= ~BIT(node->id);
 
-	if (node_group->streaming_map == 0) {
-		/*
-		 * If all nodes have stopped streaming release all jobs
-		 * without holding the lock.
-		 */
-		list_splice_init(&pispbe->job_queue, &tmp_list);
+	/*
+	 * If a node has stopped streaming release all jobs belonging to the
+	 * node group immediately.
+	 */
+	list_for_each_entry_safe(job, temp, &pispbe->job_queue, queue) {
+		if (job->node_group == node->node_group) {
+			list_del(&job->queue);
+			kfree(job);
+		}
 	}
 	spin_unlock_irq(&pispbe->hw_lock);
-
-	list_for_each_entry_safe(job, temp, &tmp_list, queue) {
-		list_del(&job->queue);
-		kfree(job);
-	}
 
 	pm_runtime_mark_last_busy(pispbe->dev);
 	pm_runtime_put_autosuspend(pispbe->dev);
--- a/drivers/media/platform/raspberrypi/pisp_be/pisp_be.c
+++ b/drivers/media/platform/raspberrypi/pisp_be/pisp_be.c
@@ -594,9 +594,7 @@ err_return_buffers:
 	return -ENODEV;
 }
 
-static void pispbe_schedule(struct pispbe_dev *pispbe,
-			    struct pispbe_node_group *node_group,
-			    bool clear_hw_busy)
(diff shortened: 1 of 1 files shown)


drivers/media/platform/raspberrypi/pisp_be/pisp_be.c
```diff
@@ -968,20 +968,18 @@ static void pispbe_node_stop_streaming(struct vb2_queue *q)
...
 	node_group->streaming_map &= ~BIT(node->id);
 
-	if (node_group->streaming_map == 0) {
-		/*
-		 * If all nodes have stopped streaming release all jobs
-		 * without holding the lock.
-		 */
-		list_splice_init(&pispbe->job_queue, &tmp_list);
+	/*
+	 * If a node have stopped streaming release all jobs belonging to the
```
JamesH65: have->has?

naushir: Ack.

njhollinghurst:
Is it possible to be in a situation where jobs are queued for this node-group, which don't reference the node that was stopped? (It seems unlikely.)

naushir:
> Is it possible to be in a situation where jobs are queued for this node-group, which don't reference the node that was stopped? (It seems unlikely.)

I think that would be a malformed job and probably fail validation?

naushir:
Sadly, this does not seem to be enough to fix https://github.com/raspberrypi/rpicam-apps/issues/849.

naushir:
With https://github.com/raspberrypi/linux/pull/6999/commits/1506a5efe815e8cfbb578af657952800100b3281, I'm now pretty confident that https://github.com/raspberrypi/rpicam-apps/issues/849 ought to be resolved.

naushir:
@njhollinghurst this is ready to be reviewed and merged if you are happy with it.

njhollinghurst:
LGTM. I assume the checkpatch error is just an artifact of a shallow clone?

6by9:
> LGTM. I assume the checkpatch error is just an artifact of a shallow clone?

Yes, it's one of the current limitations of the checkpatch job. 

Looks like we can squash that one by adding `--ignore UNKNOWN_COMMIT_ID` to the checkpatch invocation - I'll sort a PR.

jmondi:
Hi Naush, thanks for fixing this. Indeed I introduced this when moving the patches from the non-multi-group mainline version to the multi-group BSP version of the driver. Sorry about that, I recall I tested with 2 cameras, but I don't recall seeing this (however, it should happen all the times one camera is stopped while the other is left running ?)

https://github.com/raspberrypi/linux/pull/6999/commits/1506a5efe815e8cfbb578af657952800100b3281 is certainly good, I don't think there is anything to upport to mainline right ?

I'm a bit less sure about https://github.com/raspberrypi/linux/pull/6999/commits/3e4f99e060ba49deb3d2b0202d288e0ee9366556. `continue` is certainly wrong, but I wonder if we should `return` ? `node_group` is still a thing for the BSP driver, but I guess the check in `pispbe_schedule()` is redundant as if we have any job at the front of the queue, we shall run it regardless of the `group` that has just created a new job ?

Thanks for fixing this, and sorry if you caught me on the week I was on leave.

naushir:
> Hi Naush, thanks for fixing this. Indeed I introduced this when moving the patches from the non-multi-group mainline version to the multi-group BSP version of the driver. Sorry about that, I recall I tested with 2 cameras, but I don't recall seeing this (however, it should happen all the times one camera is stopped while the other is left running ?)

Not to worry!  This bug did take some time to trigger - and in very specific timing conditions.  So most of out testing will not have caught it.

> https://github.com/raspberrypi/linux/commit/1506a5efe815e8cfbb578af657952800100b3281 is certainly good, I don't think there is anything to upport to mainline right ?

Correct, I checked and there is no need to mainline this change.

> I'm a bit less sure about https://github.com/raspberrypi/linux/commit/3e4f99e060ba49deb3d2b0202d288e0ee9366556. continue is certainly wrong, but I wonder if we should return ? node_group is still a thing for the BSP driver, but I guess the check in pispbe_schedule() is redundant as if we have any job at the front of the queue, we shall run it regardless of the group that has just created a new job ?

My feeling is that it's best if the test is removed entirely.  Since we now rely on a single job queue across all users, we should unconditionally run the first job in the queue to maintain strict FIFO ordering.  There is no reason to jump to the "current" user's job in such cases.  Again, there is no need to upstream this as the logic is different, i.e. only handles a single user.

> Thanks for fixing this, and sorry if you caught me on the week I was on leave.

No problem at all - hope you had a good rest 😄

PR #7005, opened 2025-08-18
Enable Full Screen Splash Images

bebon901:
This will allow users to enter a cmdline value which will enable the kernel to read a PPM file, and then rendering it fullscreen at the early boot stage.
I am aware it needs a _lot_ of tidying up, but put here for others to reference!

### DIFF
--- a/drivers/video/fbdev/core/fb_logo.c
+++ b/drivers/video/fbdev/core/fb_logo.c
@@ -2,11 +2,24 @@
 
 #include <linux/fb.h>
 #include <linux/linux_logo.h>
+#include <linux/fs.h>
+#include <linux/slab.h>
+#include <linux/string.h>
+#include <linux/uaccess.h>
+#include <linux/file.h>
+#include <linux/kernel.h>
+#include <linux/firmware.h>
 
 #include "fb_internal.h"
 
 bool fb_center_logo __read_mostly;
 int fb_logo_count __read_mostly = -1;
+static int fullscreen_logo_enabled;
+static char *fullscreen_logo_path;
+
+struct image_palette {
+	u8 colors[224][3];
+};
 
 static inline unsigned int safe_shift(unsigned int d, int n)
 {
@@ -76,6 +89,22 @@ static void  fb_set_logo_truepalette(struct fb_info *info,
 				 safe_shift((clut[1] & greenmask), greenshift) |
 				 safe_shift((clut[2] & bluemask), blueshift));
 		clut += 3;
+	}
+}
+
+static void fb_set_logo_RGB_palette(struct image_palette *palette,
+						 u32 *palette_to_write, int current_rows)
+{
+	// Set the kernel palette from an array of RGB values
+	uint32_t color_code;
+	int i;
+
+	// Color format is RGB565, remove LSB 3 bits, and move to correct position
+	for (i = 0; i < current_rows; i++) {
+		color_code = ((((uint16_t)palette->colors[i][0]) >> 3) << 11) |
+			     ((((uint16_t)palette->colors[i][1]) >> 2) << 5) |
+			     (((uint16_t)palette->colors[i][2]) >> 3);
+		palette_to_write[i+32] = color_code;
 	}
 }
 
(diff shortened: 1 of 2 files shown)


drivers/video/fbdev/core/fb_logo.c
```diff
@@ -275,84 +319,544 @@ static void fb_do_show_logo(struct fb_info *info, struct fb_image *image,
...
+	fullscreen_logo_enabled = 1;
+	fullscreen_logo_path = str;
+	printk(KERN_ALERT "Fullscreen Logo Enabled: %d", fullscreen_logo_enabled);
+	printk(KERN_ALERT "Fullscreen Logo Path: %s", fullscreen_logo_path);
+	return 1;
+}
+
+__setup("fullscreen_logo=", fullscreen_logo_setup);
+
+ssize_t read_file_from_initramfs(const char *filepath, char **buffer, size_t max_size)
```
pelwell: Why make the name initramfs-specific? Doesn't this work just as well with files in a regular FS?

drivers/video/fbdev/core/fb_logo.c
```diff
@@ -275,84 +319,544 @@ static void fb_do_show_logo(struct fb_info *info, struct fb_image *image,
...
+
+    // Null-terminate the buffer
+    (*buffer)[bytes_read] = '\0';
+    printk(KERN_INFO "Successfully read %zd bytes from %s:\n%s\n",
+           bytes_read, filepath, *buffer);
+
+    ret = bytes_read; // Return actual bytes read
+
+out_free:
+    if (ret < 0 && *buffer) {
```
pelwell: We already know that `*buffer` is non-NULL

drivers/video/fbdev/core/fb_logo.c
```diff
@@ -275,84 +319,544 @@ static void fb_do_show_logo(struct fb_info *info, struct fb_image *image,
...
+    size_t file_len = i_size_read(file_inode(file));
+    if (file_len > max_size) {
+        printk(KERN_WARNING "File %s is larger than max_size (%zu vs %zu)\n",
+               filepath, file_len, max_size);
+        file_len = max_size; // Limit read to max_size
+    }
+
+    // 2. Allocate buffer
+	printk(KERN_ALERT "File Length: %d", file_len);
+    *buffer = kvmalloc(file_len + 1, GFP_KERNEL); // +1 for null terminator
```
pelwell: It would be nicer to use a temporary pointer for the allocated buffer, and only assign to `*buffer` in the success path.

drivers/video/fbdev/core/fb_logo.c
```diff
@@ -275,84 +319,544 @@ static void fb_do_show_logo(struct fb_info *info, struct fb_image *image,
...
+	}	
+	if (fullscreen_logo_enabled) {
+		// Lets try and read a file eh
+		char *file_content = NULL;
+		ssize_t len;
+
+		printk(KERN_INFO "Attempting to read file from initramfs... --------------------------------------------------------------------------------------------------\n");
+
+		long width = 0, height = 0, max_color_value = 0;
+		long image_palette[224][3];
```
pelwell: It would be nice to declare a typedef for this type and use it throughout.

timg236:
I think it would be preferable (more efficient and smaller code) to use a binary format rather than parsing PPM. TGA is very simple and supports palettes.

6by9:
There may already be part of this implemented for uefi in drivers/video/fbdev/efifb.c for allowing a vendor specific bitmap to be passed in such that it can be merged into the boot screens. That's how you can get say "Asus" as the background whilst Ubuntu is starting.
I can't remember the full runes, but the BMP file gets exposed as /sys/firmware/acpi/bgrt

Loading the file using the `request_firmware` calls to load from /lib/firmware (which can be in the initramfs) is far cleaner than messing with `filp_open` calls for yourself.

pelwell (commented):
I'm not going to give this a full review until you've tidied it up some - it's too tiring otherwise.

A few kernel coding rules (see https://www.kernel.org/doc/html/v6.12/process/coding-style.html for more):
1. Use TABs for indentation, with tab stops set to 8. Spaces can be used to precise alignment, but only after any TABs.
2. You have the preferred brace style and no trailing whitespace, so that's good.
3. Your local variable declarations are also pretty good, but for full marks put them in "reverse Christmas tree" format, longest to shortest, making sure there is a blank line before any statements.
4. Make all local-scope functions `static`, and ideally with a module-specific prefix.
5. Prefer `pr_info` (or `dev_info`) over `printk(KERN_INFO `.

`scripts/checkpatch.pl -g <from-commit>..<to-commit>` is your fussy friend.

bebon901:
> There may already be part of this implemented for uefi in drivers/video/fbdev/efifb.c for allowing a vendor specific bitmap to be passed in such that it can be merged into the boot screens. That's how you can get say "Asus" as the background whilst Ubuntu is starting. I can't remember the full runes, but the BMP file gets exposed as /sys/firmware/acpi/bgrt
> 
> Loading the file using the `request_firmware` calls to load from /lib/firmware (which can be in the initramfs) is far cleaner than messing with `filp_open` calls for yourself.

I've had a look around the efifb.c file, and it looks like it's ultimately expecting the data from the ACPI Boot Graphics Resource Table - which following a conversation with Tom, looks like a path we don't want to go down.
Will have a look into the `request_firmware` option though!

6by9:
> I've had a look around the efifb.c file, and it looks like it's ultimately expecting the data from the ACPI Boot Graphics Resource Table - which following a conversation with Tom, looks like a path we don't want to go down. Will have a look into the `request_firmware` option though!

I was more thinking if we have the BMP parser etc already in the kernel, it'd be nicer for that to be abstracted out and reused.
I will say that I haven't looked at the detail for whether it actually achieves the same goal as you have here.

bebon901:
Have now tidied up, and is ready for review @pelwell 

Also, it complains about creating a new typedef - what would you like to do about this? - I know a previous comment suggested creating `typedef long ImagePalette[224][3]`

pelwell:
This is looking much better. You could replace the typedef with a structure whose only member is the two-dimensional array. And on that subject, is there any reason not to make it an array of `u8`/`uint8_t`s? Using longs is wasteful and confusing, in the sense that it makes the reader wonder where you use values large enough to justify their use.

bebon901:
Thanks! I've updated it to include your suggestions.

hex0id:
when are you guys merging this in?

drivers/video/fbdev/core/fb_logo.c
```diff
@@ -275,6 +304,164 @@ static void fb_do_show_logo(struct fb_info *info, struct fb_image *image,
 	}
 }
 
+static int __init fb_fullscreen_logo_setup(char *str)
+{
+	fullscreen_logo_enabled = 1;
+	fullscreen_logo_path = str;
+	pr_info("Fullscreen Logo Enabled: %d", fullscreen_logo_enabled);
```
pelwell: This will only ever show "Fullscreen Logo Enabled: 1", so I think it could be combined with the "Path" line.

drivers/video/fbdev/core/fb_logo.c
```diff
@@ -275,6 +304,164 @@ static void fb_do_show_logo(struct fb_info *info, struct fb_image *image,
...
+	}
+	len = fw->size;
+	file_content = fw->data;
+
+	if (len > 0) {
+		int current_rows = 0;
+		const char *current_ptr = file_content;
+		const char *end_ptr = file_content + len;
+
+		if (len < 18) {
```
pelwell, line 363: I don't understand why checkpatch hasn't complained about this code before the second group of declarations. You could move it down to before the image_type check, or merge it with the other `len` error case after this code block.

drivers/video/fbdev/core/fb_logo.c
```diff
@@ -275,6 +304,164 @@ static void fb_do_show_logo(struct fb_info *info, struct fb_image *image,
...
+			pr_info("Invalid logo file: TGA file too small for header\n");
```
pelwell: Error messages should be output at `ERROR` (`pr_err`) or `WARN` (`pr_warn`) level to make it easier to find them (e.g. `dmesg -l err,warn`). This applies in several places in this file.

drivers/video/fbdev/core/fb_logo.c
```diff
@@ -275,6 +304,164 @@ static void fb_do_show_logo(struct fb_info *info, struct fb_image *image,
...
+		if (image_type != 2 || pixel_depth != 24) {
+			pr_info("Unsupported TGA logo format: Type=%d, Depth=%d (only support Type=2, Depth=24)\n",
+				image_type, pixel_depth);
+			goto cleanup;
+		}
+		// Skip header + ID field
+		current_ptr = file_content + 18 + id_length;
+
+		read_logo = kmalloc_array(width, height, GFP_KERNEL);
+		if (read_logo == NULL)
```
pelwell: `if (!read_logo)` is the preferred form in the kernel.

drivers/video/fbdev/core/fb_logo.c
```diff
@@ -275,6 +304,164 @@ static void fb_do_show_logo(struct fb_info *info, struct fb_image *image,
...
+			goto cleanup;
+
+		image->data = read_logo;
+
+		// TGA pixels are stored bottom-to-top by default, unless bit 5 of
+		// image_descriptor is set
+		bool top_to_bottom = (image_descriptor & 0x20) != 0;
```
pelwell: More mid-block declarations...

drivers/video/fbdev/core/fb_logo.c
```diff
@@ -275,6 +304,164 @@ static void fb_do_show_logo(struct fb_info *info, struct fb_image *image,
...
+		// Only supports uncompressed true-color images (type 2) with 24-bit depth
+		if (image_type != 2 || pixel_depth != 24) {
+			pr_info("Unsupported TGA logo format: Type=%d, Depth=%d (only support Type=2, Depth=24)\n",
+				image_type, pixel_depth);
+			goto cleanup;
+		}
+		// Skip header + ID field
+		current_ptr = file_content + 18 + id_length;
+
+		read_logo = kmalloc_array(width, height, GFP_KERNEL);
```
pelwell, line 383: This will need to be freed in the error paths. It's common to end a function with a list of labelled cleanup steps in reverse order, with error handling code jumping to the appropriate one.

pelwell (commented):
Just a few last-minute suggestions. I nearly hit the Merge button, but this is so close to being good...

PR #7002, opened 2025-08-14
6.17 fixups

6by9:
@popcornmix 

Still some more to check out.

PR #7006, opened 2025-08-19
My rpi-6.17 fixups

pelwell:
This PR builds on #7002, adding in sufficient extra patches to make Pi 5 usable along with a few other forward ports where the merges were non-trivial.

My approach to the recent upstream RP1 commits has been to revert or partially undo them and restore the downstream equivalent, the aim being to get something working ASAP. We can then work on more refined downstream changes, with a view to using them instead at a later date.


popcornmix:
This worked for me:
```
pi@pi500:~ $ uname -a
Linux pi500 6.17.0-rc1-v8-16k #5 SMP PREEMPT Tue Aug 19 15:04:10 BST 2025 aarch64 GNU/Linux
```
I'll merge this now, and the "upstream" rp1 support can come on top.

PR #7007, opened 2025-08-19
iommu: bcm2712-iommu: Allocate tables on demand; add OF properties

njhollinghurst:
Allocate space for level-2 IOMMU translation tables on demand. This should save memory in most cases but means that map_pages() can now fail with -ENOMEM. Unused pages are retained for re-use.

Add OF properties to override the default aperture size (2GB) and base address (40GB); this doesn't include any dma-iova-offset.

Various tidy-ups. The internal representation of aperture limits *does* include dma_iova_offset, as that is more commonly useful. Clarify the distinction between Linux pages and IOMMU table pages. Fix wrong definition of MMMU_CTRL_PT_INVALID_EN flag.

### DIFF
--- a/drivers/iommu/bcm2712-iommu.c
+++ b/drivers/iommu/bcm2712-iommu.c
@@ -2,14 +2,16 @@
 /*
  * IOMMU driver for BCM2712
  *
- * Copyright (c) 2023 Raspberry Pi Ltd.
+ * Copyright (c) 2023-2025 Raspberry Pi Ltd.
  */
 
 #include "bcm2712-iommu.h"
+#include "iommu-pages.h"
 
 #include <linux/dma-mapping.h>
 #include <linux/err.h>
 #include <linux/iommu.h>
+#include <linux/minmax.h>
 #include <linux/of_platform.h>
 #include <linux/platform_device.h>
 #include <linux/spinlock.h>
@@ -26,8 +28,9 @@
 #define MMMU_CTRL_CAP_EXCEEDED_EXCEPTION_EN    BIT(24)
 #define MMMU_CTRL_PT_INVALID                   BIT(20)
 #define MMMU_CTRL_PT_INVALID_ABORT_EN          BIT(19)
-#define MMMU_CTRL_PT_INVALID_EXCEPTION_EN      BIT(18)
-#define MMMU_CTRL_PT_INVALID_EN                BIT(17)
+#define MMMU_CTRL_PT_INVALID_INT_EN            BIT(18)
+#define MMMU_CTRL_PT_INVALID_EXCEPTION_EN      BIT(17)
+#define MMMU_CTRL_PT_INVALID_EN                BIT(16)
 #define MMMU_CTRL_WRITE_VIOLATION              BIT(12)
 #define MMMU_CTRL_WRITE_VIOLATION_ABORT_EN     BIT(11)
 #define MMMU_CTRL_WRITE_VIOLATION_INT_EN       BIT(10)
@@ -66,6 +69,8 @@
 #define MMMU_ILLEGAL_ADR_OFFSET                0x30
 #define MMMU_ILLEGAL_ADR_ENABLE                BIT(31)
 
+#define MMMU_VIO_ADDR_OFFSET                   0x34
+
 #define MMMU_DEBUG_INFO_OFFSET                 0x38
 #define MMMU_DEBUG_INFO_VERSION_MASK           0x0000000Fu
 #define MMMU_DEBUG_INFO_VA_WIDTH_MASK          0x000000F0u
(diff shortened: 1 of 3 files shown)


njhollinghurst:
Would anyone like to review this? I know it's quite a sprawling change...
Or critique the "lazy allocation / never free" strategy? @pelwell @6by9 @popcornmix @jc-kynesim 

The most visible difference is that it should save about 4MBytes of memory in most cases. With the second commit it ought to cope better with 2 x 64MPixel cameras, though I haven't tested that.

jc-kynesim:
You end up allocating many IOMMU pages individually (presumably 1024 of them by the end), would there be any benefit in trying to allocate in bigger chunks - or is that just pointlessly tricky?

njhollinghurst:
That's what the old version did, using the scatter-gather-list style of DMA mapping (so as not to require contiguous memory).

I suppose one could allocate in chunks that were bigger than one Linux page but still smaller than the total. I did wonder about this. But introducing yet another "page/chunk size" would make the code even messier.

One 16K Linux page is enough for 16M of IOVA, which didn't seem ridiculously small.

jc-kynesim:
You seem to have got rid of all the dma_sync_sgtable_for_cpu calls that were previously in the code. Are those now done in some other function or have they just gone? Having just been reminded by upstream of their importance I feel that they should be included even though they probably end up doing nothing in the current world.

njhollinghurst:
Yes, that (along with the assumption that dma_addr == phys_addr) was copied from another IOMMU driver, but it might not have been one for ARM64 so perhaps not a good example.

Might redundant/repeated sync calls be even worse than not having them (e.g. if they invalidate wanted cache contents)? If so, it might require keeping track of the current sync status of each page. Bother.

I see that [this driver](https://github.com/raspberrypi/linux/blob/rpi-6.12.y/drivers/iommu/exynos-iommu.c#L883) does it for every single write! Maybe I should do that...

jc-kynesim:
My understanding goes that it should always look like this when writing (if cached - if not then no-one cares):

dma_sync_sgtable_for_cpu(ptr, DMA_TO_DEVICE)
\<fill table at ptr with stuff\>
dma_sync_sgtable_for_device(ptr, DMA_TO_DEVICE)
\<do whatever is needed to the the h/w to read the stuff just written\>

You can add as much as you like between the syncs but it shouldn't be stuff that will be read by the h/w until after the sync_for_device. As it happens dma_sync_sgtable_for_cpu(ptr, DMA_TO_DEVICE) will no-op on most architectures including arm64 so it is all a bit hypothetical but one could conceive of an optimization which no-ops the dma_sync_sgtable_for_device if there was no preceding dma_sync_sgtable_for_cpu so it is probably worth having.

AFAIK calls should always be paired. It is theoretically either wrong or pointless to do otherwise though in practice you will get away with it if you pick the correct call+direction.

njhollinghurst:
I've moved `dma_sync_*_for_*()` into the map and unmap calls, which makes it easier to ensure they are balanced.

Also removed a couple of redundant ones from probe, because DMA documentation assures me that a newly-created mapping will be synchronized for the device.

jc-kynesim:
That looks good to me

njhollinghurst:
Unfortunately something is intermittently going wrong now I've switched back from VEC to HDMI. Possibly because the latter uses multi-hugepage-sized coherent blocks, which weren't adequately tested earlier.

Heisenbug. Not necessarily caused by this change (though I bet it is somehow related).

drivers/iommu/bcm2712-iommu.c
```diff
@@ -241,25 +258,61 @@ static int bcm2712_iommu_map(struct iommu_domain *domain, unsigned long iova,
...
+			     i < (lxp + 1) << (PAGE_SHIFT - IOMMU_PAGE_SHIFT); i++) {
+				mmu->top_table[i] = u++;
+			}
+			dirty_top = true;
+		}
+	}
+	if (dirty_top)
+		dma_sync_single_for_device(mmu->dev, virt_to_phys(&mmu->top_table),
+					   PAGE_SIZE, DMA_TO_DEVICE);
+	if (lxp <= (p_last >> TABLES_LXPAGE_SHIFT)) {
```
njhollinghurst: Oops, the shift probably should be `LX_PAGEWORDS_SHIFT`, as at line 287.

njhollinghurst: Fortunately this never happens (and the wrong shift makes it happen even less).

drivers/iommu/bcm2712-iommu.c
```diff
@@ -241,25 +258,61 @@ static int bcm2712_iommu_map(struct iommu_domain *domain, unsigned long iova,
...
+	     lxp <= (p_last >> LX_PAGEWORDS_SHIFT); lxp++) {
+		if (!mmu->tables[lxp]) {
+			unsigned int i, u;
+
+			u = bcm2712_iommu_getpage(mmu, &mmu->tables[lxp]);
+			if (!u)
+				break;
+			u |= MMMU_PTE_VALID;
+			if (!dirty_top)
+				dma_sync_single_for_cpu(mmu->dev, virt_to_phys(&mmu->top_table),
```
njhollinghurst: Aargh! That `&` is wrong!

drivers/iommu/bcm2712-iommu.c
```diff
@@ -241,25 +258,61 @@ static int bcm2712_iommu_map(struct iommu_domain *domain, unsigned long iova,
...
+							PAGE_SIZE, DMA_TO_DEVICE);
+			for (i = lxp << (PAGE_SHIFT - IOMMU_PAGE_SHIFT);
+			     i < (lxp + 1) << (PAGE_SHIFT - IOMMU_PAGE_SHIFT); i++) {
+				mmu->top_table[i] = u++;
+			}
+			dirty_top = true;
+		}
+	}
+	if (dirty_top)
+		dma_sync_single_for_device(mmu->dev, virt_to_phys(&mmu->top_table),
```
njhollinghurst: You would not believe how long that has taken to debug.

pelwell:
I see this PR is no longer a draft  - is it ready to merge?

njhollinghurst:
it's certainly ready to review, and has had a decent amount of testing now. But given the potential for errors I feel it should skip any imminent release, to allow even more time to test. Its main benefit is to save about 4MB of memory.

pelwell:
SInce we've made the stable kernel releases, let's get this merged.

PR #7010, opened 2025-08-19
RP1 Video Out: Hack for 50fps/60fps buffer-flips in interlaced modes

njhollinghurst:
By avoiding VEC's built-in interlaced DMA modes, and sending individual fields (adapting the "field wobble" trick from the DPI driver), it looks like we can get VEC to update at full field rate.

This changes the pattern of VSync pulses on the VEC's FE->BE interface. Some jiggery-pokery with the NTSC vertical settings was required before the BE would tolerate this (my guess is that the offset between "frame datum" and VSync, which is nonzero in conventional NTSC line numbering, made it more sensitive to FE VSync timings?)

Needs more testing to ensure it's reliable in all modes (including the "vintage" ones). Unlike the DPI case we can't use PIO to infer correct field polarity. If it  ever gets out of step, it won't recover!

To clarify: The first of these two commits relates to _progressive_ modes, and is unrelated.

### DIFF
--- a/drivers/gpu/drm/rp1/rp1-vec/rp1_vec_hw.c
+++ b/drivers/gpu/drm/rp1/rp1-vec/rp1_vec_hw.c
@@ -102,13 +102,18 @@ static const struct rp1vec_ipixfmt my_formats[] = {
  * See "vec_regs.h" for further descriptions of these registers and fields.
  * Driver should adjust some values for other TV standards and for pixel rate,
  * and must ensure that ((de_end - de_bgn) % rate) == 0.
+ *
+ * To support 60fps update in interlaced modes, we now do ISR-based field-flip.
+ * The FIELDS_PER_FRAME_MINUS1 flag in "misc" is no longer set. Some vertical
+ * timings have been rotated wrt conventional line-numbering (to ensure a gap
+ * between the last active line and nominal end-of-field).
  */
 
 struct rp1vec_hwmode {
 	u16  max_rows_per_field;   /* active lines per field (including partial ones)  */
 	u16  ref_vfp;		   /* nominal (vsync_start - vdisplay) when max height */
 	bool interlaced;	   /* set for interlaced			       */
-	bool first_field_odd;	   /* depends confusingly on line numbering convention */
+	bool first_field_odd;	   /* true if odd-indexed scanlines go to first field  */
 	s16  scale_v;		   /* V scale in 2.8 format (for power-of-2 CIC rates) */
 	s16  scale_u;		   /* U scale in 2.8 format (for power-of-2 CIC rates) */
 	u16  scale_y;		   /* Y scale in 2.8 format (for power-of-2 CIC rates) */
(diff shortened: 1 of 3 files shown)


njhollinghurst:
Simulations suggest that the issue with NTSC timings may be the small distance between the last active (or FE-inserted "vfp") line and the field datum (rather than the vertical sync sequence per se).

It seems the BE stops accepting pixels just before it gets to see the first pixel of the second field, the one with VSync. In the native interlace mode it won't be expecting a VSync here, but in the hacked mode it seems to lock up if one is not present!

Fortunately (and still rather mysteriously) the workaround works.

drivers/gpu/drm/rp1/rp1-vec/rp1_vec_hw.c
```diff
@@ -525,32 +574,48 @@ void rp1vec_hw_setup(struct rp1_vec *vec,
 
 void rp1vec_hw_update(struct rp1_vec *vec, dma_addr_t addr, u32 offset, u32 stride)
 {
+	unsigned long flags;
+
+	spin_lock_irqsave(&vec->hw_lock, flags);
```
pelwell: Could the spin_lock be claimed later (more for readability than efficiency)?

drivers/gpu/drm/rp1/rp1-vec/rp1_vec_hw.c
```diff
@@ -525,32 +574,48 @@ void rp1vec_hw_setup(struct rp1_vec *vec,
...
+	 * the next vertical sync event. In interlaced mode, we need to adjust
+	 * the address and stride to display only the current field, saving
+	 * the original address (so it can be flipped for subsequent fields).
 	 */
-	u64 a = addr + offset;
-
-	if (vec->fake_31khz) {
-		a += stride;
+	addr += offset;
+	vec->last_dma_addr = addr;
```
pelwell, line 554: Here, perhaps?

njhollinghurst: It could skip one line `addr += offset;` but it's then needed to protect the atomicity of `last_dma_addr` and `last_stride` as well as the register writes.

njhollinghurst:
Do we have a policy about "chicken parameters"? e.g.  a module parameter called `fast-update` or some such, default true.

Even once 60fps is (we hope) proven reliable, it's just conceivable some users might want to keep the old 30fps rate (though VC4 was always 60fps).

pelwell:
> Do we have a policy about "chicken parameters"?

Use them when necessary? Module parameters have the advantage that they can be set in their own module.conf file, which makes maintenance easier for distros.

njhollinghurst:
Done some more testing.
- Inducing VEC under-runs using the AXI burst breaker, seems benign. The picture breaks up but VEC FE maintains sync, as it should, and interrupts continue.
- Dropping an RP1 VEC interrupt causes hideous vertical jitter (as predicted) but nothing fatal; the display remains usable for many purposes. It persists until the next mode change (or another dropped interrupt).
 - Soak testing with `kmstest --flip` in different modes with random (odd and even) numbers of flips, seems fine.

I might not bother with a parameter.

drivers/gpu/drm/rp1/rp1-vec/rp1_vec_hw.c
njhollinghurst: Moved

PR #7011, opened 2025-08-21
dmaengine: dw-axi-dmac: report per-channel max_burst via device_caps

nbuchwitz:
The patch "dmaengine: dw-axi-dmac: add per-channel AXI burst length support" programs ARLEN/AWLEN from the snps,axi-max-burst-len array but still exposed a single max_burst value via dma_get_slave_caps(). As a result all channels reported 8 even when limited to 4, leading to warnings:

  dma dma2chan5: requested source burst length 8 exceeds supported 4

Add a .device_caps callback to return the correct per-channel max_burst. This allows drivers like amba-pl011 to clamp burst lengths properly.

Fixes: 0e4e6a0c4f4e ("dmaengine: dw-axi-dmac: add per-channel AXI burst length support") https://github.com/raspberrypi/linux/pull/6994

### DIFF
--- a/drivers/dma/dw-axi-dmac/dw-axi-dmac-platform.c
+++ b/drivers/dma/dw-axi-dmac/dw-axi-dmac-platform.c
@@ -1476,6 +1476,16 @@ static int __maybe_unused axi_dma_runtime_resume(struct device *dev)
 	return axi_dma_resume(chip);
 }
 
+static void dw_axi_dma_device_caps(struct dma_chan *dchan,
+				   struct dma_slave_caps *caps)
+{
+	struct axi_dma_chan *chan = dchan_to_axi_dma_chan(dchan);
+	struct dw_axi_dma *dw = chan->chip->dw;
+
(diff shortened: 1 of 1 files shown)


nbuchwitz:
To be discussed: Keep `dw->dma.max_burst = hdata->axi_rw_burst_len[0];` which set the default burst length to 8 or better use min on ` hdata->axi_rw_burst_len`?

pelwell:
Apart from the checkpatch indentation complaints, this looks good.

> To be discussed: Keep dw->dma.max_burst = hdata->axi_rw_burst_len[0]; which set the default burst length to 8 or better use min on  hdata->axi_rw_burst_len?

Don't we drop the assignment completely now we have the `device_caps` method?

nbuchwitz:
> Don't we drop the assignment completely now we have the device_caps method?

Probably can do, yes. The array should be populated for all channels, so the fallback isn't really needed. I will amend my commit when I take care of checkpatch.

pelwell:
Thanks!

PR #7015, opened 2025-08-26
ronbo panel updates - CM4

ORR99:
- Modified the panel-ronbo-rb070d30.c driver
- Added vc4-kms-dsi-ronbo-rb070d30.dts
- Added .config file which is the /boot/config-6.12.25+rpt-rpi-v8 from my CM4 and modified this line : CONFIG_DRM_PANEL_RONBO_RB070D30=m
- Add "extra" directory which contains : /boot/firmware/config.txt and /etc/modules-load.d/modules.conf


drivers/gpu/drm/panel/Makefile
```diff
@@ -104,3 +104,4 @@ obj-$(CONFIG_DRM_PANEL_WAVESHARE_TOUCHSCREEN) += panel-waveshare-dsi.o
 obj-$(CONFIG_DRM_PANEL_WAVESHARE_TOUCHSCREEN_V2) += panel-waveshare-dsi-v2.o
 obj-$(CONFIG_DRM_PANEL_WIDECHIPS_WS2401) += panel-widechips-ws2401.o
 obj-$(CONFIG_DRM_PANEL_XINPENG_XPP055C272) += panel-xinpeng-xpp055c272.o
+obj-$(CONFIG_DRM_PANEL_RONBO_RB070D30) += panel-ronbo-rb070d30.o
```
6by9: Already present at line 65.

drivers/gpu/drm/panel/panel-ronbo-rb070d30.c
```diff
@@ -25,210 +25,257 @@
...
-	struct mipi_dsi_device *dsi;
-	struct regulator *supply;
-
-	struct {
-		struct gpio_desc *power;
-		struct gpio_desc *reset;
-		struct gpio_desc *updn;
-		struct gpio_desc *shlr;
-	} gpios;
+        struct drm_panel panel;
```
6by9, line 28: Linux coding style is to use tabs, and tabs to be set at 8 space indentation.

Because your editor wasn't set up for that, the diff appears much larger than it actually is.

extra/modules.conf
```diff
@@ -0,0 +1,8 @@
+# /etc/modules: kernel modules to load at boot time.
+#
+# This file contains the names of kernel modules that should be loaded
+# at boot time, one per line. Lines beginning with "#" are ignored.
+# Parameters can be specified after the module name.
+
+i2c-dev
+panel_ronbo_rb070d30
```
6by9, line 8: Not needed if systemd or similar is running. Userspace will match the udev notification of a node with the compatible "ronbo,rb070d30" being registered, and load the relevant driver automatically.

extra/config.txt
```diff
@@ -0,0 +1,55 @@
...
+otg_mode=1
+
+[cm5]
+dtoverlay=dwc2,dr_mode=host
+
+[all]
+enable_uart=1
+ignore_lcd=1
+dtoverlay=vc4-kms-dsi-ronbo-rb070d30
+dtparam=dsi1_escape_clk_div=4
```
6by9: No such dtparam.

.config
```diff
@@ -0,0 +1,9451 @@
+#
```
6by9, line 1: .config changes should be made to arch/arm64/configs/bcm2711_defconfig. They then get actioned by the `make bcm2711_defconfig` step in the [build docs](https://www.raspberrypi.com/documentation/computers/linux_kernel.html).

6by9 (commented):
Link to https://forums.raspberrypi.com/viewtopic.php?t=390710

General comments made as I'll have to take manual steps to test this.

extra/config.txt
pelwell: And no such file or directory - what is the purpose of this addition?

ORR99: I was just trying something

6by9: In the now linked forum thread I'd asked for them to push their kernel tree so I can see what the DSI analyser makes of their DSI panel configuration. It looks like we've got various extra lumps as well, which don't help.

ORR99: Do you want me to clean the tree up based on your feedback, and push a new version of it so we can test again?

drivers/gpu/drm/panel/panel-ronbo-rb070d30.c
ORR99: but this should not cause a problem, right?

6by9: It makes reviewing the changes significantly harder.

extra/config.txt
6by9: I can take the changes and get them to work, but it's just more work so may not get done immediately.

ORR99: The changes in the driver:
- I commented the gpio_set functions in prepare as gpios in my case are external
- mipi_dsi-dcs_exit_sleep_mode() is now called in prepare function
- Enable function does not get called and is commented in drm_panel_funcs struct
- "ctx->panel.prepare_prev_first = true;" was added to the probe function as advised
- in probe function regulator and gpio functions are now optional: devm_regulator_get_optional,devm_gpiod_get_optional (since they are external)

ORR99: I’ve modified my tree and pushed the updates. If you need me to, I can open a new draft PR — just let me know.

pelwell:
Something has gone horribly wrong with this PR it now contains over 3000 "modified" files.

ORR99:
> Something has gone horribly wrong with this PR it now contains over 3000 "modified" files.

I ran "make bcm2711_defconfig"

6by9:
But you'd also removed .gitignore and then added all the files that had changed.
.gitignore used to include ignoring include/config and various other directories and file types used as part of the build process.

Leave it alone and I should be able to make sense of it from the first commit.

ORR99:
Hello 6by9, how is it going with the kernel tree? Did you manage to run it yet?

6by9:
I've tidied up your commits and pushed them to https://github.com/6by9/linux/tree/rpi-6.12.y-dsi-ronbo

Capture made on both a CM4 and Pi5. Files attached.
[RonboCaptures.zip](https://github.com/user-attachments/files/22011077/RonboCaptures.zip)

The main capture files should be readable if you install the software from https://www.teledynelecroy.com/protocolanalyzer/mipi/envision-x84-analyzer/resources/analysis-software. I don't believe you need a licence for viewing (it's tied to the analyser hardware for capturing/generation).

I did notice that Pi5 was driving the display in sync event mode (as per the panel driver's request), whilst CM4 wasn't supporting that and was using sync pulse mode. There's a quick patch to the vc4 driver to add sync event mode, and that appears to work (second capture taken in that mode).

The images confirm that the video timings from both are almost exactly identical.

The timings for entering/exiting HS mode aren't that interesting/useful. Tprepare+Tpreamble is potentially shorter than I'd expect on CM4, but I haven't checked the relevant section of the D-PHY spec.

Yes CM4 and Pi5 differ in whether they send NULL packets or drop to LP mode when idle. Either is permitted in the spec - they're referenced as BLLP for "Blanking or Low Power", with description "DSI Packet: Arbitrary sequence of non-restricted DSI packets or Low Power Mode including optional BTA". A NULL packet is a non-restricted DSI packet.

Welcome to the wonderful world of DSI.

Please note that I'm not back in the office until Monday.

6by9:
Hmm, quick search for the datasheet on the display to see which modes the controller states it supports.

https://ronboe.com/vdetail/392.html would appear to be the manufacturer. They state
> Controller: HX8282-A&HX8696-A

Search for those part numbers, and HX8282-A01 is an LVDS TCON - https://www.displayfuture.com/Display/datasheet/controller/HX8282-A01.pdf. Impressive feeding MIPI DSI into an LVDS interface :-(
HX8696-A is just an LCD Gate Driver, so not of interest.

ORR99:
@6by9 Thanks for you time! I didn't try it just yet, but I’m hopeful it’ll work

ORR99:
Hello @6by9 ,

First I download the tree that you modified, then I compiled the vc4.ko module, and replaced the new module with the old one, but nothing changed.

Then, since you said that the panel driver requests to drive the display in sync event mode, I modified the driver to request sync pulse mode
	dsi->mode_flags = MIPI_DSI_MODE_VIDEO | MIPI_DSI_MODE_VIDEO_SYNC_PULSE | MIPI_DSI_MODE_LPM;
but also nothing changed, the screen is still blacked out.

could it be that I am doing something wrong while loading the new vc4.ko? which I don't think this can be the case but I added some printk function in the new vc4_dsi.c but none of these logs were found in dmesg or journalctl

Also the LCD that im using is not ronbo, it is AD070IP532NG06-A, but its very similar ronbo panel, It uses EK79007 display IC, which should support:
- Non-Burst Mode with Sync Pulses
- Non-Burst Mode with Sync Event
- Brust Mode

Thanks in Advance!

PR #7016, opened 2025-08-26
overlays: Rebuild 5inch display from the 7inch

pelwell:
Reimplement the vc4-kms-dsi-ili9881-5inch display overlay by applying a few changes to the vc4-kms-dsi-ili9881-7inch version. In doing so, it inherits the rotation parameter that was previously absent.


pelwell:
FYI, @foonerd.

6by9:
LGTM

PR #7021, opened 2025-08-27
Splash Screen: bug fix

bebon901:
Prevents fullscreen logos from being drawn multiple times. With small logos, the image would be drawn multiple times across the screen.


drivers/video/fbdev/core/fb_logo.c
```diff
@@ -476,6 +476,7 @@ static int fb_show_logo_line(struct fb_info *info, int rotate,
 	if (fullscreen_logo_enabled) {
 		fb_set_logo_from_file(info, fullscreen_logo_path,
 			&image, palette);
+		n = 1;
```
pelwell: Although this works, I think it's more friendly to the reader if you place it in the next `if (fullscreen_logo_enabled)` block - perhaps at the end, just before the call to `fb_do_show_logo`.

pelwell:
Bonus points for the comment!

Issue #7019, opened 2025-08-27
Dynamic loading of UART overlays does not create /dev/ttyAMA


webmeister:
### Describe the bug

If I set `dtoverlay=uart3` in `/boot/config.txt` on a Raspberry Pi 4 and reboot, `/dev/ttyAMA3` exists. If I don't set `dtoverlay=uart3` in `/boot/config.txt` and instead run `dtoverlay uart3`, no `/dev/ttyAMA3` appears.

This is different from how other overlays work. For example, running `dtoverlay spi0-1cs` causes `/dev/spidev0.0` to appear immediately, similar to how it exists after a reboot when instead setting `dtoverlay=spi0-1cs` in `/boot/config.txt`.

A similar issue has been mentioned in https://raspberrypi.stackexchange.com/questions/143151/pi4-enable-uart4-without-using-boot-config-txt, but I was unable to find a solution.

### Steps to reproduce the behaviour

1. Ensure there are no `dtoverlay=uart*` entries in `/boot/config.txt`
2. Run `dtoverlay uart3`
3. Check whether a corresponding `/dev/ttyAMA3` exists

### Device (s)

Raspberry Pi 4 Mod. B

### System

Arch Linux ARM

```
$ vcgencmd version
May 14 2025 12:23:36
Copyright (c) 2012 Broadcom
version 17084b403fb60475b8ee2641c26049a7d54bf153 (clean) (release) (start)
```

```
$ uname -a
Linux rpi 6.12.43-1-rpi #1 SMP PREEMPT Tue Aug 26 13:22:07 MDT 2025 aarch64 GNU/Linux
```

pelwell:
Dynamic changes to Device Tree are not universally supported in Linux. They often rely on a parent node or bus registering to be notified of changes, and that doesn't happen for all devices. The GPIO, SPI and I2C subsystems do register for notifications, as does the "platform" bus type.

The Device Trees for Raspberry Pis (along with almost all other devices), declare the UART as an AMBA bus device (`"arm,pl011"`) because that's what the driver expects. (There is a platform bus variant, but with tweaks for one specific manufacturer). Well, that's what the _upstream_ driver expects - for Pi 5 support we added an AXI variant (compatible string `"arm,pl011-axi"`) in the downstream kernel tree that registers as a normal platform bus device.

Q. What happens if we change the compatible string for all the Pi PL011 UARTs to `"arm,pl011-axi"`?
[...]

PR #7023, opened 2025-08-29
ARM: dts: Select the PL011 platform driver

pelwell:
The PL011 driver in this downstream kernel tree supports an extra compatible string - arm,pl011-axi - for use by RP1. This registers as a platform driver, not an AMBA driver, and has the advantage of responding to dynamic Device Tree changes such as loading one of the "uart<n>" overlays.

Change all of the downstream Raspberry Pi dts files to use the new compatible string. At the same time, remove the override of the periphid as the upstream code now has the correct value.

See: https://github.com/raspberrypi/linux/issues/7019

### DIFF
--- a/arch/arm/boot/dts/broadcom/bcm270x.dtsi
+++ b/arch/arm/boot/dts/broadcom/bcm270x.dtsi
@@ -239,6 +239,7 @@
 };
 
 &uart0 {
+	compatible = "arm,pl011-axi";
 	/* Enable CTS bug workaround */
 	cts-event-workaround;
 };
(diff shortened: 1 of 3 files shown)


pelwell:
Any objections to merging this, @popcornmix?

nullr0ute:
So uart10 is on the bcm2712 SoC itself and not on the RP1 chip so I don't think the change to uart10 is correct @pelwell but it should just be for the RP1 uarts.

pelwell:
Although prompted by RP1, I think the new compatible string is equally applicable to 2712 and 2711. There is no AMBA bus on any of these chips, and we'd rather have the benefit of responding to dynamic DT changes that you get from using a platform driver.

nullr0ute:
You should leave the original compatibles so there's fall back, eg this breaks with an upstream kernel.

pelwell:
See https://github.com/raspberrypi/linux/pull/7278.

Issue #7018, opened 2025-08-26
7" touchscreen not working when drivers are compiled in into the kernel (Rpi5)


nooploop:
### Describe the bug

I'm trying to build a kernel with compiled in drivers for 7" official touchscreen (V1) for Rpi5. But there seems to be some i2c initialization issue and the touchscreen fails to initialize. When compiling touchscreen related drivers as modules and loading them manually after boot everything works.

### Steps to reproduce the behaviour

Build kernel without loadable modules and support for the 7" touchscreen: [config_kern.txt](https://github.com/user-attachments/files/21992984/config_kern.txt)
---> touchscreen does not work:  [dmesg_kern.txt](https://github.com/user-attachments/files/21993000/dmesg_kern.txt)

Build kernel with loadable modules: [config_mod.txt](https://github.com/user-attachments/files/21993025/config_mod.txt)
and loading the modules manually afterwards: [lsmod_mod.txt](https://github.com/user-attachments/files/21993028/lsmod_mod.txt)
---> touchscreen works: [dmesg_mod.txt](https://github.com/user-attachments/files/21993033/dmesg_mod.txt)

### Device (s)

Raspberry Pi 5

### System

kernel 6.12.42 (made with buildroot)

6by9:
Building the kernel with your config results in the system not fully booting. A diff between that and our normal defconfig shows many more changes than just switching from modules to built-in.

Trying the simplistic edit of .config to search for `=m` and replace with `=y` throws a wobbler in the build and restarts the config.

Taking our defconfig and switching the relevant modules to being built in is still giving me a working display and touch screen.
[config_working.txt](https://github.com/user-attachments/files/22005817/config_working.txt)
[kernel_log.txt](https://github.com/user-attachments/files/22005818/kernel_log.txt)
Based on lsmod I don't believe I've missed any modules out.

nooploop:
Yes, the config deviates from the defconfig in the kernel. I stripped it down by a lot, because i don't need a lot of stuff. But the only change between the two configs i supplied is enabling modules and making touchscreen related drivers modules:
```
CONFIG_TOUCHSCREEN_EDT_FT5X06=m
CONFIG_I2C_BRCMSTB=m
CONFIG_I2C_DESIGNWARE_CORE=m
CONFIG_I2C_DESIGNWARE_PLATFORM=m
CONFIG_REGULATOR_RASPBERRYPI_TOUCHSCREEN_ATTINY=m
CONFIG_DRM=m
[...]

PR #7027, opened 2025-09-01
mfd: rp1: Fix device links when modules disabled

pelwell:
Device links are used to keep track of suppliers and consumers of resources, adding some control over the ordering of device probes other than returning -EPROBE_DEFER. The way the RP1 device is created breaks this mechanism in the rare case that the use of modules has been completely disabled, thanks to some opimisations within the device link code.

Fix this glitch by giving the corresponding fwnode a pointer to the device, taking the opportunity to remove a pointless check on the validity of the rp1_node pointer.

See: https://github.com/raspberrypi/linux/issues/7018


pelwell:
Merging as low risk.

PR #7031, opened 2025-09-02
dt: Add eth_max_speed override for CM4, Pi4, and Pi400

6by9:
max-speed is a generic property for ethernet PHYs, so should be supported by the PHY on CM4/Pi4/Pi400.

Add the override and update the documentation accordingly.

### DIFF
--- a/.github/workflows/kernel-build.yml
+++ b/.github/workflows/kernel-build.yml
@@ -67,7 +67,7 @@ jobs:
         else
           sudo apt-get install gcc-arm-linux-gnueabihf;
         fi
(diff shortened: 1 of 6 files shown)


6by9:
https://forums.raspberrypi.com/viewtopic.php?p=2335473

6by9:
Added the parameter for Pi5/Pi500/CM5 as it works there too.

6by9:
Thanks for the overlay README fixup - I got the failure email just as I'd shut down last night, so it was going to be the first thing I fixed this morning.

PR #7038, opened 2025-09-05
brcmfmac: Add power_save_disable module parameter

pelwell:
The power_save_disable module parameter allows power management to be disabled from start of day, e.g. from a .conf file, without needing to run any commands from e.g. a systemd service. It also prevents any subsequent modifications unless the parameter is changed at runtime.

N.B. With power management disabled in this way, commands such as "iw wlan0 get power_save" may indicated that it is enabled, but the output in the kernel log says otherwise:

    brcmfmac: brcmf_cfg80211_set_power_mgmt: power save disabled

See: https://github.com/RPi-Distro/raspi-config/issues/282


popcornmix:
Is this preferable to NetworkManager config option `wifi.powersave = 2`? (if we are looking for a means to set this through raspi-config).

An option that disables power-save but incorrectly reports it is enabled may cause confusion.

pelwell:
I was humouring @XECDesign, but with that config option that everyone's talking about this is unnecessary.

PR #7042, opened 2025-09-10
workflows: Use native arm64 runners

nbuchwitz:
Use native arm64 runners to speed up build process. Cross compile is still used for arm targets, but also benefit from the arm64 runner architecture. Overall build time will be reduced by 25 to 30 minutes by this.

For comparison:

| arch | defconfig | time x86_x64 | time arm64 |
| ---- | --------- | ------------ | ---------- |
| arm | bcm2835 | 10:59 | 6:13 |
| arm | bcmrpi | 31:27 | 18:25 |
| arm | bcm2709 | 33:13 | 18:35 |
| arm64 | arm64 | 40:10 | 20:04 |
| arm64 | bcm2711 | 53:42 | 21:56 |
| arm64 | bcm2711_rt | 42:07 | 22:55 |
| arm64 | bcm2712 | 42:54 | 22:46 |

x86_64: https://github.com/raspberrypi/linux/actions/runs/17582418198/job/49949837669 (total 53:48)

arm64: https://github.com/RevolutionPi/linux-raspberrypi/actions/runs/17626854174 (total 23:01)

### DIFF
--- a/.github/workflows/kernel-build.yml
+++ b/.github/workflows/kernel-build.yml
@@ -16,7 +16,7 @@ env:
 
 jobs:
   build:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04-arm
     strategy:
       matrix:
         include:
@@ -56,17 +56,13 @@ jobs:
             kernel: kernel_2712
 
     steps:
-    - name: Update install
-      run:
+    - name: Install armhf crossbuild toolchain
+      if: matrix.arch == 'arm'
+      env:
+        DEBIAN_FRONTEND: noninteractive
+      run: |
         sudo apt-get update
-
-    - name: Install toolchain
-      run:
-        if [[ "${{matrix.arch}}" == "arm64" ]]; then
-          sudo apt-get install gcc-aarch64-linux-gnu;
-        else
-          sudo apt-get install gcc-arm-linux-gnueabihf;
-        fi
+        sudo apt-get install -y gcc-arm-linux-gnueabihf
       timeout-minutes: 15
 
     - uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


6by9:
Interesting.

With all the jobs running in parallel it's only the highest time that really matters, except potentially for billing.
x86_64 runners are free on public repos such as this one. I can't immediately see anything about billing for arm64 runners one way or the other. Have you found any information on that?

nbuchwitz:
As far as I understand the minutes in our usage metrics, the arm runners are accounted equally as the x86_64 runners. They do not run on the large runners, which would be billed differently based in the used minutes (even for public repos).

EDIT: I've just found this page: https://github.blog/changelog/2025-01-16-linux-arm64-hosted-runners-now-available-for-free-in-public-repositories-public-preview/

6by9:
Please note that editing Github comments doesn't notify people of the change, however that is a useful blog post to confirm that there should be no cost difference.

Minor annoyance that changes to just the workflow files are deliberately ignored.
I've thrown a patch of mine with and without your change to my rpi-6.12.y-drm2 branch, so we can see what difference we get in build times and output.

6by9:
Confirmed from my own builds that it is still classed under the 100% reduction for public repos.
arm64 minutes are cheaper too at $0.005/unit instead of $0.008 for x86_64.

I messed up my test branch as having the two overlapping builds (x86_64 and arm64) meant that I couldn't control which one rpi-update would pick up, so I've created #7044 just to confirm that the output is valid.

Otherwise it looks like a useful little saving in time for CI builds.

@pelwell @popcornmix any comments or objections?

pelwell:
If it's still free, runs quicker, and uses less energy, what's not to like?

popcornmix:
Sounds good to me

pelwell:
Thanks, Nicolai. I'll have to get used to writing "In about 20 minutes...".

PR #7044, opened 2025-09-11
Test PR to test arm64 build workflows.

6by9:
The PR doesn't get a kernel build if it is only changing the workflow files.
I want the artifact files to confirm that they do actually work.


6by9:
I have my artifacts, and they work, so all good.
Closing.

PR #7049, opened 2025-09-15
dtoverlays: adjust inbound windows for MIP1 on Pi 5 with 32-bit PCIe DMA

P33M:
see #7046 

Note that trying this with an NVMe drive as a test will result in swiotlb exhaustion - compiling with CONFIG_SWIOTLB_DYNAMIC fixed this but the vc4 driver won't link - which is a separate issue.


arch/arm/boot/dts/overlays/pcie-32bit-dma-pi5-overlay.dts
```diff
@@ -19,8 +19,21 @@
...
+				     <0x02000000 0x00 0xfffff000 0x10 0x00131000
+				      0x00 0x00001000>;
 		};
 	};
 
+	fragment@1 {
+		target = <&mip1>;
+		__overlay__ {
+			/*
+			 * The MIP uses the ranges property to derive the target
```
pelwell: `reg property`?

PR #7051, opened 2025-09-17
DRM command line parser fix

6by9:
https://forums.raspberrypi.com/viewtopic.php?t=391784

`video=720x480,reflect_x,rotate=270` wouldn't parse the `reflect_x` correctly as it found the `=` from `rotate=270` and assumed that was all one option.

### DIFF
--- a/drivers/gpu/drm/drm_modes.c
+++ b/drivers/gpu/drm/drm_modes.c
@@ -2173,12 +2173,15 @@ static int drm_mode_parse_cmdline_options(const char *str,
 
 	option = str;
 	do {
+		sep = strchr(option, ',');
 		delim = strchr(option, '=');
 		if (!delim) {
-			delim = strchr(option, ',');
-
-			if (!delim)
+			if (!sep)
 				delim = option + strlen(option);
+			else
+				delim = sep;
+		} else if (sep && sep < delim) {
+			delim = sep;
 		}
 
 		if (!strncmp(option, "rotate", delim - option)) {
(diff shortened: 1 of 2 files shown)


popcornmix (approved):

popcornmix:
I assume this is suitable for upstreaming?

6by9:
> I assume this is suitable for upstreaming?

Yes, it'll be added to the small pile I'll be sending upstream in due course.

In running the unit tests I did notice that we have a few failures due to max_bpc selection.
Upstream defaults a connector to the max bit depth possible. That means that booting to the console at 4k60 you'd get YUV422 at 12bpc rather than RGB888 at 8bpp.

I'd written https://github.com/raspberrypi/linux/commit/7ff3c4a9cf344c08cdac97ee82d986c5ee007605 which drops it to 8bpc, and that's why we have the failures. That'll be a fun one to argue, but I don't believe it affects anyone else as all other users of the helpers max out at 8bpc.

Likewise the mainline kernel doesn't try YUV422, so https://github.com/raspberrypi/linux/commit/56fe89b9e66b52499b8c051f0fee8b7eb426311a causes a failure.

Ho hum.

Then again https://lore.kernel.org/dri-devel/20250911130739.4936-1-marius.vlad@collabora.com/ adding a userspace "output format" control to allow forcing RGB/YCbCr444/YCbCr422/YCbCr420  appears to be gaining traction now which may help, or may just confuse matters further.

Issue #7048, opened 2025-09-15
Feature Request: Add support for VEML6030 and VEML7700 sensors


Jibun-no-Kage:
Feature Request: Add support for VEML6030 and VEML7700 sensors.  Thanks.

6by9:
(This is not a firmware issue - dtoverlays are part of the Linux kernel, hence https://github.com/raspberrypi/linux).

The module VEML6030 for 6030, 6035, and 7700 isn't built in the standard Pi kernel configs, so it would need a defconfig change as well as the overlay.

Before we jump through the hoops to add this one, are there any others that you're after?

Jibun-no-Kage:
Ah yes!  Thanks for transferring it.

The 6040, 6070, 6075 are already supported.  Which is wonderful!  However, these requested additions extend the support of the family of sensors... in that some are UV sensors, some spectral (color), and some Lux only.  The pricing and sourcing changes all the time, as batch runs are completed, then have to wait for the batch runs to complete again, so for one project 6030 might available, later only the 7700 is available... then later 6030, etc., etc.

Just for comparison... same is happening with the BMP280, BME280 being replaced, via the BMP580, BMP581, and BMP585 are available now.  There is prototype support for BMP580 support via dtoverlay, per 6.12.45, which I tested and it did not seem to work, i2cdetect did not show UU state. I reported this on another issue.

pelwell:
I'm just preparing a PR for this, but I've found that the 6035 and 7700 support doesn't appear until 6.16. Back-porting may be possible, but it's non-trivial. Would you settle for VEML6030 in 6.12, with the others appearing in 6.16 and later?

pelwell:
See: https://github.com/raspberrypi/linux/pull/7052

Jibun-no-Kage:
Yes!  No need for jump through hoops!  VEML6030 in 6.12... and other is in 6.16 is fine.  Again... Thank you.

pelwell:
You can install a trial build of the new kernel and updated overlay using `sudo rpi-update pulls/7052` (or build it yourself) - just be careful if you are relying on the initramfs for a non-standard root filesystem, etc.

Jibun-no-Kage:
True, true.  I have been looking for an opportunity to learn the kernel build process, maybe this is just that.  I trust that you will keep the updated support in queue or back burner for 6.16.  So I will close the request for now.

Jibun-no-Kage:
[...]

PR #7052, opened 2025-09-17
configs: Add CONFIG_VEML6030=m

pelwell:
Enable support for the VEML6030 light sensor family.

See: https://github.com/raspberrypi/linux/issues/7048


6by9:
Was there meant to be an overlay to go alongside the defconfig addition?

pelwell:
You say _overlay_ - does it need more than an entry in `i2c-sensor-common.dtsi` (which it doesn't yet have)?

pelwell:
Now with a veml6030-capable i2c-sensor overlay. I'm curious about the best default address, and whether configuring the interrupt line is preferred.

6by9:
> You say _overlay_ - does it need more than an entry in `i2c-sensor-common.dtsi` (which it doesn't yet have)?

No, that was all I meant.
The PR only updating the defconfigs appeared to be missing a bit.

> Now with a veml6030-capable i2c-sensor overlay. I'm curious about the best default address, and whether configuring the interrupt line is preferred.

Seeing as the driver can work with or without interrupt, I'd go with leaving it disabled by default. Those who actually wire it up can enable it with an override. We have that with mcp23017 and probably a few others.

The options for address appear to be 0x48 (ADDR = high) or 0x10 (ADDR low). Your choice of 0x48 seems perfectly reasonable to me.

pelwell:
> We have that with mcp23017 and probably a few others.

Unfortunately that's not how the i2c-sensor overlay is structured. Devices that support interrupts have it enabled permanently - the only option is to change the address. We have never said that `int_pin` has a default value, so we could demand the use of `int_pin` to enable them. Sadly it's not as simple as creating a dormant node that adds the necessary properties, but that would mean that interrupt-supporting devices need a second fragment (the targets are device-specific, and the interrupt type varies).

PR #7056, opened 2025-09-18
configs: Add CONFIG_W1_SLAVE_DS2430=m

nbuchwitz:
The Maxim DS2430 is a popular 256-bit 1-Wire EEPROM that can be accessed in a similar way to the already supported DS2431 1-Wire EEPROM. It is, for example, used by a German federal research agency in a custom hardware board that is used with a CM4S-based Revolution Pi controller.


markusrudolf:
It will be useful for our application if the module is build and included in the default config. Space should also not be a real problem

pelwell:
In about 25 minutes you should be able to install a trial build of the kernel by running `sudo rpi-update pulls/7056` . Back up any important data first, and be extra careful if you are making use of the initramfs - test with `auto_initramfs=0` in config.txt first if unsure.

markusrudolf:
Just tried it on a Pi4 with 2GB RAM, works like a charm:

```
pi@largecalibscreen 14-000007ab3213 $ cat eeprom | xxd -u -g1
00000000: 01 00 8B 00 00 D9 00 23 00 0A 00 C1 00 23 11 C6  .......#.....#..
00000010: 32 00 01 00 01 00 01 00 00 CB 00 00 00 00 00 8D  2...............
pi@largecalibscreen 14-000007ab3213 $ 
```

pelwell:
Thanks - feedback is always helpful.

PR #7057, opened 2025-09-18
4lane support for IMX708

6by9:
This gets IMX708 working in 4 lane mode.

It hasn't yet changed the frame rates achievable, but has made it easier in computing the PLL register settings based on the configured values.
Whilst we could try to do something clever in automatically computing values, I'd probably lean towards having an array for `line_length_pix` and `pixel_rate` for each mode, with one set of values for 2 lanes and another for 4 lanes. At least that way we can guarantee that the value combinations work, but does limit some of the gain in increasing the link frequency.

I haven't added any validation on the link frequency actually making a usable system. Adding a check later that the line time of the pixel array is greater than that of the MIPI link would probably be beneficial.


6by9:
Bah, I hate the handling of 64bit division on 32bit kernels. Hopefully resolved now.

drivers/media/i2c/imx708.c
```diff
@@ -240,434 +238,434 @@ static const struct imx708_reg_list link_freq_regs[] = {
...
+	{CCI_REG8(0x3092), 0x00},
+	{CCI_REG8(0x3419), 0x00},
+	{CCI_REG8(0xBCF1), 0x02},
+	{CCI_REG8(0x3094), 0x01},
+	{CCI_REG8(0x3095), 0x01},
+	{CCI_REG8(0x3362), 0x00},
+	{CCI_REG8(0x3363), 0x00},
+	{CCI_REG8(0x3364), 0x00},
+	{CCI_REG8(0x3365), 0x00},
+	{CCI_REG8(0x0138), 0x01}, //REG_TEMP_SENS_CTL
```
naushir: C style comment?

naushir (approved):

drivers/media/i2c/imx708.c
6by9: Removed.

PR #7060, opened 2025-09-22
net: phy: broadcom: Add led-swap property support

pelwell:
Extend the LED control features of the led-modes property by adding a led-swap property. This allows the same led-modes values to be used across designs where the LED assignments differ.


timg236:
LGTM

PR #7063, opened 2025-09-24
media: i2c: imx219: fix binning and rate_factor for 480p and 1232p

naushir:
At a high FPS with RAW10, there is frame corruption for 480p because the rate_factor of 2 is used with the normal 2x2 bining [1]. This commit ties the rate_factor to the selected binning mode. For the 480p mode, analog 2x2 binning mode with a rate_factor of 2 is always used. For the 1232p mode the normal 2x2 binning mode is used for RAW10 while analog 2x2 binning mode is used for RAW8.

[1] raspberrypi#5493

Reworked due to upstream changes

### DIFF
--- a/drivers/media/i2c/imx219.c
+++ b/drivers/media/i2c/imx219.c
@@ -148,6 +148,12 @@
 #define IMX219_PIXEL_ARRAY_WIDTH	3280U
 #define IMX219_PIXEL_ARRAY_HEIGHT	2464U
 
+enum binning_bit_depths {
+	BINNING_IDX_8_BIT,
+	BINNING_IDX_10_BIT,
+	BINNING_IDX_MAX
+};
+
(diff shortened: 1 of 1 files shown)


drivers/media/i2c/imx219.c
```diff
@@ -322,24 +325,40 @@ static const struct imx219_mode supported_modes[] = {
 		.width = 3280,
 		.height = 2464,
 		.vts_def = 3526,
+		.binning = {
+			[BINNING_IDX_8_BIT] = 0,
```
pelwell: Would this (and other zeroes) be better as `IMX219_BINNING_NONE` (which I assume is 0)?

naushir: Yes!

pelwell (approved):

PR #7065, opened 2025-09-25
drm/rp1: rp1-dsi: Add support for inverting lane polarities

6by9:
The D-PHY on RP1 support lane polarity swapping, and there is a standard device tree mechanism for configuring this, so tie the two together.

https://forums.raspberrypi.com/viewtopic.php?t=388879


drivers/gpu/drm/rp1/rp1-dsi/rp1_dsi_dsi.c
```diff
@@ -145,11 +145,18 @@
 #define DSI_PHYRSTZ_FORCEPLL_LSB 3
 #define DSI_PHYRSTZ_FORCEPLL_BITS  BIT(DSI_PHYRSTZ_FORCEPLL_LSB)
 
+#define DPHY_CLK_PN_SWAP		0x35
```
pelwell, line 146: I could understand putting this register out in front if the other registers were in numerical order, but they aren't.

6by9: Registers now in order, and lots of whitespace issues and other cruft sorted.

PR #7066, opened 2025-09-26
[TESTING] kbingham/rpi 6.16.y/imx283 fixes

kbingham:
Some redesign of the imx283 mode configuration that allows us to add modes and configurations more easily. Ultimately - I hope to progress this sometime to support free arbitrary cropping on the sensor side with the driver handling all the alignment and restrictions correctly - but that will take some more time still yet.

Crucially in this series is some changes to the handling of the user clamp area which contains the VOB but isn't actually part of the sensor vertical position coordinates. Previously - this means that a 16 line offset has been occuring in the output image.

Another artifact is that the sensor introduces an extra line at the top of the image which we believe is to support or prevent bayer reordering on vflip. But the additional line produces an undesirable artifact in the output. This series uses the ability to adjust the clamp area/VOB to extend *over* the additional line (and one more to maintain bayer order) to remove the artifact. To prevent this causing the image to be offset - an extra two lines are requested on top of the mode first.

Finally - three clear new modes are added
  - native array - outputs the full 5592x3710 pixel count - which is interesting as you see the bayer reorder occur - and the various HOB/VOB areas.
  - a 'full active' pixels mode produces all output from the sensor /excluding/ the HOB/VOB - but this *includes* the additional line artifact.
  - The 'effective array' is the datasheet pixel area that includes a 12 pixel margin on all sides to allow for color processing to occur in the ISP scalers without issues at the edges.

I'd still like to get the binning and timing values out of the crop mode structure - but that will take some more work - and we don't yet have an api to control binning / skipping which makes it difficult to convey in the scanout mode structures.

### DIFF
--- a/drivers/media/i2c/imx283.c
+++ b/drivers/media/i2c/imx283.c
@@ -509,6 +509,21 @@ static const struct imx283_reg_list link_freq_reglist[] = {
 /* Mode configs */
 static const struct imx283_mode supported_modes_12bit[] = {
 	{
+		/* Full Native pixel array, including HOB/VOB. 5592x3710 */
+		.scan = &imx283_scan_modes[IMX283_MODE_0],
+
+		.width = 5592,
+		.height = 3710, /* 3694 + 16 additional lines for VOB */
+		.min_hmax = 5914, /* 887 @ 480MHz/72MHz */
+		.min_vmax = 3793, /* Lines */
+
+		/* 20.00 FPS */
+		.default_hmax = 6000, /* 900 @ 480MHz/72MHz */
+		.default_vmax = 4000,
+
+		.crop = imx283_native_area,
+	},
+	{
 		/* 20MPix 21.40 fps readout mode 0 */
 		.scan = &imx283_scan_modes[IMX283_MODE_0],
 
--- a/drivers/media/i2c/imx283.c
+++ b/drivers/media/i2c/imx283.c
@@ -1144,14 +1144,30 @@ static int imx283_start_streaming(struct imx283 *imx283,
 	/* Vertical Configuration */
 	{
 		u32 y_out_size = mode->crop.height / mode->scan->vbin_ratio;
-		u32 write_v_size = y_out_size + mode->scan->vertical_ob;
 
-		s16 top = mode->crop.top;
+		/*
+		 * The CLAMP region contains the Vertical Optical Black (VOB)
+		 * lines, which are not included in the effective image height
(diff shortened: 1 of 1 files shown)


kbingham:
I haven't posted these upstream yet - sharing here to get them tested by known IMX283 users on RPi ... but no specific desire to merge this into the RPi tree until reviewed upstream. But they could be if testers find the series beneficial.

will127534:
SRGGB12_CSI2P,5592x3710 mode doesn't seems right, but all other modes are working fine:
5592x3710:
IMX283_5592_3710


5496x3672:
IMX283_5496_3672

will127534:
Ah it looks like SRGGB10_CSI2P,5472x3648/0 this mode isn't working also, there is no error message but the frame is empty.

will127534:
Here are the jpeg images for all the modes 
=========12-bit Mode ============
5592x3710_12:
5592x3710_12
5496x3694_12:
5496x3694_12
5496x3672_12:
5496x3672_12
5472x3648_12:
5472x3648_12
2736x1824_12:
2736x1824_12
1824x1216_12:
1824x1216_12

=========10-bit Mode ============
5472x3648_10:
5472x3648_10

kbingham:
Thank you - I hadn't noticed 10 bit was broken - but thanks to your test I can see the bug. I'll fix that.

kbingham:
> 5592x3710_12:
> 5592x3710_12

I think this is almost expected behaviour - I was expecting 'pink' because the bayer order ends up being mis-interpreted/mis-aligned - though I don't know why it goes quite so psychedelic through the ISP ...
But it's also a bit too specialised a mode to actually advertise anyway I think - so I'll just drop this one.

pelwell:
What's the state of this PR? rpi-6.16.y is EOL, and rpi-6.17.y can't be far behind. rpi-6.18.y must still be quite close to upstream, and we'd want this on rpi-6.18.y before we'd consider merging, so perhaps it's time to rebase your branch.

kbingham:
absolutely - no desire for this to be merged for RPI it's testing only - I've only pushed this so that I could get some RPI testers the most recent updates on a working Pi kernel, where they reported some issues.

I'll be posting these upstream to mainline, and where appropriate I'll share backports - but I wanted to be sure everything's working, and work through some issues that were reported in the community.

kbingham:
preparing to post upstream - so closing this pr! So long and thanks for all the fish! (and the tests)

Issue #6068, opened 2024-03-25
TC358743 produces BGR instead of RGB


mdevaev:
### Describe the bug

I'm trying to use RGB24 format with TC358743 instead of UYVY, and noticed that it produces BGR sequence in the raw data instead of RGB. I used the blue fullscreen image as an HDMI source and made a stream dump that confirms this, getting something like `[255, 0, 0]` for first 3 bytes.

If you have a top secret datasheet under Toshiba NDA, you can check pages 36/261 and 37/261. This confirms that UYVY is implemented correctly, but RGB has the reverse sequence. Apparently, this bug went unnoticed because everyone was using UYVY. I need RGB to output the image in DRM and encode in M2M at the same time.

### Steps to reproduce the behaviour

Use any capture software with TC358743 in RGB24 format and dump the raw image.

### Device (s)

Raspberry Pi CM4, Raspberry Pi CM4 Lite

### System

* Arch Linux ARM, but it doesn't matter
* `version c14213886ae1e31951022b74ce776ce7ee7023ee (clean) (release) (start)`
* `Linux pikvm 6.6.21-4-rpi #1 SMP Wed Mar 13 20:36:42 UTC 2024 armv7l GNU/Linux`
* Apparently, the problem is relevant for any version of kernel.

### Additional context

I found this bug when encoding JPEG in M2M and using libjpeg. It's the same with M2M H.264. Interestingly, because of the endianness, DRM_FORMAT_RGB888 is actually BGR, so the rendering of the TC* chip output in DRM was correct colors, while the encoding was not.

I don't know if it is possible to fix this with little blood without loss of performance, given that it looks like you need to use the reverse order to put bytes in memory after receiving from CSI. This is not a big problem and I can make a userspace workaround for this, but it would probably be better to fix it upstream if the fix is not complex or fatal to performance.

6by9:
Life with component ordering is always complicated.

tc358743 is an upstream driver with relatively minimal changes. It was written by Cisco for some of their video conferencing kit. IIRC it was against an ST SoC, but it may have been TI.

They've defined that the media bus code is MEDIA_BUS_FMT_RGB888_1X24

Checking the [definition for MEDIA_BUS_FMT_RGB888_1X24](https://www.kernel.org/doc/html/latest/userspace-api/media/v4l/subdev-formats.html#packed-rgb-formats), it is putting B in the LSBs of the bus. CSI-2 does send the data LSB and blue first, so would make sense as a mapping, 

[...]

PR #7070, opened 2025-09-29
Upstream patches for TC358743 colour swap

6by9:
Upstream patches proposed to solve #6068 

Draft PR for testing.


6by9:
Tested and does the right thing.

I'm happy for this to be merged now, so draft status dropped.

PR #7075, opened 2025-10-03
drivers: pcie: Hailo: Remove Hailo PCIe driver

naushir:
This driver will be installed through DKMS going forward.


6by9:
IIRC out-of-tree modules will set the kernel taint flag, so using DKMS could cause some grief. You'll also have the fun and games of needing to support multiple kernel versions in the driver. However it does make maintenance easier in some ways.

pelwell:
Do you still want this gone?

naushir:
Yes, I'll merge this now.

pelwell:
I just noticed that the arm64 defconfigs were not touched in this PR. Was that a simple oversight?

naushir:
Oops, yes, this was indeed missed out.  I'll raise a PR.

PR #7082, opened 2025-10-08
Suggest `raspinfo | pastebinit` in bug_report.yml

lurch:
Output of `raspinfo` is now over 65,000 characters, which is more than GitHub allows in a single form field! Also adds Pi 500+ to list of models.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -72,8 +72,8 @@ body:
   attributes:
     label: System
     description: |
-      Copy and paste the results of `raspinfo | pastebinit` in to this section.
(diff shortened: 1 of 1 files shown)


timg236 (approved):

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -71,7 +72,7 @@ body:
   attributes:
     label: System
     description: |
-      Copy and paste the results of the raspinfo command in to this section.
+      Copy and paste the results of `raspinfo | pastebinit` in to this section.
```
popcornmix: Maybe make it clearer the wanted text is now a URL. e.g.

Copy and paste the URL returned from `raspinfo | pastebinit`

PR #7088, opened 2025-10-13
fbcon: Add defensive coding to logo loader

6by9:
There were various points where the loader was using uninitialised data, had the potential to run off the end of an array, or was handling core functions incorrectly. Fix these up.

(It still goes quite badly wrong if the framebuffer isn't 16bpp)


6by9:
Now fixed up to work with 24 and 32bpp framebuffer devices.

Amusingly Pi5 with `framebuffer_depth=24` appears to have set the hardware to BGR888, but always sends `r8g8b8` to simple-framebuffer. The code appears to look at `framebuffer_swap`, but somehow it isn't applied. (I ought to look at the dlists).

`framebuffer_depth=16` ignores `framebuffer_swap` totally and always sets r5g6b5. 

Both are semi-understandable as simple-framebuffer didn't have the BGR variants defined. I've added them, so we probably want to update the firmware to make use of this.

pelwell (approved):

popcornmix (approved):

Issue #7078, opened 2025-10-07
Paced DMA on RP1 not possible


goldwaving:
### Describe the bug

Setting up paced DMA with PWM flow control (or any other flow control peripheral) fails to work on RP1.  In `dw-axi-dmac-platform.c` the `snps,sel-require` entry is read from the device tree. However that entry does not exist.  So in  the `dw_axi_dma_of_xlate` function, the `dw->sel_required` values are always zero, resulting in a zero `chan_mask`.  No channel can be selected.

Also the DMA handshake table in the `RP1 Peripherals` document is missing entirely, even though it is mentioned in the text.

Would adding a `snps,sel-require` entry in the device tree solve this problem?  If so, what would be the appropriate values for DREQ setup?

### Steps to reproduce the behaviour

Attempt to set up paced DMA using PWM on RP1

### Device (s)

Raspberry Pi 5

### System

Raspberry Pi reference 2025-05-13
Generated using pi-gen, https://github.com/RPi-Distro/pi-gen, 5dabc7dc940059dfbc46af5d97b60a1e812523dd, stage2

2025/05/08 15:13:17
Copyright (c) 2012 Broadcom
version 69471177 (release) (embedded)

Linux precisionwall 6.12.47+rpt-rpi-2712 #1 SMP PREEMPT Debian 1:6.12.47-1+rpt1~bookworm (2025-09-16) aarch64 GNU/Linux

pelwell:
> However that entry does not exist.

Yes it does:
```
~$ od -Ax -tx4 --endian=big /proc/device-tree/axi/pcie@1000120000/rp1/dma@188000/snps,sel-require
000000 00000100 00000100 00000000 00000000
000010 00000000 00000000 00000000 00000000
000020
```
Unless, of course, you are getting your Device Tree from somewhere other than the one installed with the kernel, then loaded and passed on by the firmware.

goldwaving:
So with those values, a kernel module with a device tree overlay having `dmas = <&rp1_dma 0x118>` should connect to the PWA flow control (assuming line 24), but it doesn't.  Actually the DMA transfer doesn't run at all and gives no errors.  Even with a `dmas = <&rp1_dma 0>` the transfer does not run.  With `dmas = <&rp1_dma 0xFFFFFF>`, no errors occur and it does not run.  Only a value 0x2XX give an error (precluded).  Using `dmas = <&dma40 0>` or `<&dma32 0>` runs, but unpaced.

[...]

PR #7089, opened 2025-10-14
Improve the RP1 DMA channel allocation scheme

pelwell:
As well as being bugged (#7078), the previous RP1 DMA channel allocation scheme would not make optimal use of the channels in the event that many were required. The new scheme gives each channel a score as to how well it matches the client's requirements, but when availability is low even low-scoring channels are eventually allocated.

Despite the improved functionality, the DT declaration is simpler. The driver should also behave the same with old-style DTBs.

See: https://github.com/raspberrypi/linux/issues/7078


pelwell:
Users of RP1 DMA are few and far between, but I've tested this to my satisfaction, so merging.

PR #7095, opened 2025-10-20
rpi-6.18.y: ilitek ili9881 driver cleanup

6by9:
Deals with the changes in upstream to handle some of the issues that our downstream driver was dealing with.

Please drop / squash patches appropriately.


popcornmix:
Rebased with reverted commits dropped.

PR #6881, opened 2025-06-04
add rtw88/rtw89 drivers

macmpi:
as per https://github.com/raspberrypi/linux/issues/6874

PR #7096, opened 2025-10-21
Add RTW89 USB modules to Pi defconfigs

macmpi:
8851BU, 8852BU
following https://github.com/raspberrypi/linux/pull/6881
Also applicable to 6.18 and on


pelwell:
Have you tested this? Even with this PR, 8851BU is not being enabled - there's probably a missing dependency. And, as with #6881, these are not proper defconfig files.

pelwell:
> there's probably a missing dependency

No, it's just a typo - you want `CONFIG_RTW89_8851BU`, not `CONFIG_RTW88_8851BU`.

macmpi:
> No, it's just a typo - you want `CONFIG_RTW89_8851BU`, not `CONFIG_RTW88_8851BU`.

Fixed, sorry.

pelwell:
Those now give the correct results, but the arm64 versions have the additions in the wrong places - 8852BU should be immediately after 8852BE. You should be running `make ... savedefconfig` to get the canonical result.

pelwell:
That's looking better. In about 25 minutes you should have a trial kernel to test: `sudo rpi-update pulls/7096`.

macmpi:
Thanks

Issue #7108, opened 2025-10-25
Cannot disable gamma corection for LEDs (RPi5 ws2812-pio-rp1 module)


awawa-dev:
### Describe the bug

Hi
Kernel module for Raspberry Pi 5 PIO ws2812-pio-rp1 module that allows control over ws2812 sk6812 LED strips:
https://github.com/raspberrypi/linux/blob/2e376fd6b1f2fe1ab785fa607f36db6669d0bb8b/drivers/misc/ws2812-pio-rp1.c#L154

Gamma correction is hard-coded. We cannot disable it with any settings. So instead of rendering 256 colors per channel we are able to render max 100-120 due to the gamma compression.

### Steps to reproduce the behaviour

https://github.com/awawa-dev/HyperHDR/pull/1317

### Device (s)

Raspberry Pi 5

### System

System Information
------------------

Raspberry Pi 5 Model B Rev 1.0
PRETTY_NAME="Debian GNU/Linux 13 (trixie)"
NAME="Debian GNU/Linux"
VERSION_ID="13"
VERSION="13 (trixie)"

Linux rpi5 6.12.47+rpt-rpi-2712 #1 SMP PREEMPT Debian 1:6.12.47-1+rpt1 (2025-09-16) aarch64 GNU/Linux

pelwell:
That's a fair point. It would be easy to use a magic brightness value - 0, say - to disable the gamma curve. Would that work for you?

pelwell:
https://github.com/raspberrypi/linux/pull/7110 is an untested implementation of my idea above. To bypass the gamma curve, add the parameter `brightness=0` to the dtoverlay line, e.g.:
```
dtoverlay=ws2812-pio,gpio=18,num_leds=30,rgbw,brightness=0
```

By the time you read this you should be able to install a trial kernel using `sudo rpi-update pulls/7110`. Be careful to back up any important data first, and be extra cautious if you rely on `auto_initramfs=1` in order to boot (try booting with it set to 0 before installing the new kernel).

awawa-dev:
Hi Phil,
Thank you so much for the quick response. It works great now, I'm using a magic byte to set the brightness to 0 to disable gamma. Updated measurements made using a logic analyzer: https://github.com/awawa-dev/HyperHDR/pull/1317

pelwell:
This issue has now been resolved.

PR #7110, opened 2025-10-26
misc: ws2812-pio-rp1: Allow gamma to be disabled

pelwell:
Modify the gamma logic such that setting the brightness in DT to zero causes the supplied pixel values to be passed through unmodified.

See: https://github.com/raspberrypi/linux/issues/7108


pelwell:
Updated with a change to the README explaining the new behaviour.

popcornmix (approved):

PR #7123, opened 2025-11-03
drivers: i2c: imx477: Use pm_runtime_use_autosuspend()

naushir:
Switch the power management in the imx477 device driver to use auto- suspend with a 5s timeout.

This improves mode switching time that avoids additional regulator switch-on delays and common register I2C writes.


6by9 (approved):
Based only on comparing to the imx290 equivalent, it looks good.

PR #7124, opened 2025-11-03
drm/vc4: plane: Swap Cb/Cr offsets for YVU formats

6by9:
Follow up to "drm/vc4: plane: Swap Cb/Cr pointers for YVU formats" Swap the offsets as well as the buffer objects for YVU formats.

Draft as I want to double check the async update path too.


6by9:
Rightly or wrongly, https://github.com/raspberrypi/linux/blob/rpi-6.12.y/drivers/gpu/drm/vc4/vc4_plane.c#L2571 only lets ptr0_offset[0] change with an async update. ptr0_offset[1] and ptr0_offset[2] can't change.

I think that means that we can't (currently) update the chroma planes of a YUV/YVU frame with an async update, so there's no issue. When someone complains about async YUV updates I'll take another look.

Dropping the draft status.

PR #7127, opened 2025-11-06
overlays: Add vec-gpio-pi5-overlay

njhollinghurst:
Add an overlay to allow composite video to be output on GPIOs 4-11 on Raspberry Pi 5, 500, 500+ or CM5 only, with an optional 108 MHz clock on GPIO 0 and duplicate MSB on GPIO 27.

Requires composite video to be enabled and DPI to be disabled.

### DIFF
--- a/arch/arm/boot/dts/overlays/README
+++ b/arch/arm/boot/dts/overlays/README
@@ -5832,12 +5832,21 @@ Info:   See vc4-kms-v3d-pi4 (this is the Pi 5 version)
 
 
 Name:   vc4-kms-vga666
-Info:   Enable the VGA666 (resistor ladder ADC) for the vc4-kms-v3d driver.
+Info:   Enable the VGA666 (resistor ladder DAC) for the vc4-kms-v3d driver.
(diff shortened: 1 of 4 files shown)


arch/arm/boot/dts/overlays/README
```diff
@@ -5832,14 +5832,23 @@ Info:   See vc4-kms-v3d-pi4 (this is the Pi 5 version)
...
 Params: ddc                     Enables GPIOs 0&1 as the I2C to read the EDID
                                 from the display. NB These are NOT 5V tolerant
                                 GPIOs, therefore level shifters are required.
 
 
+Name:   vec-gpio-pi5
+Info:   Overlay to output composite video as an 8-bit digital code over GPIO
+        pins 4-11, for Raspberry Pi 5 series only. Requires composite video
+        to be enabled; incompatible with any use of DPI or VGA666.
+Load:   dtoverlay=vec-gpio-pi5
```
pelwell: The checker requires `,<param>`

njhollinghurst: Fixed

PR #7128, opened 2025-11-07
drivers: i2c: imx708: Use pm_runtime_use_autosuspend()

naushir:
Switch the power management in the imx708 device driver to use auto- suspend with a 5s timeout.

This improves mode switching time that avoids additional regulator switch-on delays and common register I2C writes.

### DIFF
--- a/drivers/media/i2c/imx708.c
+++ b/drivers/media/i2c/imx708.c
@@ -1236,7 +1236,8 @@ static int imx708_set_ctrl(struct v4l2_ctrl *ctrl)
 		break;
 	}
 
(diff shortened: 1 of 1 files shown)


6by9 (approved):
Matches the change for imx477 from last week.

Be aware on this sensor that because the lens driver will have a reference on the same regulator, calling `imx708_power_off` does NOT mean that the sensor has been reset. We also don't have a software reset obviously in the register set (I can't see one documented in the datasheet).
As long as we do have the same register writes in all mode tables, then it should be fine.

